PR #68, opened 2015-01-01
Fix 'ls' command

yath:
CmdLs needs to implement flags.Commander, so change (CmdLs) Execute to
only take a []string to get actually called.


fd0:
lgtm

PR #81, opened 2015-01-31
Order of parameters fixed.

kamudadreieck:
restic -r /tmp/backup_loaction <command>


fd0:
Merged, thanks for the report!

PR #114, opened 2015-03-21
compile on windows

thesyncim:
not really useful but i would like to help on supporting windows.
can you guide me?
will start with restic.Node

### DIFF
--- a/cache_windows.go
+++ b/cache_windows.go
@@ -1,0 +1,45 @@
+package restic
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/restic/restic/debug"
+)
+
+// GetCacheDir returns the cache directory
+func GetCacheDir() (string, error) {
+	AppData := os.Getenv("AppData")
+
+	if AppData == "" {
+		return "", errors.New("unable to locate cache directory")
+	}
+
+	cachedir := ""
+	if AppData != "" {
+		cachedir = filepath.Join(AppData, "restic")
+	}
+
+	fi, err := os.Stat(cachedir)
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(cachedir, 0700)
+		if err != nil {
+			return "", err
+		}
+
+		fi, err = os.Stat(cachedir)
+		debug.Log("getCacheDir", "create cache dir %v", cachedir)
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	if !fi.IsDir() {
+		return "", fmt.Errorf("cache dir %v is not a directory", cachedir)
+	}
+
+	return cachedir, nil
+}
--- a/node_windows.go
+++ b/node_windows.go
@@ -17,3 +17,7 @@ func (node *Node) createCharDevAt(path string) error {
 func (node *Node) createFifoAt(path string) error {
 	return nil
 }
+
+func (node *Node) isNewer(path string, fi os.FileInfo) bool {
+	return false
+}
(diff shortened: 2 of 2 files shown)


fd0:
Thanks for trying this. Please have a look at `isNewer()` and `fill_extra()` in `node_linux.go`, this needs to be implemented in order to fully support windows. Are you aware of any extra attributes files and directories have? I'm thinking about things like ACL, owner, group, etc.

thesyncim:
thanks for the feedback 

node.ChangeTime and node.AccessTime 
```
stat, ok := fi.Sys().(*syscall.Win32FileAttributeData)
	if !ok {
		return
	}

node.ChangeTime = time.Unix(0, stat.LastWriteTime.Nanoseconds())
node.AccessTime = time.Unix(0, stat.LastAccessTime.Nanoseconds())
```

i´m not a windows expert but i think we can use GetSecurityInfo to get owner and group. 
https://msdn.microsoft.com/en-us/library/windows/desktop/aa446654(v=vs.85).aspx 

I´m a bit worried about you do we manage to copy open files maybe we can use VSS

fd0:
VSS would be the right thing to do. I'm not a windows expert either, so someone has to find out how VSS works ;)

thesyncim:
we can use vshadow.exe
`VShadow is a command-line tool that you can use to create and manage volume shadow copies.`
https://msdn.microsoft.com/en-us/library/windows/desktop/bb530725(v=vs.85).aspx





 but we need different binaries for different windows arch/versions 
also vshadow.exe needs administrative privileges 
binaries for major windows arch/versions (https://github.com/thesyncim/insync/tree/master/util/vshadow)

example:
vshadow.exe  -exec=(program to execute after snapshot created) -script=(File To save Snapshot information)  C:

Here is a example wrapper service for vss to choose right vshadow.exe based on windows arch/version
https://github.com/thesyncim/insync/blob/master/util/vss/vss.go (needs admin privileges)

and here is the example script to execute once snapshot is ready (https://github.com/thesyncim/insync/blob/master/util/copy/copy.go)

thesyncim:
we may need some API changes so we can support windows.

node.fill_extra(string,os.FileInfo) must be node.fill_extra(string, *os.File) in order to get access to *os.File.Fd() (file Handle).

any thoughts on this?

fd0:
Oh, why is that? This needs some more changes, as the file isn't opened yet at the time the node is created. Can't we just open it in `fill_extra()`?

btw: The latest commit is not properly formatted with `gofmt`, maybe you can install the pre-commit hook for git as suggested in the `README.md`.

thesyncim:
sry about that will fix in the next commit

currently the file in open in archive.SaveFile (https://github.com/restic/restic/blob/1c0e76ccd6db5925e2a1c6db10abe6e3366842e2/archiver.go#L167) and closed only after the call to callNodeFromFileInfo.
we can close it before enter NodeFromFileInfo but we will have an extra syscall.

what do you suggest ?
thanks

fd0:
That's the easy part. What about the call in `fileWorker()`? 
https://github.com/restic/restic/blob/1c0e76ccd6db5925e2a1c6db10abe6e3366842e2/archiver.go#L405

fd0:
Maybe we should postpone windows support until afte the alpha release, so that the internal structures are more stable. What do you think?

thesyncim:
ok sounds good to me

thesyncim:
Meanwhile I will work on a go interface for vss snapshot

fd0:
That's a great idea! Thanks for your support and braveness to attempt working on the windows support for restic! ;)

fd0:
I'll close this PR for now.

Issue #105, opened 2015-03-14
Find latest parent snapshot automatically


fd0:
Also add command line parameter to force re-reading all data.

PR #119, opened 2015-04-04
Take last snapshot as parent by default and allow force re-reading

rakoo:
This solves #105

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -13,7 +13,8 @@ import (
 )
 
 type CmdBackup struct {
-	Parent string `short:"p" long:"parent"    description:"use this parent snapshot (default: not set)"`
+	Parent string `short:"p" long:"parent"    description:"use this parent snapshot (default: last snapshot in repo)"`
+	AsNew  bool   `short:"n" long:"asnew" description:"consider the target as new and re-read it. Overrides the \"parent\" flag"`
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


fd0:
Hey, thanks for your contribution! This looks like a good start, here are my comments:
 * I think it would be good to change `--asnew` to `--force`, this makes it clearer
 * Why did you move the search for a parent snapshot after the scan phase? restic should abort as soon as possible without doing any work if there are wrong command line parameters.
 * At the moment the code only looks at cached snapshots, this should be changed to use the list of snapshots from the repository, the local cache should only be queried before reading the snapshot from the repo.
 * If `--force` is given, do not search for a snapshot, just create a new one

You can re-submit the commit after reworking it byy pushing it to your master branch with `git push -f`, it will appear here in this pull request and overwrite the old commit.

Thanks!

rakoo:
Hey fd0,

I thought it would be best to get content from the cache after it was refreshed, since at that moment we have all the trees locally and we wouldn't have to fetch them from the repository a 2nd time (if I understand restic correctly), possibly on a remote location.

I thought the lookup for a parent also needed that, but it actually doesn't query the cache.

fd0:
Another thought: The snapshots should also be checked whether the same directories/files have been specified as to be backed up. If I create backups of `~/foo` and `~/bar` separately, it should find the matching latest parent commit.

rakoo:
I created a proper branch and re-submitted a PR in #121

PR #121, opened 2015-04-07
Take last snapshot as parent by default and allow force re-reading

rakoo:
A rework of #119 with a proper branch, following your indications.


fd0:
Thanks for your contribution!

Issue #56, opened 2014-11-18
Bundle small files together for storage


fd0:
Transfer of blobs is much more efficient for large blobs, but restic stores small files into their own blob.

Implement bundling of small files into larger blobs.

fd0:
Need more discussion how to handle this.

cfcs:
How are you looking to structure this? By finding the largest incomplete (<8MB) chunk that fits with the file size and adding it to the end, and storing an index / offset + length with the filename/inode information?
By using dedicated chunks for small files you could avoid storing offset / lengths for regular chunks (used to store chunks of large files).

You said you'd like discussion about this feature, do you have any problems in mind which you'd like to discuss?

fd0:
Packing of small blobs into larger ones and storing the offset in the map structure in a tree currently interferes with the archiver pipeline, as the tree structure of a directory depends on the exact location of all data blobs and all subtrees.

Therefore I think that packing must be done on a lower layer and the tree structure shouldn't include any information if blobs are packed. This could e.g. be realised by introducing pack-files with the extension ".pack" that include an encrypted and signed header which lists the chunks that are packed together in this file. For each chunk, the header lists the storage hash, offset and length. This information can be cached locally. If the cache is not available, restic has to read the header of all pack files to find out which blocks are stored within.

That's a rough draft on how this might be implemented. Any thought on that?

andrewchambers:
You need an unencrypted index file if you want fast random access for fuse etc. Though the decrypted index file could be held in memory by the file server I suppose.

cfcs:
Ah, right, that makes sense @ one-pass pipeline.

The index / header you're describing would have linear access times, don't you think a hash table or similar data structure allowing more efficient random access would be better?
Perhaps you already do this for the normal chunks? Or do you just scan the entire thing and put it in a hash map in memory when restoring the backup? (That would work too).

Potential small optimization: If you store the length of the pack-file and the offsets, you don't need to store the length of each individual chunk in the file since you can calculate that by subtracting the current offset from the next offset (or the length of the file, if it's the last chunk).

ON THE OTHER HAND, if you keep the (length + offset) fields, and you make those fields constant size, you can sort the index alphabetically and find blobs efficiently through a binary search of the index, which may be a useful feature to have.

You're much more familiar with the internals of restic, so I'm not sure whether my suggestions are of any use, but I think that what you proposed sounds reasonable, and relatively clean.

[...]

PR #122, opened 2015-04-07
Add fuse mount for browsing snapshots

### LINKED ISSUE #27: Implement fuse file system for snapshots

rakoo:
For consideration only; I don't expect this to be merged as-is.

Here's a very simple `mount` command to mount a restic repo on some mountpoint through fuse.

To use, do this:

``` shell
$ mkdir /some/mount/path
$ make
$ ./restic -r /path/to/repo mount /some/mount/path
(do your stuff in /some/mount/path: cat, mplayer, thunar, whatever)
(stop the mount with ^C)
$ sudo fusermount -u /some/moutn/path # this step is important, I don't know the gritty details of fuse/filesystems so I couldn't explain why
```

Some notes about how it works:
- Everything is read-only (dirs are executable to be browseable)
- It seems to work ok for most files...
- ...except for videos, it is extremely laggy. A quick look at `iotop` shows almost no disk usage; further profiling is needed
- I think using `mount` as a command name makes more sense than using `fuse`, because it is the actual thing the user wants to do whereas `fuse` is just the implementation.

Here's the hierarchy (inspired by bup):

```
mountpath
  `- snapshots (a pseudo dir)
        |- <snapshot date1> (another pseudo dir)
        |- <snapshot date2>
        | # there will be a "latest" symlink here that points to the latest snapshot directly
        `- <snapshot date3>
               `- <snapshot top dir, such as /home/rakoo/Documents>
                       |- dir1 (here we start real dirs)
                       |   |- dir 11
                       |   |    `- file 111
                       |   `- file 12
                       `- dir2
```

Closes #27


fd0:
Wow, this already looks really good! I'll refactor things a bit while implementing #56, afterwards we can tackle the fuse mount. If you are interested in a chat, you can find me in #restic on irc.freenode.net.

fd0:
Hey, are you still interested? Bundling of small blobs together has been implemented, so it's not a blocker any more.

I'll reopen this pull request in the hope that you have the motivation to implement this :+1:

rakoo:
Oh, I'm sorry, this fell completely out of my radar ! I'll have a look as soon as I have some time.

fd0:
Great, I'm looking forward to it!

fw42:
This is super cool. Can't wait for this to be merged.

rakoo:
New version ! Use it like that:

```shell
$ restic -r /path/to/your/repo mount /path/to/mountpoint
Now serving /path/to/your/repo under /path/to/mountpoint
Don't forget to umount after quitting !
```

Now you can access `/path/to/mountpoint` anyway you like. You can even add new snapshots here, they will be visible right away. After you're done with the mount:

```shell
$ sudo umount -l /path/to/mountpoint
```

Some comments:
* The build fails for `openbsd/386` and `openbsd/amd64` because of fuse.
* I was thinking that #60 could be solved with a standard webserver pointing to a mounted repo, but Windows doesn't have fuse ...

Future possible improvements:
* add a `/snapshots/latest` that is a symlink to latest snapshot
* add a `/recent` that has, say, the 500 last modified files

fd0:
Thanks for your work, this already looks really really good! I've addressed a few minor things with comments, the main things we need to work on to get this merged are:
 * Either disable fuse support for OpenBSD or make it work with it (should be not a big problem)
 * Add a test for the mount command

Very cool!

fd0:
Hm, it looks like we can't run CI tests requiring fuse on the Travis container infrastructure...

fd0:
I've opened https://github.com/bazil/fuse/pull/104 which adds support for openbsd to fuse.

fd0:
I just noticed that you vendored an old version of the fuse library from April, could you use the current one?

fd0:
Looks good! :+1:

fw42:
If you rebase this branch then the OS X tests shouldn't hang anymore

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+			global: &globalOpts,
+			ready:  make(chan struct{}, 1),
+		})
+	if err != nil {
+		panic(err)
+	}
+}
+
+func (cmd CmdMount) Usage() string {
+	return "MOUNTPOINT"
```
fw42: what's this?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func (cmd CmdMount) Execute(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("wrong number of parameters, Usage: %s", cmd.Usage())
```
fw42: Usage should be lowercase here I guess?

fw42: also I think "arguments" instead of "parameters" is more clear

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	err = repo.LoadIndex()
+	if err != nil {
+		return err
+	}
+
+	mountpoint := args[0]
+	if _, err := os.Stat(mountpoint); err != nil {
+		if os.IsNotExist(err) {
```
fw42: since this is the only thing you check here, you could move this into the other `if`, i.e.:
`if _, err := os.Stat(mountpoint); os.IsNotExist(err) { ... }`.

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+			err = os.Mkdir(mountpoint, os.ModeDir|0700)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	c, err := fuse.Mount(
+		mountpoint,
+		fuse.ReadOnly(),
+		fuse.FSName("restic"),
```
fw42: is this a unique identifier or something like that? just out of curiosity, but what happens if you mount two repositories (or two snapshots from the same repository) at once? i.e. with two restic processes. Does that work?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+		return err
+	}
+
+	root := fs.Tree{}
+	root.Add("snapshots", &snapshots{
+		repo:           repo,
+		knownSnapshots: make(map[string]*restic.Snapshot),
+	})
+
+	cmd.global.Printf("Now serving %s under %s\n", repo.Backend().Location(), mountpoint)
```
fw42: nitpick: `s/under/at/` maybe?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+	cmd.global.Printf("Don't forget to umount after quitting !\n")
```
fw42: s/ !/!/

cmd/restic/cmd_mount.go
rakoo: I did the same as was in other commands, such as `cmd_backup.go`: the expected argument is a path to the mountpoint.

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+		return err
+	}
+
+	<-c.Ready
+	return c.MountError
+}
+
+// These lines statically ensure that a *snapshots implement the given
+// interfaces; a misplaced refactoring of the implementation that breaks
+// the interface will be catched by the compiler
```
fw42: Sorry, but I don't understand this comment (pretty much none of it). Can you explain or maybe rephrase it so that it's more clear?

fw42: If I'm reading this right, this says "this is here so that possible future refactorings in the future wont break the interface" or something like that? That seems like something that should be in a unit test, not in the code.

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+var _ = fs.NodeStringLookuper(&snapshots{})
+
+type snapshots struct {
+	repo *repository.Repository
+
+	// snapshot timestamp -> snapshot
+	knownSnapshots map[string]*restic.Snapshot
+}
+
+func (sn *snapshots) Attr(ctx context.Context, a *fuse.Attr) error {
```
fw42: s/a/attr/ maybe for clarity?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+// interfaces; a misplaced refactoring of the implementation that breaks
+// the interface will be catched by the compiler
+var _ = fs.HandleReadDirAller(&snapshots{})
+var _ = fs.NodeStringLookuper(&snapshots{})
+
+type snapshots struct {
+	repo *repository.Repository
+
+	// snapshot timestamp -> snapshot
+	knownSnapshots map[string]*restic.Snapshot
```
fw42: why not make this `map[time.Time]*restic.Snapshot`?

fw42: I don't know how the go fuse package you are using is implemented in terms of concurrency, but do we need to mutex the access to this map so that it can safely be used concurrently? for example if ReadDirAll or Lookup are called concurrently?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+func (sn *snapshots) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
+	ret := make([]fuse.Dirent, 0)
+	for id := range sn.repo.List(backend.Snapshot, ctx.Done()) {
+		snapshot, err := restic.LoadSnapshot(sn.repo, id)
+		if err != nil {
+			return nil, err
+		}
+		sn.knownSnapshots[snapshot.Time.Format(time.RFC3339)] = snapshot
+		ret = append(ret, fuse.Dirent{
+			Inode: binary.BigEndian.Uint64(id[:8]),
```
fw42: Why do we care about the endianess?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+	snapshot := sn.knownSnapshots[name]
+	tree, err := restic.LoadTree(sn.repo, snapshot.Tree)
+	if err != nil {
+		return nil, err
+	}
+	return &dir{
+		repo: sn.repo,
+		tree: tree,
+	}, nil
+	return nil, fuse.ENOENT
```
fw42: this line is dead code

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// Statically ensure that *dir implement those interface
```
fw42: again, this should be in a test, imho

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+	return c.MountError
+}
+
+// These lines statically ensure that a *snapshots implement the given
+// interfaces; a misplaced refactoring of the implementation that breaks
+// the interface will be catched by the compiler
+var _ = fs.HandleReadDirAller(&snapshots{})
+var _ = fs.NodeStringLookuper(&snapshots{})
+
+type snapshots struct {
```
fw42: I feel like this type (and all it's methods) should be in a separate file, not in the cmd file. And maybe be renamed to something like FuseSnapshots? Does that make sense?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+			Name:  snapshot.Time.Format(time.RFC3339),
+		})
+	}
+
+	return ret, nil
+}
+
+func (sn *snapshots) Lookup(ctx context.Context, name string) (fs.Node, error) {
+	if _, ok := sn.knownSnapshots[name]; !ok {
+		// At least this snapshot is not cached. We use this opportunity to
```
fw42: What does "at least" mean? Are the other snapshots cached? Why is that? And why is it important here? Does this comment really add value? Would the code be harder to understand without it? Can the code be changed in a way that it's understandable without the comment?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+		tree: tree,
+	}, nil
+	return nil, fuse.ENOENT
+}
+
+// Statically ensure that *dir implement those interface
+var _ = fs.HandleReadDirAller(&dir{})
+var _ = fs.NodeStringLookuper(&dir{})
+
+type dir struct {
```
fw42: Hmm I think this should be a separate file too (and it's methods). Not a fan of having the cmd stuff and two separate types all in one big 300 line file.

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+			Name:  node.Name,
+		})
+	}
+
+	return ret, nil
+}
+
+func (d *dir) Lookup(ctx context.Context, name string) (fs.Node, error) {
+	for _, node := range d.tree.Nodes {
+		if node.Name == name {
```
fw42: hmm, maybe premature, but is there no better way to find a node with a given name other than to iterate over all nodes?

cmd/restic/cmd_mount.go
rakoo: It's alright to have multiple file systems with the same name, here's the output of my `df -a`:

```
proc                       0         0          0    - /proc
sys                        0         0          0    - /sys
dev                  1529256         0    1529256   0% /dev
run                  1531952       592    1531360   1% /run
/dev/sda4          144064184 131102472    5620608  96% /
securityfs                 0         0          0    - /sys/kernel/security
tmpfs                1531952       176    1531776   1% /dev/shm
devpts                     0         0          0    - /dev/pts
tmpfs                1531952         0    1531952   0% /sys/fs/cgroup
cgroup                     0         0          0    - /sys/fs/cgroup/systemd
pstore                     0         0          0    - /sys/fs/pstore
cgroup                     0         0          0    - /sys/fs/cgroup/net_cls
cgroup                     0         0          0    - /sys/fs/cgroup/freezer
cgroup                     0         0          0    - /sys/fs/cgroup/cpu,cpuacct
cgroup                     0         0          0    - /sys/fs/cgroup/devices
cgroup                     0         0          0    - /sys/fs/cgroup/blkio
cgroup                     0         0          0    - /sys/fs/cgroup/cpuset
cgroup                     0         0          0    - /sys/fs/cgroup/memory
systemd-1                  -         -          -    - /proc/sys/fs/binfmt_misc
hugetlbfs                  0         0          0    - /dev/hugepages
mqueue                     0         0          0    - /dev/mqueue
tmpfs                1531952    199780    1332172  14% /tmp
debugfs                    0         0          0    - /sys/kernel/debug
configfs                   0         0          0    - /sys/kernel/config
fusectl                    0         0          0    - /sys/fs/fuse/connections
tmpfs                 306392        12     306380   1% /run/user/1000
restic                     0         0          0    - /tmp/restic-mount
restic                     0         0          0    - /tmp/restic-mount2
binfmt_misc                0         0          0    - /proc/sys/fs/binfmt_misc
```

So, yeah, mounting the same repo on 2 different mount points (`/tmp/restic-mount` and `/tmp/restic-mount2` here) works.

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+			switch {
+			case node.Mode.IsDir():
+				subtree, err := restic.LoadTree(d.repo, node.Subtree)
+				if err != nil {
+					return nil, err
+				}
+				return &dir{
+					repo:  d.repo,
+					tree:  subtree,
+					inode: binary.BigEndian.Uint64(node.Subtree[:8]),
```
fw42: This is here for the second time now. Would it make sense to refactor it into it's own method? Something like `inodeForNode()` maybe?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+// Statically ensure that *file implements the given interface
+var _ = fs.HandleReader(&file{})
+
+type file struct {
+	repo *repository.Repository
+	node *restic.Node
+
+	sizes []uint32
+
+	// cleartext contents
```
fw42: instead of this comment, why not rename the variable to `cleartextContent` and make it self-explanatory?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+	clearContent [][]byte
+}
+
+func makeFile(repo *repository.Repository, node *restic.Node) (*file, error) {
+	sizes := make([]uint32, len(node.Content))
+	for i, bid := range node.Content {
```
fw42: Maybe `backendID` instead of `bib`?

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+		_, _, _, length, err := repo.Index().Lookup(bid)
+		if err != nil {
+			return nil, err
+		}
+		sizes[i] = uint32(length) - crypto.Extension
```
fw42: can you explain the `- crypto.Extension`? somehow I feel like this doesn't belong here (not in the sense that the size will be wrong but in the sense that this file here should not be concerned with any crypto things and this here is the wrong level of abstraction for this).

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadResponse) error {
+	off := req.Offset
+	content := make([]byte, req.Size)
+	allContent := content
+
+	for i := range f.node.Content {
+		if off >= int64(f.sizes[i]) {
+			off -= int64(f.sizes[i])
```
fw42: can you explain this? I don't get it

cmd/restic/cmd_mount.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func (f *file) Attr(ctx context.Context, a *fuse.Attr) error {
+	a.Inode = f.node.Inode
+	a.Mode = f.node.Mode
+	a.Size = f.node.Size
+	return nil
+}
+
+func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadResponse) error {
```
fw42: I find this method hard to understand... is there any way we could make it easier? maybe by breaking it down into smaller ones with descriptive names?

cmd/restic/integration_fuse_test.go
```diff
@@ -0,0 +1,101 @@
...
+		Assert(t,
+			len(namesInSnapshots) == len(snapshotIDs),
+			"Invalid number of snapshots: expected %d, got %d", len(snapshotIDs), len(namesInSnapshots))
+
+		for i, id := range snapshotIDs {
+			snapshot, err := restic.LoadSnapshot(repo, id)
+			OK(t, err)
+			Assert(t,
+				namesInSnapshots[i] == snapshot.Time.Format(time.RFC3339),
+				"Invalid snapshot folder name: expected %s, got %s", snapshot.Time.Format(time.RFC3339), namesInSnapshots[i])
```
fw42: `s/folder/directory/` maybe?

.travis.yml
```diff
@@ -9,7 +9,7 @@ os:
   - linux
   - osx
 
-env: GOX_OS="linux darwin openbsd freebsd" GOX_ARCH="386 amd64 arm"
+env: GOX_OS="linux darwin openbsd freebsd" GOX_ARCH="386 amd64 arm" RESTIC_TEST_FUSE="0"
```
fw42: why is this necessary?

cmd/restic/cmd_mount.go
rakoo: The other way is to have a `map[name string]*restic.Node` at the `dir` level. At first glance it doesn't make the code a lot more complicated; I'll see what it looks like with it.

cmd/restic/cmd_mount.go
rakoo: I'm partial to having this check done by the compiler, because it happens earlier in the code -> compile -> test chain, and it allows us to write less code, which is a big win for me.

cmd/restic/cmd_mount.go
rakoo: Regarding using `time.Time` as a key: [the doc](https://golang.org/pkg/time/#Time) says that to do that you must ensure that the location is correct. Since I don't think the location can realistically change after the mount started, I believe it's safe to do it.

Regarding concurrency: I don't know if fuse allows parallel calls, but even if it does, the key->value mapping will always happen; can we consider that the go runtime can somehow half-fill a map ? If not, it's safe to leave it like that.

cmd/restic/cmd_mount.go
fw42: I guess I don't understand what this does then. What could break if those two lines (below) are removed? Why would the compiler not catch that anyway even without those two lines?

cmd/restic/cmd_mount.go
rakoo: We don't; we just need some unique `uint64`, and the most unique thing we have here is the id, so I took the simplest mapping from []byte to uint64 I could think of.

cmd/restic/cmd_mount.go
rakoo: Correction: instead of trying to be clever, I also think it's safer to mutex the access to this map here.

cmd/restic/cmd_mount.go
rakoo: "At least" means the snapshot we are looking for doesn't exist in the cache, and there might ever be other. We don't really care about those here, we're just helping future callers by reloading the cache.

We could introduce a `func (sn *snapshots) updateCache()` method here, I guess it is self-explaining.

cmd/restic/cmd_mount.go
rakoo: Yeah, this calls for a `Index.LookupSize(id)`.

cmd/restic/cmd_mount.go
rakoo: Because the 2 interfaces are optional (only the implementation of `Attr` is mandatory); the compiler will not complain, but the Readdir and the Lookup will not return anything. Which will be catched by the tests anyway.

fw42: Interesting, I didn't know that "optional interfaces" are a thing

cmd/restic/cmd_mount.go
fd0: That's a different issue (and I'm the one to blame), it's the same for the other cmds.

cmd/restic/cmd_mount.go
fd0: As far as I know this is just an identifier for the type of file system, like `tmpfs` or `sshfs`. Using `restic` is fine.

cmd/restic/cmd_mount.go
fd0: fuse and the Go fuse implementation indeed server requests concurrently, that's why there's a Context object which can cancel the request.

I think we should use a `sync.RWMutex` here, you can embed that into the `snapshot` struct, e.g.:
```go
type snapshots struct {
	repo *repository.Repository

	// snapshot timestamp -> snapshot
	knownSnapshots map[string]*restic.Snapshot
	sync.RWMutex
}
```

It can then be used like this for reading:
```go
func (sn *snapshots) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
	sn.Lock()
	defer sn.Unlock()
	[...]
}
```

and for writing:
```go
func (sn *snapshots) Lookup(ctx context.Context, name string) (fs.Node, error) {
	sn.RLock()
	defer sn.RUnlock()
	[...]
}

File: cmd/restic/cmd_mount.go (hunk 10 above)
Comment (fd0) [inline cmd/restic/cmd_mount.go]: Clever idea, actually. I like it.

File: cmd/restic/cmd_mount.go (hunk 21 above)
Comment (fd0) [inline cmd/restic/cmd_mount.go]: It's a read request starting at a given offset. `node.Content` is a list of blobs and offsets, this code skips over all the blobs that lie before `req.Offset` in the file.

File: cmd/restic/fuse/dir.go [hunk 25]
```diff
@@ -0,0 +1,100 @@
...
+	}
+	children := make(map[string]*restic.Node)
+	for _, node := range tree.Nodes {
+		children[node.Name] = node
+	}
+
+	return &dir{
+		repo:     repo,
+		children: children,
+		inode:    binary.BigEndian.Uint64(snapshot.ID),
```
fw42: in the other place, you are using `binary.BigEndian.Uint64(snapshot.ID[:8])`. Where is the `[:8]` here?

fw42: I still don't understand why we need big endian here. What would break if we would just do `inode: snapshot.ID`?

fd0: It won't compile. `inode` is an uint64, `snapshot.ID` a `[]byte`. It doesn't matter if it's little or big endian, as long as the result is an uint64.

fw42: ah ok that makes sense. Thanks for the explanation.

rakoo: That's a mistake, it should also be [:8].

Also, inode is a `uint64` while snapshot.ID is a `backend.ID` which is a `[]byte` in disguise. It can't be used as a uint64.

fd0:
Got it, the following patches fixes running the tests on Darwin:
 * https://gist.github.com/b5faeafb8b449099c6a6
 * https://gist.github.com/4a78458f13a1b8745baf (already pushed to master, you just need to rebase)

The first adds a timed test for the subdir `snapshots`, it's a race condition between sending to channel `ready` in `CmdMount.Execute()` and continuing the test. It seems that this only occurs in virtual machines, I haven't seen it on my Linux machine here.

The second patch fixes installing `osxfuse` on darwin. Here, the problem was that the osxfuse version installed with `brew` doesn't include a signed kernel module, so loading it fails. This is mitigated by using `cask` to install the official release of `osxfuse`, then it works.

Running the code on a real OS X VM with vagrant was invaluable for debugging this. I've pushed the `Vagrantfile` in 9441442a4d84313b993b1aa6df0729cb85a48566.

You can see that the mount suceeds here: https://travis-ci.org/restic/restic/builds/71675868

The error (with that revision of the fuse mount code) also occurs on my machine on Linux.

fw42:
:heart:

fd0:
The darwin tests pass! awesome!

rakoo:
Congrats everyone on this epic PR ! I thought I was prepared for harsh code reviews, I was not :)

fw42:
Sorry for being harsh :)

fd0:
@rakoo you did great work, I just verified that the fuse mount works like charm, even on freebsd :)

rakoo:
Sweet !

PR #127, opened 2015-04-24
gitignore binary files

fw42:
I keep accidentally `git add .`ing those binary files... :-/

This PR adds the files that are generated by `make` to the gitignore list.


fd0:
FYI: This can also be solved by adding the paths to `.git/info/excludes`.

PR #129, opened 2015-04-24
Prefer mixedCaps over underscore

fw42:
https://golang.org/doc/effective_go.html#mixed-caps

> Finally, the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.


fd0:
Thanks for going through all this. I've used `snake_case` all my life and I'm still adapting to the Go coding style. How did you rename all this? Are you aware of gorename and vim-go?

PR #130, opened 2015-04-24
Consistent writing

fw42:
Super nitpick, but kinda bugged me :-) This makes things more consistent with

```
enter password for new key
```

```
enter password again
```

```
enter password for new backend
```

(all lowercase)

I noticed some other inconsistencies: Sometimes log messages begin with the first word being capitalized, sometimes they don't. And sometimes they end with periods and sometimes they don't.

Not a huge deal I guess, but looks kinda "unpolished" (which I guess it is :-))

Maybe open an issue to make all user-faced logging use a consistent style?

### DIFF
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -146,7 +146,7 @@ func OpenRepo() (restic.Server, error) {
 
 	s := restic.NewServer(be)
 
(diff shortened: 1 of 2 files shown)


fd0:
Heh, you beeing nitpick got me through the math exams. Talking about Fundamentalmaschen. Merged.

PR #131, opened 2015-04-25
O_NOATIME on Linux

### LINKED ISSUE #53: Preserve atime on backup
khepri should not touch anything in the file system when making a backup. This especially includes not modifying the atime of files. Implement opening files with the O_NOATIME flag.

fw42:
Fixes https://github.com/restic/restic/issues/53.

FYI: You might have to remount your filesystem with the `strictatime` option to actually see the behaviour that this PR is fixing, since Linux seems to mount with `relatime` by default nowadays (which disables updating the atime unless it's newer than the mtime), so on many (most?) Linux systems, this isn't actually an issue. But it's an easy fix, so here it is.


fd0:
Thanks, could you add a test for it?

PR #132, opened 2015-04-25
rename "change" to "passwd"

### LINKED ISSUE #76: rename key management command "change"
rename change command:

``` restic key change 63b0509a ```

Change implies modifications to the key, e.g. hostname and others, in fact it is only the password. Suggestion: rename 'change' to 'changepwd'.

fw42:
Closes https://github.com/restic/restic/issues/76.

I think `passwd` makes more sense than the suggested `changepwd`, in analogy to the unix `passwd(1)` command.


fd0:
Good idea, merged.

PR #137, opened 2015-04-25
Fix floating point presentation

fw42:
If your backup is 0 bytes, you can see this:

```
[0:00] NaN%  0B/s  0B / 0B  1 / 0 items  ETA 0:00
duration: 0:00, NaNMiB/s
```

and if your backups is >0 bytes, but very fast, you see this:

```
duration: 0:00, +InfMiB/s
```

This PR fixes that by avoiding floating point divisions by 0. If a percentage is calculated and the denominator is 0, just return an empty string. The "Inf" is fixed by not converting the number of seconds to an int before the division.


fd0:
Oh, that's a really annoying issue, thanks for tackling it!

PR #141, opened 2015-04-25
Fix progress off-by-1

fw42:
Fixes this:

```
[0:00] NaN%  0B/s  0B / 0B  2 / 1 items  ETA 0:00
```

The number of "done" items is always one bigger than the number of "todo" items.

Looks like the root node is counted twice. I didn't quite understand the code, but I guess you are creating one node with an empty Path() and then another node below that with the name of the actual directory? 

Not sure if this is the best way to fix it, but it does seem to work :-) Please suggest something else if there is a cleaner solution. This does feel kinda hacky.


fd0:
I like your proposal, merged.

PR #142, opened 2015-04-26
Fix O_NOATIME permission error

fw42:
NOATIME is only allowed if you are the owner of a file (or root).

This PR retries without the flag if it fails because of permission errors.

### DIFF
--- a/node_linux.go
+++ b/node_linux.go
@@ -12,7 +12,11 @@ import (
 )
 
 func (node *Node) OpenForReading() (*os.File, error) {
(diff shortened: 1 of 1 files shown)


fw42:
Alternatively, we could check if we are the owner (or root) before trying to open the file. Not sure which approach is better (faster?). Thoughts?

fd0:
I like your proposal, merged.

fd0:
I think a test for ownership is easily integrated, we have this information at hand in the node structure, but I'm not sure if it's worth it.

PR #143, opened 2015-04-26
Count errors separately and not as files that are "done"

fw42:
Currently, files that are raising errors while trying to be backed up will be counted towards the number of files that are "done", which I think is confusing. Instead, count "errors" separately.


fw42:
damn it. clicked some wrong buttons somewhere.

PR #147, opened 2015-04-28
golint

fw42:
Fixes a few minor style issues that [golint](https://github.com/golang/lint) is complaining about, like consistent naming for receivers, etc.

Everything that's still left is about missing documentation on exported fields/functions.


fd0:
This fails due to an unused package. Please consider using e.g. vim-go with goimports auto-format on save, in addition to the pre-commit hook for gofmt from the README. Maybe we could adapt this hook to use goimports...

fw42:
fixed

PR #149, opened 2015-04-29
merge cache_darwin and cache_linux

fw42:
Those two files were exactly identical, so there is no need to have them as two separate files if the same code is compatible with both platforms anyway.


fd0:
Yeah, this was introduced for the windows port, which has been removed in the meantime. It can probably be implemented in a more sane way via built tags later (only build this file for linux and darwin, not windows).

fw42:
Cool so it can be removed for now? I hate seeing so much code duplication. The platform-specific parts should be as tiny as possible, imho.

fd0:
I think that's a good idea.

Issue #148, opened 2015-04-29
Darwin tests broken


fw42:
Getting this on Darwin:
```
run test testsuite/test-backup-incremental.sh

file node "file" of tree Blob<adbaa711 (947) -> c4c17a09 (979)> has size 38, but only 9 bytes could be found
snapshot 439a6dc6c22ab6ca5156c26976a3a69eae7533aec2f925b7307614311873a75d:
  error for tree Blob<1cea3eb5 (491) -> 04bd78a8 (523)>:
    file node "file" of tree Blob<adbaa711 (947) -> c4c17a09 (979)> has size 38, but only 9 bytes could be found
check for snapshot 439a6dc6c22ab6ca5156c26976a3a69eae7533aec2f925b7307614311873a75d failed
file node "file" of tree Blob<0d33fe11 (947) -> d570bc06 (979)> has size 67, but only 9 bytes could be found
snapshot be26a97b91ae9c9378272b999b6cb96b988bd6004c33d1ea7ee81f93436b0c0b:
  error for tree Blob<85c3f586 (491) -> 9624bf3c (523)>:
    file node "file" of tree Blob<0d33fe11 (947) -> d570bc06 (979)> has size 67, but only 9 bytes could be found
check for snapshot be26a97b91ae9c9378272b999b6cb96b988bd6004c33d1ea7ee81f93436b0c0b failed
file node "file" of tree Blob<0d33fe11 (947) -> d570bc06 (979)> has size 67, but only 9 bytes could be found
test-backup-incremental failed!
```

Reverting https://github.com/restic/restic/commit/1d0f8fa80ebce0289b9f4791e87c4324072471f7 makes the tests happy again. Not sure why exactly.

@fd0 @rakoo

fd0:
Please test it again later this evening when the code for packing blobs has been push (see #56), as this area of code has been heavily modified.

fw42:
not sure why or how (maybe somehow due to #150), but tests are passing now for me on Darwin

PR #150, opened 2015-04-29
Refactor node.go

fw42:
This refactors node.go in a couple of different ways. I find the changes make the file much easier to read (and I removed some code that I think was redundant).

I made the following changes. Sorry it's so many in one PR. I can try to break it into several smaller PRs if necessary.
- `nodeTypeFromFileInfo` was taking a `path` argument that wasn't used. I removed it.
- Renamed `CreateNodeAt(node, things, path)` to `node.CreateAt(path, things)`, since I think that's more idiomatic.
- Split the `CreateAt` function into several smaller functions, such that `CreateAt` itself operates on only one level of abstraction. I think this makes the function easier to understand.
- Remove a bunch of obvious (and therefore imho useless) comments as well as a few TODO comments. (I think TODO comments are a code smell and those things should be moved into GitHub issues, not pollute the codebase).
- Get rid of a couple of useless `Chown` calls (for example the previous code was calling `Lchown` when creating directories and files and then a few lines later down it was calling `Chown` again. The new code always just calls `Lchown` once, since I think that should work as expected for all types).
- The old code was calling `Utimesnano` and then later calling `Chtimes` again. That was probably redudant too, right? I removed the `Chtimes` call.
- Break some part of the `Equals` function out into a `sameContent` function.
- Delete the `SameContent` function since it wasn't being called anywhere.
- Change a few other small style issues that I consider to be cleaner now.
- Change a `panic` to be an error instead. Not sure if this is ok, but the issue seemed to me not severe enough to crash the whole program (unknown node type).

GitHub is not doing the greatest job at displaying the diff. Maybe try the side-by-side view... Or just look at the new file on it's own. Probably easier to review.

### DIFF
--- a/node.go
+++ b/node.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/user"
 	"strconv"
 	"syscall"
 	"time"
@@ -336,4 +337,50 @@ func (node *Node) isNewer(path string, fi os.FileInfo) bool {
 
 	debug.Log("node.isNewer", "node %v is not newer", path)
 	return false
+}
+
+func (node *Node) fillUser(stat *syscall.Stat_t) error {
+	node.UID = stat.Uid
+	node.GID = stat.Gid
+
+	u, err := user.LookupId(strconv.Itoa(int(stat.Uid)))
+	if err != nil {
+		return err
+	}
+
+	node.User = u.Username
+	return nil
+}
+
+func (node *Node) fillExtra(path string, fi os.FileInfo) error {
+	stat, ok := fi.Sys().(*syscall.Stat_t)
+	if !ok {
+		return nil
+	}
+
+	node.Inode = stat.Ino
+
+	node.fillUser(stat)
+	node.fillTimes(stat)
+
+	var err error
+
+	switch node.Type {
+	case "file":
+		node.Size = uint64(stat.Size)
+		node.Links = uint64(stat.Nlink)
+	case "dir":
+	case "symlink":
+		node.LinkTarget, err = os.Readlink(path)
+	case "dev":
+		node.fillDevice(stat)
+	case "chardev":
+		node.fillDevice(stat)
+	case "fifo":
+	case "socket":
+	default:
+		err = fmt.Errorf("invalid node type %q", node.Type)
+	}
+
+	return err
 }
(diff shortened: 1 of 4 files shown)


node_darwin.go
```diff
@@ -26,24 +26,19 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) (err error) {
 	node.UID = stat.Uid
 	node.GID = stat.Gid
 
-	// TODO: cache uid lookup
-	if u, nil := user.LookupId(strconv.Itoa(int(stat.Uid))); err == nil {
+	if u, err := user.LookupId(strconv.Itoa(int(stat.Uid))); err == nil {
```
fw42: I think this was a bug (`u, nil := ...`), the error would have been swallowed, right?

fw42:
Also move the `isNewer` function from `_darwin` and `_linux` into `node.go` since the function was almost identical in both (I extracted the tiny part that's actually different). See 91a2a5f90a941aa3b2c00d64d5e79f7a6eca718d. This also fixes a bug in the boolean logic (compare the `return` values in the darwin and linux version. One has `return true` where the other one has `return false` and vice versa).

fw42:
Also moved `fillExtra` into `node.go` and extracted only the tiny piece that's actually different on Darwin. This strips down the platform-specific files even more and avoids duplication.

node.go
```diff
@@ -96,155 +98,114 @@ func nodeTypeFromFileInfo(path string, fi os.FileInfo) string {
...
-			syscall.NsecToTimespec(node.ModTime.UnixNano()),
-		}
-		err = syscall.UtimesNano(path, utimes)
-		if err != nil {
-			return arrar.Annotate(err, "Utimesnano")
+		if err := node.createDirAt(path); err != nil {
+			return err
 		}
 	case "file":
-		// TODO: handle hard links
```
fw42: should open a github issue for this

node.go
```diff
@@ -270,7 +231,6 @@ func (node *Node) UnmarshalJSON(data []byte) error {
 }
 
 func (node Node) Equals(other Node) bool {
-	// TODO: add generatored code for this
```
fw42: should open a github issue if you still care, although I think it's fine as it is

fw42:
Hu, weird, for some reason this PR fixes https://github.com/restic/restic/issues/148 for me. Not sure why.

fd0:
Thanks for your cleanup work, I'll have a look at it this evening.

node_darwin.go
fd0: I think you are right. Good catch!

PR #151, opened 2015-04-29
Pack blobs

fd0:
This implements bundling of smaller chunks into bigger files (#56). Unfortunately, this is a big change that touches a lot of code in different places, and breaks the repository format and fsck. Let me know what you think!


archiver.go
```diff
@@ -215,75 +148,71 @@ func (arch *Archiver) SaveFile(p *Progress, node *Node) (server.Blobs, error) {
...
+	if len(results) != chunks {
+		return fmt.Errorf("chunker returned %v chunks, but only %v blobs saved", chunks, len(results))
 	}
 
 	var bytes uint64
 
-	node.Content = make([]backend.ID, len(blobs))
+	node.Content = make([]backend.ID, len(results))
 	debug.Log("Archiver.Save", "checking size for file %s", node.path)
-	for i, blob := range blobs {
```
fw42: I liked `blob` better than `b` :-)

server/server.go
```diff
@@ -65,59 +77,112 @@ func (s *Server) Load(t backend.Type, blob Blob) ([]byte, error) {
...
 
-	// decrypt
-	buf, err = s.Decrypt(buf)
+	return plain, nil
+}
+
+// LoadBlob tries to load and decrypt content identified by t and id from a
+// pack from the backend.
+func (s *Server) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
+	debug.Log("Server.LoadPack", "load %v with id %v", t, id.Str())
```
fw42: should this say LoadBlob, not LoadPack? (same below)

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+// Packer is used to create a new Pack.
+type Packer struct {
+	blobs []Blob
+
+	bytes uint
+	k     *crypto.Key
+	wr    io.Writer
+	hw    *backend.HashingWriter
+
+	m sync.Mutex
```
fw42: I think you might be able to improve performance/contention if you turn this into a RWMutex, which allows arbitrary numbers of readers to hold a lock as long as no writer is holding one. As far as I can see it, there is a couple of different places which need to lock for reading, but only one place that needs to lock for writing. No reason why the readers should block each other.

fw42: (assuming things are actually happening in parallel here)

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+// calling ID().
+func (p *Packer) Finalize() (int64, error) {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	// n tracks the number of bytes written for the header
+	var n int64 = int64(p.bytes)
+
+	// create writer to encrypt header
+	wr := crypto.EncryptTo(p.k, p.hw)
```
fw42: You never call `Close()` on this writer if there is an error when writing the header. Is that ok?

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+// Finalize writes the header for all added blobs and finalizes the pack.
+// Returned are the complete number of bytes written, including the header.
+// After Finalize() has finished, the ID of this pack can be obtained by
+// calling ID().
+func (p *Packer) Finalize() (int64, error) {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	// n tracks the number of bytes written for the header
+	var n int64 = int64(p.bytes)
```
fw42: I think the type declaration is unnecessary here, you can just use `n := int64(p.bytes)`

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+
+// Finalize writes the header for all added blobs and finalizes the pack.
+// Returned are the complete number of bytes written, including the header.
+// After Finalize() has finished, the ID of this pack can be obtained by
+// calling ID().
+func (p *Packer) Finalize() (int64, error) {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	// n tracks the number of bytes written for the header
```
fw42: instead of having comments like this, why not call it `bytesWritten` instead of `n`, then it's obvious

pack/pack.go
fd0: * Florian Weingarten <notifications@github.com> wrote:
> > +		return nil, err
> > +	}
> > +
> > +	return io.LimitReader(rd, int64(e.Length)), nil
> > +}
> > +
> > +// Packer is used to create a new Pack.
> > +type Packer struct {
> > +	blobs []Blob
> > +
> > +	bytes uint
> > +	k     *crypto.Key
> > +	wr    io.Writer
> > +	hw    *backend.HashingWriter
> > +
> > +	m sync.Mutex
> 
> I think you might be able to improve performance/contention if you turn this into a RWMutex, which allows arbitrary numbers of readers to hold a lock as long as no writer is holding one. As far as I can see it, there is a couple of different places which need to lock for reading, but only one place that needs to lock for writing. No reason why the readers should block each other.

I don't think this is called in parallel a lot apart from the `Add()`
function, and that needs to change `[]blobs`.

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+	for _, b := range p.blobs {
+		entry := headerEntry{
+			Type:   b.Type,
+			Length: b.Length,
+		}
+		copy(entry.ID[:], b.ID)
+
+		err := binary.Write(wr, binary.LittleEndian, entry)
+		if err != nil {
+			return int64(n), err
```
fw42: `n` is already a `int64`, I don't think the typecast here is necessary

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+		return int64(n), err
+	}
+
+	// account for crypto overhead
+	n += crypto.Extension
+
+	// write length
+	err = binary.Write(p.hw, binary.LittleEndian, uint32(len(p.blobs)*entrySize+crypto.Extension))
+	if err != nil {
+		return int64(n), err
```
fw42: same, typecast unnecessary

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	n += int64(binary.Size(uint32(0)))
+
+	p.bytes = uint(n)
```
fw42: can this overflow? maybe `p.bytes` should be `uint64` too?

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+	ls := binary.Size(uint32(0))
+
+	// reset to the end to read header length
+	_, err = rd.Seek(-int64(ls), 2)
+	if err != nil {
+		return nil, fmt.Errorf("seeking to read header length failed: %v", err)
+	}
+
+	// read length
+	var l uint32
```
fw42: same as above.. if you call it `length` instead of `l` then the code is more readable and the comment is unnecessary :-)

pack/pack_test.go
```diff
@@ -0,0 +1,111 @@
...
+		_, err := io.ReadFull(rand.Reader, b)
+		OK(t, err)
+		h := sha256.Sum256(b)
+		bufs = append(bufs, Buf{data: b, id: h[:]})
+	}
+
+	file := bytes.NewBuffer(nil)
+
+	// create random keys
+	k := crypto.NewKey()
```
fw42: calling it `crypto.NewRandomKey()` would be more obvious and the comment would be unnecessary ;-)

(I'm gonna stop with the naming nitpicks now ;-))

server/index.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Count returns the number of blobs of type t in the index.
+func (idx *Index) Count(t pack.BlobType) (n uint) {
+	debug.Log("Index.Count", "counting blobs of type %v", t)
+	idx.m.Lock()
+	defer idx.m.Unlock()
+
+	for id, blob := range idx.pack {
+		if blob.tpe == t {
```
fw42: Hm this doesn't seem super efficient, especially if being called multiple times with different arguments `t`. Maybe it's worth it to run this only once and then cache the counts for all types so that those `Count()` queries can be instant? Or is this usually only a small number of blobs anyway?

pack/pack.go
fd0: No, not at the moment. Packs are made of chunks, which are 8MiB max, and the sum of all blobs in a pack doesn't exceed 16MiB.

fw42: just confused why we do any casting at all then... if everything will always fit into a `uint` (even when added up), why not use `uint` everywhere (instead of `uint64`)?

server/index.go
fd0: Good idea, but unnecessary: `Count()` is only used in the tests.

pack/pack.go
fd0: "Historisch gewachsen". `Finalize()` was once called `WriteTo()` and there is an interface for it in the stdlib which returns int64: https://golang.org/pkg/io/#WriterTo

I decided to rename the function because the semantic doesn't quite fit, the return type stayed.

In Go, `uint` is either `uint32` or `uint64`, depending on the architecture, so your considerations are important.

cache.go
```diff
@@ -389,3 +243,5 @@ func GetCacheDir() (string, error) {
 
 	return cachedir, nil
 }
+
+// TODO: implement RefreshIndex()
```
fw42: :rage:

cmd/restic/cmd_backup.go
```diff
@@ -278,16 +263,7 @@ func (cmd CmdBackup) Execute(args []string) error {
...
-	if err != nil {
-		return err
-	}
-
-	fmt.Printf("loading blobs\n")
-	err = arch.Preload()
-	if err != nil {
-		return err
-	}
+	// TODO: load index
```
fw42: what's this? this sounds like important functionality :-) is this ok to be missing?

cmd/restic/cmd_fsck.go
```diff
@@ -153,11 +148,11 @@ func fsckTree(opts CmdFsck, s *server.Server, blob server.Blob) error {
...
-	for _, id := range tree.Map.IDs() {
-		if seenIDs.Find(id) != nil {
-			return fmt.Errorf("tree %v: map contains unused ID %v", blob.ID, id)
-		}
-	}
+	// for _, id := range tree.Map.IDs() {
+	// 	if seenIDs.Find(id) != nil {
+	// 		return fmt.Errorf("tree %v: map contains unused ID %v", id, id)
+	// 	}
+	// }
```
fw42: what's this? can this be deleted?

fd0: fsck is not yet fixed for the new repository structure with packs, I have to work on that.

cmd/restic/cmd_fsck.go
```diff
@@ -249,40 +245,26 @@ func (cmd CmdFsck) Execute(args []string) error {
...
-				err := s.Remove(d.tpe, name)
-				if err != nil {
-					return err
-				}
-			}
+			fmt.Printf("removing orphaned %v blob %v\n", blob.Type, blob.ID)
+			// err := s.Remove(d.tpe, name)
+			// if err != nil {
+			// 	return err
+			// }
```
fw42: can this be deleted?

cmd/restic/cmd_fsck.go
```diff
@@ -249,40 +245,26 @@ func (cmd CmdFsck) Execute(args []string) error {
...
-		tpe  backend.Type
-		set  *backend.IDSet
-	}{
-		{"data blob", backend.Data, cmd.o_data},
-		{"tree", backend.Tree, cmd.o_trees},
-	}
-
-	for _, d := range l {
-		debug.Log("restic.fsck", "checking for orphaned %v\n", d.desc)
+	cnt := make(map[pack.BlobType]*backend.IDSet)
```
fw42: what does `cnt` stand for? can you use a more descriptive name?

cmd/restic/cmd_backup.go
fd0: Leftover comment, the index is loaded in line 183.

doc/Design.md
```diff
@@ -99,39 +92,47 @@ A repository can be initialized with the `restic init` command, e.g.:
...
-is encrypted and signed. The Pack's structure is as follows:
+The Pack's structure is as follows:
 
-    NONCE || Header_Length ||
-    IV_Header || Ciphertext_Header || MAC_Header  ||
-    IV_Blob_1 || Ciphertext_Blob_1 || MAC_Blob_1 ||
-    [...]
-    IV_Blob_n || Ciphertext_Blob_n || MAC_Blob_n ||
-    MAC
+    EncryptedBlob1 || ... || EncryptedBlobN || EncryptedHeader || Header_Length
```
fw42: why's the header at the end?

doc/Design.md
```diff
@@ -99,39 +92,47 @@ A repository can be initialized with the `restic init` command, e.g.:
...
 
-`NONCE` consists of 16 bytes and `Header_Length` is a four byte integer in
-little-endian encoding.
+At the end of the Pack is a header, which describes the content and is
+encrypted and signed. `Header_Length` is the length of the encrypted header
+encoded as a is a four byte integer in little-endian encoding.
```
fw42: "as a is a"

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	// n tracks the number of bytes written for the header
+	var n int64 = int64(p.bytes)
+
+	// create writer to encrypt header
+	wr := crypto.EncryptTo(p.k, p.hw)
+
+	// write header
```
fw42: unless it's a super critical performance hot path, I would prefer to not have comments like this but instead of a separate `p.writeHeader(wr)` function. That makes the code more readable and easier to refactor, imho.

pack/pack.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// Unpacker is used to read individual blobs from a pack.
+type Unpacker struct {
+	rd      io.ReadSeeker
+	Entries []Blob
+	k       *crypto.Key
+}
+
+// NewUnpacker returns a pointer to Unpacker which can be used to read
```
fw42: nitpick but I would argue that the fact that it's a pointer is an implementation detail that's kinda irrelevant. I wouldn't mention it.

pack/pack_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/restic/restic/backend"
+	"github.com/restic/restic/crypto"
+	"github.com/restic/restic/pack"
+	. "github.com/restic/restic/test"
+)
+
+var lengths = []int{23, 31650, 25860, 10928, 13769, 19862, 5211, 127, 13690, 30231}
+
+// var lengths = []int{200}
```
fw42: delete this?

server/index.go
```diff
@@ -0,0 +1,258 @@
...
+
+type indexEntry struct {
+	tpe    pack.BlobType
+	packID backend.ID
+	offset uint
+	length uint
+	old    bool
+}
+
+// NewIndex returns a new index.
```
fw42: wow :)

server/index.go
```diff
@@ -0,0 +1,258 @@
...
+func (idx *Index) Has(id backend.ID) bool {
+	_, _, _, _, err := idx.Lookup(id)
+	if err == nil {
+		return true
+	}
+
+	return false
+}
+
+// Merge loads all items from other into idx.
```
fw42: ... unless they already exist?

PR #157, opened 2015-04-30
Small refactorings and cosmetic changes

fw42:
A few more cosmetic changes... I will comment on each one inline on the diff.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -24,9 +24,11 @@ const (
 	maxConcurrency        = 10
 	maxConcurrencyPreload = 20
 
-	// chunkerBufSize is used in pool.go
 	chunkerBufSize = 512 * chunker.KiB
 )
+
+var archiverAbortOnAllErrors = func(str string, fi os.FileInfo, err error) error { return err }
+var archiverAllowAllFiles = func(string, os.FileInfo) bool { return true }
 
 type Archiver struct {
 	s *server.Server
@@ -37,24 +39,20 @@ type Archiver struct {
 	Filter func(item string, fi os.FileInfo) bool
 }
 
-func NewArchiver(s *server.Server) (*Archiver, error) {
-	var err error
+func NewArchiver(s *server.Server) *Archiver {
 	arch := &Archiver{
 		s:         s,
 		blobToken: make(chan struct{}, maxConcurrentBlobs),
 	}
 
-	// fill blob token
 	for i := 0; i < maxConcurrentBlobs; i++ {
 		arch.blobToken <- struct{}{}
 	}
 
-	// abort on all errors
-	arch.Error = func(string, os.FileInfo, error) error { return err }
-	// allow all files
-	arch.Filter = func(string, os.FileInfo) bool { return true }
+	arch.Error = archiverAbortOnAllErrors
+	arch.Filter = archiverAllowAllFiles
 
-	return arch, nil
+	return arch
 }
 
 func (arch *Archiver) Save(t pack.BlobType, id backend.ID, length uint, rd io.Reader) error {
@@ -78,22 +76,18 @@ func (arch *Archiver) Save(t pack.BlobType, id backend.ID, length uint, rd io.Re
 }
 
 func (arch *Archiver) SaveTreeJSON(item interface{}) (backend.ID, error) {
-	// convert to json
 	data, err := json.Marshal(item)
-	// append newline
-	data = append(data, '\n')
 	if err != nil {
 		return nil, err
 	}
+	data = append(data, '\n')
 
 	// check if tree has been saved before
 	id := backend.Hash(data)
-
 	if arch.s.Index().Has(id) {
 		return id, nil
 	}
 
-	// otherwise save the data
 	return arch.s.SaveJSON(pack.Tree, item)
 }
 
@@ -106,7 +100,7 @@ func (arch *Archiver) SaveFile(p *Progress, node *Node) error {
 		return err
 	}
 
-	// check file again
+	// check file again, since it could have disappeared by now
 	fi, err := file.Stat()
 	if err != nil {
 		return err
@@ -116,14 +110,12 @@ func (arch *Archiver) SaveFile(p *Progress, node *Node) error {
 		e2 := arch.Error(node.path, fi, errors.New("file was updated, using new version"))
 
 		if e2 == nil {
-			// create new node
 			n, err := NodeFromFileInfo(node.path, fi)
 			if err != nil {
 				debug.Log("Archiver.SaveFile", "NodeFromFileInfo returned error for %v: %v", node.path, err)
 				return err
 			}
 
-			// copy node
 			*node = *n
 		}
 	}
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -121,10 +121,10 @@ func archiveDirectory(b testing.TB) {
 	key := SetupKey(b, server, "geheim")
 	server.SetKey(key)
 
-	arch, err := restic.NewArchiver(server)
-	OK(b, err)
+	arch := restic.NewArchiver(server)
 
 	_, id, err := arch.Snapshot(nil, []string{*benchArchiveDirectory}, nil)
+	OK(b, err)
 
 	b.Logf("snapshot archived as %v", id)
 }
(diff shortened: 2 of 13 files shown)


progress.go
```diff
@@ -79,12 +94,17 @@ func (p *Progress) Report(s Stat) {
 	cur := p.cur
 	p.curM.Unlock()
 
-	// update progress
-	if p.OnUpdate != nil {
-		p.fnM.Lock()
-		p.OnUpdate(cur, time.Since(p.start), false)
-		p.fnM.Unlock()
+	p.updateProgress(cur, false)
```
fw42: moved this stuff into a new function since there was some duplication (the almost exact same code existed below again)

progress.go
```diff
@@ -111,40 +126,23 @@ func (p *Progress) reporter() {
...
-	p.curM.Unlock()
-}
-
 // Done closes the progress report.
 func (p *Progress) Done() {
 	if p == nil || !p.running {
 		return
 	}
 
-	if p.running {
```
fw42: this check was unnecessary since you already check for `!p.running` above and then `return`

restorer.go
```diff
@@ -19,57 +19,31 @@ type Restorer struct {
 	Filter func(item string, dstpath string, node *Node) bool
 }
 
-// NewRestorer creates a restorer preloaded with the content from the snapshot snid.
-func NewRestorer(s *server.Server, snid backend.ID) (*Restorer, error) {
-	r := &Restorer{s: s}
+var abortOnAllErrors = func(str string, node *Node, err error) error { return err }
+
+// NewRestorer creates a restorer preloaded with the content from the snapshot id.
+func NewRestorer(s *server.Server, id backend.ID) (*Restorer, error) {
```
fw42: you call this `id` (not `snid`) everywhere else, so I renamed it

restorer.go
```diff
@@ -19,57 +19,31 @@ type Restorer struct {
 	Filter func(item string, dstpath string, node *Node) bool
 }
 
-// NewRestorer creates a restorer preloaded with the content from the snapshot snid.
-func NewRestorer(s *server.Server, snid backend.ID) (*Restorer, error) {
-	r := &Restorer{s: s}
+var abortOnAllErrors = func(str string, node *Node, err error) error { return err }
```
fw42: I think having it in a variable like this is nicer than adding a comment (also there was a bug: since this was defined as a closure, it was referencing the `err` from the surrounding context, not the error that was being passed into the function (it wasn't even named))

restorer.go
```diff
@@ -19,57 +19,31 @@ type Restorer struct {
...
 	}
 
-	// abort on all errors
-	r.Error = func(string, *Node, error) error { return err }
-
 	return r, nil
 }
 
-func (res *Restorer) to(dst string, dir string, treeID backend.ID) error {
+func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error {
```
fw42: This function was quite long. I broke out the core part (the loop body for non-directories) into it's own function (`restoreNodeTo`). I think this increases readability.

snapshot_test.go
```diff
@@ -2,29 +2,17 @@ package restic_test
...
 	. "github.com/restic/restic/test"
 )
 
-func testSnapshot(t *testing.T, s *server.Server) {
-	var err error
-	sn, err := restic.NewSnapshot([]string{"/home/foobar"})
-	OK(t, err)
-	// sn.Tree, err = restic.Blob{ID: backend.ParseID("c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2")}
-	// ok(t, err)
-	sn.Time, err = time.Parse(time.RFC3339Nano, "2014-08-03T17:49:05.378595539+02:00")
```
fw42: I don't know what this is... was this even testing anything? I removed it... also removed the commented lines here.. if it's important that this gets done, we should add a github issue

tree.go
```diff
@@ -74,7 +73,7 @@ func (t *Tree) Insert(node *Node) error {
 	return nil
 }
 
-func (t Tree) find(name string) (int, *Node, error) {
+func (t Tree) binarySearch(name string) (int, *Node, error) {
```
fw42: renamed this for clarity, since it took me a few minutes to realize why this is not super slow

tree.go
```diff
@@ -60,9 +60,8 @@ func (t Tree) Equals(other *Tree) bool {
 }
 
 func (t *Tree) Insert(node *Node) error {
-	pos, _, err := t.find(node.Name)
+	pos, _, err := t.binarySearch(node.Name)
 	if err == nil {
-		// already present
```
fw42: I think the name of the error already says what this is doing, no need to have the comment

tree_test.go
```diff
@@ -21,8 +21,7 @@ var testFiles = []struct {
 	{"bar/bla/blubb", []byte("This is just a test!\n")},
 }
 
-// prepareDir creates a temporary directory and returns it.
```
fw42: again, renaming the function is more concise than having a comment on a function with an ambiguous name, imho.

tree_test.go
```diff
@@ -89,7 +88,7 @@ func TestNodeComparison(t *testing.T) {
 	n2 := *node
 	Assert(t, node.Equals(n2), "nodes aren't equal")
 
-	n2.Size -= 1
+	n2.Size--
```
fw42: golint doesn't like `-= 1` and `+= 1`

walk.go
```diff
@@ -30,15 +30,15 @@ func walkTree(s *server.Server, path string, treeID backend.ID, done chan struct
 		if node.Type == "dir" {
 			walkTree(s, p, node.Subtree, done, jobCh)
 		} else {
-			jobCh <- WalkTreeJob{Path: p, Node: node, Error: err}
```
fw42: `err` is always `nil` at this point, so no need to explicitly pass it

walk_test.go
```diff
@@ -30,11 +30,6 @@ func TestWalkTree(t *testing.T) {
 	// flush server, write all packs
 	OK(t, server.Flush())
 
-	// start benchmark
-	// t.ResetTimer()
-
-	// for i := 0; i < t.N; i++ {
```
fw42: unused code

archiver.go
```diff
@@ -37,24 +39,20 @@ type Archiver struct {
...
 		blobToken: make(chan struct{}, maxConcurrentBlobs),
 	}
 
-	// fill blob token
 	for i := 0; i < maxConcurrentBlobs; i++ {
 		arch.blobToken <- struct{}{}
 	}
 
-	// abort on all errors
-	arch.Error = func(string, os.FileInfo, error) error { return err }
```
fw42: FYI I don't think this would have aborted anything, it would have returned `nil` in every case, even on errors...

fd0: Oh, nice catch!

restorer.go
fd0: Great catch btw!

cache.go
```diff
@@ -90,14 +82,14 @@ func (c *Cache) Load(t backend.Type, subtype string, id backend.ID) (io.ReadClos
 	return os.Open(filename)
 }
 
-func (c *Cache) Purge(t backend.Type, subtype string, id backend.ID) error {
+func (c *Cache) purge(t backend.Type, subtype string, id backend.ID) error {
```
fd0: Why did you change the visibility of `Purge()`?

archiver_test.go
```diff
@@ -121,10 +121,10 @@ func archiveDirectory(b testing.TB) {
 	key := SetupKey(b, server, "geheim")
 	server.SetKey(key)
 
-	arch, err := restic.NewArchiver(server)
-	OK(b, err)
+	arch := restic.NewArchiver(server)
 
 	_, id, err := arch.Snapshot(nil, []string{*benchArchiveDirectory}, nil)
+	OK(b, err)
```
fd0: This is another good catch!

fd0:
Looks good to me. I'll merge it with one exception: I'll dropped the visibility change for `Purge()`.

fw42:
> I'll dropped the visibility change for Purge().

There was no reason for it to be public, it wasn't being called from anywhere (except other methods of the same type).

PR #158, opened 2015-04-30
crypto.go nitpicks

fw42:
Some small changes...
- Refactored duplicated code into a new `poly1305PrepareKey` function
- Remove useless `json:...` annotations on a struct that is never being marshalled
- Rename a few variables and functions for clarity
- Remove some obvious comments and dead code
- Consistently rename "signature" to "MAC" (I think it's important to be consistent so that people don't get confused and I think MAC is more technically correct since "signature" could mean a digital public key signature as well).


crypto/crypto_int_test.go
```diff
@@ -84,19 +85,6 @@ func decodeHex(s string) []byte {
 	return d
 }
 
-// returns true if function called panic
-func shouldPanic(f func()) (didPanic bool) {
```
fw42: Not sure if this is actually dead code, but unless there is some fancy introspection test magic happening, I don't think this is being called anywhere?

crypto/crypto.go
```diff
@@ -258,28 +240,28 @@ func Encrypt(ks *Key, ciphertext, plaintext []byte) ([]byte, error) {
 // Decrypt verifies and decrypts the ciphertext. Ciphertext must be in the form
 // IV || Ciphertext || MAC. plaintext and ciphertext may point to (exactly) the
 // same slice.
-func Decrypt(ks *Key, plaintext, ciphertext []byte) ([]byte, error) {
+func Decrypt(ks *Key, plaintext []byte, ciphertextWithMac []byte) ([]byte, error) {
```
fw42: I found it super confusing that `ciphertext` is being overwritten several times (first with mac and iv, then just with mac, then just ciphertext). So unless that has some performance reasons, I think it's better to be explicit here, since the crypto code is easier to audit the more readable it is.

fw42: might even call this ciphertextWithIVAndMac, although that's maybe getting a bit too verbose

crypto/crypto_int_test.go
fd0: This is dead code. Once `Decrypt()` panicked when the mac couldn't be verified, but that's long gone.

PR #159, opened 2015-04-30
Refactor debug.go

fw42:
The `initDebugLogger` function was pretty big and doing several different unrelated things. This PR breaks that function down into three: `initDebugLogger`, `initDebugTags`, and `initDebugBreaks`, each of which does only one thing.

### DIFF
--- a/debug/debug.go
+++ b/debug/debug.go
@@ -14,101 +14,109 @@ import (
 )
 
 var opts struct {
+	logger *log.Logger
 	tags   map[string]bool
 	breaks map[string]bool
 	m      sync.Mutex
 }
 
-var debugLogger = initDebugLogger()
-
-func initDebugLogger() (lgr *log.Logger) {
-	opts.tags = make(map[string]bool)
-	opts.breaks = make(map[string]bool)
+func init() {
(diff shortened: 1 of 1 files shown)


debug/debug.go
```diff
@@ -14,101 +14,109 @@ import (
 )
 
 var opts struct {
+	logger *log.Logger
 	tags   map[string]bool
 	breaks map[string]bool
 	m      sync.Mutex
 }
 
-var debugLogger = initDebugLogger()
```
fd0: There is a reason that the `debugLogger` variable is not in the `opts` struct
and that it is initialized this way: In Go, variables are initialized first,
then the `init()` functions are called.

Initializing `debugLogger` this way makes sure that it is already available in
`init()` functions of other packages.

Relevant part of the spec is at https://golang.org/ref/spec#Package_initialization

fd0:
This looks good, I'll add a commit that runs the debug init function before the other init() functions. Thanks for your contribution!

fd0:
The commit is 9d29a61861ca0b058144abeb31fc9108b4793624

PR #162, opened 2015-05-01
debug: Add location to log, improve formatting

fd0:
Before:

```
2015/05/02 00:38:46 [restic] main []string{"./restic", "backup", "."}
```

After:

```
2015/05/02 00:47:42 [  restic]   1 main.go:176    main []string{"./restic", "backup", "-f", "."}
```

### DIFF
--- a/debug/debug.go
+++ b/debug/debug.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


debug/debug.go
```diff
@@ -130,12 +153,14 @@ func Log(tag string, f string, args ...interface{}) {
 		f += "\n"
 	}
 
+	formatString := fmt.Sprintf("[% 25s] %-20s %s", tag, getPosition(), f)
```
fw42: instead of 25, you could take the maximum width of all the tags that are specified...

fw42:
One comment, rest looks good :+1:

debug/debug.go
fd0: good idea

PR #163, opened 2015-05-01
replace github.com/juju/arrar by /errors

fd0:


fw42:
What problem does this solve? What does this package do better than the other one? (Why do we need either one?)

fd0:
https://bugsnag.com/blog/go-errors explains it. Basically, when you return an error, you don't get any information on where it was originally encountered. In Go, these errors tend to bubble up until a function decides to handle them (print out, terminate program, etc), but this doesn't give any context where this error occured originally. And for some errors, the message attached to the error is something very generic, like "Permission Denied", without the information what has been accessed.

I think it's a good idea to annotate things to errors so that the is a bit more context. We could do that by using `fmt.Errorf()` so that instead of `return err` we use `return errors.Errorf("unable to open file %s: %v", path, err)`, but this wraps the original error in a way that you cannot get it out again in a higher function, and e.g. compare it with `io.EOF`.

The `juju/errors` package solves this problem, allowing to wrap an error, annotate it with context, and still allow accessing the original error.

PR #164, opened 2015-05-02
Break Archiver.SaveFile() method down into smaller ones

fw42:
Breaks `SaveFile()` down into
- `SaveFile()`
- `reloadFileIfChanged()`
- `saveChunk()`
- `waitForResults()`
- `updateNodeContent()`

I think this increases readability drastically. The old function was doing way too many things on way too many different levels of abstraction and was therefore hard to understand.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/juju/arrar"
 	"github.com/restic/restic/backend"
+	"github.com/restic/restic/chunker"
 	"github.com/restic/restic/debug"
 	"github.com/restic/restic/pack"
 	"github.com/restic/restic/pipe"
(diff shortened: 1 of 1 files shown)


fw42:
GitHub is terrible again and displaying the diff...

Best to just look at the new file to see what's happening now.

fd0:
I'm having a look. Btw: using https://github.com/codegram/vim-codereview together with https://github.com/chrisbra/vim-diff-enhanced makes viewing such changes much more pleasant:
Before (with myers diff algorithm)
-----------------------------------------
 befdore(diff-myers)

After (with patience diff algorithm)
------------------------------------------
after(diff-patience)

The pictures have been stitched together from two screenshots each, and I'm not using Terminus at 8 points ;)

PR #165, opened 2015-05-02
Cleanups

fd0:
This is a collection of a lot of minor cleanups, each in its own commit.


archiver.go
```diff
@@ -28,6 +28,7 @@ const (
 var archiverAbortOnAllErrors = func(str string, fi os.FileInfo, err error) error { return err }
 var archiverAllowAllFiles = func(string, os.FileInfo) bool { return true }
 
+// Archiver is used to backup a set of directories.
```
fw42: Where does the Archiver get that set of directories from? it's stored in the `*server.Server` somehow? Maybe you can say "Archiver is used to back up a set of directories as references by a Server" or something like that?

archiver.go
```diff
@@ -619,7 +623,9 @@ func (j archiveJob) Copy() pipe.Job {
 	return j.new
 }
 
-func (arch *Archiver) Snapshot(p *Progress, paths []string, pid backend.ID) (*Snapshot, backend.ID, error) {
+// Snapshot creates a snapshot of the given paths. If parentID is set, this is
+// used to compare the files with.
```
fw42: "compare the files with" sounds kinda weird. "compare the files to"?

backend/generic.go
```diff
@@ -128,6 +128,8 @@ func (l *blobReader) Close() error {
 	return nil
 }
 
-func LimitReader(f io.ReadCloser, n int64) *blobReader {
-	return &blobReader{f: f, rd: io.LimitReader(f, n)}
+// LimitReadCloser returns a new reader wraps r in an io.LimitReader, but also
```
fw42: "new reader **that** wraps r"?

cache.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/restic/restic/server"
 )
 
+// Cache is used to handle the local cache.
```
fw42: Local cache of what? What's stored in it?

cache.go
```diff
@@ -51,6 +53,7 @@ func (c *Cache) Has(t backend.Type, subtype string, id backend.ID) (bool, error)
 	return true, nil
 }
 
+// Store returns an io.WriteCloser that is used to save new information to.
```
fw42: "new information **to the cache**"?

cache.go
```diff
@@ -73,6 +76,7 @@ func (c *Cache) Store(t backend.Type, subtype string, id backend.ID) (io.WriteCl
 	return file, nil
 }
 
+// Load returns information from the cache.
 func (c *Cache) Load(t backend.Type, subtype string, id backend.ID) (io.ReadCloser, error) {
```
fw42: What about Close()ing of the WriteCloser and ReadCloser that are returned in all those methods? Does the caller have to close them? Is that important? What happens if I don't do that? Maybe add a comment "It's the caller's responsibility to close the returned reader" or something like that (if it's important)?

chunker/chunker.go
```diff
@@ -52,7 +53,7 @@ func (c Chunk) Reader(r io.ReaderAt) io.Reader {
 	return io.NewSectionReader(r, int64(c.Start), int64(c.Length))
 }
 
-// A chunker internally holds everything needed to split content.
+// Chunker split content with Rabin Fingerprints.
```
fw42: s/split/splits/

fw42:
This looks great, much clearer :heart:

archiver.go
fd0: It's given to `Snapshot()`...

Issue #123, opened 2015-04-08
Implement repository configuration file


fd0:
At the moment, the files `version` and `id` are not signed. Before the first
release it should be decided whether a signature is needed for those files too.

Maybe it's a good idea to combine these two into a `config` file that is signed
and encrypted with the master keys and also includes the polynomial used for
the chunker.

fw42:
What's the disadvantage of signing them? (Other than having to write and maintain the code for that :-))

fd0:
It's a change that needs to be done before the alpha release, after that I'd like to have backwards-compatible repositories. It's basically a change in the repository structure and some added code.

fd0:
I propose to add a file named 'config' to the repository containing an encrypted (and authenticated) json structure with the following fields:
 * chunker polynomial
 * repository id
 * repository version

After opening a key with a password, this file is read, the config JSON structure is parsed and the next thing the code does is chenking the version number. This way we can extend the config structure later with additional fields and increment the version field, so that restic can decide if it can handle this version repository or not. Comments?

fw42:
Hmm.. Why encrypt the version number? Seems like that could make it potentially more annoying to make versioned changes to the encryption code (like, version 1 means AES, but now version 2 uses something much better than AES, but you can't know that because you already removed the AES code from that code path).

fw42:
> chunker polynomial (contained redundantly in the key files at the moment, **that's not ideal**)

What's wrong with embedding the polynomial with the keys?

fd0:
There may be more than one key (more than one password), so each one contains the chunker polynomial again. That's redundant. In addition, if an attacker has access to several different repos where the same password is used (but  a different chunker polynomial), she can swap the key-files and restic would (probably) use the other chunker polynomial... I'm not sure how relevant this attack is, though.

fd0:
So, maybe leave the version out? This would make it easy for an attacker to downgrade the repository version...

fw42:
I would MAC the version number, but not encrypt it...

fd0:
I think we should include the repository version number in the encrypted config file.

Suppose we exchange AES for something else (e.g. Chacha20), keeping the MAC (Poly1305).

The relevant scenarios are:

 * Users with a new version of restic accessing an older repository: restic
   tries to authenticate the config file (that works), but decryption fails.
   Afterwards it can try to decrypt with AES, that works again. Display a
   message to the user that the repository uses an older format, with the option
   to upgrade it.

[...]

PR #166, opened 2015-05-03
Add config file to repository

fd0:
This PR tracks my progress on implementing #123, more commits will be added. The first commit just documents what I'm planning to do.


fd0:
This is done and can be tested.

backend/local/local.go
```diff
@@ -36,68 +30,19 @@ func Open(dir string) (*Local, error) {
...
-		return nil, err
-	}
-
-	return &Local{p: dir, ver: version, id: id}, nil
+	return &Local{p: dir}, nil
 }
 
 // Create creates all the necessary files and directories for a new local
-// backend at dir.
+// backend at dir. Afterwards a new config blob should must created.
```
fw42: "should must created"?

backend/local/local.go
```diff
@@ -108,15 +53,10 @@ func Create(dir string) (*Local, error) {
...
 	}
 
-	// test if files already exist
-	_, err := os.Lstat(versionFile)
-	if err == nil {
-		return nil, errors.New("version file already exists")
-	}
-
-	_, err = os.Lstat(idFile)
+	// test if config file already exist
```
fw42: s/exist/exists/

backend/local/local.go
```diff
@@ -265,6 +167,10 @@ func (b *Local) Create() (backend.Blob, error) {
 
 // Construct path for given Type and name.
 func filename(base string, t backend.Type, name string) string {
+	if t == backend.Config {
```
fw42: why is this a special case here and not implemented in the `dirname` function like the other types?

backend/sftp/sftp.go
```diff
@@ -90,64 +84,18 @@ func Open(dir string, program string, args ...string) (*SFTP, error) {
...
-
-	sftp.id = strings.TrimSpace(string(buf))
 	sftp.p = dir
-
 	return sftp, nil
 }
 
 // Create creates all the necessary files and directories for a new sftp
-// backend at dir.
+// backend at dir. Afterwards a new config blob should must created.
```
fw42: same as above, "should must created" sounds weird

backend/sftp/sftp.go
```diff
@@ -389,6 +294,10 @@ func (r *SFTP) Create() (backend.Blob, error) {
 
 // Construct path for given backend.Type and name.
 func (r *SFTP) filename(t backend.Type, name string) string {
```
fw42: hm this function looks identical to the one in `local.go`? the `dirname` function also looks similar.. why is this code duplicated and not shared by both backends?

doc/Design.md
```diff
@@ -21,49 +21,65 @@ been backed up at some point in time. The state here means the content and meta
 data like the name and modification time for the file or the directory and its
 contents.
 
+*Storage ID*: A storage ID is the hash of the content of a file stored in the
+repository. This ID is needed in order to load the file from the repository.
+The storage hash is the SHA-256 hash of the content.
```
fw42: I'm confused now... "Storage ID is the hash of the content" and "storage hash is the hash of the content". What's the difference between "storage ID" and "storage hash"? If it's the same thing, why mention both?

doc/Design.md
```diff
@@ -21,49 +21,65 @@ been backed up at some point in time. The state here means the content and meta
...
-like disk read errors by simply running the program `sha256sum` and comparing
-its output to the file name. If the prefix of a filename is unique amongst all
-the other files in the same directory, the prefix may be used instead of the
-complete filename.
-
-Apart from the files `version`, `id` and the files stored below the `keys`
-directory, all files are encrypted with AES-256 in counter mode (CTR). The
-integrity of the encrypted data is secured by a Poly1305-AES message
-authentication code (sometimes also referred to as a "signature").
+Repositories consists of several directories and a file called `config`. For
```
fw42: s/consists/consist/

server/server.go
```diff
@@ -147,7 +151,7 @@ func (s *Server) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
 
 // LoadJSONEncrypted decrypts the data and afterwards calls json.Unmarshal on
```
fw42: update function name here too

server/server.go
```diff
@@ -525,15 +529,75 @@ func (s *Server) loadIndex(id string) error {
...
+	return err
+}
+
+func (s *Server) loadConfig(cfg *Config) error {
+	err := s.LoadJSONUnpacked(backend.Config, nil, cfg)
+	if err != nil {
+		return err
+	}
+
+	if !cfg.ChunkerPolynomial.Irreducible() {
```
fw42: I don't think this is the right place to check this stuff... other things will fail later anyway, wont they?

server/server.go
```diff
@@ -525,15 +529,75 @@ func (s *Server) loadIndex(id string) error {
...
+		return errors.New("invalid chunker polynomial")
+	}
+
+	return nil
+}
+
+// SearchKey tries to find a key for which the supplied password works,
```
fw42: instead of "for which the supplied password works", maybe say "which can be decrypted using the supplied password"?

fw42: or just "find a key with the supplied password"

server/key.go
```diff
@@ -229,46 +211,6 @@ func AddKey(s *Server, password string, template *Key) (*Key, error) {
 	return newkey, nil
 }
 
-// Encrypt encrypts and authenticates data with the master key. Stored in
-// ciphertext is IV || Ciphertext || MAC. Returns the ciphertext, which is
-// extended if necessary.
-func (k *Key) Encrypt(ciphertext, plaintext []byte) ([]byte, error) {
```
fw42: :+1: to removing those methods

server/server.go
```diff
@@ -147,7 +151,7 @@ func (s *Server) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
 
 // LoadJSONEncrypted decrypts the data and afterwards calls json.Unmarshal on
 // the item.
-func (s *Server) LoadJSONEncrypted(t backend.Type, id backend.ID, item interface{}) error {
+func (s *Server) LoadJSONUnpacked(t backend.Type, id backend.ID, item interface{}) error {
```
fw42: I kinda liked the old name better.. it's not clear to me anymore that the data is encrypted.. I would change it back (or actually I would prefer LoadEncryptedJSON, which is even more clear, imho).

server/server.go
```diff
@@ -525,15 +529,75 @@ func (s *Server) loadIndex(id string) error {
...
+	}
+
+	newID := make([]byte, repositoryIDSize)
+	_, err = io.ReadFull(rand.Reader, newID)
+	if err != nil {
+		return err
+	}
+
+	s.Config.ID = hex.EncodeToString(newID)
+	s.Config.Version = RepositoryVersion
```
fw42: This is never read... If I go and change the version to 0 or 2 or 17, nothing complains.

fw42: I think `loadConfig` should check that `RepositoryVersion` is included in some kind of list of "supported repository versions" or something like that.

server/server.go
```diff
@@ -525,15 +529,75 @@ func (s *Server) loadIndex(id string) error {
...
+	}
+
+	key, err := createMasterKey(s, password)
+	if err != nil {
+		return err
+	}
+
+	s.key = key.master
+	s.keyName = key.Name()
+	return s.createConfig()
```
fw42: This seems super weird to me... Why does the `CreateMasterKey` method create the configuration? Why are those two things "coupled" together? That doesn't seem like something I would expect a function called `CreateMasterKey` to do. I think either `createConfig` should happen somewhere else or you should rename the function (`CreateMasterKeyAndConfig` maybe).

cmd/restic/main.go
```diff
@@ -74,14 +74,13 @@ func (cmd CmdInit) Execute(args []string) error {
 	}
 
 	s := server.NewServer(be)
-
-	_, err = server.CreateKey(s, pw)
+	err = s.CreateMasterKey(pw)
```
fw42: I think you should split the two functions (see [above](https://github.com/restic/restic/pull/166/files#r29563676)) and just add a call to `CreateConfig(key)` here.

backend/local/local.go
fd0: Because `dirname()` is also used to construct glob patterns for listing types and the config file isn't stored in a directory below the repository. So it's just a completely different case.

backend/sftp/sftp.go
fd0: Historical reasons, this should be refactored.

server/server.go
fd0: No, they won't, at least not for the polynomial. The chunker will run with anything, but all the nice properties depend on that the polynomial is irreducible, so this is checked.

server/server.go
fd0: Good catch! At the moment I think it's sufficient to check repository version for the constant `RepositoryVersion`, we'll change that to be a list later on.

server/server.go
fw42: Well then I think this is the wrong place for this check even more. If there is easy ways to use the chunker with a "bad" polynomial, then the chunker itself should do this check, not the thing that is using the chunker (some other thing that will also be using the chunker might forget to add this check).

server/server.go
fd0: These are tightly coupled: The config is encrypted with the master key. I'll rename the function to `Init`, as this describes what the method does.

fw42: I would make this three methods then. CreateMasterKey, CreateConfig, and something like Init (which calls the other two).

server/server.go
fd0: I've just added another benchmark, running this irreducibility check costs 300ms, this would be run for every file, so I think it's too expensive to run it on every `chunker.New()`. Thoughts?

server/server.go
fd0: @fw42 agreed on IRC that this is okay for now. ;)

fd0:
I think this PR is ready to be merged. I'll shortly add another PR which refactors `server.go` (so that the methods have proper names etc.)

fw42:
:tractor: :dash:

PR #168, opened 2015-05-05
initial openbsd support added

ckeen:
This is one commit that adds a mostly verbatim copy of the node_linux.go for OpenBSD. The only difference is the absence of the NOATTIME flag.

With the latest commits the unit tests and the test suite pass. Note that to run on OpenBSD you need the modifications in the openbsd_tests branch. Some of those are portable so they could move upstream until the suite has been rewritten in a proper programming language :see_no_evil:

Thanks for the awesome work on restic!

### DIFF
--- a/node_openbsd.go
+++ b/node_openbsd.go
@@ -1,0 +1,36 @@
+package restic
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+func (node *Node) OpenForReading() (*os.File, error) {
+	file, err := os.OpenFile(node.path, os.O_RDONLY, 0)
+	if os.IsPermission(err) {
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
+	}
+	return file, err
+}
+
+func (node *Node) createDevAt(path string) error {
+	return syscall.Mknod(path, syscall.S_IFBLK|0600, int(node.Device))
+}
+
+func (node *Node) createCharDevAt(path string) error {
+	return syscall.Mknod(path, syscall.S_IFCHR|0600, int(node.Device))
+}
+
+func (node *Node) createFifoAt(path string) error {
+	return syscall.Mkfifo(path, 0600)
+}
+
+func (node *Node) fillTimes(stat *syscall.Stat_t) {
+	node.ChangeTime = time.Unix(stat.Ctim.Unix())
+	node.AccessTime = time.Unix(stat.Atim.Unix())
+}
+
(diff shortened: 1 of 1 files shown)


fd0:
looks good, thanks for your contribution!

node_openbsd.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+func (node *Node) OpenForReading() (*os.File, error) {
+	file, err := os.OpenFile(node.path, os.O_RDONLY, 0)
+	if os.IsPermission(err) {
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
```
fw42: this is kinda useless... If you get permission denied, you just retry the exact same thing with the exact same flags? why would it work the second time? If OpenBSD doesn't have NOATIME, then this function should be simplified to
```golang
func (node *Node) OpenForReading() (*os.File, error) {
  return os.OpenFile(node.path, os.O_RDONLY, 0)
}
```

node_openbsd.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (node *Node) createDevAt(path string) error {
+	return syscall.Mknod(path, syscall.S_IFBLK|0600, int(node.Device))
+}
+
+func (node *Node) createCharDevAt(path string) error {
+	return syscall.Mknod(path, syscall.S_IFCHR|0600, int(node.Device))
+}
+
+func (node *Node) createFifoAt(path string) error {
```
fw42: why does this function exist? it's identical to the one in node.go.

node_openbsd.go
```diff
@@ -0,0 +1,36 @@
...
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
+	}
+	return file, err
+}
+
+func (node *Node) createDevAt(path string) error {
+	return syscall.Mknod(path, syscall.S_IFBLK|0600, int(node.Device))
+}
+
+func (node *Node) createCharDevAt(path string) error {
```
fw42: this and createDevAt are also identical to the generic ones in node.go. Why do we need BSD-specific copies that don't do anything BSD-specific?

fw42:
I don't like the code duplication :-(

fw42:
Ah sorry, my bad, I didn't see https://github.com/restic/restic/commit/1723be5a506a832f364ae257610aae97b2c54bf5. Cool.

fd0:
Heh, I already thought that you'd complain about the code duplication ;)

PR #169, opened 2015-05-05
Add node to build on freebsd.

howeyc:
Also note that inode is 32 bit in go freebsd stat struct. Assumed
to be 64bit in restic.

### DIFF
--- a/node_freebsd.go
+++ b/node_freebsd.go
@@ -1,0 +1,24 @@
+package restic
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+func (node *Node) OpenForReading() (*os.File, error) {
+	file, err := os.OpenFile(node.path, os.O_RDONLY, 0)
+	if os.IsPermission(err) {
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
+	}
+	return file, err
+}
+
+func (node *Node) fillTimes(stat *syscall.Stat_t) {
+	node.ChangeTime = time.Unix(stat.Ctimespec.Unix())
+	node.AccessTime = time.Unix(stat.Atimespec.Unix())
+}
+
+func changeTime(stat *syscall.Stat_t) time.Time {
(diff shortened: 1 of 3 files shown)


howeyc:
Not sure if you want to take as is, as I had to cast inode to 64bit.

Also, I can backup and restore and it seems fine, but output from ls doesn't seem to do anything. I'm planning to take a look, but wondering what I should expect (intuitively I expect I should see a bunch of paths?).

```sh
> restic -r bkup/ list snapshots
cfe024710bcda9ad7af86fc8f48da5e4402157bbeb70242b0b8cf9734e7b83d4                                                                  
> restic -r bkup/ ls cfe02471 /
snapshot of [/xx/xx/xx/xxx] at 2015-05-05 18:01:35.937230606 -0500 CDT:
id not found
```

node_freebsd.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+func (node *Node) OpenForReading() (*os.File, error) {
+	file, err := os.OpenFile(node.path, os.O_RDONLY, 0)
+	if os.IsPermission(err) {
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
```
fw42: same comment here as in https://github.com/restic/restic/pull/168. This retry logic here is useless if it doesn't have NOATIME (if it failed the first time, it will just fail again the second time). Just simplify the function to only make the OpenFile call once.

fw42:
Looks good to me besides the one comment :+1:

howeyc:
Yeah, I didn't even notice that function. I just fixed the others to get it working.

Updated.

fd0:
This failed, but I don't see why. I've restarted the Travis build.

fd0:
This looks good, Only remaining thing is to add FreeBSD to `.travis.yml` (two places).

fd0:
Merged, thanks for your contribution!

PR #170, opened 2015-05-06
Fix ls command

howeyc:
Added missing loadindex call. Also fixed the recursive call to
use the subtree to stop infinte recursion.


fw42:
hu interesting. good catch. @fd0 we should have a regression test for this.

fd0:
Agreed, I'll ad one this evening.

fd0:
Merged, thanks for your contribution!

fd0:
Test was added in c86474663c16cd2fd6b2b88db700682c64e89e53

PR #174, opened 2015-05-06
testsuite: Add test for 'ls' command

fd0:


testsuite/test-backup.sh
```diff
@@ -6,10 +6,17 @@ run restic backup "${BASE}/fake-data"
 run restic restore "$(basename "$RESTIC_REPOSITORY"/snapshots/*)" "${BASE}/fake-data-restore"
 dirdiff "${BASE}/fake-data" "${BASE}/fake-data-restore/fake-data"
 
-SNAPSHOT=$(run restic list snapshots)
+SNAPSHOT=$(restic list snapshots)
```
fw42: why remove the `run`?

testsuite/test-backup.sh
```diff
@@ -6,10 +6,17 @@ run restic backup "${BASE}/fake-data"
...
 run restic backup -p "$SNAPSHOT" "${BASE}/fake-data"
 run restic restore "$(basename "$RESTIC_REPOSITORY"/snapshots/*)" "${BASE}/fake-data-restore-incremental"
 dirdiff "${BASE}/fake-data" "${BASE}/fake-data-restore-incremental/fake-data"
 
+# first line contains snapshot id
+# second line contains timestamp of directory creation
+# so compare from line three on
+echo "snapshot id is $SNAPSHOT"
+restic ls "$SNAPSHOT" | tail -n +3 > "${BASE}/test-ls-output"
```
fw42: delete the "test-ls-output" file after? or just pipe to diff directly instead of generating a temporary file?

fw42: oh or is stuff in $BASE automatically cleaned up or something?

fw42:
Looks good to me :+1:

testsuite/test-backup.sh
fd0: So that `$SNAPSHOT` even contains an id when not in debug mode. Don't ask. :disappointed:

testsuite/test-backup.sh
fd0: Yes, when all tests pass.

PR #176, opened 2015-05-09
Improve CI with coveralls

fd0:


.travis.yml
```diff
@@ -3,7 +3,6 @@ language: go
 go:
   - 1.3.3
   - 1.4.2
-  - release
```
fw42: why remove this?

fd0:
It didn't work and ran the tests with Go 1.4.1, because that's the system-wide version. So I removed it.

Issue #173, opened 2015-05-06
godoc developer documentation


fw42:
https://godoc.org/github.com/restic/restic

Is kinda incomplete. We should add a package description and documentation for every public function and method as well as some examples on how to use the restic package API.

[golint](https://github.com/golang/lint) can be used to generate a list of undocumented (but exported) functions or variables.

fd0:
Good point, thanks for creating this issue. It's not super critical, but valuable.

Maybe we can add just a few words to the top-level package that describes what restic is and referers to the project page at Github, just as a placeholder until the documentation is complete (and this issue is closed).

PR #177, opened 2015-05-09
Add code documentation

fd0:
These commits add doc stubs to all the packages. This partly addresses #173.

### DIFF
--- a/backend/doc.go
+++ b/backend/doc.go
@@ -1,2 +1,2 @@
-// Package backend provides local and remote storage for restic backups.
+// Package backend provides local and remote storage for restic repositories.
 package backend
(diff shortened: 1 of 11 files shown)


fd0:
Ugh, I don't like the comments by coveralls, so I've disabled that. The new test coverage for pull requests can still be seen in the ci-status.

fw42:
> I don't like the comments by coveralls

Is it possible to have comments only when coverage decreases (by x%), but not if it stays the same?

fd0:
No, it's not. Just on or off. We can however configure that the coverall test fails either when the coverage falls below a threshhold or decreases by an amount.

PR #178, opened 2015-05-09
Rename Server -> Repository

fd0:
This pull request contains several commits that rename the struct `Server` to `Repository`. I think this makes sense and increases readability, as the name `Server` wasn't good at all.

The dependency of e.g. `Restorer` can be expressed as follows, (`->` means "depends on"):

```
Restorer -> Repository -> Backend
```

I'm looking forward to comments, @fw42 @ckeen


fw42:
I think the package and directory should be called "repository", not "repo" (since that's what the type is called and I think the convention is to keep those the same?).

Besides that, looks good to me! :+1:

fd0:
Hm, but then let's call it `repo` and `Repo`, that's less typing.

Issue #87, opened 2015-02-06
Implement external test suite in Go


fd0:
At the moment, the external test suite is implemented in shell an only runs under Linux. This issue tracks the progress of implementing a test suite in Go that runs on Linux, darwin and Windows.

fw42:
What's the problem with implementing it as shell scripts (or anything else other than Go)? Is this mostly about cross-platform compatibility?

andrewchambers:
Go tests give code coverage, profiling, race detection and is cross platform.

fd0:
Writing tests in shell is really tedious and error-prone. Even writing the few tests that are there was a pain. The other arguments @andrewchambers pointed out.

I can also live with having the test suite in another language, e.g. Ruby, but not shell.

fd0:
This was fixed with #179.

Issue #171, opened 2015-05-06
Make testsuite more portable


ckeen:
This is related to #87; As the shell scripts are currently, they will not run under OpenBSD (and probably other BSDs as well). There are a couple of things that could be done to make them work under all supported OSs, such as the du changes or the mktemp changes in my [ckeen/openbsd_tests](https://github.com/ckeen/restic/tree/openbsd_tests) branch.

Unfortunately my base system does not ship with the timeout command, so that's still another open issue for me.

As I can see the others could go in to at least improve the situation for my OS a bit. I have not tested whether that breaks anything in MacOS X, as that machine has been occupied so far, so I will update this issue when I get to that.

fd0:
This has been addressed in #179.

PR #179, opened 2015-05-09
Add integration test with the go testing framework

fd0:
This is an example on how integration tests can be run with the Go testing framework. In order to run this test, run the following command in `cmd/restic`:

```
$ go test -v -tags integration -test.datafile ../../testsuite/fake-data.tar.gz
```

Sample output:

```
$ go test -v -tags integration -test.datafile ../../testsuite/fake-data.tar.gz
=== RUN TestBackup-4                                                           
created restic backend 558fe89396 at /tmp/restic-test-358226688/repo
Please note that knowledge of your password is required to access the repository.
Losing your password means that your data is irrecoverably lost.
scan [/tmp/restic-test-358226688/testdata]
scanned 14 directories, 1224 files in 0:00
[0:00] 100.00%  0B/s  19.074 MiB / 19.074 MiB  1239 / 1238 items  0 errors  ETA 0:00 
duration: 0:00, 40.21MiB/s
snapshot b08fe909 saved
--- PASS: TestBackup-4 (2.54s)
    integration_test.go:22: repository initialized at /tmp/restic-test-358226688/repo
    integration_test.go:27: backing up [/tmp/restic-test-358226688/testdata]
PASS
ok      github.com/restic/restic/cmd/restic     2.549s
```

What do you think? Is this the way to go for the integration tests?


cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	. "github.com/restic/restic/test"
+)
+
+var TestDataFile = flag.String("test.datafile", "", `specify tar.gz file with test data to backup and restore (required for integration test)`)
+
+func cmd_init(t testing.TB) {
```
fw42: cmdInit

cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+	cmd := &CmdInit{}
+	OK(t, cmd.Execute(nil))
+
+	t.Logf("repository initialized at %v", opts.Repo)
+}
+
+func cmd_backup(t testing.TB, target []string) {
```
fw42: cmdBackup

cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+		OK(t, os.RemoveAll(tempdir))
+	}()
+
+	// use cache dir within tempdir
+	OK(t, os.Setenv("RESTIC_CACHE", filepath.Join(tempdir, "cache")))
+
+	// configure environment
+	opts.Repo = filepath.Join(tempdir, "repo")
+	OK(t, os.Setenv("RESTIC_PASSWORD", *TestPassword))
+	cmd_init(t)
```
fw42: seems like everything in this function above this line doesn't really belong into this test since it's very generic, not specific to what is being tested here, and we will do the same thing for lots of other tests... maybe we can hide this behind some kind of setup function?

cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+
+	// extract data file
+	datadir := filepath.Join(tempdir, "testdata")
+	err = system("sh", "-c", `mkdir "$1" && (cd "$1" && tar xz) < "$2"`,
```
fw42: can you put this into a helper function with a descriptive name like `setupTarTestFixture(datadir, *TestDataFile)` or something like that?

cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+		"sh", datadir, *TestDataFile)
+	OK(t, err)
+
+	cmd_backup(t, []string{datadir})
```
fw42: Do we not need to assert that this actually worked? This test so far only gives us confidence that cmd_init and cmd_backup don't panic, but we don't know that they do what we expect them to.

cmd/restic/integration_test.go
```diff
@@ -1,3 +1,72 @@
...
+func cmd_backup(t testing.TB, target []string) {
+	cmd := &CmdBackup{}
+	t.Logf("backing up %v", target)
+
+	OK(t, cmd.Execute(target))
+}
+
+func system(command string, args ...string) error {
+	cmd := exec.Command(command, args...)
+	cmd.Stdout = os.Stdout
```
fw42: I think we can live without this line. It's very verbose and I don't think it adds much value (unless you are debugging something). I would set this to /dev/null for now (maybe unless you explicitly set a debug flag).

fw42:
:+1: I like the idea (testing the cmd things in Go, not in sh) in general. I think we need to discipline ourselves to not use `system()` too much unless it's really necessary (and abstract it away so it's invisible and discouraged), otherwise we will just end up with a Go test that looks like a shell script :-)

fd0:
Sure, i agree. I just needed a quick way to test this approach. Looks good to me.

cmd/restic/integration_test.go
fd0: Uhm, this isn't the line that causes restic to output a comprehensive progress report, since the archiver isn't started with `system()`.

fw42: Right, sorry, my bad. Maybe we should stop using `fmt` and instead add something that implements the logger interface, so that in test, we can inject a null logger. I can work on that if you want.

fd0: I've just pushed some more commits, have a look at the `verbosePrintf()` function in `main.go`. We should also make this functionality available to the user, I've added `--quiet` for that.

fw42:
For later reference: This PR is related to / part of https://github.com/restic/restic/issues/171.

test/backend.go
```diff
@@ -25,10 +24,6 @@ func SetupBackend(t testing.TB) *server.Server {
 	b, err := local.Create(filepath.Join(tempdir, "repo"))
 	OK(t, err)
 
-	// set cache dir below temp dir
-	err = os.Setenv("RESTIC_CACHE", filepath.Join(tempdir, "cache"))
```
fw42: is this not needed anymore?

fd0: Removed in the next chunk of commits.

fd0:
I've just rebased the commits to master. Once #194 is merged, the messages about the 'orphaned' data blobs will vanish. I'll refactore the other integration tests in a bit.

fd0:
I've split out three minor refactorings into #196.

fd0:
This is finally done, thanks for all your patience ;)

ceh:
Closes #87 and #171?

fd0:
Indeed, thanks for pointing it out!

Issue #38, opened 2014-11-16
Vendor code


fd0:
Use godep https://github.com/tools/godep to include all third-party library code into the khepri repository.

fd0:
There is currently a discussion about a new built tool called `gb`, which solves this problem more elegantly than to vendor code and rewrite import paths: https://github.com/constabulary/gb

Slides: http://go-talks.appspot.com/github.com/davecheney/presentations/reproducible-builds.slide

fd0:
I need to get an overview over existing tools and how they work before I'm able to decide this.

From the top of my head, I have the following requirements:
 * Source code should be contained in a subdir of this repository, so that cloning it gives us all the code needed to build restic (apart from the stdlib)
 * No import path rewriting, that's ugly
 * Ideally, restic is still 'go-gettable', although this then doesn't use the vendored dependencies (but that's okay)
 * Ideally, users who would like to compile restic shouldn't need to be bothered with setting up `$GOPATH` etc, just clone the repository, run a command, and that's it.

There is also `wgo`, which might be a solution: https://github.com/skelterjohn/wgo

fd0:
This is addressed in #197. We're still using Godeps for vendoring, but that can easily be exchanged for something else. End users only need to run `go run build.go`, without having to set up a GOPATH. So I consider this issue closed once #197 is merged.

PR #181, opened 2015-05-10
Vendor using godep, allow compiling from checkout

fd0:
The commits in this PR allow compiling and running all the tests etc. from a checkout of the repository:

```
$ make restic
mkdir -p .gopath/src/github.com/restic
ln -sf ../../../.. .gopath/src/github.com/restic/restic
cd /home/fd0/tmp/restic/.gopath/src/github.com/restic/restic && \
    go build -a -ldflags "-s" -o restic ./cmd/restic

$ make restic.debug
cd /home/fd0/tmp/restic/.gopath/src/github.com/restic/restic && \
    go build -a -tags debug -o restic ./cmd/restic

$ make test
cd /home/fd0/tmp/restic/.gopath/src/github.com/restic/restic && \
    go test -v ./...
[...]

$ make test-integration
[...]
```

Travis now also uses this, not yet integrated is the old testsuite.

The basic idea is taken from docker: Create a temporary directory for gopath at `.gopath`, and create a symlink at the correct location to the current repository. Together with the vendored dependencies, this makes up the `GOPATH`.

If this works, this addresses #38 and makes it really easy to compile restic from a checkout ouf the repository simply by calling `make`. This should then also be addressen in `README.md`.

In this branch, the only requirements for compiling restic are bash, make and Go, no need to setup a `GOPATH` or so. This is a step forward.


fw42:
Why are the integration tests a separate make target and not part of the regular test run?

Makefile
```diff
@@ -1,15 +1,55 @@
...
+GOX_OS ?= linux darwin openbsd freebsd
+SFTP_PATH ?= /usr/lib/ssh/sftp-server
+
+export GOPATH GOX_OS
+
+all: restic
+
+.gopath:
+	mkdir -p .gopath/src/github.com/restic
+	ln -sf ../../../.. .gopath/src/github.com/restic/restic
```
fw42: hm I don't understand this part. Why `../../../..`?

Makefile
```diff
@@ -1,15 +1,55 @@
...
+	rm -rf .gopath restic *.cov restic_*
 	go clean ./...
+
+test: .gopath
+	cd $(BASEPATH) && \
+		go test $(GOTESTFLAGS) ./...
+
+bench: .gopath
+	cd $(BASEPATH) && \
+		go test GOTESTFLAGS) bench ./...
```
fw42: looks like a missing `$(` here?

Makefile
fd0: This is the structure after `.gopath` has been created:
```
$ tree .gopath 
.gopath
└── src
    └── github.com
        └── restic
            └── restic -> ../../../..
```

And `GOPATH` is set to the following value, this can be seen with `make env`:
```
$ make env
export GOPATH="/home/fd0/tmp/restic/.gopath:/home/fd0/tmp/restic/Godeps/_workspace"
```

So when the go tool is run and encounters the path `github.com/restic/restic/chunker`, the following path is constructed from the first element of `GOPATH`:
```
/home/fd0/tmp/restic/.gopath/src/github.com/restic/restic/chunker
```

Since `restic` is a symlink, it is resolved to:
```
/home/fd0/tmp/restic/.gopath/src/github.com/restic/../../../../chunker
```

Which is the same as:
```
/home/fd0/tmp/restic/chunker
```

That's why it is `../../../..`. Tadaa!

Makefile
fd0: You're right, good catch!

fd0:
"integration test" at this point just means the test of the sftp backend, this needs an sftp server binary (and the complete path to it), which isn't always available.

Makefile
fw42: thanks for the explanation :)

Issue #8, opened 2014-11-15
Storage Backend: Amazon S3


scoddy:
Implement Storage Backend Amazon S3

fd0:
This has been added when #184 was merged.

PR #184, opened 2015-05-12
Add Amazon S3 backend.

howeyc:
Currently it is slow to backup as only one connection to s3 at a time. I did this because I had tls timeouts when attempting a bunch of connections at a time, the lock is in Finalize().

Also, need 2 environment variables to access to Amazon AWS:
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY

The s3 url supported is of the form:
s3://<region name>/<bucketname>
ie s3://us-west-1/backupbucket

Also, the bucket must already exist in s3. I did not look into how (or if it's even possible) to create a bucket using the library.


fw42:
How many concurrent connections did you try before you ran into timeouts?

howeyc:
My first attempt was without limit and I got timeouts. I tried 5 and no
timeout issues, but it didn't transfer any faster for me. So I stopped
trying and just went back to one since it was simple to do.

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+var ErrWrongData = errors.New("wrong data returned by backend, checksum does not match")
+
+func s3path(t backend.Type, name string) string {
+	if t == backend.Config {
+		return string(t)
+	}
+	return string(t) + "/" + name
+}
+
+type S3DB struct {
```
fw42: I think it's idiomatic to name the type according to the package name and filename? So maybe call this just `S3`?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"strings"
+	"sync"
+
+	"github.com/mitchellh/goamz/aws"
+	as3 "github.com/mitchellh/goamz/s3"
```
fw42: what does the "a" stand for?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+func (bb *s3Blob) Finalize(t backend.Type, name string) error {
+	if bb.final {
+		return errors.New("Already finalized")
+	}
+
+	bb.final = true
+
+	path := s3path(t, name)
+	bb.b.mput.Lock()
+	err := bb.b.bucket.Put(path, bb.buf.Bytes(), "binary/octet-stream", "private")
```
fw42: Hm not sure I understand why this needs to be mutexed. Can you explain?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+	return &blob, nil
+}
+
+func (b *S3DB) get(t backend.Type, name string) (*s3Blob, error) {
+	blob := &s3Blob{
+		b: b,
+	}
+
+	path := s3path(t, name)
+	data, err := b.bucket.Get(path)
```
fw42: ... or rather: why does this not need to be mutexed? Why only the `Put()`?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+	blob.buf = bytes.NewBuffer(data)
+	return blob, err
+}
+
+// Get returns a reader that yields the content stored under the given
+// name. The reader should be closed after draining it.
```
fw42: Are those comments copied from local/sftp? We should add those generic docs to the interface, not to the implementation, imho. This should only document S3-specific stuff I think. @fd0 thoughts?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+
+// List returns a channel that yields all names of blobs of type t. A
+// goroutine is started for this. If the channel done is closed, sending
+// stops.
+func (b *S3DB) List(t backend.Type, done <-chan struct{}) <-chan string {
+	ch := make(chan string)
+
+	prefix := string(t) + "/"
+
+	listresp, err := b.bucket.List(prefix, "/", "", 10000)
```
fw42: for clarity: can we call this "response", not "listresp"?

fw42: what's the magic 10000? Can we put that in a constant with a descriptive name?

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+	}()
+
+	return ch
+}
+
+// Delete removes the repository and all files.
+func (b *S3DB) Delete() error {
+	return b.bucket.DelBucket()
```
fw42: Is deleting a bucket equivalent to deleting all files in the bucket? Sounds like something more destructive (emptying the bucket vs. deleting the bucket itself).

backend/s3/s3.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+	"sync"
+
+	"github.com/mitchellh/goamz/aws"
+	as3 "github.com/mitchellh/goamz/s3"
+
+	"github.com/restic/restic/backend"
+)
+
+var ErrWrongData = errors.New("wrong data returned by backend, checksum does not match")
```
fw42: this isn't used anywhere, is it?

fw42:
Looks very promising so far.

goamz has server [mocks](https://github.com/mitchellh/goamz/blob/master/s3/s3test/server.go) that we can use to write some unit tests.

fw42:
>  I did not look into how (or if it's even possible) to create a bucket using the library.

Looks like there is a [PutBucket](https://github.com/mitchellh/goamz/blob/0391bda5052267a61c7ba3ee358c970dbc5f6790/s3/s3.go#L136-L139) method for that.

backend/s3/s3.go
fd0: I thinkt that's the implementation for "use just one connection". For using an arbitrary number, have a look at the `blobToken` channel in `archiver.go`. Basically, it's a channel with a limited buffer capacity which is filled with token (just an empty struct, `struct{}`). The code can then receive one token from the channel, which blocks if the channel is empty (aka: there are no token available at the moment), and put the token back once the goroutine is done.

backend/s3/s3.go
fd0: Absolutely. I'll start refactoring this once the integration tests are done (I expect tomorrow, since that's a bank holiday in Germany).

fd0:
Thanks for your contribution! I'll have a closer look this evening. The integration tests failed btw because the dependency isn't vendored. It looks like you have your own `GOPATH` set up, you need to run `godep save` and commit the additional files. You can test if it works by running `make test`, that sets up `GOPATH` so that only the vendored depedencies are used.

fw42:
Fixes https://github.com/restic/restic/issues/8.

backend/s3/s3.go
howeyc: This was just a little internal weirdness to me to use a package s3 while writing package s3. I'll take it off.

backend/s3/s3.go
howeyc: @fd0 is right, this is the "one connection" deal I mentioned.

backend/s3/s3.go
howeyc: good point, I didn't have any tls timeouts so I didn't bother with a mutex here.

backend/s3/s3.go
howeyc: the type returned from the s3 library is a list response, so that's the name I used for the variable. As to the 10000, it's just a number i pulled out of the air. You need to pass how many keys can be returned with each "List" request.

Now that you mention it, I didn't handle a case where the list is too long for one request and you need to request again with a parameter changed to "continue" from the last element in the response.

fd0:
@howeyc did you see the announcement of the official Go aws sdk? https://github.com/aws/aws-sdk-go

fd0:
I've had a look at the s3 backend, and here are my comments:
 * At the moment this is limited to one connection due to the `mput` mutex in the `S3` struct. This should be changed to a constant of about 10 or so (make this configurable?), by using a channel (see comment above).
 * In `Finalize()`, checking if a key already exists and writing to it afterwards is racy. What happens when data is written to an already existing key?
 * The implementation buffers all data written to S3 in memory (isn't there a function to rename a key?), that's okay from my point of view, but may lead to massive memory usage. We should reuse the buffers that are used here, similar to the buffer pool in the crypto functions (see `DecryptFrom()` in `crypto/reader.go` and `crypto/buffer_pool.go`).
 * For reading data, the implementation always reads the data into memory and returns a `bytes.Buffer`. It would be better to use the `bucket.GetReader()` method to obtain an `io.Reader` and pass that one on, because when data is decrypted, it is also read and buffered (to check the MAC), so the same data ends up in memory twice.
 * Can we add a custom prefix (`restic`) to all keys, and remove only the keys with that prefix on `Delete()`? I find deleting the complete bucket too invasive, we don't know what other things the user put in there.

Some small style nitpicks:
 * Use a coherent naming style for the receiver of methods. `b` is not a good name for a pointer to `S3` (it's even used again in the blob struct), maybe rename the `S3` struct to `Backend` (or `S3Backend` and use `backend` instead of `b` as the receiver. In addition, you can change `Open()` to return `backend.Backend`, so that the name of your struct is hidden outside of that package. Naming things is hard :)
 * The comment for `Delete()` talks about files, that's incorrect.
 * Can you please rebase onto the current master branch?

Thanks a lot for your contribution!

howeyc:
I don't have an answer for you Finalize() question. Not sure what happens there.

Still working on your comments. I uploaded what I've done so far.

howeyc:
I think I have handled the comments you made.

Not sure what's up with the tests failing. Something about goveralls api key I think.

fd0:
The failing test is indeed caused by coveralls.io, the remaining tests pass!

PR #186, opened 2015-05-14
Fix ModTime for directories

### LINKED ISSUE #24: Correctly restore timestamps
Some timestamps for directories are not restored, this should be corrected.

fw42:
Fixes https://github.com/restic/restic/issues/24.

But a similar issue exists with symlinks. It seems that it's currently not possible to call the [lutimes](http://linux.die.net/man/3/lutimes) system call to change symlink timestamps (without dereferencing the link). I'm gonna open a separate GitHub issue for that after this PR is merged.


fw42:
A fix for this has been merged upstream 8 days ago (https://go-review.googlesource.com/#/c/9379/).

fw42:
Looks like I made coverails unhappy :-/

I don't think it's worth it to add a unit test for this new function. The code has just as much test coverage as before via the shell scripts.

fd0:
I've set the failure threshhold to 5%, I hope that this is the right value. The meaning of the fields in the coveralls settings panel aren't explained anywhere.

This should also be gone once #179 is done, afterwards the integration tests count towards the coverage.

fw42:
Does this solution make sense to you? Iterate over the directories a second time at the end. Or is that too inefficient?

fd0:
Looks good to me. We could also fix the timestamps for dirs right after all nodes have been restored to, before returning from `restoreTo()`, what do you think?

fw42:
Hm not sure what you mean by that. We need to do this **after** all non-dirs have already been restored.

fd0:
Yeah, I know, but we can do this directly after all the nodes in a dir have been created, no need to iterate a second time.

fw42:
You mean right after [this](https://github.com/restic/restic/blob/1fb8e332ce031e92a210bab5ac6e00b720eac0c9/restorer.go#L57) returns?

fw42:
oh, you already did that.

fd0:
Exactly. ;)

fw42:
works for me :rocket:

fd0:
since travis and coveralls are happy, I'm merging this.

PR #191, opened 2015-05-14
Fix restoring symlink timestamps for linux

fd0:
This adds a test and fixes restoring symlink timestamps for Linux. We need to figure out how to do this on darwin and *bsd.

### DIFF
--- a/node.go
+++ b/node.go
@@ -171,7 +171,7 @@ func (node Node) RestoreTimestamps(path string) error {
 
 	if node.Type == "symlink" {
 		if err := node.restoreSymlinkTimestamps(path, utimes); err != nil {
-			return errors.Annotate(err, "UtimesNano")
+			return err
 		}
 
 		return nil
--- a/node.go
+++ b/node.go
@@ -146,13 +146,11 @@ func (node Node) restoreMetadata(path string) error {
 		return errors.Annotate(err, "Lchown")
 	}
 
-	if node.Type == "symlink" {
-		return nil
-	}
-
-	err = os.Chmod(path, node.Mode)
-	if err != nil {
-		return errors.Annotate(err, "Chmod")
+	if node.Type != "symlink" {
+		err = os.Chmod(path, node.Mode)
+		if err != nil {
+			return errors.Annotate(err, "Chmod")
+		}
 	}
 
 	if node.Type != "dir" {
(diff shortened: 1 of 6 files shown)


node_linux.go
```diff
@@ -22,3 +24,36 @@ func (node *Node) fillTimes(stat *syscall.Stat_t) {
...
+
+const AT_SYMLINK_NOFOLLOW = 0x100
+
+func utimensat(dirfd int, path string, times *[2]syscall.Timespec, flags int) (err error) {
+	var _p0 *byte
+	_p0, err = syscall.BytePtrFromString(path)
+	if err != nil {
+		return
+	}
+	_, _, e1 := syscall.Syscall6(syscall.SYS_UTIMENSAT, uintptr(dirfd), uintptr(unsafe.Pointer(_p0)), uintptr(unsafe.Pointer(times)), uintptr(flags), 0, 0)
```
fw42: this looks scary. do we really want to do this?

fd0: I'm not sure yet. It's the same they do in the stdlib for Go 1.5...

node.go
```diff
@@ -166,12 +164,20 @@ func (node Node) restoreMetadata(path string) error {
...
-	var utimes = []syscall.Timespec{
+	var utimes = [...]syscall.Timespec{
 		syscall.NsecToTimespec(node.AccessTime.UnixNano()),
 		syscall.NsecToTimespec(node.ModTime.UnixNano()),
 	}
 
-	if err := syscall.UtimesNano(path, utimes); err != nil {
+	if node.Type == "symlink" {
+		if err := node.restoreSymlinkTimestamps(path, utimes); err != nil {
+			return errors.Annotate(err, "UtimesNano")
```
fw42: I feel like this annotation should happen in `restoreSymlinkTimestamps`, not here

node_linux.go
fw42: why do you need this `e1`? why not overwrite the other `err`?

node_linux.go
```diff
@@ -22,3 +24,36 @@ func (node *Node) fillTimes(stat *syscall.Stat_t) {
...
+	if e1 != 0 {
+		err = e1
+	}
+	return
+}
+
+//sys	utimensat(dirfd int, path string, times *[2]Timespec, flags int) (err error)
```
fw42: I don't think we need this comment

fw42:
I guess this works, but I'm not a big fan. I hope we can get rid of this soon and use something from the standard library.

node_linux.go
fd0: I've copied that from the generated documentation, but yeah, isn't necessary.

node.go
fd0: I agree.

fd0:
Yeah, me neither. I've fixed the error annotation and will merge this once travis agrees.

PR #192, opened 2015-05-14
Fix restic on i386

fd0:
This fixes restic for Linux/i386 and instructs travis to run the tests also on i386.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -71,10 +71,10 @@ func (sn Snapshot) ID() backend.ID {
 func (sn *Snapshot) fillUserInfo() error {
 	usr, err := user.Current()
 	if err != nil {
-		return err
+		return nil
 	}
-
 	sn.Username = usr.Username
+
 	uid, err := strconv.ParseInt(usr.Uid, 10, 32)
 	if err != nil {
 		return err
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,9 @@ script:
   - make restic
   - make gox
   - make test
+  - GOARCH=386 make test
   - make test-integration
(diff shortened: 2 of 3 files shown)


fw42:
I don't understand what the problem is and I don't understand why this PR fixes it. Can you explain?

fw42:
is this about the id being uint32 and not uint64 or something like that?

snapshot.go
```diff
@@ -71,10 +71,10 @@ func (sn Snapshot) ID() backend.ID {
 func (sn *Snapshot) fillUserInfo() error {
 	usr, err := user.Current()
 	if err != nil {
-		return err
+		return nil
```
fw42: I don't understand this change

fd0:
Oh, I'm sorry: I discovered that restic stopped working on Linux/i386, because several functions from the stdlib weren't implemented on that platform (e.g `user.Current()` and `user.LookupId()`). So the purpose of this PR is twofold: Add tests on i386 to travis, and make most cases with these lookup-functions not fail hard.

snapshot.go
fd0: If the current user cannot be determined, do not fill in any user info. This information isn't  required for restic to function (it's just an information after all, nothing that you'd really need to restore a backup), so don't fail hard when `user.Current()` returns an error.

fd0:
Ah, I asked about these not-implemented functions (because it bothered me a lot), turns out, this functionality needs CGO (because it uses the libc functions), and that's disabled when cross-compiling: https://github.com/golang/go/issues/10868

fw42:
Ok I get it now. Weird. But ok. :rocket:

PR #194, opened 2015-05-17
Fix fsck orphaned check, add debug command 'dump', cleanup

fd0:
These commits add a command 'dump' that's available in debug builds. 'dump'
writes internal data structes pretty-printed as JSON to stdout for (manual)
inspection.

In contrast to the 'cat' command, this one prints the data structures as
there are interpreted by restic, not as they are stored in the repository.
This means that only the merged index from all the index files is printed
out.

The first few commits are just little cleanup.


fw42:
What was the bug with fsck?

fd0:
Since 29ead95c96536325a90cf2d8dc8d63ae03eba6b0, `restic fsck --orphane` reported every blob as being orphaned, that's wrong obviously. I noticed this problem during development for #179.

PR #205, opened 2015-06-28
backend: fix doc typos

ceh:


fd0:
Thanks for your time and work!

PR #209, opened 2015-07-02
Force sudo: false to run on containers

badboy:


fw42:
:+1:

fd0:
Ok, this could significantly speed up Travis jobs...

PR #216, opened 2015-07-06
Command `snapshots`: do not list first dir twice

### LINKED ISSUE #215: First directory is listed twice
Currently, `restic snapshots` lists the first directory twice if more than one directory has been backed up:
```
$ restic snapshots
ID        Date                 Source      Directory
9d35d148  2015-07-05 11:10:44  foo        /home/fd0/shared
e4d7e9f7  2015-07-05 12:27:54  foo        /home/fd0/shared
                                          /home/fd0/shared
                                          /home/fd0/work
8ebac95a  2015-07-05 23:38:32  foo        /home/fd0/shared
```

fd0:
This fixes #215.

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -145,7 +145,7 @@ func (cmd CmdSnapshots) Execute(args []string) error {
 		tab.Rows = append(tab.Rows, []interface{}{sn.ID()[:plen/2], sn.Time.Format(TimeFormat), sn.Hostname, sn.Paths[0]})
 
 		if len(sn.Paths) > 1 {
-			for _, path := range sn.Paths {
+			for _, path := range sn.Paths[1:] {
 				tab.Rows = append(tab.Rows, []interface{}{"", "", "", path})
 			}
 		}


cmd/restic/cmd_snapshots.go
```diff
@@ -145,7 +145,7 @@ func (cmd CmdSnapshots) Execute(args []string) error {
 		tab.Rows = append(tab.Rows, []interface{}{sn.ID()[:plen/2], sn.Time.Format(TimeFormat), sn.Hostname, sn.Paths[0]})
 
 		if len(sn.Paths) > 1 {
-			for _, path := range sn.Paths {
+			for _, path := range sn.Paths[1:] {
```
fw42: why is it in the list twice to begin with?

fw42: seems like a bandaid solution without fixing the root cause...

fd0: The first line for each snapshot contains all the other fields (ID, Date etc), the following lines are just added if there is more than one dir. So this indeed fixes the root cause.

PR #217, opened 2015-07-06
cmd_backup: filter non-existing target files/dirs

### LINKED ISSUE #128: Snapshot is created even if backup source path doesn't exist
Trying to backup a directory ("foobar") that does not exist: 

```
$ cmd/restic/restic -r /tmp/restic backup foobar
Enter Password for Repository:
scan [/Users/flo/go/src/github.com/restic/restic/foobar]
error for /Users/flo/go/src/github.com/restic/restic/foobar: lstat /Users/flo/go/src/github.com/restic/restic/foobar: no such file or directory
scanned 0 directories, 0 files in 0:00
loading blobs
[0:00] NaN%  0B/s  0B / 0B  1 / 0 items  ETA 0:00
duration: 0:00, NaNMiB/s
snapshot b319c02e saved
```
gives an error, but still creates an empty snapshot.

Is that on purpose or a bug?

fd0:
This fixes #128.

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


fw42:
This triggers one additional system call (stat) for every single file. That sounds potentially slow. Isn't there a better way? I think there is a syscall that gives you all stats at once in one go ([getdents](http://man7.org/linux/man-pages/man2/getdents.2.html) maybe?).

fd0:
This only triggers an additional `stat` syscall for each file listed on the command line as an argument for `restic backup`, nothing more. Especially there are no additional syscalls for all entries below.

PR #218, opened 2015-07-06
Fix restore filter

### LINKED ISSUE #202: Restore filter is broken
The command `restore` accepts a pattern of files that should be restored. At the moment, restic doesn't seem to use it at all, fix that.

fd0:
Internally rename restorer.Filter -> restorer.SelectForRestore to make
semantic clear.

In addition, swap parameters to filepath.Match() so that the pattern can
really be matched.

Limitation: The filter only works on the filename, not on any path
component, e.g. `*.go` selects all go files, `subdir/foo*` doesn't
select anything.

This fixes #202.

### DIFF
--- a/restorer.go
+++ b/restorer.go
@@ -26,7 +26,10 @@ var restorerAbortOnAllErrors = func(str string, node *Node, err error) error { r
 
 // NewRestorer creates a restorer preloaded with the content from the snapshot id.
 func NewRestorer(repo *repository.Repository, id backend.ID) (*Restorer, error) {
-	r := &Restorer{repo: repo, Error: restorerAbortOnAllErrors}
+	r := &Restorer{
+		repo: repo, Error: restorerAbortOnAllErrors,
+		SelectForRestore: func(string, string, *Node) bool { return true },
+	}
 
 	var err error
 
@@ -45,14 +48,9 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error 
 	}
 
 	for _, node := range tree.Nodes {
-		selectedForRestore := true
-		if res.SelectForRestore != nil {
-			debug.Log("Restorer.restoreTo", "running select filter for %v", node.Name)
-
-			selectedForRestore = res.SelectForRestore(filepath.Join(dir, node.Name),
-				filepath.Join(dst, dir, node.Name), node)
-			debug.Log("Restorer.restoreNodeTo", "SelectForRestore returned %v", selectedForRestore)
-		}
+		selectedForRestore := res.SelectForRestore(filepath.Join(dir, node.Name),
+			filepath.Join(dst, dir, node.Name), node)
+		debug.Log("Restorer.restoreNodeTo", "SelectForRestore returned %v", selectedForRestore)
 
 		if selectedForRestore {
 			err := res.restoreNodeTo(node, dir, dst)
--- a/restorer.go
+++ b/restorer.go
@@ -45,9 +45,20 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error 
 	}
 
 	for _, node := range tree.Nodes {
-		excluded, err := res.restoreNodeTo(node, dir, dst)
-		if err != nil {
-			return err
+		selectedForRestore := true
+		if res.SelectForRestore != nil {
+			debug.Log("Restorer.restoreTo", "running select filter for %v", node.Name)
+
+			selectedForRestore = res.SelectForRestore(filepath.Join(dir, node.Name),
+				filepath.Join(dst, dir, node.Name), node)
+			debug.Log("Restorer.restoreNodeTo", "SelectForRestore returned %v", selectedForRestore)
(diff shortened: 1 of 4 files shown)


restorer.go
```diff
@@ -73,19 +77,30 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error
...
+func (res *Restorer) restoreNodeTo(node *Node, dir string, dst string) (excluded bool, err error) {
+	debug.Log("Restorer.restoreNodeTo", "node %v, dir %v, dst %v", node.Name, dir, dst)
 	dstPath := filepath.Join(dst, dir, node.Name)
 
-	if res.Filter != nil && res.Filter(filepath.Join(dir, node.Name), dstPath, node) {
-		return nil
+	if res.SelectForRestore != nil {
+		debug.Log("Restorer.restoreNodeTo", "running select filter for %v", node.Name)
+
+		if !res.SelectForRestore(filepath.Join(dir, node.Name), dstPath, node) {
```
fw42: I kinda feel like the filtering should be up to the caller and not happen in `restoreNodeTo`. Feels like this is not this method's responsibility.

restorer.go
```diff
@@ -44,7 +45,8 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error
 	}
 
 	for _, node := range tree.Nodes {
-		if err := res.restoreNodeTo(node, dir, dst); err != nil {
+		excluded, err := res.restoreNodeTo(node, dir, dst)
```
fw42: I think maybe this is where the filtering should happen and then `continue` the loop before even calling `restoreNodeTo`. Thoughts?

fw42:
> Limitation: The filter only works on the filename

Why is that?

fw42:
not a fan of all this `debug.Log` clutter :-( is that really that helpful?

fd0:
The limitation comes from using `filepath.Match`, which doesn't handle separator chars `/` well, see https://godoc.org/path/filepath#Match

The `debug.Log` stuff helps a lot, in my opinion. This is how I debugged it, and I think it's good to leave it in. This allows users (who are able to recompile restic with `-tags debug`) to create a really verbose logfile of all the internal things restic does, so we can remotely figure out what went wrong.

Do you have a better idea?

restorer.go
fd0: I think you're right.

restorer.go
fd0: Agreed.

fd0:
I've moved the filter selection to `restoreTo` as you suggested. We need a test for this functionality, afterwards the PR can be merged I think. I can live with the file name filter limitation for now, and I'll create a new issue for it so we can track the progress.

Any other thoughts?

restorer.go
```diff
@@ -44,8 +45,20 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error
 	}
 
 	for _, node := range tree.Nodes {
-		if err := res.restoreNodeTo(node, dir, dst); err != nil {
-			return err
+		selectedForRestore := true
+		if res.SelectForRestore != nil {
```
fw42: maybe it would be nicer to initialize `res.SelectForRestore` to `func(...) { return true }`, then you wouldn't have to do this kind of check... dunno if that's necessarily better. just an idea.

fd0: Hm, I like that. I suppose this doesn't really introduce any performance problems.

restorer.go
```diff
@@ -62,10 +75,12 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID backend.ID) error
 				}
 			}
 
-			// Restore directory timestamp at the end. If we would do it earlier, restoring files within
-			// the directory would overwrite the timestamp of the directory they are in.
-			if err := node.RestoreTimestamps(filepath.Join(dst, dir, node.Name)); err != nil {
-				return err
+			if selectedForRestore {
```
fw42: Good catch. I feel like this might deserve a test (a file that was not selected to be restored is not only not restored but also it's metadata is not restored).

fw42:
:+1: Looks good. Tests would be A+.

restorer.go
fd0: That's an interesting corner case, we can leave it for now (just don't restore metadata for dirs that aren't selected for restore), but I think we can add an option for it later, so that the user can decide to restore the metadata of directories restored implicitely because they contain restored files.

fw42: Sorry, maybe that was confusing. What I meant to say is: I think this is the correct behaviour, but it worries me that no test will break if I remove this line and break this behaviour.

fd0: So you suggest we add a test that checks if meta data has been restored for an intermediate directory that'd not have been selected by the filter?

fw42: yeah, exactly (assert that meta data will not get restored if the file wasn't selected)

fd0:
Test has been added, this can be merged. I'll rebase it to master and run travis again.

Issue #133, opened 2015-04-25
Define process for reporting security-critical bugs


fd0:
Security vulnerabilities should be reported privately in a responsible-disclosure process. Describe this process and create a GPG-key for it.

PR #220, opened 2015-07-07
Note about security bugs

fw42:
Closes https://github.com/restic/restic/issues/133.

Thoughts?


fw42:
Should probably have a more generic email like restic@restic.com or whatever, but this is better than nothing for now and probably good enough for the next little while.

fd0:
Great, thanks for adding this. We should also add it to the website.

I've registered restic.net, we should at some point set up email and web for it (haven't had the time to do it), then we can create an email alias and generate a new pgp key (or something like that).

fd0:
I've added the same text to the website.

PR #221, opened 2015-07-08
Count non-regular files for progress as well

### LINKED ISSUE #206: More items than exists processed
I ran a few backups yesteday, over different directories, and a couple of them gave these results:

~~~
[5:51] 100.00%  1.995 MiB/s  700.156 MiB / 700.156 MiB  28200 / 28192 items  0 errors  ETA 0:00
[47:20] 100.00%  10.192 MiB/s  28.268 GiB / 28.268 GiB  129944 / 128915 items  4 errors  ETA 0:00
~~~

As you can see, there are more items processed than what was initially scanned.

It's possible that files in these folders change during the backup, especially the latter one.

fw42:
This should fix https://github.com/restic/restic/issues/206.

I think the source of the bug is that when generating the "todo list", we only count things as "file" if they are matching `fi.Mode()&(os.ModeType|os.ModeCharDevice) == 0`, where `ModeType = ModeDir | ModeSymlink | ModeNamedPipe | ModeSocket | ModeDevice`, so this excludes symbolic links.

This PR fixes that by counting everything as "file" which is not a "directory".

@fd0, does this make sense?

cc @rawtaz

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -503,7 +503,7 @@ func (j archiveJob) Copy() pipe.Job {
 	}
 
 	// handle files
-	if isFile(j.new.Info()) {
+	if isRegularFile(j.new.Info()) {
 		debug.Log("archiveJob.Copy", "   job %v is file", j.new.Path())
 
 		// if type has changed, return new job directly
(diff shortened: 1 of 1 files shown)


fw42:
(Tests are passing on Linux, but failing on OS X. They already fail on master though I think, so that's probably unrelated to my changes.)

fd0:
This makes perfect sense and I'm ok with it :+1:

PR #222, opened 2015-07-08
More useful error messages

fw42:
Tests are currently failing on OS X. This PR makes it more clear why:

```
node_test.go:144: file: ModTime dosn't match (2015-05-14 17:07:23.111 -0400 EDT != 2015-05-14 17:07:23 -0400 EDT)
```

Looks like there is an issue with the nanoseconds not being read correctly (I'm curious if the same problem exists on FreeBSD/OpenBSD).

Other people ran into similar problems before, maybe it's the same root cause and fix as [this](https://lists.gnu.org/archive/html/bug-make/2011-10/msg00030.html).

I suggest we merge this PR for now so that we get better errors and address the actual OS X test fail in a separate PR.

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -129,20 +129,20 @@ func TestNodeRestoreAt(t *testing.T) {
 		OK(t, err)
 
 		Assert(t, test.Name == n2.Name,
-			"%v: name doesn't match", test.Type)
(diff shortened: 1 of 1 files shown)


fw42:
Also interesting: http://comments.gmane.org/gmane.comp.lang.go.devel/52782

fd0:
LGTM

PR #225, opened 2015-07-09
Fix darwin tests

fw42:
This fixes tests on Darwin.

There is three different issues here:
- The default permissions of symbolic links on darwin are `lrwxr-xr-x` while on Linux they are `lrwxrwxrwx`. (Those permissions can't be changed, neither on Darwin nor on Linux, as far as I know, so it doesn't really make sense to assert on them - we [don't](https://github.com/restic/restic/blob/cf37b619fd903cea3371a7f2bf1456eccddad654/node.go#L149-L150) try to restore it anyway). This PR skips that check.
- It's currently not possible to change the timestamps of symlinks on Darwin (I'm not sure if this is a limitation of Darwin or of the Go runtime). This PR skips the symlink timestamp checks on Darwin.
- HFS+ doesn't support sub-second resolution for timestamps (the nanoseconds returned will be set to 0). This PR ignores the nanoseconds on Darwin and only checks up to second precision.

### DIFF
--- a/cmd/restic/integration_helpers_test.go
+++ b/cmd/restic/integration_helpers_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"syscall"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


cmd/restic/integration_helpers_test.go
```diff
@@ -55,25 +56,32 @@ func walkDir(dir string) <-chan *dirEntry {
...
 	stat2, _ := other.fi.Sys().(*syscall.Stat_t)
 
-	if stat.Uid != stat2.Uid || stat2.Gid != stat2.Gid {
-		fmt.Fprintf(os.Stderr, "%v: UID or GID do not match\n", e.path)
+	if stat.Uid != stat2.Uid {
+		fmt.Fprintf(os.Stderr, "%v: UID does not match (%v != %v)\n", e.path, stat.Uid, stat2.Uid)
+		return false
+	}
+
+	if stat.Gid != stat2.Gid {
```
fw42: also fixes a tiny bug here :-) this used to say `stat2.Gid != stat2.Gid`.

fw42:
screen shot 2015-07-09 at 11 35 17 am

table flip

fw42:
:green_heart: :dancer:

fd0:
Heh, great! I'll have a look now

cmd/restic/integration_helpers_test.go
fd0: great catch!

fd0:
Looks good! :+1:

PR #227, opened 2015-07-09
Disable gox cross compile on darwin

fd0:


fw42:
Hm this is getting super ugly... I would prefer if all this logic would live in two scripts (`ci/install` and `ci/run` or something like that) and Travis only calls those two scripts.

fw42:
That would also make it easier to run the same steps locally or on another CI system or in a docker container or whatever.

fw42:
I guess the Travis test output wouldn't look as nice though :/

fd0:
Yeah, the Travis config file is ugly. I've reworked the commit, maybe that's better.

fd0:
A problem with the script approach is, that we currently let Travis set up the development environment, that must be replicated with something else.

PR #231, opened 2015-07-13
Add filter implementation for files

### LINKED ISSUE #86: Allow ignore/exclude files and directories
Implement several methods for excluding files and directories:
 * glob pattern
 * regex pattern
 * cache-ignore-files (".CACHEDIR"?)

### LINKED ISSUE #161: Define filter function semantic
It's not the same in restorer and archiver

### LINKED ISSUE #223: Improve filter function for restore
In #218 we discussed the limitations of the currently used function `filepath.Match` for selecting files that should be restored. This filter is based on the shell globbing (e.g. `*.go`) and only applies to the file name, it doesn't handle separator chars (`/`) at all.

This issue should track the progress on improving the filters on what is restored. I'm thinking about something like zsh globbing, so that users can easily do the following:
 * Filter `/foo/*/*.go` restores all files ending with `.go` from any subdirs of `foo` in the backup
 * Filter `**/*.go` restores all `.go` files from all dirs in the backup

etc. Any other thoughts on these filter expressions?

We could just use a regex, but that feels like overkill for me.

### LINKED ISSUE #226: User Interface: Ignoring files/dirs for backup and restore?
I'd like to propose a user interface for selecting and ignoring files to backup and restore (this connects to #86):

Add `--exclude` and `--include` to `backup` and `restore`, both parameters can be added multiple times and each specifies a pattern.
- If neither `--exclude` nor `--include` is specified, all files are backed up or restored.
- If only `--include` is specified (at least once), only the items that match the given pattern are backed up or restored.
- If only `--exclude` is specified (at least once), all items except those that match the given pattern are backed up or restored.
- If both `--exclude` and `--include` are specified, only those items are backed up that either
  - don't match any exclude pattern
  - match an exclude pattern and at least one include pattern

As for the pattern, I'd like to implement something that is easy to use. Some examples:
- `/proc` matches all items below `/proc`, e.g. `/proc/kcore`
- `foo/*` matches all items in a subdirectory of a directory called `foo`, e.g. `/home/user/foo/test.txt`
- `foo/**/bar*/*.c` matches all items with the extension `.c` in a directory whose name begins with `bar` that somewhere has a parent directory called `foo`.

So, the rules are roughly:
- If the pattern starts with a slash, the following name is meant absolute
- If the pattern matches a directory, it matches all subdirs/items below, e.g. `/proc` is the same as `/proc/*`
- Similar to shell globbing, `*` matches everything

Usage examples:
- Backup everything except `/proc`,`/sys` and all vim swap files: `restic backup --exclude /proc --exclude /sys --exclude "*.swp" /`
- Backup only c source code files ignoring everything else: `restic backup --include "*.c" /home/user`
- Backup the user's home directory, excluding everything except c source code from the directory `work`: `restic backup --exclude /home/user/work --include "*.c" /home/user`

I think this can be implemented fairly easy by using `filepath.Match`.

What do you think about this approach? Are you aware of any use cases that aren't covered?

@restic/devs, @ceh, @rubenv, @howeyc, @rakoo

connects to #223 #161 #86

fd0:
I consider this PR done when full filter functionality is implemented for backup and restore.

TODO:
- [x] Add double-wildcard support (`foo/**/bar.go`)
- [x] Use `filter` for backup
- [x] Use `filter` for restore

Closes #226, closes #223, closes #161, closes #86


fd0:
This is done, `backup` and `restore` are now able to exclude files/dirs with patterns as described in #226. Follow-up issue is #233.

fd0:
I've removed `TestRestoreNoMetadataOnIgnoredIntermediateDirs` for now, this is only interesting once we allow mixing include and exclude patterns. I'll add that to the followup issue #233, so that we can reintroduce the test then.

fd0:
Ah, I've added the option to have only include patterns for restore, so `TestRestoreNoMetadataOnIgnoredIntermediateDirs` is back in. That should close this issue.

PR #234, opened 2015-07-18
osxfuse for travis on darwin

fw42:
This might fix the Darwin tests on https://github.com/restic/restic/pull/122

@fd0 @rakoo


fw42:
Hm the travis log says this:
```
The new osxfuse file system bundle needs to be installed by the root user:
  sudo /bin/cp -RfX /usr/local/opt/osxfuse/Library/Filesystems/osxfusefs.fs /Library/Filesystems/
  sudo chmod +s /Library/Filesystems/osxfusefs.fs/Support/load_osxfusefs
```
not sure if that's necessary (or even possible) on Travis.

fw42:
Ok looks like the sudo worked... This doesn't make the tests pass (they fail for some other reason), but at least they don't hang anymore.

PR #237, opened 2015-07-19
Allow more than one cleanup handler

### LINKED ISSUE #236: `os.Exit(0)` in lock.go prevents clean exit in main command
Because of [this line](https://github.com/restic/restic/blob/3f8f391d0b917d3daa711d17421f3c35f8a756bc/cmd/restic/lock.go#L139), an interrupt signal automatically closes the application and a command can't have any proper cleanup of its own

fd0:
This fixes #236

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	"github.com/restic/restic/debug"
+)
+
+var cleanupHandlers struct {
+	sync.Mutex
+	list []func() error
+}
+
+var stderr = os.Stderr
+
+func init() {
+	c := make(chan os.Signal)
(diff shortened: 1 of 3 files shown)


cmd/restic/cleanup.go
```diff
@@ -0,0 +1,65 @@
...
+		err := f()
+		if err != nil {
+			fmt.Fprintf(stderr, "error in cleanup handler: %v\n", err)
+		}
+	}
+}
+
+// CleanupHandler handles the SIGINT signal.
+func CleanupHandler(c <-chan os.Signal) {
+	for s := range c {
```
fw42: super nitpick, but this could (in the future) potentially handle other signals as well. maybe you should check if `s` is actually not just any `os.Signal` but in fact SIGINT?

fd0: Hm, what does that change? If `c` also receives other signal, it means that the same action (cleanup and exit) is requested... I'd leave it this way for now.

PR #238, opened 2015-07-20
Refactor skipping symlink timestamp checks on *bsd and darwin

fd0:
This refactors skipping symlink timestamp checks and adds this to the tests on
freebsd and openbsd.

### DIFF
--- a/test/helpers.go
+++ b/test/helpers.go
@@ -88,7 +88,7 @@ func RandomReader(seed, size int) *bytes.Reader {
 
 // SetupTarTestFixture extracts the tarFile to outputDir.
 func SetupTarTestFixture(t testing.TB, outputDir, tarFile string) {
-	err := System("sh", "-c", `(cd "$1" && tar xz) < "$2"`,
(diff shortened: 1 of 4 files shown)


test/helpers.go
```diff
@@ -88,7 +88,7 @@ func RandomReader(seed, size int) *bytes.Reader {
 
 // SetupTarTestFixture extracts the tarFile to outputDir.
 func SetupTarTestFixture(t testing.TB, outputDir, tarFile string) {
-	err := System("sh", "-c", `(cd "$1" && tar xz) < "$2"`,
+	err := System("sh", "-c", `(cd "$1" && tar xzf - ) < "$2"`,
```
fw42: what's this?

fd0: On OpenBSD and FreeBSD, `tar` doesn't read from stdin by default (which it does on Linux and darwin), but from `/dev/sa0`. This instructs tar to read from stdin with `f -`.

fw42:
Cool :+1:

PR #239, opened 2015-07-21
Web server, for discussion only

rakoo:
I open this PR for discussion, I don't expect it to be merged as-is.

Fixes #60

### DIFF
--- a/cmd/restic/cmd_web.go
+++ b/cmd/restic/cmd_web.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"net/http"
 	"os"
 	"strconv"
@@ -43,8 +42,10 @@ func (cmd CmdWeb) Execute(args []string) error {
 	}
 	repo.LoadIndex()
 	httpRepo := &httpRepo{
-		repo:      repo,
-		snapshots: make(map[string]*restic.Snapshot),
+		repo: repo,
+		objects: map[string]http.File{
+			"": &httpRoot{repo: repo},
+		},
 	}
 
 	portStr := args[0]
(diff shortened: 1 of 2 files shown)


cmd/restic/cmd_web.go
```diff
@@ -0,0 +1,269 @@
...
+		if !found {
+			return nil, fmt.Errorf("%s not found", name)
+		}
+		children, err := object.(childrener).loadChildren()
+		if err != nil {
+			return nil, err
+		}
+		for _, child := range children {
+			if _, ok := hr.objects[child.Name()]; !ok {
+				hr.objects[child.Name()] = child
```
rakoo: High risk of collision here: we only use the node's name, we should use its full path

fd0:
Thanks for your contribution. This already looks really good, although the web pages aren't really conforming to HTML ;)

rakoo:
I'd like to keep using go's http stuff here, and maybe ask upstream for a better dirList function (the one that turns a list of entries to HTML).

fd0:
Hm, it doesn't look hard to implement this with a template...

fd0:
Hey, what's the status of this pull request, what are your plans?

In my opinion, we should just render the directory listing ourselves, and not depend on the stdlib stuff. It's not that hard...

fd0:
@rakoo ping

I'd like to include the webserver in the first release, do you think we can get this in a usable state until Thursday?

rakoo:
Hey @fd0 I'm really sorry but I can't work on it at the moment.

fd0:
Okay, no problem. Shall we suspend the PR until you have time, or should I continue to work on it?

xor-gate:
I have tried this on current master, but alot of internal API changes have been submitted. I think this pull request should be closed until we have a new web frontend which is mergable.

fd0:
I agree, too much has changed, so I'll close this PR for now.

Issue #248, opened 2015-07-26
fuse mount: high memory usage


fd0:
I discovered that restic seems to keep all data that is read over the fuse mount in memory, so reading a 2GiB file needs 2GiB of memory. Could you take a look @rakoo?

fd0:
It looks like this is caused by https://github.com/restic/restic/blob/master/cmd/restic/fuse/file.go#L49-L61, which loads blobs into memory and keeps the reference to the slice, so that it is never garbage collected.

rakoo:
Do you think an LRU at the file level would be enough ? Since the same blob is often read many times at different offset and most use cases are forward reading, I'd think about that first.

fd0:
I'm working on this at the moment, it's a combination of two different things:
 * `repository.LoadBlob()` always creates a new buffer
 * the fuse code doesn't release blobs (by removing the pointer to it), so they are never garbage collected, this can be fixed by implementing the `HandleReleaser` interface

rakoo:
AFAICS `HandleReleaser` only works on fuse-visible stuff, so you would release the whole 2GB file or none of it. We'll have to manually GC old blobs.

fd0:
In my tests, `Release()` was nicely called when the file was closed, so we can at least release the blobs then. Please have a look at #249.

fd0:
I think this is enough for now.

PR #249, opened 2015-07-26
Reduce memory usage for fuse mount

fd0:
This changes `repository.LoadBlob()` so that a destination buffer must be provided, which enables the fuse code to use a buffer from a `sync.Pool`. In addition, release the buffers when the file is closed. At the moment, the max memory usage is defined by the max file size that is read in one go (e.g. with `cat`). It could be further optimized by implementing a LRU caching scheme.

Also refactor the fuse `Read` method a bit.

Connects #248.


fd0:
I think we also need tests for all the fuse stuff...

fd0:
I felt very unsure about the magic happening in `file.Read()`, so I refactored this a bit to be able to mock the repository, and added tests for `Read()`.

cmd/restic/cmd_cat.go
```diff
@@ -162,9 +162,18 @@ func (cmd CmdCat) Execute(args []string) error {
 		return err
 
 	case "blob":
-		data, err := repo.LoadBlob(pack.Data, id)
-		if err == nil {
-			_, err = os.Stdout.Write(data)
+		_, tpe, _, length, err := repo.Index().Lookup(id)
```
fw42: "tpe"? what is a "tpe"? maybe "blobType"?

repository/repository.go
```diff
@@ -97,6 +97,10 @@ func (r *Repository) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
 		return nil, err
 	}
 
+	if length > uint(cap(buf))+crypto.Extension {
+		return nil, errors.New("buf is too small")
```
fw42: maybe include the number of bytes it needs to be able to hold, to make the error message more useful?

repository/repository.go
```diff
@@ -111,7 +115,9 @@ func (r *Repository) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
 		return nil, err
 	}
 
-	buf, err := ioutil.ReadAll(rd)
+	// make buffer that is large enough for the complete blob
+	cbuf := make([]byte, length)
```
fw42: can you think of a more descriptive variable name than `cbuf`? I'm assuming this is a buffer that holds ciphertext? how about `ciphertextBuf`?

repository/repository.go
```diff
@@ -122,17 +128,17 @@ func (r *Repository) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
 	}
 
 	// decrypt
-	plain, err := r.Decrypt(buf)
+	buf, err = r.decryptTo(buf, cbuf)
```
fw42: and `plaintextBuf` for consistency then? I think that would make this code a bit easier to follow

cmd/restic/cmd_cat.go
fd0: Yes, that's better.

repository/repository.go
```diff
@@ -87,8 +86,9 @@ func (r *Repository) LoadAndDecrypt(t backend.Type, id backend.ID) ([]byte, erro
 }
 
 // LoadBlob tries to load and decrypt content identified by t and id from a
-// pack from the backend.
-func (r *Repository) LoadBlob(t pack.BlobType, id backend.ID) ([]byte, error) {
+// pack from the backend, the result is stored in buf, which must be large
+// enough to hold the complete blob.
+func (r *Repository) LoadBlob(t pack.BlobType, id backend.ID, buf []byte) ([]byte, error) {
```
fw42: what's the reasoning for this method still returning the buffer? Isn't that unnecessary now that it's a parameter? The one that will be returned is always the same as the one that was passed, right? (except for error cases)

repository/repository.go
fd0: okay

repository/repository.go
fd0: Indeed, it is.

fw42:
some bitching, but looks mostly good :-) :+1:

repository/repository.go
fd0: That's correct, but `len` of `buf` could've been modified (and this happens). A slice in Go is call-by-value regarding `len`, but call-by-reference regarding the underlying array. You can think of it as `struct {len int, cap int, data *[200]byte}`, ie. the struct is passed by value, but `data` is a pointer.

So we could either explicitely return `n` (like `Read`), or return the buf slice back to the caller.

PR #252, opened 2015-07-26
WIP: Repack blobs

### LINKED ISSUE #243: Add blob repacking
restic bundles together several blobs into pack files. There are several reasons why blobs aren't referenced any more (e.g. because a snapshot is removed), but it is likely that other blobs in the same pack are still in use, so restic cannot just remove the pack (and all blobs within).

When a garbage-collection process is run manually or more than half of all bytes in a pack aren't referenced, restic should read the remaining blobs and create a new pack. Afterwards, each Index that references blobs in the old pack should be replaced by a new Index which superseds the old one (i.e. lists the storage ID of the old index in the field `obsolete`), so that all blobs are correctly referenced to the new pack. Once this is done, the superseded Indexes and packs are removed from the repository.

fd0:
This is work in progress on the code that repacks blobs into new packs.

The code requires the new index format (that was already documented in `Design.md`), the code takes care to also read the old format and output a warning on stderr. When the `check` command is run, it converts all indexes to the new format, so users can upgrade and we can remove the code handling the old index format eventually.

Closes #243

### DIFF
--- a/backend/idset.go
+++ b/backend/idset.go
@@ -66,5 +66,5 @@ func (s IDSet) String() string {
 		return "{}"
 	}
 
-	return "{" + str[1:len(str)-2] + "}"
+	return "{" + str[1:len(str)-1] + "}"
 }
--- a/backend/idset.go
+++ b/backend/idset.go
@@ -1,5 +1,7 @@
 package backend
 
+import "sort"
+
 // IDSet is a set of IDs.
 type IDSet map[ID]struct{}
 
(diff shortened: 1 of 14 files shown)


fd0:
I've extracted a few commits from this into #294 in order to break this PR into smaller ones.

fd0:
I've extracted more commits into #310.

fd0:
This PR now introduces the `optimize` command and changes the behaviour of the `check` command:
 * `check` only tests for errors
 * `optimize` converts index format if necessary and rebuilds pack files so that unused blobs can be dropped

Issue #23, opened 2014-11-15
Implement HTTP Server/Client and backend


fd0:
Add a custom HTTP server (REST) and add that a as a storage backend

fd0:
This was implemented in #253.

PR #253, opened 2015-08-04
Basic REST backend

bchapuis:
I started working on a very basic REST backend (#23) and test server ([server.go](https://gist.githubusercontent.com/bchapuis/16d9dd0a8ae1ba80bcdd/raw/054dd9dec3360f63a98a682715b729614f7dfa44/server.go)). At this point, the test server must be started in the background before running the tests. The I'm not really familiar with go and I don't really know how to stop the server after the execution. Feedbacks appreciated... ;)

Connects #23


fd0:
Hey, thanks for your contribution! We need to get this PR in shape before it can be merged, but that shouldn't be too hard. I'll have a look later today.

bchapuis:
It wasn't really intended for a direct merge at this point. The current PR doesn't care about HTTP Auth which is probably the minimal requirement, except if the backend is for demo purpose only. Do you have something in mind regarding this point?

fd0:
I don't have the time to look at the code directly. The server component should be split out as far as possible and the tests could be done with a mock-server (should be easy to implement). I'll have a look at this later. HTTP auth is a good idea.

bchapuis:
Ok, the tests are now passing. Regarding the basic authentication, is seems possible to add the credentials in the url (http://username:password@host/)instead of the Http Headers. It's probably a good option to keep things as simple as possible.

fd0:
Cool, sounds good. I didn't yet have the time to look at your code.

fd0:
In the following, I'll dump my thoughts I had today:

Could you give a high-level overview of how the REST server should look like and what it should do? E.g., what are the paths, methods etc? We should document this below `doc/` somewhere so people can easily reimplement the server part in something else than Go. In addition, I'd like to support different 'paths', so it should be possible to specify the REST server URL like this `http://user:password@host/foo/bar/repopath`, so e.g. the config can be accessed at the path `/foo/bar/repopath/config`.

I'm thinking about adding a 'server' command to restic that can later be used to start different servers, e.g. REST, a remote server for stdin/stdout etc. Are you also interested in implementing the REST server part? Like, run `restic server --repo /tmp/repo --bind-address 127.0.0.1 --port 8080 --user foo --password bar` or so, which starts an HTTP server that serves the repo at `/tmp/repo` via REST (but doesn't try to open the repo, just serving). This can also be done in a different PR.

fd0:
Ok, first round of comments is done :grin:

backend/backend_test.go
```diff
@@ -108,6 +108,7 @@ func testBackend(b backend.Backend, t *testing.T) {
 				OK(t, err)
 
 				found, err := b.Test(tpe, id.String())
+
```
fd0: What's this empty line?

backend/rest/rest.go
```diff
@@ -0,0 +1,212 @@
...
+func (rb *RestBlob) Close() error {
+	rb.final = true
+	rb.buf.Reset()
+	return nil
+}
+
+// Finalize moves the data blob to the final location for type and name.
+func (rb *RestBlob) Finalize(t backend.Type, name string) error {
+	if rb.final {
+		return errors.New("Already finalized")
```
fd0: Error string should not start with an upper-case character, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings

(I won't add any further comments to this kind of style issue)

backend/rest/rest.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	rb.final = true
+
+	// Check key does not already exist
+	resp, err := http.Get(restPath(rb.b.url, t, name))
```
fd0: I think that this is not a good idea. If the the blob at that location is 8MiB, wouldn't that download the whole blob? Instead, maybe `HEAD` is better? Or skip this check and `POST` the new content, and expect an error from the server when the resource already exists (this expected behaviour must be documented)? REST uses `PUT` to overwrite an already-existing resource...

backend/rest/rest.go
```diff
@@ -0,0 +1,212 @@
...
+
+// GetReader returns an io.ReadCloser for the Blob with the given name of
+// type t at offset and length.
+func (b *Rest) GetReader(t backend.Type, name string, offset, length uint) (io.ReadCloser, error) {
+	rc, err := b.Get(t, name)
+	if err != nil {
+		return nil, err
+	}
+
+	n, errc := io.CopyN(ioutil.Discard, rc, int64(offset))
```
fd0: Hm, I think this can be implemented with an HTTP range request. If I'm just interested in the last few bytes of a large blob, this code downloads and discards the rest. I'm aware that the S3 backend also uses this code, and I think we should check if S3 supports range request and fix the code there, too.

backend/rest/rest.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	data, e2 := ioutil.ReadAll(resp.Body)
+	if e2 != nil {
+		close(ch)
+		return ch
+	}
+
+	var list []string
+	e3 := json.Unmarshal(data, &list)
```
fd0: This code expects a listing in a JSON format, that should be documented in seperate file below `doc/`, together with the rest of the REST api description.

backend/s3/s3.go
```diff
@@ -66,7 +66,7 @@ type s3Blob struct {
 
 func (bb *s3Blob) Write(p []byte) (int, error) {
 	if bb.final {
-		return 0, errors.New("blob already closed")
+		return 0, errors.New("Blob already closed")
```
fd0: Same as above: don't start error strings with a capital letter, and don't fix unneeded things in an unrelated commit ;)

cmd/restic/global.go
```diff
@@ -132,7 +133,10 @@ func (o GlobalOptions) OpenRepository() (*repository.Repository, error) {
 // * s3://region/bucket -> amazon s3 bucket
 // * sftp://user@host/foo/bar -> remote sftp repository on host for user at path foo/bar
 // * sftp://host//tmp/backup -> remote sftp repository on host at path /tmp/backup
+// * http://host -> remote and public http repository
 func open(u string) (backend.Backend, error) {
+	fmt.Println(u)
```
fd0: Please remove this obviously forgotten debug function call.

backend/rest_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	for _, d := range dirs {
+		os.MkdirAll(d, backend.Modes.Dir)
+	}
+
+	// Routing the repository requests
+	r := mux.NewRouter()
+
+	// Exists
+	r.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
fd0: Most of this code can probably refactored into a working REST server, see my comment at https://github.com/restic/restic/pull/253#issuecomment-128130604

backend/rest_test.go
```diff
@@ -0,0 +1,172 @@
...
+			http.Error(w, "File not found", 404)
+		}
+	}).Methods("GET")
+
+	// Put file
+	r.HandleFunc("/{file}", func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		file := filepath.Join(path, vars["file"])
+		if _, err := os.Stat(file); err == nil {
+			fmt.Fprintf(w, "Blob already uploaded", 404)
```
fd0: How does this work? `fmt.Fprintf` writes `Blob already uploaded` to w, and the additional parameter `404` is ignored... I suspect that replacing `fmt.Fprintf` with `http.Error` could work.

backend/rest_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}).Methods("POST")
+
+	// Delete file
+	r.HandleFunc("/{file}", func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		file := filepath.Join(path, vars["file"])
+		if _, err := os.Stat(file); err == nil {
+			os.Remove(file)
+		} else {
+			fmt.Fprintf(w, "File not found", 404)
```
fd0: Same as above, probably meant `http.Error` instead of `fmt.Fprintf`.

backend/rest_test.go
```diff
@@ -0,0 +1,172 @@
...
+		file := filepath.Join(path, vars["file"])
+		if _, err := os.Stat(file); err != nil {
+			http.Error(w, "File not found", 404)
+		}
+	}).Methods("HEAD")
+
+	// Get file
+	r.HandleFunc("/{file}", func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		file := filepath.Join(path, vars["file"])
```
fd0: Ah, if we're going to use this code, we must carefully check it for path traversals (`../../` etc). This should even be checked for the tests, because while the test is running the server is available, so anybody who can connect to the test REST server could `GET` or `DELETE` arbitrary files from the system. That's not good :)

Cleanup might look something like this: https://github.com/golang/tools/blob/master/godoc/server.go#L238-L239

bchapuis:
I'm not sure if the cleanup is necessary since the router is already extracting variables from the path. In this case, a relative path can not be injected in the variable.

fd0:
Even if it isn't necessary I'd like to implement it, this approach follows the "defense-in-depth" or "defensive programming" paradigm and I think it's a good idea. It's not much additional code (a call to `filepath.Clean()` and possibly checking if the filename is a valid ID), we can just write a function for that.

backend/rest/rest.go
fd0: This link is helpful: http://stackoverflow.com/a/27846510

fd0:
Another thought I had today: The REST backend can (with minor modifications) directly be used to restore a backup that is served via HTTP, e.g. by an Apache web server. In this case locking wouldn't work (because adding new files doesn't work), but we could ignore that with a command-line flag. Very cool!

bchapuis:
From a backend perspective the more defensive approach would be to test the types and the blobIDs. Since we know the types and the blobIDs are sha256 sums it shouldn't be too hard to implement very strong checks on the server side.

I just tried an approach that validates the "type" and the "blobID" but it seems that the generic tests are not using valid sha256 sums. I don't want to change the tests at the moment since it will have an impact on the other backends. A different PR would be more suitable. What do you think?

```golang
var blobTypes = []string{"data", "snapshot", "index", "lock", "key", "temp"}

func isBlobType(blobType string) bool {
	for _, t := range blobTypes {
		if blobType == t {
			return true
		}
	}
	return false
}

func isBlobID(blobID string) bool {
	if len(blobID) != sha256.Size {
		return false
	}
	if _, e := hex.DecodeString(blobID); e != nil {
		return false
	}
	return true
}
```

bchapuis:
Regarding the command line for starting a server in restic, I think it would make sense to have a very small read only http backend over a local directory. However, I wonder if it is the right place to build a bigger backend, since it will probably come with other dependencies. What do you think?

fd0:
> From a backend perspective the more defensive approach would be to test the types and the blobIDs. Since we know the types and the blobIDs are sha256 sums it shouldn't be too hard to implement very strong checks on the server side.

I agree. We could just use `filepath.Clean` additionally (even if it isn't required).

> I just tried an approach that validates the "type" and the "blobID" but it seems that the generic tests are not using valid sha256 sums.

Oh, they are valid, you just have a small bug in your code: In the representation as a string, the string length is `2*32 == 64`, each byte is represented by two characters in hex.

The function `isBlobType()` should be implemented in https://github.com/restic/restic/blob/master/backend/interface.go#L8-L15 (because that's where the blob types are defined), and it should be exported then (`IsBlobType()`), or maybe move the blob type definition and this function to its own file. When comparing, use `backend.IDSize` instead of `sha256.Size` https://github.com/restic/restic/blob/master/backend/id.go#L11

`isBlobID()` is nearly the same as `ParseID()` https://github.com/restic/restic/blob/master/backend/id.go#L17 so we can just use that.

> Regarding the command line for starting a server in restic, I think it would make sense to have a very small read only http backend over a local directory. However, I wonder if it is the right place to build a bigger backend, since it will probably come with other dependencies. What do you think?

I haven't made my mind about that. The read-only HTTP backend has at least the problem that locking isn't possible because for that a blob of type `lock` must be created, so we need to have a special case for that. But I think a basic server (read and write) isn't so much code at all, and I think we don't need so many dependencies outside the stdlib. I agree that this server part should be added in a different PR, so we can concentrate on the backend here.

What server implementation do you use for testing at the moment?

backend/rest/rest.go
```diff
@@ -0,0 +1,215 @@
...
+
+// GetReader returns an io.ReadCloser for the Blob with the given name of
+// type t at offset and length.
+func (b *Rest) GetReader(t backend.Type, name string, offset, length uint) (io.ReadCloser, error) {
+	req, err := http.NewRequest("GET", restPath(b.url, t, name), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Range", "bytes="+string(offset)+"-"+string(offset+length))
```
fd0: This doesn't work, see http://play.golang.org/p/uVA6JAf1qD

You can just use `fmt.Sprintf("bytes=%d-%d", offset, offset+length)`.

backend/rest/rest.go
```diff
@@ -0,0 +1,215 @@
...
+		return ch
+	}
+
+	data, errd := ioutil.ReadAll(resp.Body)
+	if errd != nil {
+		close(ch)
+		return ch
+	}
+
+	var list []string
```
fd0: `List()` should return the IDs in lexicographical order. We should add `sort.Strings(list)` here.

backend/rest_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	for _, d := range dirs {
+		os.MkdirAll(d, backend.Modes.Dir)
+	}
+
+	// Initialize the router for the repository requests.
+	r := mux.NewRouter()
+
+	// Check if the repository has already been initialized.
```
fd0: This comment is wrong. How does the code defining a `HandleFunc` check if the repository has been initialized?

For restic, we try to build the code in a way that it is obvious what the code does (e.g. by having good function names), so a comment isn't needed and cannot be wrong if the codes is changed. It takes a while to get used to. Especially comments like in line 45 could be just left out.

fd0:
Thanks again for your contribution! I like it, and the documentation for the REST server interface is good for now. :smiley:

backend/rest/rest.go
```diff
@@ -0,0 +1,215 @@
...
+
+// Finalize moves the data blob to the final location for type and name.
+func (rb *RestBlob) Finalize(t backend.Type, name string) error {
+	if rb.final {
+		return errors.New("already finalized")
+	}
+
+	rb.final = true
+
+	// Check key does not already exist.
```
fd0: Hm, do you think we can just skip the HEAD request here and just execute the POST below? In the (rather unlikely) event that a blob already exists, the POST request returns an error. This way we could save a round-trip (just one HTTP request), but maybe upload a large amount of data before receiving an errorr...

bchapuis: I don't know the internals of the golang http server. However, if the call on HandleFunc is performed immediately when the http headers are received, it may be possible to cancel the POST request before receiving the full Body of the request (multipart). It would be something similar to the HEAD request with one round-trip. I will remove this call for now.

fd0: I agree. This saves a round-trip for each newly added blob. Because the error condition we're trying to avoid here (uploading a duplicate blob) is unlikely, this should speed up the regular case (new blob) greatly.

backend/rest_test.go
bchapuis: The comment intended to describe the goal of the HTTP method. Should I remove all the following comments as well(// Save the configuration, etc... )?

fd0: Ah, now I understand. I suggest that you decide for yourself and remove the comments that aren't needed (e.g. I'd remove the one in line 45), and leave the others in. :+1:

fd0:
Other minor things I noticed:

 * Usually, git commit messages have a very terse summary in the first line of the commit message, followed by an empty line, followed by a more verbose description or a list of changed things.

 * If you change/add multiple different things that aren't related at all, try to make several smaller commits (e.g. one commit for the documentation changes, one for adding the methods for the blobs etc.). This is much easier to review.

("Minor things" here means "you don't need to change existing commits, but try to apply the tips next time")

backend/interface.go
```diff
@@ -14,6 +17,28 @@ const (
...
+	if s == string(Snapshot) {
+		return Snapshot, nil
+	}
+	if s == string(Index) {
+		return Index, nil
+	}
+	if s == string(Config) {
+		return Config, nil
+	}
+	return "", errors.New("invalid type")
```
fd0: If you care about style and look, this could be shortened to:

```Go
	switch s {
		case string(Data): return Data, nil
		case string(Key): return Data, nil
		case string(Lock): return Lock, nil
		case string(Snapshot): return Snapshot, nil
		case string(Index): return Index, nil
		case string(Config): return Config, nil
	}
	return "", errors.New("invalid type")
```

bchapuis:
Ok, thanks for the tips, I will try to improve... the last commit message was particularly long... :)

fd0:
In the FROSCON lineup I've just discovered http://swagger.io/, that looks really nice for documenting REST APIs, maybe you take it for a test drive?

bchapuis:
I made a little research about api documentation. I wonder if a integrating the api doc in jekyll such as in [carte](https://github.com/devo-ps/carte) would not be a simpler approach. In the case of restic, we don't really need to let the user make api calls from the browser at the moment. :) We could create a special collection in jekyll for this type and also use something similar for the user manual as the documentation grows.

fd0:
Also looks fine to me. :)

bchapuis:
I just tried to rebase the commits from the master in this PR but I'm encountering a strange issue with appveyor. Any idea?

fd0:
Just a wild guess: You need to remove the write-protection bit before removing a file for the REST test server in Windows, similar to https://github.com/restic/restic/commit/264472219804ac7c62ab1f90d3cf4a156b89495b

bchapuis:
Thank you for the hint. Apparently, I wasn't closing the file correctly before reading and only windows bothered. :)

bchapuis:
Hey, the last commit 35c3873 tries to cache the tree objects aggressively. It fixes #282 and really improves the performances when dealing with remote backends, it also improves the local backend performences. It would make sense to integrate this commit with the cache, but I don't really no how to proceed. Any advice?

fd0:
I"ll have a look in the evening.

fd0:
I had a look, the commit indeed caches trees aggressively, but this needs A LOT of RAM (I cancelled my test backup after ~6GB)...

bchapuis:
Ok, looks pretty bad. I only tested it with small backups (~3GB) and didn't noticed this memory issue. This is strange, in that case the files loaded in memory represents only 35MB. Let's focus on the REST backend for now.

fd0:
I had a look, and the backend works really well. I came across the following issues:

 * REST server URLs are only accepted with at least one slash `/` character in the path, it was very irritating that `http://localhost:4567` wasn't accepted as a valid HTTP url.
 * When accessing and listing things, the singular form of the type is required, i.e. `/keys/` instead of `/key/`. We should add this to the REST documentation. What about '/key' (without the trailing slash)?

I tested the REST backend against a quickly thrown-together REST server in ruby, see https://gist.github.com/d83e9da3ab9364a1d912

Could you remove the aggressive caching commit and rebase against the current master, then I'll merge the PR.

fd0:
Next issue: when the REST server is not reachable, a `nil` pointer is dereferenced:

    ./restic -r http://localhost:2323/ backup ~/shared 
    debug enabled
    parsed url http://localhost:2323/
    [signal 0xb code=0x1 addr=0x0 pc=0x596b5c]

    goroutine 1 [running]:
    github.com/restic/restic/backend/rest.(*Rest).List(0xc8200a9260, 0x912750, 0x3, 0xc820076300, 0x0)
        /tmp/restic-build-809839120/src/github.com/restic/restic/backend/rest/rest.go:189 +0x37c
    github.com/restic/restic/repository.SearchKey(0xc820073960, 0xc820079ad0, 0x3, 0x0, 0x0, 0x0)
        /tmp/restic-build-809839120/src/github.com/restic/restic/repository/key.go:102 +0xe4
    github.com/restic/restic/repository.(*Repository).SearchKey(0xc820073960, 0xc820079ad0, 0x3, 0x0, 0x0)
        /tmp/restic-build-809839120/src/github.com/restic/restic/repository/repository.go:559 +0x5f
    main.GlobalOptions.OpenRepository(0x7fff8f6def70, 0x16, 0x0, 0x0, 0x0, 0xc820079ad0, 0x3, 0x7f7688531198, 0xc82008e008, 0x7f7688531198, ...)
        /tmp/restic-build-809839120/src/github.com/restic/restic/cmd/restic/global.go:122 +0x344
    main.CmdBackup.Execute(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xb62500, 0xc820074480, 0x1, 0x4, ...)
        /tmp/restic-build-809839120/src/github.com/restic/restic/cmd/restic/cmd_backup.go:253 +0x48c
    main.(*CmdBackup).Execute(0xc820074380, 0xc820074480, 0x1, 0x4, 0x0, 0x0)
        <autogenerated>:12 +0xc9
    github.com/jessevdk/go-flags.(*Parser).ParseArgs(0xc820090870, 0xc820090010, 0x4, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0)
        /tmp/restic-build-809839120/src/github.com/jessevdk/go-flags/parser.go:278 +0x8e3
    github.com/jessevdk/go-flags.(*Parser).Parse(0xc820090870, 0x0, 0x0, 0x0, 0x0, 0x0)
        /tmp/restic-build-809839120/src/github.com/jessevdk/go-flags/parser.go:154 +0x9b
    main.main()
        /tmp/restic-build-809839120/src/github.com/restic/restic/cmd/restic/main.go:26 +0x1bf

    goroutine 17 [syscall, locked to thread]:
    runtime.goexit()
        /usr/lib/go/src/runtime/asm_amd64.s:1696 +0x1

    goroutine 20 [syscall]:
    os/signal.loop()
        /usr/lib/go/src/os/signal/signal_unix.go:22 +0x18
    created by os/signal.init.1
        /usr/lib/go/src/os/signal/signal_unix.go:28 +0x37

    goroutine 21 [chan receive]:
    github.com/restic/restic.init.1.func1.1()
        /tmp/restic-build-809839120/src/github.com/restic/restic/lock.go:262 +0x184
    created by github.com/restic/restic.init.1.func1
        /tmp/restic-build-809839120/src/github.com/restic/restic/lock.go:265 +0x2b

    goroutine 3 [select, locked to thread]:
    runtime.gopark(0x9e8b08, 0xc82002ef28, 0x9139c0, 0x6, 0x44c618, 0x2)
        /usr/lib/go/src/runtime/proc.go:185 +0x163
    runtime.selectgoImpl(0xc82002ef28, 0x0, 0x18)
        /usr/lib/go/src/runtime/select.go:392 +0xa64
    runtime.selectgo(0xc82002ef28)
        /usr/lib/go/src/runtime/select.go:212 +0x12
    runtime.ensureSigM.func1()
        /usr/lib/go/src/runtime/signal1_unix.go:227 +0x353
    runtime.goexit()
        /usr/lib/go/src/runtime/asm_amd64.s:1696 +0x1

    goroutine 22 [chan receive]:
    main.CleanupHandler(0xc8200760c0)
        /tmp/restic-build-809839120/src/github.com/restic/restic/cmd/restic/cleanup.go:58 +0x71
    created by main.init.1
        /tmp/restic-build-809839120/src/github.com/restic/restic/cmd/restic/cleanup.go:25 +0x12a
    Command exited with non-zero status 2

fd0:
And the fuse mount with a REST backend doesn't work (at least with my test server, but I don't see any errors):

    ls -al ~/mnt/test/snapshots
    ls: cannot access 2015-07-05T11:10:44+02:00: Input/output error
    ls: cannot access 2015-09-13T15:46:34+02:00: Input/output error
    ls: cannot access 2015-05-14T23:31:09+02:00: Input/output error
    ls: cannot access 2015-07-10T22:53:04+02:00: Input/output error
    ls: cannot access 2015-07-07T00:18:01+02:00: Input/output error
    ls: cannot access 2015-06-30T21:42:28+02:00: Input/output error
    ls: cannot access 2015-07-19T12:02:09+02:00: Input/output error
    ls: cannot access 2015-07-05T23:38:32+02:00: Input/output error
    ls: cannot access 2015-07-12T17:31:55+02:00: Input/output error
    ls: cannot access 2015-07-05T12:27:54+02:00: Input/output error
    ls: cannot access 2015-07-09T00:01:11+02:00: Input/output error
    ls: cannot access 2015-05-14T18:30:54+02:00: Input/output error
    ls: cannot access 2015-05-14T18:08:49+02:00: Input/output error
    ls: cannot access 2015-07-19T00:19:21+02:00: Input/output error
    ls: cannot access 2015-07-10T22:44:36+02:00: Input/output error
    ls: cannot access 2015-07-13T23:54:27+02:00: Input/output error
    ls: cannot access 2015-07-12T22:28:06+02:00: Input/output error
    ls: cannot access 2015-07-10T22:55:54+02:00: Input/output error
    total 0
    d????????? ? ? ? ?            ? 2015-05-14T18:08:49+02:00
    d????????? ? ? ? ?            ? 2015-05-14T18:30:54+02:00
    d????????? ? ? ? ?            ? 2015-05-14T23:31:09+02:00
    d????????? ? ? ? ?            ? 2015-06-30T21:42:28+02:00
    d????????? ? ? ? ?            ? 2015-07-05T11:10:44+02:00
    d????????? ? ? ? ?            ? 2015-07-05T12:27:54+02:00
    d????????? ? ? ? ?            ? 2015-07-05T23:38:32+02:00
    d????????? ? ? ? ?            ? 2015-07-07T00:18:01+02:00
    d????????? ? ? ? ?            ? 2015-07-09T00:01:11+02:00
    d????????? ? ? ? ?            ? 2015-07-10T22:44:36+02:00
    d????????? ? ? ? ?            ? 2015-07-10T22:53:04+02:00
    d????????? ? ? ? ?            ? 2015-07-10T22:55:54+02:00
    d????????? ? ? ? ?            ? 2015-07-12T17:31:55+02:00
    d????????? ? ? ? ?            ? 2015-07-12T22:28:06+02:00
    d????????? ? ? ? ?            ? 2015-07-13T23:54:27+02:00
    d????????? ? ? ? ?            ? 2015-07-19T00:19:21+02:00
    d????????? ? ? ? ?            ? 2015-07-19T12:02:09+02:00
    d????????? ? ? ? ?            ? 2015-09-13T15:46:34+02:00

bchapuis:
Thank you for the feedbacks, I will be very busy during the next days but I will try to solve these issues one evening by the end of the week.

fd0:
Sure, no need to rush anything. Thanks for your contribution!

bchapuis:
Regarding the singular form I mainly used it to avoid the mapping between the directories and the types. The same approach is adopted by the S3 backend. But it's not an issue to change this behavior, what do you think?

Regarding the trailing slashes, it seems [common](https://en.wikipedia.org/wiki/Representational_state_transfer#Applied_to_web_services) to use them to distinguish a group of resources from a specific resource.

The fuse mount works with my server implementation but I will check what happening with in your case.

fw42:
:+1: to using plural, that's more common

fd0:
Hm, changing the behavior of the s3 backend should be backwards-compatible. I'd suggest leaving the s3 backend alone for now, but still use the plural form for the REST backend.

bchapuis:
I made some changes that address all the comments. Regarding fuse, it seems to work with the go server but I haven't been able to identify what goes wrong with the ruby server. I also removed completely the dependency to gorilla mux.

fd0:
Thanks a lot!

Just a quick heads-up: I'm up to my neck in other work, I don't have the time to look at it right now ;)

xor-gate:
@bchapuis I tried to build your test server but it failed (debian jessy 8 amd64):
```
jerry@backupsy001:~/git/restic/cmd$ go build server.go
# command-line-arguments
./server.go:172: undefined: addr
./server.go:172: undefined: time
jerry@backupsy001:~/git/restic/cmd$ go version
go version go1.3.3 linux/amd64
```

backend/rest/rest.go
fawick: Hi @fd0, I'm investigating the idea behind this PR for my own purposes (cf. #456).
is that comment still valid? I understood that from restic/backend.Lister.List() docs that it is no longer required to send the strings in a sorted order.

fd0: Oh no, that's no longer necessary. I've just yesterday started to port the REST backend to the most recent version of the code in master, and I'll continue today. Maybe we'll have an up to date version in the evening. If you like, drop by in `#restic` on freenode IRC.

fd0:
I've started porting the REST backend to the current code, this PR is therefore superseded by #464

Issue #26, opened 2014-11-15
Support MS Windows


fd0:

ThomasWaldmann:
While I don't need it personally, I'ld like to see windows support because it's always a pain to support the existing crappy windows backup software. Also, I use some crappy windows imaging software.

So, I'ld offer testing something that can create and restore windows full-disk images. It could be based on linux (boot-cd) or winpe or both. Biggest issue there is software-raid support.

fd0:
I've removed windows support in e0fbd83a2d756d04ec1df3108fe5b9c69f0f183f for now, until the base is stable enough an someone volunteers to do it.

klauspost:
I wouldn't mind helping here. There is probably some work needed to have [UNC path](https://en.wikipedia.org/wiki/Path_(computing)#Uniform_Naming_Convention) support, since the Go runtime doesn't do that by itself.

fd0:
Feel free to give it a try :)

It may be a good idea to start with the basics, and add UNC path support later.

fd0:
This has been implemented with #259 by @klauspost

PR #257, opened 2015-08-12
Add Windows support

klauspost:
WIP for #26 - DO NOT MERGE!

Compiles, but all tests fail. 

Requires "RESTIC_TEST_CLEANUP=0" to avoid panics, since files are not closed before the test directory is deleted.

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"github.com/restic/restic/backend"
+	"runtime"
 )
 
 var ErrWrongData = errors.New("wrong data returned by backend, checksum does not match")
@@ -146,6 +147,9 @@ func (lb *localBlob) Finalize(t backend.Type, name string) error {
 		return err
 	}
 
+	if runtime.GOOS == "windows" {
+		return nil
+	}
 	return os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
 }
 
@@ -258,15 +262,16 @@ func (b *Local) Test(t backend.Type, name string) (bool, error) {
 // Remove removes the blob with the given name and type.
 func (b *Local) Remove(t backend.Type, name string) error {
 	// Close all open files we may have.
+	fn := filename(b.p, t, name)
 	b.mu.Lock()
-	open, _ := b.open[filename(b.p, t, name)]
+	open, _ := b.open[fn]
 	for _, file := range open {
 		file.Close()
 	}
-	b.open[filename(b.p, t, name)] = nil
+	b.open[fn] = nil
 	b.mu.Unlock()
 
-	return os.Remove(filename(b.p, t, name))
+	return os.Remove(fn)
 }
 
 // List returns a channel that yields all names of blobs of type t. A
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 
 	"github.com/restic/restic/backend"
 )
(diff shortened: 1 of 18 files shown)


fd0:
For the ansi output we can maybe use https://github.com/Bitbored/go-ansicon, there was a talk about that at FOSDEM: https://youtu.be/oM-MfeflUZ8?t=46m41s

klauspost:
@fd0 - good idea. Is there ANSI other places? (Otherwise it might  be a bit overkill to add a dependency for just nice-looking tests)

Right now I get many failures like:
```
archiver_test.go:108: unexpected error: strconv.ParseInt: parsing "S-1-5-21-1075215122-993071141-2159816414-1000": invalid syntax
```

Any idea where should I start looking?

klauspost:
Ah - a lot of places are just converting the os/user/User.Uid/Gid to an integer value. :/

fd0:
> @fd0 - good idea. Is there ANSI other places? (Otherwise it might be a bit overkill to add a dependency for just nice-looking tests)

The tests are the most boring thing, much more interesting is the progress information that is updated via ANSI escape codes, e.g. https://github.com/restic/restic/blob/master/cmd/restic/cmd_backup.go#L112

I'd suggest disabling color output for tests on windows, and test if `cmd.exe` is able to work with the other escape sequences. We might have a look at how git does status updates...

> Ah - a lot of places are just converting the os/user/User.Uid/Gid to an integer value. :/

Since restic was started with Unix/Linux in mind, `uid` and `gid` fields in the json data structures are integers. What do they look like in Windows? Strings?

What do you think about having additional fields `user` and `group` for that? This way, restic can easily distinguish these.

klauspost:
I have just set them to 0 for now.

The biggest issue right now is files that aren't closed before they are deleted. If you have a file open, it cannot be deleted on Windows. I modified the local file storage, but there are still a lot of tests that are failing because of that. Do you know additional places I should look for that?

fd0:
Hm, that's interesting. I don't have any idea where that is coming from, I thought all the code always takes care to close before removing. What files are removed in the tests? What tests fail exactly?

fd0:
btw, I don't know if you're aware of that, but there's also #restic on irc.freenode.net for realtime debuggung ;)

fd0:
I've just added some words about git commits to the contibuting guidelines, it'd be awesome if you could finde the time to read and apply them https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits

I'm aware that this looks like nitpicking, but please believe me it makes viewing the history so much nicer. :)

klauspost:
@fd0 - I have read it, believe me ;) 

I made some one-line commits, to trigger the travis CI build to test I haven't broken other platforms.

As I wrote in the top, this is WIP. I will make a separate PR once this is running.

fd0:
Great, I'm looking forward to it!

klauspost:
All tests pass now. I will proceed test actual functionality now.

klauspost:
The ansi library is broken (doesn't compile).

https://github.com/Bitbored/go-ansicon/issues/1

I made a small adjustment, and it works mostly ok. The automatic wrapping detection in [cmd_backup.go](https://github.com/restic/restic/blob/master/cmd/restic/cmd_backup.go#L155) has to be adjusted by one, but with that, it looks good.

klauspost:
It seems we are in a "good enough" state.  ANSI still needs fixing, but godep is not being helpful.

I will resubmit a "reduced" PR, with fewer commits tomorrow, and make sure stuff is documented. Finally it should also be possible to add windows/amd64 to the CI tests.

fd0:
Cool, thanks a lot for your contribution! You don't need to open a new PR though, just force-push the commits into your branch, and this PR will reflect your changes. Force-push in a PR is okay :)

klauspost:
Replaced by #259

PR #259, opened 2015-08-14
Add Windows support

klauspost:
Cleaned up version of #257 
- Compiles
- Passes all tests
- Operations complete (init/backup/restore/snapshots, etc)

Things NOT covered in the PR:
- ANSI output doesn't work.
- Check cross-platform backup/restore (mainly UID/GID).

PS. Sorry - my "git-fu" skills are not up to "force-push" the old PR.

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"runtime"
 	"sort"
 	"sync"
 
@@ -147,16 +146,7 @@ func (lb *localBlob) Finalize(t backend.Type, name string) error {
 		return err
 	}
 
-	// set file to readonly, except on Windows,
-	// otherwise deletion will fail.
-	if runtime.GOOS != "windows" {
-		err = os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
-		if err != nil {
-			return err
-		}
-	}
-
-	return nil
+	return setNewFileMode(f, fi)
 }
 
 // Create creates a new Blob. The data is available only after Finalize()
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -7,7 +7,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
+	"sync"
 
 	"github.com/restic/restic/backend"
 )
@@ -15,7 +17,9 @@ import (
 var ErrWrongData = errors.New("wrong data returned by backend, checksum does not match")
 
 type Local struct {
-	p string
+	p    string
+	mu   sync.Mutex
+	open map[string][]*os.File // Contains open files. Guarded by 'mu'.
 }
 
 // Open opens the local backend at dir.
@@ -37,7 +41,7 @@ func Open(dir string) (*Local, error) {
 		}
 	}
 
-	return &Local{p: dir}, nil
+	return &Local{p: dir, open: make(map[string][]*os.File)}, nil
 }
 
 // Create creates all the necessary files and directories for a new local
@@ -143,7 +147,16 @@ func (lb *localBlob) Finalize(t backend.Type, name string) error {
 		return err
 	}
 
-	return os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
+	// set file to readonly, except on Windows,
+	// otherwise deletion will fail.
+	if runtime.GOOS != "windows" {
+		err = os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // Create creates a new Blob. The data is available only after Finalize()
@@ -161,6 +174,11 @@ func (b *Local) Create() (backend.Blob, error) {
 		f:       file,
 		basedir: b.p,
 	}
+
+	b.mu.Lock()
+	open, _ := b.open["blobs"]
+	b.open["blobs"] = append(open, file)
+	b.mu.Unlock()
 
 	return &blob, nil
 }
@@ -198,7 +216,15 @@ func dirname(base string, t backend.Type, name string) string {
 // Get returns a reader that yields the content stored under the given
 // name. The reader should be closed after draining it.
 func (b *Local) Get(t backend.Type, name string) (io.ReadCloser, error) {
-	return os.Open(filename(b.p, t, name))
+	file, err := os.Open(filename(b.p, t, name))
+	if err != nil {
+		return nil, err
+	}
+	b.mu.Lock()
+	open, _ := b.open[filename(b.p, t, name)]
+	b.open[filename(b.p, t, name)] = append(open, file)
+	b.mu.Unlock()
+	return file, nil
 }
 
 // GetReader returns an io.ReadCloser for the Blob with the given name of
@@ -208,6 +234,11 @@ func (b *Local) GetReader(t backend.Type, name string, offset, length uint) (io.
 	if err != nil {
 		return nil, err
 	}
+
+	b.mu.Lock()
+	open, _ := b.open[filename(b.p, t, name)]
+	b.open[filename(b.p, t, name)] = append(open, f)
+	b.mu.Unlock()
 
 	_, err = f.Seek(int64(offset), 0)
 	if err != nil {
@@ -236,7 +267,17 @@ func (b *Local) Test(t backend.Type, name string) (bool, error) {
 
 // Remove removes the blob with the given name and type.
 func (b *Local) Remove(t backend.Type, name string) error {
-	return os.Remove(filename(b.p, t, name))
+	// close all open files we may have.
+	fn := filename(b.p, t, name)
+	b.mu.Lock()
+	open, _ := b.open[fn]
+	for _, file := range open {
+		file.Close()
+	}
+	b.open[fn] = nil
+	b.mu.Unlock()
+
+	return os.Remove(fn)
 }
 
 // List returns a channel that yields all names of blobs of type t. A
@@ -283,7 +324,22 @@ func (b *Local) List(t backend.Type, done <-chan struct{}) <-chan string {
 }
 
 // Delete removes the repository and all files.
-func (b *Local) Delete() error { return os.RemoveAll(b.p) }
+func (b *Local) Delete() error {
+	b.Close()
+	return os.RemoveAll(b.p)
+}
 
-// Close does nothing
-func (b *Local) Close() error { return nil }
+// Close closes all open files.
+// They may have been closed already,
+// so we ignore all errors.
+func (b *Local) Close() error {
+	b.mu.Lock()
+	for _, open := range b.open {
+		for _, file := range open {
+			file.Close()
+		}
+	}
+	b.open = make(map[string][]*os.File)
+	b.mu.Unlock()
+	return nil
+}
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -10,7 +10,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sort"
-	"syscall"
 
 	"github.com/juju/errors"
 	"github.com/pkg/sftp"
@@ -28,8 +27,6 @@ type SFTP struct {
 	cmd *exec.Cmd
 }
 
-var sysProcAttr syscall.SysProcAttr
-
 func startClient(program string, args ...string) (*SFTP, error) {
 	// Connect to a remote host and request the sftp subsystem via the 'ssh'
 	// command.  This assumes that passwordless login is correctly configured.
@@ -39,7 +36,7 @@ func startClient(program string, args ...string) (*SFTP, error) {
 	cmd.Stderr = os.Stderr
 
 	// ignore signals sent to the parent (e.g. SIGINT)
-	cmd.SysProcAttr = &sysProcAttr
+	cmd.SysProcAttr = ignoreSigIntProcAttr()
 
 	// get stdin and stdout
 	wr, err := cmd.StdinPipe()
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -28,6 +28,8 @@ type SFTP struct {
 	cmd *exec.Cmd
 }
 
+var sysProcAttr syscall.SysProcAttr
+
 func startClient(program string, args ...string) (*SFTP, error) {
 	// Connect to a remote host and request the sftp subsystem via the 'ssh'
 	// command.  This assumes that passwordless login is correctly configured.
@@ -37,7 +39,7 @@ func startClient(program string, args ...string) (*SFTP, error) {
 	cmd.Stderr = os.Stderr
 
 	// ignore signals sent to the parent (e.g. SIGINT)
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
+	cmd.SysProcAttr = &sysProcAttr
 
 	// get stdin and stdout
 	wr, err := cmd.StdinPipe()
--- a/build.go
+++ b/build.go
@@ -259,15 +259,15 @@ func main() {
 
 	version := getVersion()
 	compileTime := time.Now().Format(timeFormat)
+	output := "restic"
+	if runtime.GOOS == "windows" {
+		output = "restic.exe"
+	}
+
 	args := []string{
 		"-tags", strings.Join(buildTags, " "),
 		"-ldflags", fmt.Sprintf(`-s -X main.version %q -X main.compiledAt %q`, version, compileTime),
-	}
-	if runtime.GOOS != "windows" {
-		args = append(args, "-o", "restic", "github.com/restic/restic/cmd/restic")
-
-	} else {
-		args = append(args, "-o", "restic.exe", "github.com/restic/restic/cmd/restic")
+		"-o", output, "github.com/restic/restic/cmd/restic",
 	}
 
 	err = build(gopath, args...)
--- a/build.go
+++ b/build.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 )
@@ -261,8 +262,14 @@ func main() {
 	args := []string{
 		"-tags", strings.Join(buildTags, " "),
 		"-ldflags", fmt.Sprintf(`-s -X main.version %q -X main.compiledAt %q`, version, compileTime),
-		"-o", "restic", "github.com/restic/restic/cmd/restic",
 	}
+	if runtime.GOOS != "windows" {
+		args = append(args, "-o", "restic", "github.com/restic/restic/cmd/restic")
+
+	} else {
+		args = append(args, "-o", "restic.exe", "github.com/restic/restic/cmd/restic")
+	}
+
 	err = build(gopath, args...)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "build failed: %v\n", err)
--- a/lock.go
+++ b/lock.go
@@ -12,7 +12,6 @@ import (
 	"github.com/restic/restic/backend"
 	"github.com/restic/restic/debug"
 	"github.com/restic/restic/repository"
-	"runtime"
 )
 
 // Lock represents a process locking the repository for an operation.
@@ -195,21 +194,11 @@ func (l *Lock) Stale() bool {
 		return true
 	}
 
-	proc, err := os.FindProcess(l.PID)
-	if err != nil {
-		debug.Log("Lock.Stale", "error searching for process %d: %v\n", l.PID, err)
+	// check if we can reach the process retaining the lock
+	exists := l.processExists()
+	if !exists {
+		debug.Log("Lock.Stale", "could not reach process, %d, lock is probably stale\n", l.PID)
 		return true
-	}
-	defer proc.Release()
-
-	// Windows does not have SIGHUP
-	if runtime.GOOS != "windows" {
-		debug.Log("Lock.Stale", "sending SIGHUP to process %d\n", l.PID)
-		err = proc.Signal(syscall.SIGHUP)
-		if err != nil {
-			debug.Log("Lock.Stale", "signal error: %v, lock is probably stale\n", err)
-			return true
-		}
 	}
 
 	debug.Log("Lock.Stale", "lock not stale\n")
--- a/lock.go
+++ b/lock.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"os/signal"
 	"os/user"
-	"strconv"
 	"sync"
 	"syscall"
 	"time"
@@ -117,15 +116,8 @@ func (l *Lock) fillUserInfo() error {
 	}
 	l.Username = usr.Username
 
-	// We ignore the error. On Windows Uid is not a number
-	uid, _ := strconv.ParseInt(usr.Uid, 10, 32)
-	l.UID = uint32(uid)
-
-	// We ignore the error. On Windows Gid is not a number
-	gid, _ := strconv.ParseInt(usr.Gid, 10, 32)
-	l.GID = uint32(gid)
-
-	return nil
+	l.UID, l.GID, err = uidGidInt(*usr)
+	return err
 }
 
 // checkForOtherLocks looks for other locks that currently exist in the repository.
--- a/lock.go
+++ b/lock.go
@@ -13,6 +13,7 @@ import (
 	"github.com/restic/restic/backend"
 	"github.com/restic/restic/debug"
 	"github.com/restic/restic/repository"
+	"runtime"
 )
 
 // Lock represents a process locking the repository for an operation.
@@ -116,16 +117,12 @@ func (l *Lock) fillUserInfo() error {
 	}
 	l.Username = usr.Username
 
-	uid, err := strconv.ParseInt(usr.Uid, 10, 32)
-	if err != nil {
-		return err
-	}
+	// We ignore the error. On Windows Uid is not a number
+	uid, _ := strconv.ParseInt(usr.Uid, 10, 32)
 	l.UID = uint32(uid)
 
-	gid, err := strconv.ParseInt(usr.Gid, 10, 32)
-	if err != nil {
-		return err
-	}
+	// We ignore the error. On Windows Gid is not a number
+	gid, _ := strconv.ParseInt(usr.Gid, 10, 32)
 	l.GID = uint32(gid)
 
 	return nil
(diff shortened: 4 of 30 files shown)


backend/local/local.go
```diff
@@ -143,7 +147,16 @@ func (lb *localBlob) Finalize(t backend.Type, name string) error {
 		return err
 	}
 
-	return os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
+	// set file to readonly, except on Windows,
+	// otherwise deletion will fail.
+	if runtime.GOOS != "windows" {
+		err = os.Chmod(f, fi.Mode()&os.FileMode(^uint32(0222)))
```
fd0: I don't like that this is evaluated every time the code is called. What about adding a new function `setNewFileMode()` that does the `Chmod()` on Unix and nothing on Windows?

backend/sftp/sftp.go
```diff
@@ -28,6 +28,8 @@ type SFTP struct {
 	cmd *exec.Cmd
 }
 
+var sysProcAttr syscall.SysProcAttr
```
fd0: This solution works, but I don't like the added global state. What about having a function (like for the `Chmod()` above) that sets the right options? That's much clearer and you can directly see where the options in `cmd.SysProcAttr` come from, instead of having to dig through `init()` methods...

build.go
```diff
@@ -261,8 +262,14 @@ func main() {
 	args := []string{
 		"-tags", strings.Join(buildTags, " "),
 		"-ldflags", fmt.Sprintf(`-s -X main.version %q -X main.compiledAt %q`, version, compileTime),
-		"-o", "restic", "github.com/restic/restic/cmd/restic",
 	}
+	if runtime.GOOS != "windows" {
```
fd0: This could be much shorter, by setting the output file name depending on `GOOS` ;)

lock.go
```diff
@@ -116,16 +117,12 @@ func (l *Lock) fillUserInfo() error {
 	}
 	l.Username = usr.Username
 
-	uid, err := strconv.ParseInt(usr.Uid, 10, 32)
-	if err != nil {
-		return err
-	}
+	// We ignore the error. On Windows Uid is not a number
+	uid, _ := strconv.ParseInt(usr.Uid, 10, 32)
```
fd0: This should also have its own function that just does nothing on Windows, instead of ignoring the error for all platfoms. Maybe something like `fillTimes()` from `node.go`, e.g. for Linux: https://github.com/restic/restic/blob/master/node_linux.go#L21

lock.go
```diff
@@ -116,16 +117,12 @@ func (l *Lock) fillUserInfo() error {
...
 	l.UID = uint32(uid)
 
-	gid, err := strconv.ParseInt(usr.Gid, 10, 32)
-	if err != nil {
-		return err
-	}
+	// We ignore the error. On Windows Gid is not a number
+	gid, _ := strconv.ParseInt(usr.Gid, 10, 32)
```
fd0: Same here, I don't like globally ignoring the error.

lock_test.go
```diff
@@ -124,7 +124,7 @@ var staleLockTests = []struct {
 	{
 		timestamp: time.Now(),
 		stale:     true,
-		pid:       os.Getpid() + 500,
+		pid:       os.Getpid() + 500000,
```
fd0: What's the reason for this change?

lock.go
```diff
@@ -207,17 +204,20 @@ func (l *Lock) Stale() bool {
...
 	}
+	defer proc.Release()
 
-	debug.Log("Lock.Stale", "sending SIGHUP to process %d\n", l.PID)
-	err = proc.Signal(syscall.SIGHUP)
-	if err != nil {
-		debug.Log("Lock.Stale", "signal error: %v, lock is probably stale\n", err)
-		return true
+	// Windows does not have SIGHUP
+	if runtime.GOOS != "windows" {
```
fd0: Same as above: This is maybe better in a specific function.

node.go
```diff
@@ -381,9 +386,19 @@ func (node *Node) isNewer(path string, fi os.FileInfo) bool {
...
-	extendedStat := fi.Sys().(*syscall.Stat_t)
-	inode := extendedStat.Ino
-	size := uint64(extendedStat.Size)
+	size := uint64(fi.Size())
+
+	extendedStat, ok := toStatT(fi.Sys())
+	if !ok {
+		if node.ModTime != fi.ModTime() ||
+			node.Size != size {
+			debug.Log("node.isNewer", "node %v is newer: timestamp, size or inode changed", path)
```
fd0: The message is wrong: The inode isn't tested here, as far as I can see.

node.go
```diff
@@ -473,3 +488,32 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 
 	return err
 }
+
+type statT interface {
```
fd0: I like this interface :)

node_test.go
```diff
@@ -135,14 +138,16 @@ func TestNodeRestoreAt(t *testing.T) {
...
-		Assert(t, test.UID == n2.UID,
-			"%v: UID doesn't match (%v != %v)", test.Type, test.UID, n2.UID)
-		Assert(t, test.GID == n2.GID,
-			"%v: GID doesn't match (%v != %v)", test.Type, test.GID, n2.GID)
-
-		if test.Type != "symlink" {
-			Assert(t, test.Mode == n2.Mode,
-				"%v: mode doesn't match (%v != %v)", test.Type, test.Mode, n2.Mode)
+
+		if runtime.GOOS != "windows" {
```
fd0: Using `runtime.GOOS` in tests is perfectly fine, in my opinion.

node_unix.go
```diff
@@ -0,0 +1,34 @@
+//
```
fd0: The first two lines are unneeded as far as I can see.

node_unix.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"syscall"
+)
+
+var mknod = syscall.Mknod
+var lchown = os.Lchown
+
+type statUnix syscall.Stat_t
+
+func toStatT(i interface{}) (statT, bool) {
```
fd0: Nice solution! Makes the code much more readable!

node_windows.go
```diff
@@ -0,0 +1,66 @@
+package restic
+
+import (
+	"os"
+	"syscall"
+)
+
+func (node *Node) OpenForReading() (*os.File, error) {
+	file, err := os.OpenFile(node.path, os.O_RDONLY, 0)
```
fd0: This can be shortened, the second call to `os.OpenFile()` will most certainly fail exactly as the first.

node_windows.go
```diff
@@ -0,0 +1,66 @@
...
+		return os.OpenFile(node.path, os.O_RDONLY, 0)
+	}
+	return file, err
+}
+
+// mknod() creates a filesystem node (file, device
+// special file, or named pipe) named pathname, with attributes
+// specified by mode and dev.
+var mknod = func(path string, mode uint32, dev int) (err error) {
+	panic("mknod not implemented")
```
fd0: This mustn't panic, but return an error that device nodes can't be created on windows.

backend/local/local.go
klauspost: As a package level function pointer? That way we avoid the string compare and we don't have to do separate files. Seems the neatest way to get rid of it.

snapshot.go
```diff
@@ -76,16 +76,12 @@ func (sn *Snapshot) fillUserInfo() error {
 	}
 	sn.Username = usr.Username
 
-	uid, err := strconv.ParseInt(usr.Uid, 10, 32)
-	if err != nil {
-		return err
-	}
+	// We ignore the error. On Windows Uid is not a number
+	uid, _ := strconv.ParseInt(usr.Uid, 10, 32)
```
fd0: Same as above: Ignoring the error to cover one special case is not good.

fd0:
Okay, first round of comments done! I haven't had the time to actually try it, but it looks really well already. I especially like what you did with the interface for the stat!

What do you think about adding `User` and `Group` attributes to nodes for windows, and fill in the user and group strings there? It'd be a different attribute to `UID` and `GID`, so it doesn't conflict with the unix attributes...

build.go
klauspost: ?

backend/local/local.go
fd0: No, as a function. One in `local_windows.go` that does nothing, and one in `local_unix.go` (with appropiate build constraints of course) that does the `Chmod`.

lock_test.go
klauspost: See https://github.com/klauspost/restic/commit/3804bc7493711782d022164f9dc95b8e1c79a700

build.go
fd0: Like this:
```go
output := "restic"
    output = "restic.exe"
}

args = append(args, "-o", output, "github.com/restic/restic/cmd/restic")
```

lock_test.go
fd0: Ah, thanks.

klauspost:
In the long term, yes.

UID/GID: 0/0 is usually 'root' on Linux AFAIK, so I am not sure if setting it to 0/0 is a such a good idea long term.

Maybe if the type of the fields were changed from `uint32` to `int`, it could set 
to -1 to indicate "no value". That would allow restic to handle crossplatform backup/restores better.

fd0:
> UID/GID: 0/0 is usually 'root' on Linux AFAIK, so I am not sure if setting it to 0/0 is a such a good idea long term.

We could also change the field in the struct to `*uint32`, so we have the special value `null` to signal that this information isn't available.

> Maybe if the type of the fields were changed from uint32 to int, it could set
> to -1 to indicate "no value". That would allow restic to handle crossplatform backup/restores better.

I don't like that solution, abusing the special value -1.

Another solution would be a combination, like having `UID *uint32` with a tag `json:"uid,omitempty"` and a second field `User *string` with a tag of `json:"user,omitempty"`.

This should make it possible to distinguish all possible situations without having to abuse special values ;)

node_windows.go
klauspost: Maybe this should actually just return nil, so it will be possible to restore snapshots from other OS'es to Windows.

klauspost:
okie dokie! 

I have changed the things you requested. I found #260 which I suspect isn't intentional, but I didn't want to fix it in this PR.

fd0:
Good idea, I think I've fixed this in #261

fd0:
LGTM! :+1:

fd0:
Hm, I get a bunch of errors for the tests.

I've checked out your branch to `c:\gopath\src\github.com\restic\restic` and used the following batch file:

    SET GOPATH=\gopath;\gopath\src\github.com\restic\restic\Godeps\_workspace
    go test ./...
    pause

These are the errors when running `run_tests.bat` from `cmd.exe`:

    C:\gopath\src\github.com\restic\restic>SET GOPATH=\gopath;\gopath\src\github.com\restic\restic\Godeps\_workspace

    C:\gopath\src\github.com\restic\restic>go test ./...
    ←[31mcache_test.go:15: unexpected error: unable to locate cache directory (XDG_CACHE_HOME and HOME unset)←[39m

    --- FAIL: TestCache (1.80s)
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic        40.673s
    ok      _/C_/gopath/src/github.com/restic/restic/backend        0.407s
    ?       _/C_/gopath/src/github.com/restic/restic/backend/local  [no test files]
    ?       _/C_/gopath/src/github.com/restic/restic/backend/s3     [no test files]
    ?       _/C_/gopath/src/github.com/restic/restic/backend/sftp   [no test files]
    ←[31mhelpers.go:93: unexpected error: exec: "sh": executable file not found in %PATH%←[39m

    --- FAIL: TestCheckRepo (0.01s)
    ←[31mhelpers.go:93: unexpected error: exec: "sh": executable file not found in %PATH%←[39m

    --- FAIL: TestMissingPack (0.00s)
    ←[31mhelpers.go:93: unexpected error: exec: "sh": executable file not found in %PATH%←[39m

    --- FAIL: TestUnreferencedPack (0.00s)
    ←[31mhelpers.go:93: unexpected error: exec: "sh": executable file not found in %PATH%←[39m

    --- FAIL: TestUnreferencedBlobs (0.00s)
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/checker        0.051s
    # _/C_/gopath/src/github.com/restic/restic/cmd/restic
    cmd\restic\integration_fuse_test.go:56: undefined: CmdMount
    cmd\restic\integration_helpers_test.go:89: undefined: syscall.Stat_t
    cmd\restic\integration_helpers_test.go:90: undefined: syscall.Stat_t
    FAIL    _/C_/gopath/src/github.com/restic/restic/cmd/restic [build failed]
    ←[31mcrypto_test.go:54: unexpected error: open /dev/urandom: The system cannot find the path specified.←[39m

    --- FAIL: TestSmallBuffer (0.00s)
    ←[31mcrypto_test.go:79: unexpected error: open /dev/urandom: The system cannot find the path specified.←[39m

    --- FAIL: TestSameBuffer (0.00s)
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/crypto 1.208s
    ?       _/C_/gopath/src/github.com/restic/restic/debug  [no test files]
    --- FAIL: TestMatch (0.00s)
            filter_test.go:87: test 6: filter.Match("foo*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 7: filter.Match("bar*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 9: filter.Match("bar/*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 11: filter.Match("bar/test.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 12: filter.Match("bar/*.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 13: filter.Match("ba*/*.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 15: filter.Match("test.*", "/foo/bar/test.go"):expected true, got false
            filter_test.go:87: test 17: filter.Match("bar/*", "/foo/bar/baz"): expected true, got false
            filter_test.go:87: test 18: filter.Match("bar", "/foo/bar"): expected true, got false
            filter_test.go:87: test 19: filter.Match("bar", "/foo/bar/baz"): expected true, got false
            filter_test.go:87: test 20: filter.Match("bar", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 21: filter.Match("/foo/*test.*", "/foo/bar/test.go"): expected false, got true
            filter_test.go:87: test 32: filter.Match("baz/test.*", "bar/baz/test.go"): expected true, got false
            filter_test.go:87: test 34: filter.Match("test.go", "bar/baz/test.go"):expected true, got false
            filter_test.go:87: test 37: filter.Match("sdk", "/foo/bar/sdk"): expected true, got false
            filter_test.go:87: test 38: filter.Match("sdk", "/foo/bar/sdk/test/sdk_foo.go"): expected true, got false
            filter_test.go:87: test 40: filter.Match("foo/**/bar/*.go", "/home/user/foo/work/special/project/bar/test.go"): expected true, got false
            filter_test.go:87: test 41: filter.Match("foo/**/bar/*.go", "/home/user/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 42: filter.Match("foo/**/bar/*.go", "x/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 43: filter.Match("foo/**/bar/*.go", "foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 46: filter.Match("foo/**/bar/*.go", "/foo/bar/main.go"): expected true, got false
            filter_test.go:87: test 48: filter.Match("foo/**/bar", "/home/user/foo/x/y/bar"): expected true, got false
            filter_test.go:87: test 49: filter.Match("foo/**/bar", "/home/user/foo/x/y/bar/main.go"): expected true, got false
            filter_test.go:87: test 51: filter.Match("user/**/hidden*/**/c", "/home/user/work/x/y/hidden/z/a/b/c"): expected true, got false
    --- FAIL: TestMatchList (0.00s)
            filter_test.go:133: test 6: filter.MatchList(["?" "x"], "/foo/bar/x"): expected true, got false
    --- FAIL: TestFilterPatternsFile (0.24s)
            filter_test.go:194: wrong number of hits for pattern "sdk": want 22186,got 0
            filter_test.go:194: wrong number of hits for pattern "sdk/*/cpp/*/*vars.html": want 3, got 0
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/filter 0.273s
    # bazil.org/fuse
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\error_std.go:27: undefined: errNoXattr
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse.go:1277: undefined: attr
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse_kernel.go:171: undefined: syscall.O_ACCMODE
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse_kernel.go:397: undefined: attr
    FAIL    _/C_/gopath/src/github.com/restic/restic/fuse [build failed]
    ok      _/C_/gopath/src/github.com/restic/restic/pack   0.042s
    ok      _/C_/gopath/src/github.com/restic/restic/pipe   0.054s
    ←[31mhelpers.go:93: unexpected error: exec: "sh": executable file not found in %PATH%←[39m

    --- FAIL: TestLoadIndex (0.00s)
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/repository     15.624s
    ?       _/C_/gopath/src/github.com/restic/restic/test   [no test files]

    C:\gopath\src\github.com\restic\restic>pause
    Press any key to continue . . .

And these when run from within the cygwin shell:

    $ ./run_tests.bat

    C:\gopath\src\github.com\restic\restic>SET GOPATH=\gopath;\gopath\src\github.com                                                                             \restic\restic\Godeps\_workspace

    C:\gopath\src\github.com\restic\restic>go test ./...
    ok      _/C_/gopath/src/github.com/restic/restic        42.176s
    ok      _/C_/gopath/src/github.com/restic/restic/backend        0.402s
    ?       _/C_/gopath/src/github.com/restic/restic/backend/local  [no test files]
    ?       _/C_/gopath/src/github.com/restic/restic/backend/s3     [no test files]
    ?       _/C_/gopath/src/github.com/restic/restic/backend/sftp   [no test files]
    ok      _/C_/gopath/src/github.com/restic/restic/checker        7.781s
    # _/C_/gopath/src/github.com/restic/restic/cmd/restic
    cmd\restic\integration_fuse_test.go:56: undefined: CmdMount
    cmd\restic\integration_helpers_test.go:89: undefined: syscall.Stat_t
    cmd\restic\integration_helpers_test.go:90: undefined: syscall.Stat_t
    FAIL    _/C_/gopath/src/github.com/restic/restic/cmd/restic [build failed]
    crypto_test.go:54: unexpected error: open /dev/urandom: The system cannot find the path specified.

    --- FAIL: TestSmallBuffer (0.00s)
    crypto_test.go:79: unexpected error: open /dev/urandom: The system cannot find the path specified.

    --- FAIL: TestSameBuffer (0.00s)
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/crypto 1.247s
    ?       _/C_/gopath/src/github.com/restic/restic/debug  [no test files]
    --- FAIL: TestMatch (0.00s)
            filter_test.go:87: test 6: filter.Match("foo*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 7: filter.Match("bar*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 9: filter.Match("bar/*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 11: filter.Match("bar/test.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 12: filter.Match("bar/*.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 13: filter.Match("ba*/*.go", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 15: filter.Match("test.*", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 17: filter.Match("bar/*", "/foo/bar/baz"): expected true, got false
            filter_test.go:87: test 18: filter.Match("bar", "/foo/bar"): expected true, got false
            filter_test.go:87: test 19: filter.Match("bar", "/foo/bar/baz"): expected true, got false
            filter_test.go:87: test 20: filter.Match("bar", "/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 21: filter.Match("/foo/*test.*", "/foo/bar/test.go"): expected false, got true
            filter_test.go:87: test 32: filter.Match("baz/test.*", "bar/baz/test.go"): expected true, got false
            filter_test.go:87: test 34: filter.Match("test.go", "bar/baz/test.go"): expected true, got false
            filter_test.go:87: test 37: filter.Match("sdk", "/foo/bar/sdk"): expected true, got false
            filter_test.go:87: test 38: filter.Match("sdk", "/foo/bar/sdk/test/sdk_foo.go"): expected true, got false
            filter_test.go:87: test 40: filter.Match("foo/**/bar/*.go", "/home/user/foo/work/special/project/bar/test.go"): expected true, got false
            filter_test.go:87: test 41: filter.Match("foo/**/bar/*.go", "/home/user/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 42: filter.Match("foo/**/bar/*.go", "x/foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 43: filter.Match("foo/**/bar/*.go", "foo/bar/test.go"): expected true, got false
            filter_test.go:87: test 46: filter.Match("foo/**/bar/*.go", "/foo/bar/main.go"): expected true, got false
            filter_test.go:87: test 48: filter.Match("foo/**/bar", "/home/user/foo/x/y/bar"): expected true, got false
            filter_test.go:87: test 49: filter.Match("foo/**/bar", "/home/user/foo/x/y/bar/main.go"): expected true, got false
            filter_test.go:87: test 51: filter.Match("user/**/hidden*/**/c", "/home/user/work/x/y/hidden/z/a/b/c"): expected true, got false
    --- FAIL: TestMatchList (0.00s)
            filter_test.go:133: test 6: filter.MatchList(["?" "x"], "/foo/bar/x"): expected true, got false
    --- FAIL: TestFilterPatternsFile (0.32s)
            filter_test.go:194: wrong number of hits for pattern "sdk": want 22186, got 0
            filter_test.go:194: wrong number of hits for pattern "sdk/*/cpp/*/*vars.html": want 3, got 0
    FAIL
    FAIL    _/C_/gopath/src/github.com/restic/restic/filter 0.352s
    # bazil.org/fuse
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\error_std.go:27: undefined: errNoXattr
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse.go:1277: undefined: attr
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse_kernel.go:171: undefined: syscall.O_ACCMODE
    \gopath\src\github.com\restic\restic\Godeps\_workspace\src\bazil.org\fuse\fuse_kernel.go:397: undefined: attr
    FAIL    _/C_/gopath/src/github.com/restic/restic/fuse [build failed]
    ok      _/C_/gopath/src/github.com/restic/restic/pack   0.043s
    ok      _/C_/gopath/src/github.com/restic/restic/pipe   0.041s
    ok      _/C_/gopath/src/github.com/restic/restic/repository     17.674s
    ?       _/C_/gopath/src/github.com/restic/restic/test   [no test files]

    C:\gopath\src\github.com\restic\restic>pause
    Press any key to continue . . .

When going throught the test fails, this is what we need to do in order to make them pass:

 * Figure out what the cache location on Windows should be when `XDG_CACHE_HOME` isn't set
 * Find out why there's no `sh` binary although cygwin is installed
 * Remove `/dev/urandom` from the crypto tests, maybe use `rand.Reader` instead (I think there already is code for that somewhere)
 * Replace `/` with `\` in the filter tests on Windows before running the filters
 * Insert build constraints for all fuse related stuff (maybe also applies to openbsd)

This seems to be quite some stuff, but I think we're on a good track and just need to fine-tune everything. :)

Did the tests pass on your system?

klauspost:
I just tested main dir and building the server.

>Figure out what the cache location on Windows should be when XDG_CACHE_HOME isn't set

I assume the cache location is permanent, so %APP_DATA% is probably the way to go.

>Find out why there's no sh binary although cygwin is installed

Does sftp rely on it? In that case, I would suggest to move to using a [library that doesn't](https://godoc.org/github.com/pkg/sftp). Though for now I would just disable it.

I don't have any sh binary either, 
>Remove /dev/urandom from the crypto tests, maybe use rand.Reader instead (I think there already is code for that somewhere)

Yeah - should be an easy fix.

> Replace / with \ in the filter tests on Windows before running the filters

Yes, there is "os.PathSeparator", but paths can have both `/` and `\`, so both should be supported, or everything should be `/`, and be converted to that, so there is just one on-disk format.

>Insert build constraints for all fuse related stuff (maybe also applies to openbsd)

It is only the dependencies that are failing. The command

fd0:
I think your comment was cut short.

The `sh` binary is just used in the tests, that should be easy to fix. The sftp library you're suggesting doesn't depend on `sh` and is precisely that restic already uses.

> Yes, there is "os.PathSeparator", but paths can have both / and \, so both should be supported, or everything should be /, and be converted to that, so there is just one on-disk format.

That's what I meant.

I think we can merge this code and have a second PR that fixes all the test related stuff. What do you think?

klauspost:
Yes - they aren't affected by this, and it compiles and runs basic operations fine.

klauspost:
I menat to say "It is only the dependencies that are failing. The mount command is disabled, and doesn't import it on Windows". It is annoying that FUSE gives compile errors on Windows, but I am not sure of a good way to fix it on our side.

fd0:
The compile errors for fuse are easily explained: The files in `fuse/` don't have a build constraint for `!windows` and import the fuse lib, so the lib is compiled and fails.

fd0:
I've pushed a few fixes to #262.

klauspost:
I think with the latest changes to this and #262, we should have fixed all tests.

fd0:
I'll have a look in the evening today.

fw42:
Is travis gonna run any windows tests? Or at least cross-compile?

fd0:
We can add back cross-compiling, but I doubt travis offers any windows-based tests.

fw42:
so now we have windows support but no way for non-windows developers to run the windows tests? that kinda sucks :-(

fd0:
Do you have an idea to solve this? I've tried using vagrant yesterday, with limited success to be honest.

fd0:
Any experiences with appveyor?

PR #261, opened 2015-08-16
locks: fix testing stale locks created on other hosts

### LINKED ISSUE #260: Lock.Stale doesn't operate as advertised
In the [Lock.Stale](https://github.com/restic/restic/blob/master/lock.go#L199) function the documentation says:

>A lock is stale if [...] it was created on the current machine and the process isn't alive any more.

However, I see no tests in the code to check if it was created on the current machine. Instead it seems like it will always return as stale if it was created on another machine.

This should probably be added:
```Go
	// check if lock was created on this host
	hn, err := os.Hostname()
	if err != nil {
		debug.Log("Lock.Stale", "could not retrieve current host name: %v\n", err)
	}

	if l.Hostname != hn {
		debug.Log("Lock.Stale", "lock created on different host, probably not stale\n")
	}
```

fd0:
Closes #260

Blocked by #259

### DIFF
--- a/lock.go
+++ b/lock.go
@@ -206,6 +206,19 @@ func (l *Lock) Stale() bool {
 		return true
 	}
 
+	hn, err := os.Hostname()
+	if err != nil {
+		debug.Log("Lock.Stale", "unable to find current hostnanme: %v", err)
+		// since we cannot find the current hostname, assume that the lock is
+		// not stale.
+		return false
+	}
+
+	if hn != l.Hostname {
+		// lock was created on a different host, assume the lock is not stale.
+		return false
+	}
+
 	proc, err := os.FindProcess(l.PID)
 	defer proc.Release()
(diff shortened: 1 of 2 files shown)


klauspost:
Could you wait merging this until #259 lands? Re-basing it not my strongest point, and this is very likely to conflict.

fd0:
Yes, that's certainly possible. :)

What do you think about the code changes and the additional tests?

fd0:
When commenting changes in a pull request, make sure that you click on the "Files changed" tab before adding a comment. Otherwise, you end up adding the comment to a specific commit (fbe746c26124858b286d04a5fca3145ff867d8d1 in your case), which vanishes if the commit is reworked before merging this PR.

I don't know why GitHub's user interface makes it so easy to add comments to commits which might vanish that way. :(

PR #262, opened 2015-08-16
Fix tests

fd0:
This PR includes a few fixes suggested in https://github.com/restic/restic/pull/259#issuecomment-131579346

Connects #259

### DIFF
--- a/fuse/dir.go
+++ b/fuse/dir.go
@@ -1,3 +1,5 @@
+// +build !openbsd
+
 package fuse
 
 import (
--- a/fuse/dir.go
+++ b/fuse/dir.go
@@ -1,4 +1,5 @@
 // +build !openbsd
+// +build !windows
 
 package fuse
 
--- a/test/helpers.go
+++ b/test/helpers.go
@@ -88,19 +88,18 @@ func RandomReader(seed, size int) *bytes.Reader {
 
 // SetupTarTestFixture extracts the tarFile to outputDir.
 func SetupTarTestFixture(t testing.TB, outputDir, tarFile string) {
-	err := System("sh", "-c", `(cd "$1" && tar xzf - ) < "$2"`,
-		"sh", outputDir, tarFile)
+	f, err := os.Open(tarFile)
+	defer f.Close()
 	OK(t, err)
-}
 
-// System runs the command and returns the exit code. Output is passed on to
-// stdout/stderr.
-func System(command string, args ...string) error {
-	cmd := exec.Command(command, args...)
+	cmd := exec.Command("tar", "xzf", "-")
+	cmd.Dir = outputDir
+
+	cmd.Stdin = f
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
-	return cmd.Run()
+	OK(t, cmd.Run())
 }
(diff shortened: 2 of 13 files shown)


fuse/dir.go
```diff
@@ -1,3 +1,5 @@
+// +build !openbsd
```
klauspost: Adding `// +build !windows` (as a separate line) will do the same for Windows. So while you are at it we might add it as well.

test/helpers.go
```diff
@@ -88,19 +88,18 @@ func RandomReader(seed, size int) *bytes.Reader {
...
+	f, err := os.Open(tarFile)
+	defer f.Close()
 	OK(t, err)
-}
 
-// System runs the command and returns the exit code. Output is passed on to
-// stdout/stderr.
-func System(command string, args ...string) error {
-	cmd := exec.Command(command, args...)
+	cmd := exec.Command("tar", "xzf", "-")
```
klauspost: Could we use archive/tar instead?

fuse/dir.go
fd0: Yes, that's the plan.

test/helpers.go
fd0: I'm using tar files for tests in two cases:
* Providing sample repositories e.g. to test the check functions
* Provide test data to backup

The latter case includes symlinks, hard links, device nodes etc., all that must be extracted from the tar and properly set up so that the tests can run. I think it's better to rely on an external `tar` for this, at least for now.

klauspost: No problem. The only issue I am seeing is that os.RemoveAll(tempdir) fails because the tar file contains readonly files, which Windows cannot delete due to https://github.com/golang/go/issues/9606

klauspost: There is some workaround code here:

https://github.com/golang/go/blob/3b7841b3aff9204f054ffabbe4dd39d3e3dd3e91/src/cmd/go/go_test.go#L557

klauspost: ... and here it is, so it can be included in helpers.go

```Go
// resetReadOnlyFlagAll resets windows read-only flag
// set on path and any children it contains.
// The flag is set by git and has to be removed.
// os.Remove refuses to remove files with read-only flag set.
func resetReadOnlyFlagAll(t testing.TB, path string) {
	fi, err := os.Stat(path)
	if err != nil {
		t.Fatalf("resetReadOnlyFlagAll(%q) failed: %v", path, err)
	}
	if !fi.IsDir() {
		err := os.Chmod(path, 0666)
		if err != nil {
			t.Fatalf("resetReadOnlyFlagAll(%q) failed: %v", path, err)
		}
	}
	fd, err := os.Open(path)
	if err != nil {
		t.Fatalf("resetReadOnlyFlagAll(%q) failed: %v", path, err)
	}
	defer fd.Close()
	names, _ := fd.Readdirnames(-1)
	for _, name := range names {
		resetReadOnlyFlagAll(t, path+string(filepath.Separator)+name)
	}
}
```

If I insert `resetReadOnlyFlagAll(t, tempdir)` right after `	SetupTarTestFixture(t, tempdir, repoFixture)` it helps a lot.

Issue #263, opened 2015-08-17
Add rate limit to progress updates.


rubenv:
Might want to rate-limit the progress updates (more than once every 16.6 ms would be > 60fps and thus pointless): https://youtu.be/AuYHXvgwKE8

rubenv:
This hunger for redraws causes this:

![screen shot 2015-08-17 at 14 20 04](https://cloud.githubusercontent.com/assets/42904/9304698/233e4d12-44eb-11e5-838b-a9d6e5b7faf7.png)

Which in turn means that restic becomes ridiculously slow, because CPU cycles are wasted elsewhere :-)

fd0:
Indeed, that's a good idea.

rubenv:
Mind you, this is mostly a problem because the terminal in OSX is somewhat crappy

But anything faster than what we can see isn't really needed either

fd0:
I agree, but since the OSX terminal is widely in use, I consider this a bug. Adding a rate limit isn't that hard to do.

fd0:
This has been fixed in #266.

PR #266, opened 2015-08-18
Update the progress status with the ticker

bchapuis:
Apparently, the progress reporter already uses a ticker initialized at 1second, however, the updateProgress function is called every time new stats are added. I just tried to update the progress with the ticker only (60fps), to see if it has an effect on cpu usage on mac OS X. I may miss some points but it doesn't looks like the root cause of issue #263.

### DIFF
--- a/progress.go
+++ b/progress.go
@@ -6,19 +6,22 @@ import (
 	"time"
 )
 
+const minTickerTime = time.Second / 60
+
 type Progress struct {
 	OnStart  func()
 	OnUpdate ProgressFunc
 	OnDone   ProgressFunc
 	fnM      sync.Mutex
 
-	cur    Stat
-	curM   sync.Mutex
-	start  time.Time
-	c      *time.Ticker
-	cancel chan struct{}
-	o      sync.Once
-	d      time.Duration
+	cur        Stat
+	curM       sync.Mutex
+	start      time.Time
+	c          *time.Ticker
+	cancel     chan struct{}
+	o          sync.Once
+	d          time.Duration
+	lastUpdate time.Time
 
 	running bool
 }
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for trying to fix this! Unfortunately, this isn't a good fix. It updates the status 60 times per second although there may be nothing to update. I'd suggest leaving the ticker at a second (so that the progress is updated at least once per second). Then add a new field `lastUpdate time.Time` to the `Progress` struct, check this in `Report()` and only call `updateProgress()` when since the last update more than 16ms have passed. e.g.:

```go
func (p *Progress) Report(s Stat) {
	[...]
	needUpdate := false

	p.curM.Lock()
	p.cur.Add(s)
	cur := p.cur
	if time.Since(p.lastUpdate) > 16 * time.Millisecond {
		p.lastUpdate = time.Now()
		needUpdate = true
	}
	p.curM.Unlock()

	if needUpdate {
		p.updateProgress(cur, false)
	}
}
```

bchapuis:
Ok, your code worked as it is :)

fd0:
Cool, I didn't expect that. I'll merge this later today. If you have the time, you could squash both commits.

Maybe @rubenv can give this code a try.

rubenv:
Looks like a perfect fix. CPU usage is back where it should be: within restic.

LGTM

progress.go
```diff
@@ -92,9 +95,17 @@ func (p *Progress) Report(s Stat) {
 	p.curM.Lock()
 	p.cur.Add(s)
 	cur := p.cur
+	needUpdate := false
+	if time.Since(p.lastUpdate) > minTickerTime {
```
fw42: how about extracting this into a new method?
```
func (p *Progress) needUpdate() { ... }
```
I think that would improve readability. And then use `if p.needUpdate() { ... }` below.

fw42: also right now this code that you are adding is mutexed by `curM`. Is that intentional/necessary? I don't see why.

fd0: That is intentional. `Report()` is called from multiple goroutines all read (and update) `p.lastUpdate`.

PR #267, opened 2015-08-18
test the GetReader() function

### LINKED ISSUE #254: backend tests: Add test for GetReader
At the moment, the `GetReader()` function for the backends isn't used in any tests at all. Implement a test that tests the `GetReader()` function.

bchapuis:
Added tests for the GetReader() function in order to fix #254.

### DIFF
--- a/backend/backend_test.go
+++ b/backend/backend_test.go
@@ -57,9 +57,20 @@ func testBackend(b backend.Backend, t *testing.T) {
 			Assert(t, rd != nil, "Get() returned nil")
 
 			// try to read it out again
-			r, err := b.GetReader(tpe, test.id, 0, uint(len(test.data)))
+			reader, err := b.GetReader(tpe, test.id, 0, uint(len(test.data)))
 			OK(t, err)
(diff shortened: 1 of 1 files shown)


backend/backend_test.go
```diff
@@ -52,6 +56,11 @@ func testBackend(b backend.Backend, t *testing.T) {
 			OK(t, err)
 			Assert(t, rd != nil, "Get() returned nil")
 
+			// try to read it out again
+			r, err := b.GetReader(tpe, test.id, 0, uint(len(test.data)))
+			OK(t, err)
+			Assert(t, r != nil, "GetReader() returned nil")
```
fw42: is that all we care about? that it's not `nil`? don't we also want to assert that you can actually use this reader to read things?

fd0:
Heh, just as I read what @fw42 commented, @bchapuis pushed another commit. Keep up the great work! :+1:

bchapuis:
Agree it was a bit minimalistic :) Wonder if the close method must be tested as well. @fw42 What do you think?

PR #268, opened 2015-08-18
CI: Add Windows

fd0:
Add windows to the list of cross compilation OS.

### DIFF
--- a/lock_test.go
+++ b/lock_test.go
@@ -134,7 +134,7 @@ var staleLockTests = []struct {
 		timestamp:        time.Now(),
 		stale:            true,
 		staleOnOtherHost: false,
-		pid:              os.Getpid() + 500,
+		pid:              os.Getpid() + 500000,
 	},
 }
 
@@ -204,7 +204,7 @@ func TestRemoveAllLocks(t *testing.T) {
 	id2, err := createFakeLock(repo, time.Now().Add(-time.Minute), os.Getpid())
 	OK(t, err)
 
-	id3, err := createFakeLock(repo, time.Now().Add(-time.Minute), os.Getpid()+500)
+	id3, err := createFakeLock(repo, time.Now().Add(-time.Minute), os.Getpid()+500000)
 	OK(t, err)
 
 	OK(t, restic.RemoveAllLocks(repo))
--- a/run_integration_tests.go
+++ b/run_integration_tests.go
@@ -77,6 +77,19 @@ func (env *TravisEnvironment) RunTests() {
 	runGofmt()
 }
 
+type AppveyorEnvironment struct{}
+
+func (env *AppveyorEnvironment) Prepare() {
+	msg("preparing environment for Appveyor CI\n")
+
+	// install tar, gzip, bzip2
+}
+
+func (env *AppveyorEnvironment) RunTests() {
+	// run the build script and the tests
+	run("go", "run", "build.go", "-v", "-T")
+}
+
 // findGoFiles returns a list of go source code file names below dir.
 func findGoFiles(dir string) (list []string, err error) {
 	err = filepath.Walk(dir, func(name string, fi os.FileInfo, err error) error {
@@ -153,12 +166,18 @@ func isTravis() bool {
 	return os.Getenv("TRAVIS_BUILD_DIR") != ""
 }
 
+func isAppveyor() bool {
+	return runtime.GOOS == "windows"
+}
+
 func main() {
 	var env CIEnvironment
 
 	switch {
 	case isTravis():
 		env = &TravisEnvironment{}
+	case isAppveyor():
+		env = &AppveyorEnvironment{}
 	default:
 		fmt.Fprintln(os.Stderr, "unknown CI environment")
 		os.Exit(1)
--- a/run_integration_tests.go
+++ b/run_integration_tests.go
@@ -30,7 +30,7 @@ func (env *TravisEnvironment) Prepare() {
 	if runtime.GOOS == "darwin" {
 		// install the libraries necessary for fuse
 		run("brew", "install", "caskroom/cask/brew-cask")
-		run("brewcask", "install", "osxfuse")
+		run("brew", "cask", "install", "osxfuse")
 	}
 
 	// only test cross compilation on linux with Travis
--- a/run_integration_tests.go
+++ b/run_integration_tests.go
@@ -22,7 +22,7 @@ type TravisEnvironment struct {
 }
 
 func (env *TravisEnvironment) Prepare() {
-	fmt.Printf("preparing environment for Travis CI\n")
+	msg("preparing environment for Travis CI\n")
 
 	run("go", "get", "github.com/mattn/goveralls")
 	run("go", "get", "github.com/mitchellh/gox")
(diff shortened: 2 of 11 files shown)


fd0:
Ok, that's enough for this PR. No we're cross-compiling for Windows on Travis and we're running the integration tests on Windows with Appveyor.

lock_test.go
```diff
@@ -134,7 +134,7 @@ var staleLockTests = []struct {
 		timestamp:        time.Now(),
 		stale:            true,
 		staleOnOtherHost: false,
-		pid:              os.Getpid() + 500,
+		pid:              os.Getpid() + 500000,
```
fw42: what's this?

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+type CIEnvironment interface {
```
fw42: I think the Go naming convention would prefer this to be called "CiEnvironment"? (lowercase i)

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+	RunTests()
+}
+
+type TravisEnvironment struct {
+	goxArch []string
+	goxOS   []string
+}
+
+func (env *TravisEnvironment) Prepare() {
+	msg("preparing environment for Travis CI\n")
```
fw42: I don't see anything specific to Travis in either this function nor the other one.. How about we do a `s/TravisEnvironment/UnixTestEnvironment/g` on this whole file?

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+			command, strings.Join(args, " "), err)
+		os.Exit(3)
+	}
+}
+
+func isTravis() bool {
+	return os.Getenv("TRAVIS_BUILD_DIR") != ""
+}
+
+func isAppveyor() bool {
```
fw42: same reasoning here... this stuff has actually nothing to do with Appveyor.. I would `s/AppveyorEnvironment/WindowsTestEnvironment/g` since this code could just as well run on any other windows system.

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+	return runtime.GOOS == "windows"
+}
+
+func main() {
+	var env CIEnvironment
+
+	switch {
```
fw42: I think instead of this switch, I would prefer to split this whole file into two files: `run_integration_tests_windows.go` and `run_integration_tests_unix.go`. Then you could get of all uses of `runtime.GOOS`.

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+type AppveyorEnvironment struct{}
+
+func (env *AppveyorEnvironment) Prepare() {
+	msg("preparing environment for Appveyor CI\n")
+
+	// install tar, gzip, bzip2
+}
+
+func (env *AppveyorEnvironment) RunTests() {
+	// run the build script and the tests
```
fw42: I think the name of the function is obvious enough to make this comment unnecessary

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+
+	runGofmt()
+}
+
+type AppveyorEnvironment struct{}
+
+func (env *AppveyorEnvironment) Prepare() {
+	msg("preparing environment for Appveyor CI\n")
+
+	// install tar, gzip, bzip2
```
fw42: what's this? is there something missing here? why are tests currently passing if this is missing?

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (env *TravisEnvironment) RunTests() {
+	// run fuse tests on darwin
+	if runtime.GOOS != "darwin" {
+		msg("skip fuse integration tests on %v\n", runtime.GOOS)
+		os.Setenv("RESTIC_TEST_FUSE", "0")
+	}
+
+	// compile for all target architectures with tags
```
fw42: I would get rid of all the comments in this function and instead break the code out into smaller functions, i.e. `runFuseTests()`, `compileAllTargetArchitectures()`, `runBuildScript()`, and `gatherCoverageInformation()`. I think that would be more readable.

fw42:
I think this PR is a great idea, but conceptually I think you are tied too much to the idea of a CI environment. imho there should be no mention of travis or Appveyor anywhere in the codebase (other than in the yml files that configure those environments).

Also I think it would be great if the conventional way to run all of the restic tests locally would use the exact same code as CI uses (so the runner code itself should work locally just as well as it does on the CI server).

Might even add "go run run_integration_tests.go" to the Makefile so that "make test" does the same as CI.

run_integration_tests.go
```diff
@@ -0,0 +1,189 @@
...
+	case isTravis():
+		env = &TravisEnvironment{}
+	case isAppveyor():
+		env = &AppveyorEnvironment{}
+	default:
+		fmt.Fprintln(os.Stderr, "unknown CI environment")
+		os.Exit(1)
+	}
+
+	for _, f := range []func(){env.Prepare, env.RunTests} {
```
fw42: how about making those two things (prepare and run) two separate commands (using a command line argument switch or something)? then you could do those two steps separately in the CI environment and at least make use of the separate output grouping etc., and it would also make it easier to run only one but not the other (maybe I want to run the tests locally, but I don't need to prepare locally).

lock_test.go
fd0: The correction of a merge error of the following commit:
https://github.com/restic/restic/commit/3804bc7493711782d022164f9dc95b8e1c79a700

I've also stumbled upon this in https://github.com/restic/restic/pull/259#discussion_r37098341

run_integration_tests.go
fd0: Nop: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

run_integration_tests.go
fd0: Whoops, the comment is wrong. `tar` is installed by the `appveyor.yml` file, and `gzip` and `bzip2` aren't needed any more (switched to the Go stdlib implementations).

run_integration_tests.go
fd0: That's not easily possible:
 * GOOS is used to decide whether to call `brew` and install the osx fuse libs and to decide whether to use gox to cross-compile
 * The file `run_integration_tests.go` has a build constraint of `ignore`, so it ignored by `go build` and run directly with `go run`. The latter makes it hard to split functions into different files, it would need to be called with `go run run_integration_tests.go foo_unix.go` etc. That's really ugly, I prefer just checking for `runtim.GOOS` over that.

run_integration_tests.go
fd0: Agreed.

run_integration_tests.go
fd0: I've thought about doing that, but in the first step I decided to just move all that cruft from the travis config file to the Go program. Maybe we can find a nice way to separate this later.

fd0:
Good idea in general, but let's tackle that in a second PR and merge this for now.

PR #269, opened 2015-08-19
Allow building without git

fw42:
Right now, you can't build restic unless the source that you have is a git repository. That requirement effectively breaks the "Download ZIP" button on the GitHub repository page. This PR fixes that (by not linking a version number at all if one can't be determined).

@fd0


fd0:
LGTM :+1:

PR #281, opened 2015-08-25
build.go: use new combined version string

fd0:
Previously, when a VERSION file exists it takes precendence over the git version. This is unfortunate because all restic binaries compiled from a git checkout will just identify as the latest release (e.g. '0.1.0'), regardeless of any commits on top of it.

This commit adds a combined version string by using the contents of the VERSION file, and append the current git version returned by `git describe` if available, e.g.:

```
0.1.0 (v0.1.0-6-gb188217-dirty).
```

### DIFF
--- a/build.go
+++ b/build.go
@@ -200,35 +200,41 @@ func test(gopath string, args ...string) error {
 	return cmd.Run()
 }
 
+// getVersion returns the version string from the file VERSION in the current
+// directory.
+func getVersionFromFile() string {
+	buf, err := ioutil.ReadFile("VERSION")
+	if err != nil {
+		verbosePrintf("error reading file VERSION: %v\n", err)
+		return ""
+	}
+
+	return strings.TrimSpace(string(buf))
+}
+
 // getVersion returns a version string which is a combination of the contents
 // of the file VERSION in the current directory and the version from git (if
 // available).
 func getVersion() string {
-	buf, err := ioutil.ReadFile("VERSION")
-	if err != nil {
-		verbosePrintf("error reading file VERSION: %v\n", err)
-	}
-	versionFile := strings.TrimSpace(string(buf))
-	verbosePrintf("version from file 'VERSION' is %q\n", versionFile)
+	versionFile := getVersionFromFile()
+	versionGit := getVersionFromGit()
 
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -200,17 +200,30 @@ func test(gopath string, args ...string) error {
...
+// available).
 func getVersion() string {
-	v, err := ioutil.ReadFile("VERSION")
-	version := strings.TrimSpace(string(v))
-	if err == nil {
-		verbosePrintf("version from file 'VERSION' is %s\n", version)
-		return version
+	buf, err := ioutil.ReadFile("VERSION")
+	if err != nil {
+		verbosePrintf("error reading file VERSION: %v\n", err)
```
fw42: shouldn't this `return`? will `string.TrimSpace(string(buf))` work in the error case? will that just be an empty string?

build.go
```diff
@@ -200,17 +200,30 @@ func test(gopath string, args ...string) error {
...
+	}
+	versionFile := strings.TrimSpace(string(buf))
+	verbosePrintf("version from file 'VERSION' is %q\n", versionFile)
+
+	versionGit := gitVersion()
+	verbosePrintf("version from file 'VERSION' is %q\n", versionGit)
```
fw42: this should say "version from git" I guess?

build.go
```diff
@@ -200,17 +200,30 @@ func test(gopath string, args ...string) error {
...
+		return "unknown"
+	}
+
+	version := versionFile
+	if versionGit != "" {
+		version += " (" + versionGit + ")"
 	}
 
-	return gitVersion()
+	return version
```
fw42: if VERSION doesn't exist, this will be `" (git things)"` (with parentheses and the leading space). Kinda awkward...

build.go
fd0: Uh, good point. I don't know what `ReadFile()` returns in case of an error...

build.go
fd0: Agreed.

build.go
fd0: That's the most obscure case, there should always be a `VERSION` file. But we can just use the git version without the parentheses...

Issue #283, opened 2015-08-26
Fix build on darwin


fd0:
Since last night, the tests on Travis for darwin fail, even the master branch (that was successfully built before) doesn't work any more: https://travis-ci.org/restic/restic/jobs/76800443

Someone needs to investigate that.

fd0:
I've emailed the support, let's see what they say.

bchapuis:
Just gave it a try on my mac and apparently the problem may come from one of the tests in the cmd/restic directory.

fd0:
Hm, but why did it work before?

fd0:
Could you run the tests with `go test -v ./...`?

bchapuis:
Here is the output:

```
=== RUN   TestArchivePipe
--- PASS: TestArchivePipe (0.00s)
=== RUN   TestArchiveDirectory
--- PASS: TestArchiveDirectory (2.02s)
	archiver_test.go:110: snapshot archived as dc70118799a27f127366ec2366478d089bdcedf7f55a8b36c7f93cf2713a192f
=== RUN   TestArchiveDedup
--- PASS: TestArchiveDedup (2.28s)
	archiver_test.go:147: archived snapshot 6c31ca42
	archiver_test.go:154: packs 13, data blobs 459, tree blobs 91
	archiver_test.go:158: archived snapshot 9e0a7946
	archiver_test.go:165: packs 16, data blobs 459, tree blobs 181
	archiver_test.go:175: archived snapshot afcbaa63, parent 9e0a7946
	archiver_test.go:182: packs 18, data blobs 459, tree blobs 266
=== RUN   TestCache
--- PASS: TestCache (2.29s)
=== RUN   TestLock
--- PASS: TestLock (1.00s)
=== RUN   TestDoubleUnlock
--- PASS: TestDoubleUnlock (1.02s)
=== RUN   TestMultipleLock
--- PASS: TestMultipleLock (0.97s)
=== RUN   TestLockExclusive
--- PASS: TestLockExclusive (1.04s)
=== RUN   TestLockOnExclusiveLockedRepo
--- PASS: TestLockOnExclusiveLockedRepo (0.76s)
=== RUN   TestExclusiveLockOnLockedRepo
--- PASS: TestExclusiveLockOnLockedRepo (1.56s)
=== RUN   TestLockStale
--- PASS: TestLockStale (0.00s)
=== RUN   TestLockWithStaleLock
--- PASS: TestLockWithStaleLock (0.60s)
[...]

PR #284, opened 2015-08-26
wait for integration tests

bchapuis:
#283 check what this travis_wait function does


fd0:
Thanks for helping out, I think we don't need this PR any more :)

PR #285, opened 2015-08-26
Use new version of s3 library, Fixes #276

### LINKED ISSUE #276: Support Amazon S3 signature version 4?
I tried creating a repository in the eu-west-1 region, and got:

    restic -r s3://eu-central-1/my.backup.bucket init
    creating key in backend at s3://eu-central-1/my.backup.bucket failed:
    The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.

I guess Frankfurt was an unlucky choice, since it's the one [region](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region) that doesn't support the old V2. Still, maybe consider switching to version 4?

P.S. I saw the video of today's restic presentation, very cool. I look forward to trying it out further!

howeyc:
This more up-to-date library is missing the GetKey function on a bucket. So I used the GetReader and the error value to test for a Key that exists. I'm not sure how much data is returned this way, so it may be slower than the other library that did a "HEAD" request.

I did take a look at the Amazon library/SDK but the interface looks to be very different. I'll investigate further at another time.


fd0:
Thanks for your work!

Please ignore build errors for darwin on travis, these are broken at the moment (see #283), the other errors: please fix ;)

fd0:
I've restarted the travis build, should pass now.

fd0:
LGTM! :+1:

PR #287, opened 2015-08-26
Remove tests for directories

### LINKED ISSUE #279: using an empty password on initial creation will create an unusable repository
# how to reproduce

init a new repository and enter an empty password

```
$ ./restic -r foo  backup ~/tmp
an empty password is not a password
```

So just try it again, this time willing to enter a non-empty password

# expected result

Can do this again and enter a password this time.

# actual result

```
$ ./restic -r ~/tmp  init
creating backend at /Users/mirko/tmp failed: dir /Users/mirko/tmp/data already exists
```

I can do nothing except deleting ```~/tmp```. This maybe a problem with other backends.

# proposal

Do the *i want a non empty password* check before initializing the repository.

fd0:
For testing whether a repository already exists it is sufficient to test if the config file (and therefore the master key) exists.

Closes #279

### DIFF
--- a/backend/local/local.go
+++ b/backend/local/local.go
@@ -62,13 +62,6 @@ func Create(dir string) (*Local, error) {
 		return nil, errors.New("config file already exists")
 	}
 
-	// test if directories already exist
-	for _, d := range dirs[1:] {
-		if _, err := os.Stat(d); err == nil {
-			return nil, fmt.Errorf("dir %s already exists", d)
-		}
-	}
-
 	// create paths for data, refs and temp
 	for _, d := range dirs {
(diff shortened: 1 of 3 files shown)


backend/local/local.go
```diff
@@ -57,18 +57,11 @@ func Create(dir string) (*Local, error) {
 	}
 
 	// test if config file already exists
-	_, err := os.Lstat(backend.Paths.Config)
+	_, err := os.Lstat(filepath.Join(dir, backend.Paths.Config))
```
fw42: why was this join not necessary before?

fd0: It was necessary before, that's why this change fixes a bug ;)

PR #296, opened 2015-09-06
testify dependency

bchapuis:
If we remove completely the dependency between the chunker and restic, testify is necessary for restic to compile.


fd0:
I don't understand what this pull-request does, it only adds testify to the Godeps file, but doesn't add the vendored source code, and doesn't even use testify... can you elaborate?

bchapuis:
Sorry, I misunderstood the way Godeps work and thought this was necessary to compile the chunker in the vendor directory, but apparently, it is not.

fd0:
Ah, thanks for the clarification.

Issue #211, opened 2015-07-04
Storage Backend: Google Cloud Storage


e2b:
Google's [Cloud Storage Nearline](https://cloud.google.com/storage-nearline/) has a pricing similar to Amazon Glacier, but with ~3 sec response times and free ingress (upload) traffic.

kurin:
I'd like to add this feature.  I'll send a pull request in a few days.

fd0:
Cool, please go ahead! :+1:

fd0:
As just said in the PR #449: I'd like to move forward with adding GCS backend support natively. What is needed in order to access it? A client ID or a client secret?

fd0:
According to @gebi, no such thing is needed, so that's good news.

PR #302, opened 2015-09-16
Add support for Google Cloud Storage (including Nearline).

kurin:
This adds support for GCS per #211.

There are a few improvements that can be made on top of this, especially fixing GetReader so that it doesn't throw away extraneous bytes.


fw42:
This is impossible to review (the diff is too big for GitHub to display). Can you split this into two branches (and two PRs)? One that only has the dependencies and then another PR (against that first branch) that has your actual code contributions to restic?

kurin:
Can do.

fd0:
I must say that I find the large amount of code this backend depends on frightening...

fd0:
Code is in #304, Deps in #303.

abourget:
Any news on this ? I'd be glad to use restic with Google Storage too.

PR #303, opened 2015-09-16
Modify godeps for GCS.

kurin:


fd0:
This is not needed any more, thanks a lot @kurin

PR #304, opened 2015-09-16
Add backend support for Google Cloud Storage.

kurin:


fd0:
This is expected to fail, the dependencies are not included in the PR (see #302)

fw42:
That's why I asked this to be PRed against the other branch, not against master :-/ Then it wouldn't fail...

fd0:
Yeah, but then we'd still see the commit adding the dependecies and GitHub won't show the diff.

fw42:
nope you wouldn't

fw42:
The GitHub PR UI shows the diff between your branch and the branch the PR is made against... So if you PR against a branch that already has the dependencies (and this branch has it as well), then you only see what you see here right now (but the tests will work).

fd0:
Interesting, I'll talk to @kurin.

fd0:
Done, let's see.

fd0:
For me, it still shows `370 changed files`...

fd0:
@kurin Just a quick reminder: Thank you for your contribution, I haven't forgotten about it. I'll have a look as soon as I have time, probably over the holidays :)

kurin:
FYI, the Google storage library has changed significantly since this CL was opened.  If you want to keep in sync, more work will have to be done.

fd0:
Ok, good to know.

abourget:
Keep up the good work folks :) Anything I could help with ?

fd0:
Maybe you could have a look at the API changes and forward-port the code in this PR.

Is there a way to get easy access for me to Nearline storage to test this code?

kardianos:
@fd0 google cloud offers a free trial you could use as a test for initial merging: https://cloud.google.com/free-trial/

broady:
https://github.com/GoogleCloudPlatform/gcloud-golang/wiki/Storage-Package-Update

fd0:
@broady thanks, I'll have a look.

kurin:
Just FYI I actually have a CL open with the cloud storage package to provide a LimitedReader, which means you can read from part of an object without downloading the rest of it, which would be nice.

Once I get that in I can revisit this, if you want, or if you have a better way to approach this (without adding all of GCS to the dependencies) I could look at that.  My personal backup strategies have moved from restic to zfs, so I no longer really need to push this though, but I'm happy to see it in if anyone else has a need.

mappu:
> The Google Cloud Storage XML API is interoperable with some cloud storage tools and libraries that work with services such as Amazon Simple Storage Service (Amazon S3) 

So `restic` can already back up to Google Cloud Storage using existing minio backend. Am I wrong? or what does this patch add?

fd0:
@mappu This PR was opened way before we switched to minio, I'll close it for now.

PR #308, opened 2015-09-24
fix:restic#292 Prevent concurrent processing of the same blob

episource:
Prevent concurrent processing of the same blob by first adding a preliminary index entry: Adding a preliminary index entry fails if another goroutine is already working on the same blob. In contrast to my proposal in restic#292, index.Store has not been changed. Instead a new method index.StoreInProgress has been added.

### DIFF
--- a/repository/index.go
+++ b/repository/index.go
@@ -2,6 +2,7 @@ package repository
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"sync"
@@ -43,7 +44,8 @@ func (idx *Index) store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
 	}
 }
 
-// Store remembers the id and pack in the index.
+// Store remembers the id and pack in the index. An existing entry will be
+// silently overwritten.
 func (idx *Index) Store(t pack.BlobType, id backend.ID, pack *backend.ID, offset, length uint) {
 	idx.m.Lock()
 	defer idx.m.Unlock()
@@ -52,6 +54,26 @@ func (idx *Index) Store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
 		pack.Str(), id.Str(), t, offset, length)
 
 	idx.store(t, id, pack, offset, length, false)
+}
+
+// StoreInProgress adds a preliminary index entry for a blob that is about to be
+// saved. The entry must be updated using Store once the the blob has been
+// written to a pack. Adding an preliminary index fails if there's an existing
+// entry associated with the same id.
+func (idx *Index) StoreInProgress(t pack.BlobType, id backend.ID) error {
+	idx.m.Lock()
+	defer idx.m.Unlock()
+
+	if _, hasID := idx.pack[id]; hasID {
+		errorMsg := fmt.Sprintf("index already contains id %v (%v)", id.Str(), t)
+		debug.Log("Index.StoreInProgress", errorMsg)
+		return errors.New(errorMsg)
+	}
+
+	idx.store(t, id, nil, 0, 0, false)
+	debug.Log("Index.StoreInProgress", "preliminary entry added for id %v (%v)",
+		id.Str(), t)
+	return nil
 }
 
 // Remove removes the pack ID from the index.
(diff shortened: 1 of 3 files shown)


episource:
I tested the patch by making a backup of the data set described in restic#292, that initally made me discover the issue addressed by the pull request. During that test restic check did not find any errors. Besides the analysis in restic#292, this is a strong indicator, that the issue is fixed by the patch. However, due to the non-deterministic nature of the problem, it's no guarantee.

fd0:
Awesome, thanks a lot! I'll have a look on the weekend.

repository/index.go
```diff
@@ -43,7 +44,7 @@ func (idx *Index) store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
 	}
 }
 
-// Store remembers the id and pack in the index.
+// Store remembers the id and pack in the index; an existing entry will be silently overwritten
```
fd0: Make this full sentences, e.g. `Store remembers the id and pack in the index. An existing entry will be silently overwritten.`

repository/index.go
```diff
@@ -54,6 +55,27 @@ func (idx *Index) Store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
 	idx.store(t, id, pack, offset, length, false)
 }
 
+// Adds a preliminary index entry for a blob that is about to be saved. The entry must be updated using Store once
+// the the blob has been written to a pack. Adding an preliminary index fails if there's an existing entry associated
+// with the same id. Due to locking the index this will ensure only one goroutine processes the the same blob
+// concurrently.
```
fd0: Comments for functions should begin with the name of the function they describe and wrapped at 80 characters. Suggestion:
```
// StoreInProgress adds a preliminary index entry for a blob that is about to
// be saved. The entry must be updated using Store once the blob has been
// written to a pack. Adding an preliminary index fails if there's an existing
// entry associated with the same id. Due to locking the index this will ensure
// only one goroutine processes the same blob concurrently.
```

repository/index.go
```diff
@@ -54,6 +55,27 @@ func (idx *Index) Store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
...
+func (idx *Index) StoreInProgress(t pack.BlobType, id backend.ID) error {
+	// Note: Cannot use idx.Has(...) or idx.Store(...) as these would lock the index again and Go mutexes seem not to be
```
fd0: Comment line is too wide, please wrap at 80 chars. Is this comment really necessary?

repository/index.go
```diff
@@ -54,6 +55,27 @@ func (idx *Index) Store(t pack.BlobType, id backend.ID, pack *backend.ID, offset
...
+	// reentrant
+	idx.m.Lock()
+	defer idx.m.Unlock()
+
+	if _, hasId := idx.pack[id]; hasId {
```
fd0: Use `hashID` instead of `hashId`, see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

fd0:
Thanks again for your contribution, I added my comments. These may seem to be only small nitpicks, but it's important for us to keep the code quality high (this includes the comments).

In addition to the comments above there are only two things that should be changed before I will merge this PR:
 * Please reformat the commit message, following the format described here: https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits
 * Add a test for `StoreInProgress()` that tries to add the same ID multiple times.

episource:
> Please reformat the commit message, following the format described here: https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits
Mmh. Now I'm in trouble, somewhat: Should I rewrite history in a public repository?

fd0:
Yes, you should rewrite history. This is okay (and common practice) in a topic-branch for a PR, it is never acceptable in e.g. the `master` (or `release` or `stable`) branch.

fd0:
This looks good, are you finished? :)

episource:
Well, I should have addressed all your comments... so, yes :)

fd0:
ok, cool, thanks for the contribution!

PR #310, opened 2015-09-27
resume interrupted backups

### LINKED ISSUE #59: Resume interrupted backups
When a backup is interrupted and restarted later, restic writes all data again, newly encrypted, because the index is written at the end so restic doesn't know about what blobs have already been stored.

This can be fixed for example by regularly (e.g. every X minutes or every Y files or so) uploading a preliminary index that is removed after the next preliminary index (which is a superset of the first one) has been uploaded. Once the final index was successfully uploaded, all preliminary indexes from this session are removed.

Connects #243

fd0:
This breaks out some commits from #252 (which was clearly too large to be done in one PR). The code in this PR adds resuming of interrupted backups by regularly writing an index (utilizing the new supersedes field) and reading these when a backup should be resumed.

Closes #59


fd0:
@episource I've handled the bug you discovered, thanks again.

episource:
Is a RWMutex reantrant? The documentation does not say anything about this...

Context: 
`AddInFlight` acquires a RW-Lock (`mi.inFlight.Lock()`) and then calls `Has`, which tries to acquire a Read-Lock (`mi.idxMutex.RLock()`). Without RWMutex being reantrant this should fail/dead-lock.

Assuming go locks to be reantrant I tried the same while implementing my draft of `StoreInProgress`. Due to that I learned the hard way, that my assumption was wrong and Mutex (not RWMutex) is not reantrant.

fd0:
Uhoh. I'll have a look, thanks again!

episource:
No, just forget my comment. It's plain wrong.

fd0:
Hm, I'd expect that acquiring the r lock while an r/w lock is already held leads to a deadlock...

fd0:
Ah, indeed, now I see it.

fd0:
Any other comments? Otherwise I'll merge this branch now :)

episource:
Not from my side...

fd0:
cool

PR #319, opened 2015-10-11
Handle null subtree IDs

### LINKED ISSUE #311: checker.filterTrees does not evaluate treeJob.error
**Restic version:** e738d35c4e

**Observed behavior:** During discussion of restic#292 it has been observed, that [`checker.filterTrees`](https://github.com/restic/restic/blob/e738d35c4ebb3fb26afb5b880357fbe3286cc7d4/checker/checker.go#L431-L442) does not check treeJob.error (channel `in`, local variable `j`) before accessing members of the treeJob struct. This might cause e.g. a nil pointer dereference error to happen.

**Expected behavior:** `treeJob.error`should be checked before reading any other member of the treeJob struct. Not sure what should be done with items that have `treeJob.error != nil`. Most probably they should be passed further along the pipe, just like done by the current implementation. In this case [`checker.checkTreeWorker`](https://github.com/restic/restic/blob/e738d35c4ebb3fb26afb5b880357fbe3286cc7d4/checker/checker.go#L338) and/or [`checker.checkTree`](https://github.com/restic/restic/blob/e738d35c4ebb3fb26afb5b880357fbe3286cc7d4/checker/checker.go#L489) might need to evaluate treeJob.error, too.

### LINKED ISSUE #314: panic when checking repo
Hi, first of all thank you for this great piece of software :+1: 

**Restic version** : e738d35c compiled 2015-10-01 11:59:10 with go version go1.5.1 linux/amd64

**Expected behavior:** report the integrity and consistency of the backup without panic and unlock the repository when executing `$ restic check` 

**Observed behavior:**  having this message in the stdout
```
Load indexes
Check all packs
```
and a panic [attached - debug enabled](https://gist.github.com/aminehmida/0e193889022465749c6c) in the stderr without unlocking the repo (need to run `$ restic unlock` )

My repo is stored in a usb drive formated with NTFS if this can help.

fd0:
This PR fixes handling of repositories with trees that have a subtree ID of 000000. It seems that #308 introduced a bug that results in a null subtree ID for the empty directory when this is archived concurrently.

Closes #311
Closes #314

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -334,6 +334,10 @@ func (arch *Archiver) dirWorker(wg *sync.WaitGroup, p *Progress, done <-chan str
 
 				if node.Type == "dir" {
 					debug.Log("Archiver.dirWorker", "got tree node for %s: %v", node.path, node.blobs)
+
+					if node.Subtree.IsNull() {
+						panic("invalid null subtree ID")
+					}
 				}
 			}
 
@@ -359,6 +363,9 @@ func (arch *Archiver) dirWorker(wg *sync.WaitGroup, p *Progress, done <-chan str
 				panic(err)
 			}
 			debug.Log("Archiver.dirWorker", "save tree for %s: %v", dir.Path(), id.Str())
+			if id.IsNull() {
+				panic("invalid null subtree ID return from SaveTreeJSON()")
+			}
 
 			node.Subtree = &id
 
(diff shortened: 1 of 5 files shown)


fd0:
@episource your PR contained a small bug that was triggered mostly only when creating a backup an empty subdir, cc7acba02b60d1bd82f0f8a26c319f94b226eec2 fixes this.

episource:
We should probably add a (integration) test, that triggers the former error...

fd0:
Yes, indeed, but that's not as easy as it sounds ;)

PR #335, opened 2015-11-02
Always use forward slashes in SFTP (Fixes #334)

### LINKED ISSUE #334: Windows SFTP Path Handling
Trying to Backup to an SFTP repository on Windows like this:

    > .\restic.exe -r sftp://user@host//files/Backup/restic backup "C:\Users\Jacob1123"
    \files\Backup\restic\data does not exist

Note that the error message contains Backslashes instead of forward slashes.

Also, when I init a new repository via SFTP I get the following folders instead of a single folder "restic" containing the others:

    restic/
    restic\\keys/
    restic\\locks/
    restic\\data/
    restic\\snapshots/
    restic\\index/
    restic\\tmp/

I also tried using a minGW-Terminal, but got the same results.

JaCoB1123:
Add custom Join func that always uses forward slashes in SFTP

### DIFF
--- a/backend/sftp/sftp.go
+++ b/backend/sftp/sftp.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/juju/errors"
 	"github.com/pkg/sftp"
@@ -74,12 +75,12 @@ func Open(dir string, program string, args ...string) (*SFTP, error) {
 	// test if all necessary dirs and files are there
 	items := []string{
 		dir,
-		filepath.Join(dir, backend.Paths.Data),
-		filepath.Join(dir, backend.Paths.Snapshots),
-		filepath.Join(dir, backend.Paths.Index),
-		filepath.Join(dir, backend.Paths.Locks),
-		filepath.Join(dir, backend.Paths.Keys),
-		filepath.Join(dir, backend.Paths.Temp),
+		Join(dir, backend.Paths.Data),
+		Join(dir, backend.Paths.Snapshots),
+		Join(dir, backend.Paths.Index),
+		Join(dir, backend.Paths.Locks),
+		Join(dir, backend.Paths.Keys),
+		Join(dir, backend.Paths.Temp),
 	}
 	for _, d := range items {
 		if _, err := sftp.c.Lstat(d); err != nil {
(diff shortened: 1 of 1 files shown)


backend/sftp/sftp.go
```diff
@@ -101,16 +102,16 @@ func Create(dir string, program string, args ...string) (*SFTP, error) {
...
+		Join(dir, backend.Paths.Snapshots),
+		Join(dir, backend.Paths.Index),
+		Join(dir, backend.Paths.Locks),
+		Join(dir, backend.Paths.Keys),
+		Join(dir, backend.Paths.Temp),
 	}
 
 	// test if config file already exists
-	_, err = sftp.c.Lstat(filepath.Join(dir, backend.Paths.Config))
+	_, err = sftp.c.Lstat(dir + backend.Paths.Config)
```
fd0: Why did you remove the `path.Join()` here? is `dir` guaranteed to always end with a slash?

backend/sftp/sftp.go
```diff
@@ -286,13 +287,33 @@ func (r *SFTP) Create() (backend.Blob, error) {
 	return &blob, nil
 }
 
+func Join(parts ...string) string {
```
fd0: This function's body can be replaced by calling `path.Clean(path.Join(parts...))`.

backend/sftp/sftp.go
```diff
@@ -286,13 +287,33 @@ func (r *SFTP) Create() (backend.Blob, error) {
...
+		result = result + "/" + s
+	}
+	return result
+}
+
 // Construct path for given backend.Type and name.
 func (r *SFTP) filename(t backend.Type, name string) string {
 	if t == backend.Config {
-		return filepath.Join(r.p, "config")
+		return Join(r.p + "config")
```
fd0: Again, I think `+` may be wrong here, since `r.p` may not end with a slash. Call `Join(r.p, "config")` instead.

fd0:
Thanks for your contribution! Please see my code review comments ;)

JaCoB1123:
Thanks. I fixed the issues.
The missing joins were leftovers of my first try to fix it ;-)

JaCoB1123:
Just noticed the Travis build failed. I can't figure out what the log says though.

fd0:
This is a really strange issue that sometimes occurs on Darwin. I've restarted the test.

fd0:
The test suceeded, thanks for your contribution!

PR #345, opened 2015-11-06
Switch s3 library to allow for s3 compatible backends. Fixes #315

### LINKED ISSUE #315: creating key in backend at s3://192.168.122.62/test-bucket failed:  unsupported protocol scheme ""
I have a local S3 cluster, both riak cs and swift deployted in the lab, not on AWS

s3cmd and other tools can read/write/make/delete buckets and objects without issues
i set  AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment to match
my credentials, when running

restic init -r s3://192.168.122.62/test-bucket
enter password for new backend:
enter password again:
creating key in backend at s3://192.168.122.62/test-bucket failed: Put /test-bucket/restic/key/e2130523be50a1ced505d5607b8e35124c21d378cb901bd2de9952d550e72c30: unsupported protocol scheme ""

howeyc:
One thing to note is the library I used is Apache License 2.0.

From what I can tell by reading the license all that's required is that we distribute the license with any binaries produced. Something to keep in mind for when that day comes.

Other than that, I think this should help with other s3 compatible backends.


fd0:
Thanks for your work! To be honest, I don't like the test dependency on the external service. I'm regularly coding without access to the Internet, and that test would fail then. Is there a way to start a s3 mock service as before?

howeyc:
Yeah, I probably could.

Minio has a server implementation. Maybe I could have the test start one up.

harshavardhana:
First of all I really like this project and see to be quite useful for others as well. 

I am the primary author for the aforementioned library, thanks for considering it. Happy to work with you guys to get this in properly. 

The problem with using https://github.com/minio/minio-go is that it is Signature V4 only, wouldn't work with some of the S3 compatible solutions out there namely Swift, Ceph etc. 

Minio server on the other hand is Signature V4 so minio-go will not have any issue.  

Additionally we also have https://github.com/minio/minio-go-legacy for Signature V2 - servers. 

But here is the plan to make things simple, i am going to merge. 

 - https://github.com/minio/minio-go
 - https://github.com/minio/minio-go-legacy 

This would allow you guys to transparently do signature V2 or V4 depending upon the server you are using.  I will get it ready by tentative tomorrow my time i.e PST. Please do not merge these changes before that.

Thanks

harshavardhana:
https://github.com/minio/minio-go/pull/195 - here we go - merged with signature v2 support.

harshavardhana:
Looks like you guys also need a proper a fake testing server to be used along with minio-go, let me provide that rather than using play.minio.io:9000

fd0:
> Looks like you guys also need a proper a fake testing server to be used along with minio-go, let me provide that rather than using play.minio.io:9000

I was thinking more along the lines of running a local instance for the backend integration tests, so that no Internet connection is required. We can however have another test that is only run on demand which tests against a live minio test server.

fd0:
@howeyc what's the problem with `GetPartialObject()`?

harshavardhana:
> I was thinking more along the lines of running a local instance for the backend integration tests, so that no Internet connection is required. We can however have another test that is only run on demand which tests against a live minio test server.

Even that is sufficient, you can run a local 'https://github.com/minio/minio' instance - its a single binary a simple "go get" will work. 

We will make sure not to break anything for you guys :-)

harshavardhana:
> @howeyc what's the problem with GetPartialObject()?

My question as well! - is it an issue with `play.minio.io:9000` ?

harshavardhana:
> @howeyc what's the problem with GetPartialObject()?

Guessing just fixed a problem in how it was interpreting the situation where offset is 0, and length N - in this case GetPartialObject() would return back the last N bytes, rather than the first 0 - (N-1) bytes.

howeyc:
I'll play with it some more, but I was getting a lot of errors in the tests.

I'll see if I can find the cause, or if it is now fixed as mentioned in a later comment.

howeyc:
GetPartialObject looks like it's working now. I've even tested this against my real restic repo on s3.

Not sure if I was seeing something else, or if latest version fixed it. Either way it's working for me.

Regarding the tests, they will fail. I'm not sure where/how to place the hooks to download an s3 server and start/stop it.

harshavardhana:
> Regarding the tests, they will fail. I'm not sure where/how to place the hooks to download an s3 server and start/stop it.

What i am going to do is write a simpler fake version of all the handlers so that it becomes easier to do regular tests during build here. Rather than starting a server and having a large dependency such as that. 

For full integration tests i think @fd0 has opened a new issue #346

harshavardhana:
> GetPartialObject looks like it's working now. I've even tested this against my real restic repo on s3.

Yes that's what i expected :-) ..

harshavardhana:
> What i am going to do is write a simpler fake version of all the handlers so that it becomes easier to do regular tests during build here. Rather than starting a server and having a large dependency such as that.

Actually after looking at this approach, i am thinking it is pointless to write a large fake server when you actually have a proper working version. 

What we need to do is run "minio" micro services binary itself directly, during these backend tests. `minio` micro services binary is small enough to run for these backend tests. 

What you guys think about this approach?

fd0:
I like this approach. However, downloading and compiling the minio server binary should be done outside of the tests itself, and only the URL should be passed in. I'll prepare a proposal later.

harshavardhana:
> I like this approach. However, downloading and compiling the minio server binary should be done outside of the tests itself, and only the URL should be passed in. I'll prepare a proposal later.

Do you need any help here?

fd0:
> Do you need any help here?

Not at the moment, but I'll definitely contact you ;)

We already have a script (`run_integration_tests.go`) that configures the environment and runs the tests, that'd be ideal to modify to download and run the minio binary.

fd0:
I'll close this PR, it is superseded by #366

PR #350, opened 2015-11-10
Allow check/restore read-only, without locking

### LINKED ISSUE #318: Allow checking a read-only repository (e.g. disable locking)
**Restic version:** e738d35

**Expected behavior:** `restic check` should support checking repositories stored on a read only mounted filesystem.

**Observed behavior:** restic fails to acquire exclusive lock, but the lock is not actually necessary due to the filesystem being read only
```
$ restic -r /media/backup-intern/restic/limucs-main.repo check
open /media/backup-intern/restic/limucs-main.repo/tmp/temp-278668902: read-only file system
```

fd0:
This code adds the option `--no-lock` to the `check`, `list` and `restore` commands so that (local) repositories on a read-only medium can be accessed. For all commands, locking is the only modification that is done to the repo, so when locking is disabled, the repo is not written to at all.

This fixes #318


episource:
Tried `check --no-lock` and it works for me.

Looking at the code you didn't implement --no-lock for other read-only operations than check, list and restore. What about ls, find, cat, ...?

fd0:
I'll add that.

fd0:
I've made `--no-lock` a global option, the other commands `find`, `cat` and `ls` do not lock the repo.

Issue #358, opened 2015-11-24
Regression: local backup of large data set - pack not referenced in any index


episource:
Possible regression: The issue in #292 is back again! #292 was finally solved in cc7acba, but that was some while ago and lot changed in the meanwhile.  Sadly I can't tell at which commit after cc7acba the issue reappeared. Due to the non-deterministic nature of this issue it's hard to tell. 

**Restic version:** restic 0.1.0 (v0.1.0-147-g867fc5b), compiled at 2015-11-21 09:32:12 with go1.5.1

**Expected behavior:** restic check passes successfuly and hashdeep confirms file integrity.

**Observed behavior:** pack not referenced in any index, chance of observing this errors seems to increase with backup set size and data duplication; see also #292.

Backup can be restored without errors. This has been verified using a hashdeep md5 audit file.

```
restic -r /media/backup-intern/restic/limucs-main.repo --cache-dir=/root/.cache/restic check
Create exclusive lock for repository
Load indexes
Check all packs
pack a2de8d66e12c28b36c63827a0a8e75583cc8f4186e7004fc68e2e7d6c328a5d8: not referenced in any index
Check snapshots, trees and blobs
repository contains errors
```

**Steps to reproduce:** Backup a large data set (mine is ~500G) with duplication. I will repeat my analysis from #292 as soon as possible, but can't provide a timeline.

episource:
## Analysis
[...]

PR #365, opened 2015-12-06
Add a test reproducing #358 by concurrently saving duplicated chunks

episource:
This commit adds an integration test, that calls Archiver.Save from many goroutines processing several duplicated chunks concurrently. The test asserts, that after all chunks have been saved, there are no
unreferenced packs in the repository.

The test has been checked to give the expected results:
1. Running the test with maxParallel=1 (all chunks are processed sequentially) has been verified not to produce any unreferenced packs. Consequently the test passes.
2. Running the test with unbounded parallelism (maxParallel=math.MaxInt32) has been verified to produce unreferenced packs all the time (at least 25 test runs). Consequently the test fails due to #358.

Note: The test is **expected to fail**, as the underlying issue #358 has not yet been solved. Hopefully it does not fail due to running **out of memory**. In it's current form, the test allocates a lot of memory. So it might require some reworking.

### DIFF
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -245,7 +245,7 @@ func BenchmarkLoadTree(t *testing.B) {
 // Saves several identical chunks concurrently and later checks that there are no
 // unreferenced packs in the repository. See also #292 and #358.
 func TestParallelSaveWithDuplication(t *testing.T) {
-	for seed := 0; seed < 5; seed++ {
+	for seed := 0; seed < 10; seed++ {
 		testParallelSaveWithDuplication(t, seed)
 	}
 }
@@ -254,7 +254,7 @@ func testParallelSaveWithDuplication(t *testing.T, seed int) {
 	repo := SetupRepo()
 	defer TeardownRepo(repo)
 
-	dataSizeMb := 92
+	dataSizeMb := 128
 	duplication := 7
 
 	arch := restic.NewArchiver(repo)
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -4,13 +4,16 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"io"
+	"math"
 	"testing"
 
 	"github.com/restic/chunker"
 	"github.com/restic/restic"
 	"github.com/restic/restic/backend"
+	"github.com/restic/restic/checker"
 	"github.com/restic/restic/crypto"
 	"github.com/restic/restic/pack"
+	"github.com/restic/restic/repository"
 	. "github.com/restic/restic/test"
 )
 
@@ -19,6 +22,11 @@ var testPol = chunker.Pol(0x3DA3358B4DC173)
 type Rdr interface {
 	io.ReadSeeker
 	io.ReaderAt
+}
+
+type chunkedData struct {
+	buf    []byte
+	chunks []*chunker.Chunk
 }
 
 func benchmarkChunkEncrypt(b testing.TB, buf, buf2 []byte, rd Rdr, key *crypto.Key) {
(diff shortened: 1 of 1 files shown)


episource:
Interesting. All but one test runs failed as expected. One passed however: [Job #892.1](https://travis-ci.org/restic/restic/jobs/95211692). Probably I should further increase the number of duplicates!? Any idea on how to increase the chance of #358 to happen?

episource:
I'm trying to diagnose why the test consumes so much memory. Therefore I'm running the test with [gctrace=1](https://golang.org/pkg/runtime/): `env GODEBUG=gctrace=1 go test -v ./archiver_test.go`

When I do this for the unmodified TestParallelSaveWithHighDuplication test from above (all other tests have been deleted from archiver_test.go), I get an output similiar to:
```
=== RUN   TestParallelSaveWithHighDuplication
gc 1 @0.007s 4%: 0.015+0.16+0.10+0.25+0.18 ms clock, 0.031+0.16+0+0/0.11/0+0.37 ms cpu, 64->64->64 MB, 4 MB goal, 2 P
gc 2 @2.054s 0%: 0.41+1.5+0.15+0.88+0.45 ms clock, 0.82+1.5+0+0.16/0.012/0+0.91 ms cpu, 128->192->128 MB, 128 MB goal, 2 P
gc 3 @2.063s 0%: 0.049+0.15+3.6+0.18+0.35 ms clock, 0.049+0.15+0+0.097/0.017/0+0.35 ms cpu, 160->160->160 MB, 128 MB goal, 2 P
gc 4 @6.689s 0%: 0.015+1.7+0.005+13+0.26 ms clock, 0.031+1.7+0+0.14/1.4/0+0.53 ms cpu, 219->233->191 MB, 320 MB goal, 2 P
gc 5 @6.919s 0%: 0.029+2.0+0.007+1.3+0.30 ms clock, 0.058+2.0+0+0.19/0.99/0+0.60 ms cpu, 292->300->236 MB, 356 MB goal, 2 P
gc 6 @7.263s 0%: 0.055+2.1+0.006+16+4.6 ms clock, 0.11+2.1+0+0.26/0.98/0+9.2 ms cpu, 425->438->345 MB, 457 MB goal, 2 P
gc 7 @7.784s 0%: 0.079+3.6+0.008+19+0.36 ms clock, 0.15+3.6+0+0.71/0.78/0+0.73 ms cpu, 643->661->508 MB, 664 MB goal, 2 P
gc 8 @8.575s 0%: 0.038+5.8+0.009+3.9+1.2 ms clock, 0.077+5.8+0+0.97/0.89/0+2.5 ms cpu, 955->980->754 MB, 979 MB goal, 2 P
gc 9 @9.390s 0%: 0.12+7.9+0.008+3.1+0.37 ms clock, 0.24+7.9+0+0.29/1.4/0+0.74 ms cpu, 1422->1428->1121 MB, 1459 MB goal, 2 P
gc 10 @10.893s 0%: 0.066+10+0.010+9.6+0.45 ms clock, 0.13+10+0+0.74/1.3/0+0.91 ms cpu, 2178->2237->1729 MB, 2232 MB goal, 2 P
gc 11 @13.734s 0%: 0.063+12+0.007+27+1.3 ms clock, 0.12+12+0+0.24/2.2/0+2.6 ms cpu, 3260->3295->2384 MB, 3343 MB goal, 2 P
gc 12 @17.014s 0%: 0.11+14+0.008+2.3+0.83 ms clock, 0.22+14+0+0/2.2/0+1.6 ms cpu, 4582->4582->3187 MB, 4699 MB goal, 2 P
gc 13 @24.859s 0%: 0.14+13+0.008+14+1.1 ms clock, 0.29+13+0+0/2.5/0+2.3 ms cpu, 6215->6222->4538 MB, 6374 MB goal, 2 P
```
This shows, that the test allocates quite a lot of memory. The peek live heap size of ~4500Mb is nearly 30 times the amount of memory required to store the 5 (pseudo) random blobs of 32Mb each.

Now I do the same for the test modified, such that archiver.Save is not called:
```
@@ -14,3 +14,3 @@
 	"github.com/restic/restic/checker"
-    "github.com/restic/restic/pack"
+
 	"github.com/restic/restic/repository"
@@ -49,3 +49,3 @@
 
-	arch := restic.NewArchiver(repo)
+	restic.NewArchiver(repo)
 	data := getRandomData(seeds)
@@ -68,3 +68,3 @@
 
-					err := arch.Save(pack.Data, id, c.Length, c.Reader(bytes.NewReader(*buf)))
+					var err error = nil
 					<-barrier
```

After that change, the peak memory usage droped by more than 90%:
```
=== RUN   TestParallelSaveWithHighDuplication
gc 1 @0.007s 5%: 0.022+0.18+0.004+0.17+0.29 ms clock, 0.044+0.18+0+0/0.088/0.005+0.58 ms cpu, 64->64->64 MB, 4 MB goal, 2 P
gc 2 @0.921s 0%: 0.34+0.14+1.4+0.73+0.33 ms clock, 0.68+0.14+0+0.10/0.023/0+0.66 ms cpu, 128->224->160 MB, 128 MB goal, 2 P
gc 3 @2.718s 0%: 0.016+0.15+0.005+0.25+0.30 ms clock, 0.033+0.15+0+0/0.12/0+0.61 ms cpu, 161->161->161 MB, 128 MB goal, 2 P
```
This time the peak live heap size is 161Mb, which is about the memory required to store the 5 32Mb blobs. Actually I'd expect a little bit higher heap allocation, but well, it's possible.

Interesting, isn't it? Am I doing something wrong? Does arch.Save really require so much memory? Why? Many copies of the input data?

fd0:
Heh, I think you're looking at an intelligent compiler. When Save() is not called, the buffers may be allocated on the stack and not on the heap.

If you're interested where the memory is used/allocated, run `go test` with `-memprofile` and use `go tool pprof` to inspect it afterwards. Very helpful! See also https://blog.golang.org/profiling-go-programs

episource:
The test code allocates only about 160Mb. See [memprofile](https://github.com/restic/restic/files/55797/out.pdf).

**Here are some more details:**
The test creates 5 blobs with 32Mb each. The chunker splits these into a total of 128 chunks which leads to an average of 1.25Mb per chunk. For each chunk the test currently creates 15 goroutines processing that chunk. This makes a total of 1920 goroutines each processing 1.25 Mb of data. Hence I'd expect at least ~2500Mb heap size. [Restic did use ~4500Mb](https://github.com/restic/restic/pull/365#issuecomment-162661735). Cannot really tell whether this is good or bad.

**How I will develop the test further:**
* Create `N` goroutines processing a chunk concurrently. `N` should probably be less than 15 as much more goroutines than there are cores probably doesn't make sense. 10 could be a good choice. Let's see.
* Wait for these goroutines to finish before processing the next chunk in the same manner. 

=> Processing any other chunks in parallel probably won't help provoking #358. It only increases memory usage. ... also after these changes one 32Mb blob might be sufficient...

episource:
> How I will develop the test further:
>
> *  Create N goroutines processing a chunk concurrently. N should probably be less than 15 as much more goroutines than there are cores probably doesn't make sense. 10 could be a good choice. Let's see.
> * Wait for these goroutines to finish before processing the next chunk in the same manner.
>
> => Processing any other chunks in parallel probably won't help provoking #358. It only increases memory usage. ... also after these changes one 32Mb blob might be sufficient...

This assumption was wrong: After that change I'm not able to reproduce #358 in 100% of the test runs on my local machine. Nevertheless, I'll commit an updated test soon, that reduces memory usage by lowering the number of concurrent saves.

episource:
> How I will develop the test further:
>
> *  Create N goroutines processing a chunk concurrently. N should probably be less than 15 as much more goroutines than there are cores probably doesn't make sense. 10 could be a good choice. Let's see.
> * Wait for these goroutines to finish before processing the next chunk in the same manner.
>
> => Processing any other chunks in parallel probably won't help provoking #358. It only increases memory usage. ... also after these changes one 32Mb blob might be sufficient...

This assumption was wrong: After that change I'm not able to reproduce #358 in 100% of the test runs on my local machine. Nevertheless, I'll commit an updated test soon, that reduces memory usage by lowering the number of concurrent saves.

episource:
Mmh. This is a hard one. Not all builds failed. This time the windows build passed... I'll increase the total number of chunks, even if that'll raise memory usage again. The current test is somewhere around ~450Mb.

episource:
This looks promising. Nevertheless, the test won't be 100% reliable. I think no dynamic test will be. Multithreading issues like #358 are just too hard to reproduce reliably.

fd0:
Thanks for your work, I've planned a restic hack day on the weekend. @episource Will you maybe attend 32c3 in Hamburg? We'll have a hack-meeting there.

episource:
I didn't plan going to 32c3. As I'am living at germany's southern end, there also won't be the possibility for a spontaneous visit.

PR #366, opened 2015-12-06
rebase: Switch s3 library to allow for s3 compatible backends

fd0:
I've rebased #345 on top of the current master and added a few commits to get it working.

TODO:
- [x] run minio server for integration tests
- [x] Clarify URL parsing rules

I've also added an issue to the minio server repository in order to be able to specify the config directory: https://github.com/minio/minio/issues/997

This supersedes #345.

Ping @howeyc


howeyc:
I'm not sure that you want Delete to also remove the bucket. A while back you talked me into adding all restic-related keys under the "restic/" prefix in case there was other data in the bucket already. What if there is other data besides restic in a bucket? Deleting it might not be the best idea.

I suppose we could document somewhere that the bucket should be for restic data only?

The rest of it looks good.

fd0:
Oh, you're right. I'd then opt to not remove the bucket itself on Delete(). For consistency I'd the also ignore errors returned from CreateBucket() in Create(), so that it does not fail when the bucket already exists. Does that sound good to you?

howeyc:
Sounds good to me.

fd0:
This looks good. I've also added a dockerfile that can be used to build a container which resembles the Travis CI environment.

fd0:
FYI: I'm currently reworking the URL/repository location parsing code, as that's too loosely defined right now. My approach is roughly the following:

 * When the repo location string starts with the name of a backend, followed by a colon, a function in that backend is called to parse the string. I'll make sure that these backend-specific functions are backwards-compatible to the current syntax.
 * Otherwise the local backend is used and tries to open/create a repository at a directory location given by the string.

harshavardhana:
Let me know @fd0 - if you need any help from my end.

fd0:
Just a quick heads-up: I'm going back and forth with the minio-go people, I found a few bugs in the library. Still working on it.

fd0:
Oi, this really seems to work! Creating a backup on a minio server, and on S3 directly :)

harshavardhana:
> Oi, this really seems to work now! Creating a backup on a minio server, and on S3 directly :)

:+1:

JaCoB1123:
I just tested restic with minio and got the following error from minio (see minio/minio#1024)
[log.txt](https://github.com/minio/minio/files/97335/log.txt)

On that error, restic freezes completely and I have to kill it via the task manager.

fd0:
Interesting, could you describe how restic was started exactly?

JaCoB1123:
```
RESTIC_PASSWORD=.. AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=... restic.exe -r s3:http://127.0.0.1:9000/restic backup D:\test
```

JaCoB1123:
I got some output once:
[log_restic.txt](https://github.com/restic/restic/files/97356/log_restic.txt)

howeyc:
Try removing the "http:"

It should be implicit, we may need to change that.

JaCoB1123:
I tried it without the http first, but then restic uses https

fd0:
He's right, default is HTTPS. I'm still working on implementing a way to pass additional parameters to backends. It will look like the following:
```
$ restic -r s3 -o s3.endpoint=http://10.0.0.1:9000 -o s3.prefix=/restic -o s3.bucket=foobar-backup
```

But this is not yet done.

harshavardhana:
Looks like minhttp bug on windows. Will take a look.

harshavardhana:
Thanks @JaCoB1123 this is a Golang runtime bug and perhaps a known issue will be fixed only in Go 1.6 release which will happen in couple of weeks relevant issue to be noted here.

```
runtime: free list of span 0x26d7100:
0xc082580900 -> 0x5400410044 (BAD)
fatal error: free list corrupted

runtime stack:
runtime.throw(0x9cbe00, 0x13)
    c:/go/src/runtime/panic.go:527 +0x97
runtime.mSpan_Sweep(0x26d7100, 0x2100000000, 0xc000000001)
    c:/go/src/runtime/mgcsweep.go:186 +0x80e
runtime.sweepone(0x46e389)
    c:/go/src/runtime/mgcsweep.go:97 +0x162
runtime.gosweepone.func1()
    c:/go/src/runtime/mgcsweep.go:109 +0x28
runtime.systemstack(0xc08203a1c0)
    c:/go/src/runtime/asm_amd64.s:262 +0x7e
runtime.mstart()
    c:/go/src/runtime/proc1.go:674

goroutine 3 [running]:
runtime.systemstack_switch()
    c:/go/src/runtime/asm_amd64.s:216 fp=0xc082029f38 sp=0xc082029f30
runtime.gosweepone(0x1)
    c:/go/src/runtime/mgcsweep.go:110 +0x44 fp=0xc082029f60 sp=0xc082029f38
runtime.bgsweep(0xc082036000)
    c:/go/src/runtime/mgcsweep.go:54 +0xc4 fp=0xc082029f98 sp=0xc082029f60
runtime.goexit()
    c:/go/src/runtime/asm_amd64.s:1696 +0x1 fp=0xc082029fa0 sp=0xc082029f98
created by runtime.gcenable
    c:/go/src/runtime/mgc.go:206 +0x5a

goroutine 1 [select]:
github.com/minio/minio/vendor/github.com/minio/minio-xl/pkg/minhttp.ListenAndServeLimited(0x0, 0xc08202a270, 0x1, 0x1, 0x8)
    C:/Git/go/src/github.com/minio/minio/vendor/github.com/minio/minio-xl/pkg/minhttp/http_windows.go:205 +0x4bd
main.startServer(0xc08202e4e0, 0x5, 0x0, 0xc082002280, 0x8, 0xa, 0x0, 0x0, 0x0, 0x0, ...)
    C:/Git/go/src/github.com/minio/minio/server-main.go:151 +0x110
main.serverMain(0xc0820d60f0)
```

golang/go#13287

PR #408, opened 2016-02-04
checker: Validate pack checksums before unpacking

benmur:
This avoids reading a possibly invalid size at the end of a corrupted pack


fd0:
This is a great contribution, thanks! :+1:

PR #409, opened 2016-02-04
s3: Disable keepalive on minio HTTP client

benmur:
This avoids connections being used for too long and being terminated by the
S3 endpoint.

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"net/http"
 	"strings"
 
 	"github.com/minio/minio-go"
@@ -38,6 +39,11 @@ func Open(cfg Config) (backend.Backend, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	tr := &http.Transport{
+	    DisableKeepAlives: true,
+	}
+	client.SetCustomTransport(tr)
 
 	be := &s3{client: client, bucketname: cfg.Bucket}
 	be.createConnections()


benmur:
I don't know if there is a better way than disabling keepalive, if there is, please disregard this PR :) It made things a lot better for me for very long upload sessions but it may be a bit extreme. I have another PR coming that implements retries.

fd0:
Hi, thanks for raising this issue and even proposing a fix!

I've had a chat with the nice people from https://minio.io (we use their s3 client library), and it seems the problems is in the HTTP library of Go:
https://github.com/golang/go/issues/3514

The discussion can be found here: https://gitter.im/minio/minio?at=56b44a675cc201794d498cc9

This issue is fixed in Go 1.6, could you maybe give it a try with Go 1.6rc2 (and without disabling keepalive)?

benmur:
I've launched a lengthy check --read-data, with go 1.6rc2 and without disabling keepalive, I'll let you know how it goes.

fd0:
That's a good start, I'm looking forward to it.

If I understood the underlying issue correctly, the problem is that the Amazon service disconnects slow clients, especially when uploading data. Do you use an asymmetric connection to upload data to S3? If this is the case, then you'll probably need to do another backup to reproduce the issue.

fd0:
Disabling keepalive means that for every request a new connection is made to the server, including a TCP and TLS handshake. This is very inefficient and should be only enabled on request of the user, e.g. with a special option. I'll implement having free-form options in the form `restic -o s3.keepalive=no` shortly, if the issue persists we can add such an option.

fd0:
The integration tests fail because a file is not correctly formatted with `gofmt`: https://travis-ci.org/restic/restic/jobs/107102123#L223

If you want to make your own life easier, consider adding `fmt-check` as a pre-commit hook from https://github.com/edsrzf/gofmt-git-hook as described in
https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches

benmur:
Thanks for the fmt tip!

So here are the number of transient errors (500, EOF, connection reset by peer, ...) encountered with different versions of go, and with and without keepalive:
go 1.4.2, keepalive: about 1 every hour
go 1.4.2, no keepalive: about 1 every 4 hours (30% slowdown)
go 1.6rc2, keepalive: about 1 every 15 minutes
go 1.6rc2, no keepalive: about 1 every hour  (30% slowdown)

So unfortunately using go 1.6 made things worse for me. I'm even starting to wonder if something's wrong with my bucket.

About the connection, it's 400Mbps down, 20Mbps up, so it's asymmetric but I'm not sure each of the 10 concurrent uploads can be called slow - however I don't know after how long connections are recycled.

I agree that it would be very inefficient to disable keep alive by default, but your proposed option sounds great - I would implement it once you've added free-form options.

Thanks

fd0:
So, let's continue the discussion in an issue, and close this PR for now. I've created issue #415 for it.

PR #410, opened 2016-02-04
s3: Retry with an exponential backoff delay

benmur:
This is more of an RFC, since I don't know if retries are to be attempted all over the backend client code or in backends. Also, I'm having problems vendoring the backoff dependency (and I'm not sure of the maintainer wants it :)).

Anyway, this has helped a lot on my long backups (along with my other PR about keepalive), where S3 would fail in one way or another every few hundred requests, panicking on error.

Thanks for your comments!

### DIFF
--- a/backend/s3/s3.go
+++ b/backend/s3/s3.go
@@ -4,7 +4,11 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"log"
 	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff"
 
 	"github.com/minio/minio-go"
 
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for your contribution! I'm planning to implement retries (and exponential backoff sounds like a good idea, with some maximum waiting time), but it should be implemented on a different layer so that all backends can benefit from the code and we have less code duplication.

In addition, the minio s3 library may implement retries specifically for s3 at some later point.

If you're interested in contributing generic retries for backends, I'm happy to work with you :)

I'm closing this PR for now.

PR #418, opened 2016-02-07
Handle pack loading errors in rebuild-index

benmur:
Errors returned from backend.LoadAll() were not handled, leading to
these fatal errors from the unpacker trying to read the size from the end of
an empty buffer:

`seeking to read header length failed: bytes.Reader.Seek: negative position`

This change takes care of returning on error, as well as showing which pack
failed to load and validating pack integrity.

### DIFF
--- a/cmd/restic/cmd_rebuild_index.go
+++ b/cmd/restic/cmd_rebuild_index.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"fmt"
 
 	"github.com/restic/restic/backend"
 	"github.com/restic/restic/debug"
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for your contribution. Looks good and I'd like to merge it, could you run `gofmt` on the files so the integration test will succeed?

benmur:
Sorry about that. I actually ran `gofmt -w **/*.go` but somehow it did not do anything so I assumed all was good. I'll have a look in a few.

fd0:
Great, thanks. I'd suggest installing the `fmt-check` hook, but you already know about that one ;)

benmur:
@fd0 branch updated, thanks for your patience!

fd0:
Of course, thanks for your contribution!

Issue #415, opened 2016-02-06
backend: server disconnect leads to panic


fd0:
@benmur described an undesired behavior in #409: When the s3 server disconnects a client, restic panics and aborts the backup process. For the solution it was proposed to introduce re-trying at the s3 backend level. It looks like the minio library will add this functionality sooner or later.

We agreed that it is a good idea to have retry-functionality at the backend level, so all backends can benefit.

fd0:
@benmur I've drafted up a small patch that allows writing all communication between restic and s3 to a file, could you please apply that patch and try to reproduce the dropped connections again?

Please find the patch here: https://gist.github.com/3f5172b9037f3819d160, then run restic with the environment variable `RESTIC_S3_TRACE=/tmp/s3trace.log`.

benmur:
Thanks @fd0 that's great! I'll try that tomorrow. Do you prefer any kind of operation or is backup best?

fd0:
Whatever produced the errors most reliably ;)

benmur:
I have to let a rebuild-index finish first (opened #423 about that) but will definitely run with tracing on later.

Bregor:
It seems, I crashed in this issue too.
Restic built just now from commit `3d27751b6923bd7513439725195f5478edf55b24`
Doing first full backup.

Full output here: https://gist.github.com/4589159dc1df8ab8749d

cdrx:
minio/minio-go#330 has added support for retrying failed requests upstream

harshavardhana:
> minio/minio-go#330 has added support for retrying failed requests upstream

Currently the retry retries s3 specific errors i.e which come about in the s3 error response. Will be adding more network related errors to the list as well.

dominikschulz:
Is there anyone working on this? It's really annoying and I'd be willing to look into it.

The S3 backend probably should fall back to resumeable multipart uploads even if the file is smaller the 5MB for very slow connections and retry several times.

fd0:
@dominikschulz you can try fetching the latest version of minio-go (with `gb vendor`) and submit this as a PR, maybe they've already fixed it.

harshavardhana:
> @dominikschulz you can try fetching the latest version of minio-go (with gb vendor) and submit this as a PR, maybe they've already fixed it.
> 
> You can also try to implement retries in restic itself.
> 
> As far as I know, nobody is currently working on this.

[...]

PR #423, opened 2016-02-07
Use a new buffer for each pack in rebuild-index

benmur:
The same buffer was being used for downloading multiple packs and
when a bigger-sized buffer was reused to download a new pack, the call
to `io.ReadFull` from some backends would fail with an `unexpected EOF`
error because the number of bytes read would be smaller than the buffer
size.


fd0:
Thanks for bringing this up. I've stumbled across this error myself a day a go, but haven't gotten around to pushing the fix. I'm closing this PR, as the underlying issue must be fixed, no the symptom.

Please see  #415 for the fix.

Issue #387, opened 2016-01-12
Support many backups inside single s3 bucket


vshalts:
Hi,
Can you add support of path prefix for s3 backup?

Number of buckets is limited.  At the same time it is useful to have backups with different policies/lifecycle rules/storage classes etc. For example I want some backups to go to Glacier, but other stay in Infrequent Access Storage Class. Also in Glacier it is better to not keep everything in single backup due recovery of big backup will cost a lot. So, better to split backups by topics (photos, projects, etc). It is low probability to have shared data between photos and projects, so there is no win to keep them in single backup.

I want something like:
```
$ restic -r s3://region_name/bucket_name/backup_prefix_path init
```
will create backup in bucket bucket_name under backup_prefix_path.
it should be enough to support all required functionality.

fd0:
This is a good idea, thanks for bringing it up.

haswalt:
I too like this idea. We've just found here that we don't have enough buckets to backup our system whilst evaluating Restic

ckemper67:
I just wrote the code for this over the weekend while trying to improve the integration for GCS. 

Unfortunately those two changes may not be cleanly separated and I made a bunch of other changes at the same time. What is the best way to get these to you? Would a single pull request be ok or do you prefer me to tease those apart?

fd0:
Thanks for stepping up, please create a PR just for one functionality, so we can review it properly.

fd0:
#443 was merged a few minutes ago, it introduces custom prefixes for s3.

PR #430, opened 2016-02-11
Introduce a configurable object path prefix for s3 repositories.

ckemper67:
Fix for issue #387 

Prepends the object path prefix to all s3 paths and allows to have multiple independent
restic backup repositories in a single s3 bucket.

Removed the hardcoded "restic" prefix from s3 paths.

Use "restic" as the default object path prefix for s3 if no other prefix gets specified.
This will retain backward compatibility with existing s3 repository configurations.

Simplified the parse flow to have a single point where we parse the bucket name and the prefix within the bucket.

Added tests for s3 object path prefix and the new default prefix to config_test and location_test.


fd0:
This PR contains too many changes, please create smaller ones.

In addition: Why did copy the s3 code to gcs? This creates a lot code unnecessary code duplication. It is already possible to specify a different endpoint for the minio library...

I'm sorry, I'm unable to merge this PR as it is.

ckemper67:
That functionality was not supposed to end up in this branch quite yet. I
am still struggling a bit with git and must have accidentally merged my GCS
branch into the s3 branch. I will try to clean it.

I wanted to give a simpler syntax for handling GCS based repositories.
Ideally I can just create another S3 configuration object and hand that to
the s3 code.


On Sat, Feb 13, 2016 at 2:14 PM, Alexander Neumann <notifications@github.com

fd0:
> That functionality was not supposed to end up in this branch quite yet. I am still struggling a bit with git and must have accidentally merged my GCS branch into the s3 branch. I will try to clean it.

Ah, that explains a lot. Thanks for your effort!

> I wanted to give a simpler syntax for handling GCS based repositories. Ideally I can just create another S3 configuration object and hand that to the s3 code.

Is that really necessary? I think I need to understand what the difference between S3 and GCS is. Isn't it sufficient to just give the s3 backend a different endpoint, like `s3:https://gcsendpointhost.google.com:bucket` or something similar?

PR #443, opened 2016-02-13
Introduced a configurable object path prefix for s3 repositories to address #387

ckemper67:
Prepends the object path prefix to all s3 paths and allows to have multiple independent
restic backup repositories in a single s3 bucket.

Removed the hardcoded "restic" prefix from s3 paths.

Use "restic" as the default object path prefix for s3 if no other prefix gets specified.
This will retain backward compatibility with existing s3 repository configurations.

Simplified the parse flow to have a single point where we parse the bucket name and the prefix within the bucket.

Added tests for s3 object path prefix and the new default prefix to config_test and location_test.


fd0:
I'm investigating what's wrong with the s3 tests. Additionally, the files `config.go` and `s3.go` in `backend/s3` aren't properly formatted with gofmt. I'd suggest installing the pre-commit hook locally, https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches

fd0:
The integration tests fail because the s3 tests call `Open()` directly with a config, and there the prefix is not specified, so it defaults to the empty string. This does not work with the current code, an empty prefix generates URLs with paths like `/restictestbucket//data/0000000000000000000000000000000000000000000000000000000000000000-0`. For such paths, the minio server returns a 301 HTTP status code. I'd suggest using the following function for `s3path()`:

```go
func s3path(prefix string, t backend.Type, name string) string {
	var path string

	if prefix != "" {
		path = prefix + "/"
	}
	path += string(t)

	if t == backend.Config {
		return path
	}
	return path + "/" + name
}
```

While we're at it, you could make that function a method of the `s3` struct, so you don't have to pass the prefix as a parameter every time.

backend/s3/config.go
```diff
@@ -13,53 +13,28 @@ type Config struct {
...
-
-	if len(rest) == 2 {
-		// assume that just a region name and a bucket has been specified, in
-		// the format region/bucket
-		cfg.Endpoint = rest[0]
-		cfg.Bucket = rest[1]
-	} else {
-		// assume that a URL has been specified, parse it and use the path as
-		// the bucket name.
+	       	path = strings.SplitN(s, "/", 3)
```
fd0: I think you've removed too many tests here. Using t

backend/s3/config.go
```diff
@@ -13,53 +13,28 @@ type Config struct {
...
+		cfg.Endpoint = path[0]
+		path = path[1:]
+        } else if strings.HasPrefix(s, "s3:http") {
```
fd0: Use `switch` instead of repeated `if ... else if ...`.

backend/s3/config.go
```diff
@@ -13,53 +13,28 @@ type Config struct {
...
-		// assume that just a region name and a bucket has been specified, in
-		// the format region/bucket
-		cfg.Endpoint = rest[0]
-		cfg.Bucket = rest[1]
-	} else {
-		// assume that a URL has been specified, parse it and use the path as
-		// the bucket name.
+	       	path = strings.SplitN(s, "/", 3)
+		cfg.Endpoint = path[0]
+		path = path[1:]
```
fd0: This one is clever, I like it! Makes the verbose tests I wrote unnecessary.

backend/s3/config.go
```diff
@@ -73,9 +48,25 @@ func ParseConfig(s string) (interface{}, error) {
 		if url.Scheme == "http" {
 			cfg.UseHTTP = true
 		}
-
-		cfg.Bucket = url.Path[1:]
+		path = strings.SplitN(url.Path[1:], "/", 2)
+	} else if strings.HasPrefix(s, "s3:") {
+	       s = s[3:]
```
fd0: I think we can combine this test case with the first one, it's the same code.

backend/s3/config.go
```diff
@@ -73,9 +48,25 @@ func ParseConfig(s string) (interface{}, error) {
...
+	cfg.Bucket = path[0];
+	if len(path) > 1 {
+	   cfg.Prefix = path[1]
+	} else {
+	   cfg.Prefix = defaultPrefix
+        }
 
 	return cfg, nil
 }
+
```
fd0: This file is not properly formatted with `gofmt`, please run it on the file and force-push a new commit.

backend/s3/config_test.go
```diff
@@ -9,18 +9,38 @@ var configTests = []struct {
...
 		Bucket:   "foobar",
+		Prefix:   "restic",
+		UseHTTP:  true,
+	}},
+	{"s3:http://hostname:9999/bucket/prefix/directory", Config{
+		Endpoint: "hostname:9999",
+		Bucket:   "bucket",
+		Prefix:   "prefix/directory",
 		UseHTTP:  true,
 	}},
```
fd0: We should add a test for the empty prefix, e.g. with `s3:http://hostname:999/bucket/`.

backend/s3/s3.go
```diff
@@ -13,13 +13,12 @@ import (
 )
 
 const connLimit = 10
-const backendPrefix = "restic"
 
-func s3path(t backend.Type, name string) string {
+func s3path(prefix string, t backend.Type, name string) string {
```
fd0: Make this function a method on `s3`, so we don't have to explicitly pass the prefix every time.

fd0:
First round of comments is done, thanks!

backend/s3/config.go
ckemper67: I tried, but I cannot figure out a simple way to handle the different prefix length that I have to strip off.

backend/s3/config_test.go
ckemper67: I have done that and changed s3path to allow the empty prefix now.

But we now have to consider if an empty prefix should be left empty in s3path or if I should always expand it to "restic" in s3path as the original code did.

i.e. do we want to allow to write to the root of the bucket or do we always want to fore the "restic" prefix?

ckemper67: It occurred to me that we have a similar problem if you specify a trailing slash in the object prefix. I have decided to trim it off and added tests for that as well.

The simplest way would probably be to always trim trailing slashes off the end of the input string. That would handle the trailing slash in the bucket name (it would force the "restic" prefix) and also handle this new case.

backend/s3/config.go
fd0: Something along the lines of:
```go
if strings.HasPrefix(s, "s3:") {
    s = s[3:]
    if strings.HasPrefix(s, "//") {
        s = s[2:]
    }
   [...]
}
```

ckemper67: That works :-)

I also pulled out the creation of the config into a separate method.

backend/s3/config_test.go
fd0: I suggest using `path.Clean()` on the path. This will trim the trailing slash and will also resolve things such as `../` and `./` and `foo//bar`.

Regarding the empty prefix: I think we should allow it in the code, but the parsing routine should use `restic` instead of the empty string, so we stay compatible with older backups.

ckemper67: Done.

I like the suggestion of enforcing "restic" in parsing but allow the empty string in code. It really bugged me that the code was not backward compatible.

fd0: Being backwards compatible is very important, for the project and for me personally :)

fd0:
LGTM!

PR #447, opened 2016-02-14
Fix #442: set blocks and blocksize in fuse

### LINKED ISSUE #442: "du" displays "0 bytes" for all files in FUSE mounted snapshots
I tried mounting the snapshots and verify with "du" if certain directories contain plausible amounts of data and if there might be any directories with a lot of big files (like games) that I could exclude. That didn't work in the first attempt as "du" shows 0 bytes for every file:

    # du /mnt/t/snapshots/2016-02-13T09:08:18+01:00/bin
    0       /mnt/t/snapshots/2016-02-13T09:08:18+01:00/bin

The original directory:

    # du /bin
    11468   /bin

Google found me an explanation at https://github.com/MediaFire/mediafire-fuse/issues/15 and the recommended flag did work here, too:

    # du --apparent-size /mnt/t/snapshots/2016-02-13T09:08:18+01:00/bin
    11289   /mnt/t/snapshots/2016-02-13T09:08:18+01:00/bin

Maybe you can fix this by providing a faked blocksize. If not a note regarding --apparent-size in the User Manual would be nice.

ckemper67:
I added code to set the blocks and blocksize similar to [ MediaFire/mediafire-fuse#15](https://github.com/MediaFire/mediafire-fuse/issues/15). Also added a test.

I cannot test if this fixes the issue on my machine, since I cannot reproduce the original problem. OS X does not report 0 file sizes, even when using the "du" from the GNU coreutils. Maybe the OS X fuse layer does a similar calculation?

Maybe someone can test this on another OS?

### DIFF
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -15,6 +15,9 @@ import (
 	"golang.org/x/net/context"
 )
 
+// The default block size to report in stat
+const blockSize = 512
+
 // Statically ensure that *file implements the given interface
 var _ = fs.HandleReader(&file{})
 var _ = fs.HandleReleaser(&file{})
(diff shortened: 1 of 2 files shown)


fd0:
Hey, thanks for your work! Could you rebase this branch on current master, so the first two commits aren't included in the PR and we can have a proper review?

fd0:
I've just run the fuse tests on Linux, looks good!

PR #451, opened 2016-02-15
Support custom SSH ports in URL of SFTP-repo

fawick:
Allow supplying an override for the default port (or the one configured in SSH config) similar to the way the user override is possible. When needing to specify a custom port,  the sftp.//... format must be used.


fd0:
This is a great PR, thanks a lot for your work!

For all other options for the SSH connection to the sftp server, the ssh config file should be used. Alternatively, I'll add free-form options for the backend soon^TM (in the form `restic -o sftp.port=1234`)

Issue #212, opened 2015-07-04
Storage Backend: Amazon Cloud Drive


e2b:
The "Unlimited Everything" plan of [Amazon Cloud Drive](https://www.amazon.com/clouddrive/home/) is a quite affordable backup storage option. Amazon Cloud Drive has its own RESTful API.

klauspost:
Just thought the same thing.

I may look at this once #21 is in place - uploading without compression seems like a waste of bandwidth.

fd0:
That depends on your use case ;)

stapelberg:
There’s some proof of concept code I found: http://sprunge.us/fdQF — it requires that an oauth token is in /tmp/token.json, but seems to work for me.

Motivated people could turn that into a clean backend for ACD :).

jsimonetti:
+1 from me :)
I compile the backend @stapelberg found, and it indeed appears to work and use code from rclone.

kisscool:
+1 too
I'm trying to use Restic + acd_cli (FUSE python client for ACD), but it's very unreliable for now : some operations do not behave as expected (file truncate, rename) and Restic randomly panics as a result.
A working ACD storage backend would do wonder.

fd0:
I'm currently reworking the interface to the backends, this includes a radical simplification. This is basically done, but not yet merged. For the plan, see #383, the PR is #395.

Afterwards it will be much easier to implement new backends.

Before implementing many new backends, I'd like to have a list of rules that services we write backends for must fulfill, this may include that a test instance of the service must be available that we can run the integration tests against.

Do you by chance know whether there is a test service for ACD we can use for tests?

kisscool:
As far as I know, there is no test instance for ACD. No mention of such a thing here : https://developer.amazon.com/public/apis/experience/cloud-drive/content/restful-api

But the https://github.com/ncw/rclone project already did an ACD backend in go. It seems to be fairly reusable as demonstrated by the proof of concept shared by @stapelberg .

klauspost:
Actually, looking at the revised interface, it would be reasonably easy to do a full wrapper for rclone filesystems. Maybe that way separate implementations isn't needed?

kisscool:
I don't know what @fd0 vision for Restic future is, but it would seem logical to focus the project on the backup intelligence instead of re-implementing a ton of remote filesystems one by one. Besides both project licenses are compatible. 
It would also solve the worries about how to test all those backends.

[...]

PR #461, opened 2016-02-21
Onedrive backend

chbmuc:
This is a first draft of a new storage backend for Microsoft Onedrive. Some things are still on my TODO-List, but it is already working:
- [x] Folder creation on first access
- [x] List() is limited to 200 items (could be increased easily, but I think support for paging would be better)

Apart from that, review comments or other feedback would be appreciated.

Onedrive uses OAUTH for authentication. At the moment you need to provide a config file with the token credentials, Client-ID and Client-Secret. I think it would be best, to provide a helper application with a local webserver to ease the generation of the token. 

Client-ID and Client-Secret can be obtained in the Microsoft Developer Center. I've seen that rclone is providing them in the source code,  but I'm unsure if this is admissible.

### DIFF
--- a/src/cmds/restic/global.go
+++ b/src/cmds/restic/global.go
@@ -10,6 +10,7 @@ import (
 
 	"restic/backend"
 	"restic/backend/local"
+	"restic/backend/onedrive"
 	"restic/backend/rest"
 	"restic/backend/s3"
 	"restic/backend/sftp"
@@ -251,6 +252,9 @@ func open(s string) (backend.Backend, error) {
 		return s3.Open(cfg)
 	case "rest":
 		return rest.Open(loc.Config.(rest.Config))
+	case "onedrive":
+		debug.Log("open", "opening ondrive repository")
+		return onedrive.Open(loc.Config.(onedrive.Config))
 	}
 
 	debug.Log("open", "invalid repository location: %v", s)
@@ -286,6 +290,9 @@ func create(s string) (backend.Backend, error) {
 		return s3.Open(cfg)
 	case "rest":
 		return rest.Open(loc.Config.(rest.Config))
+	case "onedrive":
+		debug.Log("open", "opening ondrive repository")
+		return onedrive.Open(loc.Config.(onedrive.Config))
 	}
 
 	debug.Log("open", "invalid repository scheme: %v", s)
--- a/src/restic/backend/onedrive/config.go
+++ b/src/restic/backend/onedrive/config.go
@@ -1,0 +1,45 @@
+package onedrive
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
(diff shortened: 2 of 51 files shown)


fd0:
Hey, thanks a lot for your contribution! Is there a test service that we can use during CI tests for this backend? What's the difference between `Client-ID`, `Client-Secret` and the OAuth token?

chbmuc:
- The problem with CI tests is, that I don't know of any onedrive service except the one from Microsoft. To upload files, we need an account. Free accounts are available, but we have to keep the credentials secret, to avoid abuse.
- Every "onedrive client application" needs to be registered with Microsoft. You can do this in the developer center, but you have to agree to the Terms of Service. Upon registration, you get a Client-ID and a Client-Secret for you application (see also: https://auth0.com/docs/connections/social/microsoft-account).
- Access to a users onedrive account is authenticated with a token. For permanent access we need a "refresh token" (see https://dev.onedrive.com/auth/msa_oauth.htm - chapter "Code flow"). I'll provide a helper application for this as an example.
- Reading the configuration from environment variables should be possible. I just thought, that generating a file would improve usability.

chbmuc:
- Folder creation is now working
- Paging for long List() is implemented

fd0:
May I suggest that you install the `check-fmt` pre-commit hook as described here https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches

fd0:
So, in order to make the current CI tests pass, you'll need to reformat the following files with `gofmt`:
```
not formatted with `gofmt`:
src/restic/cmd/restic/global.go
src/restic/location/location.go
```

fd0:
I've asked the rclone people about their decision to just include the Client-Secret in the source code: https://github.com/ncw/rclone/issues/372

chbmuc:
I've fixed the formatting and did a rebase

fd0:
Just a quick heads-up: I haven't forgotten, I just don't have any time at the moment.

dbadrian:
Has there been any progress on this lately? I am very interested in this feature.

I wish I could help (myself) and help finish this, but I have no time to learn Go first XD. I am happy to help in testing, if there is any need.

fd0:
I'm sorry to acknowledge that there hasn't been any progress. The blocker here is that we (as a project) need to decide how we handle backends for third-party services, especially in terms of testing. This also applies to the request for an Amazon OneDrive backend, you can find the discussion here: https://github.com/restic/restic/issues/212#issuecomment-230400007

gngeorgiev:
Hey guys, I am not sure why you have problems with the CI, maybe I am missing something. You can have a free account right? Then, in travis you can encrypt the credentials or provide them with environment variables https://docs.travis-ci.com/user/environment-variables.

fd0:
I'm going to close this PR for now, since the questions outlined #212 need to be answered before I'll merge any code. We can always open the PR again (or start a new one).

Issue #55, opened 2014-11-18
Add tags to snapshots


fd0:

mappu:
[snapshot-comments-v1.diff.txt](https://github.com/restic/restic/files/133561/snapshot-comments-v1.diff.txt) here's patch.

Use optional `-c` when performing backup to set comment on snapshot. Comment is optional (default emptystring). Comment is displayed in `restic snapshots` output.

fd0:
Could you please create a PR, so we can do a proper code review? If you need assistance, please let me know :)

fd0:
Hey, any news here? If you need any assistance creating a PR, please let me know! You can also drop by `#restic` on `irc.freenode.net` if you like :grinning:

fawick:
#463 is the relevant PR.

fd0:
Thanks for your help!

This is not exactly the same, the code by @mappu (#463) adds a `comment` field, which is slightly different from the way I imagined the feature: Tags (or labels) can be specified multiple times, whereas the comment field is just one string. So we'll keep this issue open for now.

mappu:
Thank you @fawick .

This patch has some downsides (hidden paths in snapshots output during there's no control over output format, ...). It should be very easy to take idea from this patch and make a tags system instead, but, display of tag on `snapshots` output should be discussed.

PR #463, opened 2016-02-21
Cleaned up patch for #55.

fawick:
Changes to _test.go were not part of the patch. I added them to be able
to run the tests after patching. 

All other changes are from #55, https://github.com/restic/restic/files/133561/snapshot-comments-v1.diff.txt by @mappu

### DIFF
--- a/src/restic/cmd/restic/cmd_snapshots.go
+++ b/src/restic/cmd/restic/cmd_snapshots.go
@@ -82,7 +82,7 @@ func (cmd CmdSnapshots) Execute(args []string) error {
 	}
 
 	tab := NewTable()
-	tab.Header = fmt.Sprintf("%-8s  %-19s  %-10s  %s", "ID", "Date", "Source", "Directory")
+	tab.Header = fmt.Sprintf("%-8s  %-19s  %-10s  %s", "ID", "Date", "Source", "Comment")
 	tab.RowFormat = "%-8s  %-19s  %-10s  %s"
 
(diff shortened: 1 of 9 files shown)


src/restic/cmd/restic/cmd_snapshots.go
```diff
@@ -119,13 +119,7 @@ func (cmd CmdSnapshots) Execute(args []string) error {
 			continue
 		}
 		id := sn.ID()
-		tab.Rows = append(tab.Rows, []interface{}{hex.EncodeToString(id[:plen/2]), sn.Time.Format(TimeFormat), sn.Hostname, sn.Paths[0]})
-
-		if len(sn.Paths) > 1 {
```
fd0: This is strange. @mappu why did you remove the other paths from the table?

fd0:
Looks good, thanks for your work @fawick!

We'll need to refactor this sometime later, we can't just keep adding parameters to the `Snapshot()` function...

src/restic/cmd/restic/cmd_snapshots.go
mappu: With 80 character default terminal width it was unreadable to have all fields visible.

fd0:
I've thought about this patch, and I've decided not to merge it in this form. Adding the comment field is nice, but removing the most important thing from the snapshot output, the directories that have been backed up, is not good.

Thanks for your work!

PR #464, opened 2016-02-21
Add REST backend

fd0:
This is a port of the original work by @bchapuis in
https://github.com/restic/restic/pull/253

TODO:
- [x] Improve documentation

### DIFF
--- a/doc/REST_backend.md
+++ b/doc/REST_backend.md
@@ -1,0 +1,56 @@
+REST Backend
+============
+
+Restic can interact with HTTP Backend that respects the following REST API.
+
+## HEAD /config
+
+Returns "200 OK" if the repository has a configuration,
+an HTTP error otherwise.
+
+## GET /config
+
+Returns the content of the configuration file if the repository has a configuration,
+an HTTP error otherwise.
+
+Response format: binary/octet-stream
+
+## POST /config
+
+Returns "200 OK" if the configuration of the request body has been saved,
+an HTTP error otherwise.
+
+## GET /{type}/
+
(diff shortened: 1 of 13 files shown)


doc/REST_backend.md
```diff
@@ -0,0 +1,56 @@
...
+an HTTP error otherwise.
+
+Response format: binary/octet-stream
+
+## POST /config
+
+Returns "200 OK" if the configuration of the request body has been saved,
+an HTTP error otherwise.
+
+## GET /{type}/
```
fawick: The doc should list what would be acceptable values for {type}.

src/restic/backend/rest/backend_test.go
```diff
@@ -0,0 +1,87 @@
+// DO NOT EDIT, AUTOMATICALLY GENERATED
```
fawick: Should the source and the generation recipe for the auto-generation be part of the repo?

fd0: Oh, it is. Here you can find the go generate comment: https://github.com/restic/restic/pull/464/files#diff-6397b881264cb4356775a0e5a3be197fR15

The file is generated from the file https://github.com/restic/restic/blob/master/src/restic/backend/test/tests.go by the program https://github.com/restic/restic/blob/master/src/restic/backend/test/generate_backend_tests.go

doc/REST_backend.md
fd0: Good point.

fd0:
This works with my toy Ruby server and passes all the tests: https://gist.github.com/07c8a0cda34e257565e9

PR #465, opened 2016-02-21
Merging restic-server

fawick:
This PR aims to merge bchapuis/restic-server as a subtree (src/restic/cmd/restic-server). I also took the liberty of `gb vendor`ing all requirements for the tests. So far, the server is not modified.

The benefit of using the "github.com/stretchr/testify/require" in the test cases package seems small to me in comparison of how much code needs to be vendored, as 'require' needs even more external packages that would not be used otherwise. Hence, I suggest that before merging this we try to get rid of the dependency.


fd0:
Superseded by #467

PR #466, opened 2016-02-21
Cleaned up the sftp parsing logic.

ckemper67:
Simplified and cleaned up the sftp parsing logic. Added support to path.Clean the directory. Added additional tests.

### DIFF
--- a/src/restic/backend/sftp/config.go
+++ b/src/restic/backend/sftp/config.go
@@ -3,6 +3,7 @@ package sftp
 import (
 	"errors"
 	"net/url"
+	"path"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


src/restic/backend/sftp/config.go
```diff
@@ -11,58 +12,47 @@ type Config struct {
 	User, Host, Dir string
 }
 
-// ParseConfig extracts all information for the sftp connection from the string s.
+// ParseConfig parses the string s and extracts the sftp config. The two
+// supported configuration formats are sftp:user@host:port/directory and
+// sftp:user@host:directory. The directory will be path Cleaned and can be
+// an absolute path.
```
fd0: I think the comment is not accurate: What about the format `sftp://user@host`? The prefix isn't mentioned at all...

fd0:
Hey, thanks for the contribution! Looks really good already, just a small nitpick: Could you correct the comment?

ckemper67:
I corrected the comment and added a sentence about how to pass an absolute
directory since that is not completely obvious.

PR #468, opened 2016-02-21
Added missing handle to the s3.Stat log message output

ckemper67:
I enabled logging to better understand the s3 interaction flow and noticed that the handle was missing in Stat().


fd0:
Oh, that's a nice catch. Thanks for your contribution!

PR #471, opened 2016-02-22
Refactor src/restic/cmd to src/cmds

fawick:
This PR is to satisfy the usual folder structure of a gb project.

src/cmd/\* will be copied to src/restic/cmd duing build.go runs.


fd0:
I like where this is going, but you'll need to update the path in `run_integration_tests.go`. :+1:

fawick:
Oh, I did not even notice that this is even run during CI...

fawick:
`go help packages` states the following reason why `cmd` was a bad name:

```
There are four reserved names for paths that should not be used  
for packages to be built with the go tool:  
  
- "main" denotes the top-level package in a stand-alone executable.  
  
- "all" expands to all package directories found in all the GOPATH 
  trees. For example, 'go list all' lists all the packages on the local system.  
  
- "std" is like all but expands to just the packages in the standard Go library.  
  
- "cmd" expands to the Go repository's commands and their internal libraries.  
```

fd0:
I think it's fine now. We could've also have it in `src/cmd` and copy that to `restic/cmd` in `build.go`, but this way (with `cmds` instead of `cmd`) less people will run into strange things with the reserved name `cmd`.

Issue #438, opened 2016-02-12
Feedback after entering the repo password


Mebus:
Hi!

After entering a password restic does not give any feedback if the password was correct. 
Please print something like "password correct, starting".

Mebus

fw42:
Isn't the fact that it doesn't say "password is wrong" feedback enough that it was right?

Mebus:
How do I know then if it isn't waiting for my network connection or something else to tell me 15 minutes later, that the password was wrong.

rawtaz:
I agree that there should be some confirmation that downloading the index has started. I am now and then on a connection where it takes like 5 minutes to download the index before the actual backup starts. I have with time learnt that unless is says something else, it's probably fine once I entered the repo password, but I have many times wondered about the current status, and checked the network monitor to make sure it appears to do something.

I definitely think there should be some feedback.

Regards, Leo

12 feb 2016 kl. 19:32 skrev Mebus <notifications@github.com>:

fd0:
We can probably add a line saying `password is correct`, but only for interactive terminals and when `--quiet` is not specified.

If anybody want's to implement this, I'll gladly accept a PR.

mholt:
I'm a first-time restic user tonight, and I typed my password wrong (on purpose, just for kicks), and this is the output:

```
$ restic -r /tmp/backup backup .
enter password for repository: 
unable to open repo: no key could be found
```

"no key could be found" scares me, because to me, that means that my key was deleted or something, and my data is inaccessible. I would rather see "incorrect password" with the ability to retry.

As for a "correct password" notice: so far haven't seen it being necessary (but I haven't gotten through the user guide yet either!). In fact I think it would be weird to see one since most programs I type a password for don't say that I typed it correctly; they just print some output when they start a slow or long-running process.

rawtaz:
The last point you raise is what the issue is about; ‎after entering the password restic starts downloading the index or something, and this can take a while. In some cases it can take 5-10 minutes. But even If it takes just one minute, it's weird that one doesnt know what's going on after entering the password correctly (when there's no output).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        Från: Matt HoltSkickat: torsdagen den 10:e mars 2016 04:03Till: restic/resticSvara till: restic/resticKopia: rawtazÄmne: Re: [restic] Feedback after entering the repo password (#438)I'm a first-time restic user tonight, and I typed my password wrong (on purpose, just for kicks), and this is the output:

[...]

PR #484, opened 2016-03-11
Change ErrNoKeyFound message

mholt:
For #438. I was just going to change it to "wrong password" but then I saw that it might actually be the case that no key could be found, so I changed it to what I did. Let me know if you'd like something different!

### DIFF
--- a/src/restic/repository/key.go
+++ b/src/restic/repository/key.go
@@ -16,7 +16,7 @@ import (
 
 var (
 	// ErrNoKeyFound is returned when no key for the repository could be decrypted.
-	ErrNoKeyFound = errors.New("no key could be found")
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for your contribution! And especially thank you for writing caddy, I love it and use it regularly!

mholt:
Glad you like it!

I'm really excited about restic in that it, in some ways, solves the same security/usability problems for backup like Caddy does for web. I know this PR was just a petty change, but I'll keep using restic as it matures and hope to be back for more. :wink:

And separating the error types sounds like a good idea.

fd0:
Oh, awesome! I'm looking forward to receiving more bug reports and PRs from you ;)

Many such small PRs and improvements can add up to quite a lot...

PR #489, opened 2016-03-31
add debian stable help

xmaka:
add a little clue to debian stable users, to install 'go' from the repositories


fd0:
Thanks for your contribution!

PR #490, opened 2016-04-01
add backup --exclude-file

### LINKED ISSUE #416: Allow reading exclude patterns from a file
It was proposed on IRC that it is easier to have restic read the exclude patterns for backup from a file, so that the command line does not get so messy. This issue is about implementing an `--exclude-file` parameter that reads the patterns from a file, one per line.

Thor77:
Read exclude-patterns from a file (one pattern per line), lines starting with `#` are ignored.

Closes #416


fd0:
Looks good, thanks for your contribution!

Issue #494, opened 2016-04-13
minor: `gb test` leaves temporary `mount` behind


jzacsh:
## Output of `restic version`
```sh
$ ./restic version
restic 0.1.0 (v0.1.0-479-g32a5778)
compiled at 2016-04-13 12:53:04 with go1.6
```

## Expected behavior
`gb test` should leave no state behind _(at least for successful test runs)_. Feel free to just close this bug if you disagree :)

## Actual behavior
`gb test` left a temporary mount (`$TMPDIR/restic-test-mount-987358968`) after all tests passed and `gb test` exited 0.

## Steps to reproduce the behavior
1. `git clone https://github.com/restic/restic`
2. `go get github.com/constabulary/gb/...` _(exits 0)_
3. `gb build` _(exits 0)_
4. `gb test`, which output:

  ```sh
  restic/debug
  restic/fs
  restic/test
  restic/location
  cmds/restic-server
  restic/fuse
  restic/filter
  restic/worker
  restic/crypto
  restic/pack
  restic/pipe
  restic/backend/local
  restic/backend/mem
  restic/backend
  restic/backend/s3
  restic/backend/rest
  restic/backend/sftp
  restic/backend/test
  restic/checker
  restic/repository
  restic
  cmds/restic
  ```
  _(I **think** the mount appeared at `restic/test`, but I haven't poked around at any code yet)_

5. `mount | grep restic` shows test mount is still around. This time, it was:

  ```sh
  [MY_CLONE_DIR]/restic on $TMPDIR/restic-test-mount-987358968 type fuse (ro,nosuid,nodev,relatime,user_id=1000,group_id=1000)
  ```

---

**Actually**, I ran a few extra [commands I saw here](https://github.com/restic/restic/issues/368#issuecomment-208815375), before simply running the step 1 admit to above _(which I found in [the CONTRIBUTING doc](https://github.com/restic/restic/blob/32a577860226522b4e879edc7dfd98e061db06ca/CONTRIBUTING.md#development-environment))_. I assume this is harmless to reproducing, but just in case, here's the **actual** last 17 commands in my shell:
```sh
git clone https://github.com/restic/restic && cd restic && go run build.go
go run build.go
./restic --help
./restic help
./restic --help  backup
./restic backup --help
go get github.com/constabulary/...
go get github.com/constabulary/bg/...
go get github.com/constabulary/gb/...
gb build
bin/restic version
gb test
mount
mount | grep restic
umount restic
sudo umount restic
[...]

PR #496, opened 2016-04-13
doc: point to correct `gb` arg for `go get`

jzacsh:
per side-chat in https://github.com/restic/restic/issues/494#issuecomment-209591073


fd0:
Looks good to me, thanks for your contribution!

PR #497, opened 2016-04-16
Expand environment-variables in exclude-files

Thor77:


fd0:
LGTM, let's wait for the CI tests.

PR #500, opened 2016-04-17
Fix exclude filters with trailing slash

### LINKED ISSUE #499: --exclude seems confused by leading & trailing slashes
## Output of `restic version`
built from HEAD....
```sh
$ git rev-parse HEAD
5b33a7a903fc0c6d834be20c7399fe3f5ba1f210
$ bin/restic version
restic compiled manually
compiled at unknown time with go1.6
```

## Setup
Runnig my backup with:
```sh
export TMPDIR=/tmp/
restic/restic backup \
  --repo $HOME/bak/restic \
  --exclude-file $HOME/.config/restic-exclude
  $HOME
```

## bad report :(

Sorry for the poor bug-report quality, below, but i just spent **way** too long writing up my exact steps in a "micro" case, and then found opposite behavior _(still broken, though)_....

- In my real-world backup (of `$HOME`), a leading slash confuses `--exclude-file` contents
- In my test-case backup _(of a `mktemp -d` w/2 dirs and 2 files)_, a leading slash has no effect _(restic still backups everything indiscriminately)_ but removing the **trailing** slash from  `--exclude-file` lines solves the problem.

So I'm not sure exact reproduction steps, unfortunately :( I only know that:
- **expected**: simple[1] paths should be excluded by their presence in an `--exclude-file`
- **actual**: said paths are almost always - _except above-mentioned attempts_  - backed up anyway

---

[1]: _(no globs, no env. variables, just simple paths)_

fd0:
fixes #499


jzacsh:
nice, `filepath.Clean` ftw. :+1:

jzacsh:
fyi, just tested this branch, and sure enough it works as advertised.

PR #509, opened 2016-05-08
Allow reading data from stdin

### LINKED ISSUE #255: Feature: Backup data read from stdin
This issue tracks implementing the feature that restic reads data from stdin and saves it to the repo (with dedup etc).

I can imagine several use cases, e.g.:
 * Saving SQL dumps directly: `mysqldump db | restic backup -`
 * Saving complete devices: `dd if=/dev/sda | restic backup -`

Before implementing this, we need to decide how to save this in the json data structures. I'd like restic to behave consistent with this, e.g. when the repo is mounted with the fuse backend, data read from stdin should appear as a file. I'd also like to be able to write this data to a file.

Any other thoughts or use cases that haven't been mentioned so far? Do you like the syntax (`restic backup -`), or should we rather use `restic backup --stdin`?

fd0:
This PR adds the options `--stdin` and `--stdin-filename` to the `backup` command. It allows reading backup data from stdin, so the following use case becomes possible:

```
$ mysqldump | restic backup --stdin --stdin-filename production.sql
```

It's not yet fully optimized, it only reads chunks sequentially.

TODO:
- [x] Add a test
- [x] Add documentation

Closes #255

### DIFF
--- a/doc/Manual.md
+++ b/doc/Manual.md
@@ -182,6 +182,22 @@ see [`filepath.Match`](https://golang.org/pkg/path/filepath/#Match) for syntax.
 Additionally `**` exludes arbitrary subdirectories.  
 Environment-variables in exclude-files are expanded with [`os.ExpandEnv`](https://golang.org/pkg/os/#ExpandEnv).
 
+## Reading data from stdin
+
+Sometimes it can be nice to directly save the output of a program, e.g.
+`mysqldump` so that the SQL can later be restored. Restic supports this mode of
+operation, just supply the option `--stdin` to the `backup` command like this:
+
+    $ mysqldump [...] | restic -r /tmp/backup backup --stdin
+
+This creates a new snapshot of the output of `mqsqldump`. You can then use e.g.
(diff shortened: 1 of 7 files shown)


pvgoran:
Does this also include the ability to backup a single special file (for example, a disk partition)?

fd0:
Yes, if you're feeding this into restic, e.g. with `dd`:
```
$ dd if=/dev/sda1 | restic backup --stdin --stdin-filename sda1.raw
```

pvgoran:
So just specifying the special file(s) in command line will still not work?

fd0:
That works, but will not read from the device. Restic is smart enough to recognize the block device file and will only backup meta information about it. This is a sane default because it allows doing e.g. `restic backup /`.

`borg` has a mode to treat all arguments as files (just open and read their contents):
```
  --read-special        open and read special files as if they were regular
                        files
```

If you feel this is necessary (I don't think we need something like this), please open an issue and describe your use case.

Issue #513, opened 2016-05-09
Files are read again, despite unchanged timestamp/size


viric:
I'm testing current master.

I run restic backup with absolute paths, like this:
```
restic backup /srv/repo /srv/repo2 /srv/repo3 /srv/mydir/other/data
```

Mounting with fuse, I notice that the root directory is like "/srv", so the snapshot contains directly repo, repo2, ...

So I've seen (by stracing) that it keeps on reading a big 5GB file not modified since weeks ago, and we have a snapshot of yesterday: /srv/repo/bigfile.fossil

Is it possible that there is a bug in checking node existence by path? Maybe it stores "repo/bigfile.fossil", but checks for existence of a node "/srv/repo/bigfile.fossil" or so.

viric:
I guess that it is not using a parent snapshot (I don't see any line about parent snapshot at least).

If I specify the snapshot from yesterday (-p 76510eda) it does say that it uses it, but it also reads the whole bigfile.fossil.

viric:
We have a daily backup. With "restic cat snapshot ...." I see that none of our daily backups have parent, but we run it every day with the same cmdline. It is a script in cron.

viric:
Point 1: the samePaths() in cmd_backup are broken, because the snapshots have the path sorted, and the cmdline not. And they are compared index-to-index.

viric:
I think it'd be slightly ggbetter if we looked for the newest snapshot that *contains* all the paths requested. I see that I also had troubles because I was supplying a path *twice*, and it goes by length. I'll do a PR for this.

fd0:
Thanks for your contribution! This is indeed a bug :)

viric:
I think I found the other part, about archivePipe. You try to traverse the fs and the parent snapshot in parallel, so the order of the paths is important because that determines if you advance the parent file or not.

So, if I supply to backup: "/srv/b /srv/c /srv/hello/a", these look in order in absolute path. The NEW files will be traversed in that order. But the parent snapshot will be traversed by the 'basename' of the input path: "/srv/hello/a /srv/b /srv/c"  (a < b < c). That gives a big lack of matches in archivePipe.

Does it look right? I don't have a fix, but that's what I think is happening, seeing the comparisons in ArchivePipe.compare.

fd0:
Hey, is this already solved? Can we close this issue?

PR #514, opened 2016-05-09
Better backup parent snapshot search. Part of #513

viric:
I look for the newest snapshot that contains all supplied paths to backup.

This fixes the finding of good parent if 1) there are a different number of paths supplied, but contained in a snapshot, and 2) if cmdline paths are not given in alphabetical order.

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -180,11 +180,15 @@ func samePaths(expected, actual []string) bool {
 		return true
 	}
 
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for your contribution! However, this does not work as advertised, this returns true:
```go
samePaths([]string{"foo", "bar"}, []string{"bar", "foo", "baz"})
```

The code tests if `expected` is a subset of `actual`. Was that your intention?

fd0:
Ah, testing for subset is actually better than before. I like this.

PR #515, opened 2016-05-09
Traverse paths in the same order as parent snapshot

viric:
This is the 2nd partial fix to #513.

The archivepipe requires the snapshot paths and the backup paths to be
traversed in the same order, and they were sorted differently: the backup paths
by full path, and the snapshot by basename path.


fd0:
This looks good, thanks for the code! Before merging it, it would be great to have a test for the bad behavior...

PR #518, opened 2016-05-11
Implement prune

### LINKED ISSUE #92: Implement "optimize repository" command
Should check for optimal deduplication.

### LINKED ISSUE #479: Add 'prune' command
In #459, the `optimize` command was removed due to a bug. This issue is about adding a new command called `prune` which basically does the same as `optimize`, just with a better name (and hopefully without the bug).

### LINKED ISSUE #572: User Interface: Snapshot Expire Options?
Hi,

I'd like to collect some opinions on the user interface for removing snapshots from the repo. The `prune` command is nearly done (see #518), the last remaining part is a good user interface.

What the command does is listing all snapshots, then going through all the data blobs stored in the repo and removing (by repacking) those that aren't referenced any more. It's not as efficient as it could be (and it's rather slow), but it works and is correct (as far as I can see).

So I just need to add a function that lists the snapshots and removes those that are to be deleted. All options described below are applied to snapshots grouped by host and list of directories separately. If you have snapshots from two different hosts in the repo, the snapshots for each host will be processed separately, e.g. if only the last 10 snapshots should be kept, this means 10 backups per host. The same is applied for list of dirs, if you have snapshots for `/foo` and `/bar, /baz`, you'll end up with the last 10 backups of `/foo` and the last 10 backups of `/bar, /baz`.

I've come up with the following options for `prune`:
 * `--host=X` only remove snapshots that were created for host `X` (to be specified in combination with the other options)
 * `--keep-last=n` by using this option restic will only remove snapshots if there are more than `n` for each host/dirs.
 * `--keep-all=DUR` keep all backups made during `DUR`, e.g. `24h`, this will not touch any backups made in the last 24 hours.
 * `--keep-hourly=DUR` keep hourly backups for `DUR`, which can for example be `2d`. This instructs restic to go through the snapshots in reverse chronological order (newest to oldest) from the current time two days back (48 hours). For each hour, it removes all but the last backup. If we have three backups at 23:12, 23:30 and 23:55 restic will remove all but the last.
 * `--keep-daily=DUR` keep daily backups for `DUR`, e.g. `2w`. This will keep at most one backup per day for the last two weeks.
 * `--keep-monthly` and `--keep-yearly` are treated the same.
 * @Thor77 on IRC suggested another setting `--max-repo-size`, which is harder to implement (because it requires traversing all snapshots) and will maybe implemented later.

Is this enough/sufficient?

Is it a good idea to have defaults for the values (e.g. 24h hourly snapshots, 4w daily, 1y monthly, 10y yearly) so the user doesn't have to bother specifying the command-line options every time? Or is that too invasive, because just running `prune` without any options will probably remove some snapshots/data?

I'm looking forward to your comments! :smiley:

fd0:
This is my work in progress branch while implementing a prune function to remove old data from the repository.

TODO:
- [X] Implement repacking of blobs into new packs
- [x] Find out which blobs aren't in use any more by traversing all trees.
- [x] Combine the two into the `prune` command
- [x] Add `forget` command (for details see #572) to remove snapshots
- [x] Add documentation for the new commands

Closes #479
Closes #92
Closes #572


fd0:
I've nearly finished the implementation of a `prune` command and need some feedback on how to specify which snapshots to remove/keep, please add your comments to #572. Thanks!

fd0:
Phew, this is done. Is there anybody willing to test it? :smiley:

fd0:
This is merged now, thanks for your comments!

PR #525, opened 2016-05-24
Added Minio.io configuration steps to run as backend for restic.

koolhead17:


doc/Manual.md
```diff
@@ -357,6 +357,29 @@ You can then easily initialize a repository that uses your Amazon S3 as a backen
 For an S3-compatible repository without TLS available, use the alternative URI
 protocol `s3:http://server:port/bucket_name`.
 
+# Create a Minio Server repository
```
fd0: I think this would fit better as a sub-heading to the s3 backend example, could you add another `#`?

doc/Manual.md
```diff
@@ -357,6 +357,29 @@ You can then easily initialize a repository that uses your Amazon S3 as a backen
...
+
+[Minio](https://www.minio.io) is an Open Source Object Storage, written in Go and compatible with AWS S3 API.
+
+### Pre-Requisites
+
+* Download and Install [Minio Client](https://minio.io/download/) & [Minio Server](https://minio.io/download/).
```
fd0: Does a user really have to download the client?

fd0:
Thanks for adding this!

doc/Manual.md
koolhead17: sure

doc/Manual.md
koolhead17: sure. Will clean the client part. :)

fd0:
Merged, sorry for the delay.

PR #548, opened 2016-07-25
idset.go: micro-optimise away redundant scan

mappu:
I saw this small thing when studying the code.


fd0:
This is a nice finding, thank you!

PR #552, opened 2016-07-29
Explicitly Close() obj after ReadFull()

benagricola:
Restic makes heavy use of short reads from S3 when performing incremental backups. It runs `GetObject` from `minio-go` and reads part of the returned stream (e.g. 45 bytes from a specific offset) and then returns.

This leaves the HTTP connection managed by `minio-go` containing unread data, and in this situation the `minio-go` library will create a new connection for subsequent requests even though restic is finished with the previous object. Performing an incremental backup of a 7GiB filesystem I noticed connection counts from Restic to the S3 endpoint of up to 30,000 when the open file limit was raised high enough to allow this.

Explicitly calling `Close()` after `io.ReadFull()` causes the remaining data to be discarded and allows `minio-go` to handle connection reuse effectively. The same incremental backup described above will use on average 15 to 30 connections for the duration of the backup.

### DIFF
--- a/src/restic/backend/s3/s3.go
+++ b/src/restic/backend/s3/s3.go
@@ -97,7 +97,17 @@ func (be s3) Load(h backend.Handle, p []byte, off int64) (int, error) {
 	defer func() {
 		be.connChan <- struct{}{}
 	}()
-	return io.ReadFull(obj, p)
+
+	// This may not read the whole object, so ensure object
+	// is closed to avoid duplicate connections.
+	n, err := io.ReadFull(obj, p)
+	if err != nil {
+		obj.Close()
+	} else {
+		err = obj.Close()
+	}
+	return n, err
+
 }
 
 // Save stores data in the backend at the handle.


fd0:
Hey, thanks for giving this a try!

src/restic/backend/s3/s3.go
```diff
@@ -97,7 +97,12 @@ func (be s3) Load(h backend.Handle, p []byte, off int64) (int, error) {
 	defer func() {
 		be.connChan <- struct{}{}
 	}()
-	return io.ReadFull(obj, p)
+
+	// This may not read the whole object, so ensure object
+	// is closed to avoid duplicate connections.
+	n, err := io.ReadFull(obj, p)
+	obj.Close()
```
fd0: This discards the error that `Close()` may have returned. What do you think about returning the error from `Close()` in case `ReadFull()` did not return any error?

src/restic/backend/s3/s3.go
```diff
@@ -97,7 +97,17 @@ func (be s3) Load(h backend.Handle, p []byte, off int64) (int, error) {
 	defer func() {
 		be.connChan <- struct{}{}
 	}()
-	return io.ReadFull(obj, p)
+
+	// This may not read the whole object, so ensure object
+	// is closed to avoid duplicate connections.
+	n, err := io.ReadFull(obj, p)
+	if err != nil {
+		obj.Close()
```
benagricola: I assume in this situation that the important part is the Read error rather than the close error.

I wonder if it's possible for a Read error to cause `obj` to be already-closed, in which case this would throw an error attempting to close it twice...

benagricola:
Also one thing to note, there's been a number of changes to `minio-go` upstream which help to reduce the concurrent connection count (also to do with connections being recreated instead of reused). Would be good to have updated minio-go pulled in.

fd0:
Yes, I agree. Thanks for your contribution!

PR #558, opened 2016-08-01
Detect a devel version correctly in LDFlags()

vrischmann:
I use a tip build where `runtime.Version()` return something like this

```
devel +111d590 Fri Jul 29 01:09:55 2016 +0000
```

This patch detects these builds.


fd0:
Oh, good catch, thanks! I need to reverse this test rather soon now, since the `key=value` syntax is the new standard, so it's easier to test for old Go versions.

PR #559, opened 2016-08-01
Fix the debug environment variable name in the manual

vrischmann:
The debug variable name is wrong in the manual :)


fd0:
Thank you very much!

PR #562, opened 2016-08-03
Issue-535: restic respect GOMAXPROCS env variable depending on go version

### LINKED ISSUE #535: restic should respect GOMAXPROCS env var
In current master, and since long, restic overrides GOMAXPROCS to NumCPU().

It would be nice to just respect GOMAXPROCS env var, if it is present. And going for NumCPU() is already the default in recent go versions, iirc.

damekr:
I've tested it on go1.4 and 1.6. It works, but I don't like expression like "if version < "go1.5". If you have any suggestions please let me know, I'll change them.

Closes #535

### DIFF
--- a/src/cmds/restic/main.go
+++ b/src/cmds/restic/main.go
@@ -2,17 +2,23 @@ package main
 
 import (
 	"fmt"
-	"os"
-	"runtime"
-
 	"github.com/jessevdk/go-flags"
+	"os"
 	"restic"
 	"restic/debug"
+	"runtime"
 )
 
 func init() {
 	// set GOMAXPROCS to number of CPUs
-	runtime.GOMAXPROCS(runtime.NumCPU())
+	if runtime.Version() < "go1.5" {
+		gomaxprocs := os.Getenv("GOMAXPROCS")
+		debug.Log("restic", "read GOMAXPROCS from env variable, value: %s", gomaxprocs)
+		if gomaxprocs == "" {
+			runtime.GOMAXPROCS(runtime.NumCPU())
+		}
+	}
+
 }
 
 func main() {


fd0:
Hey, thanks for giving this a try.

I've got the following remarks:
 * Please shorten the git commit message's first line, there's usually no need to mention the issue, just include a line "Closes #535" in the body of the commit message, see https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits and http://chris.beams.io/posts/git-commit/. Mentioning the issue ID automatically creates a link to the issue.
 * Comparing the Go version as a string works this way, that's okay. String comparison in Go is well defined.
 * It seems you implemented my suggestions wrong: If the environment variable is set (`gomaxprocs != ""`), do nothing, the runtime will take care of setting `GOMAXPROCS`. Your code does it just the other way around, so exchange `!=` for `==` in the if.

You can amend your commit (with `git commit --amend`) and force-push (`git push -f`) to your branch, the new commit will then be shown here.

damekr:
Hey, sorry I've done this too fast :) Now should be correct.

fd0:
Heh, no problem. In this PR, you can push/edit/amend as often as you want. Just let us know when you think the PR is ready, then we'll look at the code.

fd0:
Could you please run `gofmt` and amend the commit? This fails here: https://travis-ci.org/restic/restic/jobs/149567058#L256

Have a look at https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches, there we've listed hints for providing patches. This also includes a pre-commit hook for git which checks gofmt locally.

damekr:
Hi,
Now should be ok.

fd0:
Looks good, thanks!

Issue #567, opened 2016-08-09
Backup also files on windows with longer pathnames than 255 chars (e.g. from node).


stuertz:
## Output of `restic version`
restic.exe version
restic 0.2.0 (v0.2.0-8-ge4a9905)
compiled at 2016-08-08 17:37:01 with go1.6.3

## Expected behavior
Backup also files on windows with longer pathnames than 256 chars (e.g. from node).

## Actual behavior
> restic backup c:\work
[...]
work\applications\cs.activitystream\trunk\cs\activitystream\web\base\node_modules\autolinker\node_modules\gulp-header\node_modules\gulp-util\node_modules\dateformat\node_modules\meow\node_modules\read-pkg-up\node_modules\find-up\node_modules\path-exists\readme.md: GetFileAttributesEx c:\work\applications\cs.activitystream\trunk\cs\activitystream\web\base\node_modules\autolinker\node_modules\gulp-header\node_modules\gulp-util\node_modules\dateformat\node_modules\meow\node_modules\read-pkg-up\node_modules\find-up\node_modules\path-exists\readme.md: Das System kann den angegebenen Pfad nicht finden.
[...]

Files are not backuped, but logged.

## Steps to reproduce the behavior
Generate a path with long pathnames and try to backup it on Windows.

## Estimated solution
Use the right windows API Funtion, that is aware of (Unicode) Long Pathnames

stuertz:
see also https://github.com/golang/go/issues/3358

fd0:
Thanks for reporting this issue. It was already known, but buried deep in the comments of another issue: https://github.com/restic/restic/issues/339#issuecomment-211831435

A workaround is mentioned in https://github.com/restic/restic/issues/339#issuecomment-226207599

I'll repeat my statement from the other issue so that we can track progress here:
Properly handling long paths on Windows is not easy to do with Go. I don't have a Windows development environment, and the Go stdlib does not yet support long paths, see https://github.com/golang/go/issues/3358

So, if someone wants to step up implementing this for restic, I'm glad to assist.

stuertz:
Hi, I would like to help out with the issue, as far as I do have such an environment :-)
But I'm currently very new into go and also with restic (did a lot of python and c/c++ before). I tried already a bit around with patching the os.FileOpen in src/restic/fs/file_all.go. But that didn't help here.

[...]

PR #570, opened 2016-08-13
Fix 567

stuertz:
This patches #567.
Currently not all tests run on windows as expected (some don't work on windows before :-) ). See the attached log file for details. And an explicit test for long pathnames is also currently missing.

Before I dig into the testing issues, I'd like to get some feedback on the change.

As Alexander mentioned, this could be a starting point to implement a layer between the the native go functions and the functions used by restic. I implemeted a layer to most of the os.\* functions (at least those that uses a pathname)  and replaced every usage I found in the code base, even those used only in the unix files. 

In he end my use case, backup a directory with long path names already works with the patch.


fd0:
Hey, thanks for giving this a try!

I've thought about this and I think we should proceed as follows:
 * There already is a top-level `restic/fs` package, let's use this for all file-related operations
 * Create wrapper-functions for the file/dir functions in the package `fs`
 * Make `fixpath()` a private function (that is not exported), and call it only when run on Windows
 * Replace all calls to functions in package `os` in restic code (not in tests) that take a path by calls to the wrapper functions in package `fs`

What do you think?

stuertz:
OK. Sounds good. Its not too much. Your list is identical to patch.go
But: 
1. there are more os dependencies in restic than you quoted. I like to agree to replace only the listed functions. But that means you don't get rid of the os. Is 
2. There is also filepah.Walk calls, that need to be replaced in archiver.go. I will add them also to fs.


Regarding the errors: This is the log from a clean checkout:
[15:35:51,74] C:\work\go\src\restic-trunk\restic>c:\work\go\bin\gb test
restic/debug
restic/fs
restic/test
restic/location
restic/backend/mem
cmds/restic-server
restic/backend/rest
restic/backend/test
restic/backend/s3
restic/backend/sftp
restic/filter
restic/worker
restic/crypto
restic/pack
restic/pipe
restic/backend/local
restic/backend
restic/checker
restic/repository
# restic
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestDelayedWalkTree (0.03s)
FAIL
# cmds/restic
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestBackupNonExistingFile (0.82s)
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestBackupMissingFile1 (0.80s)
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestBackupMissingFile2 (0.92s)
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestBackupDirectoryError (0.96s)
index d5697fe0 has old format
index d5697fe0 has old format
←[31mhelpers.go:184: unexpected error: write |1: Die Pipe wurde beendet.←[39m

--- FAIL: TestFind (1.69s)
        integration_test.go:46: repository initialized at c:\Temp\restic-test-987033399\repo
FAIL
FATAL: command "test" failed: [c:\Temp\gb700567763\cmds\restic\testmain\_test\restic.test.exe]: exit status 1

Could that be a problem with the tar util, which is not a standard tool on windows. I currently an old version from gnuwin32-utils?

fd0:
> there are more os dependencies in restic than you quoted. I like to agree to replace only the listed functions. But that means you don't get rid of the os.

True, but we're abstracting filesystem interaction, which means it's not necessary to wrap all functions in `os`.

> There is also filepah.Walk calls, that need to be replaced in archiver.go. I will add them also to fs.

Indeed, good catch.

Your test failures may be cause by `tar`, for the CI tests I'm using a freshly downloaded version from sourceforge: https://github.com/restic/restic/blob/master/appveyor.yml#L20

fd0:
Just to leave this here for you to read: I'm planning to restructure the restic source code somewhat similar to the structure this article describes: https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1

This is also helpful if you're just getting started in writing Go: https://medium.com/@benbjohnson/structuring-applications-in-go-3b04be4ff091

stuertz:
Thanks for pointing me to the docs. Currently had some problems with understanding the fs.File interface, but hope I got it right.

Using an actual tar, doesn't fix the tests on my machine. In case of good tests on Appveyor, this must be a problem with my local environment.

Any comments are welcome.

fd0:
Thanks, I'll have a look later today. I'll probably merge your commits into one (squashing) because we don't need the Work In Progress commits in the git history.

For the future I have the following suggestions for you:
 * Configure your editor to automatically run `gofmt` or `goimports` on the code on each save. Install `fmt-check` as a pre-commit git hook locally: https://github.com/edsrzf/gofmt-git-hook This helps you to not check in code that is not properly formatted.
 * Try to write your git commits in the same style as the rest of the project, this is described here: https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits
 * When reworking this, use `git rebase` to reorder and/or squash commits so we have a clean history

fd0:
Looks good, thanks!

PR #583, opened 2016-08-21
Fix progress output on Windows

stuertz:
The windows cmd shell is not aware of ANSI escape sequences and does print them uninterpreted to the console. This is ugly. Added a function to generate platform specific string for the escape sequence. On Windows this will be 79 white spaces with a trailing \r.

This is a very simple cosmetic change, so I decided to not open an issue, but fix it directly. If its required by the project to write down an issue for this, I can for sure create one.

### DIFF
--- a/src/cmds/restic/cleanup.go
+++ b/src/cmds/restic/cleanup.go
@@ -57,7 +57,7 @@ func RunCleanupHandlers() {
 func CleanupHandler(c <-chan os.Signal) {
 	for s := range c {
 		debug.Log("CleanupHandler", "signal %v received, cleaning up", s)
-		fmt.Println("\x1b[2KInterrupt received, cleaning up")
+		fmt.Printf("%sInterrupt received, cleaning up\n", ClearLine())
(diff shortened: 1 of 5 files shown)


fd0:
Looks good, I'll have a closer look in the evening. Thanks for your contribution!

I've thought about switching to e.g. https://github.com/cheggaaa/pb at some point, or at least have a look how that progress bar is implemented on the different OS.

stuertz:
Yep. Nowerdays there should be a full implemented cross platform solution available for simple problems like a progressbar. ;-) Nothing, that should be reinvented.

fd0:
May I suggest installing the pre-commit hook to check gofmt again? :wink:

stuertz:
Sorry. Yes for sure. I currently work on several different machines with different emacs installations. I will check all my repo instances tomorrow and fix the code. Again, sorry.

fd0:
You don't have to apologize, just let me know when the PR is ready :smiley:

fd0:
LGTM, thanks!

PR #607, opened 2016-09-04
Add local index for metadata

### LINKED ISSUE #386: Slow backup speed for large (incremental) backups
Hi,

I notice that `restic` doesn't seem to be optimised for unchanged files which makes backing up very large repos (~500GBs for example with some large files) quite time consuming.

Not sure if this would work internally, but you must be keeping a map from file->chunks so can't you notice that a file is the "same" (e.g. same hashcode, date/time access and file size?) and then simply clone the file->chunks map which should be significantly quicker?

I hesitate to raise this because having seen the calibre of this tool I can't believe this hasn't already occurred to you :-).

Thanks!

### LINKED ISSUE #454: Slow, although it is just checking timestamps
I have 164809 files to backup regularly (about 60GB)... Every time I run "restic backup" the report doesn't go beyond 33MB/s and checking with strace it's only doing lstat() calls.

This renders about 20 minutes per backup. I wonder what is restic doing, because being almost all files unmodified, it shows the steady 33MB/s and I understand that it only needs to lstat() them, which is exactly what restic already does at the first step of backup just to show the total size, in 6 or 7 seconds.

Is it just the cpu time spent in checking the contents for that same file/timestamp is already present in a restic previous snapshot?

fd0:
The code in this PR adds a local metadata cache for tree blobs, index and snapshot files.

Closes #386
Closes #454

### DIFF
--- a/src/cmds/restic/global.go
+++ b/src/cmds/restic/global.go
@@ -13,6 +13,7 @@ import (
 	"restic/backend/rest"
 	"restic/backend/s3"
 	"restic/backend/sftp"
+	"restic/cache"
 	"restic/debug"
 	"restic/location"
 	"restic/repository"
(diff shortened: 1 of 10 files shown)


fd0:
I'm closing this PR for now, it is not ready.

PR #626, opened 2016-09-18
Add "-x", "--one-file-system" option

### LINKED ISSUE #370: Add --exclude-other-filesystem support
From duplicity:

> --exclude-other-filesystems
>   Exclude files on file systems (identified by device number) other than the  file  system  the  root  of  the source directory is on.

Given

> /dev/mapper/sys-root on / 
> proc on /proc
> /dev/mapper/sys-home on /home

The following would backup **/** but not **/home** nor **/proc**

> restic backup / --exclude-other-filesystem

rfjakob:
Equivalent to rsync's "-x" option.

Closes #370

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -19,13 +19,14 @@ import (
 )
 
 type CmdBackup struct {
-	Parent        string   `short:"p" long:"parent"                  description:"use this parent snapshot (default: last snapshot in repo that has the same target)"`
-	Force         bool     `short:"f" long:"force"                   description:"Force re-reading the target. Overrides the \"parent\" flag"`
-	Excludes      []string `short:"e" long:"exclude"                 description:"Exclude a pattern (can be specified multiple times)"`
-	ExcludeFile   string   `long:"exclude-file"                      description:"Read exclude-patterns from file"`
-	Stdin         bool     `long:"stdin"                             description:"read backup data from stdin"`
-	StdinFilename string   `long:"stdin-filename"    default:"stdin" description:"file name to use when reading from stdin"`
-	Tags          []string `long:"tag"                               description:"Add a tag (can be specified multiple times)"`
+	Parent         string   `short:"p" long:"parent"                    description:"use this parent snapshot (default: last snapshot in repo that has the same target)"`
+	Force          bool     `short:"f" long:"force"                     description:"Force re-reading the target. Overrides the \"parent\" flag"`
+	Excludes       []string `short:"e" long:"exclude"                   description:"Exclude a pattern (can be specified multiple times)"`
+	ExcludeOtherFS bool     `short:"x" long:"one-file-system"           description:"Exclude other file systems"`
+	ExcludeFile    string   `long:"exclude-file"                        description:"Read exclude-patterns from file"`
+	Stdin          bool     `long:"stdin"                               description:"read backup data from stdin"`
+	StdinFilename  string   `long:"stdin-filename"    default:"stdin"   description:"file name to use when reading from stdin"`
+	Tags           []string `long:"tag"                                 description:"Add a tag (can be specified multiple times)"`
 
 	global *GlobalOptions
 }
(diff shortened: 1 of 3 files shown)


rfjakob:
Closes #370

fd0:
Hey, that already looks great! Let me know if you need any assistance.

pvgoran:
Maybe use `--one-file-system` as the long option name, like `rsync` and `tar` do?

rfjakob:
Slowly getting there ;)

rfjakob:
The problem with `--one-file-system` is that it may be more than one, because we can have more than one source directory! Duplicity uses `--exclude-other-filesystems`.

fd0:
Hm. Usually I'm all for using the same arguments as other popular tools, but in this case I think `--one-file-system` is not a good fit since there may be several file systems. Running `restic backup --exclude-other-filesystems / /mnt/foo /mnt/bar` should backup at least from the three FS that are listed.

So, I'd go for `--exclude-other-filesystems`. This also matches the other exclude options.

fd0:
Ok, we agree, great!

rfjakob:
BTW while I am waiting for the CI to complete: Can I run the test suite locally? This is what I get from a naive attempt:
```
$ go run run_integration_tests.go
unknown CI environment
exit status 1
```

fd0:
The `run_integration_tests.go` is just for CI, it sets up things and so on. You can use `gb test` locally. It won't run things that depend on external services, e.g. the s3 backend isn't tested, but for this issue it should work just fine.

rfjakob:
Found it, it's documented in CONTRIBUTING.md, thanks!

fd0:
The last error is about the stdlib `errors` package, use `restic/errors` for awesome stacktraces (provided by https://github.com/pkg/errors)

pvgoran:
Well, IDK, both `tar` and `rsync` can have multiple path arguments as well. For `rsync`, the man page even explicitly says that `--one-file-system` does not interfere with specifying paths from different filesystems.

BTW, `attic` uses `--do-not-cross-mountpoints`, which is a more verbose and non-ambiguous variant.

rfjakob:
CI passes, should be good to merge.

I don't care too much about the name, but I do care about having the feature, so let's not bikeshed this too much.

rfjakob:
That tar and rsync both use "--one-file-system" and both allow multiple source directories actually convinced me to rename it. @fd0 you file with that?

fd0:
I'm okay with that ;)

fd0 (approved):
LGTM :+1:

fd0:
Hm, I think there's a somewhat obscure issue with the current implementation. I'm not sure if it's worth opening an issue about (and fixing it), so maybe we can discuss it here.

At the moment we have a list of allowed device IDs, and when the dirs given on the command line are walked, restic checks that items are located on one of the allowed devices.

But what happens when one of the allowed devices is mounted below one of the other directories? Let's say, `/dev/sda1` is mounted at `/mnt/foo` and again at `/mnt/bar/baz/foo`, then running `restic backup --one-file-system /mnt/foo /mnt/bar` would walk down to `/mnt/bar/baz/foo`, because it is in the list of allowed devices. The same may happen when there's a bind mount... Do we need to handle this case?

ping @rfjakob

pvgoran:
IMHO it should be handled. Adding an additional backup target to the command line should not have such far-reaching consequences as modifying the scope of existing backup targets.

pvgoran:
I'd prefer even bind mounts to be detected and skipped, but it may be difficult to implement this well (device ID will no longer help us), and rsync doesn't do it. :)

rfjakob:
Maybe we should check if the device id is different from that of the
respective parent dir. This effectively skips all mountpoints.

pvgoran:
Except for inner-filesystem binds.

fd0:
What about having just one allowed device ID per directory passed to the `backup` command? We could make `map[string]uint64` and select the longest matching prefix with the current path from the keys of that map to find the current device ID.

pvgoran:
I don't know how the scanning process work, but maybe it can remember the "root" directory (the backup target passed to restic) from where the file/directory that's currently being scanned originated?

fd0:
I've created a new issue for this, let's move the discussion there: https://github.com/restic/restic/issues/645

PR #637, opened 2016-10-02
Use path.Join to create the s3 object name within the bucket.

ckemper67:
path.Join already automatically skips empty path segments when
joining, so this simplifies the s3Path code.

### DIFF
--- a/src/restic/backend/s3/s3.go
+++ b/src/restic/backend/s3/s3.go
@@ -3,6 +3,7 @@ package s3
 import (
 	"bytes"
 	"io"
+	"path"
 	"restic"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


fd0:
Looks good, simplifications are always desirable! Why did you rename the variable from `path` to `objName` in the code?

ckemper67:
I needed to do the rename because Join is imported from path and the
identifiers collided.

fd0:
Oh, okay, that's a very good reason. Thanks for your contribution!

fd0:
Ah, for the record: it would've also worked without the rename, but I think this way it's clearer. Thanks!

PR #638, opened 2016-10-05
Added long paths fix for samba network shares

hmsdao:
Also check if long paths expression is already included in paths.

### DIFF
--- a/src/restic/fs/file.go
+++ b/src/restic/fs/file.go
@@ -28,19 +28,19 @@ func fixpath(name string) string {
 		abspath, err := filepath.Abs(name)
 		if err == nil {
 			// Check if \\?\UNC\ already exist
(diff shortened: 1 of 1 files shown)


fd0:
Thanks for your contribution! I just have one (minor) request, then this can be merged!

fd0 (changes requested):
Please use backticks instead of double quotes to delimit the strings with backslashes, this way the code remains readable. Backticks are used to delimit raw strrings and the backslash character is a normal character thath doesn't need escaping. Example here: https://play.golang.org/p/iXD-ow_xqa

hmsdao:
Am I suppose to do any more work on this PR?

fd0 (approved):

fd0:
Nop, looks fine, thanks!

PR #646, opened 2016-10-14
don't print status info when running in the background

### LINKED ISSUE #381: Do not output to terminal, if 'bg'
It'd be nice that, if the process is put into background (ctrl-z, bg), it stopped printing numbers on stdout, unless it is fg again.

There are posix function calls that allow this, or signals. I don't remember. scp does this, if I remember correctly.

stkw0:
Try to fix #381 . It will only work for linux for now. 

For checking if we are running in the background I've  used the idea from http://stackoverflow.com/a/14894437

### DIFF
--- a/src/cmds/restic/background.go
+++ b/src/cmds/restic/background.go
@@ -1,0 +1,9 @@
+// +build !linux
+
+package main
+
+// IsProcessBackground should return true if it is running in the background or false if not
+func IsProcessBackground() bool {
+	//TODO: Check if the process are running in the background in other OS than linux
+	return false
+}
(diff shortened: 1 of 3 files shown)


fd0 (changes requested):
Nice work so far. Make a new file called `background_linux.go` and put the function `IsProcessBackground` in there. Make a second new file `background.go` with a build tag of `// +build !linux` and put an empty function `Is ProcessBackground` which always returns false into that file.

fd0 (approved):

fd0:
LGTM, thanks for your contribution!

Issue #272, opened 2015-08-20
Storage Backend: OpenStack Object Storage (Swift)


ar-jan:
Could OpenStack Object Storage be supported by restic? For example RunAbove object storage looks interesting, since it's cheaper than Amazon S3, and especially outgoing traffic is much cheaper.

fd0:
According to https://www.openstack.org/software/openstack-storage/, it already supports the S3 API, so it should work with restic.

ar-jan:
Ah, excellent!

ar-jan:
It seems this needs some more work and/or that it would be useful to support native Swift. Apparently OpenStack providers can choose not to enable S3 emulation, [like RunAbove](https://community.runabove.com/share/topic/s3-emulation-enabled-on-storage).

I'm also not sure if they did support S3 emulation how exactly one would use it currently - would you substitute AWS_ACCESS_KEY_ID with the Swift username (email address), and AWS_SECRET_ACCESS_KEY with the password? I suppose there'd also need to be a way to provide other service endpoints than AWS regions.

fd0:
did you try to use restic with the s3 backend with swift? did it work?

ar-jan:
I tried, though all RunAbove examples use Swift, and require other environment variables than S3, e.g. Tenant name ([example](https://community.runabove.com/kb/en/object-storage/how-to-store-your-backup-with-duplicity-and-openstack-swift.html)). I tried filling the AWS environment variables with RunAbove username and password, and using  s3://auth.runabove.io/v2.0/container_name, and got "unsupported protocol scheme". But I guess that's expected, since restic's S3 backend interprets the this URL as an Amazon S3 region name.

How could one currently specify an s3:// backend with a non-Amazon endpoint?

fd0:
Ok, this needs to be investigated. I'll reopen the issue for now.

retrack:
Hi there,

Tested with non-Amazon endpoint at Exoscale ( S3 storage backed with Pithos.io)

It fails with the following message at `init`stage:

    unsupported protocol scheme

Using debug mode did not provide more information.

fd0:
This is a limitiation of the s3 backend code (you cannot specify a server other than Amazon at the moment), and it's tracked in #315

byo:
Hi,

I was able to create a base implementation of native swift support which you can find here: https://github.com/byo/restic/tree/swift-backend.

[...]

PR #648, opened 2016-10-18
Implement OpenStack swift support

byo:
This PR implements support for OpenStack Swift backends requested in #272.


fd0:
Thanks a lot for your work! Before merging this at least we'd need to do the following things:
 * Add documentation to the manual
 * Setup CI tests for it

jayme-github:
Thanks for this implementation.
I did some tests and get a lot of HTTP 408 error messages (with "OVH Cloud Storage") although the data seems to get uploaded correctly:

```
2016/10/31 09:56:19 Unsolicited response received on idle HTTP channel starting with "HTTP/1.0 408 Request Time-out\r\nCache-Control: no-cache\r\nConnection: close\r\nContent-Type: text/html\r\n\r\n<html><body><h1>408 Request Time-out</h1>\nYour browser didn't send a complete request in time.\n</body></html>\n"; err=<nil>
2016/10/31 09:56:19 Unsolicited response received on idle HTTP channel starting with "HTTP/1.0 408 Request Time-out\r\nCache-Control: no-cache\r\nConnection: close\r\nContent-Type: text/html\r\n\r\n<html><body><h1>408 Request Time-out</h1>\nYour browser didn't send a complete request in time.\n</body></html>\n"; err=<nil>
 ```

byo:
@fd0:

Hi, sorry for a bit of delay here. I'll try to add docs and resolve issues this or next week.

Regarding lack of hostname, in swift the hostname can actually mean different things depending on the authentication method. If keystone server is being used, we only give the URL of this auth server. After authentication is done, keystone server does give the list of swift server URLs (one auth server could be used for multiple swift regions). Alternative authentication methods (legacy one and preauth) require actual swift server URL to be used.

Passing the configuration through environment is also consistent with official swift client. What's only passed through restic URL is the name of container so that one swift server can be used for many independent repositories. To let the hostname (and more robust URLs) be used in the future, I currently require empty one - that's why there are three slashes in swift url: 'swift:///...'.

Regarding CI setup - what are current guidelines about using external services? Can we spawn some internal swift server for tests only or should we use external provider for that? If external provider, how can we securely store credentials to such servers?

byo:
@jayme-github: I'll try to directly contact some people in OVH about this, will see if this is specific to their implementation or some bug. Even though I also observed those warnings, I didn't see any corruption caused by this yet.

rledisez:
@byo @jayme-github I'm from the OVH Cloud Storage team. I want to give some explanation to the 408 problem.

We use haproxy as a load balancer in front of our Swift cluster. In haproxy we configured:
- the use of keep alive;
- timeout the inactive connections after 10 seconds.

Problem is that the HTTP library from golang manage of pool of connections. So, sometimes it opens connections that will not be used. When a connections is inactive for 10 seconds, haproxy close it with the response "408 Request Time-out". This is not a problem at all as the connection was not in use. But golang print it. AFAIK, there is no way to tell golang to ignore this "non-error".

So, it's not really a problem on our side, but more a golang one, displaying error that are not really errors. FYI, rclone or svfs are also "impacted"

You can read some here: http://blog.haproxy.com/2014/05/26/haproxy-and-http-errors-408-in-chrome/

If our customers complain too much, we will consider implementing the workaround. But we would prefer if the real cause was fixed (golang HTTP lib)

jayme-github:
@rledisez thanks for clarifying that!

@byo I' seeing quite slow "restore" speeds here (around 1.5MiB/s via a 10MBit/s connection) while "backup" speed is much higher. Can you confirm that?

fd0:
@jayme-github restore still runs sequentially for all chunks, so it's not yet as  fast as it could be. This is not backend-specific.

jayme-github:
@fd0 I see, thanks!

jayme-github:
I've rebased and adapted to #741 (PR https://github.com/byo/restic/pull/1).

@fd0 what's your opinion regarding integration tests against a remote service? I might be wrong but setting up swift looks pretty complex to me.
I did already ask the guys from http://trystack.org regarding their swift service (as is is not working for me atm). @rledisez would OVH offer some king of "sponsoring" for restics integration tests?

byo:
@jayme-github thanks for taking care of that, I merged your branch and rebased on top of current master, history is much cleaner now. I haven't yet run testes on some real swift backend, hope to run something tonight.

jayme-github:
I did run integration tests against couple of times (they say I've used 600GB bandwidth - sounds quite a lot)

byo:
Still works with OVH after merge and rebase.

fd0:
Thanks for working on this, all of you. I've thought about it and I think running the integration tests against a live service is a good idea. We can tune down the amount of traffic for the swift tests though.

Looking forward to hearing from OVH. @rledisez are you interested in providing test credentials we can run our CI tests against?

jayme-github:
Alternatively I've set up integration tests via docker on travis but that only works for linux builds (because OSX is missing docker support)

fd0:
FYI: I think it's fine to have CI tests for swift only on Linux.

rledisez:
Sorry for the delay @fd0, @jayme-github. Let me get back to you in few days about sponsorship.

jayme-github:
I've added integration tests via docker at https://github.com/byo/restic/pull/2

byo:
@jayme-github Thanks again for your help, swift backend being tested as I write this :)

jayme-github:
@fd0 is there anything left we can do to get this merged?

fd0:
@jayme-github the next step is that I find some time to review the code and build a local test setup to try the backend. Then there will be feedback, and after one or two rounds of code review we can merge this. I'll also have a look at your docker-on-travis solution for the CI tests.

@rledisez Any feedback regarding test credentials we may use for integration testing?

jayme-github:
> the next step is that I find some time to review the code and build a local test setup to try the backend. Then there will be feedback, and after one or two rounds of code review we can merge this. I'll also have a look at your docker-on-travis solution for the CI tests.

Thanks for getting back, didn't mean to rush you!

fd0:
FYI: OVH provided a voucher so I can use their Swift service.

fd0 (changes requested):
So, I played with the backend a bit, looks already really good! There are several issues we need to talk about:
 * It uses the same path names as the s3 backend, e.g. `key` and `snapshot` instead of `keys` and `snapshots`. I'd like to standardize the names, e.g. in a section in the `Design.md`, so that the plural names are used. The s3 backend introduced the singular names, unfortunately I overlooked that and now we're stuck with it. I'll probably change the s3 backend to (by default) create the new names, but also accept the old names as a fallback. I'll also add code to the local backend shortly so that it can be used to read all formats (local, s3 with old names via fuse mount, s3 with new names via fuse mount). You can find a function which constructs the names [here](https://github.com/restic/restic/blob/master/src/restic/backend/local/local.go#L74-L102), I'll probably also move that from the local backend to the `backend` package.
 * Unfortunately changing the path names will make the existing repos created with this branch unusable until the paths are renamed manually. Can that be done  with the `swift` tool? Or maybe `rclone`?
 * For the integration tests, I'd like to try using the OVH service instead of running a local instance of openstack for now, let's see how that goes. Can you please remove the commits which add the docker instance to Travis?
 * Could you please cleanup the history and maybe merge all code into two commits: One adding the dependencies to `vendor/`, and one adding the real code? I'm not sure how to handle @jayme-github's contributions correctly, maybe add a note to the commit message?
 * I think the merge conflicts will go away when you remove the Swift integration tests, if not please let me know when you need help.
 * For the repository specification format for `--repo` I'd like to suggest using `swift:container-name:/path`. Let's see if that's enough for people to use. I'm planning to add free-form config parameters in the form `-o foo=bar`, so we may add parameters like `-o swift.hostname=foobar` later.

jayme-github:
> FYI: OVH provided a voucher so I can use their Swift service. Thanks @rledisez!

Cool, nice to hear!

> Unfortunately changing the path names will make the existing repos created with this branch unusable  [...]

Should be no problem to fix that using `swift` client or a small script. Swift supports "renaming" by server-side-copy + remove.

>  Can you please remove the commits which add the docker instance to Travis?

I think it would be nice (at least for running tests locally) to keep the swift stuff in `run_integration_tests.go`. What do you think?

> I'm not sure how to handle @jayme-github's contributions correctly, maybe add a note to the commit message?

No overhead here @byo. I'm okay with squashing my commits!

fd0:
> Should be no problem to fix that using swift client or a small script. Swift supports "renaming" by server-side-copy + remove.

Ok, that's great. It's just renaming `key` to `keys` and `snapshot` to `snapshots`.

> I think it would be nice (at least for running tests locally) to keep the swift stuff in run_integration_tests.go. What do you think?

I'd like to remove it if we're not going to use it with Travis. We can document running the integration tests with all the dependent services somewhere, this also goes for the Minio server.

byo:
@fd0 Thanks for review, I'll try to address those this week, unfortunately my 3-weeks-old daughter is in charge of all my free time now so it all depends on her ;)

I see extended options already in master - it looks like adding support for them would be trivial, let's see how it goes. Are those persisted when new repository is being created or it has to be specified on each restic invocation? Having some persistence here could be beneficial similarly to how one does not have to use repo URL each time.

@jayme-github I'll definitely acknowledge you in the commit message.

fd0:
Oh wow, congratulations @byo! Take all the time you need, and let us know when you don't have the time, so someone else (like me) can pick up. Thanks!

byo:
@fd0: I think I've addressed your comments, all seems to be working fine. In order to setup OVH Swift server, please set those environmental variables in travis config:
* ```RESTIC_TEST_SWIFT_AUTH_SERVER``` (should be ```https://auth.cloud.ovh.net/v2.0/```),
* ```RESTIC_TEST_SWIFT_AUTH_USERNAME```
* ```RESTIC_TEST_SWIFT_AUTH_PASSWORD```
* ```RESTIC_TEST_SWIFT_TENANT_NAME```
* ```RESTIC_TEST_SWIFT_REGION_NAME```
Those can easily be extracted from OpenStack rc file which you can download directly from OVH manager. Please let me know if you have any issues here.

I did not experiment yet with extended options, I'm not sure if I'll have enough time to take a look at it. I think it may even be better to add those in a separate PR.

jayme-github:
Hey @byo thanks for updating!
Is there any special reason that you choose using auth server and tenant stuff over storage url and token?

byo:
Hi @jayme-github,
Tokens will have expiration date, I don't think it would be easy to generate long-lived tokens without admin approval. That's why I think it's safer to generate token in each integration tests run.

Actually, tests themselves are using storage server and token, it's the ```run_integration_tests.go``` environment that is responsible to get those from somewhere.

jayme-github:
> Tokens will have expiration date

Oh, I see. Was not aware of that (my OpenStack/Swift knowledge is pretty limited). From the OVH management interface it seemed as if those tokens would not expire but that assumption is just based on the fact that they don't show any expiry date :smile:

jayme-github:
> Should be no problem to fix that using swift client or a small script. Swift supports "renaming" by server-side-copy + remove.

As swift client has no move wrapper I decided to stick with github.com/ncw/swift and hacked together this little tool one could use to migrate to plural path names: https://gist.github.com/jayme-github/8b7a3fd6e2b5a3a73d23bd8234623132

Probably not the smartest or fastes way to do this but it was good enough for my repos.

fd0:
Thanks for your work. I'll take over from here :)

fd0:
Hey, just a quick heads up: I'm currently working on #898, afterwards I'll look at this backend.

@byo can we use this branch for further development?

byo:
@fd0 great to hear. I believe this branch is in a pretty good state now, I'm using it for my private backups and see no issues so far.

fd0:
FYI: I'm currently working on this.

byo:
@fd0 thanks for looking at this, if you need any help, please let me know.

fd0:
Another quick heads up: I'm still working on this backend. First, I rewrote the test suite for backends (#961), then we've decided to unify the repository layout (#965, #966), so it unfortunately you'll have to do a bit of renaming before you'll be able to use the backend with a restic release eventually. Plus side: You can also access the repo via e.g. a fuse mount. So, please stay tuned!

byo:
Thanks @fd0 for the update. I guess this PR won't get into 0.6 release?

fd0:
No, this won't go into 0.6.0, but I'm aiming for the next release :)

fd0:
This PR is superseded by #975, I needed to make a PR in a branch in the restic organisation so I could test the live integration tests against OVH, this works now. I'm going to close this PR.

@byo @jayme-github could you please test the code in #975? You need to rename the files below `data/` to the same structure as in the default layout used for local/sftp. Maybe adapt the rename script?

fd0:
@byo @jayme-github ping? No time to test #975 yet? :)

jayme-github:
Sorry, I'm a little busy lately. Probably next weekend or next week :disappointed:

byo:
I'll try to start some tests today.

fd0:
Great :)

PR #657, opened 2016-11-03
Read files to backup from a file

AlexanderThaller:
I had the need to backup a lot of files that where pre filtered by another command. I tried using xargs but that only would lead to multiple executions of restic and thus multiple snapshots.

This pull request adds a new flag that will read the files to backup from a file instead of the command line.

Usage example:
restic -r /tmp/backups backup --files-file /tmp/input_files

### DIFF
--- a/doc/Manual.md
+++ b/doc/Manual.md
@@ -99,6 +99,7 @@ Usage:
 Flags:
   -e, --exclude pattern         exclude a pattern (can be specified multiple times)
       --exclude-file string     read exclude patterns from a file
+      --files-from string       read the files to backup from file (can be combined with file args)
   -f, --force                   force re-reading the target files/directories. Overrides the "parent" flag
   -x, --one-file-system         Exclude other file systems
       --parent string           use this parent snapshot (default: last snapshot in the repo that has the same target files/directories)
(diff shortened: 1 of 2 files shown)


fd0:
Thanks for your contribution! Unfortunately, I'm not convinced that we need this function. The command `xargs` can be used as follows to achieve what you're after. Suppose the files/dirs to be backed up are saved in `/tmp/list`, call `xargs` and `restic` as follows:
```
$ cat /tmp/list | xargs -d '\n' restic backup 
```

You can test that with the command `ruby -e 'puts ARGV.inspect'`, which prints out the arguments as an array. For example if you have a line in the `list` file which contains white space, that will be handled gracefully:
```
$ cat /tmp/list
1
2
3
4
foo bar baz
end

$ xargs -d '\n' ruby -e 'puts ARGV.inspect' < /tmp/list
["1", "2", "3", "4", "foo bar baz", "end"]
```

On the other hand I think that `--files-file` is not a good name for the command-line option.

fd0:
I've had a look at `tar` and `rsync`, both have an option called `--files-from` to read a list of files/dirs from a file.

AlexanderThaller:
I'm gonna try again maybe I just got confused by the interface. Thx

AlexanderThaller:
And yeah i wasn't sure what the naming should be so i just oriented myself by the excludes-file flag. What tar and rsync do looks more reasonable. 

I still would think that an option like this would me more convenient for users instead of the xargs approach.

fd0:
Turns out, for very long argument lists this is not sufficient and using xargs in this way will not work for all cases. I've thought about this and changed my mind: We should add an option like this.

Would you like to rework your PR, or should I have a look first?

AlexanderThaller:
I will rework it.

fd0:
Let me know if you're ready for a review

AlexanderThaller:
This should be fine the big if that differentiates between reading from file or not seems a bit ugly but not sure how to fix that.

fd0:
What about the following: Add a new function `func readFromFile(filename string) ([]string, error)`. If `filename` is the empty string, just return `nil, nil`. Then use this function to append the filenames read from the file to `args` like this (at the beginning of `runBackup()`):
```go
func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
   target, err := readFromFile(opts.FilesFrom)
   if err != nil {
      return err
   }

   target = append(target, args)

   if len(args) == 0 {
     return errors.Fatalf("wrong number of parameters")
   }
```

fd0:
You can also squash the commits into one then.

AlexanderThaller:
Ok I made the changes and squashed the commits. Seems good.

AlexanderThaller:
I also updated the documentation.

fd0:
Looks good, thanks! :+1:

fd0 (approved):

Issue #29, opened 2014-11-15
Implement Metadata Cache


scoddy:
Metadata should be cached locally for multiple operations.
- Feature to limit cache size

fd0:
This has already been started.

fd0:
Blob lists for snapshots are now cached in a cache directory. More metadata caching will be added once it's needed.

fd0:
The cache was removed, so this issue is open again.

r3dey3:
Not sure if I should create a new issue for this, but I'm using add-cache branch (I rebased d9d22d6 onto 0ee1650 but I'm guessing that's not the cause).
 If a metadata file cannot be retrieved (eg a snapshot) because of a permissions issue, restic panics but not before caching an invalid file. I was able to recreate the problem using a test repo and changing the ownership of a snapshot file to root after doing a couple of backups.

The error message from the initial failure:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x47dc12]

goroutine 1 [running]:
io.copyBuffer(0xdfd760, 0xc4222960a0, 0x0, 0x0, 0xc42227c000, 0x8000, 0x8000, 0xab0ae0, 0xc422328900, 0xdfd760)
	/usr/local/go/src/io/io.go:392 +0x102
io.Copy(0xdfd760, 0xc4222960a0, 0x0, 0x0, 0xc4222960a0, 0xe010a0, 0xc4222960a0)
	/usr/local/go/src/io/io.go:362 +0x68
github.com/restic/restic/internal/cache.(*Cache).Save(0xc420052ac0, 0xac8477, 0x8, 0xc4223289c0, 0x40, 0x0, 0x0, 0x0, 0x0)
	src/github.com/restic/restic/internal/cache/file.go:97 +0x181
github.com/restic/restic/internal/cache.(*Backend).Load(0xc4200e4620, 0xe04160, 0xc420016120, 0xac8477, 0x8, 0xc4223289c0, 0x40, 0x0, 0x0, 0x6e1aad, ...)
	src/github.com/restic/restic/internal/cache/backend.go:113 +0x50f
github.com/restic/restic/internal/backend.LoadAll(0xe04160, 0xc420016120, 0xe07400, 0xc4200e4620, 0xac8477, 0x8, 0xc4223289c0, 0x40, 0x0, 0x0, ...)
	src/github.com/restic/restic/internal/backend/utils.go:13 +0xda
[...]

PR #671, opened 2016-11-13
Implement a progress bar for restores

### LINKED ISSUE #426: Progress bar for restore
Backup have nice progress bar, but restore is not.
In real life I expect to do restore not that often (now I play with restic and with restore a bit too often).
But it will be useful to have progress for restore too (for big repo I would like to see some estimation about how much time it will take).

Novex:
Very similar to how the progress bar for taking snapshots works, this PR implements one for restores.

restic-restore-progress

Closes #426

### DIFF
--- a/src/cmds/restic/cmd_restore.go
+++ b/src/cmds/restic/cmd_restore.go
@@ -106,7 +106,7 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return nil
 	}
 
-	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) bool {
+	selectExcludeFilter := func(item string, node *restic.Node) bool {
 		matched, err := filter.List(opts.Exclude, item)
 		if err != nil {
 			Warnf("error for exclude pattern: %v", err)
@@ -115,7 +115,7 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return !matched
 	}
 
-	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) bool {
+	selectIncludeFilter := func(item string, node *restic.Node) bool {
 		matched, err := filter.List(opts.Include, item)
 		if err != nil {
 			Warnf("error for include pattern: %v", err)
--- a/src/cmds/restic/cmd_restore.go
+++ b/src/cmds/restic/cmd_restore.go
@@ -47,6 +47,10 @@ func init() {
 	flags.StringSliceVar(&restoreOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path` for snapshot ID \"latest\"")
 }
 
+func newRestoreProgress(gopts GlobalOptions, todo restic.Stat) *restic.Progress {
+	return newArchiveProgress(gopts, todo)
+}
+
 func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 	if len(args) != 1 {
 		return errors.Fatalf("no snapshot ID specified")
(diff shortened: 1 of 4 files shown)


fd0:
Hey, thanks for your contribution! I'll have an in-depth look sometime this evening (rather busy right now).

One thing I noticed while scrolling over the code: For high-latency remote repositories, it'll take some time to collect the information (in the scan phase) needed to display a progress (files, dirs, sizes), so this will slow down the whole operation. Did you think about that?

Novex:
All good :) I have to admit I haven't tested with a remote repository yet.

I kind of assumed the metadata was downloaded in one hit and available locally after the call to `res.repo.LoadTree()`, so it would be reasonably cheap to query after the fact.

If it isn't downloaded all in one hit, after it is downloaded is there any caching?
* If there is then the restore should take no more time because it will now be reading locally
* If not, would you consider it as an option to improve performance?

Also I did notice that the archiver implemented the progress bar for its scan process - if it is an action that takes a while I can implement the scanning progress bar for this area as well?

fd0:
The metadata information (stored in `tree` blobs in the repo) is fetched on demand and not yet cached. A cache is planned, but not yet implemented (see #29).

Novex:
Ah yep I see - in that case this probably isn't such a good idea without a cache because we are indeed doubling the amount of work taken to get the tree blobs. Which I imagine is significant for larger repositories.

You mentioned in #29 that the cache was implemented, but then removed - just wondering in what way it was implemented and how come it was removed?

As a caching strategy to mitigate this, I was thinking:
* Check/populate the blob cache in the `loadBlob` function in the `Repository` package
* Delete the cache on exit/when the command is finished
  * So it's only useful for avoiding this performance penalty, but we don't have to worry about expiry, etc
* Just naively cache all backends for now. Alternatively, we could have some way to determine whether a backend is remote (and should be cached) or local (and shouldn't), but I'm not sure of the most idiomatic way of doing this?

Keen to hear your thoughts.

fd0:
We had a cache implementation really early on before we added bundling of smaller blobs into pack files, so it was before the repository format was stable. That was easier, because a file in the local cache directly corresponds to a file in the repo. That's not the case any more, so we removed the cache before implementing blob packing, and didn't yet add it back.

Adding a cache in memory is IMHO not a good idea, repositories get quite large, and restic already has a high memory usage (unfortunately, there's a bug which causes this, and I didn't have the time to debug this further).

A cache should be independent of the backend repository, but maybe we can have a global flag to disable it completely. The cache must also be encrypted with the repository key, so that no meta data is leaked to the file system the user's home directory is located.

Novex:
Oops sorry I should have been more clear - I was just going to store the encrypted blob parts (essentially [this result](https://github.com/restic/restic/blob/master/src/restic/repository/repository.go#L113)) in a temporary area of the filesystem as files like `<hash>_<offset>_<length>` rather than in memory.

This way they're stored encrypted and we don't have to worry about memory bloat. Good point on having the flag to disable it.

fd0:
Hm, that'd be a special solution just for this use case. I'd rather like to solve this globally, so that other operations (e.g. `check` and `prune`) can benefit.

Novex:
Yeah that would be a better way of doing it, but I don't have a clear picture in my head of how to do so :)

I'm happy to take a look and try and work something out if you have any high level pointers?

Novex:
So I had a bit of a look into it, I'm not sure if this is solvable globally at this point in time.

The common point which would apply this globally to all commands is the `Load` function of the backends. However, there's no way in the `DataFile` blob type to differentiate between:
* directory metadata, which we want to cache to speed this and the other operations up

vs

* actual data, which we do not want to cache because it would double the amount of disk space needed for restores, it's generally only needed once, etc :(

The only real alternative I can think of is to manually apply the caching to the relevant areas which could benefit from them, Eg. The directory walking in the restorer and the relevant parts of `check`, `prune` and any other commands (which I'm not 100% sure of, but would be able to dig around and find). But that's pretty much back to square one.

Any other ideas?

fd0:
I'm sorry that I didn't respond earlier.

It seems to me that you are not aware of a layer in restic's design: Blobs are either data (parts of files, `DataBlob`) or JSON with metadata (`TreeBlob`). These blobs are bundled together into files, that are stored in the repo. At the file level, there's just a `DataFile`, which may contain both types of blobs.

I'm working on adding a local cache for the tree blobs, until that is merged I'm going to close this PR for now. It does not mean that this PR won't get merged, it just means that it won't be merged now. Thanks for your work!

Issue #200, opened 2015-06-26
Feature request: "Restore newer/missing datasets" / "synchronize"


cfcs:
It would be useful to have a feature that allows restoring files/datasets that are not currently in a "checked out" state locally while avoiding doing "double work" by downloading data that already resides in the local target directory.

This feature would enable asynchronous synchronization of files between multiple machines (not necessarily online at the same time), enabling users to have a full, shared, encrypted backup on an untrusted (always online) machine that the individual machines can sync from/to when they are online.

I envision an implementation that resembles the `restore` command that:
1) Takes a designated (local) target directory
2) loops over all snapshots - pseudo-code:

``` bash
for snapshot in `./restic list snapshots|egrep -o '[a-f0-9]{64}'`
  do ./restic ls $snapshot
done
```

3) Checks if any of the files in the snapshots are newer than what currently resides in the target directory, or if there are files present in the snapshot that do not exist in the local checkout.
4) Selectively does a `restic restore $snapshot` to extract the files

**NOTE:** One issue that could arise from the implementation of this feature is that it becomes impossible to delete a file as long as the snapshots still exists on the remote end.
A mechanism to "ignore" specific files/trees that you do not want restored / have been deleted would probably be desirable. This is already on the drawing board: https://github.com/restic/restic/issues/86

Thoughts? :-)

fd0:
I like this feature, it could replace `unison` in my current setup, but it's rather far away from the core function of restic (namely: saving and restoring backups accurately). We'll leave this issue open for later.

tf198:
I've got some proof of concept code for this for you to have a look at.  In my case I want to sync both a directory containing 100K+ images and a 2GB database file with minor changes each day.
My approach is:
* Move the current file out the way
* Run the chunker on it
* If all the chunk IDs match then just move it back and continue
* Start the restore but check the chunk index before using the repo.
* If it matches then copy the bytes from the current file rather than the repo

In my tests with a 2GB direcory across a network with about 3MB/s transfer the initial restore takes 13 minutes.  If I modify a few files and then restore again it comes down below 1 minute with the network saturated for the first 30 seconds while it pulls the metadata and then 30 seconds of CPU as it checks all the existing files.

https://github.com/restic/restic/compare/master...tf198:sync_restore
I haven't created a pull request yet as it is a bit rough and ready but I can't see it having too many adverse effects. Comments welcome...

MichaelEischer:
[...]

PR #672, opened 2016-11-14
Only restore blobs which don't exist locally

### LINKED ISSUE #407: Make restore resumeable.

Novex:
* If we can, calculate the hash of the part of the local file that the blob corresponds to
  * If it matches then we already have the data and can skip fetching it
  * This should significantly improve the speed of remote restores when a local dataset already exists

This should allow for resumable restores, which resolves #407 and possibly also #200 (if I understand it correctly)

### DIFF
--- a/src/restic/node.go
+++ b/src/restic/node.go
@@ -192,7 +192,7 @@ func (node Node) createDirAt(path string) error {
 }
 
 func (node Node) createFileAt(path string, repo Repository) error {
-	f, err := fs.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0600)
+	f, err := fs.OpenFile(path, os.O_CREATE|os.O_RDWR, 0600)
 	defer f.Close()
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


fd0:
Thanks again for your contribution! I think it's a good idea, but we'll need to handle a few corner cases. I'll add more comments in the evening (Europe here).

Novex:
No worries! Looking forward to them :)

Novex:
No pressure, just following up as to whether you had a chance to take a look at this?

fd0:
Thanks for asking: I was a bit busy this week and did not have the time yet, sorry.

Novex:
No worries at all - I know that feeling :)

fd0:
So, I have some feedback for you: I've thought a lot about how to handle this and I think that modifying files directly is a bad idea because it can easily go wrong. Suppose a user restores their home directory while a program is still running and both restic and the program modify the same file. The result will almost always be a corrupt file, which is a bad idea.

I had a look what other programs do and I like the approach implemented by rsync: By default, if a destination file already exists, it creates a new temp file with a random name right next to it, starts reading the already present file and copies any valid content from the old to the temp file (for restic this means running a chunker on the old file and copying any blobs that should be present in the new file). All blobs not present in the old file are fetched from the repo and written to the new file. Once the new file is complete, rename the new (temp) file to the old one so that it is atomically replaced.

What do you think? Are you interested in implementing this approach? I'd be glad to assist!

I'm going to close this PR since I won't merge the code in this form.

Novex:
Fantastic point! I'm happy to have a crack at implementing it.

Just so I have this clear:
* The current default behaviour of restic restore when a file exists is to overwrite each chunk of the file currently being restored in-place
  * Which leads to many of the risks you mentioned, eg. another program modifying the file/inconsistent results should restic fail
* We want to change it so the default behaviour is to instead restore the the file currently being restored beside the existing one, then atomically replace the existing one on success
  * A safer option, but has the downside of temporarily requiring twice as much disk space as the size of the currently restored file
* In addition, we want to add a new flag `--inplace` to restore the original default behaviour with the optimisation of only overwriting chunks which have changed

Does that sound about right?

fd0:
That's mostly right, just a small correction to the first point: the current behavior is to create a new file which replaces any old file that may exist via the flag `O_CREATE` to the `os.OpenFile` call here: https://github.com/restic/restic/blob/master/src/restic/node.go#L195 The flag tells the kernel to either create the file (file does not exist yet) or truncate it (file already exists).

fd0:
Let's create a new issue and transfer the discussion and todo list there.

PR #683, opened 2016-12-03
Omit "archived as %v" messages in quiet mode.

jannic:
The message "archived as %v" is shown for backups from stdin, even in quiet mode. This leads to unwanted mails if run from cron. As quiet mode doesn't output anything for normal backups, I think the same should be done for backups from stdin.


fd0 (approved):
Great catch, thanks!

PR #690, opened 2016-12-10
Even if file changes size during backup, still save it

### LINKED ISSUE #689: Files not backed up at all
Beside #604, where the backup run would end up with zeroed files and slightly corrupted repo, there's another similar bug, where files would not end up in the repo at all. The case is where some file is open and appended to during the backup run (think logs and similar). Currently restic just skips those files. What it should do instead is warn the operator and save what it can, i.e. do its best and keep going, not just skip such files.

I'll send a patch to fix this.

zcalusic:
Previously such files (typically log files) wouldn't be backed up at all!

The proper behaviour is to backup what we can, and warn the operator that file is possibly not complete. But it is a warning, not an error.

Closes #689

### DIFF
--- a/src/restic/archiver/archiver.go
+++ b/src/restic/archiver/archiver.go
@@ -197,7 +197,7 @@ func updateNodeContent(node *restic.Node, results []saveResult) error {
 	}
 
 	if bytes != node.Size {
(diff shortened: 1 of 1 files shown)


fd0 (approved):
LGTM!

PR #694, opened 2016-12-12
Backblaze B2 backend support

### LINKED ISSUE #512: feature request: add backblaze b2 storage backend
See https://www.backblaze.com/b2/docs and the go client/bindings at https://github.com/kothar/go-backblaze

arithmetric:
This PR implements support for Backblaze B2 as a storage backend using kurin's [blazer](https://github.com/kurin/blazer) package. The code is based on the S3 backend.

Like the S3 backend, the B2 account ID and key credentials should be specified by environment variables (`B2_ACCOUNT_ID` and `B2_ACCOUNT_KEY`), while the B2 bucket name and path prefix are specified through the repository URL in the format: `b2:<bucket name>` or `b2:<bucket name>:<prefix path>`.

If you register with Backblaze B2, you should be able to run the full test suite against the B2 API at least once daily and stay within the free limits. By raising the caps to a modest amount (less than $1 total), you would be able to run the full test suite multiple times.

Resolves #512.

### DIFF
--- a/src/restic/backend/b2/b2.go
+++ b/src/restic/backend/b2/b2.go
@@ -1,0 +1,288 @@
+package b2
+
+import (
+	"io"
+	"path"
+	"restic"
+	"strings"
+
+	"restic/debug"
+	"restic/errors"
+
+	blazerb2 "github.com/kurin/blazer/b2"
+
+	"golang.org/x/net/context"
+)
+
+// b2 is a backend which stores the data on a B2 endpoint.
+type b2 struct {
+	client     *blazerb2.Client
+	bucket     *blazerb2.Bucket
+	bucketName string
+	prefix     string
+	context    context.Context
+}
+
+// Open opens the B2 backend. The bucket is created if it does not exist yet.
+func Open(cfg Config) (restic.Backend, error) {
+	debug.Log("open, config %#v", cfg)
+
+	ctx := context.Background()
+
+	client, err := blazerb2.NewClient(ctx, cfg.AccountID, cfg.Key)
+	if err != nil {
+		return nil, errors.Wrap(err, "blazerb2.NewClient")
+	}
+
+	bucket, err := client.NewBucket(ctx, cfg.Bucket, blazerb2.Private)
+	if err != nil {
+		return nil, errors.Wrap(err, "blazerb2.NewBucket")
+	}
+
+	be := &b2{client: client, bucket: bucket, bucketName: cfg.Bucket, prefix: cfg.Prefix, context: ctx}
+
+	return be, nil
+}
+
+// Resolve the backend path for based on file type ane name.
+func (be *b2) b2path(t restic.FileType, name string) string {
+	if t == "key" && name == "config" {
+		t = restic.ConfigFile
+	}
+	if string(t) == restic.ConfigFile || name == "" {
+		return path.Join(be.prefix, string(t))
+	}
+	return path.Join(be.prefix, string(t), name)
+}
+
+// Location returns this backend's location (the bucket name).
+func (be *b2) Location() string {
+	return be.bucketName
+}
+
+// Load returns the data stored in the backend for h at the given offset
+// and saves it in p. Load has the same semantics as io.ReaderAt.
+func (be b2) Load(h restic.Handle, p []byte, off int64) (n int, err error) {
+	debug.Log("Load: %v, offset %v, len %v", h, off, len(p))
+	objName := be.b2path(h.Type, h.Name)
+
+	obj := be.bucket.Object(objName)
+	if err != nil {
+		debug.Log("  err %v", err)
+		return 0, errors.Wrap(err, "bucket.Object")
+	}
+
+	info, err := obj.Attrs(be.context)
+	if err != nil {
+		return 0, errors.Wrap(err, "obj.Attrs")
+	}
+
+	// handle negative offsets
+	if off < 0 {
+		// if the negative offset is larger than the object itself, read from
+		// the beginning.
+		if -off > info.Size {
+			off = 0
+		} else {
+			// otherwise compute the offset from the end of the file.
+			off = info.Size + off
+		}
+	}
+
+	// return an error if the offset is beyond the end of the file
+	if off > info.Size {
+		return 0, errors.Wrap(io.EOF, "")
+	}
+
+	var nextError error
+
+	// manually create an io.ErrUnexpectedEOF
+	if off+int64(len(p)) > info.Size {
+		newlen := info.Size - off
+		p = p[:newlen]
+
+		nextError = io.ErrUnexpectedEOF
+
+		debug.Log("  capped buffer to %v byte", len(p))
+	}
+
+	r := obj.NewRangeReader(be.context, off, int64(len(p)))
+	defer r.Close()
+
+	bufsize := int64(len(p))
+	var pos int64 = 0
+
+	// loop to read chunks until the buffer is full or EOF is reached
+	for {
+		n, err = r.Read(p[pos:])
+		pos += int64(n)
+		if pos == info.Size-off || pos == bufsize || err != nil {
+			if errors.Cause(err) == io.EOF {
+				err = nil
+			}
+			break
+		}
+	}
+
+	if err == nil {
+		err = nextError
+	}
+
+	return int(pos), err
+}
+
+// Save stores data in the backend at the handle.
+func (be b2) Save(h restic.Handle, p []byte) (err error) {
+	debug.Log("Save: %v with len %d", h, len(p))
+	if err := h.Valid(); err != nil {
+		return err
+	}
+
+	objName := be.b2path(h.Type, h.Name)
+
+	obj := be.bucket.Object(objName)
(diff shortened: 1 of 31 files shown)


src/restic/backend/b2/b2.go
```diff
@@ -0,0 +1,288 @@
...
+	return restic.FileInfo{Size: info.Size}, nil
+}
+
+// Test returns true if a blob of the given type and name exists in the backend.
+func (be *b2) Test(t restic.FileType, name string) (bool, error) {
+	found := false
+	objName := be.b2path(t, name)
+	obj := be.bucket.Object(objName)
+	_, err := obj.Attrs(be.context)
+	if err == nil {
```
tscs37: IIRC This might result in a problem if the file is currently being uploaded.

This should additionally check if `attrs.Status == b2.Uploaded`, otherwise it might return true for files that haven't been fully uploaded yet.

src/restic/backend/b2/b2.go
```diff
@@ -0,0 +1,288 @@
...
+		cur := &blazerb2.Cursor{Name: prefix}
+
+		for {
+			objs, c, err := be.bucket.ListCurrentObjects(be.context, 1000, cur)
+			if err != nil && err != io.EOF {
+				return
+			}
+			for _, obj := range objs {
+				info, err := obj.Attrs(be.context)
+				if err != nil {
```
tscs37: Similar to above, it should check `attrs.Status == b2.Uploaded`, otherwise treat the file as not-existing.

src/restic/backend/b2/b2.go
```diff
@@ -0,0 +1,288 @@
...
+	objName := be.b2path(h.Type, h.Name)
+
+	obj := be.bucket.Object(objName)
+	if err != nil {
+		debug.Log("  err %v", err)
+		return 0, errors.Wrap(err, "bucket.Object")
+	}
+
+	info, err := obj.Attrs(be.context)
+	if err != nil {
```
tscs37: This needs to check for `attrs.Status == b2.Uploaded`, otherwise the read might fail.

tscs37 (commented):
I've added some comments regarding some checks concerning the existence of files.

Backblaze allows you to access objects that have not been uploaded yet, ie a multipart upload, the file could also be in an unknown state or be a hider.

The B2 part looks solid otherwise.

src/restic/backend/b2/b2.go
arithmetric: I've made this change.

src/restic/backend/b2/b2.go
arithmetric: I've made this change.

src/restic/backend/b2/b2.go
arithmetric: Instead of making this change, I've removed the call to `obj.Attrs()` in `List()`, which is no longer necessary for getting the object's name due to your change to blazer in: https://github.com/kurin/blazer/pull/3

This change benefits performance and API usage costs by not making an attributes request for every object returned in a list operation. Additionally, I think it's appropriate for the list method to return objects which exist but are still being uploaded.

arithmetric:
Thanks for the review @tscs37! I've responded to your comments and made most of the suggested changes. I also pulled in your update to blazer.

src/restic/backend/b2/b2_test.go
```diff
@@ -0,0 +1,60 @@
...
+func init() {
+	if os.Getenv("B2_ACCOUNT_ID") == "" || os.Getenv("B2_ACCOUNT_KEY") == "" {
+		SkipMessage = "No B2 credentials found. Skipping B2 backend tests."
+		return
+	}
+
+	cfg := b2.Config{
+		AccountID: os.Getenv("B2_ACCOUNT_ID"),
+		Key:       os.Getenv("B2_ACCOUNT_KEY"),
+		Bucket:    "restic-test",
```
kurin: Bucket names need to be globally unique, and there's already a "restic-test" out there.  You should probably append a random string to this.

mholt:
Ooo, this is timely. I'm looking into restic for my backups on B2 since I'm having lock issues with duplicity. Hope this merges soon!

tscs37:
There is probably a bug somewhere in the code for extracting the prefix for the repository.

I compiled restic with the changes of the PR and when using a bucket like `b2:test-bucket:example`
it seems to try and use it for the bucket name, returning 

```
unable to open repo at b2:test-bucket:test: blazerb2.NewBucket: b2_create_bucket: 400: Invalid characters in bucketName: must be alphanumeric or '-'
```

tscs37:
There also seems to be a problem indicating the upload progress, as uploading a larger repository seems to halt in progress for a while.

fd0:
Hey, thanks for the work so far! Does anyone of you have a contact at backblaze? Maybe we can ask them if they'd be willing to donate an account that can be used for integration testing?

mholt:
Backblaze B2 is free up to 10 GB.

src/restic/backend/b2/b2_test.go
arithmetric: Thanks for the review @kurin! I've updated the test code to append a random string as you suggested.

arithmetric:
@tscs37 The repository location format that I implemented for B2 is `b2:[bucket name]/[prefix]`. So try `b2:test-bucket/example` instead of `b2:test-bucket:example`.

arithmetric:
@fd0 Thanks for the feedback. I'll submit a ticket to Backblaze support to inquire about a free or sponsored account.

@mholt Thanks for your interest in this feature. The 10 GB free storage is more than enough for the tests, however one run of the e2e tests consumes 50% of the free daily download allocation (~500 MB) and 20% of the free requests (~500 class B and C requests).

mholt:
> however one run of the e2e tests consumes 50% of the free daily download allocation (~500 MB) and 20% of the free requests (~500 class B and C requests).

Holy smokes, that's a lot. Didn't realize how expensive those tests were. I'm sure this is obvious, but maybe in case some support rep doesn't understand the implications, make sure to tell them that if they can help raise the limits for a test account, that more people will be purchasing B2 service -- myself included.

fd0:
We can certainly lower the amount of requests/bytes for the CI tests for a Backblaze backend...

@arithmetric Thanks for contacting Backblaze.

tscs37:
I ran it again using `b2:bucket/prefix` and it works perfectly now.

As a side note; I've found that the directory structure on backblaze does not reflect the on-disk format, making both compatible as much as possible would be a nice additional feature. That way existing backups can be uploaded and reused.

mholt:
What's left on this? Just a test account? Is there any way I can help?

kurin:
fwiw I made an API-breaking change to blazer to implement bucket attributes, but it's a small one and this patch vendors blazer anyway.

fd0 (changes requested):
I've just merged a PR which changes the backend API slightly (#741), could you please have a look and update this PR? Then I'll review the code and we can work towards getting this PR merged.

mholt:
Ping. :) Is there anything I can do to help get this merged?

fd0:
There is at least one merge conflict, and the backend API has changed since this code was submitted. Since there wasn't any activity for a few weeks, I'm tempted to close this PR for now. @mholt if you're interested in giving this a try and reworking the code, I'd be glad to assist!

mholt:
For reworking the whole PR I might need more time -- right now I have to get a new website and backend finished, but if this still needs help after that's done, I might be able to help, yeah. It just might be a month or two...

fd0:
Ok, then I'll close this PR for now, please let me know when/if you (or somebody else) have time and motivation to work on this. I promise I'll help :)

kurin:
I could poke at it, but it felt like this PR spent a long time in a state where it was basically done, but we didn't have a good story about how to run the integration tests.  I don't want to complete a PR against the new backend only to have it sit for months in the same place.

What do you need from backends in terms of tests, so that this isn't as big an obstacle next time?

fd0:
Ideally, I'd like to run the backend tests for each commit/PR, just as the other tests. For now we're running an sftp-server (for the sftp backend) and a minio s3 server (for the s3 backend) locally on Travis, and use that for tests. The upside is that we don't depend on any third-party services.

I think it's okay to use a third-party service for running the CI tests (I suspect that there will be more backends like backblaze in the future which we cannot just run locally). I understood that there is a test account with some free space for backblaze that may be suitable to run the test suite against for each commit/PR.

What are the constraints for using the free backblaze account? We can certainly reduce the amount of data for each test, if that is still a concern. @arithmetric you said that you contacted backblaze, did they respond?

ibib:
[Source](https://www.backblaze.com/b2/cloud-storage-pricing.html)
Free Tier: 
Storage: 10GB   
Upload: Free   
Download: 1GB/Day
Transactions: 2500 free Downloads/day 2500 other Transactions/day

fd0:
Hm, I think that the free tier may work, if we reduce the amount of data a bit.

fd0:
Let's continue the discussion in #512. I've submitted a new request with Backblaze.

Issue #654, opened 2016-11-01
Pathnames depend on backend


jayme-github:
I was trying to move a file based repository to S3 by copying the files but I'm unable to access it ("unable to open repo: wrong password or no key found")

Looking closer reveals that the file backend uses plural path names ("snapshots", "keys", "locks" probably from `src/restic/backend/paths.go`) while (at least) S3 ans Swift backend uses the singular `FileType` (`src/restic/file.go`) directly ("snapshot", "key", "lock").

It would be cool to have the same path names in every backend so we can move the repos around.
`restic check --read-data` finishes successfully when I rename the directories to their singular names and remove the sub directory structure below `data`:

### Original
```
./data
./data/e3
./data/e3/e37d9fc8074d7032010a5876df3aaa4008c404227fea426ee4f46b02762ff396
./snapshots
./snapshots/0a8fd93c7b58c254d255e1b796b2b4a005f2e6b2a4d954cad0c465e56733a1ae
./index
./index/975c562709214460143b85fdf3f335761dcb08a8ea9944416a22e86ce20140ac
./locks
./keys
./keys/e8a7991a36bc46d51a5a59a3fd218396dc679f5406ddedc9ae535fa83b6792c0
./tmp
./config
```

### Modified
```
./data
./data/e37d9fc8074d7032010a5876df3aaa4008c404227fea426ee4f46b02762ff396
./index
./index/975c562709214460143b85fdf3f335761dcb08a8ea9944416a22e86ce20140ac
./config
./snapshot
./snapshot/0a8fd93c7b58c254d255e1b796b2b4a005f2e6b2a4d954cad0c465e56733a1ae
./key
./key/e8a7991a36bc46d51a5a59a3fd218396dc679f5406ddedc9ae535fa83b6792c0
```

## Output of `restic version`
restic 0.3.0 (v0.3.0-16-g011aee1)
compiled at 2016-10-31 08:38:12 with go1.6.2 on linux/amd64

fd0:
Thanks for bringing this up. When we first added the s3 backend, this just slipped in. Probably we'll always have slightly backend-specific filenames. I've thought about adding an option to synchronize (copy) data within backends, #323 mentions this. What do you think about that?

zcalusic:
I agree that synchronizing backend file structure and keeping it in sync would be great. 'Cause what I saw by playing with minio S3 backend, behind all the bells and whistles is just a bunch of files.

[...]

PR #695, opened 2016-12-12
Align path names generated for file and S3 backend

jayme-github:
This aligns the path names generated for S3 backend to the ones used by
the file backend allowing S3 objects to be used as file backend and
vice versa.

Dirname and Filename generation logic have moved from file backend to
tha backend package.

Added a environment variable (AWS_LEGACY_PATHS) to S3 backend which
cat be set to true to switch to legacy pathnames (to be used with
existing repositories).

See #654 for a discussion about this.
I decided to step in with a very simple proposal on how to solve this. Especially because there are some pull requests for new backend that will use the same path names as S3.
This is probably not the best (or even wanted) solution as it requires S3 users to set a additional environment variable to be able to use their repository but I think it's the easiest one and I wanted to push the discussion forward.

We could provide an additional tool for "one-shot" migrations from "legacy S3" to "new style S3" backend paths so that restic does not have to maintain two different path styles forever.

(Please be lenient with me as I'm in no way familiar with go :smiley: )

### DIFF
--- a/src/cmds/restic/global.go
+++ b/src/cmds/restic/global.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"restic"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -326,6 +327,7 @@ func open(s string) (restic.Backend, error) {
 		if cfg.Secret == "" {
 			cfg.Secret = os.Getenv("AWS_SECRET_ACCESS_KEY")
 		}
+		cfg.LegacyPaths, _ = strconv.ParseBool(os.Getenv("AWS_LEGACY_PATHS"))
 
 		debug.Log("opening s3 repository at %#v", cfg)
 		be, err = s3.Open(cfg)
--- a/src/restic/backend/local/local.go
+++ b/src/restic/backend/local/local.go
@@ -71,36 +71,6 @@ func (b *Local) Location() string {
 	return b.p
 }
 
-// Construct path for given Type and name.
-func filename(base string, t restic.FileType, name string) string {
-	if t == restic.ConfigFile {
-		return filepath.Join(base, "config")
-	}
-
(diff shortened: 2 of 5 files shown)


fd0:
Thanks for the suggestion. I think this is good for new repositories on s3, but since it is a breaking change: How do we handle migration of existing repos?

jayme-github:
> How do we handle migration of existing repos?

Easiest thing I can think of would be an extra tool that does the migration on S3. I'm happy to provide one (despite being slow) but wanted to make sure you would agree on the idea first.

I'm with you regarding your second point but I think it would be way more work to provide a generic "backend-to-backed" migration system that to just keep the structure aligned where it is possible and of no harm.

fd0:
I have an idea: What about recording information about the backend in the repo config? E.g., have a field for subdir depth, the name of the backend used to create the repo. It would allow detecting the directory name/structure used and we can implement a `migrate` command that does the right thing[tm]. Let me think about this for a bit.

jayme-github:
Souds good for me. Would also allow the backeds to do further migrations if needed in future.

Would it be okay for you if I submit the non S3 specific parts of this as a separate PR so the file backend structure can be easily used by new backends like !694?

fd0:
> Would it be okay for you if I submit the non S3 specific parts of this as a separate PR so the file backend structure can be easily used by new backends like #648 and #694?

Yes, I think that's a good idea. I propose the following steps:
 * Extract the functions to construct the path name for a file into generic functions, and pass the number of sub-dirs as a parameter. In the first step, this can be hardcoded per backend (local/sftp -> 1, s3 -> 0). Let's get that PR merged.
 * Then we'll add infrastructure and a config option to set this parameter, so we can easily override it, so people can use it already.
 * The last step is adding a field to the repo config.

What do you think?

PR #703, opened 2016-12-18
Avoid duplicate backup paths

sjoerdsimons:
Target directories from the from-files argument get added to the command
line args, after which all command line args were appended to the same
variable again causing duplicates. Split the used variables to avoid
this.

Signed-off-by: Sjoerd Simons <sjoerd@luon.net>


fd0:
Good catch, thanks for your contribution!

fd0 (approved):

PR #709, opened 2016-12-28
Make sure cleanup is executed before exiting

### LINKED ISSUE #708: Lock doesn't appear to be released after failed restore
# Output of `restic version`
restic 0.3.2 (v0.3.2-4-gb0997d0)
compiled with go1.7.4 on darwin/amd64

## Expected behavior
An exclusive lock should not be held after an operation is complete and the process terminates.

## Actual behavior

```
$ restic restore latest --path "path/that's/not/found" --target ~/restoretest/
latest snapshot for criteria not found: no snapshot found Paths:[path/that's/not/found] Host:

$ restic check
repository is already locked by PID 57750 on Shadowfax.home by matt (UID 501, GID 20)
lock was created at 2016-12-27 08:33:22 (50.993236973s ago)
storage ID 082d53a5
the `unlock` command can be used to remove stale locks
```

As far as I know, just try to do a restore on a path that's not found, and then run `restic check` and it seems the lock didn't release after the failed restore. The process with ID 57750 had terminated nearly a minute earlier.

The `unlock` command did the trick but it seems a bit unnecessary? Should I just run `restic unlock` after every restic command or is that a bad idea. (Going for unsupervised execution here.)

fd0:
Closes #708

### DIFF
--- a/src/cmds/restic/cleanup.go
+++ b/src/cmds/restic/cleanup.go
@@ -62,8 +62,13 @@ func CleanupHandler(c <-chan os.Signal) {
 	for s := range c {
 		debug.Log("signal %v received, cleaning up", s)
 		fmt.Printf("%sInterrupt received, cleaning up\n", ClearLine())
-		RunCleanupHandlers()
-		fmt.Println("exiting")
-		os.Exit(0)
+		Exit(0)
 	}
+}
+
+// Exit runs the cleanup handlers and then terminates the process with the
+// given exit code.
+func Exit(code int) {
+	RunCleanupHandlers()
+	os.Exit(code)
 }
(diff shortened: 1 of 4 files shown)


src/cmds/restic/cleanup.go
```diff
@@ -62,8 +62,13 @@ func CleanupHandler(c <-chan os.Signal) {
...
-		os.Exit(0)
+		Exit(0)
 	}
 }
+
+// Exit runs the cleanup handlers and then terminates the process with the
+// given exit code.
+func Exit(code int) {
+	RunCleanupHandlers()
+	os.Exit(code)
```
mholt: I think this is a good way to do it. :+1:

fd0: Thanks :)

PR #714, opened 2016-12-31
Added support for a configurable directory prefix for blob names.

ckemper67:
Extracted a Filename and Dirname function and used it in the s3 and
sftp backends. Added support for setting the length of the directory
prefix in the s3 configuration. Introduced a NewConfig method for
creating the s3 configuration.

### DIFF
--- a/src/restic/backend/paths.go
+++ b/src/restic/backend/paths.go
@@ -1,6 +1,10 @@
 package backend
 
-import "os"
+import (
+	"os"
+	"path"
+	"restic"
+)
 
 // Paths contains the default paths for file-based backends (e.g. local).
 var Paths = struct {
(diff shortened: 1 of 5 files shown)


ckemper67:
I saw your comment on #695. I had already written similar code for my GCS backend, so I followed your suggestion and extracted a method for the filename that can be reused.

So far I implement a simple integer for the length of the directory name (with sftp using a prefixLength of 2). I know you suggested to allow multiple directories for nesting and I can take a look at that tomorrow.

I am not really happy that I cannot easily reuse the methods in the local backend since that needs to use filepath.Join instead of path.Join.  Any idea how to handle that nicely? The ideas I had so far were to either always use forward '/' internally and substitute when accessing the filesystem or to handle file path as arrays of segment names in the Filename and Dirname functions and use Join in the respective backends.

fd0:
Hey, thanks for giving this a try. I'm wondering though: Do we need to make the prefix length configurable? I think having a variable number of subdirs is sufficient...

I also have thought about how to make it more manageable in the code, I'll outline my idea below:
 * In the `backend` package, create a new struct `Localizer` (name to be discussed, I didn't come up with anything better):
```go
type Localizer struct {
    subdirs int
    subdirLength int
    dirname func(string) string
    join func(string) string
}
```
 * Then implement the needed functions as methods on this struct, so we can use it in the backend code, without referencing the `backend` package and supplying lots of parameters all over the place.

What do you think?

fd0:
Since I haven't heard anything I'm going to close the PR for now. Please let me know when you want to resume working on it. Thanks a lot for your work!

Issue #710, opened 2016-12-28
Panic after 2 days into backup


mholt:
## Output of `restic version`
restic 0.3.2 (v0.3.2-4-gb0997d0)
compiled with go1.7.4 on linux/arm

## Expected behavior
No panic.

## Actual behavior

```
panic(0x44ca90, 0x274353d0)
        /usr/local/go/src/runtime/panic.go:500 +0x33c
restic/repository.(*Index).Store(0x109e05a0, 0x1, 0x1988f0, 0xaca72265, 0x36646422, 0xbbeb2592, 0x8818bdc4, 0x20aea0a4, 0xcb7eb498, 0x47dc953f, ...)
        /var/folders/86/75b471wx6sn8bjtd4jz48kbm0000gn/T/restic-build-922213292/src/restic/repository/index.go:104 +0xc4
restic/repository.(*Repository).savePacker(0x1098e990, 0x1ad69680, 0x0, 0x0)
        /var/folders/86/75b471wx6sn8bjtd4jz48kbm0000gn/T/restic-build-922213292/src/restic/repository/packer_manager.go:144 +0xab4
restic/repository.(*Repository).SaveAndEncrypt(0x1098e990, 0x33ed3d01, 0x22bca000, 0x3d3f63, 0x5dc000, 0x33ed3e20, 0x0, 0x0, 0x0, 0x0, ...)
        /var/folders/86/75b471wx6sn8bjtd4jz48kbm0000gn/T/restic-build-922213292/src/restic/repository/repository.go:209 +0x4f4
restic/repository.(*Repository).SaveBlob(0x1098e990, 0xcc454801, 0x22bca000, 0x3d3f63, 0x5dc000, 0xcc454851, 0x748d6d56, 0xaa8da3c, 0xb9d7ea2c, 0xb09b1972, ...)
        /var/folders/86/75b471wx6sn8bjtd4jz48kbm0000gn/T/restic-build-922213292/src/restic/repository/repository.go:561 +0x114
restic/archiver.(*Archiver).Save(0x10b301b0, 0x3d3f01, 0x22bca000, 0x3d3f63, 0x5dc000, 0xcc454851, 0x748d6d56, 0xaa8da3c, 0xb9d7ea2c, 0xb09b1972, ...)
        /var/folders/86/75b471wx6sn8bjtd4jz48kbm0000gn/T/restic-build-922213292/src/restic/archiver/archiver.go:103 +0x270
[...]

PR #717, opened 2017-01-02
Only add entries to indexes inside PackerManager

### LINKED ISSUE #367: Restic panic("store new item in finalized index")s during backup.
Restic crashes with a panic() during backups after a few gigabyte.
- OS: FreeBSD
- FS: ZFS
- Version: restic 0.1.0 (v0.1.0-168-g55d9c5f) compiled at 2015-12-07 00:21:27 with go1.5.1

[restic.txt](https://github.com/restic/restic/files/53925/restic.txt)

fd0:
This was a nasty bug. Users reported (#367, #710) that restic aborts with panic:

    panic: store new item in finalized index

The code calling panic() is in the Store() method of an index and guards
the failure case that an index is to be modified while it has already
been saved in the repo.

What happens here (at least that's what I suspect): PackerManager calls
Current() on a MasterIndex, which yields one index A. Concurrently,
another goroutine calls Repository.SaveFullIndex(), which in turn calls
MasterIndex.FullIndexes(), which (among others) yields the index A. Then
all indexes are marked as final. Then the other goroutine is executed
which adds an entry to the index A, which is now marked as final. Then
the panic occurs.

The commit solves this by removing MasterIndex.Current() and adding a
Store() method that stores the entry in one non-finalized index. This
method uses the same RWMutex as the other methods (e.g. FullIndexes()),
thereby ensuring that the full indexes can only be processed before or
after Store() is called.

### DIFF
--- a/src/restic/repository/master_index.go
+++ b/src/restic/repository/master_index.go
@@ -119,16 +119,14 @@ func (mi *MasterIndex) Remove(index *Index) {
 	}
 }
 
-// Current returns an index that is not yet finalized, so that new entries can
-// still be added. If all indexes are finalized, a new index is created and
-// returned.
-func (mi *MasterIndex) Current() *Index {
+// Store remembers the id and pack in the index.
+func (mi *MasterIndex) Store(pb restic.PackedBlob) {
 	mi.idxMutex.RLock()
-
 	for _, idx := range mi.idx {
 		if !idx.Final() {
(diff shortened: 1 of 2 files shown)


mholt:
I'm running a full backup now with restic built from this branch, will let you know how it goes. Might take a few days.

fd0:
I'm going to merge this now, but we'll leave issue #367 open until you've reported back.

PR #718, opened 2017-01-02
CLI options now override env vars

mholt:
In testing restic, I had an env variable RESTIC_REPOSITORY that was my "main" or "real" backup location, but then I ran `restic --repo /other/repo/for/testing init` and it still tried to use the value in RESTIC_REPOSITORY. This change re-prioritizes CLI flags over env vars, which in my opinion, is more intuitive.

I hope it's on the right track, I noticed that the `parseEnvironment()` function didn't use any of its parameters and I didn't see why it needed to happen outside of the `init()`.

Let me know if you need anything in the way of testing this.

### DIFF
--- a/src/cmds/restic/global.go
+++ b/src/cmds/restic/global.go
@@ -25,18 +25,6 @@ import (
 
 var version = "compiled manually"
 
-func parseEnvironment() {
-	repo := os.Getenv("RESTIC_REPOSITORY")
-	if repo != "" {
-		globalOptions.Repo = repo
-	}
-
-	pw := os.Getenv("RESTIC_PASSWORD")
-	if pw != "" {
-		globalOptions.password = pw
-	}
-}
-
 // GlobalOptions hold all global options for restic.
 type GlobalOptions struct {
 	Repo         string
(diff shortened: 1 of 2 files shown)


fd0:
Oh, that's really embarrassing. Thanks for raising this issue. The order should obviously be command-line params before environment.

This change however breaks restic: The environment is never considered. The struct `globalOptions` is filled by `parseEnvironment()`, but the flag definitions seem to reset the values to the default values specified in the call to `StringVarP()`. With `parseEnvironment()` in the `PersistentPreRunE` hook, it is run after `init()`, but before any command code is run.

So, for a proper fix I suggest removing `RESTIC_REPOSITORY` from `parseEnvironment()` and changing the line defining the `--repo` flag as follows:
```go
	f.StringVarP(&globalOptions.PasswordFile, "password-file", "p", os.Getenv("RESTIC_REPOSITORY"), "read the repository password from a file")
```

That should fix it. Would you like to the change?

mholt:
Oops, I thought I took care of that too, but after I re-exported the env variable I didn't hit [return] to run the restic command and check; so you're right. Now I'm embarrassed :smile:

Pushed another commit.

fd0 (approved):
Cool, thanks!

PR #723, opened 2017-01-12
Added latest keyword in ls command.

ulziibuyan:
The restore command has the special "latest" snapshot keyword. Why not have same for listing files and directories? :-)

### DIFF
--- a/src/cmds/restic/cmd_ls.go
+++ b/src/cmds/restic/cmd_ls.go
@@ -17,6 +17,8 @@ var cmdLs = &cobra.Command{
 	Short: "list files in a snapshot",
 	Long: `
 The "ls" command allows listing files and directories in a snapshot.
+
+The special snapshot-ID "latest" can be used to list files and directories of the latest snapshot in the repository.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return runLs(globalOptions, args)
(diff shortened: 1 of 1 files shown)


src/cmds/restic/cmd_ls.go
```diff
@@ -86,9 +88,20 @@ func runLs(gopts GlobalOptions, args []string) error {
...
 
-	id, err := restic.FindSnapshot(repo, args[0])
-	if err != nil {
-		return err
+	snapshotIDString := args[0]
+	var id restic.ID
+	var paths []string
+
+	if snapshotIDString == "latest" {
+		id, err = restic.FindLatestSnapshot(repo, paths, "")
```
fd0: Have a look at how `FindLatestSnapshot` is used in the restore command. There, you can specify the list of paths and a hostname. I think we should add that here, too, to make the user interface consistent.

fd0 (changes requested):
Thanks for the PR, please add the options for paths and hostname.

ulziibuyan:
Options added. Should I open issue for this PR? Actually, just read CONTRIBUTING.MD

fd0:
It's not necessary to add an additional issue, the PR is enough for now.

fd0 (approved):

fd0:
LGTM, thanks for your work!

PR #730, opened 2017-01-18
Progress interval cli option

trbs:
This PR will allow a user to optionally specify the update interval for restic.NewProgress.


fd0:
Thanks for taking a stab at this! I must apologize: I haven't left any feedback for you because I needed to think about what to do with this feature first. I must admit that I do not like restic printing progress information when run as a cron job at all, so I planned to remove this periodic progress output for non-terminal runs completely.

What do you think? Do you need this feature? Or did you implemented this option to disable progress reporting for non-terminals?

trbs:
My use case is that the backups are schedules and run via Jenkins. Some are fairly large so it's nice to have some output logged but then every 30s or 60s is good enough. Think the same would go for cron jobs for having the output of the current run (or using logrotate).

I find this useful to check progress if needed when Restic is running on the server or for looking at the logs later to keep an eye on how backups are doing.

Restic already have a `-q` option right for when people want it to be silent (except for errors) during cron jobs ?

zcalusic:
I share the opinion with @fd0, that progress output for unattended runs is completely unneeded. Post-mortem progress just provides no useful information, and actually hides the warnings 'cause it's interspersed in the output. That doesn't mean setting interval is useless, just provide an option that completely disables progress logging, for cases where it's not useful.

Option `-q` is not the solution, it completely hides the output, and I wan't a few lines in the email to check that backup has run properly, how long it took, any warnings, someday how much data it added to the repo, etc...

trbs:
If we want I can add a PR (or add to this PR) for `--disable-progress`.

fd0:
Let me think about this for a bit. The problem with all command-line switches is that we cannot easily take them back, because that will break user's scripts. That's a bad thing and I'd like to prevent it if possible :)

trbs:
Turns out Restic will already slow down the progress interval to every 10s if it does not run on a terminal.

For situations where you want to force this behaviour or where a terminal is "faked" one can use:

```
$ restic backup ... | cat
```

For that reason I'm closing this PR.

PR #731, opened 2017-01-20
Improve memory usage

fd0:
This is a collection of commits which improve memory consumption.

### DIFF
--- a/src/cmds/restic/cmd_prune.go
+++ b/src/cmds/restic/cmd_prune.go
@@ -103,11 +103,13 @@ func runPrune(gopts GlobalOptions) error {
 		return err
 	}
 
+	blobs := 0
 	for _, pack := range idx.Packs {
 		stats.bytes += pack.Size
+		blobs += len(pack.Entries)
 	}
 	Verbosef("repository contains %v packs (%v blobs) with %v bytes\n",
-		len(idx.Packs), len(idx.Blobs), formatBytes(uint64(stats.bytes)))
+		len(idx.Packs), blobs, formatBytes(uint64(stats.bytes)))
 
 	blobCount := make(map[restic.BlobHandle]int)
 	duplicateBlobs := 0
@@ -164,14 +166,17 @@ func runPrune(gopts GlobalOptions) error {
 
 	// find packs that need a rewrite
 	rewritePacks := restic.NewIDSet()
-	for h, blob := range idx.Blobs {
-		if !usedBlobs.Has(h) {
-			rewritePacks.Merge(blob.Packs)
-			continue
-		}
+	for _, pack := range idx.Packs {
+		for _, blob := range pack.Entries {
+			h := restic.BlobHandle{ID: blob.ID, Type: blob.Type}
+			if !usedBlobs.Has(h) {
+				rewritePacks.Insert(pack.ID)
+				continue
+			}
 
-		if blobCount[h] > 1 {
-			rewritePacks.Merge(blob.Packs)
+			if blobCount[h] > 1 {
+				rewritePacks.Insert(pack.ID)
+			}
 		}
 	}
 
--- a/src/restic/archiver/archive_reader_test.go
+++ b/src/restic/archiver/archive_reader_test.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"math/rand"
 	"restic"
+	"restic/crypto"
 	"restic/repository"
 	"testing"
 )
@@ -44,7 +45,7 @@ func checkSavedFile(t *testing.T, repo restic.Repository, treeID restic.ID, name
 			t.Fatal(err)
 		}
 
-		buf := make([]byte, int(size))
+		buf := make([]byte, int(size), int(size)+crypto.Extension)
 		n := loadBlob(t, repo, id, buf)
 		if n != len(buf) {
 			t.Errorf("wrong number of bytes read, want %d, got %d", len(buf), n)
--- a/src/restic/archiver/archive_reader_test.go
+++ b/src/restic/archiver/archive_reader_test.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"math/rand"
 	"restic"
-	"restic/crypto"
 	"restic/repository"
 	"testing"
 )
@@ -45,7 +44,7 @@ func checkSavedFile(t *testing.T, repo restic.Repository, treeID restic.ID, name
 			t.Fatal(err)
 		}
 
-		buf := make([]byte, int(size), int(size)+crypto.Extension)
+		buf := restic.NewBlobBuffer(int(size))
 		n := loadBlob(t, repo, id, buf)
 		if n != len(buf) {
 			t.Errorf("wrong number of bytes read, want %d, got %d", len(buf), n)
--- a/src/restic/buffer.go
+++ b/src/restic/buffer.go
@@ -1,0 +1,21 @@
+package restic
+
+import "restic/crypto"
+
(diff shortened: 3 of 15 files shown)


mholt:
Nice! I see you added some benchmarks; I should probably just run those, but just curious (on a very high level), what kind of improvements did you see, or for what operations?

fd0:
Good question. I started with the `prune` operation because I noticed for one of my servers that restic uses A LOT of memory (~5GiB) during prune. There were several "gotcha" moments:
 * https://github.com/restic/restic/pull/731/commits/055668758459d73e390ea5d23f2f21d8b7a550a0 `Index.LookupSize()` was used a lot for non-existing blobs, and each time it was run it returned an error. We're using Dave Cheney's [errors](https://github.com/pkg/errors) package, which for each error allocated memory for a complete stack trace. That was way more memory than expected. So testing for the existence of a blob first (which does not return an error when the blob is not present) reduced the allocations.
 * https://github.com/restic/restic/pull/731/commits/9a5b9253c4bce73a68b08e9f6bd6ac205d8ebc8f changes the buffer handling for `repository.LoadBlob()`: the buffer passed to that functions needs to be large enough to hold the encrypted data (including IV and MAC), so we do not need to allocate a temporary buffer for the encrypted data. This again reduced allocations.
 * https://github.com/restic/restic/pull/731/commits/e571b6a6567a1a5fdc24741621f27b68d2f49103 also use the buffer of the encrypted data to hold the plaintext data after decryption.
 * https://github.com/restic/restic/pull/731/commits/caabc4ec44414d23e504ceb7b8a2a9d10e861a8d This one is especially unexpected: Turns out that for large objects like the index, `json.Unmarshal()` uses WAY less memory than `json.Decoder`. I've got no idea why, but it works.
 * https://github.com/restic/restic/pull/731/commits/73ad3d418d49ba12c219a3063a0164a6bec0f5bc When a new index is built from scratch, this commit fixes a double assignment for a new entry in the index. This is just a bug, I think the compiler even optimized it away because there was no change in memory usage.
 * https://github.com/restic/restic/pull/731/commits/d40f566e419fed34d108338e263af3221ec78d57 switches the index from maps to slices, this reduces the memory by about 1/2.
 * https://github.com/restic/restic/pull/731/commits/cd9b5262038f2ca29312fef805218aa8b95bec95 optimizes loading a pack header by pre-allocating the slice of blob entries.

Overall I got a lot of insight into how Go manages memory. Most important tool (besides `go test -bench . -benchmem` was [`gcvis`](https://github.com/davecheney/gcvis) (again by Dave)

mholt:
That was more detail than I was expecting :smile: - thanks for the writeup! These are good tricks to know, some definitely unexpected. Mind if I share this? Some good pointers here (pun intended).

fd0:
Not at all, please share this. If you think it's worth it, I could also write a blog entry for [the restic development blog](https://restic.github.io/blog).

mholt:
If you have the time, that'd be interesting. But I wouldn't want it to take much away from actual development either.

cstockton:
Hello @fd0, cool project, this is my first time checking it out after coming across it from a reddit post. I would suggest reverting https://github.com/restic/restic/pull/731/commits/e571b6a6567a1a5fdc24741621f27b68d2f49103 .. As a general rule of thumb, you shouldn't share buffers for cipher text and plain text without being very careful because it's more difficult to invariants because they may rely on weaker ones like error propagation. In this specific case- giving a identical buffer for a cipher and plaintext to a library which does not explicitly grant you permission to do so I would consider unsafe usage of the library. In this particular instance it may work but you are relying on implementation details of the XOR rounds. For example if the Go team decided to implement a more efficient algorithm in asm, or perhaps use segments of the buffers for scratch space.

fd0:
Thanks for your comment, it's always good to get feedback on my code!

The documentation for [the `cipher.Stream`](https://godoc.org/crypto/cipher#Stream) interface, which is returned by [`cipher.NewCTR()`](https://github.com/restic/restic/blob/master/src/restic/crypto/crypto.go#L312-L314) states that the ciphertext and plaintext may point to the same memory:
```
type Stream interface {
    // XORKeyStream XORs each byte in the given slice with a byte from the
    // cipher's key stream. Dst and src may point to the same memory.
    // If len(dst) < len(src), XORKeyStream should panic. It is acceptable
    // to pass a dst bigger than src, and in that case, XORKeyStream will
    // only update dst[:len(src)] and will not touch the rest of dst.
    XORKeyStream(dst, src []byte)
}
```

On amd64, the Go stdlib implementation for AES already is implemented in assembler.

Can you explain what you mean by saying: "it's more difficult to prove your invariants because they may rely on weaker ones like error propagation."?

The ciphertext is protected by a MAC (using Poly1305-AES), which is always checked first, with explicit error handling. Do you think that this is not enough?

cstockton:
My apologies, I stand corrected in regards to:
>  In this specific case- giving a identical buffer for a cipher and plaintext to a library which does not explicitly grant you permission to do so I would consider unsafe usage of the library.

Since they grant you permission explicitly, the usage of XORKeyStream should be considered safe as that negates the two examples I gave you for why it should be avoided.

> On amd64, the Go stdlib implementation for AES already is implemented in assembler.

The block cipher is surely implemented in ASM, but I would be surprised if the counter mode of operation was (though may stand correct again!). This doesn't matter though since they grant explicit permission to pass the same buffer.

> Can you explain what you mean by saying: "it's more difficult to prove your invariants because they may rely on weaker ones like error propagation."?

Yes- in general in a security context you want to be concise so it's easy to prove your implementation is correct. Here it's hard to do so and raised some immediate red flags to me. It was not clear to as https://github.com/restic/restic/blob/e571b6a6567a1a5fdc24741621f27b68d2f49103/src/restic/repository/repository.go#L115 -> https://github.com/restic/restic/blob/e571b6a6567a1a5fdc24741621f27b68d2f49103/src/restic/repository/repository.go#L411 looks like a "OOPS?" as you cross API boundaries into a system that has "plaintext" and "ciphertext" as the names of the buffers with both being "plaintext" at the call site.

> The ciphertext is protected by a MAC (using Poly1305-AES), which is always checked first, with explicit error handling. Do you think that this is not enough?

This is more centered around object reuse in general, which there is lots of reading material so I won't deep dive that. Just to reiterate I don't think your implementation is incorrect or you have made any mistakes, you're doing some fantastic work here! It's refreshing to see security taken seriously here. I just reported this because it looked like a mistake at first glance and I wanted to advocate against buffer reuse in general. Have a good evening (or day).

fd0:
Hey Chris,

Thanks for the explanation. To be explicit: I appreciate your comments and thoughts about the design and code of restic! I think it's good practice to comment on an explain things, especially if the code in question triggers any red flags in regards to security. After all, restic is built with [security in mind](https://github.com/restic/restic/blob/master/doc/Design.md#threat-model), so I like to have a productive code review and exchange of ideas/criticism and set the tone (here in the bug tracker and in the IRC channel) in a way that makes it productive.

I think you made a good point, the crypto code needs to be cleaned up and the variables and API made more concise and explicit.

Thanks again.

zcalusic:
FWIW, I see 40% reduction in memory usage with these changes. Good job @fd0. 👍

For repo with 11055 packs (2182938 blobs) with 49.525 GiB bytes memory usage (rss) goes down from 2.3GB to 1.4GB.

fd0:
For which operation? `prune`?

zcalusic:
Yes, prune, forgot to add...

fd0:
Ok, thanks!

PR #746, opened 2017-01-24
fuse: Improve memory usage

### LINKED ISSUE #480: Memory leak in fuse mount
If I do a "ls -lRa" through multiple snapshots, the amount of memory used by the restic process increases. I stopped it when it took 7GB of RAM.
## Output of `restic version`

restic 0.1.0 (v0.1.0-451-g4dac6d4)
compiled at 2016-03-02 17:56:59 with go1.5.3
## Expected behavior

Keep the amount of memory used low, for an ls.
## Actual behavior

Too much memory used

fd0:
Discard blobs that aren't in use any more. This greatly reduces memory
usage and will probably only trigger on sequential read (e.g. for
restore via fuse).

Closes #480

### DIFF
--- a/src/restic/fuse/file.go
+++ b/src/restic/fuse/file.go
@@ -91,6 +91,11 @@ func (f *file) getBlobAt(i int) (blob []byte, err error) {
 		return f.blobs[i], nil
 	}
 
+	// release earlier blobs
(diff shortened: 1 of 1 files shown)


zcalusic:
Good job, @fd0!

PR #750, opened 2017-01-28
Document creating a new repo via REST

### LINKED ISSUE #736: Document creating a new REST backend
As agreed with @zcalusic: Amend the REST protocol description and add an explicit `POST {path}?create=true` to create the repository on the server side.

fd0:
Closes #736

CC @zcalusic


zcalusic (commented):
Commenting this part "If the repo already exists, the HTTP error "409 Conflict" is returned." of `POST {path}?create=true`. Currently os.MkdirAll() is used to create directories, which will not complain if they already exist. Also, the directory creation is a non-destructive operation, so I wonder do we really need "409 Confict" in the spec?

fd0:
Hm, it's not strictly necessary, we could also just test if there is a config file...

Issue #726, opened 2017-01-18
Add --json option to "restic snapshots" (feature)


yhafri:
## Output of `restic version`

Any

## Expected behavior

Being able to embed the result of `snapshots` in other services (ex. daily nice formatted HTML report).

```
$ restic --json -r /tmp/backup snapshots
{
  [ "id:"40dc1520", "date":"2015-05-08 21:38:30", "host":"kasimir", "tag":"", "directory":"/home/user/work" ],
  [ "id:"79766175", "date":"2015-05-08 21:40:19", "host":"kasimir", "tag":"", "directory":"/home/user/work" ]
}
```

## Actual behavior

```
$ restic -r /tmp/backup snapshots
ID        Date                 Host    Tags   Directory
----------------------------------------------------------------------
40dc1520  2015-05-08 21:38:30  kasimir        /home/user/work
79766175  2015-05-08 21:40:19  kasimir        /home/user/work
```

fd0:
Interesting idea, thanks for reporting it!

yhafri:
Fantastic job @fd0. 
I'll start testing and report any issue I may found

yhafri:
And special thanks to @ibib for the PR

yhafri:
Hi guys

Can we extend this *--json* feature to the *find* command?
Something like:

`$ restic --json find -s 82a6e9de *.pdf`

@ibib @fd0 your thoughts on this please ?

ibib:
Sounds great. My masterplan is to implement a JSON output for every command.

The Problem is that i cant give you a timeline for it.

yhafri:
@ibib excellent
As I've told @fd0 before, we can donate for it if you're interested.

fd0:
I think it's a good idea (I've talked about it with @ibib already). I'm going to close this issue, since the option for the `snapshots` command is already implemented.

PR #751, opened 2017-01-28
Added json ouput of snapshots

ibib:
This PR adds json support as requested in [#726](https://github.com/restic/restic/issues/726)

### DIFF
--- a/src/cmds/restic/cmd_snapshots.go
+++ b/src/cmds/restic/cmd_snapshots.go
@@ -85,11 +85,14 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 
 	}
 
-	if gopts.JSON == true {
-		jsonoutput(list)
-	} else {
-		naturaloutput(list)
+	if gopts.JSON {
+		err := jsonoutput(list)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error printing snapshot: %v\n", err)
+		}
+		return nil
 	}
+	naturaloutput(list)
 
 	return nil
 }
--- a/src/cmds/restic/cmd_snapshots.go
+++ b/src/cmds/restic/cmd_snapshots.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"restic"
+	"time"
 )
 
 var cmdSnapshots = &cobra.Command{
@@ -151,35 +152,47 @@ func printSnapshotsReadable(list []*restic.Snapshot) {
 	return
 }
 
-//Snapshot provides struct to store snapshots
+// Snapshot helps to print Snaphots as JSON
 type Snapshot struct {
-	ID          string   `json:"id"`
-	Date        string   `json:"date"`
-	Host        string   `json:"host"`
-	Tags        []string `json:"tags"`
-	Directories []string `json:"directories"`
+	Time     time.Time `json:"time"`
+	Parent   string    `json:"parent,omitempty"`
+	Tree     string    `json:"tree,omitempty"`
+	Paths    []string  `json:"paths"`
+	Hostname string    `json:"hostname,omitempty"`
+	Username string    `json:"username,omitempty"`
+	UID      uint32    `json:"uid,omitempty"`
+	GID      uint32    `json:"gid,omitempty"`
+	Excludes []string  `json:"excludes,omitempty"`
+	Tags     []string  `json:"tags,omitempty"`
+	ID       string    `json:"id"`
 }
 
 // printSnapshotsJSON writes the JSON representation of list to stdout.
 func printSnapshotsJSON(list []*restic.Snapshot) error {
 
 	enc := json.NewEncoder(os.Stdout)
-	//set JSON prefix
+	// set JSON prefix
 	fmt.Fprint(os.Stdout, "[")
 
 	for i, sn := range list {
 
-		//set JSON delimiter
+		// set JSON delimiter
 		if i != 0 {
 			fmt.Fprint(os.Stdout, ",")
 		}
 
 		k := Snapshot{
-			ID:          sn.ID().Str(),
-			Date:        sn.Time.Format(TimeFormat),
-			Host:        sn.Hostname,
-			Tags:        sn.Tags,
-			Directories: sn.Paths}
+			Time:     sn.Time,
+			Parent:   sn.Parent.PrintStr(),
+			Tree:     sn.Tree.PrintStr(),
+			Paths:    sn.Paths,
+			Hostname: sn.Hostname,
+			Username: sn.Username,
+			UID:      sn.UID,
+			GID:      sn.GID,
+			Excludes: sn.Excludes,
+			Tags:     sn.Tags,
+			ID:       sn.ID().String()}
 
 		err := enc.Encode(k)
 		if err != nil {
@@ -188,7 +201,7 @@ func printSnapshotsJSON(list []*restic.Snapshot) error {
 
 	}
 
-	//set JSON postfix
+	// set JSON postfix
 	fmt.Fprint(os.Stdout, "]")
 
 	return nil
--- a/src/cmds/restic/cmd_snapshots.go
+++ b/src/cmds/restic/cmd_snapshots.go
@@ -86,19 +86,19 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 	}
 
 	if gopts.JSON {
-		err := jsonoutput(list)
+		err := printSnapshotsJSON(list)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "error printing snapshot: %v\n", err)
 		}
 		return nil
 	}
-	naturaloutput(list)
+	printSnapshotsReadable(list)
 
 	return nil
 }
 
-//naturaloutput provides human redability
-func naturaloutput(list []*restic.Snapshot) {
+//printSnapshotsReadable provides human redability
+func printSnapshotsReadable(list []*restic.Snapshot) {
 
 	tab := NewTable()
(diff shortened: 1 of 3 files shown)


src/cmds/restic/cmd_snapshots.go
```diff
@@ -26,6 +27,7 @@ The "snapshots" command lists all snapshots stored in a repository.
 type SnapshotOptions struct {
 	Host  string
 	Paths []string
+	Json  bool
```
fd0: Make this a global option, so we can add it to other commands as well.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -87,6 +86,22 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 
 	}
 
+	if opts.Json == true {
```
fd0: The `==` is unneeded, please remove it.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -87,6 +86,22 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+		jsonoutput(list)
```
fd0: Add a `return` after this and remove the `else` branch.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -87,6 +86,22 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+	} else {
+		naturaloutput(list)
+	}
+
+	return nil
+}
+
+//naturaloutput provides human redability
+func naturaloutput(list []*restic.Snapshot) {
```
fd0: This function needs a new name, all functions in the restic binary share the same namespace. What about `printSnapshots()`?

src/cmds/restic/cmd_snapshots.go
```diff
@@ -121,5 +136,40 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+	return
+}
+
+//Snapshot provides struct to store snapshots
+type Snapshot struct {
+	ID        string   `json:"id"`
+	Date      string   `json:"date"`
+	Host      string   `json:"host"`
+	Tags      []string `json:"tags"`
+	Directory []string `json:"directory"`
```
fd0: Please also use the plural `Directories` here.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -121,5 +136,40 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+}
+
+//jsonoutput provides machine redability
+func jsonoutput(list []*restic.Snapshot) error {
```
fd0: This function also needs a different name. What about `printSnapshotsJSON()`?

src/cmds/restic/cmd_snapshots.go
```diff
@@ -121,5 +136,40 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+			ID:        sn.ID().Str(),
+			Date:      sn.Time.Format(TimeFormat),
+			Host:      sn.Hostname,
+			Tags:      sn.Tags,
+			Directory: sn.Paths}
+
+		response = append(response, k)
+	}
+
+	output, _ := json.Marshal(response)
```
fd0: Use a `json.Decoder` here, like `dec := json.NewDecoder(os.Stdout)` and then use its `Decode()` method.

fd0 (changes requested):
Thanks for your work, please see my review comments.

ibib:
Thank you for your annotations.
Sorry that this PR is a bit ugly, this is my first opensource contribution.

I am working now on the JSON encode thingy and will then look in how to resolve conflicts.

The describing text at the global JSON option was choosen on purpose as it only works with the snapshot option at the moment. `print snapshots in json format`

I cant use `printSnapshots()` as function name as it is already in use [here](https://github.com/restic/restic/blob/master/src/cmds/restic/cmd_forget.go#L63).
Would this be a candidate for refactoration?

fd0:
> Sorry that this PR is a bit ugly, this is my first opensource contribution.

Oh, that's great, I'm honored!

> The describing text at the global JSON option was chosen on purpose as it only works with the snapshot option at the moment. print snapshots in json format

Okay, we'll leave it for now.

> I cant use printSnapshots() as function name as it is already in use here.
> Would this be a candidate for refactoration?

Yes, but not in this PR.

I'll add further comments now.

Are you aware that you accidentally merged the master branch in again? I know that git and GitHub are a hard to understand for beginners.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -87,6 +84,25 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+			fmt.Fprintf(os.Stderr, "error printing snapshot: %v\n", err)
+		}
+		return nil
+	}
+	printSnapshotsReadable(list)
+
+	return nil
+}
+
+//printSnapshotsReadable provides human redability
```
fd0: Please describe what the function does, e.g.:

```
// printSnapshotsReadable prints a text table of the snapshots in list to stdout.
```

Also add a space character after the comment (`//`), and make sure that the description is a complete sentence, as described in https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences.

src/cmds/restic/cmd_snapshots.go
```diff
@@ -132,5 +148,49 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 
 	tab.Write(os.Stdout)
 
+	return
+}
+
+//Snapshot provides struct to store snapshots
```
fd0: Actually I'd prefer to use the JSON representation we're storing in the repository, it is defined on the `Snapshot` struct here (so you do not need a special struct at all): https://github.com/restic/restic/blob/master/src/restic/snapshot.go#L13-L27

src/cmds/restic/cmd_snapshots.go
```diff
@@ -132,5 +148,49 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+type Snapshot struct {
+	ID          string   `json:"id"`
+	Date        string   `json:"date"`
+	Host        string   `json:"host"`
+	Tags        []string `json:"tags"`
+	Directories []string `json:"directories"`
+}
+
+//printSnapshotsJSON provides machine redability
```
fd0: Add a space after `//`, then describe what the function does, e.g.:
```
// printSnapshotsJSON writes the JSON representation of list to stdout.
```

src/cmds/restic/cmd_snapshots.go
```diff
@@ -132,5 +148,49 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
...
+func printSnapshotsJSON(list []*restic.Snapshot) error {
+
+	enc := json.NewEncoder(os.Stdout)
+	//set JSON prefix
+	fmt.Fprint(os.Stdout, "[")
```
fd0: Are you aware that you can just pass a `[]Snapshot` to the encoder? Like this:
```go
err := json.NewEncoder(os.Stdout).Encode(list)
```

The encoder prints the list delimiters (`[` and `]`) automatically.

src/cmds/restic/global.go
```diff
@@ -53,6 +54,7 @@ func init() {
 	f.StringVarP(&globalOptions.PasswordFile, "password-file", "p", "", "read the repository password from a file")
 	f.BoolVarP(&globalOptions.Quiet, "quiet", "q", false, "do not output comprehensive progress report")
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repo, this allows some operations on read-only repos")
+	f.BoolVarP(&globalOptions.JSON, "json", "", false, `print snapshots in json format`)
```
fd0: Use double quotes (`"`) for the help text. What do you think about the following help text: `set output mode to JSON for commands that support it`

fd0 (changes requested):

fd0:
Hey, do you have any questions? Is there anything I can do to help you? Or do you just not have any spare time at the moment? :)

ibib:
I tried to use [`err := json.NewEncoder(os.Stdout).Encode(list)`](https://gist.github.com/ibib/f705fb8982bcf0295f9e75f7c22ce3f1) as suggested but i couldn't work out how to print the id's.
The latest commit should provide the same result.

fd0:
You can create a new struct `Snapshot` that [embeds](https://golang.org/ref/spec#Struct_types)  `*restic.Snapshot`, and add an `ID` field like this:
```go
// Snapshot helps to print Snaphots as JSON
type Snapshot struct {
	*restic.Snapshot

	ID string `json:"id"`
}
```

It has all the fields `restic.Snapshot` has, and an additional `ID` field, and you don't even need to repeat all field names!

This new struct can then be used to construct a list of `Snapshot` like this:
```go
	var snapshots []Snapshot
	for _, sn := range list {
		k := Snapshot{
			Snapshot: sn,
			ID:       sn.ID().String(),
		}

		snapshots = append(snapshots, k)
	}
```
Finally, just print this list:
```go
	return json.NewEncoder(os.Stdout).Encode(snapshots)
```

ibib:
Wow. That is totally more easy. 

Thank you very much for your help.

fd0 (approved):

fd0 (changes requested):
Very good! Now before this can get merged, there are a couple of smaller things that need to be done:
 * Squash your changes into one commit
 * Write a short paragraph for the manual, perhaps titled "Scripting restic", which mentions `--json` and shows an example usage with `snapshots`.

Thanks for your work!

PR #752, opened 2017-01-28
Support for hardlinks with the restore command

### LINKED ISSUE #152: Support hard links
At the moment, restic saves enough meta data to restore hard links, but this is not implemented yet.

jgfrm:
Requires repository change to store the device a files is stored

Closes #152

### DIFF
--- a/src/restic/fs/file.go
+++ b/src/restic/fs/file.go
@@ -102,6 +102,12 @@ func Symlink(oldname, newname string) error {
 	return os.Symlink(fixpath(oldname), fixpath(newname))
 }
 
+// Hardlink creates newname as a hard link to oldname.
+// If there is an error, it will be of type *LinkError.
+func Hardlink(oldname, newname string) error {
+	return os.Link(fixpath(oldname), fixpath(newname))
+}
+
 // Stat returns a FileInfo structure describing the named file.
 // If there is an error, it will be of type *PathError.
 func Stat(name string) (os.FileInfo, error) {
--- a/src/restic/hardlinks_index.go
+++ b/src/restic/hardlinks_index.go
@@ -5,17 +5,19 @@ import (
 	"fmt"
 )
 
-type hardlinkKey struct {
+// HardlinkKey is a composed key for finding inodes on a specific device
+type HardlinkKey struct {
     Inode, Device uint64
 }
 
-type hardlinkData struct {
+// HardlinkData contains to number of links for an inode, and the filename
+type HardlinkData struct {
     Links uint64
     Name string
 }
 
 var (
-	hardLinkIndex      = make(map[hardlinkKey]*hardlinkData)
+	hardLinkIndex      = make(map[HardlinkKey]*HardlinkData)
 	hardLinkIndexMutex = sync.RWMutex{}
 )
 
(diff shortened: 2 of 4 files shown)


jgfrm (commented):
changed exports and comments

jgfrm:
export datatypes

jgfrm (commented):
export datatypes

fd0:
FYI: The tests on Travis pass, it fails because some code is not properly formatted with `gofmt`: https://travis-ci.org/restic/restic/jobs/196162594#L254

I'd suggest installing the precommit-hook as outlined here: https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches
It will prevent you from committing anything that is not properly formatted, even before you push your changes to GitHub

src/restic/fs/file.go
```diff
@@ -102,6 +102,12 @@ func Symlink(oldname, newname string) error {
 	return os.Symlink(fixpath(oldname), fixpath(newname))
 }
 
+// Hardlink creates newname as a hard link to oldname.
+// If there is an error, it will be of type *LinkError.
+func Hardlink(oldname, newname string) error {
```
fd0: Why is this function called `Hardlink()`, an not just `Link()` as the corresponding function in package `os`?

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// HardlinkData contains to number of links for an inode, and the filename
+type HardlinkData struct {
+    Links uint64
+    Name string
+}
+
+var (
+	hardLinkIndex      = make(map[HardlinkKey]*HardlinkData)
```
fd0: Having just on global instance is a bad idea, make this a separate type and embed the mutex. Also, please just use a simple `Mutex`, not a `sync.RWMutex`, that's not really needed here.

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,99 @@
...
+	hardLinkIndexMutex = sync.RWMutex{}
+)
+
+// ExistsLink checks wether the link already exist in the index
+func ExistsLink(inode uint64, device uint64) bool {
```
fd0: All functions in this file needed to be attached to the new type.

src/restic/node.go
```diff
@@ -482,6 +500,7 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	case "file":
 		node.Size = uint64(stat.size())
 		node.Links = uint64(stat.nlink())
+		node.Device = uint64(stat.dev())
```
fd0: Ok, that was the missing part for being able to restore hardlinks.

fd0 (changes requested):
Thanks a lot for your work!

I've read through your code. I think we can do with less code. First, please use a standard `sync.Mutex`, not a `sync.RWMutex`, the latter is way more complicated to reason about and not really necessary here.

Please make the hardlink index a new type and do not just use a single global instance.

I think we do not need to decrement the link count, or even keep track of the link count during restore. There are plenty of reasons why a counterpart of a restored file was not restored (e.g. a filter), so it's impossible to reason why a hardlink could not be restored, so I think we should not print anything.

In order for this to be merged please add tests for the new type.

These are my comments so far. If you have any questions, feel free to ask, I'm happy to assist!

jgfrm:
Thanks for the feedback.
I have a few questions:
1. "Having just on global instance is a bad idea, make this a separate type". Can you pinpoint me to an example in the source I can have a look at? I need to create the link index on beforehand, and I thought a global instance is the way to do that.
2. "All functions in this file needed to be attached to the new type". Some questions: Is there an example in the code I can look at?

Thanks,

JG

fd0:
Hey, you can have a look at the `Lock` struct for example, defined here: https://github.com/restic/restic/blob/master/src/restic/lock.go#L26, here's and example for a method of that type: https://github.com/restic/restic/blob/master/src/restic/lock.go#L118

jgfrm:
I had a look at the source, but I do not precisely understand the issue.
Just started with Go

+var (
+	hardLinkIndex      = make(map[HardlinkKey]*HardlinkData)
Having just on global instance is a bad idea, make this a separate type and embed the mutex.

fd0:
I'd suggest going through the excellent [Go Tour](https://tour.golang.org/) if you haven't done already, especially the section on ["more types"](https://tour.golang.org/moretypes/2) and ["methods"](https://tour.golang.org/methods/1).

As for a concrete proposal, make a new type like this:
```go
type HardlinkIndex struct {
    m sync.Mutex
    Index map[HardlinkKey]string // holds [inode,device] -> filename
}
```

and attach the functions to the new type, e.g. a `Has()` method:
```go
func (idx *HardlinkIndex) Has(inode, device uint64) bool {
    idx.m.Lock()
    defer idx.m.Unlock()
    _, ok := idx.Index[HardlinkKey{Inode: inode, Device: device}]
    return ok
}
```

Let me know if you need more assistance. And thanks again for jumping into the cold water and trying this yourself!

jgfrm:
Thanks for your help.
The problem is however that the index shoudl be created once for the full restore process.
So, the node cannot create the index.
This was the motivation to use a global variable for the index

jgfrm:
Hi,

I have xattrs for files working.
There are however two problems.

1.	I need a full path to the file (in case the node is for a file which is there) and a full path to the directory (in case the node is for a dir).
What happens is that fileWorker uses a full path:
node, err := restic.NodeFromFileInfo(e.Fullpath(), e.Info())
where dirWorker uses a relative path
n, err := restic.NodeFromFileInfo(dir.Path(), dir.Info())

My question:

*	how can I obtain the full path of a dir, specifically in node.fillExtra

2.	I also tried to implement xattr for fuse.
To call the proper syscall of the fuse package (syscallx.Setxattr), I need also the full path of the file or dir.
Again here: how can I obtain the full path?

Thanks

fd0:
So, then I'd suggest adding a field for the index to [the Restorer struct](https://github.com/restic/restic/blob/master/src/restic/restorer.go#L14-L16) and an additional parameter to [CreateAt()](https://github.com/restic/restic/blob/master/src/restic/node.go#L100) for the index.

fd0:
This PR is superseded by #763, closing.

PR #754, opened 2017-01-29
systemFuse.AllowOther() in cmd_mount.go

Thamster:
This small change, allows restic fuse mounts to be browsed on NAS shares by non root users. The issue was that although, mounted subfolders were correctly chowned by the user executing the restic mount command, the mount point itself was owned by root and not browsable by non root users.


fd0:
Hey, thanks for giving this a try and even contributing code!

It's a security feature that, by default, only the user running the process which implements the fuse filesystem can read it. So I'm not going to merge a PR that will always allow root to read the filesystem mounted provided by restic, therefore I'm closing this PR (as for now).

I'm however willing to merge a PR which adds the options `--allow-others` and `--allow-root` to the `mount` command (including proper documentation in the manual), so that you can enable this behavior (similar to `sshfs` options `-o allow_other` and `-o allow_root`).

If you need any help, please let me know.

PR #763, opened 2017-01-30
Support hard links

jgfrm:
Made struct for HardlinkIndex
Reworked global variable and added creation of HardlinkIndex in restorer
Bind functions to type


fd0:
Does this PR supersede #752?

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,57 @@
+package restic
+
+import (
+	"sync"
+)
+
+// HardlinkKey is a composed key for finding inodes on a specific device
```
fd0: Make all comments which describe something full sentences, see https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,57 @@
...
+
+// NewHardlinkIndex create a new index for hard links
+func NewHardlinkIndex() *HardlinkIndex {
+	return &HardlinkIndex{
+		Index: make(map[HardlinkKey]string),
+	}
+}
+
+// ExistsLink checks wether the link already exist in the index
+func (idx *HardlinkIndex) ExistsLink(inode uint64, device uint64) bool {
```
fd0: I'd prefer this function to be named `Has()` instead of `ExistsLink()`, that's the same style as [the index](https://godoc.org/github.com/restic/restic/src/restic/repository#Index.Has).

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,57 @@
...
+	idx.m.Lock()
+	defer idx.m.Unlock()
+	_, ok := idx.Index[HardlinkKey{inode, device}]
+
+	return ok
+}
+
+// AddLink adds a link to the index
+func (idx *HardlinkIndex) AddLink(inode uint64, device uint64, name string) {
```
fd0: This function can just be named `Add()`.

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,57 @@
...
+	defer idx.m.Unlock()
+	_, ok := idx.Index[HardlinkKey{inode, device}]
+
+	if !ok {
+		idx.Index[HardlinkKey{inode, device}] = name
+	}
+}
+
+// GetLinkName obtains the filename from the index
+func (idx *HardlinkIndex) GetLinkName(inode uint64, device uint64) string {
```
fd0: Hm, `GetFilename` would be more appropriate as a function name I think.

This function also needs to return an error, what if there isn't a file in the index with this inode and device?

src/restic/hardlinks_index.go
```diff
@@ -0,0 +1,57 @@
...
+	idx.m.Lock()
+	defer idx.m.Unlock()
+	return idx.Index[HardlinkKey{inode, device}]
+}
+
+// RemoveLink removes a link from the index
+func (idx *HardlinkIndex) RemoveLink(inode uint64, device uint64) {
```
fd0: Remove the `Link` from the function name, `Remove()` is clear enough I think.

src/restic/hardlinks_index_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestHardLinks(t *testing.T) {
+
+	idx := restic.NewHardlinkIndex()
+
+	idx.AddLink(1, 2, "inode1-file1-on-device2")
+	idx.AddLink(2, 3, "inode2-file2-on-device3")
+
+	var sresult string
+	sresult = idx.GetLinkName(1, 2)
+	Assert(t, sresult == "inode1-file1-on-device2",
```
fd0: Use `Equals()` instead of `Assert()`, sample is here: https://github.com/restic/restic/blob/master/src/restic/repository/index_test.go#L75

vendor/src/golang.org/x/crypto/ssh/agent/example_test.go
```diff
@@ -6,20 +6,20 @@ package agent_test
 
 import (
 	"log"
-	"os"
 	"net"
+	"os"
 
-        "golang.org/x/crypto/ssh"
```
fd0: This does not belong into the PR, did you accidentally run `gofmt` on the `vendor` directory?

vendor/src/golang.org/x/crypto/ssh/keys_test.go
```diff
@@ -309,14 +309,14 @@ func TestInvalidEntry(t *testing.T) {
 }
 
 var knownHostsParseTests = []struct {
-	input     string
```
fd0: Same here, do not change the vendored libraries.

src/restic/node.go
```diff
@@ -191,7 +191,19 @@ func (node Node) createDirAt(path string) error {
 	return nil
 }
 
-func (node Node) createFileAt(path string, repo Repository) error {
+func (node Node) createFileAt(path string, repo Repository, idx *HardlinkIndex) error {
+	if node.Links > 1 {
+		if !idx.ExistsLink(node.Inode, node.Device) {
+			idx.AddLink(node.Inode, node.Device, path)
```
fd0: I think the node should only be added to the hardlink index when the restore was successful, so you need to move this to the end of the function. Otherwise restoring a file may fail, and with the current code all hardlinked versions of that file will also not be restored because the hardlink index contains a filename that may not exist (or even point to different content).

fd0 (changes requested):
I've added my review comments.

jgfrm:
Can someone explain why the CI failed (and how I can see that)?

fd0:
Click on the red `x`, then on `Details`, e.g. for Travis. Then you'll get to the build's page, e.g. here:
https://travis-ci.org/restic/restic/builds/197350087

The tests for Travis fail at the moment because the infrastructure for running tests on Darwin/OS X is broken: https://www.traviscistatus.com/incidents/k79mjcv403c4

Let's just wait a bit.

jgfrm:
should I do something about the gofmt on the vendor libs?

fd0:
Yes: Remove your changes of the files in vendor/. Do you need help doing that?

jgfrm:
If you can give me a clue how to do that would be more than welcome.

 

Van: Alexander Neumann [mailto:notifications@github.com] 
Verzonden: donderdag 2 februari 2017 10:02
Aan: restic/restic <restic@noreply.github.com>
CC: jgfrm <frm@gordijn.org>; Author <author@noreply.github.com>
Onderwerp: Re: [restic/restic] Changed Hardlink() in Link() (#763)

 

Yes: Remove your changes of the files in vendor/. Do you need help doing that?

fd0:
First, find out which commit(s) touch the files in the vendor directory (this assumes you are in the branch `issue25`):
```
$ git log --stat master..HEAD -- vendor/
commit 369f9804114d45876078cb26593bc5552e1ccf54
Author: Jaap Gordijn <jaap@intra.gordijn.org>
Date:   Tue Jan 31 00:14:20 2017 +0100

    Changed Hardlink() in Link()
    Made struct for HardlinkIndex
    Reworked global variable and added creation of HardlinkIndex in restorer
    Bind functions to type

 vendor/src/golang.org/x/crypto/ssh/agent/example_test.go | 14 +++++++-------
 vendor/src/golang.org/x/crypto/ssh/keys_test.go          | 20 ++++++++++----------
 2 files changed, 17 insertions(+), 17 deletions(-)
```

Start an interactive rebase:
```
$ git rebase --interactive master
```

This opens an editor with a few lines at the top, one for each commit. Replace the word `pick` in the line for the commit `369f980` with the word `edit`, then save and quit the editor.

It then looks like this:
```
$ git rebase --interactive master
Stopped at 369f980... Changed Hardlink() in Link() Made struct for HardlinkIndex Reworked global variable and added creation of HardlinkIndex in restorer Bind functions to type
You can amend the commit now, with

	git commit --amend 

Once you are satisfied with your changes, run

	git rebase --continue
```

Then discard the previous commit (so we can create a new one):
```
$ git reset HEAD~
Unstaged changes after reset:
M	src/restic/fs/file.go
M	src/restic/node.go
M	src/restic/node_test.go
M	src/restic/restorer.go
M	vendor/src/golang.org/x/crypto/ssh/agent/example_test.go
M	vendor/src/golang.org/x/crypto/ssh/keys_test.go
```

Discard all changes to files in the `vendor/` directory:
```
$ git checkout vendor/
$ git status
[...]

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/restic/fs/file.go
	modified:   src/restic/node.go
	modified:   src/restic/node_test.go
	modified:   src/restic/restorer.go

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	src/restic/hardlinks_index.go
	src/restic/hardlinks_index_test.go
```

Then add the files and make a commit as usual. This time, please adhere to the [commit message format](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits):
```
$ git add src
$ git commit
```

Finish the rebase:
```
$ git rebase --continue
Successfully rebased and updated refs/heads/issue25.
```

Then push the changes to github, overriding the old ones (hence the `--force`):
```
$ git push --force
```

That should do it :)

jgfrm:
I have tried what you suggested.
However, the step
`it reset HEAD~`
results in 
`M       src/restic/hardlinks_index_test.go`
and does not show the checked in vendor files.

Also: what am I supposed to do at the 
`git commit --amend`
step?

thanks

fd0:
Hm, interesting. Thanks for trying, though. Just leave it in, I'll fix that myself later. You should be able to abort the rebase by running `git reset --hard` and then `git rebase --abort`.

jgfrm:
Ok. Thanks.

fd0:
Ok, your changes look good, thanks! The last thing that I'd like to discuss with you is: I think we need an option for `restore` that allows users not to restore the hard linked files. What do you think?

fd0 (approved):

jgfrm:
Wellm rsync has the option --hard-links. Then you have to opt in to have hard link support.

pvgoran:
I believe that by default a backup application should capture and restore as much details as possible. `rsync` is a copy tool, it's expected to have a different set of defaults.

P.S. I'm glad that hardlinks and xattrs finally get support in Restic, without this I couldn't consider it a viable solution for my backup needs.

zcalusic:
I share the sentiment with @pvgoran, xattr support is greatly appreciated @jgfrm. :+1:

jgfrm:
Above commit fixes some issues with hardlinks and the repo. The repo did not store hardlink info for dev and symlink.
Hardlinks for fuse work also, but is not included in this commit.
Once this commit in included in master, I will upload a new PR for fuse support

src/restic/node.go
```diff
@@ -495,13 +495,17 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 		node.Size = uint64(stat.size())
 		node.Links = uint64(stat.nlink())
 	case "dir":
+		node.Links = uint64(stat.nlink())
```
fd0: Directories cannot be hard-linked, so this line is not needed.

fd0 (changes requested):
Ok, I think this is good so far. Thanks for your comments, I agree that restoring hardlinks should be the default. We can add an option when users request that.

So from my point of view the only thing missing is an integration test which archives hardlinked files, restores them again and verifies that the restored files are indeed hard links.

src/restic/node.go

jgfrm:
Can you give me a clue where to add the integration test?

fd0:
Ah, it seems you found a place for the integration test. :)

fd0:
So, just format the remaining files with `gofmt` (that's why Travis failed), then we're good to merge I think :)

jgfrm:
Thanks.
A question: is it possible to run the integration test locally?

jgfrm:
I supposed that the code was already merged into the master branch, but apparently, it is not. Am I supposed to do anything?

fd0:
Thanks for asking, I didn't find the time yet to review the latest commits you pushed. I'll add a comment when something needs to be changed. Once your code is merged, I'll close this PR.

In order to cleanup the commit history, I will squash all your commits into one, and remove the changes `gofmt` did on the files in the `vendor` folder.

src/cmds/restic/integration_helpers_unix_test.go
```diff
@@ -37,5 +37,15 @@ func (e *dirEntry) equals(other *dirEntry) bool {
 		return false
 	}
 
+	if stat.Nlink != stat2.Nlink {
+		fmt.Fprintf(os.Stderr, "%v: Number of links doe not match (%v != %v)\n", e.path, stat.Nlink, stat2.Nlink)
```
fd0: Typo: "doe not match" -> "do not match"

src/cmds/restic/integration_test.go
```diff
@@ -1011,3 +1011,40 @@ func TestPrune(t *testing.T) {
...
+			"expected one snapshot, got %v", snapshotIDs)
+
+		testRunCheck(t, gopts)
+
+		// restore all backups and compare
+		for i, snapshotID := range snapshotIDs {
+			restoredir := filepath.Join(env.base, fmt.Sprintf("restore%d", i))
+			t.Logf("restoring snapshot %v to %v", snapshotID.Str(), restoredir)
+			testRunRestore(t, gopts, restoredir, snapshotIDs[0])
+			Assert(t, directoriesEqualContents(env.testdata, filepath.Join(restoredir, "testdata")),
```
fd0: This test is not enough: It only tests that the link count for each item in the the restored directory is the same as in the archived directory. This only verifies that files which were hardlinks before, are hardlinks afterwards, but does not test whether the same files are hardlinks of each other.

Suppose we have the files `A` and `B`, and the files `a`, which is a hard link of `A` and `b`, which is a hard link of `B`: `A` <-> `a` and `B` <-> `b`.

The test will succeed for `A` <-> `a` and `B` <-> `b` and also for `A` <-> `b` and `B` <-> `a`, because only the link count is verified.

I think we need to make a list before and after backup and check that the right files are linked together.

Do you agree?

src/restic/node.go
```diff
@@ -191,7 +191,17 @@ func (node Node) createDirAt(path string) error {
 	return nil
 }
 
-func (node Node) createFileAt(path string, repo Repository) error {
+func (node Node) createFileAt(path string, repo Repository, idx *HardlinkIndex) error {
+	if node.Links > 1 {
+		if idx.Has(node.Inode, node.Device) {
```
fd0: You can combine the two `if`.

fd0 (changes requested):
So, next round of comments is done. Just a few minor things, mainly the tests needs an additional check :)

src/cmds/restic/integration_helpers_unix_test.go
jgfrm: done

src/restic/node.go

src/cmds/restic/integration_test.go
jgfrm: done

fd0 (approved):
Looks good, thanks a lot for your work!

fd0:
Merged in 6300c8df56ff386c831eff331296b175583c9595, congrats :)

src/restic/node.go
```diff
@@ -191,7 +191,15 @@ func (node Node) createDirAt(path string) error {
 	return nil
 }
 
-func (node Node) createFileAt(path string, repo Repository) error {
+func (node Node) createFileAt(path string, repo Repository, idx *HardlinkIndex) error {
+	if node.Links > 1 && idx.Has(node.Inode, node.Device) {
+		err := fs.Link(idx.GetFilename(node.Inode, node.Device), path)
```
ThomasWaldmann: if you do a partial extract and the "hardlink master" was not extracted, you can not link to it.

fd0: Indeed, but then the file is extracted normally, because `idx` does not contain the other file in this case.

ThomasWaldmann: ah, ok. and the current item not only has the reference (by name) back to the "hardlink master", but also the list of chunks, so it can be normally extracted?

fd0: Exactly.

PR #766, opened 2017-02-02
Support for xattr/acl (only restore command)

jgfrm:


fd0:
Please leave a comment when you're done and I can review the changes :)

jgfrm:
Yes, you can review the changes.
The code does not include support for fuse, but I consider this as a separate issue.

fd0:
Fuse is indeed a separate issue, let's get save/restore done first :)

src/restic/node.go
```diff
@@ -34,6 +41,7 @@ type Node struct {
 	Size       uint64      `json:"size,omitempty"`
 	Links      uint64      `json:"links,omitempty"`
 	LinkTarget string      `json:"linktarget,omitempty"`
+	Xattrs     []Xattr     `json:"xattrstore,omitempty"`
```
fd0: I think the name `xattrstore` for the JSON structure is not a good fit. What about `extended_attributes`? And use `ExtendedAttributes` for the structure variable name, what do you think?

src/restic/node.go
```diff
@@ -14,10 +14,17 @@ import (
 
 	"runtime"
 
+	"bytes"
 	"restic/debug"
 	"restic/fs"
 )
 
+// Xattr is a tuple storing the xattr name and value.
+type Xattr struct {
```
fd0: This name is not a good fit, what about `ExtendedAttribute`? Or even just `Attribute`?

src/restic/node.go
```diff
@@ -14,10 +14,17 @@ import (
...
+	XattrName  string `json:"xattrname"`
```
fd0: No need to repeat yourself, use `Name` (`json:"name"`) and `Value` (`json:"value")`.

This is a lot easier to read in the code:
```go
attr.XattrnNme := "foo"
```
versus
```go
attr.Name := foo
```

src/restic/node.go
```diff
@@ -162,6 +170,26 @@ func (node Node) restoreMetadata(path string) error {
 		}
 	}
 
+	switch runtime.GOOS {
+	case "windows", "openbsd":
```
fd0: Why the switch here? For the Node stuff, we're using different files to implement functions differently based on the OS (you've already discovered that), so you can just call `restoreAcls()` even on Windows and this switch is not needed (because `Setxattr()` for Windows is just a no-op).

src/restic/node.go
```diff
@@ -162,6 +170,26 @@ func (node Node) restoreMetadata(path string) error {
...
+	err = node.restoreAcls(path)
+	if err != nil {
+		debug.Log("error restoring acls for %v: %v", path, err)
+		return err
+	}
+
+	return nil
+}
+
+func (node Node) restoreAcls(path string) error {
```
fd0: This function name is not a good fit, because it restores more than just ACL (think SELinux domain markers or other stuff). What about `restoreExtendedAttributes()` or `restoreAttributes()` (based on how you name the struct above, `Attribute` or `ExtendedAttribute`, please make it consistent)?

src/restic/node.go
```diff
@@ -378,7 +409,48 @@ func (node Node) sameContent(other Node) bool {
 			return false
 		}
 	}
+	return true
+}
 
+func (node Node) sameAcl(other Node) bool {
```
fd0: Same here, a better name would be `sameExtendedAttributes` or `sameAttributes`.

src/restic/node.go
```diff
@@ -378,7 +409,48 @@ func (node Node) sameContent(other Node) bool {
...
+	if node.Xattrs == nil {
```
fd0: This is not a good test. When one node has the attribute set to nil, and the other to the empty slice (which is `!= nil`), it is considered differently. Just use the length test in line 422 below, `len(node.Attributes) != len(other.Attributes)`.

src/restic/node.go
```diff
@@ -378,7 +409,48 @@ func (node Node) sameContent(other Node) bool {
...
+		return other.Xattrs == nil
+	}
+	if other.Xattrs == nil {
+		return false
+	}
+	if len(node.Xattrs) != len(other.Xattrs) {
+		return false
+	}
+	for _, thisAttr := range node.Xattrs {
```
fd0: Split this into its own function that compares two `[]Attribute` slices by building a `map[Attribute]struct{}` and comparing those, see https://github.com/restic/restic/blob/master/src/restic/idset.go#L47 for a sample on the `IDSet` type (which is just a `[]ID`).

src/restic/node.go
```diff
@@ -496,9 +568,35 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 		err = errors.Errorf("invalid node type %q", node.Type)
 	}
 
+	switch runtime.GOOS {
+	case "windows", "openbsd":
```
fd0: Same as above, you don't need this switch. `fillAcls()` will call functions that are no-ops on unsupported OS.

src/restic/node.go
```diff
@@ -496,9 +568,35 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
...
+		return err
+	}
+	if err = node.fillAcls(path); err != nil {
+		return err
+	}
+
 	return err
 }
 
+func (node *Node) fillAcls(path string) error {
```
fd0: Same as above, the function name `fillExtendedAttributes()` is probably better suited.

src/restic/node.go
```diff
@@ -496,9 +568,35 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
...
+	xattrs, e := Listxattr(path)
+	// ignore paths for which no acl can be obtained silently
+	if e == nil {
```
fd0: Why is this error ignored? When the attributes for a file cannot be listed, isn't that an error which should at least be displayed to the user?

src/restic/node.go
```diff
@@ -496,9 +568,35 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
...
+		node.Xattrs = make([]Xattr, len(xattrs))
+		for i, attr := range xattrs {
+			attrVal, err := Getxattr(path, attr)
+			if err != nil {
+				errors.Errorf("can not obtain extended attribute %v for %v:\n", attr, path)
```
fd0: This line makes no sense, I guess you'd like to remove the next line and replace this one with:
```go
return errors.Errorf("can not obtain extended attribute %v for %v:\n", attr, path)
```

fd0 (changes requested):
So, first round of comments done. Thanks for your work!

src/restic/node.go

src/restic/node.go

src/restic/node.go

src/restic/node.go
jgfrm: removed

src/restic/node.go

src/restic/node.go

src/restic/node.go

src/restic/node.go
jgfrm: I do not precisely underdstand what you mean here

src/restic/node.go

src/restic/node.go

jgfrm:
I processed all comments except this one:

> Split this into its own function that compares two []Attribute slices by building a map[Attribute]struct{} and comparing those, see https://github.com/restic/restic/blob/master/src/restic/idset.go#L47 for a sample on the IDSet type (which is just a []ID).

It is entirely clear to me what you mean here.

src/restic/node.go
```diff
@@ -12,31 +12,37 @@ import (
...
-	"runtime"
-
+	"bytes"
 	"restic/debug"
 	"restic/fs"
 )
 
+// ExtendedAttribute is a tuple storing the xattr name and value.
+type ExtendedAttribute struct {
+	XattrName  string `json:"name"`
```
fd0: Rename the fields to `Name` and `Value`.

src/restic/node.go
```diff
@@ -12,31 +12,37 @@ import (
...
+	ChangeTime time.Time           `json:"ctime,omitempty"`
+	UID        uint32              `json:"uid"`
+	GID        uint32              `json:"gid"`
+	User       string              `json:"user,omitempty"`
+	Group      string              `json:"group,omitempty"`
+	Inode      uint64              `json:"inode,omitempty"`
+	Size       uint64              `json:"size,omitempty"`
+	Links      uint64              `json:"links,omitempty"`
+	LinkTarget string              `json:"linktarget,omitempty"`
+	Xattrs     []ExtendedAttribute `json:"extended_attributes,omitempty"`
```
fd0: Rename this field to `ExtendedAttributes`. The JSON name is fine.

fd0 (changes requested):

fd0:
For the comparison function, I was thinking about something like this: https://gist.github.com/fd0/32e52723ac791043c9553b2b1e6c5f4c

src/restic/node.go

src/restic/node.go

jgfrm:
thanks for the Equal stuff. Included.

src/restic/node.go
```diff
@@ -493,12 +559,35 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
...
+
+	if err = node.fillExtendedAttributes(path); err != nil {
+		return err
 	}
 
 	return err
 }
 
+func (node *Node) fillExtendedAttributes(path string) error {
+	if node.Type == "symlink" && node.LinkTarget == "invalid" {
```
fd0: What is `LinkTarget == "invalid"` used for?

fd0 (changes requested):

src/restic/node.go
jgfrm: The extended attribute library tries to follow the link. If the link does not exist, an error is raised.
Actually, node_test, TestNodeRestoreAt fails then for LinkTarget: "invalid.

fd0: I've deliberately included a symlink node with a non-existing target in the tests for the `RestoreAt()` function. restic should be able to restore even those symlinks which do not point to a valid target.

Can symlinks have their own extended attributes? According to the tests that I just did locally (on Linux), they cannot. So restic does not need to save extended attributes for symlinks, and it must not try to restore the extended attributes for them.

I think that's a bug in the code you contributed. What do you think?

src/restic/node.go
```diff
@@ -227,10 +249,6 @@ func (node Node) createFileAt(path string, repo Repository) error {
 }
 
 func (node Node) createSymlinkAt(path string) error {
-	// Windows does not allow non-admins to create soft links.
```
fd0: I think you accidentally removed this check for windows here, could you please put it back in?

src/restic/node.go
jgfrm: I can create softlinks with acls (debian jessie):
```
touch /tmp/test

ln -s /tmp/test /tmp/softlink-with-acl

setfacl /tmp/softlink-with-acl

getfacl /tmp/softlink-with-acl
Removing leading '/' from absolute path names
# file: tmp/softlink-with-acl
# owner: root
# group: root
user::rw-
group::r--
other::r--

setfacl -m u:nobody:r /tmp/softlink-with-acl

getfacl /tmp/softlink-with-acl
getfacl: Removing leading '/' from absolute path names
# file: tmp/softlink-with-acl
# owner: root
# group: root
user::rw-
user:nobody:r--
group::r--
mask::r--
other::r--

setfattr -n user.comment -v "this is a comment" /tmp/softlink-with-acl
getfattr -d /tmp/softlink-with-acl
getfattr: Removing leading '/' from absolute path names
# file: tmp/softlink-with-acl
user.comment="this is a comment"
```

So you can set acls and extended attributes on softlinks.
I do not know if it is a bug of the xattr lib I used (the other library has problems on OS X, that's why I used this one). Perhaps the syscall returns an error if the link can not followed.

fd0: That's not correct, the ACL and the comment are not set for the symlink, but for the file the symlink points to:
```
# getfattr -d softlink-with-acl                                    
# file: softlink-with-acl
user.comment="this is a comment"

getfattr -d test
# file: test
user.comment="this is a comment"
```

src/restic/node.go
jgfrm: restored

jgfrm:
btw: how do you run locally test for all platforms (like appveyor is doing)?

fd0:
You can use `vagrant`, there's a config file in the root directory of the project, but that's not so easy to setup...

fd0 (changes requested):
So, one minor detail to go :)

Thanks a lot for your work!

src/restic/node.go
```diff
@@ -566,6 +566,10 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	default:
 		return errors.Errorf("invalid node type %q", node.Type)
 	}
+	
+	if err {
```
fd0: This doesn't compile. Modify the code for the `symlink` case like this:
```go
case "symlink":
    node.LinkTarget, err = fs.Readlink(path)
    if err != nil {
        return err
    }
```

fd0 (changes requested):

jgfrm:
Any progress on this?

fd0:
Not yet, sorry, I was away at a conference the last couple of days. I'll look into this as soon as I find some time.

fd0 (approved):

fd0:
Merged, thank you very much!

PR #771, opened 2017-02-03
Return an error if password is not set for stdin

### LINKED ISSUE #770: Disable password reading when backup is used with `--stdin`
## Output of `restic version`

restic 0.4.0 (v0.4.0-4-gf5faff9)
compiled with go1.8rc2 on linux/amd64

## Expected behavior

restic shows password prompt and upon entering the password backups the contents of the standard input.

## Actual behavior

No password prompt; instead restic  prints an error:

`unable to open repo: wrong password or no key found`

```
$ restic init -r /tmp/test
$ echo hello | restic backup --stdin -r /tmp/test
unable to open repo: wrong password or no key found
```
It seems that restic tries to use `hello` as a password. It can be verified by using the correct password for the repository in the `echo` command.

fd0:
Closes #770

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -235,6 +235,10 @@ func readBackupFromStdin(opts BackupOptions, gopts GlobalOptions, args []string)
 		return errors.Fatalf("when reading from stdin, no additional files can be specified")
 	}
 
+	if gopts.password == "" && gopts.PasswordFile == "" {
+		return errors.Fatal("Unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
+	}
+
 	repo, err := OpenRepository(gopts)
 	if err != nil {
 		return err


src/cmds/restic/cmd_backup.go
```diff
@@ -235,6 +235,10 @@ func readBackupFromStdin(opts BackupOptions, gopts GlobalOptions, args []string)
 		return errors.Fatalf("when reading from stdin, no additional files can be specified")
 	}
 
+	if gopts.password == "" && gopts.PasswordFile == "" {
+		return errors.Fatal("Unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
```
opennota: The error messages are inconsistent: the previous `errors.Fatalf` message starts with a lower-case letter, this one starts with a capital letter.

fd0: Oh, you're right. Thanks!

Issue #769, opened 2017-02-03
Allow --files-from to take a dash (-) for stdin


opennota:
The `backup` command currently does not recognize `-` as a synonym for stdin:

```
$ find -name '*.go' | restic backup --files-from -
open -: no such file or directory
```

Whereas one can use `/dev/stdin` for that, I would like restic to recognize the dash, because `-` as a filename to mean stdin is a convention used by a great many programs.

If both `--files-from -` and `--stdin` are used, restic should of course print a warning.

fd0:
Thanks for raising this as a different issue. If someone starts working on this, please leave a comment. Also if you need guidance on what to do.

barthr:
I would like to start work on this tomorrow. I think I already know what needs to be done. Will PR tomorrow

opennota:
@barthr OK. I'll stop working on it, then.

barthr:
See PR: https://github.com/restic/restic/pull/772

barthr:
See PR: #772

barthr:
See PR: #772 for discussion

opennota:
Also PR #773

PR #772, opened 2017-02-04
Allow --files-from to take a dash (-) for stdin #769

barthr:
See code, added errors when both --stdin and --files from - are used together.


barthr:
I'm not entirely sure what you mean. Could you elaborate more on that?

barthr:
Do you mean for example that if you have the command:
```
$ find -name '*.go' | restic backup --files-from -
open -: no such file or directory
``` 
That the `-` automatically gets converted to os.Stdin?

opennota:
I thought it would be something like this: https://github.com/restic/restic/pull/773/files

barthr:
I interpreted it the wrong way, excuses. I will close my PR.

fd0:
Thanks for trying :)

PR #773, opened 2017-02-04
Allow --files-from to take a dash for stdin (fixes #769)

### LINKED ISSUE #769: Allow --files-from to take a dash (-) for stdin

ghost:
Closes #769

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"restic"
@@ -28,6 +29,10 @@ The "backup" command creates a new snapshot and saves the files and directories
 given as the arguments.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if backupOptions.Stdin && backupOptions.FilesFrom == "-" {
+			return errors.Fatal("cannot use both `--stdin` and `--files-from -`")
+		}
+
 		if backupOptions.Stdin {
 			return readBackupFromStdin(backupOptions, globalOptions, args)
 		}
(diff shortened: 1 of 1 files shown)


src/cmds/restic/cmd_backup.go
```diff
@@ -293,6 +303,10 @@ func readLinesFromFile(filename string) ([]string, error) {
 }
 
 func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
+	if opts.FilesFrom == "-" && gopts.password == "" && gopts.PasswordFile == "" && os.Getenv("RESTIC_PASSWORD") == "" {
```
fd0: It's enough to test for `gopts.password == ""`, that field is filled with the contents of the environment variable on startup (here: https://github.com/restic/restic/blob/master/src/cmds/restic/global.go#L46-L49), so you can remove the test for `os.Getenv(...) == ''`.

fd0 (changes requested):
ok, looks good so far, just remove the one condition from the `if`. Thanks for your contribution!

opennota:
Updated.

fd0 (approved):

fd0:
Thanks, merged!

src/cmds/restic/cmd_backup.go
```diff
@@ -28,6 +29,10 @@ The "backup" command creates a new snapshot and saves the files and directories
 given as the arguments.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if backupOptions.Stdin && backupOptions.FilesFrom == "-" {
```
gerardbosch: Hi there!
Question: What about `backupOptions.FilesFrom == "/dev/stdin"`? Shouldn't it be checked some way?
Cheers!

src/cmds/restic/cmd_backup.go
```diff
@@ -293,6 +303,10 @@ func readLinesFromFile(filename string) ([]string, error) {
 }
 
 func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
+	if opts.FilesFrom == "-" && gopts.password == "" && gopts.PasswordFile == "" {
```
gerardbosch: Does the following error should only raise when files to back up come from stdin (-)? I am very new to this code and probably I am missing things, but... I ask just in case. Thx!

fd0: Try it with the latest version, we've recently changed this part of the code. if you discover anything odd with it, please create a new issue! Comments to old pull requests tend to get lost :)

src/cmds/restic/cmd_backup.go
fd0: No, because that'll work out of the box. Restic opens `/dev/stdin` as a regular file and reads from it, nothing special here.

gerardbosch:
Oh, thanks, you're right. I'm rather new to this workflow. Thanks for pointing that :)

src/cmds/restic/cmd_backup.go
gerardbosch: Oh, thanks, you're right. I'm rather new to this workflow. Thanks for pointing that :)

src/cmds/restic/cmd_backup.go
gerardbosch: So I assume it is still valid to call restic with`--stdin` and `--files-from /dev/stdin`. Is it right?

fd0: I haven't tried it, but I think that will work just fine.

PR #793, opened 2017-02-10
backup: Add `--hostname` parameter

### LINKED ISSUE #788: Add --host flag for backup command
Please add `--host` flag to `backup` command to be used with `--stdin` flag. 

This is useful when pulling/pushing files for backup from other hosts. See workarounds mentioned in #784. If a backup from another host was made without this flag, then `forget` command can't be used reliably.

~~~[There's only one place where host is set](https://github.com/restic/restic/blob/master/src/cmds/restic/cmd_backup.go#L379), so this shouldn't take too much effort.~~~

And no, unfortunately one couldn't set a hostname with an envvar. [Go gets it directly from the kernel at least in Linux.](https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/os/sys_linux.go#L9)

fd0:
Closes #788

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -51,6 +51,7 @@ type BackupOptions struct {
 	Stdin          bool
 	StdinFilename  string
 	Tags           []string
+	Hostname       string
 	FilesFrom      string
 }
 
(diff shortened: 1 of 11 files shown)


src/cmds/restic/cmd_backup.go
```diff
@@ -264,7 +272,7 @@ func readBackupFromStdin(opts BackupOptions, gopts GlobalOptions, args []string)
 		return err
 	}
```
sanmai: `os.Hostname` does quite a bit of stuff internally... Not like we're going to die if `os.Hostname` spends extra 60µs doing its job, but what if we only load hostname if empty here?

sanmai: if opts.Hostname == "" {
		hostname, err := os.Hostname()
		if err == nil {
			opts.Hostname = hostname
		}
	}

fd0: I've actually thought about doing it this way, but the problem here is that during `init()` the flags are just registered, nothing is parsed yet. So `opts.Hostname` will always be empty.

I think we can leave it this way, having the hostname read at startup won't be noticeable. But good point (and good proposition)!

PR #799, opened 2017-02-11
Hardlinks fuse

jgfrm:
Report the right number of hardlinks for directories in fuse mounts


jgfrm:
How do we solve this?

fd0:
Uhm, what code was added in this PR that was not in the code that I merged yesterday?

Besides, I added a very small commit on top of your code, so that the number of hardlinks is correctly displayed in the fuse mount: https://github.com/restic/restic/commit/6275d69a3612c1fd8c14cb099e67ddd140081af4

As far as I can see, this already works, this is in the fuse mount for two archived hard linked files `test` and `test-hard`:
```
$  stat test test-hard 
  File: test
  Size: 4         	Blocks: 1          IO Block: 512    regular file
Device: 2ah/42d	Inode: 8391094     Links: 2
Access: (0644/-rw-r--r--)  Uid: ( 1000/     fd0)   Gid: (  100/   users)
Access: 2017-02-10 21:14:04.554814170 +0100
Modify: 2017-02-10 21:14:04.554814170 +0100
Change: 2017-02-10 21:14:07.018138855 +0100
 Birth: -

  File: test-hard
  Size: 4         	Blocks: 1          IO Block: 512    regular file
Device: 2ah/42d	Inode: 8391094     Links: 2
Access: (0644/-rw-r--r--)  Uid: ( 1000/     fd0)   Gid: (  100/   users)
Access: 2017-02-10 21:14:04.554814170 +0100
Modify: 2017-02-10 21:14:04.554814170 +0100
Change: 2017-02-10 21:14:07.018138855 +0100
 Birth: -
```

jgfrm:
I have added the correct report of the number of hardlinks for directories in case of fuse mounts.
The currect number is 2 (. plus the dir itself) plus the number of subdir (as .. is implement as a hard link to the parent).

fd0:
As far as I know that's not always the case: First, there are only very few file systems that even allow hard linked directories (problems outlined here: http://askubuntu.com/a/525129), second the link count for directories returned from the `stat` syscall varies by file system, so I think we can just leave it at one.

Is there anything else in this code that needs to be merged? I must admit I don't have an overview any more...

jgfrm:
No, it was only the dir part.
Normally, you can not use hard links for reasons as described by the article you refering to.
However, usually the filesystem is implemented such that the number of hardlinks of a directory equals two plus the number of subdirs contained by that dir. Two because because the dir name and ```.``` is a hardlink and ```..``` in all the subdirs are also hardlinks. Just use ```ls -al``` and you will see

fd0:
So, do you think we really need to implement this for the fuse mount? Or is this just for completeness' sake?

jgfrm:
Well, it is up to you.
While comparing the fuse mount with my test backup dir, I was confused that the link numbers in fuse where not matching the link numbers in my test dir.
Actually, for the same reason, it is not entirely clear to me why ```.``` and ```..``` are not present.
Personally, I have no problems with that but users might be confused.

fd0:
Huh, you're right! I didn't notice that until now!

Ok, you convinced me, I'll extract your work on the fuse directory implementation and merge that. I'll open a new PR just with the commits shortly, and I'll close this PR for now.

fd0:
I've extracted your code to PR #800, please have a short look if everything is in there.

pvgoran:
I think `.` an `..` should definitely be there, unless adding them is a major issue.

fd0:
I'll add an issue to track this.

fd0:
See #802.

PR #800, opened 2017-02-11
fuse: correct nlink count for directories

fd0:
Extracted code from #799


jgfrm:
looks fine with me

fd0:
Okay then!

PR #806, opened 2017-02-13
Fix inconsistencies in CLI

oysols:
I did a small review of the help text, and also found a bug in the number of arguments accepted for the key command.

The changes are small, so I didn't create a separate issue. Let me know if I should.


fd0 (approved):
Awesome, thanks a lot for taking the time to do this!

PR #817, opened 2017-02-21
Add `--prune` switch to `forget`

fd0:
This allows automatically running `prune` when snapshots have been removed by `forget`.


zcalusic:
This is a very nice feature! 👍

Niondir:
Is it faster than running `prune` separately after `forget`?

zcalusic:
No, the prune step itself takes the same amount the time, whichever way it was called. The optimization comes from the fact that if forget hasn't removed any snapshots, then prune step can be completely avoided, if called this way.

PR #847, opened 2017-03-04
Display the proper amount of bytes we will be pruning from the repo.

### LINKED ISSUE #846: Nit: output of prune does not line up with what it does.
## Output of `restic version`
restic 0.4.0 (v0.4.0-86-g90bc187-dirty)
compiled with go1.8 on linux/amd64
(note my version is slightly different, I added a progress bad for the repacking; i'll make a patch for that later)

## Expected behavior
During the phase were it find the duplicate blobs, it detects 11GB worth of them. (Very likely, I had a large number of OOMing backups as restic is really really memory hungry. Not the point however).
Then, on repack and remove line, it claims it will only recover 200MB !?!?
Note however that is DID free up over 11GB of data.

I would have expected it to tell me the remove and repack will gain me 11.2GB ...

## Actual behavior
[/share/Backups/restic] # du -sh *
4.0K	config
65G	data
100M	index
12K	keys
4.0K	locks
72K	snapshots
4.0K	tmp

\# time /usr/local/sbin/restic -r rest:http://qnap.********:8000 prune
counting files in repo
building new index for repo
[2:52] 100.00%  13654 / 13654 packs
repository contains 13654 packs (865740 blobs) with 64.833 GiB bytes
processed 865740 blobs: 14571 duplicate blobs, 11.038 GiB duplicate
load all snapshots
find data that is still in use for 17 snapshots
[18:22] 100.00%  17 / 17 snapshots
found 835833 of 865740 data blobs still in use, removing 29907 blobs
will delete 0 packs and rewrite 2800 packs, this frees 199.554 MiB
[22:04] 100.00%  2800 / 2800 packs rewriten
creating new index
[2:08] 100.00%  11512 / 11512 packs
saved new index as 796a7b7f
done

real	45m53.795s
user	7m19.354s
sys	1m46.655s

[/share/Backups/restic] # du -sh *
4.0K	config
54G	data
96M	index
12K	keys
4.0K	locks
72K	snapshots
4.0K	tmp

Em. get yourself a large repo of 8 machines running the same distro (lots of binary overlap) and have one of them OOM at ~11% a bunch of times when running restic :)

middelink:
Closes #846

### DIFF
--- a/src/cmds/restic/cmd_prune.go
+++ b/src/cmds/restic/cmd_prune.go
@@ -184,7 +184,7 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 		}
 	}
 
-	removeBytes := 0
+	removeBytes := duplicateBytes
 
 	// find packs that are unneeded
 	removePacks := restic.NewIDSet()


fd0:
FYI: you don't need to add the PR number to the comment, the PR will automatically be closed.

fd0 (approved):
Thanks for taking the time!

PR #852, opened 2017-03-05
Improve output formatting and increase testability of cmd_snapshots.

middelink:
- Small changes to cmd_snapshot to use the global stdout, so it can be used from the testing environment.
- Change width of host and tag column in its output dynamically.


middelink:
I'm pulling this request, not happy how the refactoring went with the integration test I needed this for. I'll submit 2 other PRs

PR #853, opened 2017-03-05
restic backup --tag integration test

middelink:
Make sure the backup command does indeed add the required tags when asked.

### DIFF
--- a/src/cmds/restic/integration_test.go
+++ b/src/cmds/restic/integration_test.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"crypto/rand"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


src/cmds/restic/integration_test.go
```diff
@@ -160,6 +161,33 @@ func testRunFind(t testing.TB, gopts GlobalOptions, pattern string) []string {
...
+		globalOptions.stdout = os.Stdout
+		globalOptions.JSON = gopts.JSON
+	}()
+
+	opts := SnapshotOptions{}
+
+	OK(t, runSnapshots(opts, globalOptions, []string{}))
+
+	snapshots := []Snapshot{}
+	OK(t, json.Unmarshal(buf.Bytes(), &snapshots))
```
fd0: Nice idea, parsing the JSON. I like it :)

src/cmds/restic/integration_test.go
```diff
@@ -602,6 +630,31 @@ func TestIncrementalBackup(t *testing.T) {
...
+	withTestEnvironment(t, func(env *testEnvironment, gopts GlobalOptions) {
+		datafile := filepath.Join("testdata", "backup-data.tar.gz")
+		testRunInit(t, gopts)
+		SetupTarTestFixture(t, env.testdata, datafile)
+
+		opts := BackupOptions{}
+
+		testRunBackup(t, []string{env.testdata}, opts, gopts)
+		testRunCheck(t, gopts)
+		newest, _ := testRunSnapshots(t, gopts)
```
fd0: Why do you return the snapshot map if you're not going to use it here (the second return value of `testRunSnapshots()`)?

fd0 (changes requested):
Looks already very good, just a minor thing :)

src/cmds/restic/integration_test.go
middelink: I wanted to be generous to other potential testcases which might have a need to get all snapshots. I can change it's signature and leave it to others to add it back if you like?

fd0: Ok, good point, on a second thought that can be quite useful. Looked like old code for me when I first read it, so leave it in.

fd0 (approved):
Thanks for your work!

PR #855, opened 2017-03-05
Add `tag` command to restic cli to manipulate tags on existing snapshots.

### LINKED ISSUE #851: [FR] Add a new command to manipulate tags on existing snapshots.
Sometimes it is handy to manipulate snapshots based on their tags, however to make that universal useful we need a way to change them on existing snapshots.

Example I was thinking about was progressive insight into snapshots are managed by some enterprising sysadmin. Without the possibility to change tags retroactively, they would have to make exceptions to their newly thought of schema, till the old snapshots are forgotten.

Unavoidable implementation detail, due to the tags changing, and therefore the sha256 hash, the snapshots will get a new ID assigned. All other data stays the same however.

Possible cmdline:
```console
$ ../../../bin/restic -r rest:http://qnap.********:8000 tag --help
The "tag" command allows you to modify tags on existing snapshots.

You can either set/replace the entire set of tags on a snapshot, or
add tags to/remove tags from the existing set.

When no snapshot-ID is given, all snapshots matching the host, tag and path filter criteria are modified.

Usage:
  restic tag [flags] [snapshot-ID ...]

Flags:
      --add tag       tag which will be added to the existing tags, can be given multiple times
  -H, --host string   only consider snapshots for this host, when no snapshot ID is given
      --path path     only consider snapshots which include this (absolute) path, when no snapshot-ID is given
      --remove tag    tag which will be removed from the existing tags, can be given multiple times
      --set tag       tag which will replace the existing tags, can be given multiple times
      --tag tag       only consider snapshots which include this tag, when no snapshot-ID is given


Global Flags:
      --json                   set output mode to JSON for commands that support it
      --no-lock                do not lock the repo, this allows some operations on read-only repos
  -p, --password-file string   read the repository password from a file
  -q, --quiet                  do not output comprehensive progress report
  -r, --repo string            repository to backup to or restore from (default: $RESTIC_REPOSITORY)
```

middelink:
Closes #851

### DIFF
--- a/src/cmds/restic/cmd_tag.go
+++ b/src/cmds/restic/cmd_tag.go
@@ -1,0 +1,189 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+
+	"restic"
+	"restic/debug"
+	"restic/errors"
+	"restic/repository"
+)
+
+var cmdTag = &cobra.Command{
+	Use:   "tag [flags] [snapshot-ID ...]",
+	Short: "modifies tags on snapshots",
+	Long: `
+The "tag" command allows you to modify tags on exiting snapshots.
+
+You can either set/replace the entire set of tags on a snapshot, or
+add tags to/remove tags from the existing set.
+
+When no snapshot-ID is given, all snapshots matching the host, tag and path filter criteria are modified.
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runTag(tagOptions, globalOptions, args)
+	},
+}
+
+// TagOptions bundles all options for the 'tag' command.
+type TagOptions struct {
+	Host       string
+	Paths      []string
+	Tags       []string
+	SetTags    []string
+	AddTags    []string
+	RemoveTags []string
+}
+
+var tagOptions TagOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdTag)
+
+	tagFlags := cmdTag.Flags()
+	tagFlags.StringSliceVar(&tagOptions.SetTags, "set", nil, "`tag` which will replace the existing tags (can be given multiple times)")
+	tagFlags.StringSliceVar(&tagOptions.AddTags, "add", nil, "`tag` which will be added to the existing tags (can be given multiple times)")
+	tagFlags.StringSliceVar(&tagOptions.RemoveTags, "remove", nil, "`tag` which will be removed from the existing tags (can be given multiple times)")
+
+	tagFlags.StringVarP(&tagOptions.Host, "host", "H", "", `only consider snapshots for this host, when no snapshot ID is given`)
+	tagFlags.StringSliceVar(&tagOptions.Tags, "tag", nil, "only consider snapshots which include this `tag`, when no snapshot-ID is given")
+	tagFlags.StringSliceVar(&tagOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")
+}
+
+func changeTags(repo *repository.Repository, snapshotID restic.ID, setTags, addTags, removeTags, tags, paths []string, host string) (bool, error) {
+	var changed bool
+
+	sn, err := restic.LoadSnapshot(repo, snapshotID)
+	if err != nil {
+		return false, err
+	}
+	if (host != "" && host != sn.Hostname) || !sn.HasTags(tags) || !restic.SamePaths(sn.Paths, paths) {
+		return false, nil
+	}
+
+	if len(setTags) != 0 {
+		// Setting the tag to an empty string really means no more tags.
+		if len(setTags) == 1 && setTags[0] == "" {
+			setTags = nil
+		}
+		sn.Tags = setTags
+		changed = true
+	} else {
+		for _, add := range addTags {
+			found := false
+			for _, tag := range sn.Tags {
+				if tag == add {
+					found = true
+					break
+				}
+			}
+			if !found {
+				sn.Tags = append(sn.Tags, add)
+				changed = true
+			}
+		}
+		for _, remove := range removeTags {
+			for i, tag := range sn.Tags {
+				if tag == remove {
+					// https://github.com/golang/go/wiki/SliceTricks
+					sn.Tags[i] = sn.Tags[len(sn.Tags)-1]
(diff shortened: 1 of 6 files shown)


src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+	RemoveTags []string
+}
+
+var tagOptions TagOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdTag)
+
+	tagFlags := cmdTag.Flags()
+	tagFlags.StringSliceVar(&tagOptions.SetTags, "set", nil, "`tag` which will replace the existing tags, can be given multiple times")
```
fd0: Add parentheses around the `can be given multiple times`, that's the style we use for the other options, e.g. here: https://github.com/restic/restic/blob/master/src/cmds/restic/cmd_backup.go#L77

Also for the other options below, obviously.

src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+
+	sn, err := restic.LoadSnapshot(repo, snapshotID)
+	if err != nil {
+		return false, err
+	}
+	if (host != "" && host != sn.Hostname) || !sn.HasTags(tags) || !restic.SamePaths(sn.Paths, paths) {
+		return false, nil
+	}
+
+	if len(setTags) != 0 {
```
fd0: Maybe split the code below into a function on the `Snapshot` struct, in the `restic` package?

src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+		// Setting the tag to an empty string really means no more tags.
+		if len(setTags) == 1 && setTags[0] == "" {
+			setTags = nil
+		}
+		sn.Tags = setTags
+		changed = true
+	} else {
+		for _, add := range addTags {
```
fd0: This is a candidate for a new function, `Snapshot.AddTags([]string)` or so.

src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+					found = true
+					break
+				}
+			}
+			if !found {
+				sn.Tags = append(sn.Tags, add)
+				changed = true
+			}
+		}
+		for _, remove := range removeTags {
```
fd0: Same here, `Snapshot.RemoveTags([]string)`

src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	repo, err := OpenRepository(gopts)
+	if err != nil {
+		return err
+	}
+
+	if !gopts.NoLock {
+		Verbosef("Create exclusive lock for repository\n")
+		lock, err := lockRepoExclusive(repo)
```
fd0: I think a non-exclusive lock is sufficient, what do you think?

src/cmds/restic/cmd_tag.go
```diff
@@ -0,0 +1,189 @@
...
+		defer unlockRepo(lock)
+		if err != nil {
+			return err
+		}
+	}
+
+	var ids restic.IDs
+	if len(args) != 0 {
+		// When explit snapshot-IDs are given, the filtering does not matter anymore.
+		opts.Host = ""
```
fd0: I feel it may be better to first create a list of snapshot IDs that are to be modified, either by taking the explicit list of IDs (if given) or by listing all snapshots and then applying the filters. Then iterate through that list and modify the tags. This way, `changeTags()` won't need the filter criteria at all.

fd0 (changes requested):

fd0:
Ah, and here you can drop the commits from the other PRs.

src/cmds/restic/cmd_tag.go
middelink: Wait, that is what I do, no? Build a list of snapshotIDs. The problem is, if I want to do filtering I need to Load the snapshot which, sure, I can do in `runTag()`, but then I need to lift the Loading code also into the code path for the explicit ids. And I really really would like to keep the ids.Uniq() there.

src/cmds/restic/cmd_tag.go
middelink: Done.

src/cmds/restic/cmd_tag.go
middelink: Hmmm. will multiple non-exclusive Locks serialize each other? Because otherwise there will be a race with say `restic snapshots`, or `restic forget` or.. well. all functions which load the list of snapshotIDs to work on.

fd0: Good point, I retract my comment :)

src/cmds/restic/cmd_tag.go
fd0: I see what you mean. Leave it ;)

fd0:
Ok, so the code already looks good, we can leave it this way. Two more things need to be done:
 * Add the new field (and a short paragraph which explains it) to `doc/Design.md`
 * Describe the new command in a section in `doc/Manual.md`

fd0 (approved):

fd0:
When you rebase to the current master branch and force-push the commits, git should drop the duplicate commits from the other PRs...

fd0:
Small nitpick: Use present tense for commit descriptions and subject lines, i.e. "add foobar" instead of "added foobar"

middelink:
Well, the git rebase did not make things better, now we have 19 commits in the PR. My current understand of git means... hands off till it is submitted :(

fd0:
This should fix it:
```
git reset --hard 3a1ecae0204bc205a0adbd710022237fdb77bd0d
git rebase master
git push -f
```

(I have a backup of your commits locally)

fd0:
Exactly, when you rebase, you're rewriting history (a "non fast-forward commit" in git terms), so in that special case you need `push -f`.

fd0:
Hm, now you cannot use my instructions, so I'll fix it for you before merging.

middelink:
And... shoot. I did not see your comment after I pushed the documentation commit. Can I still run that rebase now? Or will it completely hose the world?

PR #861, opened 2017-03-06
Refactor cmd_forget

### LINKED ISSUE #841: Limit forget to specific path
## Output of `restic version`
restic 0.4.0
compiled with go1.6.2 on linux/amd64

## Expected behavior
On a given host, it should be possible to keep differing numbers of snapshots for different paths. Eg retain all `/etc`, but only keep a few of `/home`

## Actual behavior
`restic forget` only offers match by hostname and tag

middelink:
Add `--path` for snapshot filtering by path.

Add `--groupby-tags` for grouping on host,tags,dirs instead of host,dirs.

Borrow the snapshot printing from cmd_snapshot.

Closes #841

### DIFF
--- a/src/cmds/restic/cmd_forget.go
+++ b/src/cmds/restic/cmd_forget.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"encoding/hex"
-	"fmt"
-	"io"
+	"encoding/json"
 	"restic"
+	"sort"
 	"strings"
+
+	"restic/errors"
 
 	"github.com/spf13/cobra"
 )
@@ -34,11 +36,13 @@ type ForgetOptions struct {
 
 	KeepTags []string
 
-	Hostname string
-	Tags     []string
+	Host  string
+	Tags  []string
+	Paths []string
 
-	DryRun bool
-	Prune  bool
+	GroupByTags bool
+	DryRun      bool
+	Prune       bool
 }
 
 var forgetOptions ForgetOptions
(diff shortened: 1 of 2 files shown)


fd0:
Hm. Shouldn't `forget` group by `host,paths,tag` by default? isn't that the sensible thing that users would expect?

middelink:
My gut feeling: no. The ability to change tags have just recently been introduced, so ppl are not used to them. I think.

It also depend completely on your tag use-case. For example, I tag my backups with the location the VM runs in [`CH`,`NL`]. So using that in grouping can make sense. However another use-case would be to tag a backup as `super-important` and have a `--keep-tag super-important` as a policy. No grouping wanted here..

src/cmds/restic/cmd_forget.go
```diff
@@ -54,54 +56,17 @@ func init() {
 	f.IntVarP(&forgetOptions.Monthly, "keep-monthly", "m", 0, "keep the last `n` monthly snapshots")
 	f.IntVarP(&forgetOptions.Yearly, "keep-yearly", "y", 0, "keep the last `n` yearly snapshots")
 
-	f.StringSliceVar(&forgetOptions.KeepTags, "keep-tag", []string{}, "always keep snapshots with this `tag` (can be specified multiple times)")
-	f.StringVar(&forgetOptions.Hostname, "hostname", "", "only forget snapshots for the given hostname")
-	f.StringSliceVar(&forgetOptions.Tags, "tag", []string{}, "only forget snapshots with the `tag` (can be specified multiple times)")
+	f.StringSliceVar(&forgetOptions.KeepTags, "keep-tag", []string{}, "keep snapshots with this `tag` (can be specified multiple times)")
+	f.BoolVarP(&forgetOptions.GroupByTags, "groupby-tags", "G", false, "Group by host,paths,tags instead of just host,paths")
+	// Sadly the commonly used shortcut `H` is already used.
+	f.StringVar(&forgetOptions.Host, "host", "", "only consider snapshots with the given hostname")
```
fd0: This option was changed from `--hostname` to `--host`, this will break people's shell scripts. I see that `--host` is used for all other commands and I like having a consistent UI, can we maybe have an alias so that both work (at least for some time)?

src/cmds/restic/cmd_forget.go
```diff
@@ -54,54 +56,17 @@ func init() {
 	f.IntVarP(&forgetOptions.Monthly, "keep-monthly", "m", 0, "keep the last `n` monthly snapshots")
 	f.IntVarP(&forgetOptions.Yearly, "keep-yearly", "y", 0, "keep the last `n` yearly snapshots")
 
-	f.StringSliceVar(&forgetOptions.KeepTags, "keep-tag", []string{}, "always keep snapshots with this `tag` (can be specified multiple times)")
-	f.StringVar(&forgetOptions.Hostname, "hostname", "", "only forget snapshots for the given hostname")
-	f.StringSliceVar(&forgetOptions.Tags, "tag", []string{}, "only forget snapshots with the `tag` (can be specified multiple times)")
+	f.StringSliceVar(&forgetOptions.KeepTags, "keep-tag", []string{}, "keep snapshots with this `tag` (can be specified multiple times)")
+	f.BoolVarP(&forgetOptions.GroupByTags, "groupby-tags", "G", false, "Group by host,paths,tags instead of just host,paths")
```
fd0: I'd prefer `--group-by-tags`, is that okay for you?

src/cmds/restic/cmd_forget.go
```diff
@@ -114,37 +79,33 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
-			Warnf("could not find a snapshot for ID %q, ignoring\n", s)
-			continue
-		}
-
-		if !opts.DryRun {
-			h := restic.Handle{Type: restic.SnapshotFile, Name: id.String()}
-			err = repo.Backend().Remove(h)
+	// Process all snapshot IDs given as arguments.
+	if len(args) != 0 {
+		for _, s := range args {
```
fd0: This is a candidate for its own function, like `forgetSnapshotIDs()`, which will also shorten the current function a lot.

src/cmds/restic/cmd_forget.go
```diff
@@ -157,51 +118,69 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
+		if opts.Host != "" && sn.Hostname != opts.Host {
 			continue
 		}
 
 		if !sn.HasTags(opts.Tags) {
 			continue
 		}
 
-		k := key{Hostname: sn.Hostname, Dirs: strings.Join(sn.Paths, ":")}
+		if !sn.HasPaths(opts.Paths) {
```
fd0: This is nice code, good work!

src/cmds/restic/cmd_forget.go
```diff
@@ -157,51 +118,69 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
+			continue
+		}
+
+		tags := ""
+		if opts.GroupByTags {
+			tags = strings.Join(sn.Tags, ":")
+		}
+		k := key{Hostname: sn.Hostname, Tags: tags, Paths: strings.Join(sn.Paths, ":")}
```
fd0: This approach has several problems:
 * What happens when the a tag includes `:`?
 * `sn.Tags` may not be sorted, so `["foo", "bar"]` and `["bar", "foo"]` are two different groups

Quick idea for a solution: sort `sn.Tags` then use the JSON representation of the tags for the key. Feels dirty, but should work.

fd0 (changes requested):

src/cmds/restic/cmd_forget.go
middelink: Done.

src/cmds/restic/cmd_forget.go
middelink: Done

src/cmds/restic/cmd_forget.go
middelink: You mean... Move it into the restic library? Not 100% sure what the scope of this your remark is here.

fd0: No, just make a new function in `cmd_forget.go`, so that this function becomes shorter.

src/cmds/restic/cmd_forget.go
middelink: Re 1: Nothing. It is simply a separator for when the key is printed later on. It will look ugly, sure ^^
Re 2: Sorting, ok. Good catch there. I think the Paths need a sort for the same reason, on the same host I can backup /etc,/home and later on /home,/etc. Unless they get sorted elsewhere.

The JSON representation just for tags is overkill IMO. Either struct equivalence works or it doesn't. If it doesn't we need to do something for the whole struct, not just the tags.

src/cmds/restic/cmd_forget.go
middelink: Ah, i see. If you don't mind, I hold off on that one, the (unpublished) changelist for widespread-tags has a much nicer fix for that.

src/cmds/restic/cmd_forget.go
fd0: The paths are sorted somewhere, but I think it makes sense to sort them here again, just to be one the safe side.

The problem with just joining the tags by `:` is that you get the same key for the tags `["host:test", "production"]` and tags `["host", "test:production"]`.

src/cmds/restic/cmd_forget.go
fd0: Ok, even better.

middelink: I presume this PR is now ok? Not sure how I can mark it as ready for review...

src/cmds/restic/cmd_forget.go
middelink: How about I make the separator a `,`? Since that is the character pflags separates slices on, I assume there is no (easy) way to enter it from the cmdline. Nah. take it back. it uses the encoding/csv parser, so it understands about quoting. (`--set 'A,",CH"'`)

More thinking required.

middelink: Fixed it by just wrapping the entire key struct in json. it's about the same amount of work of just doing it for Tags alone, and it gives us the nice benefit we have the struct available later for displaying.

src/cmds/restic/cmd_forget.go
fd0: Just add a comment (to the PR, not the code), then I'll have another look.

PR #866, opened 2017-03-07
Add --tag filtering to every command, where applicable

### LINKED ISSUE #863: Make cli filtering more consistent
Currently few commands accept all 3 filters (`--host`, `--path` and `--tag`) and a lot of other commands accept just one or two. Can we try and make things more consistent? For a casual user consistency rules. They don't want to look at the help each time they need to run a command.

My first attempt at solving this. Not a PR yet (expecting loads of comments ^^ and will have merge conflicts with the 2 other PRs I have inflight):
  https://github.com/middelink/restic/compare/master...middelink:widespread-tags

middelink:
Add helper function `FindFilteredSnapshots()` to handle the bulk of iterating
over the snapshots and perform filtering.

Started using contexts (which means minimal requirement is now Go 1.7)

Fixed long silent runs of `find` by outputting results as we find them.
(We lost the ability to print how many matches we found however.)

Made `mount` more interesting, it now applies filters when filling the
`snapshots` directory.

Closes #863

### DIFF
--- a/src/cmds/restic/cmd_find.go
+++ b/src/cmds/restic/cmd_find.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"path/filepath"
 	"strings"
 	"time"
@@ -28,8 +29,12 @@ repo. `,
 type FindOptions struct {
 	Oldest          string
 	Newest          string
-	Snapshot        string
+	Snapshots       []string
 	CaseInsensitive bool
+	ListLong        bool
+	Host            string
+	Paths           []string
+	Tags            []string
 }
 
 var findOptions FindOptions
@@ -40,19 +45,19 @@ func init() {
 	f := cmdFind.Flags()
 	f.StringVarP(&findOptions.Oldest, "oldest", "o", "", "oldest modification date/time")
 	f.StringVarP(&findOptions.Newest, "newest", "n", "", "newest modification date/time")
-	f.StringVarP(&findOptions.Snapshot, "snapshot", "s", "", "snapshot ID to search in")
+	f.StringSliceVarP(&findOptions.Snapshots, "snapshot", "s", nil, "snapshot `id` to search in (can be given multiple times)")
 	f.BoolVarP(&findOptions.CaseInsensitive, "ignore-case", "i", false, "ignore case for pattern")
+	f.BoolVarP(&findOptions.ListLong, "long", "l", false, "use a long listing format showing size and mode")
+
+	f.StringVarP(&findOptions.Host, "host", "H", "", "only consider snapshots for this `host`, when no snapshot ID is given")
+	f.StringSliceVar(&findOptions.Tags, "tag", nil, "only consider snapshots which include this `tag`, when no snapshot-ID is given")
+	f.StringSliceVar(&findOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")
 }
 
 type findPattern struct {
 	oldest, newest time.Time
 	pattern        string
 	ignoreCase     bool
-}
-
-type findResult struct {
-	node *restic.Node
-	path string
 }
 
 var timeFormats = []string{
@@ -79,14 +84,14 @@ func parseTime(str string) (time.Time, error) {
 	return time.Time{}, errors.Fatalf("unable to parse time: %q", str)
 }
 
-func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, path string) ([]findResult, error) {
+func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, prefix string, snapshotID *string) error {
 	debug.Log("checking tree %v\n", id)
+
 	tree, err := repo.LoadTree(id)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	results := []findResult{}
 	for _, node := range tree.Nodes {
 		debug.Log("  testing entry %q\n", node.Name)
 
@@ -97,7 +102,7 @@ func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, path
 
 		m, err := filepath.Match(pat.pattern, name)
 		if err != nil {
-			return nil, err
+			return err
 		}
 
 		if m {
@@ -112,46 +117,32 @@ func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, path
 				continue
 			}
 
-			results = append(results, findResult{node: node, path: path})
+			if snapshotID != nil {
+				Verbosef("Found matching entries in snapshot %s\n", *snapshotID)
+				snapshotID = nil
+			}
+			Printf(formatNode(prefix, node, findOptions.ListLong) + "\n")
 		} else {
 			debug.Log("    pattern does not match\n")
 		}
 
 		if node.Type == "dir" {
-			subdirResults, err := findInTree(repo, pat, *node.Subtree, filepath.Join(path, node.Name))
-			if err != nil {
-				return nil, err
+			if err := findInTree(repo, pat, *node.Subtree, filepath.Join(prefix, node.Name), snapshotID); err != nil {
+				return err
 			}
-
-			results = append(results, subdirResults...)
 		}
 	}
 
-	return results, nil
+	return nil
 }
 
-func findInSnapshot(repo *repository.Repository, pat findPattern, id restic.ID) error {
-	debug.Log("searching in snapshot %s\n  for entries within [%s %s]", id.Str(), pat.oldest, pat.newest)
+func findInSnapshot(repo *repository.Repository, sn *restic.Snapshot, pat findPattern) error {
+	debug.Log("searching in snapshot %s\n  for entries within [%s %s]", sn.ID(), pat.oldest, pat.newest)
 
-	sn, err := restic.LoadSnapshot(repo, id)
-	if err != nil {
+	snapshotID := sn.ID().Str()
+	if err := findInTree(repo, pat, *sn.Tree, string(filepath.Separator), &snapshotID); err != nil {
 		return err
 	}
-
-	results, err := findInTree(repo, pat, *sn.Tree, string(filepath.Separator))
-	if err != nil {
-		return err
-	}
-
-	if len(results) == 0 {
-		return nil
-	}
-	Verbosef("found %d matching entries in snapshot %s\n", len(results), id)
-	for _, res := range results {
-		res.node.Name = filepath.Join(res.path, res.node.Name)
-		Printf("  %s\n", res.node)
-	}
-
 	return nil
 }
 
@@ -160,21 +151,21 @@ func runFind(opts FindOptions, gopts GlobalOptions, args []string) error {
 		return errors.Fatal("wrong number of arguments")
 	}
 
-	var (
-		err error
-		pat findPattern
-	)
+	var err error
+	pat := findPattern{pattern: args[0]}
+	if opts.CaseInsensitive {
+		pat.pattern = strings.ToLower(pat.pattern)
+		pat.ignoreCase = true
+	}
 
 	if opts.Oldest != "" {
-		pat.oldest, err = parseTime(opts.Oldest)
-		if err != nil {
+		if pat.oldest, err = parseTime(opts.Oldest); err != nil {
 			return err
 		}
 	}
 
 	if opts.Newest != "" {
-		pat.newest, err = parseTime(opts.Newest)
-		if err != nil {
+		if pat.newest, err = parseTime(opts.Newest); err != nil {
 			return err
 		}
 	}
@@ -192,33 +183,14 @@ func runFind(opts FindOptions, gopts GlobalOptions, args []string) error {
 		}
 	}
 
-	err = repo.LoadIndex()
-	if err != nil {
+	if err = repo.LoadIndex(); err != nil {
 		return err
 	}
 
-	pat.pattern = args[0]
-
-	if opts.CaseInsensitive {
-		pat.pattern = strings.ToLower(pat.pattern)
-		pat.ignoreCase = true
-	}
-
-	if opts.Snapshot != "" {
-		snapshotID, err := restic.FindSnapshot(repo, opts.Snapshot)
-		if err != nil {
-			return errors.Fatalf("invalid id %q: %v", args[1], err)
-		}
-
-		return findInSnapshot(repo, pat, snapshotID)
-	}
-
-	done := make(chan struct{})
-	defer close(done)
-	for snapshotID := range repo.List(restic.SnapshotFile, done) {
-		err := findInSnapshot(repo, pat, snapshotID)
-
-		if err != nil {
+	ctx, cancel := context.WithCancel(gopts.ctx)
+	defer cancel()
+	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, opts.Snapshots) {
+		if err = findInSnapshot(repo, sn, pat); err != nil {
 			return err
 		}
 	}
--- a/src/cmds/restic/cmd_find.go
+++ b/src/cmds/restic/cmd_find.go
@@ -24,7 +24,7 @@ repo. `,
 	},
 }
 
-// FindOptions bundle all options for the find command.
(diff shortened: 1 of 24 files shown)


fd0:
Hm. I appreciate the switch to `context.Context`, but doesn't that cause problems with the fuse library (which still imports the old context library)?

src/cmds/restic/cmd_find.go
```diff
@@ -192,33 +178,14 @@ func runFind(opts FindOptions, gopts GlobalOptions, args []string) error {
...
-		return findInSnapshot(repo, pat, snapshotID)
-	}
-
-	done := make(chan struct{})
-	defer close(done)
-	for snapshotID := range repo.List(restic.SnapshotFile, done) {
-		err := findInSnapshot(repo, pat, snapshotID)
-
-		if err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
```
fd0: Use `context.TODO()` instead of `context.Background()`, the latter should only be used in the `main()` (or equivalent). The former is made specifically to allow refactoring a code base piece by piece and easily find the places where a new context has been introduced that'd need to be derived from a parent context.

src/cmds/restic/cmd_find.go
```diff
@@ -192,33 +178,14 @@ func runFind(opts FindOptions, gopts GlobalOptions, args []string) error {
...
+	defer cancel()
+	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, opts.Snapshots) {
```
fd0: Maybe bundle the filter criteria for snapshots into a new struct, e.g.:
```go
type SnapshotFilter struct {
    Host string
    Tags []string
    Paths []string
    IDs []restic.ID
}
```

src/cmds/restic/cmd_forget.go
```diff
@@ -55,8 +55,9 @@ func init() {
 	f.IntVarP(&forgetOptions.Yearly, "keep-yearly", "y", 0, "keep the last `n` yearly snapshots")
 
 	f.StringSliceVar(&forgetOptions.KeepTags, "keep-tag", []string{}, "always keep snapshots with this `tag` (can be specified multiple times)")
-	f.StringVar(&forgetOptions.Hostname, "hostname", "", "only forget snapshots for the given hostname")
-	f.StringSliceVar(&forgetOptions.Tags, "tag", []string{}, "only forget snapshots with the `tag` (can be specified multiple times)")
+	f.StringVar(&forgetOptions.Host, "host", "", "only consider snapshots for this `host`")
```
fd0: Same comment as in the other PR: we're changing from `--hostname` to `--host`, we need to find a workaround for people using `--hostname` in their scripts.

src/cmds/restic/cmd_forget.go
```diff
@@ -114,38 +115,36 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
+				}
+				Verbosef("removed snapshot %v\n", sn.ID())
+			} else {
+				Verbosef("would have removed snapshot %v\n", sn.ID())
 			}
-
-			Verbosef("removed snapshot %v\n", id.Str())
 		} else {
-			Verbosef("would remove snapshot %v\n", id.Str())
+			// Not cool, we lose the benefit of yielding, e.g. only one snapshot in memory.
```
fd0: What is this comment about? What should it tell future readers of this source code?

fd0 (changes requested):
First round of comments, gotta go now, I'll finish this review later!

src/cmds/restic/cmd_find.go
middelink: Uhm. This is `main()` or equivalent. Unless cobra is going to supply us with a context, this is as high in the tree as we can get. So `context.Background()` felt like the right choice there.

For me, `context.TODO()` would appear in say refactoring the backends, when the restic library itself does not plumb the context through, so you would have to make one up.

src/cmds/restic/cmd_forget.go
middelink: Ack, but... I made the change in the other PR, so if that gets submitted, I can resolve the conflicts (it's going to be more than just this one.)

src/cmds/restic/cmd_forget.go
middelink: That using a channel to hand over a Snapshot one by one is completely defeated by putting them all in a list. But if you don't like the remainder, I can certainly make it go the way of the dodo's ^^

middelink:
Regarding the use of the old context in fuse. I don't think it matters much, they call us with a context and none of the restics fuse code uses it.

Then I found this:
  "However, I'm not sure how important it is to transition right away. The two context types interoperate (the interfaces are equivalent). It's only an issue if you've got it in a func/interface signature that you're trying to use as a type, and that's not very common."

So while not having a definite answer, it seems not a super big issue out there.

src/cmds/restic/cmd_forget.go
middelink: I actually rebased this tree on fix-841. So it should be fine now.

src/cmds/restic/cmd_find.go
fd0: Hm, I'd rather pass down a context from `main()`, so we can have the cleanup handlers cancel that context. And I think in `main()` is the right place for this to do (as we only need to do it once and not in each function). So I still think `context.TODO()` is the better choice here. But that should not be a merge blocker for now ;)

fd0:
Ah, thanks for looking it up, I forgot that `context.Context` is an interface. We need to also update the docs (readme and manual).

src/cmds/restic/find.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"context"
+
+	"restic"
+	"restic/repository"
+)
+
+// FindFilteredSnapshots yields Snapshots, either given explicitly by `args` or filtered from the list of all snapshots.
+func FindFilteredSnapshots(ctx context.Context, repo *repository.Repository, host string, tags []string, paths []string, args []string) <-chan *restic.Snapshot {
```
fd0: I'd suggest renaming `args` to `snapshotIDs` so it becomes clear what should be in there from the function signature.

src/cmds/restic/find.go
```diff
@@ -0,0 +1,78 @@
...
+			for _, s := range args {
+				if s == "latest" {
+					id, err = restic.FindLatestSnapshot(repo, paths, tags, host)
+					if err != nil {
+						Warnf("Ignoring %q, no snapshot matched given filter (Paths:%v Tags:%v Host:%v)\n", s, paths, tags, host)
+						usedFilter = true
+						continue
+					}
+				} else {
+					id, err = restic.ParseID(s)
```
fd0: The code here now requires passing a complete (non abbreviated) snapshot ID, that's not a good user experience:
```
$ bin/restic snapshots       
ID        Date                 Host        Tags        Directory
4b122706  2017-03-07 11:11:52  mopped                  /home/fd0/shared/work/restic/restic

$ bin/restic tag --add foo 4b122706
Ignoring "4b122706", it is not a snapshot id
No snapshots were modified

$ bin/restic list snapshots
4b12270680254642a8fe140dba8e7ede140e40af855899768133f4e01c0c875e

$ bin/restic tag --add foo 4b12270680254642a8fe140dba8e7ede140e40af855899768133f4e01c0c875e
Modified tags on 1 snapshots
```

There's `restic.FindSnapshot()` which finds the snapshot ID for from a prefix, it was removed in 63db3045c2bb8661a41ebd357984df6794e62d0d.

fd0 (changes requested):
I like what you did here, good work!

Some things still need to be done, from my point of view these are:
 * Add the new requirement for Go 1.7 to the readme and the manual
 * Add `--tag` to the `snapshots` command, that seems to be missing
 * Allow passing in a snapshot ID prefix (add a call to `restic.FindSnapshot()` for each ID

src/cmds/restic/cmd_find.go
middelink: Done. I had to add it to the `GlobalOptions`, but at least that makes it accessable to all cmd's

src/cmds/restic/find.go
middelink: Args....  You are so right.
I think some rebase mucked up my code. Both here and cmd_snapshot are weird. In fact, cmd_snapshot looks rather pristine. :(
Fixed.

src/cmds/restic/find.go
middelink: I named then `ids` as it is not sure yet they are actual `snapshotIDs`

middelink:
Addressed your latest comments (i think; it getting hard to see...)
-> documentation
-> refresh of cmd_snapshot
-> FindSnapshot in find.go

src/cmds/restic/cmd_find.go
fd0: Cool :)

src/cmds/restic/cmd_forget.go
```diff
@@ -83,32 +80,40 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
+				Verbosef("would have removed snapshot %v\n", sn.ID().Str())
+			}
+		} else {
+			var tags []string
+			if opts.GroupByTags {
+				tags = sn.Tags
+				sort.StringSlice(tags).Sort()
 			}
+			sort.StringSlice(sn.Paths).Sort()
+			k, _ := json.Marshal(key{Hostname: sn.Hostname, Tags: sn.Tags, Paths: sn.Paths})
```
fd0: Uh, you're throwing away an error here, that's not a good idea. I agree that there shouldn't be an error most of the time, but if there is one, `json.Marshal()` may return an empty buffer, so `string(buf)` will be the empty string, which means that all snapshots are sorted into the same group. That's not a good idea.

src/cmds/restic/global.go
```diff
@@ -49,6 +51,13 @@ func init() {
 		globalOptions.password = pw
 	}
 
+	var cancel context.CancelFunc
+	globalOptions.ctx, cancel = context.WithCancel(context.Background())
```
fd0: Nice!

fd0 (changes requested):
Ok, looks great!

Still open things:
 * Add proper error handling for `json.Marshal()`
 * What about my idea of bundling together filter criteria in a separate struct? Future work?
 * The option `--group-by-tags` is on by default (-> bug in the code, see comment there)
 * The `find` command does not print the snapshot the results are found in any more, was that intentional? Sample usage for find on a repo with several snapshots:
```
$ bin/restic find Dockerfile  
/restic/Dockerfile
/restic/Dockerfile
```
 * The `ls` command used to return an error when no snapshot IDs are given, now it lists ALL snapshots. Bug or feature?

fd0:
So, I went through the code and listed all open issues in the comment above. I know that it's hard to do, but it would be great if you could split such huge changes into sensible smaller commits. Reviewing one huge commit is really hard if so much is changed at once... Maybe have a look at how to rework and merge commits with `git rebase --interactive`?

src/cmds/restic/cmd_forget.go
middelink: Yeah, both fixed. Thanks!

middelink:
> The find command does not print the snapshot the results are found in any more, was that intentional?
Now I print a header when it finds a match:
```console
$ restic -r rest:http://qnap.**.********.**:8000 find -l bin
Found matching entries in snapshot 0a4ce6e3
Lrwxrwxrwx     0     0      0 2016-12-14 12:54:30 /bin -> usr/bin
dr-xr-xr-x     0     0      0 2016-12-14 12:59:29 /usr/bin
Lrwxrwxrwx     0     0      0 2016-07-03 14:41:12 /usr/lib/debug/bin -> usr/bin
drwxr-xr-x     0     0      0 2016-07-03 14:41:12 /usr/lib/debug/usr/bin
drwxr-xr-x     0     0      0 2016-07-03 14:44:42 /usr/lib64/pm-utils/bin
drwxr-xr-x     0     0      0 2016-11-05 16:38:36 /usr/local/bin
drwxr-xr-x     0     0      0 2016-11-05 16:38:37 /usr/share/locale/bin
Found matching entries in snapshot 0f65f72a
Lrwxrwxrwx     0     0      0 2016-12-14 10:23:13 /bin -> usr/bin
drwxrwxr-x  1000  1000      0 2016-10-13 20:53:56 /home/middelin/.atom/packages/build/node_modules/cross-spawn-async/node_modules/which/bin
drwxrwxr-x  1000  1000      0 2016-10-13 20:53:56 /home/middelin/.atom/packages/build/node_modules/cson-parser/node_modules/coffee-script/bin
drwxrwxr-x  1000  1000      0 2016-10-13 20:53:56 /home/middelin/.atom/packages/build/node_modules/getmac/bin
drwxrwxr-x  1000  1000      0 2016-10-13 20:53:56 /home/middelin/.atom/packages/build/node_modules/js-yaml/bin
drwxrwxr-x  1000  1000      0 2016-10-13 20:53:56 /home/middelin/.atom/packages/build/node_modules/js-yaml/node_modules/esprima/bin
...
```

> The ls command used to return an error when no snapshot IDs are given, now it lists ALL snapshots

Not sure. I can make a case for either ^^
1. most other commands play smart and iterate. (snapshots, forget, tag)
2. if we want to allow for filtering, we need to allow for no snapshot ids. Which, granted, does not mean that no filtering means all snapshots.

Alright, lets not rock the boat too much. I put in a check for it.

> What about my idea of bundling together filter criteria in a separate struct? Future work?

Uhm. I don't like it very much. First it will make the call to the function not at all easier to understand:
`FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args)` vs `FindFilteredSnapshots(ctx, repo, restic.Filter{Host:opts.Host, Tags:opts.Tags, Paths:opts.Paths, SnapshotIDs: args})`
Not much gain, granted, you get some idea what the arguments mean, but if that is the reason you should do it with every function in the API... And worse, if we decide to add a new filter in some future world, one would change the function signature and the compiler would find all call paths for you. Now you need to do greps and whatnot to find them all.
(was not able to find any preference indication when Googling though; so the internet is for once silent)

> I know that it's hard to do, but it would be great if you could split such huge changes into sensible smaller commits

Ouch. So one PR with a commit for say find.go and then all the cmd_xxx's separate? Aside some ordering issues where I reuse some functions from other commands, I think it should be do-able. Or did you mean a PR per commit? If you want smaller reviews, I think you meant that one...

fd0:
> > What about my idea of bundling together filter criteria in a separate struct? Future work?

> Uhm. I don't like it very much. First it will make the call to the function not at all easier to understand:
FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args) vs FindFilteredSnapshots(ctx, repo, restic.Filter{Host:opts.Host, Tags:opts.Tags, Paths:opts.Paths, SnapshotIDs: args})
Not much gain.

Hm, I think it's way easier to read:
```go
FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args)
```
versus:

```go
f:= restic.Filter{Host:opts.Host, Tags:opts.Tags, Paths:opts.Paths, SnapshotIDs: args}
FindFilteredSnapshots(ctx, repo, f)
```

> Granted, you get some idea what the arguments mean, but if that is the reason you should do it with every function in the API...

And for this reason I've for example used this pattern for the snapshot filter for `forget`:
https://github.com/restic/restic/blob/master/src/restic/snapshot_filter.go#L30-L35
https://github.com/restic/restic/blob/master/src/restic/snapshot_filter.go#L38

I don't like functions with long argument lists, I'd rather prefer to group them into logical chunks via custom types.

> And worse, if we decide to add a new filter in some future world, one would change the function signature and the compiler would find all call paths for you.

That is indeed true, but:

> Now you need to do greps and whatnot to find them all.

That's not necessary, there's [`guru`](https://godoc.org/golang.org/x/tools/cmd/guru), from vim-go you can just use `:GoCallers` to get all callsites for a function under the cursor.

fd0:
But we can also leave the filter thing as an optimization for the future.

fd0 (approved):

fd0 (approved):

fd0:
Thanks, merged!

PR #867, opened 2017-03-07
Optimize List() pipeline

middelink:
... I wanted to use contexts here too, but that requires too much plumbing right now.


fd0 (commented):
Ok, that code was probably a bit too complicated. Cancelling the sending is handled by `backend.List()`, so it's fine.

PR #871, opened 2017-03-09
Refactor output of `find` to allow for json and normal output.

### LINKED ISSUE #869: Add JSON ouput to find command
## Output of `restic version`

`Any`

## Expected behavior

This is exactly the same idea behind the new option *--json* for [snapshots](https://github.com/restic/restic/pull/751) command.
We would like to get the output as *JSON* instead of raw/not-formatted text:

```
$ restic find --json -s 82a6e9de 554*
[ 
 {"snapshot": "82a6e9dec437fd736f6a59cdd1eb11f68ab2fde01a7976050d80a6a5109aad27"}:
   [ 
     {"hits": 1},
     {"matches": [
        {"match": [ 
          {"path":"test\554a2d0ab44d5.pdf"},
          {"permissions": "-rw-rw-rw-"},
          {"size": 3171662},
          {"date": "2015-05-06 11:02:36 -0400 EDT"} 
        ]}
      ]}
    ]
]
```

or even simpler if the above isn't possible (or hard to achieve):

```
$ restic find --json -s 82a6e9de 554*
[ 
 {"snapshot": "82a6e9dec437fd736f6a59cdd1eb11f68ab2fde01a7976050d80a6a5109aad27"}:
   [ 
     {"hits": 1},
     {"matches": [ 
          {"match": "-rw-rw-rw-     0     0 3171662 2015-05-06 11:02:36 -0400 EDT test\554a2d0ab44d5.pdf"} 
       ]}
    ]
]
```

## Actual behavior

```
$ restic find -s 82a6e9de 554*
found 1 matching entries in snapshot 82a6e9dec437fd736f6a59cdd1eb11f68ab2fde01a7976050d80a6a5109aad27
  -rw-rw-rw-     0     0 3171662 2015-05-06 11:02:36 -0400 EDT test\554a2d0ab44d5.pdf
```

I'm willing to _sponsor_ this feature. Please reach me out if interested.

middelink:
Rather complicated solution becaused I wanted to retain the streaming
character of the output, which means for json I have to manually add
headers and footers per snapshot scanned + a list around the whole
set.

Closes #869

### DIFF
--- a/src/cmds/restic/cmd_find.go
+++ b/src/cmds/restic/cmd_find.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"path/filepath"
 	"strings"
 	"time"
@@ -84,7 +85,94 @@ func parseTime(str string) (time.Time, error) {
 	return time.Time{}, errors.Fatalf("unable to parse time: %q", str)
 }
 
-func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, prefix string, snapshotID *string) error {
+type statefulOutput struct {
+	ListLong bool
+	JSON     bool
+	inuse    bool
+	newsn    *restic.Snapshot
+	oldsn    *restic.Snapshot
+	hits     int
+}
+
+func (s *statefulOutput) PrintJSON(prefix string, node *restic.Node) {
+	type findNode restic.Node
+	b, err := json.Marshal(struct {
+		// Add these attributes
+		Path        string `json:"path,omitempty"`
+		Permissions string `json:"permissions,omitempty"`
+
+		*findNode
+
+		// Make the following attributes disappear
+		Name               byte `json:"name,omitempty"`
+		Inode              byte `json:"inode,omitempty"`
+		ExtendedAttributes byte `json:"extended_attributes,omitempty"`
+		Device             byte `json:"device,omitempty"`
+		Content            byte `json:"content,omitempty"`
+		Subtree            byte `json:"subtree,omitempty"`
+	}{
+		Path:        filepath.Join(prefix, node.Name),
+		Permissions: node.Mode.String(),
+		findNode:    (*findNode)(node),
+	})
+	if err != nil {
+		Warnf("Marshall failed: %v\n", err)
+		return
+	}
+	if !s.inuse {
+		Printf("[")
+		s.inuse = true
+	}
+	if s.newsn != s.oldsn {
+		if s.oldsn != nil {
+			Printf("],\"hits\":%d,\"snapshot\":%q},", s.hits, s.oldsn.ID())
+		}
+		Printf(`{"matches":[`)
+		s.oldsn = s.newsn
+		s.hits = 0
+	}
+	if s.hits > 0 {
+		Printf(",")
+	}
+	Printf(string(b))
+	s.hits++
+}
+
+func (s *statefulOutput) PrintNormal(prefix string, node *restic.Node) {
+	if s.newsn != s.oldsn {
+		if s.oldsn != nil {
+			Verbosef("\n")
+		}
+		s.oldsn = s.newsn
+		Verbosef("Found matching entries in snapshot %s\n", s.oldsn.ID())
+	}
+	Printf(formatNode(prefix, node, s.ListLong) + "\n")
+}
+
+func (s *statefulOutput) Print(prefix string, node *restic.Node) {
+	if s.JSON {
+		s.PrintJSON(prefix, node)
+	} else {
+		s.PrintNormal(prefix, node)
+	}
+}
+
+func (s *statefulOutput) Finish() {
+	if s.JSON {
+		// do some finishing up
+		if s.oldsn != nil {
+			Printf("],\"hits\":%d,\"snapshot\":%q}", s.hits, s.oldsn.ID())
+		}
+		if s.inuse {
+			Printf("]\n")
+		} else {
+			Printf("[]\n")
+		}
+		return
+	}
+}
+
+func findInTree(repo *repository.Repository, pat *findPattern, id restic.ID, prefix string, state *statefulOutput) error {
 	debug.Log("checking tree %v\n", id)
 
 	tree, err := repo.LoadTree(id)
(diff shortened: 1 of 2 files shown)


fd0 (changes requested):
We need a test for this, no way we can merge this and not notice when it breaks (e.g. invalid json output)...

middelink:
Awesome idea, testing included !

fd0:
FYI: The test fails on windows because (as far as I can see) because either the `testfile-hardlink` or `testfile-symlink` are not extracted from the `tar.gz` and therefore not archived, so the snapshot only contains one file called `testfile` that the second pattern `testfile*` matches.

Maybe use the pre-packaged `small-repo.tar.gz` and search for `bar` and `bar*` in snapshot `06c5a25c`:
```
$ restic -r ./repo find 'bar*'
Found matching entries in snapshot 6619a8ad
/testdir/bar
Found matching entries in snapshot d187480a
/testdir/bar
Found matching entries in snapshot 2145b769
/testdir/bar
Found matching entries in snapshot 06c5a25c
/testdir/bar
/testdir/bar2
```

middelink:
Hmm, I don't think you read that right. Under windows it seems to extract symlinks as real files (either tar or restore), because I get three files back, the normal, the hardlink and the symlink. And since 2 != 3 .. boom

Fixed edgecase by having printJSON handle symlinks now as well.

[{3 e5343e07cd055f251a9f1603c773d7642f3717239a4b8b4dcd9e943da826b58c [{\testdata\0\tests\testfile -rw-rw-rw- 21 2014-08-09 12:14:20 +0000 UTC 0 0} {\testdata\0\tests\testfile-hardlink -rw-rw-rw- 21 2014-08-09 12:14:20 +0000 UTC 0 0} {\testdata\0\tests\testfile-symlink -rw-rw-rw- 21 2014-08-09 12:14:28 +0000 UTC 0 0}]}]

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,93 @@ func parseTime(str string) (time.Time, error) {
...
+	JSON     bool
+	inuse    bool
+	newsn    *restic.Snapshot
+	oldsn    *restic.Snapshot
+	hits     int
+}
+
+func (s *statefulOutput) PrintJSON(prefix string, node *restic.Node) {
+	if node.Type != "file" && node.Type != "dir" {
+		// How to represent links and devices in json?
```
fd0: Add a `type` field to the JSON output and populate it with the type, so the caller can decide whether or not they are interested in other things as files.

fd0:
I wonder: shouldn't we just output the raw `restic.Node` as the hit, complemented by a `path` field with the absolute path (in addition to the `name` field)? The output of `find --json` will be read by a program (e.g. `jq`), so the caller can decide what fields they are interested in...

middelink:
Let me figure it out. Last time I tried, Node contained a lot of cruft, so I decided against it. But you are right, it is machine read...

middelink:
I did decide to filter some stuff. subtree, content, extended_attributes etc have no use in the find output IMO.

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,95 @@ func parseTime(str string) (time.Time, error) {
 	return time.Time{}, errors.Fatalf("unable to parse time: %q", str)
 }
 
-func findInTree(repo *repository.Repository, pat findPattern, id restic.ID, prefix string, snapshotID *string) error {
+type findNode restic.Node
```
fd0: I think it's better to declare that type within the function `PrintJSON()` where it is needed.

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,95 @@ func parseTime(str string) (time.Time, error) {
...
+type omit *struct{}
```
fd0: You do not need this, simply use a small data type like `byte` or or `string` instead of `omit`.

I think the generic way to declare something like your `omit` type would be the empty interface (`interface{}`), a pointer to the empty struct is rather unusual ;)

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,95 @@ func parseTime(str string) (time.Time, error) {
...
+			Printf("],\"hits\":%d,\"snapshot\":%q},", s.hits, s.oldsn.ID())
+		}
+		Printf(`{"matches":[`)
+		s.oldsn = s.newsn
+		s.hits = 0
+	}
+	if s.hits > 0 {
+		Printf(",")
+	}
+	b, err := json.Marshal(struct {
```
fd0: Nice trickery, I like it :smile:

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,95 @@ func parseTime(str string) (time.Time, error) {
...
+		// Add these attributes
+		Path        string `json:"path,omitempty"`
+		Permissions string `json:"permissions,omitempty"`
+
+		*findNode
+
+		// Make the following attributes disappear
+		Name               omit `json:"name,omitempty"`
```
fd0: Why did you decide to hide the name field?

src/cmds/restic/cmd_find.go
```diff
@@ -84,7 +85,95 @@ func parseTime(str string) (time.Time, error) {
...
+		Device             omit `json:"device,omitempty"`
+		Content            omit `json:"content,omitempty"`
+		Subtree            omit `json:"subtree,omitempty"`
+	}{
+		Path:        filepath.Join(prefix, node.Name),
+		Permissions: node.Mode.String(),
+		findNode:    (*findNode)(node),
+	})
+	if err != nil {
+		return
```
fd0: Before this return it'd be nice to finalize the JSON string we already printed, so that even in the case of an error, the output is valid JSON (at least).

In addition you're silently ignoring an error here, use `Warnf()` to print it to stderr instead.

fd0 (changes requested):
Nice work!

src/cmds/restic/cmd_find.go
middelink: "omit" has a certain charm in that it makes clear I do not want to field to show. But sure, I can make them bytes, so the struct becomes as small as possible :)

src/cmds/restic/cmd_find.go
middelink: Done.

src/cmds/restic/cmd_find.go
middelink: Because you suggested a while back :P
In seriousness, Path already has the information.

src/cmds/restic/cmd_find.go
middelink: Warnf() goes to stderr, have to remember that ^^

I will move the json code up, before setting up the surrounding fragments.That way I don't need to make up fake json for the error case.

src/cmds/restic/cmd_find.go
fd0: Oh, I did? Hm ;)

src/cmds/restic/cmd_find.go
fd0: Ok, even better. I like it!

fd0 (approved):

PR #880, opened 2017-03-12
Optimize by keeping node.ExtendedAttributes as map, not as list

middelink:
Make access to the extended attributes slightly easier by handeling them as a
map (each attribute name can only exist once anyway).

Handle backwards compatibility by unmarshaling of the old json data and
converting to the new one when needed.

Mark node.Content as optional in json data, to suppress "content":null fragments
when not needed. It seems for a single check, the code is already Content == nil resistant.

NOTICE: These json changes affect the content of a Treeblob and therefore its
ID. So while a backup against a parent before this change will not store the
content of dirs/files again, each directory will create a new treeblob.


middelink:
Djeez... that's annoying, the tests in src/restic won't run when my cwd is src/cmds/restic ... Ok, back to the drawing board.

fd0:
How do you run the tests? you can use either `gb test .` or `go test .`, provided you set the `GOPATH` accordingly. If you install `direnv`, the provided `.envrc` will take care of that.

src/restic/checker/checker.go
```diff
@@ -587,10 +587,6 @@ func (c *Checker) checkTree(id restic.ID, tree *restic.Tree) (errs []error) {
 	for _, node := range tree.Nodes {
 		switch node.Type {
 		case "file":
-			if node.Content == nil {
```
fd0: This check was introduced to recognize an error condition where the archiver won't properly save a list of blobs, but rather leave the `.Content` set to `nil`. I would have left the check, even if it seems redundant.

src/restic/node.go
```diff
@@ -18,36 +18,29 @@ import (
...
+	ChangeTime         time.Time         `json:"ctime,omitempty"`
+	UID                uint32            `json:"uid"`
+	GID                uint32            `json:"gid"`
+	User               string            `json:"user,omitempty"`
+	Group              string            `json:"group,omitempty"`
+	Inode              uint64            `json:"inode,omitempty"`
+	Size               uint64            `json:"size,omitempty"`
+	Links              uint64            `json:"links,omitempty"`
+	LinkTarget         string            `json:"linktarget,omitempty"`
+	ExtendedAttributes map[string][]byte `json:"extended_attributes,omitempty"`
```
fd0: What's the rationale behind this change? Apart from the slightly nicer structure (I get that one).

src/restic/node.go
```diff
@@ -331,11 +321,26 @@ func (node *Node) UnmarshalJSON(data []byte) error {
 	type nodeJSON Node
 	nj := (*nodeJSON)(node)
 
-	err := json.Unmarshal(data, nj)
-	if err != nil {
-		return errors.Wrap(err, "Unmarshal")
+	if err := json.Unmarshal(data, nj); err != nil {
+		// Retry conversion with list of ExtendedAttributes
```
fd0: Hm, now we have a slightly nicer structure and the comparison code is shorter, but `Unmarshal` is way more complex. I must admit that I do not really like it.

src/restic/node.go
```diff
@@ -437,45 +442,20 @@ func (node Node) sameExtendedAttributes(other Node) bool {
 	if len(node.ExtendedAttributes) != len(other.ExtendedAttributes) {
 		return false
 	}
-
-	// build a set of all attributes that node has
```
fd0: If it is only this check that bothers you, we can also just create a `map[string][]byte` for each node on the fly and use your new code, but leave the JSON structure as it is.

fd0 (changes requested):

middelink:
I normally just run `gb test` and figured that would set GOPATH properly.

src/restic/node.go
middelink: Quicker access to its elements. if you look at the fuse code, you see it returns a list of keys, and then gets a call per attribute. Walking the list of attributes again can become costly that way. O(n^2)

src/restic/node.go
middelink: That's done in integration testing, right? if it can still read small-repo.tar.gz and consorts, the legacy path is fine.

Also, from a bigger picture, isn't this the best way forward? How else would you make json effecting changes?

src/restic/node.go
middelink: That feels like postponing the inevitable. See previous answer :)

src/restic/node.go
fd0: But this is a matter of representation in memory, not on disk. For one, in my experience there are only a few (like max 10 or so) different extended attributes in active use, so even n^2 is still small. Second, if that's a concern, then we can also have a `map[string][]byte` for the extended attributes in the `file` struct, here: https://github.com/restic/restic/blob/master/src/restic/fuse/file.go#L31

Did you measure the impact of this change? Is it worth it?

src/restic/checker/checker.go
middelink: Well, that would defeat the purpose of the change, as Content can now be legit nil.

src/restic/node.go
fd0: For this special case I'd suggest a unit test for just the `Unmarshal()` function.

> That's done in integration testing, right? if it can still read small-repo.tar.gz and consorts, the legacy path is fine.

That depends on the code, I'm afraid. There are several archived repos that are tested, but there isn't any explicit test if the extended attributes of a node are restored correctly. So the failure case "extended attributes in the old format are not read at all, it always returns an empty map" won't be detected.

> Also, from a bigger picture, isn't this the best way forward? How else would you make json effecting changes?

Very, very carefully. If in doubt I think we should postpone optimizations for a future repository version. For the repo v2 work, I'll add an `upgrade` command (name tbd) that will migrate the data to the new format.

For all changes of the JSON format I'd like to have solid reasoning, there's a lot that may go wrong. Adding things is usually easy (e.g. introducing extended attributes at all), but changing things offers a lot of room for bugs and errors.

src/restic/checker/checker.go
fd0: Ah, I see what you did. Acknowledged.

middelink:
Alright, I close this for now, but keep it humming around in the back of my mind if you ever open up the repo-v2 :)

fd0:
agreed :)

PR #883, opened 2017-03-14
Cache size of last ReadAt on S3 for performance

trbs:
Each `obj.Stat()` call adds another request to the S3 endpoint
for some commands a lot of ReadAt calls are made for the same
object in S3. This patch essentially cuts the number of calls
to S3 in this case in half. Speeding up the progress and lowering
costs to S3.

It is assumed that the object in S3 should not change and therefor
the cached size should be valid. (If it would not be then the entire
`restic` process would have false information since it does two distinct
reads on a file which has changed in the mean time ?)

Note: in my case the `s3-backend.log` show a lot of
(almost) consecutive byte ranges like:
```
Range: bytes=3352327-3356150
Range: bytes=3356151-3357129
```

So there is a further improvement to be made there I think :)

### DIFF
--- a/src/restic/backend/s3/s3.go
+++ b/src/restic/backend/s3/s3.go
@@ -20,10 +20,12 @@ const connLimit = 40
 
 // s3 is a backend which stores the data on an S3 endpoint.
 type s3 struct {
-	client     *minio.Client
-	connChan   chan struct{}
-	bucketname string
-	prefix     string
+	client       *minio.Client
+	connChan     chan struct{}
+	bucketname   string
+	prefix       string
+	cacheObjName string
+	cacheSize    int64
 }
 
 // Open opens the S3 backend at bucket and region. The bucket is created if it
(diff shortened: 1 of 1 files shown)


trbs:
Unscientific comparison runs of `restic prune` 
```
[13:58] 1.96%  1 / 51 snapshots
vs
[2:58] 1.92%  1 / 52 snapshots
```

fd0 (changes requested):
This is a great idea! Thanks for sending a PR!

You are right: Files stored in a restic backend are never modified, only read and removed. So when we have the size once, it will not change. So I think we can even improve your idea: Just cache all sizes of all files that are queried. I propose the following changes:

 * Add a `map[string]int64` to the `s3` struct so that we can cache all sizes
 * Add a `sync.Mutex` to the struct so that we can serialize accesses. At the moment there's a race condition on `cacheObjName` and `cacheSize` accesses.

fd0 (approved):

mappu:
The original patch only kept stat history for one file. Despite that, it was enough for a dramatic improvement in `prune` time.

The updated patch does keep the history for longer. It's definitely better, but, for a very large restore task it might cause large memory growth..

I think it's better to refactor each `backend.Load` to not support negative offset. The feature is only used in one place. Then `s3.Load` does not need to call `stat` at all.

PR #902, opened 2017-04-04
storage of device id of files

### LINKED ISSUE #893: device id is not saved for files
The current master does not save the device id field of the `stat` struct, but that's needed for proper restore of hard links.

jgfrm:
Closes #893

### DIFF
--- a/src/restic/node.go
+++ b/src/restic/node.go
@@ -581,6 +581,7 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	case "file":
 		node.Size = uint64(stat.size())
 		node.Links = uint64(stat.nlink())
+		node.Device = uint64(stat.dev())
 	case "dir":
 	case "symlink":
 		node.LinkTarget, err = fs.Readlink(path)
--- a/src/restic/node.go
+++ b/src/restic/node.go
@@ -568,6 +568,7 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	}
 
 	node.Inode = uint64(stat.ino())
+	node.Device = uint64(stat.dev())
 
 	node.fillTimes(stat)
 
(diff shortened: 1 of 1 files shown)


src/restic/node.go
```diff
@@ -581,6 +581,7 @@ func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	case "file":
 		node.Size = uint64(stat.size())
 		node.Links = uint64(stat.nlink())
+		node.Device = uint64(stat.dev())
```
middelink: I wonder if this (and the other 2 places) should be moved up, under the node.Inode =
After all, the inode is useless without the device, and if we fill one unconditionally, we should fill both.

fd0:
Thanks a lot!

src/restic/node.go
middelink: I am afraid we're mixing up things here. Device is now used twice. Both as st_dev and as st_rdev. Which means if I make a hardlinked /dev/null, restic will store 518 (1,3) in the Device field and during restore this will never match any of the stored inodes in the HardlinkIndex, right?

fd0: Huh, I think you're right. Good catch! I'll prepare a PR, and add tests (so we can catch it).

jgfrm: This would mean we need two fields: one for st_dev and one st_rdev.
My suggestion would be to have Device for st_dev and Rdevice for st_rdev.

fd0: No, that's not possible: It would change the semantic of the `Device` field, which is set to `st_rdev` right now, that would break existing repos. I'm already working on it, we'll add a new field `DeviceID` for `st_dev`.

fd0: See #903 (this only adds the new field and the still failing tests)

fd0: And the second commit adds the fixes. @middelink thanks for looking out!

PR #903, opened 2017-04-05
Add new field DeviceID and tests

fd0:


fd0:
@middelink would you mind giving the code a look? :)

src/restic/node_unix_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return fi, true
+}
+
+func checkFile(t testing.TB, stat *syscall.Stat_t, node *restic.Node) {
```
middelink: Re disabling the test for darwin,
If youi change `stat` to a `*statUnix`, you can use `mtim()` and consorts to test times, no?

Well, that and some casts to Uint64...

fd0: Hm, possible. Then we'll need to change the package for the test from `restic_test` to `restic`, to be able to access `statUnix`. My intention with the test was to reuse only very little code...

middelink: True. Just a bit worrying that now both windows and mac are excluded from a rather crucial backup/restore test.

fd0: The problem is rather that the things we're accessing and saving in our data structures are very different in different OS, and very low-level as well.

fd0:
I've reworked the tests so that they are also run on darwin.

PR #908, opened 2017-04-13
Check allowed devices per path

### LINKED ISSUE #645: Handle filesystems mounted several times with --one-file-system
This is an extension of #626. At the moment, restic keeps a list of allowed device IDs for the directories passed to the `backup` command in order to implement `--one-file-system` and detect filesystem boundaries. When a file system is mounted multiple times, e.g. at `/mnt/foo` and `/mnt/bar/baz/foo`, and restic is called with `restic backup --one-file-system /mnt/foo /mnt/bar`, the fs at `/mnt/bar/baz/foo` is walked and archived (twice). This issue tracks handling this special case.

The discussion was initially started here: https://github.com/restic/restic/pull/626#issuecomment-253809059

### LINKED ISSUE #702: backup --one-file-system and 'nested' mount points.
## Output of `restic version`
debug enabled
restic 0.3.1 (a507374b121)
compiled at 2016-12-16 06:57:37 with go1.7.4 on openbsd/amd64

## Expected behavior
Example disk layout:

	/			(Mountpoint_1)
	/storage
		/disk1		(Mountpoint_2)
			/home
			/yyy
			/zzz
	/usr
	/whatever

Expected behaviour of 'restic backup -x / /storage/disk1/home' is that:

- Only data on Mountpoint_1 in / is backed up;
- Only data on Mountpoint_2 in /storage/disk1/home is backed up.

## Actual behavior
- Data on Mountpoint_1 in / is backed up;
- All data on Mountpoint_2 is backed up.

Repeat the above on nested mount points...

## Workaround
restic backup -x / && restic backup -x /storage/disk1/home

## Additional
cmd_backup.gatherDevices returns a map of allowed devices. cmd_backup.runBackup uses this map to exclude files. Items on this map are not related to the original files/dirs which have been selected for backup.

fd0:
Closes #645
Closes #702

### DIFF
--- a/src/cmds/restic/cmd_backup.go
+++ b/src/cmds/restic/cmd_backup.go
@@ -220,8 +220,8 @@ func filterExisting(items []string) (result []string, err error) {
 
 // gatherDevices returns the set of unique device ids of the files and/or
 // directory paths listed in "items".
-func gatherDevices(items []string) (deviceMap map[uint64]struct{}, err error) {
-	deviceMap = make(map[uint64]struct{})
+func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
+	deviceMap = make(map[string]uint64)
 	for _, item := range items {
 		fi, err := fs.Lstat(item)
 		if err != nil {
@@ -231,7 +231,7 @@ func gatherDevices(items []string) (deviceMap map[uint64]struct{}, err error) {
 		if err != nil {
 			return nil, err
 		}
-		deviceMap[id] = struct{}{}
+		deviceMap[item] = id
 	}
 	if len(deviceMap) == 0 {
 		return nil, errors.New("zero allowed devices")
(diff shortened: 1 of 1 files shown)


zx2c4:
The CPU usage increases enormously when scanning through directories now. As I have 64 gigs of ram, most of the filesystem info is cached, so in 0.5, a whole scan takes around 15 seconds. However, at the moment, I'm still waiting for it to finish. I'll report back on whether or not it worked if it ever completes.

There should be a more efficient algorithm you can use for this, though. For example, just make everything an absolute path, and then check if one is a prefix of the other.

fd0:
I've had a look at what `du` does, that's unfortunately not possible with the current archiver code (but I'll keep it mind when I'm reworking that code). I'll have a look if there's a way to make it more efficient.

zx2c4:
Yowza! Please disregard what I said above. I had compiled in debug mode.

fd0:
Ha, I was just writing a Benchmark function for that, because it looked wrong :P 

The `map` type in Go has a lookup of O(1) (=constant), and the number of lookups for each path is limited to just a few, so it shouldn't take so much cpu time...

zx2c4:
Great. I just tried this out, and I think it does what we want.

fd0:
Thanks for the feedback.

btw: thanks a lot for `pass`! I've been using it for years :)

zx2c4:
Horrah! My pleasure.

fd0:
So, then I'm gonna merge this.

PR #943, opened 2017-05-04
Small changes to cmd_forget and snapshot_filter

middelink:
Like all other CLI commands, cmd_forget should use the global context.

Update some variable names in snapshot_filter.apply to better show their behavior.


fd0:
I like it, good work!

PR #945, opened 2017-05-06
Add bash autocompletion generation

### LINKED ISSUE #942: bash autocompletion
It would be great to have autocompletion for restic.

https://github.com/spf13/hugo/blob/master/commands/genautocomplete.go could be used as a basis.

middelink:
Fix #942

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -16,6 +16,7 @@ Usage help is available:
       restic [command]
 
     Available Commands:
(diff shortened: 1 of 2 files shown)


middelink:
Meh, looks  like the s3 test timed out...

fd0:
I've restarted the test. Thanks!

fd0:
FYI: The s3 timeout was probably caused by https://github.com/minio/minio/issues/4275

PR #957, opened 2017-05-11
Change backup policy to be inclusive

### LINKED ISSUE #953: [RFC] Implement consistent backup/forget policy
This is a request for comments to improve (streamline, make more consistent) the current forget expire policy work.

## Background

Currently restics ExpirePolicy allows one to expire on Tag, Last, Hourly, Daily, Weekly, Monthly and Yearly policies. For each of the given elements one can specify how many of each one wants.

## Issue

The items in the ExpirePolicy are implemented serially, so first the Tags are processed, then the first (most recent) "Last" snapshots are retained and then the "Hourly" etc etc. This means, given an policy of say { Last: 1, Daily: 7, Weekly: 5, Monthly 99 }, and a year worth of running a backup each day, one would expect to see Monthly backups from the last backup of the Month, so 2016-01-31, 2016-02-29, 2016-03-31 etc etc. However, due to the way the ExpirePolicy is executed, the result is:
`[...<2016-12-25 (M)> <2016-11-27 (M)> <2016-10-30 (M)> <2016-09-25 (M)> <2016-08-28 (M)> <2016-07-31 (M)> <2016-06-26 (M)> <2016-05-29 (M)> <2016-04-24 (M)> <2016-03-31 (M)> <2016-02-29 (M)> <2016-01-31 (M)>]`
Note the for example April 24, definitely not the last day of the month...

The reason this happens is that snapshots are marked off one by one, so we keep the most recent snapshot (due to Last); then 7 daily ones; then 5 weekly ones. Note these weekly ones are already removing snapshots, so by the time the 5 weeks are gone, the Monthly checker can only work with the remaining snapshots of the month.

## Idea

Instead of applying policies sequentially, do them in parallel. Go over the (sorted) snapshots one by one and mark snapshots as matching one or more policies which apply. This means in my example that the most recent snapshot matches both Last, Daily, Weekly and Monthly and then start counting the policies down. In the end this results in a much more expected list:
`[... <2016-12-31 (M)> <2016-11-30 (M)> <2016-10-31 (M)> <2016-09-30 (M)> <2016-08-31 (M)> <2016-07-31 (M)> <2016-06-30 (M)> <2016-05-31 (M)> <2016-04-30 (M)> <2016-03-31 (M)> <2016-02-29 (M)> <2016-01-31 (M)> ]`

Also note that the property of counting snapshots down instead of looking at their actually date is retained. So if we have 6 backups made on Sundays, and one would use --keep-daily 5, only the last backup would be removed, not 5.

## Rationale

Consistency is all in backups. If one makes regular daily backups, one expects snapshots of the last day of the week, month, year. Not some arbitrary snapshot which fell out of some complicated formula.

### Demo program
Attached you find the source of my demo program. It starts out by creating 400 daily snapshots, the most recent one a year old, applying the policy, (use case for running forget once in a blue moon) and then for a full year, adding a single snapshot and applying the policy per day (use case for folks with more than one machine to backup...) Note for demo purposes, the program assigns a reason to each snapshot, so we can see why it was retained.
[policy.zip](https://github.com/restic/restic/files/984810/policy.zip)

old policy:
```console
[<2017-05-08 (L)> <2017-05-07 (L)> <2017-05-06 (L)> <2017-05-05 (D)> <2017-05-04 (D)> <2017-05-03 (D)> <2017-05-02 (W)> <2017-04-30 (W)> <2017-04-23 (W)> <2017-04-16 (W)> <2017-04-09 (W)> <2017-04-02 (M)> <2017-03-26 (M)> <2017-02-26 (M)> <2017-01-29 (M)> <2016-12-25 (M)> <2016-11-27 (M)> <2016-10-30 (M)> <2016-09-25 (M)> <2016-08-28 (M)> <2016-07-31 (M)> <2016-06-26 (M)> <2016-05-29 (M)> <2016-04-24 (M)> <2016-03-31 (M)> <2016-02-29 (M)> <2016-01-31 (M)> <2015-12-31 (M)> <2015-11-30 (M)> <2015-10-31 (M)> <2015-09-30 (M)> <2015-08-31 (M)> <2015-07-31 (M)> <2015-06-30 (M)> <2015-05-31 (M)> <2015-04-30 (M)>]
```

new policy:
```console
[<2017-05-08 (LDWM)> <2017-05-07 (LDW)> <2017-05-06 (LD)> <2017-04-30 (WM)> <2017-04-23 (W)> <2017-04-16 (W)> <2017-03-31 (M)> <2017-02-28 (M)> <2017-01-31 (M)> <2016-12-31 (M)> <2016-11-30 (M)> <2016-10-31 (M)> <2016-09-30 (M)> <2016-08-31 (M)> <2016-07-31 (M)> <2016-06-30 (M)> <2016-05-31 (M)> <2016-04-30 (M)> <2016-03-31 (M)> <2016-02-29 (M)> <2016-01-31 (M)> <2015-12-31 (M)> <2015-11-30 (M)> <2015-10-31 (M)> <2015-09-30 (M)> <2015-08-31 (M)> <2015-07-31 (M)> <2015-06-30 (M)> <2015-05-31 (M)> <2015-04-30 (M)>]
```

middelink:
Fix #953


doc/manual.rst
```diff
@@ -788,6 +788,9 @@ All the ``--keep-*`` options above only count
 hours/days/weeks/months/years which have a snapshot, so those without a
 snapshot are ignored.
 
+All snapshots are evaluated counted against all matching keep-* counts. A
+single snapshot on 30-09-2017 (Sun) will count as a daily, weekly and monthly.
```
fd0: I'd suggest formatting the date as 2017-09-30 so that US users aren't confused.

doc/manual.rst
```diff
@@ -796,6 +799,14 @@ is a safety feature: it prevents restic from removing many snapshots
 when no new ones are created. If it was implemented otherwise, running
 ``forget --keep-daily 4`` on a Friday would remove all snapshots!
 
+Another example: Suppose you make daily backups for 100 years. Then
+``forget --keep-daily 7 --keep-weekly 5 --keep-monthly 12 --keep-yearly 75``
+will keep the most recent 7 daily snapshots, then 4 (remember, 7 dailies
+already include a week!) last-day-of-the-weeks and 11 or 12
+last-day-of-the-months. (11 or 12 depends if the 5 weeklies cross a month).
```
fd0: Remove the dot after "months", the text in the braces belongs to the sentence before.

doc/manual.rst
```diff
@@ -796,6 +799,14 @@ is a safety feature: it prevents restic from removing many snapshots
...
+And ofcourse 75 last-day-of-the-year snapshots. All other snapshots are
```
fd0: "ofcourse"?

src/restic/snapshot_filter.go
```diff
@@ -124,84 +77,16 @@ func y(d time.Time) int {
...
-		f.Remove = append(f.Remove, sameBucket[1:]...)
-	}
-}
-
-// keepTags marks the snapshots which have all tags as to be kept.
-func (f *filter) keepTags(tags []string) {
-	if len(tags) == 0 {
-		return
-	}
+var a int
```
fd0: I don't like this singleton, singletons are almost always trouble. What if snapshots are filtered in parallel in several goroutines in some future version?

src/restic/snapshot_filter.go
```diff
@@ -124,84 +77,16 @@ func y(d time.Time) int {
...
-			f.Keep = append(f.Keep, sn)
-			continue
-		}
-		unprocessed = append(unprocessed, sn)
-	}
-	f.Unprocessed = unprocessed
+// always retuns a unique number for d.
+func always(d time.Time) int {
+	a++
+	return a
```
fd0: Since we have nanosecond precision in the snapshot timestamps, I think we can just use `int(d.UnixNano())` here. That should be unique, always (even if the `int64` is casted to an `int32`). This removes the need for the singleton `a`.

src/restic/snapshot_filter_test.go
```diff
@@ -235,57 +164,58 @@ var expireTests = []restic.ExpirePolicy{
 	{Yearly: 10},
 	{Daily: 7, Weekly: 2, Monthly: 3, Yearly: 10},
 	{Tags: []string{"foo"}},
+	{Tags: []string{"foo", "bar"}},
 }
 
 func TestApplyPolicy(t *testing.T) {
 	for i, p := range expireTests {
-		keep, remove := restic.ApplyPolicy(testExpireSnapshots, p)
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
```
fd0: FYI (no need to change this here): you can leave the name for the subtest empty, then the testing library will automatically assign a sequential number that can then be used for filtering subtests.

fd0 (changes requested):
Thank you very much for your contribution, I like the code a lot! I especially like that the filtering algorithm is much more elegant than I could've done it. And it's more consistent.

I've found a couple of smaller things that I'd like you to address, the singleton is (at least from my point of view) the most important one.

We need to add a warning to the release notes on the next release so that users are aware something in the forget settings has changed, I'm planning to release a new version of restic next week.

doc/manual.rst
middelink: Done.

doc/manual.rst
middelink: Done.

doc/manual.rst
middelink: "finally"

src/restic/snapshot_filter_test.go
middelink: It let me use `Fatalf()` on a per test basis instead of `Errorf()/continue`, but it was not really needed. Reverted.

src/restic/snapshot_filter.go
middelink: I agree, but I did not see a right other solution. See below.

src/restic/snapshot_filter.go
middelink: It breaks the tests. ParseUTC("2014-11-13 10:20:30 +0000 UTC") has zeroed nanoseconds and the tests contain multiple instances in the same hour, which now be collapsed to one entry.
In real world scenarios this will be a non-issue, but for tests...

Two solutions, I unfold `always()` into the loop like I did with the Tags or I add subseconds to the tests cases? I'm open for both solutions.

fd0: Great, that's really specific for those tests ;)

src/restic/snapshot_filter_test.go
fd0: Oh, why that? Actually, I really like subtests and slowly convert the codebase to using them, it gives me an easy method to just run one of the sub tests. So, put the subtests back in?

middelink:
BTW, I was super happy to find --update on the tests to update the golden files :) I was so not looking forward to hand editing them ^^

fd0:
Yeah, that's indeed a very handy pattern. Not my own invention, though, I've seen it on https://speakerdeck.com/mitchellh/advanced-testing-with-go by @mitchellh (thanks for sharing the awesome ideas!)

fd0:
What about the subtests? Leave it out, or put it back in?

middelink:
I removed it. They were not essential to the PR.

fd0:
merged, thanks!

PR #967, opened 2017-05-16
Fix a number of warnings from go report

middelink:


fd0:
Looks great, thanks for your work!

PR #974, opened 2017-05-25
Remove regular status printing for non terminals

fd0:


dirk777:
Is there any chance to activate the progress report for windows when running in background?

fd0:
Nop, nothing built in. As far as I know, Windows doesn't have signals (in the Unix sense). Sorry.

xcezzz:
I believe restic needs an actual logging facility that is not just a transient output to stdout or stderr and can be referenced without needing to just redirect these or turn on debug. 

Making this structured we can include information about the files, the progress and other information if necessary that can be read/parsed later without having to invoke restic (file, influxdb, etc).

dirk777:
I highly appreciate this approach. This would improve restic' usability and makes it more developer friendly

PR #990, opened 2017-06-02
Fix the links in the readme to render in RST

tmcarr:
Someone added markdown links in RST, went ahead and fixed that.


fd0:
Oh, good catch, thanks!

Issue #820, opened 2017-02-23
fuse: speed up listing very large files


fbkopp:
## Output of `restic version`
root@user-linux:~# restic version
restic 0.4.0 (v0.4.0-70-g3d1dc63-dirty)
compiled with go1.7.4 on linux/amd64

## Expected behavior
no delay to list directory contents

## Actual behavior
it take more than 12 minutes

## Steps to reproduce the behavior

```
root@user-linux:~# restic mount -r /data/teste2/ /mnt/ --allow-other --owner-root
enter password for repository:
Now serving the repository at /mnt/
Don't forget to umount after quitting!
```

****in another terminal
```
root@user-linux:~# cd /mnt/

root@user-linux:/mnt# ls
snapshots

root@user-linux:/mnt# cd snapshots/2017-02-22T08\:30\:31-03\:00/

root@user-linux:/mnt/snapshots/2017-02-22T08:30:31-03:00# ls
Backup

root@user-linux:/mnt/snapshots/2017-02-22T08:30:31-03:00# time ls
Backup

real    0m0.001s
user    0m0.000s
sys     0m0.000s

root@user-linux:/mnt/snapshots/2017-02-22T08:30:31-03:00# cd Backup/SERVER/WindowsImageBackup/SERVER/Backup\ 2017-02-21\ 213024/

root@user-linux:/mnt/snapshots/2017-02-22T08:30:31-03:00/Backup/SERVER/WindowsImageBackup/SERVER/Backup 2017-02-21 213024# time ls -la                                                      total 857713444
-rwxr-xr-x 1 root root          776 Fev 21 21:22 4a66920e-938b-49ab-8ebd-ece661f14c23_AdditionalFilesc3b9f3c7-5e52-4d5e-8b20-19adc95a34c7.xml
-rwxr-xr-x 1 root root        37498 Fev 21 21:22 4a66920e-938b-49ab-8ebd-ece661f14c23_Components.xml
-rwxr-xr-x 1 root root         5520 Fev 21 21:22 4a66920e-938b-49ab-8ebd-ece661f14c23_RegistryExcludes.xml
-rwxr-xr-x 1 root root         3138 Fev 21 21:22 4a66920e-938b-49ab-8ebd-ece661f14c23_Writer0bada1de-01a9-4625-8278-69e735f39dd2.xml
[...]

PR #998, opened 2017-06-05
fuse: Add cache for blob sizes

fd0:
Closes: #820

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Important Changes in 0.X.Y
    https://github.com/restic/restic/issues/989
    https://github.com/restic/restic/pull/993
 
+ * Improved performance for the fuse mount: Listing directories which contain
+   large files now is significantly faster.
+   https://github.com/restic/restic/pull/998
+
 Important Changes in 0.6.1
 ==========================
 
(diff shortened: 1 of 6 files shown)


middelink:
Should we run the tests with -race? maps aren't thread safe and I'm not 100% sure we cannot have multiple file listings in progress. aka, callbacks coming in from fuse.

Side note, this should be abstracted out, no? We now have multiple backends and cmd's where we are adding some form of caching to make things perform. Seems to me we should do this more central. For exmaple, I can imagine this commit not doing much on S3, as it already has a size cache...

fd0:
Good point, but it's not relevant here:
 * Maps aren't thread safe, and multiple request from fuse can come in in parallel, that's both true. However, concurrent read from maps is thread safe, and we'll only create the size cache once (when the snapshots dir is constucted), and then it's read only access. So that's not an issue here.
 * The map caches sizes for all **blobs** in the repo, the data is extracted from the index. The size cache for the s3 backend caches the size of **files**. It's a different level of abstraction.

fd0:
I've wrapped it in a struct and added a `Lookup()` method to make it clearer that the map is only queried, not changed concurrently.

PR #1003, opened 2017-06-07
Fix relative link to CONTRIBUTING.md

fwilhe:
The link does not work at the moment, not on github and not in the generated docs. This commit makes it work on github, but still the file is not part of the generated docs. Maybe we should just put an absolute link on the file in the repo on `master` branch there?

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -9,7 +9,7 @@ new feature. This way, duplicate work is prevented and we can discuss
 your ideas and design first.
 
 More information and a description of the development environment can be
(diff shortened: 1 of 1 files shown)


fd0:
Yeah, good point. Could you please just insert an absolute link? Thanks!

fd0 (changes requested):
Yeah, good point. Could you please just insert an absolute link? Thanks!

fwilhe:
> Yeah, good point. Could you please just insert an absolute link? Thanks!

@fd0 Updated.

fd0:
Thanks!

PR #1012, opened 2017-06-08
Travis: use latest versions of Go and rename tip to master

zcalusic:


zcalusic:
But, after checking which Go versions Travis really used (1.7.5 & 1.8.1), it seems this feature is not yet production ready. Nice idea (to not have to update .travis.yml after every new Go release), yet we'll have to wait until it really starts using the latest version of Go as [documented](https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use).

fd0:
Nice idea, I wasn't aware of the `x`!

PR #1019, opened 2017-06-10
ls: Print names with percent correctly

### LINKED ISSUE #1017: restic ls shows MISSING files
## Output of `restic version`
restic 0.6.1

restic -r /mnt/media/restictest ls latest | grep MISSING

local

## Expected behavior
the output from ls should match what is on the filesystem, excluding excludes

## Actual behavior
weirdly named files e.g. `./.config/gconf/desktop/gnome/interface/%!g(MISSING)conf.xml`

```
$ restic -r /mnt/media/restictest ls latest | grep MISSING
./.config/gconf/apps/%!g(MISSING)conf.xml
./.config/gconf/apps/GnomeBaker/%!g(MISSING)conf.xml
./.config/gconf/apps/GnomeBaker/Devices/%!g(MISSING)conf.xml
```
So `./.config/gconf/apps/GnomeBaker/%gconf.xml` becomes `./.config/gconf/apps/GnomeBaker/%!g(MISSING)conf.xml`

Maybe restic is mishandling the `%` character.

fd0:
Closes #1017

### DIFF
--- a/src/cmds/restic/cmd_ls.go
+++ b/src/cmds/restic/cmd_ls.go
@@ -52,7 +52,7 @@ func printTree(repo *repository.Repository, id *restic.ID, prefix string) error 
 	}
 
 	for _, entry := range tree.Nodes {
-		Printf(formatNode(prefix, entry, lsOptions.ListLong) + "\n")
+		Printf("%s\n", formatNode(prefix, entry, lsOptions.ListLong))
 
 		if entry.Type == "dir" && entry.Subtree != nil {
 			if err = printTree(repo, entry.Subtree, filepath.Join(prefix, entry.Name)); err != nil {


src/cmds/restic/cmd_ls.go
```diff
@@ -52,7 +52,7 @@ func printTree(repo *repository.Repository, id *restic.ID, prefix string) error
 	}
 
 	for _, entry := range tree.Nodes {
-		Printf(formatNode(prefix, entry, lsOptions.ListLong) + "\n")
+		Printf("%s\n", formatNode(prefix, entry, lsOptions.ListLong))
```
middelink: Wow, I cannot believe not one of the lint, vet or other obnoxious tools let that slip. Using format strings with user supplied strings is bad.... (and yes, pointing finger at self here...)

fd0: Happens to the best :)

Issue #979, opened 2017-05-29
runtime: out of memory: cannot allocate 536870912-byte block (556793856 in use)


middelink:
Likely supersedes  #526 

## Output of `restic version`
```console
# restic.debug version
debug enabled
restic 0.6.0-rc.1 (v0.6.0-rc.1-1-g37b107b)
compiled with go1.8.3 on linux/arm
```

## How did you start restic exactly? (Include the complete command line)
```console
# /usr/local/sbin/restic.debug -r rest:http://qnap.ch.polyware.nl:8000 backup -x --tag CH / /boot --exclude "/var/log/*,/var/tmp/*,/tmp/*" --mem-profile=/tmp
```

## What backend/server/service did you use?
rest

## Expected behavior
Making a backup.

## Actual behavior
Failed backup. out of memory.
```console
# /usr/local/sbin/restic.debug -r rest:http://qnap.ch.polyware.nl:8000 backup -x --tag CH / /boot --exclude "/var/log/*,/var/tmp/*,/tmp/*" --mem-profile=/tmp
debug enabled
enter password for repository: 
runtime: out of memory: cannot allocate 536870912-byte block (556793856 in use)
fatal error: out of memory

runtime stack:
runtime.throw(0x4c85ae, 0xd)
	/usr/local/go/src/runtime/panic.go:596 +0x70
runtime.largeAlloc(0x1ffffe00, 0x6ae01, 0x711f50)
	/usr/local/go/src/runtime/malloc.go:809 +0xe0
runtime.mallocgc.func1()
	/usr/local/go/src/runtime/malloc.go:702 +0x30
runtime.systemstack(0x10a1ca80)
	/usr/local/go/src/runtime/asm_arm.s:264 +0x80
runtime.mstart()
	/usr/local/go/src/runtime/proc.go:1132

goroutine 117 [running]:
runtime.systemstack_switch()
	/usr/local/go/src/runtime/asm_arm.s:209 +0x4 fp=0x10ba7a7c sp=0x10ba7a78
runtime.mallocgc(0x1ffffe00, 0x44ea60, 0x10b66e01, 0x10adab18)
	/usr/local/go/src/runtime/malloc.go:703 +0x9e0 fp=0x10ba7ad8 sp=0x10ba7a7c
runtime.makeslice(0x44ea60, 0x1ffffe00, 0x1ffffe00, 0x960, 0x0, 0x0)
[...]

PR #1022, opened 2017-06-10
WIP: Add cache for snapshots

fd0:
The code in this PR is Work In Progress to explore the possibilities a local cache for some data in the repo can give us.

**DO NOT USE THIS CODE**

### DIFF
--- a/doc/cache.rst
+++ b/doc/cache.rst
@@ -1,0 +1,58 @@
+Local Cache
+===========
+
+In order to speed up certain operations, restic manages a local cache of data.
+This document describes the data structures for the local cache. This document
+describes the cache with version 1.
+
+Versions
+--------
+
+The cache directory is selected according to the `XDG base dir specification
+<http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html>`. It
+contains a file named `version`, which contains a single ASCII integer line
+that stands for the current version of the cache. If a lower version number is
+found the cache is recreated with the current version. If a higher version
+number is found the cache is ignored and left as is.
+
+Snapshots
+---------
+
+The sub-directory `snapshots` stores cached data for each snapshot in a single
+file named after the complete snapshot ID. The snapshot ID is either the
+storage ID for the snapshot or, if it has been amended (e.g. by adding tags)
+the contents of the JSON field `original`.
+
+A snapshot cache file consists of several blocks, each prepended by a four byte
+length field (`uint32`) in little-endian encoding. All blocks are encrypted and
+authenticated with the repository master keys.
+
+The first block contains the snapshot data itself, encoded in JSON as it is
+saved in the repo. Afterwards follow all (encrypted) tree objects of this
+repository, sorted by the node name in depth-first order. The tree objects are
+embedded in the following JSON structure:
+
+.. code:: json
+
+    {
+        "path": "/",
+        "id": "2da81727b6585232894cfbb8f8bdab8d1eccd3d8f7c92bc934d62e62e618ffdf",
+        "tree": {
+          "nodes": [
+            {
+              "name": "testdata",
+              "type": "dir",
+              "mode": 493,
+              "mtime": "2014-12-22T14:47:59.912418701+01:00",
+              "atime": "2014-12-06T17:49:21.748468803+01:00",
+              "ctime": "2014-12-22T14:47:59.912418701+01:00",
+              "uid": 1000,
+              "gid": 100,
+              "user": "fd0",
+              "inode": 409704562,
+              "content": null,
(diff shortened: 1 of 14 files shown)


middelink:
First off, not a big fan of local caches for non static data. They tend to get corrupt and out of date and lead to call kind of unexpectedness. Also, I think it's an optimization restic does not really need right now, there are other -in repo - ways to tweak storage to make these things faster better stronger. If we start using caches right now, it papers over inefficient storage which will pop up during cache regeneration.

Without deep diving too much into the code, how does this operate for a repo in which many-ish (for me ~10) clients make their backups? As each makes a snapshot roughly 30 mins after the other, will each one have to update it's cache  - which likely needs a super expensive tree walk - when it discovers a new snapshot? Or only if it uses the snapshot or only if it writes the snapshot... or? :)

BTW, what about simply adding the entire tree into the snapshot itself? Same lifetime after all and no need to do complicated caching as the snapshot file itself is the cache. And yes, I realize you would need to version the snapshot file to make it work properly, but I fear a day like that is coming any way.

fd0:
Heh, I did not expect a review comment so early, but here we go (good questions by the way):

First, I'd like to explore what a cache can do and play with it a bit, especially without changing the repository format at all. Once we've explored that area a bit, we can think about changing the repository format to maybe match the format we're using for the on disk cache. Changing the repository format is a breaking change, which is bad for a backup program. And we will support repo v1 for quite some time.

> what about simply adding the entire tree into the snapshot itself?

I've experimented with it a bit, and large JSON documents suck (you can see that in #979). For this PR I'd like to explore how chaining together several (separately encrypted) JSON documents in a stream work (four byte length, uint32 little endian, followed by the encrypted data). This has several nice properties: We can process the (smaller) documents one by one, and even skip over them or seek to the correct location within the file if we know the position. For the snapshot cache, the documents consist of the snapshot JSON document itself, followed by the tree objects for each directory, sorted by names, depth first. If this is a good format, we can think about saving the snapshots directly in it in a future version of the repository format.

> If we start using caches right now, it papers over inefficient storage which will pop up during cache regeneration.

We're already stuck with a (non optimal) repository format which pops up during `prune`, that annoys not only our users but also me personally. `prune` runtime is bad and I'd like to improve it.

> Without deep diving too much into the code, how does this operate for a repo in which many-ish (for me ~10) clients make their backups? As each makes a snapshot roughly 30 mins after the other, will each one have to update it's cache - which likely needs a super expensive tree walk - when it discovers a new snapshot? Or only if it uses the snapshot or only if it writes the snapshot... or? :)

What I have in mind right now works as follows, for one of your hosts:
 * On start, restic lists all snapshots, and removes the snapshot cache files for non-existing snapshots. That's the way the cache is cleared.
 * `restic backup` is called. It first downloads all indexes, snapshots, looks for a parent snapshot. If there is one (and `--force` hasn't been run) it looks if there is a snapshot cache file for it. If there isn't it falls back to fetching the metadata objects on demand. Otherwise the cached snapshot is used, so no interaction with the repo is necessary to find out which files/dirs have been changed.
 * During the backup, restic writes a new snapshot cache file for the current snapshot. At the end (when the ID for the snapshot is known), it moves it into the right position (`~/.cache/restic/<repoid>/snapshots/<id>`) so if that snapshot is used as a parent the next time the cache file is already there.
 * The other hosts don't care about snapshots of other hosts unless they are selected as a parent snapshot.
 * For the one host running `prune`, the cache files are built for all snapshot. That's basically the same tree walk that `prune` does right now, except the result is preserved, so when a particular snapshot is preserved for the next run of `prune`, the cache file is already there so there's no need for an expensive walk.
 * Since the snapshot ID is unique (and even preserved when the tags have changed), restic knows when a snapshot has changed. In addition, the tree ID is included in the cached snapshot file, so when that has changed we know the cache file is invalid and can throw it away.

All in all we get the following:
 * For `backup`: When a cache file is there, the backup process should be **much** faster. During backup, a new cache file for the next backup run is saved, but that should not influence runtime at all. So it's either much faster or as slow as before.
 * For `prune`: The first run of `prune` without any cache will take the same amount of time as before. Any subsequent runs of `prune` will be much faster, since a lot of snapshots that have been preserved before are now cached.

> Also, I think it's an optimization restic does not really need right now, there are other -in repo - ways to tweak storage to make these things faster better stronger.

Do you have any ideas on how to improve the storage, without breaking the repository format?

fd0:
So, I'm abandoning this PR.

Caching the data for the snapshot is not viable, that's way too much data: A single snapshot cache entry, containing the snapshot itself, and a list of all tree objects referenced in the snapshot, for a directory of ~23GB with 60k dirs and 255k files is about 110MB uncompressed, and 60MB (compressed with snappy). That is too much data.

My experiment also shows that it is not a good idea to just add a single tree to the snapshot object: The snapshot files would be very very large.

I'll rework my approach.

Issue #996, opened 2017-06-05
GCS: Prune / check fails?


cafuego:
## Output of `restic version`

restic 0.6.1
compiled with go1.8.3 on linux/amd64

## How did you start restic exactly? (Include the complete command line)

`./restic_0.6.1_linux_amd64 prune`

## What backend/server/service did you use?

Google Cloud (REST via the s3 driver)

## Expected behaviour

Remove unreferenced packs, possibly print status of what was removed, then exit.

## Actual behaviour

2.5MB of output along the lines of:

```
counting files in repo
building new index for repo
unable to list pack 000a70db: ReadAt: offset is negative
unable to list pack 000579f2: ReadAt: offset is negative
...
...
...
unable to list pack fffb6ac3: ReadAt: offset is negative
[1:44] 100.00%  46535 / 46535 packs

repository contains 0 packs (0 blobs) with 0B bytes
processed 0 blobs: 0 duplicate blobs, 0B duplicate
load all snapshots
unable to list pack fffbae0c: ReadAt: offset is negative
unable to list pack fffd9eaa: ReadAt: offset is negative
unable to list pack fffc363a: ReadAt: offset is negative
unable to list pack fff58a7d: ReadAt: offset is negative
unable to list pack fffebe29: ReadAt: offset is negative
unable to list pack fff9d516: ReadAt: offset is negative
unable to list pack fffec873: ReadAt: offset is negative
unable to list pack fff955c0: ReadAt: offset is negative
find data that is still in use for 7 snapshots
id 49d7d23ef708943a01d3e2c33798ac1bddf0a4dda27d8b5072aec0beb135c0d5 not found in any index
restic/repository.(*MasterIndex).LookupSize
	src/restic/repository/master_index.go:53
restic/repository.(*Repository).LoadTree
	src/restic/repository/repository.go:535
restic.FindUsedBlobs
	src/restic/find.go:9
main.pruneRepository
	src/cmds/restic/cmd_prune.go:155
main.runPrune
	src/cmds/restic/cmd_prune.go:75
main.glob..func13
	src/cmds/restic/cmd_prune.go:24
github.com/spf13/cobra.(*Command).execute
	src/github.com/spf13/cobra/command.go:631
github.com/spf13/cobra.(*Command).ExecuteC
	src/github.com/spf13/cobra/command.go:710
github.com/spf13/cobra.(*Command).Execute
	src/github.com/spf13/cobra/command.go:669
main.main
	src/cmds/restic/main.go:60
[...]

PR #1038, opened 2017-06-15
Improve GCS support

fd0:
This is a followup PR for #996. The code prevents that an *os.File is closed by the s3 backend, which seems to happen when used with GCS. In addition, a bit of custom code is removed that was also present in the stdlib, and the backend Remove() method for s3 ignores errors for non-existing files.


fd0:
@cafuego Could you please test this?

PR #1040, opened 2017-06-15
Add local cache

### LINKED ISSUE #29: Implement Metadata Cache

### LINKED ISSUE #282: Archiver generates one GetReader requests per directory
When walking the file tree, the archiver generates one LoadJSONPack call per directory, each of which generates a GetReader request on the backend with a length and an offset. These requests seems to usually occur on a small number of pack files that contain blobs of type pack.Tree. When dealing with remote backends, it may be more efficient to load and cache these files.

### LINKED ISSUE #738: Prune performance
## Output of `restic version`

restic 0.3.3 (v0.3.3-21-g36276c4)
compiled with go1.8rc2 on linux/amd64

## Expected behavior

Prune command is quite slow. It could use some parallelism I think, especially for remote repos, where latency dominates.

## Actual behavior

For a 50GB remote repo, it takes 1h to check, but 16-24h hours to prune. I suppose prune should last somewhat longer than check, hopefully not that much longer eventually. :)

Just run prune on a big remote repo.

fd0:
The code in this PR adds a local cache for metadata. It may not be fully effective until the next time `prune` is run.
TODO:
 * [x] Add a basic cache for newly saved data (snapshots, index files)
 * [x] Make `check` cache aware
 * [x] Add code to purge old cache files
 * [x] Add code to `prune` to untangle mixed pack files
 * [x] Create new cache files with `O_EXCL`
 * [x] Fix strange error on Windows, "cannot remove file it is in use by a different process"
 * [x] Add documentation to the manual
 * [x] Exclude the cache from a backup
 * [x] Tag the directory by creating a `CACHEDIR.tag` file
 * [x] Fix error as reported in https://github.com/restic/restic/issues/29#issuecomment-327021065
 * [x] Decide where to add an explicit switch to update the cache

closes #282 
closes #29 
closes #738

### DIFF
--- a/doc/cache.rst
+++ b/doc/cache.rst
@@ -1,0 +1,26 @@
+Local Cache
+===========
+
+In order to speed up certain operations, restic manages a local cache of data.
+This document describes the data structures for the local cache with version 1.
+
+Versions
+--------
+
+The cache directory is selected according to the `XDG base dir specification
+<http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html>`__.
+Each repository has its own cache sub-directory, consting of the repository ID
+which is chosen at ``init``. All cache directories for different repos are
+independent of each other.
+
+The cache dir for a repo contains a file named ``version``, which contains a
+single ASCII integer line that stands for the current version of the cache. If
+a lower version number is found the cache is recreated with the current
+version. If a higher version number is found the cache is ignored and left as
+is.
+
+Snapshots and Indexes
+---------------------
+
+Snapshot, Data and Index files are cached in the sub-directories ``snapshots``,
+``data`` and  ``index``, as read from the repository.
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -387,6 +387,16 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 		return err
 	}
 
+	// exclude restic cache
+	if repo.Cache != nil {
+		f, err := rejectResticCache(repo)
+		if err != nil {
+			return err
+		}
+
+		rejectFuncs = append(rejectFuncs, f)
+	}
+
 	err = repo.LoadIndex(context.TODO())
 	if err != nil {
 		return err
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -19,6 +19,9 @@ var cmdCheck = &cobra.Command{
 	Long: `
 The "check" command tests the repository for errors and reports any errors it
 finds. It can also be used to read all data and therefore simulate a restore.
+
+By default, the "check" command will always load all data directly from the
+repository and not use a local cache.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -30,6 +33,7 @@ finds. It can also be used to read all data and therefore simulate a restore.
 type CheckOptions struct {
 	ReadData    bool
 	CheckUnused bool
+	WithCache   bool
 }
 
 var checkOptions CheckOptions
@@ -40,6 +44,7 @@ func init() {
 	f := cmdCheck.Flags()
 	f.BoolVar(&checkOptions.ReadData, "read-data", false, "read all data blobs")
 	f.BoolVar(&checkOptions.CheckUnused, "check-unused", false, "find unused blobs")
+	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use the cache")
 }
 
 func newReadProgress(gopts GlobalOptions, todo restic.Stat) *restic.Progress {
@@ -75,6 +80,11 @@ func newReadProgress(gopts GlobalOptions, todo restic.Stat) *restic.Progress {
 func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
 	if len(args) != 0 {
 		return errors.Fatal("check has no arguments")
+	}
+
+	if !opts.WithCache {
+		// do not use a cache for the checker
+		gopts.NoCache = true
 	}
 
 	repo, err := OpenRepository(gopts)
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -85,6 +85,25 @@ func runPrune(gopts GlobalOptions) error {
 	return pruneRepository(gopts, repo)
 }
 
+func mixedBlobs(list []restic.Blob) bool {
+	var tree, data bool
+
+	for _, pb := range list {
+		switch pb.Type {
+		case restic.TreeBlob:
+			tree = true
+		case restic.DataBlob:
+			data = true
+		}
+
+		if tree && data {
+			return true
+		}
+	}
+
+	return false
+}
+
 func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 	ctx := gopts.ctx
 
@@ -191,6 +210,11 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 	// find packs that need a rewrite
 	rewritePacks := restic.NewIDSet()
 	for _, pack := range idx.Packs {
+		if mixedBlobs(pack.Entries) {
+			rewritePacks.Insert(pack.ID)
+			continue
+		}
+
 		for _, blob := range pack.Entries {
 			h := restic.BlobHandle{ID: blob.ID, Type: blob.Type}
 			if !usedBlobs.Has(h) {
--- a/cmd/restic/exclude.go
+++ b/cmd/restic/exclude.go
@@ -12,6 +12,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/fs"
+	"github.com/restic/restic/internal/repository"
 )
 
 // RejectFunc is a function that takes a filename and os.FileInfo of a
@@ -175,5 +176,29 @@ func rejectByDevice(samples []string) (RejectFunc, error) {
 		}
 
 		panic(fmt.Sprintf("item %v, device id %v not found, allowedDevs: %v", item, id, allowed))
+	}, nil
+}
+
+// rejectResticCache returns a RejectFunc that rejects the restic cache
+// directory (if set).
+func rejectResticCache(repo *repository.Repository) (RejectFunc, error) {
+	if repo.Cache == nil {
+		return func(string, os.FileInfo) bool {
+			return false
+		}, nil
+	}
+	cacheBase := repo.Cache.BaseDir()
+
+	if cacheBase == "" {
+		return nil, errors.New("cacheBase is empty string")
+	}
+
+	return func(item string, _ os.FileInfo) bool {
+		if fs.HasPathPrefix(cacheBase, item) {
+			debug.Log("rejecting restic cache directory %v", item)
+			return true
+		}
+
+		return false
 	}, nil
 }
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -19,6 +19,7 @@ import (
 	"github.com/restic/restic/internal/backend/s3"
 	"github.com/restic/restic/internal/backend/sftp"
 	"github.com/restic/restic/internal/backend/swift"
+	"github.com/restic/restic/internal/cache"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/options"
 	"github.com/restic/restic/internal/repository"
@@ -38,6 +39,8 @@ type GlobalOptions struct {
 	Quiet        bool
 	NoLock       bool
 	JSON         bool
+	CacheDir     string
+	NoCache      bool
 
 	ctx      context.Context
 	password string
@@ -68,7 +71,8 @@ func init() {
 	f.BoolVarP(&globalOptions.Quiet, "quiet", "q", false, "do not output comprehensive progress report")
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repo, this allows some operations on read-only repos")
 	f.BoolVarP(&globalOptions.JSON, "json", "", false, "set output mode to JSON for commands that support it")
-
+	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache directory")
+	f.BoolVar(&globalOptions.NoCache, "no-cache", false, "do not use a local cache")
 	f.StringSliceVarP(&globalOptions.Options, "option", "o", []string{}, "set extended option (`key=value`, can be specified multiple times)")
 
 	restoreTerminal()
@@ -320,6 +324,17 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 	err = s.SearchKey(context.TODO(), opts.password, maxKeys)
 	if err != nil {
 		return nil, err
+	}
+
+	if opts.NoCache {
+		return s, nil
+	}
+
+	cache, err := cache.New(s.Config().ID, opts.CacheDir)
+	if err != nil {
+		Warnf("unable to open cache: %v\n", err)
+	} else {
+		s.UseCache(cache)
 	}
 
 	return s, nil
--- a/cmd/restic/integration_helpers_test.go
+++ b/cmd/restic/integration_helpers_test.go
@@ -199,6 +199,7 @@ func withTestEnvironment(t testing.TB) (env *testEnvironment, cleanup func()) {
 	env.gopts = GlobalOptions{
 		Repo:     env.repo,
 		Quiet:    true,
+		CacheDir: env.cache,
 		ctx:      context.Background(),
 		password: TestPassword,
 		stdout:   os.Stdout,
--- a/doc/man/restic-autocomplete.1
+++ b/doc/man/restic-autocomplete.1
@@ -41,8 +41,16 @@ $ sudo restic autocomplete
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
--- a/doc/man/restic-backup.1
+++ b/doc/man/restic-backup.1
@@ -79,8 +79,16 @@ given as the arguments.
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
--- a/doc/man/restic-cat.1
+++ b/doc/man/restic-cat.1
@@ -26,8 +26,16 @@ The "cat" command is used to print internal objects to stdout.
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
--- a/doc/man/restic-check.1
+++ b/doc/man/restic-check.1
@@ -18,6 +18,10 @@ restic\-check \- Check the repository for errors
 The "check" command tests the repository for errors and reports any errors it
 finds. It can also be used to read all data and therefore simulate a restore.
 
+.PP
+By default, the "check" command will always load all data directly from the
+repository and not use a local cache.
+
 
 .SH OPTIONS
 .PP
@@ -32,11 +36,23 @@ finds. It can also be used to read all data and therefore simulate a restore.
 \fB\-\-read\-data\fP[=false]
     read all data blobs
 
+.PP
+\fB\-\-with\-cache\fP[=false]
+    use the cache
+
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
--- a/doc/man/restic-dump.1
+++ b/doc/man/restic-dump.1
@@ -27,8 +27,16 @@ is used for debugging purposes only.
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
--- a/doc/man/restic-find.1
+++ b/doc/man/restic-find.1
@@ -59,8 +59,16 @@ repo.
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
 .PP
+\fB\-\-cache\-dir\fP=""
+    set the cache directory
+
+.PP
 \fB\-\-json\fP[=false]
     set output mode to JSON for commands that support it
+
+.PP
+\fB\-\-no\-cache\fP[=false]
+    do not use a local cache
 
 .PP
 \fB\-\-no\-lock\fP[=false]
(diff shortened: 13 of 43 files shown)


doc/cache.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Blobs
+-----
+
+The sub-directory ``Blobs`` stores Blobs combined to larger files. The directory
+contains one level of sub-directories, named ``00`` to ``ff``. These
+sub-directories store files with names between two and 64 hexadecimal
+characters. A file may only store Blobs whose SHA-256 hash starts with the file
+name. For example, consider the following cache directory for the repo with the
+ID starting with ``263a9c7bbc``:
```
rumpelsepp: You can put the `::` at the end if the line.

``` rst
ID starting with ``263a9c7bbc``::
```

is the same as

``` rst
ID starting with ``263a9c7bbc``:

::
```

mholt:
One of the appealing draws of restic is that there is no local cache regarding repo state - this was a major problem for us with duplicity. Are there guarantees that this cache won't get large? I'm gearing up to try a restic backup on all 7+ TB of our data at our ML lab (as a test run) (on second thought, I will probably start smaller on a subset of the files) and I'm worried about any aspect of this (or of index files) that won't scale to work well with large repos...

doc/cache.rst
fd0: Thanks, I didn't know that!

fd0:
Good question, thanks for asking. I've spent a lot of time thinking about the repository format and how we can achieve good performance without local state. I came to the conclusion that in order to fulfill the efficiency goal, we need at least some local state.

I did not use duplicity in a long time, and I don't know how they handle their state, but for restic it will follow some of the principles that worked so well for the repository.

For example, it's inefficient to download the index files every time. At the start, restic can just check which index files are there (that's already happening) and download them, and just keep the files in a local directory. When an index files isn't present in the repo any more, restic can purge it from the local cache. That's a very simple optimization which makes my personal backups at least 10-15 seconds faster at startup.

At the moment, the index files are enormous JSON documents that take some time (and memory) to parse. With a local cache, we can experiment parsing them just once, and leaving their contents in a local file in a much more digestible state. If we retain the original filename, we can easily tell when the cached files are outdated. Furthermore, this allows rapid prototyping, as the cache will also have a version number. So whenever we decide to change something, we increase the cache version number, throw away the old cache and build a new from scratch. This way, we can be backwards compatible to the repo v1 format for a long time, but still be efficient and fast.

As a principle restic (in its current form at least) will always be able to work just fine when the local cache is corrupt or does not exist. So if in doubt or the space is needed, users can always just throw away the cache.

I hope this answers at least a few of your questions/fears :)

zcalusic:
Thank you for a very nice explanation, @fd0!

Maybe even consider caching optional? Turned on with some switch, if/when you need it. That way you could cover both grounds, people which mostly fight latency (like me), and people who have issues with disk space. I definitely wouldn't mind another switch, needed to speed up things for remote backups.

mholt:
@fd0 That sounds good and reasonable. My main concern is how big the cache will get before it's rotated out. Duplicity's problem was it kept a cache basically indefinitely and to some uncontrollable size. Frankly, from my perspective, it's OK if backups take a while, but they shouldn't take more space than necessary.

I support the idea of a cache as long as 1) it's somewhat ephemeral, and 2) it contains a representation of ONLY what is needed for the current operation in a format that is optimized for the operation (like binary). Duplicity just kept everything around, it seems. I honestly have no idea what it was doing but it expended probably 500 GB of storage.

I am still in the process of my first full backup; do you have any idea how much space a typical cache might use? Is most of the cache going to be the indexes? (I'm assuming what makes a cache grow large is having lots of files being back up. Our ML datasets contain lots of small files...)

fd0:
I don't have any experience yet, let me come back to you with some numbers as soon as I have them.

lloeki:
> Duplicity just kept everything around, it seems. I honestly have no idea what it was doing but it expended probably 500 GB of storage.

It's not just that it kept things around, it was ballooning straight to space. Over here backing up a 40GB disk on a VM that was half filled, throwing away duplicity's cache it would rebuild it from scratch by fetching its index files and it would end up taking ~10GB. That's downright ridiculous when the index comes to be half the size of the backed up data.

fd0:
So, I've decided to implement just a very basic, simple cache for now. The commits need so me cleanup, but you can already test drive it. No guarantees, though.

mholt:
Thanks @fd0! Is the local cache opt-in, or will it be? (I'm mobile, so I haven't looked at the code change, so forgive my perhaps obvious question.)

fd0:
I'm not sure yet, I'm thinking about having it on by default (so most users will experience better performance) and having a switch to deactivate the cache. The cache will only store files that are in the repo anyway: The index and snapshot files are cache when they are read (that alone speeds up things a lot!)

In addition, I've changed the code so that now pack files either contain only data blobs or only tree blobs. The latter type of pack files is then also cached, so we can read the metadata directly from the cache and don't waste space with any other data.

That should give us the best of both worlds:
 * Files in the cache directly correspond to files in the repo
 * Only metadata is cached (index, snapshots, pack files with tree blobs)
 * Listings are never cached, the repo is always queried for that
 * When a file is not in the cache, it will be loaded from the repo, so throwing away the cache works just fine.

fd0:
I've played with the cache a bit, it's super awesome :)

rawtaz:
Nice work @fd0! Regarding your question, I think most people would agree that such migration is a step that should not happen automatically but be triggered by the user. Perhaps when running restic and the cache being enabled restic could tell the user that in order for the caching to be possible, the repository must be migrated (using command X)? Then the user will have the option of either disabling the cache, or do the migration in order to start using the cache.

That said, are we talking about a new version of the repo structure? How are "upgrades" to the repo structure dealt with normally, are they done automatically and behind the scenes?

fd0:
Ah, maybe I should clarify a bit: Being able to cache files (which contain only tree blobs/metadata) does not require any change to the repository structure and is fully backwards compatible. It used to be the case that restic just mixes whatever data comes along (data/metadata), bundles it together and uploads it to the repo. New files uploaded by the new code in this PR will create files that contain only metadata or only data, not a mix of both. And the former files can easily be cached. So it's just a matter of reorganizing the data that is already in the repo, what `prune` already does.

zcalusic:
I've tested this yesterday, in hope it would shorten 18h prune cycle on one of my repos. But, as we discussed in the meantime, it couldn't possibly work, because cache is populated at backup time, and this repo already has 27 snapshots without cache, so the prune still took almost the whole day to run. I hope you can implement cache populating for old snapshots (via check/prune or similar).

I've also peeked in the cache folder, and there's some more bad news there. A single backup adds about 550MB of cached data. Times number of backups, it looks like this repo would have about 15GB of cache for 75GB of data backed up. That's a lot (20% of data cached locally). I hope this will also be improved when you separate data & tree blobs? I'd say no more than 5% of the repo full size cached locally would be reasonable.

fd0:
I highly doubt that each of your snapshots will add 550MB. Have a look at the size of the individual subdirs in the cache dir please and report back.

zcalusic:
This is what it looks like:
```
1106	45ba74d09e656a6ce2a9a2216c1cdbb837b910eda9e4271b4d071eb2560a270f
26	b1024b269eb07fffe6f33346325015a685e7e02a4ba47b79e404d72e1350da98
```
Usage is in MB

zcalusic:
This first repo is the interesting one. I did one extra backup run yesterday, it ended with 576MB in cache folder. Then I run prune. Then a second backup ran this morning and that's the outcome. It's safe to assume that not many files have changed in the meantime, but as we don't have statistics yet, that's only an estimate.

fd0:
Ah, sorry, please look at the size of the subdirs of the per-repo cache dir ;]

zcalusic:
Ah, I see. That makes sense. This is how it looks like:
```
607	data
499	index
1	snapshots
```

I'm leaving that machine running with cache enabled, so we'll compare after a few days.

fd0:
Ok, that looks more reasonable. Please be aware that the cache does not contain any data that is not also stored in the repo. All data stored in the cache would otherwise be downloaded from the repo on demand.

For the indexes, I'd like to experiment with other file formats and/or compression (after the first version of the cache code is done)

zcalusic:
That cache contains only data from the repo is good news, it means the logic is simple and thus maintainable. Not good, great! :) Caches can be a bitch when it comes to data synchronization, keeping things simple is probably the most important thing.

Yeah, I suppose some space optimizations can be done later. Let's make it work first! Right now a feature to populate cache for previous snapshots would be most useful, IMHO. I feel like it shouldn't be too much trouble, am I wrong?

fd0:
> Right now a feature to populate cache for previous snapshots would be most useful, IMHO. I feel like it shouldn't be too much trouble, am I wrong?

Not at all, although that includes repacking data files, I'll build that one next. I've found an ugly bug yesterday and had to fix that first, see #1115

mappu:
Right now the cache (e.g. for indexes) is cleared as part of an index rebuild. It's special-purpose.

I think it would be better to have a general-purpose cache cleanup every time `backend.List()` is called. Anything not appearing in the backend list result can be cleared from the cache.

Although it means that the cache directory would need to be specialised on a per-repository basis. `.cache/sha256(repository_uri)/` perhaps? or `.cache/sha256(ChunkerPolynomial)/` or some other repository unique identifier.

zcalusic:
Another backup and data cache size grew from 607 MB to 720 MB, and index cache as expected stayed about the same size. That it very good already.

fd0:
@mappu most of what you suggested is already implemented, it just isn't pushed yet.

> Right now the cache (e.g. for indexes) is cleared as part of an index rebuild. It's special-purpose.

In the current code, index files are removed from the cache when they are removed from the repo (1) and (2) when the list of index files is requested on startup.

> I think it would be better to have a general-purpose cache cleanup every time backend.List() is called. Anything not appearing in the backend list result can be cleared from the cache.

That's not feasible for all data types: Index files and snapshot files are listed regularly, but data files are not. Basically, the code as it is now removes index and snapshot files when list() is called, and data files based on the index. When a data file is not contained in the index any more, the cache copy is removed as well. That's a conservative approach: If it is needed later (and still exists in the repo) it is simply downloaded (and cached) again.

> Although it means that the cache directory would need to be specialised on a per-repository basis. .cache/sha256(repository_uri)/ perhaps? or .cache/sha256(ChunkerPolynomial)/ or some other repository unique identifier.

That's already the case: Each repo has a unique ID, and I'm using that to create a new subdir below `~/.cache/restic`: https://github.com/restic/restic/pull/1040/files#diff-13762a446c34629dc970fe74455799d5R53 and https://github.com/restic/restic/pull/1040/files#diff-e693f6d151e99806d3169ef6aa780688R326

zcalusic:
@fd0, could you please rebase this branch so it includes bd31281f1e134c453ae451800fce3261adc0b45b commit?

I'd like to test the cache on a brand new repo, and that commit seems a bit frightening, and is missing in this PR.

Thanks.

fd0:
Updated.

@zcalusic Please remove the old cache before testing, I've discovered that the cache on `prune` may save too much data. Should be fixed now.

fd0:
Ah, and the `prune` code now untangles data and tree pack files, so expect a lot of data to be re-written, but afterwards all metadata should be in the local cache!

zcalusic:
I ❤️ this WIP, it's lots of fun. :)

I noticed a strange disk usage increase on my very first test of this code, yet it also dropped back suddenly. I was a bit lucky that it haven't ended with a disk full condition, it was really close. That happened overnight, so I don't know at which phase. Good job finding it, before it makes more damage! 😄

I'll continue testing the cache, both on some existing and one brand new repo. Thanks for the rebase, I appreciate it! 👍

zcalusic:
To give some more perspective and explain why it's fun... :)

I'm currently running an initial full backup of a VM in one country, sending to a rest-server in another country, which then saves the data on a networked distributed file system with several servers in a third country, with a replication factor of 2 (my decision).

Strange, but it all works. Double encrypted (restic + TLS), lots of latency along the way, and still plenty fast. Magic! 😄

tamalsaha:
I am excited to see this cache in restic, as I understand that it will improve upload performance for remote backends. Forgive me for asking naive questions that might have been answered above. I have some questions:

- Is there a way to tell restic where this cache will be stored? Is this using the restic TMPDIR env var?
- Does this only have metadata? Is there a bound on the size of the cache? Does this also cache actual data?
- Is this a 'cache" in the proper sense of the cache? Meaning if I lose access to the cache data at anytime, will my repository be safe ? What will be the performance impact next time?
- Does this cache handle multiple invocation of restic at the same time? Say, I open 2 terminals and run the exact same restic command? Are these restic commands going to fight over the cache? What actually happens?

Thanks.

fd0:
> Is there a way to tell restic where this cache will be stored? Is this using the restic TMPDIR env var?

There's the standard env variable `XDG_CACHE_HOME` (see the [xdg spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)), then there's the `--cache DIR` option. By default the cache is active, unless the directory is not writable or so. You can manually disable the cache with `--no-cache`.

> Does this only have metadata? Is there a bound on the size of the cache? Does this also cache actual data?

It caches only metadata: Index files (that previously were always downloaded from the repo), snapshot files and data files which only contain metadata. "Actual data", which I presume means "contents of files in the backup" is not stored in the cached. There is no upper limit that you can configure (at least yet), the natural upper limit is the amount of metadata in the repo.

> Is this a 'cache" in the proper sense of the cache? Meaning if I lose access to the cache data at anytime, will my repository be safe?

You can safely just throw the cache away and it will be rebuilt next time it is needed. restic now has opportunistic caching, which means that when a file in the repo (e.g. an index) is accessed and it is not yet in the cache, it will be stored there so that next time it can be read from the cache. When reading from the cache fails in any way (file is not there, corrupted, unreadable, [...]) restic prints en error message and falls back to just read the data from the repo.

> What will be the performance impact next time?

A non-initial backup will be slower because the metadata is fetched on demand. We'll need to see how we can rebuild the cache over time automatically.

> Does this cache handle multiple invocation of restic at the same time? Say, I open 2 terminals and run the exact same restic command? Are these restic commands going to fight over the cache? What actually happens?

Hm, good question. Each repo has its own cache, so only one instance of restic can have an exclusive lock (for operations like `prune` or `forget`), which means that when two instances run on the same repo in parallel it can only happen for a non-destructive operation such as `backup`. For this situation, only new data is added to the repo, the files have a random name with a very low probability for a collision. So new  files are going to be saved to the cache with new names, that's not an issue. It may happen that two instances try to cache the same file from the repo (e.g. an index or snapshot file) at the same time. That's something I didn't consider (yet), we can easily fix that by creating cache files with `O_EXCL` so that only one process can create the file. The other prints an error and carries on.

zcalusic:
In the meantime, this is how the first run with cache enabled looked like, on a "serious" existing repo. No problems at all. Can't wait to see what will happen next. :)

```
counting files in repo
building new index for repo
[6:52] 100.00%  16727 / 16727 packs
repository contains 16727 packs (3164956 blobs) with 74.833 GiB bytes
processed 3164956 blobs: 0 duplicate blobs, 0B duplicate
load all snapshots
find data that is still in use for 27 snapshots
[14:53:39] 100.00%  27 / 27 snapshots
found 3078657 of 3164956 data blobs still in use, removing 86299 blobs
will remove 0 invalid files
will delete 91 packs and rewrite 6662 packs, this frees 1.042 GiB
[5:43:34] 100.00%  6662 / 6662 packs rewritten
counting files in repo
[9:02] 100.00%  16489 / 16489 packs
finding old index files
saved new index as a5980711
remove 14 old index files
[9:06] 100.00%  6753 / 6753 packs deleted
done
restic prune  3371.59s user 1047.83s system 5% cpu 21:20:39.69 total
```

And the cache ended up with 1016 files in it, taking up this much space (values in MB):
```
4095    data
349     index
1       snapshots
```

mholt:
@fd0 Since caching is on by default, but there is no XDG_CACHE _HOME variable set or `--cache` option specified, where does the cache go, then?

Also, according to @zcalusic's experiment, it looks like a ~75 GB repo created a cache of ~4.5 GB. That's about 6% of the repo size. Do we expect that to be a somewhat normal ratio for cache sizes? I know it depends on metadata, but my knowledge of how metadata correlates to repo size is limited.

Like @jsissom, we'll be backing up multiple TB, and while I'm glad that the cache can be disabled, it seems like it would be most helpful when the repo size is huge. If we can pinpoint how much space, at maximum, the cache will use, I can try to keep that much free on the disk. Can the cache size be limited  by how much free space there is on the disk? I.e. if the disk grows to 90-95% full, can restic stop caching new files and just use whatever cache it can fit?

zcalusic:
@mholt, as you can see, this repo contains lots of blobs. It is not my biggest repo by size, but definitely my most painful one, due to lots of files.

So, depending on the average size of your files, you might end up with a lot smaller cache.

For me, I can definitely live with 6% cache if it will allow to run prune cycle right after every backup run , just like I have for all other repos except this one. Also, as @fd0 mentioned, there are possible improvements later on, switching from json to some more compact format, maybe compression etc...

I suppose we need to get this in first, then optimize later.

fd0:
> @fd0 Since caching is on by default, but there is no XDG_CACHE _HOME variable set or --cache option specified, where does the cache go, then?

Then it uses the default value for the XDG_CACHE_HOME variable, according to the xdg spec (linked above): "If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used."

> Also, according to @zcalusic's experiment, it looks like a ~75 GB repo created a cache of ~4.5 GB. That's about 6% of the repo size. Do we expect that to be a somewhat normal ratio for cache sizes? I know it depends on metadata, but my knowledge of how metadata correlates to repo size is limited.

The cache/metadata size depends a lot on how many files are stored in the repo. The more dirs/files, the more metadata, the more cache size. I have a repo with 43GB of data, and the cache size is ~660MB, or 1.4%.

> Like @jsissom, we'll be backing up multiple TB, and while I'm glad that the cache can be disabled, it seems like it would be most helpful when the repo size is huge. If we can pinpoint how much space, at maximum, the cache will use, I can try to keep that much free on the disk. Can the cache size be limited by how much free space there is on the disk? I.e. if the disk grows to 90-95% full, can restic stop caching new files and just use whatever cache it can fit?

It can, but there's no code (yet) to do that. On Linux/Unix, we could use `statfs` on the cache directory to get the free space, but I'm not sure how that works on Windows. However, when restic runs out of disk space for the cache, it will happily continue saving data to the repo. I agree that the situation is not optimal, but we need to start somewhere :)

zcalusic:
I ran restic check and got this out:
```
Load indexes
Check all packs
error for tree 69691d30:
  tree 69691d30, blob 6b9a777b: not found in index
error for tree c3d459e7:
  tree c3d459e7, blob 6b9a777b: not found in index
error for tree 28a1f822:
  tree 28a1f822, blob 6b9a777b: not found in index
restic check  2442.84s user 425.11s system 97% cpu 49:07.31 total
```

It seems like I have one extra blob. :)

mholt:
Ah, thanks - I hadn't heard of XDG before, and I should have clicked on the link. (Although the link doesn't tell me how restic's logic prioritizes that variable vs. its default value vs. `-cache`, so thanks for clarifying.)

I agree, go ahead forward with this, then optimize later. Just note that I probably won't personally be using the caching until we can set a hard size, either in bytes, or by bounding the % of disk space to remain free. (We have tens of millions of files, so that does not sound like it will bode well.)

tamalsaha:
I run restic in a Docker image inside Kubernetes cluster https://github.com/appscode/stash . Should there be a more non-desktop specific variable for location of cache dir?

zcalusic:
JFTR, I found "not found in index" errors on 3 other repos, which have not been running with cache enabled. So, it seems those errors are not cache specific.

fd0:
@tamalsaha Can't you just pass in a desired cache location via `XDG_CACHE_DIR` (that's also what other programs use) or call restic with `--cache <DIR>`?

fd0:
@zcalusic Let's open a new issue and debug this there. I'm eager to find out why there's a blob missing for you.

tamalsaha:
We can do `restic with --cache <DIR>` flag.

zcalusic:
I'm running prune with the populated cache for the first time. It's still very slow. To try to understand where it's spending time, I briefly traced the running restic process. What I observe is that lots of stat() calls, cache lookups obviously, get ENOENT (No such file or directory). What's worse, restic then repeatedly makes several calls to the remote end, without pulling that file locally first.

I admit I haven't yet looked at your code, 'cause I need to first do homework and better understand various data structures. But, even in my ignorance :), I'd expect that if a needed file is not in the cache, that cache logic first brings it in (caches it), then use the local copy to do the rest of the work. That would mean one cache MISS, and then dozens of cache HITS. Right now i see way too many accesses of the remote file.

Of course, this might be something completely different, we'll see...

zcalusic:
Showing only the most interesting spec:
```
find data that is still in use for 27 snapshots
[3:16:04] 100.00%  27 / 27 snapshots
found 3122078 of 3156601 data blobs still in use, removing 34523 blobs
```

So, it is about 5 times faster with cache. I think we will do much better than that when we understand what is not cached properly and should be.

zcalusic:
Took a brief detour, educating myself about various restic internals & data structures. Armed with this new knowledge :), I parsed the index contents, and found out that of 1298 packs containing tree nodes, only 1201 were cached.

Then decided to do a little experiment, to see just once how much speedup the cache ought to provide when we fix bugs. So, I manually copied missing packs to the cache folder and run prune again to see this:
```
find data that is still in use for 27 snapshots
[3:56] 100.00%  27 / 27 snapshots
found 3124939 of 3124938 data blobs still in use, removing -1 blobs
```
BTW, @fd0, that -1 reminds me of my missing blob :)

So, restic was finally CPU bound, not latency bound at that stage, using 100% of a single CPU core. And the speedup is about 225x, which is really great. All that for ~7% of content cached locally. Works for me. :)

Of course, in the future, 1000x will be realistic, by using all CPU cores. And those 7% could be less if we find a way to pack metadata better.

zcalusic:
I'm in the process of converting all my repos to use the index. And I also turned on cache support for check command, which I'm also adding to my regular backup cycle (after prune command). Basically, I'm gonna be stuck on this branch, until it's part of the master. ;)

On the positive side, on many other repos cache is really small, at around 1% of the whole repo size. The speedup continues to be amazing, what was measured in hours, now takes minutes.

On the negative side, during all conversions some packs do not get cached. I've written a custom program which finds those and fetches them manually. However small percentage of them is missing, it still has enormous impact on the performance, so I decided to fix that manually, until the time when it is properly fixed.

zcalusic:
Also closes #738

fd0:
So, this is basically done!

I'll wait for a few people to test it and then the cache code can be merged.

A few days ago I merged #1265, from then on restic will only create pack files that either only contain data (non-cachable) or metada (cachable). With the code in this PR, the next run of `prune` will be a bit larger because mixed-type packs will be split up so that they can be efficiently cached.

The cache is active by default (unless `--no-cache` is specified) and deactivated for the `check` command (unless `check --with-cache` is specified). When a file is accesses that only contains metadata, it is automatically pulled from the repo and cached locally, so the cache is automatically filled as it is needed.

zcalusic:
Tested the latest code on ALL my repositories, all went OK. I'd say this is ready to be merged. 👍 

Here's just one thing that might be interesting to investigate. For various reasons, 2 of my repos are not set up to email reports after backup run. Here's how I learned that is a very bad idea. One of them got locked somehow (interrupted prune or something), and the lock file remained for full 2 months without me noticing. This of course means that backup haven't run for about 2 months. 😞 

What this also means, in context of this PR, is that I converted that repo to use cache today, using the latest code from this branch (all the others repos were converted 1.5 months ago). It went OK, check comes out clean, still... there were some warnings during the process which you might wanna inspect.

```
scanned 651 directories, 1051629 files in 0:04
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
error loading tree: ReadFull(<data/2f590eadfa>): EOF
[16:35] 100.00%  0B/s  226.488 GiB / 226.488 GiB  1052280 / 1052280 items  0 errors  ETA 0:00 

duration: 16:35, 233.06MiB/s
snapshot f453ced0 saved
snapshots for (host [somehost], paths [/somepath]):
```

But, please do merge this PR soon, it's really ready!

fd0:
So, it's merged! Thank you everyone!

zcalusic:
And that is how we approach restic 0.8.0 😄 

Then 0.9.0 with compression support, and we'll have 1.0.0 in no time, for total world domination. 😀 

Good work, @fd0, as always!

armhold:
I got a screenful of warnings similar to @zcalusic :
```
scanned 133964 directories, 511998 files in 1:52
error loading tree: ReadFull(<data/22a6ed12a6>): EOFB  299805 / 645962 items  0 errors  ETA 0:40 
error loading tree: ReadFull(<data/22a6ed12a6>): EOFB  299847 / 645962 items  0 errors  ETA 0:40 
error loading tree: ReadFull(<data/22a6ed12a6>): EOF
```

fd0:
@armhold That sounds like a bug. Can you please describe what you did? Check the file `data/22/22a6ed12a6` in the repo and in the cache, is it maybe empty or too small?

fd0:
Ah, got it, it's a concurrency issue. I'll deal with it. Thanks!

fd0:
The bug is tracked in #1278, no need to add further comments here. Thanks for testing!

PR #1044, opened 2017-06-16
Improve restore

### LINKED ISSUE #982: Restore code is inefficient and produces inconsistent timestamps/permissions
## Output of `restic version`

```
> restic version
restic 0.6.0-rc.1 (v0.6.0-rc.1-0-g17ff41af)
```

See repro steps below.

Local + S3.

## Expected behavior

- timestamps should be restored consistently for intermediate directories
- rights should be restored consistently for intermediate directories
- paths that are known to be non-matching should not be recursively descended into (e.g `--include /test/foo` should not explore `/test/bar`)

## Actual behavior

- timestamps and rights are not restored correctly depending on the pattern and the fact that the path matches or not.
- the tree is always fully traversed

```
mkdir test/{foo,bar}
chmod 700 test/foo
touch test/{foo,bar}/{what,wut}
restic -r repo init
restic -r repo backup test
restic -r repo restore f608e300 --include 'wut' --target restored
rm -rf restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/bar  # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/foo  # timestamp is at restore date, rights are 0700
restic -r repo restore f608e300 --include 'foo' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
rm -rf restored
restic -r repo restore f608e300 --include 'test' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0755
ls -ld restored/test/bar  # timestamp is at backup date, rights are 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
restic -r repo restore f608e300 --include '/test/**' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/bar  # timestamp is at backup date, rights are 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
rm -rf restored
```

Yes. In fact I repro'd this following some code exploration looking into what could be done to improve `restore` performance and me stumbling on `selectedForRestore` then going down the rabbit hole :D. Along the way I annotated the files as seen in the diff below. I can start some work on this if you feel this is the right direction. There are a couple notes about parallelizing, which is a separate issue but might influence an updated design.

```diff
diff --git a/src/cmds/restic/cmd_restore.go b/src/cmds/restic/cmd_restore.go
index b19b580..dcb5775 100644
--- a/src/cmds/restic/cmd_restore.go
+++ b/src/cmds/restic/cmd_restore.go
@@ -119,6 +119,7 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return !matched
 	}
 
+	// This could return a `childMayMatch` bool and do some additional testing using filter.Match or something alike
 	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) bool {
 		matched, err := filter.List(opts.Include, item)
 		if err != nil {
diff --git a/src/restic/restorer.go b/src/restic/restorer.go
index 56916f3..90fa277 100644
--- a/src/restic/restorer.go
+++ b/src/restic/restorer.go
@@ -50,6 +50,7 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID ID, idx *HardlinkI
 		debug.Log("SelectForRestore returned %v", selectedForRestore)
 
 		if selectedForRestore {
+			// when parallelizing, enqueue this op
 			err := res.restoreNodeTo(node, dir, dst, idx)
 			if err != nil {
 				return err
@@ -61,6 +62,8 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID ID, idx *HardlinkI
 				return errors.Errorf("Dir without subtree in tree %v", treeID.Str())
 			}
 
+			// FIXME: unconditionally recursing, even if subp could be known to never match (esp. with abs. path)
+			// the parent `if` could use childMayMatch from the SelectFilter
 			subp := filepath.Join(dir, node.Name)
 			err = res.restoreTo(dst, subp, *node.Subtree, idx)
 			if err != nil {
@@ -70,6 +73,9 @@ func (res *Restorer) restoreTo(dst string, dir string, treeID ID, idx *HardlinkI
 				}
 			}
 
+			// FIXME: timestamp is restored only for final directory matching the pattern, e.g only for baz in foo/bar/baz, if baz is a dir, and in no other case
+			// instead, restoreTo could return a `hasRestorable` flag (along with the existing err) set to true if something was indeed found to be restored during recursion
+			// When parallelizing: watch out, this should happen after all descendant ops have been dequeued and completed
 			if selectedForRestore {
 				// Restore directory timestamp at the end. If we would do it earlier, restoring files within
 				// the directory would overwrite the timestamp of the directory they are in.
@@ -96,6 +102,9 @@ func (res *Restorer) restoreNodeTo(node *Node, dir string, dst string, idx *Hard
 	if err != nil && os.IsNotExist(errors.Cause(err)) {
 		debug.Log("create intermediate paths")
 
+		// FIXME: not mkdir str, restore node for each path component (which requires passing the current node stack), else attrs are incorrect
+		// also, one should take care of not doing restore ops twice or more if the node path component is already there.
+		// not sure this is worth parallelizing (due to each op depending on the previous one)
 		// Create parent directories and retry
 		err = fs.MkdirAll(filepath.Dir(dstPath), 0700)
 		if err == nil || os.IsExist(errors.Cause(err)) {
@@ -120,6 +129,7 @@ func (res *Restorer) restoreNodeTo(node *Node, dir string, dst string, idx *Hard
 // Before an item is created, res.Filter is called.
 func (res *Restorer) RestoreTo(dst string) error {
 	idx := NewHardlinkIndex()
+	// When parallelizing: good place to create a FIFO queue and n consumers
 	return res.restoreTo(dst, string(filepath.Separator), *res.sn.Tree, idx)
 }
 
```

lloeki:
This improves restore performance by several orders of magniture by not going through the whole tree recursively when we can anticipate that no match will ever occur.

Here is a sample run restoring one file from a `/` backup of a Debian machine on S3:

```
time (DEBUG_LOG=/tmp/restic-debug.log ./restic restore --include /etc/resolv.conf 1141ffca --target restore)
debug log file /tmp/restic-debug.log
debug enabled
restoring <Snapshot 1141ffca of [/] at 2017-06-15 09:25:24.915676881 +0200 CEST by root@test> to restore
ignoring error for restore/etc/resolv.conf: Lchown: lchown restore/etc/resolv.conf: operation not permitted
There were 1 errors

real	0m5.075s
user	0m0.927s
sys	0m0.183s
```

- time to reach the target file is significantly improved since it doesn't go down every directory before as well as inside `/etc`
- time after having restored the file is very short since it doesn't go down every other directory

(The error about `lchown` is spurious since I'm not running this as root)

I have not included time from the current version since it just takes literally forever.

Timestamps and rights not being restored correctly will be the next unit of work once this is dealt with.

Fixes #982

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -417,7 +417,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 	}
 
 	selectFilter := func(item string, fi os.FileInfo) bool {
-		matched, err := filter.List(opts.Excludes, item)
+		matched, _, err := filter.List(opts.Excludes, item)
 		if err != nil {
 			Warnf("error for exclude pattern: %v", err)
 		}
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -113,22 +113,32 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return nil
 	}
 
-	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) (bool, bool) {
-		matched, childMayMatch, err := filter.List(opts.Exclude, item)
+	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) (selectedForRestore bool, childMayBeSelected bool) {
+		matched, _, err := filter.List(opts.Exclude, item)
 		if err != nil {
 			Warnf("error for exclude pattern: %v", err)
 		}
 
-		return !matched, childMayMatch
+		// An exclude filter is basically a 'wildcard but foo',
+		// so even if a childMayMatch, other children of a dir may not,
+		// therefore childMayMatch does not matter, but we should not go down
+		// unless the dir is selected for restore
+		selectedForRestore = !matched
+		childMayBeSelected = selectedForRestore && node.Type == "dir"
+
+		return selectedForRestore, childMayBeSelected
 	}
 
-	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) (bool, bool) {
+	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) (selectedForRestore bool, childMayBeSelected bool) {
 		matched, childMayMatch, err := filter.List(opts.Include, item)
 		if err != nil {
 			Warnf("error for include pattern: %v", err)
 		}
 
-		return matched, childMayMatch
+		selectedForRestore = matched
+		childMayBeSelected = childMayMatch && node.Type == "dir"
+
+		return selectedForRestore, childMayBeSelected
 	}
 
 	if len(opts.Exclude) > 0 {
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -113,22 +113,22 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return nil
 	}
 
-	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) bool {
-		matched, err := filter.List(opts.Exclude, item)
+	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) (bool, bool) {
+		matched, childMayMatch, err := filter.List(opts.Exclude, item)
 		if err != nil {
 			Warnf("error for exclude pattern: %v", err)
 		}
 
-		return !matched
+		return !matched, childMayMatch
 	}
 
-	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) bool {
-		matched, err := filter.List(opts.Include, item)
+	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) (bool, bool) {
+		matched, childMayMatch, err := filter.List(opts.Include, item)
 		if err != nil {
 			Warnf("error for include pattern: %v", err)
 		}
 
-		return matched
+		return matched, childMayMatch
 	}
 
 	if len(opts.Exclude) > 0 {
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -40,6 +40,51 @@ func Match(pattern, str string) (matched bool, err error) {
 	return match(patterns, strs)
 }
 
+// ChildMatch returns true if children of str can match the pattern. When the pattern is
+// malformed, filepath.ErrBadPattern is returned. The empty pattern matches
+// everything, when str is the empty string ErrBadString is returned.
+//
+// Pattern can be a combination of patterns suitable for filepath.Match, joined
+// by filepath.Separator.
+func ChildMatch(pattern, str string) (matched bool, err error) {
+	if pattern == "" {
+		return true, nil
+	}
+
+	pattern = filepath.Clean(pattern)
+
+	if str == "" {
+		return false, ErrBadString
+	}
+
+	// convert file path separator to '/'
+	if filepath.Separator != '/' {
+		pattern = strings.Replace(pattern, string(filepath.Separator), "/", -1)
+		str = strings.Replace(str, string(filepath.Separator), "/", -1)
+	}
+
+	patterns := strings.Split(pattern, "/")
+	strs := strings.Split(str, "/")
+
+	return childMatch(patterns, strs)
+}
+
+func childMatch(patterns, strs []string) (matched bool, err error) {
+	if patterns[0] != "" {
+		// relative pattern can always be nested down
+		return true, nil
+	}
+
+	// match path against absolute pattern prefix
+	l := 0
+	if len(strs) > len(patterns) {
+		l = len(patterns)
+	} else {
+		l = len(strs)
+	}
+	return match(patterns[0:l], strs)
+}
+
 func hasDoubleWildcard(list []string) (ok bool, pos int) {
 	for i, item := range list {
 		if item == "**" {
@@ -102,21 +147,29 @@ func match(patterns, strs []string) (matched bool, err error) {
 
 // List returns true if str matches one of the patterns. Empty patterns are
 // ignored.
-func List(patterns []string, str string) (matched bool, err error) {
+func List(patterns []string, str string) (matched bool, childMayMatch bool, err error) {
 	for _, pat := range patterns {
 		if pat == "" {
 			continue
 		}
 
-		matched, err = Match(pat, str)
+		m, err := Match(pat, str)
 		if err != nil {
-			return false, err
+			return false, false, err
 		}
 
-		if matched {
-			return true, nil
+		c, err := ChildMatch(pat, str)
+		if err != nil {
+			return false, false, err
+		}
+
+		matched = matched || m
+		childMayMatch = childMayMatch || c
+
+		if matched && childMayMatch {
+			return true, true, nil
 		}
 	}
 
-	return false, nil
+	return matched, childMayMatch, nil
 }
--- a/internal/filter/filter_test.go
+++ b/internal/filter/filter_test.go
@@ -124,6 +124,77 @@ func TestMatch(t *testing.T) {
 	}
 }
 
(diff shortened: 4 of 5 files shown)


lloeki:
Please ignore `SAVEPOINT` commit, this is just WIP archival.

src/restic/filter/filter.go
```diff
@@ -37,9 +38,57 @@ func Match(pattern, str string) (matched bool, err error) {
...
+	}
+
+	// convert file path separator to '/'
+	if filepath.Separator != '/' {
+		pattern = strings.Replace(pattern, string(filepath.Separator), "/", -1)
+		str = strings.Replace(str, string(filepath.Separator), "/", -1)
+	}
+
+	patterns := strings.Split(pattern, "/")
+	strs := strings.Split(str, "/")
```
lloeki: There is duplication with `Match` here. This will be dealt with.

src/restic/filter/filter_test.go
```diff
@@ -124,6 +124,77 @@ func TestMatch(t *testing.T) {
...
+	{"/foo/bar/baz", "/foo", true},
+	{"/foo/bar/baz", "/foo/bar", true},
+	{"/foo/bar/baz", "/foo/baz", false},
+	{"/foo/**/baz", "/foo/bar/baz", true},
+	{"/foo/**/qux", "/foo/bar/baz/qux", true},
+	{"/baz/bar", "/foo", false},
+	{"/foo", "/foo/bar", true}, // maybe this should be false
+	{"/*", "/foo", true},       // maybe this should be false
+	{"/*", "/foo/bar", true},   // maybe this should be false
+	{"/foo", "/foo/bar", true}, // maybe this should be false
```
lloeki: There may be some semantic choice to be made here: `/foo` and `/foo/*` are equivalent to `/foo/**` currently. Maybe a single `*` could mean "restore just files at that level and `foo` or `/foo` mean "restore `foo` but if it's a dir, don't go down". I kept the existing behavior as is.

lloeki:
I have an integration test failing locally, but I can't make sense of it yet (I haven't looked into it, but I will).

lloeki:
Good, Appveyor [reproduced the same failure](https://ci.appveyor.com/project/fd0/restic/build/1769-master#L103)

src/cmds/restic/cmd_restore.go
```diff
@@ -112,22 +112,22 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
...
-	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) bool {
-		matched, err := filter.List(opts.Exclude, item)
+	selectExcludeFilter := func(item string, dstpath string, node *restic.Node) (bool, bool) {
+		matched, childMayMatch, err := filter.List(opts.Exclude, item)
 		if err != nil {
 			Warnf("error for exclude pattern: %v", err)
 		}
 
-		return !matched
+		return !matched, !childMayMatch
```
lloeki: Reverting that `!` on `childMayMatch` fixes the integration test.

fd0: Do you know that you can just force-push reworked commits, and they will replace the commits here?

.gopath
```diff
@@ -0,0 +1 @@
+/Users/lloeki/Workspace/go
```
fd0: Remove this file from the PR

src/cmds/restic/cmd_restore.go
```diff
@@ -112,22 +112,23 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
...
 	}
 
-	selectIncludeFilter := func(item string, dstpath string, node *restic.Node) bool {
-		matched, err := filter.List(opts.Include, item)
+	// This could return a `childMayMatch` bool and do some additional testing using filter.Match or something alike
```
fd0: Is this a work-in-progress comment?

src/restic/filter/filter_test.go
fd0: You're right, I like your suggestion. Make it so :) We'll need to document this, though.

fd0 (changes requested):
First few comments are in, having an in depth look now.

src/cmds/restic/cmd_restore.go
lloeki: Yes I do, serial abuser of `rebase -i` over here :)

lloeki: Had to leave pronto just as I found the fix but wanted to let you know (AFK for now)

.gopath
lloeki: Yep, SAVEPOINT commit

src/cmds/restic/cmd_restore.go
lloeki: Yep, SAVEPOINT commit

fd0:
I haven't found any other issues, good work already!

src/cmds/restic/cmd_restore.go
fd0: Heh, nice.  :)

lloeki:
Regarding this:

```
	{"/foo", "/foo/bar", true}, // maybe this should be false
	{"/*", "/foo", true},       // maybe this should be false
	{"/*", "/foo/bar", true},   // maybe this should be false
	{"/foo", "/foo/bar", true}, // maybe this should be false
```

> You're right, I like your suggestion. Make it so :) We'll need to document this, though.

I'd rather handle it in a separate issue/PR as this seems a bit more involved and I don't want to hold this PR back.

lloeki:
> Timestamps and rights not being restored correctly will be the next unit of work once this is dealt with.

@fd0 Would you like me to split this fix in a separate PR? I have not worked on it yet although I have some ideas for a fix.

lloeki:
Hmm, I had a hunch and tested this some more, and the behaviour for exclude filters is incorrect. `./restic restore --exclude /etc/resolv.conf 37b9c56d` skips the whole of `/etc`

lloeki:
Just to be sure I tested multiple include filters and it behaves as expected: `./restic restore --include /bin/bash --include /etc/resolv.conf 37b9c56d --target restored`.

lloeki:
The exclusion case is now fixed and its performance improved as well. I made some variable names more explicit (notably semantic of matched vs selected) to make it easier to reason about the code.

fd0:
Cool, let me know once you're satisfied with the PR and I'll have a look :)

lloeki:
Sorry this fell off the radar, go for it!

fd0:
Hey, just a quick question: We're switching to the idiomatic Go repository layout (see #1116). May I convert your changes?

lloeki:
Sure :)

fd0:
Hm, it seems I'm not allowed to push into your branch. Could you please check if you've enabled this feature? https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

lloeki:
Weird, that setting is active:

screen shot 2017-08-16 at 15 20 07

Anyway, I rebased.

fd0:
Thanks a lot, I'll have a look!

fd0:
Please excuse the long time I sat on this PR. It looks great and I've merged it. Thanks a lot for your work!

lloeki:
Awesome, many thanks!

fd0:
Would you like to continue working on the timestamps issue?

lloeki:
If time is *not* of the essence on that front, yes I can get my hands on it.

fd0:
It isn't, but we should probably document that in an issue (and that you'll work on it sometime).

lloeki:
OK, I created the issue.

fd0:
Awesome, thanks!

PR #1061, opened 2017-06-23
add docker image

bclermont:


docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.6
+
+COPY restic /usr/bin
```
tamalsaha: I believe ca-certificates should be installed.
```
RUN set -x \
  && apk update \
  && apk add ca-certificates \
  && rm -rf /var/cache/apk/*
```

bclermont: b7251dbea520cbd69d0abdd7c51439a1f6c371ef

docker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:3.6
```
jannickfahlbusch: I think, you could use `FROM scratch` here.

We don't need a whole Linux system to use restic.

tamalsaha: I don't think scratch includes ca-certificates. You need ca-certificates to connect to https urls.

bclermont: I also added `fuse`

jannickfahlbusch: You can copy the ca-certificates from the host  to `/etc/ssl/certs`.

jannickfahlbusch: I saw @bclermont too late, to reply in the same comment.

Didn't thought about `fuse`. Should be fine to use `alpine` rather than `scratch` then :)

bclermont: @jannickfahlbusch then you create a dependency on the host system. and what about docker on osx?

jkahrs: Assuming you have docker version 17.05+ you should be able to use multistage builds

bclermont: @jkahrs I know, but it's not even in `stable` repository, which still hold 17.03.1

jkahrs: @bclermont Just wanted to mention it as a possibilty for future work on this.
As of today 17.06 is in rc5 and should arrive soon(ish?) as next stable release.

tamalsaha: I for one is in favor of waiting at lease next patch release of Docker before using new features.

fd0:
Hey, thanks for your contribution. I'm not sure I understand the intention here: What's the use case for such a container? If it's only for building restic, we've already published the Dockerfile for the container we're using to produce the release binaries here: https://github.com/restic/builder

bclermont:
> use case for such a container

I want to keep everything as containers, I don't want to install backup software on the docker host. I want to have restic backup volume from other containers.

> https://github.com/restic/builder

I didn't saw this one.
btw, there is officially maintained images that have https://hub.docker.com/_/golang/

> Isn't it trivial to just write your own?

to not duplicate efforts, make it available easily for mass of users.
which is the point to opensource a backup tool like restic, to not have everyone implement their owns.

docker/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+COPY restic /usr/bin
```
quantonganh: This will failed because `restic` is in the root directory of repo while `Dockerfile` is put in `docker`.

The thing is we have another `Dockerfile` already: https://github.com/restic/restic/blob/master/Dockerfile

bclermont: @quantonganh https://github.com/restic/restic/blob/master/Dockerfile install Golang and bunch of dependencies for build and testing. that are totally unrelated to run the binary/daemon. My docker image is super small compared to the other, it's just alpine + restic binary.

bclermont: > This will failed because restic is in the root directory of repo while Dockerfile is put in docker.

```
$ docker run restic/restic
restic is a backup program which allows saving multiple revisions of files and
directories in an encrypted repository stored on different backends.

Usage:
  restic [command]
```

why it works for me?

quantonganh: I'm sorry. You're right.

```
+ /usr/local/bin/docker build --rm=true -f docker/Dockerfile -t 7529baea7f9e7b304ab8641ab2b86f08286e8279 . --pull=true
Sending build context to Docker daemon 60.63 MB
...

Step 2/4 : COPY restic /usr/bin
 ---> a7b7e770bc93
```

So, the build's context = current working directory = Git repo location.

fd0:
Thanks for the explanation. I'll need to look into this, but I think we can publish a docker image.

fd0:
Merged, thanks!

fd0:
And I've also pushed the image with restic 0.7.1 (just released) to https://hub.docker.com/r/restic/restic/.

PR #1086, opened 2017-07-05
Add support for IAM instance profile

### LINKED ISSUE #1067: Use IAM instance profiles for S3 authentication
Instead of passing key/secret, there should be a possibility to use metadata server to retrieve key/secret from IAM instance profile. Minio supports this approach, but there's no documentation how to configure that and it doesn't work out of the box with s3:s3.amazonaws.com/bucket URL.

Probably requires moving from minio.New to minio.NewWithCredentials.

kamsz:
This is probably a very ugly way to do it, but I'm not experienced in Golang. I've tested it with IAM instance profile attached and with regular environment variables/credential file.

Closes #1067

### DIFF
--- a/src/restic/backend/s3/s3.go
+++ b/src/restic/backend/s3/s3.go
@@ -43,7 +43,7 @@ func open(cfg Config) (*Backend, error) {
 	var err error
 
 	if cfg.KeyID == "" || cfg.Secret == "" {
-		debug.Log("iam")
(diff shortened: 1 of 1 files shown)


fd0 (changes requested):
Thanks for the contribution! Make the debug messages a bit more verbose, then it works for me.

fd0 (approved):

PR #1090, opened 2017-07-08
Update HasTags() and HasPaths() to follow #1081 feature request

### LINKED ISSUE #1081: forget --keep-tag is ambigous
A user reported that `forget --keep-tag` is not clearly defined. The manual says:

> `--keep-tag` keep all snapshots which have all tags specified by this option (can be specified multiple times).

 The help text is:
```
--keep-tag tag        keep snapshots with this tag (can be specified multiple times)
```

At the moment it's implemented this way: When just one tag to keep is specified (`--keep-tag foo`) restic won't touch any snapshots that have tag `foo`. For multiple tags to keep (e.g. `--keep-tag foo --keep-tag bar`) restic won't remove snapshots which have **both** tag `foo` and tag `bar`, but can forget snapshots which just have tag `foo`.

The manual states this (more or less clearly), but the help text is not clear.

I'd like to propose the following change: `--keep-tag foo` will work as is. `--keep-tag foo,bar` instructs restic to keep all snapshots that have both `foo` and `bar`. `--keep-tag foo --keep-tag bar` will not touch any snapshots that have either `foo` or `bar` set.

@middelink What do you think?

middelink:
Replace all but 3 occurrences of `StringSliceVar` to `StringArrayVar`. This will
prevent the flag parser to interpret the given values as a CSV string.

Both --tag, --keep-tag and --path can be given multiple times, the command will
match snapshots matching ANY of the tags/paths. Only when a value is given which
contains a comma separated list of tags/paths, ALL elements need to match.

Closes #1081


fd0:
Works great. I have a few ideas on how to improve the code, for example by adding a type `Tags`, so we can pass functions a `[]Tags`. I also think that splitting the arguments to `--path` and `--keep-tag` should be done in `cmds/restic`, or maybe in a separate function (like `NewTags([]string)` or so).

May I push a few commits on top of your commit here?

middelink:
Sure, I'm just not sure where you see the splitting of `--path` or ``--keep-tag`. pflag already gives us an []string. And if you refer to ... oh I see. You want the functions to have essentially [][]string. In that case, perhaps we should build an Any(...Tags) and All(...Tags) functions allowing them to nest. And then I'm sure someone wants a Not(), to denote something like all tags with CH but not TEST... ^^

My gut feeling is that would be optimizing too much. The splitting of the tag/path strings in the `HasTag()` and `HasPath()` functions dwarfs compared to the CPU time needed to parse the snapshot itself.

fd0:
Have a look at the code, I've added a type `TagList`.

middelink:
I'm very tempted to create a custom flag type for this, hiding all tag/path complexity from the cmds.

BTW, you're not allowing for ANDing path here, right? E.g. give me all the snapshots containing / and /boot paths.

fd0:
Hm, great idea actually! For the paths, it should always be AND, hm.

src/restic/tag_list.go
```diff
@@ -0,0 +1,62 @@
...
+	return l
+}
+
+func (l TagLists) String() string {
+	return fmt.Sprintf("%v", []TagList(l))
+}
+
+// Set updates the TagList's value.
+func (l *TagLists) Set(s string) error {
+	*l = append(*l, SplitTagList(s))
```
middelink: This feels wrong, why not do:
 *l = SplitTagsLists(l)

src/restic/tag_list.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+)
+
+// TagList is a list of tags.
+type TagList []string
+
+// SplitTagList splits a string into a list of tags. The tags in the string
+// need to be separated by commas. Whitespace is stripped around the individual
+// tags.
+func SplitTagList(s string) (l TagList) {
```
middelink: Does this still have to be a public function?

src/restic/tag_list.go
```diff
@@ -0,0 +1,62 @@
...
+func (TagList) Type() string {
+	return "TagList"
+}
+
+// TagLists consists of several TagList.
+type TagLists []TagList
+
+// SplitTagLists splits a slice of strings into a slice of TagLists using
+// SplitTagList.
+func SplitTagLists(s []string) (l TagLists) {
```
middelink: Does this still have to be a public function?

middelink:
Well, let me rephrase my example then
"Give me the snapshots containing either / or /boot" :)

PS. I left some comments on the commits

middelink (commented):
I left some comments on the commits

fd0:
> "Give me the snapshots containing either / or /boot" :)

That's not supported (yet), and it was not supported in any version of restic. Honest question: Do we need that?

src/restic/tag_list.go
fd0: No.

src/restic/tag_list.go
fd0: No.

src/restic/tag_list.go
fd0: Because that'd overwrite the previous value.

Let's say we have a flag with a taglist, like this:
```go
var tags restic.TagLists
pflag.Var(&tags, "tags", "use these tags")
```

When the binary is called like this:
```
$ foo --tags foo --tags bar --tags baz,boo,buu
```

then the method `Set()` is called three times:
```go
Set("foo") -> *l = [ ["foo" ]
Set("bar") -> *l = [ ["foo"], ["bar"] ]
Set("baz,boo,buu") -> *l = [ ["foo"], ["bar"], ["baz", "boo", "buu"] ]
```

fd0:
Hm, shall we forbid tags with commas? I.e. return an error when a tag (`backup --tag foo,bar`) with a comma is given?

middelink:
Do we need to? While I would argue that they should be separated out, in the past this used to work, No? So we would need to "lift up" the TagList in the TagLists. Or keep using VarSliceString :)

Also, I think in true Unix style we should be forgiving when parsing the input, as it is clear what the user wants, adding both tags, not doing an AND.

src/cmds/restic/global.go
```diff
@@ -72,7 +72,7 @@ func init() {
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repo, this allows some operations on read-only repos")
 	f.BoolVarP(&globalOptions.JSON, "json", "", false, "set output mode to JSON for commands that support it")
 
-	f.StringSliceVarP(&globalOptions.Options, "option", "o", []string{}, "set extended option (`key=value`, can be specified multiple times)")
+	f.StringArrayVarP(&globalOptions.Options, "option", "o", []string{}, "set extended option (`key=value`, can be specified multiple times)")
```
middelink: Same here. I feel -o key1=something,key2=somethingelse would be a valid argument.

src/cmds/restic/cmd_tag.go
```diff
@@ -43,9 +43,9 @@ func init() {
 	cmdRoot.AddCommand(cmdTag)
 
 	tagFlags := cmdTag.Flags()
-	tagFlags.StringSliceVar(&tagOptions.SetTags, "set", nil, "`tag` which will replace the existing tags (can be given multiple times)")
-	tagFlags.StringSliceVar(&tagOptions.AddTags, "add", nil, "`tag` which will be added to the existing tags (can be given multiple times)")
-	tagFlags.StringSliceVar(&tagOptions.RemoveTags, "remove", nil, "`tag` which will be removed from the existing tags (can be given multiple times)")
+	tagFlags.StringArrayVar(&tagOptions.SetTags, "set", nil, "`tag` which will replace the existing tags (can be given multiple times)")
```
middelink: Actually, I left these on purpose. I feel that --add A,B should add both A and B, not a tag named "A,B"

fd0: ah, I see

src/cmds/restic/global.go
fd0: Then we can never have an option that includes a comma as a value. Hm.

fd0:
I've removed the last commit. Should be fine now, do you agree?

fd0:
I think this is done, I'm going to merge this. Objections?

middelink:
None.

PR #1100, opened 2017-07-14
Allow absolute target path in build.go

### LINKED ISSUE #1099: build.go: absolute path for --output is treated as relative to cwd
## Output of ~~`restic version`~~`git describe`
`v0.7.0-22-g1d2045cb`

## How did you start ~~restic~~ the build exactly? (Include the complete command line)
`go run build.go -v -o /home/user/bin/restic`

Not applicable.

## Expected behavior
Resulting binary would be in `/home/user/bin/restic`.

## Actual behavior
Resulting binary would be in `/home/user/src/restic/home/user/bin/restic`.

1. `git checkout v0.7.0-22-g1d2045cb`
2. `go run build.go -v -o $HOME/bin/restic`

The parameter value for `--output` is always joined with the current working directory in 
https://github.com/restic/restic/blob/master/build.go#L404

_(hmm, __line 404__... so maybe the "not found" for the binary was intentional... :grin: )_

fawick:
Fixes #1099

### DIFF
--- a/build.go
+++ b/build.go
@@ -401,7 +401,10 @@ func main() {
 	if err != nil {
 		die("Getwd() returned %v\n", err)
 	}
-	output := filepath.Join(cwd, outputFilename)
+	output := outputFilename
+	if !filepath.IsAbs(output) {
+		output = filepath.Join(cwd, output)
+	}
 
 	version := getVersion()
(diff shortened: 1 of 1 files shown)


fd0:
Good catch, thanks!

PR #1101, opened 2017-07-14
Update to link to new document

tobya:
Links to where the new design document is.

PR #1103, opened 2017-07-15
Update links to design.md to design.rst

tobya:
Only really one place in docs/manual.rst but also comments in two other test files which I updated for completeness.   

This follows on from #1101


fd0:
Thanks!

PR #1117, opened 2017-07-19
Set default value for password-file flag from env

donat-b:
Allows defining password file path as RESTIC_PASSWORD_FILE=/foo


donat-b:
Rebased to master.

fd0:
LGTM, thanks!

donat-b:
By the way, are these variables documented anywhere besides cli help?

fd0:
Not yet, no. The `RESTIC_PASSWORD` and `RESTIC_REPOSITORY` are documented in the manual: https://restic.readthedocs.io/en/latest/manual.html#local

PR #1130, opened 2017-07-24
Fuse testing leaves test mountpoint around.

middelink:
Move it under the testing tree which is removed after each test.


fd0 (approved):
LGTM

PR #1131, opened 2017-07-24
Rework withTestEnvironment

### LINKED ISSUE #879: Rework tests (especially integration) to be "panic safe"
This is a followup issue for #873, I'm copying @middelink's text here:

> btw, I like this way of setting up:

```
func testBla() {
env, teardown := withEnvironment(t...)
defer teardown(t, env)

... testcode ...
}

```
> defer makes sure the teardown code gets run, even in light of a panic attack.

Rework all tests to use this pattern so that cleanup is run even when the test panics.

fd0:
Switch from a function passed as a parameter to a cleanup function, which is also executed when the test function panics, so no temporary directories are left behind.

Closes #879

### DIFF
--- a/cmd/restic/integration_fuse_test.go
+++ b/cmd/restic/integration_fuse_test.go
@@ -142,44 +142,45 @@ func TestMount(t *testing.T) {
 		t.Skip("Skipping fuse tests")
 	}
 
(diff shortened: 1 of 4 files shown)


middelink:
This looks to be a fix for #879 ?

fd0:
Yes :)

PR #1133, opened 2017-07-24
Force restic to ask the password when adding a key.

### LINKED ISSUE #1132: restic key add will not ask for password when RESTIC_PASSWORD or password-file is used.
## Output of `restic version`
```shell
restic 0.7.1 (v0.7.1-27-g608adf1-dirty)
```

rest
also irrelevant to the issue.

## Expected behavior
```console
$ RESTIC_PASSWORD=****** RESTIC_REPOSITORY=rest:http://****.**.polyware.nl:8000 ./restic key add
enter password for new key: 
enter password again: 
saved new key as ...
```

## Actual behavior
```console
$ RESTIC_PASSWORD=****** RESTIC_REPOSITORY=rest:http://****.**.polyware.nl:8000 ./restic key add
saved new key as ...
```
It did not ask the password for the new key, it blindly assumed I wanted the same password as I used opening the repo.

Just run the `restic key add` on an existing repo with the environment variable set.

Yes ^^ PR in the works.

middelink:
As `restic key add` uses the same `ReadPasswordTwice()` as the
rest of restic, it is sensitive to the environment variable
RESTIC_PASSWORD or --password-file= override.

When asking for the new key, temporary remove these 2 overrides, forcing
the password to be asked.

Closes #1132

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -249,7 +249,7 @@ func readBackupFromStdin(opts BackupOptions, gopts GlobalOptions, args []string)
 		return errors.Fatal("filename for backup from stdin must not be empty")
 	}
 
-	if gopts.password == "" && gopts.PasswordFile == "" {
+	if gopts.password == "" {
 		return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
 	}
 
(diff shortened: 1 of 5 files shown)


pvgoran:
Sounds like this solution will not allow non-interactive invocation of this subcommand.

Possibly a more universal solution would be to use a different environment variable and a different command line option for password that is added by `restic key add`. Like `RESTIC_NEW_PASSWORD` and `--new-password-file`. Moreover, these new environment variable/command line option could be also used by `restic init`; this way, a password to open the repository will be always `RESTIC_PASSWORD`/`--password-file`, and a "new" password (for new repository or new key) will be always `RESTIC_NEW_PASSWORD`/`--new-password-file`.

middelink:
Can you please file a new feature request for this? I'm just trying to fix the broken existing behavior (==bug) here and would be happy to discuss (and share my concerns ^^) on your feature request.

fd0 (approved):
Great, thanks!

PR #1149, opened 2017-08-05
Add Azure blob storage as backend

### LINKED ISSUE #609: Azure Blob Storage as repository
Is there a plan to add Azure Blob Storage as repository?

fd0:
Main work by @diptadas, thank you very much! Successor of #1059

TODO:
 * [x] Move files to new repo structure
 * [x] Vendor dependencies with `dep`
 * [x] Configure backend tests
 * [x] Test the backend thoroughly
 * [x] Add documentation to the manual (see https://github.com/restic/restic/pull/1059#issue-237783794)

Closes #609


fd0:
This is good work, thank you @diptadas again! And we need to switch to Go 1.8 at minimum, otherwise a library the Azure lib depends on won't compile. https://golang.org/doc/go1.8#net_http

flo82:
i tested the implementation a little bit. works functionally fine. thank you very much for the merge and pull request :-) excellent work.

i would like to mention one little thing:

the contents of the buckets are different in s3 vs. azure layout. Example:

if using s3 with repository address: RESTIC_REPOSITORY=s3:https://minio:9000/restic
you get a bucket restic and inside the bucket a folder named restic

if using azure with repository address: RESTIC_REPOSITORY=azure:restic:/
you get a bucket restic and inside the bucket there is NO folder named restic. it will start with the repository folders data, index ....

don't know if this right - but in my mind the s3 implementation could be optimized here.

Regarding the performance i got a 50mbit upload of files during the whole backup process. That maxed out my current internet connection and is on par with my current solution through minio in gateway mode. "check" and "prune" are also very fast - these functions seem to depend on the calling CPU owner.

fd0:
Yes, the default path within a bucket for s3 is (historically) `/restic`, for azure it's the path you specify (`/` in your example).

PR #1156, opened 2017-08-13
Readme: Correct link formatting

dimejo:


fd0:
You're right, thanks!

PR #1157, opened 2017-08-13
internal: check error before deferring file Close()

ceh:
If there is an error, file will be `nil`. We should check the returned error before deferring file `Close()`.


fd0:
Great catch, thanks!

PR #1164, opened 2017-08-17
Change key rm command to key remove

### LINKED ISSUE #1163: Key removal commands incorrect in manual
## Output of `restic version`

```
restic 0.7.1
```

`$ restic -r sftp:machine:tmp/restic key remove`

## Issues

1.  This command (`key remove`) prompts me for the repo password but does nothing.
2.  The manual says there are `key remove` and `key passwd` commands but does not explain them.
3.  A commit from few years ago added `key rm`, but it does not give helpful errors anymore:
```
$ restic -r sftp:machine:tmp/restic key
Fatal: wrong number of arguments

$ restic -r sftp:machine:tmp/restic key rm
Fatal: wrong number of arguments
```
4.  Trying to remove a key with `key remove` doesn't work:
```
$ restic -r sftp:machine:tmp/restic key remove dd9d2f10
Fatal: wrong number of arguments
```
5.  Removing a key with `key rm` does work, but this is not the command given in the manual:
```
$ restic -r sftp:machine:tmp/restic key rm dd9d2f10
removed key dd9d2f107513422794de5b8dca03796e600ed93737cd527ef2bd82a3c420872c
```

P.S.  Thank you for your work on restic!  I have been using Obnam for a few years, and since Lars is discontinuing it, I am looking for a replacement that can do what Obnam can.  I'm really excited about restic so far.  

It even addresses one of the few significant drawbacks that Obnam has, by grouping host-directory pairs when forgetting old backups.  This means that I can, e.g. backup my homedir daily, and backup a few important directories within it hourly, all to the same repo.  Obnam can do this, but it doesn't distinguish between backup roots when forgetting, so it's possible that, e.g. the whole-homedir backup could be pruned, leaving the subdir-only backup as the sole snapshot for that day, which is obviously bad.  So to avoid that, I have to backup into separate repos, and that means I don't benefit from deduplication across those backups.

Also, performance is very impressive!  One of Obnam's significant issues is slowness caused by network round-trip latency--even on a LAN it can turn a no-op backup into 30-60 seconds, while with restic, the same no-op backup takes *less than one second*.  I haven't tried restic over the Internet yet, but I'm optimistic!  Although the lack of compression makes it somewhat inappropriate for backups over the Internet.  Can't wait for that feature.  :)

ricardoseriani:
Change key rm command to key remove, to follow manual and other commands

Closes #1163

### DIFF
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -12,7 +12,7 @@ import (
 )
 
 var cmdKey = &cobra.Command{
-	Use:   "key [list|add|rm|passwd] [ID]",
+	Use:   "key [list|add|remove|passwd] [ID]",
 	Short: "manage keys (passwords)",
 	Long: `
 The "key" command manages keys (passwords) for accessing the repository.
@@ -124,7 +124,7 @@ func changePassword(gopts GlobalOptions, repo *repository.Repository) error {
 }
 
 func runKey(gopts GlobalOptions, args []string) error {
-	if len(args) < 1 || (args[0] == "rm" && len(args) != 2) || (args[0] != "rm" && len(args) != 1) {
+	if len(args) < 1 || (args[0] == "remove" && len(args) != 2) || (args[0] != "remove" && len(args) != 1) {
 		return errors.Fatal("wrong number of arguments")
 	}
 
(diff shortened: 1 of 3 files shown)


fd0:
Hey, thanks for your contribution, you'll need to update the man pages by running:
```
restic manpage --output-dir doc/man
```

ricardoseriani:
Man pages generated, as you advise.
Thanks for helping me contributing to this project.
Keep the awesome work.

fd0 (approved):
Looks great, thanks!

PR #1166, opened 2017-08-18
implement constrained restore that only walks paths specified in the --include pattern

armhold:
If the user is doing restore using "--include" paths, and all of the
given paths are absolute (begin with '/'), and no glob match chars are
present, we can prevent walking the entire restore tree, and instead
visit only subdirs that match the given "--include" paths.

This is a huge speedup for large/remote repos. In my testing with s3,
I went 53m18s to 1m17s. Prior to this patch, the restore often failed
to complete at all due to timeouts.

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -7,6 +7,8 @@ import (
 	"github.com/restic/restic/internal/restic"
 
 	"github.com/spf13/cobra"
+	"path/filepath"
+	"strings"
 )
 
 var cmdRestore = &cobra.Command{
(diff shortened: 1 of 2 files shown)


armhold:
There are a couple of Printf() calls at in cmd_restore.go @ L146 and L150 you might want to elide, or send to logging. I'm not totally sure what the preferred conventions are in restic.

fd0:
Hey, thanks for trying to make restic more efficient. I'm not sure if you're aware that a similar PR is still open (and I need to find time to review it) here: #1044 Would you mind having a look over there?

armhold:
My bad, I guess I didn't do a good job checking for similar issues. The proposed solution by @lloeki is much better than mine. In fact I tried to go down that route ("childMayMatch"), but punted on the difficulty, and so my PR only tries to handle absolute, non-globbed paths. Happy to close my PR, thanks!

PR #1170, opened 2017-08-19
Add option to exclude contents of cache directories

### LINKED ISSUE #317: Allow automatic exclusion of cache directories
I need to exclude specific directories from the backup, and could not find a way to do so.

Rather than adding cumbersome lists, an easy and widespread way is supporting the [Cache Directory Tagging Standard](http://fileformats.archiveteam.org/wiki/Cache_Directory_Tagging_Standard).

I currently use [Obnam](http://obnam.org/) and this is the one thing preventing me from switching to Restic. Which is quite impressive, both theoretically and practically: thank you for it.

fawick:
The option is named --exclude-caches and behaves as the likewise-named
option of tar(1): Directories are recognized as cache if they contain a
file named "CACHEDIR.TAG. The contents of the cache directories are not
to be backed up, except for CACHEDIR.TAG itself.

Closes #317.

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -53,16 +54,18 @@ given as the arguments.
 
 // BackupOptions bundles all options for the backup command.
 type BackupOptions struct {
-	Parent         string
-	Force          bool
-	Excludes       []string
-	ExcludeFiles   []string
-	ExcludeOtherFS bool
-	Stdin          bool
-	StdinFilename  string
-	Tags           []string
-	Hostname       string
-	FilesFrom      string
+	Parent           string
+	Force            bool
+	Excludes         []string
+	ExcludeFiles     []string
+	ExcludeOtherFS   bool
+	ExcludeIfPresent string
+	ExcludeCaches    bool
+	Stdin            bool
+	StdinFilename    string
+	Tags             []string
+	Hostname         string
+	FilesFrom        string
 }
 
 var backupOptions BackupOptions
@@ -76,6 +79,8 @@ func init() {
 	f.StringArrayVarP(&backupOptions.Excludes, "exclude", "e", nil, "exclude a `pattern` (can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.ExcludeFiles, "exclude-file", nil, "read exclude patterns from a `file` (can be specified multiple times)")
 	f.BoolVarP(&backupOptions.ExcludeOtherFS, "one-file-system", "x", false, "exclude other file systems")
+	f.StringVar(&backupOptions.ExcludeIfPresent, "exclude-if-present", "", "takes filename[:header], exclude contents of directories containing filename (except filename itself) if header of that file is as provided")
+	f.BoolVar(&backupOptions.ExcludeCaches, "exclude-caches", false, `excludes cache directories that are marked with a CACHEDIR.TAG file`)
 	f.BoolVar(&backupOptions.Stdin, "stdin", false, "read backup from stdin")
 	f.StringVar(&backupOptions.StdinFilename, "stdin-filename", "stdin", "file name to use when reading from stdin")
 	f.StringArrayVar(&backupOptions.Tags, "tag", nil, "add a `tag` for the new snapshot (can be specified multiple times)")
@@ -416,6 +421,18 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 		opts.Excludes = append(opts.Excludes, readExcludePatternsFromFiles(opts.ExcludeFiles)...)
 	}
 
+	if opts.ExcludeCaches {
+		if opts.ExcludeIfPresent != "" {
+			return fmt.Errorf("cannot have --exclude-caches defined at the same time as --exclude-if-present")
+		}
+		opts.ExcludeIfPresent = "CACHEDIR.TAG:Signature: 8a477f597d28d172789f06886806bc55"
+	}
+
+	excludeByFile, err := excludeByFile(opts.ExcludeIfPresent)
+	if err != nil {
+		return err
+	}
+
 	selectFilter := func(item string, fi os.FileInfo) bool {
 		matched, _, err := filter.List(opts.Excludes, item)
 		if err != nil {
@@ -424,6 +441,11 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 
 		if matched {
 			debug.Log("path %q excluded by a filter", item)
+			return false
+		}
+
+		if excludeByFile(item) {
+			debug.Log("path %q excluded by tagfile", item)
 			return false
 		}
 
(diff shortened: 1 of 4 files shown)


cmd/restic/cmd_backup.go
```diff
@@ -427,6 +429,18 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 			return false
 		}
 
+		if opts.ExcludeCaches && fi.Name() != "CACHEDIR.TAG" {
+			d := filepath.Dir(item)
+			_, err := os.Stat(filepath.Join(d, "CACHEDIR.TAG"))
```
fd0: This is not enough, the file also needs to start with the following string: 
```
Signature: 8a477f597d28d172789f06886806bc55
```

Details here: http://www.brynosaurus.com/cachedir/spec.html

I'd suggest adding a separate function like this:
```
func isCacheFile(filename string) bool {}
```

The function then tests if there is a valid `CACHEDIR.TAG` in the same directory (and the file is not the cachedir tag file itself). That way the logic is nicely abstracted away from the select function. What do you think?

If we think it is too expensive to read the file every time we could also add a global cache (protected by a mutex) of known cache dirs and non-cache dirs. But let's add this first and worry about performance when it becomes an issue.

fd0 (commented):
Thanks!

cmd/restic/cmd_backup.go
fawick: Oh, right, I forgot about the header. I should have read the relevant section of `tar(1)` instead of only the excerpt in #317. 😁 
I fully agree that it's best to refactor the check into a function and will proceed accordingly.

fawick: Done.

rawtaz:
I hate to downvote something @fawick put time into, but see #317 for some thoughts on why we shouldn't implement this (especially when there's alternatives).

fawick:
The latest commit implements what was decided on here: https://github.com/restic/restic/issues/317#issuecomment-326807674

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+	if excludeIfPresent != "" {
+		colon := strings.Index(excludeIfPresent, ":")
+		switch {
+		default:
+			tf = excludeIfPresent[:colon]
+			tc = excludeIfPresent[colon+1:]
+		case colon < 0:
+			tf = excludeIfPresent
+		case colon == 0:
+			return nil, fmt.Errorf("no name for exclusion tagfile provided")
```
fd0: Here too: Handle the error case directly after the call to `Index()`

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+		}
+	}
+	if tf != "" {
```
fd0: This case cannot happen (you've handled `colon == 0` earlier already), drop the `if`.

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+func isExcludedByFile(filename, tagFilename, header string) bool {
+	if tagFilename == "" {
+		return false
+	}
+	dir, base := filepath.Split(filename)
+	if base == tagFilename {
+		return false // do not exclude the tagfile itself
+	}
+	tf := filepath.Join(dir, tagFilename)
+	_, err := os.Stat(tf)
```
fd0: Use `github.com/restic/restic/internal/fs` instead of `os`, for Windows that's important to correctly handle long paths.

In addition: I think it is necessary to use `Lstat()` here: If the tag file is a symlink, `Stat()` returns an error if the target does not exist. What do you think?

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+	// Warnings will be generated so that the user is informed that the
+	// indented ignore-action is not performed.
+	f, err := os.Open(tf)
+	if err != nil {
+		Warnf("could not open exclusion tagfile: %v", err)
+		return false
+	}
+	defer f.Close()
+	var buf bytes.Buffer
+	_, err = io.CopyN(&buf, f, int64(len(header)))
```
fd0: This code can be simplified: Use `buf := make([]byte, len(header))` and use `io.ReadFull(f, buf)`.

fd0 (changes requested):
Great work so far! I've added a few comments.

I'm not sure if we need a cache for the directories taht have already been checked though: For each file in a directory that is to be saved, the code here again tests if an exclude file is present. So if we have a directory with 1000 entries, the check is run 1000 times...

fd0:
Also add an entry to the changelog, along the lines of:
```
 * A new option `--exclude-caches` was added that allows excluding cache
   directories (that are tagged as such). This is a special case of a more
   generic option `--exclude-if-present` which excludes a directory if a file
   with a specific name (and contents) is present.
   https://github.com/restic/restic/issues/317
   https://github.com/restic/restic/pull/1170
```

cmd/restic/cmd_backup.go
```diff
@@ -76,6 +79,8 @@ func init() {
 	f.StringArrayVarP(&backupOptions.Excludes, "exclude", "e", nil, "exclude a `pattern` (can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.ExcludeFiles, "exclude-file", nil, "read exclude patterns from a `file` (can be specified multiple times)")
 	f.BoolVarP(&backupOptions.ExcludeOtherFS, "one-file-system", "x", false, "exclude other file systems")
+	f.StringVar(&backupOptions.ExcludeIfPresent, "exclude-if-present", "", "takes filename[:header], exclude contents of directories containing filename (except filename itself) if header of that file is as provided")
+	f.BoolVar(&backupOptions.ExcludeCaches, "exclude-caches", false, `alias for --exclude-if-present "CACHEDIR.TAG:Signature: 8a477f597d28d172789f06886806bc55`)
```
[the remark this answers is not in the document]
fawick: Fair point. Done.

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
 	}
 	return excludes
 }
+
+// FilenameCheck is a function that takes a filename and returns a boolean depending on arbitraty check
```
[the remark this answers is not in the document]
fawick: Done.

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+type FilenameCheck func(filename string) bool
+
+// genExcludeByFile returns a FilenameCheck which itself returns whether a path
+// should be excluded. The FilenameCheck considers a file to be excluded when
+// it resides in a directory with an exclusion file, that is either specified
+// by excludeIfPresent in the form "filename[:content]".  The returned error is
+// non-nil if the filename component of excludeIfPresent is empty.
+func genExcludeByFile(excludeIfPresent string) (FilenameCheck, error) {
```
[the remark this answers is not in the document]
fawick: Done.

cmd/restic/cmd_backup.go
```diff
@@ -522,3 +544,87 @@ func readExcludePatternsFromFiles(excludeFiles []string) []string {
...
+	fn := func(string) bool { return false }
+	tf, tc := "", ""
+	if excludeIfPresent != "" {
```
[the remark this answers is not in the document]
fawick: Of course, you're right, thanks! Done.

cmd/restic/cmd_backup.go
fawick: Done.

cmd/restic/cmd_backup.go
fawick: Thanks for the heads-up regarding `internal/fs`, I wasn't aware of that. Done.
`Lstat` also done.

cmd/restic/cmd_backup.go
fawick: Done.

cmd/restic/cmd_backup.go
fawick: Done.

fd0 (approved):

fd0:
I've cancelled the Darwin build for Travis (it's broken at the moment), the remaining Linux builds completed successfully. Thanks for your work!

mathstuf:
Thanks! Is there a reason that the flag is only allowed once?

fawick:
Yes: I was thinking too small. But worry not, https://github.com/restic/restic/pull/1224 is lifting that clunky restriction.

mathstuf:
Thanks!

fawick:
@mathstuf #1224 was merged last night, so if you pull master you will be able to use --exclude-if-present multiple times.

PR #1174, opened 2017-08-24
Update minimum Go version

pwaring:
Minimum version is now 1.8 according to build.go (from latest master, cloned a few minutes ago):

```
paul@voga:~/third-dev/restic$ go run build.go 
Go version go1.7.4 detected, restic requires at least Go 1.8
exit status 1
```


fd0 (approved):
You're right, great catch! Thanks!

Issue #396, opened 2016-01-25
Manual: Document include/exclude examples for backup and and restore


fd0:
This issue is about adding examples for `--exclude` and `--include` to the manual for the backup and restore commands.

Information is currently hidden in the following issues:
https://github.com/restic/restic/issues/373
https://github.com/restic/restic/issues/226
https://github.com/restic/restic/pull/231
https://github.com/restic/restic/issues/1016
https://github.com/restic/restic/issues/1005

tmcarr:
I'm going to try to produce something for this today.

tmcarr:
@fd0 Is it possible to get this assigned? Just want to make sure work isn't duplicated for Hacktoberfest?

fd0:
Hm, I cannot assign people from outside the restic org, but it's yours. Can you please hold off a little bit, until #1300 is merged? Thanks!

karldw:
These exclude-file comments were really helpful, thanks!  A couple of additional questions that might be worth including:
- Can you have multiple `--exclude-file`s?  `restic help backup` says it can be specified multiple times, the docs say "specify [exclude patterns] with multiple `--exclude`‘s or **one** `--exclude-file" (emphasis added)
- What's the format of the exclude file?  A filter on each line?  Are comments allowed?

fd0:
Yes, you can specify `--exclude-file` multiple times, the results will be merged into one big list of excludes.

One filter pattern per line, when the line starts with `#` it's a comment.

fd0:
The section in the manual should also mention `CACHEDIR.TAG` and link to http://www.brynosaurus.com/cachedir/

ametzler:
How about the attached patch? I think it covers all the major points, much of the wording is copied from your comments in the bug-reports.
[more-examples.diff.txt](https://github.com/restic/restic/files/1584358/more-examples.diff.txt)

fd0:
Would you mind submitting a pull request? :)

ametzler:
Sure, see #1499. HTH. I have read over CONTRIBUTING and I think I have done the necessary. If I missed something please tell me.

Merry christmas, BTW. ;-)

fd0:
Thanks! :)

PR #1184, opened 2017-08-28
Doc cleanup and mention restore include/exclude

prattmic:
The second commit adds a short mention of --include/--exclude to the restore section. There is a lot more detail to be added, but this at least mentions the possibility of restoring individual files.

The first commit cleans up some inconsistencies I noticed along the way.

Updates #396


fd0:
Awesome, thank you!

PR #1187, opened 2017-08-28
internal/crypto: small simplifications

FiloSottile:
* append operates on len, not cap (not a bug since len is set to cap above, but let's avoid the confusion)
* no need to extend ciphertext again to cap after we made it big enough
* make consistent use of ciphertext[:ivSize] vs iv[:]
* make all input problems errors and impossible/catastrophic cases panics

### DIFF
--- a/internal/crypto/crypto.go
+++ b/internal/crypto/crypto.go
@@ -251,21 +251,19 @@ func (k *Key) Encrypt(ciphertext []byte, plaintext []byte) ([]byte, error) {
 
 	// extend ciphertext slice if necessary
 	if len(ciphertext) < len(plaintext)+Extension {
-		ext := len(plaintext) + Extension - cap(ciphertext)
+		ext := len(plaintext) + Extension - len(ciphertext)
 		ciphertext = append(ciphertext, make([]byte, ext)...)
-		ciphertext = ciphertext[:cap(ciphertext)]
 	}
 
 	iv := newIV()
+	copy(ciphertext, iv[:])
+
 	c, err := aes.NewCipher(k.EncryptionKey[:])
 	if err != nil {
 		panic(fmt.Sprintf("unable to create cipher: %v", err))
 	}
-
-	e := cipher.NewCTR(c, iv[:])
-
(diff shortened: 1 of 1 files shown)


internal/crypto/crypto.go
```diff
@@ -299,27 +297,24 @@ func (k *Key) Decrypt(plaintext []byte, ciphertextWithMac []byte) (int, error) {
...
 
 	// decrypt data
 	c, err := aes.NewCipher(k.EncryptionKey[:])
 	if err != nil {
 		panic(fmt.Sprintf("unable to create cipher: %v", err))
 	}
-
-	// decrypt
 	e := cipher.NewCTR(c, iv)
-	plaintext = plaintext[:len(ciphertext)]
```
fd0: What happens if `len(plaintext)` is greater than `len(ciphertext)-Extension`? I think this line needs to stay, what do you think?

fd0 (changes requested):
Hey, thanks a lot for looking at the code and suggesting simplifications! I agree with most of the changes except one.

(I expected you to look at the crypto code if you ever stumble over restic ;)

Would you mind running `gofmt` on the code? Thanks!

internal/crypto/crypto.go
FiloSottile: `XORKeyStream` is defined in that case to only operate on `len(dst)`.

I am, to be honest, not a fan of this API because it does not return the plaintext slice, only the length, and I can see an user failing to slice the plaintext in such a case, leaving trailing unauthenticated data.

FiloSottile:
Hey, dammit, I thought I did `gofmt` by eye, but I was editing from the GitHub editor. I'll clone it and `gofmt` it :)

BTW, I like the design, and indeed I'm reviewing the crypto before adopting it ;) I'm doing a quick writeup, let me know if you want to read a draft.

Thanks for the software!

fd0:
Heh, thanks a lot for the draft, added a few comments, and thanks for looking at the code! I appreciate that!

fd0 (approved):

PR #1191, opened 2017-08-29
debug: properly handle interrupted profiles

prattmic:
By default (i.e., without profile.NoShutdownHook), profile.Start listens
for SIGINT and will stop the profile and call os.Exit(0).

restic already listens for SIGINT and runs its own cleanup handlers
before calling os.Exit(0).

As is, these handlers are racing when an interrupt occurs, and in my
experience, restic tends to win the race, resulting in an unusable
profile.

Eliminate the race and properly stop profiles on interrupt by disabling
package profile's signal handler and instead stop the profile in a
restic cleanup handler.

### DIFF
--- a/cmd/restic/global_debug.go
+++ b/cmd/restic/global_debug.go
@@ -19,10 +19,6 @@ var (
 	memProfilePath      string
 	cpuProfilePath      string
 	insecure            bool
-
-	prof interface {
-		Stop()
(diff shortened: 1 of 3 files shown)


fd0 (approved):
Cool, this is indeed much nicer. Thank you very much for your contribution!

PR #1194, opened 2017-08-29
fix: bucker

mungomat:
We talked about this issue at Froscon. 

bucker "always" does not return a unique id in case of exact same timestamps (see pull request backup_time)

Tobias


internal/restic/snapshot_policy.go
```diff
@@ -90,7 +90,9 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots) {
 		{p.Yearly, y, -1},
 	}
 
+	var nr int
 	for _, cur := range list {
```
fd0: We can use the index from the list here:
```
for nr, cur := range list {
```

Then we don't need another variable.

internal/restic/snapshot_policy.go
```diff
@@ -103,7 +105,7 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots) {
 		// Now update the other buckets and see if they have some counts left.
 		for i, b := range buckets {
 			if b.Count > 0 {
-				val := b.bucker(cur.Time)
+				val := b.bucker(cur.Time, nr)
```
fd0: Good idea, I like this solution!

fd0 (changes requested):
Thanks a lot for submitting this PR. I've added a few small things. The only remaining blocker is that there is no test yet for this special case. Would you mind adding a test?

Please let me know if you need any help.

internal/restic/snapshot_policy.go
```diff
@@ -90,7 +90,8 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots) {
 		{p.Yearly, y, -1},
 	}
 
-	for _, cur := range list {
+	for nr, cur := range list {
+		nr = nr + 1
```
fd0: hm, now `nr` is always incremented, so for the first element of `list`, `nr` is set to `1`. Was that intentional?

fd0 (changes requested):

internal/restic/snapshot_policy.go
mungomat: You are right, we dont need this line

mungomat:
hm, the new tests fail - I think we have to update some policy_keep_snapshots-files?

fd0:
Ah, yes. The idea is that you run the tests once with `-update=true` so the policy files are updated. Then have a look at what has changed (and if that's correct), then commit the files.

fd0:
I've updated the files for you, thanks for your contribution, looks good now!

fd0 (approved):

fd0:
Merged, thanks!

PR #1195, opened 2017-08-29
new sub-option for backup: time

mungomat:
New option to specify the timestamp for a backup - depends on the bugfix_bucker_always

### DIFF
--- a/cmd/restic/global_debug.go
+++ b/cmd/restic/global_debug.go
@@ -19,10 +19,6 @@ var (
 	memProfilePath      string
 	cpuProfilePath      string
 	insecure            bool
-
-	prof interface {
-		Stop()
-	}
 )
 
 func init() {
@@ -54,10 +50,21 @@ func runDebug() error {
 		return errors.Fatal("only one profile (memory or CPU) may be activated at the same time")
 	}
 
+	var prof interface {
+		Stop()
+	}
+
 	if memProfilePath != "" {
-		prof = profile.Start(profile.Quiet, profile.MemProfile, profile.ProfilePath(memProfilePath))
+		prof = profile.Start(profile.Quiet, profile.NoShutdownHook, profile.MemProfile, profile.ProfilePath(memProfilePath))
 	} else if cpuProfilePath != "" {
-		prof = profile.Start(profile.Quiet, profile.CPUProfile, profile.ProfilePath(cpuProfilePath))
+		prof = profile.Start(profile.Quiet, profile.NoShutdownHook, profile.CPUProfile, profile.ProfilePath(cpuProfilePath))
+	}
+
+	if prof != nil {
+		AddCleanupHandler(func() error {
+			prof.Stop()
+			return nil
+		})
 	}
 
 	if insecure {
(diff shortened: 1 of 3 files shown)


fawick:
appveyor and travic fail since the tests can no longer be compiled (`not enough arguments`). It would be great if you could fix the function calls in the test.

mungomat:
appveyor is ok
could not resolve travis conflicts:
"restic manpage --output-dir doc/man" changed manual dates (ok), but it also added some local paths which I did not commit

fawick:
Could you rebase this to the latest master an rerun `restic manpage`? A couple of hours ago the date was changed to always read Jan 2017 in the manpages.

cmd/restic/cmd_backup.go
```diff
@@ -81,6 +82,7 @@ func init() {
 	f.StringArrayVar(&backupOptions.Tags, "tag", nil, "add a `tag` for the new snapshot (can be specified multiple times)")
 	f.StringVar(&backupOptions.Hostname, "hostname", "", "set the `hostname` for the snapshot manually")
 	f.StringVar(&backupOptions.FilesFrom, "files-from", "", "read the files to backup from file (can be combined with file args)")
+	f.StringVar(&backupOptions.TimeStamp, "time", "", "time of the backup (ex. 2012-11-01T22:08:41+00:00) (default: now)")
```
fd0: I'm not sure we should use this rather complicated format. What about using the time format that is used in the `snapshots` command? It would require proper quoting (since there's a space in it) but is much easier to get right.

https://github.com/restic/restic/blob/a3f6bf3e5a07949e0a30f1fd05c6d75affe51c08/cmd/restic/table.go#L46

cmd/restic/cmd_backup.go
```diff
@@ -471,7 +473,16 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 		Warnf("%s\rwarning for %s: %v\n", ClearLine(), dir, err)
 	}
 
-	_, id, err := arch.Snapshot(context.TODO(), newArchiveProgress(gopts, stat), target, opts.Tags, opts.Hostname, parentSnapshotID)
+	timeStamp := time.Now()
+	if opts.TimeStamp != "" {
+		timeStamp, err = time.Parse(time.RFC3339, opts.TimeStamp)
+		if err != nil {
+			Verbosef("wrong time format\n")
+			return err
```
fd0: Please return a proper error message using `errors.Fatalf`, like this:
https://github.com/restic/restic/blob/a3f6bf3e5a07949e0a30f1fd05c6d75affe51c08/cmd/restic/cmd_cat.go#L56

You could also include the requested time format in the error (maybe in a second line?) so that users see how it should look like.

What do you think?

fd0 (changes requested):
I've added a few comments. If you have problems with rebasing this branch on master (switch to master branch, run `git pull`, switch to branch `backup_time` again, then run `git rebase master`) let us know, we can do that for you.

Thanks for solving this!

PR #1196, opened 2017-08-29
forget: group-by-tags-only

mungomat:
This is not complete - what if "restic forget --group-by-tags --group-by-tags-only" is given?

### DIFF
--- a/cmd/restic/cmd_forget.go
+++ b/cmd/restic/cmd_forget.go
@@ -6,8 +6,8 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/restic"
 	"github.com/spf13/cobra"
 )
 
@@ -40,9 +40,9 @@ type ForgetOptions struct {
 	Paths []string
 
 	// Grouping
-	GroupBy		string
-	DryRun          bool
-	Prune           bool
+	GroupBy string
+	DryRun  bool
+	Prune   bool
 }
 
 var forgetOptions ForgetOptions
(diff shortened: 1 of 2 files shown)


fawick:
Travis fails because the manpage is not up-to-date in the PR. In your branch, could you please call `restic manpage --output-dir doc/man` and add this to the PR? Thanks!

fd0:
And it needs a rebase, just like the other PR. Let us know if you need help with the rebase.

middelink:
Can I suggest to resolve this in a different way? Instead of adding --group-by-tags-only and possibly --group-by-path-ony or --group-by-host-only, make a flag specifying what your ordering needs to be --group-by=[host,path,tags]. Default to "host,path,tags" to keep functionality the same.

fd0:
@middelink that is an excellent idea, thank you very much! We can just fill the JSON key according to what's present in `--group-by`.

@mungomat Do you think you can tackle this? Or shall I do that?

fd0:
The CI test fail because `cmd_forget.go` is not properly formatted. May I suggest installing the script
`fmt-check` from https://github.com/edsrzf/gofmt-git-hook locally as a
pre-commit hook? It checks formatting before committing automatically, just copy
this script to `.git/hooks/pre-commit`.

cmd/restic/cmd_forget.go
```diff
@@ -91,6 +92,14 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
 	}
 	snapshotGroups := make(map[string]restic.Snapshots)
 
+	var GroupByTag      bool
+	var GroupByHost     bool
+	var GroupByPath     bool
+
+	GroupByTag = strings.Contains( opts.GroupBy, "tag" )
```
fd0: That is very vague. I'd go for splitting `opts.GroupBy` by `,`, then iterate over the results, and use a `switch` statement that reacts on the strings `tag`, `host` and `path` and returns an error if the string is unknown.

What do you think?

fd0 (changes requested):

fd0 (approved):

fd0:
Good work, thanks!

PR #1198, opened 2017-08-30
Add env variable to force progress interval

icgood:
Adds optional RESTIC_PROGRESS_INTERVAL environment variable to force interval'ed progress output during non-terminal backups.

----

The use-case for this is a GUI that runs `restic backup` and wants to render the progress. It could send `USR1` signals, unless it runs the backup as root -- then every signal would need to be sent as root as well.

### DIFF
--- a/internal/restic/progress.go
+++ b/internal/restic/progress.go
@@ -3,6 +3,7 @@ package restic
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


icgood:
Sounds like work/thought has gone into this already, I'll see where it plays out in other issues or PRs.

fd0:
Thanks for trying! :)

PR #1201, opened 2017-08-31
Fixed word-o. ("Package", not "packet".)

gjmf:
Fixed a word-o. homebrew is a package manager, not a packet manager. :)


fd0:
Thanks! Even such small contributions help :)

fd0:
Thanks! Good catch!

PR #1203, opened 2017-09-02
Handle SIGINFO on all supported platforms

valpackett:
Not just darwin (#1082)


gregf:
Correct me if I'm wrong, but doesn't Linux support this to? I think they are using this functionality in the newer coreutils for dd. If you pass dd progress=status and hit ^t you get a progress update.

myfreeweb:
@gregf I guess your terminal is set up to send `SIGUSR1` on Ctrl-T. Linux does not have `SIGINFO`.

gregf:
Interesting, thanks for clearing that up.

fd0:
Looks great!

PR #1209, opened 2017-09-04
Resolve name collisions

### LINKED ISSUE #549: Unable to backup/restore files/dirs with same name
## Output of `restic version`

restic 0.1.0
compiled at 2016-07-20 12:42:43 with go1.6.3
## Expected behavior

After restore all directories should be restored.
## Actual behavior

Only one directory is restored.
1. Create files

/tmp/restic/FILESNEW01/Dir01/Test01.txt
/tmp/restic/FILESNEW01/Dir01/Test02.txt
/tmp/restic/FILESNEW01/Dir01/Test03.txt

/tmp/restic/FILESNEW02/Dir01/Test01.txt
/tmp/restic/FILESNEW02/Dir01/Test02.txt
/tmp/restic/FILESNEW02/Dir01/Test03.txt

content of files:
**`cat /tmp/restic/FILESNEW01/Dir01/Test0*`**
`Content file. /tmp/restic/FILESNEW01/Dir01/Test01.txt`
`Content file. /tmp/restic/FILESNEW01/Dir01/Test02.txt`
`Content file. /tmp/restic/FILESNEW01/Dir01/Test03.txt` 

**`cat /tmp/restic/FILESNEW02/Dir01/Test0*`**
`Content file. /tmp/restic/FILESNEW02/Dir01/Test01.txt`
`Content file. /tmp/restic/FILESNEW02/Dir01/Test01.txt`
`Content file. /tmp/restic/FILESNEW02/Dir01/Test03.txt`

I want backup
- /tmp/restic/FILESNEW01/Dir01/
- /tmp/restic/FILESNEW02/Dir01/

Commands:
Initiate repository in /tmp/restic/BACKUP directory
- **restic -r /tmp/restic/BACKUP/ init**

Make backup 
-  **restic backup /tmp/restic/FILESNEW01/Dir01 /tmp/restic/FILESNEW02/Dir01 -r /tmp/restic/BACKUP/**
  
  `scan [/tmp/restic/FILESNEW01/Dir01 /tmp/restic/FILESNEW02/Dir01]`
  `scanned 2 directories, 6 files in 0:00`
  `[0:00] 16.67%  0B/s  51B / 306B  0 / 8 items  0 errors  ETA 0:00
  duration: 0:00, 0.01MiB/s`
  `snapshot 4d197b90 saved`

Checking if backup exists in repository
- **restic -r /tmp/restic/BACKUP/ snapshots**

`ID        Date                 Host        Directory`

`4d197b90  2016-07-26 14:14:43  nebss       /tmp/restic/FILESNEW01/Dir01
                                           /tmp/restic/FILESNEW02/Dir01`

Restore backup
- **restic -r /tmp/restic/BACKUP/ restore 4d197b90  -t /tmp/restic/RESTORE/**

`restoring <Snapshot 4d197b90 of [/tmp/restic/FILESNEW01/Dir01 /tmp/restic/FILESNEW02/Dir01] at 2016-07-26 14:14:43.208840145 +0300 EEST> to /tmp/restic/RESTORE/`

Checking if directories/files exists
- **ls /tmp/restic/RESTORE/**
  `Dir01`
- **cat /tmp/restic/RESTORE/Dir01/Test0***
  `Content file. /tmp/restic/FILES01/Dir01/Test01.txt`
  `Content file. /tmp/restic/FILES01/Dir01/Test02.txt`
  `Content file. /tmp/restic/FILES01/Dir01/Test03.txt`

### LINKED ISSUE #1179: restic hides and destroys directories with the same name (data is lost)
I might be doing something just stupid but I guess this is what users generally do.

If I backup two different paths that have the same final directory name, restic simply backups the first one and ignores all the other ones. Or better, it do save them, but all commands that could read data ignore them and prune gets rid of them.

## Output of `restic version`
```
restic 0.7.1
```

```
echo 123 > restic.key
mkdir repo
mkdir -p home/test1/data
date > home/test1/data/testfile1.txt
mkdir -p home/test2/data
date > home/test2/data/testfile2.txt
mkdir -p home/test3/data
date > home/test3/data/testfile3.txt
restic -p restic.key -r repo init
...

restic -p restic.key -r repo backup home/test1/data home/test2/data home/test3/data
scan [/root/a/home/test1/data /root/a/home/test2/data /root/a/home/test3/data]
scanned 3 directories, 3 files in 0:00
[0:00] 100.00%  0B/s  90B / 90B  6 / 6 items  0 errors  ETA 0:00 
duration: 0:00, 0.00MiB/s
snapshot 8bfeec2b saved
```
Until here, everything was OK. It saved 6 items (3 directories and 3 files).
Wait, 3 directories? Wasn't it supposed to be 6 directories?

```
restic -p restic.key -r repo snapshots
ID        Date                 Host        Tags        Directory
8bfeec2b  2017-08-26 00:52:00  zipper              ┌── /root/a/home/test1/data
                                                   │   /root/a/home/test2/data
                                                   └── /root/a/home/test3/data
```
Still OK.

```
restic -p restic.key -r repo ls latest
snapshot 8bfeec2b of [/root/a/home/test1/data /root/a/home/test2/data /root/a/home/test3/data] at 2017-08-26 00:52:00.186855416 -0300 -03):
/data
/data/testfile1.txt
```

I guess that wasn't what I really asked it to backup. Did it really saved only data from test1?

```
restic -p restic.key -r repo find testfile1.txt
Found matching entries in snapshot 8bfeec2b4574226f411df235a6ef51d26da8ee635dd670d862950bce8206a371
```

testfile1.txt is there, but testfile2.txt ...

```
restic -p restic.key -r repo find testfile2.txt
<nothing returned!>
```

At least, mount, ls, find and specially, restore, never listed testefile2.txt or testfile3.txt.
Let's debug:

```
restic -p restic.key -r repo list blobs
tree ab1808987f20ebf517b2f2b906d73b74f0be79d701786979f2f525f7281d902e
data 711db6309812a19a5535d1b4c2302ced3e2f4f1d656b39ff0c088d4eabfe7b1a
data 7ba9ce1f2c12857a31b629c2c1f9826a0b45147900024534159cb908a2576737
tree c67214e81e8380335629696f2056575dc324ae2980f97098a2219c29e268a57e
tree 6b257a61f0197ff8ca9a82b1358dcedf08b9ec6d1744d28348ef992b75da2bf8
tree d147d18cfeeac8146d00e1f98925b29e179193b54561983a77c244b53dbd0d4b
```

Four dirs and 2 files? Maybe two of them are equal?

```
md5sum home/test*/*/*
7dd0c1b72dc0f35688241f7ac4e42504  home/test1/data/testfile1.txt
7dd0c1b72dc0f35688241f7ac4e42504  home/test2/data/testfile2.txt
42b2d1966ff6645925190b3db15b32e9  home/test3/data/testfile3.txt
```

Yes, it did deduplicated it nicely. Let's show all of them:

```
for blob in $(restic -p restic.key -r repo list blobs | awk '{ print $2 }'); do restic -p restic.key -r repo cat blob $blob; done
{"nodes":[{"name":"data","type":"dir","mode":2147484141,"mtime":"2017-08-26T00:51:50.489170628-03:00","atime":"2017-08-26T00:52:00.185075187-03:00","ctime":"2017-08-26T00:51:50.489170628-03:00","uid":0,"gid":0,"inode":12486783,"device_id":23,"content":null,"subtree":"d147d18cfeeac8146d00e1f98925b29e179193b54561983a77c244b53dbd0d4b"}]}
sáb ago 26 00:51:50 -03 2017
sáb ago 26 00:51:49 -03 2017
{"nodes":[{"name":"testfile2.txt","type":"file","mode":420,"mtime":"2017-08-26T00:51:49.713178185-03:00","atime":"2017-08-26T00:52:00.185075187-03:00","ctime":"2017-08-26T00:51:49.713178185-03:00","uid":0,"gid":0,"inode":12486781,"device_id":23,"size":30,"links":1,"content":["7ba9ce1f2c12857a31b629c2c1f9826a0b45147900024534159cb908a2576737"]}]}
{"nodes":[{"name":"testfile1.txt","type":"file","mode":420,"mtime":"2017-08-26T00:51:49.709178223-03:00","atime":"2017-08-26T00:52:00.189075147-03:00","ctime":"2017-08-26T00:51:49.709178223-03:00","uid":0,"gid":0,"inode":12486778,"device_id":23,"size":30,"links":1,"content":["7ba9ce1f2c12857a31b629c2c1f9826a0b45147900024534159cb908a2576737"]}]}
{"nodes":[{"name":"testfile3.txt","type":"file","mode":420,"mtime":"2017-08-26T00:51:50.49317059-03:00","atime":"2017-08-26T00:52:00.189075147-03:00","ctime":"2017-08-26T00:51:50.49317059-03:00","uid":0,"gid":0,"inode":12486784,"device_id":23,"size":30,"links":1,"content":["711db6309812a19a5535d1b4c2302ced3e2f4f1d656b39ff0c088d4eabfe7b1a"]}]}
```

Everything is correctly saved, except that there is only one 'data' directory. And finally, let's try prune:

```
restic -p restic.key -r repo prune
counting files in repo
building new index for repo
[0:00] 100.00%  1 / 1 packs
repository contains 1 packs (6 blobs) with 1.842 KiB bytes
processed 6 blobs: 0 duplicate blobs, 0B duplicate
load all snapshots
find data that is still in use for 1 snapshots
[0:00] 100.00%  1 / 1 snapshots
found 3 of 6 data blobs still in use, removing 3 blobs
will remove 0 invalid files
will delete 0 packs and rewrite 1 packs, this frees 818B
[0:00] 100.00%  1 / 1 packs rewritten
counting files in repo
[0:00] 100.00%  1 / 1 packs
finding old index files
saved new index as be5f649a
remove 1 old index files
[0:00] 100.00%  1 / 1 packs deleted
done
```

There goes all my files that "restic ls" didn't list...

I guess this is **really** not expected.

fd0:
Resolve name collisions

At the moment when two items to be saved have the same directory name,
restic only saves the first one to the repo. Let's say we have a
structure like this:

    dir1
    └── subdir
        └── file
    dir2
    └── subdir
        └── file

When restic is run on `dir1/subdir` and `dir2/subdir`, it will only save
the first `subdir`:

    $ restic backup dir1/subdir dir2/subdir
    [...]

    $ restic ls -l latest
    drwxr-xr-x  1000   100      0 2017-08-27 20:56:39 /subdir
    -rw-r--r--  1000   100     17 2017-08-27 20:56:39 /subdir/file

That's obviously a bad thing, caused by an early decision to strip the
full path to the files/dirs to save and only leave the last directory.

This commit partly resolves this by handling colliding names and
resolving the conflicts. Restic will now append a counter to the file
(`-123`) until the conflict is resolved. So in the example above, we'll
end up with the following structure:

    $ restic ls -l latest
    drwxr-xr-x  1000   100      0 2017-08-27 20:56:39 /subdir
    -rw-r--r--  1000   100     17 2017-08-27 20:56:39 /subdir/file
    drwxr-xr-x  1000   100      0 2017-08-27 20:56:46 /subdir-1
    -rw-r--r--  1000   100     17 2017-08-27 20:56:46 /subdir-1/file

This partly addresses #549 and closes #1179.

At first I thought that the obvious correction would be to archive the
full path. But it turns out that collisions may still occur: Suppose you
have a file named `foo` in the current directory, and the parent directory
also contains a file `foo`. Archiving these with restic also causes a
collision, since restic strips the `../` from the first file:

    $ restic backup ../foo foo

This also happens with `tar`, which does not handle the collision and
will happily archive two files called `foo`.

So, the best way forward is to handle name collisions and archive the
whole path. The latter will be tackled in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ Important Changes in 0.X.Y
    included) in a restore, they are not loaded from the repo any more.
    https://github.com/restic/restic/pull/1044
 
+ * Name collisions are now resolved by appending a counter.
+   https://github.com/restic/restic/issues/1179
+   https://github.com/restic/restic/pull/1209
+
 
 Small changes
 -------------
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -383,7 +383,22 @@ func (arch *Archiver) dirWorker(ctx context.Context, wg *sync.WaitGroup, p *rest
 						panic("invalid null subtree restic.ID")
 					}
 				}
-				tree.Insert(node)
+
+				// insert node into tree, resolve name collisions
+				name := node.Name
+				i := 0
+				for {
+					i++
+					err := tree.Insert(node)
+					if err == nil {
+						break
+					}
+
+					newName := fmt.Sprintf("%v-%d", name, i)
+					fmt.Fprintf(os.Stderr, "%v: name collision for %q, renaming to %q\n", filepath.Dir(node.Path), node.Name, newName)
+					node.Name = newName
+				}
+
 			}
 
 			node := &restic.Node{}
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -310,5 +313,56 @@ func TestArchiveEmptySnapshot(t *testing.T) {
 
 	if sn != nil {
 		t.Errorf("expected null snapshot for empty snapshot, got %v", sn)
+	}
+}
+
+func chdir(t testing.TB, target string) (cleanup func()) {
+	curdir, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("chdir to %v", target)
+	err = os.Chdir(target)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return func() {
+		t.Logf("chdir back to %v", curdir)
+		err := os.Chdir(curdir)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestArchiveNameCollision(t *testing.T) {
+	repo, cleanup := repository.TestRepository(t)
+	defer cleanup()
+
+	dir, cleanup := TempDir(t)
+	defer cleanup()
+
+	root := filepath.Join(dir, "root")
+	OK(t, os.MkdirAll(root, 0755))
+
+	OK(t, ioutil.WriteFile(filepath.Join(dir, "testfile"), []byte("testfile1"), 0644))
+	OK(t, ioutil.WriteFile(filepath.Join(dir, "root", "testfile"), []byte("testfile2"), 0644))
+
+	defer chdir(t, root)()
+
+	arch := archiver.New(repo)
+
+	sn, id, err := arch.Snapshot(context.TODO(), nil, []string{"testfile", filepath.Join("..", "testfile")}, nil, "localhost", nil)
+	OK(t, err)
+
+	t.Logf("snapshot archived as %v", id)
+
+	tree, err := repo.LoadTree(context.TODO(), *sn.Tree)
+	OK(t, err)
+
+	if len(tree.Nodes) != 2 {
+		t.Fatalf("tree has %d nodes, wanted 2: %v", len(tree.Nodes), tree.Nodes)
 	}
 }


rawtaz:
This is great stuff. Clean and simple and the expected way for it to be dealt with when there's conflicting file/directory names. Also will only happen on the first level in the restore, which is already a place where you know you'll have duplicate names to address anyway.

luizluca:
@fd0 ,

The fix helps when there is a collision but it looks like it treats the "collision" as temporary situation. It's better than previous behavior as it does not ignores/removes files from backup but it will still confuses the restore. Which subdir applies to which path? It might get untreatable. For example, image a user wants to backup its /etc/ directory. However, it uses opensuse that uses btrfs and snapshots periodically root fs. The user could ask restic to backup

```
restic backup /.snapshots/*/snapshot/etc
```

Now I'll get a bunch of etc-xxx. I guess restic would even reuse the counter (or even /etc) when the snapshot that used it get pruned. So, a new /etc could be newer than /etc-1. I was just a cheap example. It could be many more.

I guess the correct solution would be to expand any relative path and save the full/absolute path. Go-lang libraries could do it for you.

In the example you mentioned,

```
machine1:/xxx/yyy $ restic backup ../foo foo
```

Should never collide as they should be saved as:

* machine1:/xxx/foo
* machine1:/xxx/yyy/foo

I didn't get the idea why using only dirname was a good solution. What was the situation it was trying to solve?

fd0:
You're right, this PR does not solve the issue completely. The second part (which will be done in a separate PR) will change the code so that the complete path (if available) is saved.

In this your case, running

    $ restic backup /.snapshots/*/snapshot/etc

will yield a snapshot that contains:

    .snapshots/foo/snapshot/etc
    .snapshots/bar/snapshot/etc
    [...]

So there's no collision any more. That's what `tar` does, and it works very well.

The problem with the other example is that determining what `..` means is not so easy as it may sound. So we'll probably keep the current behavior: When restic is instructed to save `../foo` and `foo`, it will still rename the second `foo` to `foo-1`. Users are warned during backup that this happened, and they can easily remedy the situation by providing an absolute path, e.g. `/home/user/foo` and `/home/user/work/foo` so the collision does not occur.

Another option that I'd like to implement (again, similar to `tar`) is `--strip-components`, so that a number of path elements are removed. Your example again:

    $ restic backup --strip-components=1 /.snapshots/*/snapshot/etc

will yield a snapshot that contains:

    foo/snapshot/etc
    bar/snapshot/etc
    [...]

But when users then instruct restic to strip e.g. three path components, collisions happen again (several `etc` directories at the top level).

As far as I can see there are several options what restic can do:
 * Bail out, print an error message, exit with an error code
 * Silently archive several directories with the same name at the same level. That's possible in the data structures we use.
 * Warn the use, archive the files, but make sure the names are unique.

I don't like the first option, it's not user friendly. The second option is bad because it makes restore very awkward: Which `etc` directory should be restored? We'd need to find a way (maybe with include/exclude filters?) to address a specific directory when there are several with the same name. That's not good.

The third option is what I like best so far, and that's what we're going to build (at least that's the plan).

luizluca:
@fd0 , this is the only (known) issue that prevent me from using restic. The fullpath feature seems to be a very long-time requested feature. Isn't just a matter of storing all the path passed as arg in the field where today only the directory name is stored?

If / or . are special character, simply replace it with any other character (like '-' or \1). It's hacky but way better than simply append a different number for each conflicted backup. If the user asked to backup '../../etc', warn the user, replace ".." with "__" and go on, storing it as __-__-etc. I'll give you time to refactor the repo structure for a long-term solution. Users need a deterministic "backup path", not dir-1, dir-2, dir-3... that depends on the order of args. If dir-2 is gone, dir-3 becomes dir-2. It's a maintenance nightmare. It would be better to only apply this kind of replace-string solution where it is needed. If repo can store ../../etc or /etc/, save it just like the user passed. Leave the hacks for where it is really needed (fuse mount?)

A feature like '--fullpath' that resolves relative path into full paths is an needed feature for any serious backup solution. I think that it is the way to go. Keep the "last directory name" as default behavior for now if you wish but allow the user to ask restic to resolve the path (I still prefer to simply stores what user passed as arg). Any language library allow to convert relative path into fullpath for existing paths, and so does go-lang. You can easily resolve any path in unix with "readllnk -f relative-or-absolute-path", even with non-existing path (-m).

fd0:
FYI: I'm already working on this.

PR #1210, opened 2017-09-04
filter: document recursive wildcards

prattmic:
Match/ChildMatch accept a ** pattern which is not noted in the doc
string, nor do any of the docs or tests specify whether the match is
greedy (i.e., can 'foo/**/bar' match paths with additional intermediate
bar directories?).

Add a note to the doc string and add test cases for greedy matches.


fd0:
Thanks!

PR #1223, opened 2017-09-09
Add --compact option to snapshots

### LINKED ISSUE #1218: Compact snapshots view
It would be nice if the `snapshots` command had an option to give a compact view of the snapshots. Currently, it lists all the directories passed to the `backup` command when the snapshot was made. When there is a dozen of these in each snapshot, the screen real estate is very quickly wasted. If there was a `--compact` option that listed just the first line, without the directory list, it would be much easier to scan what's available in the repository.

mrzv:
With --compact, snapshots doesn't list directories and puts all tags on a single
line. This way each snapshot takes up exactly one line. Fixes #1218.


fd0 (approved):
Good idea actually!

PR #1224, opened 2017-09-09
Allow multiple exclude-if-present

fd0:


cmd/restic/exclude.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/filter"
+	"github.com/restic/restic/internal/fs"
+)
+
+// RejectFunc is a function that takes a filename and os.FileInfo of a
+// file that would be included in the backup. The function returns true if it
+// should be excluded (rejected) from the backup.
+type RejectFunc func(filename string, fi os.FileInfo) bool
```
fawick: I'd rather `s/filename/path/` here

fawick (approved):
Great code, as always. I remember from FrOSCon that you were aiming to unify all the RejectFuncs, and it turned out great. I especially like how selectFilter shrank to a very simple loop. The only issue I had is bikeshedding over parameter names.

cmd/restic/exclude.go
fd0: Works for me :)

PR #1232, opened 2017-09-13
doc: remove broken link

prattmic:
The link to SFTP is broken, remove it.

I don't even bother to put a new link, since SFTP is literally the next section.


fd0 (approved):
Fine with me, thanks for taking the time to correct this!

Issue #533, opened 2016-06-12
Support unattended encrypted backups without key disclosure


robbat2:
## Expected behavior

restic should allow unattended encrypted backups, without disclosing any key material that can lead to decrypting backups. This does tie into #187 as asymmetric keys can be used to general session keys for example.
## Actual behavior

restic requires a password or unsecure key material to do a new backup.

fd0:
Hi, thanks for suggesting this feature. Can you please elaborate a bit on how this issue is different than #187?

robbat2:
Asymmetric keys are only one of the ways this can be accomplished. Another is immediately pushing the backup to write-only destination (eg a S3 bucket where you can put new content but not read or modify anything already present). Asymmetric key usage (eg generate a session key for the data, use GPG or RSA to protect that key) do ensure that the attacker can't decrypt the backups if they do get their hands on them.

anarcat:
couldn't we implement this by simply storing the password in a file and use `--password-file`? that file could then be encrypted with `$WHATEVER` mechanism you prefer. for example, to encrypt the password file with GnuPG, you could do:

```bash
restic backup --password-file <(gpg --decrypt keyfile.gpg)
```

then of course, it's only as unattended as gpg will allow it, but at least there's some level of automation there... i guess that storing the password in standard keyrings (e.g. gnome-keyring and so on) might be another solution...

but yes, #187 would probably be better in the long term.

jcollie:
The ability to use HashiCorp's Vault https://www.vaultproject.io/ to store the password would be awesome!

osamuaoki:
For convenience, I like to see Gnome keyring supported :-)

FYI: I think the C code  "char *msmtp_password_callback ..." in src/msmtp.c in the source of msmtp gives good example for how to implement such.  It also has code for MAC and WINDOWS equivalents. 
 See code http://msmtp.sourceforge.net/

robbat2:
@osamuaoki there are multiple implementations already for accessing keyring: https://github.com/tmc/keyring

osamuaoki:
Oh, great!  Let's use the existing golang code!

Thanks for the pointer


2017/10/27 午前3:04 "Robin H. Johnson" <notifications@github.com>:

@osamuaoki <https://github.com/osamuaoki> there are multiple
implementations already for accessing keyring:
https://github.com/tmc/keyring

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
[...]

PR #1245, opened 2017-09-16
add explanation of restic automation

anarcat:
every time i look at restic, i block on this and figured it may be
useful for others

this will at least partly address #533.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,7 +19,7 @@ import os
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = ['sphinx.ext.extlinks']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
@@ -104,3 +104,7 @@ html_static_path = ['_static']
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'resticdoc'
+
+extlinks = {
+    'issue': ('https://github.com/restic/restic/issues/%s', '#'),
+}
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -30,7 +30,7 @@ cause of this bug is not yet known.
 How can I specify encryption passwords automatically?
 -----------------------------------------------------
 
-When you run ``restict create``, you need to enter the passphrase on
+When you run ``restic backup``, you need to enter the passphrase on
 the console. This is not very convenient for automated backups, so you
 can also provide the password through the ``--password-file`` option
 or ``RESTIC_PASSWORD`` environment. A discussion is in progress over
@@ -43,6 +43,9 @@ implementing unattended backups happens in :issue:`533`.
                directly and they will be readable to all users on a
                system. Using export in a shell script file should be
                safe, however, as the environment of a process is
-               `accessible only to that user`_.
+               `accessible only to that user`_. Please make sure that
+               the permissions on the files where the password is
+               eventually stored are safe (e.g. `0600` and owned by
+               root).
 
 .. _accessible only to that user: https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -26,3 +26,23 @@ The message means that there is more data stored in the repo than
 strictly necessary. With high probability this is duplicate data. In
 order to clean it up, the command ``restic prune`` can be used. The
 cause of this bug is not yet known.
+
+How can I specify encryption passwords automatically?
+-----------------------------------------------------
+
+When you run ``restict create``, you need to enter the passphrase on
+the console. This is not very convenient for automated backups, so you
+can also provide the password through the ``--password-file`` option
+or ``RESTIC_PASSWORD`` environment. A discussion is in progress over
+implementing unattended backups happens in :issue:`533`.
+
+.. important:: Be careful how you set the environment; using the env
+               command, a `system()` call or using inline shell
+               scripts (e.g. `RESTIC_PASSWORD=password borg ...`)
+               might expose the credentials in the process list
+               directly and they will be readable to all users on a
+               system. Using export in a shell script file should be
+               safe, however, as the environment of a process is
+               `accessible only to that user`_.
+
+.. _accessible only to that user: https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009


doc/conf.py
```diff
@@ -104,3 +104,7 @@
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'resticdoc'
+
+extlinks = {
+    'issue': ('https://github.com/restic/restic/issues/%s', '#'),
```
fd0: Nice, I didn't know that this is possible. Is it compatible with readthedocs?

doc/faq.rst
```diff
@@ -26,3 +26,23 @@ The message means that there is more data stored in the repo than
 strictly necessary. With high probability this is duplicate data. In
 order to clean it up, the command ``restic prune`` can be used. The
 cause of this bug is not yet known.
+
+How can I specify encryption passwords automatically?
+-----------------------------------------------------
+
+When you run ``restict create``, you need to enter the passphrase on
```
fd0: Here's a typo: `s/restict/restic/`, and I suspect you mean `restic backup`?

doc/faq.rst
```diff
@@ -26,3 +26,23 @@ The message means that there is more data stored in the repo than
...
+the console. This is not very convenient for automated backups, so you
+can also provide the password through the ``--password-file`` option
+or ``RESTIC_PASSWORD`` environment. A discussion is in progress over
+implementing unattended backups happens in :issue:`533`.
+
+.. important:: Be careful how you set the environment; using the env
+               command, a `system()` call or using inline shell
+               scripts (e.g. `RESTIC_PASSWORD=password borg ...`)
+               might expose the credentials in the process list
```
fd0: Is that the case, are you sure? How does that work? Do you have a sample for me?

fd0 (changes requested):

doc/conf.py
anarcat: yes, it is.

doc/faq.rst
anarcat: oops, confused with borg, sorry, will reword.

doc/faq.rst
anarcat: i must admit i coped this verbatim from the borg warning. but it fits with my experience. e.g.

```
$ env FOO=secretkey sleep 1 & sudo -u nobody ps axfue | grep -q FOO=secretkey && echo oops. secret leaked.
[2] 18143
oops. secret leaked.
```

calls with `system("FOO=secret bar")` will also be a problem, as this will call `sh -c "FOO=secret bar"` which will show up in the process list.

fd0: Hm. Your example does not work: The `grep -q` returns exit code 0 because it sees itself in the process list, not the `env`:
```
$ env FOO=secretkey sleep 10 & sudo -u nobody ps axfue | grep FOO
[1] 14446
fd0      14448  0.0  0.0  12432  2420 pts/8    S+   20:37   0:00          \_ grep FOO
```

The program `env` will have the secret value in its command line, which is world readable, so I agree that in principle this problem exists. But an instance of `env` is extremely short-lived: It just sets the environment variables from the arguments, then runs `exceve()` (or something similar) and the old process is replaced by the new one, and the entry in `/proc/PID/cmdline` is also replaced. At least that's what I just verified with `gdb`.

The case with `system()` I just tried to reproduce with ruby/irb, and for `system("FOO=x sleep 500")` I was unable to see the intermediate `sh -c`: The `sleep` binary was a direct child to the `irb` process.

Please don't get me wrong: I appreciate your contribution and the added FAQ entry, but I'd like to make sure that this warning is really accurate and we don't scare users needlessly.

anarcat: i can't speak about ruby as i'm less familiar with it, but i can reproduce with a Python `system()` call, however. the faulty script:

    python -c 'import os; os.system("RESTIC_PASSWORD=secretkey sleep 1")'

the sniffer that catches it:

```
#!/usr/bin/perl -w

use Proc::ProcessTable;

$| = 1;

my $i = 0;
my $start = time();
while (1) {
  $i++;
  my $delta = time() - $start;
  if ($delta > 0) {
    printf("\r%d loops per second", ($i/$delta));
    $start = time();
    $i = 0;
  }
  my $t = new Proc::ProcessTable;
  foreach my $p ( @{$t->table} ){
    if ($p->cmndline =~ m/RESTIC_PASSWORD=(\w+)/) {
      printf("\npass: $1, cmd: %s\n", $p->cmndline);
    }
  }
}
```

output of the sniffer:

```
$ ./sniff-cli.pl
66 loops per second
pass: secretkey, cmd: python -c import os; os.system("RESTIC_PASSWORD=secretkey sleep 1")

pass: secretkey, cmd: python -c import os; os.system("RESTIC_PASSWORD=secretkey sleep 1")

pass: secretkey, cmd: sh -c RESTIC_PASSWORD=secretkey sleep 1

pass: secretkey, cmd: python -c import os; os.system("RESTIC_PASSWORD=secretkey sleep 1")

pass: secretkey, cmd: sh -c RESTIC_PASSWORD=secretkey sleep 1

[... repeated about 80 times ...]
63 loops per second^C
```

the workaround, in Python, is of course to set `os.environ` directly and call the `subprocess` module with `shell=False`. i think the above covers both the `system()` and "inline shell script" descriptions.

i must admit i am having trouble reproducing with `env`: it's pretty fast at doing that execve so it's hard to catch it in the act. my attack script can only leverage about 60Hz (e.g. 60 samples of the process list per second) which is very low - i suspect there are more efficient ways of doing this that *would* catch `env` in the act because it *is* passing through the process list. "extremely short lived" is not good enough for something that leaks through the process list, in my opinion. the risk is low, but the impact is so catastrophic that i think such practices should be discouraged.

note the sniffer is in perl for historical reasons, but I also wrote a Python version to see if it would be faster - using some bypassing tricks, i can go up to about 160Hz, but i still can't hit the `env` calls.

anarcat: i obsessed over this a little more and wrote this dumb C program that achieves better performance:

```
#include <proc/readproc.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

int main(int argc, char *argv[]){
  pid_t* pidlist;
  int flags;
  PROCTAB* ptp;
  static proc_t buf;
  int i, j;
  time_t t0, t1;
  double diff;
  flags = PROC_FILLCOM;
  time(&t0);
  for (i = 0; 1; i++) {
    time(&t1);
    diff = (int) difftime(t1, t0);
    if (diff > 0) {
      printf("\r%0.0f loops per second", (i/diff));
      fflush(stdout);
      time(&t0);
      i = 0;
    }
    ptp = openproc(flags, pidlist);
    while (readproc(ptp, &buf)) {
      for (j = 0; buf.cmdline && buf.cmdline[j]; j++) {
        if (strstr(buf.cmdline[j], "RESTIC_PASSWORD")) {
          printf("\n%s\n", buf.cmdline[j]);
          exit(0);
        }
      }
    }
    closeproc(ptp);
  }
}
```

this gets between 700 and 900 Hz (ie. loop iterations per second), and it *does* catch `env`, *if I run `env` in a tight loop like this*:

```
while true; do
    /bin/env RESTIC_PASSWORD=secret true
done
```

now. i know what you'll say: this is unfair to poor `env`, we're mistreating it totally. but consider this: even if I run the loop like this:

```
while true; do
    /bin/env RESTIC_PASSWORD=secret sleep 10
done
```

... it *eventually* finds it. on my first try, when trying this, it found it in 3 seconds (so on the first try!):

```
$ time ./sniff-cli
1001 loops per second
RESTIC_PASSWORD=secret
0.43user 2.07system 0:02.52elapsed 99%CPU (0avgtext+0avgdata 2608maxresident)k
0inputs+0outputs (0major+143minor)pagefaults 0swaps
```

pure coincidence of course. in general, it's going to take a few attempts to get it. for example, on my second try, it took 53 seconds to catch the env call... then on the third try, 43 seconds. so i'd say i've got around 10-20% chance of hitting the password, based on that very low sample. it's very thing, but it's non-zero.

now, arguably, this attack require arbitrary code execution from another user, using a dedicated CPU rolling full speed, something which could be detected.

but i think this is something we should be worried about for something as critical as a crypto passphrase. just don't shove this on the commandline. environment is okay, as the comment says, but commandline is not.

(hopefully, this is the last time I need to explain this to anyone - every time i need to build a more optimized version of the code as CPUs speed up... i guess that, eventually, i'll just lose that battle completely :p)

anarcat: oh, and to complete our little programming language tour, the motivated reader may want to write their own version in golang, with coroutines coordinating the listing and inspection of the `cmdline` field to max out CPUs.

in the end, on linux, the best we can do is iterate real hard on `/proc` - it's what ps and `readproc(2)` is doing anyways... anything better would need root/kernel module access, at which point the game is over anyways.

fd0: You convinced me to keep this paragraph as is, great work! ;)

donat-b: How about saving password in a file with rights 600 owned by user which runs the restic process? Doesn't it completely eliminate the said problem?

doc/faq.rst
anarcat: done.

doc/faq.rst
anarcat: @donat-b i added a comment about safe file permissions. it does resolve the problem, which is why we warn about passing environment on the commandline (and not a password file).

now what *i* would like to see is the possibility of encrypting said password file with GPG or some other external command. that would neatly solve #533 at low implementation cost... but that belongs in that other thread i guess. :)

fd0 (approved):
LGTM

PR #1250, opened 2017-09-17
travis: update go versions

dvrkps:


zcalusic:
I tried this already in #1012, but unfortunately versions of Go Travis uses in that case are not the latest one. :(

dvrkps:
Please look again, i think that travis use correct versions now.

zcalusic:
Yeah, judged by my other project, the versions are now up to date. We can only hope it stays that way, otherwise 1.x.y notation doesn't make much sense.

fd0 (approved):
Thanks!

Issue #1137, opened 2017-07-25
Add advice to run `restic check` regularly


fd0:
Followup issue to #1136:
 * [ ] Add a few prominent advice texts to the restic manual to regularly run `restic check`
 * [ ] Add messages to the code that prints something along the lines of "please remember: it is a good idea to run `restic check` regularly.

More ideas?

dionorgua:
Just an idea: some sort of counter somewhere (maybe in cache dir when it'll be available) and print such msg to stderr if 'check' was executed more than X days/backups ago. And reset it after `restic check`

dionorgua:
[ sorry not read 'please-upgrade' page ]. This was already mentioned

rumpelsepp:
Git runs `git gc` automatically. Does it make sense for you to automatically run (like git does) `restic check`, let's say every 10 backups. A config switch to opt-out from automatic checks seems useful.

jkahrs:
I like the idea of opt-out automatic runs.
However I'd suggest to run the check based on time since last check. This way people who are running hourly backups won't receive a penalty for using their repo more frequently.
Another thing to think about is on what commands the check should be forced. Can it be safe to only force a check on forget/prune?

zcalusic:
I'd rather that check is not forced upon me at all! Warn me that check hasn't run in a long time - OK, but I'd rather my backup schedule is predictable and without sudden surprises (oh, I decided to run extra 2 hour check, sorry user).

Having said that, I'm currently implementing backup+prune+check (this last one is new) for some of my repos, but for that to finish in adequate time, I had to revert the patch that disables caching for check command. So far, so good. It reports all the same repo errors I observe without cache, but does it much much faster. 😇 

I strongly feel that caching should be implemented first, and also for check command, by default. Together with --force or --no-cache option, when you feel like you want a brute force check the repo itself, bypassing cache. Then all commands would behave the same - by default caching is on - and if you're short on disk space, you can turn that off and trade time for space.

jniggemann:
@fd0: The first todo on the list is fixed, can you please add a checkmark so we see the one that's remaining?

fd0:
done.

PR #1259, opened 2017-09-21
adds advice to run restic check regularly

jniggemann:
First shot at partly fixing #1137...
Ideas / comments?

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -512,6 +512,10 @@ find the old snapshot in the repo and by default only reads those files
 that are new or have been modified since the last snapshot. This is
 decided based on the modify date of the file in the file system.
 
+Now is a good time to run ``restic check`` to verify that all data
+is properly stored in the repository. You should run this command regularly
+to make sure the internal structure of the repository is free of errors.
+
 You can exclude folders and files by specifying exclude-patterns. Either
 specify them with multiple ``--exclude``'s or one ``--exclude-file``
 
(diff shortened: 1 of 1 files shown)


zcalusic (changes requested):
3 typos in the last paragraph. Other than that, I like it. 👍

jniggemann:
Argh, will fix the typos ...

fd0 (changes requested):
Looks very good, thanks a lot for your contribution!

I know this sounds a bit petty, but may I ask you to reformat your commits so that they conform to the general style we used for restic? It's documented here: https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits and here https://chris.beams.io/posts/git-commit/

Especially:
 * Start the terse subject line with an upper case letter (except when there's a module prefix like `backend: ` or so)
 * Use imperative for the verb: Describe what the commit does when applied
 * Suash the two commits into one, there's no value in keeping the typos ;)
 
 So I'd suggest using the following subject for the commit: `Add advice to run check regularly`
 
 Thanks!

jniggemann:
Petty? Just a tad ;-) So - no problem, it's your project so you set the guidelines.
My PRs will be about documentation, should I prefix them with `doc:`?

fd0:
Good idea, for commits that only change documentation. Thanks!

fd0 (approved):
Thanks!

Issue #1014, opened 2017-06-09
Add information on special items (symlinks, device files, ...) to the manual


LaszloHont:
Different backup programs handle symlinks differently, some backup the symlink, some follow the symlink and backup the content, most will stay under the location you start your backup at. Could this be added to the docs at https://restic.readthedocs.io/ ?

An example: what happens if you are in /my/path/ and there is a symlink to /my/ ? Will restic also backup /my/ even though it's outside of the backup root?

fd0:
Hi, thanks for bringing this up. We're archiving symlinks as symlinks, we don't follow them. So when you restore, you'll get the same symlink again, with the same link target and the same timestamps. I think we can add that to a section in the manual which describes how we handle items in the file system that aren't files and directories (sockets, pipes, devices, symlinks etc.)

LaszloHont:
So if you have a hardlink, there is the potential for the backup to go outside of the backup root (which is kind of expected I think, hardlinks screw everything up). A bind mount could do as well I suppose. I can't think of any other cases where that could occur though (is this right?)

fd0:
I think we need to define what "outside of the backup root" means. For file that is hard linked somewhere else, speaking in the semantics of a file system, the file is **in** the backup root. restic will only ever descend into the directories you give it on the command line after `backup`. It won't follow symlinks. If there is a bind-mount below a directory that is to be saved, restic descends into it. If you manage to build a loop with bind mounts (not sure if that's possible) then restic will probably keep on descending forever. This is a case that is not handled.

Please be aware that `-x` does not always work with bind mounts. The help says:
```
  -x, --one-file-system         exclude other file systems
```
This means that restic will stay within the same **file system**, not necessarily on the same **mount point**. That's subtly different.

Internally, restic does a `stat()` on all items. This includes an identifier for the filesystem the item is stored on. If `-x` is given, it consults a list of allowed file systems (created at start from the list of files/dirs to backup) in order to decide if it should be saved or not.

LaszloHont:
Thanks.

alphapapa:
Please do document this.  I'm glad I caught it while testing, otherwise I might not have noticed until I needed to run a restore and ended up with a symlink instead of the directory it points to.

PR #1260, opened 2017-09-21
Doc add info on special items

jniggemann:
Document behaviour for symlinks and bind-mounts.

### DIFF
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -574,6 +574,11 @@ and the same timestamps.
 
 If there is a **bind-mount** below a directory that is to be saved, restic descends into it.
 
+**Device files** are saved and restored as device files. This means that e.g. ``/dev/sda`` is
+archived as a block device file and restored as such. This also means that the content of the
+corresponding disk is not read, at least not from the device file.
+
+
 Reading data from stdin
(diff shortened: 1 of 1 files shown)


fd0 (changes requested):
Thanks for taking a stab at the issue.
 * Same advice as in #1259 applies for the title of the commit. Summary line suggestion: `manual: Add info on special items (symlinks)`
 * What about device files? Those are saved (and restored) as device files, so e.g. `sda` is archived as a block device file and restored as such, and the contents of the first hard disk is not read, at least not from the device file
 * There's a second commit (the one that adds the hint for `restic check`) that shouldn't be in here
 * When closing an issue (like #1014 here), either mention `Closes: #1014` in the body of the commit message or in the body of the pull request. The terse subject line should say what the commit does, not which issue it solves. Here it's best to at least repeat the text in the body of the Pull Request, then it is automatically linked from the issue.

If you need help with that, please let us know! :)

jniggemann:
- removed superfluous commit
- Fixed commit message

I'll add another commit with information regarding device files.

fd0 (approved):
Thanks!

PR #1266, opened 2017-09-22
Add note about rescan to hostname flag (fixes #1221)

### LINKED ISSUE #1221: Changing the hostname results in expensive rescan
## Output of `restic version`

restic 0.7.1

I have a restic setup where I use a local rest-server [1] storing the files on a locally attached disk. I then run restic on a remote host via SSH using the rest transport (```RESTIC_REPOSITORY=rest:https://...```)

Typically a backup of that remote system is rather fast (a few minutes).
Today I decided that the hostname I had chosen for the remote system was not ideal, so I changed it via ```restic --hostname=```.
The result was that restic took a really long time to complete. I didn't use a stopwatch, but it was around an hour or so. During that time, almost no data was transmitted.
I'm not entirely sure what was going on and why.
If this behaviour is expected, could this please be documented in the man page for the --hostname flag.

[1] https://github.com/restic/rest-server

JaCoB1123:


fd0 (approved):
Fine with me, thanks!

PR #1269, opened 2017-09-23
Add --compact option to forget

mrzv:
This should have been done as part of #1223. It adds `--compact` option to `forget` to have the same one-per-line snapshot output as in `snapshots`.


fd0 (approved):

fd0:
Nice, thank you!

PR #1270, opened 2017-09-23
sftp: Allow password entry

### LINKED ISSUE #448: request: SFTP with password prompt
Hi guys, I'm using a service which ask for the password and I can't handle the auth using public keys, could be possible than for future releases restic handle sftp with password?...maybe setting the values as a env var or passing as argument?...thanks!.

fd0:
This was a bit tricky: We start the ssh binary, but we want it to ignore SIGINT. In contrast, restic itself should process SIGINT and clean up properly. Before, we used `setsid()` to give the ssh process its own process group, but that means it cannot prompt the user for a password because the tty is gone.

So, now we're passing in two functions that ignore SIGINT just before the ssh process is started and re-install it after start.

Closes #448


fd0:
@mebus @rawtaz can maybe one of you test this PR?

rawtaz:
Works great here, for all I can tell it now asks me for the password of my key when the ssh-agent isn't loaded. Great!

fd0:
Nice, thanks for the feedback!

aadrian:
@fd0 how to use this feature? The documentation still mentions only the key based config: https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#sftp

Thanks in advance.

fd0:
Hm, you should have been asked for the password interactively. If that doesn't work, please open a new issue so we can triage it. Thanks!

PR #1275, opened 2017-09-24
Always sort snapshots lists ascending by timestamp

### LINKED ISSUE #1219: restic prune --keep-* reverses sort order compared to restic snapshots
## Output of `restic version`

restic 0.7.1

Running ```restic snapshots``` lists the snaphosts in ascending order, i.e. oldest snapshots come first and newest are shown last.

When running ```restic forget --keep-*``` this order is reversed. Not sure if there is a good reason for that that, but I found that a bit confusing when your are used to the output of ```restic snapshots```

fawick:
Fixes #1219.


fd0 (approved):
nice, thanks!

PR #1276, opened 2017-09-24
Add REST backend option to use CA root certificate

### LINKED ISSUE #1114: Restic doesn't accept self-signed certificates
## Output of `restic version`

restic 0.7.0

restic init --repo "rest:https://backup:backup@192.168.1.250:8888/repo_name"

Caddy's plugin

## Expected behavior

Repo inited and I can do backups over HTTPS

## Actual behavior

create backend at rest:https://backup:backup@192.168.1.250:8888/repo_name failed: Post https://backup:backup@192.168.1.250:8888/repo_name?create=true: x509: certificate signed by unknown authority

There is no option to accept self-signed certificates. I can't easily issue LE certificate because this VM isolated from WAN.

Create Caddyfile with HTTP and HTTPS (or just HTTP) enabled, specify restic's plugin options, start Caddy. Issue this command on client:

restic init --repo "rest:https://backup:backup@192.168.1.250:8888/repo_name"

fawick:
Fixes #1114.

### DIFF
--- a/internal/backend/http_transport.go
+++ b/internal/backend/http_transport.go
@@ -1,6 +1,10 @@
 package backend
 
 import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"time"
@@ -8,8 +12,10 @@ import (
 	"github.com/restic/restic/internal/debug"
 )
 
-// Transport returns a new http.RoundTripper with default settings applied.
-func Transport() http.RoundTripper {
+// Transport returns a new http.RoundTripper with default settings applied. If
+// a custom rootCertFilename is non-empty, it must point to a valid PEM file,
+// otherwise the function will return an error.
+func Transport(rootCertFilenames []string) (http.RoundTripper, error) {
 	// copied from net/http
 	tr := &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
@@ -25,6 +31,28 @@ func Transport() http.RoundTripper {
 		ExpectContinueTimeout: 1 * time.Second,
 	}
 
+	if rootCertFilenames == nil {
+		return debug.RoundTripper(tr), nil
+	}
+
+	p := x509.NewCertPool()
+	for _, filename := range rootCertFilenames {
+		if filename == "" {
+			return nil, fmt.Errorf("empty filename for root certificate supplied")
+		}
+		b, err := ioutil.ReadFile(filename)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read root certificate: %v", err)
+		}
+		if ok := p.AppendCertsFromPEM(b); !ok {
+			return nil, fmt.Errorf("cannot parse root certificate from %q", filename)
+		}
+	}
+
+	tr.TLSClientConfig = &tls.Config{
+		RootCAs: p,
+	}
+
 	// wrap in the debug round tripper
-	return debug.RoundTripper(tr)
+	return debug.RoundTripper(tr), nil
 }
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/backend/azure"
 	"github.com/restic/restic/internal/backend/b2"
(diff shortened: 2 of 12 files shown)


fd0:
Hm, it feels to me that the REST backend is the wrong place to add a CA certificate. What about moving the code to https://github.com/restic/restic/blob/49d397a4190de017ace5802962aeda22c518c67c/internal/backend/http_transport.go#L12?

fawick:
Fair enough. Will refactor.

fawick:
So, I moved that part to package `internal/backend`. The only one using it so far is still the REST backend, as this was specifically was #1114 was about. It might be useful for the S3 backend (when Minio is used). But probably not for GCS / Azure / Swift, I guess. What do you think, @fd0?

internal/backend/http_transport.go
```diff
@@ -1,15 +1,25 @@
...
 	"github.com/restic/restic/internal/debug"
 )
 
-// Transport returns a new http.RoundTripper with default settings applied.
-func Transport() http.RoundTripper {
+// Transport returns a new http.RoundTripper with default settings applied. If
+// a custom rootCertFilename is non-empty, it must point to a valid PEM file,
+// otherwise the function will return an error.
+func Transport(rootCertFilename string) (http.RoundTripper, error) {
+	p, err := certPoolFromFile(rootCertFilename)
```
fd0: This needs to be guarded with an `if rootCertFilename == ""`.

internal/backend/http_transport.go
```diff
@@ -23,8 +33,23 @@ func Transport() http.RoundTripper {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
+		TLSClientConfig: &tls.Config{
```
fd0: Same here: Only set this if a filename is present.

fd0 (changes requested):

fd0:
I agree that it's also valuable for the s3 backend. Hm. To be honest, I'd rather have a global option `--cacert` or so that is then used for all backends that use the `Transport`. The most straight-forward way to pass that option to the `Transport` would be to instanciate it in `global.go` and pass it as a parameter to all backends that need it. What do you think? Is that clear enough?

internal/backend/http_transport.go
fawick: Ooops, you're right, of course. I understand that you to write meant `!=`.

internal/backend/http_transport.go
fawick: Done.

fawick:
I'm not against the idea to turn this into a global option. When I started out, I concentrated purely on the REST backend and did not think of it being of use to S3 then.

I want to raise one question, though: Should we be ready for the potential situation of the future in which restic instantiates more than one backend at runtime (e.g. repo migration, repo syncing, repo duplication)? It's easy to create a hypothetical scenario, in which the REST backend uses a custom CA for a self-signed certificate, whereas the the S3/GCS backends would still rely on Amazon's/Google's official certificates. Here it were necessary to be able to set the `--cacert` on a per-backend basis. Or even per repo (AWS S3 <-> self-hosted Minio)...

fd0:
Let's worry about that use case when we get there :)

Another possible solution would be to allow more than one `--cacert` parameter and create a pool of allowed CA certificates (system and specified files)...

fawick:
I'd agree that multiple `--cacert` are the best trade-off here, so I am going to implement that.

fawick:
@fd0 Please take another look and let's discuss. I am not happy with the way I pass the filenames around, as it feels a bit clunky and reduces code readability IMHO. And I still need to come up with some tests for all that.

fd0:
The code looks good already. I'd suggest we don't pass the filenames to `Open()` and `Create()`, but instead create the `http.Transport` in `global.go` and pass it as a parameter to `Open()` and `Create()` instead. This way we can add other (global) options like timeouts, and make this generic for all backends that use HTTP, this helps debugging also with other backends.

fd0:
Do you need any more assistance? Or just not enough time? :)

fawick:
WIP - do not review yet.

fd0:
Looks finished, is it?

fawick:
Yeah, expect for a test. But I can add that later.

fd0 (approved):

PR #1277, opened 2017-09-24
Document GCS permission requirements

prattmic:
Plus a second commit with minor cleanups to the package gs comments.


fd0 (approved):
Thanks!

PR #1281, opened 2017-09-26
gs: allow backend creation without storage.buckets.get

prattmic:
If the service account used with restic does not have the storage.buckets.get permission (in the "Storage Admin" role), Create cannot use Get to determine if the bucket is accessible.

Rather than always trying to create the bucket on Get error, gracefully fall back to assuming the bucket is accessible. If it is, restic init will complete successfully. If it is not, it will fail on a later call.

Here is what init looks like now in different cases.

Service account without "Storage Admin":

Bucket exists and is accessible (this is the case that didn't work before):

```
$ ./restic init -r gs:this-bucket-does-exist:/
enter password for new backend:
enter password again:
created restic backend c02e2edb67 at gs:this-bucket-does-exist:/

irrecoverably lost.
```

Bucket exists but is not accessible:

```
$ ./restic init -r gs:this-bucket-does-exist:/
enter password for new backend:
enter password again:
create key in backend at gs:this-bucket-does-exist:/ failed:
service.Objects.Insert: googleapi: Error 403:
my-service-account@myproject.iam.gserviceaccount.com does not have
storage.objects.create access to object this-bucket-exists/keys/0fa714e695c8ecd58cb467cdeb04d36f3b710f883496a90f23cae0315daf0b93., forbidden
```

Bucket does not exist:

```
$ ./restic init -r gs:this-bucket-does-not-exist:/
create backend at gs:this-bucket-does-not-exist:/ failed:
service.Buckets.Insert: googleapi: Error 403:
my-service-account@myproject.iam.gserviceaccount.com does not have storage.buckets.create access to bucket this-bucket-does-not-exist., forbidden
```

Service account with "Storage Admin":

Bucket exists and is accessible: Same

Bucket exists but is not accessible: Same. Previously this would fail when Create tried to create the bucket. Now it fails when trying to create the keys.

Bucket does not exist:

```
$ ./restic init -r gs:this-bucket-does-not-exist:/
enter password for new backend:
enter password again:
created restic backend c3c48b481d at gs:this-bucket-does-not-exist:/

irrecoverably lost.
```

@fd0 I think the best approach for ensuring this continues to work is for you to remove the "Storage Admin" role for the service account you are using for integration testing (do so at https://console.cloud.google.com/iam-admin/iam/project).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Important Changes in 0.X.Y
    untangle mixed files automatically.
    https://github.com/restic/restic/pull/1265
 
+ * The Google Cloud Storage backend no longer requires the service account to
+   have the `storage.buckets.get` permission ("Storage Admin" role) in `restic
+   init` if the bucket already exists.
+   https://github.com/restic/restic/pull/1281
+
 Small changes
 -------------
 
--- a/doc/manual.rst
+++ b/doc/manual.rst
@@ -411,16 +411,18 @@ Restic connects to Google Cloud Storage via a `service account`_.
 
 For normal restic operation, the service account must have the
 ``storage.objects.{create,delete,get,list}`` permissions for the bucket. These
-are included in the "Storage Object Admin" role. For ``restic init``, the
-service account must also have the ``storage.buckets.get`` and
-``storage.buckets.create`` (if the bucket does not exist) permissions. These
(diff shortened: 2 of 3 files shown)


fd0 (approved):
Great work, thanks! And you even added an entry to the `CHANGELOG.md`, wohoo!

PR #1287, opened 2017-09-28
cache: OS-specific cache directories

prattmic:
Note: I haven't had a chance to test this on Windows yet, and I don't have an OS X machine.

Windows, and to a lesser extent OS X, don't conform to XDG and have
their own preferred locations for caches.

On Windows, use %LOCALAPPDATA%/restic (i.e., ~/AppData/Local/restic). I
can't find authoritative documentation from Microsoft recommending
specifically which of %APPDATA%, %LOCALAPPDATA%, and %TEMP% should be
used for caches, but %LOCALAPPDATA% is where browsers store their
caches, so it seems like a good fit.

On OS X, use ~/Library/Caches/restic, which is recommended by the Apple
documentation. They do suggest using the application "bundle identifier"
as the base folder name, but restic doesn't have one, so I just used
"restic".

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -84,7 +84,7 @@ func writeCachedirTag(dir string) error {
 // performReadahead returns true.
 func New(id string, basedir string) (c *Cache, err error) {
 	if basedir == "" {
-		basedir, err = getXDGCacheDir()
+		basedir, err = defaultCacheDir()
 		if err != nil {
 			return nil, err
 		}
--- a/internal/cache/dir.go
+++ b/internal/cache/dir.go
@@ -3,27 +3,68 @@ package cache
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/pkg/errors"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/fs"
 )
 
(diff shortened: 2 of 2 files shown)


dionorgua:
Just small note. 

`%LOCALAPPDATA%` is probably correct choose. The real difference is that in corporate environment `%APPDATA%` will be roamed when user sign to different machine (Sign in on PC1, store something in `%APPDATA%`, logout. Go to PC2 and sign in with same domain account. `%APPDATA%` will be synchronized).

The only thing I'm wondering is that `%LOCALAPPDATA%/restic` is subject to backup by other apps. Ideally `restic` should explicitly tell system that this directory has no important data by creating special registry key in `FilesNotToBackup` tree: https://msdn.microsoft.com/en-us/library/windows/desktop/bb891959(v=vs.85).aspx

prattmic:
I can report that this works swimmingly on my Windows machine.

@dionorgua good to know thanks. Setting that registry key would be clever, but I think that is outside the scope of this PR.

fd0 (approved):
LGTM!

rawtaz:
Confirmed working on macOS 10.12.6.

fd0:
Thank you very much for the confirmation. This change is better done sooner rather than later, while the number of people affected (in terms of "have an unused cache lying around") is small.

Thanks everyone!

Issue #1160, opened 2017-08-15
most of time for backup is cost in scan(unreasonable)


gluthgao:
## Output of `restic version`

debug log file /Users/gluth/go/src/github.com/restic/restic/log1.txt
debug enabled
restic 0.7.1 
compiled with go1.8.1 on darwin/amd64
## How did you run restic exactly?
Run the back up of local NAS to remote NAS
./restic -r remoteNAS backup /  NAS

## What backend/server/service did you use?
NAS

## Expected behavior
We should not scan the file system first which cost long time, i have lots of file  around 15TB, 
the total time should be spend on the backup data.
## Actual behavior
the scan cost around 6 hours !!!!

## Steps to reproduce the behavior
lots of file system like NAS, then do the backup

## Do you have any idea what may have caused this?
scan file system cost more time is normally since the fs.walk is single thread travel model which will cost long time.

## Do you have an idea how to solve the issue?
Do we necessary to scan the file system to get the total file/dir number + size?
also some file may not have permission to read or read failure that's hard to get the accurate size.

fd0:
Hi, thanks for raising this issue. First, please let me give you feedback on your style of reporting issues here at GitHub: I feel that you're writing in a style that comes across as very accusatory and heated (at least to me). Can you try writing more matter-of-factly? That'd help discussing issues a lot. Thanks!

So what I understood so far is that you have a local directory called `NAS` with about 15TiB of data that is to be saved into a local repository in the directory `remoteNAS`, is that correct? From the names of the directories I can guess that the data is really read from (and saved to) network attached storage, so how did you mount the directories? Which service and implementation is used for that?

Which type of storage is used for both the directory save and the storage location for the repository? Classical hard discs or flash-based SSDs?

The scan phase is single-threaded because in my benchmarks having a multi-thread scanning phase did not perform any better, but instead slowed finding all file and directory sizes down.

We're scanning the file system to be saved before a backup for two reasons:
 * To give the user a sense of where we are in the backup it is necessary to find the number of files and the sum of the file sizes
  * This way, on modern operating systems like Linux the directory cache is primed. The backup phase that follows right after scanning can then be much more efficient because restic and the kernel doesn't need to wait until a directory listing is loaded from disc.

gluthgao:
I‘m not good at write english:), you guys did a great job,  i'm in looking to use restic to backup NAS.
[...]

Issue #1063, opened 2017-06-25
Support --json flag with backup command


tamalsaha:
We would like collect info about a backup action from script. But --json flag is not supported for `backup` command. Can this be supported?

fd0:
Good point, I can add this when I refactor the archiver.

xcezzz:
Hi,

I actually had already started this since we planned to wrap restic within a GUI application.

Although a bit off from being worthy of a pull request, my fork allows for --json and will output data in the format...

```json
{
"type":"archive",
"duration":332406583,
"directories":28,
"files":324,
"completed":130878418,
"total":11437998853,
"bps":0,
"completed_bytes":130878418,
"items_done":106,
"items_total":352,
"total_bytes":11437998853,
"eta":0,
"errors":0,
"seconds":0
}
```

type: switches between "archive" and "scanner" to differentiate what process it is in the middle of... 

Currently though you should be aware that sometimes output buffering within the console or whatever you plan on having read this information can cause it to delay and then burst with so much info... using unbuffer from expect fixes this.

https://github.com/xcezzz/restic/tree/json-backup-status

@fd0 in an effort to help out with this what would you recommend being done so that I can get this merged in as it seems to be something a lot of people want...

middelink:
I think you should integrate that progress_output.go file into cmd_backup. It is completely private to the backup function. E.g. it is not a generic helper function.
Also you need to sync and remove some debugging? That commented out return looks suspicious.

Oh, and for commenting it's easier if you make it a PR on this repo...

whereisaaron:
The JSON output idea for non-interactive use is a nice idea, but could it be without line breaks please? That way is also suitable for non-interactive logging output also.

agrover:
Yeah, I'm looking at adding support for Restic to [DejaDup](https://wiki.gnome.org/Apps/DejaDup) (Restic's cmdline is tantalizingly close to Duplicity, which is currently used by DejaDup) but I need something that can be parsed while running, so either incremental json or some other way for programmatic interaction is needed to do this reasonably.

jasperjorna:
What is the latest status on this? I see that there is a`--json` flag for 'some' commands, but `backup` doesn't seem to be supported as of `0.9.4` yet?

fd0:
Ah, this should have been closed! Support for streaming JSON status for `backup` was added in #1944, which is contained in master but not in any release yet. But all the [beta binaries](https://beta.restic.net/?sort=time&order=desc) have it.

[...]

PR #1293, opened 2017-09-28
json backup status - offload progress output code

xcezzz:
This is the rebase for the new repo layout

This contains the json output status functionality, and possibly a more configurable progress output module. 

Referencing issues #1063 and #1160


cmd/restic/backup_progress.go
```diff
@@ -0,0 +1,176 @@
...
+func init() {
+	templateFunctions := template.FuncMap{
+		"duration": formatDuration,
+		"percent":  formatPercent,
+		"bytes":    formatBytes,
+		"seconds":  formatSeconds,
+		"rate":     formatRate,
+	}
+
+	scanProgressUpdateTemplate, _ = template.New("scanProgressUpdateTemplate").Funcs(templateFunctions).Parse(ScanProgressUpdateTemplate)
```
fd0: This throws away the error, use `template.Must()` around it (which will panic if an error occurs).

cmd/restic/backup_progress.go
```diff
@@ -0,0 +1,176 @@
...
+
+	f := bufio.NewWriter(os.Stdout)
+	fmt.Fprintf(f, "%s\n", js)
+	f.Flush()
+}
+
+// PrintScannerProgress print the scanner content
+func (ps *ProgressStatus) PrintScannerProgress() {
+	var result bytes.Buffer
+	_ = scanProgressUpdateTemplate.Execute(&result, ps)
```
fd0: Don't just throw away the error here: If you're sure that the error never occurs, add something like `if err != nil { panic(err) }` so that we notice when it somehow becomes non-nil.

cmd/restic/backup_progress.go
```diff
@@ -0,0 +1,176 @@
...
+	PrintProgress(result.String())
+}
+
+// PrintScannerDone will print the scanner done content
+func (ps *ProgressStatus) PrintScannerDone() {
+	var result bytes.Buffer
+	_ = scanProgressDoneTemplate.Execute(&result, ps)
```
fd0: Same here, do not just ignore the error.

cmd/restic/backup_progress.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+
+	return status1
+}
+
+// PrintArchiveDoneProgress will print current archive progress
+func (ps *ProgressStatus) PrintArchiveDoneProgress() {
+	var result bytes.Buffer
+	_ = archiveProgressDoneTemplate.Execute(&result, ps)
```
fd0: Same here, check the error.

cmd/restic/backup_progress.go
```diff
@@ -0,0 +1,176 @@
...
+	// PrintProgress(result.String())
```
fd0: Leftover debug comment?

xcezzz:
Hey, had a chance to resolve your changes.

Been busy implementing more json functionality with other functions...

fd0:
Thanks. Can you please remove the merge commit and the commit that fixes the broken merge? Those two don't need to be in the patch set.

Let me know if you need help.

fd0:
There hasn't been any progress here for the last few weeks so I'm closing this PR. Just leave a comment if you're still interested in continuing the work (or need help), thanks!

xcezzz:
Sorry not very well versed with Git as I would like to be so am not sure how to go about updating some of these things. There are other changes I have been making regarding output of JSON information using the client when listing files and other status information.

PR #1299, opened 2017-09-30
fuse: Add a snapshot IDs directory (#1102)

### LINKED ISSUE #1102: Enhancement: Add a snapshot IDs directory to the FUSE mount
Currently we can see snapshots by hosts, date and tags.
It would be nice to be able to browse snapshots per ID in a FUSE mount, with  a special directory.

E.g.: `ls -l /mnt/restic/ids/3662613c`

mungomat:
Closes #1102

### DIFF
--- a/internal/fuse/snapshots_dir.go
+++ b/internal/fuse/snapshots_dir.go
@@ -66,6 +66,26 @@ func NewSnapshotsDir(root *Root, inode uint64, snapshots restic.Snapshots) *Snap
 	return d
 }
 
+// NewSnapshotsIDSDir returns a new directory containing snapshots named by ids.
+func NewSnapshotsIDSDir(root *Root, inode uint64, snapshots restic.Snapshots) *SnapshotsDir {
+	debug.Log("create snapshots ids dir with %d snapshots, inode %d", len(snapshots), inode)
+	d := &SnapshotsDir{
+		root:      root,
+		inode:     inode,
+		snapshots: snapshots,
+		names:     make(map[string]*restic.Snapshot, len(snapshots)),
+	}
+
+	for _, sn := range snapshots {
+		name := sn.ID().Str()
+
+		d.names[name] = sn
+		debug.Log("  add snapshot %v", name)
+	}
+
+	return d
+}
+
 // Attr returns the attributes for the root node.
 func (d *SnapshotsDir) Attr(ctx context.Context, attr *fuse.Attr) error {
 	attr.Inode = d.inode
--- a/internal/fuse/root.go
+++ b/internal/fuse/root.go
@@ -56,6 +56,7 @@ func NewRoot(ctx context.Context, repo restic.Repository, cfg Config) (*Root, er
 		"snapshots": NewSnapshotsDir(root, fs.GenerateDynamicInode(root.inode, "snapshots"), snapshots),
 		"tags":      NewTagsDir(root, fs.GenerateDynamicInode(root.inode, "tags"), snapshots),
 		"hosts":     NewHostsDir(root, fs.GenerateDynamicInode(root.inode, "hosts"), snapshots),
+		"ids":       NewSnapshotsIDSDir(root, fs.GenerateDynamicInode(root.inode, "ids"), snapshots),
 	}
 
 	root.MetaDir = NewMetaDir(root, rootInode, entries)


internal/fuse/snapshots_dir.go
```diff
@@ -66,6 +66,41 @@ func NewSnapshotsDir(root *Root, inode uint64, snapshots restic.Snapshots) *Snap
...
+		snapshots: snapshots,
+		names:     make(map[string]*restic.Snapshot, len(snapshots)),
+	}
+
+	// Track latest Snapshot
+	var latestTime time.Time
+	d.latest = ""
+
+	for _, sn := range snapshots {
+		name := fmt.Sprintf("%.8s", sn.ID().String())
```
fd0: Did you discover `ID.Str()` yet? That's the method we're using everywhere to shorten an ID.

internal/fuse/snapshots_dir.go
```diff
@@ -66,6 +66,41 @@ func NewSnapshotsDir(root *Root, inode uint64, snapshots restic.Snapshots) *Snap
...
+	d := &SnapshotsDir{
+		root:      root,
+		inode:     inode,
+		snapshots: snapshots,
+		names:     make(map[string]*restic.Snapshot, len(snapshots)),
+	}
+
+	// Track latest Snapshot
+	var latestTime time.Time
+	d.latest = ""
```
fd0: Do we need the `latest` symlink for the IDs subdir? I don't think so, but I'm not against it...

fd0 (changes requested):

internal/fuse/snapshots_dir.go
mungomat: You are right. We already have /mnt/snapshots/latest

internal/fuse/snapshots_dir.go
mungomat: Thank you

fd0:
Open question from #1102: Should we list the short or the long snapshot IDs?

rawtaz:
My opinion:

- All over restic, make IDs be in the short format by default.
- Have a global switch/argument like --long-ids that when enabled changes the format of IDs so that they are instead printed in long format.

That way the convenient and most often needed/used format is the one that will be used by default, and when/if someone needs the long format, they can make that happen when they need it.

mungomat:
Good question. My arguments for choosing short ids were:
- The issue wished for it
- "restic snapshot" as well as "restic backup" show the short versions

On the other hand we want to avoid collisions. As can be seen in https://en.wikipedia.org/wiki/Birthday_attack  the short versions ids (32Bit) for 93 snapshots is expected to collide in 1 in a Million. So it will happen eventually.

That is why I also like your idea of having global option --long-ids. But maybe this is another issue since this will affect several places in restic?

maxkueng:
> "restic snapshot" as well as "restic backup" show the short versions

But `restic snapshots --json` uses the long IDs. So if the short Ids are used everywhere then the JSON output should use the same. Or at least provide an additional `{..."short_id": "3662613c",...}` field in the output.

rawtaz:
As far as I'm concerned, JSON output is definitely a programmatic interface, so in that case (and other JSON output) it can be logn IDs by default. But in general and when it's not clearly meant for programming, it might be better to default to short IDs, with a switch to make them long.

Regarding this specific PR I think that while the mount can be used in a scripted way, it's not necessarily always the case, so I think the general default should be applied here too.

maxkueng:
What about having both `/mnt/restic/ids` and `/mnt/restic/long_ids`, or the other way around: `/mnt/restic/short_ids` and `/mnt/restic/ids`?

I really like the structure of `/dev/disk`. It looks so neat and ordered.

```
⌁ [max:~] % ls -l /dev/disk 
total 0
drwxr-xr-x 2 root root 380 Oct  3 14:48 by-id
drwxr-xr-x 2 root root  60 Oct  3 14:48 by-label
drwxr-xr-x 2 root root  60 Oct  3 14:48 by-partlabel
drwxr-xr-x 2 root root 120 Oct  3 14:48 by-partuuid
drwxr-xr-x 2 root root 180 Oct  3 14:48 by-path
drwxr-xr-x 2 root root 160 Oct  3 14:48 by-uuid
```

So for restic this could look like this:

```
⌁ [max:~] % ls -l /mnt/restic 
total 0
drwxr-xr-x 2 root root 300 Oct  3 11:01 by-id
drwxr-xr-x 2 root root 100 Oct  2 10:42 by-shortid
drwxr-xr-x 2 root root 140 Oct  3 11:01 by-date
drwxr-xr-x 2 root root 140 Oct  2 10:42 by-host
drwxr-xr-x 2 root root 140 Oct  2 10:42 by-tag
```

But obviously this would be a breaking change.

fd0:
Thank you all for your comments, I'd like to use the short IDs in the fuse mount and I'll add a field to the JSON output for it.

fd0 (approved):

fd0:
See https://github.com/restic/restic/pull/1320

mungomat:
I like maxkuengs idea

PR #1304, opened 2017-10-01
Show the number of snapshots

jannickfahlbusch:
When viewing the list of snapshots I'd like to see the total amount of them too.

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -163,6 +163,8 @@ func PrintSnapshots(stdout io.Writer, list restic.Snapshots, compact bool) {
 		}
 	}
 
+	tab.Footer = fmt.Sprintf("Amount of snapshots: %d", len(list))
+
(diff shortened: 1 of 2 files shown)


cmd/restic/cmd_snapshots.go
```diff
@@ -163,6 +163,8 @@ func PrintSnapshots(stdout io.Writer, list restic.Snapshots, compact bool) {
 		}
 	}
 
+	tab.Footer = fmt.Sprintf("Amount of snapshots: %d", len(list))
```
fd0: What about "Number of snapshots"? Or just "%d snapshots"? I think that's clearer

fd0 (changes requested):

cmd/restic/cmd_snapshots.go
jannickfahlbusch: Done

fd0 (approved):

PR #1311, opened 2017-10-02
"dump" command no longer has a "trees" option

felix9:


rawtaz:
Which issue is tracking this?

felix9:
no issue, it seemed trivial. I can open one.

rawtaz:
No need for something as trivial as this, just thought there was one.

It's not really clear from your comments why you are doing this. Was the dump command never supporting trees, so you're removing some cruft?

felix9:
I don't know the history; I'm new to restic. I was exploring the commands with "restic help" and got a failure with "restic dump trees", which is obviously because the code doesn't have an option for trees, and I assume that someone deleted that option without updating the docs.

felix9:
The dump trees option was removed in these commits:
f7ae0cb78f4bbe469daf384351a0026ede5c013a
0045f2fb61ed0d75007d7f09d6e1266509deab22

fd0:
There used to be a `restic dump trees`, indeed. Thanks!

fd0 (approved):

PR #1312, opened 2017-10-02
nit: please remove dot imports #1119

### LINKED ISSUE #1119: nit: please remove dot imports
Many (all?) restic tests use dot imports such as `import . "restic/test"`. This makes it harder than necessary to find the definition of identifiers, for example while debugging test failures for #1085.

The Go wiki gives similar advice: https://github.com/golang/go/wiki/CodeReviewComments#import-dot

Please get rid of dot imports.

ghost:
This pull request was created to remove all dot imports found in `cmd` and `internal` files.

Closes #1119

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /restic
 /.vagrant
 /doc/_build
-/.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /restic
 /.vagrant
 /doc/_build
+/.vscode
--- a/cmd/restic/integration_fuse_test.go
+++ b/cmd/restic/integration_fuse_test.go
@@ -13,7 +13,7 @@ import (
 
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
-	"github.com/restic/restic/internal/test"
+	rtest "github.com/restic/restic/internal/test"
 )
 
 const (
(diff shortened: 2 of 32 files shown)


hgfischer:
I don't get why the difference with Codecov. I've actually haven't added or removed code.

hgfischer:
And it looks like Travis failed because of timeouts.

prattmic:
+1 to removing dot imports. The uses here don't meet the circular dependency bar: https://github.com/golang/go/wiki/CodeReviewComments#import-dot

The code coverage complaint is normal. Travis skips integration tests that use keys to cloud storage providers in pull requests, which confuses CodeCov.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /restic
 /.vagrant
 /doc/_build
+/.vscode
```
fd0: This does not belong here. If you want to ignore that dir in your checkout of the repo, add the directory to `.git/info/exclude`. We don't collect ignores for various editors here :)

fd0 (changes requested):
Good work so far!

.gitignore
hgfischer: I honestly think it is an exaggeration to be picky with `.gitignore`, but, whatever. It is your project.

fd0:
@hgfischer The backend tests cannot be run because for security reasons Travis doesn't expose the needed credentials, don't worry about that.

fd0 (approved):

PR #1300, opened 2017-09-30
doc: Refactors the documentation

jniggemann:
This commit refactors the documentation according to my proposal in #1273
and the discussion I had with fd0 on IRC.

The bits from the manual that I could not immediately put into the new
structure are contained in manual_rest.rst Anything else is still there,
nothing has been deleted.

I changed the heading markup to follow the convention used in Python’s
Style Guide for documentation, this convention is explained in a comment
at the top of every file.

I also added a paragraph on installing restic on Debian.

PR #1315, opened 2017-10-03
Clarify cache location documentation

### LINKED ISSUE #1309: XDG_CACHE_HOME is ignored on macOS
## Output of `restic version`

restic 0.7.3 (v0.7.3-87-g3afd974d)
compiled with go1.9 on darwin/amd64

`restic -r ~/restic-cache-debug backup ~/backup-fodder` 

Local & B2.

## Expected behavior

If `XDG_CACHE_HOME` is defined, and `--cache-dir` is not specified, then restic should use `$XDG_CACHE_HOME ` for cache.

## Actual behavior

On macOS, restic is ignoring `XDG_CACHE_HOME`, in favor of either or `--cache-dir` flag or `darwinCacheDir()`. From a quick read of the code, Windows has this problem as well.

Set XDG_CACHE_HOME to a valid dirname, and run a backup as above.

I think it was 21338691276a138f6e0bb2ead9068683c38ddf9f.

If `XDG_CACHE_HOME` is only meant to be supported on Linux (perhaps reasonable to do so?) then the documentation should reflect that. Else, maybe [something like this](https://github.com/armhold/restic/commit/8968fd7fdfd3625d77bca05903e15e004b12a9b0) (I have not tested that code.)

prattmic:
PR #1287 changed the default cache location on darwin and windows.
Update the changelog and manual to reflect the new behavior.

Since the cache hasn't been included in an official release yet, I've
just changed the main cache changelog entry.

Fixes #1309

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@ Important Changes in 0.X.Y
  * We've added a local cache for metadata so that restic doesn't need to load
    all metadata (snapshots, indexes, ...) from the repo each time it starts. By
    default the cache is active, but there's a new global option `--no-cache`
-   that can be used to disable the cache. The cache location is
-   `~/.cache/restic` by default, which can be overridden with `--cache-dir` or
-   the environment variable `XDG_CACHE_HOME`. The cache will automatically
-   populate, indexes and snapshots are saved as they are loaded.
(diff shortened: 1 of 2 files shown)


fd0:
I'll port this over to #1300 once it is merged.

PR #1319, opened 2017-10-03
Add explicit OK output to check command

### LINKED ISSUE #1303: The check command should output "No errors found."
Hallo,

the check command is not explicit about whether it found errors.

Please output something like "No errors were found." in addition to

```
Load indexes
Check all packs
```
when it did not find an error, so that I know whats going on.

Mebus

antonlindstrom:
This adds additional output to the check command when no errors were
found. It means that when all checks have been completed, the following
output is displayed:

	No errors were found

The output is added to make sure that it is easier to understand that no
errors were found.

Full example output:

	Load indexes
	Check all packs
	No errors were found

Fixes #1303 

Thanks!


fd0 (approved):
Nice, thanks!

PR #1320, opened 2017-10-03
snapshots: Add short ID to JSON output

fd0:
We're using the short ID in all output to users, so it should also be included in the JSON output of the `snapshots` command.

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -172,7 +172,8 @@ func PrintSnapshots(stdout io.Writer, list restic.Snapshots, compact bool) {
 type Snapshot struct {
 	*restic.Snapshot
 
-	ID *restic.ID `json:"id"`
+	ID      *restic.ID `json:"id"`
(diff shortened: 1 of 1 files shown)


rawtaz:
This is really ugly. The data provided by JSON is pure raw data, it's not meant for human consumption. Hence it makes sense to keep it "complete" and in this case the full ID.

If someone wants a shorter version they can just cut that out, just like they can pick and choose whatever they want from the rest of the data.

IMO this pollutes the JSON output and adds little value.

fd0:
I disagree: At any point we can decide to e.g. make the short IDs longer or replace it with something else, which will then break everything downstream that parses the JSON and creates the short IDs by truncating the long IDs. When we expose this in the JSON, the other programs can just use the short ID as provided.

Do you have a different proposal?

rawtaz:
I think the JSON output should provide the long/complete ID in the 'id' key, that's all. I'm not fond of the 'short_id' key, personally I would remove it, at the very least until there's a real actual long term use case for having it there.

Anyway, carry on ;)

PR #1324, opened 2017-10-04
Add password successful feedback

### LINKED ISSUE #438: Feedback after entering the repo password

antonlindstrom:
This adds some feedback when entering the password on the command line.
When the password is entered and supplied through stdin (and stdout is a
terminal) then the a message saying `password is correct` if correct is
printed.

The pull request aims to fix #438 

I'm not sure if it's the correct place to output in the `OpenRepository` function
or that my assumption is correct in when it should be outputted.

Thanks!

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -132,20 +132,6 @@ func restoreTerminal() {
 	})
 }
 
-// shouldOutputPasswordSuccessful returns true if the password is supplied
-// interactively and the stdout is a terminal.
-func shouldOutputPasswordSuccessful() bool {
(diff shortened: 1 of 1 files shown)


rawtaz:
I have a thought on this: Why would you not output the text just because the password happens to come from an environment variable or password file, rather than the terminal? I mean, it's still a password, and whether it's correct or not is still relevant. In short, I think you should output the text regardless of the source of the password.

fd0 (approved):
Nice, that was easy.

PR #1325, opened 2017-10-04
Add group name in fillUser

### LINKED ISSUE #154: Implement getgrnam() for group names
Or use a library, e.g. https://github.com/kless/osutil

antonlindstrom:
This does a lookup of the group name from the GID and adds it to the
Node.

Fixes #154 

Thank you so much for the hard work!


fd0 (approved):
Nice, thanks!

PR #1326, opened 2017-10-04
Cache evaluated directories in isExcludedByFile

### LINKED ISSUE #1271: --exclude-if-present leads to lots of lstat-operations for each directory
## Output of `restic version`

restic 0.7.3 (v0.7.3-28-g2080afd-dirty)

export RESTIC_REPOSITORY=/srv/storage/restic
export RESTIC_PASSWORD_FILE=$HOME/.config/restic-password

ionice -c 3 ./restic backup --exclude-caches --exclude /mnt/backup-snapshot/_backup --force /mnt/backup-snapshot

local mount of ZFS via CIFS

## Expected behavior
lstat each CACHEDIR.TAG once per directory

## Actual behavior
CACHEDIR.TAG is lstat'ed once per file 

```
futex(0xc4204cf610, FUTEX_WAKE, 1)      = 1 <0.000007>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/7969-tiny-816f5f5ad4039bdb6eb0bdeebe3fe69ed4f83c3a", {st_mode=S_IFREG|0600, st_size=7425, ...}) = 0 <0.000024>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/CACHEDIR.TAG", 0xc545816518) = -1 ENOENT (No such file or directory) <0.000008>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/7969-tiny-816f5f5ad4039bdb6eb0bdeebe3fe69ed4f83c3a", {st_mode=S_IFREG|0600, st_size=7425, ...}) = 0 <0.000010>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/CACHEDIR.TAG", 0xc5458166b8) = -1 ENOENT (No such file or directory) <0.000007>
epoll_wait(4, {}, 128, 0)               = 0 <0.000006>
futex(0xe3feb0, FUTEX_WAIT, 0, NULL)    = 0 <0.008781>
epoll_wait(4, {}, 128, 0)               = 0 <0.000010>
pselect6(0, NULL, NULL, NULL, {0, 3000}, 0) = 0 (Timeout) <0.000064>
futex(0xc420108490, FUTEX_WAKE, 1)      = 1 <0.000013>
futex(0xc420108490, FUTEX_WAKE, 1)      = 1 <0.000008>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/8236-81177c7d3b2c70e6a9a60bc53bdbc29b5a857be8", {st_mode=S_IFREG|0600, st_size=587571, ...}) = 0 <0.000068>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/CACHEDIR.TAG", 0xc5fbc7f628) = -1 ENOENT (No such file or directory) <0.000009>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/8236-81177c7d3b2c70e6a9a60bc53bdbc29b5a857be8", {st_mode=S_IFREG|0600, st_size=587571, ...}) = 0 <0.000007>
lstat("/mnt/backup-snapshot/[..sensitive pathname..]/cache/81/CACHEDIR.TAG", 0xc5fbc7f7c8) = -1 ENOENT (No such file or directory) <0.000008>
```

run restic with --exclude-caches or own --exclude-if-present

From looking at the code I'd judge that the problem is, that the exclude filter is only applied to entries of each path that is walked in Pipe.Walk. If it would be applied to the path itself, too, and skip walking the entries if the path itself is excluded, this should not happen. Otherwise this seems not the complete solution, as if it were true, the problem should only happen in the first level of path traversal and not in "deeper" levels. So I'd say my analysis is flawed somehow.

I see multiple possible solutions:

1. split exclude filters for paths and entries. This seems to be the most effective solution: if the exclude filters are split (--exclude-if-present goes only with path and not with entry filters) then there is no need to check the presence of these files for each entry, but rather for each path
2. implement a temporary cache for the files presence(s)

fawick:
Fixes #1271.

### DIFF
--- a/cmd/restic/exclude.go
+++ b/cmd/restic/exclude.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
@@ -14,6 +15,50 @@ import (
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/repository"
 )
+
+type rejectionCache struct {
+	m   map[string]bool
+	mtx sync.Mutex
+}
+
+// Lock locks the mutex in rc.
+func (rc *rejectionCache) Lock() {
+	if rc != nil {
+		rc.mtx.Lock()
+	}
+}
+
+// Unlock unlocks the mutex in rc.
+func (rc *rejectionCache) Unlock() {
+	if rc != nil {
+		rc.mtx.Unlock()
+	}
+}
+
+// Get returns the last stored value for dir and a second boolean that
+// indicates whether that value was actually written to the cache. It is the
+// callers responsibility to call rc.Lock and rc.Unlock before using this
+// method, otherwise data races may occur.
+func (rc *rejectionCache) Get(dir string) (bool, bool) {
+	if rc == nil || rc.m == nil {
+		return false, false
+	}
+	v, ok := rc.m[dir]
+	return v, ok
+}
+
+// Store stores a new value for dir.  It is the callers responsibility to call
+// rc.Lock and rc.Unlock before using this method, otherwise data races may
+// occur.
+func (rc *rejectionCache) Store(dir string, rejected bool) {
+	if rc == nil {
+		return
+	}
+	if rc.m == nil {
+		rc.m = make(map[string]bool)
+	}
+	rc.m[dir] = rejected
+}
 
 // RejectFunc is a function that takes a filename and os.FileInfo of a
 // file that would be included in the backup. The function returns true if it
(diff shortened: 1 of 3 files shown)


fawick:
@phillipp I'd be interested to learn whether this PR improves your usecase from #1271

cmd/restic/exclude.go
```diff
@@ -60,22 +68,43 @@ func rejectIfPresent(excludeFileSpec string) (RejectFunc, error) {
...
+// directory based on previous visits.
+func isExcludedByFile(filename, tagFilename, header string, rc *rejectionCache) bool {
 	if tagFilename == "" {
 		return false
 	}
 	dir, base := filepath.Split(filename)
 	if base == tagFilename {
 		return false // do not exclude the tagfile itself
 	}
+	if rc != nil {
```
fd0: It'd be nice if this were a method on the `rc` struct, and it should also work if `rc` is nil, so we can abstract that away, like `rejected, ok := rc.Get(dir); if ok { return rejected }`. What do you think?

cmd/restic/exclude.go
```diff
@@ -60,22 +68,43 @@ func rejectIfPresent(excludeFileSpec string) (RejectFunc, error) {
...
+		rc.mtx.Lock()
+		defer rc.mtx.Unlock()
+		if rejected, visited := rc.m[dir]; visited {
+			return rejected
+		}
+	}
+
+	rejected := isDirExcludedByFile(dir, tagFilename, header)
+	if rc != nil {
```
fd0: This one too, e.g. `rc.Store(dir, rejected)`.

fd0 (changes requested):

cmd/restic/exclude.go
fawick: Good idea. The mutex must remain locked until the function returnes, though, so I wrap its `Lock()` and `Unlock()` methods, too.

cmd/restic/exclude.go
fawick: Done.

fawick (commented):
@fd0 The changes you requested have been implemented.

fd0 (approved):

PR #1334, opened 2017-10-05
backup should use latest parent regardless of tags

### LINKED ISSUE #1143: using tags slows down backup process
## Output of `restic version`

restic 0.7.1
compiled with go1.8.3 on linux/386

time restic -r /dev/shm/test backup --tag **$RANDOM** /usr/share/

a local tmpfs at /dev/shm/

## Expected behavior

restic should finish as fast as without using tags

## Actual behavior

restic is much slower when using a tag name **that wasn't used before**

time restic -r /dev/shm/test backup /usr/share/  \#fast: ~5s
time restic -r /dev/shm/test backup --tag new1 /usr/share/  \#slow: ~30s
time restic -r /dev/shm/test backup --tag new1 /usr/share/  \#fast: ~5s
time restic -r /dev/shm/test backup --tag new1 /usr/share/  \#fast: ~5s
time restic -r /dev/shm/test backup --tag new2 /usr/share/  \#slow: ~30s
time restic -r /dev/shm/test backup --tag new2 /usr/share/  \#fast: ~5s
time restic -r /dev/shm/test backup --tag new2 /usr/share/  \#fast: ~5s

unfortunately not

felix9:
Backup was choosing a parent snapshot that had the same tags, which
makes backup unnecessarily slow when there are newer snapshots with
different tags.

There's no reason parent has to have the same tags.

This change makes backup choose the newest snapshot instead.

Fixes #1143

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -416,7 +416,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, args []string) error {
 
 	// Find last snapshot to set it as parent, if not already set
 	if !opts.Force && parentSnapshotID == nil {
-		id, err := restic.FindLatestSnapshot(context.TODO(), repo, target, []restic.TagList{opts.Tags}, opts.Hostname)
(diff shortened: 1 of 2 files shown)


fd0 (approved):
I agree, thanks!

Issue #1216, opened 2017-09-07
limit bandwidth option


gregf:
It would be nice if restic offered an option like rsync has to limit bandwidth usage. In particular when uploading. I have pretty bad internet where I live and when I upload it really kills my download connection. It would be great if I could limit the upload bandwidth being used by restic.

fd0:
Great idea, I'll keep this issue as a reminder.

flo82:
i'm using traffic shaping with trafficcontrol/iptables.
have a look at this link: https://help.atmail.com/hc/en-us/articles/201566464-Throttling-Bandwidth-using-Traffic-Controller-for-Linux

works perfect :-)

nibble:
An integrated bandwith limiting feature would be great.

In the meantime, what I use to limit traffic for the sftp backend is to pipe the connection through `pv`:

```
Host myserver-restic
        ProxyCommand pv -q -L 100k | nc myserver.com 22
```

apirogov:
I would also really like to have this. My upload speed is about 120KB/s and for an initial backup of 300GB to BackBlaze I would need a month at full upload speed, so I'd rather make two months at half speed out of it and use my internet normally.

Off-topic - is restic even designed for what I want to do? Can the upload be interrupted and continued? I can not assume to have my laptop powered and online for such a long time.

fd0:
> Off-topic - is restic even designed for what I want to do? Can the upload be interrupted and continued? I can not assume to have my laptop powered and online for such a long time.

While it's not an ideal situation, I'm confident that restic will handle this just fine. It just might take longer to upload the data, and it will upload duplicate data.

During upload, even if the backup is not finished, restic will write intermediate index files that contain a list of blobs uploaded so far. At a minimum, such an index file is written every five minutes.

When you abort the backup, and re-run it later, restic will first download all the index files so it knows which blobs are already in the repo. This includes the intermediate index files.

So, eventually you'll finish uploading all data.

armhold:
This would be a really nice feature to have. 

> While it's not an ideal situation, I'm confident that restic will handle this just fine. It just might take longer to upload the data, and it will upload duplicate data.

This is in fact how I uploaded my 185GB to B2. Since it took several days of saturating my Internet, I would ^C the process during the day, and re-run the backup at night. Not ideal, but it worked OK after a few days and then some prune/check sessions.

For pruning, I've found that merely suspending the process (^Z) and then "fg" works fine, even after several hours of suspension.

[...]

PR #1336, opened 2017-10-06
Add basic rate limiting to backup

ghost:
Introduces the --rate-limit flag for the backup subcommand. This sets a
package global indicating the KB/s rate for uploads.

Tested with iftop

References #1216

### DIFF
--- a/internal/limiter/limiter.go
+++ b/internal/limiter/limiter.go
@@ -1,0 +1,17 @@
+package limiter
+
+import (
+	"io"
+)
+
+// Limiter defines an interface that implementors can use to rate limit I/O
+// according to some policy defined and configured by the implementor.
+type Limiter interface {
+	// Upstream returns a rate limited reader that is intended to be used in
+	// uploads.
+	Upstream(r io.Reader) io.Reader
+
+	// Downstream returns a rate limited reader that is intended to be used
+	// for downloads.
+	Downstream(r io.Reader) io.Reader
+}
--- a/internal/limiter/static_limiter.go
+++ b/internal/limiter/static_limiter.go
@@ -1,0 +1,53 @@
+package limiter
+
+import (
+	"io"
+
+	"github.com/juju/ratelimit"
+)
+
+type staticLimiter struct {
+	upstream   *ratelimit.Bucket
+	downstream *ratelimit.Bucket
+}
+
+// NewStaticLimiter constructs a Limiter with a fixed (static) upload and
+// download rate cap
+func NewStaticLimiter(uploadKb, downloadKb int) Limiter {
+	var (
+		upstreamBucket   *ratelimit.Bucket
+		downstreamBucket *ratelimit.Bucket
+	)
+
+	if uploadKb > 0 {
+		upstreamBucket = ratelimit.NewBucketWithRate(toByteRate(uploadKb), int64(toByteRate(uploadKb)))
+	}
+
+	if downloadKb > 0 {
+		downstreamBucket = ratelimit.NewBucketWithRate(toByteRate(downloadKb), int64(toByteRate(downloadKb)))
+	}
+
+	return staticLimiter{
+		upstream:   upstreamBucket,
+		downstream: downstreamBucket,
+	}
+}
+
+func (l staticLimiter) Upstream(r io.Reader) io.Reader {
+	return l.limit(r, l.upstream)
+}
+
+func (l staticLimiter) Downstream(r io.Reader) io.Reader {
+	return l.limit(r, l.downstream)
+}
+
+func (l staticLimiter) limit(r io.Reader, b *ratelimit.Bucket) io.Reader {
+	if b == nil {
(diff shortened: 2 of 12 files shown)


Gopkg.toml
```diff
@@ -72,3 +72,7 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/sys"
+
+[[constraint]]
```
fd0: Hm, did you add this yourself? The way I understood `dep` works is that it should figure out a dependency by inspecting the code, so this entry shouldn't be needed.

fd0 (changes requested):
This looks really good already! However, I'd like to have a discussion on how to integrate it.

fd0:
At the moment this PR conflicts with #1276, which also changes the way the `http.Transport` is passed to the backends, so we need to resolve that once #1276 (which is basically ready) is merged.

Then I'd like to talk about:
 * Integrating this with the `sftp` and `local` backends (yes, also for the `local` backend, in case people use that with other fuse-based file systems like `sshfs` or `cifs`)
 * Getting rid of the global variables in the `backend` package
 * I'd like to do the same for download
 * Finding a better name for the option, `--upload-limit`? And add a separate option for download

It feels to me that the HTTP Transport, while used for several backends, is the wrong place for the limiter. Maybe it should be a backend config thing, so that each backend takes care of it (and we can still provide a wrapper around `http.Transport`).

What do you think?

Gopkg.toml
rmdashrf: Yeah, this is unnecessary. Still getting used to `dep` -- I forget that the Gopkg.lock file is checked in and will contain this revision anyway.

rmdashrf:
@fd0 - I agree with all of your points. I whipped this up quickly because my B2 backup was killing my internet speed, so it was mainly a proof of concept. I think Golang even explicity states that `http.RoundTripper` implementations should not mess with the request. :)

Without globals, I think the best way to approach this is to define something like
```
type Limiter interface {
   LimitReader(r io.Reader) io.Reader
   LimitWriter(r io.Writer) io.Writer
}
```
which would be initialized by the CLI and passed as parameters to all of the backends. The backend implementations would be responsible for wrapping the readers/writers when appropriate.

I have no strong opinions on the command flag, although I do prefer using "limit" as the prefix (e.g. `--limit-upload` and `--limit-download`) for easier copy-pasta. :)

fd0:
Good idea with the interface, but I think defining that for an `io.Reader` is sufficient. We'll see that during the implementation. I'd rather like to differentiate between up and downstream, e.g. like this:
```go
type Limiter interface {
    Upstream(io.Reader) io.Reader
    Downstream(io.Reader) io.Reader
}
```

fawick:
@rmdashrf (Sorry for side-tracking this issue discussion) Out of curiosity, did you consider using golang.org/x/time/rate, too? If so, what are the drawbacks when compared to the juju package?

rmdashrf:
@fawick The juju implementation already implements a tested rate limited `io.Reader`. It seems like the `golang.org/x/time/rate` package would require more work and more testing.

fawick:
I see, thank you!

fd0:
Hm, I've had a look, and it works well for upstream but fails for downstream with the sftp backend. Any ideas?

internal/limiter/limiter.go
```diff
@@ -0,0 +1,15 @@
+package limiter
+
+import (
+	"io"
+)
+
+type Limiter interface {
```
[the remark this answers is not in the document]
fd0: Please add a short description here.

internal/limiter/static_limiter.go
```diff
@@ -0,0 +1,53 @@
...
+		upstreamBucket   *ratelimit.Bucket
+		downstreamBucket *ratelimit.Bucket
+	)
+
+	if uploadKb > 0 {
+		upstreamBucket = ratelimit.NewBucketWithRate(toByteRate(uploadKb), int64(toByteRate(uploadKb)))
+	}
+
+	if downloadKb > 0 {
+		upstreamBucket = ratelimit.NewBucketWithRate(toByteRate(downloadKb), int64(toByteRate(downloadKb)))
```
fd0: Here's a bug: That should be `downstreamBucket` instead of `upstreamBucket`, then it works.

fd0 (changes requested):

internal/limiter/static_limiter.go
rmdashrf: /facepalm

fd0 (approved):

fd0:
Thanks!

zcalusic:
This is really great work, @rmdashrf! 

Even if I don't have immediate need for it, I know the time will come when it'll be indispensable, 'cause it already happened multiple times in the past with rsync, wget, and other apps... IMHO, it's a great feature for a backup program.

Thank you for your contribution!

Issue #1330, opened 2017-10-05
fuse mount: parallel execution of mount and backup


mungomat:
Currently after being mounted the directory does not recognize new snapshots. So the user has to remount the repository after every backup.

I made some promising tests where the mounted directory could show newly generated snapshots. No need to unmount for every backup. Unfortunately this requires a reloading of the index every time the user does a "ls /mnt/snapshots" which takes time.

Questions:
- Would it be a useful feature?
- What about other operations parallel to a mount (forget, prune)?
- Is there a faster way to recognize a repository-change despite reloading the index?

fd0:
That'd be useful indeed, but will require significant changes to the fuse code. It'd need to create a non-exclusive lock in the repo, so that `backup` works fine, but `forget` and `prune` won't.

The best way to find out if there is new data is to reload the list of snapshots.

benschumacher:
This appears to be closed with #1337.

fd0:
Indeed, thanks for the hint! I'm closing this issue.

PR #1337, opened 2017-10-06
fuse: mount and backup in parallel (#1330)

mungomat:
The locking mechanism is missing.

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -67,6 +67,12 @@ func mount(opts MountOptions, gopts GlobalOptions, mountpoint string) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


fd0:
This looks good! One thing I noticed was that the `latest` symlink is not updated.

And just add a lock in `cmd_mount.go`, e.g. like this:
```go
	lock, err := lockRepo(repo)
	defer unlockRepo(lock)
	if err != nil {
		return err
	}
```

mungomat:
Added the locking.

The latest symlink is updated after 1-2min. I don't know why. Can you confirm it?

fd0:
Ah, interesting, that can be caused by fuse or the kernel caching the result of the `Readlink` syscall.

fd0 (approved):

Issue #351, opened 2015-11-11
Add code to auto-detect a read-only file system for local (and possibly) sftp backend


fd0:
In #318 we discussed and added code in #350 to disabe locking so that restic can be used with a read-only backend (e.g. no lock is created). This might be implemented by opening a file with O_WRONLY and check the returned error for EROFS, see https://github.com/restic/restic/issues/318#issuecomment-147466249

fawick:
I try to tackle this over the next few days.

hosebei:
At least I could find this :)
Would be appreciated to have such a feature.

PR #1342, opened 2017-10-08
Add code to autodetect whether backends are read-only or can be written to

fawick:
This comment is an attempt to address #351.

It adds the `restic.Writabler` interface which wraps the basic `Writable() bool` method, that has been added to some of the backends: local, REST, SFTP, and B2. Once `restic.Writabler` is implemented by all `restic.Backend` implementations, `Writable() bool` can (and should) be added to `restic.Backend`. 
To add the functionality to the missing backends, one would only need to implement `restic.Writabler` for them on a case-by-case basis.

The PR also adds `cmd_writable.go` which serves as a usage example. `restic writable` may not a very useful command (or perhaps it is for scripts?), but it served me well for testing the code and may be of interest for checking things out during reviews. In any case, I am happy to remove it before an eventual merge to upstream.

The name `Ẁritable() bool` was chosen over `ReadOnly() bool` because I prefer the affirmative result to return `true` and because otherwise the interfache name were `type ReadOnlier interface` and I did not like that. :)

### DIFF
--- a/internal/backend/rest/rest.go
+++ b/internal/backend/rest/rest.go
@@ -1,6 +1,7 @@
 package rest
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -22,6 +23,9 @@ import (
 
 // make sure the rest backend implements restic.Backend
 var _ restic.Backend = &restBackend{}
+
+// make sure the rest backend implements restic.Backend
+var _ restic.Writabler = &restBackend{}
 
 type restBackend struct {
 	url    *url.URL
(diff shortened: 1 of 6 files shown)


internal/backend/rest/rest.go
```diff
@@ -349,3 +353,48 @@ func (b *restBackend) Close() error {
...
+// Writable implements restic.Writabler for the REST backend and returns
+// whether the backend can be written to or must be considered as read-only.
+//
+// The methods probes the backend by writing to a temporary file and deleting
+// it wright away. If that fails for any reason, the backend is considered to
+// be read-only.
+func (b *restBackend) Writable() bool {
+	debug.Log("Writable()")
+
+	const probeFilename = ".resticWriteCheck"
```
fd0: This can fail when two instances of restic create this file simultaneously. Or restic failed, but left this file behind. The name must be unpredictable.

fd0 (changes requested):
I've left a few comments at specific places, but I must say that I don't like this approach: It's error-prone for most backends, and having an optional interface does not work well when we wrap the backend (e.g. for the cache, for the rate limiting etc).

I'd like to propose a different approach: Try to create a lock, if that failed (and the operation is possible without locking), retry without a lock. That can be easily implemented without touching the backends at all, e.g. by returning a specific `ErrUnableToCreateLock` in https://github.com/restic/restic/blob/9ce2a73fc5fe576e4f459bb2641793e6c6d7ddfd/cmd/restic/lock.go#L22

What do you think?

fawick:
Well, I better scrap this then. :-)

PR #1343, opened 2017-10-09
fuse mount: speedup

mungomat:
Before reading the indexes, restic now checks whether the count of snapshots in the repository changed

A better way would be to have a hash for the whole repository?


fd0:
Looks like a good approximation.

Issue #510, opened 2016-05-08
Allow dumping files to stdout


fd0:
This is a followup issue to #255: Find a place and implement a command or option that allows dumping a file from a snapshot to stdout.

This allows the following use case:
- Backup an SQL dump with restic: `mysqldump | restic backup --stdin --stdin-filename production.sql`
- Restore it directly: `restic cat-file-to-stdout $SNAPSHOTID production.sql | mysql`

cfcs:
Also somewhat useful for workarounds like the one suggested in https://github.com/restic/restic/issues/949

fd0:
This is done in #1346, there's the new `dump` command. I'm closing this issue.

rawtaz:
Should have been `restore --stdout` instead of `dump`, to be coherent with `backup --stdin`.

PR #1346, opened 2017-10-09
Add dump sub-command for dumping files to stdout

fawick:
I took a stab at addressing #510. This PR adds the `dump` subcommand which allows dumping the contents of a backed-up file to standard output.

Edit: renamed from `catfile` to `dump`


fd0:
I did not have the time to look at the code yet, but I'd like to find a better name instead of `catfile`. And I'd like to propose switching the order of the parameters, "<cmd> <ID> <filename>". What do you think?

fawick:
Thanks for the suggestions! I am going to change the parameter order.

As for the name, I am not attached to  `catfile` which I derived by shortening the clunky `cat-file-to-stdout` that was suggested in #510.  I did not take the time to look for a short one that isn't taken yet (e.g. `cat` or `dump`). Maybe `dumpfile`? I don't really like `printfile`. `catcontent` is all over the internet already, so that's out of the question. ;)

fd0 (changes requested):
Hm, did you forget to push? The commit in this PR still has `cmd_catfile.go` and the command is named `catfile` instead of `dump`?

pvgoran:
I suppose it's already late for this, but I think "catfile" is actually much better than "dump". The latter is too vague, and it invokes associations with functionality that produces lengthy debugging output or something low-level and machine-oriented like this, rather than actual data. Whereas the former unambiguously represents the meaning and intent of the command.

fawick:
Yeah, I committed with `-m`and without adding first or using `-a` and then the push had nothing to do. :-) Corrected now.

fd0:
Hm, @fawick you need to rebase on master first, otherwise there's the old debug `dump` command still present.

I don't really like `catfile`, I'd prefer `dump`. After all, it may just be a huge binary virtual machine image that is written to stdout.. What about `print`?

pvgoran:
> What about print?
"print" suffers from the same drawbacks as "catfile", and doesn't have its advantages. :)

"dumpfile" looks like an acceptable combination of "dump" and "catfile" to me. A bit long, but straightforward.

> otherwise there's the old debug dump command still present.
See what I meant? :)

fd0:
I've removed the `dump` debug command (which should be only active for debug builds) in master a few days ago, so that's not a problem any more. @fawick what do you think about `dumpfile`?

fawick:
According to https://github.com/fawick/restic/commits/cmdCatFile that should already be the case...

I'm only on mobile right now so cannot check what missing here.

fawick:
I missed to test the build with `-tags debug` and did not notice the variable and function name clash. I chose to rename the previous `cmdDump` to `cmdDebugDump` and `runDump` to `runDebugDump`, resp.

fd0 (approved):

fd0:
We'll go with `dump`. Thanks!

PR #1358, opened 2017-10-15
gs: add option to set chunk size

prattmic:
By default, the GCS Go packages have an internal "chunk size" of 8MB,
used for blob uploads.

Media().Do() will buffer a full 8MB from the io.Reader (or less if EOF
is reached) then write that full 8MB to the network all at once.

This behavior does not play nicely with --limit-upload, which only
limits the Reader passed to Media. While the long-term average upload
rate will be correctly limited, the actual network bandwidth will be
very spikey.

e.g., if an 8MB/s connection is limited to 1MB/s, Media().Do() will
spend 8s reading from the rate-limited reader (performing no network
requests), then 1s writing to the network at 8MB/s.

This is bad for network connections hurt by full-speed uploads,
particularly when writing 8MB will take several seconds.

Add a backend option to allow specifying the ChunkSize media option,
which sets this internal buffer size in increments of 256KB, with 256KB
as the smallest chunk size.

My connection is around 1.5MB/s up, with nominal ~15ms ping times to
8.8.8.8.

Without this change, --limit-upload 1024 results in several seconds of
~200ms ping times (uploading), followed by several seconds of ~15ms ping
times (reading from rate-limited reader).

With this change and a chunk size of 256KB, --limit-upload 1024 results
in a relatively constant ~30-40ms ping times.

It may eventually make sense to automatically reduce the chunk size if
--limit-upload is set, but at this point I don't have a good heuristic
for what the size should be.

As far as I can tell, --limit-download is not affected by this problem,
as Get().Download() returns the real http.Response.Body without any
internal buffering.

Updates #1216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Important Changes in 0.X.Y
    `--limit-download` flags.
    https://github.com/restic/restic/issues/1216
    https://github.com/restic/restic/pull/1336
+   https://github.com/restic/restic/pull/1358
 
  * Failed backend requests are now automatically retried.
    https://github.com/restic/restic/pull/1353
--- a/internal/backend/gs/gs.go
+++ b/internal/backend/gs/gs.go
@@ -214,10 +214,37 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd io.Reader) (err
 
 	debug.Log("InsertObject(%v, %v)", be.bucketName, objName)
 
+	// Set chunk size to zero to disable resumable uploads.
+	//
+	// With a non-zero chunk size (the default is
+	// googleapi.DefaultUploadChunkSize, 8MB), Insert will buffer data from
+	// rd in chunks of this size so it can upload these chunks in
+	// individual requests.
+	//
+	// This chunking allows the library to automatically handle network
+	// interruptions and re-upload only the last chunk rather than the full
+	// file.
+	//
+	// Unfortunately, this buffering doesn't play nicely with
+	// --limit-upload, which applies a rate limit to rd. This rate limit
+	// ends up only limiting the read from rd into the buffer rather than
+	// the network traffic itself. This results in poor network rate limit
+	// behavior, where individual chunks are written to the network at full
+	// bandwidth for several seconds, followed by several seconds of no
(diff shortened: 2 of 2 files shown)


fd0:
Oh, good point, thanks for digging into the problem!

I'd like to avoid exposing this option to end users, is there any harm in using 256KiB by default?

From the backend, it's not possible to directly see if the upload limit is configured, so we'd need to pass that in when opening the backend. The upload limit divided by four or eight should be a good heuristic.

What do you think?

prattmic:
Actually, looking into it more, I think it might be reasonable to set the chunk size to 0 by default.

The GCS library is doing this chunking to implement [resumable uploads](https://cloud.google.com/storage/docs/json_api/v1/how-tos/resumable-upload). It uploads the object in chunk size chunks and has automatic retry and exponential backoff if chunk upload fails.

The main purpose of resumable uploads is to avoid having to re-upload lots of data if a request fails. (Max re-upload is the chunk size). However, the default chunk size is 8MB, and if I understand correctly, the maximum pack size is 12MB (4MB min pack size + 8MB max blob size), so the default chunk size isn't much more than the maximum object size we'll upload anyways.

Setting the chunk size to 0 disables resumable uploads altogether and uploads the entire object in a single request. The passed Reader is used more directly, without explicit large buffers.

This wouldn't eliminate all buffering; the GCS libraries still io.Copy the Reader to an io.Pipe(), and net/http does a fair amount of io.Copy'ing itself, but I suspect this would get the best rate limiting behavior we can easily achieve.

Thoughts?

fd0:
I think we can set the chunk size to zero.

prattmic:
PTAL

fd0 (approved):
LGTM!

fd0:
`benchcmp` agrees:
```
benchmark                                               old bytes     new bytes     delta
BenchmarkBackendGS/BenchmarkLoadFile-4                  373032        344088        -7.76%
BenchmarkBackendGS/BenchmarkLoadPartialFile-4           72112         72680         +0.79%
BenchmarkBackendGS/BenchmarkLoadPartialFileOffset-4     72336         84205         +16.41%
BenchmarkBackendGS/BenchmarkSave-4                      8581304       206912        -97.59%
```

PR #1365, opened 2017-10-18
Fix failure to detect some legacy s3 repos

### LINKED ISSUE #1068: Restic 0.6.0 onwards fails to run with legacy S3 repo
#### FROM
restic 0.5.0 (v0.5.0-0-gf678c97)
compiled with go1.8 on linux/amd64
#### TO
restic 0.6.1

./restic_0.6.1_linux_amd64 check

AWS S3

## Expected behavior
To check the repo for errors

## Actual behavior or output
unable to open repo at s3://s3.amazonaws.com/myrepo: ReadDir: invalid key name restic/key/, removing prefix restic/key/ yielded empty string

Use restic 0.6.0 onwards on a 0.5.0 formatted S3 repo.

Not sure. Maybe no longer allows using the legacy repo.

## What have I tried
Compiled latest restic and attempted to run the migrate command on repo. Migrate produces the same error as above like all other commands.

felix9:
Sometimes s3 listobjects for a directory includes an entry for that
directory. The restic s3 backend doesn't expect that and returns
an error.

Symptom is:
  ReadDir: invalid key name restic/key/, removing prefix
     restic/key/ yielded empty string

I'm not sure when s3 does that; I'm unable to reproduce it myself.

But in any case, it seems correct to ignore that when it happens.

Fixes #1068

### DIFF
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -172,8 +172,9 @@ func (be *Backend) ReadDir(dir string) (list []os.FileInfo, err error) {
 		}
 
 		name := strings.TrimPrefix(obj.Key, dir)
+		// Sometimes s3 returns an entry for the dir itself. Ignore it.
 		if name == "" {
-			return nil, errors.Errorf("invalid key name %v, removing prefix %v yielded empty string", obj.Key, dir)
+			continue
 		}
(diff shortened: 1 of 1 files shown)


fd0 (approved):
LGTM, thanks!

Issue #1370, opened 2017-10-20
Concurrent restorer


is:
## Output of `restic version`

```
debug enabled
restic 0.7.3 (v0.7.3-172-ga8dc7934-dirty)
compiled with go1.9.1 on linux/amd64
```

## How did you run restic exactly?

```
restic restore -c 40 --target _ 18a2cde
```

## What backend/server/service did you use?
Any

## Actual behavior
Unlike backup, restore is a very slow process, especially when a lot of small files there.

## Steps to reproduce the behavior
restore a big snapshot (10G+, 40000+ dirs and files) to local filesystem.

## Do you have any idea what may have caused this?
The default restorer is worked in single thread.

## Do you have an idea how to solve the issue?
I try to rewrite a restorer, with worked in multiple goroutines pool.

background: I run restic on a stronger server, with 128G ram and 10G NIC.
I hope to finish backup/restore as soon as possible, without thinking to
much about CPU/memory consumption.

creckx:
:+1: 

I would like to say this is the only thing which keeps us from using restic in production. Backup is superfast into our own minio cluster or SFTP connection, but restore is annoyingly slow. It's fine if you are not in hurry like with photos, but usually we are, because if we need the backup it means the service is down.

fd0:
@is thanks for tackling this issue! I'll try to find time to review your PR the next couple of days and give you feedback.

dionorgua:
I've found that `restic mount` is currently much faster than restore for me (at least when I need to restore just one particular directory that is less than ~10% of snapshot)

fd0:
This has been resolved, the restorer is concurrent now. I'm closing this issue.

PR #1371, opened 2017-10-20
Concurrent restorer.

is:
for #1370. Pass by some tests for actual data snapshots recovery.

add a command line option -c/--worker to set concurrent workers number.

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -33,6 +33,7 @@ type RestoreOptions struct {
 	Host    string
 	Paths   []string
 	Tags    restic.TagLists
+	Workers int
 }
 
 var restoreOptions RestoreOptions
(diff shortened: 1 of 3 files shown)


fd0 (changes requested):
Hey, thanks a lot for giving this a try, and please accept my apologies for the long delay. I'd like to get this merged into restic, but before we can do that the following changes need to be done:

 * Please rebase the PR onto the latest master branch
 * Remove the old restorer, insert the new, concurrent one in place, rename it to `Restorer` etc.
 * Add tests which restore with different levels of concurrency
 * Format the commit message according to our general style, e.g. "Add concurrent restorer" (describe what the commit does, no punctuation)

Once you've done that we can have an in-depth look at the code.

Please let me know if you need any help and/or clarification. Thanks!

is:
@fd0 I amended the code, it will pass all existed cases. 

I will duplicate the cases in restorer_test, and set more concurrent worker, is it OK?.

I am not very familiar with go test, any sugguestions?

fd0:
I'm closing this for now, it is superseded by #1719. Thanks!

PR #1381, opened 2017-10-24
Remove Markdown syntax from documentation

tbm:
Convert a link from Markdown to RST syntax.  This link must
have been missed during the recent conversion.


fd0:
Thanks a lot!

PR #1384, opened 2017-10-25
Reduce ticker refresh rate with RESTIC_PROGRESS_FPS env

wojas:
With the old refresh rate of 60 fps, iTerm2 (macOS) typically uses ~40% CPU to keep up. This changes the refresh rate to 5 fps.

This makes the ticker feel less 'smooth', but the CPU usage can be better utilized for hashing files or other tasks.

### DIFF
--- a/internal/restic/progress.go
+++ b/internal/restic/progress.go
@@ -3,16 +3,30 @@ package restic
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-const minTickerTime = time.Second / 60
+// minTickerTime limits how often the progress ticker is updated. It can be
+// overridden using the RESTIC_PROGRESS_FPS (frames per second) environment
+// variable.
+var minTickerTime = time.Second / 60
 
 var isTerminal = terminal.IsTerminal(int(os.Stdout.Fd()))
 var forceUpdateProgress = make(chan bool)
+
+func init() {
+	fps, err := strconv.ParseInt(os.Getenv("RESTIC_PROGRESS_FPS"), 10, 64)
+	if err == nil && fps >= 1 {
+		if fps > 60 {
+			fps = 60
+		}
+		minTickerTime = time.Second / time.Duration(fps)
+	}
+}
 
 // Progress reports progress on an operation.
 type Progress struct {


fd0:
Huh, that is odd, I've never received any similar reports. I'm reluctant to merge this as is because with just 5 fps maximum frame rate restic feels much more sluggish, not so snappy any more. I've deliberately used an upper bound of 60fps...

Can somebody else please try to confirm your observations? Is your setup special in any way (like, unusual iterm2 settings)? In order to reproduce it, this needs a subsequent backup in which restic churns through many files very fast, otherwise there will be less than 60 updates per second.

mannp:
How are you proposing to test it? to be able to see the specific refresh cpu load compared to restic going about its usual tasks to do a backup?

fd0:
Make a backup of a sufficiently large data set (not sure what that is), then run restic again in iterm2 and observe the CPU load? I'm not so sure myself.

mannp:
But doesn't that mean we must be able to change the refresh rate between backups to compare.

Maybe the OP can shed some light on how they came to the 40% on refresh alone (if I have understood that correctly)

wojas:
I did some more testing. One way for me to reproduce it is to start a local backup of a large directory with many files.  The CPU usage for iTerm2 will jump to over 30%. With my patch, this drops to a few percent.

I have font anti-aliasing enabled in iTerm2. If I disable it, the CPU usage drops to a few percent. If I run it in Terminal.app, it uses about 8%.

You could argue that I should disable anti-aliasing, but running restic with a progress indicator is the only thing that exhibits issue that I know of. There is no other console tool that updates this frequently and restic can run for hours for a huge backup. Catting a large logfile might trigger it, but only for a few seconds.

Checking iStat Menus, the typical screen update frequency that is reported for my GPU under normal usage is 3-10 fps. This jumps to ~56 fps with the restic progress indicator at the old value, indicating that the feature is at least working as intended.

Affected hardware: MacBook Pro (Retina, 13-inch, Late 2013)

wojas:
Giving it some more thought, I think my pick of 5 fps is actually a bad choice: it is fast enough for the brain to expect realtime updates, but not fast enough to blur the last fast updating digits to make it feel smooth and make it easy to absorb the information. 

So I would actually argue for a default update interval of 1 second:

- This is the common choice for other utilities
- It does does not give the realtime expectation of higher rates
- It updates slow enough for a human to actually be able to read the numbers
- It minimizes CPU usage, while providing adequate updates on the progress

In order not to complicate restic with unnecessary command line flags, I would suggest not to add one to revert to the old behavior. But I could add something like a `RESTIC_PROGRESS_FPS` environment var to allow users to revert to the old behavior, as I fully admit it makes demos look way cooler. 😄

fd0:
Hm, I'd be willing to merge a PR which adds code to set the maximum number of frames per second via `RESTIC_PROGRESS_FPS`, with a default of 60. What do you think?

wojas:
I added a commit that reverts the default update rate to 60 fps and allows overriding it with an env var. I'm not sure where to add this to the documentation.

I still feel the default is a bit high, as even the 8% CPU usage with Terminal.app exceeds the idle CPU usage of all my other processes combined.

fd0 (changes requested):
LGTM, would you mind squashing the commits? The first one isn't really relevant. Thanks!

wojas:
@fd0 I squashed the commits. Thanks!

fd0 (approved):

PR #1387, opened 2017-10-26
Add --last flag to snapshots command

wojas:
Add --last flag to snapshots command to only show the last entry for any (hostname, paths) combination.

This makes it easier to check when various paths were last backed up.


fd0 (approved):
LGTM

Issue #1385, opened 2017-10-25
small cleanup, "lint" issues


armhold:
I'd like to contribute a few minor code cleanups. Mostly cosmetic, but with the goal of shining light on potential bugs, ignored errors, etc. Sample commit [here](https://github.com/armhold/restic/commit/bcdebfb84e249738d6573fd678f8ff480f09e610).

If this kind of thing is useful, I propose to discuss & track such changes here. Thanks!

fd0:
Please do so, thanks!

armhold:
How do you feel about this ignored error from f.Close() [in a defer](https://github.com/restic/restic/blob/master/internal/restic/node.go#L256)? 

My feeling is that since there's a write, errors from that close are significant, and should probably be returned to the caller. (Feel free to talk me out of that... am I wrong?)

I know the resulting error handling in this situation is kind of ugly. We could:

1) Change the func signature to have a named error, and have the defer check if that named err is nil, and if so, return the result of the close.

2) Remove the defer, put `f.Close()` in front of every return statement in the for loop.

3) Remove the defer, refactor that for loop into its own func (maybe even inline anonymous), and move the `f.Close()` after the error check from the call to *that* func.

fd0:
Oh, good catch, I feel the ignored error here is bug! :)

I'd go for option 3, with a named function, which should improve readability.

armhold:
OK this one might be significant for B2 users. [defaultListMaxItems](https://github.com/restic/restic/blob/master/internal/backend/b2/b2.go#L28) is currently unused in the codebase. Shouldn't this be used like so?

```
	be := &b2Backend{
		client: client,
		bucket: bucket,
		cfg:    cfg,
		Layout: &backend.DefaultLayout{
			Join: path.Join,
			Path: cfg.Prefix,
		},
		listMaxItems: defaultListMaxItems,  // added this line
		sem: sem,
	}
```

Otherwise it defaults to zero, which has the effect of using the [B2 default of 100](https://www.backblaze.com/b2/docs/b2_list_file_names.html) for listing files, rather than the "restic" default of 1000.

If it really is intentional to default to 0/100 (maybe it was perf-tested to work well?) then I would make that explicit in the const, and pass in the setting as above.

armhold:
Is there a good reason for the receivers on [Node](https://github.com/restic/restic/blob/master/internal/restic/node.go) to not all be pointer receivers? 

[...]

PR #1390, opened 2017-10-26
small cleanup

armhold:
As discussed in gh-1385.

- be explicit when discarding returned errors from .Close(), etc.
- remove named return values from funcs when naked return not used
- fix some "err" shadowing when redeclaration not needed

### DIFF
--- a/internal/backend/b2/b2.go
+++ b/internal/backend/b2/b2.go
@@ -218,7 +218,7 @@ func (be *b2Backend) Load(ctx context.Context, h restic.Handle, length int, offs
 }
 
 // Save stores data in the backend at the handle.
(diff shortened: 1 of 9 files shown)


fd0 (approved):
Very nice, thanks a lot!

fd0:
A few files are not formatted with `gofmt`. 
I'd like to suggest installing `fmt-check` from https://github.com/edsrzf/gofmt-git-hook as a pre-commit hook (just copy it to `.git/hooks/pre-commit`).

armhold:
Sorry about that. I actually did take care to run "gofmt", but I ran `gofmt -w **/*.go`, which produced no output (if you squint I think you can see the problem there. 😄)

I will install the pre-commit hook as you suggested.

PR #1393, opened 2017-10-27
detect errors from fs.Walk() in local backend List()

armhold:
I found another place where failing to check an error might hide a bug, or at least a problematic repo. 

The scenario where this matters is when there is a problem in a local repo, e.g. permissions setting on a data/FOO directory. This will cause "list packs" and similar code paths to silently stop listing data when it encounters the problematic dir. 

My solution here isn't really satisfactory, as it merely logs it, and only to the debug log (not really user-visible). What do you think?

Related to gh-1385.

### DIFF
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -252,23 +252,27 @@ func (b *Local) List(ctx context.Context, t restic.FileType) <-chan string {
 	go func() {
 		defer close(ch)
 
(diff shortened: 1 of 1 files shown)


fd0:
Yeah, that's not ideal but better than not logging it at all. I've got plans to make the `List()` backend function return an error somehow, maybe via the channel. I'm not sure yet. For now this commit improves the code.

fd0 (approved):

PR #1394, opened 2017-10-27
Use lowercase in messages from check/tag commands, too

zcalusic:
Almost all commands output in lowercase. Let's bring check & tag commands to that standard, too.


fd0 (approved):
LGTM!

PR #1397, opened 2017-10-28
crypto: Make crypto.Key implement cipher.AEAD

### LINKED ISSUE #1190: internal/crypto: Decrypt overlap rules are a bit off
There seems to be a problem with the `Decrypt` overlap rules:

> plaintext and ciphertext may point to (exactly) the same slice.

The `cipher.Stream` interface implemented by CTR mode says this at the method `XORKeyStream`:

> Dst and src may point to the same memory.

Which isn't that helpful. There is already a discussion about it at golang/go#21279 but I think it means that `dst` and `src` may overlap entirely (without misalignment) or not at all.

However, calling `Decrypt` with the same buffer for plaintext and ciphertext (as done throughout the code and allowed by the docs) means that `buf[16:]` is decrypted into `buf[:]` because of the IV. That's theoretically(?) not ok.

I don't think it's a problem right now, but it might become one as the CTR implementation [is optimized](https://github.com/golang/go/issues/20967).

Also, I'm not sure what this is supposed to mean in the `Encrypt` docs, but overlapping buffers are never used in the codebase anyway.

> ciphertext and plaintext may not point to (exactly) the same slice or non-intersecting slices.

fd0:
This was suggested by @FiloSottile in #1190

Fixes #1190


internal/crypto/crypto.go
```diff
@@ -233,79 +235,127 @@ func (k *EncryptionKey) Valid() bool {
...
 
+	dst = dst[:cap(dst)]
+	plaintextLength := len(ciphertext) - k.Overhead()
+
 	// check buffer length for plaintext
-	plaintextLength := len(ciphertextWithMac) - Extension
-	if len(plaintext) < plaintextLength {
-		return 0, errors.Errorf("plaintext buffer too small, %d < %d", len(plaintext), plaintextLength)
+	if len(dst) < plaintextLength {
+		return nil, errors.Errorf("plaintext buffer too small, %d < %d", len(dst), plaintextLength)
```
FiloSottile: Open is an append-like API. It should not require a preallocated buffer, dst can be nil.

internal/checker/checker.go
```diff
@@ -724,15 +724,21 @@ func checkPack(ctx context.Context, r restic.Repository, id restic.ID) error {
...
 
-		n, err := r.Key().Decrypt(buf, buf)
+		if len(buf) < r.Key().NonceSize()+r.Key().Overhead() {
+			errs = append(errs, errors.Errorf("blob %v: invalid encrypted data, too short", i))
+			continue
+		}
+
+		nonceSize := r.Key().NonceSize()
+		nonce, ciphertext := buf[:nonceSize], buf[nonceSize:]
+		plaintext, err := r.Key().Open(ciphertext, nonce, ciphertext, nil)
```
FiloSottile: Also, since it’s an append API it should not overwrite the slice, but add data after it. That’s why you use ciphertext[:0].

FiloSottile (changes requested):
I had a very quick look, and it’s a good change, but I think the API is a little off.

internal/crypto/crypto.go
fd0: Ah, you're right.

internal/checker/checker.go
fd0: Ok, need to make that sure. Thanks!

fd0:
Thanks a lot for your comments, I'll fix it.

fd0:
Thank you very much for your comments, I indeed misunderstood the API. I've corrected the implementation, and added tests for the corner cases.

Please take another look!

internal/repository/key.go
```diff
@@ -87,12 +87,12 @@ func OpenKey(ctx context.Context, s *Repository, name string, password string) (
 	}
 
 	// decrypt master keys
-	buf := make([]byte, len(k.Data))
-	n, err := k.user.Decrypt(buf, k.Data)
+	nonce, ciphertext := k.Data[:k.user.NonceSize()], k.Data[k.user.NonceSize():]
+	buf := make([]byte, 0, len(ciphertext))
+	buf, err = k.user.Open(buf, nonce, ciphertext, nil)
```
FiloSottile: Here and elsewhere you could skip the `make([]byte, 0, len(ciphertext))` and pass nil instead, as that will be done automatically.

internal/crypto/crypto.go
```diff
@@ -233,91 +235,142 @@ func (k *EncryptionKey) Valid() bool {
...
-// Encrypt encrypts and authenticates data. Stored in ciphertext is IV || Ciphertext ||
-// MAC. Encrypt returns the new ciphertext slice, which is extended when
-// necessary. ciphertext and plaintext may not point to (exactly) the same
-// slice or non-intersecting slices.
-func (k *Key) Encrypt(ciphertext []byte, plaintext []byte) ([]byte, error) {
+// validNonce checks that nonce is not all zero.
+func validNonce(nonce []byte) bool {
+	sum := 0
+	for b := range nonce {
+		sum += b
```
FiloSottile: This might be faster with `|=` and `var sum byte`.

internal/crypto/crypto.go
```diff
@@ -233,91 +235,142 @@ func (k *EncryptionKey) Valid() bool {
...
+	ciphertextLength := len(plaintext) + k.Overhead()
+	pos := len(dst)
+
+	capacity := cap(dst) - len(dst)
+	if capacity < ciphertextLength {
+		dst = dst[:cap(dst)]
+		dst = append(dst, make([]byte, ciphertextLength-capacity)...)
+	} else {
+		dst = dst[:pos+ciphertextLength]
+	}
```
FiloSottile: The common pattern here is `sliceForAppend`.

https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/crypto/aes/aes_gcm.go#L81-L94

FiloSottile (approved):

internal/crypto/crypto.go
fd0: You're right, it's ever so slightly faster, at least on amd64:
```
$ benchcmp bench-int bench-byte
benchmark                 old ns/op     new ns/op     delta
BenchmarkNonceValid-4     9.45          8.22          -13.02%
```

But, much more importantly: After I made `sum` a `byte`, the compiler suddenly complained that `b` had the wrong type: `int` instead of `byte`. Turns out, the function has a very subtle bug: It didn't sum up the elements of `nonce`, but the indexes. :disappointed: 

So, thanks again for the review, it helped find a bug! (And I've also added tests for thatt).

internal/crypto/crypto.go
fd0: Oh, I wasn't aware of that. Thanks for the pointer! The pattern is really nice and reduces the complexity of the code in Seal/Open :)

internal/repository/key.go
fd0: Good point, although I did only find this one occurrence. I've replaced it.

fd0:
Thank you very much for your help @FiloSottile!

internal/crypto/crypto.go
FiloSottile: Woo, good catch!

PR #1399, opened 2017-10-29
prevent deadlock in List() for B2 when b2.connections=1

### LINKED ISSUE #1188: restic hangs when b2.connections is set to 1
Setting b2.connections >= 2 overloads my internet connection. Setting it = 1 causes restic to hang.

## Output of `restic version`
```
restic compiled manually
```

(was built from restic/restic master on Aug 10).

I have set RESTIC_PASSWORD,  B2_ACCOUNT_ID and  B2_ACCOUNT_KEY appropriately. 

Then I use the following bash alias: 

`alias restic-b2='caffeinate restic -r b2:MY-BUCKET:restic backup ~ /Volumes/data1 --hostname george-iMac --exclude-file ~/restic-excludes --exclude-file ~/s3-excludes -o b2.connections=1'`

NB: `caffeinate` is a standard macOS program that prevents the computer from going to sleep while running another shell command.

B2

## Expected behavior

restic should not hang when using only a single b2.connection. 

## Actual behavior

Restic seems to hang, does not even get to the "scan" phase (as far as I can tell) and does not produce a single line of output after several hours.

Set vars as above to real credential values, then run `restic-b2` as a non-root user in bash.

Wondering if perhaps there is a deadlock situation regarding worker threads.

I can run a debug build if you think that would be helpful.

armhold:
This is a fix for the following situation (gh-1188):

List() grabs a semaphore token upon entry, starts a goroutine, and
does not release the token until the routine exits (via a defer).

The goroutine iterates over the results from ListCurrentObjects(),
sending them one at a time to a channel, where they are ultimately
processed by be.Load().

Since be.Load() also needs a token, this will result in deadlock if
b2.connections=1.

This fix changes List() so that the token is only held during the call
to ListCurrentObjects().

Closes #1188

### DIFF
--- a/internal/backend/b2/b2.go
+++ b/internal/backend/b2/b2.go
@@ -307,18 +307,17 @@ func (be *b2Backend) List(ctx context.Context, t restic.FileType) <-chan string 
 
 	ctx, cancel := context.WithCancel(ctx)
 
-	be.sem.GetToken()
-
 	go func() {
 		defer close(ch)
 		defer cancel()
-		defer be.sem.ReleaseToken()
 
 		prefix := be.Dirname(restic.Handle{Type: t})
 		cur := &b2.Cursor{Prefix: prefix}
 
 		for {
+			be.sem.GetToken()
 			objs, c, err := be.bucket.ListCurrentObjects(ctx, be.listMaxItems, cur)
+			be.sem.ReleaseToken()
 			if err != nil && err != io.EOF {
 				return
 			}


fd0 (approved):
Very nice, that'll fix it. Thanks!

Did you maybe have a look at the other backends, if the same problem exists there also?

armhold:
Yes I did look there, and I meant to bring that up with you. In the rest backend, it follows the same pattern as my fix: lock briefly during the "remote list" portion.

For the others (azure, gs, s3, swift) they don't seem to lock at all during List(); only during Load/Save. And sftp doesn't seem to have this "# of connections" feature at all.

I'm not super familiar with those backends yet, so I don't know if List() is a "cheap" operation for them or not. So I'm not sure if this was an intentional design decision or an oversight.

fd0:
Limiting the number of connections was a first attempt at controlling the load for remote backends. I'm not entirely sure that `List()` should even fall under that limit, it was mainly intended to limit large upload/download connections (`Save()` and `Load()`).

SFTP does not have a notion of "connections", since all data and requests are multiplexed in one TCP connection.

I'd like to unify behavior across the HTTP based backends (all except local and sftp), I think that's a good idea. What's your take on this? How would you proceed?

armhold:
Yeah, the `limit-upload` and `limit-download` enhancements are a massive improvement, and for me at least, they mostly obviate the need to micro-manage http connection counts. I just wasn't familiar enough with the codebase to fix the issue at the time I reported this bug.

> I'm not entirely sure that List() should even fall under that limit,

I really have no idea if List() can ever take a "long" time. If it can, it should probably still be managed by the semaphore. Either way, I can imagine server scenarios where concurrent http connections are precious and metered in some fashion. In that case I think you'd want the semaphore here, even if it's not held for long.

As for unifying, I need to think on this a bit. My first thought was to incorporate semaphore acquisition/release into something like a wrapReader/wrapWriter struct. But that doesn't cover cases like obj.Attrs(ctx), which is being used to check that a file exists or not in B2, prior to saving a file. There's no reader/writer exposed for that call. There are similar complications with the other backends. Tough to find a general solution here.

fd0:
Ah, "unifying" can also mean "fix the code so that it behaves the same in all backends", no need to find a generic solution in which backends are wrapped.

armhold:
Ah, ok good, because a wrapping abstraction does not seem to add much value here (I think it would,
if it could be uniformly applied).

Codewise, it's easy to see what needs to change: just trace calls in the current code to
`sem.GetToken()` and it's obvious where the difference in APIs is:

semaphore-api-calls

I'm happy to make the required changes to make all the backends behave uniformly across the various
API calls, but I am somewhat reluctant to commit/PR code that I cannot personally test (swift,
azure, gcs)... do we have people using those backends willing to spend time doing some thoughtful
testing?

As for how to "unify"- I think you need to decide if this feature is primarily a bandwidth knob, or
primarily an http connection knob. I ran some (simplistic) [timing tests](https://github.com/restic/restic/files/1427193/timing.log). Note that this is for B2,
and I have no idea if the other backends behave similarly.

You are correct that `List()` seems to be a rounding error vs almost all the other calls. But there
were a couple of surprises for me:

> Stat took: 959ms (timing/config)
> Remove took: 916ms (timing/locks/53391e94231fd1c388b669a3a2663fb99f375d4d8ba412d1b17487f23215ad1a)

My guess is that these are delays in the B2 API, and probably have close to zero impact on
bandwidth. So I would ask again: do you want to limit bandwidth or simultaneous connections with
this feature?

Side observation: it's CRAZY how many repeat `Load()` calls there are for the same data file with
`--no-cache`!

fd0:
I've made sure that I have access credentials for all the different backends, and the CI also runs those tests when the when the credentials are available, so we'll see if something breaks.

I think the knobs here are primarily meant to control the number of concurrent connections, the bandwidth can be limited with the `--limit-*` options.

armhold:
I ran into a couple of issues while unifying the backend behavior.

1. In the current code for the swift backend, Load() releases the semaphore on return, but it
   returns a ReadCloser. So the sempahore is not really in force for the duration of the download.

   This can be fixed similar to B2 by using a wrapReader; Should I duplicate this code, or can we
   move wrapReader elsewhere so the other backends can use it?

2. In s3, the underlying ListObjects() api returns a channel for results. Adding a semaphore around
   that call would create a deadlock situation just like the one I fixed in b2. I don't see a clear
   solution here...

fd0:
to 1): I consider this a bug, and maybe we can make the wrapReader a function of the semaphore, which returns a token when it sees `io.EOF` or `Close()`?

to 2): I don't have an idea either at the moment.

armhold:
OK, I did my best with this. Is there a way to "re-open" a PR for my recent commits above, or should I create a new one and just reference this discussion from there?

PR #1409, opened 2017-11-02
[WIP] Add --subtree flag to find command to limit find to a specific subtree.

asannes:
This just adds a --subtree option to restic find command to limit where find is looking.


asannes:
If this is okay, should I do a similar thing to the ls command in a separate PR or just add it here?

asannes:
Going to change how the subtree is found, to make it a bit more userfriendly.

rawtaz:
@asannes For those who don't know, what problem does this solve, what is the use case? Is there an issue or similar where it's been discussed?

fd0:
The build on AppVeyor fails, do you need help here?

fd0:
Are you still interested in continuing this? Do you need help?

asannes:
Not stuck, just ran out of time for a while, need to fix that test and for that I need a windows computer I think.

fd0:
FYI, there's a merge commit in the list of commits that doesn't belong here. Can you rebase this on master and squash all commits into one?

fd0:
There hasn't been any activity here for the last few weeks so I'm closing this PR. Just leave a comment if you're still interested in continuing the work (or need help), thanks!

PR #1410, opened 2017-11-02
unify behavior for max http connections across backends

armhold:
As we discussed in gh-1399, this PR aims to unify the behavior of the http backends with regard to the number of allowed simultaneous http connections.


fd0:
Thanks!

PR #1389, opened 2017-10-26
Fix modtime tests on APFS: allow 1us difference

wojas:
Fixes #1087.

This fixes the tests failing on APFS as reported in #1087  by allowing a difference in modtime of up to 1 microsecond.

I'm not sure if this is the right approach. Maybe we should increase the precision that we store in backups instead?

Issue #1307, opened 2017-10-01
test fails on OpenBSD


bket:
Regression test fails on OpenBSD with the error below. This error is caused by a comparison in restic/cmd/restic/integration_helpers_unix_test.go:

```
        if !sameModTime(e.fi, other.fi) {
                fmt.Fprintf(os.Stderr, "%v: ModTime does not match (%v != %v)\n", e.path, e.fi.ModTime(), other.fi.ModTime())
                return false
        }
```

It seems that e.fi.ModTime() uses nanosecond accuracy while other.fi.ModTime() uses microsecond accuracy. Comparing the two will result in false, which is negated thus the if-statement will return false.

Is it by design that restic's make test uses two different time resolutions? Is this also the case for restic itself?

```
===>  Regression tests for restic-0.7.3
cd /home/code/ports/pobj/restic-0.7.3/go/src/github.com/restic/restic && /usr/bin/env -i PATH="/home/code/ports/pobj/restic-0.7.3/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/X11R6/bin" TMPDIR="/home/code/ports/pobj/restic-0.7.3/build-amd64" GOPATH="/home/code/ports/pobj/restic-0.7.3/go:/usr/local/go-pkg" PATH="/home/code/ports/pobj/restic-0.7.3/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/X11R6/bin" go run build.go --test
ok  	github.com/restic/restic/internal/archiver	101.494s
ok  	github.com/restic/restic/internal/backend	1.417s
ok  	github.com/restic/restic/internal/backend/azure	0.009s
ok  	github.com/restic/restic/internal/backend/b2	0.009s
ok  	github.com/restic/restic/internal/backend/gs	0.009s
ok  	github.com/restic/restic/internal/backend/local	2.181s
ok  	github.com/restic/restic/internal/backend/location	0.038s
ok  	github.com/restic/restic/internal/backend/mem	1.359s
ok  	github.com/restic/restic/internal/backend/rest	0.009s
ok  	github.com/restic/restic/internal/backend/s3	0.054s
ok  	github.com/restic/restic/internal/backend/sftp	9.276s
ok  	github.com/restic/restic/internal/backend/swift	0.009s
ok  	github.com/restic/restic/internal/backend/test	0.803s
ok  	github.com/restic/restic/internal/checker	6.167s
[...]

Issue #1087, opened 2017-07-06
Restic on APFS needs small adjustments


lloeki:
## Output of `restic version`

```
$  ./restic version
debug enabled
restic 0.6.1 (v0.6.1-73-gaabe2a0-dirty)
compiled with go1.8.1 on darwin/amd64
```

## How did you start restic exactly? (Include the complete command line)

```
$ gb test cmds/restic
```

## What backend/server/service did you use?

Local FS.

## Expected behavior

Tests should pass.

## Actual behavior

Using 10.13 beta on APFS, there is a precision discreptancy. I don't know what the deal is with the lchown.

```
$ gb test cmds/restic
# cmds/restic
Now serving the repository at /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-mount-242244809
Don't forget to umount after quitting!
Now serving the repository at /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-mount-242244809
Don't forget to umount after quitting!
Now serving the repository at /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-mount-242244809
Don't forget to umount after quitting!
Now serving the repository at /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-mount-242244809
Don't forget to umount after quitting!
Now serving the repository at /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-mount-055822315
Don't forget to umount after quitting!
error for testdata/0/0/9/37: Lstat: lstat /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-149018899/testdata/0/0/9/37: no such file or directory
error walking dir testdata/0/0/9/37: Lstat: lstat /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-049004756/testdata/0/0/9/37: no such file or directory
error walking dir 0/9: Open: open /var/folders/8n/00j6g_xx08qb21g22s_jj7sh0000gn/T/restic-test-184104134/testdata/0/0/9: no such file or directory
 ID          User        Host        Created
----------------------------------------------------------------------
*dc3fff31    lloeki      Galileo.local  2017-07-06 17:28:03
foo: ModTime does not match (2017-07-06 17:28:04.513803084 +0200 CEST != 2017-07-06 17:28:04.513804 +0200 CEST)
[...]

PR #1414, opened 2017-11-05
Fix test failing on OpenBSD #1307

bket:
Is seems that #1307 is similar to #1087, which describes a comparable
observation on Apple's new filesystem. #1389 Has been committed and
fixes the problem on Darwin.

Although I'm not sure if the root cause of the issue is the same the
solution is similar for OpenBSD, and leverages #1389.


fd0 (approved):
Fine with me.

PR #1415, opened 2017-11-05
also handle SIGPIPE in cleanup routines

### LINKED ISSUE #1413: restic fails to cleanup locks when bash pipeline fails
I think I've solved the mystery of why my restic repos sometimes get mysteriously locked. If you run restic in a bash pipe, and the pipe fails for some reason (e.g. typo), restic fails to clean up its lock.

## Output of `restic version`

```
debug enabled
restic 0.7.3 (v0.7.3-235-gf39f7c76)
compiled with go1.9.2 on darwin/amd64
```

I was trying to cat an index through 'jq' but I made a typo (notice the lack of space between `jq` and its arg in my command line):

`restic -r ~/repo cat index 8a3288733c47ccb02e830fc3f980ae4c0f019a2c94ea6d6d09dc4026879662ee |jq'.' | head`

local

## Expected behavior

Restic should not leave locks behind on SIGPIPE.

## Actual behavior

Restic leaves the lock behind.

Pipe restic `cat index` into a bogus bash command, e.g.:

`restic -r ~/repo cat index 8a3288733c47ccb02e830fc3f980ae4c0f019a2c94ea6d6d09dc4026879662ee | bogus`

Not handling SIGPIPE.

Perhaps adding SIGPIPE to the list of signals ignored in lock.go:

```
signal.Notify(c, syscall.SIGPIPE)
```

armhold:
fixes gh-1413: restic fails to cleanup locks when bash pipeline fails.

Something else to consider: is it still appropriate to exit(0) when a SIGPIPE is received, as we currently do for SIGINT?

If the bash pipeline fails, even if it's not the fault of restic, perhaps restic should maybe exit(1)? I think it would be bad for restic to give the false impression of success (0) when in fact it had a problem. I don't really know what standard practice is.

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -12,27 +12,29 @@ import (
 
 var cleanupHandlers struct {
 	sync.Mutex
-	list     []func() error
-	done     bool
-	sigintCh chan os.Signal
+	list []func() error
+	done bool
+	ch   chan os.Signal
 }
 
 var stderr = os.Stderr
 
 func init() {
-	cleanupHandlers.sigintCh = make(chan os.Signal)
-	go CleanupHandler(cleanupHandlers.sigintCh)
+	cleanupHandlers.ch = make(chan os.Signal)
+	go CleanupHandler(cleanupHandlers.ch)
 	InstallSignalHandler()
 }
 
-// InstallSignalHandler listens for SIGINT and triggers the cleanup handlers.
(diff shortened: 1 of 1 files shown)


fd0:
Hmm, good point. I'm also not sure, but I lean towards using a non-zero exit code in this case.

armhold:
> Cloning into 'restic/restic'...
fatal: unable to access 'https://github.com/restic/restic.git/': Could not resolve host: github.com

This looks like some kind of transitory DNS failure I think.

PR #1417, opened 2017-11-06
Docker improvements

thomasleveil:
The main change is to allow [docker multi stage build](https://docs.docker.com/engine/userguide/eng-image/multistage-build/).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+*
+!cmd
+!internal
+!vendor
+!*.go
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,13 @@
+# Docker 17.05 or higher
+FROM golang:1.8.3-alpine as builder
+COPY . /go/src/github.com/restic/restic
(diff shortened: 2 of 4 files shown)


fd0:
Hi, thanks for trying to optimize the docker build process, and please accept my apologies for the long delay. I don't want to merge this PR as it is, but instead I'd like to:
 * Move the Dockerfile and the docker build process to a separate repository, https://github.com/restic/docker (to be created)
 * The restic build process (which we're using to build reproducible binaries for our releases) is already in a Docker container, https://github.com/restic/builder, so we don't need that here
 * The new Dockerfile should just take a `restic` binary from the current directory (e.g. the official build of `restic`) and build an Image, that we can then publish as https://hub.docker.com/r/restic/restic/. We should also look at the (very popular) image https://hub.docker.com/r/jsmitsnl/docker-restic/ and try to convince them to collaborate with us :)

I'm going to close this PR for now, but I'd be very glad if you're interested in building the new Docker image (and maybe improving the build image) for restic. Thanks!

Issue #1422, opened 2017-11-16
IAM roles, supported in 0.7.3, appear to no longer work in master


TheWalkers:
## Output of `restic version`

v0.7.3-237-g06bd606d

## How did you run restic exactly?

/usr/local/bin/restic -r s3:s3.amazonaws.com/restic-test-repo -p /root/.restic_pass init
create backend at s3:s3.amazonaws.com/restic-test-repo failed: client.MakeBucket: Anonymous access is forbidden for this operation

Running 0.7.3 with the same command line options on the same instance:

/usr/local/bin/restic_0.7.3_linux_amd64 -r s3:s3.amazonaws.com/restic-test-repo -p /root/.restic_pass init
created restic backend ffc0ca4517 at s3:s3.amazonaws.com/restic-test-repo

Please note that knowledge of your password is required to access
the repository. Losing your password means that your data is
irrecoverably lost.

## What backend/server/service did you use?

AWS S3

## Expected behavior

restic should use IAM roles if no credentials are provided.

## Do you have an idea how to solve the issue?

I will be reviewing commits after 0.7.3 to see if I can spot a likely culprit.

TheWalkers:
git bisect shows the breaking commit @ 042adeb5d0e4cc1771da05a56e8b868292d7c037.

TheWalkers:
The problem appears to be that the credential chaining is not working as expected. 

                &credentials.Static{
                        Value: credentials.Value{
                                AccessKeyID:     cfg.KeyID,
                                SecretAccessKey: cfg.Secret,
                        },
                },

causes minio to try to access S3 with an empty AccessKeyID and SecretAccessKey, if those aren't defined.  It never falls back to IAM.

Happy to submit a PR that only includes the static credentials in the chain if they exist, but not sure if that's the correct approach.  I welcome thoughts from folks who understand minio-go credential chaining better on whether that's appropriate.

fd0:
Thanks for the report. @harshavardhana would you mind having a quick look? If I remember correctly, IAM profiles were briefly removed from minio-go, is that correct?

harshavardhana:
> Thanks for the report. @harshavardhana would you mind having a quick look? If I remember correctly, IAM profiles were briefly removed from minio-go, is that correct?

No @fd0 restic has IAM profiles supported , looks like a bug in chaining as mentioned by @TheWalkers

harshavardhana:
[...]

PR #1423, opened 2017-11-16
Fix chaining of credentials for minio-go

harshavardhana:
chaining failed because chaining provider
was only looking for subsequent credentials
provider after an error. Writer a new
chaining provider which proceeds to fetch
new credentials also under situations where
providers do not return but instead return
no keys at all.

https://github.com/restic/restic/issues/1422

### DIFF
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -33,6 +33,48 @@ var _ restic.Backend = &Backend{}
 
 const defaultLayout = "default"
 
+type chain struct {
+	Providers []credentials.Provider
+	curr      credentials.Provider
+}
+
+// FIXME: Remove this code once restic migrates to minio-go 4.0.x
+func newChainCredentials(providers []credentials.Provider) *credentials.Credentials {
+	return credentials.New(&chain{
+		Providers: append([]credentials.Provider{}, providers...),
+	})
+}
+
+// Retrieve returns the credentials value or error if no provider returned
+// without error.
+//
+// If a provider is found it will be cached and any calls to IsExpired()
+// will return the expired state of the cached provider.
+func (c *chain) Retrieve() (credentials.Value, error) {
+	for _, p := range c.Providers {
+		creds, _ := p.Retrieve()
+		// If anonymous proceed to the next provider if any.
+		if creds.SignerType.IsAnonymous() {
+			continue
+		}
+		c.curr = p
+		return creds, nil
+	}
+	return credentials.Value{
+		SignerType: credentials.SignatureAnonymous,
+	}, nil
+}
+
+// IsExpired will returned the expired state of the currently cached provider
+// if there is one.  If there is no current provider, true will be returned.
+func (c *chain) IsExpired() bool {
+	if c.curr != nil {
+		return c.curr.IsExpired()
+	}
+
+	return true
+}
+
 func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 	debug.Log("open, config %#v", cfg)
 
@@ -47,7 +89,7 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 	// configured ec2 instances)
 	// AWS env variables such as AWS_ACCESS_KEY_ID
 	// Minio env variables such as MINIO_ACCESS_KEY
-	creds := credentials.NewChainCredentials([]credentials.Provider{
+	creds := newChainCredentials([]credentials.Provider{
 		&credentials.Static{
 			Value: credentials.Value{
 				AccessKeyID:     cfg.KeyID,


ay:
+1 on the issue, but -1 on this fix. This looks like a bug in minio-go, not restic.

I just ran into this when trying to add a new credential provider to the credential chain. It seems like credential provider chaining is broken in minio-go because some credential providers (namely `credentials.EnvAWS`, `credentials.EnvMinio`, and `credentials.Static`) [don't](https://github.com/minio/minio-go/blob/9a9f8cb9f5946200b7f0b223d97725b30876bb00/pkg/credentials/env_aws.go#L40-L66) [return](https://github.com/minio/minio-go/blob/9a9f8cb9f5946200b7f0b223d97725b30876bb00/pkg/credentials/env_minio.go#L39-L57) [errors](https://github.com/minio/minio-go/blob/9a9f8cb9f5946200b7f0b223d97725b30876bb00/pkg/credentials/static.go#L55-L58) when key ID and secret are empty. This in turn causes the chaining implemenation to always stop looking for credentials after it encounters one of those providers in the chain, even if it doesn't actually provide any credentials.

FWIW, aws-sdk-go, which minio-go seems to be based on, implements this correctly by [always](https://github.com/aws/aws-sdk-go/blob/14a64cf1ee5fbc5ff6ebaffac89732b492f2697b/aws/credentials/env_provider.go#L58-L64) [returning](https://github.com/aws/aws-sdk-go/blob/14a64cf1ee5fbc5ff6ebaffac89732b492f2697b/aws/credentials/static_provider.go#L42-L44) an error if the key ID or secret parsed from environment variables or passed in as static credentials are empty.

A fix for this in minio-go might involve doing what aws-sdk-go does and have the environment variable credential providers return errors instead of empty strings when the provided credentials are empty (or the environment variables aren't set). Another way is to do what's in this pull request and change the chaining implementation to keep looking for credentials if it encounters non-error empty-string credentials.

Until this gets fixed in minio-go, instead of copying types to restic that already exist in minio-go, there should just be a workaround: build the chain as usual (with the IAM credential provider for example) but exclude any `credentials.EnvAWS`, `credentials.EnvMinio`, and `credentials.Static` providers, and inject a `credentials.Static` to the beginning of the chain *only if* `cfg.KeyID` and `cfg.Secret` are non-empty strings. Once it gets fixed, just delete the injection workaround and add `credentials.EnvAWS` and `credentials.EnvMinio` to the beginning of the chain.

harshavardhana:
> Until this gets fixed in minio-go, instead of copying types to restic that already exist in minio-go, there should just be a workaround: build the chain as usual (with the IAM credential provider for example) but exclude any credentials.EnvAWS, credentials.EnvMinio, and credentials.Static providers, and inject a credentials.Static to the beginning of the chain only if cfg.KeyID and cfg.Secret are non-empty strings. Once it gets fixed, just delete the injection workaround and add credentials.EnvAWS and credentials.EnvMinio to the beginning of the chain.

The reason we do not return error is because returning error for anonymous credentials is a wrong thing to do, we just treat these things alike.  The fix here is just a different implementation of the same chaining provider which implements functionality by prioritizing non anonymous credentials. 

I do not agree that we should return errors, aws-sdk-go has different requirements than minio-go.

ay:
> I do not agree that we should return errors, aws-sdk-go has different requirements than minio-go.

Makes sense. It looks like minio/minio-go#877 implements your fix here to minio-go itself, right?

I still think there's no need to copy that type to restic and a workaround should work fine until that pull request gets merged into minio-go and updated here. There should at least be a comment to remove the extra chain type/workaround after minio-go gets patched.

harshavardhana:
> Makes sense. It looks like minio/minio-go#877 implements your fix here to minio-go itself, right?
> 
> I still think there's no need to copy that type to restic and a workaround should work fine until that pull request gets merged into minio-go and updated here. There should at least be a comment to remove the extra chain type/workaround after minio-go gets patched.

This is an interim workaround for now, once restic moves to perhaps 4.0.x release it can be migrated. Thanks good idea, will add a FIXME perhaps to indicate that we need to remove it.

fd0:
Thanks for the contributions!

fd0:
@harshavardhana I've updated restic to minio-go v4, would you mind having a look what can be removed from the s3 backend here: https://github.com/restic/restic/blob/4f46b4f3935286221f05000575bd5e97fd1a1151/internal/backend/s3/s3.go#L85-L107

harshavardhana:
@fd0  We can basically remove 

```diff
diff --git a/internal/backend/s3/s3.go b/internal/backend/s3/s3.go
index f15a2ddc..d7119648 100644
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -33,48 +33,6 @@ var _ restic.Backend = &Backend{}
 
 const defaultLayout = "default"
 
-type chain struct {
-       Providers []credentials.Provider
-       curr      credentials.Provider
-}
-
-// FIXME: Remove this code once restic migrates to minio-go 4.0.x
-func newChainCredentials(providers []credentials.Provider) *credentials.Credentials {
-       return credentials.New(&chain{
-               Providers: append([]credentials.Provider{}, providers...),
-       })
-}
-
-// Retrieve returns the credentials value or error if no provider returned
-// without error.
-//
-// If a provider is found it will be cached and any calls to IsExpired()
-// will return the expired state of the cached provider.
-func (c *chain) Retrieve() (credentials.Value, error) {
-       for _, p := range c.Providers {
-               creds, _ := p.Retrieve()
-               // If anonymous proceed to the next provider if any.
-               if creds.SignerType.IsAnonymous() {
-                       continue
-               }
-               c.curr = p
-               return creds, nil
-       }
-       return credentials.Value{
-               SignerType: credentials.SignatureAnonymous,
-       }, nil
-}
-
-// IsExpired will returned the expired state of the currently cached provider
-// if there is one.  If there is no current provider, true will be returned.
-func (c *chain) IsExpired() bool {
-       if c.curr != nil {
-               return c.curr.IsExpired()
-       }
-
-       return true
-}
-
 func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
        debug.Log("open, config %#v", cfg)
 
@@ -89,7 +47,7 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
        // configured ec2 instances)
        // AWS env variables such as AWS_ACCESS_KEY_ID
        // Minio env variables such as MINIO_ACCESS_KEY
-       creds := newChainCredentials([]credentials.Provider{
+       creds := credentials.NewChainCredentials([]credentials.Provider{
                &credentials.Static{
                        Value: credentials.Value{
                                AccessKeyID:     cfg.KeyID,
```

fd0:
Would you like to do a PR? Or should I do that?

harshavardhana:
> Would you like to do a PR? Or should I do that?

i can do ..

fd0:
Please, do, thanks!

harshavardhana:
> Please, do, thanks!

done

PR #1428, opened 2017-11-18
use default brew formula

stephengroat:
restic has been added as a `homebrew-core` formula, the https://github.com/restic/homebrew-restic repo can also be deleted if necessary


fd0:
Hey, thanks for letting us know! What do we need to do when a new release is published? How do we update the default formula?

PR #1429, opened 2017-11-18
add homebrew test

stephengroat:


fd0:
I'm not sure this is a good idea. It would test not restic itself, but the installation via homebrew. That'd add many different external dependencies, at a minimum the homebrew servers etc.

PR #1462, opened 2017-12-03
Add diff command

### LINKED ISSUE #11: Compare snapshots
Implement feature to compare two or more snapshots, show differences of files + size
- Show added files/directories
- Show deleted files/directories
- Show modified files/directories 

Use case: find a specific version of a file, compare directory structure of multiple snapshots, etc.

### LINKED ISSUE #1460: Feature request: snapshots diff
It would be very useful to be able to list the metadata (not files content) differences between two snapshots, to better understand e.g. what needs to be excluded from backups.

(This is a feature of rsnapshot.)

fd0:

The commits in this PR adds the `diff` command that compares two snapshots and prints a line for each difference, similar to what the Unix `diff` utility does.

https://github.com/restic/restic/issues/11
https://github.com/restic/restic/issues/1460

### Checklist

- [x] There's an entry in the `CHANGELOG.md` file that describe the changes for our users

Closes #11
Closes #1460

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ released version of restic from the perspective of the user.
 Important Changes in 0.X.Y
 ==========================
 
+ * The new command `diff` was added, it allows comparing two snapshots and
+   listing all differences.
+   https://github.com/restic/restic/issues/11
+   https://github.com/restic/restic/issues/1460
+   https://github.com/restic/restic/pull/1462
+
 
 Small changes
 -------------
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -1,0 +1,340 @@
+package main
+
+import (
+	"context"
+	"path"
+	"reflect"
+	"sort"
+
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/repository"
+	"github.com/restic/restic/internal/restic"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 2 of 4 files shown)


fd0:
Sample output:

```
$ ./restic diff ea 2a
password is correct
comparing snapshot ea657ce5 to 2ab627a6:
 C   /restic/cmd_diff.go
+    /restic/foo
 C   /restic/restic
Files:           0 new,     0 removed,     2 changed
Dirs:            1 new,     0 removed
Others:          0 new,     0 removed
Data Blobs:     14 new,    15 removed
Tree Blobs:      2 new,     1 removed
  Added:   16.403 MiB
  Removed: 16.402 MiB
```

lingej:
Hi Alexander

tested the diff command with version v0.8.0-44-g06cb3f70 amd64

a new directory is shown with flag "+" but the new files inside are not recognized 

Kind regards 
Joerg

fd0:
Yes, I thought it is sufficient to show that the directory has been added. Do you think it's valuable to traverse into the newly added dirs?

lingej:
in my opinion? yes it is ;)

the new files in B are not part of A so they should be listed by diff

IMHO it is very useful to get a list of changed/new files after each backup run.

fd0:
Ah, hm, you're right, and I think there's even a bug in the `diff` command: Files in added/removed subdirs aren't included in the stats. I'll have a look later today.

fd0:
You were right, there was indeed a bug. So diff now descends into added/removed dirs and prints them, the PR is here: https://github.com/restic/restic/pull/1471

Issue #361, opened 2015-11-26
Mount in windows


tigerclawn82:
Hi,
   As instructed in the User Manual, i have snapshot created and backup repository available. When i tried to mount the repository it gives following error.

```
D:\Workspace\Go-Ide\GOPATH\src\restic>restic.exe version
restic 0.1.0 (v0.1.0-148-g2c95772)
compiled at 2015-11-25 18:53:15 with go1.3.1

D:\Workspace\Go-Ide\GOPATH\src\restic>restic.exe -r D:\Restic-Backups mount D:\Restic-Mount
Unknown command `mount'. Please specify one command of: backup, cache, cat, check, find, init, key, list, ls, optimize, rebuild-index, restore, snapshots, unlock or version
```

Any help would be appreciated. 

Thanks

fd0:
Unfortunately it is not possible to mount the repository on Windows. I'll add that to the manual.

tseel:
Dokan (https://dokan-dev.github.io) provides a fuse library on windows which can be used.

Matroschker:
I tried to use Dokan 1.0.0 on Win 10 Pro for mounting, currently I run into an [error](https://github.com/dokan-dev/dokany/issues/177).

fd0:
I don't use windows and I have no idea what's happening here, sorry.

Matroschker:
Ok, but can you explain what to mount exactly with dokan? Restic creates a directory structure, should be mounted the root folder or maybe the data folder?

fd0:
As  far as I can see, dokan is a library for windows which implements something similar to fuse, but is not directly compatible.

For restic this means that someone needs to add the library to restic and write code to handle the dokan-specific file-system abstraction. I'm not using Windows, so someone else needs to step up to implement this.

Until dokan support is added to windows, this won't work.

Liryna:
Hi @fd0 

It is true that dokan have is own file-system abstraction but also provide a FUSE wrapper.
Does restic use FUSE ? If it is the case, restic will have to be build with cygwin against the dokanfuse.dll and it should work by is own.

As exemple, there is a FUSE Mirror FS that is include in the repository that come directly from the libfuse repository that worked with dokanfuse.dll without changes.

fd0:
If I understood the documentation correctly (sorry, only had a quick glance today), dokan is "API compatible" to fuse. Which means that when a project uses the original fuse API provided by libfuse, you can probably just compile it against dokan instead of fuse and it works.

[...]

PR #1474, opened 2017-12-08
supports mount on windows

comicfans:
these patches support mount on windows through winfsp

#361 have some discuss on dokan , but I prefer [winfsp](http://www.secfs.net/winfsp/) instead, reasons:

* its official go bind [cgofuse](https://godoc.org/github.com/billziss-gh/cgofuse/fuse) have similar interface to bazil fuse,  and it also targets linux/osx,  makes port easily
* it's used by [rclone](https://rclone.org/commands/rclone_mount/) mount on windows
* winfsp's author takes [test](http://www.secfs.net/winfsp/develop/tests/) heavily
* winfsp suggests better [performance](http://www.secfs.net/winfsp/develop/perf-tests/) 
* in my test environment (with antivirus and data leak prevention software), dokan based encfs4win cause BSOD , but winfsp works without problem (with it's memfs example). and winfsp never crash during my restic development (even with my incorrect code)

this implementation is not a "native" port , it's an adapter layer which reuses current implementation .for that  I've moved minimal set codes to separate file in bazil fuse to make it built on windows.   it should not change behavior for other OS , and we can evaluate this on windows. 

### DIFF
--- a/cmd/restic/cmd_cgofuse.go
+++ b/cmd/restic/cmd_cgofuse.go
@@ -1,0 +1,140 @@
+// +build windows
+
+package main
+
+import (
+	"context"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	cgofuse "github.com/billziss-gh/cgofuse/fuse"
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+	resticfs "github.com/restic/restic/internal/fs"
+	"github.com/restic/restic/internal/fuse"
+	"github.com/restic/restic/internal/restic"
+)
+
+var cmdMount = &cobra.Command{
+	Use:   "mount [flags] mountpoint",
+	Short: "Mount the repository",
+	Long: `
+The "mount" command mounts the repository via fuse to a directory. This is a
+read-only mount.
+`,
+	DisableAutoGenTag: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runMount(mountOptions, globalOptions, args)
+	},
+}
+
+var host *cgofuse.FileSystemHost
+
+// MountOptions collects all options for the mount command.
+type MountOptions struct {
+	OwnerRoot  bool
+	AllowRoot  bool
+	AllowOther bool
+	Host       string
+	Tags       restic.TagLists
+	Paths      []string
+}
+
+var mountOptions MountOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdMount)
+
+	mountFlags := cmdMount.Flags()
+	mountFlags.BoolVar(&mountOptions.OwnerRoot, "owner-root", false, "use 'root' as the owner of files and dirs")
+	mountFlags.BoolVar(&mountOptions.AllowRoot, "allow-root", false, "allow root user to access the data in the mounted directory")
+	mountFlags.BoolVar(&mountOptions.AllowOther, "allow-other", false, "allow other users to access the data in the mounted directory")
+
+	mountFlags.StringVarP(&mountOptions.Host, "host", "H", "", `only consider snapshots for this host`)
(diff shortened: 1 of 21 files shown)


fd0:
Thanks for giving this a try. I must say I'm very skeptical, I had several bad experiences with cgo and (as already mentioned in https://github.com/restic/restic/issues/361#issuecomment-192171157 I'd like to keep restic Go-only.

comicfans:
I haven't tried webdav mount on windows so I'm not sure it works as well as local filesystem (file number/size limit for example),  benifits of local filesystem is that it can be shared through samba with windows authorization (which is useful for us), I don't know if webdav can provide similar functional. and  what's the plan of #60 ? it have similar functional to webdav,  maybe we should consider most user scenario and decide which one to implement  ?

fd0:
So, what do we do with this issue? Is there a way to realize fuse on Windows using Go-only code, without `cgo`?

Do you have some more background information for me on the limitations of the Windows webdav client? As far as I understood the issue you linked, it's only an issue on Windows 7? Since Windows 7 isn't supported any more since the beginning of January 2015, I don't think that's a huge issue. Which Windows version are you running?

fd0:
Ah, the webdav limitations are not an issue, there's an webdav client implementation for winfsp that you could use (instead of the internal Windows webdav client): https://github.com/FrKaram/KS2.Drive

rawtaz:
I realize it means third party software, but Cyberduck is a free and open source file transfer client with great support for many backens, including WebDAV: https://github.com/iterate-ch/cyberduck and https://cyberduck.io

They also have Mountain Duck which I am not sure if it's open source however. But my point is that there *is* third party software that will let the user use WebDAV if that is provided by restic.

comicfans:
winfsp's cgofuse needs cgo to build , that means restic also needs cgo to build (on windows only). currently we're running windows 7 (even its not supported any more ).  I've found some webdav limitation but didn't confirm that by myself. [here](https://github.com/cryptomator/cryptomator/issues/357) and [here from microsoft, file cannot larger than 4GB](https://social.technet.microsoft.com/Forums/en-US/7cc8aaf5-3b68-47dc-8dc8-0f51d955f189/i-need-to-remove-the-filesizelimitsinbytes-registry-key?forum=w8itpronetworking).

and I agreed with @rawtaz 's point.  If third party software still needed to access restic, it didn't help too much for windows users. I think most linux/osx user will still prefer fuse instead of webdav .

fd0:
> winfsp's cgofuse needs cgo to build , that means restic also needs cgo to build (on windows only)

Yeah, that's not an option. It'd mean reworking our build process completely, adding new dependencies (even a cross-compilation environment), that's not something we (as a spare-time Free Software project) can do.

To clarify this: I proposed adding a webdav server **in addition** to the fuse mount, so on non-Windows you have fuse+webdav, on Windows you only have webdav. Users can then either use the default Windows webdav client (for small files), or the Duck software thing @rawtaz mention, or the webdav client implementation based on winfsp. I think having users on Windows install an additional software package to use this is not an issue, it would be the same if we were to merge the winfsp-based fuse mount, since that requires having winfsp installed.

So, all an all, thank you very much for giving this a try, but I won't merge this PR. The way forward here in my opinion is to implement a webdav server which works on all platforms. Please let us know if you'd maybe like to try it. Thanks!

comicfans:
That's OK, but I think a web interface is more useful than webdav, for example btsync/syncthing/transmission

fd0:
@comicfans that's a somewhat related, but different issue.

rawtaz:
I think @comicfans has a good point - instead of WebDAV, perhaps that coding time is better spend on a web server in restic. But then again, it will not yield the same features in terms of how you can interact with it, so on second thought I think neither of these two features are more important than the other one. Never mind :)

PR #1478, opened 2017-12-12
Skip subdirs on local backends for keys

ckeen:
In cases where there are subdirectories present in the keys/ dir the
path returned is stripped off the component.

Keys should reside only in the top level of keys/ so skip the
subdirectories for now.

The use case causing this patch is a local storage mounted via cifs on
some webdav server that litters .DAV subdirectories everywhere.

### DIFF
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -257,6 +257,12 @@ func (b *Local) List(ctx context.Context, t restic.FileType) <-chan string {
 				return err
 			}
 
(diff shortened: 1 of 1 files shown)


ckeen:
Updated the patch, passes tests on linux now...

fd0:
The issue is deeper, I'm on it. Thanks for giving it a try!

fd0:
Superseded by #1482

Issue #1477, opened 2017-12-10
S3 fails on Buckets with permissions scoped to prefixes


advanderveer:
Thank you for this great piece of software! I've run into an issue on S3 Buckets with tightly controlled permissions.

## Output of `restic version`
```
restic 0.8.0
compiled with go1.9 on darwin/amd64
```

## How did you run restic exactly?
env (I cleared all other variables):
```
AWS_ACCESS_KEY_ID=<key> 
AWS_SECRET_ACCESS_KEY=<secret>
AWS_SESSION_TOKEN=<token>
```

```
restic init --password-file=<my-password-file> -r=s3:s3.amazonaws.com/<my-bucket>/<my-prefix>
```

```
create backend at s3:s3.amazonaws.com/<my-bucket>/chunks/<my-user>/hhh failed: client.BucketExists: Access Denied
```

## What backend/server/service did you use to store the repository?
hosted s3

## Expected behavior
I created the bucket by hand and tightly controlled the permissions for clients. The s3:HeadBucket operations ([client.BucketExists](https://github.com/minio/minio-go/blob/3419703769e0de7e06fab876ce54605df621f017/api-stat.go)) will return permission denied in these cases even though all operations necessary for backup/restore are allowed. 

I would expect restic to allow to disable this behaviour through an option and trust I setup the bucket correctly myself.  

## Actual behavior
It checks for existence in order to try and [create a bucket](https://github.com/restic/restic/blob/1ebf0e8de8f406f11c3d35b8479810433a1c0b71/internal/backend/s3/s3.go#L143), both calls will fail in tightly controlled s3 environments.

## Steps to reproduce the behavior
A user with the following permissions would cause the issue:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:HeadBucket"
            ],
            "Resource": [
                "arn:aws:s3:::<buket-name>"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::<bucket-name>"
            ],
            "Condition": {
                "StringLike": {
                    "s3:prefix": [
                        "chunks/my-user"
                    ]
                }
            }
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetObject",
                "s3:PutObject",
                "s3:DeleteObject"
            ],
            "Resource": [
[...]

PR #1479, opened 2017-12-12
Allow AWS S3 backend to use temporary and path-scoped credentials

advdv:
It is currently not possible to use the restic S3 backend with IAM credentials that are temporary and scoped to a subdirectory. The first one requires restic to accept AWS_SESSION_TOKEN credentials, the second requires that the s3 backend doesn't try to create whenever the 'bucket exists' check fails due to permission issues. 

This pull request aims to solves the issue as described here #1477 

### Checklist

- [ ] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -428,6 +428,10 @@ func parseConfig(loc location.Location, opts options.Options) (interface{}, erro
 			cfg.Secret = os.Getenv("AWS_SECRET_ACCESS_KEY")
 		}
 
+		if cfg.Token == "" {
+			cfg.Token = os.Getenv("AWS_SESSION_TOKEN")
+		}
+
 		if err := opts.Apply(loc.Scheme, &cfg); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


fd0 (changes requested):
Thanks for giving this a try. We can certainly handle the situation better.

I've read throught the code and it seems unfinished (you also did not check the "I'm ready" box).

A few thoughts:
 * The code for using a session token for authentication is unrelated, it should be added in a separate PR
 * I'd like to not introduced another option. Instead, my idea is to inspect the error that is returned by `BucketExists()`, and return silently if it is `PermissionDenied`, so that restic skips trying to create the bucket. If the bucket does not exist, it'll fail later on. What's your take on this?
 * I think we don't need a separate test case for the unprivileged account (`TestUnprivilegedS3Credentials`), the tests are mainly used during CI (by Travis), and I don't have any way to setup a token and an unprivileged account.
 * For this PR to be merged, please format the patches according to the contribution guidelines. In this case, prefix the terse summary line with `s3:` and use present tense.

As always, please let us know if you need any help. Thanks!

fd0:
There hasn't been any activity here for the last few weeks so I'm closing this PR. Just leave a comment if you're still interested in continuing the work (or need help), thanks!

Issue #1473, opened 2017-12-07
Cache directory and the root user


dsommers:
## Output of `restic version`
restic 0.8.0 (v0.8.0-0-gc02923fb)
compiled with go1.8.3 on linux/amd64

## Issue
When running restic as the root user to take a system backup, the default cache directory is /root/.cache/restic. I propose to use /var/cache/restic as the cache directory when restic is run as root.

## Reason
/root is often part of the root file system.  This file system might not be as big, particularly on servers where more advanced partitioning schemes are used.  Those places /var is more commonly more spacious and some might even have split /var over multiple partitions as well to avoid system havoc if some lesser critical partitions goes empty; like /var/cache.  Filling up / can easily cause issues - and with restic repositories with thousands of files with lots of changes, the cache directory can easily grow quite a bit.

zcalusic:
Thinking a bit about your proposition, it makes a lot of sense. /var/cache is standard FHS location for various system caches. Thanks for opening this issue.

Not to mention that I've already symlinked /root/.cache/restic to another bigger partition with more space on ALL of my restic client hosts. ;)

But, even if it's easy to symlink, sane defaults are always welcome. Probably restic could also emit a line saying Using cache: /var/cache/restic (or /home/user/.cache/restic for ordinary users), so there's no confusion. And user always knows where the cache ended.

zcalusic:
As always, migration is hard, @fd0, you could probably put in code that works for both locations, but for newly created repos, /var/cache would be chosen. So, restic client would first check the current /root/.cache/restic, and use that if found. Maybe another info line (not warning 😄) that would suggest user to move cache to the new default /var/cache/restic.

fd0:
To be honest, I'd like to leave it as is for now. Adding this requires more code and even more special cases, even checking two different cache dirs or a migration. If needed, people can easily set `XDG_CACHE_HOME` to `/var/cache` and be done with it.

dsommers:
That approach is fine when only considering non-admin users running backups.  But once starting to use this for system wide backups, where the backup tools needs to be run as root to have access to all files being backed up, this will be a growing issue.

dstosberg:
Just a note on the side: If you are on Linux, you don't have to run `restic` as root to make a full system backup. Yes, restic needs to be able to read all files, but it does not need any of the other privileges that root has. On Linux you can assign that specific single capability to the restic binary and run it as an ordinary user:

```bash
# create restic user
root@a3e580b6369d:/# useradd -m restic

[...]

PR #1483, opened 2017-12-14
document how to create a full backup without running restic as root

dstosberg:

Add an example section to the documentation that explains how one can use restic to create a full backup without having to run it as root.

The idea occurred in discussing issue #1473.

### Checklist

### DIFF
--- a/doc/080_examples.rst
+++ b/doc/080_examples.rst
@@ -270,3 +270,81 @@ restored:
 
 The snapshot was successfully restored. This concludes the tutorial.
 
+
+*****************************************************
+Backing up your system without running restic as root
+*****************************************************
+
+Motivation
+==========
+
+Creating a complete backup of a machine requires a privileged process
+that is able to read all files. On UNIX-like systems this is
+traditionally the ``root`` user. Processes running as root have
+superpower. They cannot only read all files but do also have the power
+to modify the system in any possible way.
+
+With great power comes great responsibility. If a process running as
+root malfunctions, is exploited, or simply configured in a wrong way it
+can cause any possible damage to the system. This means you only want
+to run programs as root that you trust completely. And even if you
+trust a program, it is good and common practice to run it with the
+least possible privileges.
+
+Capabilities on Linux
+=====================
+
+Fortunately, Linux has functionality to divide root's power into
+single separate *capabilities*. You can remove these from a process
+running as root to restrict it. And you can add capabilities to a
+process running as a normal user, which is what we are going to do.
+
+Full backup without root
+========================
+
+To be able to completely backup a system, restic has to read all the
+files. Luckily Linux knows a capability that allows precisely this. We
+can assign this single capability to restic and then run it as an
+unprivileged user.
+
+First we create a new user called ``restic`` that is going to create
+the backups:
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# useradd -m restic
+
+Then we download and install the restic binary into the user's home
+directory.
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# mkdir ~restic/bin
+   root@a3e580b6369d:/# curl -L https://github.com/restic/restic/releases/download/v0.8.0/restic_0.8.0_linux_amd64.bz2 | bunzip2 > ~restic/bin/restic
+
+Before we assign any special capability to the restic binary we
+restrict its permissions so that only root and the newly created
+restic user can execute it. Otherwise another - possibly untrusted -
+user could misuse the privileged restic binary to circumvent file
+access controls.
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# chown root:restic ~restic/bin/restic
+   root@a3e580b6369d:/# chmod 750 ~restic/bin/restic
+
+Finally we can use ``setcap`` to add an extended attribute to the
+restic binary. On every execution the system will read the extended
+attribute, interpret it and assign capabilities accordingly.
+
+.. code-block:: console
+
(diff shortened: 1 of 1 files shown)


doc/080_examples.rst
```diff
@@ -270,3 +270,82 @@ restored:
...
+files. Luckily Linux knows a capability that allows precisely this. We
+can assign this single capability to restic and then run it as an
+unprivileged user.
+
+First we create a new user called ``restic`` that is going to create
+the backups:
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# useradd -m restic
```
fawick: If `sudo` is always used, would it make sense to restrict the login shell  (`-s /bin/false`), too?

doc/080_examples.rst
```diff
@@ -270,3 +270,82 @@ restored:
...
+.. code-block:: console
+
+   root@a3e580b6369d:/# setcap cap_dac_read_search=+ep ~restic/bin/restic
+
+From now on the user ``restic`` can run restic to backup the whole
+system.
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# sudo -u restic -H -s
```
fawick: There may be a good reason for it which I cannot see, but why two commands instead of, e.g. `sudo -u restic ~restic/bin/restic -r /tmp --exclude /tmp backup /`

Likewise, what is the rationale `-x`? I'd assume  backing up the whole system may  require to access multiple file systems, may it not? In case you really want to have `-x` it needs to be after `backup` as it is not a global flag.

Should `--exclude /tmp` as the repository be added here? Or how about adding `--exclude={"/dev/*","/proc/*","/sys/*","/tmp/*","/run/*","/lost+found"}`

fawick (commented):
This PR is a great idea and very helpful, thanks for contributing to the documentation! I have some questions below.

doc/080_examples.rst
```diff
@@ -270,3 +270,82 @@ restored:
...
+   root@a3e580b6369d:/# chown root:restic ~restic/bin/restic
+   root@a3e580b6369d:/# chmod 750 ~restic/bin/restic
+
+Finally we can use ``setcap`` to add an extended attribute to the
+restic binary. On every execution the system will read the extended
+attribute, interpret it and assign capabilities accordingly.
+
+.. code-block:: console
+
+   root@a3e580b6369d:/# setcap cap_dac_read_search=+ep ~restic/bin/restic
```
eliasp: Although I agree, that restic ideally shouldn't be run as root, assigning `cap_dac_read_search` to the binary directly is dangerous as this would allow now anyone able to execute restic to also ready any system file.

I'd rather handle this by letting systemd control all those aspects.

My WIP service instance unit for restic: `/etc/systemd/system/restic@.service`
```INI
[Unit]
Description=Backup %I using 'restic'
ConditionFileNotEmpty=/etc/restic/%i/restic.env
ConditionPathExists=/etc/restic/%i/backup.exclude

[Service]
ExecStart=/usr/bin/restic backup --cache-dir=/var/cache/restic/%i --one-file-system --exclude-file=/etc/restic/%i/backup.exclude --exclude=/var/cache/* --exclude=/tmp/* --exclude=/var/tmp/* --exclude=/home/*/.cache/* --exclude=/run/* --quiet /
Type=oneshot
EnvironmentFile=/etc/restic/%i/restic.env
DynamicUser=yes
PrivateDevices=yes
CacheDirectory=restic/%i
CacheDirectoryMode=0700
ConfigurationDirectory=restic/%i
ConfigurationDirectoryMode=0700
#SystemCallFilter=@basic-io @file-system @network-io @io-event @ipc @cpu-emulation
#SystemCallErrorNumber=EPERM
AmbientCapabilities=CAP_DAC_READ_SEARCH

# TODO
# fix SystemCallFilter= usage, which currently fails, although the required tls syscall should be included in @default:
#   -- cannot set up thread-local storage: cannot set %fs base address for thread-local storage
# - does restic support sd_notify()? if not, create an RFC or PR
# - ensure systemd's ProtectSystem= (+ related options) is utilized to the fullest
# - set low io class
```

fd0: > Although I agree, that restic ideally shouldn't be run as root, assigning cap_dac_read_search to the binary directly is dangerous as this would allow now anyone able to execute restic to also ready any system file.

Indeed, and that's why the permissions are set like this:

```
   root@a3e580b6369d:/# chown root:restic ~restic/bin/restic
￼   root@a3e580b6369d:/# chmod 750 ~restic/bin/restic
￼```

Did you see that?

eliasp: Missed that - sorry about the noise.

fd0: No worries :)

fd0:
ping @dstosberg, are you still interested in continuing the work here, or should someone else take over?

fd0 (changes requested):
Please address the issues in the comments, thanks!

doc/080_examples.rst
dstosberg: Yes, we could easily put `sudo` and the call to `restic` a single command.

The intention for the `-x` was to prevent a clueless reader from backing up his /proc or /sys which might never actually terminate. But you are right, a proper generic exclude list, that works for most people, is a better fit. And `-x` might exclude `/home`, which is quite commonly on a separate filesystem.

doc/080_examples.rst
dstosberg: It would not be a bad idea to do that. But personally, I find it very convenient that I can run a shell as user `restic` and use it examine or restore backups.

There are a lot of things that you could do to harden or optimize the setup. But I think that is out of the scope of this chapter.

fd0 (approved):
LGTM

PR #1511, opened 2017-12-29
Improve limiting HTTP based backends

### LINKED ISSUE #1506: Potential issue or different implementation with --limit-upload
When uploading data (at least to B2) using --limit-upload 90 (in kB), data flows are bursting to sometimes 1MiB/s or higher, with long delays typically between each large burst.

I suspect the --limit-upload was done by averaging upload data over time.  That would result in 90kB/s over time, but there is no rate limiting of the actual upload itself in real time.  Is this a correct understanding of how it was implemented, or was it supposed to work another way and therefore a bug?

There are other mechanisms by which I could see a more smooth data flow, but it'd require modifications to the firewall rules or a tool to shape restic network traffic.

## Restic version details
restic 0.8.0 (b9441a1)

## Output of `restic v0.80`
(incomplete -- lots of data to upload -- still in progress)
scan [/some/dir /some/dir]
[0:06] 17482 directories, 139992 files, 1.011 TiB
scanned 17482 directories, 139992 files in 0:06

(removed some specific directories -- not relevant)
/usr/bin/restic --cache-dir=/some/dir/restic-cache --limit-upload 90 backup --one-file-system /some/dir /some/dir --exclude /some/dir --exclude /some/dir

Backblaze B2

## Expected behavior
Smooth, consistent data flows according to --limit-upload.  Expected doesn't imply it is what was intended by the developers.

## Actual behavior
Bursting, with long delays between bursts

Just run the command with a lot of files :)

I suspect the implementation of the upload limit was done over time and does not actually limit rate.

Implement rate/throughput limit differently

This is a great tool for cloud backup.

fd0:
Limit the bandwidth on the HTTP request via the RoundTripper

Closes #1506

### DIFF
--- a/changelog/0.8.2/issue-1506
+++ b/changelog/0.8.2/issue-1506
@@ -1,0 +1,4 @@
+Bugfix: Limit bandwith at the http.RoundTripper for HTTP based backends
+
+https://github.com/restic/restic/issues/1506
+https://github.com/restic/restic/pull/1511
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -323,14 +323,9 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 		return nil, errors.Fatal("Please specify repository location (-r)")
 	}
 
-	be, err := open(opts.Repo, opts.extended)
+	be, err := open(opts.Repo, opts, opts.extended)
 	if err != nil {
 		return nil, err
-	}
-
-	if opts.LimitUploadKb > 0 || opts.LimitDownloadKb > 0 {
-		debug.Log("rate limiting backend to %d KiB/s upload and %d KiB/s download", opts.LimitUploadKb, opts.LimitDownloadKb)
-		be = limiter.LimitBackend(be, limiter.NewStaticLimiter(opts.LimitUploadKb, opts.LimitDownloadKb))
 	}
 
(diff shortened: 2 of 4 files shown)


eliasp:
@fd0 Seems to have no effect here. Trying to minimize the impact on my crappy 1Mbit/s upstream connection by using `restic backup --limit-upload=500 […]`, but it still fully saturates the upstream at ~1Mbit/s.

fd0:
did you rebuild the binary? What does `restic version` report?

eliasp:
Yes, rebuilt it, version looks good:
```
restic 0.8.1 (v0.8.1-11-ga99637c6)
```

eliasp:
Can I provide any kind of useful debug information here?

fd0:
Is the bandwidth limited at all? can you try with a small value, like 10?

fd0:
I cannot reproduce what you're seeing, I see a slight overshoot at the beginning, but then it drops back to the limited bandwidth very quickly.

fd0:
Ah! You said you have 1M**bit**, the value you need to pass restic is in K**byte** per second. So in order to not saturate your upstream bandwidth, you'll need to pass something less than "125"... Can you try that?

eliasp:
Damn! I feel embarrassed for not paying attention to this - it works - thanks a ton for this!

fd0:
Awesome, thanks for the feedback!

Issue #1522, opened 2018-01-02
HTTPS TLS Client Certificate Authentication


brycied00d:
I propose adding support for TLS client certificate authentication with HTTPS connections.

The change is relatively straight-forward as it only requires feeding the paths to a certificate/key pair to [`tls.LoadX509KeyPair`](https://golang.org/pkg/crypto/tls/#LoadX509KeyPair) and appending the result to [`tls.Config.Certificates`](https://golang.org/pkg/crypto/tls/#Config).

I've already written and tested a branch with this feature implemented.

Background: We're using rest-server behind nginx (terminating and validating TLS) to receive and store backups from LAN-connected computers. We wanted to expand this to external devices, however corporate policy requires the use of TLS and client certificates across untrusted networks.

I checked the forums and the GitHub issue/PR trackers but it seems nobody else has needed this capability. I'm creating this "issue" first so that I can reference it in the PR, 

## Did restic help you or made you happy in any way?

Yes! I don't think I would be suggesting features to improve it, much less to fit it into our corporate environment, if Restic didn't make me happy.

fd0:
Nice proposal, I'll have a look at the PR now.

bjoe2k4:
Looks like it is implemented in the meantime in #1524. Close?

fd0:
Oh, indeed, thanks for the reminder, closing!

PR #1524, opened 2018-01-02
Support for TLS client certificate authentication

brycied00d:

This adds `--tls-client-cert` and `--tls-client-key` parameters and enables use
of that certificate/key pair when connecting to https servers.

First, I must apologize if this isn't adhering to Go conventions. I don't write Go, I only pieced this together from the Restic source, [an example](https://gist.github.com/michaljemala/d6f4e01c4834bf47a9c4) I found Googling, and Go's [pkg/crypto/tls documentation](https://golang.org/pkg/crypto/tls/). I welcome any and all feedback.

Feature request: #1522

### Checklist

- [X] There's an entry in the `CHANGELOG.md` file that describe the changes for our users

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -48,7 +48,6 @@ type GlobalOptions struct {
 	NoCache       bool
 	CACerts       []string
 	TLSClientCert string
-	TLSClientKey  string
 	CleanupCache  bool
 
 	LimitUploadKb   int
@@ -86,8 +85,7 @@ func init() {
 	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache directory")
 	f.BoolVar(&globalOptions.NoCache, "no-cache", false, "do not use a local cache")
 	f.StringSliceVar(&globalOptions.CACerts, "cacert", nil, "path to load root certificates from (default: use system certificates)")
-	f.StringVar(&globalOptions.TLSClientCert, "tls-client-cert", "", "path to a TLS client certificate")
-	f.StringVar(&globalOptions.TLSClientKey, "tls-client-key", "", "path to a TLS client certificate key")
+	f.StringVar(&globalOptions.TLSClientCert, "tls-client-cert", "", "path to a file containing PEM encoded TLS client certificate and private key")
 	f.BoolVar(&globalOptions.CleanupCache, "cleanup-cache", false, "auto remove old cache directories")
 	f.IntVar(&globalOptions.LimitUploadKb, "limit-upload", 0, "limits uploads to a maximum rate in KiB/s. (default: unlimited)")
 	f.IntVar(&globalOptions.LimitDownloadKb, "limit-download", 0, "limits downloads to a maximum rate in KiB/s. (default: unlimited)")
@@ -545,7 +543,11 @@ func open(s string, gopts GlobalOptions, opts options.Options) (restic.Backend, 
 		return nil, err
 	}
 
-	rt, err := backend.Transport(globalOptions.CACerts, globalOptions.TLSClientCert, globalOptions.TLSClientKey)
+	tropts := backend.TransportOptions{
+		RootCertFilenames:        globalOptions.CACerts,
+		TLSClientCertKeyFilename: globalOptions.TLSClientCert,
+	}
+	rt, err := backend.Transport(tropts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Summary
  * Enh #1436: Add code to detect old cache directories
  * Enh #1439: Improve cancellation logic
  * Enh #11: Add the `diff` command
+ * Enh #1522: Add support for TLS client certificate authentication.
```
fd0: I'm sorry about that, but can you please remove the changes from `CHANGELOG.md`, it'll be generated only shortly before the release. We've built our own changelog generator (called [calens](https://github.com/restic/calens)) and now you need to create a file in a subdir of `changelog/`.

I've also updated the PR template.

CHANGELOG.md
```diff
@@ -73,6 +74,14 @@ Details
    https://github.com/restic/restic/issues/1460
    https://github.com/restic/restic/pull/1462
 
+ * Enhancement #1522: Add support for TLS client certificate authentication.
```
fd0: This needs to be in a separate file in `changelog/0.8.2/issue-1522`

CHANGELOG.md
```diff
@@ -73,6 +74,14 @@ Details
...
+
+   Support has been added to the http backend for providing a TLS client certificate/key
+   pair using `--tls-client-cert` and `--tls-client-key` when connecting to https resources
+   that perform TLS client certificate authentication.
+
+   https://github.com/restic/restic/issues/1522
```
fd0: Add the URL to the PR here (in the separate file)

cmd/restic/global.go
```diff
@@ -84,6 +86,8 @@ func init() {
 	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache directory")
 	f.BoolVar(&globalOptions.NoCache, "no-cache", false, "do not use a local cache")
 	f.StringSliceVar(&globalOptions.CACerts, "cacert", nil, "path to load root certificates from (default: use system certificates)")
+	f.StringVar(&globalOptions.TLSClientCert, "tls-client-cert", "", "path to a TLS client certificate")
```
fd0: Hm, can we have that in a single file? It looks like it's possible to use the `pem` decoding library to split a file into the key and the certificate(s). Let me know what you think and if you need help with that.

internal/backend/http_transport.go
```diff
@@ -29,6 +29,15 @@ func Transport(rootCertFilenames []string) (http.RoundTripper, error) {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
+		TLSClientConfig:       &tls.Config{},
+	}
+
+	if tlsClientCert != "" && tlsClientKey != "" {
```
fd0: We should add a check that complains when only one (key or cert) is set, but not the other, so users are notified that both options are needed. Or just use one file for both.

fd0 (changes requested):
Looks already very good so far, let me know what you think!

cmd/restic/global.go
brycied00d: I would even prefer a single file (cURL supports it, among others) however `tls.LoadX509KeyPair` definitely requires separate files, and reading a single file in and splitting it into certificate and key for use with `tls.X509KeyPair` is beyond my Go abilities, I'm sorry.

brycied00d:
Thanks for the feedback, @fd0 

When I was updating the documentation with the new switches, that seemed like something else that would be automated in a similar way to the changelog but I didn't see any documentation referencing that (just "I have added documentation for the changes (in the manual)" in the PR template). Did I overlook anything?

I still need to devise and implement tests (obviously), I just need to find the time to figure this out, and maybe the single file cert/key pair parsing too.

fd0:
No, you did not overlook anything, adding the switches to the manual is still a manual process (and I don't have any good idea on how to solve it). Let me know if you need any help or get stuck.

fd0:
Do you need help? The code is mostly done, and splitting a file into key and cert is easy to do...

brycied00d:
I appreciate the offer, @fd0 , I would love help! The learning curve and lack of time have meant this has been sitting on my back-burner.

fd0:
I've added some code, please have a look!

fd0:
And I also took the liberty of squashing your commits. I'd like to merge this as two commits (yours and mine) so your contribution is recorded :)

fd0 (approved):

internal/backend/http_transport.go
```diff
@@ -3,19 +3,66 @@ package backend
...
+	RootCertFilenames []string
+
+	// contains the name of a file containing the TLS client certificate and private key in PEM format
+	TLSClientCertKeyFilename string
+}
+
+// readPEMCertKey reads a file and returns the PEM encoded certificate and key
+// blocks.
+func readPEMCertKey(filename string) (certs []byte, key []byte, err error) {
+	data, err := ioutil.ReadFile(os.Args[1])
```
brycied00d: I just noticed this is reading `os.Args[1]` and __not__ `filename`. Oops! Opening issue + PR shortly.

fd0: Whoopsi, I didn't catch that ;) Looking forward to the PR

PR #1538, opened 2018-01-08
repository/index: Avoid creating an error when checking if an id is in the index.

MJDSys:

When backing up several million files (>14M tested here) with few changes,
a large amount of time is spent failing to find an id in an index and creating
an error to signify this.  Since this is checked using the Has method,
which doesn't use this error, this time creating the error is wasted.

Instead, create a new function lookupInternal that doesn't create this error.
Change Has to use this function and change Lookup to use lookupInternal and
create the error if it fails to find the id.

This results in saving about one hour of wall time (about half the backup
time) and about six hours of cpu time on my system.

I'm not sure if you want Lookup just changed to work with my lookupInternal, and then fixup all the reverse dependencies.  If this is wanted, I'll update my PR appropriately.  I also wasn't sure if you wanted a changelog entry, but I'm happy to add that too if wanted.

No.

### Checklist

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -169,12 +169,13 @@ func (idx *Index) ListPack(id restic.ID) (list []restic.PackedBlob) {
 
 // Has returns true iff the id is listed in the index.
 func (idx *Index) Has(id restic.ID, tpe restic.BlobType) bool {
-	_, err := idx.Lookup(id, tpe)
-	if err == nil {
-		return true
-	}
+	idx.m.Lock()
+	defer idx.m.Unlock()
 
-	return false
+	h := restic.BlobHandle{ID: id, Type: tpe}
+
+	_, ok := idx.pack[h]
+	return ok
 }
 
 // LookupSize returns the length of the plaintext content of the blob with the
--- a/internal/repository/index_test.go
+++ b/internal/repository/index_test.go
@@ -2,6 +2,7 @@ package repository_test
 
 import (
 	"bytes"
+	"math/rand"
 	"testing"
 
 	"github.com/restic/restic/internal/repository"
@@ -378,4 +379,58 @@ func TestIndexPacks(t *testing.T) {
 
 	idxPacks := idx.Packs()
 	rtest.Assert(t, packs.Equals(idxPacks), "packs in index do not match packs added to index")
+}
+
+const maxPackSize = 16 * 1024 * 1024
+
+func createRandomIndex() (idx *repository.Index, lookupID restic.ID) {
+	idx = repository.NewIndex()
+
+	// create index with 200k pack files
+	for i := 0; i < 200000; i++ {
+		packID := restic.NewRandomID()
+		offset := 0
+		for offset < maxPackSize {
+			size := 2000 + rand.Intn(4*1024*1024)
+			id := restic.NewRandomID()
+			idx.Store(restic.PackedBlob{
+				PackID: packID,
+				Blob: restic.Blob{
+					Type:   restic.DataBlob,
+					ID:     id,
+					Length: uint(size),
+					Offset: uint(offset),
+				},
+			})
+
+			offset += size
+
+			if rand.Float32() < 0.001 && lookupID.IsNull() {
+				lookupID = id
+			}
+		}
+	}
+
+	return idx, lookupID
+}
+
+func BenchmarkIndexHasUnknown(b *testing.B) {
+	idx, _ := createRandomIndex()
+	lookupID := restic.NewRandomID()
+
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		idx.Has(lookupID, restic.DataBlob)
+	}
+}
+
+func BenchmarkIndexHasKnown(b *testing.B) {
+	idx, lookupID := createRandomIndex()
+
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		idx.Has(lookupID, restic.DataBlob)
+	}
 }
--- a/internal/repository/index_test.go
+++ b/internal/repository/index_test.go
@@ -434,3 +434,52 @@ func BenchmarkIndexHasKnown(b *testing.B) {
 		idx.Has(lookupID, restic.DataBlob)
 	}
 }
+
+func TestIndexHas(t *testing.T) {
+	type testEntry struct {
+		id             restic.ID
+		pack           restic.ID
+		tpe            restic.BlobType
+		offset, length uint
+	}
+	tests := []testEntry{}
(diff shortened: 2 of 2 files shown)


fd0 (changes requested):
Hey, thanks for proposing the enhancemente! What I'd love to see is a benchmark showing the improvement... I can already imagine that creating the error takes CPU time (and memory), since we're using the `github.com/pkg/errors` library, which capture stack traces.

I'd like to propose a slightly different approach, which will use even less CPU time and memory: Undo your changes, then modify the `Has()` method to try the lookup in `idx.pack` directly (after locking of course). That will be even more efficient because bundling the locations where blobs are stored isn't needed.

What are your thoughts?

MJDSys:
Regarding the benchmark, I posted a note about the speed up I get running a backup of my system (saving about an hour of time in a 2-3 hour backup).  Is there a more established way to do the benchmark?  If it helps, I found this by doing some CPU profiling of my backup.

Regarding your suggestion, that does sound great.  If I understand correctly, I just have the one outer if statement?  That sounds like it could help some more (lookupInternal still dominates the cpu profile).  I'll give it a test here and see how it goes.

fd0:
Hey, for Go software it's very easy to write a micro benchmark of just the `Has()` function for the index:
```go
const maxPackSize = 16 * 1024 * 1024

func createRandomIndex() (idx *repository.Index, lookupID restic.ID) {
	idx = repository.NewIndex()

	// create index with 200k pack files
	for i := 0; i < 200000; i++ {
		packID := restic.NewRandomID()
		offset := 0
		for offset < maxPackSize {
			size := 2000 + rand.Intn(4*1024*1024)
			id := restic.NewRandomID()
			idx.Store(restic.PackedBlob{
				PackID: packID,
				Blob: restic.Blob{
					Type:   restic.DataBlob,
					ID:     id,
					Length: uint(size),
					Offset: uint(offset),
				},
			})

			offset += size

			if rand.Float32() < 0.001 && lookupID.IsNull() {
				lookupID = id
			}
		}
	}

	return idx, lookupID
}

func BenchmarkIndexHasUnknown(b *testing.B) {
	idx, _ := createRandomIndex()
	lookupID := restic.NewRandomID()

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		idx.Has(lookupID, restic.DataBlob)
	}
}

func BenchmarkIndexHasKnown(b *testing.B) {
	idx, lookupID := createRandomIndex()

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		idx.Has(lookupID, restic.DataBlob)
	}
}
```

You can then run just this benchmark and look at the memory allocation:
```
$ go test -v -run xxxxx -bench IndexHas -benchmem | tee before
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/repository
BenchmarkIndexHasUnknown-4   	 1000000	      1464 ns/op	     608 B/op	      10 allocs/op
BenchmarkIndexHasKnown-4     	 5000000	       280 ns/op	     128 B/op	       5 allocs/op
PASS
ok  	github.com/restic/restic/internal/repository	20.428s
```

So, we can see that each call of `Has()` takes about 300ns and allocates 128 Byte for known blobs, and 1500ns and 600 byte for unknown blobs.

With your change cherry picked, this looks already much better:

```
$ go test -v -run xxxxx -bench IndexHas -benchmem | tee after
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/repository
BenchmarkIndexHasUnknown-4   	10000000	       143 ns/op	      16 B/op	       2 allocs/op
BenchmarkIndexHasKnown-4     	 5000000	       276 ns/op	     128 B/op	       5 allocs/op
PASS
ok  	github.com/restic/restic/internal/repository	22.466s
```

We can use `benchcmp` to compare it:

```
$ benchcmp before after
benchmark                      old ns/op     new ns/op     delta
BenchmarkIndexHasUnknown-4     1464          143           -90.23%
BenchmarkIndexHasKnown-4       280           276           -1.43%

benchmark                      old allocs     new allocs     delta
BenchmarkIndexHasUnknown-4     10             2              -80.00%
BenchmarkIndexHasKnown-4       5              5              +0.00%

benchmark                      old bytes     new bytes     delta
BenchmarkIndexHasUnknown-4     608           16            -97.37%
BenchmarkIndexHasKnown-4       128           128           +0.00%
```

With my suggestions, we can even improve further:

```
$ go test -v -run xxxxx -bench IndexHas -benchmem | tee after2
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/repository
BenchmarkIndexHasUnknown-4   	20000000	        76.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkIndexHasKnown-4     	20000000	        72.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/restic/restic/internal/repository	22.371s
```

And benchcmp agrees: :smile:
```
$ benchcmp before after2
benchmark                      old ns/op     new ns/op     delta
BenchmarkIndexHasUnknown-4     1464          76.4          -94.78%
BenchmarkIndexHasKnown-4       280           72.9          -73.96%

benchmark                      old allocs     new allocs     delta
BenchmarkIndexHasUnknown-4     10             0              -100.00%
BenchmarkIndexHasKnown-4       5              0              -100.00%

benchmark                      old bytes     new bytes     delta
BenchmarkIndexHasUnknown-4     608           0             -100.00%
BenchmarkIndexHasKnown-4       128           0             -100.00%
```

So, if it's okay for you, I'd like to take over from here. Thank you very very much for pointing me to this improvement opportunity!

MJDSys:
Ah, I didn't realize you meant that kind of benchmark.  Sorry about that.

Feel free to take over this pr if you want.  I pushed my changes to the Has function (undoing my other changes).  I also whipped up a quick test for the Has function (though it seems thoroughly tested by all the other tests), so feel free to grab that as well.  I have checked the "Allow Maintainers to make changes" button if you want to just add to this pr.

I think there is still some optimizations available to get my backup to run faster, so I'll look into those as well.  If I find any, I'll try to add a mirco-benchmark for those as well next time.

fd0:
Cool, thanks! I'll reformat the title of your commit, but otherwise leave it intact so that you get the credit for the change. Cheers!

jojomi:
Thanks to @fd0 and @MJDSys for so nicely working together in the way only OSS can enable to the benefit of us all. Nice improvement you found there and I would love to see more of these "little things" improved :).

fd0:
@MJDSys FYI, this PR cut my backup time on a larger server (~250GB data size, ~400GB repo size) in half: from ~45 min to ~20 min. Thanks again!

Issue #1541, opened 2018-01-10
Unnecessary backend.Test calls during check


ifedorenko:
Current `restic check` implementation first performs [`backend.Test`](https://github.com/restic/restic/blob/v0.8.1/internal/checker/checker.go#L194) to verify all pack files referenced in the index are indeed available in the backend, then it performs [`backend.List`](https://github.com/restic/restic/blob/v0.8.1/internal/checker/checker.go#L245) to locate orphaned pack files. 

For 500GB repository this results in ~100K remote http requests (assuming 5MB average pack file size), and with average 100ms http roundtrip time `restic check` on 5 concurrent connections takes ~33 minutes.

I think the same check can be implemented with single `backend.List` to get set of all available packs, which can then be matched with packs referenced in the index to determine missing and orphaned packs without additional remote calls.

fd0:
Thanks for the hint!

fd0:
Resolved in #1548

PR #1548, opened 2018-01-12
checker: Optimize checker.Packs()

ifedorenko:

Use result of single repository.List() to find both missing and
orphaned data packs. For 500GB repository this eliminates ~100K
repository.Test() calls and improves check time by >30M in my
environment (~45min before this change and ~7min after).

See #1541

### Checklist

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -30,8 +30,7 @@ type Checker struct {
 		sync.Mutex
 		M map[restic.ID]uint
 	}
-	indexes       map[restic.ID]*repository.Index
-	orphanedPacks restic.IDs
+	indexes map[restic.ID]*repository.Index
 
 	masterIndex *repository.MasterIndex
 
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -183,75 +182,35 @@ func (e PackError) Error() string {
...
+	for id := range c.repo.List(ctx, restic.DataFile) {
+		repoPacks.Insert(id)
 	}
 
-	for id := range c.packs {
-		seenPacks.Insert(id)
-		IDChan <- id
+	// we were asked to leave
+	select {
+	case <-ctx.Done():
```
fd0: This check needs to be moved to the loop above (loading the list of IDs from the backend may take some time, and we want to check for the operation being cancelled during the load).

internal/checker/checker.go
```diff
@@ -183,75 +182,35 @@ func (e PackError) Error() string {
...
+	default:
 	}
-	close(IDChan)
 
-	debug.Log("waiting for %d workers to terminate", defaultParallelism)
-	workerWG.Wait()
-	debug.Log("workers terminated")
+	// orphaned: present in the repo but not in c.packs
+	for orphanID := range repoPacks.Sub(c.packs) {
+		errChan <- PackError{ID: orphanID, Orphaned: true, Err: errors.New("not referenced in any index")}
```
fd0: Use `select` here and check `ctx.Done()`.

internal/checker/checker.go
```diff
@@ -183,75 +182,35 @@ func (e PackError) Error() string {
...
-			c.orphanedPacks = append(c.orphanedPacks, id)
-			select {
-			case <-ctx.Done():
-				return
-			case errChan <- PackError{ID: id, Orphaned: true, Err: errors.New("not referenced in any index")}:
-			}
-		}
+	// missing: present in c.packs but not in the repo
+	for missingID := range c.packs.Sub(repoPacks) {
+		errChan <- PackError{ID: missingID, Err: errors.New("does not exist")}
```
fd0: Again, use `select` and check `ctx.Done()`.

fd0 (changes requested):
Awesome idea, great catch, I like it a lot! I've annotated a few minor things, afterwards it's ready to be merged!

ifedorenko:
I made the requested changes, thank you for the feedback. 

PS: I just realized CONTRIBUTING.md asks for new commits added to the pull request instead of `git commit --amend` I am used to. My apologies, will use new commits the next time.

fd0:
Thanks! Don't worry about amending the patches, I've just changed the text in CONTRIBUTING.md slightly, amending commits is also fine. What's the best solution (amending or adding new commits) depends on the situation, use common sense. It's fine here :)

fd0 (commented):
LGTM!

PR #1549, opened 2018-01-12
More optimizations to avoid calling Index.Lookup()

MJDSys:

Similar to #1538, avoid calling Index.Lookup in a couple more cases.  This doesn't have the same performance benefit as #1538, but it does reduce cpu usage some on my desktop.

The first commit changes the way MasterIndex preforms lookups in its child indices, by first checking if the index has the given blob.  On repositories with large number of indices (my two backups have 445 and 763 each), the creation of the error for each attempted lookup easily shows up in the cpu profile.  I don't think this change has any practical threading related issues, even though the lock for the index is lost and regained between the Has and Lookup call.

The second commit changes the archiver to use Has over Lookup.  The archiver was using Lookup as a Has function call, so this was a straightforward change.  I don't have a test/benchmark for it, as the function is unexposed and I wasn't sure how you wanted to handle that case.  Should I make a separate test file (archiver_internal_test.go) for it, or just use the Save function and benchmark that?  If a test/benchmark is wanted, let me know which way you prefer and I can whip one up.

No

### Checklist

### DIFF
--- a/changelog/0.8.2/pull-1549
+++ b/changelog/0.8.2/pull-1549
@@ -1,0 +1,9 @@
+Enhancement: Speed up querying across indices and scanning existing files
+
+This change increases the whenever a blob (part of a file) is searched for in a
+restic repository.  This will reduce cpu usage some when backing up files already
+backed up by restic.  Cpu usage is further decreased when scanning files.
+
+https://github.com/restic/restic/issues/1234
+https://github.com/restic/restic/pull/55555
+https://forum.restic/.net/foo/bar/baz
--- a/internal/repository/master_index.go
+++ b/internal/repository/master_index.go
@@ -25,17 +25,13 @@ func (mi *MasterIndex) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic
 	mi.idxMutex.RLock()
 	defer mi.idxMutex.RUnlock()
 
-	debug.Log("looking up id %v, tpe %v", id.Str(), tpe)
-
 	for _, idx := range mi.idx {
 		blobs, found = idx.Lookup(id, tpe)
 		if found {
-			debug.Log("found id %v: %v", id.Str(), blobs)
 			return
 		}
 	}
 
-	debug.Log("id %v not found in any index", id.Str())
 	return nil, false
 }
 
--- a/internal/repository/master_index.go
+++ b/internal/repository/master_index.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"sync"
 
-	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 
 	"github.com/restic/restic/internal/debug"
@@ -22,36 +21,36 @@ func NewMasterIndex() *MasterIndex {
 }
 
 // Lookup queries all known Indexes for the ID and returns the first match.
-func (mi *MasterIndex) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic.PackedBlob, err error) {
+func (mi *MasterIndex) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic.PackedBlob, found bool) {
 	mi.idxMutex.RLock()
 	defer mi.idxMutex.RUnlock()
 
 	debug.Log("looking up id %v, tpe %v", id.Str(), tpe)
 
 	for _, idx := range mi.idx {
-		blobs, err = idx.Lookup(id, tpe)
-		if err == nil {
+		blobs, found = idx.Lookup(id, tpe)
+		if found {
 			debug.Log("found id %v: %v", id.Str(), blobs)
 			return
 		}
 	}
 
 	debug.Log("id %v not found in any index", id.Str())
-	return nil, errors.Errorf("id %v not found in any index", id)
+	return nil, false
 }
 
 // LookupSize queries all known Indexes for the ID and returns the first match.
-func (mi *MasterIndex) LookupSize(id restic.ID, tpe restic.BlobType) (uint, error) {
+func (mi *MasterIndex) LookupSize(id restic.ID, tpe restic.BlobType) (uint, bool) {
 	mi.idxMutex.RLock()
 	defer mi.idxMutex.RUnlock()
 
 	for _, idx := range mi.idx {
-		if idx.Has(id, tpe) {
-			return idx.LookupSize(id, tpe)
+		if size, found := idx.LookupSize(id, tpe); found {
+			return size, found
 		}
 	}
 
-	return 0, errors.Errorf("id %v not found in any index", id)
+	return 0, false
 }
 
 // ListPack returns the list of blobs in a pack. The first matching index is
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -165,8 +165,8 @@ func runCat(gopts GlobalOptions, args []string) error {
 
 	case "blob":
 		for _, t := range []restic.BlobType{restic.DataBlob, restic.TreeBlob} {
-			list, err := repo.Index().Lookup(id, t)
-			if err != nil {
+			list, found := repo.Index().Lookup(id, t)
+			if !found {
 				continue
 			}
 			blob := list[0]
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -135,9 +135,9 @@ func updateBlobs(repo restic.Repository, blobs restic.BlobSet, stats *DiffStat) 
 			stats.TreeBlobs++
 		}
 
-		size, err := repo.LookupBlobSize(h.ID, h.Type)
-		if err != nil {
-			Warnf("unable to find blob size for %v: %v\n", h, err)
+		size, found := repo.LookupBlobSize(h.ID, h.Type)
+		if !found {
+			Warnf("unable to find blob size for %v\n", h)
 			continue
 		}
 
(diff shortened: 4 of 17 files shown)


internal/repository/master_index.go
```diff
@@ -29,10 +29,12 @@ func (mi *MasterIndex) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic
 	debug.Log("looking up id %v, tpe %v", id.Str(), tpe)
 
 	for _, idx := range mi.idx {
-		blobs, err = idx.Lookup(id, tpe)
-		if err == nil {
-			debug.Log("found id %v: %v", id.Str(), blobs)
-			return
+		if idx.Has(id, tpe) {
```
fd0: This change won't cause any issues, since blobs are only added to an index, so it cannot happen that `Has()` succeeds and subsequent `Lookup()` fails. But now you have the map lookup twice.

I think we can revert this change and do better: Since the only error `Lookup()` ever returns is `not found`, how about we change the function signature like like this:
```go
func (idx *Index) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic.PackedBlob, found bool) {
```

That would mean touching every place where `Lookup()` is called, but shouldn't be too hard to do.

Now the lookup is done only once, and there's no error built in any case.

What do you think?

fd0 (changes requested):

fd0 (changes requested):
And I'd strongly suggest installing the `fmt-check` pre-commit hook from here: https://github.com/edsrzf/gofmt-git-hook

Helped me a lot :)

internal/repository/master_index.go
MJDSys: Sounds good to me, I'll take a look into doing that this weekend.  Hopefully I'll have some good news/better benchmarks!

MJDSys:
Hey @fd0,

I've implemented your suggestion.  For the most part, the performance didn't change.  However, I did notice with your suggestion that the debug logging doubled the time taken and dramatically increased the allocations, even when disabled.  I tracked it down the the ID.Str() function, which allocates a string on call even if not used.

I made a second version that delays the calculation until used, for use with debug printing.  I replaced the version used in Lookup, which helped the micro benchmark.  I didn't see it show up in the backup profiling, so I'm not sure how much you want to take that change.

I originally tried replacing Str with DebugStr, but it broke various for loops in the frontend.  If you wish, I can go back to that design and audit/test all those loops.  But I'm not sure the performance advantage will beat out the risk of breakage (though it should be relatively visible).  Alternately, I could just put those statements in a if statement, and have that be used only when the binary is compiled for debugging.  Let me know what you prefer.

fd0:
> However, I did notice with your suggestion that the debug logging doubled the time taken and dramatically increased the allocations, even when disabled. I tracked it down the the ID.Str() function, which allocates a string on call even if not used.

Hm, is that behavior new with the changes you made? That shouldn't be the case in my opinion, it probably was already present before.

Is it also there when compiled without the `debug` tag? If it isn't I don't really care, as the debug logs will also cost performance in other places. If it is also present without the `debug` tag, then remove the debug messages and move them up the call stack to some higher function (or just remove them, we'll add them back if needed).

MJDSys:
Ugh, my sleepy brain wasn't clear with what was happening.  So the changes in this PR (with or without the Id.DebugStr change) are faster then the previous version.  However, when I changed the Lookup function to return a boolean, that caused the benchmarks to slow down slightly (but still generally faster then the previous version).  The difference was that Index.Lookup has logging, which slowed down the not found case compared to the Has function which has no logging.  Also this is all *without* the debug tag, so the rest of the logging is disabled.

I'll work to remove those log calls, but insert a few statements in some useful cases up the call stack.  I just want to avoid these calls in the usual case, as you still pay the price.

Just to be clear, the reason this is an issue is that when Str() is called, there is some cpu time/memory allocations to return the string immediately.  This happens regardless of whether or not the logging framework is enabled or does anything.  I think currently that overhead is small enough to generally ignore, but in a very hot path there is some measurable overhead.  I brought it up here since it showed up in my profiling.

fd0:
> However, when I changed the Lookup function to return a boolean, that caused the benchmarks to slow down slightly (but still generally faster then the previous version).

Thanks for the explanation. Can you show me the benchmarks?

I've thrown together a quick benchmark for `debug.Log()`:
```
$ go test -v -bench . -benchmem
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/debug
BenchmarkLogStatic-4     	2000000000	         1.77 ns/op	       0 B/op	       0 allocs/op
BenchmarkLogIDString-4   	300000000	         5.58 ns/op	       0 B/op	       0 allocs/op
BenchmarkLogIDStr-4      	100000000	        12.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/restic/restic/internal/debug	7.242s

$ go test -v -tags debug -bench . -benchmem

debug enabled
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/debug
BenchmarkLogStatic-4     	  200000	      8567 ns/op	     456 B/op	      20 allocs/op
BenchmarkLogIDString-4   	  200000	      8626 ns/op	     504 B/op	      21 allocs/op
BenchmarkLogIDStr-4      	  200000	      8644 ns/op	     488 B/op	      21 allocs/op
PASS
ok  	github.com/restic/restic/internal/debug	5.441s
```

We can see that the debug logging costs something, but there aren't any allocations when the `debug` tag isn't set. And since it's just a few nanoseconds (without debug) I wonder if it's worth optimizing further.

If you find good places up the call stack that's fine, otherwise just throw the calls to `debug.Log()` away for now.

MJDSys:
> Thanks for the explanation. Can you show me the benchmarks?
I added a commit containing some benchmarks.  The results I get are noted below, and match about what I expect considering the speeds I got with my commits dealing with the ID.Str.  I'm not sure why you're benchmarks report no difference between the three, that seems really odd to me.  I do see the allocations and time associated when calling with .Str, and see a huge difference when using the debug tag.

If you don't want these benchmarks to stay with the code, please feel free to just pop the commit off this branch and before merging.

```shell
$ go test -run xx -bench . -benchmem -v
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/debug
BenchmarkLogStatic-6            1000000000               2.41 ns/op            0 B/op          0 allocs/op
BenchmarkLogIDStr-6             20000000                94.7 ns/op            16 B/op          2 allocs/op
BenchmarkLogIDString-6          10000000               219 ns/op             128 B/op          2 allocs/op
PASS
ok      github.com/restic/restic/internal/debug 7.089s
$ go test -run xx -bench . -benchmem -v -tags debug
debug enabled
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/debug
BenchmarkLogStatic-6              100000             12826 ns/op             432 B/op         20 allocs/op
BenchmarkLogIDStr-6               100000             14069 ns/op             440 B/op         23 allocs/op
BenchmarkLogIDString-6            100000             14058 ns/op             552 B/op         23 allocs/op
PASS
ok      github.com/restic/restic/internal/debug 4.523s
```

I also took a quick look at places where .Lookup is called (just doing a quick manual check with grep).  Any interesting calls seem to already have a Log call if needed, so I just removed the calls instead.  I also removed the whole commit adding DebugStr from this PR.

fd0:
Nah, leave the benchmarks in. Can you please fix the code (there's an import missing for the `github.com/restic/restic/internal/errors` package)?

fd0 (approved):

changelog/0.8.2/pull-1549
```diff
@@ -0,0 +1,9 @@
+Enchancement: Speed up querying across indices and speed up scanning existing files
```
fd0: Typo: "Enhancement". That'll also fail on Travis when [calens](https://github.com/restic/calens) is run to generate the changelog file:
```
$ calens
file changelog/0.8.2/pull-1549: entry type "Enchancement" is invalid, valid types: map[Enhancement:4 Security:1 Bugfix:2 Change:3]
```

fd0: Also, the title is too long, max is 80 chars. I'd suggest removing the second "speed up":
```
Enhancement: Speed up querying across indices and scanning existing files
```

fd0:
Ah, in master, the `errors` package was removed from several files, especially in `fuse/`, but this code reintroduces it, so the merged commit does not compile.

I suggest rebasing on the master branch, and then running `goimports -w` on the files.

MJDSys:
Code rebased and changelog updated!  Hopefully that's it :)

Out of curiosity, do you see benchmark results similar to mine with my benchmark code?  I'm just curious why you don't see the difference in yours.

fd0 (approved):
LGTM!

fd0:
FYI, Travis is currently broken and does not run our tests on Darwin, so it looks like the tests failed (but they did not).

changelog/0.8.2/pull-1549
```diff
@@ -0,0 +1,9 @@
+Enhancement: Speed up querying across indices and scanning existing files
+
+This change increases the whenever a blob (part of a file) is searched for in a
+restic repository.  This will reduce cpu usage some when backing up files already
+backed up by restic.  Cpu usage is further decreased when scanning files.
+
+https://github.com/restic/restic/issues/1234
+https://github.com/restic/restic/pull/55555
+https://forum.restic/.net/foo/bar/baz
```
fd0: I've corrected the changelog entry, somehow the default URLs were still in it.

fd0:
Ah, I forgot to run the benchmark again, sorry. I'll try to do that later today (also curious).

fd0:
Hm, I think I did something wrong in my benchmarks (I don't have them any more), I'm seeing the same results as you did.

fd0:
I've opened #1582 which removes all manual calls of `id.Str()` in `debug.Log()` and calls it afterwards in `Log()` via reflection.

PR #1552, opened 2018-01-12
Automatically load Google auth

lawrencejones:

This change removes the hardcoded Google auth mechanism for the GCS
backend, instead using Google's provided client library to discover and
generate credential material.

Google recommend that client libraries use their common auth mechanism
in order to authorise requests against Google services. Doing so means
you automatically support various types of authentication, from the
standard GOOGLE_APPLICATION_CREDENTIALS environment variable to making
use of Google's metadata API if running within Google Container Engine.

Not that I'm aware of, via cursory search of issues.

### Checklist

  - I don't _think_ this is required, as we already have tests that exercise GOOGLE_APPLICATION_CREDENTIALS which will cover this and the rest of the behaviour I think we can rely on Google to have implemented properly

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -440,18 +440,6 @@ func parseConfig(loc location.Location, opts options.Options) (interface{}, erro
 			cfg.ProjectID = os.Getenv("GOOGLE_PROJECT_ID")
 		}
 
-		if cfg.JSONKeyPath == "" {
-			if path := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"); path != "" {
-				// Check read access
-				if _, err := ioutil.ReadFile(path); err != nil {
-					return nil, errors.Fatalf("Failed to read google credential from file %v: %v", path, err)
-				}
-				cfg.JSONKeyPath = path
-			} else {
-				return nil, errors.Fatal("No credential file path is set")
-			}
-		}
-
 		if err := opts.Apply(loc.Scheme, &cfg); err != nil {
 			return nil, err
 		}
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -362,8 +362,14 @@ key file and the project ID as follows:
     $ export GOOGLE_PROJECT_ID=123123123123
     $ export GOOGLE_APPLICATION_CREDENTIALS=$HOME/.config/gs-secret-restic-key.json
 
-Then you can use the ``gs:`` backend type to create a new repository in the
-bucket `foo` at the root path:
+We use Google's client library to generate [default authentication
+material](https://developers.google.com/identity/protocols/application-default-credentials),
+which means if you're running in Google Container Engine or are otherwise
+located on an instance with default service accounts then these should work out
+the box.
+
+Once authenticated, you can use the ``gs:`` backend type to create a new
+repository in the bucket `foo` at the root path:
 
 .. code-block:: console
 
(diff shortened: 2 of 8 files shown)


lawrencejones:
Going to make sure this works in Google then I'll check the ready for review box

cmd/restic/global.go
```diff
@@ -438,18 +438,6 @@ func parseConfig(loc location.Location, opts options.Options) (interface{}, erro
 			cfg.ProjectID = os.Getenv("GOOGLE_PROJECT_ID")
 		}
 
-		if cfg.JSONKeyPath == "" {
```
lawrencejones: We don't need this anymore, as Google's library handles various auth mechanisms for us. We'll be less explicit about why we're failing but we gain support for several authentication methods- swings and roundabouts!

internal/backend/gs/gs_test.go
```diff
@@ -100,8 +99,8 @@ func TestBackendGS(t *testing.T) {
 
 func BenchmarkBackendGS(t *testing.B) {
 	vars := []string{
+		"GOOGLE_APPLICATION_CREDENTIALS",
```
lawrencejones: Because we don't control the authentication now, we have to use the non-namespaced environment variable.

lawrencejones:
Forgot to say- this is working for us absolutely fine from infrastructure both within and outside of Google Cloud Platform. The instances inside GCP are making use of their default service accounts and the metadata API to auth, while the other apps are using a service account JSON credential.

This makes things way easier for us- no doubt it's a huge help for others too!

lawrencejones:
@fd0 are you good on this approach? If you are then I'll fixup the conflicts- keen to get it in so more things don't change under me!

changelog/0.8.2/pull-1552
```diff
@@ -0,0 +1,12 @@
+Feature: Use Google Application Default credentials
```
fd0: This file needs to be moved to `changelog/0.8.3` now, we've released 0.8.2 yesterday :)

cmd/restic/global.go
fd0: Fine with me, if it causes problems we can add more error messages.

doc/030_preparing_a_new_repo.rst
```diff
@@ -362,8 +362,14 @@ key file and the project ID as follows:
     $ export GOOGLE_PROJECT_ID=123123123123
     $ export GOOGLE_APPLICATION_CREDENTIALS=$HOME/.config/gs-secret-restic-key.json
 
-Then you can use the ``gs:`` backend type to create a new repository in the
-bucket `foo` at the root path:
+We use Google's client library to generate [default authentication
```
fd0: Replace `We` with `Restic`. That's more consistent with the rest of the text.

internal/backend/gs/gs_test.go
fd0: Ok, fine with me.

fd0 (changes requested):
Hey, thanks for the contribution! I'm sorry that I did not find the time to review this earlier.

I've marked a few minor changes, then we can merge this! And you'll need to rebase this onto master ;)

Let me know if you need help!

fd0:
Hey, are you still interested in this PR? There's just a few minor things to edit, then I can merge it...

lawrencejones:
Hey @fd0, really sorry for the delay- this dropped off my radar since I first pr'd the change!

Will tidy this up and get it ok to merge today.

lawrencejones:
Should be good now!

fd0 (approved):

Issue #1497, opened 2017-12-24
option to check subset of repository data blobs


ifedorenko:
## Output of `restic version`

    restic 0.8.0
    compiled with go1.9 on darwin/amd64

## How did you run restic exactly?
`restic check --read-data`

## What backend/server/service did you use to store the repository?
any remote backend

## Expected behavior
I have 50Mbps Internet connection and 1TB remote backup repository.  I'd like to be able to check a subset of data blobs for few hours every night, but still check each blob at least once a month.

## Actual behavior
There is no way to limited how much work `check --read-data` does. In my case, repository data check will take more than 2 days to complete when using all available Internet connection bandwidth, and much longer if I throttle download speed.

## Steps to reproduce the behavior
run `restic check --read-data`

## Do you have any idea what may have caused this?
This is a feature request, not a bug report.

## Do you have an idea how to solve the issue?
One way to implement this is to introduce two new `check --read-data` flags: total number of blob "buckets" and index of the bucket to check. Blobs can be assigned to buckets based on the first byte of blob's sha256, for example. My desired check schedule (or any other periodic check schedule) can be trivially implemented in bash or other scripting language. I can try to provide a pull request that implements this idea, if you are interested.

## Did restic help you or made you happy in any way?

go looks weird to a java developer :-)

fd0:
That's a nice idea, thanks for reporting it! We need to come up with a good partition algorithm though, to ensure that it is guaranteed that each blob is checked.

ifedorenko:
If `totalPartitions` is the total number of partitions and `currentPartition` is the partition to check (`0 <= currentPartition <  totalPartitions`),  then each block can be assigned to a partition by simple `(id[0] % totalPartitions) == currentPartition` check on the first byte of the block id. https://github.com/ifedorenko/restic/commit/3a91af94cfedda3c752530e93684454858db6065 implements this approach and I can open pull request if you a interested. I am still debating if I want a separate `restic check-data` command, however...

yorkday:
I have been testing the --read-data-subset command and believe there may be an issue and a possible enhancement option.

**Possible issue:**
When choosing a large number of totalPartitions for a small backup, and checking only one partition, the command may not actually test any data.  For example, in a small 10Mb backup, I choose to test a single random partition out of totalPartitions = 100 (effectively testing a random 1%).  The test did not appear to validate any data - perhaps because the partition had no blocks assigned?
[...]

PR #1556, opened 2018-01-15
Add --read-data-subset flag to check command

ifedorenko:

This change introduces ability to check integrity of a subset of repository 
data packs. This can be used to spread integrity check of larger repositories
over period of time.

See #1497

### Checklist

### DIFF
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -26,13 +28,17 @@ repository and not use a local cache.
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return runCheck(checkOptions, globalOptions, args)
 	},
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		return checkFlags(checkOptions)
+	},
 }
 
 // CheckOptions bundles all options for the 'check' command.
 type CheckOptions struct {
-	ReadData    bool
-	CheckUnused bool
-	WithCache   bool
+	ReadData       bool
+	ReadDataSubset string
+	CheckUnused    bool
+	WithCache      bool
 }
 
 var checkOptions CheckOptions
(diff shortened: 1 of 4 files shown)


rawtaz:
Will there ever be a `--write-data-subset`? I guess there won't be, making the "read" part in `--read-data-subset` pretty redundant?

Also, the "data" part might be redundant too, unless there's a forecast of introducing this feature for only metadata or similar.

So in short, I'm wondering if `--subset` wouldn't be enough to have as the option name, considering it is local to the `check` command.

ifedorenko:
I named the new flag after existing `read-data` but don't have strong opinion about the flag name. I do think "data" part of the name is important because the flag does not affect index and snapshots checks.

changelog/0.8.2/issue-1497
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add --read-data-subset flag to check command
```
fd0: This file needs to be moved to `changelog/0.8.3`, we've released 0.8.2 yesterday :)

cmd/restic/cmd_check.go
```diff
@@ -42,10 +46,27 @@ func init() {
...
 	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use the cache")
 }
 
+func checkFlags(opts CheckOptions) error {
+	if opts.ReadData && opts.ReadDataSubset != nil {
+		return errors.Errorf("check flags --readData and --read-data-subset cannot be used together")
+	}
+	if opts.ReadDataSubset != nil {
+		if len(opts.ReadDataSubset) != 2 {
+			return errors.Errorf("check flag --read-data-subset must have two values")
```
fd0: Use `errors.Fatalf()` here, so that no stack trace is printed (we need to rework error handling in restic, badly). Also, maybe add an example?

cmd/restic/cmd_check.go
```diff
@@ -42,10 +46,27 @@ func init() {
 
 	f := cmdCheck.Flags()
 	f.BoolVar(&checkOptions.ReadData, "read-data", false, "read all data blobs")
+	f.UintSliceVar(&checkOptions.ReadDataSubset, "read-data-subset", nil, "read subset of data packs")
```
fd0: Hm, maybe make it a string, and split it ourselves? The `*SliceVar` functions of `pflag` have some magic that bit us before, e.g. `--read-data-subset 23 --read-data-subset 24` is the same as `--read-data-subset 23,24`.

Maybe use `/` as the  delimiter? `1/5` reads (at least to me) as `one of five`)?

doc/045_working_with_repos.rst
```diff
@@ -87,3 +87,29 @@ yield the same error:
     Load indexes
     ciphertext verification failed
 
+By default, ``check`` command does not check data packs integrity.
```
fd0: replace "data packs integrity" with "that files stored in the repository are unmodified". This is end-user documentation, being less technical is a feature :)

fd0 (changes requested):
Nice idea overall, thank you very much for your contribution!

I've added a few comments. The PR needs to be rebased on master, which should not cause any issues.

Please accept my apologies for having taken so long to review this!

doc/045_working_with_repos.rst
ifedorenko: Maybe too fine a point here, but I think most users will assume that "files stored in the repository" mean "_user_ files backed up to the repository", which ``check`` does not actually check (see #1620). I purposely used very technical "data packs" to avoid this confusion. I now changed the text to use less technical "repository data files", which I hope will be less likely confused with "user files stored in the repository". Let me know if you prefer different wording.

ifedorenko:
Thank you for the feedback, I think I made all the requested changes but let me know if I missed anything.

PS: You have nothing to apologize for. I fully understand it takes time and effort to do review and provide feedback about external contributions. And from my experience `restic` is one of the better projects when it comes to reviewing and accepting external contributions.

fd0 (approved):

fd0:
Cool, thanks!

Issue #527, opened 2016-06-04
Storage Backend: Google Drive


Intensity:
This is a feature request for Google Drive as a storage backend.

fd0:
Thanks for the suggestion.

middelink:
FYI
https://developers.google.com/drive/v3/web/quickstart/go
https://github.com/google/google-api-go-client/blob/master/drive/v3/drive-gen.go

rxng:
any short term plans to implement this?

fd0:
To be honest with you, there are no short-term plans for this. Sorry.

xor-gate:
The people from the `rclone` tool have created an drive backend here: https://github.com/ncw/rclone/tree/master/drive

naffit:
Hi, any possible news on this?
The latest Crashplan changes have made me look for an alternative and the unlimited space available when using G Suite for Business is really interesting - and restic seems like a better alternative than Arq or Duplicati since I can run it natively on my FreeBSD box..

fd0:
There are no news here, it's one of the more problematic backends, requiring a client secret (that should be kept secret) which needs to be embedded in the Go source code.

You can make a local backup and move that to Google Drive with e.g. rclone.

tamalsaha:
@fd0, from https://developers.google.com/identity/protocols/OAuth2InstalledApp , it seems to me that if this  mode is used the client secret do not need to be kept confidential.

naffit:
@fd0 Alas, that's my initial problem - with 6TB+ of data that is not really an option..

As @tamalsaha mentions others are using OAuth without problems.
The team behind Duplicati has even published this explanation: https://www.duplicati.com/articles/OAuth/ and uses https://duplicati-oauth-handler.appspot.com/ to request/revoke tokens. 

The explanation is based on OneDrive, but it's the same procedure with Google Drive and their OAuth-handler lists a lot of providers based on OAuth.

xor-gate:
Hi guys, I started to play around some time ago with the restic source code to give it a shot to integrate gdrive API package. But there is a lot of duplication going on (needed to change 3 places to add a backend). I have not proceeded because I moved on to some other stuff. https://github.com/xor-gate/restic/compare/master...add-gdrive

bherila:
With Google File Stream you might be able to back up restic today using just the GFS mount point?

naffit:
[...]

PR #1560, opened 2018-01-17
Retry all repository file download errors

ifedorenko:

Reworked Backend.Load API to support retry errors during ongoing download.

See #527 and #1600

### Checklist

- [x] There's an entry in the `CHANGELOG.md` file that describe the changes for our users

### DIFF
--- a/internal/backend/utils.go
+++ b/internal/backend/utils.go
@@ -10,24 +10,11 @@ import (
 
 // LoadAll reads all data stored in the backend for the handle.
 func LoadAll(ctx context.Context, be restic.Backend, h restic.Handle) (buf []byte, err error) {
-	rd, err := be.Load(ctx, h, 0, 0)
-	if err != nil {
-		return nil, err
-	}
-
-	defer func() {
-		_, e := io.Copy(ioutil.Discard, rd)
-		if err == nil {
-			err = e
-		}
-
-		e = rd.Close()
-		if err == nil {
-			err = e
-		}
-	}()
-
-	return ioutil.ReadAll(rd)
+	err = be.Load(ctx, h, 0, 0, func(rd io.Reader) (ierr error) {
+		buf, ierr = ioutil.ReadAll(rd)
+		return ierr
+	})
+	return buf, err
 }
 
 // LimitedReadCloser wraps io.LimitedReader and exposes the Close() method.
@@ -45,4 +32,20 @@ func (l *LimitedReadCloser) Read(p []byte) (int, error) {
 // exposes the Close() method.
 func LimitReadCloser(r io.ReadCloser, n int64) *LimitedReadCloser {
 	return &LimitedReadCloser{ReadCloser: r, Reader: io.LimitReader(r, n)}
+}
+
+// DefaultLoad implements Backend.Load using lower-level openReader func
+func DefaultLoad(ctx context.Context, h restic.Handle, length int, offset int64,
+	openReader func(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error),
+	fn func(rd io.Reader) error) error {
+	rd, err := openReader(ctx, h, length, offset)
+	if err != nil {
+		return err
+	}
+	err = fn(rd)
+	if err != nil {
+		rd.Close() // ignore secondary errors closing the reader
+		return err
+	}
+	return rd.Close()
 }
--- a/internal/repository/repack.go
+++ b/internal/repository/repack.go
@@ -2,14 +2,11 @@ package repository
 
 import (
 	"context"
-	"crypto/sha256"
 	"fmt"
-	"io"
 	"os"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/fs"
-	"github.com/restic/restic/internal/hashing"
 	"github.com/restic/restic/internal/pack"
 	"github.com/restic/restic/internal/restic"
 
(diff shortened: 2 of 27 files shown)


ifedorenko:
I am done with this prototype and would appreciate some feedback.

The main idea of the change is to let `Backend.Load` retain ownership of the file Reader, which allows the backend implement download-retry logic and guarantee open readers are never leaked by the client.

For example, here is how updated checker code looks like:

```go
packfile, err := fs.TempFile("", "restic-temp-check-")
...
var hash restic.ID
var size int64
err = r.Backend().Load(ctx, h, 0, 0, func(rd io.Reader) (err error) {
	_, err = packfile.Seek(0, io.SeekStart)
	if err == nil {
		err = packfile.Truncate(0)
	}
	if err != nil {
		return err
	}
	hrd := hashing.NewReader(rd, sha256.New())
	size, err = io.Copy(packfile, hrd)
	hash = restic.IDFromHash(hrd.Sum(nil))
	return err
})
if err != nil {
	return errors.Wrap(err, "checkPack")
}
```

Note that content consumer callback must be idempotent to handle download retries.

fd0:
This is a very interesting idea, I'll think about it! It's astonishing what you can do with closures.

fd0:
Also needs a rebase to master, sorry ;)

fd0:
I've thought about your idea, and I like it! This can also be done for `Save()`...

ifedorenko:
Apart from few TBD method comments, anything you'd like me to fix/change in the pull request?

I don't particularly like how `DefaultLoaderBackend` is injected in backends, but I don't know how to do this better/cleaner in Go. Any suggestions?

Also, I prefer to deal with `Save` separately. I don't see how to apply callback to Save, or at least what advantages that would provide, to be honest.

internal/restic/backend.go
```diff
@@ -55,3 +55,23 @@ type FileInfo struct {
...
+// DefaultLoaderBackend TBD
+type DefaultLoaderBackend struct {
+	LoaderBackend
+}
+
+// Load implements Backend using lower-level OpenReader func
+func (be *DefaultLoaderBackend) Load(ctx context.Context, h Handle, length int, offset int64, consumer func(rd io.Reader) error) error {
+	rd, err := be.OpenReader(ctx, h, length, offset)
+	if err != nil {
+		return err
```
fd0: Add `_ = rd.Close()` here to make it explicit (remove the `defer` below)

internal/restic/backend.go
```diff
@@ -55,3 +55,23 @@ type FileInfo struct {
...
+	}
+	defer rd.Close() // ignore errors closing the reader
```
fd0: Remove the defer, I've been bitten by defer before, and I prefer to have this dealt with explicitly.

internal/restic/backend.go
```diff
@@ -55,3 +55,23 @@ type FileInfo struct {
...
+	return consumer(rd)
```
fd0: Then we need to deal with the error here explicitly also, e.g.:
```go
err = consumer(rd)
if err != nil {
  _ = rd.Close()
  return err
}
return rd.Close()
```

internal/backend/utils.go
```diff
@@ -10,24 +10,11 @@ import (
...
-
-		e = rd.Close()
-		if err == nil {
-			err = e
-		}
-	}()
-
-	return ioutil.ReadAll(rd)
+	err = be.Load(ctx, h, 0, 0, func(rd io.Reader) error {
+		buf, err = ioutil.ReadAll(rd)
```
fd0: Hm, I'm not sure if we should use the same `err` variable within the func and outside. That may lead to problems when the func is called repeatedly. What are your thoughts?

What about this:
```go
err = be.Load(ctx, h, 0, 0, func(rd io.Reader) (ierr error) {
￼		buf, ierr = ioutil.ReadAll(rd)
        return ierr
￼￼})
```

This happens a lot further down in the commit.

internal/repository/repack.go
```diff
@@ -32,23 +32,26 @@ func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, kee
...
 		}
 
-		beRd, err := repo.Backend().Load(ctx, h, 0, 0)
-		if err != nil {
-			return nil, err
-		}
-
-		hrd := hashing.NewReader(beRd, sha256.New())
-		packLength, err := io.Copy(tempfile, hrd)
+		// TODO very similar code in checker, consider moving to utils.go
```
fd0: Don't leave "TODO" comments, I consider this a code smell ;)

fd0 (changes requested):
Let's leave `Save` out for now, we'll talk about it in a different issue.

I also don't like how the `DefaultLoaderBackend` is used, it touches way too much points for so few lines of code. IMHO, the abstraction is not worth it and makes the code more complicated. At the moment, I'd prefer to just copy the few lines into every backend. We should add tests for that into the backend tests, to make sure the backends properly deal with the function we pass into `Load()` (close the file on error, passes back the error returned by the func etc.)

The modification of the `Backend` `Load()` method looks fine to me, and I also like the way it improves the code using a backend!

So, what we need to do in my opinion:
 * Remove the `DefaultLoaderBackend`, copy the code into each backend
 * Process the commits that I've added inline

ifedorenko:
@fd0 I replaced `restic.DefaultLoaderBackend` with `backend.DefaultLoad` func. 

I still don't like the idea of copy&pasting the same Load implementation in all backends, especially given how many comments you had around error handling, but not sure if the new `backend.DefaultLoad` is much better compared to my original proposal. What do you think?

(I'll update the pull request according to your other comments hopefully today, but may take few days)

ifedorenko:
I am done with the code changes but still need to do git history grooming. I'll prepare pull request for merge tomorrow if you don't see any other problems. Please let me know.

fd0 (approved):
LGTM!

I don't really like having to call `backend.DefaultLoad` everywhere, but I failed to come up with a better idea, so we're going to use that.

ifedorenko:
Okay I'll finish the pull request later tonight.

fd0 (approved):

PR #1564, opened 2018-01-17
sftp: Prompt for password, don't terminate on SIGINT

fd0:
This is a follow-up on fb9729fdb9bf1e15dc6c90383df85a8968f6ddb0, which runs the `ssh` in its own process group and selects that process group as the foreground group. After the sftp connection is established, restic switches back to the previous foreground process group. 

This allows `ssh` to prompt for the password, but it won't receive the interrupt signal (SIGINT, ^C) later on, because it is not in the foreground process group any more, allowing a clean tear down.

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -22,17 +22,7 @@ var stderr = os.Stderr
 func init() {
 	cleanupHandlers.ch = make(chan os.Signal)
 	go CleanupHandler(cleanupHandlers.ch)
-	InstallSignalHandler()
-}
-
-// InstallSignalHandler listens for SIGINT, and triggers the cleanup handlers.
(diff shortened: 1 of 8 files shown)


rawtaz:
I cross-compiled restic using `go run build.go --goos freebsd --goarch amd64` after checking out the `sftp-password-prompt` branch in this repository, and then ran it on a FreeBSD machine:

~~~
% ./restic -r sftp://HOST/restic unlock
rawtaz@HOST's password:
password is correct
successfully removed locks
~~~

Is that what you wanted to see, that it/SSH asks and successfully retrieves the password for the host? It seems to do just that. This was with a plain password.

fd0:
Cool, thanks for checking it!

Issue #1567, opened 2018-01-19
Possible index.New() opimizations


ifedorenko:
Current `index.New()` implementation can be abbreviated as following
1. list all pack file names in the repository
2. for each pack file
    - determine file size
    - download pack header length (last 4 bytes of the pack file)
    - download pack header (header-length number of bytes before the header length)
3. index blob info from pack headers

For 500GB repository with 100K pack files this results in 300K repository requests. With averate request time 100ms and 5 concurrent backend connections, it takes ~100 minutes to execute the requests.

I'd like to propose two optimizations to the current implementation:

* Change `backend.List()` to return both file name and file size. I believe this can be implemented efficiently for all backends and will eliminate per-pack backend call to determine pack size (saving ~100K requests and ~33 minutes for the 100K-pack repository)

* Change `Repository.ListPack()` to speculatively request header length + fixed number of extra bytes with single `backend.Load()` request. In my test repository 98% of all pack file headers are 365 bytes or less (I can provide raw data if anyone is interested). So if `Repository.ListPack()` unconditionally downloaded 365+4 byte for each pack file, it would only need 102K requests to download all pack headers, saving another ~32 minutes for the 100K-packs repository. (note that typical http header is 700-800 according to [google spdy paper](http://dev.chromium.org/spdy/spdy-whitepaper), so eliminating one per pack request will almost certainly offset any additional traffic caused by the extra 365 bytes download. at least for http-based backends). 

What do you think?

fd0:
Good ideas, thank you! I think we can do that, shouldn't be too hard to do.

I already thought about changing the `List()` implementation so that it can properly return an error, we can do that in one run. What are your thoughts about it?

ifedorenko:
I'd just change `List()` to return set of file info objects and an error. To support repositories with millions of packs we probably want to be smart about memory and offload the set to disk, but I am not sure there are much benefits to the current channel-based API. In case you are worried about concurrency, backend can always run List on multiple goroutings internally, [like I already do for my OneDrive backend](https://github.com/ifedorenko/restic/blob/947f4822b2027901524f76ca6241dcbd0edb7eca/internal/backend/onedrive/onedrive.go#L748-L774).

[...]

PR #1571, opened 2018-01-21
Rework backend list

fd0:

It changes the signature of the `List()` functions of the `Backend` and `Repository` interfaces to take a function (similar to `filepath.Walk()`) and return not only a name/ID, but also a size.

https://github.com/restic/restic/issues/1567

### TODO

 * [x] change the `List()` function for `Repository` and `Backend`
 * [x] fix all uses of `List()`
 * [x] change all backends to provide size information
 * [x] amend the REST protocol to provide size information directly in the list response
 * [x] implement the REST v2 protocol in the backend

### DIFF
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -249,17 +249,17 @@ func (s *Suite) TestList(t *testing.T) {
 	b := s.open(t)
 	defer s.close(t, b)
 
-	list1 := restic.NewIDSet()
+	list1 := make(map[restic.ID]int64)
 
 	for i := 0; i < numTestFiles; i++ {
-		data := []byte(fmt.Sprintf("random test blob %v", i))
+		data := test.Random(rand.Int(), rand.Intn(100)+55)
 		id := restic.Hash(data)
 		h := restic.Handle{Type: restic.DataFile, Name: id.String()}
 		err := b.Save(context.TODO(), h, bytes.NewReader(data))
 		if err != nil {
 			t.Fatal(err)
 		}
-		list1.Insert(id)
+		list1[id] = int64(len(data))
 	}
 
 	t.Logf("wrote %v files", len(list1))
@@ -272,7 +272,7 @@ func (s *Suite) TestList(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(fmt.Sprintf("max-%v", test.maxItems), func(t *testing.T) {
-			list2 := restic.NewIDSet()
+			list2 := make(map[restic.ID]int64)
 
 			type setter interface {
 				SetListMaxItems(int)
@@ -288,7 +288,7 @@ func (s *Suite) TestList(t *testing.T) {
 				if err != nil {
 					t.Fatal(err)
 				}
-				list2.Insert(id)
+				list2[id] = fi.Size
 				return nil
 			})
 
@@ -298,9 +298,22 @@ func (s *Suite) TestList(t *testing.T) {
 
 			t.Logf("loaded %v IDs from backend", len(list2))
 
-			if !list1.Equals(list2) {
-				t.Errorf("lists are not equal, list1 %d entries, list2 %d entries",
-					len(list1), len(list2))
+			for id, size := range list1 {
+				size2, ok := list2[id]
+				if !ok {
+					t.Errorf("id %v not returned by List()", id.Str())
+				}
+
+				if size != size2 {
+					t.Errorf("wrong size for id %v returned: want %v, got %v", id.Str(), size, size2)
+				}
+			}
+
+			for id := range list2 {
+				_, ok := list1[id]
+				if !ok {
+					t.Errorf("extra id %v returned by List()", id.Str())
+				}
 			}
 		})
 	}
@@ -349,8 +362,8 @@ func (s *Suite) TestListCancel(t *testing.T) {
 			return nil
 		})
 
-		if err != context.Canceled {
-			t.Fatalf("expected error not found, want %v, got %v", context.Canceled, err)
+		if errors.Cause(err) != context.Canceled {
+			t.Fatalf("expected error not found, want %v, got %v", context.Canceled, errors.Cause(err))
 		}
 	})
 
@@ -404,7 +417,7 @@ func (s *Suite) TestListCancel(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.TODO())
 		defer cancel()
 
-		timeout := 500 * time.Millisecond
+		timeout := 200 * time.Millisecond
 
 		ctxTimeout, _ := context.WithTimeout(ctx, timeout)
 
@@ -414,7 +427,7 @@ func (s *Suite) TestListCancel(t *testing.T) {
 			i++
 
 			// wait until the context is cancelled
-			time.Sleep(timeout)
+			time.Sleep(timeout + 200*time.Millisecond)
 			return nil
 		})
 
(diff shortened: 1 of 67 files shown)


zcalusic:
@fd0, I can't find where is the needed change in REST API documented in these commits. Maybe I'm just lost. :(

REST API documentation is my primary guide to what needs to be implemented in rest-server. ;)

fd0:
Sorry, I haven't had the time to propose something, so it's not yet in the code here

internal/repository/repository.go
```diff
@@ -536,22 +536,15 @@ func (r *Repository) KeyName() string {
...
-		for strID := range r.be.List(ctx, t) {
-			if id, err := restic.ParseID(strID); err == nil {
-				select {
-				case out <- id:
-				case <-ctx.Done():
-					return
-				}
-			}
+// List runs fn for all files of type t in the repo.
+func (r *Repository) List(ctx context.Context, t restic.FileType, fn func(restic.ID, int64) error) error {
```
ifedorenko: Why callback function gets two separate `restic.ID` and `int64` parameters, and not single `restic.FileInfo` parameter?

internal/restic/backend.go
```diff
@@ -32,10 +32,9 @@ type Backend interface {
 	// Stat returns information about the File identified by h.
 	Stat(ctx context.Context, h Handle) (FileInfo, error)
 
-	// List returns a channel that yields all names of files of type t in an
-	// arbitrary order. A goroutine is started for this, which is stopped when
-	// ctx is cancelled.
-	List(ctx context.Context, t FileType) <-chan string
+	// List runs fn for each file in the backend which has the type t. When an
+	// error occurs (or fn returns an error), List stops and returns it.
+	List(ctx context.Context, t FileType, fn func(FileInfo) error) error
```
ifedorenko: I would mention that `fn` is allowed to call Backend. This has implications for how backend implementations throttle concurrent remote calls. 

I would also explicitly say that `fn` will be called exactly once for each `FileInfo` during successful `List()` execution (and at most once in case of an error). This has implications for `BackendRetry` implementation.

ifedorenko (commented):
Looks  good, with couple of minor comments/questions. I'll try to update BackendRetry and my OneDrive backend to use the new API in the next few days.

internal/restic/backend.go
ifedorenko: Hmm, I think we also need to explicitly state that `fn` will be called on the calling gorouting. Otherwise each client will need to deal with synchronization issues if they want to [collect results in a map](https://stackoverflow.com/questions/11063473/map-with-concurrent-access) for example.

ifedorenko (changes requested):

internal/repository/repository.go
fd0: `restic.FileInfo` contains `Name string`, which is fine for the backend: There, the files have names. On the "repository" abstraction level, we're only dealing with `restic.ID`s. I did not see the point in combining the two parameters into a new struct, as I did not come up with a good name for it.

Do you have an idea?

Thinking about it, it may get easier to extend should we ever decide to add a third parameter. hm.

internal/restic/backend.go
fd0: Ah, good points, thanks. The `List()` backend call isn't restricted (in terms of parallel connections) in any existing backend, only the other operations like `Load()` and `Save()`.

> Hmm, I think we also need to explicitly state that fn will be called on the calling gorouting

Why do you feel that way? Sure, we can add that, but In my experience, APIs for libraries rather state if something isn't run in the calling goroutine...

internal/repository/repository.go
fd0: We could also have both, `backend.FileInfo` and `repository.FileInfo`, but that may be too confusing...

ifedorenko: Ah, didn't notice `restic.ID` vs `string` difference. In that case, I think two parameters are a good decision (no sense to worry about hypothetical third parameter until we actually need it, IMO).

internal/restic/backend.go
ifedorenko: If "on the same gorouting" is the normal/expected way of doing things in Go, then indeed no need to mention it explicitly. (sorry, I am new to Go, still learning my ways around).

ifedorenko (approved):

internal/repository/repository.go
fd0: Agreed

internal/restic/backend.go
fd0: We can still add it for clarity, I was just wondering if maybe I got something wrong :)

internal/backend/test/tests.go
```diff
@@ -312,6 +330,122 @@ func (s *Suite) TestList(t *testing.T) {
...
+		ctx, cancel := context.WithCancel(context.TODO())
+		cancel()
+
+		// pass in a cancelled context
+		err := b.List(ctx, restic.DataFile, func(fi restic.FileInfo) error {
+			t.Errorf("got FileInfo %v for cancelled context", fi)
+			return nil
+		})
+
+		if errors.Cause(err) != context.Canceled {
```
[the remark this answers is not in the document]
fd0: Ok, that's something we need to add to our `errors` package, here: https://github.com/restic/restic/blob/a3d43a92b327e9aba596ac4d4d42f6ed54989255/internal/errors/wrap.go#L6

I'll do that later.

internal/backend/test/tests.go
```diff
@@ -312,6 +330,122 @@ func (s *Suite) TestList(t *testing.T) {
...
+		if i != numTestFiles {
+			t.Fatalf("wrong number of files returned by List, want %v, got %v", numTestFiles, i)
+		}
+	})
+
+	t.Run("Timeout", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.TODO())
+		defer cancel()
+
+		timeout := 200 * time.Millisecond
```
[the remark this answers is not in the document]
fd0: Yeah, we need to find a proper solution here and not depend on timing. This test case has bitten me once already (failed on Travis on darwin), so let's do this properly.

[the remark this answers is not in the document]
ifedorenko: We have to assume all remote backend tests will fail regularly. We just need to wrap each test in retry-with-backoff loop and hope the failures are not intermittent bugs in restic. For this specific test I'd increase timeout to 500ms and make the timeout per-backend configurable.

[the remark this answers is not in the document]
fd0: Ah, I was thinking about a more robust approach, like this:

```go
        // rather large timeout
        timeout := time.Second

        ctxTimeout, _ := context.WithTimeout(ctx, timeout)

        i := 0
 		// pass in a context with a small timeout
 		err := b.List(ctxTimeout, restic.DataFile, func(fi restic.FileInfo) error {
 			i++
  
 			// wait until the context is timed out, so the function is not called again
            <-ctxTimeout.Done()
 			return nil
 		})
```

Maybe we can then even accept 0 or 1 calls of the function (so the timeout happens either before the first call of the function, or before the second call).

Thoughts?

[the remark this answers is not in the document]
ifedorenko: > accept 0 or 1 calls of the function

yes, good idea, haven't thought about it.

fd0:
I've added the changes we discussed, including a clarification for the REST protocol.

fd0:
We've had a discussion on the API protocol version selection for the REST API, and settled on using an HTTP request header (`Accept`) and a proper mime type.

This PR is done now, any objections?

PR #1574, opened 2018-01-24
Optimize pack readHeader() implementation

ifedorenko:

Load pack header length and 15 header entries with single backend
request. This eliminates separate header Load() request for most pack
files and significantly improves index.New() performance.

See #1567

### Checklist

### DIFF
--- a/internal/pack/pack.go
+++ b/internal/pack/pack.go
@@ -170,28 +170,13 @@ func (p *Packer) String() string {
 	return fmt.Sprintf("<Packer %d blobs, %d bytes>", len(p.blobs), p.bytes)
 }
 
-// readHeaderLength returns the header length read from the end of the file
-// encoded in little endian.
-func readHeaderLength(rd io.ReaderAt, size int64) (uint32, error) {
-	off := size - int64(binary.Size(uint32(0)))
-
-	buf := make([]byte, binary.Size(uint32(0)))
-	n, err := rd.ReadAt(buf, off)
-	if err != nil {
-		return 0, errors.Wrap(err, "ReadAt")
-	}
-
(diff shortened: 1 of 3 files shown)


internal/pack/pack.go
```diff
@@ -170,29 +170,14 @@ func (p *Packer) String() string {
...
-	return binary.LittleEndian.Uint32(buf), nil
-}
-
 const maxHeaderSize = 16 * 1024 * 1024
 
 // we require at least one entry in the header, and one blob for a pack file
 var minFileSize = entrySize + crypto.Extension
 
+// number of header enries to download as part of header-length request
+var eagerEntries = uint(15)
```
ifedorenko: @fd0 This number is based on stats from single 435GB repository (where >98% of all packs have 15 or less header entries, fwiw). Maybe useful to get stats from other repositories, assuming you have access or have interested users who can provide the info.

internal/pack/pack.go
```diff
@@ -170,29 +170,14 @@ func (p *Packer) String() string {
...
-		return 0, errors.New("not enough bytes read")
-	}
-
-	return binary.LittleEndian.Uint32(buf), nil
-}
-
 const maxHeaderSize = 16 * 1024 * 1024
 
 // we require at least one entry in the header, and one blob for a pack file
 var minFileSize = entrySize + crypto.Extension
```
ifedorenko: Not related to my change, but I believe `minFileSize` should be 4 bytes longer to account for header length record at the end of the file.

fd0: Oh indeed, that's right.

fd0 (approved):
Cool, thanks!

PR #1579, opened 2018-01-25
Implement RetryBackend.List()

ifedorenko:

It improves reliability of backend List() implementation.

Kinda. Mentioned this in #1567 

### Checklist

### DIFF
--- a/internal/backend/backend_retry.go
+++ b/internal/backend/backend_retry.go
@@ -128,3 +128,17 @@ func (be *RetryBackend) Test(ctx context.Context, h restic.Handle) (exists bool,
 	})
 	return exists, err
 }
+
+// List runs fn for each file in the backend which has the type t.
+func (be *RetryBackend) List(ctx context.Context, t restic.FileType, fn func(restic.FileInfo) error) error {
+	listed := make(map[string]struct{})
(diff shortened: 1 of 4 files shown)


fd0:
The idea is awesome, I like it!

internal/backend/backend_retry.go
```diff
@@ -128,3 +128,17 @@ func (be *RetryBackend) Test(ctx context.Context, h restic.Handle) (exists bool,
 	})
 	return exists, err
 }
+
+// List runs fn for each file in the backend which has the type t.
+func (be *RetryBackend) List(ctx context.Context, t restic.FileType, fn func(restic.FileInfo) error) error {
+	listed := make(map[string]bool)
```
fd0: We can improve this: The `bool` is not needed, just use an empty struct:
```go
listed := make(map[string]struct{})
```

and below:
```
listed[fi.Name] = struct{}{}
```

fd0 (changes requested):
I've got an idea for improvement, then we can merge this!

internal/backend/backend_retry.go
ifedorenko: fixed.

fd0 (approved):

PR #1582, opened 2018-01-25
Optimize debug logs

fd0:
The code in here optimizes debug.Log() calls when a restic ID is passed in. Before, the "shorten the string" method Str() was always called, which led to a function call and allocations happening even without the `debug` tag. Now, we're calling the `Str()` method via reflection on demand in the debug code, which is only active with the `debug` tag.

Idea by @MJDSys in #1549

### DIFF
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -182,11 +182,11 @@ type Finder struct {
 
 func (f *Finder) findInTree(ctx context.Context, treeID restic.ID, prefix string) error {
 	if f.notfound.Has(treeID) {
-		debug.Log("%v skipping tree %v, has already been checked", prefix, treeID.Str())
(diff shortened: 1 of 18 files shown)


MJDSys:
This seems like a better implementation then what I had, so +1 from me!

I had two thoughts while I was taking a look over it:
 - would it make sense to keep a benchmark showing the cost of ID.Str versus just passing Id?  Just to document this for the future?
 - Also, should the BenchmarkLogIDString function either keep its call to ID.String, or be removed?  I don't think it is very useful being a duplicate of ID.Str's benchmark.

Otherwise I ran the micro benchmarks locally here, and as expected they run much faster.  Everything else looks good, and a quick grep through the code suggests no tricky logging statements were missed.

fd0:
Yeah, we can improve the benchmarks. I'll have a look later.

Issue #1561, opened 2018-01-17
Using rclone as a backend


fd0:
This issue is there to discuss prototype implementations which enable restic to use rclone as a backend. This makes a lot of sense since rclone already has all the cloud backends implemented (including user configuration and so on). I've had a very productive discussion with @ncw, who is the author of rclone, about how to make restic use rclone to access data, and we've decided to continue our discussion in a public issue here on GitHub.

My idea was to add a new backend to restic which talks to a second process via stdin/stdout, using a protocol that is still to be defined. I read nice things about https://github.com/hashicorp/yamux, which allows mixing several streams in parallel over a single "connection" (very similar to what HTTP2 does). We'd need to define a protocol for accessing files on top of that, with just the basic operations that restic needs.

This backend can then be implemented in rclone, so that restic just starts `rclone serve restic-backend` (or whatever command line we need), talks to rclone, and rclone takes care of saving/loading/removing the files somewhere.

Taking this a step further, we could also implement a server side for this backend in restic itself, which can then be started e.g. via SSH on a remote server, so restic runs `ssh user@host restic serve backend`, and use that connection, which will be **way** more efficient than sftp. If we make the protocol extendable, we could also add features such as remote repacking (if supported by the "server"), so that data repacked during prune does not need to be downloaded and re-uploaded.

The protocol could then be also implemented in other programs, so we can use them as "plugins" to access data stored anywhere.

A slightly different approach would be to speak HTTP2 over stdin/stdout with a process like rclone, or use something like https://github.com/hashicorp/go-plugin and implement the backend in Go as a plugin.

So, this issue is to advanced the discussion further and play with a few sample implementations, so we can get a feel for what works best.

rawtaz:
What's been said about the redundancy that will arise in the existing backends, e.g. S3, Swift, etc?

fd0:
There hasn't been a decision yet. I don't see the point in duplicating all the work and support all the thousands of cloud services, but for the existing backends I also don't see any point in removing them. The existing backends have the nice property of not requiring the setup of a third-party tool, you can get away with restic alone, which is also desirable.

For now this is just an experiment, we'll see how it goes.

ncw:
I made a little experiment in the [serve-restic branch](https://github.com/ncw/rclone/tree/serve-restic) for rclone.  I took the restic/rest-server repo and mashed it into rclone with great force and little subtlety ;-)

Run `rclone serve restic remote:`  where remote: can be any rclone remote, or a local directory.

[...]

PR #1592, opened 2018-01-27
Make backend tests more helpful

ncw:
* In TestList check that directory is empty first
  * Improve error message in TestBackend

Make the backend test suite a bit more helpful.

The change would have helped me greatly trying to track down the extra files as reported in #1561 as I spent ages looking at the wrong test failure!  The extra files were caused by the previous test failing.  I added a comment there too for the next person :-)

### Checklist

### DIFF
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -249,6 +249,19 @@ func (s *Suite) TestList(t *testing.T) {
 	b := s.open(t)
 	defer s.close(t, b)
 
+	// Check that the backend is empty to start with
+	var found []string
+	err := b.List(context.TODO(), restic.DataFile, func(fi restic.FileInfo) error {
+		found = append(found, fi.Name)
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("List returned error %v", err)
+	}
+	if found != nil {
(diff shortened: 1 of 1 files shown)


internal/backend/test/tests.go
```diff
@@ -249,6 +249,19 @@ func (s *Suite) TestList(t *testing.T) {
...
+	var found []string
+	err := b.List(context.TODO(), restic.DataFile, func(fi restic.FileInfo) error {
+		found = append(found, fi.Name)
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("List returned error %v", err)
+	}
+	if found != nil {
+		t.Fatalf("Repo not empty at start of test - contains: %v", found)
```
fd0: `s/Repo/backend`

fd0 (changes requested):
Good idea, thanks!

internal/backend/test/tests.go
ncw: I've fixed that now :-)  Haven't learnt all the restic words yet!

fd0: Thanks!

fd0 (approved):

PR #1611, opened 2018-02-11
On OpenBSD only root can set sticky bit

qbit:

Fix tests on OpenBSD. Currently the sticky bit test in `internal/restic/node_test.go` is broken on OpenBSD. From the man page for [sticky(8)](https://man.openbsd.org/sticky), only root can set the sticky bit. (Thanks @brycied00d for the man page clue stick :D)

No.

### Checklist

### DIFF
--- a/internal/restic/node_test.go
+++ b/internal/restic/node_test.go
@@ -209,8 +209,11 @@ func TestNodeRestoreAt(t *testing.T) {
 			rtest.Assert(t, test.GID == n2.GID,
 				"%v: GID doesn't match (%v != %v)", test.Type, test.GID, n2.GID)
 			if test.Type != "symlink" {
-				rtest.Assert(t, test.Mode == n2.Mode,
-					"%v: mode doesn't match (0%o != 0%o)", test.Type, test.Mode, n2.Mode)
+				// On OpenBSD only root can set sticky bit (see sticky(8)).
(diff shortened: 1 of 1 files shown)


brycied00d:
Would it make more sense for this test to examine whether it's being run as root (on OpenBSD) and perform the test in that case still? Obviously I don't need to preach what a big no-no it is to run things as root, but it still seems reasonable to provide a way for this test to still be run under the correct circumstances that should result in success.

qbit:
I thought about making the test check for root.. but from the standpoint of integrating it into ports (which is my goal - https://github.com/jasperla/openbsd-wip/tree/master/sysutils/restic) .. tests will never run as root. I can re-work it to check permissions if that's preferred.

fd0:
Thanks for giving this a try. We can certainly disable the test for the sticky bit for OpenBSD, but rather than disabling the mode test completely for OpenBSD (which is not a great idea), let's strip the sticky bit for OpenBSD only, in both `test.Mode` and `n2.Mode`.

In addition, the commit message needs to describe what the commit _does_, not just state a fact about OpenBSD.

Let me know if you need help with any of this!

qbit:
Updated to only skip `testSticky`.

fd0 (approved):
Thanks!

PR #1613, opened 2018-02-11
Strip spaces from items read via --files-from

### LINKED ISSUE #1590: backing up files with spaces in them with  --files-from argument
## Output of `restic version`
restic 0.8.1

/usr/local/lib/restic_0.8.1_linux_amd64 backup --password-file=/usr/local/etc/chbackup/private/chbackup-passwd --repo=b2:xxx --files-from=/usr/local/etc/chbackup/dirs.conf --exclude-file=/usr/local/etc/chbackup/exceptions.conf --hostname=xxx --tag=system

B2, but I don't think it matters

## Expected behavior
I would like to be able to back up files and directories with spaces in the name

## Actual behavior
/'a\ b' does not exist, skipping
/a\ b does not exist, skipping
/'a b' does not exist, skipping

Just have a directory with a space in the name in a file referenced by the files-from argument. I tried 

I tried various quoting methods with no joy
I do love restic, but do need to be able to back up such files

fd0:
In #1590, it was mentioned that while lines read from exclude files via `--exclude-file` have leading and trailing spaces stripped, this is not the case for lines read via `--files-from`. This commit fixes that, spaces are always stripped.

Closes #1590

### DIFF
--- a/changelog/0.8.2/issue-1590
+++ b/changelog/0.8.2/issue-1590
@@ -1,0 +1,7 @@
+Bugfix: Strip spaces for lines read via --files-from
+
+Leading and trailing spaces in lines read via `--files-from` are now stripped,
+so it behaves the same as with lines read via `--exclude-file`.
+
+https://github.com/restic/restic/issues/1590
+https://github.com/restic/restic/pull/1613
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -304,7 +304,7 @@ func readLinesFromFile(filename string) ([]string, error) {
 
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
-		line := scanner.Text()
+		line := strings.TrimSpace(scanner.Text())
 		// ignore empty lines
 		if line == "" {
 			continue


pvgoran:
What if the file name actually has spaces at the beginning or at the end? As far as I remember, I encountered such file names. This is obviously a bad naming choice, but still they existed.

If leading/trailing spaces are stripped from file names, this will break certain ways of backup automation in presence of such files.

fd0:
Good point indeed, this was next on my list. I'd like to evaluate if we need to consider this case. For now, the behavior is unified between `--exclude-file` and `--files-from`, and I think that's a good idea. If it is an issue, so some users need to match files containing leading or trailing spaces, we need to find a solution for both cases. After all, spaces at the end of a file name in `--exclude-file` can be solved by using globbing (`file with spaces*`), and instead of `--files-from`, the files in question can be passed as arguments (`restic backup "file with spaces   "`).

You said that you remembered that you encountered a case where this would have been relevant. Would you mind looking it up again if this cannot be solved with the current code?

pvgoran:
> You said that you remembered that you encountered a case where this would have been relevant. Would you mind looking it up again if this cannot be solved with the current code?

I confirmed that the files themselves exist (there are at least a dozen of them). It's not that I need to mention them in `--files-from` or `--exclude-file`, so there is no immediate problem I need to solve here. These files are just a proof of concept for spaces at the beginning of name.

Sure, files like this can be passed as parameters, if restic is invoked manually. However, I'm talking about automation here. If the files passed to `--exclude-file` or `--files-from` are generated by a script, then entries with such abnormal names will be misinterpreted and ignored. Or worse, a different file will be included or excluded excluded; in my case, the same directory contains the files "1.flv" and " 1.flv", so an entry " 1.flv" would be attributed to "1.flv" after stripping spaces.

At the end of the day, stripping spaces makes these two options unsuitable for reliable automated usage.

fd0:
Thanks for the confirmation.

I see several possible ways forward, none if which is really satisfactory:
 * Not strip any spaces in both files (this will lead to usability problems when people edit the files by and and leave trailing white spaces).
 * Always strip any spaces (current behavior, no possibility to save a file with trailing whitespace when read via `--files-from`), depend on the users to build a workaround (no good usability).
 * Strip any spaces, but allow quoting lines in files. When the line starts with single or double quote, treat it as a quoted string somehow.

Any other ideas?

pvgoran:
Maybe have two kinds of options, like, `--files-from` and `--files-from-raw`? `--files-from-raw` would be intended for automated invocations, and preserve all spaces. (Not sure about the name, though.)

Although personally I don't see much problem in the option 1 "when people edit the files by and and leave trailing white spaces".

rawtaz:
I think it's really straight forward from a design perspective:
- If the line in an exclude/include file starts with a quote, treat it as quoted and use the text in the quote as filename (even if it starts or ends with whitespace).
- Otherwise, strip whitespace from beginning and end of the line, and use the result of that as the filename.
- Optionally, do not even allow whitespace in the middle of filenames - if there needs to be spaces, the entire filename/line should be quoted.

That's how it should work, and it's quite conventional to require quotes when there's whitespace in a filename, so why shouldn't we do that here? How to implement it is a different matter though.

pvgoran:
When quotes become a special character, this brings the need for escapes to include quotes as part of filenames. As a result, we'll have a specialized language for list files. This means noticeable complexity both for the software (which needs to parse the language and handle errors) and for users (which need to perform processing on the filenames before including them in the list files).

> it's quite conventional to require quotes when there's whitespace in a filename, so why shouldn't we do that here?
I'd rephrase it differently: it's conventional for command interpreters (shells) to require quotes when there is a whitespace in a parameter. This is needed because whitespaces are used to separate parameters in shell command lines. In case of list files for restic, this is not the case.

Moreover, I'd say that newline-separated text (without any additional language inside lines) is a widely used standard way of representing filename lists (or other lists) in the *nix world. `xargs` accepts this kind of text, `find` generates this kind of text, and so on.

cfcs:
My experience is that trailing whitespace, while it does happen as a result of failed quoting (especially files that end in a single newline or single space) from time to time, is extremely rare.

Leading spaces happen more often for me, for some reason, at least for me. Main culprit is `youtube-dl`.

I agree with @pvgoran's observations that introducing a syntax for the excludes/includes-from is going to be yucky.

Luckily we're not the first program to have this problem, and there's a pretty elegant and well-established solution, which is to use `0x00` as a separator instead of `0x0A` or `$IFS`. The nullbyte is the only character that is not allowed in the Linux world, I believe Windows has a whole set of characters, including the nullbyte.
This will make it easy to generate lists of files safely using `grep --null`, `find -print0` etc.

Thus, I propose keeping the human-friendly behavior of the current functions and introducing `0`-variants when the need arises to automate generation of either list.

rawtaz:
I did a quick search for files ending with `[^a-zA-Z0-9._()]` in a share that has about 81k files. It turns out that of these, 61 are files which has non-regular characters in the end of their filename.

If I were to disregard file suffixes, this number would be way way bigger, because the users of this share have a real habit of ending their filenames with invisible characters. But since they only edit the part before the file suffix, that saves the day.

I'm not sure when one has a need to exclude such specific filenames ending in spaces anyway. If there are files/folders with that name, they should be corrected IMO. But regardless, you do know about them. So it's really just a problem in the editing of the exclude/include file itself, e.g. when people accidentally enter a space on a line in them.

Restic *could* use the approach settled on by @fd0 but also warn if it fines any lines in the file that starts or ends with whitespace. That'd probably improve it somewhat. Other than that, I'm fine with that decision (to strip whitespace and not support quoting).

cfcs:
@rawtaz I think issuing warnings where a strip is being done would make sense; I would expect it to be a mistake 95% of the time.

PR #1615, opened 2018-02-13
Fix #1519 and formatting in `forget` doc

### LINKED ISSUE #1519: Document handling empty forget policy
After noticing that some files have been backed up that I don't want to be, I decided to forget all snapshots for a certain tag, but Restic won't let me:
```
$ restic -r /mnt/restic -p /mnt/restic/password forget --dry-run --tag main --keep-last 0 --group-by tags
password is correct
no policy was specified, no snapshots will be removed
```
If I use `--keep-last 1`, it offers to remove all but the latest snapshot.

This might actually be a good safety measure to prevent removing all snapshots in case of a typo, so maybe instead of changing this behavior, there should be an option like `--keep-none` that prompts the user to type `YES` as confirmation.

Thanks for your work on restic!

rawtaz:

Fixes #1519 by documenting the fact that restic refuses to `forget` when given an "empty" policy. Also fixes some small formatting errors in the same `forget` doc.

Yes, #1519.

### Checklist

- [x] I have skimmed through the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [x] I like Gophers

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -194,7 +194,7 @@ repository, restic will respond that no snapshots will be removed. To delete
 all snapshots, use ``--keep-last 1`` and then finally remove the last
 snapshot ID manually (by passing the ID to ``forget``).
 
-All snapshots are evaluated counted against all matching keep-* counts. A
+All snapshots are evaluated against all matching ``--keep-*`` counts. A
 single snapshot on 2017-09-30 (Sun) will count as a daily, weekly and monthly.
 
 Let's explain this with an example: Suppose you have only made a backup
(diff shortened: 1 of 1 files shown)


rawtaz:
Not sure why the Go 1.8 checks time out over at Travis-CI, but I'd be extremely surprised if this has anything to do with the changes I made. Can probably ignore that in the context of this PR.

fd0:
I've restarted the test. That happens on Travis sometimes, no idea why.

fd0 (approved):
Great! Thank you very much for taking the time to improve the docs!

PR #1623, opened 2018-02-17
backend: Relax requirement for new files

fd0:
Before, all backend implementations were required to return an error if the file that is to be written already exists in the backend. For most backends, that means making a request (e.g. via HTTP) and returning an error when the file already exists.

This is not accurate, the file could have been created between the HTTP request testing for it, and when writing starts. In addition, apart from the `config` file in the repo, all other file names have pseudo-random names with a very very low probability of a collision. And even if a file name is written again, the way the restic repo is structured this just means that the same content is placed there again. Which is not a problem, just not very efficient.

So, this commit relaxes the requirement to return an error when the file in the backend already exists, which allows reducing the number of API requests and thereby the latency for remote backends.

### DIFF
--- a/changelog/0.8.3/pull-1623
+++ b/changelog/0.8.3/pull-1623
@@ -1,0 +1,12 @@
+Enhancement: Don't check for presence of files in the backend before writing
+
+Before, all backend implementations were required to return an error if the
+file that is to be written already exists in the backend. For most backends,
+that means making a request (e.g. via HTTP) and returning an error when the
+file already exists.
+
+This is not accurate, the file could have been created between the HTTP request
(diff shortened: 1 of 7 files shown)


kurin:
Just to confirm, do objects with the same name contain the same contents?  That is, even if an object is overwritten, it just costs time and bandwidth?

ifedorenko (approved):

fd0:
@kurin exactly, when a file with the same name is written, it will also contain the same content (because `filename == sha256(content)`). The only exception is the `config` file, which we handle as a special case.

rawtaz:
Really good change!

PR #1624, opened 2018-02-17
Attempt to make issue template a bit clearer

rawtaz:

Make the issue template a bit more explicit or "clear" in the initial part of it's comment, to encourage more people to fill it out or use the forum.

No.

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,7 +3,7 @@
 Welcome! - We kindly ask that you:
 
   1. Fill out the issue template below - not doing so needs a good reason
-  2. Use the forum if you have a question rather than a bug or feature request.
+  2. Use the forum if you have a question rather than a bug or feature request
 
 The forum is at: https://forum.restic.net
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,13 +1,25 @@
 <!--
-NOTE: Not filling out the issue template needs a good reason, otherwise it may
-take a lot longer to find the problem! Please take the time to help us
-debugging the problem by collecting information, even if it seems irrelevant to
-you. Thanks!
 
-If you have a question, the forum at https://forum.restic.net is a better place.
-Please do not create issues for usage or documentation questions! We're using
-the GitHub issue tracker mainly for tracking bugs and feature requests.
+Welcome! - We kindly ask that you:
+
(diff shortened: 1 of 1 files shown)


fd0 (changes requested):
Very nice, I like the new text. just one minor thing: Would you mind putting the forum URL back in?

rawtaz:
I'm terribly sorry, I removed it by mistake. IT's added back now, in a way such that the list items are still kept short, and the top part of the initial comment is still "light". Do you feel it works like this?

rawtaz:
Feel free to speak your mind on the command and dot issue (third commit). I'm all for changing it if requested.

ifedorenko:
The template works well for bug reports but I find it less useful for feature requests and enhancement proposals. Just my 2c.

rawtaz:
That's a really good point @ifedorenko, I agree. I'll see what I can do to change it a bit in a new PR once this one is merged (unless @fd0 wants me to do it in this one - I just feel it's two separate things to clarify the template and to make it more adjusted to feature requests).

fd0 (approved):

fd0:
This issue template is for bug reports, feature requests aren't handled at all. We can create a second template for feature requests and add a link in the README or so which loads the alternative template, see https://github.com/blog/2495-multiple-issue-and-pull-request-templates

But this is a different issue :)

rawtaz:
Hmm, but the issue template has "This issue tracker is mainly for tracking *bugs and feature requests* ..." in it, and also there's an actual feature request every now and then for which there is an issue created to track it (often by you @fd0 :).

I don't think having a link to different "modes" of opening an issue works, as the absolute majority of people creating issues does so through the "New Issue" button. A link would only be used very rarely.

So, do you want to remove the "feature" stuff from the template? Or should it still be there, and we allow and use issues for feature requests too?

IMO feature requests are fine to have in issues as long as they are to the point - it's a neat place to track them in, and they are directly related to the development of the software (as opposed to questions etc).

fd0:
> So, do you want to remove the "feature" stuff from the template? Or should it still be there, and we allow and use issues for feature requests too?

Let's leave it in and see how people use it in the next couple of weeks. Overall, I think reporting feature requests as issues is fine, and in this case it's not strictly necessary to fill out the template. Most people get that :)

The main purpose of the issue template is to aid in debugging be requiring people to submit all possibly relevant data beforehand, and (since we have the forum now) reduce the number of issues which are really just questions.

Thanks again!

PR #1636, opened 2018-02-22
Refactor the eager-header reads for readability.

kurin:
This pulls the header reads into a function that works in terms of the
number of records requested.  This preserves the existing logic of
initially reading 15 records and then falling back if that fails.

In the event of a header with more than 15 records, it will read all
records, including the already-seen final 15 records.

This is one way of refactoring the eager-header read logic.  It pulls all of the tricky byte manipulation into one function.  In rare circumstances it rereads 15 records, but that's ~hundreds of bytes.

IRC

### Checklist

### DIFF
--- a/internal/pack/pack.go
+++ b/internal/pack/pack.go
@@ -170,26 +170,74 @@ func (p *Packer) String() string {
 	return fmt.Sprintf("<Packer %d blobs, %d bytes>", len(p.blobs), p.bytes)
 }
 
-const maxHeaderSize = 16 * 1024 * 1024
+var (
+	// size of the header-length field at the end of the file
+	headerLengthSize = binary.Size(uint32(0))
+	// we require at least one entry in the header, and one blob for a pack file
+	minFileSize = entrySize + crypto.Extension + uint(headerLengthSize)
+)
 
-// size of the header-length field at the end of the file
-var headerLengthSize = binary.Size(uint32(0))
+const (
+	maxHeaderSize = 16 * 1024 * 1024
+	// number of header enries to download as part of header-length request
+	eagerEntries = 15
+)
 
-// we require at least one entry in the header, and one blob for a pack file
-var minFileSize = entrySize + crypto.Extension + uint(headerLengthSize)
+// readRecords reads count records from the underlying ReaderAt, returning the
+// raw header, the total number of records in the header, and any error.  If
+// the header contains fewer than count entries, the return value is truncated.
+func readRecords(rd io.ReaderAt, size int64, count int) ([]byte, int, error) {
+	var bufsize int
+	bufsize += count * int(entrySize)
+	bufsize += crypto.Extension
+	bufsize += headerLengthSize
 
-// number of header enries to download as part of header-length request
-var eagerEntries = uint(15)
+	if bufsize > int(size) {
+		bufsize = int(size)
+	}
+
+	b := make([]byte, bufsize)
+	off := size - int64(bufsize)
+	if _, err := rd.ReadAt(b, off); err != nil {
+		return nil, 0, err
+	}
+
+	header := b[len(b)-headerLengthSize:]
+	b = b[:len(b)-headerLengthSize]
+	hl := binary.LittleEndian.Uint32(header)
+	debug.Log("header length: %v", hl)
+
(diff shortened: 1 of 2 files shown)


internal/pack/pack.go
```diff
@@ -170,26 +170,74 @@ func (p *Packer) String() string {
...
 
-// size of the header-length field at the end of the file
-var headerLengthSize = binary.Size(uint32(0))
+	b := make([]byte, bufsize)
+	off := size - int64(bufsize)
+	if _, err := rd.ReadAt(b, off); err != nil {
+		return nil, 0, err
+	}
+
+	header := b[len(b)-headerLengthSize:]
```
ifedorenko: I think `header` is misleading here, it's "header-length field at the end of the file". maybe call it `headerLengthEncoded`?

internal/pack/pack.go
```diff
@@ -170,26 +170,74 @@ func (p *Packer) String() string {
...
+	case int64(hl) > maxHeaderSize:
+		err = InvalidFileError{Message: "header is larger than maxHeaderSize"}
+	}
+	if err != nil {
+		return nil, 0, errors.Wrap(err, "readHeader")
+	}
 
-// we require at least one entry in the header, and one blob for a pack file
-var minFileSize = entrySize + crypto.Extension + uint(headerLengthSize)
+	c := (int(hl) - crypto.Extension) / int(entrySize)
```
ifedorenko: maybe rename `c` to `entryCount`?

internal/pack/pack.go
```diff
@@ -170,26 +170,74 @@ func (p *Packer) String() string {
...
+	if c < count {
+		recordSize := c * int(entrySize)
+		start := len(b) - (recordSize + crypto.Extension)
```
ifedorenko: I find "recordSize" too close to "entrySize". Maybe get rid of it altogether?

```go
start := len(b) - (entryCount*int(entrySize) + crypto.Extension)
```

... although the code really truncates the buffer to match actual header length, so maybe just do that directly?

```go
// truncate buffer b to match actual headerLength
start := len(b) - headerLength
b = b[start:]
```

or even
```go
// truncate buffer b to match actual headerLength
b = b[len(b) - headerLength:]
```

ifedorenko (approved):

internal/pack/pack.go
```diff
@@ -170,26 +170,74 @@ func (p *Packer) String() string {
...
-var headerLengthSize = binary.Size(uint32(0))
+	b := make([]byte, bufsize)
+	off := size - int64(bufsize)
+	if _, err := rd.ReadAt(b, off); err != nil {
+		return nil, 0, err
+	}
+
+	header := b[len(b)-headerLengthSize:]
+	b = b[:len(b)-headerLengthSize]
+	hl := binary.LittleEndian.Uint32(header)
```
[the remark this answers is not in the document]
kurin: Per https://github.com/golang/go/wiki/CodeReviewComments#variable-names I'd prefer to keep scope-limited variables with shorter names.  I'll review the function and see if I can get better semantic names though.

internal/pack/pack.go
kurin: Done.

internal/pack/pack.go
kurin: I think `c` is appropriate, since it only lives for like 5 lines.

kurin: Maybe `total` to capture the meaning of that field.

internal/pack/pack.go
```diff
@@ -170,26 +170,73 @@ func (p *Packer) String() string {
...
+// If the header contains fewer than max entries, the header is truncated to
+// the appropriate size.
+func readRecords(rd io.ReaderAt, size int64, max int) ([]byte, int, error) {
+	var bufsize int
+	bufsize += max * int(entrySize)
+	bufsize += crypto.Extension
+	bufsize += headerLengthSize
+
+	if bufsize > int(size) {
+		bufsize = int(size)
```
kurin: We could also just do

`return readRecords(rd, size, max-1)`

which passes existing tests.

fd0 (approved):
Thank you very much!

PR #1638, opened 2018-02-24
backend/retry: return worker function error and abort

fd0:
This is a bug fix: Before, when the worker function fn in List() of the RetryBackend returned an error, the operation is retried with the next file. This is not consistent with the documentation, the intention was that when fn returns an error, this is passed on to the caller and the List() operation is aborted. Only errors happening on the underlying backend are retried. 

The error leads to restic ignoring exclusive locks that are present in the repo, so it may happen that a new backup is written which references data that is going to be removed by a concurrently running `prune` operation.

The bug was reported by a user here: https://forum.restic.net/t/restic-backup-returns-0-exit-code-when-already-locked/484

CC @ifedorenko

### DIFF
--- a/changelog/0.8.3/pull-1638
+++ b/changelog/0.8.3/pull-1638
@@ -1,0 +1,16 @@
+Bugfix: Handle errors listing files in the backend
+
+A user reported in the forum that restic completes a backup although a
+concurrent `prune` operation was running. A few error messages were printed,
+but the backup was attempted and completed successfully. No error code was
+returned.
+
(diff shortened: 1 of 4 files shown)


fd0:
I'm merging this now to get the fix into master as soon as possible, but I'm still very interested in your feedback @ifedorenko, please just review the code when you have time. Thanks!

ifedorenko:
+1

fd0:
Cool, thanks!

PR #1639, opened 2018-02-24
backend: Improve/Cleanup Save()

fd0:
As mentioned in issue [#1560](https://github.com/restic/restic/pull/1560#issuecomment-364689346) this changes the signature for `backend.Save()`. It now takes a parameter of interface type `RewindReader`, so that the backend implementations or our `RetryBackend` middleware can reset the reader to the beginning and then retry an upload operation. 

The `RewindReader` interface also provides a `Length()` method, which is used in the backend to get the size of the data to be saved. This removes several ugly hacks we had to do to pull the size back out of the `io.Reader` passed to `Save()` before. In the `s3` and `rest` backend this is actively used.

New signature:
```go
// Save stores the data from rd under the given handle.
Save(ctx context.Context, h Handle, rd RewindReader) error
```

Ugly hacks removed:
 * https://github.com/restic/restic/blob/2866f3f31ce98b235010dc6c095e67fc915a83f7/internal/backend/backend_retry.go#L51-L63
 * https://github.com/restic/restic/blob/2866f3f31ce98b235010dc6c095e67fc915a83f7/internal/backend/s3/s3.go#L244-L253

### DIFF
--- a/internal/backend/backend_retry.go
+++ b/internal/backend/backend_retry.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cenkalti/backoff"
 	"github.com/restic/restic/internal/debug"
-	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 )
 
@@ -47,23 +46,9 @@ func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) e
 }
 
 // Save stores the data in the backend under the given handle.
-func (be *RetryBackend) Save(ctx context.Context, h restic.Handle, rd io.Reader) error {
-	seeker, ok := rd.(io.Seeker)
-	if !ok {
-		return errors.Errorf("reader %T is not a seeker", rd)
-	}
-
-	pos, err := seeker.Seek(0, io.SeekCurrent)
-	if err != nil {
-		return errors.Wrap(err, "Seek")
-	}
-
-	if pos != 0 {
-		return errors.Errorf("reader is not at the beginning (pos %v)", pos)
-	}
-
+func (be *RetryBackend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
 	return be.retry(ctx, fmt.Sprintf("Save(%v)", h), func() error {
-		_, err := seeker.Seek(0, io.SeekStart)
+		err := rd.Rewind()
 		if err != nil {
 			return err
 		}
--- a/internal/backend/rest/rest.go
+++ b/internal/backend/rest/rest.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 
 	"golang.org/x/net/context/ctxhttp"
@@ -105,7 +106,7 @@ func (b *restBackend) Location() string {
 }
 
 // Save stores data in the backend at the handle.
-func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd io.Reader) (err error) {
+func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
 	if err := h.Valid(); err != nil {
 		return err
 	}
@@ -114,12 +115,11 @@ func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd io.Reader) (
 	defer cancel()
 
 	// make sure that client.Post() cannot close the reader by wrapping it
-	rd = ioutil.NopCloser(rd)
-
-	req, err := http.NewRequest(http.MethodPost, b.Filename(h), rd)
+	req, err := http.NewRequest(http.MethodPost, b.Filename(h), ioutil.NopCloser(rd))
 	if err != nil {
 		return errors.Wrap(err, "NewRequest")
 	}
+	req.Header.Set("Content-Length", strconv.Itoa(rd.Length()))
 	req.Header.Set("Content-Type", "application/octet-stream")
 	req.Header.Set("Accept", contentTypeV2)
 
--- a/internal/backend/rest/rest.go
+++ b/internal/backend/rest/rest.go
@@ -119,7 +119,7 @@ func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd restic.Rewin
 	if err != nil {
 		return errors.Wrap(err, "NewRequest")
 	}
-	req.Header.Set("Content-Length", strconv.Itoa(rd.Length()))
+	req.Header.Set("Content-Length", strconv.FormatInt(rd.Length(), 10))
 	req.Header.Set("Content-Type", "application/octet-stream")
 	req.Header.Set("Accept", contentTypeV2)
 
--- a/internal/restic/backend.go
+++ b/internal/restic/backend.go
@@ -20,8 +20,8 @@ type Backend interface {
 	// Close the backend
 	Close() error
 
-	// Save stores the data in the backend under the given handle.
-	Save(ctx context.Context, h Handle, rd io.Reader) error
+	// Save stores the data from rd under the given handle.
+	Save(ctx context.Context, h Handle, rd RewindReader) error
 
 	// Load runs fn with a reader that yields the contents of the file at h at the
 	// given offset. If length is larger than zero, only a portion of the file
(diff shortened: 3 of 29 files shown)


internal/backend/backend_retry.go
```diff
@@ -47,28 +46,9 @@ func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) e
...
-
+func (be *RetryBackend) Save(ctx context.Context, h restic.Handle, length int, fn func() (io.Reader, error)) error {
 	return be.retry(ctx, fmt.Sprintf("Save(%v)", h), func() error {
-		_, err := seeker.Seek(0, io.SeekStart)
-		if err != nil {
-			return err
-		}
-
-		err = be.Backend.Save(ctx, h, rd)
+		err := be.Backend.Save(ctx, h, length, fn)
```
ifedorenko: shouldn't the old reader be closed here on retry?

internal/restic/backend.go
```diff
@@ -20,8 +20,11 @@ type Backend interface {
 	// Close the backend
 	Close() error
 
-	// Save stores the data in the backend under the given handle.
-	Save(ctx context.Context, h Handle, rd io.Reader) error
+	// Save stores the data returned by the reader returned from fn in the
+	// backend under the given handle. The function fn may be called several
+	// times to retry the operation in case an error occurs. When fn returns an
+	// error, the process is aborted and the error is returned to the caller.
+	Save(ctx context.Context, h Handle, length int, fn func() (io.Reader, error)) error
```
ifedorenko: The proposed API does not look right, to be honest. If `fn` provides the content to save, it should provide both the content reader and the content length (and the error).

I also think adding `length` parameter is orthogonal to `fn` callback.

ifedorenko (commented):
I am not sure... 

I think adding `length` parameter is useful and I am surprised more backends don't need it (http PUT/POST is usually much more efficient when content-length is known).

As for `fn` callback, I am not convinced. Personally I think it trades `seeker` ugliness for `Closer` ugliness during retry, but also adds some noise in each backend implementation. But this is mostly matter of personal preferences, at least if you agree `fn` should return both the reader and the length.

internal/backend/backend_retry.go
fd0: Which old reader?

internal/restic/backend.go
fd0: Hm. But for the backend, we **always** know beforehand how much data we're going to save, and that's not going to change on the third try. `fn` is just a way to "reset" the reader, so that is left out of the backend code.

fd0 (commented):
The main idea I had was to give the backend code a way to "reset" the reader. The length is a parameter of `Save` and not returned by `fn` because it will always be constant: For all data we save in the repo we always know the exact size, which is not changing, even on the third try.

I also thought about this interface, it feels a bit awkward though:
```go
Save(ctx context.Context, h Handle, length int, rd io.Reader, rewind func() error) error
```

I don't see any of the referenced "Closer ugliness", can you show it to me?

fd0:
Ah, the other interface (with the `rewind` method) will not work with the limiters, which needs to be wrapped around the reader again on the next try. Hm.

internal/backend/backend_retry.go
ifedorenko: The API does not specify if `fn` should return a new reader or reset the same reader on retry. We either need to specify, and ideally enforce, that `fn` must always return the same reader, or assume it'll return new reader, in which case something will need to close the old one during retry.

internal/restic/backend.go
ifedorenko: I am nitpicking, but `length` parameter and the reader returned by `fn` represent related features of the content being saved. I think the API will be easier to understand and use correctly if both of these features were accessible the same way.

Personally, I'd introduce `ContentReader` interface with `Length()`, `Read(...)` and `Reset()` methods and define save as:

```go
Save(ctx context.Context, h Handle, content ContentReader) error
```

I think this will remove all the ugliness you want to remove, without the extra noise required to use`fn` parameter.

ifedorenko:
Why can't limiter implement `rewind` (aka `reset`) method to simply pass through the call to the wrapped content reader? I haven't thought much about limiters to be honest, but assumed they just throttle traffic per unit of time and don't care whether traffic is a result of retry or not.

internal/restic/backend.go
fd0: Uh, that is a nice idea! Thank you very much! I'll try it later

fd0:
You may be right, it was late yesterday evening. I'll have a look :)

I like the plain and obvious (at least in retrospect) idea to just wrap it all in an interface, e.g.:

```go
type RewindReader interface {
    io.Reader

    // Rewind sets the Reader back to the start of the
    // data, so all data can be read again.
    Rewind() error
    
    // Length returns the complete number of bytes read
    // from the Reader after Rewind() was called.
    Length() int
}
```

I'll try that later. Thanks for the feedback already!

fd0:
I'm done, please take another look! Apart from the new implementations of `RewindReader` for a byte slice and a file (`restic.ByteReader{}` and `restic.FileReader{}`), the code became even simpler. I like it :)

internal/backend/rest/rest.go
```diff
@@ -114,12 +115,11 @@ func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd io.Reader) (
 	defer cancel()
 
 	// make sure that client.Post() cannot close the reader by wrapping it
-	rd = ioutil.NopCloser(rd)
-
-	req, err := http.NewRequest(http.MethodPost, b.Filename(h), rd)
+	req, err := http.NewRequest(http.MethodPost, b.Filename(h), ioutil.NopCloser(rd))
```
ifedorenko: Is `ioutil.NopCloser` wrapper still necessary? `RewindReader` does not implement `Closer` interface as far as I can tell, so client shouldn't be able to close the underlying file already. (same question applies to azure and s3).

internal/restic/rewind_reader.go
```diff
@@ -0,0 +1,90 @@
...
+type RewindReader interface {
+	io.Reader
+
+	// Rewind rewinds the reader so the same data can be read again from the
+	// start.
+	Rewind() error
+
+	// Length returns the number of bytes that can be read from the Reader
+	// after calling Rewind.
+	Length() int
```
ifedorenko: shouldn't this be int64? I don't think it's reasonable to assume all uploaded files will be smaller than 2G.

internal/restic/rewind_reader.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Length returns the length of the file.
+func (f *FileReader) Length() int {
+	return f.Len
+}
+
+// NewFileReader wraps f in a *FileReader.
+func NewFileReader(f io.ReadSeeker) (*FileReader, error) {
+	pos, err := f.Seek(0, io.SeekEnd)
```
ifedorenko: Out of curiosity, why use `file.Seek` and not `file.Stat` to find file length?

ifedorenko (approved):
Looks good. I did leave couple of minor comments/questions, but ignore them if they don't make sense.

internal/backend/rest/rest.go
fd0: Good question! I think it isn't strictly necessary, for the reason you mentioned (`RewindReader` does not imply that there's a `Close()` method, but there's no reason there couldn't be a `Close()`). But if we, at some point, add some implementation of `RewindReader` which also has a `Close` method, retries will suddenly stop working mysteriously because the Go stdlib HTTP client will close the body reader for e.g. `POST` requests. That's why the `NopCloser` is there.

So, although it's not strictly necessary, I'd like to keep it in to prevent future bugs.

internal/restic/rewind_reader.go
fd0: Good point, I'll change it to `int64`.

internal/restic/rewind_reader.go
fd0: The interface is smaller. We need `Seek` anyway, and it provides a way to find the length of the file, so we just use that.

ifedorenko (approved):
Thank you for the explanation.

fd0:
Merged, thanks for the feedback!

internal/restic/rewind_reader.go
fd0: It'd btw also work to pass in a `bytes.Reader`, which has `Seek` but not `Stat`.

PR #1647, opened 2018-03-01
Change priority of AWS credential providers to accept AWS_SESSION_TOKEN

duzvik:
It is currently not possible to use the restic S3 backend with IAM credentials that are temporary and scoped to a subdirectory. The first one requires restic to accept AWS_SESSION_TOKEN credentials.
The current priority of AWS credential providers ignore AWS_SESSION_TOKEN env variable.

This pull request aims to solve the issue as described here #1477
https://github.com/restic/restic/pull/1479

### Checklist

### DIFF
--- a/changelog/0.8.3_2018-02-26/pull-1647
+++ b/changelog/0.8.3_2018-02-26/pull-1647
@@ -1,0 +1,8 @@
+Enhancement: Change priority of AWS credential providers to accept AWS_SESSION_TOKEN
+
+Before, it was not possible to use s3 backend with AWS temporary security credentials(with AWS_SESSION_TOKEN).
+This change gives higher priority to credentials.EnvAWS credentials provider.
+
+https://github.com/restic/restic/issues/1477
+https://github.com/restic/restic/pull/1479
+https://github.com/restic/restic/pull/1647
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -48,6 +48,7 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 	// AWS env variables such as AWS_ACCESS_KEY_ID
 	// Minio env variables such as MINIO_ACCESS_KEY
 	creds := credentials.NewChainCredentials([]credentials.Provider{
+		&credentials.EnvAWS{},
 		&credentials.Static{
 			Value: credentials.Value{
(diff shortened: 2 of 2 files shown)


duzvik:
According to the discussion at https://github.com/restic/restic/pull/1479  there is no need for a separate test.

fd0:
hey, thanks for giving this a try. I'm pretty sure it is already supported in the current code, we're calling `credentials.EnvAWS` in `minio-go`, which will try to load a token from the environment. Can you please confirm that it does not work right now? Thanks!

duzvik:
I've tested and you are right, `credentials.EnvAWS` can solve a problem with AWS_SESSION_TOKEN

But with the current priority of available credentials in `credentials.NewChainCredentials`,  there is no chance to pass AWS_SESSION_TOKEN

> // The Chain provides a way of chaining multiple providers together
> // which will pick the first available using priority order of the
> // Providers in the list.

When I set env variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN
The first available providers `credentials.Static` is used(which ignore AWS_SESSION_TOKEN).

What you think about removing `credentials.Static` from `credentials.NewChainCredentials`?

`credentials.Static`   - use  env variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
`credentials.EnvAWS`   - use  env variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN

fd0:
@harshavardhana would you mind taking a look? I'd like your opinion on the issue :)

harshavardhana:
@fd0 @duzvik we could move the env to have higher priority than static and then fallback to I AM when all else fails ?

fd0:
Sounds good to me

fd0:
Can you please squash the commits? And there's a typo in the title of the commit which adds the changelog issue ("changlelog").

fd0:
Otherwise this change looks good. Thanks!

duzvik:
Done

fd0:
Thanks, that was fast!

Just one final question: Did you try it? Does it work?

duzvik:
Yes. I'm confirming. It's work.
Restic successfully accept AWS_SESSION_TOKEN

fd0:
Thanks!

PR #1648, opened 2018-03-02
Ignore s3 AccessDenied error, during creation of repository

duzvik:
On AWS s3 it is more common to share a bucket between users as there is an account limit on the number of buckets (100). In such cases, it makes sense to limit the operations on a certain prefix to a certain user. In such cases s3:HeadBucket will return forbidden and creation fails.

It's not possible to use s3 backend scoped to a subdirectory(with specific permissions).
Restic doesn't try to create repository in a subdirectory, when   'bucket exists' of parent directory check fails due to permission issues.

https://github.com/restic/restic/issues/1477
https://github.com/restic/restic/pull/1479

### Checklist

### DIFF
--- a/changelog/0.8.3_2018-02-26/pull-1648
+++ b/changelog/0.8.3_2018-02-26/pull-1648
@@ -1,0 +1,6 @@
+Enhancement: Ignore AWS permission denied error when creating a repository.
+
+It's not possible to use s3 backend scoped to a subdirectory(with specific permissions).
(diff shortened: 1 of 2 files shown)


fd0 (approved):
LGTM!

fd0:
Is there anything you need to change? you did not check the "I'm ready" box, was that on purpose?

duzvik:
I didn't know what to do with tests/documentation.
I've checked  "I'm ready"  now.

PR #1649, opened 2018-03-04
Minimal set of patches to get restic working on Solaris

### LINKED ISSUE #1576: support for Illumos/Solaris
## Output of `restic version`

I was trying to build latest stable release, 0.8.1.

I didn't run it, I was only trying to build it.

macOS, build was done with reproducible builder via docker image.

## Expected behavior

I would love to see restic building properly on Illumos/Solaris OS.

## Actual behavior

It fails with the following error messages:

```
docker run --rm --volume $PWD:/home/build restic/builder v0.8.1.tar.gz
Go version is go version go1.9.2 linux/amd64
path is /tmp/restic-release-JzlHL4
version is 0.8.1
outputdir is /home/build/restic-0.8.1-20180124-104215
restic_0.8.1_solaris_amd64
# github.com/restic/restic/vendor/bazil.org/fuse
github.com/restic/restic/vendor/bazil.org/fuse/error_std.go:27:20: undefined: errNoXattr
github.com/restic/restic/vendor/bazil.org/fuse/fuse.go:1345:26: undefined: attr
github.com/restic/restic/vendor/bazil.org/fuse/fuse_kernel.go:166:28: undefined: syscall.O_DIRECTORY
github.com/restic/restic/vendor/bazil.org/fuse/fuse_kernel.go:404:17: undefined: attr
# github.com/restic/restic/internal/restic
github.com/restic/restic/internal/restic/node.go:231:14: node.restoreSymlinkTimestamps undefined (type Node has no field or method restoreSymlinkTimestamps)
github.com/restic/restic/internal/restic/node_unix.go:21:18: cannot use statUnix(*s) (type statUnix) as type statT in return argument:
        statUnix does not implement statT (missing atim method)
github.com/restic/restic/internal/restic/node_xattr.go:16:10: undefined: xattr.Get
github.com/restic/restic/internal/restic/node_xattr.go:26:10: undefined: xattr.List
github.com/restic/restic/internal/restic/node_xattr.go:35:7: undefined: xattr.Set
build failed: exit status 2
exit status 1
```

Try to build restic for solaris/amd64 platform via https://github.com/restic/builder

Likely lack of support for Solaris platform.

No, but I would hope having OpenBSD and FreeBSD support in place already, that it won't prove super difficult to include one more. Especially knowing that rest-server does support it.

I honestly think it's the best backup solution out there and I would love to have it running on my SmartOS machines.

jasperla:
This effectively disables the xattr/fuse support. Also the sftp backend is disabled due to lack of `syscall.SYS_IOCTL` support in Go for Solaris.

Allow restic to be built on the golang 'solaris' target. It was built and tested on SmartOS with `go version go1.9.3 solaris/amd64`

Closes #1576 

### Checklist

- [x] ~I have added tests for all changes in this PR~ (not needed?)

### DIFF
--- a/internal/backend/sftp/foreground_solaris.go
+++ b/internal/backend/sftp/foreground_solaris.go
@@ -1,0 +1,25 @@
+package sftp
+
+import (
+	"os/exec"
+	"syscall"
+
+	"github.com/restic/restic/internal/errors"
+)
+
+func startForeground(cmd *exec.Cmd) (bg func() error, err error) {
+	// run the command in it's own process group so that SIGINT
+	// is not sent to it.
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+
+	// start the process
+	err = cmd.Start()
+	if err != nil {
+		return nil, errors.Wrap(err, "cmd.Start")
+	}
+
+	bg = func() error { return nil }
+	return bg, nil
+}
--- a/changelog/0.8.3_2018-02-26/pull-1649
+++ b/changelog/0.8.3_2018-02-26/pull-1649
@@ -1,0 +1,3 @@
+Enhancement: Add illumos/Solaris support.
+
+https://github.com/restic/restic/pull/1649
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -1,4 +1,5 @@
 // +build !openbsd
+// +build !solaris
 // +build !windows
 
 package main
--- a/cmd/restic/integration_fuse_test.go
+++ b/cmd/restic/integration_fuse_test.go
@@ -1,4 +1,5 @@
 // +build !openbsd
+// +build !solaris
 // +build !windows
 
 package main
--- a/doc/050_restore.rst
+++ b/doc/050_restore.rst
@@ -63,7 +63,8 @@ command to serve the repository with FUSE:
     Now serving /tmp/backup at /mnt/restic
     Don't forget to umount after quitting!
 
-Mounting repositories via FUSE is not possible on Windows and OpenBSD.
+Mounting repositories via FUSE is not possible on OpenBSD, Solaris/illumos
+and Windows.
 
 Restic supports storage and preservation of hard links. However, since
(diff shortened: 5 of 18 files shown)


internal/backend/sftp/foreground_solaris.go
```diff
@@ -0,0 +1,15 @@
+// +build !windows
```
fawick: `node_solaris.go` below doesn't have this constraint, why is it needed here?

jasperla: Good point; it's not needed. I've pushed a fixed file.

internal/backend/sftp/foreground_solaris.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"os/exec"
+
+	"github.com/restic/restic/internal/errors"
+)
+
+// golang on Solaris lacks SYS_IOCTL. It should be possible to workaround
+// that, however for turn these functions into no-ops.
+func startForeground(cmd *exec.Cmd) (bg func() error, err error) {
+	return nil, errors.New("sftp disabled on Solaris")
```
fd0: It should be possible to use the sftp backend on solaris just fine, just without the magic that on Linux allows us to query for a password interactively and being able to tear down nicely at the same time.

Something like this should work:
```go
// run the command in its own process group
// so that SIGINT is not sent to it
cmd.SysProcAttr = &syscall.SysProcAttr{
    Setpgid: true,
}

// start the process
err = cmd.Start()
if err != nil {
    return nil, errors.Wrap(err, "cmd.Start")
}

bg = func() error { return nil }
return bg, nil
```

fd0 (changes requested):
Hey, thanks for taking a stab at this!

We need to fix the sftp backend before this can be merged (including the comments in the docs).

What's still missing (apart from sftp) is adding `solaris` to the integration tests, so we see when it breaks:

https://github.com/restic/restic/blob/6edf28d1e187a2f94fbfeef4e91fbe5c893d2f10/run_integration_tests.go#L125

fd0 (approved):

PR #1657, opened 2018-03-07
Rclone backend

### LINKED ISSUE #1561: Using rclone as a backend

fd0:

These commits add a backend based on rclone (or, an unreleased version of rclone). The corresponding PR for rclone is here: https://github.com/ncw/rclone/pull/2116

Basically, restic starts an instance of rclone, talks to it via HTTP2 on stdin/stdout, and then uses the REST protocol.

Closes #1561

### Checklist

- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/changelog-entry.tmpl))

### TODO:
 - [x] Decide how to delete/hide files, and how to communicate this to the user
 - [x] Make number of connections configurable


internal/backend/rclone/backend_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"github.com/restic/restic/internal/backend/rclone"
+	"github.com/restic/restic/internal/backend/test"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/restic"
+	rtest "github.com/restic/restic/internal/test"
+)
+
+const rcloneConfig = `
```
ncw: You could do away with this config if you want - see below

internal/backend/rclone/backend_test.go
```diff
@@ -0,0 +1,90 @@
...
+			t.Logf("use backend at %v", dir)
+
+			repodir := filepath.Join(dir, "repo")
+			err = os.Mkdir(repodir, 0755)
+			if err != nil {
+				return nil, err
+			}
+
+			cfg := rclone.NewConfig()
+			cfg.Program = fmt.Sprintf("rclone --config %q", cfgfile)
```
ncw: You can remove `--config %q` and the saving of the file above and just use

`cfg.Remote = repodir`

rclone remotes are either remote:remote/path or /local/path.  (There is a little bit of ambiguity there I'm sure you'll notice which you can workaround with ./localdirwithcolon:/path)

internal/backend/rclone/config.go
```diff
@@ -0,0 +1,36 @@
...
+	"strings"
+
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/options"
+)
+
+// Config contains all configuration necessary to start rclone.
+type Config struct {
+	Program string `option:"program" help:"path to rclone (default: rclone)"`
+	Args    string `option:"args"    help:"arguments for running rclone (default: restic serve --stdio)"`
```
ncw: That should probably be `(default: serve restic --stdio)`

internal/backend/rclone/backend.go
```diff
@@ -0,0 +1,239 @@
...
+		return nil, err
+	}
+
+	be.Backend = restBackend
+	return be, nil
+}
+
+// Close terminates the backend.
+func (be *Backend) Close() error {
+	debug.Log("exting rclone")
```
ncw: Typo here! "exiting"

internal/backend/rclone/stdio_conn.go
```diff
@@ -0,0 +1,72 @@
+package rclone
```
ncw: It ocurred to me that this is the kind of useful little object that might go in its own repo.  Rclone could then import it too!

ncw (commented):
Please find a few minor notes attached :-)

internal/backend/rclone/backend_test.go
fd0: Ah, nice, I didn't know that.

internal/backend/rclone/config.go
fd0: Yep

internal/backend/rclone/backend.go
fd0: Good catch!

internal/backend/rclone/stdio_conn.go
fd0: That also occurred to me, but I prefer to not couple the source code too much for now. We can always move it out later (ideally, when `vgo` is merged)

fd0:
So, the code is basically done, it just needs some docs. I could use some help testing the backend :)

ncw:
> So, the code is basically done, it just needs some docs. I could use some help testing the backend :)

What sort of tests?

I could run the backend unit tests (using rclone's unit test) against all the test accounts for all the different providers quite easily?  Would that help?

fd0:
> I could run the backend unit tests (using rclone's unit test) against all the test accounts for all the different providers quite easily? Would that help?

That'd be interesting! It may transfer a non-trivial amount of data though...

fd0:
Docs are done, please let me know if there's anything odd!

rawtaz:
I commented on one small thing in https://github.com/restic/restic/commit/2beaad1.

rawtaz:
In addition to the comment I mentioned above, I suggest you also add rclone to the list of backends in the README.md file :)

fd0:
Cool, thanks!

ncw:
I ran the integration tests against all the remotes last night...

The results look good though not perfect.
 
Straight passes for 

  * AzureBlob
  * Box
  * Cache
  * Crypt + Drive
  * Crypt + Swift
  * Drive
  * Dropbox
  * FTP
  * GoogleCloudStorage
  * Pcloud
  * QingStor
  * S3
  * Sftp
  * Swift
  * Yandex

Fails of some kind for

  * B2
  * Hubic
  * OneDrive

I'll dig into these.  Here are the logs in case you are interested: [integration-tests.zip](https://github.com/restic/restic/files/1818604/integration-tests.zip)


#### B2

There were no errors in the rclone log, but this test failed

```
                --- FAIL: TestBackendRESTExternalServer (95.50s)
                    --- FAIL: TestBackendRESTExternalServer/TestBackend (23.50s)
                        tests.go:782: wrong number of IDs returned: want 4, got 3
```

I think that is worth investigating further as it seems consistent

#### Hubic

Hubic failed with service problems or eventual consistency problems.  rclone would deal with this sort of thing with lots of retries...

```
2018/03/15 22:50:54 ERROR : keys/4e54d2c721cbdb730f01b10b62dec622962b36966ec685880effa63d71c808f2: Post request put error: HTTP Error: 503: 503 Service Unavailable
2018/03/15 22:56:44 ERROR : data/2e/2e4967c43b7d3fb9d3f4b61997974240b1a5a6e6136d9756673eb03b89946126: Delete request remove error: Object Not Found
2018/03/15 22:59:26 ERROR : potato/sausage/data/24/248d6a61d20638b8e5c026930c3e6039a33ce45964ff2167f6ecedd419db06c1: Delete request remove error: Object Not Found
2018/03/15 23:04:32 ERROR : potato/sausage/data/1b/1b06f7b567c7f267319bf39f28aa39461537550f0daf6b1c92564fdfb44dda05: Delete request remove error: Object Not Found
2018/03/15 23:06:01 ERROR : keys/c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2: Couldn't delete: Object Not Found
```

#### Onedrive

Failed because of a recent commit to rclone!

ncw:
I had a look through the docs - they look great :-)

You might want to add something like this, which restic users might find easier since the restic configuration is often done with environment variables.

....

Rclone can be [configured with environment variables](https://rclone.org/docs/#environment-variables) also, so for instance if you wanted to set a bandwidth limit for rclone you could `export RCLONE_BWLIMIT=1M` rather than using `-o rclone.args="serve restic --stdio --bwlimit 1M"`

fd0:
Thanks for the comments, I've added the suggested text :)

fd0:
@ncw hm, is it possible that the error for B2 is caused by not calling rclone with `--b2-hard-delete`?

fd0:
Looks like it, running the backend tests for the same directory without `--b2-hard-delete` fails on the second and subsequent runs, with `--b2-hard-delete` it always completes.

ncw:
@fd0 It should make no difference whether `--b2-hard-delete` is called or not - the deleted file versions should be invisible...  I'm struggling trying to work out what is going on.

Apparently the test is indicating that there is an extra `key` returned from the [delayedList function](https://github.com/restic/restic/blob/master/internal/backend/test/tests.go#L671-L691).

Looking at that function I'm wondering if there might be some sort of eventual consistency thing going on as that function seems to collect up as many list entries as it can, rather than just returning the last list which had the desired entry in it.

I wonder whether something like this might fix the problem...

```patch
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -673,9 +673,10 @@ func (s *Suite) delayedRemove(t testing.TB, be restic.Backend, handles ...restic
 }
 
 func delayedList(t testing.TB, b restic.Backend, tpe restic.FileType, max int, maxwait time.Duration) restic.IDs {
-	list := restic.NewIDSet()
+	var list restic.IDSet
 	start := time.Now()
 	for i := 0; i < max; i++ {
+		list = restic.NewIDSet()
 		err := b.List(context.TODO(), tpe, func(fi restic.FileInfo) error {
 			id := restic.TestParseID(fi.Name)
 			list.Insert(id)
@@ -783,7 +784,7 @@ func (s *Suite) TestBackend(t *testing.T) {
 
 		list := delayedList(t, b, tpe, len(IDs), s.WaitForDelayedRemoval)
 		if len(IDs) != len(list) {
-			t.Fatalf("wrong number of IDs returned: want %d, got %d", len(IDs), len(list))
+			t.Fatalf("wrong number of IDs returned: want %d, got %d\n  want:\n  %v\n  got:\n%v\n", len(IDs), len(list), IDs, list)
 		}
 
 		sort.Sort(IDs)
```

I tried it - it didn't!

Any more ideas?

fd0:
@ncw I've managed to reproduce it while writing debug logs for restic and rclone, see here: 

[rclone-log.txt](https://github.com/restic/restic/files/1823090/rclone-log.txt)
[restic-log.txt](https://github.com/restic/restic/files/1823091/restic-log.txt)

Restic was started like this:
```
$ cd internal/backend/rest
$ RESTIC_TEST_REST_REPOSITORY=rest:http://localhost:8080/ DEBUG_FILES=*.go go test -tags debug -v -run External/TestBackend
```

and rclone:
```
rclone --verbose=2 --dump-bodies --dump-headers serve restic b2:restic-test-an/test-rest-ext/ 
```

Maybe you can see something. What I discovered so far was: restic creates and deletes the file `c3ab8ff1` a few times, among others in the `data/` directory. At the end it tries to list all files in `data/`, and did not find the file. It was found before and for other file types.

Any idea what's going on?

fd0:
So, what's still to do before this can be merged?

From my point of view, we need find a solution for the deletion issue on B2. I'd like to avoid users ending up with plenty of invisible files stored at B2 that they're not aware of. When restic deletes files via rclone, are they eventually deleted on B2? Is there some delayed remove for hidden files? Or do users need to manually run `rclone cleanup`?

internal/backend/rclone/backend.go
```diff
@@ -0,0 +1,278 @@
...
+
+	debug.Log("new backend created")
+
+	url, err := url.Parse("http://localhost/")
+	if err != nil {
+		return nil, err
+	}
+
+	restConfig := rest.Config{
+		Connections: 20,
```
Speedy37: Could you make Connections a backend option ? I'm testing rclone with a limit of 3 ftp connections, and this caused me troubles

internal/backend/rclone/backend.go
```diff
@@ -0,0 +1,278 @@
...
+	// then add the arguments
+	if cfg.Args != "" {
+		a, err := backend.SplitShellStrings(cfg.Args)
+		if err != nil {
+			return nil, err
+		}
+
+		args = append(args, a...)
+	} else {
+		args = append(args, "serve", "restic", "--stdio")
```
Speedy37: It would be nice to have an higher level option that add arguments to rclone without having to write "serve restic --stdio" everytime

Speedy37 (commented):
After testing this pull request with docker (https://hub.docker.com/r/speedy37/docker-restic-rclone/) on OVH ftp backup system.

The current docker image contains a hardcoded patch that limit rest connections to 1.

Otherwise, things are working nicely.

ncw:
Sorry, been a bit distracted with the rclone release!  `serve restic` is now in rclone 1.40 :-)

I'd like to get to the bottom of the b2 problem too!

The tests should pass and the fact they don't is probably an rclone (or b2) bug.

I've made a new issue to track the problem https://github.com/ncw/rclone/issues/2164

@fd0 
> From my point of view, we need find a solution for the deletion issue on B2. I'd like to avoid users ending up with plenty of invisible files stored at B2 that they're not aware of. When restic deletes files via rclone, are they eventually deleted on B2? Is there some delayed remove for hidden files? Or do users need to manually run rclone cleanup?

The default for b2 is when you delete a file you "hide" it.  Then files can be retrieved later as versions.  The google drive backend will put things into the trash when you delete them.

Is this behaviour undesirable for restic?  For rclone, this is definitely the correct choice as users hate deleting all their data by accident as I've discovered of the course of a few painful issues!

You'd disable that with `--b2-hard-delete` and `--drive-use-trash=false`

What I was thinking of doing is making a `--hard-delete` flag for rclone which disabled that feature in the backends if available: https://github.com/ncw/rclone/issues/2165

The `serve restic` command could enable hard deletes for everything easily enough.

internal/backend/rclone/backend.go
ncw: Note that you can [add options to rclone as env vars](https://rclone.org/docs/#environment-variables) which you might find easier.

Speedy37: I just figured this out after writing the review

ncw:
In https://github.com/ncw/rclone/issues/2164 I think I've worked out that this is a B2 bug - uploading the same file and deleting it in quick sucession can cause B2 to get confused about the existence of the file.

Now I think this is unlikely to happen in practice when using restic so I think this is a problem with the tests only.

I'll talk to my contact in B2 about the issue anyway.

ncw:
> I'll talk to my contact in B2 about the issue anyway.

This turns out to be [expected behaviour](https://github.com/ncw/rclone/issues/2164#issuecomment-375643892)

So either we put a few `time.Sleep(time.Second)` in a few judicious places to separate the upload and delete by more than 1 second or we could just ignore the problem.

internal/backend/rclone/backend.go
fd0: Yep, we'll do that. Thanks!

internal/backend/rclone/backend.go
fd0: We've added a section to the manual for the `rclone` backend, it also contains the hint.

fd0:
I've added a TODO section in the first post, so I can keep track of what's still to do :)

fd0:
Ping @ncw, I'd like your comment on the "piling up hidden data in cloud services" question above :) If you don't have the time right now that's fine too, I'd just like to make sure you've noticed that I'm waiting on your thoughts.

ncw:
@fd0 sorry for the delay in response

> Ah, hm. I think "undesirable" is the wrong word. I'm still not sure I've understood this completely, when files are hidden (B2) or moved to Trash (Google Drive), who will clean the files up and remove them permanently?

The answer is it depends on the cloud provider...  Drive doesn't empty the trash automatically, and b2 doesn't delete hidden files automatically.  However amazon cloud drive keeps the trash for 30 days (and you can't delete something without putting it into the trash).

However you can [set lifecycle rules for b2](https://www.backblaze.com/blog/backblaze-b2-lifecycle-rules/) to enable that auto cleanup.

> I know users can do that by running rclone cleanup manually, but it feels to me that many users of restic won't find this option, and I don't like piling up old data in hidden files or the Trash. Is there an automatic cleanup process happening after e.g. 7 days or so? Or do we need to call cleanup somewhere within restic after prune to get rid of the additional files?

I think there are three solutions
1. rely on the user to "empty the trash" or enable lifecycle rules or similar
2. Call `rclone cleanup` after `restic prune` - not a huge fan of this as cleanup is a really mixed bag for different providers - there isn't uniform semantics.  Drive empties all the trash for instance, whereas  b2 empties the trash that was in the path you passed in.
3. Don't make the trash at all.
  * either restic passes `--b2-hard-delete` and `--drive-use-trash=false` or
  * `restic serve` does this internally
  * or I implement https://github.com/ncw/rclone/issues/2165 and restic passes `--hard-delete`

I suppose the question is could it be useful for the user to have access to deleted files ever?  If so 1) is the best solution.  If not then I think 3) is the best solution.

fd0:
I think 3) is the best solution, I fear that most restic users are unaware that files are not really removed, and we don't offer any way to "undelete" the files. So I'd prefer that `rclone restic serve` implements hard delete. Is that realistic?

ncw:
> So I'd prefer that rclone restic serve implements hard delete. Is that realistic? Or should we add a workaround, having restic add the two options for B2 and Google Drive.

That sounds fine. I made an issue about this here: https://github.com/ncw/rclone/issues/2192

fd0:
Nice, thanks a lot!

fd0:
I've added the two parameters for now, we can remove them from restic once the next version of rclone is released. Thanks a lot for your work @ncw!

Once the integration tests complete I'll merge this!

PR #1660, opened 2018-03-08
backend/rest: check HTTP error response for List

ncw:
Before this change restic would attempt to JSON decode the error
message resulting in confusing `Decode: invalid character 'B' looking
for beginning of value` messages.  Afterwards it will return `List
failed, server response: 400 Bad Request (400)`

### Checklist


fd0 (approved):
Oh, good catch, thanks!

PR #1661, opened 2018-03-08
rest: Really set Content-Length HTTP header

fd0:


internal/backend/rest/rest.go
```diff
@@ -119,10 +118,13 @@ func (b *restBackend) Save(ctx context.Context, h restic.Handle, rd restic.Rewin
 	if err != nil {
 		return errors.Wrap(err, "NewRequest")
 	}
-	req.Header.Set("Content-Length", strconv.FormatInt(rd.Length(), 10))
 	req.Header.Set("Content-Type", "application/octet-stream")
 	req.Header.Set("Accept", contentTypeV2)
 
+	// explicitely set the content length, this prevents chunked encoding and
```
justinburke: s/explicitely/explicitly/

fd0: Great catch, thanks!

Issue #1605, opened 2018-02-06
Slow restore


jesseangell:
## Output of `restic version`
restic 0.8.1
compiled with go1.9.2 on linux/amd64

## How did you run restic exactly?
./restic -o b2.connections=100 restore --target /mnt/path SNAPSHOT_ID

## What backend/server/service did you use to store the repository?
Backblaze B2

## Expected behavior
Restore more than one file at a time.  Maybe take advantage of the -o b2.connections= option?

## Actual behavior
Only one file is being restored at a time - resulting in terrible throughput (about 32KB a second of IO being measured).  This box has a gigabit internet link and I'm stuck restoring at nearly dialup speeds.  Large files aren't an issue.
```
root@nas1:~# ps aux | grep rest
root     27197 16.1  6.8 329544 275744 pts/4   Sl+  07:52  57:38 ./restic -o b2.connections=100 restore --target /mnt/path SNAPSHOT_ID
ls -l /proc/27197/fd
total 0
lrwx------ 1 root root 64 Feb  6 13:33 0 -> /dev/pts/4
lrwx------ 1 root root 64 Feb  6 13:33 1 -> /dev/pts/4
lrwx------ 1 root root 64 Feb  6 13:33 2 -> /dev/pts/4
l-wx------ 1 root root 64 Feb  6 13:33 3 -> /mnt/path/archive/iMac 2014ish/user/Desktop/untitled folder 2/X-Plane 10 Demo/Resources/default scenery/1000 world terrain/terrain10/._red_grass_wrm_sdry_fl_md.ter
lrwx------ 1 root root 64 Feb  6 13:33 4 -> anon_inode:[eventpoll]
lrwx------ 1 root root 64 Feb  6 13:33 5 -> socket:[580367]
```
## Steps to reproduce the behavior?
Attempt to do a restore from B2.

## Do you have any idea what may have caused this?
Perhaps this is expected behavior - however if so that should be mentioned in the documentation. 

## Do you have an idea how to solve the issue?
Not familiar enough with Restic.

## Did restic help you or made you happy in any way?
It's been great - except for the inability to restore in a reasonable amount of time given substantial bandwidth.

zout:
have the same issue with s3 backend. It does not use the connections:
```
root@c4293c39beed:/# netstat -an
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
[...]

PR #1672, opened 2018-03-14
faster restorer musings, very much WIP

ifedorenko:

Proof-of-concept of faster restore implementation. This is meant to illustrate the idea and discuss its merits. 

[RESTORE_concurrent-pack-order.md](RESTORE_concurrent-pack-order.md) has description of the POC goals and implementation

Yes, see #1605.

### Checklist

### DIFF
--- a/RESTORE_concurrent-pack-order.md
+++ b/RESTORE_concurrent-pack-order.md
@@ -13,7 +13,7 @@ while there are packs to process
   get the pack from the backend or cache        [2]
   write pack blobs to the files that need them  [3]
   if not all pack blobs were used
-    put cache the pack cache                    [4]
+    cache the pack for future use               [4]
 ```
 
 Pack download and processing (steps [2] - [4]) runs on multiple concurrent goroutings. The POC runs all steps [2]-[4] sequentially on the same gorouting, but it is possible to split the work differently. For example, one pool of workers can handle download (step [2]) while the other pool can handle pack processing (steps [3] and [4]).
--- a/RESTORE_concurrent-pack-order.md
+++ b/RESTORE_concurrent-pack-order.md
@@ -1,0 +1,30 @@
+The main goal of this POC is two fold
+
+1. Reduce number of backend request
+  * download each pack with single backend request
+  * avoid repeated pack downloads when possible
+2. Download multiple pack files
+
+Here is high-level pseudo-code of the how the POC attempts to achieve these goals
+
+```
+while there are packs to process
+  choose a pack to process                      [1]
+  get the pack from the backend or cache        [2]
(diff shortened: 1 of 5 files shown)


fd0:
I've just skimmed over the description document, but didn't find any info on that: Does the implementation still work if there's no cache available?

ifedorenko:
The POC restorer uses temporary (as in `ioutil.TempFile`) disk space to store in-progress pack files, and this is what I called "pack cache" in the doc. In most cases only a small number of temporary files should be needed (one per worker thread, basically), but there is a hard limit on the amount of temp disk space used too. The temporary files are deleted after their blobs are used up by the restorer. 

The POC restorer does not currently use existing restic cache (as in `--cache-dir`). We can decide if and how the new restorer should use the existing cache, but I believe this is relatively minor implementation detail and would like to postpone this discussion until after we agree (or not) on overall restorer approach.

ifedorenko:
FYI, I refactored the POC to add some structure to the code. It's still far from ready, let alone pretty, but hopefully the code is now little easier to follow.

ifedorenko:
Finally got around to do some adhoc performance testing. 

Here is my setup:
* I have small-ish ~2GiB repository with little under 1800 JPEG files. 
* I use OneDrive repository backend, which I think is on par with S3 and GDrive in terms of single-threaded response time and download speed. 
* I have 50Mbps DSL connection to the Internet. I typically get ~6MB/s download speed from good remote servers.
* I am running the test on pretty old 2011 MacBook Pro with SSD. 
* I am running the following restore command:
```
     rm -rf /tmp/restic-restore; time ./restic restore --target=/tmp/restic-restore latest
```

The original/unoptimized restore took **`43m41.723s`**, which works out to ~778KB/s and way below my Internet max download speed of 6MB/s.

The optimized POC restore implementation took **`5m6.998s`**, or  ~6671KB/s, which is actually little faster than my Inet connection speed, likely thanks to restic dedup behaviour and MiB-vs-MB math :-). 

And here is what the original and POC restore look like in my internet router "Download Bandwidth Graph":

restic-restore-bandwidth

As you can see, my Inet connection is maxed out by POC implementation, so it'll likely go faster on faster connections. The original implementation restore time is almost certainly dominated by the server response time and will likely take similar time on faster connections.

ifedorenko:
@fd0 does this mean you agree with the approach in general? I am happy to work on the "real" implementation if you think the POC is convincing enough to give it a try.

fd0:
Oh yes, please go ahead! :)

ifedorenko:
This POC has served its purpose, please see #1719 with more "production ready" implementation.

PR #1675, opened 2018-03-18
Minor doc fixes/improvements

oliver:

Two minor doc fixes, one minor doc improvement.

No, not that I know of.

### Checklist

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -93,7 +93,7 @@ Basic example:
     *.go
     # exclude foo/x/y/z/bar foo/x/bar foo/bar
     foo/**/bar
-    $ restic -r /tmp/backup backup ~/work --exclude=*.c --exclude-file=exclude
+    $ restic -r /tmp/backup backup ~/work --exclude="*.c" --exclude-file=exclude
 
 Please see ``restic help backup`` for more specific information about each exclude option.
 
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -146,6 +146,8 @@ args:
 
     $ restic -r /tmp/backup backup --files-from /tmp/files_to_backup /tmp/some_additional_file
 
+Paths in the listing file can be absolute or relative.
+
 Comparing Snapshots
 *******************
 
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -126,7 +126,7 @@ backup from a file. This is especially useful if a lot of files have to
 be backed up that are not in the same folder or are maybe pre-filtered
 by other software.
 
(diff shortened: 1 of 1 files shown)


doc/040_backup.rst
```diff
@@ -93,7 +93,7 @@ Basic example:
     *.go
     # exclude foo/x/y/z/bar foo/x/bar foo/bar
     foo/**/bar
-    $ restic -r /tmp/backup backup ~/work --exclude=*.c --exclude-file=exclude
+    $ restic -r /tmp/backup backup ~/work --exclude="*.c" --exclude-file=exclude
```
fd0: For the record: This is not strictly necessary, but I think it's a good idea. Shells will resolve `--exclude *.c`, but won't resolve `--exclude=*.c`.

doc/040_backup.rst
```diff
@@ -126,7 +126,7 @@ backup from a file. This is especially useful if a lot of files have to
 be backed up that are not in the same folder or are maybe pre-filtered
 by other software.
 
-or example maybe you want to backup files that have a certain filename
+For example maybe you want to backup files that have a certain filename
```
fd0: Good catch!

doc/040_backup.rst
```diff
@@ -146,6 +146,9 @@ args:
 
     $ restic -r /tmp/backup backup --files-from /tmp/files_to_backup /tmp/some_additional_file
 
+Pathes in the listing file should be absolute. Relative pathes will be
```
fd0: The plural of "path" is "paths", can you please correct that?

I'm not sure if it's a good idea to recommend absolute paths for a file used with `--files-from`. There are plenty of use cases where it's important to specify relative paths. I'm fine with the hint that relative paths will be interpreted in context with the current directory (although that's basic knowledge the reader needs to be familiar with anyway), but I think we should not recommend that the file names in the file should be absolute. Can you remove the first sentence here? Thanks!

fd0 (changes requested):

fd0:
Thanks for taking the time to improve the documentation, it means a lot for the project! :)

oliver:
Oops, I had made a mistake when experimenting with that option, and had thought that the paths are interpreted as relative to the listing file location (not relative the CWD).

You're right, it is not necessary to document this behaviour for relative paths since they behave just as expected. I would add a note that both kinds of paths are allowed; but maybe even that note is superfluous.

fd0:
> I had made a mistake when experimenting with that option, and had thought that the paths are interpreted as relative to the listing file location (not relative the CWD).

Ah, that explains the desire to document this ;)

fd0 (approved):

PR #1676, opened 2018-03-20
Skip archiver.Scan before backup when --quiet is set

bowensong:
This PR ensures the backup command skips the `archiver.Scan()` before the backup actually starts when the quiet flag `-q` or `--quiet` is set.

The reason behind the change is because the scan result is only used for displaying the progress bar and ETA estimation, but the scan result is not used when the quiet flag is set.

By skipping the scan before backup, the backup speed improves on large directory trees, especially when the entire directory tree cannot entirely fit into page cache, in which case the directory tree was twice read from the disk before this change is made.

See issue #1160 

I don't think this PR completely closes the above issue. This is only a quick change to improve the backup performance in some use cases with a minimal change in the code and logic. 

### Checklist

- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/changelog-entry.tmpl))

### DIFF
--- a/changelog/unreleased/issue-1160
+++ b/changelog/unreleased/issue-1160
@@ -1,0 +1,5 @@
+Enhancement: Improve backup speed by skipping the initial scan when the quiet flag is set
+
+We've improved the backup speed when the quiet flag -q or --quiet is set by skipping the initial scan which gathers information for displaying the progress bar and the ETA estimation.
+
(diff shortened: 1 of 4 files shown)


fd0 (approved):
LGTM, thanks!

Issue #1694, opened 2018-03-30
restic check --with-cache does not populate the metadata-cache


kamudadreieck:
## Output of `restic version`
```
restic 0.8.3 (v0.8.3-0-g272ccec7)
compiled with go1.10 on linux/amd64
```

## How did you run restic exactly?
```bash
export RESTIC_PASSWORD=geheim
export RESTIC_REPOSITORY=rest:http://localhost:8000/rest-repo
restic check --with-cache --cache-dir /tmp/empty-folder
```
## What backend/server/service did you use to store the repository?
`rest-server 0.9.7 compiled with go1.10 on linux/amd64` listening on `127.0.0.1:8000`
`restic-server --log /tmp/restic-server.log`

## Expected behavior
`restic check --with-cache --cache-dir /tmp/empty-folder` creates a new cache and populates it, thus requesting data-blobs containing meta-data only once.

## Actual behavior
Data-blobs are requested very often, and are not cached at all (empty cache/data folder).
In the example below a total of 38773 requests are issued yet the repository consists only of 72 files.

## Steps to reproduce the behavior
Launch `rest-server 0.9.7 compiled with go1.10 on linux/amd64` listening on `127.0.0.1:8000`
Initalize new repository:
```bash
export RESTIC_PASSWORD=geheim
export RESTIC_REPOSITORY=rest:http://localhost:8000/rest-repo
restic init
```

Create some dummy backups, I used the `vendor` directory of restic since it contains many folders and small files.

```bash
for _ in `seq 20`; do restic backup ~/restic/vendor; done

find ~/restic/vendor | wc -l              
8857
find /tmp/restic/test-repo -type f | wc -l
72

du -hs /tmp/restic/test-repo ~/restic/vendor 
184M	/tmp/restic/test-repo
203M	/home/robert/restic/vendor

du -hs ~/.cache/restic/984f4f579e..
4,5M	/home/robert/.cache/restic/984f4f579e..
find ~/.cache/restic/984f4f579e../data | wc -l
7
```

Now check the repo:
```
debug log file /tmp/restic-debug.log
debug enabled
password is correct
create exclusive lock for repository
load indexes
check all packs
check snapshots, trees and blobs
no errors were found
restic check --with-cache --cache-dir /tmp/empty-folder  72,53s user 7,64s system 608% cpu 13,168 total

du -hs /tmp/empty-folder
1,1M	/tmp/empty-folder
find /tmp/empty-folder/984f4f579e../data | wc -l
1

[...]

PR #1696, opened 2018-03-31
check: Improve cache handling

### LINKED ISSUE #1665: Build new cache during check
During `check`, the cache is disabled for safety reasons, which means restic will access blobs stored in pack files as it needs them. When several blobs of a single pack file stored in the repo are needed one after another, restic accesses the same file in the backend several times. For some cloud-based backends (like B2), this behavior is billed, each download file transaction costs a very small amount of money.

Instead, we could change the behavior of `check`: Do not use the existing cache, but create a new one in a temporary directory. This means that the whole file is downloaded when the first blob is accessed, but then all further blobs in the same file can be read locally.

Shouldn't be hard to do, this is just a reminder issue to actually try that.

fd0:
For safety reasons, restic does not use a local metadata cache for the `restic check` command, so that data is loaded from the repository and restic can check it's in good condition. When the cache is disabled, restic will fetch each tiny blob needed for checking the integrity using a separate backend request. For non-local backends, that will take a long time, and depending on the backend (e.g. B2) may also be much more expensive.

This PR adds a few commits which will change the behavior as follows:

 * When `restic check` is called without any additional parameters, it will build a new cache in a temporary directory, which is removed at the end of the check. This way, we'll get readahead for metadata files (so restic will fetch the whole file when the first blob from the file is requested, this reduces the number of backend requests tremendously), but all data is freshly fetched from the storage backend.

 * When `restic check` is called with `--with-cache`, the default on-disc cache is used. This behavior hasn't changed since the cache was introduced.

 * When `--no-cache` is specified, restic falls back to the old behavior, and read all tiny blobs in separate requests.

Closes #1665, #1694

### DIFF
--- a/changelog/unreleased/issue-1665
+++ b/changelog/unreleased/issue-1665
@@ -1,0 +1,27 @@
+Enhancement: Improve cache handling for `restic check`
+
+For safety reasons, restic does not use a local metadata cache for the `restic
+check` command, so that data is loaded from the repository and restic can check
+it's in good condition. When the cache is disabled, restic will fetch each tiny
+blob needed for checking the integrity using a separate backend request. For
+non-local backends, that will take a long time, and depending on the backend
+(e.g. B2) may also be much more expensive.
+
+This PR adds a few commits which will change the behavior as follows:
+
+ * When `restic check` is called without any additional parameters, it will
+   build a new cache in a temporary directory, which is removed at the end of
+   the check. This way, we'll get readahead for metadata files (so restic will
+   fetch the whole file when the first blob from the file is requested), but
+   all data is freshly fetched from the storage backend. This is the default
+   behavior and will work for almost all users.
+
+ * When `restic check` is called with `--with-cache`, the default on-disc cache
+   is used. This behavior hasn't changed since the cache was introduced.
+
+ * When `--no-cache` is specified, restic falls back to the old behavior, and
+   read all tiny blobs in separate requests.
+
+https://github.com/restic/restic/issues/1665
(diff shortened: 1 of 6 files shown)


askielboe:
So I wanted to run a quick test against the 0.8.3 release on my B2 repo containing ~ 200 GB in 40 k files. I ran `restic check` with no additional arguments.

I ended up cancelling the 0.8.3 release after 26 hours, while a build including this pull request completed successfully in a bit under 15 minutes:

 - `master` (restic 0.8.3) - cancelled manually after 26 hours:
   - **time**: restic check  2128.88s user 653.90s system 2% cpu 25:58:24.15 total
   - **data**: sent: 824 MB (2 M packages), received: 11.69 GB (10 M packages)

 - `fix-check-cache` - completed successfully in 15 min:
   - **time**: `2553.88s user 346.09s system 325% cpu 14:51.61 total`
   - **data**: sent: 723 KB (1,824 packages), received: 1.67 GB (1,265,453 packages)

The improvements seem quite significant. :)

fd0:
Uh wow, thanks for the experience report! I suspected that in some situations the improvements may be dramatic. ;)

mathiasnagler:
This has a *dramatic* effect on me as well. 
Using a repository on Google Drive (over `rclone`), without this I cancelled the check after over 2h. With the change in place, the check finished in ~30 seconds.
Thank you very much, this is great!

stevesbrain:
Sounds like this will affect me significantly too then, as I'm same boat as @askielboe (slightly bigger repo size, but not too far different). Excited to try it out :)

PR #1709, opened 2018-04-07
Improve error message for orphaned pack files

fd0:


mholt:
@fd0 Great, thanks! Does this affect restic's exit code in the case of orphaned pack files? (From my reading it doesn't appear to, I just wanna be sure, since I will still run `prune` I think, even though it is non-critical.)

fd0:
Yes, the exit code is zero if orphaned files are present and there are no other issues.

mholt:
Hmm, okay. So I should always run prune no matter what `restic` exits with then.

fd0:
Just to make it clear: Having a few orphaned files in the repo is perfectly fine, except that they take up a tiny bit of space. That's why I changed the exit code. In most cases aborted backups leave them there.

I think the message printed in `restic check` was too strongly worded, that has been changed now.

mholt:
Thanks, yeah, I understand that -- just trying to minimize for optimal space usage, is all. Thanks!

PR #1720, opened 2018-04-15
key: Add --new-password-file flag for non-interactive password changes

### LINKED ISSUE #827: key add, key passwd: Allow new password via scripting
## Output of `restic version`
restic 0.4.0
compiled with go1.6.2 on linux/amd64

## Expected behavior
Restic should be fully scriptable to add a new key or rotate the password of a key:
* `restic -p $OLD_PASS_FILE key add --new-password-file $NEW_PASS_FILE`
* `restic -p $OLD_PASS_FILE key passwd $ID --new-password-file $NEW_PASS_FILE`

## Actual behavior
`restic -p` argument uses the same password for the new key as the existing key.

See also #826 that wants to add consistent manual selection of keys

mholt:

Makes it possible to change the password non-interactively.

I was unsuccessful trying to do it in a script by piping into stdin; restic expects 1000 bytes and the error check is wrong so it always returns an error even on early EOF; but even when corrected, I can only seem to pass one input, not two. Anyway, the whole thing doesn't matter if there's a nifty flag. :smile: 

Closes #827 

Also see https://forum.restic.net/t/changing-repo-password-without-prompt/591?u=matt

/cc @robbat2 and @HLeithner

### Checklist

(There don't seem to be tests for the commands, and this is automatically documented by `restic key -h`, so I'm marking this as done, I guess?)

### DIFF
--- a/changelog/unreleased/pull-1720
+++ b/changelog/unreleased/pull-1720
@@ -1,0 +1,7 @@
+Enhancement: Add --new-password-file flag for non-interactive password changes
+
+This makes it possible to change a repository password without being prompted.
+
+https://github.com/restic/restic/issues/827
+https://github.com/restic/restic/pull/1720
+https://forum.restic.net/t/changing-repo-password-without-prompt/591
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -3,6 +3,9 @@ package main
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
 
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/repository"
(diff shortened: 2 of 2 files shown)


fd0 (changes requested):
Hey, thanks for taking a stab at implementing this! Unfortunately, you implemented something slightly different: We try hard to not take a password as a command-line parameter (because all other users of the same system are allowed to read the command-line, so they could get the password), but to read it from a file.

Would you mind changing `--new-password` to `--new-password-file` according to the proposal in the issue, and read the password from the file?

This can then also be used with third-party programs, like `restic passwd --new-password-file <(pass restic/backup/foo)`.

Thanks!

mholt:
Sure, makes sense. I've made the change!

fd0:
Looks good, I've taken the liberty of squashing the commits, and I'll merge this after the tests ran. Thank you for your contribution!

fd0 (approved):

Issue #874, opened 2017-03-10
More stats in snapshots list


mnesrine:
It would be useful to have a more stats that are not currently in snapshots list like:
- Backup size
- Backup duration or end date

Thanks :)

middelink:
Would that be the size before dedupping or after? ^^

mnesrine:
Indeed that's a good point, by size i mean a size of 'file/dir' from local system not in remotely.
to make stats about a backup in remotely, we use another way not snaphsots list.

yhafri:
@middelink interesting observations +1

@mnesrine's idea to add more _metdata_ to `snapshots` command can be very useful.
No need to have a complicated solution here, just the bare minimum for now.

We can think of it as something complementary to the `--tag` option.
Imagine being able to retrieve:

- [x] the original size of the backup (before deduplication): this info can be easily calculated when restic's comparing local blocks with the server's ones
- [ ] the size of the backup after deduplication: this can help  compute how much disk saving we've got. Still optional for me, IMHO
- [x] the date when the backup was stored on the server: see no challenge here
- [x] the type of the backup: was it a file or a directory?

And now that we've the `--json` option in place when listing the `snapshots`, these _metadata_ can be retrieved and parsed in a nice way (thanks to [jq](https://stedolan.github.io/jq/)).

@middelink I see no conflict when a `prune` is performed, as nothing has to be updated. These are archives _metadata_. They always reflect what was true for a backup when it was stored in the first place.

fd0:
I like the idea of storing some statistics with the snapshot. Easy things:
 * Size of the data before deduplication
 * Backup duration
 * Number of new files/directories
 * Number of changed files

A bit more computationally expensive, but not too hard to do:
 * Size of the data after deduplication

@yhafri I have some questions:
 * I don't understand what you mean with "the local date when the backup was successfully stored". Each snapshot already has a time stamp, what's the difference here?
 * What's "the type of the backup"? We already store the list of things to be saved (what's passed to restic on the command line or via `--files-from`), so if I have two dirs and one file in there, what type of backup should that be? What's the type of backup for reading from stdin? Why is this information even relevant? Restic can just look at the nodes in the repo to determine if it is a file or directory for each of the backup targets, so what's the point in having the information again in the snapshot?

yhafri:
> I don't understand what you mean with "the local date when the backup was successfully stored". Each snapshot already has a time stamp, what's the difference here? Do you mean the finish time for the backup?

[...]

PR #1729, opened 2018-04-20
Implement `restic stats` command to get more info about a repository

mholt:

It adds a new command, `restic stats`, which shows the total original size of all the data in the repo, and the number of files (nodes) in the repo. This command supports the global `--json` flag.

**Feel free to help test it!** I think the counting is accurate but a second opinion would be nice.

This is my first time contributing an entire command to restic.

#874

Maybe https://forum.restic.net/t/can-we-see-snapshot-sizes/89?u=matt

### Checklist


mholt:
I think the count might be (mostly) accurate.

It seems that restic repositories are resilient to file moves, but not file moves + modifications. As a test, I took a folder of about 550 MB and backed it up. Then I moved a 150 MB file to another name, and backed up again. The size count (using this branch) didn't change much. Then I moved the file again AND corrupted it with some random bytes at the end (just a few), and backed up again. The reported size shot up to about 700 MB.

Is this expected? It makes sense, except that the moved file has only a few bytes different. It's OK either way, I am just wondering about this phenomenon.

ifedorenko:
Can you clarify what you mean by "total original size of all the data in the repo" and/or what you want to use this information for? Specifically,

If the same file (as in, "same path and content") is present in two snapshots, should the file size be counted one or two times?

If the same content (as in "file sha256") is present at two locations within the same snapshot, should the content size be counted one or two times? What if the same content is present at two different locations in two different snapshots?

PS: I like the idea of `restic stats` command and was planning to implement something similar myself, just want to understand if your usecase(s) are similar to mine or not.

fd0:
I had a quick look, and the optimization to only walk trees that haven't been walked before makes the overall statistics wrong. Easy example:

```
$ echo foo > testfile.txt
$ ./restic backup testfile.txt
```

Now the stats show 1 file with 4 byte:
```
$ ./restic stats
   Cumulative Original Size:   4B
  Total Original File Count:   1
```

Backup the same file again, still the same stats:
```
$ ./restic backup testfile.txt
$ ./restic stats
   Cumulative Original Size:   4B
  Total Original File Count:   1
```

Modify any attribute of the file (e.g. access permissions), now the tree is different (because the metadata is different) and the file is counted twice:

```
$ chmod 600 testfile.txt
$ ./restic backup testfile.txt
$ ./restic stats
password is correct
   Cumulative Original Size:   8B
  Total Original File Count:   2
```

Now we have three snapshots referencing two trees referencing a single file (with slightly different metadata).

So, I think we first need to define: What's the "Total Original File Count"? In which cases are files the same, or different?

mholt:
@ifedorenko 

> Can you clarify what you mean by "total original size of all the data in the repo" and/or what you want to use this information for? Specifically,

I'd like to count the "restore potential" of a repository (or a set of snapshots, in the future -- but for now this PR just does the whole thing).

> If the same file (as in, "same path and content") is present in two snapshots, should the file size be counted one or two times?

Once.

> If the same content (as in "file sha256") is present at two locations within the same snapshot, should the content size be counted one or two times?

Hmm, probably twice, since they are in the same snapshot, they seem to be distinct files.

> What if the same content is present at two different locations in two different snapshots?

Once, since the file could have simply been moved.

I think. 😅 

I'm basically asking, "If I were to restore all the data in the repo right now, how much would it expand to?" -- without over-counting file moves and small edits, preferably, but I can see how that's super hard to do.

@fd0 Aha, you seem to have reduced my experiments to a minimal case -- thank you for that! I think I experienced the same thing in my testing. So, I'm not 100% sure on all the details yet; I am exploring and defining this problem as I got through this. But I agree, that text file should only be counted once, and is likely why my count is so high.

I was looking at the [Node](https://godoc.org/github.com/restic/restic/internal/restic#Node) type and [ID](https://godoc.org/github.com/restic/restic/internal/restic#ID) type and trying to figure out how to identify a "unique file" as much as possible, but I could use a little guidance, or some feedback from others who have similar needs too.

I think changing the attributes of a file is probably not enough for my needs to count it as a new/different file. I guess I would prefer that a unique _filename_ be the minimum requirement to be counted again.

ifedorenko:
"restore potential" is still not specific enough to decide how to implement this, I think. Do you want to be able to restore all repository snapshot simultaneously or just any single repository snapshot? If restore multiple snapshots, do you hardlink same files (as in "path+sha256") from different snapshots or keep them separate?

And, btw, you need to account for hardlinks within snapshots, see how see how [restorer uses HardlinkIndex](https://github.com/restic/restic/blob/v0.8.3/internal/restic/restorer.go#L155).

ifedorenko:
Yes, I understand you only count, but what you count depends on how you want to restore. For example, if you ever want to be able to restore a single snapshot, then you need to know the largest snapshot in the repository. Or if you want to be able to restore some/all snapshots into separate directories, then you need to decide how to deal with cross-snapshot hardlinks and count accordingly. Restore some/all snapshots into the same directory will require yet another way to count. And if you don't actually want to restore anything and only want to get a number ;-) than counting unique path+sha256 across all snapshots seems like the way to go.

HardlinkIndex uses Node DeviceID and Inode to correlate paths that point to the same file.

mholt:
>  And if you don't actually want to restore anything and only want to get a number ;-) than counting unique path+sha256 across all snapshots seems like the way to go.

Let's go with that, for now. We can always add options / flags to make this command do more or customize it to count only certain things for certain use cases.

mholt:
I guess, I did not expect that changing metadata like permissions resulted in a totally new tree for a file. Is there a way to ignore metadata changes when counting "unique" objects?

fd0:
For me, the canonical way to define "unique files" would be to only look at the content, so I propose to define a "unique file" as the (ordered) list of data blob IDs its content consists of. This catches all things I can imagine in an intuitive way: Files with the same content in different directories and hard links. It also ignores all other items (directories, symlink), these are only metadata anyway.

When we have this definition, it's easy to count the number of "unique files" in a snapshot (or even all snapshots):
 * Define `type FileID [32]byte`
 * Make a `map[FileID]uint`, which maps a file ID to its size
 * Walk the tree, for each file, take the list of content blob IDs as a stream of bytes, and hash it with SHA-256 to get the file ID
 * If the map already contains the ID, carry on, if not record the sum of the sizes of the blobs

This way, you'll end up with a map of "unique files" and their sizes, that you can count and summarize as you wish.

Thoughts?

mholt:
@fd0 Awesome! That's exactly what I was going for; you even went the extra mile and described the details of the algorithm. That should speed up my revisions, thank you.

As for hard links, perhaps how they are counted can be configurable with a flag. I don't plan on implementing that in my PR (at least for now), but no reason it couldn't be later. I see this command as being flexible. I'm just trying to fulfill my own needs at the moment while leaving it open to further development.

fd0:
Hm. What I'm interested in (besides unique files) would be the restore size for a single snapshot, let's call that the "size" of a snapshot. No need to implement this in the `stats` command (which would be a more global view), I'd just like to dump my thoughts onto you. The command may be something akin to `git show <commit>`, `restic show <snapshot>` or so. It could even compute the size for include/exclude patterns :)

The sum of the sizes of all unique files in that snapshot is the lower bound, there may be files with the same content stored twice (or more times) in the saved directory, so they'd be restored several times. The sum of the sizes of all files is an upper bound. Walking the tree, making a sum of all file sizes while accounting for hard links should be sufficient to get the size needed for restore.

So I would define the "size" of a snapshot as the sum of the sizes of all files in it, minus the sum of the sizes of the files which are hard linked (link count > 1). We already have a data structure called `HardLinkIndex` we could use.

fd0:
> As for hard links, perhaps how they are counted can be configurable with a flag.

I'd leave them out for now, for most users, hard links are a rather obscure feature (but we need to support them obviously).

mholt:
> Hm. What I'm interested in (besides unique files) would be the restore size for a single snapshot, let's call that the "size" of a snapshot. No need to implement this in the stats command (which would be a more global view), I'd just like to dump my thoughts onto you.

I would also be interested in this, definitely! I think that would be useful, and if I may suggest, I think it's not a bad fit for the `restic stats` command. I mean, why not be able to specify a snapshot or other criteria for multiple snapshots, while defaulting to all of them? Imagine `restic stats latest` to get the restore size of the latest snapshot.

I may not implement that for the initial PR for time constraints and also to keep concerns separate, but I'm happy to change this one to make it mergeable, with future development in mind.

@fd0 I have updated the PR to use your proposed algorithm. It seems to work quite nicely so far, and is fast. I have yet to do more testing (regarding renames, deletes, metadata changes, etc.) but here's what I know so far:

My ~500 GB repository now reports 520 GB of data stored (I don't delete much), which is a marked difference from 3.3 TB. It took about 20 seconds to run over my LAN (from a raspberry pi, too! - running rest-server integrated into Caddy) and used a little less than 200 MB memory.

mholt:
Is there a cheap way to get the size of a blob by ID? I see `repository.LoadBlob` but I'm afraid that would download the entire blob.

ifedorenko:
`repo.Index().Lookup(...)` gives you indirect access to `restic.Blob`, which has blob's encrypted Length. Plaintext length is `blob.Length-crypto.Extension`. Index lookup should be reasonably cheap. 

Are you trying to count sizes of unique blobs referenced by snapshots? That's an interesting idea.

mholt:
@ifedorenko Thanks! I didn't realize those paths were possible too.

> Are you trying to count sizes of unique blobs referenced by snapshots? That's an interesting idea.

Yeah, basically. I'm just exploring this while I'm at it. I'm content with the current algorithm as committed, but a 1 byte change to a 200 MB file results in counting 400 MB because, technically, the two files are different. And I think that's justified, but again, I'm just exploring what is possible here. I'll report back whether this method by counting unique blob references works.

fd0:
I'd like to propose another metric: "raw data size". It can be applied to the repo as  a whole or to an individual snapshot. Define "raw data size" as the sum of the sizes of all referenced data blobs (for a snapshot) or all data blobs (for the repo itself). That's a pretty good metric for finding out how much data there is in the repo. Changing one byte of a 50GB file will result in +50GB and +1 "unique file", but will only result in a single new data blob (~4MB at most).

mholt:
@fd0 Sure, that's pretty similar to the size of the `data` folder but, as mentioned earlier, without the encryption overhead, etc. I'm good with flags being added to change the counting mode. I think now we have about 3. :)

mholt:
@fd0 @ifedorenko I have pushed a commit where I've implemented 4 counting modes:

- `--count-restore-size` (default) will count the size of a restore of the set of files
- `--count-files-by-contents` counts the sizes of unique files only (unique is defined to be the sequence of their Content blobs)
- `--count-blobs-per-file` counts the sizes of blobs referenced by unique files at most once (unique by content and by file path)
- `--count-raw-data` counts the sizes of each unique blob, independent of how many times they are referenced by files

A snapshot can be specified like so: `restic stats latest` to walk just one snapshot. (The `--host` flag is supported for choosing the latest snapshot for a certain host.) If no snapshot is given: `restic stats` then the whole repository is walked (all snapshots).

I think I'm ready for a final series of reviews to get this merged.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,320 @@
...
+					for _, blobID := range node.Content {
+						// ensure we have this file (by path) in our map; in this
+						// mode, a file is unique by both contents and path
+						if _, ok := stats.fileBlobs[fpath]; !ok {
+							stats.fileBlobs[fpath] = restic.NewIDSet()
+							stats.TotalFileCount++
+						}
+						if _, ok := stats.fileBlobs[fpath][blobID]; !ok {
+							// TODO: Is the blob type always 'data' in this case?
+							blobSize, found := repo.LookupBlobSize(blobID, restic.DataBlob)
```
mholt: Is it fair to assume this is a data blob? Or do I need more logic here.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,321 @@
...
+						// ensure we have this file (by path) in our map; in this
+						// mode, a file is unique by both contents and path
+						nodePath := filepath.Join(fpath, node.Name)
+						if _, ok := stats.fileBlobs[nodePath]; !ok {
+							stats.fileBlobs[nodePath] = restic.NewIDSet()
+							stats.TotalFileCount++
+						}
+						if _, ok := stats.fileBlobs[nodePath][blobID]; !ok {
+							// TODO: Is the blob type always 'data' in this case?
+							blobSize, found := repo.LookupBlobSize(blobID, restic.DataBlob)
```
mholt: Is it fair to assume this is a data blob? Or do I need more logic here. 🤔

fd0: Yes, you're only filling `fileBlobs` with blobs from `Content` attribute of a file, those are all data blobs.

mholt:
@fd0 In case it wasn't clear, I'm ready for this one to be reviewed. :) It's working well for my needs!

fd0:
FYI, I'm sick and swamped with other work.

mholt:
Yeah, I heard... :-/ Take your time; and get well.

mholt:
What do you want for the docs? Should I basically convert my comment above into a section in the manual?

fd0:
I just had another look, this looks good to me. What I'd like to discuss is the user interface: calling `restic stats --count-files-by-contents` made me wonder if we need the `count` in there. Also, we're setting an (exclusive) mode via different options. Wouldn't that be better to have e.g. `restic stats --mode files-by-contents`?

Last point: Rebasing this on master would enable replacing the custom walk function by the newly introduced tree walker: https://godoc.org/github.com/restic/restic/internal/walker#Walk

Let me know what you think!

mholt:
Thanks! I'm making those changes now. Great idea about the flags... I think at first I didn't realize they would need to be exclusive to make sense.

Question about walking with the new Walker: right now [I manually keep a set of nodes](https://github.com/restic/restic/pull/1729/files#diff-36226cfa04789be10fd460afab45619aR164) I've already visited so I don't visit them again. Will I need to do that with the new walker? I see the `ignore` return value but I'm not 100% clear on how it works from the docs alone.

fd0:
> Question about walking with the new Walker: right now I manually keep a set of nodes I've already visited so I don't visit them again. Will I need to do that with the new walker? I see the ignore return value but I'm not 100% clear on how it works from the docs alone.

The idea is that the walker keeps track of the trees you don't want to visit again for you, it's the same what you're doing manually (with [extensive tests](https://github.com/restic/restic/blob/master/internal/walker/walker_test.go#L188-L459)). `Walk` adds the trees the walk function returns `ignore=true` to the `ignoreTrees` map, so when you call `Walk()` several times in a row, you can supply the start set of ignored trees. Since this is a use case we need quite often within restic, I've written code in the lib part of it.

Actually, now that I have a look, the file `walk.go` containing the implementation is 3.5KiB, and the file `walk_test.go` is more than double the size with ~10KiB...

mholt:
@fd0 Cool, if I understand you right, then returning `true` from my walk function is what I want to do in order to not visit the same node twice.

I've pushed the changes. We now have a single `--mode` flag that accepts as values: `restore-size` (default), `files-by-content` (should it be "contents"? I waffled), `blobs-per-file`, or `raw-data`.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,296 @@
...
+	return nil
+}
+
+func statsWalkTree(npath string, node *restic.Node, nodeErr error, repo restic.Repository, stats *statsContainer) (ignore bool, err error) {
+	if nodeErr != nil {
+		return true, nodeErr
+	}
+	if node == nil {
+		return true, nil
+	}
```
mholt: These two checks at the beginning of my Walk seemed necessary; let me know if I missed something. I know I got a nil pointer dereference because sometimes the node passed in was nil.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,296 @@
...
+
+	if countMode == countModeRawData {
+		// count just the sizes of unique blobs; we don't need to walk the tree
+		// ourselves in this case, since a nifty function does it for us
+		return restic.FindUsedBlobs(ctx, repo, *snapshot.Tree, stats.blobs, stats.blobsSeen)
+	}
+
+	err := walker.Walk(ctx, repo, *snapshot.Tree, restic.NewIDSet(), func(path string, node *restic.Node, nodeErr error) (bool, error) {
+		return statsWalkTree(path, node, nodeErr, repo, stats)
```
mholt: I had to wrap the Walk function like this so that I could carry through the `repo` and `stats` variables. Hope that's OK; let me know if you know a better way to do this.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,296 @@
...
+
+	if countMode == countModeRestoreSize {
+		// as this is a file in the snapshot, we can simply count its
+		// size without worrying about uniqueness, since duplicate files
+		// will still be restored
+		stats.TotalSize += node.Size
+		stats.TotalFileCount++
+	}
+
+	return true, nil
```
mholt: I always return `true` to avoid visiting the same node twice -- that's my understanding, anyway. This should work just like we did before I revised this PR to use the Walk function, if I did it right.

cmd/restic/cmd_stats.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// statsContainer holds information during a walk of a repository
+// to collect information about it, as well as state needed
+// for a successful and efficient walk.
+type statsContainer struct {
+	TotalSize      uint64 `json:"total_size"`
+	TotalFileCount uint64 `json:"total_file_count"`
+	TotalBlobCount uint64 `json:"total_blob_count,omitempty"`
+
```
mholt: We no longer have an "idSet" field here because I rely on the Walk function to add each node to the set of ignored trees so we only visit each one once.

mholt:
I know you have a lot in your notifications, so this is a friendly reminder that this is ready to review, and probably merge, if my comments above didn't point out anything too glaring. :) Thanks again for your help!

cmd/restic/cmd_stats.go
fd0: That's true, the first time the walk function is called (for `/`), the node is nil because there's no node for the top-level tree. It's a bit weird, I agree.

cmd/restic/cmd_stats.go
fd0: Not necessarily better, but I probably would have made `statsWalkTree` return a `walker.WalkFunc`, then it becomes slightly less ugly:
```go
func statsWalkSnapshot(ctx context.Context, snapshot *restic.Snapshot, repo restic.Repository, stats *statsContainer) error {
	[...]
	err := walker.Walk(ctx, repo, *snapshot.Tree, restic.NewIDSet(), statsWalkTree(repo, stats))
	[...]
	return nil
}

func statsWalkTree(repo restic.Repository, stats *statsContainer) walker.WalkFunc {
	return func(npath string, node *restic.Node, nodeErr error) (bool, error) {
		if nodeErr != nil {
			return true, nodeErr
		}
		if node == nil {
			return true, nil
		}

		[...]

		return true, nil
	}
}
```

cmd/restic/cmd_stats.go
fd0: Look good

cmd/restic/cmd_stats.go
fd0: Ah, since I had it already, I'll add a commit which changes this.

fd0 (approved):

fd0:
So, this is ready to be merged. The last two things that I'd like to see:
 * an explanation of the `stats` command in the manual, maybe in a new section, which explains the counting modes (using the examples/explanation in the comments here)
 * a short version of that in the help text for the `stats` command. I'd like to improve the online help, and this is a good first step

Since the PR is so old already, let me know if I should merge it and you'll afterwards add the docs, or if you'd like to add them on top of the commits here. In addition, I'd like to squash your commits, is that ok?

mholt:
Oops. I have been trying a new editor and I guess I'm a little clumsy still, that rebase wasn't good first time around. I think I've fixed it up, but a second pair of eyes might help.

Added docs in both places, as requested! :star:

fd0:
I've taken the liberty of squashing the two new commits in order to find out what you changed in the code. It seems that this bit here got dropped, was that intentional?

```diff
@@ -192,14 +211,7 @@ func statsWalkTree(repo restic.Repository, stats *statsContainer) walker.WalkFun
                            if !found {
                                return true, fmt.Errorf("blob %s not found for tree %s", blobID, *node.Subtree)
                            }
-
-                           // count the blob's size, then add this blob by this
-                           // file (path) so we don't double-count it
                            stats.TotalSize += uint64(blobSize)
-                           stats.fileBlobs[nodePath].Insert(blobID)
-
-                           // this mode also counts total unique blob _references_ per file
-                           stats.TotalBlobCount++
                        }
                    }
                }
```

fd0 (approved):
Looks good, apart from the small bit :)

Let me know what you'd like me to with it, leave it out or put it back in? Then I can merge this PR (finally).

mholt:
Ah -- I don't think I intended to remove that. Oops. I'm not sure why that happened. When I drafted my last commit, I later discovered that one reason I was confused was because the commit you made apparently unintentionally indented  `if countMode == countModeRestoreSize {` into the if statement above it, which would cause this one to never be evaluated, since the modes are exclusive of each other. I fixed the indentation but have to admit was a little confused about a new "variable declared but not used" error that I guess removing those lines fixed. Anyway, it was rebase confusion -- totally my bad.

I'm pushing a fix now to restore those lines.

mholt:
Also, thank you for finding that :ok_hand:

fd0:
Ah, hm, there we are, git is hard to use. I've removed the merge commit you added, and squashed the three commits reworking the `stats` command into one. I think we don't need to document for future readers that there was a bad rebase ;)

Once the tests are done, I'll merge this. Thanks!

fd0 (approved):

mholt:
@andreasnuesslein I think it would use a lot more memory.

pepa65:
Or maybe a syntax with commas between multiple modes as the user requires?

PR #1751, opened 2018-05-01
archiver: Use lstat before open/fstat

fd0:
The previous code tried to be as efficient as possible and only do a single open() on an item to save, and then fstat() on the fd to find out what the item is (file, dir, other). For normal files, it would then start reading the data without opening the file again, so it could not be exchanged for e.g. a symlink.

This behavior starts the watchdog on my machine when /dev is saved with restic, and after a few seconds, the machine reboots.

This commit reverts the behavior to the strategy the old archiver code used: run lstat(), then decide what to do. For normal files, open the file and then run fstat() on the fd to verify it's still a normal file, then start reading the data.

The downside is that for normal files we now do two stat() calls (lstat+fstat) instead of only one. On the upside, this does not start the watchdog. :)

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -7,7 +7,6 @@ import (
 	"path"
 	"runtime"
 	"sort"
-	"syscall"
 	"time"
 
 	"github.com/restic/restic/internal/debug"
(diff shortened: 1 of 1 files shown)


wscott:
As fast as possible is to use d_type out of 'struct dirent' on systems that support it and then fall back
to lstat() on ones that don't.  Usually, that involves writing wrapper code for directory reading because windows can do the same thing it just uses different system calls.

I have written this in C. I am not a Go programmer so I don't have a suggestion of how to match the behavior.

fd0:
Thanks, but for now we'll stay with the tried-and-true lstat ;)

PR #1769, opened 2018-05-10
cat snapshot: Do not display a stack trace with invalid IDs

McKael:
Don't show a stack trace when "restic cat snapshot" is invoked with
invalid/nonexistent IDs.

This patch makes restic more user-friendly and consistent by not displaying a stack trace when the user calls `restic cat snapshot ID` with an invalid ID.

Behavior without the patch:
```
% restic cat snapshot 12345678
repository 24a9cce4 opened successfully, password is correct
no ID found
github.com/restic/restic/internal/restic.init
	src/github.com/restic/restic/internal/restic/backend_find.go:11
github.com/restic/restic/internal/archiver.init
	<autogenerated>:1
main.init
	<autogenerated>:1
runtime.main
	/opt/go/src/runtime/proc.go:186
runtime.goexit
	/opt/go/src/runtime/asm_amd64.s:2361
zsh: exit 1
```

Behavior with the patch:
```
% restic cat snapshot 12345678
repository 24a9cce4 opened successfully, password is correct
Fatal: could not find snapshot: no ID found

zsh: exit 1
```

It was briefly discussed with fd0 on IRC.

### Checklist

I'm not sure this change requires a changelog entry, please let me know if it needs one...

### DIFF
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -58,7 +58,7 @@ func runCat(gopts GlobalOptions, args []string) error {
 			// find snapshot id with prefix
 			id, err = restic.FindSnapshot(repo, args[1])
 			if err != nil {
-				return err
+				return errors.Fatalf("could not find snapshot: %v\n", err)
 			}
 		}
 	}


fd0 (approved):
Thanks!

PR #1772, opened 2018-05-11
Restore verify

ifedorenko:

Adds new restore `--verify` option to verify restored file contents

See #1719

### Checklist

### DIFF
--- a/changelog/unreleased/pull-1772
+++ b/changelog/unreleased/pull-1772
@@ -1,0 +1,6 @@
+Enhancement: Add restore --verify to verify restored file content
+
+Restore will print error message if restored file content does not match
+expected SHA256 checksum
+
+https://github.com/restic/restic/pull/1772
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -5,6 +5,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/restorer"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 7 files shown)


fd0 (approved):
LGTM, thanks!

fd0:
> Do potential users have to install restic and run restic COMMAND help to find out what options are available?

For a description of all the options, yes, users need to do that. Duplicating the description in the manual is (in my opinion) not helpful and a lot of work. The manual should provide an overview of the workflow and point users to the right direction so they can find more info. Syncing all the descriptions is also a lot of (manual) work.

Over time I'd like to add more documentation to the source code itself, so each command documents itself and the documentation is right next to the code. We already have that in some commands, e.g.:
```
$ restic mount -h

The "mount" command mounts the repository via fuse to a directory. This is a
read-only mount.

Snapshot Directories
====================

If you need a different template for all directories that contain snapshots,
you can pass a template via --snapshot-template. Example without colons:

    --snapshot-template "2006-01-02_15-04-05"

You need to specify a sample format for exactly the following timestamp:

    Mon Jan 2 15:04:05 -0700 MST 2006

For details please see the documentation for time.Format() at:
  https://godoc.org/time#Time.Format

Usage:
  restic mount [flags] mountpoint
```

Maybe we can even have a brief overview of all the options with `restic CMD --help` and only print the examples and extensive text when `restic help CMD` is run.

Would you like to work on that?

alphapapa:
Well, putting full descriptions in the manual is some work, but I think it's very important for showing potential users what restic can do.  Every option is a potential dealmaker for a new user, but if they don't know it exists, it can't make a difference, and many won't install restic to find out what it can do.

I agree that, ideally, documentation for commands and options should live in the source.  And, ideally, it could be automatically extracted into a manual.  I don't know much about Go; are there any tools that can do that for Go?

mholt:
@alphapapa You could look into how rclone generates their website documentation automatically: https://github.com/ncw/rclone/blob/master/CONTRIBUTING.md#writing-documentation

But it might require some significant changes or additions to this repository, which I'm not sure if that is ideal. I don't know of any tooling that automatically generates the kind of documentation you're looking for, without integrating directly into the source code.

Issue #1777, opened 2018-05-12
Diagnostic command to display what snapshots a blob/tree belongs to


McKael:
_This is not a bug report but a feature proposal._

A few times, I've been missing a command to list the snapshots that contain a specific tree or blob, or the snapshots affected by a corrupted pack.

I'd like to be able to use a restic command for this case (AFAIK it does not exist yet) and I've written some POC code.
I can do a PR to discuss this further if this suggestion sounds reasonable to you.

Here's a practical example.

```
% restic check
repository a04959d0 opened successfully, password is correct
create exclusive lock for repository
load indexes
check all packs
check snapshots, trees and blobs
(...)
error for tree e437ff75:
  tree e437ff75: file "checks.db": metadata size (122171392) and sum of blob sizes (122175488) do not match
Fatal: repository contains errors
```

Let's check what this tree actually is and which snapshot is affected.

```
% restic what-contains tree e437ff75
repository a04959d0 opened successfully, password is correct
Found tree e437ff75
 ... path /mikael/check-data
 ... in snapshot f120bfbf
```

Here's another example with a random blob in one of my repositories:
```
% restic what-contains blob 9dcf5f6a254fb58c8b2a7e604dc76b71fbad847f6ade751513968cf2f1f161fc
repository a04959d0 opened successfully, password is correct
Blob found in pack cf8a3fa3b2c9c7a08b6b6fc6cce9fdc2eeb4dd26ebd2d67b081d63f24dd408d8
 ... Pack cf8a3fa3: <Blob (data) 9dcf5f6a, offset 2486760, length 12214>
Found blob 9dcf5f6a254fb58c8b2a7e604dc76b71fbad847f6ade751513968cf2f1f161fc
 ... in file 1398431273.H215567P22415.mail.example.com:2,
 ... in snapshot 067f149f
Found blob 9dcf5f6a254fb58c8b2a7e604dc76b71fbad847f6ade751513968cf2f1f161fc
 ... in file 1398431273.H215567P22415.mail.example.com:2,
 ... in snapshot 0a781e61
Found blob 9dcf5f6a254fb58c8b2a7e604dc76b71fbad847f6ade751513968cf2f1f161fc
 ... in file 1398431273.H215567P22415.mail.example.com:2,
 ... in snapshot 15ab2468
(...)
```

I'd also like to be able to do
```
% restic what-contains pack cf8a3fa3
```
but I haven't had a look at it yet.

Please let me know if it would be useful to other people than me.

[...]

PR #1780, opened 2018-05-13
Add options to 'find' for searching IDs (blobs, ...)

McKael:

This PR adds a command "what-contains" that lists the snapshots containing a specific tree or blob,
or the snapshots that contain blobs belonging to a given pack.

It also displays the pack ID a blob belongs to.

A list of IDs can be given, as long as the IDs are all of the same type.

I have created the following issue:
https://github.com/restic/restic/issues/1777

### Checklist

### DIFF
--- a/changelog/unreleased/pull-1780
+++ b/changelog/unreleased/pull-1780
@@ -1,0 +1,12 @@
+Enhancement: find: Add options for searching IDs (blobs, trees...)
+
+We've enhanced the `find` command to be able to list the snapshots containing
+a specific tree or blob, or even the snapshots that contain blobs belonging
+to a given pack.
+A list of IDs can be given, as long as they all have the same type.
+
+The command `find` can also display the pack IDs the blobs belong to, if
+the `--show-pack-id` flag is provided.
+
+https://github.com/restic/restic/issues/1777
+https://github.com/restic/restic/pull/1780
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -173,12 +173,13 @@ func (s *statefulOutput) PrintPattern(path string, node *restic.Node) {
 	}
 }
 
-func (s *statefulOutput) PrintObjectJSON(kind, id, nodepath string, sn *restic.Snapshot) {
+func (s *statefulOutput) PrintObjectJSON(kind, id, nodepath, treeID string, sn *restic.Snapshot) {
 	b, err := json.Marshal(struct {
 		// Add these attributes
 		ObjectType string    `json:"object_type"`
 		ID         string    `json:"id"`
-		Path       string    `json:"path,omitempty"`
+		Path       string    `json:"path"`
+		ParentTree string    `json:"parent_tree,omitempty"`
 		SnapshotID string    `json:"snapshot"`
 		Time       time.Time `json:"time,omitempty"`
 	}{
@@ -186,6 +187,7 @@ func (s *statefulOutput) PrintObjectJSON(kind, id, nodepath string, sn *restic.S
 		ID:         id,
 		Path:       nodepath,
 		SnapshotID: sn.ID().String(),
+		ParentTree: treeID,
 		Time:       sn.Time,
 	})
 	if err != nil {
@@ -203,21 +205,22 @@ func (s *statefulOutput) PrintObjectJSON(kind, id, nodepath string, sn *restic.S
 	s.hits++
 }
 
-func (s *statefulOutput) PrintObjectNormal(kind, id, nodepath string, sn *restic.Snapshot) {
-	f := "path"
-	if kind == "blob" {
-		f = "in file"
-	}
+func (s *statefulOutput) PrintObjectNormal(kind, id, nodepath, treeID string, sn *restic.Snapshot) {
 	Printf("Found %s %s\n", kind, id)
-	Printf(" ... %s %s\n", f, nodepath)
+	if kind == "blob" {
+		Printf(" ... in file %s\n", nodepath)
+		Printf("     (tree %s)\n", treeID)
+	} else {
+		Printf(" ... path %s\n", nodepath)
+	}
 	Printf(" ... in snapshot %s (%s)\n", sn.ID().Str(), sn.Time.Format(TimeFormat))
 }
 
-func (s *statefulOutput) PrintObject(kind, id, nodepath string, sn *restic.Snapshot) {
+func (s *statefulOutput) PrintObject(kind, id, nodepath, treeID string, sn *restic.Snapshot) {
 	if s.JSON {
-		s.PrintObjectJSON(kind, id, nodepath, sn)
+		s.PrintObjectJSON(kind, id, nodepath, treeID, sn)
 	} else {
-		s.PrintObjectNormal(kind, id, nodepath, sn)
+		s.PrintObjectNormal(kind, id, nodepath, treeID, sn)
 	}
 }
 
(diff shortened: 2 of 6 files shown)


fd0:
FYI, I'm sick and swamped with other work.

fd0:
This looks great already, thanks for your work! I wonder if we can integrate the functionality into the `find` command (even if that means changing it a bit)? I've just recently merged #1839, which includes not only a reworked implementation of `find`, but also a [`walker.Walk()`](https://godoc.org/github.com/restic/restic/internal/walker#Walk), which is modeled after `filepath.Walk()` and makes it really easy to traverse a tree in the repo.

Concerning the user interface, I'm a fan of no-knobs solutions, so if restic is able to figure out which type of thing (pack, data, or tree blob) an ID references, it should do that and not prompt the user (or have them type out the type).

What do you think? I'm glad to help out :)

McKael:
Thanks for the feedback.

TBH I was not very keen on adding a new command but I could see two advantages with my approach:
- it's very user-friendly and intuitive (`what-contains tree e437ff75` -- or `what-contains e437ff75` if we want to omit the type)
- having the type given explicitly seemed to make it simple and probably more efficient to give a list of blobs or trees (they're all searched at once), but maybe I'm wrong on this.

I'm not sure how it would fit the `find` command, that currently takes a path pattern as argument if I'm not mistaken.
I'm fine with trying to adapt the patch, I'd appreciate if you could give me an example usage we could have with `find`...

fd0:
Sure: What we already have is a `find` command which takes a pattern (as the positional argument) and some options to limit which snapshots are considered for the search:
```
Usage:
  restic find [flags] PATTERN
Flags:
  -h, --help            help for find
  -H, --host host       only consider snapshots for this host, when no snapshot ID is given
  -i, --ignore-case     ignore case for pattern
  -l, --long            use a long listing format showing size and mode
  -N, --newest string   newest modification date/time
  -O, --oldest string   oldest modification date/time
      --path path       only consider snapshots which include this (absolute) path, when no snapshot-ID is given
  -s, --snapshot id     snapshot id to search in (can be given multiple times)
      --tag taglist     only consider snapshots which include this taglist, when no snapshot-ID is given (default [])
```

Now, the pattern is meant to match file/directory names from the snapshot. We could add an option, e.g. `--id`, so that the argument isn't interpreted as a pattern, but as an ID. restic can then figure out if it's a tree, data blob or pack file. It already has that information readily available in the index. So the usage would be:

```
$ restic find --host foo --id 9ccd960cbd159b310aecaf0332b3947dc1a8d1dca2db75b8516f471b17d4264d
```

Alternatively we can add `--tree`, `--data` and `--pack` as options.

What are your thoughts?

McKael:
I think I can give it a try, with only the `--id` flag and I'll see what I can come up with.

IIRC I had not been able to find out how to get the type of an object with its short ID (maybe I don't remember correctly, it was a month ago...), would you have a hint for that?

fd0:
> IIRC I had not been able to find out how to get the type of an object with its short ID (maybe I don't remember correctly, it was a month ago...), would you have a hint for that?

Have a look at the methods for the `Index` data structure, which the `repository.Index()` method returns: https://godoc.org/github.com/restic/restic/internal/repository#Repository.Index and https://godoc.org/github.com/restic/restic/internal/restic#Index, especially `Has()`. This gives you the information if the ID is a data or tree blob. For other types you need to query the backend, via `repository.Backend().Test()` https://godoc.org/github.com/restic/restic/internal/restic#Backend

Please let me know if you need more help!

mholt:
This is cool! Would this, by extension, allow us to query restic like this: "Which snapshots contain _this file_?" - so that you can examine revisions of a file over time, and make it easier to restore a specific revision of a specific file? (I realize this command requires tree IDs and not file paths but let's pretend we can get the tree ID from the file path.)

McKael:
@mholt Isn't this what `restic find PATH` already does?

mholt:
@McKael Last I checked, it didn't work the way I thought, or wasn't turning up any results, but it's been a month or two since I've tried. You might be right. I'll try again...

McKael:
@fd0 Thanks for the details -- I haven't started working on it again, but I still plan to do it (soon).

BTW, from the godoc links I'm unsure those commands (e.g. `Has()`) can use short ID forms, which was actually my issue.  Anyways, I'll dig deeper when I update my patch!

McKael:
@mholt Seems to work for me, I've just checked:
```
% restic find restic
repository 24a9dde4 opened successfully, password is correct
Found matching entries in snapshot 4ed7b134
/bin/restic
```

(It's an old backup, that's why it shows a relative path)

McKael:
The feature is now merged into the find command.

Only 1 ID can be passed at the moment, and the type has to be provided (I could not figure out an easy way to figure it out cheaply, I'll work on that next).

fd0:
I've had a look, this is almost done, but it needs a bit of work:
 * Make `--json` work (currently it breaks the output)
 * Find short blob IDs
 * I'd like to propose dropping the suffix for the options, make it `--blob` instead of `--blob-id`, making the command line easier to read: `restic find --blob 592290a9d3e996142cf7899caebd3ea8b4c8164b9e3eacf37218d44d0a91cca1`

Let me know if you need help with anything.

fd0 (changes requested):
I've had a look, this is almost done, but it needs a bit of work:
 * Make `--json` work (currently it breaks the output)
 * Find short blob IDs
 * I'd like to propose dropping the suffix for the options, make it `--blob` instead of `--blob-id`, making the command line easier to read: `restic find --blob 592290a9d3e996142cf7899caebd3ea8b4c8164b9e3eacf37218d44d0a91cca1`

Let me know if you need help with anything.

McKael:
@fd0 Thanks for the feedback

Yes I plan to fix JSON output before it's merged.

Regarding the short blob IDs, it should work (I got it working locally, but maybe I haven't pushed the latest changes yet).

I'll update the flags as proposed, it looks better indeed.

I'll try to work on it before the end of the week.  Thanks again for the comments!

fd0:
Cool, don't worry about the time, I'm a bit swamped and only slowly working through all the notifications and issue that have piled up the last couple of weeks ;)

McKael:
@fd0 I have pushed a few patches with the following changes:

* flags are now `--blob` and `--tree` instead of `--blob-id`, etc. as requested
* the JSON format is supported for all output (except `--show-pack-id` at the moment)
* several file patterns or restic IDs can now be provided
* walk.Walker now passes the parent tree ID to the WalkFunc function, as discussed on IRC (this is used in the blob search result to display the tree-id a blob belongs to)
* I've added a few usage examples

I'll clean up the patches later (I think leaving the commits as they are is more convenient to review what's new since your previous review).

McKael:
I've rebased and updated @mholt `stats` command to use the modified walker.Walk function.

@fd0 I think I've done the requested changes, can you have a look and let me know if I'm on the right track?

Regarding the WalkFunc change, does it seem reasonable to you?
If so, I can make a different PR for this change if you like.

fd0:
This looks really good already! I'm not a huge fan of adding more and more parameters to the walk func, but I can see value in having the parent tree ID at hand. Since `Walk()` operates on the restic data structures on a high level, I'd prefer to pass a `*restic.ID` instead of a string. I'll push a commit, then you can have a look.

Last thing that needs to be done now is squashing the commits to a reasonable amount, then I'll merge this!

fd0 (approved):

fd0:
I've rebased this PR onto master and added a commit, please have a look!

McKael:
Changelog file added and commits squashed as requested!

Let me know if anything's missing...

PR #1782, opened 2018-05-16
Add S3 file creds and reorder creds chain

skriss:

1. Adds support for an AWS credentials file, which is part of the default AWS credentials chain.
2. Reorders the S3 backend's credentials chain to be standard (static -> env vars -> files -> remote)

Note that my main need is #1; I'd be open to deferring #2 if there's some specific reason why the standard order for the credentials chain shouldn't be used.

Also, there are currently no tests for the credentials chain, so please provide guidance if it's necessary to add something here.

No.

### Checklist

### DIFF
--- a/changelog/unreleased/pull-1782
+++ b/changelog/unreleased/pull-1782
@@ -1,0 +1,7 @@
+Enhancement: Use default AWS credentials chain for S3 backend
+
+Adds support for file credentials to the S3 backend (e.g. ~/.aws/credentials),
+and reorders the credentials chain for the S3 backend to match AWS's standard,
+which is static credentials, env vars, credentials file, and finally remote.
+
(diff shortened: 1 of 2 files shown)


fd0:
LGTM, although I can't test this as I don't have any way to test all the different options. Is this now in "standard order"?

@harshavardhana would you mind having a look please? Thanks! :)

skriss:
For ref. here is the aws-go-sdk implementation of the default chain:

https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L92

As you can see, it goes: env vars, "shared" (which is the file), and finally remote. This matches what I've implemented for restic, while retaining "static" credentials as the first provider for testing purposes.

fd0 (approved):
Ok, fine with me.

Issue #1819, opened 2018-05-26
Background check fails when no pty, and then assumes foreground and uses terminal


donbowman:
## Output of `restic version`

restic 0.9.0 (v0.9.0-0-ge4019194) compiled with go1.10.2 on linux/amd64

## How did you run restic exactly?

/usr/bin/restic backup --verbose=9 --hostname=foobar /etc

RESTIC_PASSWORD=XXX
RESTIC_REPOSITORY=rest:https://don:XXX@backup.XXX/don

open repository
repository e0167288 opened successfully, password is correct
lock repository
load index files
using parent snapshot 74313129
start scan on [/etc]
start backup on [/etc]
panic: runtime error: slice bounds out of range

goroutine 7 [running]:
github.com/restic/restic/internal/ui/termstatus.(*Terminal).SetStatus(0xc420026de0, 0xc4200667a0, 0x1, 0x1)
	src/github.com/restic/restic/internal/ui/termstatus/status.go:310 +0x319
github.com/restic/restic/internal/ui.(*Backup).update(0xc420227c70, 0x1, 0x0, 0x4e, 0x0, 0x0, 0x0, 0x0, 0xc420062d00, 0x0)
	src/github.com/restic/restic/internal/ui/backup.go:181 +0x69e
github.com/restic/restic/internal/ui.(*Backup).Run(0xc420227c70, 0xc06fc0, 0xc420071240, 0x0, 0x0)
	src/github.com/restic/restic/internal/ui/backup.go:139 +0x5d5
main.runBackup.func2(0x0, 0x0)
	src/github.com/restic/restic/cmd/restic/cmd_backup.go:360 +0x75
github.com/restic/restic/vendor/gopkg.in/tomb%2ev2.(*Tomb).run(0xc42001ca50, 0xc42000b100)
	src/github.com/restic/restic/vendor/gopkg.in/tomb.v2/tomb.go:163 +0x2b
created by github.com/restic/restic/vendor/gopkg.in/tomb%2ev2.(*Tomb).Go
	src/github.com/restic/restic/vendor/gopkg.in/tomb.v2/tomb.go:159 +0xb9

## What backend/server/service did you use to store the repository?

rest server

## Expected behavior

In internal/ui/termstatus/background_linux.go
we have this function IsProcessBackground(). It calls syscall.TIOCGPGRP.

I am running w/o a pty and this is failing. But the function assumes this means !background,
so the terminal stuff still gets called.

Maybe assume background if this fails.

[...]

PR #1820, opened 2018-05-27
termstatus: Fix panic for non-terminal runs

### LINKED ISSUE #1803: Backup fails to continue with `panic: runtime error: slice bounds out of range`
## Output of `restic version`
`restic 0.9.0 compiled with go1.10.2 on linux/amd64`

```
export B2_ACCOUNT_ID=XXX
export B2_ACCOUNT_KEY=XXX
export RESTIC_PASSWORD=XXX
restic --repo b2:XXX:/Data backup /XXX/Data
repository XXX opened successfully, password is correct
panic: runtime error: slice bounds out of range

goroutine 34 [running]:
github.com/restic/restic/internal/ui/termstatus.(*Terminal).SetStatus(0xc42016e5a0, 0xc4204280a0, 0x1, 0x1)
	src/github.com/restic/restic/internal/ui/termstatus/status.go:310 +0x319
github.com/restic/restic/internal/ui.(*Backup).update(0xc42028d450, 0x1, 0x0, 0xd306c00, 0x0, 0x0, 0x0, 0x0, 0xc42058fd00, 0x0)
	src/github.com/restic/restic/internal/ui/backup.go:181 +0x69e
github.com/restic/restic/internal/ui.(*Backup).Run(0xc42028d450, 0xc06fa0, 0xc4200b5200, 0x0, 0x0)
	src/github.com/restic/restic/internal/ui/backup.go:139 +0x5d5
main.runBackup.func2(0x0, 0x0)
	src/github.com/restic/restic/cmd/restic/cmd_backup.go:360 +0x75
github.com/restic/restic/vendor/gopkg.in/tomb%2ev2.(*Tomb).run(0xc4202402d0, 0xc4202383e0)
	src/github.com/restic/restic/vendor/gopkg.in/tomb.v2/tomb.go:163 +0x2b
created by github.com/restic/restic/vendor/gopkg.in/tomb%2ev2.(*Tomb).Go
	src/github.com/restic/restic/vendor/gopkg.in/tomb.v2/tomb.go:159 +0xb9
```

Backblaze

## Expected behavior
Backup continues from where it stopped. Or maybe there could be an error logged to tell me to revert to the previous version of restic to finish the backup.

## Actual behavior
Backup cannot continue.

I had a backup running for a week (uploading 350GB to B2 via a crappy ADSL2 line) on restic 0.8.3 and after a reboot I updated to 0.9.0 and tried to continue my backup, unfortunately it now fails with this error.

I'm guessing that backup that started on 0.8.3 and that have not yet completed cannot be resumed with the newer restic version.

Not sure if continuing the backup between version is even thinkable...

Really liking the simplicity of restic! It is a great replacement for the convenient-but-bloated CrashPlan home!

fd0:
Closes #1803

### DIFF
--- a/internal/ui/termstatus/status.go
+++ b/internal/ui/termstatus/status.go
@@ -290,6 +290,20 @@ func (t *Terminal) Errorf(msg string, args ...interface{}) {
 	t.Error(s)
 }
 
+// truncate returns a string that has at most maxlen characters. If maxlen is
+// negative, the empty string is returned.
+func truncate(s string, maxlen int) string {
+	if maxlen < 0 {
+		return ""
+	}
+
+	if len(s) < maxlen {
+		return s
+	}
+
+	return s[:maxlen]
+}
+
 // SetStatus updates the status lines.
 func (t *Terminal) SetStatus(lines []string) {
 	if len(lines) == 0 {
@@ -297,7 +311,7 @@ func (t *Terminal) SetStatus(lines []string) {
 	}
 
 	width, _, err := getTermSize(t.fd)
-	if err != nil || width < 0 {
+	if err != nil || width <= 0 {
 		// use 80 columns by default
 		width = 80
 	}
@@ -305,11 +319,7 @@ func (t *Terminal) SetStatus(lines []string) {
 	// make sure that all lines have a line break and are not too long
 	for i, line := range lines {
 		line = strings.TrimRight(line, "\n")
-
-		if len(line) >= width-2 {
-			line = line[:width-2]
-		}
-		line += "\n"
+		line = truncate(line, width-2) + "\n"
 		lines[i] = line
 	}
 
--- a/internal/ui/termstatus/status_test.go
+++ b/internal/ui/termstatus/status_test.go
@@ -1,0 +1,32 @@
+package termstatus
+
+import "testing"
+
+func TestTruncate(t *testing.T) {
+	var tests = []struct {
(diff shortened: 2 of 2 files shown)


donbowman:
I think this fix is only part of it.
When running from e.g. systemd, cron, or docker w/o -it, it assumes the input/output is a terminal, which its not. In my case I was running from crond in busybox in docker container in kubernetes, and restic would hang (also would get the panic).

The bit I pointed out in #1819 is also incorrect, it assumes 'foreground' if the ioctl fails, which is probably false.

You might consider a flag to disable the 'status' output, its not useful in all environments.

fd0:
Ah, that's not the case, the function `IsProcessBackground()` is only used when restic was started with a terminal in the foreground, and then moved to the background in the shell with `^Z` and `bg`. I think the behavior is correct.

The code which detects if the output is a terminal is here:

https://github.com/restic/restic/blob/0183fea926f6fe801d20f53d3c677ee0bd86d6e2/internal/ui/termstatus/status.go#L67

On Linux, this is the function:

https://github.com/restic/restic/blob/0183fea926f6fe801d20f53d3c677ee0bd86d6e2/internal/ui/termstatus/terminal_unix.go#L26-L28

@donbowman would you mind testing if the issue still occurs with the fix for this issue merged? Please leave a comment (or open a new issue) if it happens again, because then it's clearly a different issue than this one. Thanks!

PR #1855, opened 2018-06-20
Allows saving files/dirs on different file systems together with `--one-file-system`.

### LINKED ISSUE #1854: Backing up multiple filesystems at different mountpoints fails with --one-file-system (-x)
## Output of `restic version`
restic 0.9.0 compiled with go1.10.2 on linux/amd64

restic -r sftp:backup:/restic backup -x -e "/mnt/Z/backups/dir1" / /mnt/Z/backups

sftp

## Expected behavior
I expected / and /mnt/Z/backups to be backed up, but /mnt/Z/backups/dir1 to be excluded from the backup. Getting a snapshot listing shows /mnt/Z/backups, but nothing from that path is visible in a listing (ls -l) of the snapshot.

## Actual behavior
Restic backed up /, without traversing any other mounted filesystems - including /mnt/Z/backups.

Run restic with multiple filesystems as command line arguments and --one-file-system. I don't believe excluding a directory has anything to do with the issue at hand.

I believe restic is using the first path on the backup command line as its source filesystem for checking against with --one-file-system instead of treating each path on the command line as its own source filesystem for traversal of that path.

No. In the past I'd backed up all paths individually but now with fixes to 0.9.0 I wanted to back them all up in a single command line. I can return to backing things up individually but that's a workaround.

I love restic!

fd0:
Unroll archive trees so that files/dirs on a different file system can be saved together with `--one-file-system`.

Closes #1854

### DIFF
--- a/changelog/unreleased/issue-1854
+++ b/changelog/unreleased/issue-1854
@@ -1,0 +1,16 @@
+Bugfix: Allow saving files/dirs on different fs with `--one-file-system`
+
+restic now allows saving files/dirs on a different file system in a subdir
+correctly even when `--one-file-system` is specified. 
+
+The first thing the restic archiver code does is to build a tree of the target
+files/directories. If it detects that a parent directory is already included
+(e.g. `restic backup /foo /foo/bar/baz`), it'll ignore the latter argument.
+
+Without `--one-file-system`, that's perfectly valid: If `/foo` is to be
+archived, it will include `/foo/bar/baz`. But with `--one-file-system`,
+`/foo/bar/baz` may reside on a different file system, so it won't be included
+with `/foo`.
+
+https://github.com/restic/restic/issues/1854
+https://github.com/restic/restic/pull/1855
--- a/internal/archiver/tree.go
+++ b/internal/archiver/tree.go
@@ -202,30 +202,53 @@ func (t Tree) String() string {
 // formatTree returns a text representation of the tree t.
 func formatTree(t Tree, indent string) (s string) {
 	for name, node := range t.Nodes {
-		if node.Path != "" {
-			s += fmt.Sprintf("%v/%v, src %q\n", indent, name, node.Path)
-			continue
-		}
-		s += fmt.Sprintf("%v/%v, root %q, meta %q\n", indent, name, node.Root, node.FileInfoPath)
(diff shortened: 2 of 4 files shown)


alphapapa:
@fd0 I'm confused, because this "fix" seems to nullify the `--one-file-system` option.  What is its intended purpose now, if it doesn't do what the [manual](https://restic.readthedocs.io/en/stable/040_backup.html?highlight=--one-file-system) currently says?  i.e.:

> By specifying the option `--one-file-system` you can instruct restic to only backup files from the file systems the initially specified files or directories reside on. For example, calling restic like this won’t backup `/sys` or `/dev` on a Linux system:

> `$ restic -r /srv/restic-repo backup --one-file-system /`

I notice that the "I have added documentation for the changes (in the manual)" item in the checklist is unchecked, as well.

I think it's very important that this option be properly documented, because users who misunderstand and misuse it may not backup the data they intend to, leading to data loss.

fd0:
I'm sorry, you're right, the text is a bit confusing. What this PR does is correct restic's behavior so that `--one-file-system` works as advertised. I'll try to explain it a bit more verbose. The manual says:

> By specifying the option --one-file-system you can instruct restic to only backup files from the file systems the initially specified files or directories reside on.

So, let's say you have different file systems at `/` and `/srv/data`. Running `restic backup --one-file-system /` will not descend into `/srv/data` because it's on a different file system than `/`. That's the easy case.

When running `restic backup --one-file-system / /srv/data/webserver`, restic should backup everything in `/`, but only the directory `/srv/data/webserver` and nothing else in `/srv/data` (on the other file system).

The bug now was that restic saw that you have `/` in your list of backup targets, and `/srv/data/webserver` is a subdir of `/`, so it would drop the extra target `/srv/data/webserver`. When both reside on the same file system or `--one-files-system` is not specified, that's a valid thing to do. In our case here, saving `/` with `--one-file-system` would not automatically also save `/srv/data` because that's on a different file system. So you'd end up with a snapshot not containing one of the directories you explicitly specified. That's not okay.

This PR corrects the behavior so that it works as users expect it to work.

> The word "on" could mean either "from" or "to." I guess it's intended to mean "from", i.e. "backing up files from a filesystem," rather than "to", i.e. "backing up files to a filesystem."


Indeed, "from" a file system. I don't think it's so ambiguous, we only talk (in the docs and on the website) about saving data to a repository, not to a file system. But I can see where you're coming from.

alphapapa:
Thanks, that explanation is very clear and helpful!

PR #1856, opened 2018-06-21
Ignore ExcludeOtherFs if Stdin is true

### LINKED ISSUE #1807: restic --one-file-system --stdin fails with "zero allowed devices"
Since upgrading from 0.8.3 to 0.9.0, backups that receive a database dump via the `--stdin` options fail with the error "zero allowed devices" (and a Go backtrace). This is a minor regression.

## Output of `restic version`

`restic 0.9.0 compiled with go1.10.2 on linux/amd64` (binary downloaded from Github)

The command is run as part of a larger script that runs the same "base command" (everything up to the `--stdin` below) for a number of paths. The exact command that fails is the first one that uses `--stdin`, below. This has happened with both B2 and sftp backends (set using `RESTIC_REPOSITORY`, B2 credentials set accordingly).

`restic backup --tag "SCR:backup-restic.sh" --tag "VER:3" --tag "CFG:backup-restic.conf" --one-file-system --stdin --stdin-filename postgres.sql`

The output:
```
zero allowed devices
main.gatherDevices
      src/github.com/restic/restic/cmd/restic/exclude.go:204
main.rejectByDevice
      src/github.com/restic/restic/cmd/restic/exclude.go:212
main.collectRejectFuncs
      src/github.com/restic/restic/cmd/restic/cmd_backup.go:194
main.runBackup
      src/github.com/restic/restic/cmd/restic/cmd_backup.go:376
main.glob..func2
      src/github.com/restic/restic/cmd/restic/cmd_backup.go:54
      src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:756
      src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:846
      src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:794
main.main
      src/github.com/restic/restic/cmd/restic/main.go:86
runtime.main
      /usr/local/go/src/runtime/proc.go:198
runtime.goexit
      /usr/local/go/src/runtime/asm_amd64.s:2361
pg_dumpall: pg_dump failed on database "stdiscosrv", exiting
```

Issue exhibited with both B2 and SFTP backends.

## Expected behavior

I expect restic to act as it did in 0.8.3, backing-up the data presented to it on stdin to the backend file named by `--stdin-filename`

## Actual behavior

(See above for error and Go backtrace)

I have observed this behaviour on at least two separate hosts since upgrading to 0.9.0, both backing-up to a mix of 3 different repositories (2 backends).

It seems that `--one-file-system`, which used to be (and should be) a no-op with `--stdin` is now failing because no directories were passed and thus it was unable to determine which devices to include/exclude during filtering. If I run the restic command directly (without the `--one-file-system` my script prepended) the backup completes.

Not yet on restic's part; I haven't been able to trace this change in behaviour back to a commit yet. For my part, I need to update my script to not use `--one-file-system` when backing-up `--stdin`

It usually does... 0.9.0, not so much.

stkw0:
Closes: #1807

Ignore one-file-system flag if stdin is used, as described on the issue:
> --one-file-system, which used to be (and should be) a no-op with --stdin

Seems that there wasn't any test for the ExcludeOtherFs option, and as this is a one line patch I'm not sure if it's needed to add tests for that.

As this is a minor regression fix, I didn't added new documentation nor updated the changelog either.

Yes, issue #1807

### Checklist

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -190,7 +190,7 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 // from being saved in a snapshot
 func collectRejectFuncs(opts BackupOptions, repo *repository.Repository, targets []string) (fs []RejectFunc, err error) {
 	// allowed devices
-	if opts.ExcludeOtherFS {
+	if opts.ExcludeOtherFS && !opts.Stdin {
 		f, err := rejectByDevice(targets)
 		if err != nil {
 			return nil, err


fd0 (commented):
LGTM, thanks!

PR #1861, opened 2018-06-24
Fix find -i (case-insensitive search)

McKael:

`restic find -i PATTERN` seems broken in tip, case-insensitive search does not work anymore.

Issue https://github.com/restic/restic/issues/1862

### Checklist


fd0 (approved):
LGTM, thanks for resolving this bug! If you're done I can merge this as is, we just need to add a changelog file...

fd0 (approved):

PR #1898, opened 2018-07-19
Fix issue 1896 and refactor rclone backend to support multiple REST-over-stdio backends (e.g. SSH)

### LINKED ISSUE #1896: Suspicious calling code for rclone backend
https://github.com/restic/restic/blob/master/cmd/restic/global.go#L652

`rclone.Open` is called here instead of `rclone.Create`

This additionally means that `rclone.Create`is unused code (except for tests).

`rclone.Open` and `rclone.Create` contain almost identical code.  The main difference is that `rclone.Create` calls `rest.Create` and `rclone.Open` calls `rest.Open`.  

There are a few other minor discrepancies that should probably be resolved (e.g., via a common helper function): https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L254 vs https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L282 and https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L260 vs https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L288

@fd0

mpsrig:

Fixes #1896 
Refactors `rclone/backend.go` to contain only rclone-specific code and move the general REST-over-HTTP2-over-stdio code to a new package `rest_stdio_http2`

#1896 
Also, the future use of REST-over-stdio to implement a future SSH backend was discussed in #1561 

### Checklist

- [x] I have added tests for all changes in this PR (none needed)
- [x] I have added documentation for the changes (in the manual) (none needed)

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -649,7 +649,7 @@ func create(s string, opts options.Options) (restic.Backend, error) {
 	case "rest":
 		return rest.Create(cfg.(rest.Config), rt)
 	case "rclone":
-		return rclone.Open(cfg.(rclone.Config), nil)
+		return rclone.Create(cfg.(rclone.Config), nil)
 	}
 
 	debug.Log("invalid repository scheme: %v", s)
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/backend/rest"
+	"github.com/restic/restic/internal/backend/rest_stdio_http2/stdio_conn"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/limiter"
@@ -32,11 +33,11 @@ type Backend struct {
 	waitCh     <-chan struct{}
 	waitResult error
 	wg         *sync.WaitGroup
-	conn       *StdioConn
+	conn       *stdio_conn.StdioConn
 }
 
 // run starts command with args and initializes the StdioConn.
-func run(command string, args ...string) (*StdioConn, *exec.Cmd, *sync.WaitGroup, func() error, error) {
+func run(command string, args ...string) (*stdio_conn.StdioConn, *exec.Cmd, *sync.WaitGroup, func() error, error) {
 	cmd := exec.Command(command, args...)
 
 	p, err := cmd.StderrPipe()
@@ -74,11 +75,7 @@ func run(command string, args ...string) (*StdioConn, *exec.Cmd, *sync.WaitGroup
 		return nil, nil, nil, nil, err
 	}
 
-	c := &StdioConn{
-		stdin:  stdout,
-		stdout: stdin,
-		cmd:    cmd,
-	}
+	c := stdio_conn.New(stdin, stdout)
 
 	return c, cmd, &wg, bg, nil
 }
@@ -86,7 +83,7 @@ func run(command string, args ...string) (*StdioConn, *exec.Cmd, *sync.WaitGroup
 // wrappedConn adds bandwidth limiting capabilities to the StdioConn by
 // wrapping the Read/Write methods.
 type wrappedConn struct {
-	*StdioConn
+	*stdio_conn.StdioConn
 	io.Reader
 	io.Writer
 }
@@ -99,7 +96,7 @@ func (c wrappedConn) Write(p []byte) (int, error) {
 	return c.Writer.Write(p)
 }
 
-func wrapConn(c *StdioConn, lim limiter.Limiter) wrappedConn {
+func wrapConn(c *stdio_conn.StdioConn, lim limiter.Limiter) wrappedConn {
 	wc := wrappedConn{
 		StdioConn: c,
 		Reader:    c,
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -1,120 +1,22 @@
 package rclone
 
 import (
-	"bufio"
-	"context"
-	"crypto/tls"
-	"fmt"
-	"io"
-	"math/rand"
-	"net"
-	"net/http"
-	"net/url"
-	"os"
-	"os/exec"
-	"sync"
 	"time"
 
 	"github.com/restic/restic/internal/backend"
-	"github.com/restic/restic/internal/backend/rest"
-	"github.com/restic/restic/internal/backend/rest_stdio_http2/stdio_conn"
 	"github.com/restic/restic/internal/debug"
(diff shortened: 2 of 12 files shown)


fd0:
Uh, wow, I'm flattered that you're trying this, but can we please take a step back and discuss it first? This change is too big already for review...

mpsrig:
Sure - let me know what you'd like to see here.

My motivation is to get a proper SSH backend working as you had suggested in #1561 . I had the same idea after reading about the rclone implementation on restic.net.

I'm on a time crunch to decide on a new backup strategy (e.g. restic vs borg), and my goal is to get a SSH backend working soon for a fair comparison to Borg when using SSH.

fd0:
FYI: the rclone backend can be used over SSH already by setting `-o 'rclone.program=ssh user@host rclone'` (see [the manual](https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#other-services-via-rclone)).

mpsrig:
Yes, thank you for the link! I did see that, which is what has prompted the work so far.

mpsrig:
I'm hoping to take rclone out of the loop and use restic/rest-server.  I also have some work done on that to add the http2-over-stdio piece.

fd0:
So, first I'd like to understand what you're trying to do. Reworking the `Open/Create` issue you described in #1896 is a small change, it's likely we can remove `Create` and replace it with `Open` without any loss of functionality.

So, why move stuff around at all?

fd0:
Hm. I must admit I don't see much value in adding this functionality. Why not use rclone, or the REST server via ssh TCP forward, or sftp for that matter?

You see, all features/functions which get implement end up having to be maintained by someone (me, at the moment), this causes work. I've found an article which explains this in a great way, here: http://neugierig.org/software/blog/2018/07/options.html

So for all new features we need to evaluate whether the long-term cost is worth it. And for adding a REST-over-stdio backend I'm not convinced it adds enough value. :)

mpsrig:
The motivation here is to follow the Single Responsibility Principle and increase modularity.  The rclone backend code was mixing rclone-specific code (such as the default argument setting, etc) with general code.  

If you would prefer that all the files continue to live under the `internal/backend/rclone` package, I think that would be ok too. I don't care so much about the go package naming.  I just wanted to separate out the reusable code into separate functions. 

I understand you may not wish to maintain an SSH backend, but I'd like to do so in a fork if not.  Having an SSH backend is important to be able to use this tool in a comparable way to rsync (or borg) from an operational/networking/firewall perspective.  There are some environments where doing TCP forwarding is not realistic, plus it adds significant overhead. I also wish to avoid SFTP for various reasons, including to potentially take advantage of the `--append-only` mode.

fd0:
Would you mind elaborating what an "SSH backend" would be, besides using SSH as the transport?

mpsrig:
As a "version 1" I would like to see something similar to the rclone backend, e.g. the child process being "ssh user@host rest-server --listen=stdio" and communicating using the REST API over HTTP2 over the stdio channel over SSH.  

FWIW this last part is extremely similar to how rsync communicates (stdio over SSH).

As a "version 2", I think it would be great to incorporate the relevant parts of rest-server into a restic binary e.g. "ssh user@host restic --stdio-server" which is again really similar to rsync where the same binary is used for client & server.  But this part is not so important to me.

fd0:
Okay, thanks for the explanation!~ What's the argument against using rclone as this server program running remotely on the server?

fd0:
So, let's say you add code to the REST server to talk to it via stdio, then you can use restic with the rclone backend as is, via `-o rclone.program='ssh user@server rest-server --stdio'`. I agree that it's not that convenient (it'll get better when we'll add a config file), why add a separate backend to restic?

mpsrig:
IMHO SSH is really important and I just think it should be a first class citizen in restic.  Additionally in your example command it would be necessary to add `-o rclone.arguments=''` because it would otherwise pass `"serve", "restic", "--stdio", "--b2-hard-delete", "--drive-use-trash=false"`.  So the point is that there is too much working around the restic-specific code from the CLI to do it this way as a production solution IMHO.

fd0:
Point taken, it's inconvenient. I agree. Here are my thoughts so far:

As you saw in the other issue, I have plans for adding a "server mode" similar to what borg does, which will support more intelligent commands such as repacking and indexing on the server. Much more than the REST server and the protocol does, and I'm fairly sure it'll be something different than the REST protocol.

I'd like this to be the `ssh` backend, so I'd like to not use the name for something else. In this case it'd be for something which can be achieved with an existing backend (albeit in an inconvenient way) and it would not really add new functionality, just another way to use the REST backend. Once introduced, removing backends will cause existing installations for our users to stop working, so I'm very reluctant to add new backends in the first place. The same goes for renaming backends.

I think a fork which adds an `ssh` backend would be unfortunate, at the very least it would confuse users and cause more work for us both.

Can you live with the inconvenience of having to specify the program and arguments for the rclone backend for the time being? Do you have ideas for improving the user interface (besides adding a new backend)?

One other option would be to add this to one of the wrappers around restic...

mpsrig:
It is great to hear you have plans for a full featured ssh backend.  I must have misunderstood the comments on the other issue; it seemed like you were discussing exactly my proposal as the future ssh backend.  

I am a bit confused why you would wish it to be radically different from the REST backend.  I thought that using HTTP2 as a framing protocol - even though not over TCP - is really a brilliant way to go; it gets away from the problems faced by home-grown wire formats (see https://rsync.samba.org/how-rsync-works.html, specifically "The Rsync Protocol" which includes "It unfortunately makes the protocol extremely difficult to document, debug or extend"). 

It seems like the additional remote features (remote repacking, etc) could be implemented as iterations on the REST protocol using one of the standard API versioning schemes out there.

mpsrig:
Regardless, I'd like to get (some form of) this PR merged because it seems like there is a real issue of #1896  with `rclone.Create` never being called. This means that `rest.Create` is never called, which means the `create=true` parameter is never sent.

I am not sure why dividing the code into more files is a bad thing? I understand you do not want to expose a new backend to the CLI; feel free to propose a directory layout you would be more comfortable with.

mpsrig:
Also, I think the code coverage report above is incorrect - it is reporting coverage changes for files that have not been changed...

fd0:
I'll come back to you with comments later, it seems we agree on almost all points (maybe my text was misleading, or the issue you've referred to was on as clearly written as I'd like) ;)

Let's discuss the future SSH backend in a new issue, and make this PR about about solving #1896 only. Can you please push a commit which either removes `Create`, or which uses it? That can then be merged quickly. Thanks!

fd0:
So, would you like to reuse this PR for solving #1896? If not, I'd like to close it.

mpsrig:
Go ahead and close it, I’d start in a new branch anyway.

fd0:
Ok, cool, thanks!

PR #1899, opened 2018-07-20
check: Use --cache-dir argument

### LINKED ISSUE #1880: check command doesn't obey the --cache-dir argument and writes to TMPDIR
## Output of `restic version`

`restic 0.9.1 compiled with go1.10.3 on linux/amd64`

```
restic -v=10 -r . --cache-dir /data/backup/cache --cleanup-cache check
```

local backend

## Expected behavior

Obey to the given `--cache-dir` where it should place the temporary check cache. For people with mounted TMPDIR as tmpfs with fixed size this can be a problem.

## Actual behavior

```
using temporary cache in /tmp/restic-check-cache-219828007
repository 6bc46502 opened successfully, password is correct
load indexes
check all packs
no errors were found
```

Pass the `--cache-dir` and see it storing the check cache in `TMPDIR`

The `--cache-dir` parameter is not used.

Use the `--cache-dir` when given instead of `TMPDIR`.

Yes, because it is written in golang and has a nice commandline interface 👍

garrmcnu:

`check` command now uses a specific cache directory if set using the `--cache-dir` argument.

Closes #1880

### Checklist

### DIFF
--- a/changelog/unreleased/issue-1880
+++ b/changelog/unreleased/issue-1880
@@ -1,0 +1,7 @@
+Bugfix: Use `--cache-dir` argument for `check` command
+
+`check` command now uses a specific cache directory if set using the `--cache-dir` argument.
+
+The `--cache-dir` argument was not used by the `check` command, instead a cache directory was created in the temporary directory.
+
+https://github.com/restic/restic/issues/1880
--- a/changelog/unreleased/issue-1880
+++ b/changelog/unreleased/issue-1880
@@ -1,7 +1,12 @@
 Bugfix: Use `--cache-dir` argument for `check` command
 
-`check` command now uses a specific cache directory if set using the `--cache-dir` argument.
+`check` command now uses a temporary sub-directory of the specified directory
+if set using the `--cache-dir` argument. If not set, the cache directory is
+created in the default temporary directory as before.
+In either case a temporary cache is used to ensure the actual repository is
+checked (rather than a local copy).
 
-The `--cache-dir` argument was not used by the `check` command, instead a cache directory was created in the temporary directory.
+The `--cache-dir` argument was not used by the `check` command, instead a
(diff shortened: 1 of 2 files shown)


fd0 (commented):
Hey, thank you very much for giving this a try! Unfortunately, the logic is not quite correct yet:

fd0 (changes requested):
Hey, thank you very much for giving this a try! Unfortunately, the logic is not quite correct yet. I'll try to explain, it's a bit more complicated than it looks at first sight ;)

`restic check` will, by default, not use the default cache. This is a safety feature: it should check the data in the repo, not the local copy. If users want, they can force restic to use the local cache by specifying `--with-cache`.

Unfortunately, not using a cache at all makes `restic check` really slow, so we're building a new cache in a temporary directory. The bug report mentioned that this is a problem for devices which may not have enough space in the temporary directory (typically `/tmp` on Linux).

So the proposal was: If `--cache-dir` is specified explicitly, use a subdir of that directory for the temporary cache instead of the system-wide temporary directory, but build a new cache from scratch and do not use the existing cache at all. When `--with-cache` is specified, the old behavior (use the existing cache) should be used as before.

What this PR implements is slightly different: when `restic check --cache-dir foo` is called, it uses the cache directory. I expect that users will have set this to their permanent cache location, so this would make `restic check` use the existing cache, which is not what we want. :)

I think we should use the `--cache-dir` given on the command-line as the input to the `ioutil.TempDir()` function, to use it as the temporary directory. If it's empty, great, use the systemd-wide temporary directory. If not, use a new subdir of that directory.

So, it'd be great if you have the motivation to try this again! Please let me know if you need any help!

fd0 (approved):
Great, looks good!

xor-gate:
Nice catch both, thanks for fixing!

PR #1900, opened 2018-07-21
restorer: Add test for restore with include filter

### LINKED ISSUE #1870: Error in restoring a specific file
When I run the command to restore a specific file, the restic return the follow error:
```
ignoring error for /restore11/usr/local/teste2.txt: OpenFile: open /restore11/usr/local/teste2.txt: is a directory
```

## Output of `restic version`
(v0.9.1-33-g44fb2a8) compiled with go1.10.3 on linux/amd64

 restic restore d84bbf60 -t /restore11/ -i "/usr/local/teste2.txt"

S3 ( DigitalOcean )

## Expected behavior

 Restic should not convert the file to directory

## Actual behavior

 
Restoring specific file from a snapshot , restic convert the first file to directory 

```
root@gbox:/# restic restore d84bbf60 -t /restore11/ -i "/usr/local/*.txt"
repository 00d74122 opened successfully, password is correct
restoring <Snapshot d84bbf60 of [/usr/local] at 2018-07-01 19:45:00.798260419 +0000 UTC by root@gbox> to /restore11/
ignoring error for /restore11/usr/local/teste2.txt: OpenFile: open /restore11/usr/local/teste2.txt: is a directory
There were 1 errors
```

Linting the file/directory

```
root@gbox:/# ls -l /restore11/usr/local/
total 12
drwx------ 2 root root 4096 Jul  1 21:41 teste2.txt
```

no

no

fd0:

Resolve an error with the restore command when an include filter is used.

#### Background:

traverseTree() was meant to call enterDir() whenever a directory is selected for restore, either explicitly or implicitly (=contains a file which is to be restored). After restoring a file, leaveDir() is called in reverse order for all intermediate directories so that the metadata can be restored.

When a directory is selected implicitly, the metadata for it is restored. This is different from the previous restorer behavior, which created implicitly selected intermediate directories with permissions 0700 (only user can read/write it).

This PR changes the behavior back to the old one. Only a directory is explicitly selected for restore, enterDir()/leaveDir() are called for it. Otherwise, only visitNode() is called, so visitNode() needs to make sure the parent directory exists. If the directory is explicitly included, leaveDir() will then restore the metadata correctly.

When we decide to change the behavior (restore metadata for all intermediate directories, even if selected implicitly), we should do that in the selection functions, not here.

This finally resolves #1870

Closes #1870

### Checklist

### DIFF
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -99,35 +99,15 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 			return err
 		}
 
-		enteredDir := false
 		if node.Type == "dir" {
 			if node.Subtree == nil {
 				return errors.Errorf("Dir without subtree in tree %v", treeID.Str())
 			}
 
-			// ifedorenko: apparently a dir can be selected explicitly or implicitly when a child is selected
-			// to support implicit selection, visit the directory from within visitor#visitNode
 			if selectedForRestore {
-				enteredDir = true
 				err = sanitizeError(visitor.enterDir(node, nodeTarget, nodeLocation))
 				if err != nil {
 					return err
-				}
-			} else {
-				_visitor := visitor
-				visitor = treeVisitor{
-					enterDir: _visitor.enterDir,
-					visitNode: func(node *restic.Node, nodeTarget, nodeLocation string) error {
-						if !enteredDir {
-							enteredDir = true
-							derr := sanitizeError(_visitor.enterDir(node, nodeTarget, nodeLocation))
-							if derr != nil {
-								return derr
-							}
-						}
-						return _visitor.visitNode(node, nodeTarget, nodeLocation)
-					},
-					leaveDir: _visitor.leaveDir,
 				}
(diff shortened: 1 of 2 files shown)


ifedorenko:
Minor. I think there is no need to precreate `dst` directory in RestoreTo any more. By removing that code all created directories will default to the same `0700` permissions, which is both safer and more consistent. https://github.com/restic/restic/pull/1900/files#diff-7e901d641519cb52e213a5d339ee6d65L200

I also wonder if it'll be more "symmetrical" if directories and other node types were created in `visitNode`. This way it should be possible to remove `enterDir`, which would simplify the code. I can open follow-up PR to illustrate this, if you are interested.

ifedorenko (approved):

fd0:
I think we should either isolate the (fairly complex) `treeVisitor` from the restorer, or replace it with the recently introduced [`walker.Walk()`](https://godoc.org/github.com/restic/restic/internal/walker#Walk). I'll give this a try so you can have a look.

PR #1905, opened 2018-07-28
HTTP reporting for backup+restore, CLI reporting gor restore

CapacitorSet:

During backups and restores, Restic will send progress data (how many files processed, how many bytes, how many folders unchanged, etc.) to an external HTTP server every few seconds. It also implements CLI reporting for restores, in a fashion similar to backups. 

Yes: https://forum.restic.net/t/new-feature-http-reporting-for-backup-restore-cli-reporting-for-restore/824

Checklist
---------

The development of this feature was sponsored by [InAsset](http://www.inasset.it/en/).

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -336,13 +336,17 @@ func findParentSnapshot(ctx context.Context, repo restic.Repository, opts Backup
 }
 
 func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Terminal, args []string) error {
+	p := ui.NewBackup(term, gopts.verbosity, gopts.StatusURL, gopts.StatusTime, gopts.StatusToken)
+
 	err := opts.Check(gopts, args)
 	if err != nil {
+		p.HTTP.Error(err)
 		return err
 	}
 
 	targets, err := collectTargets(opts, args)
 	if err != nil {
+		p.HTTP.Error(err)
 		return err
 	}
 
@@ -350,13 +354,12 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	if opts.TimeStamp != "" {
 		timeStamp, err = time.Parse(TimeFormat, opts.TimeStamp)
 		if err != nil {
+			p.HTTP.Error(err)
 			return errors.Fatalf("error in time option: %v\n", err)
 		}
 	}
 
 	var t tomb.Tomb
-
-	p := ui.NewBackup(term, gopts.verbosity)
 
 	// use the terminal for stdout/stderr
 	prevStdout, prevStderr := gopts.stdout, gopts.stderr
@@ -380,6 +383,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	p.V("open repository")
 	repo, err := OpenRepository(gopts)
 	if err != nil {
+		p.HTTP.Error(err)
 		return err
 	}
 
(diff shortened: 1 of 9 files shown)


mholt:
Interesting -- what about just streaming JSON or something to stdout with the progress data? i.e. why does this change require an HTTP server to read the updates?

CapacitorSet:
That was my first thought when I was assigned this project - "why not just pipe the output to netcat?" - but it turns out that the client needed a cross-platform solution. That's a reasonable constraint.

mholt:
I asked, because I'll probably be proposing a PR in the near-future which streams the status out in JSON (or similar) format, sans HTTP. (My change will include the current file(s) being backed up.)

Is there a way to decouple this PR from HTTP?

CapacitorSet:
Yes, it'd be sufficient to replace [`http_backup.go:34`](https://github.com/restic/restic/pull/1905/files#diff-fd61510df500881ec372b6697bd2cc97R34):

```diff
-_, err = http.Post(h.url, "text/json", bytes.NewReader(j))
-if err != nil {
-	panic(err)
-}
+os.Stdout.Write(j)
```

From there, it's a matter of removing HTTP-specific configuration (URL, token) and renaming functions and variables.

fd0:
Thanks for your contribution! I'll have to think about this for a bit, but at the moment I'm not convinced that HTTP reporting is a good idea. What @Mholt proposed (streaming JSON output) is a much better approach. In addition, the progress reporting for the restorer is a separate feature, it must be submitted in a separate PR. Since we'll rework the restorer completely in #1719, I'd like to hold on for now until that is merged, then we can add reporting properly.

mpetrazzo:
Hi,
I'm @CapacitorSet colleague.
We are developing a restic wrapper that envelop the restic exe, call it at specific time (with a scheduler) and send the output informations to a centralized server.
We use python with pyinstaller for develop and create standalone exe to distribute them to the clients (win, lin, mac).
After a lot of tests we see that on windows and a program that write to console logging (stdout) aren't so much friends: the caller (our program) can correctly read outpout from stdout only in special cases, and not always. This doesn't happen, for example, on linux.
So we thought, and develop, an alternative method for receive and parse informations from restic, that is platform independent and can be use for collect them in localhost or everywhere where there is a http server (for example also in restic rest-server)

Please consider to accept this patch that doesn't interfere with restic, but add a new possibility to parse restic (fantastic) work.

fd0:
Thank you for the praise, I love that restic helps you!

It would have been great to talk about this design before somebody actually spent time implementing it, so maybe we would have found a better solution. While it covers your use case, I fear that it's a very narrow solution for a very specific problem that (at least feels to me) could have been resolved in a way that integrates better with restic. After all, we (as a project) have to live with and maintain this code indefinitely.

Here's a great article about the options dilemma, which explains it very well: 
http://neugierig.org/software/blog/2018/07/options.html

We already have one way of getting machine-readable feedback (JSON printed to stdout) in a way that does not require any network (which is always a good thing). So I don't think it's a good idea to add a second way. Where do we stop? Add support for writing progress/status information to a MySQL database? To MongoDB and CouchDB? What about monitoring such as Nagios? It's a slippery slope adding more and more features that we as a spare-time for-fun project (at least for me) have to maintain for a long time.

Please don't get me wrong: I love that you spent time and resources to improve restic, and that you use it! But let's talk about the design first.

I'd like to start with your original problem, please open a new issue (ideally as a bug report) so we can discuss it. Do you have any hints or background information what's going on when restic is started by a Python process? Why can you just read stdout of the restic process from Python, e.g. linewise? For me, it's the first time that I hear of such problems (I'm not a Windows user myself). Is there anything we can improve from restic's side?

fd0:
I'm closing this PR for now as it cannot be merged. The way forward is to do two different things:
 * For the metric/reporting issue: Open a new issue, describe the use case and the problems with Python reading restic's status output, then we can find out what's going on
 * For the restore CLI feedback: Please hold until #1719 is merged, then maybe submit the PR again, rebased to the merged code.

PR #1912, opened 2018-07-31
Support for different kinds of select functions

### LINKED ISSUE #1909: Reduce unnecessary lstat calls on excluded files to improve scan speed
Output of `restic version`
--------------------------
restic 0.9.1 compiled with go1.10.3 on darwin/amd64

Currently restic runs `lstat` on files even if they're excluded by filename. This can result in unnecessarily long scan times when e.g. a lot of files are excluded by extension. restic needs to run `lstat` on exclude files in certain circumstances, such as when checking for `--one-file-system`, but if the files checked are already excluded by filename there should be no need to run `lstat` on those files. Reducing the use of `lstat` can speed up backup times significantly in some cases.

Quote from @fd0 on [forum.restic.com](https://forum.restic.net/t/restic-runs-lstat-on-files-excluded-by-extension-intended-behaviour/831/4):

> This is an architectural limitation restic currently has: some of the exclude functions (e.g. --one-file-system) need to have the information lstat returns, and so we’re currently running lstat on the files before checking the excludes. The solution would be to have two types of excludes:
> 
>  - The ones which can decide by name/path if a file is to be excluded
>  - The ones which need the lstat information (e.g. --one-file-system)
>
> Until this is implemented, restic runs lstat() on all files. Sorry about that. You could add a bug report to the GitHub issue tracker so we can track improving the exclude function, that’d help.

What are you trying to do?
--------------------------
This issue can significantly increase scan (and therefore backup) times when a lot of files are excluded by filename.

In one case I observed an increase in backup time of > 20 minutes when excluding 1.000.000 `.pb` files using a wildcard exclude pattern `*.pb`, rather than excluding the files by path.

It would be convenient to be able to apply general filename wildcard exclude patterns on a large number of files without having to worry about long backup times.

restic has significantly strengthened my backup strategies and made it much simpler to deploy robust backup in a number of different environments. The general approach to backup and the active community around restic makes it a pleasure to use. Thanks a lot to everyone involved!

askielboe:
Attempt to add support for different kinds of select/reject functions to allow rejecting files by path only. Should speed up scanning and backups. Very much WIP (no tests / no docs).

What it basically does is wrapping the select/reject functions in a struct (`SelectFuncs`) so they can be called separately (`SelectFuncs.SelectByPath` and `SelectFuncs.SelectByInfo`). This allows calling the reject functions that only require a path before calling lstat on the file. So if the file is rejected by path, we save an lstat call.

I kept the reject function type `(string, os.FileInfo)` even for the path-only functions, in order not to change too much of the code. I still think it's fairly readable, but maybe a bit counterintuitive why the path-only-functions would take a file info as input (they ignore it). But that is anyway how it works currently.

I'd love some feedback on this. In my tests is seems to speed up scanning a lot (by a factor of 10), but total backups time are only marginally improved. Not sure what the archiver is doing, but it seems that it's not spending most of its time filtering files.. That raises the question of whether this change is wanted/needed at all..

This is my first time writing anything in `go` and I have no idea if this is a good / bad way of doing things. Any feedback is welcomed!

This issue was raised here: https://github.com/restic/restic/issues/1909

Closes #1909

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1909
+++ b/changelog/unreleased/issue-1909
@@ -1,0 +1,14 @@
+Enhancement: Reject files/dirs by name first
+
+The current scanner/archiver code had an architectural limitation: it always
+ran the `lstat()` system call on all files and directories before a decision to
+include/exclude the file/dir was made. This lead to a lot of unnecessary system
+calls for items that could have been rejected by their name or path only.
+
+We've changed the archiver/scanner implementation so that it now first rejects
+by name/path, and only runs the system call on the remaining items. This
+reduces the number of `lstat()` system calls a lot (depending on the exclude
+settings).
+
+https://github.com/restic/restic/issues/1909
+https://github.com/restic/restic/pull/1912
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -186,18 +186,9 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 	return nil
 }
 
-// collectRejectFuncs returns a list of all functions which may reject data
-// from being saved in a snapshot
-func collectRejectFuncs(opts BackupOptions, repo *repository.Repository, targets []string) (fs []RejectFunc, err error) {
-	// allowed devices
-	if opts.ExcludeOtherFS && !opts.Stdin {
-		f, err := rejectByDevice(targets)
-		if err != nil {
-			return nil, err
-		}
-		fs = append(fs, f)
-	}
-
+// collectRejectByNameFuncs returns a list of all functions which may reject data
+// from being saved in a snapshot based on path only
+func collectRejectByNameFuncs(opts BackupOptions, repo *repository.Repository, targets []string) (fs []RejectByNameFunc, err error) {
 	// exclude restic cache
 	if repo.Cache != nil {
 		f, err := rejectResticCache(repo)
@@ -231,6 +222,21 @@ func collectRejectFuncs(opts BackupOptions, repo *repository.Repository, targets
 			return nil, err
 		}
 
+		fs = append(fs, f)
+	}
+
+	return fs, nil
+}
+
+// collectRejectFuncs returns a list of all functions which may reject data
+// from being saved in a snapshot based on path and file info
+func collectRejectFuncs(opts BackupOptions, repo *repository.Repository, targets []string) (fs []RejectFunc, err error) {
+	// allowed devices
+	if opts.ExcludeOtherFS && !opts.Stdin {
+		f, err := rejectByDevice(targets)
+		if err != nil {
+			return nil, err
+		}
 		fs = append(fs, f)
 	}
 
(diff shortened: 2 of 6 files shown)


fd0:
very cool, thank you very much for giving this a try! It's a great issue for beginners (both Go and restic) ;)

I'll give you some feedback on the general design first: If I were to resolve this issue, I would follow my (and your, apparently) instinct and have two separate function callbacks in the archiver:
```go
// SelectByNameFunc returns true for all items that should be included (files and
// dirs). If false is returned, files are ignored and dirs are not even walked.
type SelectFunc func(item string) bool

// SelectFunc returns true for all items that should be included (files and
// dirs). If false is returned, files are ignored and dirs are not even walked.
type SelectFunc func(item string, fi os.FileInfo) bool
```

This makes it clear that the functions which only operate on a name don't need a

fd0:
Ah, and by the way: the CI tests fail because the Go tests fail, you can run all tests (assuming you've checked out restic at `$GOPATH/src/github.com/restic/restic`) with:
```
$ go test ./...
# github.com/restic/restic/internal/archiver [github.com/restic/restic/internal/archiver.test]
internal/archiver/archiver_test.go:1345:9: undefined: SelectFunc
internal/archiver/scanner_test.go:20:9: undefined: SelectFunc
ok      github.com/restic/restic/cmd/restic     7.081s
FAIL    github.com/restic/restic/internal/archiver [build failed]
ok      github.com/restic/restic/internal/backend       3.695s
[...]
```

To only run the tests in one directory, `cd` into it and use `go test` or `go test -v`:
```
$ cd internal/archiver
$ go test -v
# github.com/restic/restic/internal/archiver [github.com/restic/restic/internal/archiver.test]
./archiver_test.go:1345:9: undefined: SelectFunc
./scanner_test.go:20:9: undefined: SelectFunc
./scanner_test.go:137:11: undefined: SelectFunc
FAIL    github.com/restic/restic/internal/archiver [build failed]
```

fd0:
A follow-up issue would be to move the exclude code into a new package, e.g. `internal/select` or so, and then add tests for all the different excludes. But that's a bit more advanced ;)

askielboe:
I made the changes you suggested and did a minimal fix to make the tests pass.

Do we need any new tests for this?

mholt:
> I'd love some feedback on this. In my tests is seems to speed up scanning a lot (by a factor of 10), but total backups time are only marginally improved. Not sure what the archiver is doing, but it seems that it's not spending most of its time filtering files..

My guess is that most of the cycles of the backup are spent doing the read, encrypt, and write operations. Faster scans and fewer lstats sound good to me, too! :+1:

fd0 (approved):
Awesome, thanks!

fd0:
I've taken the liberty of squashing the two commits and added an entry to changelog. Once the tests give us the green light, I'll merge this PR.

fd0:

askielboe:
yay! thanks for your help. cheers! 🎊

PR #1919, opened 2018-08-01
Update dependencies

fd0:
This updates all vendored dependencies and enables pruning for the `vendor/` folder, which shrinks it by a factor of ten.


mholt:
WOAH.

screenshot from 2018-08-01 12-24-34

That's a lot of code! :innocent:

(Just remarking, not complaining.)

fd0:
restic is standing on the shoulders of (very large) giants :)

fd0:
Hm. Something odd is going on with our vendor folder. I've used `dep ensure -v -update` and the vendor folder is huge: 228 MB in master, 300MB in this branch.

Using Go 1.11beta2 with `go mod -vendor` shrinks this down to 15MB! And restic still builds. Amazing...

mholt:
@fd0 I'm not familiar with the Go 1.11 vendoring changes, but... if the code is no longer there, does restic still build because those libraries are available elsewhere in your GOPATH (outside the vendor folder)?

fd0:
Oh yes, it still builds and the tests run! You can check by running `go run build.go -v -T`, which constructs a temporary `GOPATH` with only the restic repo (including vendor) in it at the right subdir.

fd0:
Aha! `go mod -vendor` prunes tests and test files from the vendored dependencies, from `go help mod`:
> The -vendor flag resets the module's vendor directory to include all packages needed to build and test all the module's packages. It does not include any test code for the vendored packages.

If you haven't had a look at modules, do that! It's awesome :)

fd0:
Pushing the amended commit with pruning, looking much better now!

PR #1920, opened 2018-08-01
Support Go 1.11

fd0:

Support Go 1.11 (which includes support for built-in dependency management).

This includes a bunch of things we (as a project) will do differently:
 * We'll switch from `dep` to Go modules for dependency management, with backwards-compatibility via the `vendor/` folder
 * The output of `gofmt` is not stable and has changed, so we'll modify our integration test to only run `gofmt` from the latest stable version of Go, that's the new requirement

For now, we'll keep using Go in `GOPATH` mode for the integration tests and the `build.go` program. We'll revisit this once Go 1.11 has been released and tested for some time.

nope

TODO
----

 - [x] Get it working for Go 1.11 in non-module mode
 - [x] Switch from `dep` to Go modules, manage the `vendor/` dir with `go mod vendor`
 - [x] Use `go run -mod=vendor run_integration_tests.go' for Appveyor
 - [x] Fix `run_integration_tests.go`


fd0:
Ok, done! It now works as follows:

For users:
 * With Go < 1.11, `go run build.go` builds a new `GOPATH` in a temporary directory and builds restic in it
 * With Go >= 1.11, `go run -mod=vendor build.go` copies the code to a temporary directory and builds it there using the new Go modules

For developers:
 * With Go < 1.11, put restic in the `GOPATH` at `src/github.com/restic/restic`, build in there
 * With Go >= 1.11, put restic outside the `GOPATH` and build it

fd0:
I'm reworking the CI tests when I'm already at it:
 * Remove `gotestcover` for Travis: Go >= 1.10 can now finally write coverage information for more than one package to a single file, so we don't need `gotestcover` any more
 * Enable caching the Go build cache on Travis (Linux, OS X) and Appveyor (Windows), this has the potential of speeding up the CI tests enormously

PR #1932, opened 2018-08-08
cmd: restic: clean up sort.SliceStable usage 

cyphar:
The older API is still available in the Go stdlib (and will remain there
forever, thanks to Go's backwards compatibility guarantees). By
switching back to the pre-1.8 API this allows restic to be built on
older distributions that don't have a newer Go compiler (though it turns
out some of our vendors use the newer API, this is something
distributions can patch out themselves).

It also (in my opinion) makes the code much more readable (since
time.Time implements sort.Interface, there's no need to explicitly
implement it using .After and .Before). We've been carrying this patch
in openSUSE for a couple of releases and nobody seems to have noticed an
issue with it.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -106,9 +106,7 @@ func newFilterLastSnapshotsKey(sn *restic.Snapshot) filterLastSnapshotsKey {
 // they will be joined and treated as one item.
 func FilterLastSnapshots(list restic.Snapshots) restic.Snapshots {
 	// Sort the snapshots so that the newer ones are listed first
-	sort.SliceStable(list, func(i, j int) bool {
-		return list[i].Time.After(list[j].Time)
-	})
+	sort.Stable(list)
 
 	var results restic.Snapshots
 	seen := make(map[filterLastSnapshotsKey]bool)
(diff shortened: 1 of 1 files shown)


fd0:
Hey, thanks for your contribution!

I must admit that I don't particularly like it, because the order of the list now depends on the implementation of the `sort.Interface` of the `restic.Snapshots` type, here:

https://github.com/restic/restic/blob/beb208e1597a4b7d18705c12f045413056eafcdb/internal/restic/snapshot.go#L204-L220

Which is, unfortunately, very far away from the code that uses the list. I think that `restic.Snapshots` still has the functions is more an oversight than a conscious decision 😉

That it works at all has nothing to do with `time.Time` implementing `sort.Interface` (which is not the case).

So, I'm willing to merge this patch, but please be aware that the lowest version of Go that we will support is Go 1.9, mostly because of dependencies (see https://github.com/restic/restic/commit/67933008502a72d579f325584f3566535f63ccf1).

fd0:
Ah, what's the version of Go you're using?

cyphar:
> Ah, what's the version of Go you're using?

This was originally meant to fix building on Go versions as old as Go 1.7, but since writing this patch a few months ago, openSUSE Leap now has Go 1.9 -- meaning this is no longer required for us.

Personally I still prefer using `sort.Interface` because it reads more like idiomatic Go, but if you prefer using `SliceStable` specifically because it isn't "idiomatic" that's also fine. Feel free to close this. :smile_cat:

fd0:
Ok, thanks for the feedback! I'm closing this issue :)

PR #1939, opened 2018-08-10
add JSON output support for restic ls cmd

kitone:

This PR adds support for JSON output for the `restic ls` command, using the existing global `--json` flag.

output truncated of  : `restic ls 1977f34d --json`
```json
[{
  "time": "2018-08-10T13:41:51.624627031+02:00",
  "tree": "0d7878997a8317b052105f933c2ff0920f2942ce61c4842de1b1ab067d305b35",
  "paths": ["/home/test/go/src/github.com/restic/restic/doc"],
  "hostname": "computer.local",
  "username": "test",
  "uid": 502,
  "gid": 20,
  "id": "1977f34d970e318bd3249f59ec8e87d5a8b17551cfb49664c0ba5f0d56e9a97c",
  "short_id": "1977f34d",
  "nodes": [{
    "name": "doc",
    "type": "dir",
    "path": "/doc",
    "mode": 2147484141,
    "mtime": "2018-08-10T13:07:54+02:00",
    "atime": "2018-08-10T13:07:54+02:00",
    "ctime": "2018-08-10T13:07:54+02:00"
  }, {
    "name": "010_introduction.rst",
    "type": "file",
    "path": "/doc/010_introduction.rst",
    "size": 595,
    "mode": 420,
    "mtime": "2018-08-10T13:07:54+02:00",
    "atime": "2018-08-10T13:07:54+02:00",
    "ctime": "2018-08-10T13:07:54+02:00"
  }]
}]
```

No

Checklist
---------


mholt:
Oh -- thank you! This was on my TODO list. How well does this handle listing files in large repos?

cmd/restic/cmd_ls.go
```diff
@@ -61,8 +65,81 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+
+type lsNode struct {
+	Name       string      `json:"name"`
+	Type       string      `json:"type"`
+	Path       string      `json:"path"`
+	Size       uint64      `json:"size,omitempty"`
+	Mode       os.FileMode `json:"mode,omitempty"`
+	ModTime    time.Time   `json:"mtime,omitempty"`
+	AccessTime time.Time   `json:"atime,omitempty"`
+	ChangeTime time.Time   `json:"ctime,omitempty"`
```
mholt: I seem to recall that `omitempty` doesn't work with the zero `time.Time` values -- has that changed?

cmd/restic/cmd_ls.go
```diff
@@ -61,8 +65,81 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+	return printLS(ctx, repo, gopts, opts, args)
+}
+
+type lsSnapshot struct {
+	*restic.Snapshot
+
+	ID      *restic.ID `json:"id"`
+	ShortID string     `json:"short_id"`
+	Nodes   []lsNode   `json:"nodes"`
+}
```
mholt: Since the JSON structures that get printed are different, how about a field that indicates the type of struct it is, which is common between both of them? This way, we can easily differentiate the two in code.

mholt (commented):
This is nice! What is your use case, if I may ask? I'm surprised and curious that someone else needed this feature, too.

cmd/restic/cmd_ls.go
kitone: Don't know, copied from restic.Node, I'm just beginning with go

cmd/restic/cmd_ls.go
mholt: No, I mean a field like this:

```
Type string `json:"type"`
```

With a value of either "snapshot" or "node" so we know what kind of structure to parse out when we decode it.

mholt:
> I was looking / thinking about a web ui for restic

Nice, I'm doing that too: https://relicabackup.com -- we expect to launch at the end of the summer / early fall! This is a little peek at what the UI looks like so far: https://twitter.com/relicabackup/status/1013828975506833408

So, as I understand this change, the JSON output contains all the output that `--long` shows as well, right? If so, that's probably good: I think the JSON output should contain all the info.

mholt:
FYI, my PR #1941 might conflict with this slightly but it shouldn't be too hard to resolve the conflict in either direction.

cmd/restic/cmd_ls.go
kitone: I guess we can but if I may... (sorry I like to understand)
- we already have value type on node for the kind of (file, dir...)
- why the need differentiate the two  (snaphost or node)  in code using a type value ? A Snapshot has a value named "nodes" which is an array of nodes and only nodes. isn't it redundant ?

kitone:
Thanks, nice, restic only miss a web ui !

Yes, most of. The long command list : mode, UID, GID, Size, ModTime, path and target.
I don't list the target on symlink and my last commit add the uid and gid.

cmd/restic/cmd_ls.go
mholt: I guess that's a good point -- in that case, it should be documented how to distinguish the two. I still would prefer a separate field to indicate how to decode the structure, because it makes it so easy then to know how to decode it, rather than using a heuristic like looking for a certain field that only one struct has and the other doesn't. Maybe it could be called `StructType`, or the Type field in the lsNode struct could be renamed to FileType or ItemType or something.

Latest changes look great though!

fd0:
I'd like to wait with this PR until #1941 is merged. Thanks for your contribution!

mholt:
@kitone When you have a chance, sync up with the latest `master` branch :+1: -- looking forward to it! 

(The change you're merging in allows ls to be filtered by directory.)

kitone:
I will be without computer/internet access for a few weeks

kitone:
found time before the flight...
Replaced with https://github.com/restic/restic/pull/1953

Issue #1940, opened 2018-08-10
Drastically speed up `restic ls` by limiting to just a specific folder


mholt:
Output of `restic version`
--------------------------

```
restic 0.9.2 (v0.9.2-0-gbeb208e1) compiled with go1.10.2 on darwin/amd64
```

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I'd be interested in modifying `restic ls` so that it can be scoped to just a specific directory/tree. For my purposes, I want to list the files in a certain folder -- and not subfolders -- without scanning the whole repository.

Scanning the whole repository takes about 5 minutes in my case, and the repo has only about 65 GB.

Scanning just the folder should (in theory) take only a couple seconds, based on my tests.

I am willing to submit a PR, of course. A little guidance would be appreciated about where to start in the code!

Did restic help you or made you happy in any way?
-------------------------------------------------

Yes, as usual. :) You already know it from me! :wink:

fd0:
Good idea. It always bothered me that restic works different from `ls`, which is non-recursive by default. I can imagine the following user interface:
 * `restic ls latest` works as it does not, lists files recursively
 * `restic ls latest /` lists only the root directory
 * `restic ls latest /foo` lists only the contents of `/foo`
 * `restic ls latest --recursive /home/user` lists everything below `/home/user` recursively

The only use case that does not work any more is when multiple snapshot IDs are passed.

The place to start in the code is here: https://github.com/restic/restic/blob/master/cmd/restic/cmd_ls.go#L67

You would need to parse the second argument as a path, returning ignores from the function passed to `Walk()`, and only print things when the path passed to the function is a sub-path of the requested path.

Let me know if you need more hints :)

mholt:
Sounds good! Thanks for taking the time to give feedback/hints!

A couple thoughts before I dive in:

- The Walk function -- it will still traverse the whole tree (albeit quicker since we'll immediately ignore non-matching nodes)?
- The traversal -- I think it is depth-first? This could be more efficient, at least for this particular task, if it's breadth-first. Not sure if that matters, but I figured I'd raise the question.
- The user interface -- what is the use case for listing multiple snapshots on `ls`? If we wanted to preserve that, we could always just use a specific flag for the directory filter.

fd0:
> The Walk function -- it will still traverse the whole tree (albeit quicker since we'll immediately ignore non-matching nodes)?

[...]

PR #1941, opened 2018-08-11
ls: Implement directory filter, optionally subfolders

mholt:

Changes `restic ls` so that the syntax is:

```
restic ls snapshotID [dirs...]
```

instead of multiple snapshot IDs. Adds a new flag, `--recursive`, to allow matching subfolders of the given directories.

A directory filter constrains the walk to just the folder(s) specified in the arguments.

An `ls` that used to take 5 minutes now takes 50 seconds. (Most of that time is connecting to the repo, downloading the index, and decrypting it. The actual scan is instantaneous. It also uses less memory, since there are way fewer results.)

#1940

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -125,7 +125,15 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
 					}
 				}
 				if !walk {
-					return false, walker.SkipNode
+					if node.Type == "dir" {
+						// signal Walk() that it should not descend into the tree.
+						return false, walker.SkipNode
+					}
+
+					// we must not return SkipNode for non-dir nodes because
+					// then the remaining nodes in the same tree would be
+					// skipped, so return nil instead
+					return false, nil
 				}
 
 				// this second iteration ensures that we get an exact match
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"path"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -76,6 +75,40 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
 		}
 	}
 
+	withinDir := func(nodepath string) bool {
+		if len(dirs) == 0 {
+			return true
+		}
+
+		for _, dir := range dirs {
+			// we're within one of the selected dirs, example:
+			//   nodepath: "/test/foo"
+			//   dir:      "/test"
+			if fs.HasPathPrefix(dir, nodepath) {
+				return true
+			}
+		}
+		return false
+	}
+
+	approachingMatchingTree := func(nodepath string) bool {
+		if len(dirs) == 0 {
+			return true
+		}
+
+		for _, dir := range dirs {
+			// the current node path is a prefix for one of the
+			// directories, so we're interested in something deeper in the
+			// tree. Example:
+			//   nodepath: "/test"
+			//   dir:      "/test/foo"
+			if fs.HasPathPrefix(nodepath, dir) {
+				return true
+			}
+		}
+		return false
+	}
+
 	repo, err := OpenRepository(gopts)
 	if err != nil {
 		return err
@@ -98,85 +131,31 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
 				return false, nil
 			}
 
-			// apply any directory filters
-			if len(dirs) > 0 {
-				nodeDir := path.Dir(nodepath)
+			if withinDir(nodepath) {
+				// if we're within a dir, print the node
+				Printf("%s\n", formatNode(nodepath, node, lsOptions.ListLong))
 
-				// this first iteration ensures we do not traverse branches that
-				// are not in matching trees or will not lead us to matching trees
-				var walk bool
-				for _, dir := range dirs {
-
-					// the current node path is a prefix for one of the
-					// directories, so we're interested in something deeper in the
-					// tree. Example:
-					//   nodepath: "/test"
-					//   dir:      "/test/foo"
-					approachingMatchingTree := fs.HasPathPrefix(nodepath, dir)
-
-					// we're within one of the selected dirs, example:
-					//   nodepath: "/test/foo"
-					//   dir:      "/test"
-					inMatchingTree := fs.HasPathPrefix(dir, nodepath)
-
-					// this condition is complex, but it basically requires that we
-					// are either approaching a matching tree (not yet deep enough)
-					// or: if recursive, we have entered a matching tree; if non-
-					// recursive, then that we are at exactly the right depth
-					// (we can do the walk correctly by just using the condition of
-					// "approachingMatchingTree || inMatchingTree", but it will be
-					// much slower for non-recursive queries since it will continue
-					// to traverse subtrees that are too deep and won't match -- this
-					// extra check allows us to return SkipNode if we've gone TOO deep,
-					// which skips all its subfolders)
-					if approachingMatchingTree || opts.Recursive || (inMatchingTree && dir == nodeDir) {
-						walk = true
-						break
-					}
-				}
-				if !walk {
-					if node.Type == "dir" {
-						// signal Walk() that it should not descend into the tree.
-						return false, walker.SkipNode
-					}
-
-					// we must not return SkipNode for non-dir nodes because
-					// then the remaining nodes in the same tree would be
-					// skipped, so return nil instead
-					return false, nil
-				}
-
-				// this second iteration ensures that we get an exact match
-				// according to the filter and whether we should match subfolders
-				var match bool
-				for _, dir := range dirs {
-					if nodepath == dir {
-						// special case: match the directory filter exactly,
-						// which may or may not be desirable depending on your
-						// use case (for example, this is unnecessary when
-						// wanting to simply list the contents of a folder,
-						// rather than all files matching a directory prefix)
-						match = true
-						break
-					}
-					if opts.Recursive && fs.HasPathPrefix(dir, nodepath) {
-						match = true
-						break
-					}
-					if !opts.Recursive && nodeDir == dir {
-						match = true
-						break
-					}
-				}
-				if !match {
+				// if recursive listing is requested, signal the walker that it
+				// should continue walking recursively
+				if opts.Recursive {
 					return false, nil
 				}
 			}
 
-			Printf("%s\n", formatNode(nodepath, node, lsOptions.ListLong))
+			// if there's an upcoming match deeper in the tree (but we're not
+			// there yet), signal the walker to descend into any subdirs
+			if approachingMatchingTree(nodepath) {
+				return false, nil
+			}
 
+			// otherwise, signal the walker to not walk recursively into any
+			// subdirs
+			if node.Type == "dir" {
+				return false, walker.SkipNode
+			}
 			return false, nil
 		})
+
 		if err != nil {
 			return err
 		}
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -13,12 +13,21 @@ import (
 )
 
 var cmdLs = &cobra.Command{
-	Use:   "ls [flags] [snapshot-ID] [dir...]",
+	Use:   "ls [flags] [snapshotID] [dir...]",
 	Short: "List files in a snapshot",
 	Long: `
-The "ls" command allows listing files and directories in a snapshot.
+The "ls" command lists files and directories in a snapshot.
 
-The special snapshot-ID "latest" can be used to list files and directories of the latest snapshot in the repository.
+The special snapshot ID "latest" can be used to list files and
+directories of the latest snapshot in the repository. The
+--host flag can be used in conjunction to select the latest
+snapshot originating from a certain host only.
+
+File listings can optionally be filtered by directories. Any
+positional arguments after the snapshot ID are interpreted as
+directory paths, and only files inside those directories will
+be listed. If the --recursive flag is used, then the filter
+will allow traversing into matching directories' subfolders.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


mholt:
Dang, my commit from my other pull request snuck into this one. I'll try to fix that up later.

cmd/restic/cmd_ls.go
```diff
@@ -59,19 +71,55 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+	if len(args) > 1 {
+		dirs = args[1:]
+	}
+
 	ctx, cancel := context.WithCancel(gopts.ctx)
 	defer cancel()
-	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args) {
-		Verbosef("snapshot %s of %v at %s):\n", sn.ID().Str(), sn.Paths, sn.Time)
+	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args[:1]) {
+		dirsToPrint := opts.Paths
```
fd0: Hm, I think the message should always print the source paths the snapshot was taken of. This is something completely different from the path within a snapshot. For example, consider a snapshot created with the following command:
```
$ cd /home/user
$ restic backup work
```

Then there will be a single path in the snapshot called `/home/user/work`, but the top-level directory in the snapshot will be `/work`. That's confusing for users, I think.

cmd/restic/cmd_ls.go
```diff
@@ -59,19 +71,55 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
 			if node == nil {
 				return false, nil
 			}
+
+			// apply any directory filters
+			if len(dirs) > 0 {
+				var nodeDir string
+				if !opts.Recursive {
+					// only needed for exact directory match; i.e. no subfolders
+					nodeDir = filepath.Dir(nodepath)
```
fd0: `nodpath` is always slash-separated, as it is created by `path.Join()` (not `filepath.Join`), see https://godoc.org/github.com/restic/restic/internal/walker#WalkFunc

So we need to use `path.Dir()` here to make it work properly on Windows.

cmd/restic/cmd_ls.go
```diff
@@ -11,12 +13,21 @@ import (
...
+The "ls" command lists files and directories in a snapshot.
+
+The special snapshot ID "latest" can be used to list files and
+directories of the latest snapshot in the repository. The
+--host flag can be used in conjunction to select the latest
+snapshot originating from a certain host only.
+
+File listings can optionally be filtered by directories. Any
+positional arguments after the snapshot ID are interpreted as
+directory paths, and only files inside those directories will
```
fd0: I think it'd be good to have a hint here that `dir` needs to be slash-separated, and optionally starting with a `/`. How do you want to handle that? The path passed to the function in `walker.Walk()` will be slash-separated and starting with a `/` (so it's an absolute path)...

cmd/restic/cmd_ls.go
```diff
@@ -59,19 +71,55 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+			if len(dirs) > 0 {
+				var nodeDir string
+				if !opts.Recursive {
+					// only needed for exact directory match; i.e. no subfolders
+					nodeDir = filepath.Dir(nodepath)
+				}
+				var match bool
+				for _, dir := range dirs {
+					if opts.Recursive {
+						if strings.HasPrefix(nodepath, dir) {
```
fd0: That's not enough, for a node with the path `/foo/bar/baz` it would also be printed if restic was called like this: `restic ls latest /foo/bar/b`. Instead, call `fs.HasPathPrefix(dir, nodpath)`, which does the right thing: https://godoc.org/github.com/restic/restic/internal/fs#HasPathPrefix

cmd/restic/cmd_ls.go
```diff
@@ -59,19 +71,55 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+						}
+					} else {
+						if nodeDir == dir {
+							match = true
+							break
+						}
+					}
+				}
+				if !match {
+					return true, nil
```
fd0: Ah, this may be confusing in the docs: When the walker function returns `ignore=true`, the **node** won't be visited again, which is identified by the tree ID, which may be the same for different directories at different paths. For example, consider a backup of a couple of empty dirs, e.g. `/home/foo/dir1` and `/other/dir2`, and calling `restic ls latest /other`. When the walker runs the function for `/home/foo/dir1` and `ignore=true` is returned (because it is not to be listed), it will add the tree ID for the empty directory to its list of trees to ignore, and the walker function will not be called again.

Instead, return `ignore=false` and return `walker.SkipNode` as the `error` instead.

I'm sorry that it's so confusing, the interface was adapted from `filepath.Walk`, and later I added the option to not only ignore stuff based on the path (return `SkipNode`) but also based on the tree ID (return `ignore=true`). Sigh. If you have an idea on how to improve the docs, please let me know!

fd0 (changes requested):
So, first review comments are in. I'm sorry that it's so confusing, but that's life, full of interesting corner cases ;)

cmd/restic/cmd_ls.go
mholt: Alright -- I can understand that. Reverting this change!

cmd/restic/cmd_ls.go
mholt: Sounds good.

cmd/restic/cmd_ls.go
mholt: Hmm, what do you mean _optionally_ starting with a `/`? I can mention it, I just don't quite understand what we need to communicate here yet.

cmd/restic/cmd_ls.go
mholt: Ah, good catch. Thanks. My habits from Caddy path handling (which is different in HTTP requests) is carrying over here.

cmd/restic/cmd_ls.go
mholt: Ohhh... interesting! Thanks for clarifying that for me.

I wonder, in hindsight, is my usage of these return values correct in these places?

- https://github.com/restic/restic/blob/abc923f69391ae55b0319c31ff2560083c874322/cmd/restic/cmd_stats.go#L177-L184 (I noticed that in cmd_ls.go, you return false instead of true in these cases.)
- https://github.com/restic/restic/blob/abc923f69391ae55b0319c31ff2560083c874322/cmd/restic/cmd_stats.go#L236 (now I'm not sure if I didn't want to visit the node again or if I wanted to not visit the tree again)

Update: However, when I change this line to `return false, walker.SkipNode`, the tree is not walked at all. The ls does in fact work correctly with the line as-is... 🤔

cmd/restic/cmd_ls.go
fd0: Ah, sorry, I should have been clearer: the filter functions you wrote only work when the paths passed to `restic ls` are absolute (starting with a `/`), because the `nodepath` passed to the function in `walker.Walk()` starts with a slash. So I think we should document it.

Another option would be to just prefix the dirs with a slash if it isn't there already. That's what I meant with "optionally". Thinking about it, I'd suggest just requiring the paths to be absolute.

cmd/restic/cmd_ls.go
fd0: I'm not sure what you mean by "this line", can you clarify?

fd0: FYI, the walker has many tests for all cases, maybe that can clarify it a bit, see:

https://github.com/restic/restic/blob/abc923f69391ae55b0319c31ff2560083c874322/internal/walker/walker_test.go#L216-L224

fd0: It doesn't, have a look at this repo: https://fd0.me/tmp/repo.tar.gz (the password is `foo`):

```
$ ./restic ls -l latest
repository 1271b4d6 opened successfully, password is correct
found 1 old cache directories in /home/fd0/.cache/restic, pass --cleanup-cache to remove them
snapshot 62bbf4fb of [] at 2018-08-11 22:09:31.33743769 +0200 CEST):
drwxr-xr-x  1000   100      0 2018-08-11 22:01:11 /test
drwxr-xr-x  1000   100      0 2018-08-11 22:02:58 /test/bar
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/bar/test.txt
drwxr-xr-x  1000   100      0 2018-08-11 22:04:19 /test/foo
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/foo/test.txt


$ ./restic ls -l latest /test/bar
repository 1271b4d6 opened successfully, password is correct
found 1 old cache directories in /home/fd0/.cache/restic, pass --cleanup-cache to remove them
snapshot 62bbf4fb of [/test/bar] at 2018-08-11 22:09:31.33743769 +0200 CEST):
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/bar/test.txt

$ ./restic ls -l latest /test/foo
repository 1271b4d6 opened successfully, password is correct
found 1 old cache directories in /home/fd0/.cache/restic, pass --cleanup-cache to remove them
snapshot 62bbf4fb of [/test/foo] at 2018-08-11 22:09:31.33743769 +0200 CEST):
```

The third call of `restic ls` did not print anything, although it should have printed the contents of the `test/foo` directory. The reason is that `/test/foo` and `/test/bar` have the same tree, which is on the ignore list because when `Walk()` processed `/test/bar`, it did not match the filter.

By the way, I think the directory itself (e.g. `/test/bar`) should also be printed.

fd0: > The SkipNode value seems very aggressive in what it skips... I'm still working on it though.

The semantics are modeled after `filepath.Walk()`...

> What do you mean by this? Which directory exactly, and printed where? 

I meant that if I run `restic ls latest /test/bar` then I think it's good to have the directory node `/test/bar` itself also printed, like this:
```
repository 1271b4d6 opened successfully, password is correct
found 1 old cache directories in /home/fd0/.cache/restic, pass --cleanup-cache to remove them
snapshot 62bbf4fb of [/test/bar] at 2018-08-11 22:09:31.33743769 +0200 CEST):
drwxr-xr-x  1000   100      0 2018-08-11 22:02:58 /test/bar
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/bar/test.txt
```

What do you think?

mholt:
The commit I just pushed addresses most of the feedback. The logic was very different from what I expected! Sorry for the confusion.

Two things:

1. If you choose a shallow filter, e.g. `/` or `/home`, the directory listing is still sorta slow because it traverses subfolders (but doesn't print them) instead of skipping them entirely. I'm still working on the right logic for that...

2. I think the only other point of feedback that I haven't yet addressed is whether to match the folder itself, when specified in a filter. My particular use case does not need this, since I want to see the _contents_ of the filtered folder, but I could see why including the folder would be a good idea, too. I guess it could always be ignored. I'll revise that shortly.

mholt:
I think there's a bug in SkipNode, somehow?

When I return SkipNode for a node that has the `nodepath` of a file, it skips the rest of the files in that directory.

Is that expected?

fd0:
> When I return SkipNode for a node that has the nodepath of a file, it skips the rest of the files in that directory.
> Is that expected?

Yes, because I copied the behavior from `filepath.Walk()`:

>  If the function returns SkipDir when invoked on a directory, Walk skips the directory's contents entirely. If the function returns SkipDir when invoked on a non-directory file, Walk skips the remaining files in the containing directory.

mholt:
Ahh, I misunderstood the second sentence in the docs. Totally my bad. After re-reading, I understand now. Thanks!

fd0:
I don't like this behavior at all, but until somebody spends time to revisit it, we're stuck ;)

mholt:
No worries. I think I made it work out alright. It's fast and accurate according to my tests, so I look forward to when you have a chance to try it!

fd0:
Sooo. I'm sorry, but I think I found two bugs, using the test repo I mentioned yesterday:
```
$ restic ls -l latest

drwxr-xr-x  1000   100      0 2018-08-11 22:01:11 /test
drwxr-xr-x  1000   100      0 2018-08-11 22:02:58 /test/bar
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/bar/test.txt
drwxr-xr-x  1000   100      0 2018-08-11 22:04:19 /test/foo
-rw-r--r--  1000   100      0 2018-08-11 22:02:58 /test/foo/test.txt
```

This is empty:
```
$ restic ls -l /test/foo
```

Since I could not hold my fingers still, I've already debugged the issues and will push some commits soon.

cmd/restic/cmd_ls.go
```diff
@@ -59,20 +74,89 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
 		return err
 	}
 
+	// extract any specific directories to walk
```
fd0: I'd do that before opening the repo, so users get a feedback on the dirs they specified (e.g. when they aren't absolute) early on, before any network traffic.

cmd/restic/cmd_ls.go
```diff
@@ -59,20 +74,89 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+
+			// apply any directory filters
+			if len(dirs) > 0 {
+				nodeDir := path.Dir(nodepath)
+
+				// this first iteration ensures we do not traverse branches that
+				// are not in matching trees or will not lead us to matching trees
+				var walk bool
+				for _, dir := range dirs {
+					approachingMatchingTree := fs.HasPathPrefix(nodeDir, dir)
```
fd0: This needs to use `nodepath` instead of `nodeDir`, otherwise one level is walked too much. 

When the nodepath is `/test/foo` and the filter dir is `/test/bar`, then this would still evaluate:
```go
fs.HasPathPrefix("/test", "/test/bar")
```

It should do the following:
```go
fs.HasPathPrefix("/test/foo", "/test/bar")
```

cmd/restic/cmd_ls.go
```diff
@@ -59,20 +74,89 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
...
+					// to traverse subtrees that are too deep and won't match -- this
+					// extra check allows us to return SkipNode if we've gone TOO deep,
+					// which skips all its subfolders)
+					if approachingMatchingTree || opts.Recursive || (inMatchingTree && dir == nodeDir) {
+						walk = true
+						break
+					}
+				}
+				if !walk {
+					return false, walker.SkipNode
```
fd0: We need to be careful here: When we return `SkipNode` for a node that is not a dir (e.g. a file or other type), then the remaining nodes in the same tree are skipped!

mholt:
😮 You are a _saint_. Thanks for your patience and initiative here. I know sometimes contributions can be a burden too -- sorry if this is one of those times!

I've pulled your changes and inspected the code. That is definitely cleaner than my approach, and I see now where the main bug was in mine.

Your changes work for me in all my tests, and it's pretty speedy, too. Yours even works if there's a trailing slash in the filter :) The --recursive flag and multiple filters all seem to work as expected.

I agree tests would be good, although I am not sure where to add tests for the ls command, nor quite sure a strategy for how to test it.

Thank you again! :+1:

fd0 (approved):

fd0:
> I know sometimes contributions can be a burden too -- sorry if this is one of those times!

Don't worry, it was not :)

Issue #1945, opened 2018-08-11
rebuild-index: Reading header of files in the repo is not retried


fd0:
This issue was reported in the forum:
https://forum.restic.net/t/corrupted-repo-after-rebuild-index/869

Output of `restic version`
--------------------------

```
restic 0.9.2 (v0.9.2-3-gabc923f6) compiled with go1.10.3 on linux/amd64
```

How did you run restic exactly?
-------------------------------

```
$ restic -r s3:https://us-east-1@s3.wasabisys.com/my-backup rebuild-index
repository 2c128c6e opened successfully, password is correct
counting files in repo
pack file cannot be listed 824981796608e9e4d16bd6534e461c304f7e4d876106fa4e56206381be698d0f: EOF
[35:54] 100.00%  132671 / 132671 packs
finding old index files
saved new indexes as [fa145cb3 0bc1df87 4e185daa 3fee5dd5 83d391d4 14614059 2558d313 ed0d127f de82203e f7111dc4 f15c6a1c 6750607b b2cae93a 8635cba1 b86af3cb cda9055e da279e96 6405aa99 08aca003 edce9582 9c9055e3 7447af9c c788189d 01edeb63 10a2f9ea 981cab99 492c24be 7a65bedc 9a395cab 97851de1 0757a523 b50fad91 a29bed50 7495e5d8 58f5af8b e7557c47 5d066599 ab9b2ebe a3dde579 bd455dc3 aa18e796 c4043df0 53c12dbc 0a6b256f a20178ad]
remove 1359 old index files

$ restic -r s3:https://us-east-1@s3.wasabisys.com/my-backup check
using temporary cache in /var/folders/r1/3sr1p9391ql86sbhkyzw8qk80000gn/T/restic-check-cache-920683911
repository 2c128c6e opened successfully, password is correct
create exclusive lock for repository
load indexes
check all packs
pack 82498179: not referenced in any index
1 additional files were found in the repo, which likely contain duplicate data.
You can run `restic prune` to correct this.
check snapshots, trees and blobs
error for tree 1b3c7305:
  tree 1b3c73055917aaae934bf5ec8ec2537c3bb7d60358f0661f8fc9830820180096 not found in repository
error for tree 0e02d4e1:
  tree 0e02d4e15d50157f88c1491d62c04cb8be5f75274daf15ce91f41f746e0f5fdd not found in repository
error for tree 3657ccc4:
[...]

PR #1946, opened 2018-08-11
Remove truncated files from cache

### LINKED ISSUE #1935: EOF error then exit when cache file is truncated
Output of `restic version`
--------------------------

```
restic 0.9.2 (v0.9.2-0-gbeb208e1) compiled with go1.10.2 on darwin/amd64
```

```
restic --cleanup-cache --cache-dir ../mycache ls a278f... --long
```

`rclone serve restic` to a cloud provider like S3 or B2.

Expected behavior
-----------------

In case there is a failure reading a cache file, restic should probably act as if the cache is unavailable for that particular query and use the backend instead.

Actual behavior
---------------

Restic prints EOF and exits.

See this forum thread: https://forum.restic.net/t/restic-ls-lists-some-files-then-sudden-eof/866/3?u=matt

Truncate a cache file. Not really sure how to reproduce exactly yet, sorry.

Apparently some error writing the cache file.

If there's an EOF when doing an operation from the cache, revert to the non-cache behavior.

Yes! I love restic. We're gonna move my whole family to it as soon as I get some work done with it. :)

fd0:

Remove truncated files from the cache instead of displaying an error.

Closes #1935, #1945

https://forum.restic.net/t/corrupted-repo-after-rebuild-index/869

Checklist
---------


mholt:
I have tested this change with a copy of the cache folder that exhibited the issue and verified that this works! Once it gets to the truncated file, there's a slight pause as restic downloads the data it needs, and then it continues gracefully. :tada:

mholt (approved):

Issue #1930, opened 2018-08-07
Simple shell script for installing the latest release


x86dev:
Hi,

just wanted to share a simple 4-line script to download and install the latest release of restic from Linux shell using cURL to /usr/local/bin:

```
RESTIC_TAG_LATEST=$(curl --silent "https://api.github.com/repos/restic/restic/releases/latest" | grep -Po '"tag_name": "v\K.*?(?=")')
echo "Downloading and installing restic v$RESTIC_TAG_LATEST ..."
RESTIC_URL=https://github.com/restic/restic/releases/download/v${RESTIC_TAG_LATEST}/restic_${RESTIC_TAG_LATEST}_linux_amd64.bz2
sudo curl -L --silent ${RESTIC_URL} | bunzip2 > /usr/local/bin/restic
```

Note: **Assumes** running on a x86 64-bit (x86_64) machine.

Might be handy to someone, or even worth for the docs in case there isn't any official PPA.

mholt:
Thanks for sharing this!

I'll also note that we've found this script useful when distributing Caddy: https://getcaddy.com/

And rclone does similarly: https://rclone.org/install.sh

I know a lot of people have issues piping into bash but honestly I feel like if you do it right, it's really convenient and works almost anywhere.

mfsu:
You should never download and install something without integrity and cryptographic signature checking.

mholt:
> You should never download and install something without integrity and cryptographic signature checking.

TLS takes care of this.

If you mean something else, you'll need to be more specific and specify a clear threat model.

fd0:
Thanks for posting the script. I'm not really sure what to do with it, hm.

rawtaz:
I agree, updating itself is not a concern of restic. Leave that to others to manage.

mholt:
This is really cool! I had thought about doing something like this for Caddy but never got around to it.

fd0:
I like that it's so simple, and it only uses the infrastructure we already have in place :)

mfsu:
Thanks Alexander for addressing this so quickly. I want to note that my
original intent was to draw attention to the curl based method, which was
used even by docker at some point.

x86dev:
Oh wow, very nice, fd0! I'll give it a try soon. Cheers!

fd0:
So, the overwhelming majority of users I asked were in favor, so I'm merging this command. Please be aware that until the next release this won't really work.

necenzurat:
[...]

PR #1949, opened 2018-08-12
Add command self-update

fd0:

It adds a command called `self-update` which downloads the latest released version of restic from GitHub and replacing the current binary with it. It does not rely on any external program (so it'll work everywhere), but still verifies the GPG signature using the embedded GPG public key.

I've long wanted to add such a thing, but worried about feature creep. So, today I took the opportunity and built something, and it turned out that it's rather easy to do. I hope that this command makes it much easier for users to upgrade restic, reducing a lot of the friction of the current (very manual) process. In addition, it provides users with a nice, clean built-in way that is a bit more secure than just using a shell script and curl: since the GPG key is included in the binary and used during verification, a potential Man-in-the-Middle is detected and the upgrade fails, assuming the initial restic binary has not been tampered with.

By default, the `self-update` command is hidden behind the `selfupdate` built tag, which is only set when restic is built using `build.go`. The reason for this is that downstream distributions will then not include the command by default, so users are encouraged to use the platform-specific distribution mechanism.

Sample run:

    $ restic version
    restic 0.9.1 compiled with go1.10.3 on linux/amd64
    
    $ restic self-update
    latest version is 0.9.2
    download file SHA256SUMS
    download SHA256SUMS
    download file SHA256SUMS
    download SHA256SUMS.asc
    GPG signature verification succeeded
    download restic_0.9.2_linux_amd64.bz2
    downloaded restic_0.9.2_linux_amd64.bz2
    saved 12115904 bytes in ./restic
    successfully updated restic to version 0.9.2
    
    $ restic version

So, this PR is meant to gather some feedback on the new command. What do you think?

https://github.com/restic/restic/issues/1930

Checklist
---------


chrestomanci:
Perhaps this is the wrong place to ask, but I would like it if there was a --dry-run switch on this, so that I can add a restic self-update to my daily cron, and have it stay silent if there are no updates, or send me an email if there are some so that I ran run the update when I next login.

Alternatively, just create a restic-announce mailing list that I can subscribe to to find out when there is a new version.

fd0:
Interesting idea, thanks. While there is no mailing list, we have a blog at https://restic.net/blog, and GitHub provides an XML/Atom feed for releases here: https://github.com/restic/restic/releases.atom

wscott:
It is a little funny that 'restic self-update' appear to update the restic binary in the current directory. I would have expected it to figure out which restic binary I am currently executing and update that one.  (You have to look at argv[0] and PATH to figure this out.)

And if this is the expected behavior then running 'restic self-update' in a directory without a restic binary should give an error or at least print a message that a new copy of restic is being written to the current directory rather than updating the restic the user just ran.

klemensn:
Looks like it took more than three years for someone wanting to disable?
See #4100.

PR #1950, opened 2018-08-12
Add RPM SPEC file to create restic packages on Fedora and RHEL

ldelouw:
Thanks,

Luc

### DIFF
--- a/contrib/restic.spec
+++ b/contrib/restic.spec
@@ -1,0 +1,70 @@
+Name: restic	
+Version: 0.9.2git.20180812	
+Release: 3%{?dist}
+Summary: restic is a backup program that is fast, efficient and secure.
+
+%global debug_package %{nil}
+
+Group: Applications/Archiving
+License: BSD 2-Clause
+URL: https://restic.net/	
+Source0: %{name}-%{version}.tar.gz
+
+BuildRequires:	golang 
+Requires: /bin/bash	
+
+%description
+restic is a program that does backups right. The design goals are:
+
+Easy: Doing backups should be a frictionless process, otherwise you are tempted to skip it. Restic should be easy to configure and use, so that in the unlikely event of a data loss you can just restore it. Likewise, restoring data should not be complicated.
+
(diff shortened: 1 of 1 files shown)


fd0 (commented):
Hey, thanks for your contribution! I assume you've tested the spec file, and it works, correct?

Besides that, I'd like you to:
 * Squash the commits
 * Use present tense for the commit messages, like "Add RPM spec file", don't use punctuation

Then we can merge it! :)

MarkMielke (approved):

ldelouw:
Hi,

I successfully build the RPMs (and installed them) on RHEL6, RHEL7 and Fedora28. I also briefly tested the resulting binary for functionality (sftp only)

copart:
Just wanted to chime in to say that Restic is now in the official Fedora repos.  Eventually it should also be in EPEL.

See:
https://src.fedoraproject.org/rpms/restic

ldelouw:
That is interesting... at the moment it is available only in rawhide (@ f29 as of writing this post). No clue if it is getting backported to F28 or even EPEL, I have not found a bugzilla entry for EPEL so far. 

Anyway: The SPEC file should be available in upstream (this exact repository) to not reinvent the wheel. I dont care which one, Steve Millers or mine.

See also: https://bugzilla.redhat.com/show_bug.cgi?id=1590921 and https://bugzilla.redhat.com/show_bug.cgi?id=1432901

Thanks,

Luc

copart:
ldelouw: I see no reason not to include your spec file, yours may work for all rpm systems, not just Fedora/EPEL.  The spec I had to create for Fedora required that I remove the vendor files and grab all dependencies from other Fedora Packages.  

Restic will be built for F28 later this week, I am waiting for one last dependency to be pushed to stable which will happen tomorrow.  Once Restic hits F28 I will work on making it available on EPEL 7.  EPEL 7 support will require a rewrite of the spec file. I will not be able to use the same spec for both distributions since EPEL7 does not have the latest go macros that the Fedora packaging committee required that I use.  This also goes for the dependencies I had to package myself to get Restic to build.  Fedora on EPEL7 is still a few weeks away.

All that said, I think your spec file is more generic and should be the one in upstream.

fd0:
Thanks for the feedback, and for your work @copart!

PR #1952, opened 2018-08-14
Add more sshcmd tests

tcely:
Provide more test data to hopefully catch coding errors.

Not that I am aware of.

Checklist
---------

### DIFF
--- a/internal/backend/sftp/sshcmd_test.go
+++ b/internal/backend/sftp/sshcmd_test.go
@@ -40,6 +40,11 @@ var sshcmdTests = []struct {
 		"ssh",
 		[]string{"ho:st", "-p", "123", "-l", "us:er", "-s", "sftp"},
 	},
+	{
+		Config{Host: "host:ssh", Path: "/dir/subdir"},
+		"ssh",
+		[]string{"host", "-p", "22", "-s", "sftp"},
+	},
 }
 
 func TestBuildSSHCommand(t *testing.T) {
--- a/internal/backend/sftp/sshcmd_test.go
+++ b/internal/backend/sftp/sshcmd_test.go
@@ -30,6 +30,16 @@ var sshcmdTests = []struct {
 		"ssh",
 		[]string{"host", "-p", "10022", "-l", "user", "-s", "sftp"},
 	},
+	{
+		Config{User: "us:er", Host: "ho:st:10022", Path: "/dir/sub:dir"},
+		"ssh",
+		[]string{"ho:st", "-p", "10022", "-l", "us:er", "-s", "sftp"},
+	},
+	{
+		Config{User: "us:er", Host: "ho:st:0000123", Path: "/dir/sub:dir"},
+		"ssh",
+		[]string{"ho:st", "-p", "123", "-l", "us:er", "-s", "sftp"},
+	},
 }
 
 func TestBuildSSHCommand(t *testing.T) {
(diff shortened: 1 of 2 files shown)


fd0:
Hm, I'd like to confirm something: are host names with a colon in them a thing? Do people use this, did you encounter a host name with a colon in it in the wild?

internal/backend/sftp/sshcmd_test.go
```diff
@@ -30,6 +30,21 @@ var sshcmdTests = []struct {
 		"ssh",
 		[]string{"host", "-p", "10022", "-l", "user", "-s", "sftp"},
```
fd0: I think we should remove this test (and support for passing a `host:port` string to `buildSSHCommand()`. That caused us problems in the past, to the point that we removed parsing the port at all, and added hints to the manual how users can work around this: https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#sftp

fd0:
I'm sorry, I think the comments I wrote come across as very negative. I'm criticizing the code, not your contribution! Thank you very much for trying to improve restic!

It'd be helpful for me to understand what your motivation for this change is, so I'm looking forward to your answers for the questions I asked above. :)

tcely:
@fd0 

Hostnames with a colon are unusual in my experience, but IPv6 makes heavy use of them and the naive splitting on ':' precludes specifying the IPv6 address as the host to connect to. FYI, this is typically `[::1]:22` with the IPv6 address and port.

Usernames on Unix-like systems don't contain colons either. Not everything we connect to with sftp is guaranteed to be Unix-like though. BTW, the work around you gave won't work as the first user specified wins in ssh land and providing `sftp:user@foo` means a command line with `-l user` is built.

I tried not to make this overly complicated, but in my opinion the parsing of the repository specification should be made more robust too. My suggestion is to outlaw colon characters in both the `sftp` marker at the beginning of repository specification and in the path at the end.

The parsing then becomes something along these lines:

1. find the first colon, the beginning up to the first colon is the backend type
1. find the last colon, the position after that colon to the end is used to set `Config.Path`
1. split the substring between those two colon positions on '@' if it exists to set `Config.User`
1. the remainder is used to set `Config.Host`

internal/backend/sftp/sshcmd_test.go
tcely: I don't think this is a step that should be taken. Again, making the repository parsing more robust is my preference.

All of this becomes less important when `restic` has it's own configuration file support. However, even then when you have to run `restic` as different users or on many systems, being able to specify all of the configuration in a single command line is much better. I don't like the idea of needing to ensure a specific configuration across many files or users is properly setup before running `restic` operations.

That documentation should include a working example of `-o sftp.command="foobar"` and possibly also the default sftp command that's built.

The options to be built appear to be these:
- `-o sftp.command="ssh host -p 22 -l user -s sftp"`
- `-o sftp.command="ssh host -p 22 -s sftp"`
- `-o sftp.command="ssh host -l user -s sftp"`
- `-o sftp.command="ssh host -s sftp"`

I know I recently went looking for how to get `restic` to use `krssh` and this information would have been helpful then. I ended up creating a wrapper named `ssh` then changing the `PATH` `restic` was using so that this wrapper was found first.

fd0:
Superseded by #2592, thanks a lot!

PR #1953, opened 2018-08-14
ls: Add JSON output support for restic ls cmd

kitone:

This PR adds support for JSON output for the restic ls command, using the existing global --json flag.

output truncated of : `./restic ls 1977f34d /doc/_static --json`

```json
[{
  "time": "2018-08-10T13:41:51.624627031+02:00",
  "tree": "0d7878997a8317b052105f933c2ff0920f2942ce61c4842de1b1ab067d305b35",
  "paths": ["/Users/kitone/go/src/github.com/restic/restic/doc"],
  "hostname": "hostname.local",
  "username": "kitone",
  "uid": 502,
  "gid": 20,
  "id": "1977f34d970e318bd3249f59ec8e87d5a8b17551cfb49664c0ba5f0d56e9a97c",
  "short_id": "1977f34d",
  "nodes": [{
    "name": "_static",
    "type": "dir",
    "path": "/doc/_static",
    "uid": 502,
    "gid": 20,
    "mode": 2147484141,
    "mtime": "2018-08-10T13:07:54+02:00",
    "atime": "2018-08-10T13:07:54+02:00",
    "ctime": "2018-08-10T13:07:54+02:00",
    "struct_type": "node"
  }, {
    "name": "css",
    "type": "dir",
    "path": "/doc/_static/css",
    "uid": 502,
    "gid": 20,
    "mode": 2147484141,
    "mtime": "2018-08-10T13:07:54+02:00",
    "atime": "2018-08-10T13:07:54+02:00",
    "ctime": "2018-08-10T13:07:54+02:00",
    "struct_type": "node"
  }, {
    "name": "favicon.ico",
    "type": "file",
    "path": "/doc/_static/favicon.ico",
    "uid": 502,
    "gid": 20,
    "size": 1406,
    "mode": 420,
    "mtime": "2018-08-10T13:07:54+02:00",
    "atime": "2018-08-10T13:07:54+02:00",
    "ctime": "2018-08-10T13:07:54+02:00",
    "struct_type": "node"
  }],
  "struct_type": "snapshot"
}]
```

Previous PR : https://github.com/restic/restic/pull/1939

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-1953
+++ b/changelog/unreleased/pull-1953
@@ -1,5 +1,7 @@
 Enhancement: ls: Add JSON output support for restic ls cmd
 
-This PR enables users to get the output of `restic ls` in JSON format.
+We've implemented listing files in the repository with JSON as output, just
+pass `--json` as an option to `restic ls`. This makes the output of the command
+machine readable.
 
 https://github.com/restic/restic/pull/1953
--- a/changelog/unreleased/pull-1953
+++ b/changelog/unreleased/pull-1953
@@ -1,0 +1,5 @@
+Enhancement: ls: Add JSON output support for restic ls cmd
+
+This PR enables users to get the output of `restic ls` in JSON format.
+
+https://github.com/restic/restic/pull/1953
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -2,7 +2,10 @@ package main
 
 import (
 	"context"
+	"encoding/json"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -57,6 +60,29 @@ func init() {
 	flags.Var(&lsOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot ID is given")
 	flags.StringArrayVar(&lsOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot ID is given")
 	flags.BoolVar(&lsOptions.Recursive, "recursive", false, "include files in subfolders of the listed directories")
+}
+
+type lsSnapshot struct {
+	*restic.Snapshot
+
+	ID         *restic.ID `json:"id"`
+	ShortID    string     `json:"short_id"`
+	Nodes      []lsNode   `json:"nodes"`
+	StructType string     `json:"struct_type"` // "snapshot"
+}
+
+type lsNode struct {
+	Name       string      `json:"name"`
+	Type       string      `json:"type"`
+	Path       string      `json:"path"`
+	UID        uint32      `json:"uid"`
+	GID        uint32      `json:"gid"`
+	Size       uint64      `json:"size,omitempty"`
+	Mode       os.FileMode `json:"mode,omitempty"`
+	ModTime    time.Time   `json:"mtime,omitempty"`
+	AccessTime time.Time   `json:"atime,omitempty"`
+	ChangeTime time.Time   `json:"ctime,omitempty"`
+	StructType string      `json:"struct_type"` // "node"
 }
 
 func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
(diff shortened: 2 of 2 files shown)


kitone:
/cc @fd0 and @mholt

@mholt , I switch to a "struct_type" field.
I will not be available for several days, feel free to make change.

Best regards,

mholt (approved):
This looks really good! Thanks for working on it during your travel. It's an elegant solution, especially from a developer who's new to Go.

One thing of note is that currently the `ls` output is streamed to stdout, printing matches as it finds them, without buffering the whole list in memory. This approach, for the JSON output, _does_ buffer each item from all snapshots in memory before encoding them. Most of the time this won't be a problem, but I can imagine memory bloat when a listing has tens of thousands (or more) items.

I'm inclined to accept this change, though, since I think it's a really good one, and leave a change to streaming JSON output "as-needed". Heck, I might even be willing to build that on top of this later.

fd0:
Hey, thank you very much for your contribution! I have not looked at the code yet, but I think it'd be good to not print anything about the snapshot (programs will probably have that information already, or they can call `restic snapshots --json` to get the info), then we can have `restic ls --json` stream an array of nodes, one at a time. Programs can either read the whole output at once, or consume one node at a time.

frankdugan3:
Really looking forward to this! I built a backup GUI (served via GraphQL API), and it currently relies on some poorly-performing and ugly JS regexes to parse the results of `restic ls --long --quiet --no-lock [snapshot ID]` to browse the files of a given snapshot.

mholt:
@frankdugan3 Cool, do you have your GUI anywhere? I'm interested to see it!

frankdugan3:
@mholt  As of right now it's pretty integrated into a much larger ERP app I'm working on for my employer. But I've been thinking about the best way to release a more generalized version for distribution. Currently torn between making a plugin/GraphQL middleware (in JS) just to help with getting API access, or making a more general GUI webapp binary (in Go).

mholt:
@frankdugan3 Gotcha -- that sounds like a fun project. ([I'm working on one too](https://twitter.com/relicabackup)) I've really enjoyed making restic visually accessible. I think this PR will be a great benefit for that.

frankdugan3:
@mholt Wow... That looks nice! Well, in that case I’ll sub to updates and maybe just publish a GraphQL lib!

fd0:
Hm, now I see why we need the info about the snapshots: the `ls` command still supports printing more than one snapshot at a time. Should we maybe remove this function to sharpen what it should be used for? Users can search for files using the `find` function...

fd0 (approved):
I've rebased the commits on the latest master branch and improved the changelog entry a bit, but apart from that it's great! Thanks for your contribution!

mholt:
@fd0

> Hm, now I see why we need the info about the snapshots: the ls command still supports printing more than one snapshot at a time. Should we maybe remove this function to sharpen what it should be used for? Users can search for files using the find function...

I have a couple ideas for this, since I'm still interested in streaming the results out:

1. We could print each node as its own JSON object, like you suggested, but augmented with a little bit of info about the snapshot it's in, even just the snapshot ID and/or "short ID". This is just an extra field or two on the JSON output for the node and would be pretty easy to work with

2. We could print each snapshot's JSON serialization once as we traverse into that snapshot -- assuming we list entries from each snapshot in groups, rather than mixing them. This allows the consuming program to keep simple state. Each time it encounters a snapshot object, it just updates which snapshot the following listings are for, until the next snapshot object appears. This way we can print the whole snapshot with all its information since it only prints once.

What do you think of either of those options for streaming?

@frankdugan3 

> Wow... That looks nice! Well, in that case I’ll sub to updates and maybe just publish a GraphQL lib!

Thanks! No doubt there will be users who find a GraphQL lib useful. I saw your reply on Twitter too, we'd be happy to have you help us test it out -- we'll be ready in a couple of weeks!

fd0:
I thought that you'd be interested in it ;)

I think it'd be fine to have a stream of json objects, one per line, starting with a snapshot object, then all nodes, then maybe another snapshot object. If you want to work on that, please go ahead and let's discuss the details in the next PR or issue.

PR #1962, opened 2018-08-19
ls: Stream output when using --json option

mholt:

This PR streams the output of `restic --json ls` so that large listings will be efficiently handled.

I will add the docs and changelog file after a review, to ensure accuracy in case any more changes are made.

#1953

Checklist
---------


internal/restic/node.go
```diff
@@ -50,7 +50,7 @@ type Node struct {
 
 	Error string `json:"error,omitempty"`
 
-	Path string `json:"-"`
+	Path string `json:"path"`
```
mholt: Not sure if you had a reason for omitting this originally but I took the liberty of adding it to the JSON. If this affects something critical like repository indexes, let me know and I'll revert to just copying over the desired fields for `ls`.

fd0: Uh, yes, this field should not be saved to the repository, so it's left out from the json data (it's also what breaks the tests). The reasoning is as follows:

At the moment, a `node` object in the repository describes exactly one file (or directory). So when you have a directory with a file called `test.txt` with the content `foo`, the `node` for that file will describe the metadata (timestamps, name) and the content (as a list of blob IDs). The `node` structure only describes the **target**, in this case a file.

A `tree` is just a list of nodes. So when you have two directories with exactly the same files (metadata+content), it will result in exactly the same tree. Similar to the `node` structure, a `tree` also only describes the **target** directory.

When you add the `path` where the node/directory was originally located, you add a different kind of information to the data structure: the **source** of the data.

The consequence are as follows:
 * the metadata size will be larger, since for each node there's an additional (potentially longer string) for the path
 * in some (important) use cases, restic needs to write a new `tree` for each and every directory, just because the path has changed. For example, there are users who have a file system based on btrfs or zfs, who do a new file system snapshot before making a backup. This is then available as `/mnt/snapshot/201808191017` (or something similarly unique). restic is unable to find a parent snapshot, since that directory wasn't saved before. But you can just pass the parent snapshot manually with `--parent`, and it'll work just fine. With this change, restic needs to write new tree objects for **every** node, since the path has changed.
 * We're using the `path` during backup to hold the (absolute) path to the file on the file system at that point in time. So, let's say for a file `test.txt` it's `/mnt/snapshot/201808191017/test.txt`. When restic was called like `cd /mnt/snapshot/201808191017; restic backup test.txt`, the path in the repository to the file will be `/test.txt`. So, what's the point in archiving the original path in the node? It'll be different for `ls`: `/test.txt`...

So, in summary, I'd prefer to not change the repository format, and as a safeguard this field should not be serialize. If you need this field somewhere (e.g. in `ls`), you could make a new struct which anonymously embeds `restic.Node`, but has its own `Path` field.

cmd/restic/global.go
```diff
@@ -360,7 +360,9 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 		if len(id) > 8 {
 			id = id[:8]
 		}
-		Verbosef("repository %v opened successfully, password is correct\n", id)
+		if !opts.JSON {
+			Verbosef("repository %v opened successfully, password is correct\n", id)
+		}
```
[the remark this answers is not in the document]
fd0: LGTM

cmd/restic/cmd_ls.go
```diff
@@ -64,25 +62,14 @@ func init() {
...
-	Path       string      `json:"path"`
-	UID        uint32      `json:"uid"`
-	GID        uint32      `json:"gid"`
-	Size       uint64      `json:"size,omitempty"`
-	Mode       os.FileMode `json:"mode,omitempty"`
-	ModTime    time.Time   `json:"mtime,omitempty"`
-	AccessTime time.Time   `json:"atime,omitempty"`
-	ChangeTime time.Time   `json:"ctime,omitempty"`
-	StructType string      `json:"struct_type"` // "node"
+	*restic.Node
```
fd0: We have created a different structure on purpose, to make it clear which fields are exported via `find` or `ls`, instead of the raw data from the repository. restic has two "contracts" (in terms of data structures) with the outside world: the repository (`restic.Node`) and the output of `find` and `ls`. Maybe we can reuse/refactor the code in `find` so that we'll always get the same output:

https://github.com/restic/restic/blob/d708d607fab766121d06ac64af55362daeb52a60/cmd/restic/cmd_find.go#L100-L113

fd0 (changes requested):

internal/restic/node.go
fd0: ah, I understand now that you've modified the json tags to not have a separate struct for json output for the `ls` command. My comment is then just an explanation why the `Path` is ignored ;)

mholt: Wow, excellent explanation -- thanks!

Yes, I only intended to simplify the structure for the ls output, only after I pushed did I realize that it would change the repository format, oops. 😅  - and I'm still getting used to this different editor, which I thought ran the tests but apparently not.

I've just pushed an update to revert that simplification.

mholt:
Any other changes you would like for this PR?

Thanks so much for your time, and for your hard work on restic!

fd0:
Thanks!

PR #1963, opened 2018-08-19
web: Implement api/web server for restic (WIP/PoC)

kitone:

It's an early status / proof of concept, but I want to have feedback as soon a possible.

To work and prototype fast (and maybe ugly ;), the views use bootstrap and fancytree, the treeview is lazy load using the /api.

You can access to the webserver with :
_(need to be in the restic project path for templates,js,css)_
./restic web
http://localhost:8080/web/

json api respond to :

http://localhost:8080/api/snapshots/
http://localhost:8080/api/snapshots/{snapshot_short_id}/nodes/
http://localhost:8080/api/snapshots/{snapshot_short_id}/nodes/{subtree}

My point is not de build a web ui to manage all restic functions, but the idea is to be able to navigate, stats, browse and do quick restore / download from a web page.

I think having  endpoint api build in restic would make this easier (also for other building an ui).

@fd0, somes questions :
- First of all, does it make sense ? :)
- Can we have the stats on snapshot without having to run / always compute (restic stats) ? Is this save in the repository ?
- Why a Node doesn't have an id/key to identify them ? It would be easer to ask for a node by key/id and rebuild the content then download the content from a webpage ?

Best regards,
kitone

Found issue #60 : Implement web server to view backups

Checklist
---------


kitone:
Hi @gbolo,

Just add a way to download a file.

fd0:
Hi @kitone! Thank you very much for your contribution!

I'm sorry about the rather long delay, restic is a fun project that I'm doing in my spare time, and I the last few weeks have been very busy real-life wise.

> First of all, does it make sense ? :)

Very much, at least for the use case for restoring/accessing files. The `backup` use case is very different and needs further thought, I'd like to not do that right now

So as long as the web UI you're building is restricted to restoring/accessing files you have my full support :)

> Can we have the stats on snapshot without having to run / always compute (restic stats) ? Is this save in the repository ?

That depends on the data that is to be queried: the restore size probably won't change, but is not saved anywhere, so for now it needs to be re-computed.

> Why a Node doesn't have an id/key to identify them ? It would be easer to ask for a node by key/id and rebuild the content then download the content from a webpage 

Did you find the [design document](https://github.com/restic/restic/blob/master/doc/design.rst) yet? It describes the data we store in the repo, after reading it you should have a good understanding of what's available. Only the data structures that have been saved to the repo have an ID, and since we don't save every single node by itself, only trees (collections of nodes) have IDs.

I'll try to have a look at the code shortly. Please let me know if there's anything you need! And don't hesitate to bug me here several times, GitHub issues get complicated/forgotten/convoluted very fast. I'm trying to keep up, but the project is so big already that I sometimes get lost and forget things. ;)

kitone:
Hi! Don't worry, I just come back from vacation three day ago! Take your time, I know what it is.

I just push some commit with cleanup and the ability to restore selected files locally.

McFlat:
I'd recommend a whole other application called something along the lines of `restic-web` written in go or nodejs and react or angular for the frontend, to be able to do everything you can do from the commandline interface

mholt:
I do agree that this would probably be best as a separate repo/project, not integrated into restic directly. Restic's global `--json` flag works on almost all the commands now (the last few are in progress) which make automating restic just as easy, without the added complexity of a REST API server (and integrated UI).

nioncode:
@fd0 Did you already decide on how you'd like to integrate a 'browse' functionality for restic? I.e. either add it to restic's core or implement it as a separate application?

I think it would be great to have it in restic's core similar to the fuse mount to make it easy for users on Windows to browse their repository and restore selected files / folders from an earlier snapshot. I don't know how useful the `api` parts of this PR are when almost all commands allow returning JSON by now, but the web server would be a great addition at least. If you'd chime in to give a direction where you'd like this PR to go, then I might join @kitone to get this merged.

fd0:
> @fd0 Did you already decide on how you'd like to integrate a 'browse' functionality for restic? I.e. either add it to restic's core or implement it as a separate application?

@nioncode I think it'd be valuable to have that in restic itself. I'm not sure if the web UI will live in the same repo or not, we'll have to see how that's going.

fd0:
What's the current status of this PR?

kitone:
Hi, sorry for the delay.
For now, I do not want to work on this anymore, mounting and webdav pullrequest feel more interessing if it's only about restoring.

And to be honest I don't like JavaScript and my poor skills for JS (angular, react...) don't me feel I can build a good ui for restic.

fd0:
Thanks for the feedback! I can understand that very well! I'm closing this PR for now, please feel free to add further comments. Thanks!

PR #1973, opened 2018-08-28
cache: Print message when new cache is created

fd0:

Sometimes, users run restic without retaining the local cache directories. This was reported several times in the past.

Restic will now print a message whenever a new cache directory is created from scratch (i.e. it did not exist before), so users have a chance to recognize when the cache is not kept between different runs of restic.

I don't think so.

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -376,6 +376,10 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 		return s, nil
 	}
 
+	if c.Created && !opts.JSON {
+		Verbosef("created new cache in %v\n", c.Base)
+	}
+
 	// start using the cache
 	s.UseCache(c)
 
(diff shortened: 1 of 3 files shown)


cmd/restic/global.go
```diff
@@ -376,6 +376,10 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 		return s, nil
 	}
 
+	if c.Created {
+		Verbosef("created new cache in %v\n", c.Base)
```
mholt: Will this print only when the `--json` flag is _not_ used? (Printing non-JSON breaks JSON parsers)

fd0: Uh, good point, thanks!

fd0: At some point I'd like to refactor how messages are printed in restic, so we can handle such things in a more general way...

fd0: Fixed it :)

Issue #1766, opened 2018-05-08
lchown operation not permitted


petabyteboy:
## Output of `restic version`
```
restic 0.8.1 (v0.8.1-134-g2debb5c0)
compiled with go1.9.3 on linux/amd64
```

## How did you run restic exactly?
```
rm -rf /foo/bar
mkdir -p /foo/bar
chown restic /foo/bar

sudo -u restic /usr/local/bin/restic restore latest \
    -r "sftp:backup:bar" \
    -p /home/restic/.restic-password \
    --target="/foo/bar/../" \
    --cache-dir=/home/restic/.cache/restic
```

## What backend/server/service did you use to store the repository?

sftp

## Expected behavior

Run without errors and return 0.

## Actual behavior

All permissions seem to be accurate after the restore, but restic tries to lchown the directories and files and somehow fails with that:

```
ignoring error for /bar/a: Lchown: lchown /foo/bar/a: operation not permitted
ignoring error for /bar/b: Lchown: lchown /foo/bar/b: operation not permitted
ignoring error for /bar/c: Lchown: lchown /foo/bar/c: operation not permitted
ignoring error for /bar/d: Lchown: lchown /foo/bar/d: operation not permitted
```

## Steps to reproduce the behavior

Restore a backup that contains files that are owned by a different user than the one used to restore the files. 

## Do you have an idea how to solve the issue?

I'm not sure if I am missing something here, but maybe don't call lchown on all files and directories?
Isn't lchown only for symlinks?

## Did restic help you or made you happy in any way?
Very :)

rfjakob:
I believe running `lchown` on everything is correct, as restic tries to preserve the owner. (Note that `lchown` works on everything, including symlinks, while `chown` only works on regular files)

One could argue if the error message should be suppressed if restic is not running as root. I just tested with `cp -a`, and it does call `lchown` on the files, but suppresses the error message if it is not running as root. Here's the code ( http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/copy.c#n3015 ):
```
chown_failure_ok (struct cp_options const *x)
{
  /* If non-root uses -p, it's ok if we can't preserve ownership.
     But root probably wants to know, e.g. if NFS disallows it,
     or if the target system doesn't support file ownership.  */

  return ((errno == EPERM || errno == EINVAL) && !x->chown_privileges);
}
```
It seems that `rsync -a` behaves similar - you don't get an error message if you are not running as root.

[...]

PR #1993, opened 2018-09-09
restore: suppress lchown errors when not running as root

rfjakob:

Like "cp -a" and "rsync -a" do, only report lchown errors if we run
as root.

Like cp from GNU coreutils does, we check Geteuid() to determine if
we are running as root
( http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/copy.c#n3012 ).

On Windows, lchown errors are always reported.

Fixes https://github.com/restic/restic/issues/1766

Checklist
---------

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -186,7 +186,16 @@ func (node Node) restoreMetadata(path string) error {
 	var firsterr error
 
 	if err := lchown(path, int(node.UID), int(node.GID)); err != nil {
-		firsterr = errors.Wrap(err, "Lchown")
+		// Like "cp -a" and "rsync -a" do, we only report lchown permission errors
+		// if we run as root.
+		// On Windows, Geteuid always returns -1, and we always report lchown
+		// permission errors.
+		if os.Geteuid() > 0 && os.IsPermission(err) {
+			debug.Log("not running as root, ignoring lchown permission error for %v: %v",
+				path, err)
+		} else {
+			firsterr = errors.Wrap(err, "Lchown")
+		}
 	}
 
 	if node.Type != "symlink" {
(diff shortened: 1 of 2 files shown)


internal/restic/node.go
```diff
@@ -186,7 +186,13 @@ func (node Node) restoreMetadata(path string) error {
 	var firsterr error
 
 	if err := lchown(path, int(node.UID), int(node.GID)); err != nil {
-		firsterr = errors.Wrap(err, "Lchown")
+		// Like "cp -a" and "rsync -a" do, we only report lchown errors if we run
+		// as root.
+		// On Windows, Geteuid always returns -1, and we always report lchown
+		// errors.
+		if os.Geteuid() <= 0 {
+			firsterr = errors.Wrap(err, "Lchown")
```
fd0: Hm, how about we also test if `err` is `os.ErrPermission`? Or is this not necessary?

rfjakob: Good point. Will do.

rfjakob:
Fixed in https://github.com/restic/restic/pull/1993/commits/fa81741e8e0e424eb858037b61d01c4d88c58c02 . Now, the error is only supressed if `Geteuid > 0 && IsPermission`.

fd0 (approved):
LGTM, except one last thing: The text in the changelog entry can be improved to be more end-user friendly. How about:

> Like "cp" and "rsync" do, restic now only reports errors for changing the ownership or access mode of files during restore if it is run ￼as root, on non-Windows operating systems. In the latter case, the error is reported as usual.

rfjakob:
Changelog text updated in https://github.com/restic/restic/pull/1993/commits/e42d2d1da81cd766478ee29d3fc07e80ef7cadac !

fd0 (approved):
LGTM, thanks!

PR #2003, opened 2018-09-21
Progress UI for the new optimized restore implementation

### LINKED ISSUE #426: Progress bar for restore

ifedorenko:

This PR implements periodic status report for the new optimized multithreaded `restore` command implementation. It also introduces new command-agnostic `ui.ProgressUI` interface, which is then used by `restore`, `backup` and `check` long-running commands.

See #1719

Closes #426

Checklist
---------


mholt:
Ooo this looks really nice. Do you think the ProgressUI interface would help a lot with #1944?

And would you be open to having `--json` flag support for this, similar to #1944? I could potentially contribute it, just need some guidance as to how.

ifedorenko:
My idea here is to decouple what progress is being reported from how the report is being rendered to the user. So yes, `--json` does look like a good usecase. There is some rudimentary support for different rendering in `TermstatusProgressUI.displayProgress` already (i.e. ansi-vs-dumb terminal stuff), but I think `--json` will require something more elaborate. I'll need to think about it before I can suggest anything more specific.

fd0:
I've taken the liberty of rebasing this branch on master, thereby removing the commits that are already in master :)

fd0:
I just gave this a try, the UI looks nice.

ifedorenko:
To be clear, this is still work in progress and I don't like the way 
command progress 'phase' api is currently defined. I'll try to finalize the 
api and apply it to at least one other command in the next few days. Of 
course, let me know if you have any ideas or other feedback on what I have 
so far.

--
Regards,
Igor

fd0:
Understood, I'll have a look!

ifedorenko:
as for json, I'd like to suggest the following. 

each command will be modeled as sequence of phases ("load index", "list all files to restore", "restore file content", and so on). 

every few seconds during phase execution emit json message with the following attributes:
* `title` of the phase
* `percent` of the phase completion, if known
* `eta` of the phase completion, if known
* `message` description of the phase process, for example for restore file content download it will be something like "1 out of 1024 MiB in 1 out 123345 files"

at the end of each command phase emit summary json message with the following attributes
* `start`/`end` timestamps
* `message` that says how many files/bytes were "processed" during the phase and processing speed. to continue with restore file content download example, it will be something like "downloaded  1024 MiB in 123345 files at 1000Gbps".

@mholt what do you think?

mholt:
@ifedorenko That sounds perfect -- and, preferably, if each message is encoded using a different struct type, it would be nice if there was a field to indicate the type of message or structure so we can expect certain decoded fields (`struct_type` is what I used in #1962 or `message_type` in #1944 -- hmm, maybe I should have standardized on one name). If that makes sense. Maybe there's a better way to do it, I dunno though.

Basically, if each kind of message could have a different structure, it'd be nice as a consumer of the JSON output to be able to decode all the messages into a structure that can fit all the possible values, or to know which kind of message it is so we can decode properly.

Also, the more details (like processing rate, number of files, etc) that you can put into separate fields as plain numbers, the better. (A message is fine, too, as long as it's not the only way the information is given.)

Anyway, sounds great overall :+1:

ifedorenko:
I was hoping to avoid fine-grained metrics, like number of bytes downloaded or files written ;-). No big deal, will find a way to have those too.

@fd0 do you have a preferred text template library I can use here, like [mustache for Go](https://github.com/cbroglie/mustache) for example? Not sure I'll need this yet, but one of the ideas I consider is to have a map of fine-grained metrics, which are then used to render user-visible messages via a template. Probably an overkill here, but it feels like the right way to provide uniform way to support json and user-facing UI.

fd0:
I'd prefer [`text/template`](https://godoc.org/text/template) from the standard library.

@mholt the json code is not released yet, we can change the name of the type field :)

ifedorenko:
@mholt somebody ;-) will have to change `TermstatusProgressUI` to honour `--json`. `displayProgress` and `diplaySummary` are the two methods will need to change for sure, and maybe `Set` too.

ifedorenko:
@fd0 do you mind I move all backup-specific progress reporting to `archiver` package, similarly to how I already did/prototyped for restore and check? And if you have time for some feedback on overall idea and implementation of ProgressUI API, I'll certainly appreciate that too.

fd0:
@ifedorenko I need to find time to look a this, I didn't look at it yet. Sorry for the delay!

mholt:
@ifedorenko Unfortunately (or fortunately?) I haven't made time to implement the json renderer yet -- I'm getting ready to launch (next week!) the version 1.0 of the application that will use it. After that, I can probably turn my attention to doing it. So if you want to make any (major?) changes, go ahead -- won't hurt anything I've worked on.

ifedorenko:
@fd0 I implemented all changes I planned to do here, please have a look. I'd like to groom git history before (and _if_) this is merged, but otherwise I think this is "done". Let me know if you want me to make any changes or if you disagree with the proposed approach in general.

fd0:
Will do, thanks!

bbigras:
Any progress on this?

ifedorenko:
I have no immediate plans to continue work on this PR. The approach I took here was too ambitious and the change is too big to be worth the trouble. I may provide smaller change to implement restorer progress after #2195 is merged, but can't promise at this point. Really depends how long it takes for #2195 to get reviewed and how much time I will be able to dedicate to restic after that.

Issue #555, opened 2016-07-30
Ability to override a target


wrouesnel:
When doing backups with a filesystem which supports snapshots, the backup target and the filesystem path can change (often in uncontrollable ways).

i.e. if I'm backing up my home:
`/home/will`
but doing it with ZFS snapshots then I'm actually backing up:
`/home/will/.zfs/snapshot/restic-backup-1`

and that's what gets stored as the target. On a subsequent backup with a newly numbered snapshot, restic will then insist on re-reading the content of the entire filesystem i.e. if the snapshot mount is now `/home/will/.zfs/snapshot/restic-backup-2`

In both cases, what I really want to do is to specify the *intended* target and separate that concept from the mount - i.e. `/home/will/.zfs/snapshot/restic-backup-3` is actually just targeting `/home/will`

fd0:
Hey, thanks for submitting this issue. It is in indeed an interesting feature to have.

I can imagine having sort of a "translation" rule, similar to `tar --transform` ([example](http://superuser.com/a/595512)), or just being able to strip path components and insert a custom prefix.

For your first example this could mean running restic as follows:
```
restic backup --prefix /home/will --strip-components 5 \
  /home/will/.zfs/snapshot/restic-backup-1
```
so that `/home/will/.zfs/snapshot/restic-backup-1/foo.txt` has the leading 5 path elements removed to `foo.txt` and then the prefix is applied, so that restic saves `/home/will/foo.txt`.

Is that what you meant?

This feature depends on the solution of #549, because there we'll probably change the way restic handles paths.

wrouesnel:
That would pretty much do it I think.

rumpelsepp:
I posted a similar request, maybe a duplicate here: https://github.com/restic/restic/issues/658

martin21:
fd0, as you found the cause for issue #652 you hinted me at this issue and I thought it would be exactly what I want… yet… thats not entirely true, cause my BTRFS snapshots are in a *different* base path. For example:

1) /daten is the original directory
2) /mnt/daten-zeit/daten-resticbackup-$DATE or $UID would be unique snapshot directory name

Right now I work around issue #652 by providing a fixed snapshot name like "/mnt/daten-zeit/daten-resticbackup" without date. This appears to work well.

I prefer having all snapshots separately from the subvolume I snapshot.

So what I am after would be some "--pathinrepo" option or intended target option.

log "Sichere /home."
[...]

PR #2010, opened 2018-09-25
Added support for --prefix and --strip-components options on backup

giacomocariello:

Adds --prefix and --strip-components to "restic backup" command. These options allow the user to manipulate the root directory paths saved in the snapshot.

It was previously discussed in issue #555 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -78,6 +78,8 @@ type BackupOptions struct {
 	FilesFrom        string
 	TimeStamp        string
 	WithAtime        bool
+	RootPrefix       string
+	RootStrip        int
 }
 
 var backupOptions BackupOptions
@@ -104,6 +106,8 @@ func init() {
 	f.StringVar(&backupOptions.FilesFrom, "files-from", "", "read the files to backup from file (can be combined with file args)")
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "time of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
+	f.StringVar(&backupOptions.RootPrefix, "prefix", "", "apply a prefix to target paths")
+	f.IntVar(&backupOptions.RootStrip, "strip-components", 0, "strip NUMBER leading components from target paths")
 }
 
 // filterExisting returns a slice of all existing items, or an error if no
@@ -350,7 +354,7 @@ func findParentSnapshot(ctx context.Context, repo restic.Repository, opts Backup
 
 	// Find last snapshot to set it as parent, if not already set
 	if !opts.Force && parentID == nil {
-		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, opts.Host)
+		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, opts.Host, opts.RootPrefix, opts.RootStrip)
 		if err == nil {
 			parentID = &id
 		} else if err != restic.ErrNoSnapshotFound {
@@ -501,6 +505,8 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		Time:           timeStamp,
 		Hostname:       opts.Host,
 		ParentSnapshot: *parentSnapshotID,
+		RootPrefix:     opts.RootPrefix,
+		RootStrip:      opts.RootStrip,
 	}
 
 	uploader := archiver.IndexUploader{
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -107,7 +107,7 @@ func init() {
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "time of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
 	f.StringVar(&backupOptions.RootPrefix, "prefix", "", "apply a prefix to target paths")
-	f.IntVar(&backupOptions.RootStrip, "strip-components", 0, "strip NUMBER leading components from target paths")
+	f.IntVar(&backupOptions.RootStrip, "strip-components", 0, "strip `n` leading components from target paths")
 }
 
 // filterExisting returns a slice of all existing items, or an error if no
--- a/internal/restic/snapshot.go
+++ b/internal/restic/snapshot.go
@@ -27,17 +27,45 @@ type Snapshot struct {
 	id *ID // plaintext ID, used during restore
 }
 
+func pathSplit(path string) (root string, lst []string) {
+	for root = filepath.Clean(path); ; {
+		var file string
+		root, file = filepath.Split(root)
+		if file == "" {
+			break
+		}
+		lst = append([]string{file}, lst...)
+	}
+	return
+}
+
+func pathMangle(path string, prefix string, strip int) string {
+	pathRoot, pathList := pathSplit(path)
+	if strip > 0 {
+		var stripRoot int
+		if pathRoot != "" {
+			stripRoot++
+		}
+		pathList = pathList[strip-stripRoot:]
+		pathRoot = ""
+	}
+	if prefix != "" {
+		pathRoot = prefix
+	} else {
+		absPath, err := filepath.Abs(pathRoot)
+		if err == nil {
+			pathRoot = absPath
+		}
+	}
+	return filepath.Join(append([]string{pathRoot}, pathList...)...)
+}
+
 // NewSnapshot returns an initialized snapshot struct for the current user and
 // time.
-func NewSnapshot(paths []string, tags []string, hostname string, time time.Time) (*Snapshot, error) {
+func NewSnapshot(paths []string, tags []string, hostname string, time time.Time, prefix string, strip int) (*Snapshot, error) {
 	absPaths := make([]string, 0, len(paths))
 	for _, path := range paths {
-		p, err := filepath.Abs(path)
-		if err == nil {
-			absPaths = append(absPaths, p)
-		} else {
-			absPaths = append(absPaths, path)
-		}
+		absPaths = append(absPaths, pathMangle(path, prefix, strip))
 	}
 
 	sn := &Snapshot{
--- a/internal/restic/snapshot.go
+++ b/internal/restic/snapshot.go
@@ -28,10 +28,11 @@ type Snapshot struct {
 }
 
 func pathSplit(path string) (root string, lst []string) {
-	for root = filepath.Clean(path); ; {
+        root = path
+        for {
 		var file string
-		root, file = filepath.Split(root)
-		if file == "" {
+		root, file = filepath.Split(filepath.Clean(root))
+		if file == "." || file == "" {
 			break
 		}
 		lst = append([]string{file}, lst...)
--- a/internal/restic/snapshot.go
+++ b/internal/restic/snapshot.go
@@ -63,7 +63,7 @@ func pathStripPrefix(path string, prefix string, strip int) string {
 	}
 	if prefix != "" {
 		pathRoot = prefix
-	} else {
+	} else if strip == 0 {
 		absPath, err := filepath.Abs(pathRoot)
 		if err == nil {
 			pathRoot = absPath
--- a/internal/restic/snapshot.go
+++ b/internal/restic/snapshot.go
@@ -69,7 +69,7 @@ func pathStripPrefix(path string, prefix string, strip int) string {
 			pathRoot = absPath
 		}
 	}
-	return filepath.Join(append([]string{pathRoot}, pathList...)...)
+	return filepath.Clean(filepath.Join(append([]string{pathRoot}, pathList...)...))
 }
 
 // NewSnapshot returns an initialized snapshot struct for the current user and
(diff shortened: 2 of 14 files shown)


fd0:
Hey, thank you very much for giving this a try! I don't have time to look at the code today, but in order for it to get merged eventually, it will need extensive tests in the `archiver` package, here: https://github.com/restic/restic/blob/master/internal/archiver/archiver_test.go

In my experience, adding such tests reveals many subtle bugs, so maybe you'll find the one you're looking for while adding tests :)

giacomocariello:
Okay, I did some cleanup of the commits to solve the issues you pointed out. Regarding the tests, I need some time to figure out what's expected. I posted this initial draft mainly to make sure that the way I conceived it is acceptable from your point of view.

giacomocariello:
Okay, I think I figured out the problem: I need to replicate target mangling behavior in internal/restic function FindLatestSnapshot(), otherwise it won't match parent and behave as if new snapshot were an orphan.

giacomocariello:
Okay, the problem I mentioned in previous comments is solved. Now it's only a matter of writing the tests.

fd0:
Do you need some guidance on that? I could draft the test scaffolding or so...

giacomocariello:
That would be great! I was thinking of extending TestArchiverSnapshot with prefix and strip fields, but certainly an expert drafting would really help.

cmd/restic/cmd_backup.go
```diff
@@ -98,6 +100,8 @@ func init() {
 	f.StringVar(&backupOptions.FilesFrom, "files-from", "", "read the files to backup from file (can be combined with file args)")
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "time of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
+	f.StringVar(&backupOptions.RootPrefix, "prefix", "", "apply a prefix to target paths")
+	f.IntVar(&backupOptions.RootStrip, "strip-components", 0, "strip NUMBER leading components from target paths")
```
fd0: Use `n` instead of `NUMBER` and surround it with backticks, then it'll be printed correctly in the help:
```go
f.IntVar(&backupOptions.RootStrip, "strip-components", 0, "strip `n` leading components from target paths")
```

internal/restic/snapshot.go
```diff
@@ -27,17 +27,45 @@ type Snapshot struct {
 	id *ID // plaintext ID, used during restore
 }
 
+func pathSplit(path string) (root string, lst []string) {
```
fd0: This function needs tests, I'll add a test scaffold for you.

internal/restic/snapshot.go
```diff
@@ -27,17 +27,45 @@ type Snapshot struct {
...
+		root, file = filepath.Split(root)
+		if file == "" {
+			break
+		}
+		lst = append([]string{file}, lst...)
+	}
+	return
+}
+
+func pathMangle(path string, prefix string, strip int) string {
```
fd0: We need to find a better name for this function, and it also needs extensive tests. What about `pathStripPrefix`?

internal/restic/snapshot.go
```diff
@@ -27,17 +27,45 @@ type Snapshot struct {
...
+	return filepath.Join(append([]string{pathRoot}, pathList...)...)
+}
+
 // NewSnapshot returns an initialized snapshot struct for the current user and
 // time.
-func NewSnapshot(paths []string, tags []string, hostname string, time time.Time) (*Snapshot, error) {
+func NewSnapshot(paths []string, tags []string, hostname string, time time.Time, prefix string, strip int) (*Snapshot, error) {
 	absPaths := make([]string, 0, len(paths))
 	for _, path := range paths {
-		p, err := filepath.Abs(path)
```
fd0: Hm, why did you remove the call to `filepath.Abs()`? That's probably wrong for non-absolute paths...

Or does `pathMangle` makes the paths absolute? That must be in the documentation for the function!

internal/restic/snapshot_find.go
```diff
@@ -14,17 +13,11 @@ import (
 var ErrNoSnapshotFound = errors.New("no snapshot found")
 
 // FindLatestSnapshot finds latest snapshot with optional target/directory, tags and hostname filters.
-func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string, tagLists []TagList, hostname string) (ID, error) {
+func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string, tagLists []TagList, hostname string, prefix string, strip int) (ID, error) {
 	var err error
 	absTargets := make([]string, 0, len(targets))
 	for _, target := range targets {
-		if !filepath.IsAbs(target) {
```
fd0: Same here, why no absolute paths?

cmd/restic/cmd_backup.go
```diff
@@ -98,6 +100,8 @@ func init() {
 	f.StringVar(&backupOptions.FilesFrom, "files-from", "", "read the files to backup from file (can be combined with file args)")
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "time of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
+	f.StringVar(&backupOptions.RootPrefix, "prefix", "", "apply a prefix to target paths")
```
fd0: This needs a hint what kind of path you can supply as the prefix. Does it always need to be slash-separated? Or slash/backslash depending on the OS?

fd0 (changes requested):
A few comments I noticed while reading the code. Apparently, the code does not work (yet), did you know that?

```
$ echo $PWD
/home/fd0/shared/work/go/src/github.com/restic/restic/cmd/restic

$ ./restic backup $PWD
[...]

$ ./restic ls -l latest
snapshot a11d3cd9 of [/home/fd0/shared/work/go/src/github.com/restic/restic/cmd/restic] filtered by [] at 2018-09-26 14:30:50.723768212 +0200 CEST):
drwxr-xr-x     0     0      0 2016-11-25 18:24:06 /home
drwxr-xr-x  1000  1000      0 2018-09-26 14:27:52 /home/fd0
drwxr--r--  1000   100      0 2018-08-30 22:04:44 /home/fd0/shared
drwxr-xr-x  1000   100      0 2017-08-06 17:37:27 /home/fd0/shared/work

$ ./restic backup --strip-components 2 $PWD
[...]

$ ./restic ls -l latest | head -n 5
snapshot f576eda7 of [/home/fd0/shared/work/go/src/github.com/restic/restic/cmd/restic] filtered by [] at 2018-09-26 14:31:41.992084043 +0200 CEST):
drwxr-xr-x     0     0      0 2016-11-25 18:24:06 /home
drwxr-xr-x  1000  1000      0 2018-09-26 14:27:52 /home/fd0
drwxr--r--  1000   100      0 2018-08-30 22:04:44 /home/fd0/shared
drwxr-xr-x  1000   100      0 2017-08-06 17:37:27 /home/fd0/shared/work
``` 

It's the same for `--prefix`.

fd0:
Ok, I pushed a commit with test scaffolds and a few quick test cases I came up with, some comments:
 * `TestPathSplit()` fails for a few cases because I could not find out what the function does exactly. Can you please add documentation comment to it? :)
 * `TestPathMangle()` panics for one of the test cases :)
 * Please rename `TestPathMangle()` together with the function.

giacomocariello:
Okay, I'll answer the specs question now and do the actual fixing/coding tonight.

internal/restic/snapshot.go
giacomocariello: pathStripPrefix is okay, I'll rename it.

internal/restic/snapshot.go
giacomocariello: Okay.

cmd/restic/cmd_backup.go
giacomocariello: Okay.

internal/restic/snapshot.go
giacomocariello: Before my changes, if the Abs function returned an error, the function just appended the non-absolute path to the abs list:

```
                p, err := filepath.Abs(path)
                if err == nil {
                        absPaths = append(absPaths, p)
                } else {
                        absPaths = append(absPaths, path)
                }
```
The pathMangle is applied both to NewSnapshot and FindLatestSnapshot, since the behavior should be the same to determine actual target paths. For each path:
1) split each path into a root string (drive letter for Windows, slash for Unix, etc) and a list of path components with pathSplit();
2) if root string is empty, the path is relative, otherwise it's absolute;
3) if strip > 0 and the path is absolute, blank the root string and remove N-1 leading path elements; if strip > 0 and path is relative, remove N-1 leading path elements;
4) if prefix is not empty (either relative or absolute), it is applied in place of the root string;
5) if the prefix is empty, the root string is passed through filepath.Abs, so if a prefix is not applied, the output path is made absoluted, if possible. If filepath.Abs fails, the procedure continues with root string unchanged;
6) Finally the root string and the path elements are reassembled with filepath.Join.

So, to answer your question, pathMangle makes path absolute unless prefix is relative or unless filepath.Abs fails (as it was previously).

internal/restic/snapshot_find.go
giacomocariello: See above.

cmd/restic/cmd_backup.go
giacomocariello: Prefix should follow the same syntax as a normal path. It can be relative or absolute and does not need a final path separator because it's automatically appended.

giacomocariello:
I posted a new commit to solve splitting bug, it should probably solve the wrong behavior you pointed out. Could you check it out?

internal/restic/snapshot.go
giacomocariello: Amending my previous statement, if strip > 0, path is not made absolute either.

giacomocariello:
I think newer commits fix the outstanding issues according to your expectations.

giacomocariello:
I have some problem with TestArchiverSnapshotPrefixStrip: it's complaining that the internal structure of the snapshot filesystem doesn't contain prefix and/or is not stripped, but file paths aren't supposed to change, only root directory field. Am I missing anything?

fd0:
>I have some problem with TestArchiverSnapshotPrefixStrip: it's complaining that the internal structure of the snapshot filesystem doesn't contain prefix and/or is not stripped, but file paths aren't supposed to change, only root directory field. Am I missing anything?

I think we need to talk about what this change is about. There are two places in which the path to a file/dir to backup is relevant:
 * The path recorder in the snapshot (used to detect if there has been a backup of exactly this path before)
 * The structure within the snapshot

Maybe we have a misunderstanding here: I was talking about the second point, --prefix and --strip-components should modify the structure within the snapshot. Did you mabye understand something different? :)

grawity:
Thanks for this feature. I use restic on a fileserver, and ever since trying 0.9.x I wished I could avoid its storage of absolute paths completely (just as 0.8.x used to work). The ability to `--strip-components` at backup time will be very useful, and IMHO it would be a great addition to the restore subcommand as well.

(For example, today I'm backing up `D:\Shares\Users\`, some months later it might be migrated to `E:\Users\` or even `\\anotherserver\Users\`, so I wouldn't want to store anything more than just the relative path `Users\`. More than that, when the day comes to _restore_ some data, the destination might be mounted somewhere else entirely...)

fd0:
@grawity This feature is only slightly related to what you're trying to do, as this PR won't change the paths the data is stored at in a snapshot. To do what you want, you can just pass restic a relative path:
```
D:
cd \Shares
restic backup Users
```

The top-level directory in the next snapshot will be `Users`, and you can restore this as you wish.

fd0:
> In my PR, I'm not changing the structure of the snapshot, only the recorded path. The rationale for this solution is to allow different "source" paths to look equal upon backup, in order to make parent selection work as if the backup were from the same source path.

Ah, I see. Then we did talk about different things. The problem is that it's only half of the story.

Let's say I have a file called `foo.txt` in a zfs volume called `bigdata`. I make a zfs snapshot of the volume, which is the mounted at `/mnt/bigdata/zfs-snapshot-20181016`. I can call restic in several ways to make a backup of the snapshot, this varies at which path the file `foo.txt` in the snapshot is stored:

| restic | path to foo.txt |
|--------|-----------------|
| `restic backup /mnt/bigdata/zfs-snapshot-20181016` | `/mnt/bigdata/zfs-snapshot-20181016/foo.txt` |
| `cd /mnt/bigdata; restic backup zfs-snapshot-20181016` | `/zfs-snapshot-20181016/foo.txt` |
| `cd /mnt/bigdata/zfs-snapshot-20181016; restic backup .` | `/foo.txt` |

For each of these snapshots, restic records the (same) absolute path of the content in the snapshot file in the repo. So if I were to change the call to restic (e.g. from the first to the second row in the table), restic will find a parent snapshot, but it will be useless because the path within then snapshot has changed. That's not a big deal, but the files will be re-read. The parent detection code is too simple to recognize this, and we don't record the current working directory for restic (yet).

So, let's address stripping and/or prefixing the paths recorded in the snapshot in this PR, and tackle the rest later. I think you need to remove some of my test cases then, all those which test the internal structure of the snapshot.

fd0:
Hm. Thinking about this, what do we do in the following case:
```
restic backup --strip-components 2 /home/user/work /home/other/work
```
This would produce `/work`, `/work` as the paths. Is that a good thing to have?

We also need to come up with different names for the command-line option `--path-components`, because this option is used by `tar` and operates on the content, so I think we should not use the same name for restic which does not operate on the content.

To be honest, I'm not sure if this is the right approach to "fix" parent detection.

fd0:
I'm sorry for not responding for such a long time. Are you still interested in working on this PR?

fd0:
I'll close this for now. Thanks!

PR #2017, opened 2018-10-03
mount: Add default-permissions flag to set FUSE option

kylewlacy:

This PR adds the new flag `--default-permissions` for the `mount` command, which enables the "DefaultPermissions" FUSE option. This option enforces the Unix permissions of the mounted filesystem. When this flag is disabled, any user can access any file from the restic FUSE mount if the `--allow-other` flag is set; when enabled, a user can only access a file from a snapshot if they could access it originally.

As a quick demo of this feature, I wrote up a quick gist: https://gist.github.com/kylewlacy/99ec3859955a25dbbb94a56ce1b4de42 (note that it requires root, or at least the ability to run as two different users).

No (I figured it was easier to discuss this feature by providing this small PR first, then iterating on it if needed)

Checklist
---------

N.B. manpage changes need to be regenerated still (I tried regenerating them but there were other changes too, so I figured it would be easier to leave out of this PR for now). I can regenerate the manpages if needed though

I also didn't write any tests for this PR. First, there were no tests for the other FUSE-specific options so I wasn't sure it was warranted, and second, this test would be hard to write since it would involve running at least as a root user and a non-root user. I'd be happy to try my hand at writing some tests if anyone provided some suggestions/ideas for testing this kind of feature

Other notes
-----------

I'd be more than happy to discuss and iterate on the design on this feature. Personally, finding out this caveat with `restic mount --allow-other` was definitely a surprise to me, so maybe it would make sense to invert this? As in, maybe this PR should be changed so "DefaultPermissions" is set by default, with an optional flag to _disable_ it (that would be a breaking change though)

### DIFF
--- a/changelog/unreleased/pull-2017
+++ b/changelog/unreleased/pull-2017
@@ -1,12 +1,11 @@
-Enhancement: mount: Enforce FUSE Unix permissions by default
+Enhancement: mount: Enforce FUSE Unix permissions with allow-other
 
-By default, `mount` will now respect the Unix permissions of the files within
-snapshots (this is done through the "DefaultPermissions" FUSE option).
+The fuse mount (`restic mount`) now lets the kernel check the permissions of
+the files within snapshots (this is done through the `DefaultPermissions` FUSE
+option) when the option `--allow-other` is specified.
 
 To restore the old behavior, we've added the `--no-default-permissions` option.
-This allows alll users that have access to the mountpoint to access all
-files within the snapshots. Normal FUSE rules apply, so `--allow-root`
-or `--allow-other` can be used to allow users besides the mounting user to
-access the mountpoint.
+This allows all users that have access to the mount point to access all
+files within the snapshots.
 
 https://github.com/restic/restic/pull/2017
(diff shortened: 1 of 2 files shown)


fd0 (changes requested):
Hey, thanks for your contribution! I think it'd be good to invert the behavior as you suggested: When `--allow-other` is enabled, by default enforce permissions and allow to disable this with `--no-default-permissions` on explicit request. Better safe than sorry, even if this means a breaking change. In my experience, there won't be so many users affected, as using `--allow-other` is a rather obscure use case.

fd0:
I've added a commit which changes the logic slightly: checking the permissions in the fuse mount when nobody else besides the current user can access the fuse mount does not sense. The current user has access to the repo files in addition to the password, so they can access all data regardless of what
the fuse mount does.

Enabling `--allow-root` allows the root user to access the files in the fuse mount, for this user no permission checks will be done anyway.

The code now enables `DefaultPermissions` automatically when `--allow-other` is set, it can be disabled with `--no-default-permissions` to restore the old behavior.

fd0 (approved):

PR #2019, opened 2018-10-04
cache: Recheck before downloading

fd0:

recheck if another goroutine has already downloaded to the cache before
attempting it. A user reported that restic printed errors like this:
```
Load(<data/61db3a5d4c>, 0, 0) returned error, retrying after 689.290829ms: Create: open /tmp/restic-check-cache-551233426/a14636d2d23f0d08914fe85b74cf8a7828f73b5d6e6c465b6795086221200faf/data/61/61db3a5d4c79ca52bd1ddbe0f833480d9768b6af792affdb582d927764a11af8: file exists
```

This means that a goroutine was attempting to download a file, but by
the time it got to the actual download, the file was already there. We
now recheck if the file has been downloaded in the meantime.

### DIFF
--- a/internal/cache/backend.go
+++ b/internal/cache/backend.go
@@ -107,16 +107,17 @@ func (b *Backend) cacheFile(ctx context.Context, h restic.Handle) error {
 	}
 
 	// test again, maybe the file was cached in the meantime
-	if b.Cache.Has(h) {
-		return nil
-	}
+	if !b.Cache.Has(h) {
 
-	err := b.Backend.Load(ctx, h, 0, 0, func(rd io.Reader) error {
-		return b.Cache.Save(h, rd)
-	})
-	if err != nil {
-		// try to remove from the cache, ignore errors
-		_ = b.Cache.Remove(h)
+		// nope, it's still not in the cache, pull it from the repo and save it
+
+		err := b.Backend.Load(ctx, h, 0, 0, func(rd io.Reader) error {
+			return b.Cache.Save(h, rd)
+		})
+		if err != nil {
+			// try to remove from the cache, ignore errors
+			_ = b.Cache.Remove(h)
+		}
 	}
 
 	// signal other waiting goroutines that the file may now be cached
--- a/internal/cache/backend.go
+++ b/internal/cache/backend.go
@@ -106,6 +106,11 @@ func (b *Backend) cacheFile(ctx context.Context, h restic.Handle) error {
 		return nil
 	}
 
+	// test again, maybe the file was cached in the meantime
+	if b.Cache.Has(h) {
+		return nil
+	}
+
 	err := b.Backend.Load(ctx, h, 0, 0, func(rd io.Reader) error {
 		return b.Cache.Save(h, rd)
 	})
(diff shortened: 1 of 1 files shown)


ifedorenko:
Are you sure the "other" goroutine successfully downloaded the file? From what I can tell the problem can be explained by a failure on the other goroutine.  Here is execution schedule returns success when download failed:

* gr1: start download
* gr2: block on gr1 download
* gr1: detect and return error
* gr2: unblock and return success

Or do I misunderstand how the caching works here?

ifedorenko:
Never mind. Failed download is another problem.

fd0:
Here's a bit more log:

```
cache/backend.go:174        cache.(*Backend).Load       300           performing readahead for <data/61db3a5d4c>
cache/file.go:110           cache.(*Cache).Save         300           Save to cache: <data/61db3a5d4c>
cache/file.go:89            cache.(*Cache).SaveWriter   300           Save to cache: <data/61db3a5d4c>
[...]
cache/file.go:44            cache.(*Cache).Load         300           Load from cache: <data/61db3a5d4c>
cache/backend.go:174        cache.(*Backend).Load       286           performing readahead for <data/61db3a5d4c>
cache/file.go:110           cache.(*Cache).Save         286           Save to cache: <data/61db3a5d4c>
cache/file.go:89            cache.(*Cache).SaveWriter   286           Save to cache: <data/61db3a5d4c>
Load(<data/61db3a5d4c>, 0, 0) returned error, retrying after 689.290829ms: Create: open /tmp/restic-check-cache-551233426/a14636d2d23f0d08914fe85b74cf8a7828f73b5d6e6c465b6795086221200faf/data/61/61db3a5d4c79ca52bd1ddbe0f833480d9768b6af792affdb582d927764a11af8: file exists
```

The third column is the goroutine number. What I reconstructed was:
 * Goroutine 300 got a request to read from the file `61db3a5d4c`, it is not cached yet but should be, so it downloads the file
   - it calls `cacheFile` here: https://github.com/restic/restic/blob/b1e1b71bab03a0bb8b56b94443d95ff962169c89/internal/cache/backend.go#L176
   - a new channel is inserted into the `inProgress` map and Goroutine 300 starts downloading
 * Goroutine 286 also got a request to read the same file, it is not cached yet, and the goroutine at this point executes something around line 172 here: https://github.com/restic/restic/blob/b1e1b71bab03a0bb8b56b94443d95ff962169c89/internal/cache/backend.go#L172
 * Goroutine 300 finishes the download, removes the entry from the `inProgress` map, and serves the request from the cache, the file exists now. The debug message indicating this is, `Load from cache: <data/61db3a5d4c>`
 * Goroutine 286 continues executing, it detected that the file is not yet cached, so it attempts to download it (`performing readahead [...]`)
   - `cacheFile()` does not find an entry in the `inProgress` map (the other goroutine has removed it because it was done), so it starts downloading
   - then, the file in the cache is opened with `O_EXCL`
   - it exists already (created by the other goroutine), so that fails with `file exists`
 * The generic retry code kicks in and tries to load the file again, now the file is cached and all is good

So, now we're checking if the file has been cached in `cacheFile()` again. Alternatively, we could just leave the closed channel in `inProgress` after downloading, but that felt messy to me. It's still not perfect, but the probability that this happens again is much lower.

Thank you for asking btw, I saw a small bug and will push an new commit: in `cacheFile`, we're leaving an open channel in the `inProgress` map. That should be cleaned up before returning.

PR #2025, opened 2018-10-05
b2: simplify object iteration

kurin:
Blazer is moving to a simpler object list interface, so I'm changing
this here as well.


kurin:
I can't reproduce this error.  Is this test flaky?  It happened only on macOS.

fd0:
Yes, that happens sometimes. I haven't been able to find out what's going on. I've restarted the test.

fd0:
Thanks!

PR #2029, opened 2018-10-05
Escape subcommand in docker build script

### LINKED ISSUE #2001: Docker build.sh improper escape
Output of `restic version`
--------------------------
N/A, but this occurs on current master, 6bc99ce4 and tag v0.9.2, beb208e1

N/A

N/A

Expected behavior
-----------------
When building in a path which contains spaces, `docker run` succeeds.

Actual behavior
---------------
`pwd` is not escaped, so paths with spaces cause `docker run` to fail.

Simply run `docker/build.sh` in a directory with spaces:
```bash
git clone https://github.com/restic/restic/ "res tic"
cd "res tic"
docker/build.sh
```
which prints
```
Build binary using golang docker image
docker: invalid reference format.
See 'docker run --help'.
```

Yes: unescaped pwd

Yes. 
```
`pwd`
```
should be escaped as
```
"`pwd`"
```

Yeah. I loved borg, but it does not distribute static binaries, so I had to compile it from source for android, which constantly caused problems. Having a single static binary backup utility (specifically for arm/aarch64) has been pretty great.

j6s:
This change escapes the subcall to `pwd` in `build.sh` in order for it to work in directories containing spaces.

Closes #2001
This is **not my fix**. This PR simply contains the changes proposed in that issue.

Checklist
---------

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -4,7 +4,7 @@ set -e
 
 echo "Build binary using golang docker image"
 docker run --rm -ti \
-    -v `pwd`:/go/src/github.com/restic/restic \
+    -v "`pwd`":/go/src/github.com/restic/restic \
     -w /go/src/github.com/restic/restic golang:1.8.3-alpine go run build.go
 
 echo "Build docker image restic/restic:latest"


fd0 (approved):
Awesome, thanks! It still fails for corner cases, like when the current directory contains double quotes in its name, but it improves the situation.

stonewareslord:
@fd0 When does it fail? I just tried the following in bash and it succeeds:

```
mkdir "a \" b \' c"
cd "a \" b \' c"
pwd        # Prints: /tmp/a " b \' c
cd `pwd`   # Obviously fails; too many arguments
cd "`pwd`" # Succeeds since shell escapes it
```

I would expect that this solution covers all edge cases, but I may be wrong.

fd0:
Huh, interesting. It may still be different in a `Dockerfile`, but until we get any reports that it fails let's just keep it as it is :)

PR #2032, opened 2018-10-07
Add options for case insensitive includes & excludes

### LINKED ISSUE #1895: Support a case insensitive flag for include/exclude patterns
It would be helpful if the the include/exclude patterns could support a case-insensitive flag or prefix. I'm primarily a Windows user and the case sensitivity has caught me out several times already, but other than helping to avoid such accidents, I would also find it useful generally.

### LINKED ISSUE #1985: Add adjustable utilization threshold for re-writing packs during prune
Output of `restic version`
--------------------------

Currently during a "prune" operation, restic will re-write any pack that contains any unused data. This can trigger a lot of read/write operations as it has to read the data from every affected pack and then write that data back to the backend. This can become expensive on metered backends.

I would like to see an option to the prune command that allows the user to specify the amount of unused space allowed in a pack before it is re-written. Setting this option to 0% would cause packs with any amount of unused data to be re-written (the current behavior), whereas setting this option to 100% would only re-write packs that are completely unused. 

Note this feature was originally suggested as part of a broader discussion in issue [#1470 Parallelize prune pack rewrites](1470).

What are you trying to do?
--------------------------

Adding this feature would give the user some degree of control over how much read/write traffic is generated by a prune operation, with the trade-off being how much unused data is left behind in the repository.

I discovered restic like many others after CrashPlan discontinued their "Home" plan and I needed to find a suitable alternative. restic was the perfect fit, allowing me to choose different backend services as well as being a command line based utility, which allows me to run it easily on my headless server. I have been recommending it since to friends and co-workers. Thanks! :+1:

j6s:
This change adds the option `--exclude-insensitive` to the `backup` and `restore` commands and the `--include-insensitive` option to the `restore` command.

These options allow for case insensitive filtering in those commands.
Take the following directory structure:
```
restic_source/file1.txt
restic_source/ignore_me/file2.txt
restic_source/IGNORE_ME/file3.txt
```

Running `$ restic backup --exclude 'ignore_me'` would ignore `file2.txt` but not `file3.txt` as the paths do not match.

Using the new flag and running `$ restic backup --exclude 'ignore_me' --exclude-insensitive` would ignore both paths as the exclude patter is now insensitive.

Resolves #1895 

Additional Questions
------------------------
This turned out to be a more in-depth PR than I would have thought. This PR implements these options in the commands themselves - however it may be better to move this kind of case-insensitive option to the filter methods.

Also: Should I implement a test for these flags? 
I don't really know how testing in go works but I would try my best.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -28,13 +28,15 @@ repository.
 
 // RestoreOptions collects all options for the restore command.
 type RestoreOptions struct {
-	Exclude []string
-	Include []string
-	Target  string
-	Host    string
-	Paths   []string
-	Tags    restic.TagLists
-	Verify  bool
+	Exclude            []string
+	InsensitiveExclude []string
+	Include            []string
+	InsensitiveInclude []string
+	Target             string
+	Host               string
+	Paths              []string
+	Tags               restic.TagLists
+	Verify             bool
 }
 
 var restoreOptions RestoreOptions
@@ -44,7 +46,9 @@ func init() {
 
 	flags := cmdRestore.Flags()
 	flags.StringArrayVarP(&restoreOptions.Exclude, "exclude", "e", nil, "exclude a `pattern` (can be specified multiple times)")
+	flags.StringArrayVar(&restoreOptions.InsensitiveExclude, "iexclude", nil, "same as `--exclude` but ignores the casing of filenames")
 	flags.StringArrayVarP(&restoreOptions.Include, "include", "i", nil, "include a `pattern`, exclude everything else (can be specified multiple times)")
+	flags.StringArrayVar(&restoreOptions.InsensitiveInclude, "iinclude", nil, "same as `--include` but ignores the casing of filenames")
 	flags.StringVarP(&restoreOptions.Target, "target", "t", "", "directory to extract data to")
 
 	flags.StringVarP(&restoreOptions.Host, "host", "H", "", `only consider snapshots for this host when the snapshot ID is "latest"`)
@@ -55,6 +59,8 @@ func init() {
 
 func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 	ctx := gopts.ctx
+	hasExcludes := len(opts.Exclude) > 0 || len(opts.InsensitiveExclude) > 0
+	hasIncludes := len(opts.Include) > 0 || len(opts.InsensitiveInclude) > 0
 
 	switch {
 	case len(args) == 0:
@@ -67,7 +73,7 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 		return errors.Fatal("please specify a directory to restore to (--target)")
 	}
 
-	if len(opts.Exclude) > 0 && len(opts.Include) > 0 {
+	if hasExcludes && hasIncludes {
 		return errors.Fatal("exclude and include patterns are mutually exclusive")
 	}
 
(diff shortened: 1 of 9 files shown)


rawtaz:
What about less redundantly named options, like `--iexclude` and `--iinclude`? There's a lot of repeating in the above examples.

cmd/restic/cmd_restore.go
```diff
@@ -125,11 +131,16 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 			Warnf("error for exclude pattern: %v", err)
 		}
 
+		matchedInsensitive, _, err := filter.InsensitiveList(opts.InsensitiveExclude, item)
```
fd0: Unfortunately this is inefficient and does not work:
 * The function has a side effect of converting all strings in `opts.InsensitiveExclude` to lower case over and over again for every file that is to be restored
 * The `item` needs to be converted to lower case before passing it to the function

I suggest removing `InsensitiveList`, converting `opts.InsensitiveExclude` to lower case once, and then using `filter.List(opts.InsensitiveExclude, strings.ToLower(item)` instead. Thoughts?

cmd/restic/cmd_restore.go
```diff
@@ -141,15 +152,20 @@ func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
 			Warnf("error for include pattern: %v", err)
 		}
 
-		selectedForRestore = matched
-		childMayBeSelected = childMayMatch && node.Type == "dir"
+		matchedInsensitive, childMayMatchInsensitive, err := filter.InsensitiveList(opts.InsensitiveInclude, item)
```
fd0: Same as above for include.

cmd/restic/exclude.go
```diff
@@ -88,6 +88,16 @@ func rejectByPattern(patterns []string) RejectByNameFunc {
 	}
 }
 
+// Same as `rejectByPattern` but case insensitive.
+func rejectByInsensitivePattern(patterns []string) RejectByNameFunc {
+	for index, path := range patterns {
+		patterns[index] = strings.ToLower(path)
+	}
+	return func(item string) bool {
+		return rejectByPattern(patterns)(strings.ToLower(item))
```
fd0: For every file that is to be tested during backup, `rejectByPattern()` creates a new function that is instantly evaluated. What about creating the function just once, outside the closure, and then use it inside?
```suggestion
	rejFunc := rejectByPattern(patterns)
	return func(item string) bool {
		return rejFunc(strings.ToLower(item))
```

internal/filter/filter.go
```diff
@@ -187,3 +187,12 @@ func List(patterns []string, str string) (matched bool, childMayMatch bool, err
 
 	return matched, childMayMatch, nil
 }
+
+// InsensitiveList is the same as List but case insensitive.
+func InsensitiveList(patterns []string, str string) (matched bool, childMayMatch bool, err error) {
```
fd0: I'd remove this as per my suggestion above and just use one list.

fd0 (changes requested):
Thank you very much for your contribution! Most looks great already, there are a few things that I think needs changes, I've added a few comments. Please let me know if you need any help!

I've also taken the liberty of rebasing on the latest master branch and resolving the merge conflict there.

fd0:
Hey, are you still interested in working on this?

j6s:
Sorry -

totally forgot about this PR.
Will make the requested changes shortly.

fd0:
Awesome, thanks!

fd0 (approved):
LGTM, thanks!

PR #2033, opened 2018-10-07
Output directory size in cache command

### LINKED ISSUE #2028: Feature: Print size of cache directories
Output of `restic version`
--------------------------
Current master: 
```
$  ./restic version 
restic 0.9.2 (v0.9.2-130-g0916ff71-dirty) compiled with go1.10.4 on linux/amd64
```

It would be nice if restic would print the total size of the cache in the table outputted by `restic cache`.  This would increase the usefulness of the command.

Currently the output of the command is as follows (in this instance including the output of #2027:
```
$ ./restic cache
Repo ID     Last Used     Old
-----------------------------
426f9cb6c8  244 days ago  yes
469099403b  0 days ago
-----------------------------
2 cache dirs in /home/j/.cache/restic
```

Having an additional column containing the size of the cache directory would be cool ideal.

```
$ du -h --max-depth=1 $HOME/.cache/restic/
362M	/home/j/.cache/restic/426f9cb6c8d17fa92318ac340a68a4480b685f4fb9df864e3ecbde820e5f5afd
682M	/home/j/.cache/restic/469099403b967aaba024805d666f87ec0392828656ed61ac2ff75df1ff124aa6

$ ./restic cache 
Repo ID     Last Used     Old    Size
--------------------------------------
426f9cb6c8  244 days ago  yes    326M
469099403b  0 days ago           682M
--------------------------------------
2 cache dirs in /home/j/.cache/restic
```

Restic backs up all of my data as well as most people around me. About half of my NAS is filled with restic repositories. Thankfully I have never been in the situation where I needed those backups - but restic helps me sleep at night.

j6s:
This PR adds an additional column to the table printed by `restic cache` that contains the size of the cache directory. There is also a new flag `--no-size` which disables this column as computation can be expensive for large cache directories.

```
$ ./restic cache 
Repo ID     Last Used     Old  Size
------------------------------------------
426f9cb6c8  246 days ago  yes  360.365 MiB
469099403b  1 days ago         679.077 MiB
8af3cec4e8  1 days ago           8.097 KiB
8a28008e9b  0 days ago          11.687 KiB
------------------------------------------
4 cache dirs in /home/j/.cache/restic
```

```
$ ./restic cache --no-size
Repo ID     Last Used     Old
-----------------------------
426f9cb6c8  246 days ago  yes
469099403b  1 days ago
8af3cec4e8  1 days ago
8a28008e9b  0 days ago
-----------------------------
4 cache dirs in /home/j/.cache/restic
```

Resolves #2028 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_cache.go
+++ b/cmd/restic/cmd_cache.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"sort"
 	"time"
@@ -29,6 +30,7 @@ The "cache" command allows listing and cleaning local cache directories.
 type CacheOptions struct {
 	Cleanup bool
 	MaxAge  uint
+	NoSize  bool
 }
 
 var cacheOptions CacheOptions
@@ -39,6 +41,7 @@ func init() {
 	f := cmdCache.Flags()
 	f.BoolVar(&cacheOptions.Cleanup, "cleanup", false, "remove old cache directories")
 	f.UintVar(&cacheOptions.MaxAge, "max-age", 30, "max age in `days` for cache directories to be considered old")
+	f.BoolVar(&cacheOptions.NoSize, "no-size", false, "do not output the size of the cache directories")
 }
 
 func runCache(opts CacheOptions, gopts GlobalOptions, args []string) error {
@@ -92,11 +95,16 @@ func runCache(opts CacheOptions, gopts GlobalOptions, args []string) error {
 		ID   string
 		Last string
 		Old  string
+		Size string
 	}
 
 	tab.AddColumn("Repo ID", "{{ .ID }}")
 	tab.AddColumn("Last Used", "{{ .Last }}")
 	tab.AddColumn("Old", "{{ .Old }}")
+
+	if !opts.NoSize {
+		tab.AddColumn("Size", "{{ .Size }}")
+	}
 
 	dirs, err := cache.All(cachedir)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/restic/cmd_cache.go
```diff
@@ -118,10 +126,20 @@ func runCache(opts CacheOptions, gopts GlobalOptions, args []string) error {
 			old = "yes"
 		}
 
+		var size string
+		if !opts.NoSize {
+			bytes, err := dirSize(fmt.Sprintf("%s/%s", cachedir, entry.Name()))
```
fd0: Use `filepath.Join(cachedir, entry.Name())`, that's the correct function and also works on Windows.

cmd/restic/cmd_cache.go
```diff
@@ -130,3 +148,14 @@ func runCache(opts CacheOptions, gopts GlobalOptions, args []string) error {
 
 	return nil
 }
+
+func dirSize(path string) (int64, error) {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if !info.IsDir() {
```
fd0: You need to check `err` first, return it if it's non-nil, then check if `info` is nil (which may happen) and return otherwise. and only afterwards access `info.Size()`.

fd0 (changes requested):
LGTM, two small issues. Thank you very much!

fd0 (approved):

fd0:
Thank you!

Issue #1631, opened 2018-02-19
Very slow backup of SSHFS filesystems -- need ability to skip inode-based comparison


McKael:
## Output of `restic version`

restic 0.8.2

## How did you run restic exactly?

```
% restic -r /tmp/restictest backup $SSHFSMOUNT/directory
```

The backup directory is a sshfs mounted filesystem.
The command is run several times to do several snapshots.
In mly case there is a huge amount of data (1TB+) in the remote directory.

## What backend/server/service did you use to store the repository?

Local ext4 filesystem.

## Expected behavior

I expect the first backup to be slow and transfer much data over LAN, while the next backups should be pretty fast and should not use much bandwidth.

## Actual behavior

All backups take hours (if not days) even if no files have changed.

## Steps to reproduce the behavior

- run a backup
- unmount sshfs filesystem
- re-mount sshfs filesystem
- run a new backup

It is not 100% reproducible but even with a small amount of data I could reproduce it.

The SFTP serrver logs show that files are completely retrieved even when they did not change.

## Do you have any idea what may have caused this?

Yes: restic compares inodes to double-check if files have been modified (debug message "timestamp, size or inode changed", `restic/node.go:551restic.(*Node).IsNewer11node`).

However, inodes might change across filesystem mounts with sshfs (and probaly some other filesystems).

## Do you have an idea how to solve the issue?

Commenting out the inode check solved the problem for me.

I'd like to have a way to disable this check; maybe a command-line flag?

## Did restic help you or made you happy in any way?

Sure, it's a nice piece of software!  I'm all the more pleased since I have found a workaround...
Keep up the good work!

fd0:
Thanks for the report, this is indeed caused by restic detecting that files have changed based on the inode. For fuse-based file systems, this check is not great, we should instead only check the timestamps and the file size.

In principle this can be detected automatically, so we may not even need a command-line flag.

ebastos:
> In principle this could also be detected automatically (by looking at the file system name, and keeping a black list of know-inode-unstable filesystems) so we may not even need a command-line flag.

How do you envision this? I could give it a shot...

cyphar:
#2205 was merged and is in `0.9.5`, this should be closed. :wink:

fd0:
You're right, thanks for the hint!

PR #2047, opened 2018-10-16
Add --ignore-inode option to backup cmd.

giacomocariello:

This PR adds --ignore-inode option to backup command, in order to support those filesystems that don't support fixed inodes, typically Fuse, pCloud, etc.

It was already discussed #1631.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2047
+++ b/changelog/unreleased/pull-2047
@@ -1,0 +1,9 @@
+Enhancement: Add --ignore-inode option to backup cmd
+
+This option handles backup of virtual filesystems that do not keep fixed
+inodes for files, like Fuse-based, pCloud, etc. Ignoring inode changes allows
+to consider the file as unchanged if last modification date and size
+are unchanged.
+
+https://github.com/restic/restic/pull/2047
+https://github.com/restic/restic/issues/1631
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -76,6 +76,7 @@ type BackupOptions struct {
 	FilesFrom        string
 	TimeStamp        string
 	WithAtime        bool
+	IgnoreInode      bool
 }
 
 var backupOptions BackupOptions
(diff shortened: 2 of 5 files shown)


pabs3:
This reduced the amount of time it took for converting my rdiff-backup snapshots to restic snapshots from over an hour to under 8 minutes. I would love to see it merged into restic mainline.

fd0 (changes requested):
Thank you very much for your contribution! I've had a look at the code, while it get's the job done, a *lot* of places must be touched to pass in an option of type `ModifiedIgnores`, which only has one field. That feels overly generic to me.

I'd like to propose a different approach:
 * Add a new field to the `Archiver` struct, `IgnoreInode bool`
 * Change the signature of `fileChanged` like this:
```go
func fileChanged(fi os.FileInfo, node *restic.Node, ignoreInode bool) bool
```
 * Pass the `arch.IgnoreInode` to `fileChanged`

The default value for the bool is false, so the inode is checked.

What do you think? Please let me know if you need any help!

Other nitpick: Can you please remove the trailing `.` from the summary line fro the commit messages? Thanks!

giacomocariello:
@fd0 yup, no problem on my side. I'll try and apply your requests. The idea of having a struct to handle multiple types of ignore flags was probably overengineered at this stage.

fd0:
Hey, are you still interested in working on this? Is there anything you need, or are you stuck?

robvalca:
I'm really interested in this feature, do you know if it will go in, finally? :)

fd0:
@robvalca this PR is incomplete, it looks like @giacomocariello did not have the time to rework it yet. So we need to wait until this happens (or somebody else volunteers to do it) :)

robvalca:
@fd0, @giacomocariello  do you think that I could take on this? I'm newbie in golang (and btw in github collab) but with the work already done and some help I think that I could finish this. :)

PR #2050, opened 2018-10-17
tiny formatting fix

andreasnuesslein:
I noticed this discrepancy when trying to parse the output of restic.


fd0 (approved):
Great, thanks for taking the time!

PR #2055, opened 2018-10-19
Correct ineffassign

gliptak:


internal/repository/packer_manager_test.go
```diff
@@ -91,7 +91,7 @@ func fillPacks(t testing.TB, rnd *randReader, be Saver, pm *packerManager, buf [
 			t.Fatal(err)
 		}
 
-		n, err := packer.Add(restic.DataBlob, id, buf)
+		n, _ := packer.Add(restic.DataBlob, id, buf)
```
fd0: Uh, this is actually a bug! You're right, the `err` will not be handled, but instead of ignoring it, please use the following:
```suggestion
		n, err := packer.Add(restic.DataBlob, id, buf)
		if err != nil {
			t.Fatal(err)
		}
```

fd0 (changes requested):
Thanks for taking the time to fix this!

internal/repository/packer_manager_test.go
gliptak: updated

fd0 (approved):

Issue #2085, opened 2018-11-12
--files-from cannot be specified more than once


Adam-Ant:
Output of `restic version`
--------------------------
`restic 0.9.3 compiled with go1.11.1 on linux/amd64`

How did you run restic exactly?
-------------------------------
Manually from the command line, backing up to OpenStack Swift. Restic and Swift config exposed via env vars. Exact command run is:
` restic --tag automated backup --files-from /path/to/list.1 --files-from /path/to/list.2`

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
OpenStack Swift, hosted with OVH

Expected behavior
-----------------
Restic backs up files from both lists, a similar behavior to how `--exclude-file` works.

Actual behavior
---------------
Restic only backs up files from the last list given, eg `/path/to/list.2`

Steps to reproduce the behavior
-------------------------------
Run restic with multiple `--files-from` arguments, as above

Do you have any idea what may have caused this?
-----------------------------------------------
Not a shiny shoe! :)

Do you have an idea how to solve the issue?
-------------------------------------------
Alas no, I have very little experience with Go. 

Did restic help you or made you happy in any way?
-------------------------------------------------
Restic has made my backups so much easier to manage, and has saved me a good chunk of money each month by allowing me to move to using Openstack Swift as a backend :)

fd0:
Thanks for the report! I'm not sure if this is a bug or a feature request, but I can see that users expect this to work.

frebib:
I don't use Restic but from looking at the code would `cat some/files/*.whatever | restic --files-from -` work?

fd0:
Thanks for the PR!

PR #2086, opened 2018-11-12
Allow --files-from multiple times

frebib:
Should address issue in https://github.com/restic/restic/issues/2085

Allow --files-from multiple times

Checklist
---------

There are some documentation/admin tasks I still need to tend to as above but provided the code review looks good I'll sort those before merging :)

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -45,8 +45,12 @@ given as the arguments.
 	},
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if backupOptions.Stdin && backupOptions.FilesFrom == "-" {
-			return errors.Fatal("cannot use both `--stdin` and `--files-from -`")
+		if backupOptions.Stdin {
+			for _, filename := range backupOptions.FilesFrom {
+				if filename == "-" {
+					return errors.Fatal("cannot use both `--stdin` and `--files-from -`")
+				}
+			}
 		}
 
 		var t tomb.Tomb
@@ -75,7 +79,7 @@ type BackupOptions struct {
 	StdinFilename    string
 	Tags             []string
 	Host             string
-	FilesFrom        string
+	FilesFrom        []string
 	TimeStamp        string
 	WithAtime        bool
 }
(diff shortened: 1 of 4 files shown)


cmd/restic/cmd_backup.go
```diff
@@ -175,12 +175,12 @@ func readLinesFromFile(filename string) ([]string, error) {
 
 // Check returns an error when an invalid combination of options was set.
 func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
-	if opts.FilesFrom == "-" && gopts.password == "" {
+	if opts.FilesFrom != nil && opts.FilesFrom[0] == "-" && gopts.password == "" {
```
fd0: Use `len(opts.FilesFrom) == 0` instead of the `nil` check: A slice may be non-nil and still have zero elements in it. Also, this only catches `--files-from -` when it is the first `--files-from` option... We need to loop over all elements of the slice and check for `-`.
```suggestion
	if gopts.password == "" {
		for _, filename := range opts.FilesFrom {
			if filename == "-" {
				return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
			}
		}
	}
```

cmd/restic/cmd_backup.go
```diff
@@ -175,12 +175,12 @@ func readLinesFromFile(filename string) ([]string, error) {
...
 		return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
 	}
 
 	if opts.Stdin {
-		if opts.FilesFrom != "" {
+		if opts.FilesFrom != nil {
```
fd0: A slice can be non-nil and still have zero elements, better check its length:
```suggestion
		if len(opts.FilesFrom) > 0 {
```

fd0 (changes requested):
Just a few small issues, otherwise looks good, thanks!

cmd/restic/cmd_backup.go
frebib: This new review suggestion thing is cool!
Thanks for the info on nil/empty slices, I'm new to Go so I didn't know about that difference.

From what I can see the `--files-from -` behaviour is completely broken (with and without this patch), but that's another issue entirely.

frebib:
I couldn't find anywhere to add appropriate tests and I couldn't find any tests relating to the cmdline arguments either, so I've left that.

Other than that I think this is good

cmd/restic/cmd_backup.go
fd0: Yeah, we need to fix that, but it's a different issue so leave it for now.

fd0 (approved):

fd0:
Thanks! I've taken the liberty of squashing the commits and rewording the changelog, after the tests pass it's good to merge!

cmd/restic/cmd_backup.go
fd0: Hm, I had a look, and the `--files-from -` works for me as expected. What do you think is broken?

fd0:
Hm, I had a look, and the `--files-from -` works for me as expected. What do you think is broken?

frebib:
I'm not sure. It's probably fine

PR #2087, opened 2018-11-14
Add GroupBy option to snapshots command

### LINKED ISSUE #2037: --group-by option for snapshots command
Output of `restic version`
--------------------------

restic compiled manually compiled with go1.10.3 on freebsd/amd64

What are you trying to do?
--------------------------

it would be great to have the --group-by option available for the snapshots command as well as the forget command. It makes it much easier to see the sets of backups marked with tags.

`restic snapshots -g tags`

Similar grouped output as you see when you run `restic forget -g tags ...`

```
snapshots for (tags [monthly]):
      ...
snapshots for (tags [daily]):
      ...
```

Ref: https://forum.restic.net/t/feature-group-by-option-for-snapshots-command/1070/4

It makes me happy every day to know my backups are safe and secured :)

ArcticXWolf:
This commit adds a --group-by option to the snapshots command, which
behaves similar to the --group-by option of forget. Valid option values
are "host, paths, tags". If this option is given, the output of
snapshots will be divided into multiple tables, according to the value
given (i.e. "host" will create a table of snapshots for each host, that
has a snapshot in the list). Also the JSON output will be grouped.

The default behavior (when --group-by is not given) has not changed.

More to this discussion can be found in issue #2037.

Yes, it was discussed in issue #2037.

Closes #2037

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -8,7 +8,6 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/ui/table"
 	"github.com/spf13/cobra"
@@ -50,12 +49,6 @@ func init() {
 	f.StringVarP(&snapshotOptions.GroupBy, "group-by", "g", "", "string for grouping snapshots by host,paths,tags")
 }
 
-type groupKey struct {
-	Hostname string   `json:"hostname"`
-	Paths    []string `json:"paths"`
-	Tags     []string `json:"tags"`
-}
-
 func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) error {
 	repo, err := OpenRepository(gopts)
 	if err != nil {
@@ -77,7 +70,7 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args) {
 		snapshots = append(snapshots, sn)
 	}
-	snapshotGroups, grouped, err := GroupSnapshots(snapshots, opts.GroupBy)
+	snapshotGroups, grouped, err := restic.GroupSnapshots(snapshots, opts.GroupBy)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


ArcticXWolf:
Hey, I want to work on the group-by option discussed in #2037. The main functionality is already done, whats missing are tests, documentation and changelogs.

As this is my first contribution to restic and also my first time coding in Go, I have some questions:

1. What's used to do the testing? (As I'm coming from Ruby, I know that there are many testing frameworks/DSLs)
2. Is the style of my changes okay? I'm not sure, if I violated any naming conventions or else.

I hope my contribution helps!

cmd/restic/cmd_snapshots.go
```diff
@@ -61,28 +70,87 @@ func runSnapshots(opts SnapshotOptions, gopts GlobalOptions, args []string) erro
 		}
 	}
 
+	// group by hostname and dirs
```
fd0: This is duplicate code, I would suggest extracting it into a function which takes a list of snapshots (`[]restic.Snapshot`, called a "slice" in Go) and returns a `map[string][]Snapshot` and an `error`. Let me know if you need help with this.

cmd/restic/cmd_snapshots.go
```diff
@@ -223,6 +291,40 @@ func PrintSnapshots(stdout io.Writer, list restic.Snapshots, reasons []restic.Ke
 	tab.Write(stdout)
 }
 
+// PrintSnapshotGroupHeader prints which group of the group-by option the
+// following snapshots belong to.
+// Prints nothing, if we did not group at all.
+func PrintSnapshotGroupHeader(stdout io.Writer, groupKeyJSON string, GroupByTag bool, GroupByHost bool, GroupByPath bool) error {
+	if GroupByTag || GroupByHost || GroupByPath {
```
fd0: It's common in Go to return early, so the actual action in the function is not indented so far. How about this instead?
```suggestion
	if !GroupByTag && !GroupByHost && !GroupByPath {
		return nil
	}
```

cmd/restic/cmd_snapshots.go
```diff
@@ -231,20 +333,60 @@ type Snapshot struct {
 	ShortID string     `json:"short_id"`
 }
 
+// SnapshotGroup helps to print SnaphotGroups as JSON with their GroupReasons included.
+type SnapshotGroup struct {
+	GroupKey  groupKey
```
fd0: Add so-called struct tags here to set the names in the JSON output
```suggestion
	GroupKey  groupKey `json:"group_key"`
```

cmd/restic/cmd_snapshots.go
```diff
@@ -231,20 +333,60 @@ type Snapshot struct {
...
+	Snapshots []Snapshot
```
fd0:
```suggestion
	Snapshots []Snapshot `json:"snapshots"`
```

cmd/restic/cmd_snapshots.go
```diff
@@ -45,6 +47,13 @@ func init() {
 	f.StringArrayVar(&snapshotOptions.Paths, "path", nil, "only consider snapshots for this `path` (can be specified multiple times)")
 	f.BoolVarP(&snapshotOptions.Compact, "compact", "c", false, "use compact format")
 	f.BoolVar(&snapshotOptions.Last, "last", false, "only show the last snapshot for each host and path")
+	f.StringVarP(&snapshotOptions.GroupBy, "group-by", "g", "", "string for grouping snapshots by host,paths,tags")
+}
+
+type groupKey struct {
+	Hostname string
```
fd0: Add so-called struct tags here to set the name for the JSON output (lowercase, snake_case)
```suggestion
	Hostname string `json:"hostname"`
```

fd0 (changes requested):
Hey, I've had a look at the code, thanks for your contribution (especially as a first-time contributor new to Go)!

I've got a few things that you could improve in the code, but it looks great already.

 * The Tests on Travis fail because the file `internal/archiver/testing_test.go` is not formatted properly with `gofmt` (the Go formatting standard is an implementation, rather than a document), which has changed a bit recently in the Go 1.11 release. In order to pass the tests, you need to format the code with `gofmt` on Go 1.11.
 * There are modifications to files in `vendor`/, these should not be there. Did you maybe run `gofmt` on all files, including `vendor/`? Please remove the modifications from the PR.

For testing, we're using the builtin testing framework, you can easily run all tests with `go test ./...` in the repo, or chdir to some directory and just run `go test`.

Unfortunately, we don't have tests for user-interface interactions like this in place (at least not yet).

Let me know if you need any help!

cmd/restic/cmd_snapshots.go
ArcticXWolf: Okay, I extracted it into an own function. Is the style of the new function okay? And shall I remove the duplicate code (in the forget command) in this pull request or shall I open a new one after this got merged?

ArcticXWolf:
Okay, this is nearly complete, I just have some last questions:

> For testing, we're using the builtin testing framework, you can easily run all tests with go test ./... in the repo, or chdir to some directory and just run go test.
> 
> Unfortunately, we don't have tests for user-interface interactions like this in place (at least not yet).

The tests succeed with my changes, but shall I add new ones? I have no idea how I should design such UI tests. Maybe just save the expected output into a fixture?

Furthermore I added a changelog file, whats missing now is the modification of the documentation. Where shall I mention it? Maybe at [the working with repos section](https://github.com/restic/restic/blame/master/doc/045_working_with_repos.rst#L56)?

cmd/restic/cmd_snapshots.go
ArcticXWolf: I put the shared code into the internal folder and made the forget command use it too.

ArcticXWolf:
Okay, I refactored the forget command to use the shared code and added the documentation. Tests are green for me, just waiting for the CI to confirm this. As I do not know how to do the interface tests (and we did not have them before), I believe this is ready for review.

fd0 (approved):
LGTM, thanks! Sorry for the long delay before reviewing this!

fd0:
I've rebased and fixed the conflict on `cmd_forget.go`, let's see what the tests say.

Issue #2089, opened 2018-11-16
Increase the granularity of the "keep within" snapshot retention policy


plumbeo:
Output of `restic version`
--------------------------

restic 0.9.3 compiled with go1.11.1 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

`restic forget --keep-within` currently accepts time ranges with a minimum granularity of one day. It would be nice to be able to pass to the command ranges measured in hours.

What are you trying to do?
--------------------------

I have a few clients that make a backup daily at a random time between 14 and 18 when they're powered on and connected to the internal network and I want to apply a retention policy that keeps the snapshots made in the last N days.

Neither `--keep-daily N` nor `--keep-last N` help because the number of snapshots taken in the range can change and `--keep-within=Nd` preserves the Nth day snapshot or deletes it according to the (random) times it and the last snapshot were made. 

I'd like to use `--keep-within=Nd6h` to consistently keep the Nth day snapshot.

Did restic help you or made you happy in any way?
-------------------------------------------------

Yes, it's great to be able to backup our data on local hard drives, on a local (trusted) storage and on (untrusted) cloud storage using the same software.

dimejo:
Any reason not to use `--keep-last` in this case?

plumbeo:
> Any reason not to use `--keep-last` in this case?

There's no guarantee there will be exactly N snapshots taken in the last N days, it could be fewer.

plumbeo:
https://github.com/restic/restic/pull/2090 merged

fd0:
Ah, forgot this issue, thanks!

PR #2090, opened 2018-11-16
Increase the granularity of the "keep within" snapshot retention policy

plumbeo:

Increase the minimum granularity of the time ranges accepted by the `restic forget --keep-within` command from a daily granularity to an hourly granularity.

Yes, in issue #2089.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2089
+++ b/changelog/unreleased/issue-2089
@@ -1,0 +1,9 @@
+Enhancement: increase granularity of the "keep within" retention policy
+
+The `keep-within` option of the `forget` command now accepts time ranges with
+an hourly granularity. For example, running `restic forget --keep-within 3d12h`
+will keep all the snapshots made within three days and twelve hours from the
+time of the latest snapshot.
+
+https://github.com/restic/restic/issues/2089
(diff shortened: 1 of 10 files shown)


plumbeo:
I didn't add new tests covering the change to snapshot_policy_test.go yet because I'm not sure about how to generate the testdata.

fd0:
Nice idea, I like it! Add the tests to `snapshot_policy_test.go`, then run the tests with `go test -v -update -run TestApplyPolicy`, then inspect the new (updated) files in `testdata/`. If they look good (the results are correct), commit the files. This technique uses so-called "golden" files to compare the result of an operation with the expected result without manually having to write the long and complicated files.

fd0 (changes requested):
Add the tests, then I think this change is good to be merged. Thanks!

plumbeo:
@fd0 I just pushed a modified commit that also add tests. Also, in the option description wasn't "older" supposed to be "newer"?

fd0:
> Also, in the option description wasn't "older" supposed to be "newer"?

Huh, I think you're right!

fd0 (approved):
LGTM, thanks!

PR #2094, opened 2018-11-18
Add support for reading password from external command

juergenhoetzel:
This allows reading the password from an password manager (like [pass](https://www.passwordstore.org/))

Signed-off-by: Juergen Hoetzel <juergen@archlinux.org>

No

Checklist
---------

This is inspired by a feature of [mbsync](https://wiki.archlinux.org/index.php/isync): `mbsync` supports reading the password from an external command.

Using `pinentry-gnome3` I get a nice prompt when the GnuPG agent cache expired. 

This is my systemd/timer setup I use for doing daily backups without storing the password in a plaintext file:

* Oneshot-Service
```ini
juergen@samson:~ → systemctl --user cat restic
# /home/juergen/.config/systemd/user/restic.service
[Unit]
Description=Restic Backup

[Service]
Type=oneshot
Environment=RESTIC_PASSWORD_COMMAND='pass Backup/restic@samson'
Environment=RESTIC_REPOSITORY=sftp:restic@shaun:/ext/restic
ExecStart=/home/juergen/bin/restic backup ${HOME}/
```
* Timer
```ini
juergen@samson:~ → systemctl --user cat restic.timer
# /home/juergen/.config/systemd/user/restic.timer
[Unit]
Description=Run restic backup daily

[Timer]
OnCalendar=*-*-* 14:06:00
Persistent=true

[Install]
WantedBy=timers.target
```

### DIFF
--- a/changelog/unreleased/pull-2094
+++ b/changelog/unreleased/pull-2094
@@ -1,0 +1,8 @@
+Enhancement: Run command to get password
+
+We've added the `--password-command` option which allows specifying a command
+that restic runs every time the password for the repository is needed, so it
+can be integrated with a password manager or keyring. The option can also be
+set via the environment variable `$RESTIC_PASSWORD_COMMAND`.
+
+https://github.com/restic/restic/pull/2094
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -34,6 +34,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 
 	"golang.org/x/crypto/ssh/terminal"
+	"os/exec"
 )
 
 var version = "0.9.3-dev (compiled manually)"
@@ -43,18 +44,19 @@ const TimeFormat = "2006-01-02 15:04:05"
 
 // GlobalOptions hold all global options for restic.
 type GlobalOptions struct {
-	Repo          string
-	PasswordFile  string
-	KeyHint       string
-	Quiet         bool
-	Verbose       int
-	NoLock        bool
-	JSON          bool
-	CacheDir      string
-	NoCache       bool
-	CACerts       []string
-	TLSClientCert string
-	CleanupCache  bool
+	Repo            string
+	PasswordFile    string
+	PasswordCommand string
+	KeyHint         string
+	Quiet           bool
+	Verbose         int
+	NoLock          bool
+	JSON            bool
+	CacheDir        string
+	NoCache         bool
+	CACerts         []string
+	TLSClientCert   string
+	CleanupCache    bool
 
 	LimitUploadKb   int
 	LimitDownloadKb int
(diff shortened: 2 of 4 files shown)


juergenhoetzel:
If you are in interested in this commit I can also add the documentation.

fd0:
Hey, thank you very much for contributing to restic! Why do you think this feature is needed? You can just supply the password for restic on stdin, like `pass backup/restic | restic backup /home`...

juergenhoetzel:
> Hey, thank you very much for contributing to restic! Why do you think this feature is needed? There 
are several ways to supply the password to restic with a program:

Thanks for your feedback and especially creating `restic`! 👍 
> 
> * pipe the password into restic via stdin: `pass backup/restic | restic backup /home`...

I wasn't aware `restic` can read the password from a non-terminal device. But input redirection will lead to other issues when restic tries to read from `stdin` in subcommands like:

```bash
pass Backup/restic@samson |restic key add
```
```
main.readPassword
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/global.go:281
main.ReadPassword
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/global.go:317
main.ReadPasswordTwice
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/global.go:335
main.getNewPassword
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/cmd_key.go:107
main.addKey
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/cmd_key.go:113
main.runKey
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/cmd_key.go:194
main.glob..func12
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/cmd_key.go:26
        /home/juergen/go/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:762
        /home/juergen/go/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:852
        /home/juergen/go/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:800
main.main
        /home/juergen/go/src/github.com/restic/restic/cmd/restic/main.go:86
runtime.main
        /usr/lib/go/src/runtime/proc.go:201
runtime.goexit
        /usr/lib/go/src/runtime/asm_amd64.s:1333
unable to read password
```


> * use the shell and `--password-file`: `restic backup --password-file <(pass backup/restic) /home`

Has the same issue as the above solution: `stdin` redirection. Also the shell will generate a temporary file containing  the password: Which might be readable by an attacker (`extundelete`or other filesystem tools).

> * use the environment: `RESTIC_PASSWORD=$(pass backup/restic) restic backup /home`

This seems like a good solution! The only disadvantage i see: The plaintext password stays in the  environment while `restic` is executing:

```bash
ps e $(pidof restic)
```

fd0:
The code is good, still missing is an entry for the changelog and an update to the documentation. Unfortunately, we don't have tests in place for user-interface stuff like this, so let's leave it for now.

juergenhoetzel:
> Ah, that's not the case: The shell starts the other process (e.g. `pass`) and connects a pipe to its stdout, which is then passed as a path to the file descriptor to restic:

Yep I was wrong here. I confused it with here-docs (`<<EOF`).

> Yep, but reading the environment is restricted to the current user (who could also just debug the process and dump its memory) or `root`, who could do anything.

Yep this is correct. You have to trust `root` anyway. But compared to environment variables you can prevent swapping out sensitive data (`mlock(2)`) in C: I don't know if `restic` is using this mechanism.

> Don't get me wrong, I'm not completely against adding this feature, I'm just worrying about feature creep and I'm trying to find out what benefits the additional complexity brings. :)

I fully agree. Less code, less bugs 😀

fd0:
So, what do we do? Is your use case covered by the options we discussed or do you still think it's valuable to merge this feature?

fd0 (approved):
Ok, fine with me

fd0:
I've added documentation and a changelog file.

Issue #2065, opened 2018-10-28
timezone in TimeFormat variable?


pabs3:
Output of `restic version`
--------------------------
```
$ restic version
restic 0.9.2 compiled with go1.10.3 on linux/amd64
```

Describe the issue
------------------
I'm importing backups from another backup system (rdiff-backup) that uses the local time (instead of UTC) for snapshots (like restic does).

I am setting doing the conversion like this and getting an error parsing the time:

```
$ rdiff-backup --list-increments /media/backup/rdiff/ | sponge | head -n2
Found 1234 increments:
    increments.2010-07-04T01:37:29+08:00.dir   Sun Jul  4 01:37:29 2010
$ time="2010-07-04 01:37:29+08:00"
$ rdiff-backup --restore-as-of "$time" /media/backup/rdiff /media/restore/rdiff/"$time"
$ cd /media/restore/rdiff/"$time"
$ restic backup --time "$time" --verbose --one-file-system --repo /media/backup/restic/ *
Fatal: error in time option: parsing time "2010-07-04 01:37:29+08:00": extra text: +08:00
```

I also tried passing the time in UTC and that works but it results in a snapshot timestamp of 2010-07-03T17:37:29Z which loses the timezone information that restic snapshots normally have.

If the TimeFormat variable is changed to include the timezone specification, then there will be a backwards compatibility issue and everyone using --time will have to change their code.

Another alternative would be to add a --time-format option similar to the `restic mount --snapshot-template` option.

fd0:
This is connected to #1306

MichaelEischer:
We could parse the timestamp normally first but fall back to try parsing with a timezone

pabs3:
Now my local patch to change the TimeFormat causes build failures because there is a test that checks the TimeFormat is correct.

PR #2095, opened 2018-11-20
Use local time for user-provided time of backup

plumbeo:

Snapshots created with the default `restic backup` behaviour of using the current time as time of backup use local time, but when the `--time` flag is used to force a specific time the resulting snapshot's time is set to UTC. This PR makes all snapshots use local time.

https://forum.restic.net/t/which-timezone-does-forget-use/1166/2
https://github.com/restic/restic/issues/2065

Checklist
---------


fd0 (approved):
Thank you very much, I like it!

Issue #2097, opened 2018-11-25
Allow key to be hinted


cdhowie:
Output of `restic version`
--------------------------
```
restic 0.9.3 compiled with go1.11.1 on linux/amd64
```

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
Along with the options to specify the password, it might be useful to have command-line or environmental ways to hint the key ID that should be tried first.

This would not change the behavior of the program in terms of the end result, but when there are many keys in a repository, restic can take a long time (15+ seconds) to check the password against each key.  For scripts in particular, hinting which key to try first could reduce the runtime of restic operations.

cdhowie:
#2098 merged

PR #2098, opened 2018-11-25
Add key hinting (#2097)

cdhowie:
Adds key hinting support (#2097).

No.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2097
+++ b/changelog/unreleased/issue-2097
@@ -1,0 +1,12 @@
+Enhancement: Add key hinting
+
+Added a new option `--key-hint` and corresponding environment variable
+`RESTIC_KEY_HINT`.  The key hint is a key ID to try decrypting first, before
+other keys in the repository.
+
+This change will benefit repositories with many keys; if the correct key hint
+is supplied then restic only needs to check one key.  If the key hint is
+incorrect (the key does not exist, or the password is incorrect) then restic
+will check all keys, as usual.
+
+https://github.com/restic/restic/issues/2097
--- a/internal/repository/key.go
+++ b/internal/repository/key.go
@@ -112,8 +112,25 @@ func OpenKey(ctx context.Context, s *Repository, name string, password string) (
 // given password. If none could be found, ErrNoKeyFound is returned. When
 // maxKeys is reached, ErrMaxKeysReached is returned. When setting maxKeys to
 // zero, all keys in the repo are checked.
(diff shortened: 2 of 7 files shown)


cdhowie:
It is unclear how to suitably test this code, as supplying an invalid key hint will not cause any observable error or change in behavior.  The only difference is that a correct key hint can shorten the time required to obtain the master key.

fd0:
Don't worry about the generated documentation, it's only generated for release binaries (everything else was too cumbersome for contributors).

changelog/unreleased/issue-2097
```diff
@@ -0,0 +1,12 @@
+Enhancement: Add key hinting
+
+Added a new option `--key-hint` and corresponding environment variable
+`RESTIC_KEY_HINT`.  The key hint is a key ID to try decrypting first, before
+other keys in the repository.
+
+This change will benefit repositories with many keys; if the correct key hint
+is supplied then Restic only needs to check one key.  If the key hint is
+incorrect (the key does not exist, or the password is incorrect) then Restic
```
fd0: nitpick: it's `restic`, not `Restic` :)

internal/repository/key.go
```diff
@@ -112,9 +112,26 @@ func OpenKey(ctx context.Context, s *Repository, name string, password string) (
 // given password. If none could be found, ErrNoKeyFound is returned. When
 // maxKeys is reached, ErrMaxKeysReached is returned. When setting maxKeys to
 // zero, all keys in the repo are checked.
-func SearchKey(ctx context.Context, s *Repository, password string, maxKeys int) (k *Key, err error) {
+func SearchKey(ctx context.Context, s *Repository, password string, maxKeys int, keyHint string) (k *Key, err error) {
 	checked := 0
 
+	if len(keyHint) > 0 {
```
fd0: If you extract this into its own function, the error handling becomes nicer, since you can return an error and keep the happy path non-indented. Let me know if you'd like to see a sample.

fd0 (changes requested):
In general I think this feature is helpful.

* The integration tests fail because the tests fail to compile, error is here: https://travis-ci.com/restic/restic/jobs/160640355#L229

 * Can you please remove the generated files from the commit?

internal/repository/key.go
cdhowie: I disagree here, as error handling is not necessary, except to bail out of this block.  On any error, we want to proceed with business as usual and check every key.  The hint is only a hint.

fd0: Fair enough.

fd0 (approved):

PR #2101, opened 2018-11-26
Out order restore

ifedorenko:

Significantly simplified `filerestore` implementation that fixes `not enough cache capacity` error and should also perform better when restoring large files like VM images.

Note that this PR depends on #2003 (`ProgressUI` API) so the delta looks much bigger than it really is.

* #2074
* https://forum.restic.net/t/restore-using-rclone-gdrive-backend-is-slow/1112/8

Checklist
---------


pmkane:
@ifedorenko, for folks building under 1.10, it looks like a copy of xxhash needs to be added to the vendor/ directory.

ifedorenko:
Please see #2195, which has the same restorer changes but without progress ui.

PR #2118, opened 2018-12-18
Clarification of keep-within

Triangle345:
clarification of forget help string

no

Checklist
---------


fd0:
Thanks for your contribution! Unfortunately this conflicts with #2090, where we discovered that the word in the help text should be "newer", not "older". In addition, I think repeating the exact same word "within" from the option in the help text is not really helpful. Thanks anyway! :)

PR #2124, opened 2018-12-24
Ability to dump folders to tar via stdout

### LINKED ISSUE #2123: Ability to dump folders to tar via stdout
Output of `restic version`
--------------------------
restic 0.9.3 compiled with go1.11.2 on darwin/amd64

Instead of only allowing folders to be dumped to stdout, it should also be possible to dump a whole folder to a tar file via stdout.

I've implemented a small POC for this. I'll create a merge request for this afterwards so we have a basis to discuss this. The POC currently only adds files to the tar, although tar would also support links. 

What are you trying to do?
--------------------------

Uploading restores directly to S3 without having to store any files locally. 

This is much more storage efficient on the nodes than restoring it first to a directory and then upload it to S3. We are also using restic in OpenShift/Kubernetes where we restrict the locally available.

The current implementation to solve this consists of a restic wrapper tool that lists all files in a repository and then calls `restic dump` for each one. This is very slow, complex and error prone.

Restic is just very awesome and I personally and the company I work for are using it quite extensively. So we want to donate this feature back to restic once it's cleanly implemented.

Kidswiss:

This enhances the dump command so that it will recognise if it was called on a file or a folder.

- If it's a file it will just dump its contents to stdout as before

- if it's a folder it will walk through the tree with the provided folder as its root. For each file it finds in the tree it adds it to the tar which in turn gets dumped to the stdout.

This implementation is a POC to test the viability and the speed of this solution, so it's not finished in terms of what it restores:

- currently no xattrs
- currently no links

I tested this with a repository consisting of about 500 files and about 1.3Gb data. The speed compared to a normal restore is practically the same (even a small bit faster). And I'd like this pull request to serve as a discussion ground to finish this cleanly.

This is the merge request for issue #2123.
Closes #2123

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -1,15 +1,19 @@
 package main
 
 import (
+	"archive/tar"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/walker"
 
 	"github.com/spf13/cobra"
 )
@@ -50,41 +54,18 @@ func init() {
 
 func splitPath(p string) []string {
 	d, f := path.Split(p)
-	if d == "" || d == "/" {
+	if d == "" {
 		return []string{f}
+	}
+	if d == "/" {
+		return []string{d}
 	}
 	s := splitPath(path.Clean(d))
 	return append(s, f)
 }
 
-func dumpNode(ctx context.Context, repo restic.Repository, node *restic.Node) error {
-	var buf []byte
-	for _, id := range node.Content {
-		size, found := repo.LookupBlobSize(id, restic.DataBlob)
-		if !found {
-			return errors.Errorf("id %v not found in repository", id)
-		}
+func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repository, prefix string, pathComponents []string, pathToPrint string) error {
 
-		buf = buf[:cap(buf)]
-		if len(buf) < restic.CiphertextLength(int(size)) {
-			buf = restic.NewBlobBuffer(int(size))
-		}
-
-		n, err := repo.LoadBlob(ctx, restic.DataBlob, id, buf)
-		if err != nil {
-			return err
-		}
-		buf = buf[:n]
-
-		_, err = os.Stdout.Write(buf)
-		if err != nil {
-			return errors.Wrap(err, "Write")
-		}
-	}
-	return nil
-}
-
-func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repository, prefix string, pathComponents []string) error {
 	if tree == nil {
 		return fmt.Errorf("called with a nil tree")
 	}
@@ -97,16 +78,19 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
 	}
 	item := filepath.Join(prefix, pathComponents[0])
 	for _, node := range tree.Nodes {
-		if node.Name == pathComponents[0] {
+		if node.Name == pathComponents[0] || pathComponents[0] == "/" {
 			switch {
 			case l == 1 && node.Type == "file":
-				return dumpNode(ctx, repo, node)
+				return getNodeData(ctx, os.Stdout, repo, node)
 			case l > 1 && node.Type == "dir":
 				subtree, err := repo.LoadTree(ctx, *node.Subtree)
 				if err != nil {
 					return errors.Wrapf(err, "cannot load subtree for %q", item)
 				}
-				return printFromTree(ctx, subtree, repo, item, pathComponents[1:])
+				return printFromTree(ctx, subtree, repo, item, pathComponents[1:], pathToPrint)
+			case node.Type == "dir":
+				node.Path = pathToPrint
+				return tarTree(ctx, repo, node, pathToPrint)
 			case l > 1:
 				return fmt.Errorf("%q should be a dir, but s a %q", item, node.Type)
 			case node.Type != "file":
@@ -129,7 +113,7 @@ func runDump(opts DumpOptions, gopts GlobalOptions, args []string) error {
 
 	debug.Log("dump file %q from %q", pathToPrint, snapshotIDString)
 
-	splittedPath := splitPath(pathToPrint)
+	splittedPath := splitPath(path.Clean(pathToPrint))
 
 	repo, err := OpenRepository(gopts)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


Kidswiss:
The build is currently failing due to the method I use to write the xattrs to the tar file. There are two ways to handle the xattrs: https://golang.org/pkg/archive/tar/#Header

```// Deprecated: Use PAXRecords instead.
Xattrs map[string]string // Go 1.3
```

and

```// The following are semantically equivalent:
//  h.Xattrs[key] = value
//  h.PAXRecords["SCHILY.xattr."+key] = value
PAXRecords map[string]string // Go 1.10
```

As the Xattrs map is deprecated I used PAXRecords which is only available in go 1.10+. So the build for go 1.9 fails.

Any suggestions how I should handle this? One possibility would be to use build tags and implement both versions (not a fan of this but if there's no other way). Also just for curiosity: why is there still a build that uses go 1.9? Is this for some backward compatibility?

cmd/restic/cmd_dump.go
```diff
@@ -58,30 +61,7 @@ func splitPath(p string) []string {
 }
 
 func dumpNode(ctx context.Context, repo restic.Repository, node *restic.Node) error {
```
fd0: Remove this function and call `getNodeData()` below, it's not worth being kept.
```suggestion
```

fd0: Remove `dumpNode()` and call `getNodeData()` at line 83 (the only place where it is called).

fd0 (changes requested):
Hey, thank you very much for your contribution, this function is great! Please accept my apologies for not reviewing this earlier, it's a rough time right now for me (at least concerning "spare time for restic in the evenings") ;)

As for your question: we try to support older versions of Go mainly so that people on Debian stable can build it. At the moment, the latest version for it available via the `backports` repo is 1.11, so I don't think there's a need to support Go 1.9 any more. You have my approval to remove Go 1.9 from the CI tests, but please mention it in the changelog file. Let me know if you need help with that!

A few issues I noticed while reviewing it:
 * Dumping a whole snapshot via `restic dump latest /` does not work, it prints: `cannot dump file: path "" not found in snapshot`
 * I noticed that the current code does not include any information about directories in the tar, is that intentional?
 * Something isn't quite right with extended attributes, like ACL, my test case:
```
mkdir test
echo foo > test/file
setfacl -m user:nobody:rwx test/file
chmod 700 test
```

Save this in a repository, then run `tar` on it gives:
```
./restic dump latest /test | tar --acls -tvp
-rw-rwxr--  1000/100          4 2019-02-23 11:47 test/file
```
It looks like there's no information about the ACL in it... Can you have a look? :)

Kidswiss:
Hey

Thank you very much for the feedback. Keep hanging in there, you get through that rough time :)

> * Dumping a whole snapshot via `restic dump latest /` does not work, it prints: `cannot dump file: path "" not found in snapshot`

I thought I fixed that one already... Maybe I forgot to push the commit. I'll have a look again.

> * I noticed that the current code does not include any information about directories in the tar, is that intentional?

You mean like mtime,ctime etc? To be honest I didn't really think about adding them. Good point, thx.

> * Something isn't quite right with extended attributes, like ACL, my test case:

I'll have a look at that. I developed and tested the feature under OSX where the whole tar/xattr story is somewhat weird and broken. I'll try to replicate your test and fix this under Linux.

Best Regards
Simon

Kidswiss:
I found the issue with the ACLs and tar. ACLs have to be encoded completely different for a tar, even though they are technically saved as an xattr. The ACLs are saved in some binary format in the xattrs.

Examples:
```
# Get the files ACLs
vagrant@golang:~$ getfacl backup/main.go
# file: backup.old/main.go
# owner: vagrant
# group: vagrant
user::rw-
user:root:rwx
user:nobody:rwx
group::rwx
mask::rwx
other::r--

# Get the files xattrs
vagrant@golang:~$ getfattr -d -m - backup/main.go 
# file: backup.old/main.go
system.posix_acl_access=0sAgAAAAEABgD/////AgAHAAAAAAACAAcA/v8AAAQABwD/////EAAHAP////8gAAQA/////w==
user.asd="bar"
user.foo="bar" 

# Tar header tar --create --xattrs  --acls test.tar backup/main.go
vagrant@golang:~$ cat test.tar
backup.old/PaxHeaders.6059/main.go0000644000000000000000000000047113436167053014051 xustar0030 mtime=1551429163.710267245
30 atime=1551430659.857967244
30 ctime=1551430766.967495243
# Correctly encoded ACLs in the tar header
80 SCHILY.acl.access=user::rw-
user:nobody:rwx
group::rwx
mask::rwx
other::r--
# Normal xattrs
29 SCHILY.xattr.user.foo=bar
29 SCHILY.xattr.user.asd=bar
# Even though the ACLs were written above the tar command adds a garbled mess here like my implementation...
85 SCHILY.xattr.system.posix_acl_access=?????????????? ????
backup.old/main.go0000674000175000017500000000055213436167053015556 0ustar00vagrantvagrant00000000000000package main
<SNIP>
```

While researching about this encoding I found this implementation in go: https://github.com/maxymania/go-system/blob/master/posix_acl/posix_acl.go. Unfortunately this is actually the only information I was able to find how that encoding actually works. It might be burried somewhere deep inside some POSIX specification.

I added a modified version of that code to Restic which does actually write the correct ACL headers:

```
vagrant@golang:/repos/restic/cmd/restic$ ./restic dump 1bb70a9b /home/vagrant/backup | tar --acls --xattrs -tvvp
-rw-rwxr--+ 1000/1000       362 2019-03-01 08:32 home/vagrant/backup/main.go
  a: user::rw,u:65534:rwx,group::rw,mask::rwx,other::r
  x: 3 user.asd
  x: 3 user.foo
```

The only difference here is that the user gets saved with its ID instead of the actual name provided by the system.

Kidswiss:
Sooo I finally found the time to finish this. I finished the implementation of the ACL parsing and also added docs and the changelog for the feature.

fd0:
Thank you very much! I'll try to find time to review and merge this the next couple of days!

fd0 (approved):

fd0 (changes requested):
I had a look, it looks great! I have two remarks:
 * I found it very odd that the tar output seems to drop the top-level directory name:
```
$ mkdir foo/bar/baz
$ echo test > foo/bar/baz/test

$ ./restic backup foo
[...]

$ ./restic ls latest
/foo
/foo/bar
/foo/bar/baz
/foo/bar/baz/test

$ ./restic dump latest /foo/bar/baz/test
test

$ ./restic dump latest / | tar tv
drwxr-xr-x 1000/100          0 2019-04-13 14:38 /bar
drwxr-xr-x 1000/100          0 2019-04-13 14:38 /bar/baz
-rw-r--r-- 1000/100          5 2019-04-13 14:38 /bar/baz/test

$ ./restic dump latest /foo | tar tv
drwxr-xr-x 1000/100          0 2019-04-13 14:38 /bar
drwxr-xr-x 1000/100          0 2019-04-13 14:38 /bar/baz
-rw-r--r-- 1000/100          5 2019-04-13 14:38 /bar/baz/test
```
 * I'd like to suggest adding a check that if `stdout` is a terminal, we return an error instead of throwing the (usually binary) `tar` output at the user. There's already `stdoutIsTerminal()` in `cmd/restic` that you could use. What do you think?

The rest looks great, I'll merge this after we agree on the two points above. Thank you very much for your work!

fd0:
FYI, I've pushed a commit which slightly rewords the changelog entry and tests that the Go version is at least 1.10 in `build.go`.

Kidswiss:
Thanks for the feedback!

* I had a look at the issue that the first node doesn't get tarred. That definitely a bug, that I missed during my tests. It was because `walker.Walk` actually starts with the next node, so the first one was completely ignored. I fixed that now. :)

* I like that idea and put the check at the top of `tarTree`. It now throws an error that the output should be redirected.

I rebased everything into one clean commit.

Kidswiss:
I *THINK* I now have all cases handled for this. `printFromTree` now accepts `pathToPrint` so the information is actually available where I need it.

fd0:
I've removed the `restic` binary from the patch and un-commented the error that is returned when stdout is not a terminal ;)

PR #2135, opened 2019-01-03
[WIP] Warn when stdin is empty

TheLastProject:
Stdin being empty most likely indicates the command used for stdin failing for some reason (think of a mysqldump that couldn't connect). This pull request isn't really done, and I'm not a go developer, but I wanted to run this idea by the team and making the (pseudo) code change seemed like an easier way to explain the behaviour I'm looking for.

Print an error message to stdout when stdin is empty and --stdin is used

No

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -146,6 +146,9 @@ func readLinesFromFile(filename string) ([]string, error) {
 
 	if filename == "-" {
 		data, err = ioutil.ReadAll(os.Stdin)
+		if len(data) == 0 {
+			Warnf("backing up from stdin but received 0 bytes from stdin\n")
+		}
 	} else {
 		data, err = textfile.Read(filename)
 	}


fd0 (commented):
Hey, thanks for giving this a try! Unfortunately, this is the wrong place to do it: here, the list of files to save in a backup is read from stdin (`--files-from -`). If that is empty, restic will complain later in the process that no files are to be saved, so we don't need to warn here.

Investigating why #2136 happens I found the issue, and I'll open a separate PR to fix it.

fd0:
Superseded by 2138

fd0:
Understood, thanks for the explanation. The case you've addressed in this PR is already handled (using `--files-from -` and not supplying any files). 

I agree that restic should fail when no bytes could be read from stdin, it's a separate issue.

fd0:
> The main issue here is that I ran into a backup failing due to incorrect mysqldump parameters, and not noticing because Restic doesn't consider it to be weird if it's asked to backup through stdin but doesn't receive any data at all.

I can highly recommend using the [unofficial bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) for such commands/scripts. It includes `set -o pipefail` that makes bash return a non-zero exit code when at least one of the programs in a pipe fail. The default seems to be to just take the exit code of the last (rightmost) program in the pipe. Here, restic is happy with reading zero bytes from stdin, so it returns zero as the exit code, which is then the exit code of the whole pipe.

fd0:
See #2139

dhoffend:
If you want to check if mysqldump has failed you can workaround by checking the exit status of the initial command. Usually if you run `mysqldump ... | restic backup --stdin ...` the exit value `$?` will be the return code of the last command (in this case restic). You can either change this using pipefail or better: by checking the PIPESTATUS internal bash variable.

```
#!/bin/bash
mysqldump ... | restic backup --stdin ...
[ "${PIPESTATUS[0]}" == "0" ] || { echo "mysqldump failed" && exit 1 }
[ "${PIPESTATUS[1]}" == "0" ] || { echo "restic failed" && exit 1 }
```

Hope that helps and you can integrate it into your restic backup script.

fd0:
@dhoffend that's basically what the `pipefail` option does, just a bit more verbose :)

PR #2138, opened 2019-01-06
Use processed bytes for summary

### LINKED ISSUE #2136: Using --stdin always reports 0B
Output of `restic version`
--------------------------

echo "test" | restic backup --stdin

Azure

Expected behavior
-----------------
Restic log shows "processed 1 file, 5 B in 0:00"

Actual behavior
---------------
Restic log shows "processed 1 file, 0 B in 0:00"

The "filesize" seems to never be stored if the file is from stdin

Internally store processed filesize for stdin

fd0:

When restic reads the backup from stdin, the number of bytes processed
was always displayed as zero. The reason is that the UI for the archive
uses the total bytes as returned by the scanner, which is zero for
stdin. So instead we keep track of the real number of bytes processed
and print that at the end.

Closes #2136
Supersedes #2135

Checklist
---------

### DIFF
--- a/internal/ui/backup.go
+++ b/internal/ui/backup.go
@@ -49,6 +49,7 @@ type Backup struct {
 			Changed   uint
 			Unchanged uint
 		}
+		ProcessedBytes uint64
 		archiver.ItemStats
 	}
 }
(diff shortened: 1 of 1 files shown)


mholt:
I wonder if we could/should merge #1944 first and then include the JSON reporting in this change.

fd0:
Hm, I think you're right.

PR #2178, opened 2019-02-21
WIP cmd_restore: expose filerestorer options in flags

labkode:
Expose internal tunnng options of the filerestorer on the restore command flags:

```
--average-pack-size int    average pack size in bytes (default 5242880)
--files-writer-count int   number of open files for restore (default 32)
--worker-count int        number of workers for file restore (default 8)
```

https://github.com/restic/restic/pull/2101
https://github.com/restic/restic/issues/2074

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2178
+++ b/changelog/unreleased/pull-2178
@@ -1,0 +1,15 @@
+Enhancement: expose filerestorer tunning parameters in cmd_restore command
+
+The `restic restore` command now exposes three new flags that control the 
+behaviour of the filerestorer.
+
+These new flags are the following:
+
+```
+--average-pack-size int    average pack size in bytes (default 5242880)
+--files-writer-count int   number of open files for restore (default 32)
+--worker-count int        number of workers for file restore (default 8)
+
+```
+
+https://github.com/restic/restic/pull/2178
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -1,12 +1,13 @@
 package main
 
 import (
+	"strings"
+
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/restic"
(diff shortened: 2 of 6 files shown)


fd0:
Thanks for your contribution. Unfortunately it looks like you did not discuss this change in an issue before submitting it, or did I miss anything?

I'm not a fan of exposing internal values (which are internal for a reason: they can change without notice...) to users as flags. If we need to expose something, it must be a value that is clearly understandable. The values here are specific to the current restorer implementation, but once we change that for something else, the flags make no sense any more. In my opinion, exposing them it's not the right thing to do.

In general, the project's philosophy is to have a minimal user interface which works for most of the cases. Since we cannot just remove options from restic (because that'd break user scripts), we need to be very careful with what we decide to make configurable.

Users who want to experiment with the code can change the values in the code (after reading and understanding it, of course).

I'm against merging this PR, sorry about that.

labkode:
Hi @fd0,


thanks for your comments, I understand your point of view. Would it make more sense to expose these flags as extended options ```-o, --option key=value ``` so they are not exposed directly to the user? That would allow to tune the performance of restic without having to depend on a forked version of the software and correct if I am wrong, but do extended options come with any backwards compatibility guarantee?

You didn't miss it, I missed it, I've created this issue: https://github.com/restic/restic/issues/2182
There are many related restore performance issues that may be related with the tuning of these options: 
https://github.com/restic/restic/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+restore+performance

labkode:
Ping @fd0 @ifedorenko

ifedorenko:
I am not in favour of this change, to be honest. `--average-pack-size` and `--files-writer-count` will go away once #2195 is merged. And I think it is better to get rid of `workerCount` altogether; restorer should be able to determine number of workers based on backend connection count and available cpu cores. We may eventually need to introduce configuration parameter to control local filesystem concurrency (i.e. how many concurrent reads or writes the filesystem can tolerate), but I don't think we've reached that point yet.

labkode:
@ifedorenko I close this PR then.

Issue #2183, opened 2019-02-22
Rare race on restore: ignoring error for filename: write filename: file already closed


d3zd3z:
Output of `restic version`
--------------------------
restic 0.9.4 compiled with go1.11.4 on linux/amd64

How did you run restic exactly?
-------------------------------
```sh
export RESTIC_REPOSITORY="s3:https://s3.wasabisys.com/bucketname/directory"
export AWS_ACCESS_KEY_ID="xxx"
export AWS_SECRET_ACCESS_KEY="xxx"
export RESTIC_PASSWORD="xxx"
sudo restic restore --target . 896b87d9
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Wasabi

Expected behavior
-----------------
Restore of files

Actual behavior
---------------
During restore, I got a message like:
```
ignoring error for path/to/filename: write path/to/filename: file already closed
```

After the restore, I looked at the file, and it had been created, but was 0 bytes long.

Steps to reproduce the behavior
-------------------------------
Unfortunately, it doesn't seem to happen consistently. I'm reporting this since it suggests a race where the file is closed before being restored.

Do you have any idea what may have caused this?
-----------------------------------------------
I suspect a race and one thread is closing the file before another thread has a chance to write data to it.

Do you have an idea how to solve the issue?
-------------------------------------------
Probably need to find the race.

Did restic help you or made you happy in any way?
-------------------------------------------------
All of the files have restored on subsequent restores. I will be testing restores repeatedly to see if this happens any more for me.

ifedorenko:
Thank you for the report. Will try to fix over the weekend.

The race is here https://github.com/restic/restic/blob/master/internal/restorer/fileswriter.go#L46. When file write cache overflows, it may purge writer used by concurrently running gorouting.

PR #2189, opened 2019-02-25
Fix rare 'file already closed' during restore

ifedorenko:

Reworked cache of open target files used by restorer to fix a possible race between two concurrent writeToFile requests.

See #2183

Checklist
---------


ifedorenko:
Sadly, I was not able to reproduce the race locally, at least not without adding artificial `time.Sleep` to `writeToFile`. My working theory this is due to the difference between macos and linux schedulers, but could also be due to my misunderstanding of how goroutings work. Regardless, I think the change simplifies coordination between concurrent `writeToFile` invocation and should make things more reliable.

fd0 (approved):
LGTM!

PR #2206, opened 2019-03-11
fs: Handle absolute pathname for --stdin-filename

### LINKED ISSUE #1809: Snapshot path for snapshots created with stdin
## Output of `restic version`
restic 0.9.0 (v0.9.0-1-g20bfed59) compiled with go1.10.2 on linux/amd64

```
$ restic -r /tmp/repo -p /tmp/repo_password init
[...]
$ echo foo | restic -r /tmp/repo -p /tmp/repo_password backup --stdin --stdin-filename foo
[...]
$ restic -r /tmp/repo -p /tmp/repo_password ls latest
repository fd5bd830 opened successfully, password is correct
snapshot 8d9d9ac2 of [/home/jonathan/foo] at 2018-05-24 06:00:55.208260662 +0000 UTC):
/foo
```

local

## Expected behavior
The snapshot path (`snapshot <id> of [<path>]`) for snapshots created with stdin should be sensible and not misleading.

## Actual behavior
The snapshot path for stdin snapshots is `<working dir>/<stdin filename>` which likely points to a filepath that never existed which can cause confusion and does not accurately represent the source of the data. 

_see above_

https://github.com/restic/restic/blob/v0.9.0/internal/restic/snapshot.go#L35

> Abs returns an absolute representation of path. **If the path is not absolute it will be joined with the current working directory to turn it into an absolute path**. The absolute path name for a given file is not guaranteed to be unique. Abs calls Clean on the result.
[golang.org](https://golang.org/pkg/path/filepath/#Abs)

- Use `/dev/stdin` as the source path. This should not affect how the file is stored within the snapshot (`/<stdin filename>`)

### LINKED ISSUE #2063: --stdin-filename with path separator gives cryptic error message
Output of `restic version`
--------------------------
```
```

`RESTIC_REPOSITORY` set to S3 bucket+prefix, and `RESTIC_PASSWORD` set appropriately.

```
# sudo -iu postgres pg_dumpall | restic backup -q --stdin --stdin-filename /root/postgresql-dump.sql
Fatal: unable to save snapshot: Lstat: file does not exist
```

S3

Expected behavior
-----------------
The backup should complete successfully, or at least a useful error message should be provided.  I would not expect Restic to need to stat _anything_ as part of this operation; the repository is in S3 and the file data is provided on standard input.

Actual behavior
---------------
Restic gives a "file does not exist" error message.

Attempt to upload a file with `--stdin-filename` containing a path separator.

It appears that Restic joins the provided filename to the current working directory Restic is run from; when I omit `/root/` then the file _is_ stored in the repository as `/root/postgresql-dump.sql` (cwd for the command is `/root`).

I would suspect this needs a patch to be fixed.  A workaround is to `cd` before running Restic so that it's running from the directory where you want the backed-up file to appear to be stored.

Yes, this has become my go-to backup tool for all of my machines.  My company is using it in production to backup multiple critical servers for quick recovery, while also providing inexpensive and easily-updated offsite backups.

garrmcnu:

Return valid directory info from Lstat() for parent directories of the specified filename. Previously only "/" and "." were valid directories.
Also set directory mode as this is checked by archiver.

Closes #2063
Closes #1809

Checklist
---------

### DIFF
--- a/internal/fs/fs_reader.go
+++ b/internal/fs/fs_reader.go
@@ -103,17 +103,32 @@ func (fs *Reader) Stat(name string) (os.FileInfo, error) {
 // describes the symbolic link.  Lstat makes no attempt to follow the link.
 // If there is an error, it will be of type *PathError.
 func (fs *Reader) Lstat(name string) (os.FileInfo, error) {
+	getDirInfo := func(name string) os.FileInfo {
+		fi := fakeFileInfo{
+			name:    fs.Base(name),
+			size:    0,
+			mode:    os.ModeDir | 0755,
+			modtime: time.Now(),
+		}
+		return fi
+	}
+
 	switch name {
 	case fs.Name:
 		return fs.fi(), nil
 	case "/", ".":
-		fi := fakeFileInfo{
-			name:    name,
-			size:    0,
-			mode:    0755,
-			modtime: time.Now(),
+		return getDirInfo(name), nil
+	}
+
+	dir := fs.Dir(fs.Name)
+	for {
+		if dir == "/" || dir == "." {
+			break
 		}
-		return fi, nil
+		if name == dir {
+			return getDirInfo(name), nil
+		}
+		dir = fs.Dir(dir)
 	}
 
 	return nil, os.ErrNotExist
--- a/internal/fs/fs_reader_test.go
+++ b/internal/fs/fs_reader_test.go
@@ -270,7 +270,7 @@ func TestFSReader(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				checkFileInfo(t, fi, "/", time.Time{}, 0755, false)
+				checkFileInfo(t, fi, "/", time.Time{}, os.ModeDir|0755, true)
 			},
 		},
 		{
@@ -281,7 +281,16 @@ func TestFSReader(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				checkFileInfo(t, fi, ".", time.Time{}, 0755, false)
+				checkFileInfo(t, fi, ".", time.Time{}, os.ModeDir|0755, true)
+			},
+		},
+		{
+			name: "dir/Lstat-error-not-exist",
+			f: func(t *testing.T, fs FS) {
+				_, err := fs.Lstat("other")
+				if err != os.ErrNotExist {
+					t.Fatal(err)
+				}
 			},
 		},
 		{
@@ -292,7 +301,7 @@ func TestFSReader(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				checkFileInfo(t, fi, "/", time.Time{}, 0755, false)
+				checkFileInfo(t, fi, "/", time.Time{}, os.ModeDir|0755, true)
 			},
 		},
 		{
@@ -303,7 +312,7 @@ func TestFSReader(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				checkFileInfo(t, fi, ".", time.Time{}, 0755, false)
+				checkFileInfo(t, fi, ".", time.Time{}, os.ModeDir|0755, true)
 			},
 		},
 	}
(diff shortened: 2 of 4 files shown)


fd0:
Just checking in (since you haven't checked the last box): is this PR ready for review? Are you done? Or do you need help with anything?

garrmcnu:
Thanks for the check-in, yes this is ready for review. If this approach looks good, I believe it does handle absolute pathnames. The only question I had was should relative directory paths be allowed for `--stdin-filename`?
With this fix, a relative directory will still fail because the snapshot path is always pre-pended with "/" and so will always be absolute. But I think supplying a relative path in this context doesn't make sense, so giving an error for a relative directory would seem correct. But good to get other thoughts on that?

rawtaz:
Why not just expand the relative path name so the corresponding absolute one, and use that? Then it's supported and is up to the user to decide how they want to specify the path, and it's in line with how you specify paths in other parts of restic.

fd0:
> Why not just expand the relative path name so the corresponding absolute one, and use that?

I don't think that's a good idea, then the path in the snapshot depends on the current directory, which may be surprising. I'd just take the path as an absolute path, and be done with it:
 *  `/foo/bar` becomes `/foo/bar`
 *  `foo/bar` becomes `/foo/bar`
 *  `./foo` becomes `/foo`
 *  `../../foo/bar` becomes `/foo/bar`

And so on. That's (IMHO) the least surprising thing we can do.

fd0:
Whoopsi, that was the wrong button :)

internal/fs/fs_reader_test.go
```diff
@@ -317,6 +327,49 @@ func TestFSReader(t *testing.T) {
 			test.f(t, fs)
 		})
 	}
+
+	var dirTests = []struct {
```
fd0: I'd split this out into its own test function, maybe `TestFSReaderDir`?

internal/fs/fs_reader_test.go
```diff
@@ -151,8 +152,8 @@ func verifyDirectoryContentsFI(t testing.TB, fs FS, dir string, want []os.FileIn
 }
 
 func checkFileInfo(t testing.TB, fi os.FileInfo, filename string, modtime time.Time, mode os.FileMode, isdir bool) {
-	if fi.IsDir() {
-		t.Errorf("IsDir returned true, want false")
+	if fi.IsDir() != isdir {
```
fd0: Oh, good catch!

internal/fs/fs_reader_test.go
```diff
@@ -151,8 +152,8 @@ func verifyDirectoryContentsFI(t testing.TB, fs FS, dir string, want []os.FileIn
...
+		t.Errorf("IsDir returned %t, want %t", fi.IsDir(), isdir)
```
fd0: We need another test, `os.FileInfo` defines that `Name()` returns the base of the file name, so `path.Base(filename)`:
```suggestion
		t.Errorf("IsDir returned %t, want %t", fi.IsDir(), isdir)
	}
	
	if path.Base(fi.Name()) != fi.Name() {
		t.Errorf("Name() returned is not base: want %q, got %q", path.Base(fi.Name()), fi.Name())
```

internal/fs/fs_reader.go
```diff
@@ -103,17 +103,32 @@ func (fs *Reader) Stat(name string) (os.FileInfo, error) {
 // describes the symbolic link.  Lstat makes no attempt to follow the link.
 // If there is an error, it will be of type *PathError.
 func (fs *Reader) Lstat(name string) (os.FileInfo, error) {
-	switch name {
-	case fs.Name:
-		return fs.fi(), nil
-	case "/", ".":
+	getDirInfo := func(name string) os.FileInfo {
 		fi := fakeFileInfo{
 			name:    name,
```
fd0: This needs to be `path.Base(name)`:
```suggestion
			name:    path.Base(name),
```

internal/fs/fs_reader_test.go
fd0: We also need to check `fi.Name() != path.Base(filename)` below (unfortunately I can't add a suggestion for it).

fd0 (commented):
I've added a few comments what we need to change, in general it's a great contribution! In order to handle the relative path issue, I'd suggest passing the path given as `--stdin-filename` through `path.Clean(path.Join("/", opts.StdinFilename))` in `cmd_backup.go:526`. This will do the right thing.

Thanks!

EDIT: I pressed the wrong button (again), this was meant to be a change request.

garrmcnu:
Thanks for the feedback, that sound good, will make those changes.

On the relative path, yes treating it as an absolute path looks like a clean solution and that works, thanks.
One minor point on the suggested change, I believe path.Join() already calls Clean() so don't think it needs to be called on the joined path.

internal/fs/fs_reader_test.go
garrmcnu: Ah yes, I completely missed that, thanks!

internal/fs/fs_reader_test.go
garrmcnu: yes, that's clearer.

garrmcnu:
Thanks for all the help on this.

And of course, thanks for restic, great backup solution!

Issue #2179, opened 2019-02-21
Restic uses `mtime` to detect file changes, which can miss changes.


d3zd3z:
Output of `restic version`
--------------------------
restic 0.9.4 compiled with go1.11.4 on linux/amd64

How did you run restic exactly?
-------------------------------
see below for the exact commands I used. Repo is local, and no other arguments are given to the backup command.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Local.

Expected behavior
-----------------
Follow the script, and expect restic to backup the given file after it has been modified.

Actual behavior
---------------
"Files:   0 new,  0 changed,   4 unmodified"

Steps to reproduce the behavior
-------------------------------
```sh
echo "Hello world" > a.txt
echo "hELLO WORLD" > b.txt
touch stamp
cat a.txt > hello.txt
touch -r stamp hello.txt
restic -r /tmp/test-repo -p a.txt init
restic -r /tmp/test-repo -p a.txt backup .
sleep 10
cat b.txt > hello.txt
touch -r stamp hello.txt
restic -r /tmp/test-repo -p a.txt backup .
```

Do you have any idea what may have caused this?
-----------------------------------------------
`mtime` should not be used to determine if a file needs to be backed up, `ctime` should be used. Worst case with `ctime` is that restic will needlessly read/hash the file to determine if it has changed. By using `mtime` it can skip backing up a file.

I have seen the Debian package manager, specifically, replace a file with a different file, and put the mtime back to the same value.

If I add `-f` to the backup command, the file will indeed be backed up.

Do you have an idea how to solve the issue?
-------------------------------------------
Ideally, use `ctime`. Maybe use both, or provide an option, but `ctime` is really what should be used. It would probably have to revert to `mtime` on a filesystem that doesn't have a `ctime`.

Did restic help you or made you happy in any way?
-------------------------------------------------
It makes me happy that it otherwise seems to reliably back up my files.

aldem:
Using mtime introduces another issue - changes in metadata only, like owner, group, modes, POSIX ACLs and any of extended attributes are not detected at all.

Option ```-f``` helps, but it takes ages to backup huge filesystem, as it simply is equivalent to initial backup - for instance, in my case (ca. 35G of data, over 600K files) it takes ca. 1.5h to make first backup, then less than 10m for every snapshot, but with ```-f``` it always runs more than 1h.

fd0:
[...]

PR #2207, opened 2019-03-15
add --skip-unchanged option

jmontleon:
Adds an option --skip-unchanged to restore. When set restic skips restoring files in the target location which appear to be unchanged based on size and mod time.

I have seen some forum discussion on the idea, no issues that I am aware of yet.

Checklist
---------

### DIFF
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -50,7 +50,7 @@ type treeVisitor struct {
 
 // traverseTree traverses a tree from the repo and calls treeVisitor.
 // target is the path in the file system, location within the snapshot.
-func (res *Restorer) traverseTree(ctx context.Context, target, location string, treeID restic.ID, visitor treeVisitor) error {
+func (res *Restorer) traverseTree(ctx context.Context, target, location string, skipUnchanged bool, treeID restic.ID, visitor treeVisitor) error {
 	debug.Log("%v %v %v", target, location, treeID)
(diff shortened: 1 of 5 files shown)


internal/restorer/restorer.go
```diff
@@ -90,6 +90,15 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string,
 			continue
 		}
 
+		if skipUnchanged {
+			if targetFile, err := os.Stat(nodeTarget); !os.IsNotExist(err) {
```
ifedorenko: This should be the same as https://github.com/restic/restic/blob/7b8d1dc04062bbef0c3f92652f23731773a0ac56/internal/archiver/archiver.go#L439 

Most likely need a helper somewhere, so both archiver and restorer can use it. 

Also see #2179.

ifedorenko:
This is similar/related to #2012.

internal/restorer/restorer.go
jmontleon: Thank you for the links to the other PR and issue. I'll look these over and see if I can implement this as a helper to work with the archiver as well.

fd0:
What's the status of this PR?

The "I'm ready" box hasn't been checked and there was no activity for over a year, so I'm closing it for now. Please leave a comment if you're still interested in working on it, so we can reopen the PR. Thanks a lot!

PR #2212, opened 2019-03-20
Examine file ctime when checking if files have changed.

### LINKED ISSUE #2179: Restic uses `mtime` to detect file changes, which can miss changes.

cbane:

This makes restic compare file ctimes (in addition to all of the existing checks) when it is deciding if a file has been modified.

Yes, closes #2179.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2179
+++ b/changelog/unreleased/issue-2179
@@ -1,8 +1,15 @@
-Bugfix: Use ctime when checking for file changes
+Enhancement: Use ctime when checking for file changes
 
 Previously, restic only checked a file's mtime (along with other non-timestamp
-data) to decide if a file has changed. This could cause it to not notice changes
-if something edits a file and then resets the timestamp. Restic now also checks
-the ctime, so any modification to a file should be noticed.
+metadata) to decide if a file has changed. This could cause restic to not notice
+that a file has changed (and therefore continue to store the old version, as
+opposed to the modified version) if something edits the file and then resets the
+timestamp. Restic now also checks the ctime of files, so any modifications to a
+file should be noticed, and the modified file will be backed up. The ctime check
+will be disabled if the --ignore-inode flag was given.
+
+If this change causes problems for you, please open an issue, and we can look in
+to adding a seperate flag to disable just the ctime check.
 
 https://github.com/restic/restic/issues/2179
+https://github.com/restic/restic/pull/2212
--- a/changelog/unreleased/issue-2179
+++ b/changelog/unreleased/issue-2179
@@ -1,0 +1,8 @@
+Bugfix: Use ctime when checking for file changes
+
+Previously, restic only checked a file's mtime (along with other non-timestamp
+data) to decide if a file has changed. This could cause it to not notice changes
+if something edits a file and then resets the timestamp. Restic now also checks
+the ctime, so any modification to a file should be noticed.
+
+https://github.com/restic/restic/issues/2179
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -580,21 +580,14 @@ func TestFileChanged(t *testing.T) {
 		{
 			Name: "new-content-same-timestamp",
 			Modify: func(t testing.TB, filename string) {
-				fi, _ := os.Stat(filename)
+				fi, err := os.Stat(filename)
+				if err != nil {
+					t.Fatal(err)
+				}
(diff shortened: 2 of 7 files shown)


cbane:
It looks like this change is breaking #2205 (or at least its unit tests). How do we want these to interact? The only idea I've had so far is disabling the ctime check if the ignore inode flag is on, but I'm open to anything that makes sense.

d3zd3z:
The only idea I've had so far is disabling the ctime check if the ignore inode flag is on, but I'm open to anything that makes sense.

The `ctime` value really only makes sense on the exact same inode, so this seems to me like a reasonable way of handling this.

fd0 (changes requested):
Thanks for taking a stab at this! This is a change which may leave users wondering what's happening, so we need to be extra careful to address this in the changelog. As far as I understood, there are some situations in which restic did not read the contents of the file, but does now (file with different content, but same length and `mtime` and inode). So for people depending on this (there are always people who depend on any behavior, no matter how obscure) will see that restic reads and hashes more files than before. Also, when metadata is changed, the `ctime` is updated and the file is re-read.

In my opinion, being correct and re-reading the contents if in doubt is the better strategy, so I'm in favor for this. Maybe we should add a paragraph to the changelog that people should open an issue if this breaks anything for them, so we can decide if we need a flag for it?

We need to clarify the following things before it can be merged:
 * Is it a bug or an enhancement (important for the changelog entry)? I'm currently leaning towards the latter, but I'm open to arguments.
 * The code needs to be rebased on master, the conflict with the ignore-inode option needs to be resolved
 * When `--ignore-inode` is turned on, I think the `ctime` should not be considered so I agree with @d3zd3z

cbane:
OK, I rebased against master, and it now skips the ctime check when the `--ignore-inode` option is given. I also updated the text for the changelog entry.

changelog/unreleased/issue-2179
```diff
@@ -0,0 +1,15 @@
...
+
+Previously, restic only checked a file's mtime (along with other non-timestamp
+metadata) to decide if a file has changed. This could cause restic to not notice
+that a file has changed (and therefore continue to store the old version, as
+opposed to the modified version) if something edits the file and then resets the
+timestamp. Restic now also checks the ctime of files, so any modifications to a
+file should be noticed, and the modified file will be backed up. The ctime check
+will be disabled if the --ignore-inode flag was given.
+
+If this change causes problems for you, please open an issue, and we can look in
```
fd0: I love the wording!

internal/archiver/archiver_test.go
```diff
@@ -576,6 +577,19 @@ func TestFileChanged(t *testing.T) {
 				save(t, filename, defaultContent)
 			},
 		},
+		{
+			Name: "new-content-same-timestamp",
```
fd0: Table-driven tests are awesome, easily extendable for contributors! :)

fd0 (approved):
LGTM, thanks!

DurvalMenezes:
Please excuse me, but should'nt we be giving more warning to users about this problem? Please see my rationale [here](https://github.com/restic/restic/issues/2179#issuecomment-495654982).

PR #2214, opened 2019-03-21
Add a global option --wait-lock, fixes #719

### LINKED ISSUE #719: Make restic wait when locked
## Output of `restic version`
restic 0.3.0 (v0.3.0-19-g6b88d3b)
compiled at 2016-11-03 13:39:00 with go1.7.3 on linux/amd64

## Expected behavior
When running restic prune and another backup job is already running, restic should wait for a configured amount of time for the lock to become available.

## Actual behavior
Instead restic aborts and complains about the active lock.

Run a restic backup task, then run a restic prune in parallel

julianbrost:
If the option is passed, restic will wait the specified duration of time and retry locking the repo every 10 seconds (or more often if the total timeout is relatively small).

It adds an option `--wait-lock` that can be used to make restic wait and retry to lock the repo if there is a conflicting lock.

#719

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -56,6 +56,7 @@ type GlobalOptions struct {
 	Quiet           bool
 	Verbose         int
 	NoLock          bool
+	WaitLock        time.Duration
 	JSON            bool
 	CacheDir        string
 	NoCache         bool
@@ -109,6 +110,7 @@ func init() {
 	f.BoolVarP(&globalOptions.Quiet, "quiet", "q", false, "do not output comprehensive progress report")
 	f.CountVarP(&globalOptions.Verbose, "verbose", "v", "be verbose (specify multiple times or a level using --verbose=`n`, max level/times is 3)")
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repository, this allows some operations on read-only repositories")
+	f.DurationVar(&globalOptions.WaitLock, "wait-lock", 0, "wait if the repository is already locked, takes a value like 5m or 2h (default: don't wait)")
 	f.BoolVarP(&globalOptions.JSON, "json", "", false, "set output mode to JSON for commands that support it")
 	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache `directory`. (default: use system default cache directory)")
 	f.BoolVar(&globalOptions.NoCache, "no-cache", false, "do not use a local cache")
--- a/cmd/restic/lock.go
+++ b/cmd/restic/lock.go
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive 
 		lockFn = restic.NewExclusiveLock
 	}
 
-	lock, err := lockFn(ctx, repo)
+	retrySleep := 10 * time.Second
+	if retrySleep > globalOptions.WaitLock/4 {
+		// Reduce the time between the retries if the maximum total wait
+		// duration is small so that there are at least a few attempts.
+		retrySleep = globalOptions.WaitLock / 4
+	}
+
+	var lock *restic.Lock
+	var err error
+	waitMessagePrinted := false
+	timeout := time.After(globalOptions.WaitLock)
+retryLoop:
+	for {
+		lock, err = lockFn(ctx, repo)
+		if err != nil && restic.IsAlreadyLocked(err) {
+			// repo already locked
+			select {
+			case <-timeout:
+				debug.Log("repo already locked, timeout expired")
+				break retryLoop
+			default:
+				debug.Log("repo already locked, retrying in %v", retrySleep)
+				if !waitMessagePrinted {
+					Verbosef("repo already locked, waiting up to %s for the lock\n", globalOptions.WaitLock)
+					waitMessagePrinted = true
+				}
+				time.Sleep(retrySleep)
+			}
+		} else {
+			// anything else, either a successful lock or another error
+			break retryLoop
+		}
+	}
 	if err != nil {
 		return nil, errors.WithMessage(err, "unable to create lock in backend")
 	}
--- a/cmd/restic/lock_test.go
+++ b/cmd/restic/lock_test.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"github.com/restic/restic/internal/repository"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/restic/restic/internal/test"
+)
+
+func TestLockWaitTimeout(t *testing.T) {
+	repo, cleanup := repository.TestRepository(t)
+	defer cleanup()
+
+	elock, err := lockRepoExclusive(context.TODO(), repo.(*repository.Repository))
+	test.OK(t, err)
+
+	wait := 2 * time.Second
+	oldWaitLock := globalOptions.WaitLock
+	globalOptions.WaitLock = wait
+	start := time.Now()
+	lock, err := lockRepo(context.TODO(), repo.(*repository.Repository))
+	duration := time.Since(start)
+	globalOptions.WaitLock = oldWaitLock
+	test.Assert(t, err != nil,
+		"create normal lock with exclusively locked repo didn't return an error")
+	test.Assert(t, strings.Contains(err.Error(), "repository is already locked exclusively"),
+		"create normal lock with exclusively locked repo didn't return the correct error")
+	test.Assert(t, duration >= wait,
+		"create normal lock with exclusively locked repo didn't wait for the specified timeout")
+
+	test.OK(t, lock.Unlock())
+	test.OK(t, elock.Unlock())
+}
+
+func TestLockWaitSuccess(t *testing.T) {
+	repo, cleanup := repository.TestRepository(t)
+	defer cleanup()
+
+	elock, err := lockRepoExclusive(context.TODO(), repo.(*repository.Repository))
+	test.OK(t, err)
+
+	wait := 2 * time.Second
+	time.AfterFunc(wait/2, func() {
+		test.OK(t, elock.Unlock())
+	})
+
+	oldWaitLock := globalOptions.WaitLock
+	globalOptions.WaitLock = wait
+	lock, err := lockRepo(context.TODO(), repo.(*repository.Repository))
+	globalOptions.WaitLock = oldWaitLock
+	test.OK(t, err)
+
+	test.OK(t, lock.Unlock())
+}
--- a/changelog/unreleased/pull-2214
+++ b/changelog/unreleased/pull-2214
@@ -1,0 +1,7 @@
+Enhancement: Add --wait-lock option
+
+This option allows to specify a duration for which restic will wait if there
+already exists a conflicting lock within the repository.
+
+https://github.com/restic/restic/pull/2214
+https://github.com/restic/restic/issues/719
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -542,7 +542,10 @@ that the process is dead and considers the lock to be stale.
 When a new lock is to be created and no other conflicting locks are
 detected, restic creates a new lock, waits, and checks if other locks
 appeared in the repository. Depending on the type of the other locks and
-the lock to be created, restic either continues or fails.
+the lock to be created, restic either continues or fails. If the
+``--wait-lock`` option is specified, restic will automatically retry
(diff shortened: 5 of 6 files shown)


julianbrost:
I'd like to print some message if restic actually starts to wait so that the user isn't surprised if nothing happens. However, there seems to be no way to import `Verbosef` from the package `main` (`cmd/restic/global.go`) but I'd still like it to honor `--quiet`/`--verbose`. Any suggestions?

julianbrost:
I pushed a new version that now prints a message and simplifies the diff quite a bit based on feedback from the IRC. However now the tests are missing, still have to take a look if there's a good way to test this with the new location of the logic.

cmd/restic/lock_test.go
```diff
@@ -0,0 +1,56 @@
...
+	wait := 2 * time.Second
+	oldWaitLock := globalOptions.WaitLock
+	globalOptions.WaitLock = wait
+	start := time.Now()
+	lock, err := lockRepo(repo.(*repository.Repository))
+	duration := time.Since(start)
+	globalOptions.WaitLock = oldWaitLock
+	test.Assert(t, err != nil,
+		"create normal lock with exclusively locked repo didn't return an error")
+	test.Assert(t, strings.Contains(err.Error(), "repository is already locked exclusively"),
```
julianbrost: I'm not happy with this substring check, but https://github.com/restic/restic/blob/870e7583a1128c9a6b20932178620ee2bafe0ef4/cmd/restic/lock.go#L38-L40 prevents me from using IsAlreadyLocked(err) here.

julianbrost:
Pushed another version, now containing tests, however I'm not fully happy but I don't see a better way without changing existing behavior in restic. See previous comment.

julianbrost:
I've now also tried to build a version using the backoff library already used by restic: https://github.com/julianbrost/restic/commit/c9ef43c186057a5b76333faf6e1be319aac37aa2#diff-b347952580200249c55463fe24c3e564

But I'm not really sure if I like it. It needs some hack to support not retrying at all (setting the MaxElapsedTime to a small non-zero value, in case of my code a negative one which doesn't really seem to be documented), but one could adapt this to make the default value of the `--wait-lock` option ±1ns so that `--wait-lock 0` could be used for waiting indefinitely. It also requires splitting the code over two callback functions and wrapping error into a PermanentError.

jtagcat:
Waiting indefinitely should definitely be a thing, so 0 would be indefinite, 5 would make it retry 5 times.

But also to save energy, you should also be able to set the retry frequency (every 10 seconds or every minute).

jtagcat:
A problem I see, is when cron is in play. Avoiding the same command execution on the same computer. I think it is best done localIy, before restic even checks for repository lock (so it could also be done externally, before restic execution).
_if instance is running, it is indicated by [n], when instance is waiting for lock, indicated by [ni]_
 - Instance 1, with command `x` makes lock, what contains `x` (besides other lock information).
 - [1] [2] Instance 2, command `y`, checks if there is a lock. Lock exists. The command inside the lock file differs. Check for waiting processes. There is none. Wait for next retry..
 - [1] [2i] [3] Instance 3, command `x` checks for lock. Lock exists. Lock has `x` in it, what is the same as instance 3's command. Instance 3 exits (local duplicate command).
 - [1] [2i] [4] Instance 4, command `y`, checks for lock. Lock exists. Lock has `x` in it. Command differs. Check for waiting processes. There is one with the same command. Instance 4 exits (local duplicate command).
 - [2i] Instance 1 exits successfully. Removing lock file.
 - [2] Command 2's retry found, that there is no longer a lock file. Aquieres lock with `y` in it. Command 4 executing.
 - Command 4 exits successfully. Removing lock file.

julianbrost:
I don't really understand what you're trying to describe there. Can you please describe what you think would happen and what you'd like to happen instead? Also I don't really get why command 2 should execute command 4.

My use case for this PR would be a scheduled prune. As this requires an exclusive lock, it would get skipped if an (unscheduled) restore is running at the same time and this would allow it to wait for these jobs to finish.

jtagcat:
Command 2 isn't executing 4..?
I described a way to make sure nothing ever fails bad and no more than needed instances build up (waiting).

jtagcat:
Related: [F#2829](https://forum.restic.net/t/very-old-lock-and-concerns-about-restic-unlock/2829/)

julianbrost:
I don't see how this is related. By the way, restic already has a concept of stale locks, see `restic unlock --help` and the `--remove-all` flag.

jtagcat:
> I don't see how this is related. By the way, restic already has a concept of stale locks, see `restic unlock --help` and the `--remove-all` flag.

What classifies a lock as stale (not found in docs)?

MichaelEischer:
> > I don't see how this is related. By the way, restic already has a concept of stale locks, see `restic unlock --help` and the `--remove-all` flag.
> 
> What classifies a lock as stale (not found in docs)?

A lock is regarded as stale, once it is older than half an hour.

julianbrost:
> > > I don't see how this is related. By the way, restic already has a concept of stale locks, see `restic unlock --help` and the `--remove-all` flag.
> > 
> > 
> > What classifies a lock as stale (not found in docs)?
> 
> A lock is regarded as stale, once it is older than half an hour.

Please note that there was a fix (#2391) for the lock renewal mechanism that didn't yet make it to any released version. So right now, locks are considered stale 30 minutes after they were initially created, even if the process is still running, unfortunately.

ducalex:
> Please note that there was a fix (#2391) for the lock renewal mechanism that didn't yet make it to any released version. So right now, locks are considered stale 30 minutes after they were initially created, even if the process is still running, unfortunately.

The fix made its way into 0.10.0, was that the only thing blocking this pull request?

julianbrost:
No, that fix is totally independent from this PR. As I said, I don't even understand why this was linked as related here. Mostly I'm waiting for feedback. So obviously it is currently blocked by merge conflicts, but I won't resolve them if there is no indication of this getting merged any time soon.

ducalex:
Oh, I see. Well for what it's worth I've been using your patch against 0.10.0 (with minor changes) to great success so far, thank you!

rawtaz:
Should probably be `--lock-wait` instead of `--wait-lock`.

julianbrost:
I pushed a rebased version.

Cnoru:
When will this feature be released? Very necessary.

rawtaz:
@julianbrost I spoke with Michael and we think that the option should be renamed `--retry-lock` instead. Also, the PR currently breaks JSON output (I'm told), so it will need more changes.

@Cnoru When it's complete, reviewed, merged and a new version of restic is released after that, what else.

cmd/restic/lock.go
```diff
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive
...
+	if retrySleep > globalOptions.WaitLock/4 {
+		// Reduce the time between the retries if the maximum total wait
+		// duration is small so that there are at least a few attempts.
+		retrySleep = globalOptions.WaitLock / 4
+	}
+
+	var lock *restic.Lock
+	var err error
+	waitMessagePrinted := false
+	timeout := time.After(globalOptions.WaitLock)
```
MichaelEischer, line 45: I'd prefer to pass the `globalOptions` to `lockRepository` via a parameter.

cmd/restic/lock.go
```diff
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive
 		lockFn = restic.NewExclusiveLock
 	}
 
-	lock, err := lockFn(ctx, repo)
+	retrySleep := 10 * time.Second
+	if retrySleep > globalOptions.WaitLock/4 {
+		// Reduce the time between the retries if the maximum total wait
+		// duration is small so that there are at least a few attempts.
+		retrySleep = globalOptions.WaitLock / 4
```
MichaelEischer, line 39: For short intervals one retry should be enough. My suggestion is to change this to: `globalOptions.WaitLock*2/3`.  Or just leave the retry interval as is and always add a final retry to the loop.

cmd/restic/lock.go
```diff
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive
...
+		if err != nil && restic.IsAlreadyLocked(err) {
+			// repo already locked
+			select {
+			case <-timeout:
+				debug.Log("repo already locked, timeout expired")
+				break retryLoop
+			default:
+				debug.Log("repo already locked, retrying in %v", retrySleep)
+				if !waitMessagePrinted {
+					Verbosef("repo already locked, waiting up to %s for the lock\n", globalOptions.WaitLock)
```
MichaelEischer, line 58: The wait message must not be printed when using JSON output.

cmd/restic/lock.go
```diff
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive
...
+					waitMessagePrinted = true
+				}
+				time.Sleep(retrySleep)
```
MichaelEischer, line 61: The sleep here should be cancelable by a user (via the passed context).

cmd/restic/lock_test.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"github.com/restic/restic/internal/repository"
```
MichaelEischer, line 5: That import should be moved after the golang ones

cmd/restic/lock_test.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func TestLockWaitTimeout(t *testing.T) {
+	repo, cleanup := repository.TestRepository(t)
+	defer cleanup()
+
+	elock, err := lockRepoExclusive(context.TODO(), repo.(*repository.Repository))
+	test.OK(t, err)
+
+	wait := 2 * time.Second
```
MichaelEischer, line 20: Would it be possible to reduce the wait duration further? This test and the next add nearly 4 seconds to the unit test execution.

cmd/restic/lock.go
```diff
@@ -32,7 +32,39 @@ func lockRepository(ctx context.Context, repo *repository.Repository, exclusive
...
+	var err error
+	waitMessagePrinted := false
+	timeout := time.After(globalOptions.WaitLock)
+retryLoop:
+	for {
+		lock, err = lockFn(ctx, repo)
+		if err != nil && restic.IsAlreadyLocked(err) {
+			// repo already locked
+			select {
+			case <-timeout:
```
MichaelEischer, line 52: I'd prefer a slightly different retry logic:
Try to lock the repository for the first time, then sleep for `retrySleep` or until `timeout` has expired whichever happens earlier, then try to lock the repository again. If timeout has expired this is the last try, otherwise loop and sleep again. That would avoid when waiting for e.g. 55 seconds that the last 5 seconds have no use at all.

cmd/restic/global.go
```diff
@@ -109,6 +110,7 @@ func init() {
 	f.BoolVarP(&globalOptions.Quiet, "quiet", "q", false, "do not output comprehensive progress report")
 	f.CountVarP(&globalOptions.Verbose, "verbose", "v", "be verbose (specify multiple times or a level using --verbose=`n`, max level/times is 3)")
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repository, this allows some operations on read-only repositories")
+	f.DurationVar(&globalOptions.WaitLock, "wait-lock", 0, "wait if the repository is already locked, takes a value like 5m or 2h (default: don't wait)")
```
MichaelEischer, line 113: Please rename the option to `--retry-lock`

MichaelEischer (changes requested):
@julianbrost Are you still interested in working on this PR? The branch needs a rebase and I'd like a few changes to the retry logic (see comments below).

cmd/restic/lock.go
julianbrost: That's... surprising. Shouldn't this be better handled by `Verbosef`? I had the expectation for this message to end up on stderr.

MichaelEischer: `Verbosef` prints to stdout. Only `Warnf` prints to stderr. #3009 could provide a generic solution to avoid `globalOptions.JSON` checks everywhere, but that PR is still work in progress.

fd0:
There has not been any activity in a long time, therefore I'm closing this PR for now. Let us know if you'd like to work on it again so we can reopen it. Thanks!

PR #2217, opened 2019-03-24
WIP: improve memory usage

fd0:

I'm working on restic's memory usage, this is my work-in-progress branch/PR. If you want to follow along, please do add comments :)

Each commit is meant to be viewed on its own, if you look at all the changes together that's probably way too much.

As a small real-life test, I've run `restic ls latest /` on a large (pathological) repo with **many** snapshots and files, and for now the reduction in memory usage is at about 15%, which is encouraging.

Checklist
---------


ifedorenko:
Out of curiosity, do you have memory usage breakdown that shows, say, 10 biggest users and how much memory each of them uses?

fd0:
Excellent question, it's very hard to say. What I found during profiling is that two things happen:

 1. Some functions allocate a lot of temporary buffers, which generates garbage that needs to be cleaned up
 2. The index (which keeps a giant list of which blobs are stored in what packs, including the location) uses **a lot** of memory
 
I've tried to address the first item in the commits in this PR.

Brain dump from yesterday evening:

The second step will be to rewrite the index to be much more space efficient, there's much potential for optimization. At the moment, we store at least ~350 byte for each blob in the repository. The worst case is a deeply nested directory structure containing only a single file with just a few bytes in it, then we have a data blob for each file (350 byte) and a tree blob for each subdir (another 350 byte), so if the file only contains 10 byte we need 700 byte of memory to handle it. That's obviously not a great idea ;)

I've done some tests to find out which of the two problems above is more important, using a test repo (which contains a snapshot for each Linux kernel version between 2.0 and 4.9, so a lot of files). I've modified the `list` command to abort just after loading the index files, then measured the RSS memory usage. Without modification, it needed around 2GiB of memory just to load the index files. After commenting out actually storing anything in the index (so we do the JSON unmarshal dance and allocate buffers and so on, but don't keep any of it), the memory usage goes down to below 500MiB. So I conclude the index is the problem.

Also, I see this odd effect that when `check` is run, restic just keeps using more and more memory. It could be a variant of problem 2., meaning non-space-efficient storage of intermediate results for `check`, for example keeping a set of data blob IDs that have been seen, which means at least 2*32 byte per data blob...

Btw, if you want to do some tests yourself, I can make the test repo available to you (you need ~50GiB of storage though).

ifedorenko:
Forgot to ask the most important question last night :-) Do you have specific numbers in mind you are trying to achieve, e.g. "backup so much data into repository this big while using no more than this many GiB of ram"? 

> Some functions allocate a lot of temporary buffers, which generates garbage that needs to be cleaned up

Does this affect amount memory restic requires or how much cpu cycles it uses or both? FWIW, I deliberately decided against buffer reuse in restorer implementation. Buffer reuse did not have measurable impact on restorer performance but did add complexity and "noise" to the code.

> The index (which keeps a giant list of which blobs are stored in what packs, including the location) uses a lot of memory

Have you considered moving the index off the main memory into disk-backed database? Something like [RocksDB](https://rocksdb.org) should work quite well based on some tests I ran earlier this year. Happy to share the observations from the tests if you are interested.

fd0:
> Do you have specific numbers in mind you are trying to achieve

No, I don't have specific numbers. Restic's memory usage for certain operations (especially `check`, but also `backup`) is extensive and must be reduced by a factor of 2-10 (depending on the repo structure).

The index data structure was the most naive thing I could think of at the time, so it's completely un-optimized ;)

> Does this affect amount memory restic requires or how much cpu cycles it uses or both?

Both, actually. When more memory is allocated quickly, the runtime mostly takes it from the operating system first. After a while, the allocating goroutines are  slowed down and must do some chores for the garbage collector, so there's the CPU overhead.

> Buffer reuse did not have measurable impact on restorer performance but did add complexity and "noise" to the code.

Yeah, there's a delicate balance to be found here, too. I'm fixing the most obvious rampant allocations, and I'll try to not add too much complexity. Complexity is always bad :)

> Have you considered moving the index off the main memory into disk-backed database? 

Yes, but I haven't experimented with it yet. I'll optimize the in-memory data structure first (we still need this, as some queries must be very fast), but maybe we can (in the end) have an on-disc data structure with only a tiny part of it in memory (LRU cache plus maybe a bloom-filter of all known blobs or so). There are several nice in-process key value stores (e.g. [boltdb](https://github.com/boltdb/bolt)), I just want something in Go (so I don't have to add a C/C++ dependency to restic).

fd0:
FYI, here's a great talk about Go's GC and what $GOGC means: https://www.youtube.com/watch?v=uyifh6F_7WM

fd0:
Ok, I think I figured out how to move forward. The first step is merging these changes as they already improve the situation a bit, I'll open a followup PR shortly where I describe what I have in mind :)

PR #2220, opened 2019-03-26
Add config option to set S3 storage class

### LINKED ISSUE #706: Support Amazon S3 Infrequent Access Storage
## Output of `restic version`
```
restic 0.3.2 (v0.3.2-4-gb0997d0)
compiled with go1.7 on linux/amd64
```

Hi,

It would be great if restic supports S3 Infrequent Access Storage.

https://aws.amazon.com/s3/faqs/#sia_anchor

There is a huge price difference between Standard Storage and Infrequent Access Storage. If I read correctly this setting valid for per object and not per bucket. 

> You can directly PUT into Standard – IA by specifying STANDARD_IA in the x-amz-storage-class header. 

Thank you,
Mihaly

BenoitKnecht:

The `s3.storage-class` option can be passed to restic (using `-o`) to
specify the storage class to be used for S3 objects created by restic.

The storage class is passed as-is to S3, so it needs to be understood by
the API. On AWS, it can be one of `STANDARD`, `STANDARD_IA`,
`ONEZONE_IA`, `INTELLIGENT_TIERING` and `REDUCED_REDUNDANCY`. If
unspecified, the default storage class is used (`STANDARD` on AWS).

You can mix storage classes in the same bucket, and the setting isn't
stored in the restic repository, so be sure to specify it with each
command that writes to S3.

Closes #706 

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2220
+++ b/changelog/unreleased/pull-2220
@@ -1,0 +1,16 @@
+Enhancement: Add config option to set S3 storage class
+
+The `s3.storage-class` option can be passed to restic (using `-o`) to
+specify the storage class to be used for S3 objects created by restic.
+
+The storage class is passed as-is to S3, so it needs to be understood by
+the API. On AWS, it can be one of `STANDARD`, `STANDARD_IA`,
+`ONEZONE_IA`, `INTELLIGENT_TIERING` and `REDUCED_REDUNDANCY`. If
+unspecified, the default storage class is used (`STANDARD` on AWS).
+
+You can mix storage classes in the same bucket, and the setting isn't
+stored in the restic repository, so be sure to specify it with each
+command that writes to S3.
+
+https://github.com/restic/restic/pull/2220
+https://github.com/restic/restic/issues/706
--- a/internal/backend/s3/config.go
+++ b/internal/backend/s3/config.go
@@ -18,6 +18,7 @@ type Config struct {
 	Bucket        string
 	Prefix        string
 	Layout        string `option:"layout" help:"use this backend layout (default: auto-detect)"`
+	StorageClass  string `option:"storage-class" help:"set S3 storage class (STANDARD, STANDARD_IA, ONEZONE_IA, INTELLIGENT_TIERING or REDUCED_REDUNDANCY)"`
 
 	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
 	MaxRetries  uint `option:"retries" help:"set the number of retries attempted"`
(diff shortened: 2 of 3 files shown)


BenoitKnecht:
I'm not exactly sure what's the best place to document this new option. https://github.com/restic/restic/blob/master/doc/030_preparing_a_new_repo.rst#amazon-s3? Suggestions welcome!

fd0 (approved):
Thank you very much, code looks good to me :)

tomwaldnz:
Nice work, well done!

Does this make all files in S3 IA class, including snapshots, indexes, etc, or just the data files? I'm wondering if running daily backups would read any index / snapshot / other files on S3 enough that the costs of IA class access would outweigh the savings of IA class storage. Then again Restic may keep a copy of those files locally and not need to read them on S3, in which case IA class would be fine.

fd0:
I think this applies to all files upload to the bucket.

BenoitKnecht:
@tomwaldnz Yes, it does apply to all files created by restic. You don't have to use the IA class specifically though.

Since I'm also unsure about the access patterns of restic over the course of daily backups and pruning, I'm using `INTELLIGENT_TIERING` (https://aws.amazon.com/about-aws/whats-new/2018/11/s3-intelligent-tiering/), which should do a pretty good job at using the most appropriate tier for each object over time (there's "intelligent" in the name after all, so it must be good :wink:).

tomwaldnz:
The problem with intelligent tiering is as soon as anything is accessed once it's moved to standard tier, even if it's only accessed once a month for some kind of validation. Intelligent tiering is probably the best option, I guess, but looking at file access patterns we could probably optimse it a bit more in future.

hallamjeff:
I'm going to propose including the storage class concept in the AWS S3 example of the documentation.

kalos:
It would be useful to have a environment variable for this option, like AWS_STORAGE_CLASS.
In order to avoid using the -o parameter.

Thanks!

rawtaz:
Quick question: What's the actual problem with using the -o option?

kalos:
I'm using a restic wrapper (https://github.com/fgma/rester/) that not support -o option, and i want to instruct restic to use scaleway specific class storage, instead the standard (https://stackoverflow.com/questions/61856438/scaleway-glacier-class-object-storage-with-restic).
I think is also useful for other scenario/wrapper/etc.

rawtaz:
Without dismissing your suggestion, I'd kindly suggest that you ask the author of the wrapper to support the full feature set of restic. I think that's a pretty natural thing to do by a wrapper.That said, I'm personally not too fond of introducing environment variables too easily. The -o exists for a reason, and we can't go adding env vars for most of third party libraries' options. It's better to keep restic neutral IMO.

kalos:
Okay, it seems like a legitimate choice.

Thanks so much for your work!

PR #2221, opened 2019-03-26
Fix reading password from stdin

### LINKED ISSUE #2203: Restic crashes when reading password via stdin
Output of `restic version`
--------------------------
`restic 0.9.4 (v0.9.4-37-g7b8d1dc0) compiled with go1.12 on linux/amd64`

Expected behavior
-----------------

Repository gets created with specified password.

Actual behavior
---------------

Restic crashes with the following:

```# echo "a" | restic -r foo init
read password from stdin
read password from stdin
EOF
ReadFull
main.readPassword
        cmd/restic/global.go:281
main.ReadPassword
        cmd/restic/global.go:317
main.ReadPasswordTwice
        cmd/restic/global.go:339
main.runInit
        cmd/restic/cmd_init.go:36
main.glob..func11
        cmd/restic/cmd_init.go:18
        vendor/github.com/spf13/cobra/command.go:762
        vendor/github.com/spf13/cobra/command.go:852
        vendor/github.com/spf13/cobra/command.go:800
main.main
        cmd/restic/main.go:86
runtime.main
        /usr/lib/go/src/runtime/proc.go:200
runtime.goexit
        /usr/lib/go/src/runtime/asm_amd64.s:1337
unable to read password
main.ReadPassword
        cmd/restic/global.go:322
main.ReadPasswordTwice
        cmd/restic/global.go:339
main.runInit
        cmd/restic/cmd_init.go:36
main.glob..func11
        cmd/restic/cmd_init.go:18
        vendor/github.com/spf13/cobra/command.go:762
        vendor/github.com/spf13/cobra/command.go:852
        vendor/github.com/spf13/cobra/command.go:800
main.main
        cmd/restic/main.go:86
runtime.main
        /usr/lib/go/src/runtime/proc.go:200
runtime.goexit
        /usr/lib/go/src/runtime/asm_amd64.s:1337
```

1. `echo "a" | restic -r foo init`

No

I'm not entirely fond of having to drop passwords in a file, which is why I prefer to use stdin when possible. I would use environment variables, but it's not possible to specify the new key via env var when doing `restic key add` for example, as far as I know.

Thanks for your time :)

pschultz:

Reading the password from non-terminal stdin used io.ReadFull with a
byte slice of length 1000.

We are now using a Scanner to read one line of input, independent of its
length.

Additionally, if stdin is not a terminal, the password is read only
once instead of twice (in an effort to detect typos).

Closes #2203

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2203
+++ b/changelog/unreleased/issue-2203
@@ -1,0 +1,6 @@
+Bugfix: Fix reading passwords from stdin
+
+Passwords for the `init`, `key add`, and `key passwd` commands can now be read from
+non-terminal stdin.
+
+https://github.com/restic/restic/issues/2203
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
@@ -273,15 +274,10 @@ func resolvePassword(opts GlobalOptions) (string, error) {
 
 // readPassword reads the password from the given reader directly.
 func readPassword(in io.Reader) (password string, err error) {
-	buf := make([]byte, 1000)
-	n, err := io.ReadFull(in, buf)
-	buf = buf[:n]
+	sc := bufio.NewScanner(in)
+	sc.Scan()
 
-	if err != nil && errors.Cause(err) != io.ErrUnexpectedEOF {
-		return "", errors.Wrap(err, "ReadFull")
-	}
-
-	return strings.TrimRight(string(buf), "\r\n"), nil
+	return sc.Text(), errors.Wrap(err, "Scan")
 }
 
 // readPasswordTerminal reads the password from the given reader which must be a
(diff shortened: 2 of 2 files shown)


pschultz:
I just realized that changing readPassword isn't necessary if we're only reading once from stdin. The second ReadFull call returns EOF, not ErrUnexpectedEOF. EOF isn't handled, and that causes the problem.

So there are a few options:
- Leave the PR as-is anyway
- Fix only the error handling in readPassword and continue reading the password twice, even from non-terminals (I find this very unintuitive)
- Revert the change to readPassword and keep the change to ReadPasswordTwice

fd0:
> Let me know if you want me to revert the change to readPassword.

Please leave it, I think it's much cleaner than what we had before :)

changelog/unreleased/issue-2203
```diff
@@ -0,0 +1,11 @@
+Bugfix: Fix reading passwords from stdin
+
+Passwords for `init` and `key add` can now be read from non-terminal stdin
+without having to pad the input to 1000 characters.
```
fd0: Please leave out the second half of the sentence, we're trying to write the changelog for users, and that's a (rather obscure) implementation detail :)

changelog/unreleased/issue-2203
```diff
@@ -0,0 +1,11 @@
...
+
+Additionally the password is now only read once from non-terminal stdin.
```
fd0: I'd also leave the last two paragraphs out of the changelog. Reading the password from stdin for most commands did not work properly (or it was very hard to get it working), so I think this isn't used by anybody. Also, the code is very old and it was only reported recently...

fd0 (changes requested):
The code looks great, just a few minor things in the changelog. Afterwards it can be merged by a maintainer. Thanks for solving this issue!

pschultz:
@fd0, I amended the commit with the updated changelog.

fd0 (approved):
LGTM, thanks!

PR #2231, opened 2019-03-30
find command: don't abort on tree load errors

### LINKED ISSUE #2224: The find command should not fail when objects are missing from the repository
This issue has arisen a few times on the forum:

* https://forum.restic.net/t/unable-to-check-or-prune/1239/6?u=cdhowie
* https://forum.restic.net/t/repository-damaged-after-accidentally-deleting-data-files/1516/10?u=cdhowie

Output of `restic version`
--------------------------
Various; the problem is still present on 0.9.4.

Any variant of `restic find` that looks for a specific object can exhibit this problem.

This is not relevant to the issue.

Expected behavior
-----------------
Restic should determine what objects reference the object named on the command line.

Actual behavior
---------------
Sometimes restic will crash with a message that an object is missing from the repository, which can include the object that was named.  For example:

```
tree cee759a90802904dcdbe733403dd452f8d709dc637d8fc97b966fc7c07838476 not found in repository
        /restic/internal/repository/repository.go:653
github.com/restic/restic/internal/walker.walk
        /restic/internal/walker/walker.go:107
github.com/restic/restic/internal/walker.walk
        /restic/internal/walker/walker.go:127
github.com/restic/restic/internal/walker.walk
        /restic/internal/walker/walker.go:127
github.com/restic/restic/internal/walker.walk
        /restic/internal/walker/walker.go:127
github.com/restic/restic/internal/walker.walk
        /restic/internal/walker/walker.go:127
github.com/restic/restic/internal/walker.Walk
        /restic/internal/walker/walker.go:56
main.(*Finder).findIDs
        /restic/cmd/restic/cmd_find.go:341
main.runFind
        /restic/cmd/restic/cmd_find.go:549
main.glob..func9
        /restic/cmd/restic/cmd_find.go:33
        /restic/vendor/github.com/spf13/cobra/command.go:762
        /restic/vendor/github.com/spf13/cobra/command.go:852
        /restic/vendor/github.com/spf13/cobra/command.go:800
main.main
        /restic/cmd/restic/main.go:86
runtime.main
        /usr/local/go/src/runtime/proc.go:201
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1333
```

There are not yet consistent steps to reproduce the issue since it can depend on how objects are packed.  Creating a repository, backing up enough data to generate a dozen or so packs, deleting a pack, and running `restic rebuild-index` might get the repository in this specific failure state, but is not guaranteed.

cdhowie:
As per #2224, this changes find not to abort when a tree can't be found.  Instead, the tree ID and referent snapshot ID are displayed.  This makes it easier to recover a broken repository by forgetting any snapshots that refer to missing objects.

Closes #2224

A few times on the forum:

* https://forum.restic.net/t/repository-damaged-after-accidentally-deleting-data-files/1516/17
* https://forum.restic.net/t/unable-to-check-or-prune/1239/12

Checklist
---------

Comments on above:

* There's a few cursory tests for find, but not ones that use many of the options.  This is also a particularly hard change to construct a test repository for, at least automatically.
* It was not documented before that find stops on a tree load error, so there aren't really any sensible changes to be made.

### DIFF
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -258,9 +258,13 @@ func (f *Finder) findInSnapshot(ctx context.Context, sn *restic.Snapshot) error 
 	}
 
 	f.out.newsn = sn
-	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(_ restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
+	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(parentTreeID restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
 		if err != nil {
-			return false, err
+			debug.Log("Error loading tree %v: %v", parentTreeID, err)
+
+			Printf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
+
+			return false, walker.SkipNode
 		}
 
 		if node == nil {
@@ -340,7 +344,11 @@ func (f *Finder) findIDs(ctx context.Context, sn *restic.Snapshot) error {
 	f.out.newsn = sn
 	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(parentTreeID restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
 		if err != nil {
-			return false, err
+			debug.Log("Error loading tree %v: %v", parentTreeID, err)
+
+			Printf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
+
+			return false, walker.SkipNode
 		}
 
 		if node == nil {
--- a/changelog/unreleased/issue-2224
+++ b/changelog/unreleased/issue-2224
@@ -1,0 +1,9 @@
+Bugfix: Don't abort the find command when a tree can't be loaded
+
+Change the find command so that missing trees don't result in a crash.
+Instead, the error is logged to the debug log, and the tree ID is displayed
+along with the snapshot it belongs to.  This makes it possible to recover
+repositories that are missing trees by forgetting the snapshots they are used
+in.
+
+https://github.com/restic/restic/issues/2224


cmd/restic/cmd_find.go
```diff
@@ -258,9 +258,13 @@ func (f *Finder) findInSnapshot(ctx context.Context, sn *restic.Snapshot) error
 	}
 
 	f.out.newsn = sn
-	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(_ restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
+	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(parentTreeID restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
 		if err != nil {
-			return false, err
+			debug.Log("Error loading tree %v: %v", parentTreeID, err)
+
+			Printf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
```
fd0:
```suggestion
			Warnf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
```

cmd/restic/cmd_find.go
```diff
@@ -340,7 +344,11 @@ func (f *Finder) findIDs(ctx context.Context, sn *restic.Snapshot) error {
 	f.out.newsn = sn
 	return walker.Walk(ctx, f.repo, *sn.Tree, f.ignoreTrees, func(parentTreeID restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
 		if err != nil {
-			return false, err
+			debug.Log("Error loading tree %v: %v", parentTreeID, err)
+
+			Printf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
```
fd0:
```suggestion
			Warnf("Unable to load tree %s\n ... which belongs to snapshot %s.\n", parentTreeID, sn.ID())
```

fd0 (changes requested):
Thank you very much for your contribution! The code looks great, just a minor thing: I'd like to suggest using `Warnf()` instead of `Printf()`, so the warning is printed to stderr instead of stdout. What do you think?

cdhowie:
I think this one could go either way.  Here is my take on it.

The point of having stdout and stderr is so that output can be separated from diagnostic information, particularly in the case where the output needs to be read by another program.

I'm not sure how find works with `--json` (I admittedly did not test it before submitting the PR).  If it supports that, then we have to decide if this is diagnostic information for humans (it should go to stderr) or it's relevant results from the operation (it should be added to the JSON structure).

Not running in JSON mode, the same question applies: are these errors considered important output from the requested operation, or a diagnostic that should be relegated to a side-channel?

IMO, if you're running find with `--blob` or `--tree` then your repository has problems and you're trying to figure out why.  In that case, I think that these warnings are directly related to the purpose of the command: someone is asking "what is wrong with my repository?" and therefore they should be part of the output.

But I will not argue if you'd rather have them on stderr because I think a case can be made for both options.

fd0:
You made a good point and I agree, let's leave it on stdout for now. Thanks!

fd0 (approved):

PR #2232, opened 2019-04-02
Mention Cache Directoy Tagging Standard in man page and help text

jacksgt:
This patch makes it more explicit what is meant by the CACHEDIR.TAG file
by referring to the "Cache Directory Tagging Standard".
It not only has to have this particular name, but also a specific content
(described at http://bford.info/cachedir/spec.html), which is not immediately
obvious to the user.

At least I did not realize right away that the CACHEDIR.TAG file is supposed to contain a Signature header and was confused by the error messages restic printed ("unexpected end of file").

No.

Checklist
---------

- [ ] (NOT RELEVANT) I have added tests for all changes in this PR
- [ ] (NOT RELEVANT) There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))
- [ ] (NOT RELEVANT) I have run `gofmt` on the code in all commits


jacksgt:
I hope I have added the link to all relevant files, if not please let me know.

fd0 (changes requested):
Great, thanks a lot! It should not have been necessary to make the change in the manpages, because those are generated automatically, but it does not hurt. Accepted! :)

fd0:
Ah, pressed the wrong button. Sorry about that.

fd0 (approved):

PR #2247, opened 2019-04-23
Minor docs fixes

tbm:
Just some minor documentation fixes.

Note that I updated `doc/man/restic-stats.1` manually rather than through `restic generate` since I don't have a Go build environment at the moment.  Please verify that `generate` produces the same output.


fd0 (approved):
Awesome, thanks for taking the time! Please don't worry about the generated files, they are generated for a reason :)

PR #2252, opened 2019-04-24
Read fresh metadata for unmodified files

### LINKED ISSUE #2249: Metadata-only changes are not recognized
Output of `restic version`
--------------------------

    $ echo foo > test

    $ restic backup test
    [...]
    processed 1 files, 4 B in 0:00
    snapshot 8c72321c saved

    $ restic ls -l latest
    repository 1e2b35f0 opened successfully, password is correct
    snapshot 8c72321c of [/home/fd0/shared/work/go/src/github.com/restic/restic/test] filtered by [] at 2019-04-23 20:39:59.152281409 +0200 CEST):
    -rw-r--r--  1000   100      4 2019-04-23 20:39:55 /test

    $ sudo chown root test

    $ restic backup test
    [...]
    processed 1 files, 4 B in 0:00
    snapshot 2a34805c saved

    $ restic ls -l latest
    repository 1e2b35f0 opened successfully, password is correct
    snapshot 2a34805c of [/home/fd0/shared/work/go/src/github.com/restic/restic/test] filtered by [] at 2019-04-23 20:40:12.099541936 +0200 CEST):
    -rw-r--r--  1000   100      4 2019-04-23 20:39:55 /test

    $ ls -al test
    -rw-r--r-- 1 root users 4 Apr 23 20:39 test

This affects all backends.

Expected behavior
-----------------

restic should pick up that the ownership of the file has changed, and record that in the repo

Actual behavior
---------------

restic does not recognize that the owner has changed.

See above

It's a bug: the new archiver code introduced with 0.9.0 calls `fileChanged()` [here](https://github.com/restic/restic/blob/c7762453cf3f763ae583e0ba8f6daac21427fbd4/internal/archiver/archiver.go#L388-L392) to decide whether or not to re-read the file. It mostly checks `mtime`, the inode for the file and the size and type.

If `fileChanged()` returns false, the old node is copied over, which not only (correctly) copies over the list of blobs the file consists of, but also all metadata information. The current ownership and other attributes which can be changed without changing the `mtime` are discarded.

I've verified that the old archiver code in 0.8.3 does not have this bug.

Thank you very much @aldem for being persistent and making us aware of this issue!

Do it properly: if the file hasn't changed, only copy over the list of blobs, but read the metadata anew.

fd0:

Restic took all metadata for files which were detected as unmodified, not
taking into account changed metadata (ownership, mode). This PR corrects it,
only the list of blobs (=the content) is taken from the previous file, the
metadata is read anew.

Closes #2249

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2249
+++ b/changelog/unreleased/issue-2249
@@ -1,0 +1,6 @@
+Bugfix: Read fresh metadata for unmodified files
+
+Restic took all metadata for files which were detected as unmodified, not taking into account changed metadata (ownership, mode). This is now corrected.
+
+https://github.com/restic/restic/issues/2249
+https://github.com/restic/restic/pull/2252
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -384,12 +384,19 @@ func (arch *Archiver) Save(ctx context.Context, snPath, target string, previous 
 			return FutureNode{}, true, nil
 		}
 
-		// use previous node if the file hasn't changed
+		// use previous list of blobs if the file hasn't changed
 		if previous != nil && !fileChanged(fi, previous, arch.IgnoreInode) {
-			debug.Log("%v hasn't changed, returning old node", target)
+			debug.Log("%v hasn't changed, using old list of blobs", target)
 			arch.CompleteItem(snPath, previous, previous, ItemStats{}, time.Since(start))
 			arch.CompleteBlob(snPath, previous.Size)
-			fn.node = previous
+			fn.node, err = arch.nodeFromFileInfo(target, fi)
+			if err != nil {
+				return FutureNode{}, false, err
+			}
+
+			// copy list of blobs
+			fn.node.Content = previous.Content
+
 			_ = file.Close()
 			return fn, false, nil
 		}
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -13,6 +13,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/restic/restic/internal/checker"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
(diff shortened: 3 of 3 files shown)


cbane:
The test for this fails if the system has SELinux enabled:
```
--- FAIL: TestMetadataChanged (0.01s)
    archiver_nowin_test.go:27: chdir to /tmp/restic-test-978949359                                                     
    archiver_nowin_test.go:53: metadata does not match:
        {*restic.Node}.ExtendedAttributes:
                -: []restic.ExtendedAttribute(nil)
                +: []restic.ExtendedAttribute{{Name: "security.selinux", Value: []uint8{0x75, 0x6e, 0x63, 0x6f, 0x6e, 0x66, 0x69, 0x6e, 0x65, 0x64, 0x5f, 0x75, 0x3a, 0x6f, 0x62, 0x6a, 0x65, 0x63, 0x74, 0x5f, 0x72, 0x3a, 0x75, 0x73, 0x65, 0x72, 0x5f, 0x74, 0x6d, 0x70, 0x5f, 0x74, 0x3a, 0x73, 0x30, 0x00}}}
    asm_amd64.s:522: chdir back to /home/courtney/src/restic/internal/archiver
FAIL
FAIL    github.com/restic/restic/internal/archiver      4.323s
```
Having a default ACL on /tmp will also trigger the problem, with a `system.posix_acl_access` extended attribute. (You can use e.g. `setfacl -d -m user:root:rwx /tmp` to set one; and then `setfacl -b /tmp` to remove all ACLs.)

My idea for fixing this is to copy the `security.selinux` and `system.posix_acl_access` extended attributes from `node2`/`node3` (if they exist) to `want` before checking if they match, but I haven't thought about this deeply.

fd0:
Oh, thanks for the hint. I'll fix the Windows tests first, then we'll see about these.

fd0:
FYI: I've fixed this (more or less accidentally) in #2266 by setting the field `ExtendedAttributes` to nil.

PR #2253, opened 2019-04-24
Make sure timestamps are valid

### LINKED ISSUE #2174: Invalid mtimes (year > 9999) fails snapshot
Hi,

we are evaluating  restic for performing a multi-user multi-petabyte daily backup of our storage repository. The preliminary results are very promising and we are very satisfied on the performance and friendly usage of the tool, specially the fuse based snapshot view (awesome!). From time to time we find files that has invalid mtimes because our storage system [EOS](http://eos.web.cern.ch/) allows arbitrary utimes system calls from end-users. Unfortunately, these files causes restic to not backup the user home directory.

This issue is related to https://github.com/restic/restic/issues/1173

Output of `restic version
--------------------------
```
```

```
restic --cache-dir /var/cache/restic/user0 --cleanup-cache -o s3.connections=32 -r s3:s3.xxxx/cboxback-user backup  --parent=39fb6be1 --exclude='*.sys.a#*' --exclude='*.sys.v#*' /home/user0
```

```
S3
```

Expected behavior
-----------------

Files with invalid mtimes should be backed up and the creation of the snapshot should not fail because of this

Actual behavior
---------------

With file mtimes where the year is not in the range 0-9999 the backup fails.

1) ```touch future.zip```
2) Create utime.c
```
#include <utime.h>
#include <time.h>
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

main() {
  int file_descriptor;
  char fn[]="future.zip";
  struct utimbuf ubuf;
  struct stat info;

  if ((file_descriptor = creat(fn, S_IWUSR)) < 0)
    perror("creat() error");
  else {
    close(file_descriptor);
    puts("before utime()");
    stat(fn,&info);
    printf("  utime.file modification time is %ld\n",
           info.st_mtime);
    ubuf.modtime = 9223372036854775807;       /* set modification time to max int64 */
    time(&ubuf.actime);
    if (utime(fn, &ubuf) != 0)
      perror("utime() error");
    else {
      puts("after utime()");
      stat(fn,&info);
      printf("  utime.file modification time is %ld\n",
             info.st_mtime);
    }
  }
}
```

3) ```gcc utime.c -o utime```
4) ```./utime```
5)
 ```  $ stat future.zip 
  File: `future.zip'
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 22h/34d	Inode: 798         Links: 1
Access: (0644/-rw-r--r--)  Uid: (95491/user0)   Gid: ( 2763/      it)
Access: 9223372036854775807
Modify: 9223372036854775807
Change: 2019-02-19 09:18:30.828784073 +0100
```
6) Restic outputs: ``` invalid modtime```

Yes,
https://github.com/restic/restic/blob/4c2072d875e14cf0c50ac449396e26297208e473/internal/restic/node.go#L336

Yes, but I would like to understand what are the consequences of allowing bigger year ranges in the validation or even disabling the validation. Would allowing invalid mtimes/atimes impact the backup logic?

Thanks a lot !

### PR DESCRIPTION
**Pull request by fd0**

Sometimes restic gets bogus timestamps which cannot be converted to JSON, because the stdlib JSON encoder returns an error if the year is not within [0, 9999]. We now make sure that we at least record _some_ timestamp and cap the year either to 0000 or 9999. Before, restic would refuse to save the file at all, so this improves the status quo.

closes #2174, #1173

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2174
+++ b/changelog/unreleased/issue-2174
@@ -1,0 +1,10 @@
+Bugfix: Save files with invalid timestamps
+
+When restic reads invalid timestamps (year is before 0000 or after 9999) it
+refused to read and archive the file. We've changed the behavior and will now
+save modified timestamps with the year set to either 0000 or 9999, the rest of
+the timestamp stays the same, so the file will be saved (albeit with a bogus
+timestamp).
+
+https://github.com/restic/restic/issues/2174
+https://github.com/restic/restic/issues/1173
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -332,24 +332,27 @@ func (node *Node) createFifoAt(path string) error {
 	return mkfifo(path, 0600)
 }
 
+// FixTime returns a time.Time which can safely be used to marshal as JSON. If
+// the timestamp is ealier that year zero, the year is set to zero. In the same
+// way, if the year is larger than 9999, the year is set to 9999. Other than
+// the year nothing is changed.
+func FixTime(t time.Time) time.Time {
+	switch {
+	case t.Year() < 0000:
+		return t.AddDate(-t.Year(), 0, 0)
+	case t.Year() > 9999:
+		return t.AddDate(-(t.Year() - 9999), 0, 0)
+	default:
+		return t
+	}
+}
+
 func (node Node) MarshalJSON() ([]byte, error) {
-	if node.ModTime.Year() < 0 || node.ModTime.Year() > 9999 {
-		err := errors.Errorf("node %v has invalid ModTime year %d: %v",
-			node.Path, node.ModTime.Year(), node.ModTime)
-		return nil, err
-	}
-
-	if node.ChangeTime.Year() < 0 || node.ChangeTime.Year() > 9999 {
-		err := errors.Errorf("node %v has invalid ChangeTime year %d: %v",
-			node.Path, node.ChangeTime.Year(), node.ChangeTime)
-		return nil, err
-	}
-
-	if node.AccessTime.Year() < 0 || node.AccessTime.Year() > 9999 {
-		err := errors.Errorf("node %v has invalid AccessTime year %d: %v",
-			node.Path, node.AccessTime.Year(), node.AccessTime)
-		return nil, err
-	}
+	// make sure invalid timestamps for mtime and atime are converted to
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by robvalca** [comment]
Thanks a lot Alex. 
I've quick-tested this branch and seems to do the job. Tried backup/restore. The only thing is that the affected file is always backed-up in the incrementals but this is expected as it differs from the file in the repo. But I think that this is better than not backing up anything : )

**Comment by fd0** [comment]
Yeah, that's the effect of the `mtime` being bogus. Can you maybe post a `stat` of one of the affected files? If/when we switch to using `ctime` for modification detection (see #2212), that will probably go away as the `ctime` is set by the OS.

**Comment by robvalca** [comment]
Hi Alex,
In our case the original ctime is already some kind weird but yes, seems that this would not happen if the ctime was checked instead of the mtime. 

This is the original file:
```
‘.swanproject’
Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 2ch/44d	Inode: 5           Links: 1
Access: (0644/-rw-r--r--)  Uid: (11111/aaaaa)   Gid: ( 1111/      aaaaa)
Context: system_u:object_r:fusefs_t:s0
Access: 4428626-11-10 23:45:52.007621616 +0100
Modify: 4428626-11-10 23:45:52.007621616 +0100
Change: 1970-03-30 06:29:12.-2013272080  +0100
 Birth: -
```
This is the restored file:
```
‘.swanproject’
Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: fd01h/64769d	Inode: 14680157    Links: 1
Access: (0644/-rw-r--r--)  Uid: (11111/aaaaa)   Gid: ( 1111/      aaaa)
Context: unconfined_u:object_r:user_tmp_t:s0
Access: 1816-02-08 05:16:08.073898992 +0034
Modify: 1816-02-08 05:16:08.073898992 +0034
Change: 2019-04-24 17:15:15.365003503 +0200
 Birth: -
```
I'll would be happy to help if you need something else. 
Cheers,

fd0:
Thanks for the stats, it looks like all timestamps are bogus. Huh. So I don't particularly care about that and will merge this PR now ;)

Issue #2226, opened 2019-03-29
Parse tar data backupped via stdin


Kidswiss:
Output of `restic version`
--------------------------
`restic 0.9.4 compiled with go1.11.4 on darwin/amd64`

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

If someone streams tar data to `restic` to do a backup:

`tar -cf - -C /veryimportantfolder | restic backup --stdin`

The whole thing will be saved as a single file. This will make restoring a single file very tedious as the whole tar has to be restored, it will get more painful the large the tar file gets.

If restic would parse the tar file and "convert" the entries into restic native file trees, it would be possible to create a virtual folder snapshot. This way a tar file is backed up, but single file restore is still available.

What are you trying to do?
--------------------------

We use restic quite heavily in Kubernetes and OpenShift workloads where it's not always possible to give direct filesystem access to restic. So we stream quite a lot of tar files between containers to get the backups. This creates the problem described above.

This feature would complement #2123.

What do you think? Would something like this make sense?

Did restic help you or made you happy in any way?
-------------------------------------------------

Restic rocks :)

cdhowie:
Duplicate of #437.

fd0:
Ah, I actually like the idea. We even have an abstraction layer now (`fs.FS`) which could be used to implement a `tar` file system maybe.

eikevons:
This would also help in a situation, where firewall rules forbid connection from the system-to-be-backed-up to the backup-storage but not the reverse direction. We can have a simple script on the system-to-be-backed-up that is invoked via ssh and `tar`s to stdout and by-pass the necessity to make the whole system available through sshfs.

alallier:
@fd0 does that mean Restic plans to support streaming tar data to stdin?

FiloSottile:
Something I do often on machines where I don't want to install software or credentials is `ssh machine.home.arpa tar cv ~`. It would be awesome to be able to pipe that into restic and have it understand as a filesystem.

jinnko:
This would also be great for backing up volumes from within docker which also uses tar under the hood, for example:

    docker cp running-container:/path/to/volume - | restic backup --stdin

Legion2:
I think it is also important that the tar stdin is not completely stored in ram, because if the backup is huge it would not fit. This would allow to stream data from one remote source to a backup without storing the "source" on the local file system.

cdhowie:
[...]

PR #2255, opened 2019-04-25
Fix dumping issues with / and the first sub level

### LINKED ISSUE #2254: restic dump handles root level files incorrectly
Output of `restic version`
--------------------------

Prepare a test data directory, and back it up so that
```
>>> restic ls af55bc75
repository c96b5ed2 opened successfully, password is correct
snapshot af55bc75 of [C:\<...>\restic-test\data] filtered by [] at 2019-04-24 13:31:03.9490076 -0400 EDT):
/abc
/abc/0
/abc/1
/abc/2
/abc/3
/random0
/random1
/random2
```
Then try dumping one file in the root folder
```
>>> restic dump af55bc75 /random0 > dump
```

Local filesystem

Expected behavior
-----------------
`dump = /random0`

Actual behavior
---------------
`dump` is a tarball of `/abc` including all files in it, except the directory name "abc" is replaced with "random0".

More generally, anytime when the `file` argument to `dump` is of the form `slash + one name`, it always dumps the first node in the snapshot, whatever it is.

Described above.

Probably this:
https://github.com/restic/restic/blob/fad9f65c65b1dc426bd00c962beb68834289c37c/cmd/restic/cmd_dump.go#L60-L62

return `{d, f}` when `d == "/"` in `splitPath`.

Everyday.

### LINKED ISSUE #2357: "restic dump" command fails on 0.9.5; succeeds with 0.9.4
Output of `restic version`
--------------------------
```
```

```
restic dump $snapshot /usr/local/apache2/$filepath
```

Local disk storage.

Expected behavior
-----------------
Restic should dump the file to standard output.

Actual behavior
---------------
Restic fails:

```
repository b01bdece opened successfully, password is correct
cannot dump file: path "/local" not found in snapshot
```

Note that the path in the error does not match the path I've asked to be dumped. The path begins with `/usr/local`, not `/local`.  On a hunch I tried `/usr/usr/local/apache2/$filepath` and this also failed, but with:

```
repository b01bdece opened successfully, password is correct
cannot dump file: path "/usr" not found in snapshot
```

I'm not sure what condition triggers the failure, but it's consistently reproducible on this 200+GB repo.

The command works with restic 0.9.4, so I assume that some change in 0.9.5 causes this error.

Kidswiss:
There was an issue that prevented the dump command from working
correctly when either:

* `/` contained multiple nodes (e.g. `restic backup /`)
* dumping a file in the first sublevel was attempted (e.g. `/foo`)

Fixes #2254

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -1,19 +1,16 @@
 package main
 
 import (
-	"archive/tar"
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 
 	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/dump"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
-	"github.com/restic/restic/internal/walker"
 
 	"github.com/spf13/cobra"
 )
@@ -82,7 +79,10 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
 	// If we print / we need to assume that there are multiple nodes at that
 	// level in the tree.
 	if pathToPrint == "/" {
-		return writeTar(ctx, repo, tree, pathToPrint, os.Stdout)
+		if err := checkStdoutTar(); err != nil {
+			return err
+		}
+		return dump.WriteTar(ctx, repo, tree, pathToPrint, os.Stdout)
 	}
 
 	item := filepath.Join(prefix, pathComponents[0])
@@ -91,23 +91,26 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
 		// first item it finds and dump that according to the switch case below.
 		if node.Name == pathComponents[0] {
 			switch {
-			case l == 1 && node.Type == "file":
-				return getNodeData(ctx, os.Stdout, repo, node)
-			case l > 1 && node.Type == "dir":
+			case l == 1 && dump.IsFile(node):
+				return dump.GetNodeData(ctx, os.Stdout, repo, node)
+			case l > 1 && dump.IsDir(node):
 				subtree, err := repo.LoadTree(ctx, *node.Subtree)
 				if err != nil {
 					return errors.Wrapf(err, "cannot load subtree for %q", item)
 				}
 				return printFromTree(ctx, subtree, repo, item, pathComponents[1:], pathToPrint)
-			case node.Type == "dir":
+			case dump.IsDir(node):
+				if err := checkStdoutTar(); err != nil {
+					return err
+				}
 				subtree, err := repo.LoadTree(ctx, *node.Subtree)
 				if err != nil {
 					return err
 				}
-				return writeTar(ctx, repo, subtree, pathToPrint, os.Stdout)
+				return dump.WriteTar(ctx, repo, subtree, pathToPrint, os.Stdout)
 			case l > 1:
 				return fmt.Errorf("%q should be a dir, but is a %q", item, node.Type)
-			case node.Type != "file":
+			case !dump.IsFile(node):
 				return fmt.Errorf("%q should be a file, but is a %q", item, node.Type)
 			}
 		}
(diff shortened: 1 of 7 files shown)


fd0:
Hm. What about moving the tar code into a package below `internal/` and adding the cases you tested manually as real tests? I can give it a shot if you like me to help :)

Kidswiss:
That's a great idea, especially the test cases. Do you have anything to add to the ones I already declared at #2254?

I'd appreciate a few hints on what you have in mind, so we're on the same level. Then I'll give it a shot :)

fd0:
So, what I'd do (roughly):
 * Remove all `tar` code from `printFromTree()`
 * Move the `tar` code to e.g. `internal/archive`
 * Export a function in `internal/archive`, something along the lines `func WriteTar(ctx context.Context, repo restic.Repository, tree restic.ID, dst io.Writer) error`
 * Write a function `isDir()` (or so) to decide if the target is a dir or a file
 * then either call `printFromTree()` or `archive.WriteTar()`
 * For testing, add tests in the `internal/archive` package on a test repo (look at the archiver tests if you need inspiration):
   * for each test, create a repo with a known structure, run `WriteTar()` on it writing the output to a `bytes.Buffer`
   * then read the output back with a helper function that checks the `tar` file against the expected structure

It's a big change now thinking about it, so please just let me know if you need help with anything!

It's important to split the change into as many small commits as necessary, this makes reviewing so much easier!

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
...
 			case l > 1:
 				return fmt.Errorf("%q should be a dir, but s a %q", item, node.Type)
 			case node.Type != "file":
 				return fmt.Errorf("%q should be a file, but is a %q", item, node.Type)
 			}
 		}
+
+		// If we print / we need to asume that there are multiple nodes at that
+		// level in the tree.
+		if pathToPrint == "/" {
```
fd0: I'd handle this special case first in the function (after the sanity checks), so we get it out of the way.

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
...
+			tw := tar.NewWriter(os.Stdout)
+			defer tw.Close()
```
fd0: `Close()` returns an error which isn't checked here, but needs to be. I'd remove this line and handle the case explicitly as shown below. I'd also not use `defer()`, because error checking in `defer` is very error prone.
```suggestion
```

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
...
+			for _, rootNode := range tree.Nodes {
+				rootNode.Path = pathToPrint
+				err := tarTree(ctx, repo, rootNode, pathToPrint, tw)
+				if err != nil {
+					return err
```
fd0:
```suggestion
					_ = tw.Close()
					return err
```

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
...
+				}
+			}
+			return nil
```
fd0:
```suggestion
			return tw.Close()
```

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
 				return printFromTree(ctx, subtree, repo, item, pathComponents[1:], pathToPrint)
 			case node.Type == "dir":
 				node.Path = pathToPrint
-				return tarTree(ctx, repo, node, pathToPrint)
+				tw := tar.NewWriter(os.Stdout)
+				defer tw.Close()
```
fd0: You're dropping an error returned by `tw.Close()`, which needs to be checked. I'd remove this line and handle it as shown below.
```suggestion
```

cmd/restic/cmd_dump.go
```diff
@@ -90,13 +90,31 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
...
+				return tarTree(ctx, repo, node, pathToPrint, tw)
```
fd0:
```suggestion
				err = tarTree(ctx, repo, node, pathToPrint, tw)
				if err != nil {
					_ = tw.Close()
					return err
				}
				return tw.Close()
```

cmd/restic/cmd_dump.go
```diff
@@ -194,28 +212,28 @@ func getNodeData(ctx context.Context, output io.Writer, repo restic.Repository,
...
 	}
 
-	// we know that rootNode is a folder and walker.Walk will already process
-	// the next node, so we have to tar this one first, too
 	if err := tarNode(ctx, tw, rootNode, repo); err != nil {
 		return err
 	}
 
+	// If this is a file we are finished
+	if rootNode.Subtree == nil {
```
fd0: That's not the correct way to test for a tree, use this:
```suggestion
	if rootNode.Type != "tree" {
```

cmd/restic/cmd_dump.go
```diff
@@ -54,12 +54,9 @@ func init() {
...
 func splitPath(p string) []string {
 	d, f := path.Split(p)
-	if d == "" {
+	if d == "" || d == "/" {
 		return []string{f}
 	}
-	if d == "/" {
-		return []string{d}
-	}
 	s := splitPath(path.Clean(d))
```
fd0: There's some weird stuff going on when you pass odd paths, like `restic dump latest /foo/../`, which is caused by `path.Clean()` returning either an empty string or a relative path that makes no sense. You could do the following to always have it be absolute:
```suggestion
	s := splitPath(path.Clean(path.Join("/", d)))
```

fd0 (dismissed):
Thanks a lot for tackling these issues!

I had a look and found a couple of odd things, in addition to the line comments:
 * The path in the `tar` varies whether or not the path given to `dump` is absolute:
   * running `restic dump latest /cmd` gives me paths like `/cmd/restic/acl.go` in the tar
   * running `restic dump latest cmd` gives me paths like `cmd/restic/acl.go` in the tar
   * I think the paths in the resulting `tar` file should be relative by default, what are your thoughts on that?
 * The help text for the `dump` command isn't up to date, it does not mention writing `tar` files at all

fd0:
As always, please let me know if you need help, we can also resolve this collaboratively if you want.

Kidswiss:
Hey, thanks for the feedback!

I may have some time tomorrow that I can implement your suggestions and fix the odd stuff. So I can do the refactoring on a clean base.

This feature has quite some hidden caveats that make it more complex than anticipated, but we're getting there! :)

I like the idea of possible support for other formats, too. Also just wondering what your opinion of #2226 is. As decisions we make here could have potential impact on the implementation of the feature.

cmd/restic/cmd_dump.go
Kidswiss: Hmmm... the goal is rather to test for a file.

So I guess:
```
if rootNode.Type == "file"
```

Would be the best solution here? As your suggestion would also return if the node is a folder or a link.

fd0:
I've taken the liberty of rebasing on master, so that the CI tests on Windows pass.

Kidswiss:
Sorry for the radio silence, I was on a holiday :)

I should be able to do some work here this week. But as there's much to do at work I'll have to do it during my free time. So don't expect huge leaps in progress :)

Kidswiss:
I moved everything related to `tar` to the archiver. The whole tar handling is now much more consistent and should work better.

Although I'll need to test it on a windows machine, too. Just to be sure if the original issue is now fixed.

Todo:
* implement tests

Kidswiss:
Hi @fd0 

I had a look at the test code in Archiver and I'm not quite sure what exactly I'll need, to create a test repo. I feel like I need something like `TestArchiverSaveDir` to create the repo. But as I'm not sure I'd rather ask you before I start implementing that and realise that it's not the right way todo that :)

Kidswiss:
I added some automated test for various cases where `/` is dumped. I compare the created tar file against the temporary folder created by the restic tests and if I find some differences the test fails.

Kidswiss:
@fd0 any chance you could have a look at this so I can finish this up cleanly? :) Still not quite sure my test code is really doing the right thing...

Kidswiss:
I just realised that the push on 9th August was something completely broken and basically reverted everything... I fixed that now.

Doing this I found that the CI build for MacOS fails with the new tar tests I implemented, due to timestamps not matching. This is interesting as I'm developing locally with MacOS and can't reproduce this.

Kidswiss:
@fd0 @rawtaz 

Finally got around to clean this up properly. Any chance one of you guys have time for a review? I'd really like to get this into the next release. Thanks! :)

greatroar:
Why is tar dumping implemented in internal/archiver? That package is for backing up, not for accessing backups. Wouldn't internal/dump make more sense?

Kidswiss:
@greatroar Thanks for the feedback!

It's quite a while ago since this change was started. But it was @fd0's suggestion here:  https://github.com/restic/restic/pull/2255#issuecomment-487309292. Not quite sure about his reasoning though. Maybe because `tar` is an archive format? It seemed reasonable at the time.

The actual dump logic is in the `cmd_dump.go` file. The logic in the `internal/archive` packe is solely used to create the tar archives given some arbitrary readers.

MichaelEischer (commented):
Sorry for us taking so long to get back to you. 

I've rebased this PR to the current master and restructured the commits and their commit messages a bit (but I've checked that nothing relevant did change).

Besides that the tar code now lives in `internal/dump`. @fd0's suggestion was `internal/archive` (not archiver), but that's probably quite easy to misread. This required some changes to the test as it is no longer able to access the internal of the archiver.

I've also removed the redundant call to `path.Clean` in `splitPath` (`path.Join` already cleans the path). 

I'll try whether I can get the tar test to fully work on windows and the modtime check without introducing too much complexity.

I nearly forgot to mention: I'll have to test this PR a bit more, but besides that it should be ready for merging.

Kidswiss:
Hi @MichaelEischer 

Thanks for the help with this Pull Request. 

Ah yes, I completely miss understood @fd0's suggestions then, sorry about that.

changelog/unreleased/pull-2255
```diff
@@ -0,0 +1,10 @@
+Bugfix: Fix tar issues when dumping `/`
+
+We've fixed and issue with dumping `/` to tar. As well as dumping files on the
+first sublevel e.g. `/foo`. This also fixes tar dumping issues on Windows where
```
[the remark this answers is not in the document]
Kidswiss: It's quite a while ago that I had a look at this for Windows.

The issue for Windows was basically the same as when you're dumping `/` on Linux. Where `/` under linux contains multiple nodes that weren't iterated correctly in the first version of the tar dumping code.

But I'm not really sure any more what exact circumstances caused this problem on Windows unfortunately, as that was over a year ago.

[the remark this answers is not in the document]
MichaelEischer: @Kidswiss Is there a special reason to mention windows explicitly here? The tar dumping just seems to have been as broken as on other systems?

MichaelEischer:
I think I've found the reason for the test failures: the tar header contains the timestamps rounded to the next second. `os.Stat` on the other hand provides sub-second precision. This meant that the "last modified"-timestamps were different in general. Together with a missing negation of the equality check this means that the test only succeeded if both timestamps were different, which usually was the case.

The timestamp from `os.Stat` is now also rounded to seconds and thus should correctly match the one from the tar header.

My manual tests with dumping a tar file restricted to root or the first sub-level were also successfully.

MichaelEischer:
I've fixed a few further small issues and updated the command description.

@Kidswiss Could you take a quick look at the three newest commits?

I've also done a bit of testing on windows, the generated tar files seem mostly fine. But I've noticed that 7zip failed to open the tar archives once they contained directories. But that is caused by #2319 and should by fixed in another PR.

Kidswiss:
Just had a look at the commits. Looks good to me!

Thanks for the Windows testing. I agree that the issue should be handled in #2319.

changelog/unreleased/pull-2255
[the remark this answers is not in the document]
Kidswiss: No problem :)

If you're happy with the changelog text we can resolve this discussion.

MichaelEischer (approved):
LGTM

PR #2237, opened 2019-04-13
Test CI with Azure Pipelines

fd0:
This is a test branch to try out Azure Pipelines, which offers free CI testing for public projects on Windows, Linux and macOS, replacing Travis as well as Appveyor

PR #2267, opened 2019-05-05
Check gofmt in Travis

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ install:
   - go env
 
 script:
+  - gofmt -w -s . && git diff --exit-code
   - go run run_integration_tests.go
 
 after_success:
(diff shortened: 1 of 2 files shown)


fd0:
Hm, I'm not sure what you're trying here, we already check for `gofmt` in `run_integration_tests.go` :)

And I'd like to switch from Travis/Appveyor to Azure Pipelines, see #2237, for that I've reworked the whole integration test suite. So it's not a good idea to spend time in this area right now.

What are you trying to do?

gliptak:
@fd fail the build if gofmt fails. I missed the current implementation.closing

fd0:
Ok, I understand ;)

PR #2268, opened 2019-05-05
Set go_import_path in Travis

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: go
 sudo: false
 
+go_import_path: github.com/restic/restic
+
 matrix:
   include:
     - os: linux


fd0:
What's this about? :)

gliptak:
Allows Travis builds to work for forks.

fd0:
Ah, interesting! Thanks for the explanation, I did not see this use case before. It'll also be addressed in #2237 though, we'll then run all tests using Go Modules, so the `GOPATH` is not relevant any more.

gliptak:
thanks for the pointers. i see that Azure runs Linux/Mac/Windows from a single config.

but it might have a similar issue, as forks' checkout locations isn't their "real" Github project name

PR #2272, opened 2019-05-10
Specify user and host names when adding keys

### LINKED ISSUE #2175: Specify user and/or host when adding key
Output of `restic version`
--------------------------
```
```

It would be nice to be able to override the user and/or host values when adding a host to restic.

* Sometimes I am adding a key for a new host, but from a different host.
* Sometimes the short hostname on the new machine is ambiguous, so the key is added with the host `www`.  We might have multiple hosts with the short hostname `www`, so this is not useful.

What are you trying to do?
--------------------------
Add a new key with the FQDN of the current host, which does not seem possible.

ahktenzero:

I've been working on managing my restic backups using Saltstack. Each repository has a master password used to create it, and each of the clients backing up to the repo has it's own password. What I want to do is add keys for all the clients to the repository without disclosing the master password to the client machines, which means I need to run the key add command on the backup server. As it is, this means all the keys have the same user and host details, which makes any further key management rather difficult.

I have added two new optional parameters for the key add command which allow you to specify the user and hostname to use for the new key. 

closes #2175 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2175
+++ b/changelog/unreleased/issue-2175
@@ -1,0 +1,9 @@
+Enhancement: Allow specifying user and host when creating keys
+
+When adding a new key to the repository, the username and hostname for the new
+key can be specified on the command line. This allows overriding the defaults,
+for example if you would prefer to use the FQDN to identify the host or if you
+want to add keys for several different hosts without having to run the key add
+command on those hosts.
+
+https://github.com/restic/restic/issues/2175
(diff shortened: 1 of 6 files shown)


ahktenzero:
Can someone take a look at the CI please? The failing tests (rclone on travis, archiving on appveyor) don't appear to be related to the changes I made so I'm not sure what's going on.

fd0:
Thank you very much for your contribution, and I'm sorry that it took us so long to respond. I'm closing this is favor to #2318, thanks!

PR #2285, opened 2019-05-26
Fix JSON Printf issues with format verbs

### LINKED ISSUE #2281: restic json output invalid
Output of `restic version`
--------------------------
restic 0.9.4 compiled with go1.12.1 on linux/amd64

restic -r s3:http://minioip:9000/bucket --json find * -s id | jq .

s3

Expected behavior
-----------------
jq read restic json output correctly

Actual behavior
---------------
jq return me an error:
parse error: Invalid numeric literal at line 1, column 657818146

On windows system, create a file called randomfilename.%202018 and after run restic backup
After the backup, run find command and export the output in json format. You'll see that it's malformed:
[cut]
"path":"/path/randomfilename.%!"(MISSING),"permissions":"-rw-rw-rw-",...
[cut]

no

no

tooooo much!
thanks for this very useful tool!

curiousleo:

Currently, the JSON string is output via `Printf`. This means that if there is a format string inside the JSON, there is an error - see https://github.com/restic/restic/issues/2281 for an example.

Note that I've split the PR into two commits: the first just fixes the bug, the second replaces other occurrences of `Printf` with `fmt.Print` or `fmt.Println` where appropriate.

Fixes https://github.com/restic/restic/issues/2281.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2281
+++ b/changelog/unreleased/issue-2281
@@ -1,0 +1,7 @@
+Bugfix: Handle format verbs like '%' properly in `find` output
+
+The JSON or "normal" output of the `find` command can now deal with file names
+that contain substrings which the Golang `fmt` package considers "format verbs"
+like `%s`.
+
+https://github.com/restic/restic/issues/2281
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -156,7 +156,7 @@ func (s *statefulOutput) PrintPatternJSON(path string, node *restic.Node) {
 	if s.hits > 0 {
 		Printf(",")
 	}
-	Printf(string(b))
+	Print(string(b))
 	s.hits++
 }
 
@@ -168,7 +168,7 @@ func (s *statefulOutput) PrintPatternNormal(path string, node *restic.Node) {
 		s.oldsn = s.newsn
 		Verbosef("Found matching entries in snapshot %s from %s\n", s.oldsn.ID().Str(), s.oldsn.Time.Local().Format(TimeFormat))
 	}
-	Printf(formatNode(path, node, s.ListLong) + "\n")
+	Println(formatNode(path, node, s.ListLong))
 }
 
 func (s *statefulOutput) PrintPattern(path string, node *restic.Node) {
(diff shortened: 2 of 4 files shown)


curiousleo:
I tried to add a test for this but failed. What I tried was to add an `Example`-style unit test (to capture `stdout`, see https://golang.org/pkg/testing/#hdr-Examples) in a new `cmd_find_test.go` for `PrintPatternJSON` but couldn't get it to compile because the `statefulOutput` type could not be found ...

Happy to have another go at writing a test if reviewers think that's desirable.

curiousleo:
This CI failure looks unrelated to my change - could it be flaky?

```
    --- FAIL: TestBackendRclone/TestZZZDelete (0.03s)
        suite.go:183: signal: broken pipe
    backend_test.go:47: cleanup dir /tmp/restic-test-007545188
```
https://travis-ci.com/restic/restic/jobs/203146461#L641

curiousleo:
The code is already exercised by the existing integration tests. To trigger the relevant condition, there would have to be a file with a format string in its name in one of the test backups. I'm not sure how change them - are there instructions on how the test backups used in the integration tests were created in the first place?

fd0 (approved):
Awesome, great catch and PR! Thank you very much for taking the time. I don't think we need so many tests for that. I'm sorry for the long time you had to wait for the feedback.

rawtaz:
@curiousleo Thanks a lot for your help with this. I did remove the commit that replaced Printf() in places that weren't directly needed for this bugfix (i.e. besides your first commit). I did this because Printf() seems to be what's used for printing stuff even when there's no formatting needed. In other words, I did this to keep the code coherent. Will merge soon.

rawtaz:
Weird test failure at https://ci.appveyor.com/project/fd0/restic/builds/30761753#L97 - I checked the last PRs I merged and anything having to do with archiver and its tests have gotten a green light from AppVeyor, but now all of a sudden the TestArchiverAbortEarlyOnError test for archiver fails. This PR is not even related to any archiver changes, so I fail to understand why it hasn't failed before, and does now.

Issue #2284, opened 2019-05-26
`check` slows down quadratically with number of snapshots


BenWiederhake:
3 seconds for every 200 snapshots.  At 72 snapshots per day (3 per hour), this makes `restic check` painful to run.

Output of `restic version`
--------------------------

`restic 0.9.5 compiled with go1.12.4 on linux/amd64`

How did you run restic exactly?
-------------------------------

I make backups every 20 minutes, so 500 snapshots per week.  Please consider the following numbers in this light.

I'll skip my actual setup, and dive right into the:

Steps to reproduce the behavior
-------------------------------

Call `backup` and `check` somewhat often.

[This MRE repo](https://github.com/BenWiederhake/restic-slow) does everything, and collects the data in a nice, somewhat usable format.

Specific setup:
- `RESTIC_REPOSITORY=destdir/ RESTIC_PASSWORD=1234 restic backup fromdir`
- `RESTIC_REPOSITORY=destdir/ RESTIC_PASSWORD=1234 restic check`
- Few changes to the repository, just like my actual home folder doesn't change completely all the time.  The MRE makes no changes at all.  However, this is *unrelated* to #662.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

sftp for my actual setup, and "filesystem" for the MRE.

This should be irrelevant, as I'm complaining about an excessive CPU load.

Expected behavior
-----------------

With more snapshots, `restic check` has more to do, I get it.  But I expect something along the lines of 1 ms per "empty" snapshot.

Actual behavior
---------------

It looks like each snapshots makes `check` slower by roughly 15 ms in terms of actual CPU usage as reported by `/usr/bin/time`.  Here are the measurements from the MRE:

![chart that shows quadratic growth, x axis is the number of snapshot, y axis is time](https://user-images.githubusercontent.com/2690845/58382236-26a44f80-7fc8-11e9-9c4f-d5ea34a44cbc.png)

Note that the quadratic fit is so perfect you can barely see it.  The first few checks take virtually no time, and then it blows up into tens of seconds of high CPU load.  This can be neither explained by latency, nor number of files ("hundreds" is not enough for that kind of slowdown).

I'd like to mention again: You can recreate this with the [MRE script](https://github.com/BenWiederhake/restic-slow).

Do you have any idea what may have caused this?
-----------------------------------------------

My initial guess was that there is an "accidentally quadratic" bug, so I was pretty surprised to see linear growth on the graph.  So it's probably not that.

[...]

PR #2294, opened 2019-06-03
build: Preserve debug symbols in debug build

BenWiederhake:

This change enables debug symbols in debug builds.
(It does so by disabling the stripping of debug symbols.)

No, but the impact is low, and it would have helped with #2284 a lot.

Checklist
---------

- [/] I have added tests for all changes in this PR EDIT: See below
- [/] I have added documentation for the changes (in the manual) EDIT: None necessary.
- [/] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)) EDIT: None necessary.

Does this need testing at all?  I don't think so, but you tell me what should be done.

### DIFF
--- a/build.go
+++ b/build.go
@@ -395,8 +395,12 @@ func main() {
 
 	verbosePrintf("detected Go version %v\n", goVersion)
 
+	preserveSymbols := false
 	for i := range buildTags {
 		buildTags[i] = strings.TrimSpace(buildTags[i])
+		if buildTags[i] == "debug" || buildTags[i] == "profile" {
+			preserveSymbols = true
+		}
 	}
 
 	verbosePrintf("build tags: %s\n", buildTags)
@@ -423,7 +427,11 @@ func main() {
 	if version != "" {
 		constants["main.version"] = version
 	}
-	ldflags := "-s -w " + constants.LDFlags()
+	ldflags := constants.LDFlags()
+	if !preserveSymbols {
+		// Strip debug symbols.
+		ldflags = "-s -w " + ldflags
+	}
 	verbosePrintf("ldflags: %s\n", ldflags)
 
 	var (


BenWiederhake:
(Flaky connection on appveyor, restarting.)

build.go
```diff
@@ -553,7 +557,13 @@ func main() {
 	if version != "" {
 		constants["main.version"] = version
 	}
-	ldflags := "-s -w " + constants.LDFlags()
+	var ldflags string
+	if isDebug {
+		// Preserve debug symbols.
+		ldflags = constants.LDFlags()
+	} else {
+		ldflags = "-s -w " + constants.LDFlags()
```
MichaelEischer: Could you change that to only add the "-s -w" prefix if `isDebug` is false? That would avoid the code duplication

build.go
```diff
@@ -525,8 +525,12 @@ func main() {
 
 	verbosePrintf("detected Go version %v\n", goVersion)
 
+	isDebug := false
 	for i := range buildTags {
 		buildTags[i] = strings.TrimSpace(buildTags[i])
+		if buildTags[i] == "debug" {
```
MichaelEischer: Please change this to also match the "profile" tag.

BenWiederhake:
@MichaelEischer That makes sense.  I changes it accordingly.

I don't know why it fails on MacOS, but given that it works in the other 4 configs, I assume it's not my fault, and ignore the error.

Have a nice day :)

MichaelEischer:
Thanks for the changes. The macOS run seems to have failed before it even started O.o

fd0:
It's odd, but probably something at Travis was broken. I cannot restart the build since it's too old, so I've amended the latest commit (without changing anything) and force-pushed the branch, so Travis will run again.

fd0 (approved):
LGTM, can be merged once the CI is happy.

PR #2303, opened 2019-06-11
Call b2.Reader.Verify() when reading from b2 backend

### LINKED ISSUE #2302: Intermittent b2 errors (?) cause "pack ID does not match" errors during "restic check --read-data"
Output of `restic version`
--------------------------
`restic 0.9.5 compiled with go1.12.4 on freebsd/amd64`

I have a few layers of scripts, but roughly:

```
export RESTIC_PASSWORD_FILE=...
export RESTIC_REPOSITORY=b2:...

SIZE=1000
for x in `seq 1 $SIZE`; do
  echo $x/$SIZE
  ./restic_0.9.5_freebsd_amd64 check --read-data-subset=$x/$SIZE
done
```

The output relevant to this issue is, e.g.:

```
18/1000
using temporary cache in /tmp/restic-check-cache-158036280
created new cache in /tmp/restic-check-cache-158036280
load indexes
check all packs
read group #18 of 2021 data packs (out of total 510823 packs in 1000 groups)
Pack ID does not match, want 11ec3b05, got f54a743e
[23:05] 100.00%  2021 / 2021 items

duration: 23:05
```

b2

Expected behavior
-----------------

restic should use check b2 full file downloads against the sha1 integrity information given by b2, retrying if necessary. This will help distinguish between transient errors and more serious errors.

Actual behavior
---------------

restic catches the transient corruption and reports "pack ID does not match".

Due to the nature of the bug, there's really no way to reproduce this. I saw it happen with 5 pack files in a 2.5TB repository, so perhaps running check against a large enough b2 repo will eventually encounter it.

Yes, after I had manually verified that the pack files had the correct sha256 sum (using the info from https://github.com/restic/restic/issues/2191 ), I dug into the b2 backend code and noticed that the b2 library it's using (blazer) already keeps track of sha1 info returned by b2 and exposes methods to check it, but the b2 backend doesn't use it.

Yes, I will send a PR with an initial solution.

restic is great otherwise! I've been very happy with it.

akalin:
It calls b2.Reader.Verify(), which verifies files downloaded from b2 with the sha1
information provided by b2 (which is already being calculated, but not used).

Filed https://github.com/restic/restic/issues/2302 for the underlying issue. Closes #2302.

Checklist
---------

### DIFF
--- a/internal/backend/b2/b2.go
+++ b/internal/backend/b2/b2.go
@@ -147,6 +147,22 @@ func (be *b2Backend) Load(ctx context.Context, h restic.Handle, length int, offs
 	return backend.DefaultLoad(ctx, h, length, offset, be.openReader, fn)
 }
 
+type verifyOnCloseReader struct {
+	*b2.Reader
+	name string
+}
+
+func (r verifyOnCloseReader) Close() error {
+	err := r.Reader.Close()
+	if err != nil {
+		return err
+	}
+
+	verr, hashCalculatable := r.Reader.Verify()
+	debug.Log("Verify result for %q: err=%v, hash calculatable=%t\n", r.name, verr, hashCalculatable)
+	return nil
+}
+
 func (be *b2Backend) openReader(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error) {
 	debug.Log("Load %v, length %v, offset %v from %v", h, length, offset, be.Filename(h))
 	if err := h.Valid(); err != nil {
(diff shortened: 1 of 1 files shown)


akalin:
I left 'added tests' unchecked since I'm not sure where exactly to add tests. I see the tests in `backend/test`, but that seems more like a place to put backend-agnostic tests. Suggestions?

I'm also not sure what doc changes this change would require, but I'm happy to add any.

akalin:
I did run TestBackendB2 against a test bucket manually, though, and it passed.

ifedorenko:
Personally, I am not in favour of this change. Restic already validates data sha256 checksums during  `restore`, `check` and other places where it reads the data from the backend, so the extra level of validation seems redundant.

fd0:
Thanks for your contribution! I think we have enough checks within restic in other places so I don't think we need this check. In order to not complicate the code too much, I'm therefore closing this PR. Please feel free to leave comments, thanks!

PR #2311, opened 2019-06-14
filter: ability to use negative patterns

### LINKED ISSUE #233: Allow mixing include and exclude patterns for backup and restore
This is a followup issue for #226. Up to now it is only possible to specify exclude patterns for backup and restore.

There are uses cases that aren't covered, for example: A user would like to backup her home directory at `/home/user`, excluding anything from the directory `work` except C source files.

Are there any other uses I haven't thought of?

Implementation/User Interface: Allow specifying both `--include` and `--exclude` patterns on the command line, which fill a common list (order does matter here). For each dir/file, check all patterns in the list. The action (exclude or include) of the last mattching pattern wins. The default action (which will also be used if neither include nor exclude patterns are specified) is "include".

This already contains a corner case: Should restic walk `/home/user/work` at all?

I think it should not. Instead, if this is a desirable behaviour, require the user to add a more specific pattern to signal that the excluded directory should also be walked, e.g. `restic backup --exclude /home/user/work --include /home/user/work/**/*c. /home/user`.

A better user interface would be to allow specifying a file to read the patterns from (`--pattern-file` or something like that). In this file, all lines starting with `#` are comments, empty lines ignored, all other lines must start with either `+` (include) or `-` (exclude) following a space character and a pattern. For the use case above a filter file would look like this:

```
# filter out everything from work, but include c source code files
- /home/user/work
+ /home/user/work/**/*.c
```

This issue can be closed once a definitive solution for having both include and exclude patterns is implemented.

vincentbernat:
This is quite similar to gitignore. If a pattern is prefixed by an
exclamation mark and match a file that was previously matched by a
regular pattern, the match is cancelled. Notably, this can be used
with `--exclude-file` to cancel the exclusion of some files.

Like for gitignore, once a directory is excluded, it is not possible
to include files inside the directory. For example, a user wanting to
only keep `*.c` in some directory should not use:

    ~/work
    !~/work/*.c

But:

    ~/work/*
    !~/work/*.c

I didn't write documentation or changelog entry. I would like to get
feedback if this is the right approach for excluding/including files
at will for backups. I use something like this as an exclude file to
backup my home:

    $HOME/**/*
    !$HOME/Documents
    !$HOME/code
    !$HOME/.emacs.d
    !$HOME/games
    # [...]
    node_modules
    *~
    *.o
    *.lo
    *.pyc
    # [...]
    $HOME/code/linux/*
    !$HOME/code/linux/.git
    # [...]

There are some limitations for this change:

 - Patterns are not mixed accross methods: patterns from file are
   handled first and if a file is excluded with this method, it's not
   possible to reinclude it with `--exclude !something`.

 - Patterns starting with `!` are now interpreted as a negative
   pattern. I don't think anyone was relying on that.

 - The whole list of patterns is walked for each match. We may
   optimize later by exiting early if we know no pattern is starting
   with `!`.

Fix #233

Checklist
---------


danielsharvey:
Looks good, although at present I don't think it handles this case:
```
$HOME/**/*
!$HOME/local/bin
```

i.e. the current logic excludes `$HOME/local` and so the negate for `$HOME/local/bin` is missed. 

This solution needs special handling of the `childMayMatch` combined with `negate` case.

I've worked some logic which appears to work and will add tomorrow. It needs to consider both exclude and include cases.

vincentbernat:
Yes, you are right, this is not handled correctly.

danielsharvey:
Understood, thanks, although I think highly preferable to handle this as per the two-line formulation. Thoughts?

Also, this PR does not currently handle restore presently (same filter code used in `cmd_restore.go`).

e.g. this should restore `local/bin` files starting with 'a' through 'm' but restores nothing
```
restic restore -v --include '/local/bin' --include '!/local/bin/[a-m]*' latest
```

vincentbernat:
For this example, yes. But what about:

```
$HOME/build
test.*
!test.c
```

Should we include `$HOME/build/something/something/test.c` (and then walk `$HOME/build`)?

danielsharvey:
Regarding general logic, I think you are right. The additional terms are better than the requirement to walk additional subtrees. Thanks.

In regards to the restore, I'm presuming you mean `--include` not `--exclude`. But in any case the modification of the term to `/local/bin/*` does not help.

I believe the issue is that the `childMayMatch` should not reset in the same way as the primary match term - it should retain the behaviour of triggering to `true` if *any* pattern indicates that a child may match.

So 
```
childMayMatch = childMayMatch && !c || c && !negate
```
becomes
```
childMayMatch = childMayMatch || c && !negate
```

This works for my restore case. Thoughts?

internal/filter/filter.go
```diff
@@ -177,12 +184,8 @@ func List(patterns []string, str string) (matched bool, childMayMatch bool, err
...
 		}
 
-		matched = matched || m
-		childMayMatch = childMayMatch || c
-
-		if matched && childMayMatch {
-			return true, true, nil
-		}
+		matched = matched && !m || m && !negate
+		childMayMatch = childMayMatch && !c || c && !negate
```
danielsharvey: I believe this should be:
```suggestion
		childMayMatch = childMayMatch || c && !negate
```

I believe the issue is that the `childMayMatch` should not reset in the same way as the primary match term - it should retain the behaviour of triggering to true if *any* pattern indicates that a child may match.

danielsharvey:
I also suggest another expansion of the filter tests is warranted - a few extra cases and capturing the `childMayMatch` return value in the tests. See https://github.com/danielsharvey/restic/commit/7cb1815dc70f7fd10ea841b0142de15ba02c325a

Thoughts?

vincentbernat:
You said <q>e.g. this should restore local/bin files starting with 'a' through 'm' but restores nothing</q>, but you have `--include "!local/bin/[a-m]*"`, so it should not restore files starts with 'a' through 'm', right? So, I am a bit confused by the example.

But you are right, there is something different about restore. Just using a negative pattern is enough to make it fail. We should add tests for the this as there is currently none.

danielsharvey:
> You said e.g. this should restore local/bin files starting with 'a' through 'm' but restores nothing, but you have `--include "!local/bin/[a-m]*"`, so it should not restore files starts with 'a' through 'm', right? So, I am a bit confused by the example.

Argh! Sorry - the test restore (that works with the change I suggest) is to **exclude** 'a' through 'm' and **include** 'n' through 'z'.

danielsharvey:
> But you are right, there is something different about restore. Just using a negative pattern is enough to make it fail. We should add tests for the this as there is currently none.

Yes. The tests I suggest in my commit https://github.com/danielsharvey/restic/commit/7cb1815dc70f7fd10ea841b0142de15ba02c325a and https://github.com/danielsharvey/restic/commit/6a4938affc559d16ed98e28d7f137431c5ff5ab5 that capture this and the other checks we've been discussing are:
```
{[]string{"/**/*", "!/foo", "/foo/*", "!/foo/bar"}, "/foo/other/test.go", true, true},
{[]string{"/**/*", "!/foo", "/foo/*", "!/foo/bar"}, "/foo/bar/test.go", false, false},
{[]string{"/**/*", "!/foo", "/foo/*", "!/foo/bar"}, "/foo/bar/test.go/child", false, false},
{[]string{"/**/*", "!/foo", "/foo/*", "!/foo/bar", "/foo/bar/test*"}, "/foo/bar/test.go/child", true, true},
{[]string{"/foo/bar/*", "!/foo/bar/[a-m]*"}, "/foo", false, true},
{[]string{"/foo/**/test.c"}, "/foo/bar/foo/bar/test.c", true, true},
{[]string{"/foo/*/test.c"}, "/foo/bar/foo/bar/test.c", false, false},
```

But, I still had not properly considered the `childMayMatch` case in the negation case. I think the logic needs to be:
```
matched = matched && !m || m && !negate

if negate {
  // a matching negation will 'cover' child matches
  if m && c {
    childMayMatch = false
  }
} else {
  childMayMatch = childMayMatch || c
}
```

vincentbernat:
Hey!

Sorry for the late answer, I was a bit busy lately. So you were right. I have added a commit with tests for `childMayMatch` before the change and I have updated the commit for the change to use your logic (but I have formatted to make it look like the logic for updating `match`). I agree with the results of the tests. This seems to work now as expected.

danielsharvey:
Thanks, looks good to me.

vincentbernat:
Thanks @danielsharvey for the review! I have added the documentation and the changelog entry. The PR should be ready for a formal review.

internal/filter/filter.go
```diff
@@ -177,12 +184,8 @@ func List(patterns []string, str string) (matched bool, childMayMatch bool, err
...
 		}
 
-		matched = matched || m
-		childMayMatch = childMayMatch || c
-
-		if matched && childMayMatch {
-			return true, true, nil
-		}
+		matched = matched && !m || m && !negate
+		childMayMatch = childMayMatch && !(m && c) || c && !negate
```
MichaelEischer: If I understand `match` and `childMatch` correctly, `m==true` implies `c==true`. That would simplify the last line to `childMayMatch && !m || c && !negate`.

That would lead to a simpler explanation for this bit magic: the matched/childMayMatch flag is reset if the pattern fully matches, it is set again if this is a not-negated pattern and match/childMatch were successful.

It took me a rather long time to fully understand the effects of these Boolean expressions and why these are correct. I've tried to find a formula that's more intuitive to understand, but I'm not sure whether the result is actually better:
```
matched = !negate && (matched || m) || (negate && matched && !m)
childMayMatch = !negate && (childMayMatch || c) || (negate && childMayMatch && !m)
```

So I think these two lines need a comment at the very least.

vincentbernat: While your `matched` formula is equivalent, the `childMayMatch` is not. Are the tests successful with your version?

MichaelEischer: Yes the tests are successful. In case `m==true implies c==true` is correct, this simplifies `m && c` to `m` and then you end up with my variant of the `childMayMatch` formula.

vincentbernat: Oh, OK, I didn't catch the implication. Fine by me.

MichaelEischer: Which variant of the matched, childMayMatch expressions do you find more readable? Another option would be add an if-else-block for the normal and negated version. That would be a few more lines of code, but probably the easiest to understand variant.

danielsharvey:
@fd0 Is there any process that we should follow to progress PRs like this one? I have been using this for some time and would love to see it merged. Thanks!

rawtaz:
Following up on the comments by @MichaelEischer would be the process, but in the end it boils down to the core developers' time constraints. There are many other changes that are being worked on, patience is needed :)

internal/filter/filter.go
vincentbernat: I don't have a strong opinion about this. The two options would be:

```go
matched = matched && !m || m && !negate
childMayMatch = childMayMatch && !m || c && !negate
```

Or

```go
if negate {
   matched = matched && !m
   childMayMatch = childMayMatch && !m
} else {
   matched = matched || m
   childMayMatch = childMayMatch || c
}
```

If it is correct, the latest form has some "logical look" in it. So I would go with it.

vincentbernat:
I have rebased the commit and use the second approach, suggested by @MichaelEischer

internal/filter/filter.go
danielsharvey: Thanks @vincentbernat. I agree with @MichaelEischer in finding the if-else formulation easier to read.

danielsharvey:
Thanks @rawtaz, I appreciate the challenge of volunteer efforts. And apologies, I had missed that @MichaelEischer's comments had not been fully responded to (now done).

rawtaz:
To anyone else; Has anyone else tested this in production? Please let us know :)

danielsharvey:
I'm not "anyone else", but just to add clarity to my use cases (in regular use by me):

1. I've attached one of my standard backups, which runs off `restic ... --exclude-file ~/restic-excludes-negative ~/`

2. I also have a continuous backup script which uses a file watcher to detect filesystem updates, and then run a differential backup every 15 minutes based upon a dynamically generated exclude file similiar to #1.

Attachment: [restic-excludes-negative](https://github.com/restic/restic/files/4840600/restic-excludes-negative.txt)

vincentbernat:
If there is still hope to get this merged, I can try another rebase...

changelog/unreleased/issue-233
```diff
@@ -0,0 +1,31 @@
+Enhancement: Add negative patterns for include/exclude
+
+If a pattern is suffixed by an exclamation mark and match a file that
```
fd0, line 3: It's probably "prefixed", not "suffixed"?

doc/040_backup.rst
```diff
@@ -211,6 +211,28 @@ On most Unixy shells, you can either quote or use backslashes. For example:
  * ``--exclude="foo bar star/foo.txt"``
  * ``--exclude=foo\ bar\ star/foo.txt``
 
+If a pattern is suffixed by an exclamation mark and match a file that
```
fd0, line 214: Again "prefixed" I think.

fd0 (changes requested):
I'm sorry for the long delay, I've talked with @MichaelEischer and we're inclined to include this PR. Can you please rebase again?

I've not looked at the code too deeply, but I've seen two mistakes in the documentation and changelog.

vincentbernat:
I did have a look, but there are too many changes to easily rebase. If anyone wants to give it a try, they are welcome to take over this PR.

fd0:
I understand, thanks for your work!

changelog/unreleased/issue-233
MichaelEischer: Is changed.

doc/040_backup.rst
MichaelEischer: Is changed.

MichaelEischer (commented):
@fd0 I've rebased the PR. The check whether a pattern is negated now occurs once while preparing the pattern. That required some refactoring, in separate commits, to change the Pattern type from a slice to a struct.

x80486:
Would be really nice to expedite this work — it's been done since a while 😏 ...and it's much needed.

rawtaz:
@x80486 Comments like that are neither helpful or make things happen more quickly. In particular when a core developer was attending the PR just two days ago.

danielsharvey:
@MichaelEischer @fd0 Thanks for all the work. Just checking in - any assistance required to finalise this?

fd0 (approved):
Thank you very much for your patience, I'm approving this PR!

cowwoc:
Does the new exclamation mark functionality work for inclusions using `--files-from`? The documentation only mentions using it for exclusions, not inclusions.

MichaelEischer:
No, the negative patterns are only available for exclusions. A negative pattern in files-from would actually be an exclude, so it should just work to specify it as such?

cowwoc:
@MichaelEischer Are we able to represent any sort of ordering with this approach?

Meaning, can I specify the following?

* Exclude /foo but
* Include /foo/one, /foo/two

If I remember correctly, the exclusion of the parent directory will override any inclusion of the children directories.

danielsharvey:
@cowwoc You absolutely can.

My backup selectively backs up using this mechanism:
```sh
restic <...> --exclude-file ~/backup/readme/restic-excludes-negative ~/
```

My excludes file looks like this:
```
/*
!/Users
/Users/*
!/Users/daniel

/$HOME/*

!$HOME/Desktop
!$HOME/Documents
!$HOME/Downloads

# ...
```

Note the pattern:
1. Exclude everything
2. "Unexclude" the folder you want to include
3. Exclude everything (all contents - `*`) in that folder
4. Unexclude the contents of the folder that you want to include
5. Repeat

jmalm:
Should negative exclusion work with Windows paths? If so, what is the syntax?

I'm trying to backup my G:, ignoring everything and then _not_ ignoring specific directories:

restic backup --exclude-file .\test.exclude G:

.\test.exclude:
```
G:\*
!G:\file.txt
!G:\1080p
```

Only the root directory is included in the backup, no content. So the exclusion pattern (G:\*) works fine, but not the negative patterns. I've also tested with forward slashes, same result.

PR #2316, opened 2019-06-24
Add 'epoch' field to JSON output for 'snapshot' command

### LINKED ISSUE #2315: Facilitate Monitoring (Last Backup Age)
Output of `restic version`
--------------------------
`restic 0.9.5 compiled with go1.12.4 on linux/amd64`

Enable a way to easily programatically determine the last backup time/age.

We have `--json` output for `snapshots` command, however the time of each snapshot isn't machine friendly. For example, `jq` gives the error `Cannot index array with string` if you try to sort the JSON output by the `time` attribute.

I don't think I can assume the JSON output will be produced in chronological order and just take the last object in the array?

Adding another attribute to each object in the JSON output that gives the `time` field formatted as an epoch would help.

What are you trying to do?
--------------------------
Write a script (eg, for Nagios/Monit/etc) to monitor the last backup age.

restic makes me smile like rainbow unicorn poops :D

fukawi2:
Adds JSON output for the 'snapshot' command to include the timestamp of each snapshot formatted as a unix epoch value. This makes it easier for machine processing of the timestamp (eg, sorting)

Closes #2315.

Checklist
---------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ Summary
  * Enh #2139: Return error if no bytes could be read for `backup --stdin`
  * Enh #2205: Add --ignore-inode option to backup cmd
  * Enh #2220: Add config option to set S3 storage class
- * Enh #2315: Add 'epoch' field to json output for 'snapshot' command
 
 Details
 -------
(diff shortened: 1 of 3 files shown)


fukawi2:
Just documenting discussion from IRC. @fd0 had concerns around lack of timezone in an epoch field. After investigation, the go 'time' library function `.Unix()` converts to UTC before converting to an epoch, so the epoch will always be in UTC regardless of the original timezone. Perhaps appropriate to rename the field in the JSON output to reflect this?

mholt:
Since you're scripting anyway, what's wrong with the Time field of the snapshot exactly?

fukawi2:
@mholt Oh shit. I'm an idiot. The 'time' is formatted in a sortable format :man_facepalming: 

Although not if there are snapshots from different timezones. And if that format ever changes then it possibly won't be.

fukawi2:
I no longer work at the company where this change was appropriate. This can be closed, if it's not going to be merged?

PR #2318, opened 2019-06-26
Allow specifying user and host when adding keys

### LINKED ISSUE #2175: Specify user and/or host when adding key

pschultz:
This is an alternative to #2272, which proposes a special argument format to specify username and hostname, which I find unintuitive. This version uses normal flags (`--user` and `--host`) instead.

I copied the changelog entry directly from #2272, so credit goes to ahktenzero.

Simplifies automated key managment by adding the `--user` and `--host` flags for the `key add` command.

Prior PR in #2272, closes #2175.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2175
+++ b/changelog/unreleased/issue-2175
@@ -1,0 +1,9 @@
+Enhancement: Allow specifying user and host when creating keys
+
+When adding a new key to the repository, the username and hostname for the new
+key can be specified on the command line. This allows overriding the defaults,
+for example if you would prefer to use the FQDN to identify the host or if you
+want to add keys for several different hosts without having to run the key add
+command on those hosts.
+
+https://github.com/restic/restic/issues/2175
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -32,13 +32,19 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 	},
 }
 
-var newPasswordFile string
+var (
+	newPasswordFile string
+	keyUsername     string
+	keyHostname     string
+)
 
 func init() {
 	cmdRoot.AddCommand(cmdKey)
 
 	flags := cmdKey.Flags()
 	flags.StringVarP(&newPasswordFile, "new-password-file", "", "", "the file from which to load a new password")
+	flags.StringVarP(&keyUsername, "user", "", "", "the username for new keys")
+	flags.StringVarP(&keyHostname, "host", "", "", "the hostname for new keys")
 }
 
 func listKeys(ctx context.Context, s *repository.Repository, gopts GlobalOptions) error {
@@ -120,7 +126,7 @@ func addKey(gopts GlobalOptions, repo *repository.Repository) error {
 		return err
 	}
 
-	id, err := repository.AddKey(gopts.ctx, repo, pw, repo.Key())
+	id, err := repository.AddKey(gopts.ctx, repo, pw, keyUsername, keyHostname, repo.Key())
 	if err != nil {
 		return errors.Fatalf("creating new key failed: %v\n", err)
 	}
@@ -151,7 +157,7 @@ func changePassword(gopts GlobalOptions, repo *repository.Repository) error {
 		return err
 	}
 
-	id, err := repository.AddKey(gopts.ctx, repo, pw, repo.Key())
+	id, err := repository.AddKey(gopts.ctx, repo, pw, "", "", repo.Key())
 	if err != nil {
 		return errors.Fatalf("creating new key failed: %v\n", err)
 	}
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -639,6 +639,28 @@ func testRunKeyAddNewKey(t testing.TB, newPassword string, gopts GlobalOptions) 
 	rtest.OK(t, runKey(gopts, []string{"add"}))
 }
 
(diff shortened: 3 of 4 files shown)


internal/archiver/archiver_test.go
```diff
@@ -2061,8 +2061,8 @@ func TestMetadataChanged(t *testing.T) {
 
 	// set some values so we can then compare the nodes
 	want.Content = node2.Content
+	want.ExtendedAttributes = node2.ExtendedAttributes
```
pschultz: This test failed on my machine due to selinux:

```txt
--- FAIL: TestMetadataChanged (0.00s)
    archiver_test.go:2043: chdir to /tmp/restic-test-637943804
    archiver_test.go:2069: metadata does not match:
        {*restic.Node}.ExtendedAttributes:
                -: []restic.ExtendedAttribute(nil)
                +: []restic.ExtendedAttribute{{Name: "security.selinux", Value: []uint8{0x75, 0x6e, ..., 0x30, 0x00}}}
    panic.go:406: chdir back to /home/pschultz/src/restic/internal/archiver
FAIL
FAIL    github.com/restic/restic/internal/archiver      2.598s
```

I'm not sure why this test expected the ExtendedAttributes to be nil after snapshotting. [According to the manual ExtendedAttributes are preserved](https://github.com/restic/restic/blob/master/doc/manual_rest.rst#metadata-handling).

> Restic saves and restores most default attributes, including extended attributes like ACLs.

MichaelEischer: My guess would be that it made the test work. If the node has no ExtendedAttribute, then the node create from NodeFromFileInfo has a value of `[]restic.ExtendedAttribute()`. The node loaded from the repository has `[]restic.ExtendedAttribute(nil)`. These two values differ, causing the test to fail.

The Lstat wrapper can't control the ExtendedAttributes, therefore your change seems to be the only easy way to fix the problem.

ahktenzero:
I was checking on the PR I submitted for this feature and saw this one, I'm mostly in favour of merging this instead of mine but I'm not sure about the choice of option names. Restic uses `--host` in most commands for filtering by host, which is why I chose to use `key-hostname` instead and leave the possibility of having `--user` and `--host` options for filtering in `restic key list` or `restic key remove`.

Can I suggest `--new-user` and `--new-host` as the option names?  This would also be consistent with the existing `--new-password-file` option for the key commands.

pschultz:
@ahktenzero, I have to disagree with you regarding the flag names.

Even if key filtering is added, it makes much more sense to me to use the same flag for filtering by some attribute as for setting that attribute. Having just `--user` and `--host` flags everywhere where there are user and hostnames involved is much easier to learn than a bunch of different flags that refer to the exact same thing. The `backup` and `snapshot list` commands don't use different flags for the hostname either, for instance.

The --new-password-file was necessary to avoid a collision with the global --password-file flag. When changing the password for a key one has to be able to provide both the existing and new password, so there have to be two flags. But that doesn't apply to user and hostnames.

doc/bash-completion.sh
```diff
@@ -836,6 +836,10 @@ _restic_key()
     local_nonpersistent_flags+=("--help")
     flags+=("--new-password-file=")
     local_nonpersistent_flags+=("--new-password-file=")
+    flags+=("--user")
```
fd0: This needs to be removed from the PR, the bash completion files are automatically generated for each release.

doc/man/restic-key.1
```diff
@@ -27,6 +27,13 @@ The "key" command manages keys (passwords) for accessing the repository.
 \fB\-\-new\-password\-file\fP=""
     the file from which to load a new password
 
+.PP
```
fd0: Same here, the manpage is automatically generated.

internal/archiver/archiver_test.go
fd0: Leave it in for now, I don't see an easier way.

fd0 (changes requested):
Awesome, thank you very much for your contribution! I'm very sorry that it took us so long to respond. I hope you're still interested.

Would you mind removing the changes to the man page and the completion file? And maybe rebase it to the current master branch?

Unfortunately I'm not allowed to push to the branch in your fork, otherwise I would've done that. Please let me know if you like any help. Thanks!

rawtaz:
@pschultz You can give access for core contributors to push to your PR by editing the settings for that in the PR. See: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

pschultz:
I remove the changes to doc/* and rebased on master.

internal/archiver/archiver_test.go
MichaelEischer: This problem has been fixed on master. Could you please rebase onto the current master and remove the test fix?

pschultz: Done.

samson4649:
@fd0 - is this waiting on anything?

samson4649:
@fd0 - Does this need anything more before merging?

fd0 (approved):
LGTM, thanks!

PR #2322, opened 2019-06-30
minor refactoring

brualan:

This is the first step in solving the problems found by static analyzers. The ultimate goal: reduce the amount of code, increase readability, improve performance

Checklist
---------

- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changlog/TEMPLATE))

These are the first commits I want to make. Not sure how it will be better: gradually do the integration of the PR or complete the work and merge one large PR. I use `golangci-lint run --enable-all` to find problem areas of code

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -175,11 +175,7 @@ const MaxCacheAge = 30 * 24 * time.Hour
 
 func validCacheDirName(s string) bool {
 	r := regexp.MustCompile(`^[a-fA-F0-9]{64}$`)
-	if !r.MatchString(s) {
-		return false
-	}
-
-	return true
+	return r.MatchString(s)
 }
 
 // listCacheDirs returns the list of cache directories.
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -24,7 +24,7 @@ type Cache struct {
 }
 
 const dirMode = 0700
-const fileMode = 0600
+const fileMode = 0644
 
 func readVersion(dir string) (v uint, err error) {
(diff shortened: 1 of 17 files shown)


internal/options/options.go
```diff
@@ -22,10 +22,8 @@ func Register(ns string, cfg interface{}) {
 
 // List returns a list of all registered options (using Register()).
 func List() (list []Help) {
-	list = make([]Help, 0, len(opts))
-	for _, opt := range opts {
-		list = append(list, opt)
-	}
+	list = make([]Help, len(opts))
+	copy(list, opts)
```
fd0: Oh, wow, how did I not see that!

internal/cache/cache.go
```diff
@@ -24,7 +24,7 @@ type Cache struct {
 }
 
 const dirMode = 0700
-const fileMode = 0600
+const fileMode = 0644
```
fd0: Good find!

fd0 (approved):
Awesome, thank you very much for taking the time to clean this up! When I have a bit more spare time I plan to replace Hound with golang-ci, so we get these kinds of checks for all PRs.

Also, it was very helpful that you split the changes by type, so I could easily review them!

brualan:
Hi, I’m pretty busy lately. As soon as I have the opportunity, I will continue the work. Please merge this PR, so that it does not hang a dead weight and the time spent on it was not wasted

fd0:
Will do, thanks for your work!

PR #2327, opened 2019-07-06
Skip tree serialization

MichaelEischer:

Restic currently spends a lot of time serializing unchanged TreeNodes which are dropped after hashing the the resulting blob. This change compares the created TreeNode with the previous version from the repository. In case both have matching content, the serialization step is skipped and the blob ID of the old TreeNode is reused.

The TestArchiverSaveDirIncremental testcase is adapted to partially check the new code. As the new behavior is active when backing up changed trees, it is implicitly tested by the existing testcases. The pull request also contains a slightly unrelated fix for an archiver testcase.

However, I am unsure how to best check that the tree serialization is indeed skipped. Extending the Stats object for the test only seems to be the wrong way for me.

No.

Checklist
---------

- [x] No user visible change - There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -208,7 +208,7 @@ func (arch *Archiver) loadSubtree(ctx context.Context, node *restic.Node) *resti
 
 // SaveDir stores a directory in the repo and returns the node. snPath is the
 // path within the current snapshot.
-func (arch *Archiver) SaveDir(ctx context.Context, snPath string, fi os.FileInfo, dir string, previous *restic.Tree) (d FutureTree, err error) {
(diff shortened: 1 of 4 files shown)


h0tw1r3:
Tested with a rest backend. Hangs waiting on a futex after first `uploaded intermediate index` message in output.

MichaelEischer:
Converting this to a draft, as this PR at least needs a to verify that Node.Equals works properly. It might be possible to implement using reflections.

MichaelEischer:
I no longer think this is a good idea: If we add new optional fields to a tree, then older restic versions with this PR could accidentally end up using tree blobs which contain fields that are not known to the older version.

PR #2328, opened 2019-07-06
Fix duplicate tree checks within `restic check`

MichaelEischer:

Running `restic check` for large repositories is currently really slow. I had one run for an 8 TB repository with roughly 40 million files that took approximately one week to complete (Using GOMAXPROCS=1 to avoid interference with other processes on the server)

This is partially caused by decoding for each snapshot, every referenced tree blobs. The integrity check later on checks a tree blob only a single time and ignores later repetitions. Decoding a JSON file is not particularly fast (and the necessary index lookups are even slower, see #2284 ).

This change ensures that tree blobs are only decoded once. As I had to modify the blobRefs map anyways, it is now also used to track whether a blob is listed in an index or not, thereby obsoleting the blob IDSet.

The checker now also traverses the tree blobs in depth-first order which better matches the traversal order of the backup code.

The change is related but not identical to issue #2284 .

Checklist
---------

- [x] No user visible changes - I have added documentation for the changes (in the manual)

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -27,7 +27,6 @@ type Checker struct {
 		sync.Mutex
 		M map[restic.ID]bool
 	}
-	indexes map[restic.ID]*repository.Index
 
 	masterIndex *repository.MasterIndex
 
@@ -40,7 +39,6 @@ func New(repo restic.Repository) *Checker {
 		packs:       restic.NewIDSet(),
 		blobs:       restic.NewIDSet(),
 		masterIndex: repository.NewMasterIndex(),
-		indexes:     make(map[restic.ID]*repository.Index),
 		repo:        repo,
 	}
 
@@ -152,7 +150,6 @@ func (c *Checker) LoadIndex(ctx context.Context) (hints []error, errs []error) {
 				continue
 			}
 
-			c.indexes[res.ID] = res.Index
 			c.masterIndex.Insert(res.Index)
 
 			debug.Log("process blobs")
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -535,8 +535,9 @@ func filterTrees(ctx context.Context, backlog restic.IDs, loaderChan chan<- rest
 				debug.Log("received job with nil tree pointer: %v (ID %v)", j.error, j.ID)
 				err = errors.New("tree is nil and error is nil")
 			} else {
-				debug.Log("subtrees for tree %v: %v", j.ID, j.Tree.Subtrees())
-				for _, id := range j.Tree.Subtrees() {
+				subtrees := j.Tree.Subtrees()
+				debug.Log("subtrees for tree %v: %v", j.ID, subtrees)
+				for _, id := range subtrees {
 					if id.IsNull() {
 						// We do not need to raise this error here, it is
 						// checked when the tree is checked. Just make sure
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -25,7 +25,7 @@ type Checker struct {
 	blobRefs struct {
 		sync.Mutex
 		// see flags below
-		M map[restic.ID]blobStatus
+		M map[restic.BlobHandle]blobStatus
 	}
 
 	masterIndex *repository.MasterIndex
@@ -36,9 +36,8 @@ type Checker struct {
 type blobStatus uint8
 
 const (
-	blobExists blobStatus = 1 << iota
-	blobReferenced
-	treeProcessed
+	blobStatusExists blobStatus = 1 << iota
+	blobStatusReferenced
 )
 
 // New returns a new checker which runs on repo.
@@ -49,7 +48,7 @@ func New(repo restic.Repository) *Checker {
 		repo:        repo,
 	}
 
-	c.blobRefs.M = make(map[restic.ID]blobStatus)
+	c.blobRefs.M = make(map[restic.BlobHandle]blobStatus)
 
 	return c
 }
@@ -163,7 +162,8 @@ func (c *Checker) LoadIndex(ctx context.Context) (hints []error, errs []error) {
 			cnt := 0
 			for blob := range res.Index.Each(ctx) {
 				c.packs.Insert(blob.PackID)
-				c.blobRefs.M[blob.ID] = blobExists
+				h := restic.BlobHandle{ID: blob.ID, Type: blob.Type}
+				c.blobRefs.M[h] = blobStatusExists
 				cnt++
 
 				if _, ok := packToIndex[blob.PackID]; !ok {
@@ -500,9 +500,10 @@ func (c *Checker) filterTrees(ctx context.Context, backlog restic.IDs, loaderCha
 			// use a separate flag for processed trees to ensure that check still processes trees
 			// even when a file references a tree blob
 			c.blobRefs.Lock()
-			blobFlags := c.blobRefs.M[nextTreeID]
+			h := restic.BlobHandle{ID: nextTreeID, Type: restic.TreeBlob}
+			status := c.blobRefs.M[h]
 			c.blobRefs.Unlock()
-			if (blobFlags & treeProcessed) != 0 {
+			if (status & blobStatusReferenced) != 0 {
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


internal/checker/checker.go
```diff
@@ -498,15 +487,21 @@ func filterTrees(ctx context.Context, backlog restic.IDs, loaderChan chan<- rest
 		job                     treeJob
 		nextTreeID              restic.ID
 		outstandingLoadTreeJobs = 0
+		processedTrees          = make(map[restic.ID]bool)
```
MichaelEischer: The `processedTrees` map will increase the memory usage of the check command as it is stored in addition to `blobrefs.M`. It would be possible to add `processedTrees` to `c.blobrefs` and use it exclusively to store the tree ids.  The main drawback of this would be that `UnusedBlobs` has to look into both maps.

ad-m:
Have you collected any stats regarding performance impact?

MichaelEischer:
I've noted that I forgot to mention the speedup seen for the 40 million files test set: It went from one week down to approximately one day.

And now on to something more comparable:

Preface: Github seems to mess up the displayed commit ordering, the actual order is as follows (oldest to newest). The benchmark just copies an existing snapshot and adds it with a new set of tags. I've mixed the commit list with the observed changes, the measurement data is shown below.
```
5b94a9a8 Add benchmark for checker scaling with snapshot count
<-- baseline
ced4890f checker: Decode identical tree nodes only once
<-- nearly no more slowdown with additional snapshot copies
ca3f828a checker: Reduce cost of debug log
10e2aabf checker: Remove dead index map
bf27fbf3 checker: Traverse trees in depth-first order
189f44a8 checker: Remove duplicate sets used to track existing blobs
babac588 checker: Unify processed tree and referenced blobs map
<-- slightly faster and fewer allocations / memory usage
88776b5e repository: Don't sort one element pack lists
0003b64b repository: Remove size lookup when loading blob/tree
<-- slightly faster
```

```
> git rebase -i HEAD~9 -x 'git log --oneline -n 1 ; ../go/bin/go test -bench BenchmarkChecker -benchtime 500x -benchmem github.com/restic/restic/internal/checker'
Executing: git log --oneline -n 1 ; ../go/bin/go test -bench BenchmarkChecker -benchtime 500x -benchmem github.com/restic/restic/internal/checker
5b94a9a8 (HEAD) Add benchmark for checker scaling with snapshot count
goos: linux
goarch: amd64
pkg: github.com/restic/restic/internal/checker
BenchmarkChecker-8                      	     500	  37276318 ns/op	 4208978 B/op	    2348 allocs/op
BenchmarkCheckerSnapshotScaling_50-8    	     500	  40370646 ns/op	 5584485 B/op	   20994 allocs/op
BenchmarkCheckerSnapshotScaling_100-8   	     500	  43772751 ns/op	 6960990 B/op	   39647 allocs/op
BenchmarkCheckerSnapshotScaling_200-8   	     500	  51724338 ns/op	 9718810 B/op	   76952 allocs/op
PASS
ok  	github.com/restic/restic/internal/checker	94.344s

ced4890f (HEAD) checker: Decode identical tree nodes only once
BenchmarkChecker-8                      	     500	  36961886 ns/op	 4208199 B/op	    2296 allocs/op
BenchmarkCheckerSnapshotScaling_50-8    	     500	  37743232 ns/op	 4511634 B/op	    4890 allocs/op
BenchmarkCheckerSnapshotScaling_100-8   	     500	  39472519 ns/op	 4818885 B/op	    7490 allocs/op
BenchmarkCheckerSnapshotScaling_200-8   	     500	  39081652 ns/op	 5433777 B/op	   12694 allocs/op
PASS
ok  	github.com/restic/restic/internal/checker	84.363s

babac588 (HEAD) checker: Unify processed tree and referenced blobs map
BenchmarkChecker-8                      	     500	  36746626 ns/op	 4126244 B/op	    1089 allocs/op
BenchmarkCheckerSnapshotScaling_50-8    	     500	  36839477 ns/op	 4430195 B/op	    3695 allocs/op
BenchmarkCheckerSnapshotScaling_100-8   	     500	  37969459 ns/op	 4737592 B/op	    6298 allocs/op
BenchmarkCheckerSnapshotScaling_200-8   	     500	  39273562 ns/op	 5352415 B/op	   11500 allocs/op
PASS
ok  	github.com/restic/restic/internal/checker	83.186s

0003b64b (HEAD, origin/no-repeated-checks, no-repeated-checks) repository: Remove size lookup when loading blob/tree
BenchmarkChecker-8                      	     500	  36449458 ns/op	 4126226 B/op	    1089 allocs/op
BenchmarkCheckerSnapshotScaling_50-8    	     500	  36914002 ns/op	 4430199 B/op	    3695 allocs/op
BenchmarkCheckerSnapshotScaling_100-8   	     500	  36549722 ns/op	 4737589 B/op	    6298 allocs/op
BenchmarkCheckerSnapshotScaling_200-8   	     500	  37407083 ns/op	 5352403 B/op	   11500 allocs/op
PASS
ok  	github.com/restic/restic/internal/checker	81.400s
```

Thus from my point of view this leaves primarily the following commits without a demonstrable speed-up.
``` 
bf27fbf3 checker: Traverse trees in depth-first order
88776b5e repository: Don't sort one element pack lists
```
These commits improve data locality or avoid some unnecessary syscalls, respectively. As the accessed files come from the page cache there's not much that can be shown.

MichaelEischer:
I've split off the repository related changes into separate PRs. Apart from the commits "checker: Remove duplicate sets used to track existing blobs" and "checker: Unify processed tree and referenced blobs map" which are now merged into a single commit, the PR is largely unchanged, that is the old performance measurements still apply.

The commit `checker: Traverse trees in depth-first order` is currently still part of this PR, but could be split off as it has a slightly different focus than the remainder of the PR.

rawtaz:
Interesting work! Based on what you've seen in other PRs, do you think this PR makes changes that conflict with other PRs? Just in general/off the top of your head, I'm not suggesting you sift through them all.

MichaelEischer:
The only conflict I'm aware of is with #2523 which also removes some IDSets from the checker code. The solution is rather easy in that case: just use the code from this PR as it reduces the memory usage even a bit further.

rawtaz:
I just tested this myself on a repo that's 101 GB on disk and contains 4437 snapshots.

Running `restic check` on it (without `--read-data`) with master (`restic 0.9.6 (v0.9.6-104-g58bd1652) compiled with go1.13.5 on linux/amd64`) took *145 minutes* (`real 145m7.542s, user 520m36.152s, sys 9m24.677s`).

Running it with this PR (`restic 0.9.6 (v0.9.6-146-g3f2e8b1b) compiled with go1.14 on linux/amd64`) took just *7 minutes* (`real 7m3.783s, user 24m46.927s, sys 0m5.213s`).

That's a huge improvement. I also noticed that restic was using around 395-399% CPU with this PR instead of 380% with master.

The system is a small Dell server with 7200 RPM disks on hardware RAID, but clearly that's not a relevant bottleneck in this case. Memory is 8 GB and the CPU has four cores.

The generated output from the two runs were identical in everything that matters (a bunch of unreferenced packs and one missing pack in this case).

rawtaz:
Another run with a 90+ GB repo, took 4 minutes instead of 31. Same machine, 1767 snapshots.

rawtaz:
@greatroar I cannot request a review from you using GitHub, but I'd like to ask if you could review this PR nonetheless?

internal/checker/checker.go
```diff
@@ -22,29 +22,33 @@ import (
...
 	masterIndex *repository.MasterIndex
 
 	repo restic.Repository
 }
 
+type blobBits uint8
+
+const (
+	existsFlag blobBits = 1 << iota
+	referencedFlag
```
greatroar, line 40: I would call the type blobStatus, or just status, and the values statusExists and statusReferenced. The stdlib nearly always has its enum values start with the type name.

internal/checker/checker.go
```diff
@@ -504,8 +493,21 @@ func filterTrees(ctx context.Context, backlog restic.IDs, loaderChan chan<- rest
 
 	for {
 		if loadCh == nil && len(backlog) > 0 {
+			// process last added ids first, that is traverse the tree in depth-first order
```
greatroar, line 496: This is not depth-first order. It's more like post-order, if I understand the surrouding code correctly.

MichaelEischer: If the checker only uses a single treeLoader then the traversal order is a pre-order depth-first search. This line dequeues a child subtree and sends it to the tree loader. Once the tree is loaded the subtrees are added to the backlog and the tree node is sent via the outCh to the error checking. The child nodes are then loaded next from the backlog end and so forth.

With multiple treeLoader it gets a bit more chaotic but the overall pattern still holds. The archiver code uses a post-order depth-first traversal so the traversal order in the checker is still not a perfect fit, but at least it resembles the archiving process much more closely than the breath-first search that was used previously.

internal/checker/checker.go
MichaelEischer: Thanks for the suggestion, I've change the type name to `blobStatus` and change the value names accordingly.

fd0 (approved):
LGTM!

PR #2336, opened 2019-07-13
Disable scanner for quiet backup

MichaelEischer:

The scanner's only purpose is to provide totals for file and directory counts and size. These are used to provide a sensible progress bar for the user. When running restic in quiet mode, no output is produced and thus there is no need to run the scanner. This saves the processing and syscalls required to walk over potentially millions of files.

No

Checklist
---------

- [x] Already covered by 'cmd/restic/integration_test.TestQuietBackup'. I have added tests for all changes in this PR

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -592,16 +592,18 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		targets = []string{filename}
 	}
 
-	sc := archiver.NewScanner(targetFS)
-	sc.SelectByName = selectByNameFilter
-	sc.Select = selectFilter
-	sc.Error = p.ScannerError
-	sc.Result = p.ReportTotal
+	if !gopts.Quiet {
+		sc := archiver.NewScanner(targetFS)
+		sc.SelectByName = selectByNameFilter
+		sc.Select = selectFilter
+		sc.Error = p.ScannerError
+		sc.Result = p.ReportTotal
 
-	if !gopts.JSON {
-		p.V("start scan on %v", targets)
+		if !gopts.JSON {
+			p.V("start scan on %v", targets)
+		}
+		t.Go(func() error { return sc.Scan(t.Context(gopts.ctx), targets) })
 	}
-	t.Go(func() error { return sc.Scan(t.Context(gopts.ctx), targets) })
 
 	arch := archiver.New(repo, targetFS, archiver.Options{})
 	arch.SelectByName = selectByNameFilter
--- a/changelog/unreleased/pull-2336
+++ b/changelog/unreleased/pull-2336
@@ -1,0 +1,8 @@
+Enhancement: Faster backups in quiet mode
+
+Restic calculated the total file/directory count and size which, however,
+were not displayed when running with the quiet flag set. This speeds up
+backups as the file tree is only walked once.
+
+https://github.com/restic/restic/pull/2336
+


ad-m:
Should we document performance-penalty for non-quiet output?

fd0:
Thanks for your contribution, it looks good so far! Do you have any measurements that support the "This speeds up backups as the file tree is only walked once." statement in the changelog?

I've measure this once and observed that it does not affect performance at all. I suspected this was the case because Linux caches direntry structs in memory, so the directory structure isn't read when the archiver comes around to traversing the dirs...

cmd/restic/cmd_backup.go
```diff
@@ -534,16 +534,18 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		targets = []string{filename}
 	}
 
-	sc := archiver.NewScanner(targetFS)
-	sc.SelectByName = selectByNameFilter
-	sc.Select = selectFilter
-	sc.Error = p.ScannerError
-	sc.Result = p.ReportTotal
+	if !quiet {
```
ad-m: Why not use ```gopts.Quiet``` and remove changes in definition of ```runBackup```?

cmd/restic/cmd_backup.go
```diff
@@ -60,7 +60,7 @@ given as the arguments.
 		term := termstatus.New(globalOptions.stdout, globalOptions.stderr, globalOptions.Quiet)
 		t.Go(func() error { term.Run(t.Context(globalOptions.ctx)); return nil })
 
-		err := runBackup(backupOptions, globalOptions, term, args)
+		err := runBackup(backupOptions, globalOptions, term, args, globalOptions.Quiet)
```
ad-m: It seems to me that an additional parameter is unnecessary, and the value can be read from other parameters that are passed to the function already (```runBackup((...), globalOptions, (...), globalOptions.Quiet)```).

ad-m (changes requested):

cmd/restic/cmd_backup.go
MichaelEischer: Oh, good catch, I've somehow missed that gopts is available inside the runBackup function. I've updated to commits accordingly.

cmd/restic/cmd_backup.go
MichaelEischer: see above

MichaelEischer:
I've run a few tests and the results are mixed, that is _disabling the scanner for quiet backup_ is faster in some cases but also slower in others. That is at least the changelog entry will need some rewording. If I remember correctly I've done the initial performance tests using `GOMAXPROCS=1` in which case the pull request consistently improves performance.

## Test setup
My test dataset had 1 million small files with approx. 4GB total file size. Go version is 1.13.5. As the pull request removes a filesystem walk, the tests focus on stressing that code. This is done by running an initial (not measured) backup of the dataset and then repeating the backup runs with zero modified files.
There are four variants of the tests: Either with or without the quiet flag and with and without flushing the Linux page cache between the individual test runs. The results were stable across three runs, so that I've just included the second run of each variant. The variants with and without the quiet flag were run alternately. For more details on the setup see further below.

## Results
### Ubuntu Bionic with HDD

setting | real | user | sys
----|---|---|---
backup-drop_caches | 0m45.246s | 0m51.675s | 0m10.868s
backup-quiet-drop_caches | 0m52.563s | 0m46.188s | 0m8.424s
backup | 0m37.319s | 0m54.581s | 0m8.992s
backup-quiet | 0m34.841s | 0m49.361s | 0m6.737s

### Other Ubuntu Bionic with SSD

setting | real | user | sys
----|---|---|---
backup-drop_caches | 0m34.818s | 0m49.102s | 0m10.448s
backup-quiet-drop_caches | 0m37.827s | 0m45.967s | 0m8.542s
backup | 0m33.111s | 0m47.395s | 0m7.872s
backup-quiet | 0m31.338s | 0m43.758s | 0m5.795s

### Other Ubuntu Bionic with SSD, GOMAXPROCS=1

setting | real | user | sys
----|---|---|---
backup-drop_caches | 0m49.233s | 0m38.592s | 0m7.357s
backup-quiet-drop_caches | 0m45.926s | 0m36.207s | 0m6.638s
backup | 0m43.461s | 0m37.384s | 0m5.854s
backup-quiet | 0m40.598s | 0m35.842s | 0m4.524s

### macOS Catalina with SSD
No drop_caches as I have no idea what would be the macOS equivalent.

setting | real | user | sys
----|---|---|---
backup | 2m0.370s | 1m28.464s | 1m55.670s
backup-quiet | 1m48.782s | 1m4.006s | 1m4.599s

## Analysis
Using the quiet flag in all cases leads to a noticeable decrease of _user_ and _sys_ time. However, the real time (that is total execution time) shows mixed results: For macOS the quiet flag saves 10 seconds, whereas the runtime is slightly higher on `Ubuntu Bionic with SSD` and quite bit higher for `Ubuntu Bionic with HDD`. As the latter effects don't show up when the page cache is not cleared, I assume that the scanner thread actually serves as a filesystem data prefetcher. When using GOMAXPROCS=1 the variant with the quiet flag is also faster in all cases, which also backs up the prefetch theory.

## Test scripts
create-test-data.py
```python
#!/usr/bin/env python3

import random
import os

random.seed(42)

def recurse(base_path: str, more_depth: int):
    for i in range(100):
        fn = "d{}".format(i)
        path = os.path.join(base_path, fn)
        if more_depth > 1:
            os.mkdir(path)
            recurse(path, more_depth - 1)
        else:
            with open(path, "wb") as f:
                length = max(1, int(random.expovariate(1.0/4000)))
                data = random.getrandbits(length*8).to_bytes(length, "little")
                f.write(data)

os.mkdir("test")
recurse("test", 3)
```
```bash
echo "some-super-secret-password" > restickey
(cd restic && ../go/bin/go build -mod=vendor -tags profile ./cmd/restic)
```

run-test.sh
```bash
#!/bin/bash
set -x

restic=restic/restic
repo=test-repo
mkdir -p test-cache

flushc() {
    sync; echo 3 | sudo tee /proc/sys/vm/drop_caches
}

$restic -r $repo -p restickey init
$restic -r $repo -p restickey --cache-dir test-cache backup test

for i in {1..3}; do
    flushc
    time $restic -r $repo -p restickey --cache-dir test-cache backup test
    flushc
    time $restic -r $repo -p restickey --cache-dir test-cache backup -q test
done
for i in {1..3}; do
    time $restic -r $repo -p restickey --cache-dir test-cache backup test
    time $restic -r $repo -p restickey --cache-dir test-cache backup -q test
done
```

MichaelEischer:
I've converted this to a draft, as I currently don't have an idea how to both reduce the CPU time and not hurt the backup performance.

MichaelEischer:
I'm closing this PR as it's probably not worth the trouble figure out how to solve the performance problems.

PR #2337, opened 2019-07-13
docs: switch to gopher favicon

rfjakob:

The favicon on restic.readthedocs.io still contained the old
superman-style logo.

This changes it to the 32x32 icon also used on https://forum.restic.net/ ,
just converted to .ico using Gimp.

Yes, https://forum.restic.net/t/new-logo-for-restic/394

Checklist
---------


moritzdietz:
Thanks for this PR! Quick question: Did you convert the `.ico` file yourself or did you copy the one used for the forum?
I am not sure if this is actually worth of mentioning in a changelog entry as it does not affect restic itself. It doesn't hurt but I am also not sure if it's actually needed/wanted by the maintainers.

rfjakob:
The forum uses a png, i took the png and converted to ico with gimp.

changelog/unreleased/pull-2337
```diff
@@ -0,0 +1,9 @@
+Documentation: switch to gopher favicon
+    
+The favicon on restic.readthedocs.io still contained the old
```
moritzdietz:
```suggestion
The favicon used on restic.readthedocs.io was still the old
```

changelog/unreleased/pull-2337
```diff
@@ -0,0 +1,9 @@
...
+superman-style logo.
+
+We changed it to the 32x32 gopher icon also used on https://forum.restic.net/ ,
```
moritzdietz:
```suggestion
We changed it to the 32x32 gopher icon also used on https://forum.restic.net/
```

changelog/unreleased/pull-2337
```diff
@@ -0,0 +1,9 @@
...
+just converted to .ico using Gimp.
```
moritzdietz:
```suggestion
```

moritzdietz (commented):
So if the maintainers decide to keep this CHANGELOG entry, I would word it differently.
See my suggested change. I don't think it is relevant how you modified the `.png` to get the favicon.

rawtaz (changes requested):
Wow, there certainly seems to be *a lot* to take into account with favicon, if you read https://stackoverflow.com/a/23734416. For constructive reasons I think we can just replace the file for now, if there's trouble with it down the road, let's address it then.

@rfjakob I agree with @moritzdietz that for this type of change, an entry in the changelog adds little value. Can you please remove those changes (keeping only the new favicon file), and amend/squash so there's just one commit? Let us know if you need help 👍 

Thanks!

rawtaz (approved):
I removed the changelog file (not needed for documentation changes), thank you very much for this nice little piece of contribution! :)

Issue #1988, opened 2018-09-07
Restic backup — OOM-killed on raspberry pi after backing up another computer to same repo


hmage:
On a raspberry pi 3 model B, which has 1GB of memory, restic stopped fitting into RAM after a 3TB backup was done on another machine to same repository. I suspect indexes have grown big enough to not fit into memory.

## Output of `restic version`
```
hmage@phmd:~$ restic version
restic 0.9.2 compiled with go1.10.3 on linux/arm
```

## How did you run restic exactly?
```bash
export AWS_ACCESS_KEY_ID=secret
export AWS_SECRET_ACCESS_KEY=secret
export RESTIC_PASSWORD=secret
export RESTIC_REPOSITORY=s3:https://s3.wasabisys.com/restic-hmage
source ./restic-excludes.sh
restic backup --exclude-file <(printf "%s\n" "${EXCLUDES[@]}") /`
```

Contents of `restic-excludes.sh`:
```bash
EXCLUDES=(
/dev
/proc
/run
/sys
)
EXCLUDES+=(
$'Icon\r'
$HOME/.bundle
$HOME/.cache
$HOME/.cargo
$HOME/.ccache*
$HOME/.config/chromium
$HOME/.cpan
$HOME/.dropbox
$HOME/.local/share/akonadi
$HOME/.npm
$HOME/Library/Application\ Support/Google/Chrome
$HOME/Library/Application\ Support/Telegram\ Desktop
$HOME/Library/Arq/Cache.noindex
$HOME/norm.*
**/var/cache/apt
**/var/cache/man
**/var/lib/apt/lists
**/var/lib/mlocate
.DS_Store
.DocumentRevisions-V100
.Spotlight-V100
.Trashes
.bzvol
.cache
.dropbox.cache
.fseventsd
/Volumes/Time\ Machine
/media/psf
/private/var/vm/
/srv/piwik/tmp
/srv/www/data/cache
/tmp
/usr/lib/debug
/var/lib/lxcfs
/var/swap
/var/tmp
Cache
Caches
)
```

## What backend/server/service did you use to store the repository?
Wasabi (S3 protocol)

## Expected behavior
Restic should not run out of memory no matter how big the indexes are — they should be streamed from disk/repo rather than loaded completely into RAM since it's not infinite.

Actual behavior
---------------
[...]

Issue #1723, opened 2018-04-16
restic prune: out of memory


johndoe31415:
## Output of `restic version`
restic 0.8.3 (v0.8.3-0-g272ccec7)
compiled with go1.10 on linux/amd64

## How did you run restic exactly?
RESTIC_PASSWORD=foobar
RESTIC_REPOSITORY=/data/joe/restic
restic prune

## What backend/server/service did you use to store the repository?
Direct file access.

## Expected behavior
Restic prunes the repo.

## Actual behavior
```
2018-04-16 18:33:51: Starting restic prune
counting files in repo
building new index for repo
[11:59] 100.00%  291017 / 291017 packs

incomplete pack file (will be removed): 02b612398378d4b6d7b6a77b5f51cf3c541baa78f542886810c27488fc52f8a2
incomplete pack file (will be removed): 08ce4b79b2adacd1d85373c59ad28777e215cec52b16bf9963cb03106e1d547f
incomplete pack file (will be removed): 08d76e1953f8b06d936f46679986f0af9bfb0260ca1b81e641d30262f4e2fb69
incomplete pack file (will be removed): 0ada7a6774140391104de82da908a1027eaedc97be83a2c30ae26db85093aba1
incomplete pack file (will be removed): 0f960d42718f0e0b29ea10a47fef530638811fd1620843e13e4c4c3c139c9477
incomplete pack file (will be removed): 11aa5b87e3a44ff1678de573e21280ff56e8cdf93755c363c7ba1a03537b84b0
incomplete pack file (will be removed): 13b211c7fc26e7de47e8a5387749bb93f7737217809143e01caf7f826deeea13
incomplete pack file (will be removed): 13f00bb9960f95b0d71abc9ff5e5b86f53d370cd602a4c45e03505cefab328ab
incomplete pack file (will be removed): 1b0be1e761e0c369aaed7b2d03f07163854946dfe7b6bb749d4ff5d5c62cfd61
incomplete pack file (will be removed): 1ee341e8778661a76bc0f4d2a285457851cb5984f6ac6ee51d42bbdf705025ab
incomplete pack file (will be removed): 26de4ab07458e880217accac26098edcc46622d11d20dcbb57996cfebedbc859
incomplete pack file (will be removed): 274ede1a69f1f2634abeb6f7f7a9af2a4027a7f224d7a61a6e209ef8ee4c152f
[...]

PR #2338, opened 2019-07-13
Optimize in-memory index memory consumption

MichaelEischer:
Slightly reduce the memory usage of the in-memory index that is used by most restic commands. The memory usage should be 10-15% lower than before. This is achieved by optimizing for the usual situation where each blob is contained in a single pack file.

The pull request still needs tests which trigger the code paths with multiple packs. I'm not sure whether there are currently any tests that explicitly generate blobs which are stored in multiple packs.

The problem of high memory usage shows up in several issues #2284 , #1988 , #1723 . However, the approach of this pull request was not yet discussed in an issue. The underlying scalability issue is not tackled by this pull request.

Checklist
---------

- [x] No user visible parts. I have added documentation for the changes (in the manual)

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -109,6 +109,18 @@ func (idx *Index) Store(blob restic.PackedBlob) {
 	idx.store(blob)
 }
 
+func toPacketBlob(h restic.BlobHandle, p indexEntry) restic.PackedBlob {
+	return restic.PackedBlob{
+		Blob: restic.Blob{
+			Type:   h.Type,
+			Length: p.length,
+			ID:     h.ID,
+			Offset: p.offset,
+		},
+		PackID: p.packID,
+	}
+}
+
 // Lookup queries the index for the blob ID and returns a restic.PackedBlob.
 func (idx *Index) Lookup(id restic.ID, tpe restic.BlobType) (blobs []restic.PackedBlob, found bool) {
 	idx.m.Lock()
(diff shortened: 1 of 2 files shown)


aawsome:
Thanks @MichaelEischer for this proposal! I agree that we need a solution to the memory consumption problem and you correctly have identified the index data structure as the core point to improve :+1: 
This is also what I tried to discuss in #1988 and #2523. 

About your solution I'm not confident that 10-15% improvement is enough. I like the idea of not storing duplicate information of blobs being in the index more than once (and this situation should anyway be cleaned by the next `prune` run or its alternatives, see #2513).
I would however prefer a solution where duplicates are not even saved in the in-memory index data structure. I did hardly find any code where the `[]PackedBlob` in `Lookup` was really needed (I think it was the `find` command) and would propose to change the interface to `Lookup(ID, BlobType) (*PackedBlob, bool)`. Then your PR would become much simpler while still achiving the same savings.

But I think there are even better solutions as the PackID for packs containing many blobs is still multiple times saved in memory and there is a notable overhead of all the maps which I think can be removed without loosing too much performance. Please have a look at `index_lowmem2.go` in [index-alternatives](https://github.com/aawsome/restic/tree/index-alternatives). This is however still WIP and needs more testing.

Maybe some of the restic maintainers can give directions where to work on...

MichaelEischer:
The `[]PackedBlob` array is also used in Repository.loadBlob to try to get the blob from an already cached pack. It could also be useful, when trying to fix a repository with damaged files. However, the simplified interface would make the code a lot simpler.

@aawsome : Yes, I know that 10-15% improvement is not that much. The PackIDs can definitely be stored separately which should improve memory usage by quite a lot. However, as the pull request is already several months old I can't remember why I've ignored that optimization.

MichaelEischer:
Closing as this pull request is superseded by either https://github.com/aawsome/restic/tree/refactor-index or https://github.com/MichaelEischer/restic/tree/opt-hash-idx

tcurdt:
@MichaelEischer are there PRs for those to follow?

aawsome:
@tcurdt For the `refactor-index` branch see #2523 for the discussion.
And yes a PR will follow...

PR #2350, opened 2019-07-23
Add AWS_REGION env var to specify s3 region

viewpointsa:

With the scaleway.com S3 compatible service I have an error:
`Fatal: unable to open config file: Stat: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'fr-par'`
So to override default region, I add AWS_REGION environment variable, is the good way/name ?

I not a go developer, please check that I do


altaris:
I am also in need to this fix. Using `rclone` as a backend did not work in my case...

internal/backend/s3/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 
 	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
 	MaxRetries  uint `option:"retries" help:"set the number of retries attempted"`
+	Region	     string
```
costela: You have a TAB character in there, causing the `go fmt` test to fail. Please run your file through `go fmt`.

costela (changes requested):

costela (changes requested):
hey @viewpointsa, thanks for the PR!
Even though I'm not a committer on restic, I took the liberty of suggesting a minor improvement to your PR. I hope this increases its chance of being merged!

viewpointsa:
> 
> 
> I am also in need to this fix. Using `rclone` as a backend did not work in my case...

@altaris I'm not a github expert but my PR seem to wait someone to review it. Can you do it ?

costela:
@viewpointsa unfortunately neither @altaris nor myself are contributors to this project, so I'm not sure how much our reviews would help.
Guess we have to wait for some of the [members](https://github.com/orgs/restic/people) to have some free time :wink:

jileventreur (approved):

douglascrp:
Hello.

I am also trying to use restic with the scaleway s3 compatible service, but it is failing by the same reason.

Could someone please review and merge this PR?

cfbao:
I believe it's recommended to use a rclone backend for new repositories. The other backends still exist mostly for backward compatibility reason.

I just tested Scaleway's storage with restic+rclone and it works without problem.

viewpointsa:
> I believe it's recommended to use a rclone backend for new repositories. The other backends still exist mostly for backward compatibility reason.
> 
> I just tested Scaleway's storage with restic+rclone and it works without problem.

This patch work like a charm with Scaleway, since 4 months. I don't think software stacking brings anything.

cfbao:
> This patch work like a charm with Scaleway, since 4 months.

I'm sure it works for you. But this issue can already be solved with rclone without making changes to restic.

> I don't think software stacking brings anything.

It brings rclone's much more mature and extensive support of many types of remote storage to restic without restic maintainers spending valuable time re-inventing and maintaining the "wheel". This time could be much better spent on other features and fixes that can only be handled in restic.

kudos:
@cfbao Restic already supports S3, and S3 compatible services like Minio. I can't believe you're pushing back on what is a tiny PR. Until the Restic team indicate that everthing should go through rclone, PRs like this should be getting merged.

rawtaz:
@kudos Thanks for trying to help improve and fix problems! Please note that it's not a core developer that's suggesting rclone vs restic in this specific case.

I wonder; Should this perhaps be an option (see `restic help init` instead of an environment variable? There are other options already, and this seems like something that is specific enough to perhaps warrant being in one of those instead of an env var. What do you guys think?

Please also note we need to document this change in the documentation (and changelog) Besides the restic manual, it might be needing a documentation like the others in internal/backend/s3/config.go if it becomes an option.

cfbao:
> @cfbao Restic already supports S3, and S3 compatible services like Minio. I can't believe you're pushing back on what is a tiny PR. Until the Restic team indicate that everthing should go through rclone, PRs like this should be getting merged.

My initial comment was merely meant as a helpful tip for people waiting for this PR - "no need to wait, starting using rclone and you're good to go".

My slightly more strongly worded second comment was in response to @viewpointsa's surprisingly negative and inaccurate comment on using rclone.

internal/backend/s3/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 
 	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
 	MaxRetries  uint `option:"retries" help:"set the number of retries attempted"`
+	Region      string
```
fd0: Please add an option tag as in the previous options:

fd0:
```suggestion
	Region      string `option:"region" help:"set region"`
```

fd0 (changes requested):
This only needs a minor fix and a changelog file, then it's good to go.

fd0:
Huh, wrong button, sorry.

fd0:
I've created #2484 which contains all of this including my suggestions and documentation. Sorry for closing this PR.

douglascrp:
Awesome... I am going to try this ASAP.
Thank you

rrossouw01:
Thanks! Appears like this is fixed in Oracle OCI (Oracle Cloud Infrastructure) object storage when using S3 compatibility.  I will be able to confirm better soon but a quick initial try worked.

fd0:
FYI, I've just released restic 0.9.6 which also includes this fix.

viewpointsa:
> FYI, I've just released restic 0.9.6 which also includes this fix.

Thanks for your time. This week I had an attack by a Ransomware, Restic worked very well to recover my files with the history. So a have test restic_0.9.6_windows_amd64.exe work with AWS_REGION.

fd0:
Cool! thanks for the feedback :)

PR #2354, opened 2019-07-26
Doc Addition: Wasabi repo setup instructions

HugoReeves:

This is an addition to the 'preparing a new repo' documentation that adds specific instructions for the AWS S3 conformant object storage provider [Wasabi](https://wasabi.com). **NOTE**: I am not affiliated with Wasabi and wrote this as a service to restic not to Wasabi. This PR does not necessarily change anything, it simply adds more repo specific information to the documentation. Most of the content for this documentation addition was taken from the [Wasabi support topic on restic backups](https://wasabi-support.zendesk.com/hc/en-us/articles/115002240372-How-do-I-use-Restic-with-Wasabi-).

Not to my knowledge. I have seen some discussion of Wasabi across the restic forum and I believe Wasabi is being used by many restic users, hence why I thought specific documentation would be useful.

Checklist
---------
**PR Author Note**: These have been checked for clarity however this PR only changes one documentation file.

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -234,7 +234,7 @@ credentials of your Minio Server.
     $ export AWS_ACCESS_KEY_ID=<YOUR-MINIO-ACCESS-KEY-ID>
     $ export AWS_SECRET_ACCESS_KEY= <YOUR-MINIO-SECRET-ACCESS-KEY>
 
-Now you can easily initialize restic to use Minio server as backend with
+Now you can easily initialize restic to use Minio server as a backend with
 this command.
 
 .. code-block:: console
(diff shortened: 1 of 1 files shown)


moritzdietz (approved):
LGTM.

rawtaz:
I started looking over this and had a few small adjustments to correct, but then I realized that I'm not sure we should add this provider-specific info to the docs. This service seems to be S3 compatible, and looking through the instructions they're essentially the same as for Amazon S3, just different URLs.

So, if we add this, is it actually useful, and won't we have to add a lot of other specific docs as well? It can become a never ending spiral. Perhaps instead we should keep the focus of this part of the doc on how to init a repo for the individual backends, which I think has been the purpose of the article in the first place. That is, whether an S3 server is served by company A, B or C shouldn't really matter, as long as the docs show how to configure *such a service* or *that backend/protocol*.

Thoughts? As far as *I* am concerned, it's up to @fd0. If he thinks it should be merged, I'll go ahead with a couple of small change requests before we merge.

HugoReeves:
@rawtaz This was also something I considered before authoring. I am perfectly happy if this PR is rejected and a decision is made to avoid including provider specific backend details.

I noted, before authoring these changes, that minio information is included in the docs; this make sense to me because it is a self hosted alternative to S3 whereas Wasabi is a commercial alternative. Another option would be creating an entirely new section of documentation, maybe a different subdomain, that offers very specific guides and tutorials for setting up different backends. This would present a lot more work and may be out of scope of what is desired for the restic project.
Thanks

moritzdietz:
I think adding this would not really overcrowd the docs. It would probably make it "easier" for new users to see if their platform of choice is supported right away. Like rclone does it.

fd0:
I like the proposal to split the docs into a section for each backend (local, sftp, s3, b2), and a section describing the setup for various cloud providers (AWS, Wasabi etc).

rawtaz (approved):
Looks good, we'll merge this now and probably move it to a separate section later.

fd0:
Ok, thanks!

HugoReeves:
Thanks for merging this. As a sidenote, I have been using restic frequently for 4 months now and have been impressed with the experience.

@fd0 If I have time, would a draft PR implementing this split documentation approach be appreciated?
I'm just thinking a separate sidebar section in the readthedocs - among "talks", "faq", "manual" - called "Provider specific backend instructions".

rawtaz:
@HugoReeves Please continue the discussion in the active issue #2490 - this PR is closed 🌶

PR #2373, opened 2019-08-13
Ignore username difference in TestMetadataChanged

### LINKED ISSUE #2372: TestMetadataChanged fails on systems with UID 51234 taken by a user
Output of `restic version`
--------------------------

$ ./restic version
restic 0.9.5-dev (compiled manually) compiled with go1.12.5 on linux/amd64

`go test ./...` in a fresh git checkout

n/a

Expected behavior
-----------------

tests pass

Actual behavior
---------------

```
--- FAIL: TestMetadataChanged (0.01s)
    archiver_test.go:2043: chdir to /tmp/restic-test-488507700
    archiver_test.go:2091: metadata does not match:
        {*restic.Node}.User:
        	-: ""
        	+: "<username>"
    panic.go:406: chdir back to /<...>/restic/internal/archiver
```

Run tests on a system where UID 51234 is defined and occupied by a user.

$ id 51234 > /dev/null && echo "UID exists"
UID exists

Yes, this part of the test:

```
	if runtime.GOOS != "windows" {
		want.UID = 51234
		want.GID = 51235
	}
	// no user and group name
	want.User = ""
	want.Group = ""

	// make another snapshot
	snapshotID, node3 := snapshot(t, repo, fs, snapshotID, "testfile")

	// make sure that metadata was recorded successfully
	if !cmp.Equal(want, node3) {
		t.Fatalf("metadata does not match:\n%v", cmp.Diff(want, node3))
	}
```

One of:
- don't check for username at all (check for UID)
- pick UID which is guaranteed to be unoccupied (negative? will this work?)

yup :)

vrusinov:
In some (rare) cases "fake" UID 51234 may exist in a system running a
test. When this is the case, `cmp.Equal(want, node3)` will fail based on
difference between empty string and an actual username present in a
system.

Closes #2372.

To fix github issue #2372

No, consider this to be a proposal / start discussion. I did not put all my heart into this fix and not attached to it. Happy to discuss. :)

Checklist
---------

### DIFF
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -2085,6 +2085,10 @@ func TestMetadataChanged(t *testing.T) {
 
 	// make another snapshot
 	snapshotID, node3 := snapshot(t, repo, fs, snapshotID, "testfile")
+	// Override username and group to empty string - in case underlying system has user with UID 51234
+	// See https://github.com/restic/restic/issues/2372
+	node3.User = ""
+	node3.Group = ""
 
 	// make sure that metadata was recorded successfully
 	if !cmp.Equal(want, node3) {


fd0 (approved):
Thank you very much for improving restic. Your fix looks good to me :)

Issue #79, opened 2015-01-26
archive and restore sparse files


fd0:
description from the gnu tar manual: https://www.gnu.org/software/tar/manual/html_node/sparse.html

fd0:
gnu tar manual: https://www.gnu.org/software/tar/manual/html_node/sparse.html

jannic:
At least on linux, the description in the tar manual is outdated, the part "in order to determine if the file is sparse tar has to read it before trying to archive it" isn't true any more. The lseek() option SEEK_HOLE allows to find holes without reading the whole file.

Using this from golang would involve some low-level non-portable code, but I guess for files with very large holes it could be an important optimization.

fd0:
Thanks for the hint. I have some more optimizations in mind that may only work on Linux. In general, I think that's okay to implement, as long as the other platforms are not degraded in functionality.

andrewchambers:
Long runs of zeros will be compressed perfectly anyway.

jannic:
Correct, detecting holes doesn't help reducing the backup size. It can still improve performance, as reading large amounts of zeroes and then compressing them takes longer than just skipping over them.
But what's more important: When restoring files which had large holes, it's very useful to actually recreate those holes instead of writing zeroes. It obviously saves space, and in some cases, restoring the backups without recreating holes might even lead to disk-full errors.

andrewchambers:
There is a technical tradeoff here, complicate the code for something that won't be important 99 percent of the time? I also think actually restoring backups is far more rare than storing them. I don't mind either way, I just want to point out that adding more code hurts reliability as a whole.

cfcs:
We had some relevant discussion here: https://github.com/restic/restic/issues/117
(TL;DR: `fallocate()` is superior to `posix_fallocate()` - if `fallocate()` returns `EOPNOTSUPP` for a file, that result should be cached and regular zeroes should be written instead of attempting additional `fallocate()` calls)

[Documentation for glibc's lseek() - used when detecting sparse ranges](http://linux.die.net/man/2/lseek)

I think the main benefit from supporting sparse files would be in the `restore` scenario, so I'm going to be dealing with that use-case in this comment. Detecting holes upon `backup` saves some syscalls, while in the `restore` scenario we're saving IO, which is usually much slower.

[...]

PR #2378, opened 2019-08-20
Add support for sparse file restoration

d-b:

This is a tentative solution for restoring sparse files as discussed in issue #79. Since this implementation does not include a method to track which files were sparse to begin with, some files which were not sparse before but contain long runs of zeros will be made sparse.

The implementation is based on the observation that the chunker currently always returns a digest of 0 for a window of zeros. This results in chunks of zeros of minimum chunk size which all hash to the same ID. So, by comparing a blob's ID to that magic "zeros" ID we can quickly identify long runs of zeros and seek past them in the file instead of writing the zeros and allocating disk space.

There are likely numerous issues to discuss in this implementation. I consider it very much tentative and would like some feedback.

Yes, this is discussed in https://github.com/restic/restic/issues/79

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-79
+++ b/changelog/unreleased/issue-79
@@ -1,0 +1,5 @@
+Enhancement: Support for sparse file restoration
+
+Restore files with long runs of zeros as sparse on supported platforms
+
+https://github.com/restic/restic/issues/79
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -38,6 +38,7 @@ type RestoreOptions struct {
 	Paths              []string
 	Tags               restic.TagLists
 	Verify             bool
+	SparseFiles        bool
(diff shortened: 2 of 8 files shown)


chrestomanci:
I am just a user of restic, but I think this is a desirable feature.

I also think that it should be on by default if the OS supports sparse files, as I think that in most cases the user will prefer faster restores and reduced disk space consumption.

sbromberger:
Just wondering what the status of this PR is. I'm using restic to backup some VM images to B2; I'm testing downloads and it looks like it's creating a dense image.

d-b:
I implemented this with the same issue in mind. I too was hoping to backup sparse virtual disks with restic. I am hoping to see some feedback and movement on the PR

MichaelEischer:
Superseded by #3854.

PR #2398, opened 2019-09-13
b2: Fallback to b2_hide_file when delete returns unauthorized

### LINKED ISSUE #2134: Use b2_hide_file instead of b2_delete_file_version
Output of `restic version`
--------------------------
restic 0.9.2 (v0.9.2-1-g3c8bab2b-dirty) compiled with go1.11 on windows/amd64
(yes, I know, doesn't really matter though ;))

The B2 back-end should use b2_hide_file instead of b2_delete_file_version to remove locks, and possibly for all file removals, or offer an option to enable such behavior.

What are you trying to do?
--------------------------
As many others who also created feature requests for similar suggestions, I want to prevent an attacker who succeeds in breaking into my system from deleting / corrupting my backup. One scenario is a manual attack, but it is also not unthinkable that a crypto trojan will be written to search for credentials of common cloud backup providers on machines that it infects.

This particular suggestion solves the problem in a manner specific to the B2 back-end. It may be superseded by a more general solution but is very simple. First some background though.

**1. Versioning**
B2 has a concept of [file versions](https://www.backblaze.com/b2/docs/file_versions.html). API functions such as b2_list_file_names and b2_download_file_by_name only show you the current contents of directories and files, but previous versions of files as well as the contents of removed files may still be available through the use of b2_list_file_versions and b2_download_file_by_id. To make it clear that the file contents are not actually deleted and are still available through the versions API, the file remove operation is called "b2_hide_file". It is possible to set life-cycle rules to automatically delete these old versions after a given number of days, but any version (old or current) can also be deleted manually through b2_delete_file_version.

**2. Permissions**
In B2, permissions ([capabilities](https://www.backblaze.com/b2/docs/application_keys.html#capabilities)) are set on a per-access-key-basis, and are fairly coarse-grained. Of particular interest here are the capabilities "writeFiles" and "deleteFiles". writeFiles allows the application key to write/overwrite/remove files, but not to delete anything from the version history. In particular, it does not allow b2_delete_file_version - the "deleteFiles" capability is needed for that.

In short, application keys without the deleteFiles capability cannot irrevocably destroy information from the bucket, since they cannot meddle with the version history.

**3. Ok, so?**
The versioning feature of B2 would allow me to protect my restic backup by only giving the automated backup an application key without the "deleteFiles" capability. However, restic currently uses the b2_delete_file_version call to delete locks, so this currently does not work without errors. This could be fixed though by deleting the locks with b2_hide_file instead - it would keep the lock around in the version history, but that should not really cause much bother since lock files are tiny, and you can set up your bucket to automatically delete hidden files after a while.

Maybe it would even be a good idea to embrace the versioning and replace *all* calls to b2_delete_file_version with b2_hide_file (from what I understand that would actually be much easier to implement). This way, B2 would keep a version history for all operations, which would even make it possible to use commands like prune and forget with restricted access keys, and to recover some earlier state of the bucket in case the repository gets corrupted.

Restic gives me peace of mind by safeguarding not just my data, but also that on my parents' PC. When they call about strange noises, weird popups and the thing not booting anymore at all, I can always calmly say: Relax, everything is backed up.

DanielG:

This allows using restic in an append-only mode with B2. The idea is this
would provide security in case the machine being backed up is compromised.
The attacker won't be able to just immediately delete all backups with the
API key on the machine.

To use this feature create an application key without the deleteFiles
capability. I also recommend restricting the key to just one bucket. For
example using the b2 command line tool:

    b2 create-key --bucket <bucketName> <keyName> listBuckets,readFiles,writeFiles,listFiles

---

Before I go off documenting this change I would like some feedback on whether you guys would prefer the automatic fallback like in this PR or would rather have a command line config toggle?  

Checklist
---------

### DIFF
--- a/internal/backend/b2/b2.go
+++ b/internal/backend/b2/b2.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/kurin/blazer/b2"
+	"github.com/kurin/blazer/base"
 )
 
 // b2Backend is a backend which stores its data on Backblaze B2.
@@ -28,6 +29,8 @@ type b2Backend struct {
 	listMaxItems int
 	layout.Layout
 	sem sema.Semaphore
+
+	canDelete bool
 }
 
 // Billing happens in 1000 item granlarity, but we are more interested in reducing the number of network round trips
@@ -104,6 +107,7 @@ func Open(ctx context.Context, cfg Config, rt http.RoundTripper) (restic.Backend
 		},
 		listMaxItems: defaultListMaxItems,
 		sem:          sem,
+		canDelete:    true,
 	}
 
 	return be, nil
@@ -314,11 +318,27 @@ func (be *b2Backend) Remove(ctx context.Context, h restic.Handle) error {
 	// the retry backend will also repeat the remove method up to 10 times
 	for i := 0; i < 3; i++ {
 		obj := be.bucket.Object(be.Filename(h))
-		err := obj.Delete(ctx)
-		if err == nil {
-			// keep deleting until we are sure that no leftover file versions exist
-			continue
+
+		var err error
+		if be.canDelete {
+			err = obj.Delete(ctx)
+			if err == nil {
+				// keep deleting until we are sure that no leftover file versions exist
+				continue
+			}
+
+			code, _ := base.Code(err)
+			if code == 401 { // unauthorized
+				// fallback to hide if not allowed to delete files
+				be.canDelete = false
+				debug.Log("Removing %v failed, falling back to b2_hide_file.", h)
+				continue
+			}
+		} else {
+			// hide adds a new file version hiding all older ones, thus retries are not necessary
+			err = obj.Hide(ctx)
 		}
+
 		// consider a file as removed if b2 informs us that it does not exist
 		if b2.IsNotExist(err) {
 			return nil
--- a/changelog/unreleased/issue-2134
+++ b/changelog/unreleased/issue-2134
@@ -1,0 +1,19 @@
+Enhancement: Support B2 API keys restricted to hiding but not deleting files
+
+When the B2 backend does not have the necessary permissions to permanently
+delete files, it now automatically falls back to hiding files. This allows
+using restic with an application key which is not allowed to delete files.
+This prevents an attacker to delete backups with the API key used by restic.
+
+To use this feature create an application key without the deleteFiles
+capability. It is recommended to restrict the key to just one bucket.
+For example using the b2 command line tool:
+
+	b2 create-key --bucket <bucketName> <keyName> listBuckets,readFiles,writeFiles,listFiles
+
+Alternatively, you can use the S3 backend to access B2, as described
+in the documentation. In this mode, files are also only hidden instead
+of being deleted permanently.
+
+https://github.com/restic/restic/issues/2134
+https://github.com/restic/restic/pull/2398


DanielG:
Any chance I can get a review on this? @MichaelEischer @fd0 @rawtaz (sorry not sure who to ping)

PriceChild:
FWIW https://github.com/restic/restic/compare/master...Medo42:b2-hide-option looks like a take on doing this via config instead of at runtime.

@MichaelEischer I'd really like to help push this forward, would you be able to help answer the question as to which direction is preferred?

PriceChild:
I would argue in favour of this PR for the simple reason that it is simple.

While I understand the idea of separate keys for locks or config variables, this solution has some elegance:

It's not possible to create a key with write but without delete using the backblaze UI. You have to do so manually and intentionally using their api. Thus, if restic finds itself using such a key using this PR's change, I think it's very sensible for it to fall back on hide? I'm not sure the user would ever expect anything different?

PriceChild:
Related: https://github.com/restic/restic/issues/2134

internal/backend/b2/b2.go
```diff
@@ -258,7 +261,22 @@ func (be *b2Backend) Remove(ctx context.Context, h restic.Handle) error {
 	defer be.sem.ReleaseToken()
 
 	obj := be.bucket.Object(be.Filename(h))
-	return errors.Wrap(obj.Delete(ctx), "Delete")
+
+	if be.canDelete {
+		err := errors.Wrap(obj.Delete(ctx), "Delete")
+		if err != nil {
+			code, _ := base.Code(err)
```
MichaelEischer, line 268: This can't work as the error returned by `obj.Delete` is wrapped.

DanielG: Uups. I thought I did test out the error path but it's been so long I don't even remember.

DanielG (commented):
Thanks for the review!

I've changed the code to keep around the original error to extract the code. Haven't tested this yet though.

I'm still waiting on some guidance as to whether this is the direction we want to go in or rather something in the direction of #2887 before investing any more time here.

internal/backend/b2/b2.go
```diff
@@ -258,7 +261,23 @@ func (be *b2Backend) Remove(ctx context.Context, h restic.Handle) error {
 	defer be.sem.ReleaseToken()
 
 	obj := be.bucket.Object(be.Filename(h))
-	return errors.Wrap(obj.Delete(ctx), "Delete")
+
+	if be.canDelete {
+		delete_err := obj.Delete(ctx)
+		err := errors.Wrap(delete_err, "Delete")
+		if err != nil {
```
MichaelEischer, line 268: That check is still broken (iff delete_err != nil, then err != nil). Just remove this `if` then it should work.

DanielG: Ugh indeed that's all wrong still.

I don't think just removing the `if` would work. I'm not sure but`base.Code(nil)` probably wouldn't work. So instead I just return early in the nil case now.

I've also added a log for the fallback case so this should be easier to observe working now. Though again I haven't gotten around to testing yet.

themightychris:
+1 for automatic fallback to hide, with a notice logged. Best that people's backups err on the side of working :-)

Also, I have created immutable buckets through the web UI, I'm not sure if it implementsb through the same underlying mechanism, but it's an option only in the creation form and in manifests the same errors

DanielG:
> I'm not sure if it implements through the same underlying mechanism, but it's an option only in the creation form and it manifests the sa

You mean the new "Object lock" feature? I'm not sure that works the same as restricting the application key TBH but as long as the key you create can't clear the object lock I suspect it might just work out.

Skyebold:
If we can get this merged, then restic will have ransomware protection on b2, which would be pretty cool!  I've been hopefully watching restic/b2 ransomware protection support for a few years and would love to see some progress on this front to make it finally happen :)

With that in mind, it would be great to come to a decision and either merge this PR or #2887.

My personal vote would be for this PR.  It is simpler to set up for the user, and old versions of lock files should not be too large (and can eventually be permanently deleted automatically with the b2 lifecycle rule).

In future, it would be good to have a way to restore from old versions (ie, use bucket's file versions as they existed at datetime), along with adding to the docs what minimal b2 permissions are required -- but neither of those would block this PR.

MichaelEischer:
> My personal vote would be for this PR. It is simpler to set up for the user, and old versions of lock files should not be too large (and can eventually be permanently deleted automatically with the b2 lifecycle rule).

I agree, based on the lifecycle rules screenshot posted in https://github.com/restic/restic/issues/2134#issuecomment-1066171529, it looks like this PR would be the more user friendly solution.

I'm a bit undecided on whether it is a good idea to silently fall back to just hiding files or whether we should add an option that specifies whether delete/hide is used. When using hide, then users have to setup the key permissions and lifecycle rules accordingly anyways, so it would also be possible to pass an additional option to restic. On the other hand, not having to pass an option would be more convenient.

So the question is whether there are downsides when automatically falling back to hiding. In that variant, restic will at least send one delete request that will fail and have to be repeated, but that wouldn't be much of a problem. It would also be easier to accidentally not delete data but just hide it. However, I can't really say much about how likely that is to happen or not.

Skyebold:
I am pro additional parameter.  Here's my thinking:

> how likely that is to happen or not.

In practice, it seems very unlikely someone will wind up using an API key without delete files permission accidentally.  (At least currently) the B2 admin panel does not make it easy to create an application key without delete permission.  This is what we have now:

image

If you want to set up a key without the delete file permission, the only way to do it is via the API (AFAIK).  Thus, a user has to be really commited to setting things up this way.  They're not likely to accidentally have this permission disabled but actually want files to be deleteable.  However, it's always possible this UI may change in future.

It does seem more likely that the user may forget to configure their lifecycle rules to auto-delete old files, though.

> whether there are downsides when automatically falling back to hiding

For lock files, the damage done is pretty minimal -- just a trivial amount of space.  

For running a `forget`, that could potentially be more expensive if the user has not configured lifecycle rules.  IMO that situation would be worth having a parameter for to ensure the user has configured their bucket correctly.

So I'm not sure about halting for normal operations without a parameter (other than for consistency maybe), but for `forget` in particular, that one may need some special warning since the user is trying to free up space.  Something like:

`The provided API key does not have permission to delete files.  Restic can hide files instead of deleting by using the -whatever parameter, however, this will not free up space.  Before enabling this parameter, you may wish to ensure you have configured b2 lifecycle rules to delete hidden versions of files after a set time.`

MichaelEischer:
> I am pro additional parameter. Here's my thinking:

I agree. Let's just add an addiitonal parameter that tells the B2 backend to hide files when set. The PR also needs a changelog entry

MichaelEischer:
I've rebuilt the PR to provide a backend option to hide files. The option is currently called `-o b2.hide-files=true`, although we could rename it to something else. I wonder whether some other name might be easier to understand, maybe `soft-delete`? However, as the B2 documentation speaks of hiding files, that might be the easiest to understand variant.

Btw, using the S3 backend of B2 it is also possible to achieve the same results. For S3 the default is to just hide files.

Can someone test whether a key created using `b2 create-key --bucket <bucketName> <keyName> listBuckets,readFiles,writeFiles,listFiles` works or whether we need additonal permissions?

SkyeBlep:
> I've rebuilt the PR to provide a backend option to hide files. 

Sweet, thanks!

> The option is currently called `-o b2.hide-files=true`, although we could rename it to something else. I wonder whether some other name might be easier to understand, maybe `soft-delete`? However, as the B2 documentation speaks of hiding files, that might be the easiest to understand variant.

I would advocate to being consistent with the b2 terminology, since a b2 user can look up what 'hidden file' means in b2 and realize they should check their lifecycle rules to ensure hidden files are auto deleted.  Creating a new term might not be easily searchable.

> Btw, using the S3 backend of B2 it is also possible to achieve the same results. For S3 the default is to just hide files.

Interesting; this would be a useful info point to go into documentation for setting all this up.

> Can someone test whether a key created using `b2 create-key --bucket <bucketName> <keyName> listBuckets,readFiles,writeFiles,listFiles` works or whether we need additonal permissions?

I tested this today ([console output](https://github.com/restic/restic/files/9843014/restic-key-testing2.txt)).  Highlights:

- Yes, the above key generation command will generate a key with working permissions
- Init throws a 404 error, but seems to init just fine (`b2_download_file_by_name: 404: : b2.b2err`)
- Tested backing up, forgetting, and restoring -- all worked OK when using the hide option with limited permission key
- When called without the hide option, it retries on a 401 unauth error, which is kinda weird to retry on, but no big deal 
- One weird thing (bug?) is that calling backup does not clean up the lock afterwards?  IE, I can still see the locks folder & file in not-hidden state via the web interface after running a backup, and until unlock is manually called. Multiple backups can be called in a row, but can't seem to call backup > forget.  Instead unlock has to be manually called (backup > unlock > forget).
- After messing around with this a bit, I realized this is the kind of parameter that would be good to have a way to store in session memory (ie, an environment variable like password, etc.)  Since otherwise it's quite easy to forget to pass every time, and is a thing that remains consistent with the session.  That said, that could just be an area for future enhancement rather holding up this PR.

MichaelEischer:
Thanks for testing!

> * Init throws a 404 error, but seems to init just fine (`b2_download_file_by_name: 404: : b2.b2err`)

That #3750, the "error" is essentially saying that no repository exists right now.

> * When called without the hide option, it retries on a 401 unauth error, which is kinda weird to retry on, but no big deal

That error is returned by B2: "The auth token used is valid, but does not authorize this call with these parameters."

> * One weird thing (bug?) is that calling backup does not clean up the lock afterwards?

`backup` also must be called with the `b2.hide-files` option set. Otherwise it won't be able to delete it's lock file. In general it is always necessary to specify that option, when using a restricted application key. For now the easiest way is to use a small wrapper script which always adds the necessary parameter (or create an alias).

> > Btw, using the S3 backend of B2 it is also possible to achieve the same results. For S3 the default is to just hide files.
> 
> Interesting; this would be a useful info point to go into documentation for setting all this up.

Right now we don't have a place which documents setting up append-only repositories. The part that using the S3-backend of B2 only hides files, is already subtly mentioned in https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#backblaze-b2 .

SkyeBlep:
Thanks -- and yep, that makes sense now!  Quite excited to be able to use this feature :)

MichaelEischer:
The changelog now says that the parameter is always necessary.

MichaelEischer:
I've fixed a few typos in the changelog and changed it to use the correct option name.

MichaelEischer:
After talking with @rawtaz and the unintuitive behavior regarding the lock files, I've switched the approach back to automatically falling back to hiding files if not authorized to delete them.

The main change to the previous implementation, besides making the code work with latest changes to the B2 backend, is that the fallback now works silently and does not cause restic to print a warning that the remove operation had to be retried.

Jwink3101:
Just wondering if using B2 via rclone would instead provide the "hide" rather than delete. Or does the restic interface do hard deletes too?

(Sorry, not in a position to test empirically)

MichaelEischer:
rclone itself hides files by default on B2. It is just that restic passes the `--b2-hard-delete` to rclone by default (but that can be changed).

Could someone test the current PR version which automatically falls back to hiding files if deleting is not possible?

SkyeBlep:
I gave it a quick test and it seems to be working well.  Definitely a lot easier to use now that we don't need to remember to pass the parameter.  :+1: 

image

[console output log of test](https://github.com/restic/restic/files/10204324/restic-second-test.txt)

MichaelEischer (approved):
LGTM. Thanks for testing!

the-emerald:
> rclone itself hides files by default on B2. It is just that restic passes the `--b2-hard-delete` to rclone by default (but that can be changed).
> 
> Could someone test the current PR version which automatically falls back to hiding files if deleting is not possible?

Hmm, does this mean once this lands an append-only rclone server can now use a key without the deleteFiles permission?

MichaelEischer:
> append-only mode rclone + b2 no-deleteFile should work already. How can I disable the `--b2-hard-delete` flag being passed?

call restic like this: `restic -o rclone.args="serve restic --stdio" [...]`.

the-emerald:
Is this for calling restic on the backup client side, or in place of `rclone serve restic`? My rclone host sits on another server.

MichaelEischer:
> Is this for calling restic on the backup client side, or in place of `rclone serve restic`? My rclone host sits on another server.

That example is for the case when restic starts rclone itself. If rclone already runs at some other server, then just omit the `--b2-hard-delete` flag there?

the-emerald:
But I thought you said the --b2-hard-delete flag was passed in by default?

MichaelEischer:
> But I thought you said the --b2-hard-delete flag was passed in by default?

??? When using the `rclone` backend, then restic concatenates the `rclone.program` (default `rclone`) and `rclone.args` (default `serve restic --stdio --b2-hard-delete`) and runs that command. In that case you can use the `-o rclone.args="serve restic --stdio"` from above to get rid of the `--b2-hard-delete` option.

If rclone runs on a different host than restic, then you're either connecting to it using the `rest` backend (rclone was started using `rclone serve restic --addr :8080`, that command line doesn't contain the `--b2-hard-delete` flag) or you have a setup similar to https://ruderich.org/simon/notes/append-only-backups-with-restic-and-rclone . In the latter case just make sure that the forced ssh-command doesn't contain the `--b2-hard-delete` parameter. Or when not using forced commands, then just set the already mentioned `rclone.args`.

Which of these variants matches your setup?

the-emerald:
I have a setup similar to the latter! So in that case, it's not being passed in by default...!

PR #2406, opened 2019-09-21
Docs: Reword developer note in forget.

jtagcat:
When I went trough the docs, I had to read the text-block multiple times to understand, that the changed part did not belong together with the rest.

After 7 minutes, I did not find documentation on readthedocs markdown. Not sure, if correct.

no

Checklist
---------

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -216,8 +216,8 @@ Let's explain this with an example: Suppose you have only made a backup
 on each Sunday for 12 weeks. Then ``forget --keep-daily 4`` will keep
 the last four snapshots for the last four Sundays, but remove the rest.
 
-.. note:: Restic has a safety feature, preventing restic from removing many snapshots, when no newer exist.
-   If implemented otherwise, running ``forget --keep-daily 4`` on a Friday would remove all snapshots.
+Restic has a safety feature, preventing restic from removing many snapshots, when no newer exist.
+If implemented otherwise, running ``forget --keep-daily 4`` on a Friday would remove all snapshots.
 
 Another example: Suppose you make daily backups for 100 years. Then
 ``forget --keep-daily 7 --keep-weekly 5 --keep-monthly 12 --keep-yearly 75``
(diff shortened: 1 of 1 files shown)


rawtaz:
@jtagcat Actually, unless I'm mistaken, the text you want to change actually does belong with the text above it (in the same paragraph).

The point it's making is that when you have one backup per week (on sundays) for *twelve* weeks, and then ask to remove the last four *daily* snapshots, restic will not remove all those weekly snapshots (which you'd expect it to, because all those weekly snapshots were made *before* the last four days, and hence should be removed).

This is because it's smart enough to understand that the only daily snapshots there is available for keeping, are those weekly ones. Even though they are older than the last four days. So it keeps those, and that's what the text is trying to say.

Unless I'm missing something here, I think we should either close this PR or reword the text to make it easier to understand. What do you think?

jtagcat:
Well then if it does belong with the text, I would move the note to a newline instead in the previous paragraph.

rawtaz:
> Well then if it does belong with the text, I would move the note to a newline instead in the previous paragraph.

Sure, can you change it to that (and squash)?

doc/060_forget.rst
```diff
@@ -215,10 +215,9 @@ single snapshot on 2017-09-30 (Sat) will count as a daily, weekly and monthly.
 Let's explain this with an example: Suppose you have only made a backup
 on each Sunday for 12 weeks. Then ``forget --keep-daily 4`` will keep
 the last four snapshots for the last four Sundays, but remove the rest.
-Only counting the days which have a backup and ignore the ones without
-is a safety feature: it prevents restic from removing many snapshots
-when no new ones are created. If it was implemented otherwise, running
-``forget --keep-daily 4`` on a Friday would remove all snapshots!
+
+Restic has a safety feature, preventing restic from removing many snapshots, when no newer exist.
+If implemented otherwise, running ``forget --keep-daily 4`` on a Friday would remove all snapshots.
```
fd0: Hm, now the text is technically incorrect. Thank you very much for your contribution, but I think the right way would be to add a complete example (with restic output and so on) instead of changing the text. So I'm inclined to close this PR for now.

fd0:
Thank you very much for your feedback that the text is hard to understand, I'll rework it and add an example. I'm closing this PR.

PR #2423, opened 2019-09-26
sftp: support user@domain parsing as user

streambinder:

This change allows `user@domain` to authenticate over SFTP servers.

No.

Checklist
---------

### DIFF
--- a/internal/backend/sftp/config.go
+++ b/internal/backend/sftp/config.go
@@ -56,8 +56,11 @@ func ParseConfig(s string) (interface{}, error) {
 		host = data[0]
 		dir = data[1]
 		// split user and host at the "@"
-		data = strings.SplitN(host, "@", 2)
-		if len(data) == 2 {
+		data = strings.SplitN(host, "@", 3)
+		if len(data) == 3 {
+			user = data[0] + "@" + data[1]
+			host = data[2]
+		} else if len(data) == 2 {
 			user = data[0]
 			host = data[1]
 		}
--- a/changelog/unreleased/pull-2423
+++ b/changelog/unreleased/pull-2423
@@ -1,0 +1,5 @@
+Enhancement: support user@domain parsing as user
+
+Added the ability for user@domain-like users to be authenticated over SFTP servers.
+
+https://github.com/restic/restic/pull/2423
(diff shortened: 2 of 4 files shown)


internal/backend/sftp/config.go
```diff
@@ -56,10 +58,13 @@ func ParseConfig(s string) (interface{}, error) {
...
+		//
+		// it reverts the host data to support user@domain@hostname notation
+		// that's used by several SFTP servers providers
+		data = strings.SplitN(stringutil.Reverse(host), "@", 2)
 		if len(data) == 2 {
-			user = data[0]
-			host = data[1]
+			user = stringutil.Reverse(data[1])
+			host = stringutil.Reverse(data[0])
 		}
```
rawtaz: While I see the nifty trick in reversing the string, it doesn't seem very clean.

Also doesn't feel great importing an *examples* dependency? I haven't seen that before, are they even something one can count in being there and not changing too much? Perhaps it's not too uncommon in the Go world, I have no idea.

What about something like this instead?

~~~
data = strings.SplitN(stringutil.Reverse(host), "@", 3)
if len(data) == 3 {
  user = data[0] + "@" + data[1]
  host = data[2]
} else {
  user = data[0]
  host = data[1]
}
~~~

I'm just thinking it's less magic, easier to read down the road, and one less dependency.

streambinder: Yup, I agree with you.

rawtaz:
@streambinder I requested a review from the authoritative party, let's see.

Meanwhile, can you squash all the commits into just one?

rawtaz:
@streambinder After squashing, please also rebase your master onto rustic's master branch, to see if that solves the build errors.

rawtaz:
Nice work @streambinder! Let's see what @fd0 says :)

rawtaz:
@streambinder I got word from @fd0 that he approves this change. Before merging though, we need to update the documentation. Can you please:

- Add a new changelog entry as described in one of the checkboxes you can then check.
- Update `030_preparing_a_new_repo.rst` to contain a note that one can specify username with the new format.

rawtaz (approved):
Thanks!

Issue #2364, opened 2019-08-08
Exit status 0 on backup and invocation errors; need to distinguish warnings from errors


HaleTom:
Summary
---
While having different exit codes to distinguish between the causes of non-`0` exit status is a feature request (#956), I assert that returning a `0` exit code in some situations warrants a bug.

Below I demonstrate two cases of `backup` emitting errors and exiting `0`.

In cases like this, data loss can occur as follows:
1. All data is initially backed up
2. A change occurs (eg permissions on data being backed up)
3. Restic emits errors but  still exits `0` so these errors are not presented for human inspection
4. Result: Data appears to be backed up when in fact it isn't.

--- 

Additionally, there are some cases where a `0` exit is confusing to a human, and may also confuse automation scripts:

1. Running `restic key backup /` returns `0`. I ran into this when editing a previous command line and wondered why my backup "succeeded" so quickly.

1. Running `restic mount -H Athena latest /mountpoint` doesn't exit `0` with an invocation complaint, but rather mounts at location `latest` and ignores the `/mountpoint` argument.

1. Running `restic snapshots -H athena` returned `0` but didn't list any snapshots.
Running `restic snapshots -H Athena` (note the capitalisation) produced a list of snapshots.
I'd expect that like `grep`, if nothing matches, restic would exit non-`0` to allow for automated checking if snapshots exist.

-----

Related: "Should restic mount quit with Exit code 130?": #2015

Output of `restic version`
--------------------------

    restic 0.9.5 compiled with go1.12.4 on linux/amd64

How did you run restic exactly?
-------------------------------

Invocation:

    restic backup -v --tag seed "/media/1TB-toshiba/01TB-master/d/Media/Music" "/media/1TB-toshiba/01TB-master/c/Music"

End of output:
```
error: lstat /media/1TB-toshiba/01TB-master/d/Media/Music/Media Monkey/Working/From Christine/Who Killed Amanda Palmer 320k MP3: no such file or directory
error: lstat /media/1TB-toshiba/01TB-master/d/Media/Music/Media Monkey/Working/From Christine/XTC: no such file or directory
error: lstat /media/1TB-toshiba/01TB-master/d/Media/Music/Media Monkey/Working/From Christine/ZZ Top: no such file or directory
error: lstat /media/1TB-toshiba/01TB-master/d/Media/Music/Media Monkey/Working/From Christine/__ Playlists __: no such file or directory
[...]

PR #2461, opened 2019-11-05
Add documentation on exit status codes to man pages

rigtorp:
This is step one to start defining useful exit codes for all the commands.

Add documentation on restic exit status codes to all man pages.

Yes, it's related to #2364 but doesn't resolve the issue.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -20,6 +20,11 @@ a directory.
 
 The special snapshot "latest" can be used to restore the latest snapshot in the
 repository.
+
+EXIT STATUS
+===========
+
+Exit status is 0 if the command was successful, and non-zero if there was any error.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -35,6 +35,14 @@ var cmdBackup = &cobra.Command{
 	Long: `
 The "backup" command creates a new snapshot and saves the files and directories
 given as the arguments.
+
+EXIT STATUS
+===========
+
+Exit status is 0 if the command was successful, and non-zero if there was any error.
+
+Note that some issues such as unreadable or deleted files during backup
+currently doesn't result in a non-zero error exit status.
 `,
 	PreRun: func(cmd *cobra.Command, args []string) {
 		if backupOptions.Host == "" {
--- a/cmd/restic/cmd_cache.go
+++ b/cmd/restic/cmd_cache.go
@@ -19,6 +19,11 @@ var cmdCache = &cobra.Command{
 	Short: "Operate on local cache directories",
 	Long: `
 The "cache" command allows listing and cleaning local cache directories.
+
+EXIT STATUS
+===========
+
+Exit status is 0 if the command was successful, and non-zero if there was any error.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -18,6 +18,11 @@ var cmdCat = &cobra.Command{
 	Short: "Print internal objects to stdout",
 	Long: `
 The "cat" command is used to print internal objects to stdout.
+
+EXIT STATUS
+===========
+
+Exit status is 0 if the command was successful, and non-zero if there was any error.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -25,6 +25,11 @@ finds. It can also be used to read all data and therefore simulate a restore.
 
 By default, the "check" command will always load all data directly from the
 repository and not use a local cache.
+
+EXIT STATUS
+===========
+
+Exit status is 0 if the command was successful, and non-zero if there was any error.
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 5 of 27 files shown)


rawtaz:
@rigtorp Thanks for your contribution, a small but good one!

I have a couple of questions:

1. Restic can return other exit codes than 0 and 1, so could you reword the text to be something like "Exit status is 0 if the command was successful, and non-zero if there was any error." instead?

2. Is there any particular reason why the "EXIT STATUS" is quoted, when SEE ALSO is not? Should the quotes be removed, or should the other one have them added?

Since this branch isn't merged, you can force-push amended commits to it if you want.

rigtorp:
>     1. Restic can return other exit codes than 0 and 1, so could you reword the text to be something like "Exit status is 0 if the command was successful, and non-zero if there was any error." instead?

No Restic can only return 0 or 1 currently. It would great to improve that in the future, for example to return a specific error code when backup completed with warnings vs backup failed. See #2364 

>
>     2. Is there any particular reason why the "EXIT STATUS" is quoted, when SEE ALSO is not? Should the quotes be removed, or should the other one have them added?

I've never thought about it. I've always quoted the section headers. I will remove the quotes and verify that the output still looks correct. Better stick to the existing style.

rigtorp:
Restic exit status:
https://github.com/restic/restic/blob/master/cmd/restic/main.go#L107

For reference:
- http://www.tldp.org/LDP/abs/html/exitcodes.html
- https://github.com/openbsd/src/blob/master/include/sysexits.h
- https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

rigtorp:
Actually I found this code where Restic can exit with 130:
https://github.com/restic/restic/blob/master/cmd/restic/cleanup.go#L69

rawtaz:
Thanks for elaborating :) I'm certainly not always right, very open to being proven wrong.

Yeah, that one 130, and there's also a 2, 3 and 5 if you enable debug.go I think.

If you don't mind I'd like to change the text, because that will also make it more future-proof, for whenever someone forgets to change the EXIT STATUS text properly down the line.

rigtorp:
I'm going to change it to this:
```
Exit status is 0 if the command was successful, and non-zero if there was any error.

0    Successful program execution.
1    Generic error.
130  Program execution was interrupted with SIGINT.
```

Later we can use `errors.Is()` to query for IO, DNS and other errors and return according to `/usr/include/sysexits.h`

rawtaz:
Fair enough - after all, having the various possible exit codes documented is what this is about.

fd0:
Thanks for your contribution! Unfortunately the man pages are automatically generated from the descriptions in the source code, usually before a release (so they are up to date). One example for a man page text is here:

https://github.com/restic/restic/blob/26d1f9f4ba6c106a90e2af9a96df50d8898a9bb5/cmd/restic/cmd_check.go#L23-L27

You can generate the manpages by building restic and then running `mkdir manpages && ./restic generate --man manpages`. They can be viewed using `man` like this: `man ./manpages/restic-check.1`. Inserting a new section works like this (I think it's markdown):

```go
var cmdCheck = &cobra.Command{
	Use:   "check [flags]",
	Short: "Check the repository for errors",
	Long: `
The "check" command tests the repository for errors and reports any errors it
finds. It can also be used to read all data and therefore simulate a restore.

By default, the "check" command will always load all data directly from the
repository and not use a local cache.

EXIT STATUS
===========

Exit status is 0 if command was successful, and 1 if there was any error.
`,
	DisableAutoGenTag: true,
```

rigtorp:
@fd0 Eventually the exit status will diverge between the commands. In particular I want to make the backup command return different status depending on if backup completed without warnings or it failed. Currently it returns 0 when backup completed with warnings.

rawtaz:
@rigtorp Maybe it would be a good idea to make that (separate exit codes) a new issue for tracking, and fix this PR to reflect the current state with @fd0's comments above in mind, so we can merge this.

rigtorp:
@rawtaz Yes that's what I had in mind.

rigtorp:
@fd0 I have addressed all concerns. Specifically I added a note about current behavior of backup command.

changelog/unreleased/pull-2461
```diff
@@ -0,0 +1,3 @@
+Enhancement: Add documentation on exit status codes to man pages
+
```
rawtaz: Hm, does Calens (the changeling generator) work without the middle paragraph?

cmd/restic/cmd_restore.go
```diff
@@ -1,12 +1,13 @@
...
 
 import (
+	"strings"
+
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/restorer"
-	"strings"
```
rawtaz: This change is unrelated and should be removed. I can try to do that when I have some time if you are fed up with making corrections to this PR by now :)

rigtorp: Fixed it, I think it was go fmt or something that changed it.

rigtorp:
@rawtaz Friendly reminder!

rigtorp:
@rawtaz @fd0 Can we merge this?

rawtaz:
Looked at it earlier today, just waiting for a verification on something before we merge.

But we should remove the changelog file entirely. I'll do that if it's not done by whenever I merge it.

rawtaz:
The following needs to be cleaned up, then this can be finally merged:

- All changes to files under `doc/man/` should be reverted/removed (man pages are generated upon release).
- All changes to files under `changelog/unreleased/` should be reverted/removed (not needed for doc changes).
- Squash commits into one.

rawtaz (approved):
I did the final polish. Thanks you very much Erik! Sorry it took so long.

PR #2462, opened 2019-11-06
Refactor close response body

whs-dot-hk:
Refactoring to `defer resp.Body.Close()`

like the one show in https://golang.org/pkg/net/http/

No

Checklist
---------

### DIFF
--- a/internal/selfupdate/github.go
+++ b/internal/selfupdate/github.go
@@ -129,13 +129,9 @@ func getGithubData(ctx context.Context, url string) ([]byte, error) {
 		return nil, fmt.Errorf("unexpected status %v (%v) returned", res.StatusCode, res.Status)
 	}
 
-	buf, err := ioutil.ReadAll(res.Body)
-	if err != nil {
-		_ = res.Body.Close()
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


internal/selfupdate/github.go
```diff
@@ -129,13 +129,9 @@ func getGithubData(ctx context.Context, url string) ([]byte, error) {
 		return nil, fmt.Errorf("unexpected status %v (%v) returned", res.StatusCode, res.Status)
 	}
 
-	buf, err := ioutil.ReadAll(res.Body)
-	if err != nil {
-		_ = res.Body.Close()
-		return nil, err
-	}
+	defer res.Body.Close()
```
ifedorenko: This ignores errors closing the body. Not likely to happen, but no reason to take the risk either.

ifedorenko (commented):
Can you please explain what you are trying to achieve? The change _appears_ to save three lines of code and make error handling little less robust, but I may simply misunderstand the goal.

fd0:
Like @ifedorenko, I also don't see how this makes the code better. Less code is not always better :)

whs-dot-hk:
Agree

PR #2471, opened 2019-11-18
Small fix to the forget --tag command documentation

### LINKED ISSUE #2460: The forget --tags flag documentation is incorrect (or the implementation is incorrect)
Output of `restic version`
--------------------------
`restic 0.9.5 compiled with go1.13.1 on linux/amd64`

Problem description
---------------------------
The documentation here: https://restic.readthedocs.io/en/latest/060_forget.html states:

> To only keep the last snapshot of all snapshots with both the tag foo and bar set use:
>
> `$ restic forget --tag foo,tag bar --keep-last 1`

However from testing this syntax seems to be incorrect, with the correct syntax being:

`restic forget --tag foo,bar --keep-last 1`

What was documented, and the working version stated above

Local

Expected behavior
-----------------

Remove all but the last snapshot with both foo and bar as tags

Actual behavior
---------------

Running the documented command tells you that bar is not a valid snapshot id (I think that it thinks `tag` is the other tag I want to use)

1. Add 2 backups with both foo and bar as tags
2. run the forget commands above, the documented one does not work.

N/A

Update the documentation, the current implementation is far better compared to the documented version (for automating)

Yep, my main backup solution

8176135:
Closes #2460

I don't think this needs a change log as it is just a documentation fix.

Update the documentation to be inline with the current implementation

#2460

Checklist
---------

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -197,7 +197,7 @@ To only keep the last snapshot of all snapshots with both the tag ``foo`` and
 
 .. code-block:: console
 
-   $ restic forget --tag foo,tag bar --keep-last 1
+   $ restic forget --tag foo,bar --keep-last 1
 
 All the ``--keep-*`` options above only count
 hours/days/weeks/months/years which have a snapshot, so those without a


rawtaz (approved):

rawtaz:
Thanks @8176135 for an awesome contribution! 🥇

PR #2475, opened 2019-11-18
Use GitHub Actions for CI

fd0:
This is an infrastructure PR.

Remove Travis and AppVeyor and run tests and linters on GitHub actions.

Right now, two services run the tests for each push to master and for each Pull Request:
 * Travis:
   * This includes the tests for the cloud backends which needs access credentials
   * For Linux:
     * runs the tests on all Go versions we support (1.13 to 1.15)
     * builds restic for all supported architectures and OS on all supported Go versions (1.13 to 1.15)
     * runs the fuse tests on the latest Go version
   * For macOS: runs the tests for the latest Go version (1.15)
 * AppVeyor:
   * Runs the tests on Windows (excluding the cloud backends and fuse) for the latest Go version (1.15)

I'd like to replace this infrastructure with GitHub Actions (so we have everything in one place).

Side goals:
 * Simplify the infrastructure to run the tests. At the moment we do that by running `run_integration_tests.go` which is…not great.
 * Use `golang-ci` instead of running `gofmt` etc. manually.
 * Use https://github.com/fatih/faillint to check for unwanted imports (if we still need that)
 * Enable more linters for `golang-ci` as we fix issues in the codebase (staticcheck, errcheck etc.)
 * Upload the binaries built in the cross-compilation check and upload them to https://beta.restic.net, add a comment so that people can find them and easily try the restic version from a particular PR


shtratos:
@fd0 I've fixed rest-server installation here: https://github.com/shtratos/restic/pull/1
Feel free to cherry-pick or just apply the patch.

I'll see if I can port other parts of `run_integration_tests.go` to Github Actions and golangci-lint when I have spare time.
I'm not going to fix any linter errors, this is probably best done in a separate PR.

ad-m:
Why upload binaries to external service? GitHub Actions have already support for assets.

fd0:
@ad-m oh, thanks for the hint :)

ad-m:
Here is available details for artifact per build: https://github.com/actions/upload-artifact

fd0:
I've played around with adding assets to PRs, and unfortunately it isn't suited for us: uploading multiple assets (e.g. the complied binaries for each architecture) is cumbersome and the assets can only be downloaded by authenticated users. So we'll build something else (or leave the https://beta.restic.net running and extend it to pull requests maybe).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,175 @@
...
+        # list of jobs to run:
+        include:
+          - job_name: macOS
+            go: 1.15.x
+            os: macOS-latest
+            arch: darwin-amd64
+            test_cloud_backends: false
+            test_fuse: true
+
+          # - job_name: go1.15
```
ad-m: I would like notice, because it's not clear in GitHub Actions documentation that `matrix` might be dynamic. See https://github.com/apache/airflow/blob/master/.github/workflows/ci.yml#L466 . This should simplify the expression of the supported configuration and ensure the consistency of the CI system with the documentation.

ad-m:
@fd0, I can see you are waiting for long time to configure the CI. Sometimes the changes are minor typo fixes. Maybe you would consider using https://github.com/marketplace/actions/debugging-with-tmate, which should allow it to iterate faster without full commit-push-queue-build flow ?

fd0:
@ad-m Heh, thanks for the hint! I've used it several times already, but on Windows the shell it gives me lives in a different environment (msys2?), so it's not that helpful :)

.github/workflows/tests.yml
fd0: Thanks for the hint, we can add that later once the tests start actually working ;)

fd0:
Ok, that's it for today, everything works! Let me know if you see any opportunity for improvement! :)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+
+    env:
+      GOPROXY: https://proxy.golang.org
+
+    steps:
+      - name: Set up Go ${{ matrix.go }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+        id: go
```
ad-m, line 64:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+      GOPROXY: https://proxy.golang.org
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Go ${{ env.go }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.go }}
+        id: go
```
ad-m, line 215:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+
+      - name: Check out code
+        uses: actions/checkout@v1
```
ad-m, line 218:
```suggestion
        uses: actions/checkout@v2
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+          unzip libintl.zip
+          Invoke-WebRequest https://master.dl.sourceforge.net/project/gnuwin32/libiconv/1.8-1/libiconv-1.8-1-bin.zip -OutFile libiconv.zip
+          unzip libiconv.zip
+
+          # add $USERPROFILE/tar/bin to path
+          echo $Env:USERPROFILE\tar\bin >> $Env:GITHUB_PATH
+        if: matrix.os == 'windows-latest'
+
+      - name: Check out code
+        uses: actions/checkout@v1
```
ad-m, line 138:
```suggestion
        uses: actions/checkout@v2
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+        id: go
+
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Install gox
+        run: |
+          go get github.com/mitchellh/gox
+
+      - name: Cross-Compile with gox
```
ad-m, line 224:
```suggestion
      - name: Cross-compile with gox
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,254 @@
...
+            restic/backend/s3.TestBackendMinio,\
+            restic/backend/rclone.TestBackendRclone,\
+            restic/backend/s3.TestBackendS3,\
+            restic/backend/swift.TestBackendSwift,\
+            restic/backend/b2.TestBackendB2,\
+            restic/backend/gs.TestBackendGS,\
+            restic/backend/azure.TestBackendAzure"
+        run: |
+          # prepare credentials for Google Cloud Storage tests in a temp file
+          export RESTIC_TEST_GS_APPLICATION_CREDENTIALS=$(mktemp --tmpdir restic-gcs-auth-XXXXXXX)
```
ad-m, line 183: Why this variable is required? There is no reference to `RESTIC_TEST_GS_APPLICATION_CREDENTIALS` in source code. There is commit ba75a3884cf97a273fbd07a153ba7e186874834b which use that variable, but I have impression that we don't need that middle variable and we can use `GOOGLE_APPLICATION_CREDENTIALS` directly.

ad-m (approved):

.github/workflows/tests.yml
fd0: I'll try, I've just converted the variables from what we used with Travis, without giving it much thought ;)

PR #2481, opened 2019-11-21
WIP: Build/Release process enhancement

### LINKED ISSUE #2457: Enhance build process

The build process seems to be working well for restic itself. I see that the change log generator `calens` has already been moved in its own repo.

I suggest moving the build tools in its own repo (maybe in the existing https://github.com/restic/builder repo ?), so that in the long run we can use it for the https://github.com/restic/rest-server repo and finally close https://github.com/restic/rest-server/issues/77.

@fd0 seem to be the person to ask regarding the build process, could you give us your input ?

Should I open a dedicated ticket in https://github.com/restic/rest-server in order to get the automated build process enhanced ?

What are you trying to do?
--------------------------

Enhance the global restic build process.

Not yet, but it will surely replace my existing backup system.

jooola:

I would like to facilitate this project development by enhancing the build process (e.g. dropping some custom build scripts and trying to use the Makefile as single source of truth).
I've been warned not to make a huge PR, so I spitted all changes by features in individual commits.

Documentation will follow once the technical changes are ok for everyone.

This PR will probably be dependent on #2475 

- Remove vendor folder: With Go modules we can keep track of all our dependencies without having to store them in the project. This should reduce the project size and not impact reproducible builds. Also since Go only supports n+2 major versions, and at time of writing the latest is 1.13. Version below 1.11 aren't supported anymore and I think it is safe to say we could only rely on go modules. (This change also makes a step forward in dropping the build.go script)

- Use goreleaser as release tool: Releasing new version should be done in CI. So we could use Goreleaser which provided an easy way to build a large variety of installation ways (binaries/deb/rpm/docker/and more if enabled), and ease the release process.
The integration between calens (changelog generator) and goreleaser hasn't been added yet, I'll fix this with some future commits. Also this should remove some weight from @fd0 shoulders since he won't be the one required to release a new version (or maybe this is intended ?).

- Docker multi-stage: Instead of using a build.sh script, we should use the power of multistage Docker build. This allow us to isolate the whole process and copy the binary from one builder image to the new production image. (This might also be useful for the restic/builder repo, and the reproducible builds)

- Makefile as single source of truth: I found many script in bash or go used to build, test, release. We should if possible rewrite those script in the makefile, or if too complex, reference the script in the makefile like a wrapper. This should simplify the project management. For example, running the calens changelog generation, some linting tests, and furthermore.

IMO the restic/builder repo should be kept as tool for reproducible builds, but it doesn't make sense to use it for anything else, specially if the rest is managed with make and the CI.

Since it's still a WIP, more arguments regarding this PR might follow.

To be done:
- Integrate [goreleaser with GitHub actions](https://goreleaser.com/actions/) (Since travis CI is being dropped right ?)
- Write/Fix documentation
- Some missing changes that should be part of this PR ?

closes #2457

Checklist
---------


rawtaz:
Thanks for looking into this @jooola. I think someone more involved in the core of restic need to check this, but meanwhile I'm wondering if you have discussed these changes with anyone before? It looks like a lot of big changes.

jooola:
@rawtaz Yes, I had few minutes with @fd0 yesterday on IRC. Even if we didn't discussed this in details, I suggested to open a PR and have the technical discussion here instead of IRC.

I had the feeling that @fd0  was open for such project management changes. But he had no idea what my PR was really about. Let's see what people here think about those changes.

jooola:
Here is the really quick IRC chat I had with @fd0 and I wanted to answer few of his messages:
```
<joola> Hello there, I'm working on a better build tool. And I came across the builder repo that is actually beeing updated by fd0, was wondering if my work is useless ?
<joola> @fd0 might I have your input ?
<fd0> joola: what are you building with your tool?
<fd0> joola: we're using the builder repo to build a Docker container which is used to build the official restic binaries
<fd0> joola: https://restic.readthedocs.io/en/latest/developer_information.html#building-the-official-binaries
<t77b> Title: Developer Information — restic 0.9.5  documentation
<fd0> joola: oh, you mean a build tool for restic? Interesting, why do you think another one (besides build.go) is needed?
<joola> fd0, I think some tools already exist and do this jobs really nicely
<fd0> joola: I'd like to raise the minimum Go version needed to 1.11 so we can use "go build" without any program
<joola> I already used goreleaser, wichi makes an amazing job, for building multi arch binaries, docker images, publish everything deb/rpm packages
<joola> Also I though we could get ride of the vendor folder and only use the standard go mod tools which already make the job done in build.go, copying the right version of each dependencies in a temp folder 
<joola> This removes the problematic of maintaining the build.go script
<joola> https://goreleaser.com/introduction/
<t77b> Title: Introduction · GoReleaser
<fd0> Yeah, I've looked at goreleaser, but build.go and goreleaser serve different purposes
<joola> fd0, Oh could you tell me which one ?
<joola> I mean what the differences are
<fd0> goreleaser is for pre-built binaries for official releases, the oer is for end users
<fd0> Ah
<fd0> I'm on mobile right now
<fd0> Build.go is for end users
<fd0> It was meant as an easy way for users to build restic without having to setup a GOPATH
<joola> Yeah sure but users can also build there own package using gorealeaser, without publishing the build packages
<fd0> but now we have Go Modules so we don't need that any more
<fd0> Jo
<fd0> joola: true
<joola> Should I still propose a small PR, with some changes, and keep a discussion on the newly created PR ?
<fd0> So, next steps here are making a restic release, then raising the
<joola> Also goreleaser works well with github actions
<fd0> Go version to 1.11 and drop build.go
<fd0> Small steps please :)
<fd0> I'll have another look at it
<joola> Ok, yeah , I just read you thoughts about the future of restic
<fd0> The last time I looked it didn't seem to fit for our use case
<fd0> joola: have a look at the docs for the release process
<fd0> Bbl
<joola> Ok thanks
```
> <fd0> joola: I'd like to raise the minimum Go version needed to 1.11 so we can use "go build" without any program

> <fd0> It was meant as an easy way for users to build restic without having to setup a GOPATH

This PR is going that way.

> <fd0> goreleaser is for pre-built binaries for official releases, the oer is for end users

Agreed, still end user building can now be as easy as running go build (or make build).

fd0:
Hey, thanks a lot for trying to improve the project!

It feels to me this PR is way too much bundled together, and I'd like to discuss the individual things before we implement them (ideally in different PRs so it's easier to review).

I've extracted your proposed changes below (please add the ones I've forgotten):
 1. Require Go >= 1.11
 1. Remove the `vendor/` folder
 1. Use `goreleaser` to build the binaries for a release
 1. Improve the Docker image for restic
 1. Rework the `Makefile`

Here are my thoughts so far:
 * Require Go >= 1.11: We can always build in Module mode. This would still enable building restic on Debian stable (and even oldstable via backports), so it should not affect users that much. I don't particularly care about what the Go team supports, it's important that we enable building restic for as many users as possible without sacrificing too much. My personal reference for that is Debian stable, which has Go 1.11. It's also available in Debian oldstable, so that's a plus. This change would allow us to drop `build.go` and simplify the build and CI process a lot (we don't need to test anything in `GOPATH` mode any more).
 * Improve the Docker image for restic: That's a good idea, I personally don't use Docker so much, so if there's something to improve go ahead! For review, it'd be best to extract this into a separate PR.
 * Use goreleaser: We need to clarify what benefits this has. Users building restic themselves do not need to use it, they can just run `go build ./cmd/restic` and be done (or just call `make`, we need to update the Makefile for that when we raise the Go version and remove `build.go`). At the moment, we use our own set of programs in the `helpers/` dir, the programs ensure that I don't forget anything for a release. There are a lot of small [things to do](https://github.com/restic/restic/blob/master/helpers/prepare-release/main.go#L414-L461), like:
   - Make sure there are no uncommitted changes
   - Check that the right version numbers are in the `VERSION` file (used by `build.go`) as well as in the source
   - Check that the `builder` Docker container has the correct Go version installed
   - Check that the release has a subdir in `changelog/` with the changes
   - Generate the changelog and manpages
   - Add a signed Git tag for the version
   - Create the source `.tar.gz`, build the binaries in a reproducible way
   - Create `SHA256SUMS`, sign it with my personal GPG key
   - Update the restic Docker container
   - Create a new Git commit with the `-dev` version suffix
  Apart from all these things, the release is signed with my GPG key, so at the moment I'm the only one who can make a release. We can transition the authenticity verification to something else (e.g. Minisign or even a different GPG key), but that's a rather medium-term issue. What benefits does using `goreleaser` bring?
 * Rework the `Makefile`: First, I don't like `make` for various reasons, apart from *maybe* compiling C code. We provide a very basic one so that people can just run `make` and get a compiled version of restic, but that's also mostly for backwards-compatibility. I'm ready to drop it.
 * Remove `vendor` folder: I've created a new issue (#2482) so we can discuss this a bit, it also needs to be done in a separate PR and affects the CI integration (#2475)

fd0:
I'm closing this stale PR, please feel free to add further comments. Thanks!

PR #2486, opened 2019-11-22
changelog: Add .gitkeep to persist unreleased/ folder

### LINKED ISSUE #2485: Folder changelog/unreleased/ doesn't exist when empty
When here are no pending changeling entries in the unreleased/ folder, the folder isn't tracked by git anymore. Add a .gitkeep file to the changelog/unreleased/ folder in order to make the folder always exist. This makes it less confusing for new contributors when the contributing instructions reference this folder.

rawtaz:

It adds a .gitkeep file (conventional name, everyone and their two cats does it) to the changeling/unreleased/ folder so that it is still tracked and kept around when it's empty.

Previously when there were no pending changeling files in this folder, it wasn't tracked by Git, which could be confusing to new contributors considering the folder is referenced by the contributing instructions.

Closes #2485.

Checklist
---------


fd0:
Two things:
 * We need to make sure the empty file is created again
 * We need to teach `calens` to ignore files starting with a dot

rawtaz:
Can you teach calens to ignore files starting with dot and to only move the *files* in the `unreleased` folder, instead of the folder itself?

fd0:
Ah, that's done manually, like `mv changelog/unreleased changelog/v1.0.0`, but maybe it's a good idea to let `calens` do that. But first, I'll build ignoring dot-prefix files.

fd0 (approved):
LGTM, calens is fixed

PR #2505, opened 2019-12-08
Fix repo configfile

### LINKED ISSUE #2498: repository implementation: writing/reading config file from backend is inconsistent
The backend-interface works as follows:

- Files are accessed by a Handle
- the Handle consists of a Type of the file and the ID
- usually the ID should be equal to the SHA-256 Hash of the content

For the config file, there is an exception which is implemented for each backend and also in the repository implementation.
For instance in internal/repository/repository.go func LoadAndDecrypt the Hash is not checked for ConfigFile. This is necessary as the config file is accessed without knowing the ID (i.e. with zero ID).

But in fact, in the same repository implementation in func SaveUnpacked the  ID is computed as Hash of the content.

So the backends get a Save request for (ConfigFile, ID) with the correct Hash ID and a Read request for (ConfigFile, ID{}) (empty ID).

Luckily all backends handle ConfigFile requests as special case and hence this is no real bug ATM.
However I consider this a inconsistent design and suggest to correct it.

I'm about to file a pull request and just want to discuss it in advance.

aawsome:
access to config file from repository was inconsistent and therefore unsafe for new possible backends. This is fixed in this PR.
See #2498. 

closes #2498 

Checklist
---------

- [x] I have not added documentation for the changes, as I see no need to document it for users.
- There's no new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))  - dont't think this is necessary for this minor change.

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -82,6 +82,10 @@ func (r *Repository) LoadAndDecrypt(ctx context.Context, buf []byte, t restic.Fi
 
 	debug.Log("load %v with id %v", t, id)
 
+	if t == restic.ConfigFile {
+		id = restic.ID{}
+	}
+
 	h := restic.Handle{Type: t, Name: id.String()}
 	err := r.be.Load(ctx, h, 0, 0, func(rd io.Reader) error {
 		// make sure this call is idempotent, in case an error occurs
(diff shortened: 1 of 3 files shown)


changelog/unreleased/issue-2498
```diff
@@ -0,0 +1,5 @@
+Change: Fix inconsistent acces to config file
```
greatroar: Typo: acces. Also inconsitent, below.

aawsome: I removed the changelog for this minor change.

fd0 (approved):
LGTM, thanks!

Issue #2162, opened 2019-02-04
handle large prune much more efficent


neeral85:
Output of `restic version`
--------------------------

restic 0.9.3 compiled with go1.10.4 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

In general, I like restic very much, and creating/recovery snapshots works perfectly fine.
But run restic with large repository’s is almost impossible. I have a repository with 5 TB / 30 snapshots.
Intention was to do this like a circular buffer (remove oldest, add newest). 

Adding a snapshot and removing them works perfect, but as you necessarily come to prune your repository it can take WEEKS to just free 1 TB (because of rewrite).
This makes it almost impossible to use restic anymore as you can't create new snapshots during that time.

As you mentioned already [here](https://github.com/restic/restic/issues/725#issuecomment-277277794)
you may do something to improve this.

Example: 
found 5967884 of 7336415 data blobs still in use, removing 1368531 blobs
will delete 144850 packs and rewrite 142751 packs, this frees 1.082 TiB (took 2 weeks!)

Especially on remote repository’s where you just bought storage (with ssh access) and CPU resources are limited its much faster upload the whole repository again.

pmkane:
With performance for restores for large repos/repos with large files now resolved by @ifedorenko's out-of-order branch, it looks like this is the next boulder for using restic in multi-terabyte environments where the repo isn't on local disk and isn't being accessed via the REST server on a loopback interface.

Currently, an empty prune (pruning a snapshot that was 100% identical to a previous snapshot) against a repo in an AZ-local S3 bucket on high-end AWS instances with 10gbit/sec theoretical bandwidth runs at:

1) building new index -- ~160 packs/second
2) finding data still in use -- 56 seconds total
3) rewriting packs -- ~3 packs/second

160 packs/second is slow, but still tolerable (~80 minute run-time against a 3TB repo).

But the rewrite @ 3 packs/second will run for almost 10 hours, even for my noop prune (will delete 0 packs and rewrite 111098 packs, this frees 180.699 GiB).  For a meaningful prune on a large repo, you freeze out new backups for 24+ hours.

It looks like the pack rewrites currently happen in a single-thread, so allowing that to run across multiple workers might help quite a bit, even if the current copy then purge approach is maintained.

ifedorenko:
Personally, I would not spend time optimizing current blocking prune implementation, I think non-blocking prune is better long-term solution.

cbane:
[...]

PR #2507, opened 2019-12-09
Optimize index

aawsome:

Optimize creation of new index so that already existing index files are used.
Creation of new index is used at the moment in prune and rebuild-index.

With this pull request the index creation is much faster, especially for remote repositories.

The change is similar to one part of PR #2340 and the problem is discussed in many issues, e.g. #2162 

Checklist
---------

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -85,7 +85,10 @@ func New(ctx context.Context, repo ListLoader, ignorePacks restic.IDSet, p *rest
 			}
 
 			if res, ok := oldIdx.Packs[id]; ok {
-				if err := idx.AddPack(res.ID, res.Size, res.Entries); err != nil {
+				if err := idx.AddPack(res.ID, size, res.Entries); err != nil {
+					return err
+				}
+				if err:= oldIdx.RemovePack(id); err != nil {
 					return err
 				}
 				return nil
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -84,10 +84,13 @@ func New(ctx context.Context, repo ListLoader, ignorePacks restic.IDSet, p *rest
 				return nil
 			}
 
-			if res, ok := oldIdx.Packs[id]; ok {
-				if err := idx.AddPack(res.ID, size, res.Entries); err != nil {
-					return err
+			if old, ok := oldIdx.Packs[id]; ok {
+				select {
+				case outputCh <- Result{PackID: old.ID, Entries: old.Entries, Size: old.Size}:
+				case <-ctx.Done():
+					return nil
 				}
+
 				if err := oldIdx.RemovePack(id); err != nil {
 					return err
 				}
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -271,13 +271,11 @@ func Load(ctx context.Context, repo ListLoader, p *restic.Progress) (*Index, err
 	return index, nil
 }
 
-// AddPack adds a pack to the index. If this pack is already in the index, an
-// error is returned.
+// AddPack adds a pack to the index.
 func (idx *Index) AddPack(id restic.ID, size int64, entries []restic.Blob) error {
 	if _, ok := idx.Packs[id]; ok {
-		return errors.Errorf("pack %v already present in the index", id.Str())
+		debug.Log("pack %v already present in the index", id.Str())
 	}
-
 	idx.Packs[id] = Pack{ID: id, Size: size, Entries: entries}
 
 	return nil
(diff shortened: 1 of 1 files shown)


aawsome:
Seems like the original PR broke some tests with packs in more than one index. This should be fixed now.

aawsome:
While still being confident that this PR can improve the current situation for some users, I now believe that on the long run a refactoring of `rebuild-index` and `prune` (and thus removing `internal/index`) should be favored.

irasnyd:
Hi @aawsome,

I used this PR in conjunction with #2513 during my prune of a large (12M object / ~55TB) AWS S3 backed repository. I don't know how to tell if it was faster or slower than normal, since this was my first prune of this very large repository. As far as I can tell, it worked great. Backups have continued normally without any issue since the repository was pruned (almost 2 weeks ago).

It isn't much feedback, but it's the best I can do. I wanted to give something back, since I had used and tested this PR. Thanks!

aawsome:
@irasnyd 

> I used this PR in conjunction with #2513 during my prune of a large (12M object / ~55TB) AWS S3 backed repository.

Just to clarify: As far as I understood you used `cleanup-index` and `cleanup-packs` from #2513, right?
In this case this PR was not used at all. These two commands do not use `internal/index` as there is already an index data structure in `repository/index` using both data structure multiplies the memory usage and is one of the reasons why `prune` has a memory issue (besides the performance issue due to reading all packs)

irasnyd:
> @irasnyd
> 
> > I used this PR in conjunction with #2513 during my prune of a large (12M object / ~55TB) AWS S3 backed repository.
> 
> Just to clarify: As far as I understood you used `cleanup-index` and `cleanup-packs` from #2513 instead of `prune`, right?

Yes, that is correct. I only used the `cleanup-index` and `cleanup-packs` commands from #2513. I did **not** use `prune`.

> In this case this PR was not used at all. These two commands do not use `internal/index` as there is already an index data structure in `repository/index`. Using both data structure multiplies the memory usage and is one of the reasons why `prune` has a memory issue (besides the performance issue due to reading all packs which can be improved by using this PR)

Apologies for my confusion. I understand better what is going on now. Thanks!

internal/index/index.go
```diff
@@ -76,6 +84,16 @@ func New(ctx context.Context, repo Lister, ignorePacks restic.IDSet, p *restic.P
 				return nil
 			}
 
+			if res, ok := oldIdx.Packs[id]; ok {
```
MichaelEischer: This looks like a race condition with the `for res := range outputCh` loop. Two coroutines are working on `idx` and I don't see any synchronization in the AddPack method.

Replacing `repo.ListPack` with a lookup into the old index should avoid that race condition.

internal/index/index.go
```diff
@@ -76,6 +84,16 @@ func New(ctx context.Context, repo Lister, ignorePacks restic.IDSet, p *restic.P
...
+				if err := idx.AddPack(res.ID, size, res.Entries); err != nil {
+					return err
+				}
+				if err := oldIdx.RemovePack(id); err != nil {
```
MichaelEischer: Why do you remove the old pack? Is this done to reduce memory usage?

internal/index/index.go
```diff
@@ -255,13 +271,11 @@ func Load(ctx context.Context, repo ListLoader, p *restic.Progress) (*Index, err
 	return index, nil
 }
 
-// AddPack adds a pack to the index. If this pack is already in the index, an
-// error is returned.
+// AddPack adds a pack to the index.
 func (idx *Index) AddPack(id restic.ID, size int64, entries []restic.Blob) error {
 	if _, ok := idx.Packs[id]; ok {
-		return errors.Errorf("pack %v already present in the index", id.Str())
+		debug.Log("pack %v already present in the index", id.Str())
```
MichaelEischer: What is the reason for this change? According to my understand of the code it just replaces loading a pack with reusing the old index for that pack. This shouldn't lead to any pack being added twice.

internal/index/index.go
MichaelEischer: Or better: Just send the pack from the oldIdx over the `outputCh` channel

internal/index/index.go
aawsome: Yes that was what I was thinking. I actually don't know if the GC is really freeing something in practise, though...

internal/index/index.go
aawsome: It is not the call in `New` this change is needed for but the call in `Load`. Seems there are some integration tests with index where duplicate packs are present in the index (either within one index file or over different index files)...

aawsome:
I close this PR for two reasons:

- with #2513 there is a functionality to replace `prune` which I assume to be even faster than prune with this functionality. Moreover, this PR is not needed for #2513 
- in #2523 I already proposed a new index implementation which IMO should be enhanced such that it can also be used within #2513 and then 'prune' can be replaced.

Issue #2504, opened 2019-12-07
restic with "cold" storage (here: OVH Cloud Archive)


aawsome:
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on linux/amd64

How did you run restic exactly?
-------------------------------
`restic -r rclone:ovh:test2 -p key-restic snapshots`

Output is:
```
Load(<key/1bd3e4429a>, 0, 0) returned error, retrying after 720.254544ms: <key/1bd3e4429a> does not exist
```
repeated a couple of times with changing retry times and then:
`Fatal: <key/1bd3e4429a> does not exist`

When using
`RCLONE_VERBOSE=2 restic -r rclone:ovh:test2 -p key-restic snapshots`
you see:
```
rclone: 2019/12/07 12:40:04 DEBUG : Swift container test2: GET /file-5577006791947779410
rclone: 2019/12/07 12:40:04 DEBUG : file-5577006791947779410: GET request error: object not found
rclone: 2019/12/07 12:40:04 DEBUG : Swift container test2: HEAD /config
rclone: 2019/12/07 12:40:05 DEBUG : Swift container test2: GET /keys/
rclone: 2019/12/07 12:40:05 DEBUG : keys: list request
rclone: 2019/12/07 12:40:06 DEBUG : Swift container test2: GET /keys/1bd3e4429a441fd7546efdfa1bc7c41333d548b0a1ecb5dea3fd1f9ef16c205f
rclone: 2019/12/07 12:40:07 DEBUG : 1bd3e4429a441fd7546efdfa1bc7c41333d548b0a1ecb5dea3fd1f9ef16c205f: Get request open error: try again after 2019-12-07T12:50:07.317400358+01:00 (9m59.999921857s)
Load(<key/1bd3e4429a>, 0, 0) returned error, retrying after 720.254544ms: <key/1bd3e4429a> does not exist
```

Similar behavior with other restic commands, like `backup` etc.

If you wait long enough and redo the command, restic works fine. However, the issue occurs with

- key files
- config file
- lock files

Therefore you might have to repeat and wait quite often before the requested operation is performed.

Another issue is, that the above files are not cached and therefore once you manage to get these files "warm" in the backend, you can do your operations. But if they get "cold" again, you again come in the original mentioned issue. 

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
rclone to access OVH via swift.
The rclone config is directly as provided by OVH.
[...]

PR #2515, opened 2019-12-15
Add --retries flag

### LINKED ISSUE #3223: Limit number of retries especially for forget on append-only servers
Output of `restic version`
--------------------------

restic 0.9.6 compiled with go1.12.12 on linux/amd64

A restic command like `backup` or `forget` should have an option like `--max-retries` or `--max-retry-count` after which it will stop retrying the activity.

When using restic with a restic rest-server backend, the server can be configured to be "append-only", i.e. backup is allowed but forget is not. If I then try `restic forget`, I get:

```
repository … opened successfully, password is correct
Remove(<snapshot/703fc2e649>) returned error, retrying after 409.996243ms: blob not removed, server response: 403 Forbidden (403)
Remove(<snapshot/703fc2e649>) returned error, retrying after 907.654304ms: blob not removed, server response: 403 Forbidden (403)
…
```
With the proposed option, the command would fail after n retries and not try any longer.

Restic is awesome! I really is backups done right: encrypted, deduplicated, extremely fast, so many backends, open source, wow. The speed and small size really impresses me. I had to double check if a new snapshot really contained all the large files, I changed.

aawsome:

Adds the new flag --retries to set the maximum number of retries for file operations.
If the number of retires is set higher than 20, restic will print out a warning.

See issue #2504, first open point.
Closes #3223 

Checklist
---------

- I have added no documentation for the changes (in the manual)

### DIFF
--- a/changelog/unreleased/issue-2504
+++ b/changelog/unreleased/issue-2504
@@ -1,0 +1,8 @@
+Enhancement: Add --retries flag
+
+Restic used to retry backend operation 10 times before cancelling.
+This causes problems for 'cold storage' backends where you can have a large retrieve time.
+With the new flag you can finetune the maximum number of retries.
+
+https://github.com/restic/restic/issues/2504
+https://github.com/restic/restic/pull/2515
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -59,6 +59,7 @@ type GlobalOptions struct {
 	CACerts         []string
 	TLSClientCert   string
 	CleanupCache    bool
+	Retries         uint
 
 	LimitUploadKb   int
 	LimitDownloadKb int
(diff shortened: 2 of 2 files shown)


aawsome:
Is there anything that blocks this small change from getting merged?

MichaelEischer:
I'm not really sure about the design. Is it really the right way to blindly retry an operation lots of times?

One problem I see is that with infinite retries the shutdown of restic won't work properly in case of errors: e.g. removing a lock file might retry forever preventing restic from exiting. A similar behavior shows for other operations that can fail permanently.

aawsome:
> I'm not really sure about the design. Is it really the right way to blindly retry an operation lots of times?

I do not question whether it is a good design to often retry or not. However, in the actual implementation restic retries 10 times and a specific total amount of time before giving up. As timings always can depend on the environment used, I would prefer to have user-definable setting with a sensible default value.

> One problem I see is that with infinite retries the shutdown of restic won't work properly in case of errors: e.g. removing a lock file might retry forever preventing restic from exiting. A similar behavior shows for other operations that can fail permanently.

I agree that infinite trying is a bad idea. I propose the following:
- only allow positive values to --retry
- if the number is too large (e.g. > 20), print out a warning that restic may have to wait a long time to really quit in case of errors.

aawsome:
>     * only allow positive values to --retry
> 
>     * if the number is too large (e.g. > 20), print out a warning that restic may have to wait a long time to really quit in case of errors.

This is now implemented.

aawsome:
The warning message is now corrected.

PR #2516, opened 2019-12-15
Add flag --cache-all

aawsome:
Adds the flag --cache-all
When set, all files (including key, config, lock) are cached.
To do so, when the flag is set, the cache directory does not use repo ID
but the repo string given by -r.

Note: In order to correctly cache the config file, the PR #2505 is also required!

See issue #2504. 

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -59,6 +59,7 @@ type GlobalOptions struct {
 	JSON            bool
 	CacheDir        string
 	NoCache         bool
+	CacheAll        bool
 	CACerts         []string
 	TLSClientCert   string
 	CleanupCache    bool
@@ -83,6 +84,10 @@ type GlobalOptions struct {
 	Options []string
 
 	extended options.Options
+}
+
+func (gopts GlobalOptions) RepoHash() string {
+	return "repo-" + restic.Hash([]byte(gopts.Repo)).String()
 }
 
 var globalOptions = GlobalOptions{
@@ -111,6 +116,7 @@ func init() {
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repository, this allows some operations on read-only repositories")
 	f.BoolVarP(&globalOptions.JSON, "json", "", false, "set output mode to JSON for commands that support it")
 	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache `directory`. (default: use system default cache directory)")
+	f.BoolVar(&globalOptions.CacheAll, "cache-all", false, "also cache key, config and lock files (uses repo string to identify cache)")
 	f.BoolVar(&globalOptions.NoCache, "no-cache", false, "do not use a local cache")
 	f.StringSliceVar(&globalOptions.CACerts, "cacert", nil, "`file` to load root certificates from (default: use system certificates)")
 	f.StringVar(&globalOptions.TLSClientCert, "tls-client-cert", "", "path to a `file` containing PEM encoded TLS client certificate and private key")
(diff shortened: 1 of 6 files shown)


cmd/restic/global.go
```diff
@@ -365,6 +367,21 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 	})
 
 	s := repository.New(be)
+	var c *cache.Cache
+
+	if !opts.NoCache && opts.CacheAll {
+		c, err = cache.New(opts.Repo, opts.CacheDir, cache.LayoutAll)
```
MichaelEischer: I'm a bit worried about special characters in the repository path, that could cause trouble. I'd suggest to hash the repository path and add an prefix before using it as cache id.

greatroar: Yes, paths are very likely to contain [/:]. They can also contain passwords (cf. #2617).

aawsome:
@MichaelEischer @greatroar  Thank you for your comments.

Actually I'm moving a bit towards the following  re-implementation of the `Cache` structure:

- use a `Backend` to actually store the Cached objects
- make it possible to wrap multiple Caches into a `CachedBackend` which is finally used by restic.

This approach simplifies the caching code base and would allow much more options (and hence obsoletes this PR) like 2 layers of caches where one layer (caching all) could be on local disc but also on a remote storage.

It's still WIP; if you are interessted see https://github.com/aawsome/restic/tree/cache-as-repo

MichaelEischer:
I haven't thought that much about which functionality a cache should provide. It's probably something like the normal backend functionality plus features for cache invalidation. Sooner or later restic will need a method to remove files from the cache which were apparently damaged. Performancewise there is a difference between the cache implementation and the local backend, the former does not use fsync, while the latter does. And your cache backend will need some sort of size limiting, otherwise it will just end up as another copy of the repository, which wouldn't make that much sense.

aawsome:
> Performancewise there is a difference between the cache implementation and the local backend, the former does not use fsync, while the latter does. 

Thanks for that hint! I wasn't aware of this.

> And your cache backend will need some sort of size limiting, otherwise it will just end up as another copy of the repository, which wouldn't make that much sense.

There is already a very effective sizing limit by not storing data blobs in the cache. In basically all real-life repos I've seen, the size of data blobs was more than 99% of the repo size. (see e.g. #2543)

cmd/restic/global.go
aawsome: I changed this to use `"repo-"+sha256(opts.Repo)`.
Also added these new cache dirs to the `restic cache` command.

fd0:
Is it a good idea to add this feature? What's the use case?

I was always very reluctant to cache the config file and the keys for security reasons...

aawsome:
> Is it a good idea to add this feature? What's the use case?

I'm using it in combination with my cold storage repos. But I wanted to open a new issue to openly discuss the way how to support cold storage with restic. Then we can openly discuss the best way how to implement this and discuss a roadmap.

I just updated this as I'm using it now for over a year quite successfully. But it's still marked as draft as I think we should discuss the way first.
 
> I was always very reluctant to cache the config file and the keys for security reasons...

Mhh... we have a zero trust model about the storage backend where those files are actually stored. And we assume that we can trust the environment where restic acutally runs. So from a security point of view, I actually see no issues :smiley:

rawtaz:
Using the repo string as cache directory name can't be great in the long run. Imagine if the hostname or some other part of the repo URL changes, the cache directory name will have to be changed as well. There's a reason the current design is using the non-changing repo ID as the name :)

aawsome:
closing this PR as #3235 IMO is more general and solves the same problem.

Issue #2128, opened 2018-12-28
Key metadata is stored unencrypted


nioncode:
Output of `restic version`
--------------------------
restic 0.9.2 compiled with go1.11 on linux/amd64

How did you run restic exactly?
-------------------------------
restic -r /repo init

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
local

Expected behavior
-----------------

Keys should contain no plaintext information (apart from salt, iteration, ...).

Actual behavior
---------------
Keys contain host name, username, and time of creation in plaintext.

Do you have any idea what may have caused this?
-----------------------------------------------
Restic stores key metadata unencrypted.

Do you have an idea how to solve the issue?
-------------------------------------------
Store key metadata encrypted. We could update the key format by adding e.g. a `hostname_enc` field and blanking the `hostname` field for new repos. Then old clients would show empty metadata until they are updated. We should also migrate existing keys to remove potentially old plaintext metadata from existing repos.

fd0:
We'll probably just remove the fields from the key metadata, that should be sufficient.

nioncode:
How are keys identified then, if all their metadata gets removed? Or do you just want to remove the plaintext fields and introduce encrypted ones?

fd0:
Yes, through the key ID (the filename prefix).

cdhowie:
Perhaps have this be optional. We use a key per backup host and it's nice to be able to quickly see which key belongs to which host.  Without this data in restic, we'd have to maintain the data somewhere else (company wiki, for example) -- not the worst thing, but why fix what isn't broken?

nioncode:
I agree that it's nice to have, but this leaks metadata unencrypted. That's why I proposed to have them stored encrypted, so that each key that can decrypt the repo can still read all metadata like it is now, but the not trusted backup location can not gather the metadata in plaintext. This way, there would be no user facing changes at all.

cdhowie:
> there would be no user facing changes at all.

That depends if the repository is automatically migrated to store the data encrypted.  If so, the key IDs will necessarily change and, depending on syncing processes/restrictions, the user may need to be aware that this happened.

monkeyhybrid:
If this metadata is to be removed, could we at least have an optional 'tag' attribute for each key that can contain anything we wish? Make the tag empty by default for those concerned with leaking metadata, but allow custom text that could be useful for key management, for those that are not.

ducalex:
When stripping the metadata manually I get the following:

````
[...]

PR #2525, opened 2019-12-22
Add new flag masterkeyfile

aawsome:

Add the new flag `--masterkeyfile` which allows to directly use the
master key from a file.
When `--masterkeyfile` is used from `restic init` the master key is
written to the given file and no key is created in the repo.
For all other commands, `--masterkeyfile` uses the master key in the
given file. No password is needed then.

Was discussed (don't remember the issue) and is able to solve #2128 
Also allows increased security if you do want to save the password-encrypted masterkey.
On low-end machines saving the overhead from the KDF may give you some benefit.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_init.go
+++ b/cmd/restic/cmd_init.go
@@ -33,25 +33,32 @@ func runInit(gopts GlobalOptions, args []string) error {
 		return errors.Fatalf("create repository at %s failed: %v\n", gopts.Repo, err)
 	}
 
-	gopts.password, err = ReadPasswordTwice(gopts,
-		"enter password for new repository: ",
-		"enter password again: ")
-	if err != nil {
-		return err
+	if gopts.MasterKeyFile == "" {
+		gopts.password, err = ReadPasswordTwice(gopts,
+			"enter password for new repository: ",
+			"enter password again: ")
+		if err != nil {
+			return err
+		}
 	}
 
 	s := repository.New(be)
 
-	err = s.Init(gopts.ctx, gopts.password)
+	err = s.Init(gopts.ctx, gopts.MasterKeyFile, gopts.password)
 	if err != nil {
 		return errors.Fatalf("create key in repository at %s failed: %v\n", gopts.Repo, err)
 	}
 
 	Verbosef("created restic repository %v at %s\n", s.Config().ID[:10], gopts.Repo)
 	Verbosef("\n")
-	Verbosef("Please note that knowledge of your password is required to access\n")
-	Verbosef("the repository. Losing your password means that your data is\n")
-	Verbosef("irrecoverably lost.\n")
+	if gopts.MasterKeyFile == "" {
+		Verbosef("Please note that knowledge of your password is required to access\n")
+		Verbosef("the repository. Losing your password means that your data is\n")
+		Verbosef("irrecoverably lost.\n")
+	} else {
+		Verbosef("Please note that you need the masterkey %s to access the repository\n", gopts.MasterKeyFile)
+		Verbosef("Losing your masterkey file means that your data is irrecoverably lost.\n")
+	}
 
 	return nil
 }
--- a/cmd/restic/cmd_init.go
+++ b/cmd/restic/cmd_init.go
@@ -56,7 +56,7 @@ func runInit(gopts GlobalOptions, args []string) error {
 		Verbosef("the repository. Losing your password means that your data is\n")
 		Verbosef("irrecoverably lost.\n")
 	} else {
-		Verbosef("Please note that you need the masterkey %s to access the repository\n", gopts.MasterKeyFile)
+		Verbosef("Please note that you need the masterkey %s to access the repository.\n", gopts.MasterKeyFile)
 		Verbosef("Losing your masterkey file means that your data is irrecoverably lost.\n")
 	}
 
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -351,6 +351,41 @@ func ReadPasswordTwice(gopts GlobalOptions, prompt1, prompt2 string) (string, er
 
 const maxKeys = 20
 
+func tryPassword(opts GlobalOptions, s *repository.Repository) error {
+	var err error
+
+	passwordTriesLeft := 1
+	if stdinIsTerminal() && opts.password == "" {
+		passwordTriesLeft = 3
+	}
+
+	for ; passwordTriesLeft > 0; passwordTriesLeft-- {
+		opts.password, err = ReadPassword(opts, "enter password for repository: ")
+		if err != nil && passwordTriesLeft > 1 {
+			opts.password = ""
+			fmt.Printf("%s. Try again\n", err)
+		}
+		if err != nil {
+			continue
+		}
+
+		err = s.SearchKey(opts.ctx, opts.password, maxKeys, opts.KeyHint)
+		if err == nil {
+			if !opts.JSON {
+				Verbosef("password is correct\n")
+			}
+			break
+		}
+
+		if err != nil && passwordTriesLeft > 1 {
+			opts.password = ""
+			fmt.Printf("%s. Try again\n", err)
+		}
+	}
+
+	return err
+}
+
 // OpenRepository reads the password and opens the repository.
 func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 	if opts.Repo == "" {
@@ -369,35 +404,7 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 	s := repository.New(be)
 
 	if opts.MasterKeyFile == "" {
-
-		passwordTriesLeft := 1
-		if stdinIsTerminal() && opts.password == "" {
-			passwordTriesLeft = 3
-		}
-
-		for ; passwordTriesLeft > 0; passwordTriesLeft-- {
-			opts.password, err = ReadPassword(opts, "enter password for repository: ")
-			if err != nil && passwordTriesLeft > 1 {
-				opts.password = ""
-				fmt.Printf("%s. Try again\n", err)
-			}
-			if err != nil {
-				continue
-			}
-
-			err = s.SearchKey(opts.ctx, opts.password, maxKeys, opts.KeyHint)
-			if err == nil {
-				if !opts.JSON {
-					Verbosef("password is correct\n")
-				}
-				break
-			}
-
-			if err != nil && passwordTriesLeft > 1 {
-				opts.password = ""
-				fmt.Printf("%s. Try again\n", err)
-			}
-		}
+		err = tryPassword(opts, s)
 	} else {
 		err = s.MasterKeyFile(opts.ctx, opts.MasterKeyFile)
 	}
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -49,6 +49,7 @@ type GlobalOptions struct {
 	Repo            string
 	PasswordFile    string
 	PasswordCommand string
+	MasterKeyFile   string
 	KeyHint         string
 	Quiet           bool
 	Verbose         int
@@ -97,6 +98,7 @@ func init() {
 	f.StringVarP(&globalOptions.Repo, "repo", "r", os.Getenv("RESTIC_REPOSITORY"), "repository to backup to or restore from (default: $RESTIC_REPOSITORY)")
 	f.StringVarP(&globalOptions.PasswordFile, "password-file", "p", os.Getenv("RESTIC_PASSWORD_FILE"), "read the repository password from a file (default: $RESTIC_PASSWORD_FILE)")
 	f.StringVarP(&globalOptions.KeyHint, "key-hint", "", os.Getenv("RESTIC_KEY_HINT"), "key ID of key to try decrypting first (default: $RESTIC_KEY_HINT)")
+	f.StringVarP(&globalOptions.MasterKeyFile, "masterkeyfile", "", os.Getenv("RESTIC_MASTERKEY_FILE"), "use masterkey from file (default: $RESTIC_MASTERKEY_FILE)")
 	f.StringVarP(&globalOptions.PasswordCommand, "password-command", "", os.Getenv("RESTIC_PASSWORD_COMMAND"), "specify a shell command to obtain a password (default: $RESTIC_PASSWORD_COMMAND)")
 	f.BoolVarP(&globalOptions.Quiet, "quiet", "q", false, "do not output comprehensive progress report")
 	f.CountVarP(&globalOptions.Verbose, "verbose", "v", "be verbose (specify --verbose multiple times or level `n`)")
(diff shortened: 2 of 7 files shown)


cmd/restic/cmd_init.go
```diff
@@ -33,25 +33,32 @@ func runInit(gopts GlobalOptions, args []string) error {
...
 	Verbosef("\n")
-	Verbosef("Please note that knowledge of your password is required to access\n")
-	Verbosef("the repository. Losing your password means that your data is\n")
-	Verbosef("irrecoverably lost.\n")
+	if gopts.MasterKeyFile == "" {
+		Verbosef("Please note that knowledge of your password is required to access\n")
+		Verbosef("the repository. Losing your password means that your data is\n")
+		Verbosef("irrecoverably lost.\n")
+	} else {
+		Verbosef("Please note that you need the masterkey %s to access the repository\n", gopts.MasterKeyFile)
```
shtratos: [nit] Dot is missing at the end of the line.

cmd/restic/global.go
```diff
@@ -366,26 +368,38 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
...
+					Verbosef("password is correct\n")
+				}
+				break
+			}
+
+			if err != nil && passwordTriesLeft > 1 {
+				opts.password = ""
+				fmt.Printf("%s. Try again\n", err)
+			}
 		}
```
shtratos: this whole block is rather long, please consider extracting it into a separate function or method.

doc/070_encryption.rst
```diff
@@ -49,3 +49,51 @@ per repository. In fact, you can use the ``list``, ``add``, ``remove``, and
...
+    enter password for new repository: 
+    enter password again: 
+    created restic repository a8e7f96567 at /srv/restic-repo
+
+    Please note that knowledge of your password is required to access
+    the repository. Losing your password means that your data is
+    irrecoverably lost.
+    $ restic -r  /srv/restic-repo --quiet cat masterkey > key.master 
+    enter password for repository:
+    $ chmod 400 key.master
```
shtratos: [paranoia] I'd suggest first creating the `key.master` file, chmodding it to `600`, then exporting the key and chmodding it to `400`.
That way it can not be read by other users during the time between the export and `chmod 400`.

internal/repository/repository.go
```diff
@@ -568,6 +569,33 @@ func LoadIndex(ctx context.Context, repo restic.Repository, id restic.ID) (*Inde
 	return nil, err
 }
 
+// MasterKeyFile reads the master key from a given file
+func (r *Repository) MasterKeyFile(ctx context.Context, keyfile string) error {
+
+	buf, err := ioutil.ReadFile(keyfile) // just pass the file name
```
shtratos: [paranoia] should we check `opts.MasterKeyFile` mode and refuse to accept it unless it's `600` or `400`  like SSH does?
Maybe also warn the user that the key might have been compromised?

shtratos: [nit] not sure what is the value of the `// just pass the file name` comment. Is it needed?

doc/070_encryption.rst
```diff
@@ -49,3 +49,51 @@ per repository. In fact, you can use the ``list``, ``add``, ``remove``, and
...
+in the repository. Make sure your masterkey cannot be accessed by unauthorized persons and that you  have a backup
+of you masterkey file ready (e.g. print-out)!
+
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo --masterkeyfile master.key init
+    created restic repository f855d38126 at /tmp/repo
+
+    Please note that you need the masterkey tmp.master to access the repository
```
shtratos: [typo?] Did you mean `master.key` instead of `tmp.master` here?

doc/070_encryption.rst
```diff
@@ -49,3 +49,51 @@ per repository. In fact, you can use the ``list``, ``add``, ``remove``, and
...
+    $ restic -r  /srv/restic-repo --quiet cat masterkey > key.master 
+    enter password for repository:
+    $ chmod 400 key.master
+    $ restic -r /srv/restic-repo --masterkeyfile key.master key list
+    repository a8e7f965 opened successfully
+    ID        User        Host      Created
+    ---------------------------------------------------
+    335ca24b  username    kasimir   2015-08-12 13:35:05
+    ---------------------------------------------------
+    $ restic -r /srv/restic-repo --masterkeyfile key.master key remove 335ca24b
```
shtratos: should removing a master key from the repository warn a user that the repository can't be accessed without the master key file?

shtratos (commented):
Thanks for having a stab at this! 

I'm a bit confused on what is the exact value of having the masterkey file stored outside the repository.
Could you please clarify it in the documentation?

As a restic user, it's hard for me to make a judgement which option is "better" or "more secure".
Is there a tradeoff? What is it? Which one is the recommended option?
Is the masterkey file encrypted?

cmd/restic/cmd_init.go
aawsome: done.

doc/070_encryption.rst
aawsome: Changed it to `restic ... cat ... && chmod`.

internal/repository/repository.go
aawsome: Not really sure if we should do paranoia here.. Do others have an opinion about it? If more are voting for a permission check, I'll add it.

Comment is removed - was a copy/paste error

doc/070_encryption.rst
aawsome: is now corrected. Thanks!

doc/070_encryption.rst
aawsome: I added a general warning when deleting repo keys using masterkey file.

cmd/restic/global.go
aawsome: Thanks for the suggestion! I also changed it.

fd0:

I'm going to be honest here: in my opinion, the complexity added by this feature outweights its usefulness. And #2128 can be addressed in different ways.

I couldn't find any discussion about this. Do you still remember where it was discussed?

aawsome:
Thanks @fd0 for your feedback!

As written I do not really remember where the discussion was and having searched for it I couldn't find it. I recall there was some about the possibility to just use a key instead of password but maybe my memory was misleading me.
For me this PR is absolutely useful. As for backup security is a prio one matter, I do think we should try to offer a variety of options to satisfy the user's need of security. I agree that we should think about options which enable the user to drop security. IMO this PR allows to raise security when considering some threat options like possibility of reading repos and a weakness in the KDF allowing brute-force attacks.

Personally when starting with restic I was puzzled that there was no option to just use a key as most other backups tools do offer. As a result I only use a generated password-file with enough entropy for my one-key repository. Adding the masterkeyfile option would make my setting even more secure.

About the complexity of the PR:
I do not think that adding the masterkeyfile option does add a significant complexity to the code. Can you specify which parts of the PR seems to be too complex for you?
From my point of view the majority of the changes are due to documentation, hints for users and the ability to create a masterkeyfile during repo init. I consider all of this optional and can remove it if you like. I'm also open to suggestions how to lower complexity in the few remaining parts.
Thank you!

rawtaz:
@aawsome So put in simple terms, the use case would be to sidestep restic's own encryption and directly provide a master key such that you can use your own encryption algorithms, for the purpose of when you don't trust the encryption restic uses?

aawsome:
> @aawsome So put in simple terms, the use case would be to sidestep restic's own encryption and directly provide a master key such that you can use your own encryption algorithms, for the purpose of when you don't trust the encryption restic uses?

No, this does not describe the use case correctly.

When talking about encryption of a backup tool, I would consider the encryption used to encrypt the data backed up and the cryptography used for authentication. Restic's encryption hence is AES-256 and the authentication algorithm is POLY-1305; this is not changed by own encryption/authentication when using this PR.

However when talking about the keys used for the AES encryption (and for the POLY-1305 authentication) comes the use case for this PR. Restic saves these keys in the repository using passwords and a key derivation function (KDF). Moreover it enables users to save these keys a multiple times with different passwords. There are use cases for this way of handling the encryption keys. However from a security point of view this just adds more components to the security chain. Your data is only as secure as the weakest link in the chain: AES security; KDF security; security of weakest password used.

Using this PR does not change anything about encryption or authentication of the data. It just offers the users the possibility to only rely on the security of AES-256 and POLY-1305 for the safety and authenticity of the data.

rawtaz:
> Using this PR does not change anything about encryption or authentication of the data. It just offers the users the possibility to only rely on the security of AES-256 and POLY-1305 for the safety and authenticity of the data.

But as you wrote, this is already what they are doing. Plus the KDF of course. Is your PR effectively just giving the option to skip the KDF part (including multiple passwords)?

aawsome:
> Is your PR effectively just giving the option to skip the KDF part (including multiple passwords)?

Exactly. This is what I also added to the documentation in 070_encryption.rst. Can I add/change something there to be more clear about what `--masterkeyfile` is doing?

rawtaz:
> Personally when starting with restic I was puzzled that there was no option to just use a key as most other backups tools do offer.

Can you give examples of those other backup tools and what arguments they provide to do what this PR is about?

Honestly, I fail to see what the relevant use for this feature is. You keep talking about raising security, but I'm not convinced since you're not being very clear about what actual concrete problems this solves. I wish you would have discussed this in an issue before you started implementing it. Issue first, PR second, anything else is backwards :)

aawsome:
> Can you give examples of those other backup tools and what arguments they provide to do what this PR is about?

Just consider borg. There is the option `repokey` which saves the key in the repository. This is similar to what restic is doing at the moment. And there is the option `keyfile`which allows to save the key locally. See details [here](https://borgbackup.readthedocs.io/en/stable/internals/data-structures.html#key-files)

I tried to add the option to save the key locally with minimal added complexity and thus came to this PR.

fd0:
I'll clarify what I meant with "complexity": At the moment, there's only one location for the key files (the repo) and only one code path used for accessing data in the repo: get the key file from the repo, then call the kdf with the password to get the master keys.

If we add this option, there's a second code path. Restic would need to add more checks, for example that the master keys are valid and have indeed been generated with restic, the key material is long enough and so on. I can imagine all sorts of issues which could arise from having a "shortcut" around the KDF and storing the key file in a different location outside the repo. Then it becomes an issue of safety, people need to store the file in a different location to be able to access the repo in case the machine they're saving data from becomes unaccessible.

Sure, we could just blame the users when something goes wrong and they lose access to their data, but that's not what we will do.

There are two separate components in this issue, and I'm not sure which one you're trying to solve:
 1. The key files are stored outside of the repository
 2. The (expensive) KDF is not needed for accessing the repo

I'm not convinced this is a good feature to add and I'm especially not convinced that this improves security. For restic, we've chosen to trust the algorithms (scrypt, AES, poly1305), if there's an issue we'll see how to mitigate that.

I'm going to close this issue for now, please feel free to add further comments. Thanks!

aawsome:
@fd0 It is your project, so it is your decision. I fully understand that you fear it would add a maintenance burden to you and I do appreciate that you would feel responsible for what users would do with restic.
From this point your decision is fully understandable.
At the end, it's all work we are doing in our sparse time and you and the other maintainers so far did an incredible job about restic!

However, I cannot follow your arguments about safety and what should be the user's decision and what restic should ensure. As backup tool, restic is a sysadmin's tool and I would say we should assume that the users do know what they are doing (and yes, to blame them if they use - maybe ignoring a warning - an inappropriate setting).

Actually restic allows passwords like "test" or "password" on one hand and forces users to use long generated passwords (which are usually stored in a password file) if they strongly care about security on the other hand.
Do you feel responsible if some repos saved in an unsafe storage with an unsafe password get leaked? Do you feel responsible if a repo with a generated long password gets inaccessible because the sysadmin only stored a password file which got lost? I hope not!
So I cannot follow your opinion that this optional flag worsens the situation for users, sorry.

>     1. The key files are stored outside of the repository
>     2. The (expensive) KDF is not needed for accessing the repo

As you correctly pointed out, this PR does both. I would like to have a solution for 1. The second point is then a nice side-effect when using this PR.
Would you prefer having a solution where the same key file actually stored in the repository can also be stored only locally? Then I can provide a PR for this. I think it shouldn't be much more complicated than this PR.

For me your actual decision means that I'm going to use a patched version of restic including this PR or maybe switch in future to another backup tool if a more feature-rich alternative shows up. This makes me a bit sad as I really like restic and would like to support it.

fd0:
Thank you very much for the constructive message and the praise for the project! It's seldom that people react in this way when their feature has just been declined :)

> However, I cannot follow your arguments about safety and what should be the user's decision and what restic should ensure. As backup tool, restic is a sysadmin's tool and I would say we should assume that the users do know what they are doing (and yes, to blame them if they use - maybe ignoring a warning - an inappropriate setting).

Just to give you a bit of background: I've been working as a penetration tester for the last decade, I've seen many cryptographic systems fail, so I'm familiar with how such systems fail in practice. As a side effect, I've developed an opinion on what aspects of a system involving cryptography users should be able to influence directly. So in my opinion, the key material for cryptographic functions like AES and Poly1305 should never be exposed to users.

> Actually restic allows passwords like "test" or "password" on one hand and forces users to use long generated passwords (which are usually stored in a password file) if they strongly care about security on the other hand.

Because I don't trust users to select secure passwords, restic uses a very strong Key-Derivation Function (scrypt). It is used to derive the keys for decrypting the master keys stored in the key file. When a new repository is initialized (or a password is added), restic calibrates scrypt so that each call to the KDF takes at least 500ms and uses at least 60MiB of memory ([source](https://github.com/restic/restic/blob/6ec5dc8016f1fb47c922b0c5fb421fe188cf2fc1/internal/repository/key.go#L50-L56)).

Let's assume attackers got access to the files of a repository, but they don't know the (single) password. They can try to find this using a word-list or brute force, but for each try, they'll need 500ms on a CPU and 60MiB of memory. So even on a bunch of GPUs this severely limits the number of passwords that can be tried (which, incidentally, is what scrypt is optimized for).

I think using this construction saves a lot of users who may be a bit careless with selecting a good password.

aawsome:
Thank you for your explanation. I strongly appreciate that you take security issues seriously :+1: 
I'm more from the theoretical background. So I do think about theoretical issues (and hope I can contribute from this point) but do not have as much experience in what really fails in practice..

> Interesting, that's what I guessed. Would you mind elaborating why you want that?

Its is as I told that security is always the result of a chain of methods used. Theoretically it is only as save as the weakest link in the chain.

Personally I am maybe very paranoid about data leakage. I do want to have a remote backup repository as this makes sense for back-uped data but I do not trust at all anyone who stores my backup data.
Hence I do need some kind of encryption I trust in combination with a strong key. About the possibility of loosing my key the best method for me would be to print it out on a piece of paper and store this in some physical safe...

Having restic saving the key in the repo adds - as you wrote - scrypt to the security chain.
This means, my data is safe if 

1. my password is strong enough (and not in a word-list, etc that might be within the checks an attacker can do with reasonable effort)
2. the scrypt algorithm cannot be cracked (meaning allowing to find out the derived key without knowing my password within reasonable effort for an attacker)
3. the AES algorithm cannot be cracked (meaning allowing to encrypt the data without knowing the AES key within reasonable effort for an attacker)

Now 1. is in my hands and being paranoid I'll take care about it. About 3. I made my oppinion that I will trust in AES.
About scrypt I do have to admit that I didn't check the algorithm enough to judge it. However regarding the fact that it is much younger than AES and most likely not that well analyzed (yet) I would claim that the biggest threat in the security chain lies in 2. (Please note, that I don't want to say that I believe scrypt can/will be cracked. I just say I believe - with my actual knowledge - it is more likely for scrypt to be cracked than it is for AES.)

So, as a paranoid person, I would like to eliminate risks and therefore use an option which mitigates the risk of scrypt being cracked. This is the case if the key (in whatever form) is only locally stored. Then 2. is completely eliminated and the risk of data leakage is the risk that AES is cracked.

Issue #2523, opened 2019-12-20
Comparison of approaches to tackle index memory usage


aawsome:
### NOTE TO CODE REVIEWERS / CORE CONTRIBUTORS
Please read [this comment](https://github.com/restic/restic/issues/2523#issuecomment-585940174) for a summary of changes and structured information about each of those changes. This might be a better *starting point* than reading through all of the comments in this issue.

Describe the issue
------------------

As stated in #1988 the memory consumption of the index is one of the biggest issues with memory consumption of restic. The main point is the actual implementation of in-memory index storage in `internal/repository/index.go`

I would therefore like to start a new issue to discuss strategies that could remidy the high memory consumption.

As a starting point I opened a branch [index-alternatives](https://github.com/aawsome/restic/tree/index-alternatives) which implements the following strategies:

- Standard: index as actually used in restic
- Reload: Don't save anything. For any index operation reload the file and use the actual implementation for the temporily created index.
- Low-memory index: Store only full index data for tree blobs. For data blobs only use a IDSet to save which data blobs are present. This allows most index operations. For the missing ones do as in Reload (at least for data blobs)
- Bolt: Use a DB on disc via [bbolt](https://github.com/etcd-io/bbolt)

Index strategy can be chosen by
`restic -i <default|reload|low-mem|bolt>`

The branch is considered WIP. I added a general Interface `FileIndex` which should be fulfilled by new index strategies. Also the actual implementation just loads index files (file-per-file) into the standard `Index struct` and then reloads them into a new data structure allowing the GC to clean up. Newly created index files are not affected, so there should be only an effect for repositories with data in it.

Please note that most implementations are quick-and-dirty and miss serious error handling, clean-up etc.

More index strategies can be easily added by adding a new implementation for `FileIndex` and adding them in `repository.go` (where loading takes place) and `cmd/restic/global.go` (where the flag `-i` is interpreted)

I appreciate to get feedback on the proposed index strategies.
In this task I would also like to collect good test settings where different index strategies can be compared to each other.

aawsome:
I like to start with the setup I used to test during the implementation:

I created test data with the following script
```
#!/bin/sh
mkdir data
for i in `seq 1 1000`; do
        echo $i
        mkdir data/$i
        for j in `seq 1 100`; do
                echo $i$j > data/$i/$j
        done
done
```
This creates 100.000 small different files in 1000 directories. Therefore this should give a lot of data blobs and still quite some tree blobs. Can be changed to simulate even more files..

[...]

PR #2532, opened 2020-01-02
Optimize for dirs containing many entries (Introduce Node.Subtrees)

### LINKED ISSUE #2446: memory usage when backing up directories containing many empty files
Output of `restic version`
--------------------------
restic 0.9.5 compiled with go1.12.9 on linux/amd64

See the attached test script, which I ran like this:

```
$ ./test 1000000
```
The script creates a directory containing a million empty files and then backs it up using restic. On an ext4 filesystem the directory is 22MB in size.

Filesystem

Expected behaviour
-----------------
Lower memory usage.

Actual behaviour
---------------
Higher memory usage, see the attached heap profiles with GOGC unset and GOGC=1 taken towards the end of the restic backup process. The peak RAM usage with GOGC unset is 2.8 GB.

With real-world workloads (such as large Maildirs containing more files), the RAM usage can exceed the amount of RAM on the system, leading to swap death.

Steps to reproduce the behaviour
See the attached test script.

I am guessing that restic prepares directory indexes in RAM instead of preparing them on-disk. The issue could also be inefficiently stored directory indexes.

I don't know enough about restic's internals nor Golang memory profiling.

I was happy to find a good replacement for rdiff-backup.

aawsome:

restic uses the `Tree` object to specify which files or subdirs are within a dir. A directory node specifies its contents by referring to a `Tree` object in `Node.Subtree`.
 A `Tree` is simply a list of `Nodes` and is stored in a single blob. 

This raises some problems if a directory contains many files:

- Trees need to be stored completely in memory leading to huge memory requirements
- Memory consumption is double or even 3x during operations like backup (actual tree, previous tree, json mashalling, etc) which makes the memory situation even more severe.
- Trees are usually processed sequentially disallowing efficient parallel processing
- Big blobs can be seen in the repository as they result in big packs. This can give attackers information about the repository without being able to decrypt the files

This PR introduces a new the object `Node.Subtrees` which is a list of `Tree` elements and consequently changes restics behaviour to use the new data structure. `Node.Subtree` is still be read and written if only one subtree exists. This allows this code to work with existing repositories.

As a consequence when having directories with many entries restic is faster and much less memory consuming while no difference can be seen with directories containing only few entries.

The PR is still WIP. Open points are implementation/change of tests and documentation. I would like to discuss the change before investing more time in exhaustive testing.

closes #2446 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"path"
 	"reflect"
-	"sort"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
@@ -101,11 +100,13 @@ func addBlobs(bs restic.BlobSet, node *restic.Node) {
 			bs.Insert(h)
 		}
 	case "dir":
-		h := restic.BlobHandle{
-			ID:   *node.Subtree,
-			Type: restic.TreeBlob,
+		for _, st := range node.Subtrees {
+			h := restic.BlobHandle{
+				ID:   *st,
+				Type: restic.TreeBlob,
+			}
+			bs.Insert(h)
 		}
-		bs.Insert(h)
 	}
 }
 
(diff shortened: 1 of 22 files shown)


rawtaz:
Please do not create PRs without first creating an issue if there isn't an issue for whatever the PR is supposed to address.

There needs to be a place for discussing the proposed feature/problem, and this is the issue.

Once there has been a conclusion in that issue, a PR can be filed. By doing it in this order, unnecessary work is avoided both for the core maintainer(s) and the one producing the PR.

by-cx:
I am gonna test this code on our servers where we have tens of millions of files and we had to stop using Restic because of that.

pabs3:
@by-cx: see also the GOGC=1 workaround from #2446 and the memory reduction strategies in #2523.

pabs3:
Has any more progress been made on this PR? It would be great to have lower memory requirements in restic.

aawsome:
> Has any more progress been made on this PR? It would be great to have lower memory requirements in restic.

No, it is still in the state of the start of 2020 and already suffering from being outdated w.r.t. master.
However, as it is a big change which includes changing the repo format I'm still waiting for comments on the direction of implemention.

Frankly said, I do not want to spend more time without the perspective of getting this PR merged soon after finishing my work. (and moreover in these Corona-times I actually do not have much spare time)

Please note that this PR also only applies to a very specific problem causing large memory usage (i.e. a huge number of *direct directory entries*; it does not generally apply to a huge number of entries within a *tree of directory*) - I would judge working on #2523 more fruitful as it applies to *all* settings with memory problems.

That said there are still open points (e.g. finish work on tests and docu). 
Moreover the whole direction of the solution can and should be discussed. Actually it generates the `Subtrees` data structure which collects a specific fixed number of directory entries. Having thought about this solution a bit in the last couple of weeks I actually favor more to change the `Tree` data structure into a JSON stream of `Node`s. That would allow to use the chunker to cut this down in many blobs. Many parts of this PR could still be used for that kind of format change but others would need a change.

pabs3:
Ok, thanks for the info and your work on this and restic in general.

-- 
bye,
pabs

https://wiki.debian.org/PaulWise

pabs3:
Some folks working on bup are doing content based tree splitting, so that adding one file to a large maildir only adds a small amount of metadata rather than another copy of the directory.

https://lwn.net/Articles/824632/
https://github.com/jmberg/bup/commit/44006daca4786abe31e32a969a08778133496663

rawtaz:
> It would be great to have lower memory requirements in restic.

There's other memory usage related PRs that have been merged and that you can benefit from by running the latest master build of restic that you can get from [here](https://beta.restic.net/?sort=time&order=desc) :)

pabs3:
I noticed that the other memory usage related PRs that got merged did not have much effect on the memory usage for large directories.

aawsome:
Closing this PR as I won't work anymore on this.

IMO the way to go would be to chunk tree packs using the standard chunking algorithm and then save the blob IDs which are needed to get a tree in JSON format in the `Contents` field of the node.

MichaelEischer:
> IMO the way to go would be to chunk tree packs using the standard chunking algorithm and then save the blob IDs which are needed to get a tree in JSON format in the `Contents` field of the node.

That sounds like a reasonable way to adapt the repository format. I've started some experiments in that direction, including being able to incrementally generate, chunk and save trees, but it will take some time until that code is in any way presentable. The only downside I've noticed is that `recover` will no longer work in all cases. A chunked tree just won't be valid.

PR #2536, opened 2020-01-04
Update threat model

### LINKED ISSUE #208: Update threat model
The threat model (section in `doc/Design.md`) should be updated to the latest structure, e.g. there is no `id` or `version` file any more.

Furthermore, remove implementation specific details from it as much as possible. A good example is https://pond.imperialviolet.org/threat.html

Clarify on when the local machine is trusted, e.g. only during backup or when the password is present.

MatthewVance:
Updates the threat model section within the design document.

Yes. Closes #208 

Checklist
---------

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -566,43 +566,100 @@ Threat Model
 ============
 
 The design goals for restic include being able to securely store backups
-in a location that is not completely trusted, e.g. a shared system where
-others can potentially access the files or (in the case of the system
-administrator) even modify or delete them.
+in a location that is not completely trusted (e.g., a shared system where
+others can potentially access the files) or even modify or delete them in
+the case of the system administrator.
 
 General assumptions:
 
 -  The host system a backup is created on is trusted. This is the most
-   basic requirement, and essential for creating trustworthy backups.
+   basic requirement, and it is essential for creating trustworthy backups.
+-  The user uses an authentic copy of restic.
+-  The user does not share the repository password with an attacker.
+-  The restic backup program is not designed to protect against attackers
+   deleting files at the storage location. There is nothing that can be
+   done about this. If this needs to be guaranteed, get a secure location
+   without any access from third parties.
+-  The whole repository is re-encrypted if a key is leaked. With the current
+   key management design, it is impossible to securely revoke a leaked key
+   without re-encrypting the whole repository.
+-  Advances in cryptography attacks against the cryptographic primitives used
+   by restic (i.e, AES-256-CTR-Poly1305-AES and SHA-256) have not occurred. Such
+   advances could render the confidentiality or integrity protections provided
+   by restic useless.
+-  Sufficient advances in computing have not occurred to make bruteforce
+   attacks against restic's cryptographic protections feasible.
 
 The restic backup program guarantees the following:
 
--  Accessing the unencrypted content of stored files and metadata should
-   not be possible without a password for the repository. Everything
-   except the metadata included for informational purposes in the key
-   files is encrypted and authenticated.
+-  Unencrypted content of stored files and metadata cannot be accessed
+   without a password for the repository. Everything except the metadata
+   included for informational purposes in the key files is encrypted and
+   authenticated. The cache is also encrypted to prevent metadata 
+   leaks.
+-  Modifications to data stored in the repository (due to bad RAM, broken
+   harddisk, etc.) can be detected.
+-  Data that has been tampered will not be decrypted.
 
--  Modifications (intentional or unintentional) can be detected
-   automatically on several layers:
+With the aforementioned assumptions and guarantees in mind, the following are
+examples of things an adversary could achieve in various circumstances.
 
-   1. For all accesses of data stored in the repository it is checked
-      whether the cryptographic hash of the contents matches the storage
-      ID (the file's name). This way, modifications (bad RAM, broken
-      harddisk) can be detected easily.
+An adversary with read access to your backup storage location could:
 
-   2. Before decrypting any data, the MAC on the encrypted data is
-      checked. If there has been a modification, the MAC check will
-      fail. This step happens even before the data is decrypted, so data
-      that has been tampered with is not decrypted at all.
+-  Attempt a brute force password guessing attack against a copy of the
+   repository (even more reason to use long, 30+ character passwords).
+-  Infer which packs probably contain trees via file access patterns.
+-  Infer the size of backups by using creation timestamps of repository objects.
 
-However, the restic backup program is not designed to protect against
-attackers deleting files at the storage location. There is nothing that
-can be done about this. If this needs to be guaranteed, get a secure
-location without any access from third parties. If you assume that
-attackers have write access to your files at the storage location,
-attackers are able to figure out (e.g. based on the timestamps of the
-stored files) which files belong to what snapshot. When only these files
-are deleted, the particular snapshot vanished and all snapshots
-depending on data that has been added in the snapshot cannot be restored
-completely. Restic is not designed to detect this attack.
+An adversary with network access could:
+
+-  Attempt to DoS the server storing the backup repository or the network 
+   connection between client and server.
+-  Determine from where you create your backups (i.e., the location where the
+   requests originate).
+-  Determine where you store your backups (i.e., which provider/target system).
+-  Infer the size of backups by using creation timestamps of repository objects.
+
+The following are examples of the implications associated with violating some
+of the aforementioned assumptions.
+
+An adversary who compromises (via malware, physical access, etc.) the host
+system making backups could:
+
+-  Render the entire backup process untrustworthy (e.g., intercept password, 
+   copy files, manipulate data).
+-  Create snapshots (containing garbage data) which cover all modified files
+   and wait until a trusted host has used forget often enough to forget all
+   correct snapshots.
+-  Create a garbage snapshot for every existing snapshot with a slightly different
+   timestamp and wait until forget has run, thereby removing all correct
+   snapshots at once.
+
+An adversary with write access to your files at the storage location could:
+
+-  Delete or manipulate your backups, thereby impairing your ability to restore
+   files from the compromised storage location.
+-  Determine which files belong to what snapshot (e.g., based on the timestamps
+   of the stored files). When only these files are deleted, the particular
+   snapshot vanishes and all snapshots depending on data that has been added in
+   the snapshot cannot be restored completely. Restic is not designed to detect
+   this attack.
+
+An adversary who compromises a host system with append-only access to the 
+backup repository could:
+
+-  Render new backups untrustworthy *after* the host has been compromised
+   (due to having complete control over new backups). An attacker cannot delete
+   or manipulate old backups. As such, restoring old snapshots created *before*
+   a host compromise remains possible.
+   *Note: It is **not** recommended to ever run forget automatically for an
+   append-only backup to which a potentially compromised host has access
+   because an attacker using fake snapshots could cause forget to remove
+   correct snapshots.*
+
+An adversary who has a leaked key for a repository which has not been re-encrypted
+could:
+
+-  Decrypt existing and future backup data. If multiple hosts backup into the same
+   repository, an attacker will get access to the backup data of every host.
 


hashhar (approved):

doc/design.rst
```diff
@@ -580,21 +580,12 @@ The restic backup program guarantees the following:
 -  Accessing the unencrypted content of stored files and metadata should
    not be possible without a password for the repository. Everything
    except the metadata included for informational purposes in the key
-   files is encrypted and authenticated.
-
+   files is encrypted and authenticated. The cache is also encrypted by 
+   default to also prevent metadata leaks when encrypted filesystems are 
+   mounted.
 -  Modifications (intentional or unintentional) can be detected
    automatically on several layers:
```
MichaelEischer: Is there a reason why the modification detection list was removed? The attacker examples look at tampered backups from a different angle.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
 depending on data that has been added in the snapshot cannot be restored
 completely. Restic is not designed to detect this attack.
 
+With the aforementioned assumptions and guarantees in mind, the following  are
```
MichaelEischer: There is a double whitespace between "following  are"

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with access to your backup storage location could: 
+- Delete or manipulate your backups, thereby impairing your ability to restore
+files from the compromised storage location. - Attempt a brute force password
```
MichaelEischer: The brute-force bullet point should be on a separate line.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
 depending on data that has been added in the snapshot cannot be restored
 completely. Restic is not designed to detect this attack.
 
+With the aforementioned assumptions and guarantees in mind, the following  are
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with access to your backup storage location could:
```
MichaelEischer: I'd suggest to write "read access" to differentiate it from the write access later on. Then the point on manipulation should be moved to the next section.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with access to your backup storage location could: 
+- Delete or manipulate your backups, thereby impairing your ability to restore
+files from the compromised storage location. - Attempt a brute force password 
+guessing attack against a copy of the  repository (even more reason to use 
+long, random passwords).
+
+An adversary with write access to your files at the storage location could: 
+- Delete specific files (by using timestmaps to determine which files belong to
```
MichaelEischer: Typo: "timestmaps" -> "timestamps"

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+what snapshots), resulting in snapshots vanishing and all dependent snapshots
```
MichaelEischer: "preventing" is missing between all and dependent.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+from being completely restored.
+
+An adversary who can passively monitor all Internet traffic could: 
+- Determine from where you create your backups (i.e., the location where the 
+requests originate).
+- Determine where you are storing your remote backups via HTTP headers (URLs)
```
MichaelEischer: Does "where" just refer to the path on the target system? No matter whether you use encryption an adversary will be able to tell which server you contacted.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+files from the compromised storage location. - Attempt a brute force password 
+guessing attack against a copy of the  repository (even more reason to use 
+long, random passwords).
+
+An adversary with write access to your files at the storage location could: 
+- Delete specific files (by using timestmaps to determine which files belong to
+what snapshots), resulting in snapshots vanishing and all dependent snapshots
+from being completely restored.
+
+An adversary who can passively monitor all Internet traffic could:
```
MichaelEischer: I'd suggest to move this section down to the other "network access" section.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+With the aforementioned assumptions and guarantees in mind, the following  are
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with access to your backup storage location could: 
+- Delete or manipulate your backups, thereby impairing your ability to restore
+files from the compromised storage location. - Attempt a brute force password 
+guessing attack against a copy of the  repository (even more reason to use 
+long, random passwords).
+
+An adversary with write access to your files at the storage location could:
```
MichaelEischer: "Write access" seems to primarily mean "delete files" here. Or is this intended to also include an attacker which just can add (broken) data?

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+server is one case where unsecured HTTP could be used. However, system 
+administrators have the option (and are encouraged) to configure the REST 
+server to use HTTPS.
+
+An adversary who compromises (via malware, physical access, etc.) the  host
+system making backups could: 
+- Render the entire backup process untrustworthy (e.g., intercept password, 
+copy files, manipulate data).
+
+An adversary with the ability to add data to the repo (but not remove something)
```
MichaelEischer: The adversary in this section is a very powerful one: It can add files and possesses the encryption password. In other word it is indistinguishable from a regular backup client with append-only access (via rclone with `--append-only` flag). This is this is very similar to the previous section, just with an added append-only restriction.

An attacker could also prevent restic from working by adding properly authenticated garbage to the repository. This is probably the most effective short-term attack for an attacker with append-only access.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+attempt to get a future forget operation to remove some or all existing good 
+snapshots from the repository. However, this requires the attacker to know the
+repository password or master key in order to encrypt the malicious snapshot 
+objects so that forget will be able to read them.
+
+An adversary with network access could: 
+- attempt to DoS the restic server.
+
+Advances in cryptography attacks against the cryptographic primitives used by
+restic (i.e, AES-256-CTR-Poly1305-AES and SHA-256) could render the the
```
MichaelEischer: There's one "the" too much.

doc/design.rst
```diff
@@ -606,3 +597,61 @@ are deleted, the particular snapshot vanished and all snapshots
...
+confidentiality or integrity protections provided restic useless. Likewise,
```
MichaelEischer: Provided "by" restic.

rawtaz:
Needs corrections based on @MichaelEischer's comments :)

MatthewVance:
Yes, I saw and appreciate the feedback. I was out of town when this came
through, but plan to review more closely when I have a good chance. I'll
either submit a new PR or reply with follow up questions.

rawtaz:
@MatthewVance Sounds good, thanks a lot for your efforts here. I don't mean to stress you out.

Please make changes in this PR, there's no point in opening a new one for the same thing. You can make and push corrections to the same branch that you have for this PR, just let us know if you need help with any of that Git stuff 👍

MatthewVance:
Thanks. I should be good. Just a matter of free time.

MatthewVance:
@rawtaz I was typing on mobile yesterday and hope my prior comment didn't come across the wrong way. You're not stressing me out. I want to contribute to this issue because I really appreciate the work you and the other put into this good tool. The only stress I have is not being able to get to this as quickly as I'd like.

I have some time to review this more closely and begin working on this again now.

doc/design.rst
MatthewVance: Yes, though the colon should have been replaced with a period. It was removed for a couple of reasons.
1. @fd0 had also suggested removing implementation specific details as much as possible so my thinking was explaining that modification can be detected was sufficient without including the extra details. See https://github.com/restic/restic/issues/208#issue-92063294. 
2. In the same comment, @fd0 mentioned the threat model "should be updated to the latest structure, e.g. there is no id or version file any more."  The detection list still referenced an `ID`. If I missed a detail of the structure and that's an incorrect interpretation, let me know.

doc/design.rst
MatthewVance: Good catch, but I'm thinking "preventing: should be between "and" and "all" instead (i.e., resulting in snapshots vanishing and preventing all dependent snapshots from being completely restored).

doc/design.rst
MatthewVance: I agree it could make sense to put these two sections near each other. In my mind, "an adversary with network access" is slightly different than one "who can passively monitor all Internet traffic." However, I can see arguments either way and think having only one bullet point under the general network access odd so I ended up combining them. I debated working in "passively monitor" into the two additional points, but decided against it. If you think it's needed, let me know.

doc/design.rst
MatthewVance: Both. That's what the "or manipulate your backups" after the delete intended to convey.

doc/design.rst
MatthewVance: Yes, this is a very powerful one. Like you mentioned, the main difference is related to `--append-only`. 

@MichaelEischer, for your point about adding properly authenticated garbage to the repo, how big of a deal is it? 
- Would it corrupt the entire repo or only the adversary's own backed up files? As I understand it, the password a user uses to write data to the repo is different than the actual encryption password (the master key). 
- If [different repo keys](https://restic.readthedocs.io/en/stable/070_encryption.html#manage-repository-keys) are used by each user, does this make a difference? I use Restic in a single user scenario so don't fully understand the implications of multiple keys (passwords) beyond knowing it allows a password to be changed without having to re-encrypt all data (i.e., a key can exist for each file). 
- Do the integrity protections ([ciphertext verification](https://restic.readthedocs.io/en/stable/045_working_with_repos.html#checking-a-repo-s-integrity-and-consistency)) at time of restore come into play?

MatthewVance: As I'm looking into this a bit more, I understand a user's password would allow them to decrypt the repo's master key and that in turn would allow it to read the entire repository without needing a password at all (provided the user modifies to client code to support this). I didn't catch that at first and now see what you mean by possession the encryption password (at first I was thinking password / user key). This does further underscore how powerful this adversary is. 

I guess my main question is, what are you proposing here? Are you suggestion this entire section be removed because `append-only` limitation in this scenario isn't too different from an attacker's standpoint (i.e., it renders the entire backup process untrustworthy like the prior section)?

MatthewVance:
I've been looking over some discussions in the forum and wondering if anything from these posts should be incorporated into the threat model:
* [Questions and thoughts on restic security - revocable keys and cryptoperiods](https://forum.restic.net/t/questions-and-thoughts-on-restic-security-revocable-keys-and-cryptoperiods/2392)
* [Secure Backup Strategy - Requesting Comments](https://forum.restic.net/t/secure-backup-strategy-requesting-comments/1235)

Thoughts?

doc/design.rst
MatthewVance: Good question. In my mind, the "where" includes any details leaked in the URL. At minimum, this includes the server you contacted. I agree an adversary can tell which server you contacted regardless if encryption is in play or not. While that should generally be understood, I thought it was worth explicitly calling out as something to consider. If someone is very concerned about this, backing up on a local network rather than over the Internet should be considered. What change/suggestion did you have in mind?

MatthewVance:
I went ahead and submitted a new pull request to address most of the above suggestions.

MichaelEischer (commented):
Sorry for taking so long to respond to your feedback, every time I wanted to take a look some other seemingly more pressing issue showed up :-( .

I've noticed that the thread model referenced in #208 (still accessible on archive.org) seems to be a lot longer, however, I didn't have time yet to take a closer look at it yet. The forum posts are also still on my todo list.

Could you mark the review comments which are fixed as resolved? Github doesn't allow me to.

MichaelEischer:
The "revocable keys and cryptoperiods" forum topic mentions an important point: With the current key management design it is **impossible** to securely revoke a leaked key without reencrypting the whole repository. Once a key is leaked an attacker is able to decrypt existing and future backup data. If multiple hosts backup into the same repository and attacker will get access to the backup data of every host!

The "secure backup strategy" topic should already be covered by the attacker with append-only access discussion.

I took a look at the thread model of pond and think that we want to add the following assumptions:
- The user uses an authentic copy of restic. This is more or less implied by requiring a trusted host system, but it's probably worth to state that explicitly.
- The user does not share the repository password with an attacker. 

The discussion of what an attack on restic can do actually mixes two categories of attacks: The ones like repository read access or network monitoring which do not violate any assumptions made and the ones with an attacker with write/append access or a compromised host system which break some assumptions and discuss the resulting implications. It might be a good idea to separate these two categories.

I've also noticed a few more attacks / metadata leaks:
- An attacker with read access can use creation timestamps of repository objects to infer the size of backups
- The same applies for an adversary with network access
- For a compromised host system, I'd suggest to spell out the following attack: An attacker could create snapshots which cover all modified files, but actually just contain garbage data; then an attacker just needs to wait until a trusted host has used forget often enough to forget all correct snapshots. Oh, I just realized that an attacker just needs to create a garbage snapshot for every existing snapshot with a slightly newer/older(?) timestamp and the forget will remove all correct snapshots at once. That is it's a rather bad idea to ever run forget automatically for an append-only backup to which a (potentially) compromised host has access.
- I'd rename the "An adversary with the ability to add data to the repo (but not remove something)" paragraph to "An attacker who compromises a host system with append-only access to the backup repository".

MatthewVance:
@MichaelEischer thanks for the feedback.

doc/design.rst
```diff
@@ -606,3 +597,59 @@ are deleted, the particular snapshot vanished and all snapshots
 depending on data that has been added in the snapshot cannot be restored
 completely. Restic is not designed to detect this attack.
 
+With the aforementioned assumptions and guarantees in mind, the following are
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with read access to your backup storage location could: 
+- Attempt a brute force password guessing attack against a copy of the 
+repository (even more reason to use long, random passwords).
```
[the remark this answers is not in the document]
MatthewVance: I don't know if there's a universal consensus on password length. I checked the new NIST guidelines ([800-63B](https://doi.org/10.6028/NIST.SP.800-63b)) and it only states "The minimum password length that should be required depends to a large extent on the threat model being addressed." Personally, I'm in the 30+ camp. I ended up going revising to: `...repository (even more reason to use long, 30+ character, passwords).`

doc/design.rst
```diff
@@ -580,20 +580,11 @@ The restic backup program guarantees the following:
...
-      ID (the file's name). This way, modifications (bad RAM, broken
-      harddisk) can be detected easily.
-
-   2. Before decrypting any data, the MAC on the encrypted data is
-      checked. If there has been a modification, the MAC check will
-      fail. This step happens even before the data is decrypted, so data
-      that has been tampered with is not decrypted at all.
+   automatically on several layers.
 
 However, the restic backup program is not designed to protect against
```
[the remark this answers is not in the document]
MatthewVance: I added the three sentences as a new bullet point under general assumptions. 

I didn't make any changes regarding snapshot suppression.

doc/design.rst
```diff
@@ -606,3 +597,59 @@ are deleted, the particular snapshot vanished and all snapshots
...
+- if forget with time-based policies is regularly run against the repository 
+(particularly when automated), an attacker could introduce new snapshots (that
+are empty or contain garbage data) with specifically-crafted timestamps in an 
+attempt to get a future forget operation to remove some or all existing good 
+snapshots from the repository. However, this requires the attacker to know the
+repository password or master key in order to encrypt the malicious snapshot 
+objects so that forget will be able to read them.
+
+An adversary with network access could: 
+- attempt to DoS the restic server.
```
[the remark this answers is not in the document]
MatthewVance: I see your point about "restic server" being ambiguous. Even with your suggested revision, I'm struggling how to better clarify this one.  DoS is a threat, but a very nuanced one. 

Another scenario/nit is cases when the "server storing the backup repository" is a cloud service like S3. Your average adversary isn't going to have any luck performing a DoS attack against that. The attacker would have better luck going after the endpoints attempting to upload data to the restic repo. 

I also somewhat disagree with the local backup exception. The "adversary with network access" phrase implies this doesn't apply when storing locally. Also, if an attacker can DoS a box (regardless if restic itself is network exposed to receive backups), it could cause collateral damage (albeit the DoS would not be against the restic service, but the infrastructure it runs on). I don't know that any of this is worth calling out explicitly though. 

This is a long way of saying I need help thinking through how to better explain the DoS threat.

doc/design.rst
```diff
@@ -580,20 +580,11 @@ The restic backup program guarantees the following:
 -  Accessing the unencrypted content of stored files and metadata should
    not be possible without a password for the repository. Everything
    except the metadata included for informational purposes in the key
-   files is encrypted and authenticated.
-
+   files is encrypted and authenticated. The cache is also encrypted by 
+   default to also prevent metadata leaks when encrypted filesystems are
```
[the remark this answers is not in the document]
MatthewVance: I pulled most of this wording from @fd0's https://github.com/restic/restic/issues/208#issuecomment-246323673, but I'll admit I'm no expert on restic caching so I could be missing a thing or two. The cache isn't covered well in the [design doc](https://github.com/restic/restic/blob/master/doc/design.rst) and I'm not familiar enough with the code base to confidently answer this one. My assumption is the `encrypted filesystem` is referring to the encrypted parts of the repo that can be mounted via FUSE.

Regarding your first question, I believe you're correct that this could also apply more broadly.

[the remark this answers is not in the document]
MichaelEischer: The cache is mostly a local copy of certain repository files. It basically contains a copy of the whole index and all packs containing tree blobs. The usage of the cache for the fuse mount does no really differ from the cache usage for all other commands. I'd just leave out the "when encrypted filesystems are mounted" part.

Actually "by default" is also wrong. There is currently no way to store a decrypted cache. So I think the important message here is just, that the cache only stores encrypted data which prevents metadata leaking from the cache.

doc/design.rst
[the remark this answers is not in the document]
MichaelEischer: The "attempt to" should mostly cover the part that the attacker quite likely won't be able to bring down S3. You could add that the attacker could DoS the network connection between client and server. That should be generic enough to DoS attacks on intermediate nodes or the client itself. Intermediate node or the client usually provide no external services, thus an attacker would have to resort to flooding the networking connection with too many requests.

MichaelEischer:
@MatthewVance Are you still interested in finishing the threat model update? There are still several unaddressed comments.

MatthewVance:
@MichaelEischer I'm still interested. I am actively re-reviewing the comments and making more updates. Thanks for bringing this back to my attention.

doc/design.rst
```diff
@@ -566,43 +566,99 @@ Threat Model
...
+-  Create snapshots (containing garbage data) which cover all modified files
+   and wait until a trusted host has used forget often enough to forget all
+   correct snapshots.
+-  Create garbage snapshot for every existing snapshot with a slightly different
+   timestamp and wait until forget has run, thereby removing all correct
+   snapshots at once.
+
+An adversary with write access to your files at the storage location could:
+
+-  Render the entire backup process untrustworthy. Not much is different from
```
MichaelEischer, line 641: "untrustworthy" takes it too far. An attacker with just access to the storage location can only break the backup, but not add manipulated data to it. `restic check --read-data` would be able to detect any manipulations with just that violated assumption. The previous description was much more accurate:

"Delete or manipulate your backups, thereby impairing your ability to restore files from the compromised storage location."

doc/design.rst
```diff
@@ -566,43 +566,99 @@ Threat Model
...
+-  Determine which files belong to what snapshot (e.g., based on the timestamps
+   of the stored files). When only these files are deleted, the particular
+   snapshot vanishes and all snapshots depending on data that has been added in
+   the snapshot cannot be restored completely. Restic is not designed to detect
+   this attack.
+
+An adversary who compromises a host system with append-only access to the 
+backup repository could:
+
+-  Render the entire backup process untrustworthy. Other than not being able to
```
MichaelEischer, line 653: Well, such an attacker cannot manipulate existing snapshots. That's why we have to worry about automatic forget runs in that scenario.

MichaelEischer (dismissed):
Thanks for the cleanups. I have a few additional comments: the last two scenarios discussing "untrustworthy" backups sound worse than things really are. The current wording sounds a bit like append-only backup repositories are totally useless and an attacker with write access to a backup repository is able to perform arbitrary manipulations. However, neither is the case.

I've just noticed that it is somewhat possible to combine append-only backups with forget when using the `--keep-within duration` option. An attacker cannot trick a user into deleting snapshots within "duration".

doc/design.rst
```diff
@@ -566,43 +566,99 @@ Threat Model
...
+
+An adversary who compromises (via malware, physical access, etc.) the host
+system making backups could:
+
+-  Render the entire backup process untrustworthy (e.g., intercept password, 
+   copy files, manipulate data).
+-  Create snapshots (containing garbage data) which cover all modified files
+   and wait until a trusted host has used forget often enough to forget all
+   correct snapshots.
+-  Create garbage snapshot for every existing snapshot with a slightly different
```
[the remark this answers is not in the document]
MatthewVance: I don't understand the intent here.

doc/design.rst
MatthewVance: I agree, I like the prior wording better. However, this change was made to address a prior review comment that said:

> I think the difference between an attacker with full write access vs. append-only is worth discussing. The main difference to full write access is that you can't directly delete previous backups. The only way to remove data would be to trick a forget run with full write access into deleting backups.
> 
> Besides being unable to delete old backups there's not much difference to a compromised host system. 

It's also worth noting the general assumption in the currently published threat model states:

> The host system a backup is created on is trusted. This is the most basic requirement, and essential for creating trustworthy backups.

If that's correct, an untrusted host system (i.e., a compromised one) makes it impossible to create trustworthy backups.

Given the aforementioned quotes, I inferred there wasn't much difference to a compromised host system. Am I misunderstanding the original comment?

doc/design.rst
MatthewVance: As I drafted this section, I had reservations about how the note was put in but struggled to identify a better way given the prior comment that said "Besides being unable to delete old backups there's not much difference to a compromised host system."

Like the attacker with write access scenario to the storage location, this also mentions rendering the entire backup process untrustworthy. If that was taking it too far for an attacker with write access, is that not a concern here as well? 

In some ways, I prefer the wording from the earlier commit:

> An adversary with the ability to add data to the repo (but not remove something) could: 
> - upload pack files with wrong headers, but only if the attacker knows the password to encrypt the pack. For example, an attacker could include a header saying the file contains a blob with a specific hash when that's not really the case. 
> - upload an index file claiming a specific blob IDs exist in the repository, but only if the attacker knows the password. In this case, restic will not question this information until the index is rebuilt (by rebuild-index or prune). 
> - if forget with time-based policies is regularly run against the repository (particularly when automated), an attacker could introduce new snapshots (that are empty or contain garbage data) with specifically-crafted timestamps in an attempt to get a future forget operation to remove some or all existing good snapshots from the repository. However, this requires the attacker to know the repository password or master key in order to encrypt the malicious snapshot objects so that forget will be able to read them.

I'm not sure I'm understanding all the comments on this section. I need much more clarity on the specific suggestion here.

doc/design.rst
MatthewVance: The current wording sounds a bit like append-only backup repositories are totally useless and an attacker with write access to a backup repository is able to perform arbitrary manipulations. However, neither is the case.

Would it help if a note added that stated:

> An attacker write write access to a backup repository is unable to perform arbitrary manipulations.

doc/design.rst
```diff
@@ -566,43 +566,99 @@ Threat Model
...
+
+With the aforementioned assumptions and guarantees in mind, the following are
+examples of things an adversary could achieve in various circumstances.
+
+An adversary with read access to your backup storage location could:
+
+-  Attempt a brute force password guessing attack against a copy of the
+   repository (even more reason to use long, 30+ character passwords).
+-  Infer which packs must contain trees via the cache (restic only caches tree
+   packs, thereby introducing a small side-channel).
```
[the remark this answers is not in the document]
MatthewVance: Good catch. I wanted to capture the slight risk associated with the cache, but didn't intent to imply (like I unintentionally did) that this risk was associated with the backup storage location. I believe that would only be the case if the host running the backup also served as its own storage location. That's probably a rare scenario. 

Would it make sense to add something like the following?

> An adversary with read access to your backup host could:
> -  Infer which packs must contain trees via the cache (restic only caches tree packs, thereby introducing a small side-channel).

Personally, I don't know that it makes sense to add such a section. If an attacker has read access to the backup host, it's game over (why attack the backups if you have access to the host being backed up? Thinking about this some more, I'm thinking it may be better to entirely leave out the mention about inferring which packs must contain trees via the cache. Please let me know if you disagree.

I hadn't thought about your other point about inferring info from the file access patterns. I think what I'll do is replace the cache bullet point with one that says "Infer which packs must contain trees via file access patterns."

doc/design.rst
MatthewVance: Should I just revert back to this earlier wording (with the latest revision of the second bullet point which better retains original wording)?

> An adversary with write access to your files at the storage location could:
> 
> -  Delete or manipulate your backups, thereby impairing your ability to restore
>    files from the compromised storage location. 
> -  Determine which files belong to what snapshot (e.g., based on the timestamps
>    of the stored files). When only these files are deleted, the particular
>    snapshot vanishes and all snapshots depending on data that has been added in
>    the snapshot cannot be restored completely. Restic is not designed to detect
>    this attack.

doc/design.rst
MatthewVance: @MichaelEischer thanks for all your feedback, help, and patience on this one. Before making further revisions, I wanted to get some more clarity from you.

In addition to the other questions within the review, could you elaborate on the following?

> I've just noticed that it is somewhat possible to combine append-only backups with forget when using the --keep-within duration option. An attacker cannot trick a user into deleting snapshots within "duration".

doc/design.rst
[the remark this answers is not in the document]
MichaelEischer: I agree that when an attacker has read access to the cache on the backup host, then it's probably already game over. Mentioning that explicitly would probably only clutter the threat model.

I'd weaken the remark on the file access patterns slightly to "probably contain trees". As restic caches tree packs for most commands, that definitely complicates things. Although an file access trace for a check command run, is probably enough.

doc/design.rst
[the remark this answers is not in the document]
MichaelEischer: Oh, sorry, I was too lazy to type that out and just wrote a `sed` pattern. It says that `garbage snapshot` should be replaced with `a garbage snapshot`. Or in other words please add a `a` before `garbage snapshot`.

doc/design.rst
MichaelEischer: > > I think the difference between an attacker with full write access vs. append-only is worth discussing. The main difference to full write access is that you can't directly delete previous backups. The only way to remove data would be to trick a forget run with full write access into deleting backups.
> > Besides being unable to delete old backups there's not much difference to a compromised host system.

That previous comment referred to an attacker with access to the host creating the backup. This comment is about an attacker with access to the file storage location only! And unless an attacker knows the repository password, they won't be able to create / manipulate a file such that it still passes the integrity checks. So all an attacker is able to is to delete random files, restore deleted files again etc. But essentially all of this can only cause data loss, but it is never possible to manipulate the actual content of a snapshot.

So, yes I think it's best to just revert the first bullet point to its earlier wording.

doc/design.rst
MichaelEischer: > > Besides being unable to delete old backups there's not much difference to a compromised host system.
>
> It's also worth noting the general assumption in the currently published threat model states:
> 
> > The host system a backup is created on is trusted. This is the most basic requirement, and essential for creating trustworthy backups.
> 
> If that's correct, an untrusted host system (i.e., a compromised one) makes it impossible to create trustworthy backups.

With append-only backups we have to distinguish between backups created before a host compromise (old backups) and backups created afterwards (new backups). After a compromise an attacker has complete control over new backups, which are no longer trustworthy from that point on. The important property of append-only backups, however, is that an attacker is not able to manipulate old backups and therefore it is still possible to properly restore those even in case of a compromised host.
This is what I meant with "unable to delete old backups".

My main problem with the current wording is that "Render the entire backup process untrustworthy" first states that all hope is lost before retracting that statement in parts. I'd rather prefer something like: "An attacker cannot delete or manipulate old backups. That is correctly restoring snapshots created before a host compromise remains possible. However, the creation of new backups is no longer trustworthy as described previously.".

>  *Note: It is not recommended to ever run forget automatically for an append-only backup to which a potentially compromised host has access.*

The attack with fake snapshots that could cause `forget` to remove the correct ones only works for options like `forget --keep-daily 42 --keep-last 15`. For `--keep-daily` an attacker has to add a latest snapshot for each day to remove the correct one(s). For `--keep-last` one can simply flood the repository with snapshots. `--keep-within` would keep all snapshots within a certain timespan no matter how many snapshots one adds for that timerange. However, I've just noticed that an attacker could create a snapshot in the far future which would again wipe out all correct snapshots 🤕 . So let's forget about that parameter for now.

doc/design.rst
MatthewVance: Thanks for clarifying. This makes more sense to me now. In all the re-organization shuffle and elapsed time between my earlier updates, I missed that one was regarding access to the host and other other was storage location. That certainly makes a difference!

doc/design.rst
MatthewVance: Thanks again. The new wording should better reflect the risk. 

The forget piece is worrying. 🤕

MichaelEischer (commented):
The latest version looks good to me. Please rebase the branch to the current master and squash all commits it a single one.

@fd0: As you opened the original issue, could you have a quick look at the changes?

fd0 (approved):
LGTM! thanks!

Issue #2502, opened 2019-12-07
Clarify "Modification time" in backup reference


rawtaz:
In the bullet list at https://restic.readthedocs.io/en/latest/040_backup.html#backing-up there is mention of "Modification time" as one of the things restic looks at to determine if a file changed or not.

In https://github.com/restic/restic/pull/2212 restic changed from using mtime to using ctime for detecting file changes (amongst other methods).

I'm not sure if "Modification time" is still a good description of the new behavior. In a non-technical sense it might accommodate for both mtime and ctime, but perhaps it's too coupled with mtime? In that case, we should probably change this little item in the list to something else.

Thoughts? If changed, would simply making it "Change time" be enough (still not very clear or specific, but perhaps that's how we want it in this usage type of documentation)?

MichaelEischer:
The documentation section has been reworked in PR #2823.

PR #2541, opened 2020-01-09
Issue 2502 - Improved docs describing backup change detection

dp-github:
Documentation improvements that:

- Increase the clarity of which files restic reads during a backup
- Mentions both the mtime/ctime attributes
- Moves `restic check` to its own section

https://github.com/restic/restic/issues/2502

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2502
+++ b/changelog/unreleased/issue-2502
@@ -1,0 +1,5 @@
+Documentation: Refactor backup documentation
+
+In particular, include a better description of mtime/ctime 
+
+https://github.com/restic/restic/issues/2502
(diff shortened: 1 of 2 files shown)


rawtaz:
I looked through these changes and I don't think we can merge them right now.

One thing that we need to fix is to simply remove the changelog file - it's not needed for documentation changes outside of the restic source code, and it also makes the build fail (because Calens doesn't support documentation changes).

Regarding the changes, I think it doesn't really address #2502, in the sense that the only change actually addressing what's discussed in that issue is the change of "Modification time" to "Creation and modification times". That, in combination with the rest of the text's changes doesn't really fix the issue (IMO).

Also the text changes aren't on spot, we need to adjust it a bit. But I don't have time to sum it up right now.

MichaelEischer:
Closing this in favor of #2823.

Issue #2508, opened 2019-12-10
json output for diff


rrossouw01:
Output of `restic version`
--------------------------
0.9.6

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
diff output recognize the --json flag

What are you trying to do?
--------------------------
when trying to scale restic we run --quiet on backups with millions of files on NFS mounts to eliminate the scan.  however that mean we do not have results of the backup for later processing/monitoring.  So diff provides kind of a workaround to get at least some of the statistics.  It would be nice if that was json though and possibly also an option not to print the 1000's of +, -, M lines just the file/dir etc changes.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
love it!

rrossouw01:
Even the feature request to have diff honor --json would be nice I need to clarify something.  I thought --quiet would mean no stats at all.  That is not correct if using --quiet and --json together during backup the --json flag provides the stats I need.  It may only provide it if redirecting output to a file I did not check specifically, regardless it works for us.  So I was incorrect to think we have to use diff.

rawtaz:
Is there still a use case for such a feature now that you know you have the JSON output from `backup`? I see you filed a PR, but it seems like there's no problem to solve given what you wrote in this issue :)

rrossouw01:
Correct personally I do not need diif --json for what I am doing.  Thanks for checking.

MichaelEischer:
I've noticed a small misunderstanding regarding the `--quiet` flag: Adding `--quiet` to the backup command currently doesn't disable the scanner, see #2336 .

darkdragon-001:
I do still have a need for this. Especially since the `backup` stats are wrong (#2756). Furthermore, they don't contain `others` (what is this anyways?). `diff` doesn't contain _dirs changed_ though.

dp-github:
What would you consider to be a 'dir changed' condition?

darkdragon-001:
@dp-github I just noticed it is output in the `backup` command. Meta-data (like modified date) could be a reason to qualify as _changed_, but I would prefer not. Since renames are not tracked if I know correctly, this indeed doesn't make sense. Thanks for the hint!

dp-github:
No problem. 

IIRC,  [my pull request](https://github.com/restic/restic/pull/2542#issuecomment-586637449) implementing this feature also honors --quiet.

darkdragon-001:
[...]

PR #2542, opened 2020-01-10
#2508: Adds support for `diff --json`

dp-github:
Resolves Issue #2508, adding support for `diff --json`
Note:  As discussed in #2508 this only includes summary diff statistics in the json output

* #2508

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2508
+++ b/changelog/unreleased/issue-2508
@@ -1,0 +1,5 @@
+Enhancement: Support `diff --json`
+
+`diff --json` suppresses detailed diff and reports only summary diff stats in JSON
+
+https://github.com/restic/restic/issues/2508
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"os"
 	"path"
 	"reflect"
 	"sort"
@@ -64,9 +67,12 @@ type Comparer struct {
 
 // DiffStat collects stats for all types of items.
 type DiffStat struct {
-	Files, Dirs, Others  int
-	DataBlobs, TreeBlobs int
-	Bytes                uint64
+	Files     int    `json:"files"`
+	Dirs      int    `json:"dirs"`
+	Others    int    `json:"others"`
+	DataBlobs int    `json:"data_blobs"`
+	TreeBlobs int    `json:"tree_blobs"`
+	Bytes     uint64 `json:"bytes"`
 }
 
 // Add adds stats information for node to s.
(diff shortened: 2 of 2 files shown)


rawtaz:
Can you please provide a complete and detailed example of output from this new feature?

dp-github:
You bet.  Here is a sample command-line and it's output:
````
# restic -r . diff --json 1b5d25bb 0541d4d4
{"source_snapshot":"1b5d25bb","target_snapshot":"0541d4d4","changed_files":0,"added":{"files":3,"dirs":0,"others":0,"data_blobs":0,"tree_blobs":5,"bytes":14851},"removed":{"files":0,"dirs":0,"others":0,"data_blobs":0,"tree_blobs":5,"bytes":13426}}
````
Here is the same output, json-prettified:
````
{
  "source_snapshot": "1b5d25bb",
  "target_snapshot": "0541d4d4",
  "changed_files": 0,
  "added": {
    "files": 3,
    "dirs": 0,
    "others": 0,
    "data_blobs": 0,
    "tree_blobs": 5,
    "bytes": 14851
  },
  "removed": {
    "files": 0,
    "dirs": 0,
    "others": 0,
    "data_blobs": 0,
    "tree_blobs": 5,
    "bytes": 13426
  }
}
````

**Comment by darkdragon-001** [comment]
Any progress here?

**Comment by MichaelEischer** [comment]
I've thought a bit about this PR and I don't think it's a good idea to omit a substantial portion of the diff command output when passing the `--json` flag. After all it is just expected to change the output formatting but not the content. So I think we rather should opt for a output format similar to that of the backup command, which prints one line per changed/unchanged file and a summary object at the end.

For the individual changes, we could output something like the verboseUpdate from the backup command. Although we'd probably want to limit the struct to for example `MessageType`, `Action` and `Item` or so.
```Go
type verboseUpdate struct {
	MessageType  string  `json:"message_type"` // "verbose_status"
	Action       string  `json:"action"`
	Item         string  `json:"item"`
	Duration     float64 `json:"duration"` // in seconds
	DataSize     uint64  `json:"data_size"`
	MetadataSize uint64  `json:"metadata_size"`
	TotalFiles   uint    `json:"total_files"`
}
```
The summary output should then also include a `MessageType` to allow discerning the different output lines.

Another option would be to behave like the `find` command which incrementally assemble a single large JSON object.

dp-github:
I will support whichever decision you make, but here's an argument for the
alternative:

Dont let the perfect be the enemy of the good.  This PR implements json
output where it's not currently available at all, and even if its
imperfect, that improvement. A future PR can fix the issues you raise.

darkdragon-001:
I would still vote for an implementation in combination with the `--quiet` flag (https://github.com/restic/restic/issues/2508#issuecomment-642825328). A single incrementally build json file would make parsing and error handling a lot easier for my use case (I wait for the command to be finished anyway).

MichaelEischer:
@dp-github The json output is considered part of the API of restic and thus we try to avoid breaking compatibility if possible. As the PR currently outputs a single JSON document and does not prepare for a streaming JSON output, currently the PR essentially makes the decision that the list of changed files should be part of the summary JSON message.

Regarding a single JSON document, I'm worried that this might become very large. Just consider a snapshot which contains a few million files. Then the diff output can reach a size of a few hundred MBs. In that case I'd prefer a output format that is easier to parse incrementally. Btw, to avoid that restic has to store the whole output in memory, a single JSON document would essentially  be an array with entries describing the change for each file. And that format is basically what you get when replacing the new lines in the streaming format and add square brackets around the result.

MichaelEischer:
Replaced by #3592

Issue #956, opened 2017-05-10
[FEATURE] exit code handling


unclesamwk:
Hey,

another request error/exit code handling.

Hey, nice feature is if restic abort answer with an error code more than 0 or 1.

rsync example:
0      Success
       1      Syntax or usage error
       2      Protocol incompatibility
       3      Errors selecting input/output files, dirs
       4      Requested  action not supported: an attempt was made to manipulate 64-bit files on a platform 
              that cannot support them; or an option was specified that is supported by the client and not by the server.
       5      Error starting client-server protocol
       6      Daemon unable to append to log-file
       10     Error in socket I/O
       11     Error in file I/O
       12     Error in rsync protocol data stream
       13     Errors with program diagnostics
       14     Error in IPC code
       20     Received SIGUSR1 or SIGINT
       21     Some error returned by waitpid()
       22     Error allocating core memory buffers
       23     Partial transfer due to error
       24     Partial transfer due to vanished source files
       25     The --max-delete limit stopped deletions
       30     Timeout in data send/receive
       35     Timeout waiting for daemon connection

For scripting restic its needed.

thanks

greetings
sam

fd0:
Hey, thanks for suggesting this. Can you please elaborate what your use case is? Why do you need to know exactly what the error code was? What are the cases you need to distinguish in your script and why?

wjkohnen:
Coming here because I've expected restic to exit with non-zero code when it received a signal. Otherwise it is hard to terminate a script after restic terminated on request. In my case the script just doesn't terminate, deletes data that was not backed up yet (restic said good bye, w/o backing up as requested) and the next backup job unintentionally starts up. I tried to fix this by using `trap` in my script, but that's brittle and still destroys data now and then. 

If everything exits with 0 no matter what, automation is hard. :slightly_smiling_face:

fd0:
@wjkohnen Huh, that doesn't sound right, but I'm not sure I understood. Could you elaborate on your use case a bit? I understood you're running restic in a script, then you send a signal to restic? Which signal, and why?

wjkohnen:
Well, it's quite simple. A bash script (with `set -e`) iterates in a for loop over some directories, in each loop there is a call to restic. I hit ^c, because $reason, restic clean's up due to SIGINT, restic exits with 0, for loop iterates because there's no one telling bash to stop. `trap "{ exit 27; }" SIGINT` doesn't help. 

[...]

PR #2546, opened 2020-01-12
Return an error when errors occured during backup

### LINKED ISSUE #2064: Filesystem-level errors are undetected: file subsequently omitted from backup, even though you're going to want to restore this file
Output of `restic version`
--------------------------

I'm not sure what the solution is, but a solution without solving a problem makes no sense, so let's start with the problem.

Currently, a filesystem-level checksum error (such as from Btrfs) is not detected by restic. Restic does not say anything explicitly, even with `-v`, and just omits the file from the "tree" (if I understand the output correctly).

While the "scrub" feature of such a filesystem is meant to catch checksumming errors, the last thing you want to find after scrubbing is that your backup no longer contains the file. Unfortunately, that is the current behaviour if you're not explicitly prepared for this precise failure.

The way to set this up correctly is to configure scrubbing, have any scrubbing errors emailed to you (or regularly do it manually, I guess), test it, and have a backup retention period longer than your scrub interval + however long you might take to look at it in a worst-case scenario (e.g. multi-week holiday). I feel like people are not only unaware of these necessary steps for their checksumming to be useful at all, they are also typically used to keeping only the latest backup: all you want is a copy to recover from, why even bother with keeping history? The chances of someone setting this up optimally are pretty low.

This could be made a lot more robust if restic detected the filesystem-level issue and handled it. That is the feature request, although I have no specific implementation in mind.

The only reasonable solution I can think of is to take the previous version of the file. Other ideas, like flagging the file and not removing it when someone uses `restic forget`, do not seem right. At a minimum, restic should return a non-zero exit status when it tried to read a file but failed to get its contents.

What are you trying to do?
--------------------------

I'm trying to protect my backup against bit rot. Without even having a lot of data (or maybe I just have really ambitious friends that make my couple terabytes seem like nothing), I've had a couple too many drive failures to be fully comfortable with a situation in which random errors could just propagate into backups completely undetected.

Restic generally works well and fast. I've recently started to use it for another system and while a write-only key would have been even more perfect ("most users think your software is pretty neat, but it would be perfect with this *one* extra feature!" I remember reading somewhere :-) ), it's pretty damn good compared to lots of other backup software! Especially encryption being a first-class citizen makes this really great.

### LINKED ISSUE #2364: Exit status 0 on backup and invocation errors; need to distinguish warnings from errors

### LINKED ISSUE #2526: Restic backup reports success exit status even when there were issues during backup
Related to #2364 and #2461 

Output of `restic version`
--------------------------
restic 0.9.5 compiled with go1.13beta1 on linux/amd64

```bash
$ restic init -r testrepo
enter password for new repository: 
Fatal: an empty password is not a password
$ restic init -r testrepo
enter password for new repository: 
enter password again: 
created restic repository be2d96646f at testrepo

irrecoverably lost.
$ mkdir testbak
$ touch testbak/testfile
$ chmod -r testbak/testfile 
$ restic -r testrepo backup testbak
repository be2d9664 opened successfully, password is correct
created new cache in /home/erkki/.cache/restic
error: open testbak/testfile: permission denied

Added to the repo: 494 B

snapshot ff4e4672 saved
$ echo $?
0
```

Expected behavior
-----------------

Restic returns non-zero exit status so that automated backups can be monitored for failure.

Actual behavior
---------------

Restic returns 0 exit status even though there were issues reading files during the backup process.

See above

Correct behavior was never implemented.

Backup function should return a special warning error for these types of non-fatal errors. In main.go this error should be detected with `error.Is()` and return a special warning exit status.

Everyday I receive an email informing me my scheduled backup was successful, I feel reassured my data is stored safely by restic!

Summerdave:
This PR intents to fix Issue #2526 
`Backup.Run` counts errors but does not take any action on them. This PR changes this behaviour by returning an error if errors occured previously.

Fixes #2064
Fixes #2526
Fixes #2364

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2546
+++ b/changelog/unreleased/pull-2546
@@ -1,0 +1,19 @@
+Change: Return exit code 3 when failing to backup all source data
+
+The backup command used to return a zero exit code as long as a snapshot
+could be created successfully, even if some of the source files could not
+be read (in which case the snapshot would contain the rest of the files).
+
+This made it hard for automation/scripts to detect failures/incomplete
+backups by looking at the exit code. Restic now returns the following exit
+codes for the backup command:
+
+ - 0 when the command was successful
+ - 1 when there was a fatal error (no snapshot created)
+ - 3 when some source data could not be read (incomplete snapshot created)
+
+https://github.com/restic/restic/pull/2546
+https://github.com/restic/restic/issues/956
+https://github.com/restic/restic/issues/2064
+https://github.com/restic/restic/issues/2526
+https://github.com/restic/restic/issues/2364
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -39,10 +39,9 @@ given as the arguments.
 EXIT STATUS
 ===========
 
-Exit status is 0 if the command was successful, and non-zero if there was any error.
-
-Note that some issues such as unreadable or deleted files during backup
-currently doesn't result in a non-zero error exit status.
+Exit status is 0 if the command was successful.
+Exit status is 1 if there was a fatal error (no snapshot created).
+Exit status is 3 if some source data could not be read (incomplete snapshot created).
 `,
 	PreRun: func(cmd *cobra.Command, args []string) {
 		if backupOptions.Host == "" {
@@ -98,6 +97,9 @@ type BackupOptions struct {
 }
 
 var backupOptions BackupOptions
+
+// Error sentinel for invalid source data
+var InvalidSourceData = errors.New("Failed to read all source data during backup.")
 
 func init() {
 	cmdRoot.AddCommand(cmdBackup)
@@ -557,7 +559,11 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	arch.SelectByName = selectByNameFilter
 	arch.Select = selectFilter
 	arch.WithAtime = opts.WithAtime
-	arch.Error = p.Error
+	success := true
+	arch.Error = func(item string, fi os.FileInfo, err error) error {
+		success = false
+		return p.Error(item, fi, err)
+	}
 	arch.CompleteItem = p.CompleteItem
 	arch.StartFile = p.StartFile
 	arch.CompleteBlob = p.CompleteBlob
@@ -593,6 +599,9 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	p.Finish(id)
 	if !gopts.JSON {
 		p.P("snapshot %s saved\n", id.Str())
+	}
+	if !success {
+		return InvalidSourceData
 	}
 
 	// Return error if any
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"syscall"
 	"testing"
@@ -54,7 +55,7 @@ func testRunInit(t testing.TB, opts GlobalOptions) {
 	t.Logf("repository initialized at %v", opts.Repo)
 }
 
-func testRunBackup(t testing.TB, dir string, target []string, opts BackupOptions, gopts GlobalOptions) {
+func testRunBackupAssumeFailure(t testing.TB, dir string, target []string, opts BackupOptions, gopts GlobalOptions) error {
 	ctx, cancel := context.WithCancel(gopts.ctx)
 	defer cancel()
 
@@ -69,7 +70,7 @@ func testRunBackup(t testing.TB, dir string, target []string, opts BackupOptions
 		defer cleanup()
 	}
 
-	rtest.OK(t, runBackup(opts, gopts, term, target))
+	backupErr := runBackup(opts, gopts, term, target)
 
 	cancel()
 
@@ -77,6 +78,13 @@ func testRunBackup(t testing.TB, dir string, target []string, opts BackupOptions
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	return backupErr
+}
+
+func testRunBackup(t testing.TB, dir string, target []string, opts BackupOptions, gopts GlobalOptions) {
+	err := testRunBackupAssumeFailure(t, dir, target, opts, gopts)
+	rtest.Assert(t, err == nil, "Error while backing up")
 }
 
 func testRunList(t testing.TB, tpe string, opts GlobalOptions) restic.IDs {
@@ -434,6 +442,36 @@ func TestBackupExclude(t *testing.T) {
 		"expected file %q not in first snapshot, but it's included", "foo.tar.gz")
 	rtest.Assert(t, !includes(files, "/testdata/private/secret/passwords.txt"),
 		"expected file %q not in first snapshot, but it's included", "passwords.txt")
+}
+
+func TestBackupErrors(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		return
+	}
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	datafile := filepath.Join("testdata", "backup-data.tar.gz")
+
+	rtest.SetupTarTestFixture(t, env.testdata, datafile)
+
+	testRunInit(t, env.gopts)
+
+	// Assume failure
+	inaccessibleFile := filepath.Join(env.testdata, "0", "0", "9", "0")
+	os.Chmod(inaccessibleFile, 0000)
+	defer func() {
+		os.Chmod(inaccessibleFile, 0644)
+	}()
+	opts := BackupOptions{}
+	gopts := env.gopts
+	gopts.stderr = ioutil.Discard
+	err := testRunBackupAssumeFailure(t, filepath.Dir(env.testdata), []string{"testdata"}, opts, gopts)
+	rtest.Assert(t, err != nil, "Assumed failure, but no error occured.")
+	rtest.Assert(t, err == InvalidSourceData, "Wrong error returned")
+	snapshotIDs := testRunList(t, "snapshots", env.gopts)
+	rtest.Assert(t, len(snapshotIDs) == 1,
+		"expected one snapshot, got %v", snapshotIDs)
 }
 
 const (
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -103,9 +103,13 @@ func main() {
 	}
 
 	var exitCode int
-	if err != nil {
+	switch err {
+	case nil:
+		exitCode = 0
+	case InvalidSourceData:
+		exitCode = 3
+	default:
 		exitCode = 1
 	}
-
 	Exit(exitCode)
 }
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -366,7 +366,6 @@ created as it would only be written at the very (successful) end of
 the backup operation.  Previous snapshots will still be there and will still
 work.
 
-
 Environment Variables
 *********************
 
@@ -424,3 +423,26 @@ are taken into account for various operations:
  * ``$XDG_CACHE_HOME/restic``, ``$HOME/.cache/restic``: :ref:`caching`.
  * ``$TMPDIR``: :ref:`temporary_files`.
  * ``$PATH/fusermount``: Binary for ``restic mount``.
+
+Exit status codes
+*****************
+
+Restic returns one of the following exit status codes after the backup command is run:
+
+ * 0 when the backup was successful (snapshot with all source files created)
+ * 1 when there was a fatal error (no snapshot created)
+ * 3 when some source files could not be read (incomplete snapshot with remaining files created)
+
+Fatal errors occur for example when restic is unable to write to the backup destination, when
+there are network connectivity issues preventing successful communication, or when an invalid
+password or command line argument is provided. When restic returns this exit status code, one
+should not expect a snapshot to have been created.
+
+Source file read errors occur when restic fails to read one or more files or directories that
+it was asked to back up, e.g. due to permission problems. Restic displays the number of source
+file read errors that occurred while running the backup. If there are errors of this type,
+restic will still try to complete the backup run with all the other files, and create a
+snapshot that then contains all but the unreadable files.
+
+One can use these exit status codes in scripts and other automation tools, to make them aware of
+the outcome of the backup run. To manually inspect the exit code in e.g. Linux, run ``echo $?``.


rawtaz:
Are there any tests that might be affected or need changing? Should any tests be *added*?

We should definitely add a changelog entry for this, as it might "break" some people's scripted runs of restic.

rigtorp:
Definitely add a test. See my referenced issue with simple repro steps.

rigtorp:
@rawtaz Please link this PR to issue #2526

rawtaz:
It's already linked when you mention the other issue :)

bjoe2k4:
Does it make sense to have this configurable? As a windows user backing up a profile folder, there will with utmost certainty always be a folder which is inaccessible. I'd like to ignore these error but I would like to check the return code if something really serious is going wrong. So I'm quite happy with the current behavior.

rigtorp:
@bjoe2k4 This PR is not yet complete. I think the error code needs to be different between critical error and unaccesible files error.

Summerdave:
@bjoe2k4
If you try to back up a directory and some subdirectories are inaccessible, your backup is configured incorrectly. You should ignore these subdirectories or ignore the return code altogether.

doc/040_backup.rst
```diff
@@ -366,6 +366,12 @@ created as it would only be written at the very (successful) end of
 the backup operation.  Previous snapshots will still be there and will still
 work.
 
+Errors
+*********************
+
+When running a backup, restic displays the number of errors that occured while creating the backup.
```
rigtorp: Need to mention critical errors vs source read errors. "Snapshot will be completed even if some source data could not be read correctly. Critical errors such as failure to write to the backup destination not generate a snapshot."

internal/ui/json/backup.go
```diff
@@ -107,6 +108,9 @@ func (b *Backup) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
+			if errorCount > 0 {
+				return errors.Errorf("Errors occured during backup.")
```
rigtorp: What about "Failed to read all source data during backup"? It is more descriptive.

internal/ui/backup.go
```diff
@@ -92,6 +93,9 @@ func (b *Backup) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
+			if errors > 0 {
```
MichaelEischer, line 96: This error check is, in my opinion, in the wrong place. The status UI should not decide whether an error in the archiver code has occurred or not. Or using a different terminology: The business logic should be separated from the user interface code. The need to modify both the normal text and the JSON output code is also a symptom of this problem. The fact that the error is effectively reported via a side-channel makes the control flow rather hard to follow.

I'd propose that the error check should happen here, by wrapping the `Error` function:

https://github.com/restic/restic/blob/c03bc88b297f960868d171e098771255f6f11be1/cmd/restic/cmd_backup.go#L560

MichaelEischer (changes requested):
I wonder what the fallout of this change would be. For plain cronjob runs of restic probably nothing will change as cron ignores the status code. The main difference will be for scripts / other tools that properly check the return code of restic. These should then fail with a loud bang. No subtle data loss, it just needs some change to the tooling, so this doesn't sound too bad.

However, exit code 1 can now mean two totally different things: either that some files couldn't be backed up OR that the backup run **failed completely**. I'm afraid that this will lead several users with incomplete excludes to ignore non-zero exits code and then wonder later on why their backup didn't work. So this poses the danger of exchanging files missing from a backup for completely missing backups. Yes, you can detect whether the backup worked by searching for `snapshot %s saved` in the normal text output or it's equivalent for JSON. But given that the motivation of this PR is to not have to parse the output of restic, this shouldn't count as a solution.

Solving this properly would require different exit code as proposed in #956 .

rigtorp:
> However, exit code 1 can now mean two totally different things: either that some files couldn't be backed up OR that the backup run **failed completely**. I'm afraid that this will lead several users with incomplete excludes to ignore non-zero exits code and then wonder later on why their backup didn't work. 

I agree https://github.com/restic/restic/pull/2546#issuecomment-583794488

From go 1.13 we can use the `errors.Is()` to categorize errors (https://blog.golang.org/go1.13-errors). The error would be converted into a specific return code here: https://github.com/restic/restic/blob/master/cmd/restic/main.go#L107

cmd/restic/main.go
```diff
@@ -104,7 +104,11 @@ func main() {
 
 	var exitCode int
 	if err != nil {
-		exitCode = 1
+		if errors.Is(err, errors.InvalidSourceData) {
+			exitCode = 2
```
MichaelEischer, line 108: Go already uses exit code 2 when crashing due to an uncaught panic: https://github.com/golang/go/commit/f947c4dcfe6f22e6c92e6e5a170096b518a18448

changelog/unreleased/pull-2546
```diff
@@ -0,0 +1,8 @@
+Change: Exit status 2 when errors occured while reading source data. Previously,
```
MichaelEischer, line 1: This must be a _single_ line. Currently this gets cut off in the changelog:
```
 * Chg #2482: Remove vendored dependencies
 * Chg #2546: Exit status 2 when errors occured while reading source data. Previously,
 * Chg #2600: Require Go >= 1.11
```

What about `Exit with status 2 if some source files could not be read`?

changelog/unreleased/pull-2546
```diff
@@ -0,0 +1,8 @@
+Change: Exit status 2 when errors occured while reading source data. Previously, 
+		Restic just created the incomplete snapshot while exiting with status 0.
```
MichaelEischer, line 2: "incomplete" is slightly ambiguous here: It could either mean that the snapshot creation skipped a few files or that the repository misses part of the snapshot data. The current wording is also unclear whether restic still creates a snapshot after this change or not.

cmd/restic/cmd_backup.go
```diff
@@ -557,7 +557,10 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	arch.SelectByName = selectByNameFilter
 	arch.Select = selectFilter
 	arch.WithAtime = opts.WithAtime
-	arch.Error = p.Error
+	arch.Error = func(item string, fi os.FileInfo, err error) error {
+		arch.Success = false
```
MichaelEischer, line 561: Please use a local variable. It's no good to have to look at _both_ `cmd_backup.go` and `archiver.go`.

cmd/restic/integration_test.go
```diff
@@ -436,6 +443,34 @@ func TestBackupExclude(t *testing.T) {
 		"expected file %q not in first snapshot, but it's included", "passwords.txt")
 }
 
+func TestBackupErrors(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
MichaelEischer, line 447: Why is this test skipped on windows?

cmd/restic/integration_test.go
```diff
@@ -69,7 +70,9 @@ func testRunBackup(t testing.TB, dir string, target []string, opts BackupOptions
 		defer cleanup()
 	}
 
-	rtest.OK(t, runBackup(opts, gopts, term, target))
+	failureOccured := runBackup(opts, gopts, term, target) != nil
```
MichaelEischer, line 73: Please adjust that test function to pass up the returned error and then check for the expected error type in TestBackupErrors.

doc/040_backup.rst
```diff
@@ -366,6 +366,19 @@ created as it would only be written at the very (successful) end of
 the backup operation.  Previous snapshots will still be there and will still
 work.
 
+Errors
+*********************
+
+There are two kinds of errors that can occur during backing up:
+The first kind are *source read errors*.
```
MichaelEischer, line 373: Please add a empty line before the explanation of both error types.

doc/040_backup.rst
```diff
@@ -366,6 +366,19 @@ created as it would only be written at the very (successful) end of
...
+These are errors while reading the source. Restic displays the number of
```
MichaelEischer, line 374: I'd add `files or directories` after `source`.

doc/040_backup.rst
```diff
@@ -366,6 +366,19 @@ created as it would only be written at the very (successful) end of
...
 work.
 
+Errors
+*********************
+
+There are two kinds of errors that can occur during backing up:
+The first kind are *source read errors*.
+These are errors while reading the source. Restic displays the number of 
+source read errors that occured while creating the backup.
+If any errors occured, restic still completes the snapshot but exists with a non-zero exit code.
```
MichaelEischer, line 376: This will need to be adjusted to explain the different error codes.

MichaelEischer (changes requested):
The implementation is going in the right direction, however the recent changes need some additional changes for consistency. I'm not yet sure how to best allocate exit codes to different error types.

Please also adjust the exit status code description in the `cobra.Command` command definition of the backup command.

Oh and the commits must be squashed as they add and remove a complete restic binary.

internal/errors/errors.go
```diff
@@ -26,6 +26,11 @@ var Wrapf = errors.Wrapf
 // returns nil.
 var WithMessage = errors.WithMessage
 
+var Is = func(err1, err2 error) bool { return err1 == err2 }
```
MichaelEischer, line 29: This should either be a proper `Is` function or be removed.

cmd/restic/main.go
Summerdave: Are there any other exit codes go uses or is 3 okay?

Summerdave: Are there any other exit codes that go uses or is using 3 okay?

Summerdave:
> Oh and the commits must be squashed as they add and remove a complete restic binary.

I removed the binary via interactive rebase. I would still propose to sqash the commits, as there are quite a few of them by now.

cmd/restic/integration_test.go
Summerdave: Enabling this test on windows would require creating a file which can not be read by restic. Since windows uses an ACL that is not POSIX-compliant, i don't see a straighforward way to implement this.

MichaelEischer: Seems like the archiver also skips the test with unreadable files. So skipping the test is fine.

cmd/restic/main.go
```diff
@@ -104,7 +104,11 @@ func main() {
 
 	var exitCode int
 	if err != nil {
-		exitCode = 1
+		if err == errors.InvalidSourceData {
+			exitCode = 3
+		} else {
+			exitCode = 1
+		}
```
greatroar, line 111:
```
switch err {
case nil:
	exitCode = 0
case InvalidSourceData:
	exitCode = 3
default:
	exitCode = 1
}
```

internal/errors/errors.go
```diff
@@ -26,6 +26,9 @@ var Wrapf = errors.Wrapf
 // returns nil.
 var WithMessage = errors.WithMessage
 
+// Error sentinel for invalid source data
+var InvalidSourceData = New("Failed to read all source data during backup.")
```
greatroar, line 30: Since this is only used in cmd, wouldn't it be better to just put it cmd_backup.go?

If this comes across as nitpicking: the more stuff goes in internal/restic, the harder it becomes to refactor code. I've hit import cycles on several attempts already.

Elgarni:
> I wonder what the fallout of this change would be. For plain cronjob runs of restic probably nothing will change as cron ignores the status code. The main difference will be for scripts / other tools that properly check the return code of restic. These should then fail with a loud bang. No subtle data loss, it just needs some change to the tooling, so this doesn't sound too bad.

@MichaelEischer  As I understand, the concern here is about backward compatibility. 

I am thinking what If we added a flag  (e.g. `--ignore-errors`), where it defaults to `true` to maintain backward compatibility, and a value of `false` will change the exit status.

This way `restic` would be more opinionated, and either case is supported.

What do you think?

rawtaz:
> I am thinking what If we added a flag (e.g. --ignore-errors), where it defaults to true to maintain backward compatibility, and a value of false will change the exit status.

Ignoring errors sounds like a really really bad idea. Please don't. Instead encourage people to fix their scripts, and give a nice heads-up in the release notes (which anyone should be reading, and most people with more complex deployments are reading).

Summerdave:
@Elgarni 
Anyone interested in doing that can also alias `restic` to something like
```
restic $@; [[ $? == 0 || $? == 3 ]]
```

cmd/restic/main.go
MichaelEischer: Exit code 3 seems to be fine. http://www.tldp.org/LDP/abs/html/exitcodes.html only mentions 1 and 2 as used.

rigtorp:
@MichaelEischer BSD have some standardized exit codes: https://github.com/openbsd/src/blob/master/include/sysexits.h

MichaelEischer:
I didn't find a matching exit code in the BSD ones. The closest ones are `EX_NOINPUT` and `EX_IOERR`, but even those don't really capture "partial backup". As these exit codes use values from 64-78 we won't collide with those.

MichaelEischer:
@Summerdave Could you fix the review comments by @greatroar ? `cmd_backup.go` contains a documentation of the `EXIT STATUS`. Please add a short documentation there for exit code 1 and 3.

MichaelEischer (changes requested):
Please squash the commits into a single one. 20 commits are way too much for this PR.

I've a few further remarks on the changelog entry. As this PR might break existing workflows it's rather important to have a good changelog entry.

rawtaz:
One of the Travis checks failed, but it looks like an intermittent Travis problem rather than an actual test failing, so I force-pushed to see if it works better a second time.

MichaelEischer (approved):
LGTM.

lgommans:
(One of the issues this pull request fixed was raised by me.) Thanks everyone for the work on this! Much appreciated :)

PR #2568, opened 2020-02-06
corrected option s3.storage-class in documentation

flofeld:
Took me some time to figure out this error. All that was missing was a little "-".
s3.storageclass vs. s3.storage-class


rawtaz (approved):
Nice catch!

PR #2574, opened 2020-02-10
cmd/restic: fix dropped error

alrs:
This picks up a dropped error in the `cmd/restic` package.


fd0 (approved):
Good catch, thank you very much!

PR #2575, opened 2020-02-10
restic/internal/ui: fix gofmt nit that popped up in Go 1.13

alrs:
A stray EOL semicolon is causing Travis failures on Go 1.13, this should be a fix.


rawtaz (approved):
Thanks so much!

PR #2577, opened 2020-02-12
internal: Fix Code and Test Dropped Errors

alrs:
This picks off four dropped `error` variables in the `internal` package, three in tests and one in code.


rawtaz (approved):
LGFWICT 👀

fd0:
Awesome, thanks for taking the time to do this!

PR #2581, opened 2020-02-17
Allow multiple hostnames tags

### LINKED ISSUE #1570: forget should take multiple hostnames (since it can take multiple tags already)
The documentation at http://restic.readthedocs.io/en/latest/060_forget.html#removing-snapshots-according-to-a-policy explains how you can specify multiple tags when removing old backup revision. While you technically can specify multiple --hostname values as well, only the last is actually evaluated for the command.

## Output of `restic version`
```
restic 0.8.1
```

```
backup my-remote forget --hostname hostname1 --hostname hostname2 --keep-hourly 4 --keep-daily 8 --keep-weekly 5 --keep-monthly 13 --keep-yearly 10
```

I am using https://github.com/binarybucks/restic-tools as a wrapper to easily backup to S3/minio.

S3

## Expected behavior
List snapshots to keep and delete for all hostnames.

## Actual behavior

It only lists snapshots from the last specified hostname.

see above

Looks like only the last hostname is evaluated for the forget command.

Evaluate all hostnames (just like its already the case for tags).

that's a very nice last point :-D. I've only been playing with restic for a brief time. But so far it looks very nice and extremely easy to setup. Looking forward to gather more experience with it.

aawsome:

allows multiple hostnames for `dump`, `find`, `forget`, `ls`, `mount`,
`restore`, `snapshots`, `stats` and `tag`

closes #1570 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1570
+++ b/changelog/unreleased/issue-1570
@@ -1,0 +1,9 @@
+Enhancement: Support specifying multiple host flags for various commands
+
+Previously commands didn't take more than one `--host` or `-H` argument into account, which could be limiting with e.g.
+the `forget` command.
+
+The `dump`, `find`, `forget`, `ls`, `mount`, `restore`, `snapshots`, `stats` and `tag` commands will now take into account
+multiple `--host` and `-H` flags.
+
+https://github.com/restic/restic/issues/1570
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -43,7 +43,7 @@ a path separator); paths use the forward slash '/' as separator.
 // LsOptions collects all options for the ls command.
 type LsOptions struct {
 	ListLong  bool
-	Host      string
+	Hosts     []string
 	Tags      restic.TagLists
 	Paths     []string
 	Recursive bool
@@ -56,7 +56,7 @@ func init() {
 
 	flags := cmdLs.Flags()
 	flags.BoolVarP(&lsOptions.ListLong, "long", "l", false, "use a long listing format showing size and mode")
-	flags.StringVarP(&lsOptions.Host, "host", "H", "", "only consider snapshots for this `host`, when no snapshot ID is given")
+	flags.StringArrayVarP(&lsOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
 	flags.Var(&lsOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot ID is given")
 	flags.StringArrayVar(&lsOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot ID is given")
 	flags.BoolVar(&lsOptions.Recursive, "recursive", false, "include files in subfolders of the listed directories")
@@ -84,7 +84,7 @@ type lsNode struct {
 }
 
 func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
-	if len(args) == 0 && opts.Host == "" && len(opts.Tags) == 0 && len(opts.Paths) == 0 {
+	if len(args) == 0 && len(opts.Hosts) == 0 && len(opts.Tags) == 0 && len(opts.Paths) == 0 {
 		return errors.Fatal("Invalid arguments, either give one or more snapshot IDs or set filters.")
 	}
 
@@ -186,7 +186,7 @@ func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
 		}
 	}
 
-	for sn := range FindFilteredSnapshots(ctx, repo, opts.Host, opts.Tags, opts.Paths, args[:1]) {
+	for sn := range FindFilteredSnapshots(ctx, repo, opts.Hosts, opts.Tags, opts.Paths, args[:1]) {
 		printSnapshot(sn)
 
 		err := walker.Walk(ctx, repo, *sn.Tree, nil, func(_ restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
--- a/cmd/restic/find.go
+++ b/cmd/restic/find.go
@@ -8,7 +8,7 @@ import (
 )
 
 // FindFilteredSnapshots yields Snapshots, either given explicitly by `snapshotIDs` or filtered from the list of all snapshots.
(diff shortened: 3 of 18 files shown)


changelog/unreleased/issue-1570
```diff
@@ -0,0 +1,5 @@
+Enhancement: Multiple hostnames for `dump`, `find`, `forget`, `ls`, `mount`, `restore`, `snapshots`, `stats`, `tag`.
+
+Added the possibility to specify multiple `--hostname` or `-H` flags where it makes sense.
```
rawtaz: The title line can't end with a dot, it shouldn't have any punctuation. I also suggest shortening it, please consider changing these three lines to the following (note that I also changed `--hostname` to `--host` since the former is deprecated):

~~~
Enhancement: Support specifying multiple host flags for various commands

Previously commands didn't take more than one `--host` or `-H` argument into account, which could be limiting with e.g. the `forget` command.

The `dump`, `find`, `forget`, `ls`, `mount`, `restore`, `snapshots`, `stats` and `tag` commands will now take into account multiple `--host` and `-H` flags.
~~~

rawtaz:
Nice work @aawsome! We need to adjust the changelog file a bit, once that's done I'll take a peek through the code again (just skimmed through it) and merge whenever possible.

internal/restic/snapshot_find.go
```diff
@@ -38,7 +38,12 @@ func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string,
 		if err != nil {
 			return errors.Errorf("Error loading snapshot %v: %v", snapshotID.Str(), err)
 		}
-		if snapshot.Time.Before(latest) || (hostname != "" && hostname != snapshot.Hostname) {
+
+		if snapshot.Time.Before(latest) {
```
fd0: Breaking up the complex condition makes the code so much easier to read, great idea!

internal/restic/snapshot_find.go
```diff
@@ -38,7 +38,12 @@ func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string,
...
+			return nil
+		}
+
+		if hostnames != nil && !snapshot.HasHostname(hostnames) {
```
fd0: Good name for the function! If `HasHostname()` has defined behavior for `hostnames` being an empty (or nil) slice the code here becomes even better :)

internal/restic/snapshot.go
```diff
@@ -201,6 +201,17 @@ func (sn *Snapshot) HasPaths(paths []string) bool {
 	return true
 }
 
+// HasHostname returns true if the snapshot hostname is in the list of the given hostnames.
+func (sn *Snapshot) HasHostname(hostnames []string) bool {
```
fd0: If `HasHostname()` accepts an empty slice and returns `true` in this case we can remove the `nil` checks below (needs to be mentioned in the description for the function though).

For slice emptyness, we should use `if len(hostnames) == 0`, this will match `hostnames == nil` as well as `hostnames == []string{}`.

cmd/restic/cmd_ls.go
```diff
@@ -84,7 +84,7 @@ type lsNode struct {
 }
 
 func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
-	if len(args) == 0 && opts.Host == "" && len(opts.Tags) == 0 && len(opts.Paths) == 0 {
+	if len(args) == 0 && opts.Hosts == nil && len(opts.Tags) == 0 && len(opts.Paths) == 0 {
```
fd0: Use `len(opts.Hosts) == 0` instead of checking equality to `nil`, that's more robust.

cmd/restic/find.go
```diff
@@ -39,7 +39,7 @@ func FindFilteredSnapshots(ctx context.Context, repo *repository.Repository, hos
 			}
 
 			// Give the user some indication their filters are not used.
-			if !usedFilter && (host != "" || len(tags) != 0 || len(paths) != 0) {
+			if !usedFilter && (hosts != nil || len(tags) != 0 || len(paths) != 0) {
```
fd0: Use `len(hosts) == 0` here, too

internal/restic/snapshot_find.go
```diff
@@ -92,7 +97,7 @@ func FindFilteredSnapshots(ctx context.Context, repo Repository, host string, ta
 			return nil
 		}
 
-		if (host != "" && host != sn.Hostname) || !sn.HasTagList(tags) || !sn.HasPaths(paths) {
+		if (hosts != nil && !sn.HasHostname(hosts)) || !sn.HasTagList(tags) || !sn.HasPaths(paths) {
```
fd0: Same here, the check for `nil` is not needed (and not entirely correct) as well.

fd0 (changes requested):
Awesome, thanks for your contribution! There's a few things I added comments to, in addition it'd be great if you could:
 1. Squash the commits
 2. Use a proper email address for the commits instead of `alex@thinkpad` (use `git config --global user.email` or something similar)

Thanks again!

internal/restic/snapshot_find.go
aawsome: I already thought about this and dismissed the idea for not being perfect consistent to what the name implies.. :smile: 
However, looking at `HasTagList()` there is already the same functionality. So I added this case, pointed it out in the description and also extended the description of `HasTagList()`. Now it is consistent that `Has"Object"` checks if one of the objects given in the list is satisfied and returns true for an empty list and `Has"Objects"` checks if all objects given in the list is satsified. IMO this makes sense.

internal/restic/snapshot.go
aawsome: is changed, see above.

cmd/restic/cmd_ls.go
aawsome: is changed.

cmd/restic/find.go
aawsome: is changed.

internal/restic/snapshot_find.go
aawsome: is changed.

rawtaz (approved):
Approving this after @aawsome made changes requested by @fd0.

Edit: Apparently @fd0 has to approve it as well after having requested changes, I can't merge it as it is now. That's good!

fd0 (approved):

PR #2583, opened 2020-02-18
Remove some unused or duplicated code

greatroar:

Some unused or duplicated code was removed.

No.

Checklist
---------


internal/archiver/tree.go
```diff
@@ -195,26 +195,13 @@ func (t *Tree) add(fs fs.FS, target, root string, pc []string) error {
 	return nil
 }
 
-func (t Tree) String() string {
```
MichaelEischer: This is not dead code. The String method is used via the `debug.Log` calls in `NewTree` (last function in tree)

internal/archiver/file_saver.go
```diff
@@ -53,7 +53,6 @@ type SaveBlobFn func(context.Context, restic.BlobType, *Buffer) FutureBlob
 
 // FileSaver concurrently saves incoming files to the repo.
 type FileSaver struct {
-	fs           fs.FS
```
MichaelEischer: LGTM

internal/archiver/archiver.go
```diff
@@ -628,42 +628,7 @@ func (arch *Archiver) SaveTree(ctx context.Context, snPath string, atree *Tree,
...
-
-func (fi fileInfoSlice) Swap(i, j int) {
-	fi[i], fi[j] = fi[j], fi[i]
-}
-
-func (fi fileInfoSlice) Less(i, j int) bool {
-	return fi[i].Name() < fi[j].Name()
-}
-
-func readdir(filesystem fs.FS, dir string) ([]os.FileInfo, error) {
```
MichaelEischer: I've always been confused by these two functions. It's good to have this cleaned up

internal/archiver/tree.go
```diff
@@ -195,26 +195,13 @@ func (t *Tree) add(fs fs.FS, target, root string, pc []string) error {
...
-	return s
-}
-
 // unrollTree unrolls the tree so that only leaf nodes have Path set.
 func unrollTree(f fs.FS, t *Tree) error {
 	// if the current tree is a leaf node (Path is set) and has additional
 	// nodes, add the contents of Path to the nodes.
 	if t.Path != "" && len(t.Nodes) > 0 {
 		debug.Log("resolve path %v", t.Path)
-		entries, err := fs.ReadDirNames(f, t.Path)
```
MichaelEischer: The `fs.ReadDirNames` and `readdirnames` functions have slightly different semantics. The latter sets `fs.O_NOFOLLOW` whereas the former does not. I'm right now not sure which is the correct behavior.

internal/archiver/archiver.go
```diff
@@ -216,7 +216,7 @@ func (arch *Archiver) SaveDir(ctx context.Context, snPath string, fi os.FileInfo
 		return FutureTree{}, err
 	}
 
-	names, err := readdirnames(arch.FS, dir)
+	names, err := readdirnames(arch.FS, dir, true)
```
MichaelEischer: just by looking at `readdirnames` calls, I can't tell the meaning of the third parameter. It probably better to split this up into `readdirnames` and `readdirnamessorted` (wrapping the first) functions.

MichaelEischer (changes requested):

internal/archiver/tree.go
greatroar: Ok, restored.

internal/archiver/archiver.go
greatroar: In that case, I'd rather just sort on the spot.

internal/archiver/tree.go
greatroar: I can't find any change in behavior and all the tests pass, but I can solve this by changing readdirnames's signature to

```
// flags are passed to fs.OpenFile. O_RDONLY is implied.
func readdirnames(filesystem fs.FS, dir string, flags int) ([]string, error) {
	f, err := filesystem.OpenFile(dir, flags|fs.O_RDONLY, 0)
	// ...
}
```

This is after I remove the sorted argument in favor of sorting on the spot. Is that ok?

internal/archiver/tree.go
greatroar: tree_test.go now fails to compile when this method is not present.

fd0 (approved):
LGTM, thanks a lot for taking the time to clean this up!

PR #2584, opened 2020-02-18
Cache uid and gid for top directories in internal/fuse

greatroar:

Restic mount now caches the uid and gid for use in its top-level directories (mountpoint, mountpoint/ids, etc.). Previously, it performed two system calls per Attr call.

Also, the test coverage in internal/fuse goes up from 16.7% to 38.8%.

No.

Checklist
---------

### DIFF
--- a/internal/fuse/dir.go
+++ b/internal/fuse/dir.go
@@ -98,8 +98,6 @@ func newDirFromSnapshot(ctx context.Context, root *Root, inode uint64, snapshot 
 	return &dir{
 		root: root,
 		node: &restic.Node{
-			UID:        uint32(os.Getuid()),
-			GID:        uint32(os.Getgid()),
 			AccessTime: snapshot.Time,
 			ModTime:    snapshot.Time,
 			ChangeTime: snapshot.Time,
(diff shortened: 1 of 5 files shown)


MichaelEischer (commented):
The changes look reasonable to me. Do you have any performance measurements / profiling data that show the performance gains of this change?

MichaelEischer (approved):

rawtaz:
@greatroar I'm curious too, do you have any profiling before and after?

greatroar:
@MichaelEischer @rawtaz

```
strace -c restic mount /tmp/restic
```

then find . -maxdepth 2 inside the mount point, gives (with this PR)

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
=== snip ===
  0.00    0.000000           0         2           getuid
  0.00    0.000000           0         2           getgid
=== snip ===
```

vs. (before)

```
  0.01    0.000157           2        77           getgid
  0.01    0.000076           1        77           getuid
```

This may get bigger when there are many tags, but it will likely never be noticeable. I'm still offering this PR as it reduces the code size and adds a test, but don't write a changelog entry for it and reject it if you don't think the change is worth it.

greatroar:
@MichaelEischer @rawtaz merge? @fd0 is still listed as pending reviewer, but only because I accidentally clicked "Request review".

greatroar:
Thanks! And good work cleaning out the PR queue.

rawtaz:
Thank *you*, for contributing :)

Issue #1680, opened 2018-03-22
Horrible fuse mount performance on large repo with many small files


bowensong:
## Output of `restic version`
```
restic 0.8.3 (a05bd98)
compiled with go1.10 on linux/amd64
```

## How did you run restic exactly?
`restic -r sftp:HOSTNAME:/path/to/restic/repo -p PASSWORD_FILE mount /path/to/mnt/point`
And
`restic -r sftp:HOSTNAME:/path/to/restic/repo -p PASSWORD_FILE --no-cache mount /path/to/mnt/point`

## What backend/server/service did you use to store the repository?
SFTP over Internet with high speed (100 Mbps, ping <25ms) access between the client and server

## Expected behavior
Randomly read small files with a reasonable latency (e.g. within a few seconds), and consumes reasonable amount of RAM and CPU resource.

## Actual behavior
With the cache, random reading small files (e.g.: `cat mountpoint/snapshots/latest/path/to/small/file.ext`) between a few bytes to a few MB takes many minutes because restic downloads entire pack file(s) (size between a few MB to many hundred MB) to the cache directory before serving the read request.
Without the cache, random reading small files in small directory is a lot faster. However when the directory contains millions of small files, the latency goes up to many minutes per file regardless of the files size.

Either way, the memory consummation and CPU usage are crappy, it uses all of the available memory on the client side (12.3GB out of 16GB physical RAM, the remaining memory are used by other applications), and the CPU usage often goes up to and stays at 100% for more than half a minute. The CPU usage is worse when `--no-cache` is given.

In addition to above, the disk usage with cache enabled is also concerning me, randomly access a few files resulted in a multi GB cache directory. I'm not sure if the cache size will go up linearly when more files are accessed.

## Steps to reproduce the behavior
Backup a 12TB directory with 65 million files spread in nearly 700 subdirectories. Then use `restic mount` from a different server, and randomly access the files in the mount point.

This is to simulate a disaster recovery scenario, where the main server which hosts the multi TB of data fails, and to keep the business going, we use a new server to do fuse mount and temporarily serve the files from it while we are restoring files to a different server.

## Do you have any idea what may have caused this?
For the memory issue, probably the golang GC? this is just a wild guess.
For the CPU issue, I think restic is decrypting (and then discarding?) extremely large amount of data.
For the latency, my guess is, with the cache restic spend time to download the entire pack files instead of only the needed part; without the cache restic spend time to download the index and decrypt it over and over again.

[...]

PR #2585, opened 2020-02-18
Smaller memory footprint for restic mount

greatroar:

restic mount used to cache the sizes of all blobs in the repo. By removing this cache, the memory footprint drops by 25%. To compensate for any loss of speed (which I didn't observe, but I only casually browsed through some old snapshots), the repo's index now allows concurrent access and its size lookup no longer needs to scan through all the packs.

Not that I know.

Checklist
---------


fbkopp:
Hello, please could you test the performance with large files + 50gb. I already had problems listing large files, so the file size cache was implemented.

greatroar:
I have a snapshot that contains a single 58GB file. When I ls that file, I get instant results:

```
$ time ls
bigfile.tar.bz2

real	0m0,019s
user	0m0,003s
sys	0m0,000s
```

Is that enough of a test, or do you have a suggestion for further benchmarking?

aawsome:
See also the discussion in #1680 

About your PR:

- I do agree about removing BlobSizeCache (if the index is fast enough to deliver the size). 
- Your index updates are covered (and extended a lot in means of memory usage) in the change discussed in #2523. My two cents: It doesn't help (much) to optimize here. Also MasterIndex should be optimized to prevent looping over potentially lots of index files. This leads shortly to big changes in the index implementation, this is why I proposed a re-implementation.

About the test setting: Testing one single large files is not enough as the `mount`performance is IMO mainly driven by listing dirs, i.e. loading tree blobs (of which you most likely have only one.
Index performance generally is depending on index entries and number of index files. Both are comparatively small for a repo with just one big file (average blob size should be around 8MB).

aawsome:
...when writing my last comment I realized that I should have checked whether reading tree blob is really the slow operation for listing files in FUSE. In fact it was not...
So I made #2587 which should solve the main problem and should IMO really make caching blob sizes unnecessary because it only reads the blob sizes when a file is read.

Thanks @greatroar for leading me to the right part to look at.
And sorry for proposing a "competing" PR... :-/

greatroar:
I had seen #2523, but decided to do a more localized optimization for the short term. I'd missed #1680.

Re: performance on lots of small files, it's hard to get consistent measurements because the kernel/FUSE seems to be caching the responses from restic. The first ls takes seconds for a dir of 10k small files, the second .06s, both with master and this PR. That also suggests that an LRU cache, which I considered as a potential further optimization, will be useless.

aawsome:
@greatroar Is it possible that you test your example with 10k of small files with #2587 ?
I think there also the first `ls` should be already very fast...

greatroar:
@aawsome Here's the test. Setup code:

```
export RESTIC_REPOSITORY=/tmp/testrepo
export RESTIC_PASSWORD=test

restic init
mkdir /tmp/testdata
for i in $(seq 10000); do
	touch /tmp/testdata/$i
done

restic backup /tmp/testdata
```

Then mount, and do

```
time ls $mountpoint/snapshots/latest/tmp/testinput
```

That gives, for master:

```
real	0m0,770s
user	0m0,034s
sys	0m0,102s
```

For this PR:
```
real	0m0,811s
user	0m0,036s
sys	0m0,119s
```

For #2587:
```
real	0m1,068s
user	0m0,051s
sys	0m0,153s
```

No big changes. Still, I'm closing in favor of #2587 because its changes are more localized.

PR #2586, opened 2020-02-18
Store *[]byte in internal/repository.bufPool, not []byte

greatroar:

internal/repository uses a sync.Pool to cache allocations. The following benchstat output shows what happens when I remove this pool:

```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.1ms ± 2%    36.5ms ± 2%    ~     (p=0.123 n=10+10)

name              old speed      new speed      delta
SaveAndEncrypt-8   116MB/s ± 2%   115MB/s ± 2%    ~     (p=0.128 n=10+10)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    21.1MB ± 0%    21.0MB ± 0%  -0.42%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      82.0 ± 0%      81.0 ± 0%  -1.22%  (p=0.000 n=10+10)
```

No significant drop in speed, but the number of allocations actually goes down when the pool is removed. The code in this PR uses the pool as intended, storing pointers instead of slices. The result is a tiny, but significant, speedup, as well as reduced allocation:

```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.1ms ± 2%    35.4ms ± 1%   -2.17%  (p=0.000 n=10+7)

name              old speed      new speed      delta
SaveAndEncrypt-8   116MB/s ± 2%   119MB/s ± 1%   +2.20%  (p=0.000 n=10+7)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    21.1MB ± 0%    17.0MB ± 0%  -19.34%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      82.0 ± 0%      80.0 ± 0%   -2.44%  (p=0.000 n=10+10)
```

No.

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -228,13 +228,13 @@ func (r *Repository) SaveAndEncrypt(ctx context.Context, t restic.BlobType, data
 	// get buf from the pool
 	ciphertext := getBuf()
 
-	ciphertext = ciphertext[:0]
+	*ciphertext = (*ciphertext)[:0]
 	nonce := crypto.NewRandomNonce()
-	ciphertext = append(ciphertext, nonce...)
+	*ciphertext = append(*ciphertext, nonce...)
 	defer freeBuf(ciphertext)
 
 	// encrypt blob
-	ciphertext = r.key.Seal(ciphertext, nonce, data, nil)
+	*ciphertext = r.key.Seal(*ciphertext, nonce, data, nil)
 
 	// find suitable packer and add blob
 	var pm *packerManager
@@ -254,7 +254,7 @@ func (r *Repository) SaveAndEncrypt(ctx context.Context, t restic.BlobType, data
 	}
 
 	// save ciphertext
-	_, err = packer.Add(t, *id, ciphertext)
+	_, err = packer.Add(t, *id, *ciphertext)
 	if err != nil {
 		return restic.ID{}, err
 	}
--- a/internal/repository/pool.go
+++ b/internal/repository/pool.go
@@ -6,16 +6,19 @@ import (
 	"github.com/restic/chunker"
 )
 
+// This pool stores pointers to []byte, for use in Repository.SaveAndEncrypt.
+// See the example in the sync package docs for why pointers are used.
 var bufPool = sync.Pool{
 	New: func() interface{} {
-		return make([]byte, chunker.MaxSize/3)
+		buf := make([]byte, 0, chunker.MaxSize/3)
+		return &buf
 	},
 }
 
-func getBuf() []byte {
-	return bufPool.Get().([]byte)
+func getBuf() *[]byte {
+	return bufPool.Get().(*[]byte)
 }
 
-func freeBuf(data []byte) {
+func freeBuf(data *[]byte) {
 	bufPool.Put(data)
 }
--- a/internal/repository/repository_test.go
+++ b/internal/repository/repository_test.go
@@ -93,6 +93,8 @@ func TestSaveFrom(t *testing.T) {
 }
 
 func BenchmarkSaveAndEncrypt(t *testing.B) {
+	t.ReportAllocs()
+
 	repo, cleanup := repository.TestRepository(t)
 	defer cleanup()
 


internal/repository/repository.go
```diff
@@ -228,13 +228,13 @@ func (r *Repository) SaveAndEncrypt(ctx context.Context, t restic.BlobType, data
...
-	ciphertext = ciphertext[:0]
+	*ciphertext = (*ciphertext)[:0]
 	nonce := crypto.NewRandomNonce()
-	ciphertext = append(ciphertext, nonce...)
+	*ciphertext = append(*ciphertext, nonce...)
 	defer freeBuf(ciphertext)
 
 	// encrypt blob
-	ciphertext = r.key.Seal(ciphertext, nonce, data, nil)
+	*ciphertext = r.key.Seal(*ciphertext, nonce, data, nil)
```
MichaelEischer: Shouldn't it be enough to move the `defer freeBuf(ciphertext)` line below the call to `r.key.Seal`?
It looks to me that the problem here is that the call to freeBuf puts the original `ciphertext` buffer back into the pool and not the larger one after sealing.

The effect should be similar if the `New` function for `sync.Pool` allocated a chunk with `chunker.MaxSize`?

greatroar: Good catch. If we postpone defer, the stats are:

old=master, new=later defer
```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.1ms ± 2%    35.7ms ± 2%     ~     (p=0.123 n=10+10)

name              old speed      new speed      delta
SaveAndEncrypt-8   116MB/s ± 2%   117MB/s ± 2%     ~     (p=0.123 n=10+10)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    21.1MB ± 0%    17.0MB ± 0%  -19.31%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      82.0 ± 0%      81.0 ± 0%   -1.22%  (p=0.000 n=10+10)
```

old=no pool, new=later defer
```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.5ms ± 2%    35.7ms ± 2%   -2.02%  (p=0.004 n=10+10)

name              old speed      new speed      delta
SaveAndEncrypt-8   115MB/s ± 2%   117MB/s ± 2%   +2.07%  (p=0.004 n=10+10)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    21.0MB ± 0%    17.0MB ± 0%  -18.97%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      81.0 ± 0%      81.0 ± 0%     ~     (all equal)
```

old=later defer, new=this PR
```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    35.7ms ± 2%    35.4ms ± 1%  -1.06%  (p=0.033 n=10+7)

name              old speed      new speed      delta
SaveAndEncrypt-8   117MB/s ± 2%   119MB/s ± 1%  +1.06%  (p=0.033 n=10+7)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    17.0MB ± 0%    17.0MB ± 0%  -0.03%  (p=0.029 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      81.0 ± 0%      80.0 ± 0%  -1.23%  (p=0.000 n=10+10)
```

Allocating chunker.MaxSize in addition to later defer gives:

old=master, new=later defer+bigger alloc
```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.1ms ± 2%    36.3ms ± 4%     ~     (p=0.631 n=10+10)

name              old speed      new speed      delta
SaveAndEncrypt-8   116MB/s ± 2%   116MB/s ± 4%     ~     (p=0.631 n=10+10)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    21.1MB ± 0%    17.0MB ± 1%  -19.49%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      82.0 ± 0%      81.0 ± 0%   -1.22%  (p=0.000 n=10+9)
```

old=later defer+bigger alloc, new=this PR
```
name              old time/op    new time/op    delta
SaveAndEncrypt-8    36.3ms ± 4%    35.4ms ± 1%  -2.63%  (p=0.005 n=10+7)

name              old speed      new speed      delta
SaveAndEncrypt-8   116MB/s ± 4%   119MB/s ± 1%  +2.66%  (p=0.005 n=10+7)

name              old alloc/op   new alloc/op   delta
SaveAndEncrypt-8    17.0MB ± 1%    17.0MB ± 0%    ~     (p=0.143 n=10+10)

name              old allocs/op  new allocs/op  delta
SaveAndEncrypt-8      81.0 ± 0%      80.0 ± 0%  -1.23%  (p=0.000 n=9+10)
```

In a single table:
```
name \ time/op    master        nopool        later-defer   later-defer-maxsize  pointers
SaveAndEncrypt-8   36.1ms ± 2%   36.5ms ± 2%   35.7ms ± 2%          36.3ms ± 4%   35.4ms ± 1%

name \ speed      master        nopool        later-defer   later-defer-maxsize  pointers
SaveAndEncrypt-8  116MB/s ± 2%  115MB/s ± 2%  117MB/s ± 2%         116MB/s ± 4%  119MB/s ± 1%

name \ alloc/op   master        nopool        later-defer   later-defer-maxsize  pointers
SaveAndEncrypt-8   21.1MB ± 0%   21.0MB ± 0%   17.0MB ± 0%          17.0MB ± 1%   17.0MB ± 0%

name \ allocs/op  master        nopool        later-defer   later-defer-maxsize  pointers
SaveAndEncrypt-8     82.0 ± 0%     81.0 ± 0%     81.0 ± 0%            81.0 ± 0%     80.0 ± 0%
```

fd0:
Interesting! I'm puzzled, what's going on here?

Where do the allocations come from, exactly?

If the pool is not needed (any more?), I'd rather remove it and make the code simpler and more robust instead of trying to squeeze every bit of speed out of it (after understanding what's going on here, of course). :)

It somehow feels wrong to me to save a pointer to the slice instead of the slice itself. We want to avoid allocating the (large) underlying array, so the slice structure itself is not of any interest.

If I understood [slices](https://blog.golang.org/go-slices-usage-and-internals) correctly, they consist of a pointer to the underlying array, a length and a cap field. Worst case if the structure for the slice needs to be reallocated each slice costs the runtime 24 byte (or so).

fd0:
Ah! I found the [example](https://pkg.go.dev/sync?tab=doc#example-Pool) the comment in the code refers to, and I think I understood what's going on: if a non-pointer type is used, returning it from New() and Get() in the pool copies its value.

MichaelEischer:
@fd0 `r.key.Seal` allocates a new array in the testcase as it has to handle a 4MB blob, whereas the buffer from the pool has only 8/3MB. That is the `ciphertext` slice returned by `r.key.Seal` uses a different backing buffer than the `ciphertext` slice that is passed in initially. `defer freeBuf(ciphertext)` captures a copy of the slice (just the descriptor, not the backing array) at the point in time the defer line is evaluated.

slices (that is the descriptor) in Go are by default passed by value. The only exception is when you explicitly get a pointer to a slice. In that case Go probably moves the slice from the stack to the heap. New() and Get() always return a copy of the value stored in the Pool. That can either be a copy of the slice (descriptor) or a copy of a pointer. By using pointers this PR changes the slices at which `*ciphertext` points. That way the call to defer will put the updated slice back into the pool.

greatroar:
To elaborate on @MichaelEischer's point about descriptors, [slices](https://github.com/golang/go/blob/8bdd24df6afa746cede600351feb2aa4395a63ea/src/runtime/slice.go#L13) are syntactic sugar for

```
type slice struct {
        array unsafe.Pointer
        len   int
        cap   int
}
```

These are passed around by value.

An [interface](https://github.com/golang/go/blob/8bdd24df6afa746cede600351feb2aa4395a63ea/src/runtime/runtime2.go#L200) is actually a

```
type iface struct {
        tab  *itab
        data unsafe.Pointer
}
```

which can only hold a pointer. When you store a non-pointer value in an interface value, the address of the value is stored instead, potentially causing it to be moved to the heap (depending on escape analysis). So even when the pool prevents allocation of the slice's backing array, it still requires heap allocation of the slice header.

I'll remove the pool in a separate PR.

PR #2587, opened 2020-02-18
Make `restic mount` faster and consume less memory

### LINKED ISSUE #1680: Horrible fuse mount performance on large repo with many small files

aawsome:

In the FUSE implementation restic data structure have been read (meaning loading blobs from the repository) and internal data structures have been created when a FUSE "node" was created. This was the case, for instance, for all item within a dir when this dir was accessed. This lead to quite some memory usage and made the FUSE filesystem "unresponsive". 

This PR changes this behavior and only reads restic data structure when the information is really needed. Moreover, the internal data structure have been optimized. Also concurrent operation within FUSE should now work correctly. (FUSE documentation says that access should be synced which wasn't)

closes #1680 

Thanks @greatroar for issuing #2585 which makes me think about whether the bottleneck really is only reading tree blobs.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1680
+++ b/changelog/unreleased/issue-1680
@@ -1,0 +1,8 @@
+Enhancement: Optimize `restic mount`
+
+We've optimized the FUSE implementation used within restic.
+`restic mount` is now more responsive and uses less memory.
+
+https://github.com/restic/restic/issues/1680
+https://github.com/restic/restic/pull/2587
+https://github.com/restic/restic/pull/2787
--- a/internal/fuse/dir.go
+++ b/internal/fuse/dir.go
@@ -5,6 +5,7 @@ package fuse
 import (
 	"os"
 	"path/filepath"
+	"sync"
 
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
@@ -24,6 +25,7 @@ type dir struct {
 	inode       uint64
 	parentInode uint64
 	node        *restic.Node
+	m           sync.Mutex
 }
 
 func cleanupNodeName(name string) string {
@@ -32,20 +34,10 @@ func cleanupNodeName(name string) string {
 
 func newDir(ctx context.Context, root *Root, inode, parentInode uint64, node *restic.Node) (*dir, error) {
 	debug.Log("new dir for %v (%v)", node.Name, node.Subtree)
-	tree, err := root.repo.LoadTree(ctx, *node.Subtree)
-	if err != nil {
-		debug.Log("  error loading tree %v: %v", node.Subtree, err)
-		return nil, err
-	}
-	items := make(map[string]*restic.Node)
-	for _, node := range tree.Nodes {
-		items[cleanupNodeName(node.Name)] = node
-	}
 
 	return &dir{
 		root:        root,
 		node:        node,
-		items:       items,
 		inode:       inode,
 		parentInode: parentInode,
 	}, nil
@@ -72,24 +64,6 @@ func replaceSpecialNodes(ctx context.Context, repo restic.Repository, node *rest
 
 func newDirFromSnapshot(ctx context.Context, root *Root, inode uint64, snapshot *restic.Snapshot) (*dir, error) {
 	debug.Log("new dir for snapshot %v (%v)", snapshot.ID(), snapshot.Tree)
-	tree, err := root.repo.LoadTree(ctx, *snapshot.Tree)
-	if err != nil {
-		debug.Log("  loadTree(%v) failed: %v", snapshot.ID(), err)
-		return nil, err
-	}
-	items := make(map[string]*restic.Node)
-	for _, n := range tree.Nodes {
-		nodes, err := replaceSpecialNodes(ctx, root.repo, n)
-		if err != nil {
-			debug.Log("  replaceSpecialNodes(%v) failed: %v", n, err)
-			return nil, err
-		}
-
-		for _, node := range nodes {
-			items[cleanupNodeName(node.Name)] = node
-		}
-	}
-
 	return &dir{
 		root: root,
 		node: &restic.Node{
@@ -97,10 +71,40 @@ func newDirFromSnapshot(ctx context.Context, root *Root, inode uint64, snapshot 
 			ModTime:    snapshot.Time,
 			ChangeTime: snapshot.Time,
 			Mode:       os.ModeDir | 0555,
+			Subtree:    snapshot.Tree,
 		},
-		items: items,
 		inode: inode,
 	}, nil
+}
+
+func (d *dir) open(ctx context.Context) error {
+	d.m.Lock()
+	defer d.m.Unlock()
+
+	if d.items != nil {
+		return nil
+	}
+
+	debug.Log("open dir %v (%v)", d.node.Name, d.node.Subtree)
+
+	tree, err := d.root.repo.LoadTree(ctx, *d.node.Subtree)
+	if err != nil {
+		debug.Log("  error loading tree %v: %v", d.node.Subtree, err)
+		return err
+	}
+	items := make(map[string]*restic.Node)
+	for _, n := range tree.Nodes {
+		nodes, err := replaceSpecialNodes(ctx, d.root.repo, n)
+		if err != nil {
+			debug.Log("  replaceSpecialNodes(%v) failed: %v", n, err)
+			return err
+		}
+		for _, node := range nodes {
+			items[cleanupNodeName(node.Name)] = node
+		}
+	}
+	d.items = items
+	return nil
 }
 
 func (d *dir) Attr(ctx context.Context, a *fuse.Attr) error {
@@ -133,6 +137,10 @@ func (d *dir) calcNumberOfLinks() uint32 {
 
 func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 	debug.Log("ReadDirAll()")
+	err := d.open(ctx)
+	if err != nil {
+		return nil, err
+	}
 	ret := make([]fuse.Dirent, 0, len(d.items)+2)
 
 	ret = append(ret, fuse.Dirent{
@@ -171,6 +179,12 @@ func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 
 func (d *dir) Lookup(ctx context.Context, name string) (fs.Node, error) {
 	debug.Log("Lookup(%v)", name)
+
+	err := d.open(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	node, ok := d.items[name]
 	if !ok {
 		debug.Log("  Lookup(%v) -> not found", name)
--- a/internal/fuse/file.go
+++ b/internal/fuse/file.go
@@ -5,10 +5,9 @@ package fuse
 import (
 	"sort"
 
+	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
-
-	"github.com/restic/restic/internal/debug"
 
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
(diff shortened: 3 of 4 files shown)


internal/fuse/file.go
```diff
@@ -111,6 +90,27 @@ func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadR
 	debug.Log("Read(%v, %v, %v), file size %v", f.node.Name, req.Size, req.Offset, f.node.Size)
 	offset := req.Offset
 
+	if !f.hasSizes {
```
greatroar: Shouldn't this be protected by a lock? Or does FUSE handle that?

greatroar: Never mind. FUSE calls Lookup, and Lookup returns a fresh *file every time. So this is safe.

aawsome: Thanks for researching! I wasn't sure about how FUSE handles concurrency, but `getBlobAt` also seems not to be concurrency-safe, so I assumed it would be handled by FUSE...

greatroar: Actually, I failed to spot that the bazil.org/fuse package stores the node it gets from Lookup in a cache, and it's concurrent. So there really should be locks all over internal/fuse.

greatroar: False alarm. It looks like the kernel driver will synchronize access, as long as you don't set the AsyncRead MountOption. I've empirically verified this by setting minSnapshotsReloadTime to 0, running restic mount with the race detector on, reading the snapshots directory concurrently in two loops and adding snapshots.

MichaelEischer: The documentation of bazil/fuse actually says that synchronization is required (see https://godoc.org/bazil.org/fuse#hdr-Service_Methods):

> Multiple goroutines may call service methods simultaneously; the methods being called are responsible for appropriate synchronization.

I'd assume that some caching in the kernel effectively synchronizes access, but it probably not a good idea to rely on it.

MichaelEischer:
It might be an option to implement the `NodeOpener` which is called when a file is opened. That way the chunk size collection would happen immediately when opening a file and is not delayed until reading the first bytes.

The main difference is the point in time when the user would be notified in case some blobs are missing. The current implementation in restic probably fails with an error before providing any details on the file node, with the change this is delayed until the first bytes are read. I think that the new semantic is actually what a user would expect: Reading a damaged file fails, but usually you are able to see the metadata.

internal/fuse/file.go
aawsome: I added a Mutex to synchronize access

aawsome:
@MichaelEischer Thanks for searching the manual and offering the NodeOpener option which I'm using now.
I tried to do performance tests of this new version but did not really succeed to see differences when comparing this to. In fact it seems that there a lot of variation w.r.t. performance when using FUSE (maybe some kernel caching?)...
Is there anyone who can help out with testing this change?

Memory footprint is however clearly reduced.

ProactiveServices:
@aawsome Happy to help, do you have any specific tests you'd like to perform? I have several hundred-gig repos I can test with.

aawsome:
@ProactiveServices Thank you very much! I'm especially interested in response times and CPU usage and comparison between "plain" restic and a version including this PR.
The code changes should apply for listing and reading files, e.g. `ls file` or `cat file`/`cp file somewhere`.

I used `/usr/bin/time restic mount` to get CPU timings and a simple `time` before the commands, e.g. `time ls file` to get response times.

aawsome:
With the last commit, also dir operations should be faster (and a bit less memory consuming)...

internal/fuse/file.go
```diff
@@ -20,47 +21,31 @@ import (
 const blockSize = 512
 
 // Statically ensure that *file implements the given interface
```
greatroar: This comment is now stale because one of the checks is for *openFile. You could update it, but removing it may be easier.

internal/fuse/file.go
```diff
@@ -136,9 +138,20 @@ func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadR
...
+
+	// The documentation of bazil/fuse actually says that synchronization is
+	// required (see https://godoc.org/bazil.org/fuse#hdr-Service_Methods):
+	//
+	// Multiple goroutines may call service methods simultaneously;
+	// the methods being called are responsible for appropriate synchronization.
+	//
+	// Hence Lock to ensure operation on blobs is not done concurrently
+	// (doesn't work with cached blob in getBlobAt)
+	f.m.Lock()
```
greatroar: defer f.m.Unlock()? There are multiple exits, and lack of memory can cause a panic.

aawsome:
rebased to master and added the changes suggested by @greatroar

internal/fuse/file.go
aawsome: the comment is now updated.

internal/fuse/file.go
aawsome: is changed.

aawsome:
Rebased this PR after #2790 has been merged.

aawsome:
Noticed that the new blob cache by @greatroar already implements locking; hence I removed the locks from `file.go` and just left a comment there.

aawsome:
rebased after #2787 has been merged.

internal/fuse/dir.go
```diff
@@ -20,10 +21,12 @@ var _ = fs.NodeStringLookuper(&dir{})
 
 type dir struct {
 	root        *Root
-	items       map[string]*restic.Node
 	inode       uint64
 	parentInode uint64
 	node        *restic.Node
+	isOpen      bool
```
MichaelEischer, line 27: `isOpen` seems redundant. Afaict `isOpen` == `items != nil`

internal/fuse/dir.go
```diff
@@ -72,35 +65,48 @@ func replaceSpecialNodes(ctx context.Context, repo restic.Repository, node *rest
...
+			ModTime:    snapshot.Time,
+			ChangeTime: snapshot.Time,
+			Mode:       os.ModeDir | 0555,
+			Subtree:    snapshot.Tree,
+		},
+		inode: inode,
+	}, nil
+}
+
+func (d *dir) Open() error {
```
MichaelEischer, line 81: Is this `Open` method used by any external caller or just internally by ReadDirAll and Lookup? In the latter case please change the signature to `func (d *dir) open(ctx context.Context) error`. That avoid also the problematic use of `context.TODO()`

internal/fuse/dir.go
```diff
@@ -133,6 +139,11 @@ func (d *dir) calcNumberOfLinks() uint32 {
 
 func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 	debug.Log("ReadDirAll()")
+	err := d.Open()
+	if err != nil {
+		return nil, err
+	}
+	debug.Log("called")
```
MichaelEischer, line 146: I think we can remove that `debug.Log` call

internal/fuse/dir.go
```diff
@@ -170,6 +181,10 @@ func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 }
 
 func (d *dir) Lookup(ctx context.Context, name string) (fs.Node, error) {
+	err := d.Open()
+	if err != nil {
+		return nil, err
+	}
 	debug.Log("Lookup(%v)", name)
```
MichaelEischer, line 188: please move the `debug.Log` call above `d.Open()`

changelog/unreleased/issue-1680
```diff
@@ -0,0 +1,7 @@
+Enhancement: Optimize `restic mount`
+
+We've optimized the FUSE implementation used within restic.
+`restic mount` is now more responsive and uses less memory.
+
+https://github.com/restic/restic/issues/1680
+https://github.com/restic/restic/pull/2587
```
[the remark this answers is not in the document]
aawsome: is added.

internal/fuse/dir.go
```diff
@@ -20,10 +21,12 @@ var _ = fs.NodeStringLookuper(&dir{})
 
 type dir struct {
 	root        *Root
-	items       map[string]*restic.Node
 	inode       uint64
 	parentInode uint64
 	node        *restic.Node
+	isOpen      bool
+	m           sync.Mutex
+	items       map[string]*restic.Node
```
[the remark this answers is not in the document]
aawsome: changed.

internal/fuse/dir.go
aawsome: Ah - thanks for the hint! Changed it.

internal/fuse/dir.go
aawsome: changed it to `open(context.Context)`

internal/fuse/dir.go
aawsome: yes - this was just to debug; is removed.

internal/fuse/dir.go
aawsome: is changed.

MichaelEischer (approved):
LGTM. I've force-pushed the branch to retrigger the CI. Seems like the flaky rclone test error has changed into a flaky rclone test crash, after #2855.

PR #2589, opened 2020-02-18
Replace sort.Stable by sort.Strings

greatroar:

archiver.SaveTree calls the slow, O(n lg² n) sort.Stable. But in context, it's equivalent to the more simple sort.Strings. This is meant to simplify, not optimize, but the code does get asymptotically faster.

Not that I know of.

Checklist
---------


aawsome (approved):
I agree that sort.Stable is not necessary for a string array.

fd0 (approved):
LGTM, thanks a lot!

PR #2598, opened 2020-02-22
Diff speedup

MichaelEischer:
Optimize diff calculation for shared subtrees.

When the diff calculation compares two trees with identical id then no
differences between them can ever show up. Optimize for that case by
simply traversing the tree only once to collect all referenced blobs for
a proper calculation of added and removed blobs.

Just skipping the common subtrees is not possible as this would skew the
results if the added or removed blobs are shared with one of the
subtrees.

This change can speed-up the diff calculation by up to a factor two. I'm not sure whether that's worth a changelog entry or not.

No.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2598
+++ b/changelog/unreleased/pull-2598
@@ -1,0 +1,6 @@
+Enhancement: Improve speed of diff command
+
+We've improved the performance of the diff command when comparing snapshots
+with similar content. It should run up to twice as fast as before.
+
+https://github.com/restic/restic/pull/2598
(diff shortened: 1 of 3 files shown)


greatroar:
Confirmed, this gives a big speedup in a quick test on two ~120GB snapshots with nearly equal data. Master:

```
real	0m49,034s
user	0m56,896s
sys	0m3,197s
```

This PR:

```
real	0m27,803s
user	0m34,930s
sys	0m1,678s
```

fd0 (approved):
LGTM! Thanks!

PR #2599, opened 2020-02-22
Reduce memory usage when searching for used blobs

MichaelEischer:
The seen BlobSet always contained a subset of the entries in blobs.
Thus use blobs instead and avoid the memory overhead of the second set.

No. 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_stats.go
+++ b/cmd/restic/cmd_stats.go
@@ -93,7 +93,6 @@ func runStats(gopts GlobalOptions, args []string) error {
 		uniqueInodes: make(map[uint64]struct{}),
 		fileBlobs:    make(map[string]restic.IDSet),
 		blobs:        restic.NewBlobSet(),
-		blobsSeen:    restic.NewBlobSet(),
 	}
 
 	if snapshotIDString != "" {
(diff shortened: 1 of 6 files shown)


aawsome (approved):
Thanks for proposing this PR! I also stumbled over the double BlobSet already and somewhere (I think it was in #2513) already called `FindUsedBlobs` with the seenBlobs = usedBlobs.

I think in the long term the remaining BlobSet should be integrated into the index - after #2523 is finished.
This PR however can and should be merged independently. Good job :+1:

MichaelEischer:
@aawsome I actually got the idea from your unusual call to `FindUsedBlobs` in #2513. Sorry, for not mentioning that, I first failed to find your PR again and then completely forgot about it while adding tests. You are now mentioned in a "Suggested-by:" line in the commit message.

cmd/restic/cmd_stats.go
```diff
@@ -318,9 +317,9 @@ type statsContainer struct {
 	// blobs that have been seen as a part of the file
 	fileBlobs map[string]restic.IDSet
 
-	// blobs and blobsSeen are used to count individual
+	// blobs are used to count individual
```
greatroar: are used => is used

fd0 (approved):
Great improvement :)

PR #2606, opened 2020-02-27
Add copy functionality.

### LINKED ISSUE #323: Add command to copy all data to another repository
During the discussion in #320 we discovered that functionality may be helpful to copy all data (data blobs, tree blobs, snapshots) from a repository to a new one, recreating pack files and indexes on the fly. This allows creating a new repository in a different location (e.g. moving from a local repository to an sftp-server) and using that from now on without losing any history and old snapshots.

This issues tracks the implementation of this feature and can be closed when it is implemented.

middelink:
Add a copy command to copy snapshots between repositories. It allows the user
to specify a destination repository, password, password-file, password-command
or key-hint to supply the nessecary details to open the destination repository.

You need to supply a list of snapshots to copy, snapshots which alreadys exist
in the destination repository will be skipped.

Note, when using the network this becomes rather slow, as it needs to read the
blocks, decrypt them using the source key, then encrypt them again using the
destination key before finally writing them out to the destination repository.

closes #323 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-323
+++ b/changelog/unreleased/issue-323
@@ -1,0 +1,13 @@
+Enhancement: Add command for copying snapshots between repositories
+
+We've added a copy command, allowing you to copy snapshots from one
+repository to another.
+
+Note that this process will have to read (download) and write (upload) the
+entire snapshot(s) due to the different encryption keys used on the source
+and destination repository. Also, the transferred files are not re-chunked,
+which may break deduplication between files already stored in the
+destination repo and files copied there using this command.
+
+https://github.com/restic/restic/issues/323
+https://github.com/restic/restic/pull/2606
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -30,6 +30,7 @@ their deduplication.
 // CopyOptions bundles all options for the copy command.
 type CopyOptions struct {
 	Repo            string
+	password        string
 	PasswordFile    string
 	PasswordCommand string
 	KeyHint         string
@@ -64,9 +65,13 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 	dstGopts.PasswordFile = opts.PasswordFile
 	dstGopts.PasswordCommand = opts.PasswordCommand
 	dstGopts.KeyHint = opts.KeyHint
-	dstGopts.password, err = resolvePassword(dstGopts, "RESTIC_PASSWORD2")
-	if err != nil {
-		return err
+	if opts.password != "" {
+		dstGopts.password = opts.password
+	} else {
+		dstGopts.password, err = resolvePassword(dstGopts, "RESTIC_PASSWORD2")
+		if err != nil {
+			return err
+		}
 	}
 	dstGopts.password, err = ReadPassword(dstGopts, "enter password for destination repository: ")
 	if err != nil {
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -108,11 +108,13 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		return err
 	}
 
+	visitedTrees := restic.NewIDSet()
+
 	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
 		Verbosef("snapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
 		Verbosef("  copy started, this may take a while...\n")
 
-		if err := copyTree(ctx, srcRepo, dstRepo, *sn.Tree); err != nil {
+		if err := copyTree(ctx, srcRepo, dstRepo, *sn.Tree, visitedTrees); err != nil {
 			return err
 		}
 		debug.Log("tree copied")
@@ -134,11 +136,18 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 	return nil
 }
 
-func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID) error {
+func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID, visitedTrees restic.IDSet) error {
+	// We have already processed this tree
+	if visitedTrees.Has(treeID) {
+		return nil
+	}
+
 	tree, err := srcRepo.LoadTree(ctx, treeID)
 	if err != nil {
 		return fmt.Errorf("LoadTree(%v) returned error %v", treeID.Str(), err)
 	}
+	visitedTrees.Insert(treeID)
+
 	// Do we already have this tree blob?
 	if !dstRepo.Index().Has(treeID, restic.TreeBlob) {
 		newTreeID, err := dstRepo.SaveTree(ctx, tree)
@@ -157,7 +166,7 @@ func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID re
 	for _, entry := range tree.Nodes {
 		// If it is a directory, recurse
 		if entry.Type == "dir" && entry.Subtree != nil {
-			if err := copyTree(ctx, srcRepo, dstRepo, *entry.Subtree); err != nil {
+			if err := copyTree(ctx, srcRepo, dstRepo, *entry.Subtree, visitedTrees); err != nil {
 				return err
 			}
 		}
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -113,7 +113,6 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		return err
 	}
 
-	visitedTrees := restic.NewIDSet()
 	dstSnapshotByOriginal := make(map[restic.ID][]*restic.Snapshot)
 	for sn := range FindFilteredSnapshots(ctx, dstRepo, opts.Hosts, opts.Tags, opts.Paths, nil) {
 		if sn.Original != nil && !sn.Original.IsNull() {
@@ -121,6 +120,13 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		}
 		// also consider identical snapshot copies
 		dstSnapshotByOriginal[*sn.ID()] = append(dstSnapshotByOriginal[*sn.ID()], sn)
+	}
+
+	cloner := &treeCloner{
+		srcRepo:      srcRepo,
+		dstRepo:      dstRepo,
+		visitedTrees: restic.NewIDSet(),
+		buf:          nil,
 	}
 
 	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
@@ -146,7 +152,7 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		}
 		Verbosef("  copy started, this may take a while...\n")
 
-		if err := copyTree(ctx, srcRepo, dstRepo, *sn.Tree, visitedTrees); err != nil {
+		if err := cloner.copyTree(ctx, *sn.Tree); err != nil {
 			return err
 		}
 		debug.Log("tree copied")
@@ -190,21 +196,28 @@ func similarSnapshots(sna *restic.Snapshot, snb *restic.Snapshot) bool {
 	return true
 }
 
-func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID, visitedTrees restic.IDSet) error {
+type treeCloner struct {
+	srcRepo      restic.Repository
+	dstRepo      restic.Repository
+	visitedTrees restic.IDSet
+	buf          []byte
+}
+
+func (t *treeCloner) copyTree(ctx context.Context, treeID restic.ID) error {
 	// We have already processed this tree
-	if visitedTrees.Has(treeID) {
+	if t.visitedTrees.Has(treeID) {
 		return nil
 	}
 
-	tree, err := srcRepo.LoadTree(ctx, treeID)
+	tree, err := t.srcRepo.LoadTree(ctx, treeID)
 	if err != nil {
 		return fmt.Errorf("LoadTree(%v) returned error %v", treeID.Str(), err)
 	}
-	visitedTrees.Insert(treeID)
+	t.visitedTrees.Insert(treeID)
 
 	// Do we already have this tree blob?
-	if !dstRepo.Index().Has(treeID, restic.TreeBlob) {
-		newTreeID, err := dstRepo.SaveTree(ctx, tree)
+	if !t.dstRepo.Index().Has(treeID, restic.TreeBlob) {
+		newTreeID, err := t.dstRepo.SaveTree(ctx, tree)
 		if err != nil {
 			return fmt.Errorf("SaveTree(%v) returned error %v", treeID.Str(), err)
 		}
@@ -214,29 +227,28 @@ func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID re
 		}
 	}
 
-	// TODO: keep only one (big) buffer around.
 	// TODO: parellize this stuff, likely only needed inside a tree.
 
 	for _, entry := range tree.Nodes {
 		// If it is a directory, recurse
 		if entry.Type == "dir" && entry.Subtree != nil {
-			if err := copyTree(ctx, srcRepo, dstRepo, *entry.Subtree, visitedTrees); err != nil {
+			if err := t.copyTree(ctx, *entry.Subtree); err != nil {
 				return err
 			}
 		}
 		// Copy the blobs for this file.
 		for _, blobID := range entry.Content {
 			// Do we already have this data blob?
-			if dstRepo.Index().Has(blobID, restic.DataBlob) {
+			if t.dstRepo.Index().Has(blobID, restic.DataBlob) {
 				continue
 			}
 			debug.Log("Copying blob %s\n", blobID.Str())
-			buf, err := srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, nil)
+			t.buf, err = t.srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, t.buf)
 			if err != nil {
 				return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
 			}
 
-			_, _, err = dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID, false)
+			_, _, err = t.dstRepo.SaveBlob(ctx, restic.DataBlob, t.buf, blobID, false)
 			if err != nil {
 				return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
 			}
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -120,14 +120,7 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		if err = dstRepo.Flush(ctx); err != nil {
 			return err
 		}
-		debug.Log("flushed packs")
-
-		err = dstRepo.SaveIndex(ctx)
-		if err != nil {
-			debug.Log("error saving index: %v", err)
-			return err
-		}
-		debug.Log("saved index")
+		debug.Log("flushed packs and saved index")
 
 		// save snapshot
 		sn.Parent = nil   // Parent does not have relevance in the new repo.
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -1,0 +1,203 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/restic"
+
+	"github.com/spf13/cobra"
+)
+
+var cmdCopy = &cobra.Command{
+	Use:   "copy [flags] [snapshotID ...]",
+	Short: "Copy snapshots from one repository to another",
+	Long: `
+The "copy" command copies one or more snapshots from one repository to another
+repository. Note that this will have to read (download) and write (upload) the
+entire snapshot(s) due to the different encryption keys on the source and
+destination, and that transferred files are not re-chunked, which may break
+their deduplication.
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runCopy(copyOptions, globalOptions, args)
+	},
+}
+
+// CopyOptions bundles all options for the copy command.
+type CopyOptions struct {
+	Repo            string
+	PasswordFile    string
+	PasswordCommand string
+	KeyHint         string
+	Hosts           []string
+	Tags            restic.TagLists
+	Paths           []string
+}
+
+var copyOptions CopyOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdCopy)
+
+	f := cmdCopy.Flags()
+	f.StringVarP(&copyOptions.Repo, "repo2", "", os.Getenv("RESTIC_REPOSITORY2"), "destination repository to copy snapshots to (default: $RESTIC_REPOSITORY2)")
+	f.StringVarP(&copyOptions.PasswordFile, "password-file2", "", os.Getenv("RESTIC_PASSWORD_FILE2"), "read the destination repository password from a file (default: $RESTIC_PASSWORD_FILE2)")
+	f.StringVarP(&copyOptions.KeyHint, "key-hint2", "", os.Getenv("RESTIC_KEY_HINT2"), "key ID of key to try decrypting the destination repository first (default: $RESTIC_KEY_HINT2)")
+	f.StringVarP(&copyOptions.PasswordCommand, "password-command2", "", os.Getenv("RESTIC_PASSWORD_COMMAND2"), "specify a shell command to obtain a password for the destination repository (default: $RESTIC_PASSWORD_COMMAND2)")
+
+	f.StringArrayVarP(&copyOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
+	f.Var(&copyOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot ID is given")
+	f.StringArrayVar(&copyOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot ID is given")
+}
+
+func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
+	if opts.Repo == "" {
+		return errors.Fatal("Please specify a destination repository location (--repo2)")
+	}
+	var err error
+	dstGopts := gopts
+	dstGopts.Repo = opts.Repo
+	dstGopts.PasswordFile = opts.PasswordFile
+	dstGopts.PasswordCommand = opts.PasswordCommand
+	dstGopts.KeyHint = opts.KeyHint
+	dstGopts.password, err = resolvePassword(dstGopts, "RESTIC_PASSWORD2")
+	if err != nil {
+		return err
+	}
+	dstGopts.password, err = ReadPassword(dstGopts, "enter password for destination repository: ")
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(gopts.ctx)
+	defer cancel()
+
+	srcRepo, err := OpenRepository(gopts)
+	if err != nil {
+		return err
+	}
+
+	dstRepo, err := OpenRepository(dstGopts)
+	if err != nil {
+		return err
+	}
+
+	srcLock, err := lockRepo(srcRepo)
+	defer unlockRepo(srcLock)
+	if err != nil {
+		return err
+	}
+
+	dstLock, err := lockRepo(dstRepo)
+	defer unlockRepo(dstLock)
+	if err != nil {
+		return err
+	}
+
+	debug.Log("Loading source index")
+	if err := srcRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	debug.Log("Loading destination index")
+	if err := dstRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
+		Verbosef("snapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+		Verbosef("  copy started, this may take a while...\n")
+
+		if err := copyTree(ctx, srcRepo, dstRepo, *sn.Tree); err != nil {
+			return err
+		}
+		debug.Log("tree copied")
+
+		if err = dstRepo.Flush(ctx); err != nil {
+			return err
+		}
+		debug.Log("flushed packs")
+
+		err = dstRepo.SaveIndex(ctx)
+		if err != nil {
+			debug.Log("error saving index: %v", err)
+			return err
+		}
+		debug.Log("saved index")
+
+		// save snapshot
+		sn.Parent = nil   // Parent does not have relevance in the new repo.
+		sn.Original = nil // Original does not have relevance in the new repo.
+		newID, err := dstRepo.SaveJSONUnpacked(ctx, restic.SnapshotFile, sn)
+		if err != nil {
+			return err
+		}
+		Verbosef("snapshot %s saved\n", newID.Str())
+	}
+	return nil
+}
+
+func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID) error {
+	tree, err := srcRepo.LoadTree(ctx, treeID)
+	if err != nil {
+		return fmt.Errorf("LoadTree(%v) returned error %v", treeID.Str(), err)
+	}
+	// Do we already have this tree blob?
+	if !dstRepo.Index().Has(treeID, restic.TreeBlob) {
+		newTreeID, err := dstRepo.SaveTree(ctx, tree)
+		if err != nil {
+			return fmt.Errorf("SaveTree(%v) returned error %v", treeID.Str(), err)
+		}
+		// Assurance only.
+		if newTreeID != treeID {
+			return fmt.Errorf("SaveTree(%v) returned unexpected id %s", treeID.Str(), newTreeID.Str())
+		}
+	}
+
+	// TODO: keep only one (big) buffer around.
+	// TODO: parellize this stuff, likely only needed inside a tree.
+
+	for _, entry := range tree.Nodes {
+		// If it is a directory, recurse
+		if entry.Type == "dir" && entry.Subtree != nil {
+			if err := copyTree(ctx, srcRepo, dstRepo, *entry.Subtree); err != nil {
+				return err
+			}
+		}
+		// Copy the blobs for this file.
+		for _, blobID := range entry.Content {
+			// Do we already have this data blob?
+			if dstRepo.Index().Has(blobID, restic.DataBlob) {
+				continue
+			}
+			debug.Log("Copying blob %s\n", blobID.Str())
+			size, found := srcRepo.LookupBlobSize(blobID, restic.DataBlob)
+			if !found {
+				return fmt.Errorf("LookupBlobSize(%v) failed", blobID)
+			}
+			buf := restic.NewBlobBuffer(int(size))
+			n, err := srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, buf)
+			if err != nil {
+				return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
+			}
+			if n != len(buf) {
+				return fmt.Errorf("wrong number of bytes read, want %d, got %d", len(buf), n)
+			}
+
+			newBlobID, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID)
+			if err != nil {
+				return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
+			}
+			// Assurance only.
+			if newBlobID != blobID {
+				return fmt.Errorf("SaveBlob(%v) returned unexpected id %s", blobID.Str(), newBlobID.Str())
+			}
+		}
+	}
+
+	return nil
+}
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -109,9 +109,36 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 	}
 
 	visitedTrees := restic.NewIDSet()
+	dstSnapshotByOriginal := make(map[restic.ID][]*restic.Snapshot)
+	for sn := range FindFilteredSnapshots(ctx, dstRepo, opts.Hosts, opts.Tags, opts.Paths, nil) {
+		if sn.Original != nil && !sn.Original.IsNull() {
+			dstSnapshotByOriginal[*sn.Original] = append(dstSnapshotByOriginal[*sn.Original], sn)
+		}
+		// also consider identical snapshot copies
+		dstSnapshotByOriginal[*sn.ID()] = append(dstSnapshotByOriginal[*sn.ID()], sn)
+	}
 
 	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
-		Verbosef("snapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+		Verbosef("\nsnapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+
+		// check whether the destination has a snapshot with the same persistent ID which has similar snapshot fields
+		srcOriginal := *sn.ID()
+		if sn.Original != nil {
+			srcOriginal = *sn.Original
+		}
+		if originalSns, ok := dstSnapshotByOriginal[srcOriginal]; ok {
+			isCopy := false
+			for _, originalSn := range originalSns {
+				if similarSnapshots(originalSn, sn) {
+					Verbosef("skipping source snapshot %s, was already copied to snapshot %s\n", sn.ID().Str(), originalSn.ID().Str())
+					isCopy = true
+					break
+				}
+			}
+			if isCopy {
+				continue
+			}
+		}
 		Verbosef("  copy started, this may take a while...\n")
 
 		if err := copyTree(ctx, srcRepo, dstRepo, *sn.Tree, visitedTrees); err != nil {
@@ -125,8 +152,11 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		debug.Log("flushed packs and saved index")
 
 		// save snapshot
-		sn.Parent = nil   // Parent does not have relevance in the new repo.
-		sn.Original = nil // Original does not have relevance in the new repo.
+		sn.Parent = nil // Parent does not have relevance in the new repo.
+		// Use Original as a persistent snapshot ID
+		if sn.Original == nil {
+			sn.Original = sn.ID()
+		}
 		newID, err := dstRepo.SaveJSONUnpacked(ctx, restic.SnapshotFile, sn)
 		if err != nil {
 			return err
@@ -134,6 +164,25 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		Verbosef("snapshot %s saved\n", newID.Str())
 	}
 	return nil
+}
+
+func similarSnapshots(sna *restic.Snapshot, snb *restic.Snapshot) bool {
+	// everything except Parent and Original must match
+	if !sna.Time.Equal(snb.Time) || !sna.Tree.Equal(*snb.Tree) || sna.Hostname != snb.Hostname ||
+		sna.Username != snb.Username || sna.UID != snb.UID || sna.GID != snb.GID ||
+		len(sna.Paths) != len(snb.Paths) || len(sna.Excludes) != len(snb.Excludes) ||
+		len(sna.Tags) != len(snb.Tags) {
+		return false
+	}
+	if !sna.HasPaths(snb.Paths) || !sna.HasTags(snb.Tags) {
+		return false
+	}
+	for i, a := range sna.Excludes {
+		if a != snb.Excludes[i] {
+			return false
+		}
+	}
+	return true
 }
 
 func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID, visitedTrees restic.IDSet) error {
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -168,26 +168,14 @@ func copyTree(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID re
 				continue
 			}
 			debug.Log("Copying blob %s\n", blobID.Str())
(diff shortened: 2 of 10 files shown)


cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,211 @@
...
+	// TODO: parellize this stuff, likely only needed inside a tree.
+
+	for _, entry := range tree.Nodes {
+		// If it is a directory, recurse
+		if entry.Type == "dir" && entry.Subtree != nil {
+			if err := copySnapshot(ctx, srcRepo, dstRepo, *entry.Subtree); err != nil {
+				return err
+			}
+		}
+		if len(entry.Content) != 0 {
```
greatroar: This check is unnecessary. Removing it unindents the loop body by one tab.

greatroar: This check is unnecessary. Removing it unindents the loop body by one tab.

greatroar: This check is unnecessary. Removing it unindents the loop body by one tab.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,211 @@
...
+			// Copy the blobs for this file.
+			for _, blobid := range entry.Content {
+				// Do we already have this data blob?
+				if !dstRepo.Index().Has(blobid, restic.DataBlob) {
```
greatroar: Or:

```
if dstRepo.Index().Has(blobid, restic.DataBlob) {
	continue
}
```

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,211 @@
...
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Host, opts.Tags, opts.Paths, args) {
+		Verbosef("snapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+
+		// TODO(): don't think this is sufficient, we prob need to iterate over
+		// the existing snapshots and do a full equivalance check.
+		if dstRepo.Index().Has(*sn.Tree, restic.TreeBlob) {
+			Printf("Snapshot %s already exists, skipped\n", sn.ID().Str())
+			continue
+		}
+		debug.Log("non-existence assurred")
```
greatroar: Typo: assured

cmd/restic/cmd_copy.go
middelink: Done.

Not sure what noise you are referring to. All is well.

cmd/restic/cmd_copy.go
middelink: Done.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,210 @@
...
+		debug.Log("saved index")
+
+		// save snapshot
+		sn.Parent = nil   // does Parent have relevance in the new repo?
+		sn.Original = nil // does Original have relevance in the new repo?
+		newid, err := dstRepo.SaveJSONUnpacked(ctx, restic.SnapshotFile, sn)
+		if err != nil {
+			return err
+		}
+		debug.Log("saved snapshot %v", newid.Str())
```
greatroar: I would make this a Verbosef instead of a debug.Log.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,210 @@
...
+	Verbosef("Loading destination index\n")
+	if err := dstRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Host, opts.Tags, opts.Paths, args) {
+		Verbosef("snapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+
+		// TODO(): don't think this is sufficient, we prob need to iterate over
+		// the existing snapshots and do a full equivalance check.
```
greatroar: Indeed. I just copied two snapshots that have the same tree in repo1 and the second was skipped, because its tree was just added. That's a loss of information. Since there is already duplicate checking in copySnapshot, can't we just remove this check? Or would that produce duplicate snapshots in the destination repo if restic copy is run twice with the same arguments?

middelink: Hmm, how did you create two snapshots with the same tree? The code was meant as an optimization to not have to flush indexes and all that stuff. I agree that if there is some process to easily create same snapshots trees, it should be removed.

greatroar: I just ran backups on the same directory twice, with atime recording off, and nothing had changed.

FallenWarrior2k:
I'm sorry, but may I ask what this is blocked on right now?

middelink:
@FallenWarrior2k no idea, very good question.

@fdo care to intervene? This PR has been sitting idle for almost 2 months now.

rawtaz:
@FallenWarrior2k @middelink I can only apologize. I am the one who asked @middelink to make a PR out of her work, which she so swiftly did. My intention was to have something that we could work off in terms of adjustments and reviews. We started having more of a flow in doing restic maintenance at that point in time, but then Corona etc came into the picture and since then it's been almost impossible to get anything done due to maintainers' life situations with isolation and such things requiring all available time. This PR is still on our radar though, FWIW.

FallenWarrior2k:
That's completely fair, I was just wondering. No blame on anybody during these trying times.

-------- Original Message --------

MichaelEischer (commented):
I've found a few more typos (see review comments) and a few more fundamental points which could also be deferred to a follow-up PR: The `Repository` object currently does not upload intermediate indexes on its own. That could cause the copy command to create a single gigantic index file when completing the snapshot. It would probably be a good idea to use something like the `IndexUploader` from the `backup` command.

The next point is more of a usability question: The initial idea in #323 was to support a one-time copy from one repository to another. But it would probably also be nice to ensure that when copying a snapshot multiple times, that that only creates one snapshot in the destination repository.

The current implementation won't scale for large numbers of snapshots. These usually share large parts of their filesystem trees, thus iterating the full tree for every snapshot will be rather redundant. A rather simple solution for that would be to use `FindUsedBlobs` to collect all blobs of all snapshots and then copy everything in one go. This has the slight downside that when copying a single snapshot it will load trees twice, but should be much faster when copying multiple snapshots. And it would allow adding a progress bar. As well as avoiding yet another tree walking implementation.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,202 @@
...
+var copyOptions CopyOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdCopy)
+
+	f := cmdCopy.Flags()
+	f.StringVarP(&copyOptions.Repo, "repo2", "", os.Getenv("RESTIC_REPOSITORY2"), "repository to backup to or restore from (default: $RESTIC_REPOSITORY2)")
+	f.StringVarP(&copyOptions.PasswordFile, "password-file2", "", os.Getenv("RESTIC_PASSWORD_FILE2"), "read the destination repository password from a file (default: $RESTIC_PASSWORD_FILE2)")
+	f.StringVarP(&copyOptions.KeyHint, "key-hint2", "", os.Getenv("RESTIC_KEY_HINT2"), "key ID of key to try decrypting the destination repository first (default: $RESTIC_KEY_HINT2)")
+	f.StringVarP(&copyOptions.PasswordCommand, "password-command2", "", os.Getenv("RESTIC_PASSWORD_COMMAND2"), "specify a shell command to obtain a password for the destination repository (default: $RESTIC_PASSWORD_COMMAND2)")
```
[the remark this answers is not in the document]
middelink: Done

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,202 @@
...
+
+	f.StringArrayVarP(&tagOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
```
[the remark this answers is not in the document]
middelink: Ah, good catch.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func copySnapshot(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID) error {
+	tree, err := srcRepo.LoadTree(ctx, treeID)
+	if err != nil {
+		return fmt.Errorf("LoadTree(%v) returned error %v", treeID.Str(), err)
+	}
+	// Do we already have this tree blob?
+	if !dstRepo.Index().Has(treeID, restic.TreeBlob) {
+		newtreeid, err := dstRepo.SaveTree(ctx, tree)
```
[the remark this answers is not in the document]
middelink: Done, and a bunch of names more.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,202 @@
...
+		err = dstRepo.SaveIndex(ctx)
+		if err != nil {
+			debug.Log("error saving index: %v", err)
+			return err
+		}
+		debug.Log("saved index")
+
+		// save snapshot
+		sn.Parent = nil   // does Parent have relevance in the new repo?
+		sn.Original = nil // does Original have relevance in the new repo?
```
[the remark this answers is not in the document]
middelink: Ah, that was a rhetorical question :) Rephrased

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,203 @@
...
+		// If it is a directory, recurse
+		if entry.Type == "dir" && entry.Subtree != nil {
+			if err := copySnapshot(ctx, srcRepo, dstRepo, *entry.Subtree); err != nil {
+				return err
+			}
+		}
+		// Copy the blobs for this file.
+		for _, blobID := range entry.Content {
+			// Do we already have this data blob?
+			if dstRepo.Index().Has(blobID, restic.DataBlob) {
```
MichaelEischer, line 174: This currently doesn't work as expected: Newly uploaded blobs don't show up in the repository index until their pack is uploaded. This can lead to duplicate blobs created by the copy command. The archiver code currently uses a sort of workaround to manually track these new blobs. It would be possible (and rather easy) to add similar code here. But the best solution would be move this functionality into the MasterIndex similar to the `CheckSetKnown` method from the index rewrite issue https://github.com/restic/restic/issues/2523#issuecomment-585940174 https://github.com/aawsome/restic/commit/53f8a9980cb2e52beb453677304b565e075427f5#diff-0a50833756dabb93f6aaa532c9f63ec6

MichaelEischer:
From what I can tell right now the steps to finish this PR are:

The command needs documentation and a changelog:
Probably a short explanation and usage example. And we need some documentation that backend parameters specified via environment variable or config options apply to both source and destination paths. That is it might not be possible to specify e.g. different accounts for the source and destination.

Tests:
- Check that a copied snapshot has same tree content as the old one (= identical tree hash)
- Check that the tree in the new snapshot is intact
- Check that the copy functionality doesn't add duplicate blobs to the repository

Code fixes:
- As #2640 was merged the code just needs a small tweak to reuse the buffer
- The duplicate blob upload should be fixed as this could cause lots of duplicates in pathological cases
- An intermediate index uploader could also be added after merging this.

FallenWarrior2k:
>That is it might not be possible to specify e.g. different accounts for the source and destination.

For most remotes, I _think_ it should be possible to accomplish this by going through rclone. Although it might not be possible to use e.g. `env_auth = true` (for S3) or similar in that case, meaning you'd have to temporarily persist your auth data in the rclone config. In that case, you could password-protect the file before adding them, if you want to make sure they're not visible on your filesystem.

dionorgua:
Small note: It's no longer possible to build it. I think that due to #2640 merged.

I've fixed it locally with this and tested with one-snapshot copy:
```
iff --git a/cmd/restic/cmd_copy.go b/cmd/restic/cmd_copy.go
index 4abfcdd7..337129a7 100644
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -175,18 +175,10 @@ func copySnapshot(ctx context.Context, srcRepo, dstRepo restic.Repository, treeI
 				continue
 			}
 			debug.Log("Copying blob %s\n", blobID.Str())
-			size, found := srcRepo.LookupBlobSize(blobID, restic.DataBlob)
-			if !found {
-				return fmt.Errorf("LookupBlobSize(%v) failed", blobID)
-			}
-			buf := restic.NewBlobBuffer(int(size))
-			n, err := srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, buf)
+			buf, err := srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, nil)
 			if err != nil {
 				return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
 			}
-			if n != len(buf) {
-				return fmt.Errorf("wrong number of bytes read, want %d, got %d", len(buf), n)
-			}
 
 			newBlobID, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID)
 			if err != nil {
`
```

seqizz:
This is a must-have feature for me, thanks @dionorgua :+1: Also tested & working.

greatroar:
/middelink/restic/pull/1

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,197 @@
...
+	f.StringVarP(&copyOptions.Repo, "repo2", "", os.Getenv("RESTIC_REPOSITORY2"), "repository to backup to or restore from (default: $RESTIC_REPOSITORY2)")
+	f.StringVarP(&copyOptions.PasswordFile, "password-file2", "", os.Getenv("RESTIC_PASSWORD_FILE2"), "read the destination repository password from a file (default: $RESTIC_PASSWORD_FILE2)")
+	f.StringVarP(&copyOptions.KeyHint, "key-hint2", "", os.Getenv("RESTIC_KEY_HINT2"), "key ID of key to try decrypting the destination repository first (default: $RESTIC_KEY_HINT2)")
+	f.StringVarP(&copyOptions.PasswordCommand, "password-command2", "", os.Getenv("RESTIC_PASSWORD_COMMAND2"), "specify a shell command to obtain a password for the destination repository (default: $RESTIC_PASSWORD_COMMAND2)")
+
+	f.StringArrayVarP(&copyOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
+	f.Var(&copyOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot-ID is given")
+	f.StringArrayVar(&copyOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")
```
aawsome, line 54: Just a proposal: Why not add an option to explicitely specify one or more snapshots by ID? (see `cmd_forget.go`)

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := srcRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	debug.Log("Loading destination index")
+	if err := dstRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
```
aawsome, line 111: I would define `snapshots` exactly like in `cmd_forget.go` and then loop over its entries...

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,197 @@
...
+	return nil
+}
+
+func copySnapshot(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID) error {
+	tree, err := srcRepo.LoadTree(ctx, treeID)
+	if err != nil {
+		return fmt.Errorf("LoadTree(%v) returned error %v", treeID.Str(), err)
+	}
+	// Do we already have this tree blob?
+	if !dstRepo.Index().Has(treeID, restic.TreeBlob) {
```
aawsome, line 150: IMO, you can optimize here: if `dstRepo` does have this `TreeBlob`, it should also have all entries (and sub-trees). 
At least if this repository is in a sane state :wink:  So no need to run into the recursion!

Maybe we could also add a flag `--deep-copy` to always recurse into the trees. This would optionally allow to "heal" missing blobs if they are present in another repo/snapshot which is selected for copy - at the cost of copy speed.

However, here we do have the same problem with `Index.Has` as @MichaelEischer stated below.

cmd/restic/cmd_copy.go
aawsome: As I don't see that #2523 will be tackled soon (and so far there are quite a lot open things to work on there), maybe it is best to make a small PR only implementing the 'CheckSetKnown' part of #2523. 

I can propose to prepare such an PR which would then be a prerequisite for this PR.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+		debug.Log("tree copied")
+
+		if err = dstRepo.Flush(ctx); err != nil {
+			return err
+		}
+		debug.Log("flushed packs")
+
+		err = dstRepo.SaveIndex(ctx)
```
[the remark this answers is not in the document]
aawsome: For saving an intermediate index you could follow the lines in `internal/repository/master_index.go RebuildIndex`.
Maybe start with the check after each blob write but this then could be optimized (in a future PR).

IMO it would then also help to merge #2749.

aawsome (changes requested):
I found the two major issues that @MichaelEischer also found.
I think the `Index.Has` problem should be fixed in another PR. I'll work on this.
The saving of intermediate indexes should IMO be added in this PR.

Moreover I have some enhancement suggestions and IMO found a possibility to speed `copy` up.

EDIT: I started working on the propsed PR and realized that I might easily fix both issues.

aawsome:
> EDIT: I started working on the propsed PR and realized that I might easily fix both issues.

The new PR is ready for review: #2773

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,197 @@
...
+			if err != nil {
+				return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
+			}
+
+			newBlobID, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID)
+			if err != nil {
+				return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
+			}
+			// Assurance only.
+			if newBlobID != blobID {
```
MichaelEischer, line 190: `SaveBlob` does not recalculate the blob hash, thus `blobID` and `newBlobID` will always be equal.

cmd/restic/cmd_copy.go
MichaelEischer: To be precise for the current implementation in this PR we wouldn't need an atomic test&set method for the index. It would suffice if `SaveBlob` remembered which blocks it has already saved. If we want to parallelize the implementation then we need something like the `CheckSetKnown` method.

However, there's another option: Let `SaveBlob` itself check for duplicates. This would conveniently hide the details on how to properly check for blobs that are in incomplete packs and by construction ensure that we can't accidentally call `CheckSetKnown` and then fail to actually save the chunk.

MichaelEischer:
Just noticed that I should have taken a look at #2773 first, which solves my index related concerns.

cmd/restic/cmd_copy.go
aawsome: Oh, my bad! I wasn't able to figure out that `copy` already exactly behaves like `forget` w.r.t. to the given parameters. No need to change anything here, of course!

aawsome:
I just realized that there might be another optimization:
`copy` uses `LoadBlob` for all needed blobs but this might imply that a pack is downloaded more than once. So sorting for needed packs would improve performance. I think something similar is implemented in the `restorer`.

However, I would propose to not solve this in this PR and leave it open for further optimization.

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
aawsome: Now that #2773 is merged, the `SaveIndex` can simply be removed as `Flush` also handles the index.

cmd/restic/cmd_copy.go
aawsome: This should be now resolved with #2773.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,197 @@
...
+			if dstRepo.Index().Has(blobID, restic.DataBlob) {
+				continue
+			}
+			debug.Log("Copying blob %s\n", blobID.Str())
+			buf, err = srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, buf)
+			if err != nil {
+				return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
+			}
+
+			newBlobID, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID)
```
aawsome, line 185: With #2773 the method `SaveBlob` changed; you should change this here accordingly.

aawsome (commented):
Now that #2773 was merged, there are only few open issues with this PR.

seqizz:
I think @middelink has not much time to check this PR :)

I wanted to rebase this over @aawsome 's new prune branch like mentioned above and got error on `SaveBlob` so I made this change and made the build succeed. I am still worse than /dev/urandom on Go so please correct if I understand the new param wrong:

```
diff --git a/cmd/restic/cmd_copy.go b/cmd/restic/cmd_copy.go
index 46e61ccd..5faca0d0 100644
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -182,7 +182,10 @@ func copySnapshot(ctx context.Context, srcRepo, dstRepo restic.Repository, treeI
                                return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
                        }

-                       newBlobID, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID)
+                       newBlobID, known, err := dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID, false)
+                       if known {
+                               debug.Log("Skipping since it's already in repo: %s\n", blobID.Str())
+                       }
                        if err != nil {
                                return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
                        }
```

greatroar:
middelink/restic#2

middelink:
Hi, sorry for the delay indeed. I seem to not get mail notifications anymore in github on this matter (guess I pressed a wrong button somewhere), just the emails themselves which go off into some folder I hardly look at. Anyways, excuses.

I've merged the PR, and (at least on my computer) it compiles again. Thanks!

aawsome (commented):
I just had a few remarks/suggestions for this PR

IMO there are some possibilities for future improvement which, however, should not be included in this PR:

- a common use case would be to use identical chunking parameters and keys for both repositories. This fixes the chunking issue. An optimization could check if the keys are identical and then save the re-encryption of blobs. The same optimization could be added to `RewritePacks` in `prune`.
- parallelize processing of trees, e.g. by using the code of `internal/archiver`
- check for tree blobs already in `dstRepo` either by defining a "sane" repository or by adding a flag to use this possibility on users responsibility :wink:

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+		Verbosef("  copy started, this may take a while...\n")
+
+		if err := copySnapshot(ctx, srcRepo, dstRepo, *sn.Tree, visitedTrees); err != nil {
+			return err
+		}
+		debug.Log("tree copied")
+
+		if err = dstRepo.Flush(ctx); err != nil {
```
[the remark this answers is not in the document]
MichaelEischer: My plan was to change this when parallelizing the current implementation. Right now, it probably doesn't matter much whether we flush after each snapshots or not.

changelog/unreleased/issue-323
```diff
@@ -0,0 +1,11 @@
+Enhancement: Add a command to copy a snapshot to another repository
+
+We've added a copy command, allowing you to copy snapshots from one
+repository to another.
+
+Note that this process will have to read(download) and write(upload) the
+entire snapshot(s) due to the different encryption keys used on the source
+and destination repository.
```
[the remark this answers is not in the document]
rawtaz: I think that's worthy a note, yeah. There's also a few other gotchas or know-abouts that need to be communicated. I'll add this one you mentioned to the list.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,214 @@
...
+		newID, err := dstRepo.SaveJSONUnpacked(ctx, restic.SnapshotFile, sn)
+		if err != nil {
+			return err
+		}
+		Verbosef("snapshot %s saved\n", newID.Str())
+	}
+	return nil
+}
+
+func copySnapshot(ctx context.Context, srcRepo, dstRepo restic.Repository, treeID restic.ID, visitedTrees restic.IDSet) error {
```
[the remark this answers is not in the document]
MichaelEischer: Is changed.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}
+		// Copy the blobs for this file.
+		for _, blobID := range entry.Content {
+			// Do we already have this data blob?
+			if dstRepo.Index().Has(blobID, restic.DataBlob) {
+				continue
+			}
+			debug.Log("Copying blob %s\n", blobID.Str())
+			buf, err := srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, nil)
```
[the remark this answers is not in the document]
MichaelEischer: The complexity here is that we have to (well, should) reuse a single `buf` across all `copyTree` recursions. Otherwise each recursion level could hold on to a buffer with a few megabytes each. One rather ugly option would be to pass a `buf *[]byte` parameter along the recursion. Do you have a better idea?

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
+		Verbosef("\nsnapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+
+		if newID, ok := copiedSnapshots[*sn.ID()]; ok {
+			Verbosef("skipping source snapshot %s, was already copied to snapshot %s\n", sn.ID().Str(), newID.Str())
+			continue
+		}
+		if sn.Original != nil && dstRepo.Has(restic.Handle{Type: restic.SnapshotFile, Name: sn.Original.String()}) {
```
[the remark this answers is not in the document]
MichaelEischer: My idea here was that snapshots IDs are random and therefore it is basically impossible for the same snapshot ID to appear in both repositories. Skipping original snapshots would only help if someone copied the raw files of a repository and then later on uses copy between both repositories. So it's probably rarely useful, but I guess it doesn't hurt. -> Is implemented.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,214 @@
...
+	for sn := range FindFilteredSnapshots(ctx, dstRepo, opts.Hosts, opts.Tags, opts.Paths, nil) {
+		if sn.Original != nil && !sn.Original.IsNull() {
+			copiedSnapshots[*sn.Original] = sn.ID()
+		}
+	}
+
+	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
+		Verbosef("\nsnapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
+
+		if newID, ok := copiedSnapshots[*sn.ID()]; ok {
```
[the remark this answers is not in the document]
MichaelEischer: I've used an `IDSet` here at first, however, that doesn't provide enough information for useful message to the user. The value from `copiedSnapshots` is stored in `newID`. I've change the variable name to `dstSnapshotByOriginal` and the set now stores snapshots.

With the full snapshot the check now ensures that the `Tree` of both the potential dst and src snapshots match, such that this check doesn't care when #2731 doesn't change `Original` for a snapshot.

doc/manual_rest.rst
```diff
@@ -79,10 +80,9 @@ command:
     EXIT STATUS
     ===========
 
-    Exit status is 0 if the command was successful, and non-zero if there was any error.
-
-    Note that some issues such as unreadable or deleted files during backup
-    currently doesn't result in a non-zero error exit status.
+    Exit status is 0 if the command was successful.
+    Exit status is 1 if there was a fatal error (no snapshot created).
+    Exit status is 3 if some source data could not be read (incomplete snapshot created).
```
rawtaz, line 85: Perhaps this should be in a separate commit as it's unrelated to the copy command?

changelog/unreleased/issue-323
```diff
@@ -0,0 +1,11 @@
+Enhancement: Add a command to copy a snapshot to another repository
+
+We've added a copy command, allowing you to copy snapshots from one
+repository to another.
+
+Note that this process will have to read(download) and write(upload) the
```
rawtaz, line 6: Please add a space before each of `(`.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var cmdCopy = &cobra.Command{
+	Use:   "copy [flags] [snapshotID ...]",
+	Short: "Copy snapshots from one repository to another",
+	Long: `
+The "copy" command copies one or more snapshots from one repository to another
+repository. Note that this will have to read(download) and write(upload) the
```
rawtaz, line 20: Please add a space before each `(`.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,218 @@
...
+	Paths           []string
+}
+
+var copyOptions CopyOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdCopy)
+
+	f := cmdCopy.Flags()
+	f.StringVarP(&copyOptions.Repo, "repo2", "", os.Getenv("RESTIC_REPOSITORY2"), "repository to backup to or restore from (default: $RESTIC_REPOSITORY2)")
```
rawtaz, line 47: Please change `repository to backup to or restore from` to `destination repository to copy snapshots to`.

cmd/restic/cmd_copy.go
MichaelEischer: I don't think that this definition of "sane" is a good idea. Using it, just interrupting a backup would turn a repository "insane" which I regard as highly problematic (especially for mobile clients). It would also be rather complex to implement prune in a way that can for every repository guarantee, that it never enters an "insane" state. The trees in rewritten packs would have to be laid out such that dependencies between tree packs are acyclic (a tree pack `p` depends on all packs which contain tree blobs referenced by tree blobs in `p`). If we want to avoid rewriting every tree pack, then the complexity just skyrockets. It's actually a design goal of prune to never leave the repository in a broken state (requiring repair steps beyond removing the stale lock) no matter when it is interrupted.

The definition of "sane" used by the `check` command is rather simple: A repository is "sane", if every snapshot can restored.

As restic tries to prevent users from accidentally damaging their backup, requiring a "sane" and pruned repository would require the copy command to run a _full_ check of the destination repository which sabotages the goal of speeding up the copy command.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,218 @@
...
+	f.StringVarP(&copyOptions.PasswordFile, "password-file2", "", os.Getenv("RESTIC_PASSWORD_FILE2"), "read the destination repository password from a file (default: $RESTIC_PASSWORD_FILE2)")
+	f.StringVarP(&copyOptions.KeyHint, "key-hint2", "", os.Getenv("RESTIC_KEY_HINT2"), "key ID of key to try decrypting the destination repository first (default: $RESTIC_KEY_HINT2)")
+	f.StringVarP(&copyOptions.PasswordCommand, "password-command2", "", os.Getenv("RESTIC_PASSWORD_COMMAND2"), "specify a shell command to obtain a password for the destination repository (default: $RESTIC_PASSWORD_COMMAND2)")
+
+	f.StringArrayVarP(&copyOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
+	f.Var(&copyOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot-ID is given")
```
rawtaz, line 53: Please change `snapshot-ID` to `snapshot ID`.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,218 @@
...
+	f.StringArrayVar(&copyOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")
```
rawtaz, line 54: Nitpicking again, please change `snapshot-ID` to `snapshot ID` :D

doc/045_working_with_repos.rst
```diff
@@ -82,6 +82,64 @@ Furthermore you can group the output by the same filters (host, paths, tags):
...
+
+    snapshot 4e5d5487 of [/home/user/work] at 2020-05-01 22:44:07.012113 +0200 CEST)
+    skipping snapshot 4e5d5487, was already copied to snapshot 50eb62b7
+
+The example command copies all snapshots from the source repository
+``/srv/restic-repo`` to the destination repository ``/srv/restic-repo-copy``.
+Snapshots which have previously been copied between repositories will
+be skipped by later copy runs.
+
+Note that this process will have to read(download) and write(upload) the
```
rawtaz, line 109: Please add a space before each `(`.

doc/045_working_with_repos.rst
```diff
@@ -82,6 +82,64 @@ Furthermore you can group the output by the same filters (host, paths, tags):
...
+
+The list of snapshots to copy can be filtered by host, path in the backup
+and / or a comma-separated tag list:
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo copy --repo2 /srv/restic-repo-copy --host luigi --path /srv --tag foo,bar
+
+It is also possible to explicitly specify a list of snapshots to copy between
+repositories:
```
rawtaz, line 136: I think we should change this line to `It is also possible to explicitly specify the list of snapshots to copy, in which case only these instead of all snapshots will be copied:`.

Also, a question: Is it possible to use `latest` when specifying the snapshot? If so perhaps this should be included in the example command.

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
aawsome: My intention actually was to remove this `dstRepo.Has` (which is always slow as it needs to question the backend), but I realized that this doen't work that way...

You are right that there should not be the same IDs within both repos. However, then this `if` already does not check anything: if the original snapshot exists in `dstRepo` it has not been saved with ID `sn.Original` but with a random new ID. However, if the snapshot within `dstRepo` would have set its own `sn.Original` to the original ID of this snapshot, this situation can be detected.

So instead if 
```golang
// Original does not have relevance in the new repo. Remember the source snapshot instead
		sn.Original = sn.ID()
```
below, we should use
```golang
		if sn.Originial != nil {
			sn.Original = sn.ID()
		}
```

and 
```golang
// check whether the destination has a snapshot copied from sn, whose tree is still identical
		if dstSn, ok := dstSnapshotByOriginal[*sn.ID()]; ok && dstSn.Tree.Equal(*sn.Tree) {
```
above should be changed to 
```golang
srcOriginal := *sn.ID()
if sn.Original != nil {
	srcOriginal = sn.Original
}
if dstSn, ok := dstSnapshotByOriginal[srcOriginal]; ok && dstSn.Tree.Equal(*sn.Tree) {
```

cmd/restic/cmd_copy.go
aawsome: Actually after the `vistitedTrees` optimization there is not that much to gain here by skipping the reading of trees already present in the destination repo. Yes, it will walk the tree once but detect that there is nothing to save and hence it will be as faster than a `backup` run with no modified files. Coping multiple snapshots with similar tree structure will get sped up even more by using `visitedTrees`. That should be fine :+1: 

You are right that it wouldn't help to define a "sane" state which can be only obtained safely by always pruning. Nevertheless I think a good definition of a "sane" state (and the pitfalls like being able to have "dead" trees as long as they themselves are not referenced by snapshots), maybe also the definition of an "optimal" state (in terms of e.g. not wasting space), would be good to have. This could also allow `check` to produce a summary like "your repository is sane but not optimal". But I'll stop now spoiling this PR with this unrelated topic :wink:

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
aawsome: For this implementation (with parallelizing it already on the aganda) I would just use a global `var buf []bytes` - maybe with a comment `// TODO: change this when parallelizing`.

doc/045_working_with_repos.rst
MichaelEischer: `latest` does work. I've just added it the example, without further additional explanations.

doc/045_working_with_repos.rst
MichaelEischer: Is added.

cmd/restic/cmd_copy.go
MichaelEischer: is changed.

cmd/restic/cmd_copy.go
MichaelEischer: is changed.

cmd/restic/cmd_copy.go
MichaelEischer: is changed.

cmd/restic/cmd_copy.go
MichaelEischer: is added.

changelog/unreleased/issue-323
MichaelEischer: is added.

doc/manual_rest.rst
MichaelEischer: I've split the change into a separate commit.

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
MichaelEischer: Ugh, I really dislike global variables. I've decided to move the repository objects and visitedTrees into a struct.

changelog/unreleased/issue-323
```diff
@@ -0,0 +1,11 @@
+Enhancement: Add a command to copy a snapshot to another repository
```
rawtaz, line 1: Suggest changing to `Enhancement: Add command to copy snapshots between repositories`.

doc/045_working_with_repos.rst
```diff
@@ -82,6 +82,64 @@ Furthermore you can group the output by the same filters (host, paths, tags):
...
+    skipping snapshot 4e5d5487, was already copied to snapshot 50eb62b7
+
+The example command copies all snapshots from the source repository
+``/srv/restic-repo`` to the destination repository ``/srv/restic-repo-copy``.
+Snapshots which have previously been copied between repositories will
+be skipped by later copy runs.
+
+Note that this process will have to read (download) and write (upload) the
+entire snapshot(s) due to the different encryption keys used on the source
+and destination repository.
```
rawtaz, line 111: I suggest changing this to the following (just to make it stand out in a note box):

~~~
.. note:: Note that this process will have to read (download) and write (upload) the
    entire snapshot(s) due to the different encryption keys used in the source and
    destination repository. Also, the transferred files are not re-chunked, which
    may break deduplication between files already stored in the destination repo
    and files copied there using this command.
~~~

changelog/unreleased/issue-323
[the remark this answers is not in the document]
rawtaz: Suggested in the documentation file in this PR.

changelog/unreleased/issue-323
```diff
@@ -0,0 +1,11 @@
...
+
+We've added a copy command, allowing you to copy snapshots from one
+repository to another.
+
+Note that this process will have to read (download) and write (upload) the
+entire snapshot(s) due to the different encryption keys used on the source
+and destination repository.
```
rawtaz, line 8: Suggest (as per @aawsome) adding "Also, the transferred files are not re-chunked, which may break deduplication between files already stored in the destination repo and files copied there using this command."

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+var cmdCopy = &cobra.Command{
+	Use:   "copy [flags] [snapshotID ...]",
+	Short: "Copy snapshots from one repository to another",
+	Long: `
+The "copy" command copies one or more snapshots from one repository to another
+repository. Note that this will have to read (download) and write (upload) the
+entire snapshot(s) due to the different encryption keys on the source and
+destination.
```
rawtaz, line 22: Please add `, and that transferred files are not re-chunked, which may break their deduplication` before the last dot.

changelog/unreleased/issue-323
MichaelEischer: Is changed.

doc/045_working_with_repos.rst
MichaelEischer: Is changed.

changelog/unreleased/issue-323
MichaelEischer: Is added.

cmd/restic/cmd_copy.go
MichaelEischer: Is added.

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
MichaelEischer: `dstRepo.Has` for `sn.Original` handled the case when you copy snapshots from B->A and have previously copied snapshots from A->B. The main difference of always using the `Original` field is that it provides a more consistent behavior in regard to tag modifications: My previous implementation was rather inconsequential in that regard: Assume we copy a snapshot from A to B. Changing tags on the copy in B will keep `Original` and therefore the link to A. On the other hand, changing the tags on the copy in A changes the snapshot ID referenced by `Original` and therefore breaks the link.

I've added your proposed changes in a slightly expanded version: With using the `Original` field as a persistent snapshot ID, it's possible (but probably rare) for multiple snapshots to have the same ID in their `Original` field. I've therefore changed to map to store a snapshot slice.

Regarding changes to the tags of a snapshot, my feeling is that we want to copy a snapshot again once it's tags have changed. In order to avoid playing whack-a-mole with possible future PRs, there's now a `similarSnapshots` function which checks that everything except the Parent and Original fields match.

MichaelEischer:
> a common use case would be to use identical chunking parameters and keys for both repositories. This fixes the chunking issue. An optimization could check if the keys are identical and then save the re-encryption of blobs. The same optimization could be added to RewritePacks in prune.

The re-encryption part shouldn't be too expensive on somewhat modern processors. After all a single CPU core with AES-NI is able to encrypt/decrypt over 1GB/s of data. That is I don't think it's worth to bypass the current Load/StoreBlob abstractions. Re-encrypting blobs also has the benefit, that an attacker can't learn blob boundaries within a pack file after rewriting. (Rewrite would basically shuffle lots of encrypted strings. By checking for common substrings before and after rewriting you'll learn the boundaries between most blobs) However, I can't tell in what regard that information would be useful.

For matching chunking parameters we'd probably need some way to initialize a repository with the same chunking parameters as another one. Modifying the config file of an existing repository doesn't sound right to me. (Take these remarks with a grain of salt, I haven't thought much about whether that a good idea or not)

> parallelize processing of trees, e.g. by using the code of internal/archiver

My plan was to go for an all-out parallelization in a few follow-up PRs. The first part would be to extract the parallel tree loading used by the check command into a `StreamTrees` function, then to parallelize loading data blobs (that would be the only copy command specific functionality) and finally to add the capability for parallel, asynchronous blob/pack uploads to somewhere within the repository (this functionality is also necessary to properly solve #2696).

cmd/restic/cmd_copy.go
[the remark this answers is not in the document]
MichaelEischer: @aawsome Can you take another look at the snapshot deduplication and check whether it now works as you'd expect?

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,266 @@
...
+	if err := srcRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	debug.Log("Loading destination index")
+	if err := dstRepo.LoadIndex(ctx); err != nil {
+		return err
+	}
+
+	dstSnapshotByOriginal := make(map[restic.ID][]*restic.Snapshot)
```
aawsome, line 116: In fact the usual case would be to not have more than one snapshot being "original" within one repository, but it doesn't hurt considering edge cases like deletion errors etc..

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	for i, a := range sna.Paths {
+		if a != snb.Paths[i] {
+			return false
+		}
+	}
+	for i, a := range sna.Excludes {
+		if a != snb.Excludes[i] {
+			return false
+		}
```
aawsome, line 201: IIRC Tags and Paths (and maybe Excludes) are not sorted, so this might miss some "similar" snapshots. However, as we are only comparing snapshots where one should be anyway the original snapshot of another this is ok, IMO.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,266 @@
...
+	if !sna.Time.Equal(snb.Time) || !sna.Tree.Equal(*snb.Tree) || sna.Hostname != snb.Hostname ||
+		sna.Username != snb.Username || sna.UID != snb.UID || sna.GID != snb.GID ||
+		len(sna.Paths) != len(snb.Paths) || len(sna.Excludes) != len(snb.Excludes) ||
+		len(sna.Tags) != len(snb.Tags) {
+		return false
+	}
+	for i, a := range sna.Tags {
+		if a != snb.Tags[i] {
+			return false
+		}
```
aawsome, line 191: I was thinking about the case Snapshot A is copied to dst (lets say to snapshot A'), then a Tag is added to this snapshot on src giving Snapshot B (and removing Snapshot A). Now another copy to dst including snapshot B would not identify B and A' as similar and again copy B to B' (of course realizing that there is nothing to save except the snapshot itself). This will give two Snapshots A' and B' in dst which have the same original Snapshot A...

So I think this strict check should be fine and as the command is called `Copy` and not `Sync` or similar this behavior is maybe what is expected by users.

cmd/restic/cmd_copy.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return true
+}
+
+type treeCloner struct {
+	srcRepo      restic.Repository
+	dstRepo      restic.Repository
+	visitedTrees restic.IDSet
+	buf          []byte
+}
```
aawsome, line 211: I like your solution defining this struct :+1:

aawsome (approved):
LGTM!

I think the most important optimization would be to not reload multiple blobs from identical packs (like done in restorer) and parallelize the saving of blobs (like done in archiver), but this should be done in a future PR.

aawsome:
> The re-encryption part shouldn't be too expensive on somewhat modern processors. After all a single CPU core with AES-NI is able to encrypt/decrypt over 1GB/s of data. That is I don't think it's worth to bypass the current Load/StoreBlob abstractions. Re-encrypting blobs also has the benefit, that an attacker can't learn blob boundaries within a pack file after rewriting. (Rewrite would basically shuffle lots of encrypted strings. By checking for common substrings before and after rewriting you'll learn the boundaries between most blobs) 

You are right this is a good point! In fact with this argument I would also opt for always re-encrypting blobs (even when using the same key, but always with a new random nonce)

cmd/restic/cmd_copy.go
MichaelEischer: My intention here was to rather copy a bit too much than too little. After all, if a user really doesn't want these additional copies, then it's not too difficult to workaround this behavior by just removing the unwanted snapshots.

cmd/restic/cmd_copy.go
MichaelEischer: I've changed the check for the Tags and Paths to ignore the ordering, as is done by other commands which filter snapshots.

MichaelEischer (approved):
LGTM. Thanks to everyone helping with implementing this PR and for the extensive discussions!

PR #2607, opened 2020-02-27
Honor RESTIC_CACHE_DIR on Mac and Windows

greatroar:

According to the [manual](https://restic.readthedocs.io/en/latest/manual_rest.html), RESTIC_CACHE_DIR can be used to set the cache directory, but this variable was not checked on Mac and Windows. It is now. In addition, the code has been simplified by using the Go 1.11 os.UserCacheDir function.

Not that I know.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2607
+++ b/changelog/unreleased/pull-2607
@@ -1,0 +1,7 @@
+Bugfix: Honor RESTIC_CACHE_DIR environment variable on Mac and Windows
+
+On Mac and Windows, the RESTIC_CACHE_DIR environment variable was ignored.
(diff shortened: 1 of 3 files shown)


rawtaz:
@greatroar Really nice PR, thanks for cleaning up (and fixing). Just one thing before we merge; I noticed you have so called "naked returns" in the code, which of course is nice. But in this case we'd prefer to keep the code coherent with the majority of the restic code, so could you please "denakify" the returns so it's like usual (e.g. `return cachedir, nil`)?

greatroar:
Done.

rawtaz (approved):
Great, thanks!

PR #2608, opened 2020-02-29
Simplify termstatus

greatroar:

The package internal/ui/termstatus was simplified by using golang.org/x/{crypto,sys}. Both packages were already dependencies of restic. As a result, another dependency is no longer needed and a considerable part of the code can be removed.

Not that I know of.

Checklist
---------


fd0 (approved):
Awesome, thanks!

Issue #2547, opened 2020-01-12
Discussion: Future of prune and rebuild-index


aawsome:
I would like to discuss how to improve the actual situation of `prune` (and `rebuild-index` which is used within `prune`) as there are quite some issues about it.

So far I see the following issues:

1. It uses too much memory (#1723). This is because of general memory problems with the index in `internal/repository`, see #1988 but also because `internal/index` is used additional to the first index structure. This is the reason most users start complaining about memory consumption when using `prune`.
2. It is too slow, see  #1599, #2024, #2162. This is mainly because it reads all pack files (twice?). This completely makes 'prune' unusable for remote repositories with low bandwidth. Also not parallelized (#1470).
3. It is not customizable. Repacking is slow/expensive but cannot be turned of or customized (#1985, #2305)
4. It does too many things and is too complicated (#2227). It first rebuilds the index from pack files, then walks all snapshots and all dirs within to get the used blobs, then finds the packs where to delete these blobs and rewrites all those packs. Finally the index is again re-created completely from the resulting pack files. Hence it not only removes things, but reorganizes the repository and is even able to recover from some broken repo issues. However this is one source for the performance problems.

There are already some proposals to fix specific issues, see #1994, #2340, #2507.

I also started a rebuild of prune functionalities in #2513. Here I only used the index from `internal/repository` to clean unused blobs from the index, remove unused blobs and optionally repack if requested by the user. This solves IMO Points 1. to 3. and has already been successfully used by me and others for production repostitories.

About 1.: Using #2513 and the new commands, `internal/index` is no longer used and could also be removed if `prune` is substituted by these commands.
About 4.: There is so far a command missing to "repack" the index files if wanted (to get rid of small index files) and there is no way implemented to recover the index from the pack files. (i.e. the functionality of `rebuild-index`)
I can propose to work on these issues in #2513, if wanted.

So my question is: Where should `prune` and `rebuild-index` go to?
Is it a way to complete #2513 and then use all of this functionality in a new `prune` command?
It would be great to get some direction here!

pmkane:
Interested in this as well.

Sadly we had issues with #2340 and repo corruption, so had to discontinue our testing.

eigood:
Perhaps one should treat pruning like the java GC process?  Mark/Sweep?  Or any other technology terms from that problem space?

aawsome:
[...]

Issue #2126, opened 2018-12-27
restic stats is unreasonably slow in comparison to other operations


cdhowie:
Output of `restic version`
--------------------------
```
restic 0.9.3 compiled with go1.11.1 on linux/amd64
```

How did you run restic exactly?
-------------------------------
```
restic stats --mode restore-size
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Local disk.

Expected behavior
-----------------
This mode should take no longer than the part of `restic prune` that finds unused objects (about 15-20 minutes on this repository).

Actual behavior
---------------
This analysis takes multiple hours.

Do you have any idea what may have caused this?
-----------------------------------------------
My suspicion was that, while `restic prune`'s unused-object discovery mechanism prunes trees that have already been marked as used, `restic stats` on the other hand did not prune any subtrees and therefore has to recompute the size of each tree each time it is encountered.

However, looking through the source code, it seems that this is attempted?  Perhaps this code is not functioning correctly?

Do you have an idea how to solve the issue?
-------------------------------------------
Not without digging into the source code more.

cdhowie:
Timing from `restic stats --mode restore-size`:

```
real    131m25.498s
user    123m49.552s
sys     3m59.716s
```

Timing from `restic prune` on the same repository:

```
real    14m27.800s
user    8m14.352s
sys     0m25.212s
```

ngosang:
I can reproduce this issue in Restic 0.12.1 (Linux). For me, the issue with the `stats command` is not the time but the CPU/RAM usage.

I made a Prometheus exporter for Restic and I have to call the `stats` command multiple times in order to get the size and file count for the latest snapshots. I see there is no disk/network activity but the memory and CPU is measurable. Commands are running in the same machine the local repository is. My hardware is a RaspberryPi 4 with USB 3.0 storage. Other Restic commands run fine, at least I don't consider it's wasting resources.

![image](https://user-images.githubusercontent.com/10577978/156882618-a193bced-4ebd-4728-8394-f2695fb8c2a7.png)

MichaelEischer:
> I see there is no disk/network activity but the memory and CPU is measurable.

No disk activity cannot be correct as the stat command has to read all the metadata for a snapshot.

I think the issue you are actually looking for is #693.

ngosang:
Sorry, there is disk activity but it's normal. I edited my comment.
[...]

PR #2609, opened 2020-02-29
Make FindUsedBlobs parallel

aawsome:

Make FindUsedBlobs work parallel. This can speed up `prune` and `stat --mode raw-data`.

No. However, it is based on #2599 
Performance problems with `prune` have been already discussed, see #2547.
This PR may contribute to a solution of #2126 

Actually a similar parallelization could be used for `walker.Walk` impacting much more commands.
I can propose to work on this as soon as this PR has been discussed and merged,

Checklist
---------

### DIFF
--- a/internal/restic/find.go
+++ b/internal/restic/find.go
@@ -1,39 +1,103 @@
 package restic
 
-import "context"
+import (
+	"context"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+)
 
 // TreeLoader loads a tree from a repository.
 type TreeLoader interface {
 	LoadTree(context.Context, ID) (*Tree, error)
 }
 
+const numUsedBlobsWorkers = 4
+
+// queue implements a concurrency-safe queue of IDs
+// It can be accessed by the channel given with queue.Items()
+type queue struct {
+	q chan ID
+	sync.WaitGroup
+}
+
+// NewQueue() gives a new empty queue
+func NewQueue() *queue {
+	return &queue{q: make(chan ID)}
+}
+
+// Add() adds the given IDs to the queue
+func (q *queue) Add(ids IDs) {
+	q.WaitGroup.Add(len(ids))
+	go func() {
+		for _, id := range ids {
+			q.q <- id
+		}
+	}()
+}
+
+// Items() give a channel to listen for IDs in the queue.
+// When an ID is processed queue.Done() should be called!
+func (q queue) Items() <-chan ID {
+	return q.q
+}
+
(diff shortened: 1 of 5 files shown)


internal/restic/find.go
```diff
@@ -1,39 +1,103 @@
...
+}
+
+const numUsedBlobsWorkers = 4
+
+// queue implements a concurrency-safe queue of IDs
+// It can be accessed by the channel given with queue.Items()
+type queue struct {
+	q chan ID
+	sync.WaitGroup
+}
```
greatroar: Are you sure the WaitGroup is needed? I may be mistaken, but it seems to me like the following would suffice (untested):

```
type queue struct {
	q    chan ID
	size int64
}

func (q *queue) add(ids IDs) {
	atomic.AddInt64(&q.size, int64(len(ids)))
	go func() {
		for _, id := range ids {
			q.q <- id
		}
	}()
}

func (q *queue) pop() (id ID, ok bool) {
	id, ok = <-q.q
	if ok && atomic.AddInt64(&q.size, -1) == 0 {
		close(q.q)
	}
	return id, ok
}
```

This needs neither Wait nor Done, since FindUsedBlobs is already waiting on an errgroup.Group.

aawsome: From https://golang.org/pkg/sync/atomic/:

> On ARM, x86-32, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned. 

I  do not know why the implementation of `WaitGroup` does not use the fact imposed by the last sentence. I changed the code and used it. Can anyone double-check if I understood things correctly?

greatroar: From what I can tell, the alignment guarantees were [more lax](https://github.com/golang/go/commit/fea7c43ea238d8e18f3bafaac2740b3fa8975ae5) when the current [WaitGroup](https://github.com/golang/go/commit/03a48ebe1c79ed0437dcdbc8bf3cd58e901a999d) implementation was written. Then there's the noCopy, but I guess they could move that further down the struct.

MichaelEischer:
@aawsome did you have a look at https://github.com/restic/restic/pull/2340/files#diff-2512df4970ab0a01d5ae70affdfa443b ?

aawsome:
I'm closing this PR as I'm not confident about my way of implementing the unlimited queue (actually creates one goroutine per added entry which is quite some overhead compared to a recursive stack entry). First tests also did not show a clear improvement.

PR #2610, opened 2020-03-01
Remove vendored dependencies

### LINKED ISSUE #2482: Intent to drop the vendor/ folder
I'd like to drop the `vendor/` folder containing copies of all required libraries used by restic. It was nice to be able to build restic from a checkout of the repo (using `build.go`) without having to fetch external dependencies, here's a list of downsides:
 * It blows up the repository and the source code release
 * It makes reviewing pull requests which add or update dependencies harder because **many** different files are changed
 * It creates merge conflicts for many files (in `vendor`) when master and a PR both change dependencies, in contrast to conflicts in just two files (`go.mod` and `go.sum`)
 
Before Go Modules, vendoring (directly or via a tool like `dep`) was required to manage dependencies. We don't need that any more once we raise the Version required for restic to 1.11. All libraries are fetched for the build process on demand, their integrity is verified by the Go compiler using the hashes contained in `go.mod`, so that's not an issue. For Go >= 1.13, the Go module proxy is automatically activated and we can add instructions the README for the other versions, GitHub repos going away is not a problem any more.

Several other tools have dropped the `vendor/` folder since, and I'm not aware of any incidents or negative user feedback. Did you receive any for Caddy @mholt?

Any thoughts on that?

fd0:

This PR removes the vendored dependencies (in `vendor/`) and all code and documentation which mentions it. On build (or CI), the Go compiler automatically fetches the dependencies. It will also cryptographically verify that the correct code has been fetched by using the hashes in `go.sum` (see the [doc](https://golang.org/cmd/go/#hdr-Module_downloading_and_verification)).

Closes #2482

Checklist
---------


rawtaz (approved):
AILGTM!

PR #2614, opened 2020-03-02
Simplify internal/fs

greatroar:

It introduces various simplifications to the internal/fs package. In particular, the Write method on fs.File is gone, fs has been decoupled from testing, and fixpath has been simplified to use the [stdlib's handling of long path names](https://golang.org/doc/go1.8#os) on Windows.

No.

Checklist
---------

### DIFF
--- a/internal/textfile/read_test.go
+++ b/internal/textfile/read_test.go
@@ -3,25 +3,44 @@ package textfile
 import (
 	"bytes"
 	"encoding/hex"
+	"io/ioutil"
+	"os"
 	"testing"
-
-	"github.com/restic/restic/internal/fs"
 )
 
-func writeTempfile(t testing.TB, data []byte) (fs.File, func()) {
-	f, removeTempfile := fs.TestTempFile(t, "restic-test-textfile-read-")
+// writeTempfile writes data to a new temporary file and returns its name
+// and a callback that removes it.
+func writeTempfile(t testing.TB, data []byte) (string, func()) {
+	t.Helper()
 
-	_, err := f.Write(data)
+	f, err := ioutil.TempFile("", "restic-test-textfile-read-")
+	if err != nil {
+		t.Fatal(err)
+	}
+	name := f.Name()
+
+	defer func() {
+		closeErr := f.Close()
+		if err == nil && closeErr != nil {
+			err = closeErr
+		}
+		if err != nil {
+			os.Remove(name)
+			t.Fatal(err)
+		}
+	}()
+
+	_, err = f.Write(data)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	err = f.Close()
-	if err != nil {
-		t.Fatal(err)
+	return name, func() {
+		err := os.Remove(name)
+		if err != nil {
+			t.Fatal(err)
+		}
 	}
-
-	return f, removeTempfile
 }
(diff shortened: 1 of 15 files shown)


internal/textfile/read_test.go
```diff
@@ -60,10 +68,15 @@ func TestRead(t *testing.T) {
 				want = test.data
 			}
 
-			f, cleanup := writeTempfile(t, test.data)
-			defer cleanup()
+			tempname := writeTempfile(t, test.data)
+			defer func() {
```
MichaelEischer, line 72: I'd like to keep the cleanup function as a return value of `writeTempfile`, so that a user just has to call `defer cleanup()` as before

MichaelEischer (commented):
Please split the `fixpath` changes into a separate PR. The other changes are fine.

As reported by @JaCoB1123 the go standard library only prepends `\\?\` to long filepaths. However, that way short paths are still normalized which breaks backing up files with reserved filenames such as `test.` or `con.m`.

I also think that the fix to properly handle special filenames on windows has to be more extensive: With these changes the backup command will fail for the special files when using a relative path for backup  (we could maybe normalize the path at https://github.com/restic/restic/blob/34a3adfd8d2b84a5cf121e94174ff2dc94dd4a23/internal/archiver/archiver.go#L550 ). And the restorer in turn fails to restore special filenames (https://github.com/restic/restic/blob/34a3adfd8d2b84a5cf121e94174ff2dc94dd4a23/internal/restorer/restorer.go#L194).
May idea here would be to fix the file paths to include `\\?\` as soon as possible, as afterwards these paths will just behave properly.

internal/textfile/read_test.go
greatroar: Done.

greatroar:
Force-pushed a new version with the Windows path handling removed, except for the MkdirAll case. Also removed a stray Printf from fs_reader.go. Sorry for the wait.

greatroar:
Also removed a Go < 1.12 workaround.

internal/textfile/read_test.go
```diff
@@ -3,25 +3,40 @@ package textfile
...
+
+	defer func() {
+		err := f.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
 
-	err = f.Close()
+	_, err = f.Write(data)
```
MichaelEischer, line 29: Shouldn't the test also delete the tempfile if the write fails? Maybe add a add a `shouldDelete` flag and check that in the deferred function?

MichaelEischer (commented):
> Sorry for the wait.

No problem. There are plenty other PR to keep me busy ;-)

The PR is basically ready for merging, I've just noticed a small, already existing bug in the textfile read test, which would be good to fix since the PR already modifies the code.

I'm amazed by how many workarounds for old go version you uncover all the time :-) , thanks for that. The old `MkdirAll` didn't seem to call `fixpath`, so the restorer might have actually failed to restore directories with certain special names.

internal/textfile/read_test.go
greatroar: Good catch. I found a different way of doing this.

MichaelEischer (approved):
LGTM. Thanks for the cleanup :-)

PR #2616, opened 2020-03-03
Remove practically unused cache interface

greatroar:

It removes the restic.Cache interface, which was unused in the sense that exactly one code path passed values of it around, only to type-assert the values back to *cache.Cache at the point of use. It also removes an unused error type from the cache package.

No.

Checklist
---------


internal/restic/cache.go
```diff
@@ -1,37 +0,0 @@
...
-	// Wrap returns a backend with a cache.
-	Wrap(Backend) Backend
-
-	// IsNotExist returns true if the error was caused by a non-existing file.
-	IsNotExist(err error) bool
-
-	// Load returns a reader that yields the contents of the file with the
-	// given id if it is cached. rd must be closed after use. If an error is
-	// returned, the ReadCloser is nil. The files are still encrypted
-	Load(h Handle, length int, offset int64) (io.ReadCloser, error)
```
MichaelEischer, line 19: Please lower-case `Load`, `SaveWriter` and `Remove` in `internal/restic/file.go` to private methods. These aren't used anywhere outside the cache

MichaelEischer (commented):
Sorry for taking so long to review this, I just couldn't decide whether we should keep the Cache interface or not. But as the Cache is really only ever used by the actual repository implementation, I guess there's really not much use in keeping it.

Please rebase this PR to the current master.

internal/restic/cache.go
greatroar: Done.

MichaelEischer (approved):
LGTM. Thanks!

PR #2622, opened 2020-03-05
Fix PackerManager benchmark and optimize hashing.Writer

greatroar:

I was trying to see whether a few potential optimizations to hashing.Writer would make it faster. Before I could do that, I had to fix the PackagerManager benchmark to be more realistic.

The only actual optimization that had a (very minor) effect and didn't make the code unreadable is included as the third commit.

No.

Checklist
---------

### DIFF
--- a/internal/hashing/writer.go
+++ b/internal/hashing/writer.go
@@ -15,13 +15,14 @@ type Writer struct {
 func NewWriter(w io.Writer, h hash.Hash) *Writer {
 	return &Writer{
 		h: h,
-		w: io.MultiWriter(w, h),
+		w: w,
 	}
 }
 
 // Write wraps the write method of the underlying writer and also hashes all data.
 func (h *Writer) Write(p []byte) (int, error) {
 	n, err := h.w.Write(p)
+	h.h.Write(p[:n])
 	return n, err
 }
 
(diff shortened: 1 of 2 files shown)


internal/hashing/writer.go
```diff
@@ -15,13 +15,14 @@ type Writer struct {
...
 		h: h,
-		w: io.MultiWriter(w, h),
+		w: w,
 	}
 }
 
 // Write wraps the write method of the underlying writer and also hashes all data.
 func (h *Writer) Write(p []byte) (int, error) {
 	n, err := h.w.Write(p)
+	h.h.Write(p[:n])
```
MichaelEischer: The Hash interface states that a call to `Write()` never returns an error. Does this also apply to the number of written bytes?

greatroar: I'm not exactly sure if this is what you mean, but io.Writer "must return a non-nil error if it returns n < len(p)". So if a hash.Writer never returns an error, it must always return the length of its argument.

MichaelEischer: That's exactly the bit of information I've missed. Together with "hash.Write" never returns an error this implies that all n bytes are always written.

MichaelEischer (approved):
LGTM

PR #2623, opened 2020-03-06
internal/restic: close os.File after checking for error

alrs:
This fixes a spot in `internal/restic` where a nil `os.File` could attempt to `Close()` even though `fs.Open()` returned an error.


greatroar (approved):
Doing the Close later won't actually cause a panic because os.File.Close checks for this, but it is cleaner.

greatroar:
I hope nobody minds me approving changes without having write access.

fd0:
> I hope nobody minds me approving changes without having write access.

@greatroar not at all, thanks! :)

fd0 (approved):

PR #2624, opened 2020-03-06
Simplify and optimize validNonce

greatroar:

<del>crypto.validNonce was optimized for older Go versions, but Go 1.14 is smart enough to optimize the obvious implementation to 4x the speed of the current version.</del>

Checklist
---------

### DIFF
--- a/internal/crypto/crypto.go
+++ b/internal/crypto/crypto.go
@@ -237,11 +237,12 @@ var ErrInvalidCiphertext = errors.New("invalid ciphertext, same slice used for p
 
 // validNonce checks that nonce is not all zero.
 func validNonce(nonce []byte) bool {
-	var sum byte
 	for _, b := range nonce {
-		sum |= b
+		if b != 0 {
+			return true
+		}
 	}
-	return sum > 0
+	return false
 }
 
 // statically ensure that *Key implements crypto/cipher.AEAD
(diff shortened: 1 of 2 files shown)


MichaelEischer:
The old and the new implementations yield rather similar assembly code. The main difference is in the number of loop iterations: Your optimized version leaves the function after on average 1.01 iterations, whereas the old implementation always takes exactly `len(nonce)` iterations. That is the speedup is a result of changing from a constant-time implementation to a non constant-time one.

As a constant-time implementation is usually used to avoid side-channels, this poses the question whether the change might introduce information leaks. `Key.Open` bails if the nonce is all zeros, the timing difference could be used to tell which byte of the nonce contained the first non-zero byte. In case of unpacked blobs this won't matter anyways as the nonce is always in the first few bytes and thus no new information is revealed. For packs it could speed up how an attacker could determine the blob boundaries within a pack file, more precisely they would just need one try to determine the offset `blobOffset%len(nonce)` instead of trying every offset between zero and `len(nonce)`. But AFAICT you would still need some sort of binary search to determine the overall location within a pack file. So it probably doesn't matter much whether `validNonce` uses a constant time implementation or not.

greatroar:
I should have checked the assembler. Shouldn't validNonce be using the crypto/subtle package if this is the issue? Right now, the only semantic difference between the two implementations is that the current one forces a panic if the nonce is not the right length. A more aggressive inliner in the next Go release could still produce a non-constant implementation in contexts where the size is known, notably Repository.LoadAndDecrypt.

MichaelEischer:
It probably doesn't matter much whether `validNonce` is constant-time or not. But I don't have enough insight into all possible cryptographic implications which such a change could have.

Issue #2591, opened 2020-02-19
Reading data from stdin causes password to be read from stdin


mschoettle:
Output of `restic version`
--------------------------

`restic 0.9.6 compiled with go1.13.4 on linux/arm`

How did you run restic exactly?
-------------------------------

`sudo mysqldump --single-transaction -h <host> -u <user> <db> | sudo restic -r <repo> backup --stdin --stdin-filename production.sql`

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

local for testing purposes

Expected behavior
-----------------

Ask me for the password, then backup the SQL dump as explained in the [docs](https://restic.readthedocs.io/en/latest/040_backup.html#reading-data-from-stdin)

Actual behavior
---------------

```
read password from stdin
Fatal: wrong password or no key found
```

Specifying `--password-file` works. Using an environment variable (`RESTIC_PASSWORD`) or supplying the password on request does not.

Steps to reproduce the behavior
-------------------------------

See command to execute above.

Do you have any idea what may have caused this?
-----------------------------------------------

Not really :) Seems to me that specifying something causes it to expect the password coming from stdin when it shouldn't.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yes! It is very easy to setup and I really like the features. Thanks a lot!

PR #2626, opened 2020-03-06
Specify stdin for external password command

juergenhoetzel:
This allows reading the password when terminal interaction is
required (for example, pinentry-curses).

Signed-off-by: Juergen Hoetzel <juergen@archlinux.org>

-----------------

This is just a follow up  for
[Add support for reading password from external command](https://github.com/restic/restic/commit/df7f72cdde2419937c15377d46a311e5d734a245)

currently `stdin` is connected to `/dev/null` and `pinentry-curses` fails:

```
→ export RESTIC_PASSWORD_COMMAND='pass Backup/restic@serenity'
→ restic backup ~
gpg: decryption failed: No secret key
Resolving password failed: exit status 2
```

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -268,6 +268,7 @@ func resolvePassword(opts GlobalOptions) (string, error) {
 		}
 		cmd := exec.Command(args[0], args[1:]...)
 		cmd.Stderr = os.Stderr
+		cmd.Stdin = os.Stdin
 		output, err := cmd.Output()
 		if err != nil {
 			return "", err


MichaelEischer:
Is `os.Stdin` the right choice or should it rather be `/dev/tty` (or it's windows equivalent) as in #2593? As far as I can tell, using `os.Stdin` would break in combination with backups that read from stdin.

greatroar:
Passing stdin is going to cause the trouble with backup --stdin described in #2591. This shouldn't be too hard to combine with #2593, though.

greatroar:
Regarding the example, it shows pass, but the description is about pinentry-curses. How does one use the latter program with restic?

juergenhoetzel:
> Is `os.Stdin` the right choice or should it rather be `/dev/tty` (or it's windows equivalent) as in #2593? As far as I can tell, using `os.Stdin` would break in combination with backups that read from stdin.

IMO managing and opening a specific IO device  should be the task of the password manager.
In the case of `pinentry` (used by `gpg-egent`) there are many options for example [--ttyname].(https://www.gnupg.org/documentation/manuals/gnupg/Agent-Options.html)

`pinentry-tty` also configures various terminal options via `tcsetattr`.

Other programs (`pinentry-qt`) do not use terminal devices at all:
`restic`  should not make any assumptions about how the password is read.

In the case of #2593 restic itself is responsible for reading the password, so opening the  terminal device is fine.

juergenhoetzel:
> Regarding the example, it shows pass, but the description is about pinentry-curses. How does one use the latter program with restic?

`pass` uses  `gpg`/`gpg-agent` to decrypt the password, `gpg-agent` again  uses the configured (via `/.gnupg/gpg-agent.conf`) `pinentry` program to perform the password dialog.

greatroar:
Thanks for the clarification.

> IMO managing and opening a specific IO device should be the task of the password manager.

Agreed, but in this patch, the I/O device is a file descriptor *from restic* and the password manager apparently doesn't work without restic's help. It would be nice if restic could make no assumptions at all, but it does have to assume that the password manager doesn't mess with its input stream.

juergenhoetzel:
> Agreed, but in this patch, the I/O device is a file descriptor _from restic_ and the password manager apparently doesn't work without restic's help. It would be nice if restic could make no 

`pinentry-curses` will fail when stdin is not a terminal 

```C
if (!(isatty(fileno(stdin)) && isatty(fileno(stdout))))
            {
              errno = ENOTTY;
              pinentry->specific_err = gpg_error_from_syserror ();
              pinentry->specific_err_loc = "isatty";
              return confirm_mode? 0 : -1;
            }
```

on the other hand, if the password command doesn't read from stdin, everything should work fine backuping `stdin`:

```
echo foo |./restic --password-command="echo secret" backup --stdin
repository 02bc6512 opened successfully, password is correct

Files:           0 new,     1 changed,     0 unmodified
Added to the repo: 302 B

processed 1 files, 4 B in 0:00
snapshot bac788f0 saved

```

greatroar:
Right, so even with this patch, pinentry-curses won't work. We need to combine this with #2593 to get reliable behavior.

Judging from that code, we have to give pinentry-curses a terminal on stdout, too.

greatroar:
I just pushed a branch [password-command](https://github.com/greatroar/restic/tree/password-command) to my fork that should do the trick. It's untested.

fd0:
What's the current state of this PR? Are you still interested in getting this merged/resolved? :)

juergenhoetzel:
After thinking about it again, I don't think it's worth dealing with this special case, especially because it causes problems together with the `--stdin` option.

PR #2629, opened 2020-03-07
repository: Avoid duplicate index lookup when loading blob/tree

MichaelEischer:
`LoadBlob` and `LoadTree` lookup the blob size in the index and use it to allocate a buffer for `loadBlob` which has to run another index lookup. Especially for large repositories index lookups are rather expensive, see #2523 .

This PR lets `loadBlob()` allocate the blob buffer itself and thus removes the duplicate index lookup.

I did not include a changelog entry as this PR only slightly improves performance.

No, the inefficiency showed up during work on #2328.

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -128,14 +128,25 @@ func (r *Repository) sortCachedPacks(blobs []restic.PackedBlob) []restic.PackedB
 // loadBlob tries to load and decrypt content identified by t and id from a
 // pack from the backend, the result is stored in plaintextBuf, which must be
 // large enough to hold the complete blob.
-func (r *Repository) loadBlob(ctx context.Context, id restic.ID, t restic.BlobType, plaintextBuf []byte) (int, error) {
-	debug.Log("load %v with id %v (buf len %v, cap %d)", t, id, len(plaintextBuf), cap(plaintextBuf))
+func (r *Repository) loadBlob(ctx context.Context, id restic.ID, t restic.BlobType, plaintextBuf []byte) ([]byte, int, error) {
+	if plaintextBuf == nil {
+		debug.Log("load %v with id %v (allocating buf)", t, id)
+	} else {
+		debug.Log("load %v with id %v (buf len %v, cap %d)", t, id, len(plaintextBuf), cap(plaintextBuf))
+	}
 
 	// lookup packs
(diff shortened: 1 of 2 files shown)


internal/repository/repository.go
```diff
@@ -128,14 +128,25 @@ func (r *Repository) sortCachedPacks(blobs []restic.PackedBlob) []restic.PackedB
 // loadBlob tries to load and decrypt content identified by t and id from a
 // pack from the backend, the result is stored in plaintextBuf, which must be
 // large enough to hold the complete blob.
-func (r *Repository) loadBlob(ctx context.Context, id restic.ID, t restic.BlobType, plaintextBuf []byte) (int, error) {
-	debug.Log("load %v with id %v (buf len %v, cap %d)", t, id, len(plaintextBuf), cap(plaintextBuf))
+func (r *Repository) loadBlob(ctx context.Context, id restic.ID, t restic.BlobType, plaintextBuf []byte) ([]byte, int, error) {
```
greatroar: The second return value is redundant. The []byte can be sliced to the right length before return.

internal/repository/repository.go
```diff
@@ -128,14 +128,25 @@ func (r *Repository) sortCachedPacks(blobs []restic.PackedBlob) []restic.PackedB
...
-		return 0, errors.Errorf("id %v not found in repository", id)
+		return plaintextBuf, 0, errors.Errorf("%v %v not found in repository", t, id)
+	}
+
+	if plaintextBuf == nil {
+		size := uint(restic.PlaintextLength(int(blobs[0].Length)))
+
+		debug.Log("size is %d, create buffer", size)
+		plaintextBuf = restic.NewBlobBuffer(int(size))
 	}
```
greatroar: Instead of treating nil as the special value, wouldn't it be more idiomatic to let the caller pass in arbitrary scratch space and allocate if it turns out to be too small? I can't recall a single instance where an empty []byte is treated differently from nil in the stdlib.

```
// untested
size := restic.PlaintextLength(int(blobs[0].Length))
if cap(plaintextBuf) > size {
	plaintextBuf = plaintextBuf[:size]
} else if cap(plaintextBuf) < size {
	debug.Log("size is %d, create buffer", size)
	plaintextBuf = restic.NewBlobBuffer(size)
}
```

Issue #2618, opened 2020-03-04
Backups segfault on ARMv7


richard-scott:
I build restic with Jenkins from Github master as and when commits are pushed.
Until this time I have not had any issues with it backing files up... until now ;-(

Output of `restic version`
--------------------------
```
# restic version
debug log file debug.log
debug enabled
restic 0.9.6 (v0.9.6-135-ga307797c) compiled with go1.14 on linux/arm
```

How did you run restic exactly?
-------------------------------
```
export B2_ACCOUNT_ID="XXXX"
export B2_ACCOUNT_KEY="XXXX"
export RESTIC_REPOSITORY="b2:restic-$(hostname | md5sum | awk '{print $1}')-$(hostname)"
export RESTIC_PASSWORD_FILE="/root/.restic"
export RESTIC_CACHE_DIR="/data/.restic"
restic init
restic backup /etc
restic backup /etc
```

The 1st backup of /etc will work ok, but the 2nd backup is when things segfault.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Backblaze B2

Expected behavior
-----------------
Restic should backup correctly.
My last successful backup on this host was on the 27th Feb 2020.
(I have another linux host that bacups hourly and is working just fine.)

Actual behavior
---------------
It dies with a Segmentation fault :-(
```
# restic backup /etc
debug log file debug.log
debug enabled
repository 4357d291 opened successfully, password is correct
Segmentation fault files 4.264 MiB, total 941 files 6.448 MiB, 0 errors ETA 0:47
root@ReadyNAS:~/test#
/etc/wgetrc
```
Each time I run a backup it crashes but It doesn't stop at the same point in time or file.

Steps to reproduce the behavior
-------------------------------
I'm just running a backup (well, attempting to)

Do you have any idea what may have caused this?
-----------------------------------------------
I am using a [ReadyNAS Duo](https://www.netgear.com/support/product/RND2000v1_(ReadyNAS_Duo_v1).aspx) to run this backup on. It has 512MB of RAM, and lots of swap space allocated that it never actually uses.

Do you have an idea how to solve the issue?
-------------------------------------------
Restic backs up just fine on a Linux x86_64 system with 16GB RAM. Maybe its a memory thing, but its odd that its only just stopped working.
I have deleted my repo and started again with a fresh one at Backblaze so I don't believe data to be corrupted.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Yes, and no.

[...]

PR #2643, opened 2020-03-12
Lighter init and buffered channels for signals

greatroar:

It makes restic's startup lighter by not postponing restic.Progress initialization. It also makes all channels passed to signal.Notify buffered.

The number of commits in this PR is large because it includes part of #2630, which I deemed uncontroversial, and which already fixed part of the problem. The reason that lighter initialization is part of it at all is that I first tried getting right of init functions, then figured out that signal handling is the actual problem.

The segfault reported at #2618 is very likely the result of improper signal.Notify use. This does not fix the issue entirely, but reduces it to the well-known "restic uses much memory" problem.

Checklist
---------

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -20,7 +20,7 @@ var cleanupHandlers struct {
 var stderr = os.Stderr
 
 func init() {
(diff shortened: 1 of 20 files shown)


greatroar:
Sorry, wrong branch. See #2644.

PR #2658, opened 2020-03-20
Don't echo authentication passwords (rest backend)

### LINKED ISSUE #2241: Don't echo authentication passwords
Output of `restic version`
--------------------------
`restic 0.9.4 compiled with go1.11.4 on darwin/amd64`

```
export RESTIC_REPOSITORY=rest:https://redacted-username:redacted-password@redacted-host/redacted-path/
export RESTIC_PASSWORD=redacted
restic init -v
```

`rclone serve restic` (via an nginx reverse proxy)

Expected behavior
-----------------

I have chosen to use a remote restic server (actually rclone) over TLS and requiring http authentication. However restic is a little too eager to echo the full repository URL that includes the http password. Needless to say this password is sensitive, since it allows access to push/pull data to the rclone server. It would be better for this not be visible in the output (and therefore logs etc)

Actual behavior
---------------

Full repository url is echoed that includes the remote password.

```
$ restic init
created restic repository debc97928d at rest:https://redacted-username:redacted-password@redacted-host/redacted-path/
```

Not sure, presumably repository URLs are assumed not to be sensitive, but those that include a password are.

Suggestions:
- Parse the repository url, censor the password portion when echoed
- Allow restic rest protocol credentials to be provided as environment variables (RESTIC_REST_USERNAME / RESTIC_REST_PASSWORD ?)

I love the simplicity and flexibility and I'm confident it'll be a really useful part of my backup strategy.

creativeprojects:

Bugfix: Don't echo authentication passwords

When using a password in the URL of a rest backend, in some cases
the password was displayed in the restic log:
- when creating a repository
- when restic failed to open the repository

The URL should now be displayed with "***" in place of the password.

Fixes issue https://github.com/restic/restic/issues/2241

closes #2241 

Checklist
---------

- ~I have added documentation for the changes (in the manual)~

Question before ticking the last box:
--------------

Do you know of any other place where the password could be displayed in the log? I cannot find any other place.

### DIFF
--- a/changelog/unreleased/issue-2241
+++ b/changelog/unreleased/issue-2241
@@ -1,0 +1,10 @@
+Bugfix: Hide password in REST backend repository URLs
+
+When using a password in the REST backend repository URL,
+the password could in some cases be included in the output
+from restic, e.g. when initializing a repo or during an error.
+
+The password is now replaced with "***" where applicable.
+
+https://github.com/restic/restic/issues/2241
+https://github.com/restic/restic/pull/2658
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -631,7 +631,7 @@ func parseConfig(loc location.Location, opts options.Options) (interface{}, erro
 
 // Open the backend specified by a location config.
 func open(s string, gopts GlobalOptions, opts options.Options) (restic.Backend, error) {
-	debug.Log("parsing location %v", s)
+	debug.Log("parsing location %v", location.StripPassword(s))
 	loc, err := location.Parse(s)
 	if err != nil {
 		return nil, errors.Fatalf("parsing repository location failed: %v", err)
@@ -686,13 +686,13 @@ func open(s string, gopts GlobalOptions, opts options.Options) (restic.Backend, 
 	}
 
 	if err != nil {
-		return nil, errors.Fatalf("unable to open repo at %v: %v", s, err)
+		return nil, errors.Fatalf("unable to open repo at %v: %v", location.StripPassword(s), err)
 	}
 
 	// check if config is there
 	fi, err := be.Stat(globalOptions.ctx, restic.Handle{Type: restic.ConfigFile})
 	if err != nil {
-		return nil, errors.Fatalf("unable to open config file: %v\nIs there a repository at the following location?\n%v", err, s)
+		return nil, errors.Fatalf("unable to open config file: %v\nIs there a repository at the following location?\n%v", err, location.StripPassword(s))
 	}
 
 	if fi.Size == 0 {
--- a/internal/backend/rest/config.go
+++ b/internal/backend/rest/config.go
@@ -31,10 +31,7 @@ func ParseConfig(s string) (interface{}, error) {
 		return nil, errors.New("invalid REST backend specification")
 	}
 
-	s = s[5:]
-	if !strings.HasSuffix(s, "/") {
-		s += "/"
-	}
+	s = prepareURL(s)
 
 	u, err := url.Parse(s)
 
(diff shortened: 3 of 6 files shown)


changelog/unreleased/issue-2241
```diff
@@ -0,0 +1,10 @@
+Bugfix: Don't echo authentication passwords
```
MichaelEischer, line 1: That sounds way to broad, please be a bit more specific.

changelog/unreleased/issue-2241
```diff
@@ -0,0 +1,10 @@
...
+
+When using a password in the URL of a rest backend, in some cases
+the password was displayed in the restic log:
+- when creating a repository
+- when restic failed to open the repository
+
+The URL should now be displayed with "***" in place of the password.
```
MichaelEischer, line 8: "should" or "is"?

cmd/restic/global.go
```diff
@@ -659,13 +659,13 @@ func open(s string, gopts GlobalOptions, opts options.Options) (restic.Backend,
 	}
 
 	if err != nil {
-		return nil, errors.Fatalf("unable to open repo at %v: %v", s, err)
+		return nil, errors.Fatalf("unable to open repo at %v: %v", location.StripPassword(s), err)
```
MichaelEischer, line 662: Please also mask the url in the `debug.Log` call in line 607.

internal/backend/rest/config.go
```diff
@@ -46,3 +43,37 @@ func ParseConfig(s string) (interface{}, error) {
...
+		return scheme + s
+	}
+
+	if _, set := u.User.Password(); !set {
+		return scheme + s
+	}
+
+	// A password was set: we replace the UserInfo and rebuild the string from there
+	u.User = url.UserPassword(u.User.Username(), "***")
+	stripped, err := url.QueryUnescape(u.String())
```
MichaelEischer, line 65: I'm not sure whether it's a good idea to call `QueryUnescape`?

The go standard library just replaces strings: https://github.com/golang/go/blob/d91d0762c7757c12c7d5f9e2ae3f170d5bd7ba84/src/net/http/client.go#L1004
Or less hacky: Mirror the url.Redacted() method added in go 1.15 (https://golang.org/src/net/url/url.go?s=25219:25250#L852) and replace the password with "xxxxx"

MichaelEischer:
Please rebase the PR onto the current master branch.

creativeprojects:
Sorry, this was my intention but I pushed the wrong branch 😥

MichaelEischer:
> Sorry, this was my intention but I pushed the wrong branch 😥

Do you need help with rebasing?

internal/backend/rest/config.go
creativeprojects: I remember why I was doing it, it's because the * was converted into an html character. I've changed to an 'x' instead (like in go 1.15 library)

creativeprojects:
I think I managed, it's just something I don't do every day so I forget the git commands 👍 thanks

rawtaz:
Hmm, is this specific to the REST backend, or are there other backends that it affects/fixes too? In other words, is this something that is only worth mentioning for the REST backend?

MichaelEischer:
@rawtaz The PR only affects the REST backend. I'm not aware that other backends have the option to include a password in the repository address.

changelog/unreleased/issue-2241
```diff
@@ -0,0 +1,10 @@
+Bugfix: Hide URL passwords from logs
+
+When using a password in the URL of a rest backend, in some cases
+the password was displayed in the restic log:
+- when creating a repository
+- when restic failed to open the repository
+
+The URL is now displayed with "***" in place of the password.
```
rawtaz, line 8: Since there is no "log" in restic, only its output, and the changelog generator restic uses doesn't support lists, I think it's better to formulate this as follows:

~~~
Bugfix: Hide password in REST URLs

When using a password in the REST backend repository URL,
the password could in some cases be included in the output
from restic, e.g. when initializing a repo or during an error.

The password is now replaced with "***" where applicable.
~~~

rawtaz:
I noticed that the changelog entry said the password was replaced with "***", which is what I think is most conventional and makes more sense than "xxxxx". Can you please make the code and also the tests use "***" for the obfuscation instead of "xxxxx"?

MichaelEischer:
@creativeprojects Could you also mask the repository URL in the `debug.Log` call in line 607 of `cmd/restic/global.go`?

creativeprojects:
> Could you also mask the repository URL in the `debug.Log` call in line 607 of `cmd/restic/global.go`?

I was a bit puzzled by it, and couldn't find it the other day. But I think I found it now: do you mean the one in line 617?

https://github.com/creativeprojects/restic/blob/issue-2241/cmd/restic/global.go#L617

MichaelEischer:
> I was a bit puzzled by it, and couldn't find it the other day. But I think I found it now: do you mean the one in line 617?
> 
> https://github.com/creativeprojects/restic/blob/issue-2241/cmd/restic/global.go#L617

Oh, sorry for the confusion, apparently the `debug.Log` statement moved in the meantime. I had `debug.Log("parsing location %v", s)` in mind which is now in line 634 on your branch.

rawtaz (approved):
Looks good to me, thanks a lot for your patience and quick updates after our comments!

MichaelEischer (approved):
LGTM. Thanks for the PR.
I've made one minor change and added a link to this PR to the changelog.

PR #2661, opened 2020-03-21
"Everything is ok" message after retrying

### LINKED ISSUE #1734: "Everything is ok"-Error message
How about adding a "Everything is ok"-Message, after restic printed a does not exist errormessage?

## Expected behavior

Load(<data/xxxxxxxxxx>, 0, 0) returned error, retrying after xxx.xxxxxxms: <data/xxxxxxxxxx> does not exist
Load(<data/xxxxxxxxxx>, 0, 0) found it after x retries: <data/xxxxxxxxxx> is ok

## Actual behavior

Load(<data/xxxxxxxxxx>, 0, 0) returned error, retrying after xxx.xxxxxxms: <data/xxxxxxxxxx> does not exist

Add another errormessage which tells your that the data was found.

yep, made me very happy that fd0 is always listening to my bullshit in IRC ;)

creativeprojects:

When a recoverable error is encountered, restic displays a warning message saying it's retrying, like this one:

`Save(<data/956b9ced99>) returned error, retrying after 357.131936ms: CreateBlockBlobFromReader: Put https://example.blob.core.windows.net/restic/data/95/956b9ced99...aac: write tcp 192.168.11.18:51620->1.2.3.4:443: write: connection reset by peer`

This message can be a little confusing because it seems like there was an error,
but we're not sure if it was actually fixed by retrying.

restic is now displaying a confirmation that the action was successful after retrying:

`Save(<data/956b9ced99>) operation successful after 1 retries`

issue https://github.com/restic/restic/issues/1734

Yes

closes #1734 

Checklist
---------

- [x] ~I have added documentation for the changes (in the manual)~

### DIFF
--- a/changelog/unreleased/issue-1734
+++ b/changelog/unreleased/issue-1734
@@ -1,0 +1,16 @@
+Enhancement: "Everything is ok" message after retrying
+
+When a recoverable error is encountered, restic displays a warning message
+saying it's retrying, like this one:
+
+Save(<data/956b9ced99>) returned error, retrying after 357.131936ms: CreateBlockBlobFromReader: Put https://example.blob.core.windows.net/restic/data/95/956b9ced99...aac: write tcp 192.168.11.18:51620->1.2.3.4:443: write: connection reset by peer
+
+This message can be a little confusing because it seems like there was an
+error, but we're not sure if it was actually fixed by retrying.
+
+restic is now displaying a confirmation that the action was successful after retrying:
+
+Save(<data/956b9ced99>) operation successful after 1 retries
+
+https://github.com/restic/restic/issues/1734
+https://github.com/restic/restic/pull/2661
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -439,9 +439,13 @@ func OpenRepository(opts GlobalOptions) (*repository.Repository, error) {
 		return nil, err
 	}
 
-	be = backend.NewRetryBackend(be, 10, func(msg string, err error, d time.Duration) {
+	report := func(msg string, err error, d time.Duration) {
 		Warnf("%v returned error, retrying after %v: %v\n", msg, d, err)
-	})
+	}
+	success := func(msg string, retries int) {
+		Warnf("%v operation successful after %d retries\n", msg, retries)
+	}
+	be = backend.NewRetryBackend(be, 10, report, success)
 
 	// wrap backend if a test specified a hook
 	if opts.backendTestHook != nil {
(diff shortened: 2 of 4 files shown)


ifedorenko:
I think it is better to suppress transient/recoverable error messages, at least by default, and only show errors when restic decides to fail the operation.

creativeprojects:
I was scratching my head on how to see actually it in action, and I found out that replacing the `retryNotifyErrorWithSuccess` function with this one is probably the best way:

```go
func retryNotifyErrorWithSuccess(operation backoff.Operation, b backoff.BackOff, notify backoff.Notify, success func(retries int)) error {
	if success == nil {
		return backoff.RetryNotify(operation, b, notify)
	}
	retries := 0
	operationWrapper := func() error {
		// For live testing...
		if rand.Intn(9) == 0 {
			retries++
			return errors.New("Temporary failure")
		}
		//
		err := operation()
		if err != nil {
			retries++
		} else if retries > 0 {
			success(retries)
		}
		return err
	}
	return backoff.RetryNotify(operationWrapper, b, notify)
}
```

creativeprojects:
> I think it is better to suppress transient/recoverable error messages, at least by default, and only show errors when restic decides to fail the operation.

Yes I suppose that's also a possibility 🤔

jkaberg:
I agreed with @ifedorenko, I count an successfull retry as success and idealy don't want/need an message for this. Only if it fails completly.

creativeprojects:
I've committed a change: the retry messages (and ok message) should **not** be displayed in quiet mode (-q). Only a final fatal error would be sent to _stderr_.

Retry messages (and ok messages) are still displayed in normal mode (no **-q**, no **-v**)

Does that sound better?

MichaelEischer:
I've been thinking quite a bit about this issue. I'd like to keep the retry warnings as they have proven to be necessary for debugging of for example prune problems. On the other hand, they can be ignored if everything works. However, being able to debug problems is in my opinion more important, especially as one can rely on the restic's return code to be notified about serious problems. So as long as we have no way to keep some logs for later debugging I'm against hiding the retry warnings.

Regarding the "Everything is ok" message I think that it would definitely be useful. So let's limit this PR to just add a notification after a successful retry.

MichaelEischer:
I've rebased the PR and dropped the last commit to turn the warnings into verbose print statements. The main reason for that is debugability as outlined in my last comment. In addition, switching from warnings to verbose log output moves the retry warnings from stderr to stdout and would break the json output in the progress.

creativeprojects:
> switching from warnings to verbose log output moves the retry warnings from stderr to stdout and would break the json output in the progress

That's right, sorry I didn't think of that.

I'm good with this version, thanks for doing the rebase 👍🏻

MichaelEischer (approved):
LGTM.

MichaelEischer:
The last few pushes were just attempts to fix the tests. Apparently my local `go test ./...` runs and the CI had different opinions on whether the tests compile or not.

PR #2668, opened 2020-03-26
stats: Fix crash in blobs-per-file mode on missing blob

MichaelEischer:
Restic crashes when running `restic stats --mode blobs-per-file` for a damaged repository with missing blobs.

The error message tried to dereference the subtreeID field of the current node, which is a file however. Said field is set to nil for a file thus causing a segfault when dereferenced.

Fix this by using the actual parentTreeID.

We could add a changelog entry but I doubt that it's worth it, as apparently no-one ever used that command on a repository with missing blobs before.

The crash was reported in https://forum.restic.net/t/fail-restic-stats-mode-raw-data/2601/14

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2668
+++ b/changelog/unreleased/pull-2668
@@ -1,0 +1,6 @@
+Bugfix: Don't abort the stats command when data blobs are missing
+
+Runing the stats command in the blobs-per-file mode on a repository with
+missing data blobs previously resulted in a crash.
+
+https://github.com/restic/restic/pull/2668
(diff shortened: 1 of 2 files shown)


rawtaz:
I agree with your reasoning, but let's add a bugfix changelog entry nonetheless. After all, a bug was actually fixed :)

rawtaz:
I was thinking about that message the other day.. I'm pondering whether we should suggest creating an issue in the first place? I mean, is it not more suitable for the forum (getting help)? Also, why do we put out a message like this for this type of error, when we don't do it for other errors. I'm thinking we should perhaps remove it entirely instead - if users want help with it, they will most likely post in the forum anyway.

fd0 (approved):
Oh, good catch!

Issue #2663, opened 2020-03-24
Fatal: number of used blobs is larger than number of available blobs


danielb2:
Also having a lot of errors when doing a `restic check`

Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.7 on freebsd/amd64

How did you run restic exactly?
-------------------------------
restic prune

variables:
```
set -gx AWS_ACCESS_KEY_ID secret
set -gx AWS_SECRET_ACCESS_KEY secret
set -gx RESTIC_HOME $HOME/.config/restic
set -gx RESTIC_PASSWORD_FILE $RESTIC_HOME/etc/pass
set -gx RESTIC_INCLUDE $RESTIC_HOME/include
set -gx RESTIC_REPOSITORY s3:https://secret/bkup
```
I don't remember which variables are used for restic but here are the ones I've set for myself.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
AWS

Expected behavior
-----------------

I expected prune to prune the data

Actual behavior
---------------

```
[~]$ restic prune
repository 3a6e47e7 opened successfully, password is correct
counting files in repo
building new index for repo
[17:12] 100.00%  150137 / 150137 packs
repository contains 150137 packs (676019 blobs) with 718.767 GiB
processed 676019 blobs: 0 duplicate blobs, 0 B duplicate
load all snapshots
find data that is still in use for 23 snapshots
[0:32] 100.00%  23 / 23 snapshots
Fatal: number of used blobs is larger than number of available blobs!
Please report this error (along with the output of the 'prune' run) at
https://github.com/restic/restic/issues/new
```

I was prompted to file this ticket

Steps to reproduce the behavior
-------------------------------

restic prune

Do you have any idea what may have caused this?
-----------------------------------------------

no idea. the tool told me to file the issue

Do you have an idea how to solve the issue?
-------------------------------------------

none

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

it makes me happy almost every day

dimejo:
Quote from the issue template:
> Fill out the issue template below - not doing so needs a good reason.

It probalbly doesn't make a difference in this case (as there are plenty of bug reports) but I'm still wondering why you didn't fill out the issue template. How do you expect developers to help when you don't provide any information besides the error messages?

Related issues:
#2642 
#2400 
#2385
#2334 
#2273
#2045
#2024

danielb2:
It wasn't  intentional. I didn't get prompted.

[...]

PR #2670, opened 2020-03-26
prune: Point bug report link to the bug issue template

MichaelEischer:
When prune failed with the "dreaded" `number of used blobs is larger than number of available blobs` error it showed a link pointing to a blank issue. Direct the restic user to the bug template instead.

As an alternative it would also be possible to point users to the issue template chooser. However, that would mean one extra step to take in addition to filling out the issue template.

The change was suggested in https://github.com/restic/restic/issues/2663#issuecomment-604138591 .

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -210,7 +210,7 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 	if len(usedBlobs) > stats.blobs {
 		return errors.Fatalf("number of used blobs is larger than number of available blobs!\n" +
 			"Please report this error (along with the output of the 'prune' run) at\n" +
-			"https://github.com/restic/restic/issues/new")
+			"https://github.com/restic/restic/issues/new?template=Bug.md")
 	}
 
 	Verbosef("found %d of %d data blobs still in use, removing %d blobs\n",


rawtaz:
I was thinking about that message the other day.. I'm pondering whether we should suggest creating an issue in the first place? I mean, is it not more suitable for the forum (getting help)? Also, why do we put out a message like this for this type of error, when we don't do it for other errors. I'm thinking we should perhaps remove it entirely instead - if users want help with it, they will most likely post in the forum anyway.

dimejo:
> I'm thinking we should perhaps remove it entirely instead - if users want help with it, they will most likely post in the forum anyway.

Agreed.

MichaelEischer:
I think the original idea was to get bug reports for damaged repositories. Without further guidance the bug reports will quite likely land in the forum or users will just start with a new backup. My assumption would be that asking for a bug report along with providing a link, makes it more likely that we get bug reports.

And as these prune problems show up every now and then without clear traces, some further problem reports could be helpful.

rawtaz:
But still, why do you we have it for this specific error message and not for others that too indicate a broken repository?

If we keep it, I think we should make it point to the Getting Help section in the forum instead, because that's the place were we debug broken repositories, rather than in a GitHub issue. Thoughts?

MichaelEischer:
>  But still, why do you we have it for this specific error message and not for others that too indicate a broken repository?

Hmm, that's a good question. I guess that specific prune error is not much different from other errors in prune that indicate repository damages. 

Is there already an issue that discusses how useful error messages should look like? I've only found issues that discuss logging or when restic should continue after an error.

For prune my proposal would be to point users at a forum post (or Github issue?) which would serve as a FAQ and starting point for further debugging (on a separate thread). The idea behind pointing at a post and not just the forum in general is that this would allow providing instructions on how common problems can be solved if the need arises. However, for errors with a known solution, prune should probably just print that.

MichaelEischer:
Closing this PR as it has to be redone anyways for #2683.

PR #2674, opened 2020-03-31
prune: Stricter error checks

MichaelEischer:
This PR tightens the sanity checks made by prune in the following three regards:

- Prune now exactly checks that every single used blob can be found and no longer just approximately checks this condition. This is mostly an overcautious cleanup, but might help to prevent data loss if the backend temporarily returns broken packs.
- Prune aborts repacking when it finds a different blob in a pack than expected. This usually indicates that the backup has been tampered with or hardware problems on the computer running restic.
- Prune and rebuild-index fail now if an old index file cannot be removed. For rebuild-index this is problematic as that command is usually used to repair a broken index and shouldn't leave possibly invalid index files. For prune this behavior could lead to data loss, as prune would continue and remove old packs which might still be referenced by one of the old index files that were not deleted. A later backup run would the not re-upload any blobs that are listed in these index, but which might actually be missing from the repository. This could result in a snapshot that's missing data.

Most of the code in this PR implements tests and adds support for the integration tests to wrap a repository to simplify testing repository damages.

This PRs attempts to solve possible problems in the prune implementation in the hope of possibly fixing some of the reported repository damages. The changes themselves were not discussed before.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -191,12 +191,17 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 		return err
 	}
 
+	var missingBlobs []restic.BlobHandle
 	for h := range usedBlobs {
 		if _, ok := blobCount[h]; !ok {
-			return errors.Fatalf("At least one data blob seems to be missing, aborting prune to prevent further data loss!\n" +
-				"Please report this error (along with the output of the 'prune' run) at\n" +
-				"https://github.com/restic/restic/issues/new")
+			missingBlobs = append(missingBlobs, h)
 		}
+	}
+	if len(missingBlobs) > 0 {
+		return errors.Fatalf("%v not found in the new index\n"+
+			"Data blobs seem to be missing, aborting prune to prevent further data loss!\n"+
+			"Please report this error (along with the output of the 'prune' run) at\n"+
+			"https://github.com/restic/restic/issues/new/choose", missingBlobs)
 	}
 
 	Verbosef("found %d of %d data blobs still in use, removing %d blobs\n",
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -191,10 +191,12 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 		return err
 	}
 
-	if len(usedBlobs) > stats.blobs {
-		return errors.Fatalf("number of used blobs is larger than number of available blobs!\n" +
-			"Please report this error (along with the output of the 'prune' run) at\n" +
-			"https://github.com/restic/restic/issues/new")
+	for h := range usedBlobs {
+		if _, ok := blobCount[h]; !ok {
+			return errors.Fatalf("At least one data blob seems to be missing, aborting prune to prevent further data loss!\n" +
+				"Please report this error (along with the output of the 'prune' run) at\n" +
+				"https://github.com/restic/restic/issues/new")
+		}
 	}
 
 	Verbosef("found %d of %d data blobs still in use, removing %d blobs\n",
--- a/cmd/restic/cmd_rebuild_index.go
+++ b/cmd/restic/cmd_rebuild_index.go
@@ -92,7 +92,10 @@ func rebuildIndex(ctx context.Context, repo restic.Repository, ignorePacks resti
 	Verbosef("saved new indexes as %v\n", ids)
 
 	Verbosef("remove %d old index files\n", len(supersedes))
-	DeleteFiles(globalOptions, repo, restic.NewIDSet(supersedes...), restic.IndexFile)
+	err = DeleteFilesChecked(globalOptions, repo, restic.NewIDSet(supersedes...), restic.IndexFile)
+	if err != nil {
+		return errors.Fatalf("unable to remove an old index: %v\n", err)
+	}
 
 	return nil
 }
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -1132,6 +1132,37 @@ func TestRebuildIndexAlwaysFull(t *testing.T) {
 	TestRebuildIndex(t)
 }
 
+type appendOnlyBackend struct {
+	restic.Backend
+}
+
+// called via repo.Backend().Remove()
+func (b *appendOnlyBackend) Remove(ctx context.Context, h restic.Handle) error {
+	return errors.Errorf("Failed to remove %v", h)
+}
+
+func TestRebuildIndexFailsOnAppendOnly(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	datafile := filepath.Join("..", "..", "internal", "checker", "testdata", "duplicate-packs-in-index-test-repo.tar.gz")
+	rtest.SetupTarTestFixture(t, env.base, datafile)
+
+	globalOptions.stdout = ioutil.Discard
+	defer func() {
+		globalOptions.stdout = os.Stdout
+	}()
+
+	env.gopts.backendTestHook = func(r restic.Backend) (restic.Backend, error) {
+		return &appendOnlyBackend{r}, nil
+	}
+	err := runRebuildIndex(env.gopts)
+	if err == nil {
+		t.Error("expected rebuildIndex to fail")
+	}
+	t.Log(err)
+}
+
 func TestCheckRestoreNoLock(t *testing.T) {
 	env, cleanup := withTestEnvironment(t)
 	defer cleanup()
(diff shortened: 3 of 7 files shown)


cmd/restic/cmd_prune.go
```diff
@@ -207,10 +207,12 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
 	}
 	bar.Done()
 
-	if len(usedBlobs) > stats.blobs {
-		return errors.Fatalf("number of used blobs is larger than number of available blobs!\n" +
-			"Please report this error (along with the output of the 'prune' run) at\n" +
-			"https://github.com/restic/restic/issues/new")
+	for h := range usedBlobs {
+		if _, ok := blobCount[h]; !ok {
+			return errors.Fatalf("At least one data blob seems to be missing, aborting prune to prevent further data loss!\n" +
```
rawtaz: Is it possible and useful to be more specific here, e.g. including the (shortened) blob ID, to ease troubleshooting?

MichaelEischer: I've added the ID of the first missing blob. It would also be possible to dump a possibly gigantic list with all missing blobs. Not sure whether that's a good idea.

rawtaz:
@aawsome and @greatroar could you guys review this to make sure it looks sane and shouldn't carry any unexpected results?

internal/repository/repack.go
```diff
@@ -80,7 +77,7 @@ func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, kee
 			if !id.Equal(entry.ID) {
 				debug.Log("read blob %v/%v from %v: wrong data returned, hash is %v",
 					h.Type, h.ID, tempfile.Name(), id)
-				fmt.Fprintf(os.Stderr, "read blob %v from %v: wrong data returned, hash is %v",
+				return nil, errors.Errorf("read blob %v from %v: wrong data returned, hash is %v",
```
aawsome, line 80: I see that there are error cases (like a magnetic hard disk starting to fail), where you would like to immediately stop at this point and start a forensic analysis and/or try to move all of you repository as fast as possible to some save place.

But, again, we leave the repository in a state where some new pack files + index files might have been written. This could cause problems like full disks or memory issues. Actually I think in the long run we need some "cleanup" within `prune` that tries to remove the newly generated repo files in case of an error.

For now, I agree that this improves safety.

aawsome (commented):
Most of the changes are due to using the `restic.Repository` interface in all places and added checks.

IMO this PR doesn't fully solve the index issue and #2839 should be considered instead - but merging this PR will improve the current situation.

Also this PR needs a rebase and modifications after #2773 has been merged (e.g. new `SaveBlob` syntax)

cmd/restic/integration_test.go
```diff
@@ -1023,6 +1023,46 @@ func TestRebuildIndexAlwaysFull(t *testing.T) {
...
+	restic.Backend
+}
+
+func (b *appendOnlyBackend) Remove(ctx context.Context, h restic.Handle) error {
+	return errors.Errorf("Failed to remove %v", h)
+}
+
+type appendOnlyRepository struct {
+	restic.Repository
+}
```
greatroar, line 1036: Bikeshedding maybe, but I'm kinda meh on the use of the [preemptive interfaces](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a) restic.Repository and restic.Cache (see also #2616). I know the former actually gets multiple implementations now, but they still make the code harder to navigate than it could be.

I think the appendOnlyBackend could also be installed inside the existing OpenRepository without the appendOnlyRepository. Just wrap the backend right after it's been constructed. Then the code in global.go can continue to use *repository.Repository and the changes are much more localized.

cmd/restic/integration_test.go
```diff
@@ -1086,6 +1126,61 @@ func TestPrune(t *testing.T) {
 	testRunCheck(t, env.gopts)
 }
 
+type damagedRepository struct {
+	restic.Repository
+}
```
greatroar, line 1131: Similarly, I suppose this can be replaced by a damagedBackend.

cmd/restic/integration_test.go
```diff
@@ -1086,6 +1126,61 @@ func TestPrune(t *testing.T) {
...
+	firstSnapshot := testRunList(t, "snapshots", env.gopts)
+	rtest.Assert(t, len(firstSnapshot) == 1,
+		"expected one snapshot, got %v", firstSnapshot)
+	testRunForget(t, env.gopts, firstSnapshot[0].String())
+
+	env.gopts.repositoryTestHook = func(r restic.Repository) (restic.Repository, error) {
+		return &damagedRepository{r}, nil
+	}
+
+	// create new snapshot, but loose all data
```
greatroar, line 1166: lose

MichaelEischer: Changed.

cmd/restic/cmd_prune.go
```diff
@@ -207,10 +207,13 @@ func pruneRepository(gopts GlobalOptions, repo restic.Repository) error {
...
-		return errors.Fatalf("number of used blobs is larger than number of available blobs!\n" +
-			"Please report this error (along with the output of the 'prune' run) at\n" +
-			"https://github.com/restic/restic/issues/new")
+	for h := range usedBlobs {
+		if _, ok := blobCount[h]; !ok {
+			return errors.Fatalf("%v not found in the new index\n"+
+				"At least one data blob seems to be missing, aborting prune to prevent further data loss!\n"+
+				"Please report this error (along with the output of the 'prune' run) at\n"+
+				"https://github.com/restic/restic/issues/new", h)
+		}
```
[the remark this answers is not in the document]
MichaelEischer: I've changed this to report all missing blobs.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -92,7 +92,7 @@ func rebuildIndex(ctx context.Context, repo restic.Repository, ignorePacks resti
 			Type: restic.IndexFile,
 			Name: id.String(),
 		}); err != nil {
-			Warnf("error removing old index %v: %v\n", id.Str(), err)
+			return errors.Fatalf("unable to remove old index %v: %v\n", id.Str(), err)
```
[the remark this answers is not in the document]
MichaelEischer: As commented in #2839 I think using the `superseded` field of indexes will cause more problems than it solves.

The issue with too many indexes should be rather easy to solve by removing duplicate entry while merging the index entries in memory.

internal/repository/repack.go
MichaelEischer: Hard disk errors would already be caught by the `repo.Key().Open` call a few lines earlier. This check verifies that the blob we've just read is actually the expected one. There are two possible causes for this: Either the blob was damaged before encrypting (e.g. a bit-flip in memory) or someone replaced the encrypted blob bytes with the bytes of another encrypted blob of the same size. Both scenarios indicate a potentially catastrophic repository damage.

The current error handling strategy in prune is to just stop whenever some repository damage is encountered. Changing this strategy is completely out of scope for this PR.

The full disk problematic is not affected by this error check. The other error checks in will cause `repack` to abort when it encounters a damaged blob.

cmd/restic/integration_test.go
```diff
@@ -1023,6 +1023,46 @@ func TestRebuildIndexAlwaysFull(t *testing.T) {
...
+	return errors.Errorf("Failed to remove %v", h)
+}
+
+type appendOnlyRepository struct {
+	restic.Repository
+}
+
+func (r *appendOnlyRepository) Backend() restic.Backend {
+	return &appendOnlyBackend{
+		Backend: r.Repository.Backend(),
```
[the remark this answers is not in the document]
MichaelEischer: Is added.

cmd/restic/cmd_rebuild_index.go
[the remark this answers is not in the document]
aawsome: I agree that merging this PR at least improves the open data loss problem a lot. More about superseded index files should be discussed in #2839

internal/repository/repack.go
aawsome: :+1:

aawsome (approved):
LGTM

cmd/restic/integration_test.go
MichaelEischer: I've replace the repository hook with a backend hook. This removes the code churn which modified everything to use `restic.Repository`.

cmd/restic/integration_test.go
MichaelEischer: This test now just deletes the file from the backend similar to the test for backup self-healing. That way wrapping the repository or the backend is no longer necessary.

fd0 (approved):
Awesome PR, thanks for taking the time to split this into small chunks!

PR #2686, opened 2020-04-06
cmd/restic: Fix Dropped Error

alrs:
This bug reads as if `err` was supposed to be compared to `nil`, instead of itself.


MichaelEischer:
#2630 also contains a commit "dump: Fix error check" that also fixes this error

alrs:
Thanks, I'll close this.

Issue #402, opened 2016-01-30
Canceling password prompt in linux console keeps further input hidden


wasamasa:
Repro:

- `restic -r /path/to/repo init`
- `Ctrl-c`

Expected result: After cancelling the password prompt I can type shell commands normally

Actual result: My input is hidden and requires executing `reset` blindly to restore

Version: Built yesterday from Git, running on Arch Linux.

fd0:
That's strange, this works for me. What terminal emulator are you using? How did you built restic?

wasamasa:
This happens in the linux console, aka the black thing you see when pressing `Ctrl-Alt-F2`.

I've built it with [the restic-git PKGBUILD](https://aur.archlinux.org/packages/restic-git/) from the AUR.

fd0:
could you be a bit more specific? there are different implementations, could you start a few terminals and then use `ps aux` to find out which process is running?

barsanuphe:
I can confirm this happens for me as well, in Archlinux, in a yakuake/konsole terminal.

wasamasa:
The console was started by `/sbin/agetty --noclear tty2 linux` and runs bash as shell.

fd0:
thanks for investigating

viric:
The same happens to me, in xterm.

viric:
It's not about a process being alive. It's about letting the terminal with echo disabled. I guess that on SIGINT restic does not reprogram the echo on the terminal.

PR #2689, opened 2020-04-12
Fix shutdown hang when restic is started as background job

### LINKED ISSUE #2298: Restic mount hangs in the background after mount/unmount
Output of `restic version`
--------------------------

```
# export RESTIC_PASSWORD=xxx
# restic -r /mnt/backup/restic/root mount /mnt/temp/ &
repository 2fc17b90 opened successfully, password is correct
Now serving the repository at /mnt/temp/
When finished, quit with Ctrl-c or umount the mountpoint.
```

CIFS network mount or local ZFS dataset (both exhibit the issue)

Expected behavior
-----------------

The command mounts the repo at /mnt/temp and at unmount, exits. The repo is fine, no problems accessing snapshots under the mount, etc. Note that the command is run in the background using `&`.

Actual behavior
---------------

When I unmount the repo using `umount /mnt/temp`, it's succesfully unmounted, but restic doesn't quit, it stays as a hung process in the background.

See the command issued for mounting.

Nope.

Nope.

Don't know yet, as I've just started using it for daily workloads.

MichaelEischer:
restic uses a cleanup hook to ensure that it restores the terminal configuration to a sane state, when restic is interrupted while reading a password from the terminal. However, this causes a problem, when restic runs in a background job, as reconfiguring a terminal will cause a `SIGTTOU` to be sent to restic pausing it. Therefore, restic seems to hang on shutdown when it was running in the background.

This commit changes the behavior to only restore the terminal configuration if restic was interrupted while reading a password from the terminal. As reading a password from the terminal requires that restic is in the foreground, this should avoid restic getting stopped.

See #2298 for further details on the problem. The background hang could also be addressed by another solution: `readPasswordTerminal` could temporarily register a cleanup hook and remove it afterwards. That would yield a slightly cleaner implementation, but would require much more extensive changes.

This change will result in a merge conflict with #2593, however, merging shouldn't be too complex as that PR just moves the code lines around which are modified by this PR.

The solution implemented by the PR was not discussed yet.

Closes #2298
Issue introduced in #402

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ I've no clue how a proper test for that would look like.
- ~~[ ] I have added documentation for the changes (in the manual)~~ No new feature.

### DIFF
--- a/changelog/unreleased/issue-2298
+++ b/changelog/unreleased/issue-2298
@@ -1,0 +1,8 @@
+Bugfix: Do not hang when run as a background job
+
+Restic did hang on exit while restoring the terminal configuration when it was
+started as a background job, for example using `restic ... &`. This has been
+fixed by only restoring the terminal configuration when restic is interrupted
+while reading a password from the terminal.
+
+https://github.com/restic/restic/issues/2298
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -85,6 +85,8 @@ var globalOptions = GlobalOptions{
 	stderr: os.Stderr,
 }
 
+var isReadingPassword bool
+
 func init() {
 	var cancel context.CancelFunc
 	globalOptions.ctx, cancel = context.WithCancel(context.Background())
@@ -146,7 +148,10 @@ func stdoutTerminalWidth() int {
 }
 
 // restoreTerminal installs a cleanup handler that restores the previous
-// terminal state on exit.
+// terminal state on exit. This handler is only intended to restore the
+// terminal configuration if restic exits after receiving a signal. A regular
+// program execution must revert changes to the terminal configuration itself.
+// The terminal configuration is only restored while reading a password.
 func restoreTerminal() {
 	if !stdoutIsTerminal() {
 		return
@@ -160,9 +165,17 @@ func restoreTerminal() {
 	}
 
 	AddCleanupHandler(func() error {
+		// Restoring the terminal configuration while restic runs in the
+		// background, causes restic to get stopped on unix systems with
+		// a SIGTTOU signal. Thus only restore the terminal settings if
+		// they might have been modified, which is the case while reading
+		// a password.
+		if !isReadingPassword {
+			return nil
+		}
 		err := checkErrno(terminal.Restore(fd, state))
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "unable to get restore terminal state: %#+v\n", err)
+			fmt.Fprintf(os.Stderr, "unable to restore terminal state: %v\n", err)
 		}
 		return err
 	})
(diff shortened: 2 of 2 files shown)


cmd/restic/global.go
```diff
@@ -160,9 +162,14 @@ func restoreTerminal() {
 	}
 
 	AddCleanupHandler(func() error {
+		// only restore the terminal settings when they were modified
```
rawtaz, line 165: I find this comment hard to understand easily. It talks about terminal settings being modified, but the flag that's checked is about whether or not password reading is active. They seem like two different things. For readability, can this comment be rephrased?

MichaelEischer: I've adapted the comment to better match the following check. Right now the comment and the check afterwards yield the same result, but I agree that it was a bit confusing.

changelog/unreleased/issue-2298
```diff
@@ -0,0 +1,10 @@
+Bugfix: Do not hang when run as a background job
+
+Restic did hang on exit when it was started as a background jobs, for example
```
rawtaz, line 3: s/jobs/job/

changelog/unreleased/issue-2298
```diff
@@ -0,0 +1,10 @@
...
+using `restic ... &`. restic was stopped when trying to restore the terminal
```
rawtaz, line 4: Big R after the dot.

changelog/unreleased/issue-2298
MichaelEischer: Fixed

changelog/unreleased/issue-2298
MichaelEischer: Thanks for proof reading!

changelog/unreleased/issue-2298
```diff
@@ -0,0 +1,10 @@
+Bugfix: Do not hang when run as a background job
+
+Restic did hang on exit when it was started as a background job, for example
+using `restic ... &`. Restic was stopped when trying to restore the terminal
+configuration on exit to ensure that reading a password from the terminal
+cannot leave the user with a misconfigured terminal. We have changed the
+behavior to only restore the terminal configuration when restic is interrupted
+while reading a password from the terminal.
```
rawtaz, line 8: With the risk of being picky, I think this is too technical for the average user. If they want to deep dive they can check the discussions here on GitHub. What do you think about:

Restic used to hang on exit when it was started as a background job, for example
using `restic ... &`. This has been fixed by now only trying to restore the terminal
when restic is being interrupted while reading a password from the terminal.

MichaelEischer: My changelog text felt too long while writing it, I think its in fact better to leave the technical details. I've changed it to your proposal and added a remark that the hang was caused by the terminal reconfiguration. Otherwise its not really clear how the two sentences relate to each other.

cmd/restic/global.go
```diff
@@ -160,9 +162,14 @@ func restoreTerminal() {
 	}
 
 	AddCleanupHandler(func() error {
+		// only restore the terminal settings when reading a password
+		if !isReadingPassword {
```
rawtaz, line 166: Can this be expanded to give a short explanation as to why we need to do this?

One should be able to read the code and understand why this is done, and it's not really clear now.

changelog/unreleased/issue-2298
rawtaz: I'm not seeing the remark, where did you add that?

MichaelEischer: I've sneaked it into the first sentence: `hang on exit while restoring the terminal configuration when`.

rawtaz (approved):

PR #2692, opened 2020-04-13
archiver: Fix race condition triggered by TestArchiverAbortEarlyOnError

MichaelEischer:
Fix the test failures of the TestArchiverAbortEarlyOnError test case.

The Save methods of the BlobSaver, FileSaver and TreeSaver return early on when the archiver is stopped due to an error. For that they select on both the tomb.Dying() and context.Done() channels, which can lead to a race condition when the tomb is killed due to an error: The tomb first closes its Dying channel before canceling all child contexts. Archiver.SaveDir only aborts its execution once the context was canceled. When the tomb killing is paused between closing its Dying channel and canceling the child contexts, this lets the FileSaver/TreeSaver.Save methods return immediately, however, ScanDir still reads further files causing the test case to fail.

As a killed tomb always cancels all child contexts and as the Savers always use a context bound to the tomb, it is sufficient to just use context.Done() as escape hatch in the Save functions. This fixes the mismatch between SaveDir and Save.

Adjust the tests to use contexts bound to the tomb for all interactions with the Savers.

No.

Checklist
---------

### DIFF
--- a/internal/archiver/blob_saver.go
+++ b/internal/archiver/blob_saver.go
@@ -22,8 +22,7 @@ type BlobSaver struct {
 	m          sync.Mutex
 	knownBlobs restic.BlobSet
 
-	ch   chan<- saveBlobJob
-	done <-chan struct{}
+	ch chan<- saveBlobJob
 }
 
 // NewBlobSaver returns a new blob. A worker pool is started, it is stopped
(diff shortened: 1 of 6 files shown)


fd0 (approved):
LGTM!

Issue #2727, opened 2020-05-11
restic for AIX 7.2


Mikesch-mp:
What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

i need an agent for aix 7.2 on ppc63 platform, cross compiling leads to an error.

What are you trying to do?
--------------------------

i tried to cross compile for aix/ppc64 and run into errors

```
[root@guacamole restic]# go run build.go --goos aix --goarch ppc64
# bazil.org/fuse
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/error_std.go:27:20: undefined: errNoXattr
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:171:12: undefined: mount
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:238:17: undefined: maxWrite
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:409:32: undefined: maxWrite
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:649:11: undefined: setattrIn
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:801:12: undefined: openFlags
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:819:18: undefined: openFlags
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:1357:26: undefined: attr
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse_kernel.go:404:17: undefined: attr
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse_kernel.go:430:16: undefined: attr
../go/pkg/mod/bazil.org/fuse@v0.0.0-20191225072544-27e78e7d88df/fuse.go:819:18: too many errors
# github.com/restic/restic/internal/restic
internal/restic/node.go:188:12: undefined: lchown
internal/restic/node.go:243:14: node.restoreSymlinkTimestamps undefined (type Node has no field or method restoreSymlinkTimestamps)
internal/restic/node.go:313:9: undefined: mknod
[...]

PR #2730, opened 2020-05-12
Simplify build tags for restic mount

greatroar:

restic mount can only be built on Darwin, FreeBSD and Linux (and if we upgrade bazil.org/fuse, only FreeBSD and Linux: https://github.com/bazil/fuse/issues/224).

Listing the few supported operating systems explicitly here makes porting restic to new platforms easier.

No, but #2727 shows that the number of build errors for a new platform (AIX) is longer than it should be.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -1,7 +1,4 @@
-// +build !netbsd
-// +build !openbsd
-// +build !solaris
-// +build !windows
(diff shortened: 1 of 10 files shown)


MichaelEischer (approved):
LGTM. According to `fuse.go` in `bazil.org/fuse` only darwin, freebsd and linux are/were supported.

PR #2783, opened 2020-06-13
add changelog for PR #2709

aawsome:

add a changelog for #2709 
IMO this changes deserves a changelog :smile:

### DIFF
--- a/changelog/unreleased/pull-2709
+++ b/changelog/unreleased/pull-2709
@@ -1,0 +1,6 @@
+Enhancement: Use optimized SHA-256
+
+We now use Minio's optimized version of SHA-256 (https://github.com/minio/sha256-simd/).
+This increases SHA-256 performance for many recent Intel, AMD or ARM CPUs.
+
+https://github.com/restic/restic/pull/2709


MichaelEischer:
@aawsome You've beaten me to the punch. I've also put together a changelog entry for PR #2709 (I didn't notice yours before), see https://github.com/MichaelEischer/restic/commits/minio-sha256-changelog . I didn't open a PR so far as that would look rather strange in the PR list.

As the PR also has an associated issue, the changelog entry should use that as filename. The usage of SHA-256 is more an implementation detail, not something a user normally needs to know (https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches Point 7, has a rather useful rule of thumb). So I don't think it makes a good title.

Actually the performance optimizations also help with some rather old CPUs. And even though the Minio library advertises optimizations for ARM, the `init` function seems to always use the go standard library implementation which already uses the hardware extensions (I've also tested this on my smartphone).

rawtaz:
I looked at the two changelog PRs and decided to use the one by @MichaelEischer (after some adjustments). I'll close this one, but thanks anyway @aawsome!

PR #2788, opened 2020-06-14
Reimplement LookupBlobSize without loop + allocation

greatroar:

Repository.LookupBlobSize calls Index.LookupSize, which calls Index.Lookup. That method constructs a list of all known locations for blob in a newly allocated slice, using an implicit loop in the copy function. The slice is then ignored except for the first element.

This PR changes LookupSize to just do its own lookup and return the first hit. This should speed up restic mount. The restorer can now also use LookupBlobSize instead of its own reimplementation of the size lookup, decoupling it from crypto.

This should fix any performance regression caused by #2787.

Checklist
---------


greatroar:
Reviewers: please check #2789 first.

MichaelEischer:
Closing this as the change was merged as part of #2812.

PR #2790, opened 2020-06-14
Fix quadratic file reading in restic mount

greatroar:

Doing a single read on a file in a restic mount takes time linear in the size of the file, since each read does a linear scan through an array of blob sizes, then another linear scan through an array of cached blob contents. That means that reading a whole file takes quadratic time O(n²/B), with B the block size.

This patch changes the linear scans to a binary search and a constant time operation. I measured the effect by backing up a 100GiB file of zeros (204800 blobs), then running cat inside a mount. The result is a more than 2× speedup. Before:

```
$ /usr/bin/time cat snapshots/latest/stdin >/dev/null 
2.45user 228.15system 1:07:33elapsed 5%CPU (0avgtext+0avgdata 2104maxresident)k
8inputs+0outputs (1major+132minor)pagefaults 0swaps
```

After:

```bash
$ /usr/bin/time cat snapshots/latest/stdin >/dev/null 
1.07user 138.08system 26:04.32elapsed 8%CPU (0avgtext+0avgdata 1960maxresident)k
0inputs+0outputs (0major+133minor)pagefaults 0swaps
```

That's half an hour wasted scanning arrays. Obviously, this gets much worse at terabyte scale.

The cache semantics have changed, but IMHO the old semantics didn't make much sense. Reading blob *i* would flush blobs 0 through *i*-1 from the cache, so reading backwards would cause the whole file to get cached in RAM without any eviction, while reading forwards would only cache the last block, with a linear-time eviction every time.

Existing tests pass; they fail when I introduce off-by-one errors in the new block skipping code.

#2587 optimizes reading performance, but misses the quadratic time issue.

Checklist
---------

### DIFF
--- a/internal/fuse/blobcache.go
+++ b/internal/fuse/blobcache.go
@@ -1,0 +1,87 @@
+package fuse
+
+import (
+	"sync"
+
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/restic"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+)
+
+// Crude estimate of the overhead per blob: a SHA-256, a linked list node
+// and some pointers. See comment in blobCache.add.
+const cacheOverhead = len(restic.ID{}) + 64
+
+// A blobCache is a fixed-size cache of blob contents.
+// It is safe for concurrent access.
+type blobCache struct {
+	mu sync.Mutex
+	c  *simplelru.LRU
+
+	free, size int // Current and max capacity, in bytes.
+}
+
+// Construct a blob cache that stores at most size bytes worth of blobs.
+func newBlobCache(size int) *blobCache {
+	c := &blobCache{
+		free: size,
+		size: size,
+	}
+
+	// NewLRU wants us to specify some max. number of entries, else it errors.
+	// The actual maximum will be smaller than size/cacheOverhead, because we
+	// evict entries (RemoveOldest in add) to maintain our size bound.
+	maxEntries := size / cacheOverhead
+	lru, err := simplelru.NewLRU(maxEntries, c.evict)
+	if err != nil {
+		panic(err) // Can only be maxEntries <= 0.
+	}
+	c.c = lru
+
+	return c
+}
+
+func (c *blobCache) add(id restic.ID, blob []byte) {
+	debug.Log("blobCache: add %v", id)
+
+	size := len(blob) + cacheOverhead
+	if size > c.size {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var key interface{} = id
+
+	if c.c.Contains(key) { // Doesn't update the recency list.
+		return
+	}
+
+	// This loop takes at most min(maxEntries, maxchunksize/cacheOverhead)
+	// iterations.
+	for size > c.free {
+		c.c.RemoveOldest()
+	}
+
+	c.c.Add(key, blob)
+	c.free -= size
+}
+
+func (c *blobCache) get(id restic.ID) ([]byte, bool) {
+	c.mu.Lock()
+	value, ok := c.c.Get(id)
+	c.mu.Unlock()
+
+	debug.Log("blobCache: get %v, hit %v", id, ok)
+
+	blob, ok := value.([]byte)
+	return blob, ok
+}
+
+func (c *blobCache) evict(key, value interface{}) {
+	blob := value.([]byte)
+	debug.Log("blobCache: evict %v, %d bytes", key, len(blob))
+	c.free += len(blob) + cacheOverhead
+}
--- a/internal/fuse/file.go
+++ b/internal/fuse/file.go
@@ -3,6 +3,8 @@
 package fuse
 
 import (
+	"sort"
+
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 
@@ -18,21 +20,26 @@ const blockSize = 512
 
 // Statically ensure that *file implements the given interface
 var _ = fs.HandleReader(&file{})
-var _ = fs.HandleReleaser(&file{})
 
 type file struct {
 	root  *Root
 	node  *restic.Node
 	inode uint64
 
-	sizes []int
-	blobs [][]byte
+	// cumsize[i] holds the cumulative size of blobs[:i].
+	cumsize []uint64
+
+	// Cached blob and its index in the blobs of node.
+	cached struct {
+		blob  []byte
+		index int
+	}
 }
 
 func newFile(ctx context.Context, root *Root, inode uint64, node *restic.Node) (fusefile *file, err error) {
 	debug.Log("create new file for %v with %d blobs", node.Name, len(node.Content))
 	var bytes uint64
-	sizes := make([]int, len(node.Content))
+	cumsize := make([]uint64, 1+len(node.Content))
 	for i, id := range node.Content {
 		size, ok := root.blobSizeCache.Lookup(id)
 		if !ok {
@@ -43,8 +50,8 @@ func newFile(ctx context.Context, root *Root, inode uint64, node *restic.Node) (
 			}
 		}
 
-		sizes[i] = int(size)
 		bytes += uint64(size)
+		cumsize[i+1] = bytes
 	}
 
 	if bytes != node.Size {
@@ -56,8 +63,8 @@ func newFile(ctx context.Context, root *Root, inode uint64, node *restic.Node) (
 		inode: inode,
 		root:  root,
 		node:  node,
-		sizes: sizes,
-		blobs: make([][]byte, len(node.Content)),
+
+		cumsize: cumsize,
 	}, nil
 }
 
@@ -84,13 +91,8 @@ func (f *file) Attr(ctx context.Context, a *fuse.Attr) error {
 
 func (f *file) getBlobAt(ctx context.Context, i int) (blob []byte, err error) {
 	debug.Log("getBlobAt(%v, %v)", f.node.Name, i)
-	if f.blobs[i] != nil {
-		return f.blobs[i], nil
-	}
-
-	// release earlier blobs
-	for j := 0; j < i; j++ {
-		f.blobs[j] = nil
+	if i == f.cached.index && f.cached.blob != nil {
+		return f.cached.blob, nil
 	}
 
 	blob, err = f.root.repo.LoadBlob(ctx, restic.DataBlob, f.node.Content[i], nil)
@@ -98,16 +100,16 @@ func (f *file) getBlobAt(ctx context.Context, i int) (blob []byte, err error) {
 		debug.Log("LoadBlob(%v, %v) failed: %v", f.node.Name, f.node.Content[i], err)
 		return nil, err
 	}
-	f.blobs[i] = blob
+	f.cached.blob, f.cached.index = blob, i
 
 	return blob, nil
 }
 
 func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadResponse) error {
 	debug.Log("Read(%v, %v, %v), file size %v", f.node.Name, req.Size, req.Offset, f.node.Size)
-	offset := req.Offset
+	offset := uint64(req.Offset)
 
-	if uint64(offset) > f.node.Size {
+	if offset > f.node.Size {
 		debug.Log("Read(%v): offset is greater than file size: %v > %v",
 			f.node.Name, req.Offset, f.node.Size)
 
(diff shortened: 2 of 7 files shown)


internal/fuse/file.go
```diff
@@ -84,30 +91,25 @@ func (f *file) Attr(ctx context.Context, a *fuse.Attr) error {
 
 func (f *file) getBlobAt(ctx context.Context, i int) (blob []byte, err error) {
 	debug.Log("getBlobAt(%v, %v)", f.node.Name, i)
-	if f.blobs[i] != nil {
-		return f.blobs[i], nil
-	}
-
-	// release earlier blobs
```
MichaelEischer, line 91: This loop has the interesting property that it doesn't release blobs when accessing files backwards. The main benefit here is that if a program keeps accessing files chunks before and after a blob boundary, then after one or two accesses both used blobs are kept in the cache.

It's not really optimal and I think a simple LRU cache could do a lot better.

internal/fuse/file.go
```diff
@@ -123,16 +125,15 @@ func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadR
 	}
 
 	// Skip blobs before the offset
-	startContent := 0
-	for offset > int64(f.sizes[startContent]) {
-		offset -= int64(f.sizes[startContent])
-		startContent++
-	}
+	startContent := -1 + sort.Search(len(f.cumsize), func(i int) bool {
```
MichaelEischer, line 128: My assumption would be that consecutive read calls usually access the same blob. In that case we could avoid the binary search altogether (starting from the second call) and cache the last accessed `startContent`.

internal/fuse/file.go
```diff
@@ -154,13 +155,6 @@ func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadR
 	return nil
 }
 
-func (f *file) Release(ctx context.Context, req *fuse.ReleaseRequest) error {
```
MichaelEischer, line 157: Shouldn't this clear the `cached` field? Is it that the fuse library drops the `file` struct anyways afterwards such that this isn't necessary?

MichaelEischer (commented):
I like the usage of a binary search. It could be possible to speed this up even further by caching the result of the search.

The new cache implementation looses one useful property of the old one: If file blocks are read before and after a blob boundary then the old implementation will end up with caching both blobs. As blobs are always read from the repository, this could end up in a large slowdown. (I think pdf files trigger rather erratic disk access patterns).

The proper solution here would be something like e.g. a global LRU cache. That could store the last 50-100 MB of accessed blobs.

internal/fuse/file.go
greatroar: The struct is dropped. Release is for closing files and network connections.

internal/fuse/file.go
greatroar: That's a good point. I'll install an LRU cache.

internal/fuse/file.go
greatroar: Perhaps, but I'm not convinced that this is the expensive part of the method. At 10TB and blocks of 1.5MB on average, binary search does 23 comparisons and incurs slightly fewer L1 cache misses. I can try to whip up a benchmark, but after I do the LRU thing.

internal/fuse/file.go
greatroar: Actually, when the same part of a file is accessed multiple times, shouldn't the kernel cache that?

greatroar:
I added an LRU cache. Regardless of whether the kernel caches FUSE blocks, it can still be useful when comparing files across snapshots.

greatroar:
Test failure is rclone, again.

internal/fuse/file.go
greatroar: I wrote a benchmark and optimized the binary search, so it is now more than twice as fast. The benchmark suggests that reading a 200GiB file incurs about 1.5s of binary search overhead on my box. The actual figure might be a bit higher due to cache effects.

Caching the last startContent brings us back to the question of synchronization in Fuse, which I would rather avoid.

greatroar:
@MichaelEischer Merge?

MichaelEischer (commented):
The PR looks mostly fine. I just think that there's still a small race condition regarding the LRU-cache usage.

internal/fuse/fuse_test.go
```diff
@@ -190,3 +227,41 @@ func testTopUidGid(t *testing.T, cfg Config, repo restic.Repository, uid, gid ui
...
+
+	f := file{cumsize: sizes}
+	b.Logf("%d bytes in %d chunks", total, len(sizes))
+
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Linear scan in 4kiB strides.
+		for offset := uint64(0); offset < total; offset += 4 * kiB {
+			f.findStart(offset)
```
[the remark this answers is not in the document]
greatroar: Added a check.

internal/fuse/file.go
```diff
@@ -84,30 +83,28 @@ func (f *file) Attr(ctx context.Context, a *fuse.Attr) error {
...
 	}
 
 	blob, err = f.root.repo.LoadBlob(ctx, restic.DataBlob, f.node.Content[i], nil)
 	if err != nil {
 		debug.Log("LoadBlob(%v, %v) failed: %v", f.node.Name, f.node.Content[i], err)
 		return nil, err
 	}
-	f.blobs[i] = blob
+
+	f.root.blobCache.add(f.node.Content[i], blob)
```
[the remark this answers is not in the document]
greatroar: Good catch. I replaced the groupcache LRU package with the hashicorp one, which has an API for checking whether a key is present without changing its recency. The worst that can happen now is that two goroutines load the blob concurrently.

internal/fuse/fuse_test.go
```diff
@@ -20,6 +21,48 @@ import (
...
+		rtest.Equals(t, &exp[0], &blob[0])
+		rtest.Equals(t, exp, blob)
+	}
+
+	addAndCheck(id1, make([]byte, 32*kiB))
+	addAndCheck(id2, make([]byte, 30*kiB))
+	addAndCheck(id3, make([]byte, 10*kiB))
+
+	_, ok := c.get(id2)
+	rtest.Assert(t, ok, "blob %v evicted", id2)
```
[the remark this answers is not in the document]
greatroar: It was correct, but I had trouble parsing it too, so I removed the double negation.

internal/fuse/file.go
```diff
@@ -123,16 +120,13 @@ func (f *file) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadR
 	}
 
 	// Skip blobs before the offset
-	startContent := 0
-	for offset > int64(f.sizes[startContent]) {
-		offset -= int64(f.sizes[startContent])
-		startContent++
-	}
+	startContent := f.findStart(offset)
```
aawsome, line 123: using `import "sort"` (and the changes mentioned above), this can be simply written as:
```
startContent := sort.Search(len(f.cumsize), func(i int) bool {return cumsize[i] > offset})
```

internal/fuse/file.go
```diff
@@ -18,21 +18,20 @@ const blockSize = 512
...
-	blobs [][]byte
+	// cumsize[i] holds the cumulative size of blobs[:i].
+	cumsize []uint64
 }
 
 func newFile(ctx context.Context, root *Root, inode uint64, node *restic.Node) (fusefile *file, err error) {
 	debug.Log("create new file for %v with %d blobs", node.Name, len(node.Content))
 	var bytes uint64
-	sizes := make([]int, len(node.Content))
+	cumsize := make([]uint64, 1+len(node.Content))
```
aawsome, line 34: the `1+` is not necessary, see below

internal/fuse/file.go
```diff
@@ -43,8 +42,8 @@ func newFile(ctx context.Context, root *Root, inode uint64, node *restic.Node) (
 			}
 		}
 
-		sizes[i] = int(size)
 		bytes += uint64(size)
+		cumsize[i+1] = bytes
```
aawsome, line 46: `cumsize[i] = bytes`, see below

aawsome (changes requested):
I would suggest not to re-implement binary search.

The other changes look good to me.

aawsome:
@greatroar just realized that you exactly kicked out "import sort" with you last commit.
Is standard sort really slowing things down that much? Doesn't the closure gets inlined?

internal/fuse/file.go
greatroar: That doesn't work (the existing test crashes), because we have to handle two corner cases: size == 0 and size == filesize. The following line is

    offset -= f.cumsize[startContent]

and sort.Search may return 0 through len(f.cumsize), inclusive. This can be handled by special casing, but that's not necessarily more complicated than overallocating and I really liked the one-line summary

    // cumsize[i] holds the cumulative size of blobs[:i].

It's subtle, but the semantics can be described without a +1 or -1 :)

internal/fuse/file.go
greatroar: See above.

greatroar:
The speed difference is more than a factor of 2, see the commit message for 5a315467772429b3a224661a5c2eb036a4c7f2aa. And it really is failure to inline:

```
$ go build -gcflags=-m |& grep -F Read.func
./file.go:125:51: can inline (*file).Read.func1
```

Can inline, but no "inlined" message. From what I can tell from the assembler, it's really the call (stack handling?) that's expensive.

Any way, the throughput difference will be very small compared to the sort.Search version. I wrote the custom binary search first, but only pushed it in response to @MichaelEischer's performance concerns, to avoid the who-locks-what-in-FUSE debate. So it comes down to readability and I'll happily revert the last commit if the maintainers want it.

aawsome:
> The speed difference is more than a factor of 2, see the commit message for [5a31546](https://github.com/restic/restic/commit/5a315467772429b3a224661a5c2eb036a4c7f2aa). And it really is failure to inline:

Also just did the comparison myself :wink: - the missing inlining explains it. However, I would vote for the "sort.Sort"-solution for easier readability.

If we do want to improve further, I would suggest to optimize for the sequential read case which still might be the most often used access pattern. That is, remember the last read blob (respectively index to Content) and check if the newly requested offset fits to this (or the next, if the last blob was completely read). If not, start the binary search which should be already fast enough for all other kinds of access.

internal/fuse/blobcache.go
```diff
@@ -0,0 +1,83 @@
...
+	free, size int // Current and max capacity, in bytes.
+}
+
+func newBlobCache(size int) *blobCache {
+	c := &blobCache{
+		free: size,
+		size: size,
+	}
+
+	maxEntries := size / cacheOverhead
```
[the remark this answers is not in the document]
greatroar: Done.

MichaelEischer:
> If we do want to improve further, I would suggest to optimize for the sequential read case which still might be the most often used access pattern. That is, remember the last read blob (respectively index to Content) and check if the newly requested offset fits to this (or the next, if the last blob was completely read). If not, start the binary search which should be already fast enough for all other kinds of access.

@greatroar That's what I had in mind for "caching the result of the search", but in my test with a 60GB file the binary search seemed to be fast enough. I actually prefer the 'sort.Search' solution to the reimplemented binary search.

I've noticed that PR now reuses the simplelru package instead of adding another dependency to restic and I think it's rather important to keep the amount of dependencies at a reasonable level.

So the only thing remaining is a small changelog entry which mentions the _optimized file access in the mount command_ . That changelog entry could later on also be extended to cover the other mount PRs.

aawsome:
> @greatroar That's what I had in mind for "caching the result of the search", but in my test with a 60GB file the binary search seemed to be fast enough. I actually prefer the 'sort.Search' solution to the reimplemented binary search.

I did tests all these three variants 

- sort.Search
- manual binary search implemented by @greatroar 
- saving the index which will be used in the sequentially following next read.

with a 10GB file containing zeros (thus being able to reuse the already-cached blob) and used a `time cat file >/dev/null`. This means the third variant always gets a hit here. 

However, there was no notable time difference and the searching part could not be seen in the cpu profile.
IMO this speaks for using the simple `sort.Search` solution.

MichaelEischer:
@aawsome I took a look at the blob lookup performance using the cpu profiler. Even for a 50GB file the old simple loop just consumed 10% in `fuse/file.Read` . As the slowdown is quadratic this means that the replaced code part probably didn't consume more than 1% of the runtime of the `Read` method, even with the inefficient implementation. So in total you'll only see a noticeable difference for 100GB+ files.

greatroar:
Ok, added a changelog entry and restored the sort-Search version

MichaelEischer (approved):
LGTM, thanks. I've rebased the branch onto the current master and force pushed it to restart the CI run, which once again had problems with the rclone test.

Issue #450, opened 2016-02-15
Reduce memory consumption with command line option


fawick:
Hi @fd0! Is just finished watching the video from your recent talk that is [linked in the blog](https://restic.github.io/blog/2016-02-04/CCC-Cologne-OpenChaos) and really came to like restic and the philosophy behind the tool and the development process.

At one point in the talk it is mentioned that restic currently allocates around 300MB of RAM for reading multiple files and is therefore not suited to run on, e.g., a Raspberry Pi. I wondered whether you would consider adding a command line option for manually reducing that number at runtime to allow for restic to be used on low-memory ARM devices. Next to the ubiquitous RasPi I was thinking of backing up Android and Sailfish devices.

fd0:
Hey, thanks for your interest in restic. The number of around 300MiB of ram that I mentioned in the talk is used for two things (mainly):
 * calculating scrypt(password): the constants for scrypt are hard-coded right now, and there's #17 to add code that automatically figures out how hard scrypt should be for the current system
 * having multiple buffers of blobs/packs to backup in memory: This is tightly coupled with the maximum concurrency that is allowed. For this I'd also like to have auto-tuning code, but a command line option is also likely.

To summarize my points above: It's planned and will be implemented at some point in time.

fawick:
Sounds good to me. I'd like to keep this issue open to track that feature, then.

fd0:
Agreed.

mika:
I'd be interested in this topic also from a server perspective, so not necessarily low memory but "still not enough". :) For example I've got a mail server with ~950GB of data (`scanned 64173 directories, 2147728 files in 2:41`) and with 2GB of RAM I ran into OOM with restic (running under `ionice -c 3 nice -n 19`). Are there any known numbers for calculating memory needs for restic (per file/per GB/....)? (I ran into OOM with restic in all kind of different VMs, known those needs upfront would make life easier, and if there'd be any way how to reduce memory consumption that would be great.)

Thanks!

dionorgua:
My experience is around 4GB of RAM for 1.5TB repo size (for backup). prune takes even more (~9-10GB of RAM).

Memory usage depends mostly not on data size on particular machine, but mostly on repository size. So it is impossible to backup 100KB file on machine with 2GB of RAM (no swap) if restic repository itself is ~1TB.

eliasp:
Here:
- Build:
    ```
    restic 0.8.3 (v0.8.3-11-gda77f4a2)
[...]

PR #2794, opened 2020-06-17
Add index options / make restic consume less memory for index

aawsome:

Add index options to make restic consume less memory in different situations
The index option applies when reading the index from index files into memory.

The following options are added:

- IndexOptionFull:
  saves all, identical to current implemenation
- IndexOptionDataIDsOnly
  saves all except that for data blobs only the ID is saved.
  This implies that some index operations (e.g. `Lookup()` for data blobs) are
  no longer available.
  Can be used e.g. within `restic backup`
- IndexOptionNoData
  only save tree information in index
  Here even more index operations (e.g. `Has()` for data blobs) are no
  longer available.
  Can be used e.g. for `restic find`
- IndexOptionNone
  Do not store index in memory, only provide `Each()` by loading index files
  sequentially and process them.
  Can be used e.g. for `restic list blobs`

This work can also be further extended (maybe in a following PR) to give a `--lowmem` option to chosen restic commands which would then use one of the index options to trade performance or functionality for lower memory usage, see e.g. #450 

This is WIP and I haven't added tests yet. Happy to get early feedback!

No, but I got the idea when working on #2523 and #2786.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -495,7 +495,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	if !gopts.JSON {
 		p.V("load index files")
 	}
-	err = repo.LoadIndex(gopts.ctx)
+	err = repo.LoadIndex(gopts.ctx, restic.IndexOptionDataIDsOnly)
 	if err != nil {
 		return err
 	}
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -142,7 +142,7 @@ func runCat(gopts GlobalOptions, args []string) error {
 	}
 
 	// load index, handle all the other types
-	err = repo.LoadIndex(gopts.ctx)
+	err = repo.LoadIndex(gopts.ctx, restic.IndexOptionFull)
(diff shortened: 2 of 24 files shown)


MichaelEischer (commented):
Although I generally agree that we should try to reduce the memory usage of restic, this PR introduces more complications than it's worth in my opinion.

With the optimzed index hashmap implementation `IndexOptionDataIDsOnly` probably doesn't save any memory at all it might even regress the memory usage. `IndexOptionNone` requires special handling either way.

It would probably be more useful to implement something like a generic, parallelized `StreamIndex` function which can be used by `list`, the normal index loading and the checker.

`IndexOptionNoData` breaks resolving pack-ids in `find` and for the `stats` command several modes.

This just leaves `ls` and `recover` which would consume less memory with this PR. But with the additional complexity (especially regarding which index operation can be used at a given point in the code) I don't think its worth it. `recover` is only very rarely used and if you want to use the output of `ls` in any way then you need a command which once again has to load and keep the full index in memory.

MichaelEischer:
I've decided to close this PR, as explained previously I don't think that the small gains for the ls and recover command are worth the additional code complexity.

aawsome:
Fully agree :+1:

PR #2797, opened 2020-06-18
Fix rclone backend not properly handling subprocess exit

### LINKED ISSUE #2381: restic does not detect rclone termination
Output of `restic version`
--------------------------
```
restic 0.9.5 compiled with go1.12.4 on linux/arm
```

```
GOGC=20 nice -n 19 ionice -c 3 restic -r rclone:jotta:xxx --password-file xxx.passwd backup --one-file-system -v /mnt/xxx
```

In one case the system had 2 Gbyte of RAM, and the Linux out-of-memory (OOM) killer terminated the rclone process, and in a second case the system had 1 Gbyte of RAM and the OOM killer terminated the rclone process.

In both cases, restic just hung, doing nothing.

Jottacloud, but I saw identical behavior using the sftp backend.

Expected behavior
-----------------
restic should detect that rclone has died and exit with an error.

Actual behavior
---------------
restic sat there quietly, doing nothing.

It seems like you can get similar results just by killing the rclone process by hand.

restic ignoring SIGCHLD and/or pipe file descriptor closes.

Detect the error and exit.

restic is awesome. I am nervous to have built all of my backups around it, but so far so good! 😂👍

greatroar:

When rclone exits unexpectedly, restic can hang. The problem occurs in the REST backend, which does not detect that its connection over stdin/stdout has stopped working.

This patch is a (hacky) fix for that problem: it simply closes stdin and stdout when the process is gone. Why this doesn't produce a SIGPIPE is beyond me. The error message is ugly, but at least it arrives instantly.

I have some ideas about how to fix this properly, but they may require quite a bit of code (if they even turn out correct). I suggest merging this as a hotfix.

I was hoping to fix the flaky rclone test, but I don't think this patch does that.

Fixes #2381.

Checklist
---------

### DIFF
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -184,6 +184,7 @@ func newBackend(cfg Config, lim limiter.Limiter) (*Backend, error) {
 		debug.Log("Wait returned %v", err)
 		be.waitResult = err
 		close(waitCh)
+		stdioConn.Close()
 	}()
 
 	ctx, cancel := context.WithCancel(context.Background())
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -114,7 +114,7 @@ func wrapConn(c *StdioConn, lim limiter.Limiter) wrappedConn {
 }
 
 // New initializes a Backend and starts the process.
-func New(cfg Config, lim limiter.Limiter) (*Backend, error) {
+func newBackend(cfg Config, lim limiter.Limiter) (*Backend, error) {
 	var (
 		args []string
 		err  error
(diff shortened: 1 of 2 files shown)


greatroar:
Oh, great. Tests are failing because Minio has let a certificate expire :facepalm:

MichaelEischer:
> Why this doesn't produce a SIGPIPE is beyond me.

The [default behavior](https://golang.org/pkg/os/signal/#hdr-SIGPIPE) of Go is to trigger `SIGPIPE` for stdout and stderr and to convert the signal into an `EPIPE` error for all other file descriptors.

The root cause for restic not noticing the subprocess exit is that restic currently holds on to file descriptors for both the read and write side of the stdin and stdout pipes. That prevents the pipes from closing properly. Calling `stdioConn.Close()` once the child process has exited is still necessary to free the file descriptors.

I've built a different version of this PR (including your test case) which properly fixes the pipe setup and hopefully also the rclone test failures: #2855

greatroar:
Closing in favor of #2855.

PR #2801, opened 2020-06-20
Add flag `--read-data-sample` to `restic check`

### LINKED ISSUE #2186: Increase the effective granularity of `--read-data-subset`
Output of `restic version`
--------------------------

`restic check --read-data-subset n/m` currently only uses the first byte of pack ID to divide packs into subsets. This limits the effective granularity to at most 1/256.

I hope restic can use a few more bytes instead (e.g. 2 bytes or 4 bytes) so the effective granularity can be increased. It can also help make packs more even distributed among subsets when m ≈ 100 - 200.

What are you trying to do?
--------------------------
Do data checks of random subsets more frequently without using too much data each time.

Absolutely. Having efficient daily backups is so great. Already saved my ass multiple times.

aawsome:

Add a new flag to `restic check` to allow test a random sample of given sample size of pack files
instead of all pack files (with `--read-data`) or a very specific subset (with `--read-data-subset`)

See [forum discussion](https://forum.restic.net/t/rsync-net-how-do-you-run-repo-maintanenance/2845) where I got the idea.

closes #2186 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2186
+++ b/changelog/unreleased/issue-2186
@@ -1,0 +1,7 @@
+Enhancement: Add flag `--read-data-sample` to restich check
+
+We've added a new flag to restic check which allows to read only a random sample of data files.
+
+https://github.com/restic/restic/issues/2186
+https://github.com/restic/restic/pull/2801
+https://forum.restic.net/t/rsync-net-how-do-you-run-repo-maintanenance
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"strconv"
 	"strings"
 	"time"
@@ -43,6 +44,7 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 type CheckOptions struct {
 	ReadData       bool
 	ReadDataSubset string
+	ReadDataSample uint64
 	CheckUnused    bool
 	WithCache      bool
 }
@@ -55,6 +57,7 @@ func init() {
 	f := cmdCheck.Flags()
 	f.BoolVar(&checkOptions.ReadData, "read-data", false, "read all data blobs")
 	f.StringVar(&checkOptions.ReadDataSubset, "read-data-subset", "", "read subset n of m data packs (format: `n/m`)")
+	f.Uint64Var(&checkOptions.ReadDataSample, "read-data-sample", 0, "read a random sample of data packs with given sample length")
 	f.BoolVar(&checkOptions.CheckUnused, "check-unused", false, "find unused blobs")
 	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use the cache")
 }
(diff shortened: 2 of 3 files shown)


cfbao:
I would love something like this.
Related issue - #2186

aawsome:
@cfbao Thanks for mentioning the issue - I added it to the PR description an to the changelog.
Feel free to use this PR. Worked pretty well in my tests.

changelog/unreleased/issue-2186
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add flag `--read-data-sample` to restich check
```
buschjost, line 1:
```suggestion
Enhancement: Add flag `--read-data-sample` to restic check
```

cmd/restic/cmd_check.go
```diff
@@ -293,11 +287,45 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
...
+			// of pack, update totalBucketsMax.
+			if (uint(pack[0]) % totalBuckets) == (bucket - 1) {
+				packs.Insert(pack)
+			}
+		}
+		Verbosef(fmt.Sprintf("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, len(packs), chkr.CountPacks(), totalBuckets))
+		doReadData(packs)
+	case opts.ReadDataSample > 0:
+		rand.Seed(time.Now().UnixNano())
+		maxSkip := int(chkr.CountPacks() - opts.ReadDataSample)
```
greatroar, line 309: This algorithm is not correct. Say CountPacks() = 100 and ReadDataSample = 10, then maxSkip = 90. If the RNG returns >=29 three times and >=9 once, the actual sample size is at most 3. This happens with probability (62/91)³ × 82/91 > 28%.

To do this right, you could use [reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling).

aawsome: It will always pick a sample of the given size. The number of packs to skip is recomputed after each "skipping".

But you are right, the algorithm is not correct in the sense that it collects a true random sample - in fact packs that are "later" in the `range` will be picked with a much higher probability. As `range` for maps is already kind of random, I ignored this incorrectness :wink: 

Do you now a go package that provides this sampling function? I was looking for one and did not found any - this is why I came with this "hacky" solution...

greatroar: Oh right, I missed the maxSkip-- bit.

Looks like [go-metrics has Vitter's algorithm for reservoir sampling](https://godoc.org/github.com/rcrowley/go-metrics#UniformSample).

greatroar: I have a bunch of random number utilities that I should publish somewhere. I'll see if I can add a reservoir sampler.

greatroar: [Done](https://godoc.org/github.com/greatroar/randstat/sampling). This package could do with a streaming API, but I think the function I have know can be plugged into the current code without too much of a hassle.

aawsome:
I'm closing this as the same topic is tackled in #3038.

PR #2805, opened 2020-06-23
Fix examples of --verbose with argument

jwilk:
`--verbose` has a default value, so `--verbose 2` doesn't work; only `--verbose=2` does.

BTW, there's the same mistake in the help message. But I don't speak Go, so I'll let someone else fix it. :-)

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -72,7 +72,7 @@ type GlobalOptions struct {
 	//  0 means: don't print any messages except errors, this is used when --quiet is specified
 	//  1 is the default: print essential messages
 	//  2 means: print more messages, report minor things, this is used when --verbose is specified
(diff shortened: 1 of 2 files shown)


greatroar:
Confirmed:

```
$ restic --verbose 2 snapshots
unknown command "2" for "restic"

Did you mean this?
	ls

```

The help messages are generated by pflag. I've filed spf13/pflag#267.

MichaelEischer:
Except for an extremely horrible hack, `pflag` currently always adds a space between the flagname and the argument.
A simple, short-term workaround would be to add the proper usage of verbose to its flag description:
```
[...] or level --verbose=`n`
```

MichaelEischer (approved):
LGTM.

@jwilk I've tried to push the change to the description of the verbose parameter, but it seems like you didn't enable maintainer access for the branch.

PR #2811, opened 2020-06-27
Use low memory map for in-memory index - about 45% less memory for index

aawsome:
UPDATE: This PR is only left for reference; IMO #2812 should be followed which gives slightly worse memory results but much better handling and performance.

Use a low memory map for the in-memory index.

In fact there are a few changes which come with this PR:

- Use a separate map for treeBlobs and dataBlobs as this saves 1 byte per index entry.
- Use a complete replacement for go's map which uses much less memory and doesn't show worst-case behavior on memory usage. The tradeoff is that this map is slower, needs an extra sorting step and there are performance spikes when inserting (growing is done all at once). IMO these drawbacks are acceptable within restic.
- Remove the extra handling of duplicate blobs as this "feature" is now integrated within the map.

For more details about the low-memory map, see `internal/repository/index_inmem.go`.

Here is the comparison to master (where #2781 is already merged!) of the repository benchmarks:
```
name                                     old time/op    new time/op    delta
PackerManager-4                             288ms ± 5%     262ms ± 0%    -8.94%  (p=0.000 n=10+8)
DecodeIndex-4                              18.4µs ± 4%    23.6µs ± 2%   +28.55%  (p=0.000 n=10+10)
IndexHasUnknown-4                          61.7ns ± 8%   247.0ns ± 2%  +300.19%  (p=0.000 n=10+8)
IndexHasKnown-4                            63.3ns ± 2%   279.8ns ± 2%  +341.68%  (p=0.000 n=9+9)
IndexAlloc-4                                1.08s ± 3%     1.23s ± 2%   +14.51%  (p=0.000 n=10+10)
MasterIndexLookupSingleIndex-4              173ns ± 5%     430ns ± 1%  +148.97%  (p=0.000 n=10+9)
MasterIndexLookupMultipleIndex-4            515ns ± 8%    1855ns ± 6%  +260.19%  (p=0.000 n=9+10)
MasterIndexLookupSingleIndexUnknown-4      90.9ns ± 6%   268.4ns ± 5%  +195.24%  (p=0.000 n=10+10)
MasterIndexLookupMultipleIndexUnknown-4     435ns ± 5%    1643ns ± 5%  +277.42%  (p=0.002 n=4+10)
LoadTree-4                                  459µs ± 1%     460µs ± 2%      ~     (p=0.481 n=10+10)
LoadBlob-4                                 8.19ms ± 1%    8.18ms ± 0%      ~     (p=0.549 n=10+9)
LoadAndDecrypt-4                           10.2ms ± 4%    10.1ms ± 3%      ~     (p=0.190 n=10+10)
LoadIndex-4                                41.6ms ± 2%    42.8ms ± 3%    +2.94%  (p=0.000 n=10+10)
SaveAndEncrypt-4                            127ns ± 1%     204ns ± 2%   +60.66%  (p=0.000 n=9+10)

name                                     old speed      new speed      delta
PackerManager-4                           183MB/s ± 5%   201MB/s ± 0%    +9.74%  (p=0.000 n=10+8)
LoadBlob-4                                122MB/s ± 1%   122MB/s ± 0%      ~     (p=0.534 n=10+9)
LoadAndDecrypt-4                         97.9MB/s ± 4%  99.3MB/s ± 3%      ~     (p=0.190 n=10+10)
SaveAndEncrypt-4                         33.0TB/s ± 1%  20.6TB/s ± 1%   -37.54%  (p=0.000 n=10+9)

name                                     old alloc/op   new alloc/op   delta
PackerManager-4                            91.0kB ± 0%    91.0kB ± 0%      ~     (p=1.117 n=9+10)
IndexAlloc-4                                786MB ± 0%     430MB ± 0%   -45.28%  (p=0.000 n=10+10)
SaveAndEncrypt-4                            2.00B ± 0%     3.00B ± 0%   +50.00%  (p=0.000 n=10+10)

name                                     old allocs/op  new allocs/op  delta
PackerManager-4                             1.41k ± 0%     1.41k ± 0%      ~     (p=0.917 n=9+10)
IndexAlloc-4                                 977k ± 0%      995k ± 0%    +1.90%  (p=0.000 n=10+10)
SaveAndEncrypt-4                             0.00           0.00           ~     (all equal)
```

As expected, some timings are getting worse, but these are internal timings and I didn't see impacts with restic commands (but didn't measure these, either). Note the `-45.28%` on memory usage.

Also note, that with this PR huge index speed improvements are possible when merging the indexes, see #2799 . 

The ideas come from #2523 and #2799.

Checklist
---------

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -49,7 +49,7 @@ import (
 // Index holds lookup tables for id -> pack.
 type Index struct {
 	m         sync.Mutex
-	byType    map[restic.BlobType]*lowMemMap
+	byType    []*linkedMemMap
 	packs     restic.IDs
 	treePacks restic.IDs
 	// only used by Store, StorePacks does not check for already saved packIDs
@@ -73,12 +73,12 @@ type indexEntry struct {
 // NewIndex returns a new index.
 func NewIndex() *Index {
 	ind := Index{
-		byType:        make(map[restic.BlobType]*lowMemMap),
+		byType:        make([]*linkedMemMap, len(allBlobTypes)+1),
 		packIDToIndex: make(map[restic.ID]int),
 		created:       time.Now(),
 	}
 	for _, t := range allBlobTypes {
-		ind.byType[t] = newLowMemMap()
+		ind.byType[t] = newLinkedMemMap()
 	}
 
 	return &ind
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -84,12 +84,6 @@ func NewIndex() *Index {
 	return &ind
 }
 
-func (idx *Index) Stats() {
-	for _, t := range allBlobTypes {
-		idx.byType[t].Stats()
-	}
-}
-
 // addToPacks saves the given pack ID and return the index.
 // This procedere allows to use pack IDs which can be easily garbage collected after.
 func (idx *Index) addToPacks(id restic.ID) int {
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -48,11 +48,10 @@ import (
 
 // Index holds lookup tables for id -> pack.
 type Index struct {
-	m          sync.Mutex
-	blob       map[restic.BlobHandle]indexEntry
-	duplicates map[restic.BlobHandle][]indexEntry
-	packs      restic.IDs
-	treePacks  restic.IDs
+	m         sync.Mutex
+	byType    map[restic.BlobType]*lowMemMap
+	packs     restic.IDs
+	treePacks restic.IDs
 	// only used by Store, StorePacks does not check for already saved packIDs
 	packIDToIndex map[restic.ID]int
 
@@ -61,6 +60,8 @@ type Index struct {
 	supersedes restic.IDs
 	created    time.Time
 }
+
+var allBlobTypes []restic.BlobType = []restic.BlobType{restic.TreeBlob, restic.DataBlob}
 
 type indexEntry struct {
 	// only save index do packs; i.e. packs[packindex] yields the packID
@@ -71,25 +72,22 @@ type indexEntry struct {
 
 // NewIndex returns a new index.
 func NewIndex() *Index {
-	return &Index{
-		blob:          make(map[restic.BlobHandle]indexEntry),
-		duplicates:    make(map[restic.BlobHandle][]indexEntry),
+	ind := Index{
+		byType:        make(map[restic.BlobType]*lowMemMap),
 		packIDToIndex: make(map[restic.ID]int),
 		created:       time.Now(),
 	}
-}
-
-// withDuplicates returns the list of all entries for the given blob handle
-func (idx *Index) withDuplicates(h restic.BlobHandle, entry indexEntry) []indexEntry {
-	entries, ok := idx.duplicates[h]
-	if ok {
-		all := make([]indexEntry, len(entries)+1)
-		all[0] = entry
-		copy(all[1:], entries)
-		return all
+	for _, t := range allBlobTypes {
+		ind.byType[t] = newLowMemMap()
 	}
 
-	return []indexEntry{entry}
+	return &ind
+}
+
+func (idx *Index) Stats() {
+	for _, t := range allBlobTypes {
+		idx.byType[t].Stats()
+	}
 }
 
 // addToPacks saves the given pack ID and return the index.
@@ -111,12 +109,7 @@ func (idx *Index) store(packIndex int, blob restic.Blob) {
 		offset:    uint32(blob.Offset),
 		length:    uint32(blob.Length),
 	}
-	h := restic.BlobHandle{ID: blob.ID, Type: blob.Type}
-	if _, ok := idx.blob[h]; ok {
-		idx.duplicates[h] = append(idx.duplicates[h], newEntry)
-	} else {
-		idx.blob[h] = newEntry
-	}
+	idx.byType[blob.Type].add(mapKey(blob.ID), mapValue(newEntry))
 }
 
 // Final returns true iff the index is already written to the repository, it is
@@ -135,12 +128,9 @@ const (
 
 // IndexFull returns true iff the index is "full enough" to be saved as a preliminary index.
 var IndexFull = func(idx *Index) bool {
-	idx.m.Lock()
-	defer idx.m.Unlock()
-
 	debug.Log("checking whether index %p is full", idx)
 
-	blobs := len(idx.blob)
+	blobs := idx.Len()
 	age := time.Now().Sub(idx.created)
 
 	switch {
@@ -176,6 +166,7 @@ func (idx *Index) Store(blob restic.PackedBlob) {
 	}
 
 	idx.store(packIndex, blob.Blob)
+
 }
 
 // StorePack remembers the ids of all blobs of a given pack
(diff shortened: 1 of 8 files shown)


greatroar:
Interesting.

For comparison, I've been trying custom hash tables as well, but started with a simple chaning one with the duplicates in separate links. That also seemed to solve the memory use issue, but may come with some subtle performance issues: sequential performance was down between 8% and 31% on the benchmarks, and I could see a second core running the GC at 100%, so concurrent performance might be terrible.

internal/restic/id.go
```diff
@@ -83,6 +83,28 @@ func (id ID) Equal(other ID) bool {
 	return id == other
 }
 
+func (id ID) Less(other ID) bool {
```
greatroar, line 86: Isn't this just `bytes.Compare(id[:], other[:]) < 0`?

internal/repository/index_inmem.go
```diff
@@ -0,0 +1,322 @@
...
+type mapKey restic.ID
+type mapValue indexEntry
+
+func (key mapKey) Hash() int {
+	var x int
+	var s uint
+	id := restic.ID(key)
+	for _, b := range id[:8] {
+		x |= int(b) << s
+		s += 8
```
greatroar, line 40:
```go
func (key mapKey) Hash() uint64 { return binary.LittleEndian.Uint64(key[:8]) }
```

Or BigEndian, and cast to int if you must.

internal/repository/index_inmem.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+func (key mapKey) Equal(other mapKey) bool {
+	return restic.ID(key).Equal(restic.ID(other))
+}
+
+// --- start of generic implementation
+
+// grow when each bucket has in average 64 pages
+const GrowFactor = 64
```
greatroar, line 61: This is a confusing name. The growth factor is 2.

internal/restic/id.go
greatroar: In fact, the whole Equal/LessEqual/Equal machinery can be replaced by a single Compare function, whose implementation is trivial. That way, there's less chance of inconsistency between them.

internal/repository/index_inmem.go
aawsome: Agree. Maybe maxPagesPerBucket is a better name. :smile:

internal/repository/index_inmem.go
```diff
@@ -0,0 +1,322 @@
...
+
+// getKey searches for an elements with given key; bucket must be sorted!
+// returns the first found entry and wheter found or not
+func (b *bucket) getKey(key mapKey) (mapValue, bool) {
+	if !b.sorted {
+		panic("bucket is not sorted!")
+	}
+	if k := sort.Search(b.length, func(i int) bool {
+		return key.LessEqual(b.get(i).key)
+	}); k < b.length {
```
greatroar, line 263:
```go
k := sort.Search(// ...
})
if k < b.length {
	// ...
}
```

internal/repository/index_inmem.go
```diff
@@ -0,0 +1,322 @@
...
+const pageSize = 1 << shiftBits
+const maskBits = pageSize - 1
+
+// lowMemMap - actually behaves like map[mapKey]mapEntry, but uses much less memory
+type lowMemMap struct {
+	cnt       int
+	bucketCnt int
+	mask      int
+
+	buckets []*bucket
```
greatroar, line 74: Can we get rid of the * here? bucket is a pretty small struct and I'm worried about GC pressure, given the experience with the chaining table.

internal/repository/index_inmem.go
```diff
@@ -0,0 +1,322 @@
...
+
+// Make pages of 2^5 = 32 blobEntries by default
+const shiftBits = 5
+const pageSize = 1 << shiftBits
+const maskBits = pageSize - 1
+
+// lowMemMap - actually behaves like map[mapKey]mapEntry, but uses much less memory
+type lowMemMap struct {
+	cnt       int
+	bucketCnt int
```
greatroar, line 71: I'm not sure I understand bucketCnt. Is it ever different from len(buckets)?

aawsome: oops, right; should be pageCnt instead

greatroar:
I put up my chaining version as an alternative in #2812 because its allocation rate is comparable, but it's lookup speed is a lot better than this version. It also admits merging more easily, I think, if we still want to try that some day.

aawsome:
> I put up my chaining version as an alternative in #2812 because its allocation rate is comparable, but it's lookup speed is a lot better than this version. It also admits merging more easily, I think, if we still want to try that some day.

Your linked list looks quite promising (and IMO memory consumption could be much further decreased by not using a separate "duplicates") I'm however a bit irritated: my approach should lead to at most around 5 bytes overhead per index entry, while your approach (with separate duplicates) should yield 32 extra bytes. 
So I think the "new allocs" do not indicate memory usage, but just new allocates. 

Or I did miss some issues within my implementation...

greatroar:
allocs/op is the number of calls to the memory allocator. alloc/op is the total amount of memory requested, not the peak memory usage. I'm currently trying to benchmark both PRs against master on some realistic tasks, and I'm getting conflicting results: sometimes mine wins, sometimes yours. Throughput isn't much different. Do you know what restic command is most likely to allocate indexes and not much else?

You're right about the duplicates handling, that's the part I couldn't quite get right.

greatroar:
Also, I don't really get your 5 vs. 32 bytes comment, care to elaborate?

aawsome:
> Do you know what restic command is most likely to allocate indexes and not much else?

I would say `restic cat blob` is your candidate :wink: 

> You're right about the duplicates handling, that's the part I couldn't quite get right.

IMO you should just include them into your chained list (aka linked list) and implement a `get` and `getAll` method like I did in this PR. This will save you 24 bytes per index entry.

aawsome:
> Also, I don't really get your 5 vs. 32 bytes comment, care to elaborate?

The entry size for this PR is 32 bytes  + 16 bytes = 48 bytes. (49 bytes for your approach as you also save the BlobType)

For this PR: on average 1/2 byte of overhead per index entry + 1/2 unused page per 48 total pages. 1/2 page is 16*48 bytes and 48 pages will have around 32*48 entries. This is another 2 bytes overhead.

So I actually expect 2,5 bytes overhead per entry but worst case will be about double that amount.

For your approach I just took into account the extra pointer (8 bytes) and the duplicates slice (24 bytes). This is already 32 bytes per entry. On addition there is the overhead of managing the buckets which is 8 bytes per bucket and on average 2/3 buckets per entry. So total overhead should be 37 bytes on average. If you remove the duplicates slice, it will still be 13 bytes per entry on average and a bit more in the worst case.

greatroar:
I think you're missing some bytes of malloc overhead. The Go malloc, IIIRC, allocates in 64-byte multiples, so your buckets are bigger than you think. My use of a free list takes care of some of that.

Anyway, I found a way to store the duplicates in a separate list and get the indexEntry down to 64 bytes. Debugging as we speak :)

aawsome:
> I think you're missing some bytes of malloc overhead. The Go malloc, IIIRC, allocates in 64-byte multiples, so your buckets are bigger than you think. My use of a free list takes care of some of that.

I think it is either 8 bytes or 16 byte multiples up to 288 bytes or 32 byte multiples up to 512 bytes.

greatroar:
Judging from the [malloc code](https://github.com/golang/go/blob/c875503cf79f82b7d2dbec85667f330f72ec4557/src/runtime/sizeclasses.go), you're right about the tiny sizes. Good to know.

internal/repository/index.go
```diff
@@ -48,11 +48,10 @@ import (
 
 // Index holds lookup tables for id -> pack.
 type Index struct {
-	m          sync.Mutex
-	blob       map[restic.BlobHandle]indexEntry
-	duplicates map[restic.BlobHandle][]indexEntry
-	packs      restic.IDs
-	treePacks  restic.IDs
+	m         sync.Mutex
+	byType    map[restic.BlobType]*lowMemMap
```
greatroar, line 52: This little map may well be a performance killer on the benchmark. In #2812, I inline two lazily-initialized header structures in the Index.

aawsome:
After also adding a linked list variant (approach of @greatroar) , IMO that approach should be used. I change this to draft for reference and will close it as soon as #2812 is merged.

internal/repository/index.go
aawsome: It didn't contribute too much to the performance. However, using `byType []*lowMemMap` is the way better option. This can be directly accessed with `byType[tpe]` where `tpe` is `restic.BlobType`.
I changed the implementation, so that you can use this as reference for modifying your PR (also made a comment there).

aawsome:
closed as #2812 is merged.

PR #2813, opened 2020-06-28
Fix repository_test.BenchmarkSaveAndEncrypt

greatroar:

The SaveAndEncrypt benchmark was actually testing the speed of index lookups, rather than encryption, because its result was cached after the first call in every run. Benchmark results are the most dramatic I've seen so far:

```
name              old time/op    new time/op        delta
SaveAndEncrypt-8     101ns ± 2%    31077757ns ± 1%    +30886161.78%  (p=0.001 n=10+5)

name              old speed      new speed          delta
SaveAndEncrypt-8  41.7TB/s ± 2%       0.0TB/s ± 1%         -100.00%  (p=0.001 n=10+5)

name              old alloc/op   new alloc/op       delta
SaveAndEncrypt-8     1.00B ± 0%  20989534.60B ± 0%  +2098953360.00%  (p=0.001 n=10+5)

name              old allocs/op  new allocs/op      delta
SaveAndEncrypt-8      0.00             123.00 ± 0%            +Inf%  (p=0.001 n=10+5)
```

(The actual speed is ca. 135MiB/s.)

No.

Checklist
---------

### DIFF
--- a/internal/repository/repository_test.go
+++ b/internal/repository/repository_test.go
@@ -107,8 +107,7 @@ func BenchmarkSaveAndEncrypt(t *testing.B) {
 	t.SetBytes(int64(size))
 
 	for i := 0; i < t.N; i++ {
-		// save
-		_, _, err = repo.SaveBlob(context.TODO(), restic.DataBlob, data, id, false)
+		_, _, err = repo.SaveBlob(context.TODO(), restic.DataBlob, data, id, true)
 		rtest.OK(t, err)
 	}
 }


internal/repository/repository_test.go
```diff
@@ -91,8 +91,9 @@ func TestSaveFrom(t *testing.T) {
 }
 
 func BenchmarkSaveAndEncrypt(t *testing.B) {
-	repo, cleanup := repository.TestRepository(t)
+	r, cleanup := repository.TestRepository(t)
 	defer cleanup()
+	repo := r.(*repository.Repository)
```
MichaelEischer, line 96: I'd like to stick to using the repository interface. Passing `true` instead of `false` to `SaveBlob` would be enough to force SaveBlob to do actual work. The hashmap lookup in `addPending` shouldn't make much of a difference then.

greatroar: I was under the mistaken impression that that would cause a data structure to grow linearly. Fixed now.

MichaelEischer (approved):
LGTM. I've rebased to the current master and force pushed the branch. Let's hope that the rclone test doesn't fail this time.

Issue #666, opened 2016-11-10
Clearing page cache on client machines


zcalusic:
I use restic to make backup of Linux system partitions. I noticed that restic is clearing page cache on my client machines. And it makes a lot of trouble in many cases. Libraries, databases, everything that was cached before restic run, needs to be pulled again from disk and brought into memory after restic backup passes.

The worst example is a machine where I have lots of RRD files. They're notorious of write storms, but at least they can be cached pretty well, so they don't have to be read from disk again and again. Now with restic installed on that machine, a combination of read/write storms ensues after restic has passed, really bad.

The culprit are the two invocations of unix.Fadvise(..., unix.FADV_DONTNEED) in src/restic/fs/file_linux.go. So, it seems that only Linux is hit with this, and why Linux, which has the best memory management of them all?

I would ask, if these calls need to stay, to at least provide a switch to turn that reall bad behavior OFF. If i buy memory to have big page cache, I'd like to use it. I hope we all know that free memory on Linux is WASTED memory. I might as well pull out those DIMM's if they're not going to be used. :( And Linux has very sophisticated algorithms for its page cache, which keep the most needed file blocks in memory and gets rid of others that are referenced only once (search for active/inactive lists). Now, restic is basically deactivating this nice mechanism with 2 superficial calls. Let's fix that!

zcalusic:
This is issue #666. Yeah, it's that bad. :D

wscott:
Yes FADV_DONTNEED is not what we want. It tells the kernel not only that we
don't need this piece of data again, but that no one will. For big
footprint server machines with lots of memory, this is a total performance
killer.

Do a search, bup, attic, rsync and others have all had patches to add
DONTNEED and they were either never integrated or removed later.

fd0:
I agree that at least we need to have a switch to disable `fadvise` for reading data.

At the moment, restic also uses `fadvise` for data written to a local repository, e.g. during backup. This data will usually not be needed any more, and I think the Linux kernel can figure that out. Do we need to handle this case differently?

zcalusic:
Thanks for words of wisdom, @wscott, I really appreciate it. At some moment, it felt like I was a lone warrior for this important case. :)

@fd0, the case for backend files is definitely not that problematic (compared to client side), I have no strong opinion about that. 

[...]

PR #2816, opened 2020-07-01
Set O_NOATIME flag on Linux

greatroar:

The O_NOATIME flag is set when opening a file on Linux, which prevents the access time from being updated on every read.

Citing Kerrisk, The Linux Programming Interface:

> The O_NOATIME flag is intended for use by indexing and backup programs. Its use can significantly reduce the amount of disk activity, because repeated disk seeks back and forth across the disk are not required to read the contents of a file and to update the last access time in the file’s i-node[.]

restic used to do this, but the functionality was removed along with the fadvise call in #670. That solved #666, which is unrelated to O_NOATIME.

The implementation here is different: it uses fcntl, so we don't need to open the file twice when an error occurs (leading to duplicate path traversals). Errors are expected to be common, as O_NOATIME is only available to the owner of a file or root. Also, fcntl can set O_DIRECT on FreeBSD and is needed to set F_NOCACHE on MacOS, in case we ever want to do that.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2816
+++ b/changelog/unreleased/pull-2816
@@ -1,0 +1,10 @@
+Enhancement: backup no longer updates file access times on Linux
+
+When reading files during backup, restic caused the operating system to update
+the file access times. Note that this does not apply to filesystems with
+disabled file access times.
+
+Restic now instructs the operation system not to update the file access time,
+if the user running restic is the file owner or has root permissions.
+
+https://github.com/restic/restic/pull/2816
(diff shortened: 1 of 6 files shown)


MichaelEischer (approved):
LGTM. I've rebased the PR, moved the setFlags call to fs_local.go (see commit message) and updated to test.

@greatroar Can you have a look at my modifications to the `setFlags` usages?

MichaelEischer:
@greatroar Can you have a look at my modifications to your PR?

greatroar:
Merged the t.Log message into t.Skip setflags_linux_test.go. Your changes LGTM.

MichaelEischer:
Thanks for checking.

PR #2818, opened 2020-07-04
Merge index (based on chaining index implementation)

### LINKED ISSUE #944: [RFC] Scalability improvement for large amount of indexes
This is a request for comment to improve the startup time of most CLI commands.

## Background
Currently restic interleaves tree information with their data blobs in packs. File operations do a treewalk to get to the next level of directories. Access to those blobs (any blobs for that matter) is sped up by using an index which maps blob hashes to pack file/offsets. Each backup creates a new indexes which are kept around until either one of  `rebuild_indexes` or `prune` cli commands is used.

## Issue
At the end of each backup, a new index is created which contains the set of all previous indexes plus the new blobs written for this backup. This new index has a field, telling which other indexes it supersedes. Mainly this is done so that if 2 backups run at the roughly same time, each can write it's own index without interfering with the other backup. As a consequence one ends up with a large amount of superseded indexes, which, during startup of a cli cmd, ALL need to be read and determined for applicability.

### Idea
Somewhere after reading of the available indexes, we remove indexes who are marked superseded. This removal should be non-fatal as 2 backups started at the same time, after reading all the indexes, might start removing the same indexes. Also, when reading all indexes and then failing to open them (aka, a race condition with another ReadIndexes() instance who deleted it), should be ignored, as long as in the end ReadIndexes() finds at least one non superseded index to work with.
This would not need an update of the locking strategy, a backup already has a non exclusive lock and since we are sure that the superseded indexes do not play a role in the resulting index , we do not need to upgrade our backup lock, even though it is a mutating operation.

### Rationale
These superseded indexes are no longer needed as their mapping are by definition already contained any one the indexes marked as superseding them. (Note I'm only arguing to remove indexes where their superseding index is already committed to disk). 

Some senseless ascii art.
```console
A -> B(A) -> D(B,C) -> F(D,C)         -> H(F,G)
  -> C(A) ->       E(B,C)* -> G(D,E)  -|
Note that backup E took a long time, overlapping with the start of backup F. When E finishes, F has not finished so G does not see F's index.
```

### LINKED ISSUE #1550: Improving restics performance on large mostly unchanging datasets
## Output of `restic version`
Self compiled latest master

## Identified Issue
As a follow up to #1538 and #1549, on large backups with few changed files (>14M files total, probably <10000 changed) the performance is now limited by the Index.Has function (50.53% of the total cpu usage of restic during my backup).  In an effort to try and decrease backup times further, I was looking to see what could be done to further reduce the affect of this function.  There seems to be two significant sources of cpu usage:
 - Calling Mutex.Lock() (8% of the total cpu usage of restic).  This is mostly just the overhead of having a lock, and is not caused by having the lock be contested.
 - Doing the actual lookup into the map (28.04% of the total cpu usage of restic)
Note those are the total times of the entire backup, not just in the calls to Has().

Also mixed in there is the setup/teardown time for using defer.  Considering that the Go team is looking on improving restic's exact use case (calling Mutex.Unlock), it may not be worth worrying about this.

 - For the Mutex, I think replacing it with a RWMutex may help (I need to preform more benchmarking on this to be sure).  Otherwise, I'm not sure there is a way to avoid it.  If I understand the code base correctly, finalized Indexes don't need locking as they are read only.  Is this correct?  And would it be worth trying to take advantage of this fact (with the associated complexity increase)?
 - For the map, my only thought would be to decrease the size of it.  From reading how the map lookup is done, this *might* help, but more benchmarking is needed.  Is there a reason why the index size was chosen to be what it is?
 - Alternately, restic could try to call Has() less.  Restic seems to call Has() to check if a file is present in the backup repository already.  However, I was thinking if restic sees the file in the ancestor tree, it could use that fact to avoid having to check if the repository has the file.  Does this sound reasonable?  If so, should this wait on #1494 or be done at the same time?  If this is the preferred option but you want me to wait until after #1494 before trying this option that's fine.
 - My last idea would be to reorganize the indexes so that the MasterIndex (the main direct caller of Index.Has()) would only have to search through a limited number of indexes to find the blob.  This could also help reduce memory usage by avoiding having to load all the indexes into memory at the start of restic (which consumes ~6-8G for my 14M file backup), and instead load them on demand.  But this is a large change requiring a repository format break, so I can understand if this option is currently off the table.

Alternately, I can understand if you are happy with the performance as is right now (and I am generally happy too!) and rather concentrate on other more pressing features/bugs.  I was just looking to see if there was any low hanging fruit to optimize in restic and this seems to be my largest performance bottleneck right now.  If you want to come back to this problem later, feel free to either leave it open or close this issue as suits you.

I've been looking for a backup solution to replace Crashplan Home, most importantly be able to handle my ~7TB of data with incremental backups.  Restic is the best solution I've found so far, and I'm currently planing on using it extensively.  Especially considering how easy and fast it is to use!

### LINKED ISSUE #2523: Comparison of approaches to tackle index memory usage

aawsome:

Merge in-memory indexes when they are finalized.
This increases performance especially for large repositories with many index files.

The idea comes from #2523. This is a follow-up of #2799 as the discussion in there mainly moved to index-implementation related things.

This PR is based on #2812.

closes #944
closes #1550 
closes #2799
closes #2523 

Checklist
---------

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -49,8 +49,8 @@ type Index struct {
 	// only used by Store, StorePacks does not check for already saved packIDs
 	packIDToIndex map[restic.ID]int
 
-	final      bool      // set to true for all indexes read from the backend ("finalized")
-	id         restic.ID // set to the ID of the index when it's finalized
+	final      bool       // set to true for all indexes read from the backend ("finalized")
+	ids        restic.IDs // set to the IDs of the contained finalized indexes
 	supersedes restic.IDs
 	created    time.Time
 }
@@ -393,17 +393,17 @@ func (idx *Index) Finalize() {
 	idx.packIDToIndex = nil
 }
 
-// ID returns the ID of the index, if available. If the index is not yet
+// ID returns the IDs of the index, if available. If the index is not yet
 // finalized, an error is returned.
-func (idx *Index) ID() (restic.ID, error) {
+func (idx *Index) IDs() (restic.IDs, error) {
 	idx.m.Lock()
 	defer idx.m.Unlock()
 
 	if !idx.final {
-		return restic.ID{}, errors.New("index not finalized")
+		return nil, errors.New("index not finalized")
 	}
 
-	return idx.id, nil
+	return idx.ids, nil
 }
 
 // SetID sets the ID the index has been written to. This requires that
@@ -416,12 +416,12 @@ func (idx *Index) SetID(id restic.ID) error {
 		return errors.New("index is not final")
 	}
 
-	if !idx.id.IsNull() {
+	if len(idx.ids) > 0 {
 		return errors.New("ID already set")
 	}
 
 	debug.Log("ID set to %v", id)
-	idx.id = id
+	idx.ids = append(idx.ids, id)
 
 	return nil
 }
@@ -460,6 +460,38 @@ func (idx *Index) Dump(w io.Writer) error {
 // TreePacks returns a list of packs that contain only tree blobs.
 func (idx *Index) TreePacks() restic.IDs {
 	return idx.treePacks
+}
+
+// merge() merges indexes, i.e. idx.merge(idx2) merges the contents of idx2 into idx.
+// idx2 is not changed by this method.
+func (idx *Index) merge(idx2 *Index) error {
+	idx.m.Lock()
+	defer idx.m.Unlock()
+	idx2.m.Lock()
+	defer idx2.m.Unlock()
+
+	if !idx2.final {
+		return errors.New("index to merge is not final!")
+	}
+
+	packlen := len(idx.packs)
+	// copy all index entries of idx2 to idx
+	for typ := range idx2.byType {
+		m2 := &idx2.byType[typ]
+		m := &idx.byType[typ]
+		m2.foreach(func(entry *indexEntry) bool {
+			// packIndex is changed as idx2.pack is appended to idx.pack, see below
+			m.add(entry.id, entry.packIndex+packlen, entry.offset, entry.length)
+			return true
+		})
+	}
+
+	idx.packs = append(idx.packs, idx2.packs...)
+	idx.treePacks = append(idx.treePacks, idx2.treePacks...)
+	idx.ids = append(idx.ids, idx2.ids...)
+	idx.supersedes = append(idx.supersedes, idx2.supersedes...)
+
+	return nil
 }
 
 // isErrOldIndex returns true if the error may be caused by an old index
@@ -581,7 +613,7 @@ func LoadIndexWithDecoder(ctx context.Context, repo restic.Repository, buf []byt
 		return nil, buf[:0], err
 	}
 
-	idx.id = id
+	idx.ids = append(idx.ids, id)
 
 	return idx, buf, nil
 }
--- a/internal/repository/master_index.go
+++ b/internal/repository/master_index.go
@@ -18,7 +18,12 @@ type MasterIndex struct {
 
 // NewMasterIndex creates a new master index.
 func NewMasterIndex() *MasterIndex {
-	return &MasterIndex{pendingBlobs: restic.NewBlobSet()}
+	// Always add an empty final index, such that MergeFinalIndexes can merge into this.
+	// Note that removing this index could lead to a race condition in the rare
+	// sitation that only two indexes exist which are saved and merged concurrently.
+	idx := []*Index{NewIndex()}
+	idx[0].Finalize()
+	return &MasterIndex{idx: idx, pendingBlobs: restic.NewBlobSet()}
 }
 
 // Lookup queries all known Indexes for the ID and returns the first match.
@@ -237,6 +242,31 @@ func (mi *MasterIndex) Each(ctx context.Context) <-chan restic.PackedBlob {
 	return ch
 }
 
+// MergeFinalIndexes merges all final indexes together.
+// After calling, there will be only one big final index in MasterIndex
+// containing all final index contents.
+// Indexes that are not final are left untouched.
+// This merging can only be called after all index files are loaded - as
+// removing of superseded index contents is only possible for unmerged indexes.
+func (mi *MasterIndex) MergeFinalIndexes() {
+	mi.idxMutex.Lock()
+	defer mi.idxMutex.Unlock()
+
+	// The first index is always final and the one to merge into
+	newIdx := mi.idx[:1]
+	for i := 1; i < len(mi.idx); i++ {
+		idx := mi.idx[i]
+		// clear reference in masterindex as it may become stale
+		mi.idx[i] = nil
+		if !idx.Final() {
+			newIdx = append(newIdx, idx)
+		} else {
+			mi.idx[0].merge(idx)
+		}
+	}
+	mi.idx = newIdx
+}
+
 // RebuildIndex combines all known indexes to a new index, leaving out any
 // packs whose ID is contained in packBlacklist. The new index contains the IDs
 // of all known indexes in the "supersedes" field.
@@ -267,15 +297,15 @@ func (mi *MasterIndex) RebuildIndex(packBlacklist restic.IDSet) (*Index, error) 
 			continue
 		}
 
-		id, err := idx.ID()
+		ids, err := idx.IDs()
 		if err != nil {
 			debug.Log("index %d does not have an ID: %v", err)
 			return nil, err
 		}
 
-		debug.Log("adding index id %v to supersedes field", id)
+		debug.Log("adding index ids %v to supersedes field", ids)
 
-		err = newIndex.AddToSupersedes(id)
+		err = newIndex.AddToSupersedes(ids...)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/repository/master_index_test.go
+++ b/internal/repository/master_index_test.go
@@ -57,6 +57,58 @@ func TestMasterIndexLookup(t *testing.T) {
 	rtest.Assert(t, blobs == nil, "Expected no blobs when fetching with a random id")
 }
 
+func TestMasterMergeFinalIndexes(t *testing.T) {
+	idInIdx1 := restic.NewRandomID()
+	idInIdx2 := restic.NewRandomID()
+
+	blob1 := restic.PackedBlob{
+		PackID: restic.NewRandomID(),
+		Blob: restic.Blob{
+			Type:   restic.DataBlob,
+			ID:     idInIdx1,
+			Length: 10,
+			Offset: 0,
+		},
+	}
+
+	blob2 := restic.PackedBlob{
+		PackID: restic.NewRandomID(),
+		Blob: restic.Blob{
+			Type:   restic.DataBlob,
+			ID:     idInIdx2,
+			Length: 100,
+			Offset: 10,
+		},
+	}
+
+	idx1 := repository.NewIndex()
+	idx1.Store(blob1)
+
+	idx2 := repository.NewIndex()
+	idx2.Store(blob2)
+
+	mIdx := repository.NewMasterIndex()
+	mIdx.Insert(idx1)
+	mIdx.Insert(idx2)
+
+	finalIndexes := mIdx.FinalizeNotFinalIndexes()
+	rtest.Equals(t, []*repository.Index{idx1, idx2}, finalIndexes)
+
+	mIdx.MergeFinalIndexes()
+
+	blobs, found := mIdx.Lookup(idInIdx1, restic.DataBlob)
+	rtest.Assert(t, found, "Expected to find blob id %v from index 1", idInIdx1)
+	rtest.Equals(t, []restic.PackedBlob{blob1}, blobs)
+
+	blobs, found = mIdx.Lookup(idInIdx2, restic.DataBlob)
+	rtest.Assert(t, found, "Expected to find blob id %v from index 2", idInIdx2)
+	rtest.Equals(t, []restic.PackedBlob{blob2}, blobs)
+
+	blobs, found = mIdx.Lookup(restic.NewRandomID(), restic.DataBlob)
+	rtest.Assert(t, !found, "Expected to not find a blob when fetching with a random id")
+	rtest.Assert(t, blobs == nil, "Expected no blobs when fetching with a random id")
+}
+
 func createRandomMasterIndex(rng *rand.Rand, num, size int) (*repository.MasterIndex, restic.ID) {
 	mIdx := repository.NewMasterIndex()
 	for i := 0; i < num-1; i++ {
@@ -65,6 +117,10 @@ func createRandomMasterIndex(rng *rand.Rand, num, size int) (*repository.MasterI
 	}
 	idx1, lookupID := createRandomIndex(rng, size)
 	mIdx.Insert(idx1)
+
+	mIdx.FinalizeNotFinalIndexes()
+	mIdx.MergeFinalIndexes()
+
 	return mIdx, lookupID
 }
 
--- a/internal/repository/master_index_test.go
+++ b/internal/repository/master_index_test.go
@@ -57,11 +57,19 @@ func TestMasterIndexLookup(t *testing.T) {
 	rtest.Assert(t, blobs == nil, "Expected no blobs when fetching with a random id")
 }
 
-func BenchmarkMasterIndexLookupSingleIndex(b *testing.B) {
-	idx1, lookupID := createRandomIndex(rand.New(rand.NewSource(0)))
-
+func createRandomMasterIndex(rng *rand.Rand, num, size int) (*repository.MasterIndex, restic.ID) {
 	mIdx := repository.NewMasterIndex()
+	for i := 0; i < num-1; i++ {
+		idx, _ := createRandomIndex(rng, size)
+		mIdx.Insert(idx)
+	}
+	idx1, lookupID := createRandomIndex(rng, size)
 	mIdx.Insert(idx1)
+	return mIdx, lookupID
+}
+
+func BenchmarkMasterIndexLookupSingleIndex(b *testing.B) {
+	mIdx, lookupID := createRandomMasterIndex(rand.New(rand.NewSource(0)), 1, 200000)
 
 	b.ResetTimer()
 
@@ -71,16 +79,7 @@ func BenchmarkMasterIndexLookupSingleIndex(b *testing.B) {
 }
 
 func BenchmarkMasterIndexLookupMultipleIndex(b *testing.B) {
-	rng := rand.New(rand.NewSource(0))
-	mIdx := repository.NewMasterIndex()
-
-	for i := 0; i < 5; i++ {
-		idx, _ := createRandomIndex(rng)
-		mIdx.Insert(idx)
-	}
-
-	idx1, lookupID := createRandomIndex(rng)
-	mIdx.Insert(idx1)
+	mIdx, lookupID := createRandomMasterIndex(rand.New(rand.NewSource(0)), 100, 10000)
 
 	b.ResetTimer()
 
@@ -90,11 +89,9 @@ func BenchmarkMasterIndexLookupMultipleIndex(b *testing.B) {
 }
 
 func BenchmarkMasterIndexLookupSingleIndexUnknown(b *testing.B) {
-	lookupID := restic.NewRandomID()
-	idx1, _ := createRandomIndex(rand.New(rand.NewSource(0)))
 
-	mIdx := repository.NewMasterIndex()
-	mIdx.Insert(idx1)
+	lookupID := restic.NewRandomID()
+	mIdx, _ := createRandomMasterIndex(rand.New(rand.NewSource(0)), 1, 200000)
 
 	b.ResetTimer()
 
@@ -104,14 +101,8 @@ func BenchmarkMasterIndexLookupSingleIndexUnknown(b *testing.B) {
 }
 
 func BenchmarkMasterIndexLookupMultipleIndexUnknown(b *testing.B) {
-	rng := rand.New(rand.NewSource(0))
 	lookupID := restic.NewRandomID()
-	mIdx := repository.NewMasterIndex()
-
-	for i := 0; i < 6; i++ {
-		idx, _ := createRandomIndex(rng)
-		mIdx.Insert(idx)
-	}
+	mIdx, _ := createRandomMasterIndex(rand.New(rand.NewSource(0)), 100, 10000)
 
 	b.ResetTimer()
 
@@ -123,16 +114,12 @@ func BenchmarkMasterIndexLookupMultipleIndexUnknown(b *testing.B) {
 func BenchmarkMasterIndexLookupParallel(b *testing.B) {
 	mIdx := repository.NewMasterIndex()
 
-	for _, numindices := range []int{5, 10, 20} {
+	for _, numindices := range []int{25, 50, 100} {
 		var lookupID restic.ID
 
 		b.StopTimer()
 		rng := rand.New(rand.NewSource(0))
-		for i := 0; i < numindices; i++ {
-			var idx *repository.Index
-			idx, lookupID = createRandomIndex(rng)
-			mIdx.Insert(idx)
-		}
+		mIdx, lookupID = createRandomMasterIndex(rng, numindices, 10000)
 		b.StartTimer()
 
 		name := fmt.Sprintf("known,indices=%d", numindices)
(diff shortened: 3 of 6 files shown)


aawsome:
Here are benchmark results, where old is based on commit bb65143.

```
name                                         old time/op    new time/op    delta
PackerManager-4                                 265ms ± 2%     267ms ± 3%     ~     (p=0.113 n=9+10)
DecodeIndex-4                                  22.3µs ± 4%    22.4µs ± 2%     ~     (p=0.353 n=10+10)
IndexHasUnknown-4                              74.4ns ±17%    77.2ns ±13%     ~     (p=0.342 n=10+10)
IndexHasKnown-4                                72.0ns ± 4%    73.9ns ±11%     ~     (p=0.200 n=9+10)
IndexAlloc-4                                    900ms ± 2%     891ms ± 5%     ~     (p=0.218 n=10+10)
MasterIndexLookupSingleIndex-4                  229ns ± 6%     259ns ± 3%  +13.14%  (p=0.000 n=10+10)
MasterIndexLookupMultipleIndex-4                511ns ± 6%     252ns ± 7%  -50.60%  (p=0.000 n=9+10)
MasterIndexLookupSingleIndexUnknown-4          97.8ns ±10%   125.2ns ±13%  +28.07%  (p=0.000 n=10+10)
MasterIndexLookupMultipleIndexUnknown-4         420ns ±11%     101ns ± 9%  -75.92%  (p=0.000 n=10+10)
MasterIndexLookupMultipleIndexReal-4           10.5µs ± 6%     0.3µs ± 7%  -97.52%  (p=0.000 n=10+9)
MasterIndexLookupMultipleIndexRealUnknown-4    10.2µs ± 4%     0.1µs ±11%  -99.02%  (p=0.000 n=10+10)
SaveAndEncrypt-4                                136ns ± 2%     137ns ± 2%     ~     (p=0.230 n=9+10)
LoadTree-4                                      442µs ± 2%     467µs ± 1%   +5.78%  (p=0.000 n=10+10)
LoadBlob-4                                     8.23ms ± 0%    8.34ms ± 1%   +1.36%  (p=0.000 n=9+10)
LoadAndDecrypt-4                               10.2ms ± 2%    10.2ms ± 2%     ~     (p=0.436 n=9+9)
LoadIndex-4                                    43.8ms ± 3%    43.5ms ± 4%     ~     (p=0.631 n=10+10)

name                                         old speed      new speed      delta
PackerManager-4                               199MB/s ± 2%   197MB/s ± 3%     ~     (p=0.108 n=9+10)
SaveAndEncrypt-4                             30.9TB/s ± 2%  30.7TB/s ± 2%     ~     (p=0.243 n=9+10)
LoadBlob-4                                    121MB/s ± 0%   120MB/s ± 1%   -1.34%  (p=0.000 n=9+10)
LoadAndDecrypt-4                             98.5MB/s ± 2%  98.1MB/s ± 2%     ~     (p=0.449 n=9+9)

name                                         old alloc/op   new alloc/op   delta
PackerManager-4                                91.0kB ± 0%    91.0kB ± 0%     ~     (p=0.510 n=9+10)
IndexAlloc-4                                    400MB ± 0%     400MB ± 0%     ~     (p=0.579 n=10+10)
SaveAndEncrypt-4                                2.00B ± 0%     2.00B ± 0%     ~     (all equal)

name                                         old allocs/op  new allocs/op  delta
PackerManager-4                                 1.41k ± 0%     1.41k ± 0%     ~     (p=1.000 n=10+10)
IndexAlloc-4                                    1.12M ± 0%     1.12M ± 0%     ~     (p=0.516 n=10+10)
SaveAndEncrypt-4                                 0.00           0.00          ~     (all equal)
```

aawsome:
Also checked test coverage for the changes due to this PR: only the `grow()` in `moveto` is not called in tests, but is called in the benchmarks. All other changes are covered by tests.

internal/repository/indexmap.go
```diff
@@ -0,0 +1,226 @@
...
+	for e != nil {
+		h := m.hash(e.id)
+		next := e.next
+		e.next = m.buckets[h]
+		m.buckets[h] = e
+		e = next
+	}
+}
+
+// Take transfers e and its entire chain to m.
```
greatroar, line 194: Copy-pasted comment.

internal/repository/indexmap.go
```diff
@@ -0,0 +1,226 @@
...
+func (m *indexMap) moveto(m2 *indexMap, fn func(e *indexEntry)) {
```
greatroar, line 195: Instead of a generic callback, can't we just pass a `packOffset int`? (Callbacks are a pain to debug.)

internal/repository/indexmap.go
```diff
@@ -0,0 +1,226 @@
...
+	if m2.numentries == 0 { // Lazy initialization.
+		m2.buckets = make([]*indexEntry, 64)
+	}
+
+	m2.numentries += m.numentries
+	// grow m2 if needed
+	for m2.numentries > maxLoad*uint(len(m2.buckets)) {
+		m2.grow()
```
greatroar, line 203: This reallocates in a loop. If grow would take the desired number of elements as an argument, it could then round that up to a power of two using

```go
if n&(n-1) != 0 {
	n = 1 << bits.Len(n)
}
```

internal/repository/indexmap.go
```diff
@@ -0,0 +1,226 @@
...
+		e.next = m.buckets[h]
+		m.buckets[h] = e
+		e = next
+	}
+}
+
+// Take transfers e and its entire chain to m.
+func (m *indexMap) moveto(m2 *indexMap, fn func(e *indexEntry)) {
+	if m2.numentries == 0 { // Lazy initialization.
+		m2.buckets = make([]*indexEntry, 64)
```
greatroar, line 197: m2.init()

If m2 is not initialized, its random hash key is also still zero.

internal/repository/indexmap.go
aawsome: is fixed.

internal/repository/indexmap.go
aawsome: is changed.

internal/repository/indexmap.go
aawsome: I moved the loop into `grow()`. That should not be too expensive..

internal/repository/indexmap.go
aawsome: is fixed.

aawsome:
I rebased to the current version of #2812 and fixed the issues @greatroar found.

MichaelEischer (changes requested):
I completely agree that the masterIndex has to merge indexes to achieve good performance. However, the merge procedure should not break final indexes that got merged into the main index. As there's no way to ensure to nobody still has a reference to a merged index, this can lead to all sorts of subtle race conditions. So please _remove all modifications of the source index during a merge operation_.

internal/repository/master_index_test.go
```diff
@@ -104,14 +101,29 @@ func BenchmarkMasterIndexLookupSingleIndexUnknown(b *testing.B) {
...
-		idx, _ := createRandomIndex(rng)
-		mIdx.Insert(idx)
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		mIdx.Lookup(lookupID, restic.DataBlob)
 	}
+}
+
+func BenchmarkMasterIndexLookupMultipleIndexReal(b *testing.B) {
```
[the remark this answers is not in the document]
aawsome: I now just change the existing benchmarks making them more realistic.

internal/repository/indexmap.go
```diff
@@ -156,6 +165,37 @@ func (m *indexMap) newEntry() *indexEntry {
 	return e
 }
 
+// moveto() moves each element of the indexMap into another indexMap
+// It adds packOffset to packIndex of each element.
+func (m *indexMap) moveto(m2 *indexMap, packOffset int) {
```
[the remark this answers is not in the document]
aawsome: All right - I now simply use `indexMap.add`. This simplifies this PR a lot. We could further optimize by growing only once to match the target size of the new map, but IMO this is not worth the additional effort.

internal/repository/master_index.go
```diff
@@ -237,6 +242,29 @@ func (mi *MasterIndex) Each(ctx context.Context) <-chan restic.PackedBlob {
...
+	mi.idxMutex.Lock()
+	defer mi.idxMutex.Unlock()
+
+	// The first index is always final and the one to merge into
+	newIdx := mi.idx[:1]
+	for i := 1; i < len(mi.idx); i++ {
+		idx := mi.idx[i]
+		mi.idx[i] = nil
+		switch {
+		case !idx.Final():
```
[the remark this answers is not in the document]
aawsome: is changed.

internal/repository/master_index.go
```diff
@@ -237,6 +242,29 @@ func (mi *MasterIndex) Each(ctx context.Context) <-chan restic.PackedBlob {
...
+// Indexes that are not final are left untouched.
+func (mi *MasterIndex) MergeFinalIndexes() {
+	mi.idxMutex.Lock()
+	defer mi.idxMutex.Unlock()
+
+	// The first index is always final and the one to merge into
+	newIdx := mi.idx[:1]
+	for i := 1; i < len(mi.idx); i++ {
+		idx := mi.idx[i]
+		mi.idx[i] = nil
```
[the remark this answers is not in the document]
aawsome: is added.

internal/repository/master_index.go
```diff
@@ -267,17 +295,19 @@ func (mi *MasterIndex) RebuildIndex(packBlacklist restic.IDSet) (*Index, error)
...
 		}
 
-		debug.Log("adding index id %v to supersedes field", id)
+		debug.Log("adding index ids %v to supersedes field", ids)
 
-		err = newIndex.AddToSupersedes(id)
-		if err != nil {
-			return nil, err
+		for _, id := range ids {
+			err = newIndex.AddToSupersedes(id)
```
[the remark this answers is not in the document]
aawsome: Oh - I wasn't aware that `AddToSuperseded` already accepts `ids...` - thanks for proposing! Is changed.

internal/repository/index.go
```diff
@@ -462,6 +462,39 @@ func (idx *Index) TreePacks() restic.IDs {
...
+		m.moveto(&idx.byType[typ], len(idx.packs))
+	}
+
+	idx.packs = append(idx.packs, idx2.packs...)
+	idx.treePacks = append(idx.treePacks, idx2.treePacks...)
+	idx.ids = append(idx.ids, idx2.ids...)
+	idx.supersedes = append(idx.supersedes, idx2.supersedes...)
+
+	// idx2 is no longer used
+	idx2.packs = nil
```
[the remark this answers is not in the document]
aawsome: Now `idx2` is no longer modified and left over for the GC :wink:

internal/repository/master_index_test.go
```diff
@@ -124,6 +182,8 @@ func BenchmarkMasterIndexLookupMultipleIndexReal(b *testing.B) {
 func BenchmarkMasterIndexLookupMultipleIndexRealUnknown(b *testing.B) {
 	lookupID := restic.NewRandomID()
 	mIdx, _ := createRandomMasterIndex(rand.New(rand.NewSource(0)), 100, 10000)
+	mIdx.FinalizeNotFinalIndexes()
```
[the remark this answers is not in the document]
aawsome: is changed.

internal/repository/master_index.go
```diff
@@ -237,6 +242,29 @@ func (mi *MasterIndex) Each(ctx context.Context) <-chan restic.PackedBlob {
 	return ch
 }
 
+// MergeFinalIndexes merges all final indexes together.
+// After calling, there will be only one big final index in MasterIndex
+// containing all final index contents.
+// Indexes that are not final are left untouched.
```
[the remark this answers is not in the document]
aawsome: is added.

aawsome:
Now that this PR only calls `indexMap.add()` directly, the benchmarks need an update - coming soon.

internal/repository/master_index_test.go
```diff
@@ -102,7 +158,7 @@ func BenchmarkMasterIndexLookupSingleIndexUnknown(b *testing.B) {
 
 func BenchmarkMasterIndexLookupMultipleIndexUnknown(b *testing.B) {
 	lookupID := restic.NewRandomID()
-	mIdx, _ := createRandomMasterIndex(rand.New(rand.NewSource(0)), 100, 10000)
+	mIdx, _ := createRandomMasterIndex(rand.New(rand.NewSource(0)), 5, 200000)
```
MichaelEischer, line 161: This look like a leftover change.

MichaelEischer (commented):
The changes look good on a quick glance at the code. I'll wait with the final review until #2812 is merged.

internal/repository/master_index_test.go
aawsome: is fixed.

aawsome:
The benchmarks got killed on my 8GB laptop due to memory problems; however, the important benchmarks did run. Here is the result comparing old = 93c3601 with new = 0daff59 :
```
name                                             old time/op    new time/op    delta
IndexMapHash-4                                     5.88µs ± 2%    5.87µs ± 2%     ~     (p=0.837 n=10+10)
PackerManager-4                                     270ms ± 1%     267ms ± 1%   -0.91%  (p=0.046 n=9+8)
DecodeIndex-4                                       5.45s ± 1%     5.46s ± 2%     ~     (p=0.853 n=10+10)
DecodeIndexParallel-4                               5.43s ± 0%     5.39s ± 0%   -0.80%  (p=0.000 n=9+7)
IndexHasUnknown-4                                  76.1ns ±12%    73.5ns ±16%     ~     (p=0.342 n=10+10)
IndexHasKnown-4                                    76.4ns ±10%    74.8ns ± 9%     ~     (p=0.271 n=10+10)
IndexAlloc-4                                        848ms ± 1%     847ms ± 1%     ~     (p=1.000 n=8+10)
IndexAllocParallel-4                                513ms ± 3%     514ms ± 2%     ~     (p=0.666 n=9+9)
MasterIndexLookupSingleIndex-4                      198ns ± 6%     192ns ± 7%     ~     (p=0.138 n=9+10)
MasterIndexLookupMultipleIndex-4                   10.3µs ± 3%     0.3µs ±13%  -97.51%  (p=0.000 n=10+10)
MasterIndexLookupSingleIndexUnknown-4               104ns ±16%     103ns ±15%     ~     (p=0.809 n=10+10)
MasterIndexLookupMultipleIndexUnknown-4            10.0µs ± 3%     0.1µs ± 6%  -99.06%  (p=0.000 n=10+9)
MasterIndexLookupParallel/known,indices=100-4      4.89µs ± 1%    4.89µs ± 1%     ~     (p=0.868 n=10+10)
MasterIndexLookupParallel/unknown,indices=100-4    4.55µs ± 1%    4.64µs ± 2%   +2.16%  (p=0.000 n=8+9)

name                                             old speed      new speed      delta
IndexMapHash-4                                    697MB/s ± 2%   698MB/s ± 2%     ~     (p=0.853 n=10+10)
PackerManager-4                                   195MB/s ± 1%   197MB/s ± 1%   +0.91%  (p=0.046 n=9+8)

name                                             old alloc/op   new alloc/op   delta
IndexMapHash-4                                      0.00B          0.00B          ~     (all equal)
PackerManager-4                                    91.0kB ± 0%    91.0kB ± 0%     ~     (p=0.714 n=10+10)
IndexAlloc-4                                        401MB ± 0%     401MB ± 0%     ~     (all equal)
IndexAllocParallel-4                                400MB ± 0%     400MB ± 0%     ~     (all equal)

name                                             old allocs/op  new allocs/op  delta
IndexMapHash-4                                       0.00           0.00          ~     (all equal)
PackerManager-4                                     1.41k ± 0%     1.41k ± 0%     ~     (p=0.714 n=10+10)
IndexAlloc-4                                         907k ± 0%      907k ± 0%     ~     (all equal)
IndexAllocParallel-4                                 907k ± 0%      907k ± 0%     ~     (all equal)
```

MichaelEischer:
> The benchmarks got killed on my 8GB laptop due to unsufficient memory; however, the important benchmarks did run.

Insufficient memory on a laptop with 8GB RAM sounds a bit worrying, as in that case either the tests use too large indexes or have a memory leak. Which test failed exactly? FWIW the benchmarks completed successfully on my laptop and that also has "just" 8GB RAM.

aawsome:
> > The benchmarks got killed on my 8GB laptop due to unsufficient memory; however, the important benchmarks did run.
> 
> Insufficient memory on a laptop with 8GB RAM sounds a bit worrying, as in that case either the tests use too large indexes or have a memory leak. Which test failed exactly? FWIW the benchmarks completed successfully on my laptop and that also has "just" 8GB RAM.

I had some other programs in memory which also took their share...

The memory-hungry benchmark was `MasterIndexLookupParallel` which simulates 1M, 2M and 4M pack files (the other tests had a maximum of ~1M pack files). I changed it such that it simulates 250k, 500k and 1M pack files.

Here are the benchmark results:
```
name                                             old time/op    new time/op    delta
IndexMapHash-4                                     5.77µs ± 2%    5.90µs ± 2%   +2.38%  (p=0.000 n=10+10)
PackerManager-4                                     267ms ± 2%     267ms ± 1%     ~     (p=0.720 n=10+9)
DecodeIndex-4                                       5.39s ± 0%     5.43s ± 1%   +0.75%  (p=0.015 n=8+9)
DecodeIndexParallel-4                               5.39s ± 1%     5.41s ± 1%   +0.40%  (p=0.019 n=10+10)
IndexHasUnknown-4                                  79.0ns ±26%    73.3ns ±12%     ~     (p=0.243 n=10+9)
IndexHasKnown-4                                    75.3ns ±10%    75.6ns ±15%     ~     (p=0.897 n=10+10)
IndexAlloc-4                                        837ms ± 1%     890ms ± 5%   +6.32%  (p=0.000 n=9+10)
IndexAllocParallel-4                                450ms ± 3%     511ms ± 1%  +13.54%  (p=0.000 n=9+9)
MasterIndexLookupSingleIndex-4                      203ns ± 5%     196ns ± 5%   -3.60%  (p=0.030 n=10+10)
MasterIndexLookupMultipleIndex-4                   10.1µs ± 5%     0.2µs ±10%  -97.54%  (p=0.000 n=10+10)
MasterIndexLookupSingleIndexUnknown-4               100ns ±11%     104ns ± 9%     ~     (p=0.092 n=10+10)
MasterIndexLookupMultipleIndexUnknown-4            9.94µs ± 5%    0.09µs ± 7%  -99.06%  (p=0.000 n=10+9)
MasterIndexLookupParallel/known,indices=25-4       1.50µs ± 2%    0.44µs ± 2%  -70.43%  (p=0.000 n=10+10)
MasterIndexLookupParallel/unknown,indices=25-4     1.06µs ± 1%    0.30µs ± 1%  -71.17%  (p=0.000 n=10+9)
MasterIndexLookupParallel/known,indices=50-4       2.45µs ± 1%    0.47µs ± 3%  -80.73%  (p=0.000 n=9+9)
MasterIndexLookupParallel/unknown,indices=50-4     2.14µs ± 1%    0.30µs ± 6%  -85.90%  (p=0.000 n=10+10)
MasterIndexLookupParallel/known,indices=100-4      4.92µs ± 0%    0.46µs ± 5%  -90.57%  (p=0.000 n=8+10)
MasterIndexLookupParallel/unknown,indices=100-4    4.71µs ± 1%    0.31µs ± 2%  -93.39%  (p=0.000 n=9+10)
SaveAndEncrypt-4                                   42.7ms ± 1%    43.6ms ± 1%   +2.06%  (p=0.000 n=9+10)
LoadTree-4                                          437µs ± 3%     441µs ± 2%     ~     (p=0.065 n=9+10)
LoadBlob-4                                         8.22ms ± 0%    8.43ms ± 1%   +2.54%  (p=0.000 n=8+9)
LoadAndDecrypt-4                                   8.58ms ± 1%    8.58ms ± 0%     ~     (p=0.356 n=10+9)
LoadIndex-4                                        32.7ms ± 1%    33.6ms ± 1%   +2.93%  (p=0.000 n=10+10)

name                                             old speed      new speed      delta
IndexMapHash-4                                    710MB/s ± 2%   694MB/s ± 2%   -2.33%  (p=0.000 n=10+10)
PackerManager-4                                   197MB/s ± 2%   198MB/s ± 1%     ~     (p=0.720 n=10+9)
SaveAndEncrypt-4                                 98.2MB/s ± 1%  96.3MB/s ± 1%   -2.02%  (p=0.000 n=9+10)
LoadBlob-4                                        122MB/s ± 0%   119MB/s ± 1%   -2.48%  (p=0.000 n=8+9)
LoadAndDecrypt-4                                  117MB/s ± 1%   117MB/s ± 0%     ~     (p=0.344 n=10+9)

name                                             old alloc/op   new alloc/op   delta
IndexMapHash-4                                      0.00B          0.00B          ~     (all equal)
PackerManager-4                                    91.1kB ± 0%    91.1kB ± 0%     ~     (all equal)
IndexAlloc-4                                        401MB ± 0%     401MB ± 0%     ~     (all equal)
IndexAllocParallel-4                                400MB ± 0%     400MB ± 0%   +0.00%  (p=0.000 n=9+10)
SaveAndEncrypt-4                                   21.0MB ± 0%    21.0MB ± 0%   +0.00%  (p=0.000 n=10+9)

name                                             old allocs/op  new allocs/op  delta
IndexMapHash-4                                       0.00           0.00          ~     (all equal)
PackerManager-4                                     1.41k ± 0%     1.41k ± 0%     ~     (all equal)
IndexAlloc-4                                         907k ± 0%      907k ± 0%     ~     (all equal)
IndexAllocParallel-4                                 907k ± 0%      907k ± 0%   +0.00%  (p=0.000 n=9+10)
SaveAndEncrypt-4                                      118 ± 0%       118 ± 0%     ~     (all equal)
```
(wouldn't count too much about the ~2% differences, as said this is my laptop and there are other things like a firefox running...)

internal/repository/index.go
```diff
@@ -49,8 +49,8 @@ type Index struct {
 	// only used by Store, StorePacks does not check for already saved packIDs
 	packIDToIndex map[restic.ID]int
 
-	final      bool      // set to true for all indexes read from the backend ("finalized")
-	id         restic.ID // set to the ID of the index when it's finalized
+	final      bool       // set to true for all indexes read from the backend ("finalized")
+	ids        restic.IDs // set to the ID of the index when it's finalized
```
greatroar, line 53: This comment needs an update.

MichaelEischer: Is changed.

MichaelEischer:
I've run `check` for a large repository which ran for 3 hours in total and the index lookup only took only 1.3% of the overall runtime. The garbage collector took 8.9% of the overall runtime which seems to be similar to before.

MichaelEischer (commented):
LGTM

MichaelEischer (approved):
LGTM. Two are better than one ^^ .

PR #2844, opened 2020-07-19
Add prune integration tests for many edge cases

aawsome:

Add prune integration tests for many edge cases like duplicate blobs, mixed blobs or all types of corrupt repo.

This is part of #2718 that can be merged independently.

Checklist
---------


aawsome:
Just realized that the "failing test" (it was a blob existing in the pack file but missing in the index) is in fact cured by "old" prune. So I changed this accordingly here.

cmd/restic/integration_test.go
```diff
@@ -1283,6 +1283,103 @@ func TestPruneWithDamagedRepository(t *testing.T) {
...
+	datafile := filepath.Join("testdata", "repo-index-missing.tar.gz")
+	rtest.SetupTarTestFixture(t, env.base, datafile)
+
+	rtest.Assert(t, runCheck(CheckOptions{}, env.gopts, nil) != nil,
+		"check should have reported an error")
+	rtest.Assert(t, runPrune(env.gopts) != nil,
+		"prune should have reported an error")
+}
+
+func TestIndexMissingBlob(t *testing.T) {
```
MichaelEischer, line 1299: The other tests are also identical except for the tar-file, the check parameters and expected check result. Please merge these too.

MichaelEischer: What does `IndexMissingBlob` check? That a single blob is not contained in the index, but still exists in the repository?

cmd/restic/integration_test.go
```diff
@@ -1283,6 +1283,103 @@ func TestPruneWithDamagedRepository(t *testing.T) {
...
+
+func TestObsoleteIndex(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	datafile := filepath.Join("testdata", "repo-obsolete-index.tar.gz")
+	rtest.SetupTarTestFixture(t, env.base, datafile)
+
+	// check prints out an info but doesn't give an error for this setting.
+	testRunCheck(t, env.gopts)
```
MichaelEischer, line 1365: Can we adapt the test to check something?

MichaelEischer (changes requested):
How did you generate the test data sets? I wonder if there's an easy way to generate those on the fly?

aawsome:
> How did you generate the test data sets? I wonder if there's an easy way to generate those on the fly?

Honestly, by playing a lot around with creating backups, deleting manually single files from the repo, running `prune` and `rebuild-index` and sometimes even changing the source code to get the desired test repository.
I don't feel like I'm able to code that with small effort.

Also I think that keeping the repos as tar files "hardens" the tests cases in the sense that future code changes will not spoil the test cases. 
E.g. the code-change that removed mixed pack creation effectively prevents to generate those test settings on the fly... Similar things could happen with the other cases in future.

cmd/restic/integration_test.go
aawsome: Yes - `repo-index-missing-blob.tar.gz` is a repo where the packs are fine but a single (used) blob is missing from the index. With current `prune` this will be simply fixed (as `prune` always starts with the `rebuild-index` step), but after #2718 this will result in a prune error, which has to be explicitly fixed, e.g. by `rebuild-index`.

cmd/restic/integration_test.go
```diff
@@ -1283,6 +1283,103 @@ func TestPruneWithDamagedRepository(t *testing.T) {
 	t.Log(err)
 }
 
+func TestIndexMissing(t *testing.T) {
```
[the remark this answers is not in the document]
aawsome: I now use just two functions and also added comments about what is tested.

cmd/restic/integration_test.go
aawsome: It actually tests that restic `check` and `prune` still work with an obsolete index... :wink:

cmd/restic/integration_test.go
aawsome: The tests are now all merged.

cmd/restic/integration_test.go
```diff
@@ -1283,6 +1283,67 @@ func TestPruneWithDamagedRepository(t *testing.T) {
 	t.Log(err)
 }
 
+// Test repos for edge cases
+func TestEdgeCaseRepos(t *testing.T) {
+	opts := CheckOptions{}
+
+	// repo where index is completely missing
+	// => check and prune should fail
+	testEdgeCaseRepo(t, "repo-index-missing.tar.gz", opts, false, false)
```
MichaelEischer, line 1292: Please use `t.Run(...)` to make these subtests selectable. That could come in handy when hunting a specific problem.

MichaelEischer (changes requested):
> Also I think that keeping the repos as tar files "hardens" the tests cases in the sense that future code changes will not spoil the test cases.

That will also in parts verify that we don't break repository backwards compatibility.

I'd like to have the edge case tests as subtests to allow selecting a specific one. Besides that the code is ok. (The travis build error will disappear on the next CI run.)

cmd/restic/integration_test.go
aawsome: Thanks for the hint about how to make subtests - I wasn't aware of this.
Is changed now.

MichaelEischer (approved):
LGTM. Thanks :-)

Issue #2804, opened 2020-06-23
Using restic as an embedded package


goblain:
Output of `restic version` : 
--------------------------

irrelevant

How did you run restic exactly?
-------------------------------

trying to embed restic in my golang project

Expected behavior
-----------------

We're working on an agent for our software and want to bake in restic as a library so we can call restic logic directly rather then going via OS/CLI. To avoid to much code replication I want to use some functions that are now part of `cmd/` but that is a bit tough at the moment as they are not exported, neither is cmd/ the right place to embed in external project.

I looked at possibility or restructuring them so they are moved to `internal/`, placing it directly in `internal/restic/` led to dependency loop. Hence I decided to look at how hard would it be to split components into separate packages.

Do you have an idea how to solve the issue?
-------------------------------------------

Got a quick experiment going in https://github.com/restic/restic/compare/master...tradeix:locks and I wonder if something like that could make it into upstream ?

goblain:
oops, just learned the hard way that the whole `internal/` is not accessible to me from external project.

Issue #1089, opened 2017-07-07
Question: use restic as a library


GlenDC:
## restic (repo) version

latest master: 04d27acd60270fc63979eb7c1400c0cf6806d0e8

## Question

I was checking through the restic codebase, and it doesn't seem to be possible without having to copy a lot of the code to my own codebase to simply use restic functions directly from Go, effectively using Restic as a Library.

For example, instead of requiring restic to be installed to use a cmd such as `restic backup --stdin --stdin-filename <NAME>` I would instead prefer to call such logic directly from Go code. However when I check in the cmd code, a lot of the logic seems to be hidden behind private functions. Making it pretty impossible, without having to copy a lot of that code.

Perhaps I'm the first one to want to use restic as a library, but are there any reasons why this is not possible today? Or is it just because you didn't consider that use-case before? If you like the idea of that, I could make some time available to make it happen though. What it would mean is that the binary command functions wouldn't do much more than parsing flags and options, to than call the public restic Go API themselves, as the real command logic would in this case be moved behind the public restic Go API.

tamalsaha:
You might find some useful info here: https://github.com/restic/restic/issues/833

fd0:
Hey, thanks for asking. At least for now, restic is meant to be a standalone binary, that's the use case we focus on. Exposing an API as a library (that is stable) with the same guarantees as we do for the restic CLI binary requires time and effort, which we do not have at the moment. The internal structures are changing a lot, so this would be even harder than it is for the program.

We may reconsider this later, but for now: I'm sorry, but restic is not a library, and the only interface we have is the binary.

I'm going to close this issue for now, please feel free to add further comments.

GlenDC:
@fd0 I saw you mentioning this reason before, where guaranteeing a stable API would be extra work. But I'm not sure if I fully agree with that. I'm not expecting that all internal structures are exposed. That would indeed require extra work. The binary in itself exposes an API, right? What I'm proposing is basically to have more or less the same API as a Go API, and that Go API would than be used by the binary API. So there wouldn't really be much extra work involved in having this, I would think. Or am I missing something?

Thank you for responding so fast though, appreciated greatly.

fd0:
Ah, maybe a bit more explanation: With "API" I meant calling the `restic` CLI binary. Machine-readable output as JSON has been added already to several functions, and we can work together to add this wherever it is missing.

[...]

Issue #833, opened 2017-03-02
Why not import as Traditional "github.com/restic/restic"


sadlil:
Checking the codebase of restic it occurred to me **Why not the import statement uses the traditional format like** `github.com/restic/restic`? 
I have checked that `build.go` handles the complexities required to build the project. 
But it is hard to use the `src/restic` package as a dependency to build something based on this lib. Also importing as `github.com/restic/restic` could help use `go install` and other options too.

Is there any specific reason to use the import as it is?

BTW Thanks for a this great tool.

fd0:
We've discussed this a few times already, see #444 and #660.

There are two (independent) views to this. The user's view is: It should be easy to compile restic from a checkout of the repo. Do `git clone`, then run `go run build.go` and there's your binary. Especially I think it's important not to burden users with setting up a proper `GOPATH`. Please note that this can also be achieved when the "traditional" directory structure is used, just with different code in `build.go`.

The second view is the one of a developer: We've chosen to use the `gb` build tool a while back (in #444), so this repository is basically its own `GOPATH`. The reasons are described in the issue, one of the most important ones is that you (as a developer) do not need to fiddle around with checking out the repository at the right place in your `GOPATH` and then adding your own fork to it.

As for the "import as library" thing: Although the project is structured as a library and an executable, the library part is not suitable for importing in a different project. There are no guarantees at all concerning API stability. If this were a traditionally structured repository, I'd probably use an `/internal/` path component which prevents importing in third-party projects.

I'll close this issue as your question is answered, but please feel free to add a comment.

sadlil:
Thanks for the clarification. I had an scenario where i need to run the backups and restore process from another process, instead of running `exec` restic, i wanted to use the same API restic cli uses to do the tasks. Looks like i need to find some work around.

fd0:
We're in the process of adding JSON as an output option (`--json`) to some commands, maybe that's helpful for you in order to parse the results of a command...

fd0:
I'm curious: Can you describe your use case in a bit more detail?

sadlil:
Suppose i am running an Statefull application containerized and running within `kubernetes`. In my prod deployment i am thinking to run another sidecar container that will periodically backup my application's data via restic. So in that case restic needs to run time to time. Instead of running the cli via `exec`, its better to have control over the process using the library the cli uses.  That was my intention to do.

TamalSaha:
[...]

PR #2853, opened 2020-07-25
Building restic + rclone as a library

jdeng:
Building restic + rclone as a library will allow developers to build applications to access files stored in the Cloud via restic + rclone natively on multiple platforms including mobile.

There are 3 parts in the change: a new backend `inproc` to embed rclone as an http.RoundTripper via the rest backend which is only additional, moving 3 functions in `cmd/restic/global.go` to `cmd/global/global.go` so that they can be reused, and exposing some types and functions in `cmd/global/api.go`.

This requires some small change in rclone. See https://github.com/rclone/rclone/pull/4460/files
 
A code example can be found here: https://gist.github.com/jdeng/81f64666c61d143d5d3d9fb88f1f64dd

No. The code change is relatively small (except for global.go reorganization) and mostly glue code, so I hope this can be upstreamed. 

Checklist
---------

- [ ] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [ ] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

### DIFF
--- a/cmd/global/api.go
+++ b/cmd/global/api.go
@@ -1,0 +1,29 @@
+package global
+
+import (
+	"github.com/restic/restic/internal/cache"
+	"github.com/restic/restic/internal/options"
+	"github.com/restic/restic/internal/repository"
+	"github.com/restic/restic/internal/restic"
+)
+
+const (
+	DataBlob = restic.DataBlob
+	TreeBlob = restic.TreeBlob
+)
+
+// types
+type Repository = restic.Repository
+type Node = restic.Node
+type Tree = restic.Tree
+type TagList = restic.TagList
+type Cache = cache.Cache
+type Options = options.Options
+
+// functions
+var FindLatestSnapshot = restic.FindLatestSnapshot
+var LoadSnapshot = restic.LoadSnapshot
+var NewBlobBuffer = restic.NewBlobBuffer
+var CiphertextLength = restic.CiphertextLength
+var NewRepository = repository.New
+var NewCache = cache.New
--- a/cmd/global/global.go
+++ b/cmd/global/global.go
@@ -1,0 +1,315 @@
+package global
+
+import (
+	"context"
+	"os"
+
+	"github.com/restic/restic/internal/backend"
+	"github.com/restic/restic/internal/backend/azure"
+	"github.com/restic/restic/internal/backend/b2"
+	"github.com/restic/restic/internal/backend/gs"
+	"github.com/restic/restic/internal/backend/inproc"
+	"github.com/restic/restic/internal/backend/local"
+	"github.com/restic/restic/internal/backend/location"
+	"github.com/restic/restic/internal/backend/rclone"
+	"github.com/restic/restic/internal/backend/rest"
+	"github.com/restic/restic/internal/backend/s3"
+	"github.com/restic/restic/internal/backend/sftp"
+	"github.com/restic/restic/internal/backend/swift"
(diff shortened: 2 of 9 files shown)


rawtaz:
Hmm, restic as a library has already been discussed in https://github.com/restic/restic/issues/833, https://github.com/restic/restic/issues/1089 and one-way in https://github.com/restic/restic/issues/2804. Not sure anything changed on that part.

jdeng:
I think previous discussion was aiming at a more elegant and thorough change of the code structure, while this is more like a workaround without making significant changes to the codebase. The downside is that there is no guarantee on the API stability which is still much better than no API at all, IMO.

Another bonus point is that this also allows embedding of rclone which greatly expands its usability.

rawtaz:
Embedding rclone has been discussed a couple of times as well and it's not been encouraged I'm afraid :/

jdeng:
I think the point is to provide options to developers. I don't expect embedding rclone is a popular use case, but having all pieces in one binary across multiple platforms including mobile along with golang's tooling is quite appealing at least to some developers IMO. For example, I can build an app that pulls data from my personal restic data on Google Drive. This way I can avoid my data being scanned by the cloud providers.

greatroar:
@jdeng, your use case can be satisfied by a fork as well as an API client. The API option just shifts the maintenance burden to the central project.

jdeng:
To be clear, the API change is actually just one file with some alias of types and functions. It is probably less than .1% of total lines of code in restic. The problem is that the current code structure (everything is under internal) has prohibited usage of embedding restic as a library and this is a workaround without changing it. It acts like a header file in C/C++.

This is not elegant but it will make the certain developers' life much easier.  Instead of maintaining a local copy of fork and/or maintaining a separate binary, the same functionality can be achieved with just one single import.

```
const (
	DataBlob = restic.DataBlob
	TreeBlob = restic.TreeBlob
)

// types
type Repository = restic.Repository
type Node = restic.Node
type Tree = restic.Tree
type TagList = restic.TagList
type Cache = cache.Cache
type Options = options.Options

// functions
var FindLatestSnapshot = restic.FindLatestSnapshot
var LoadSnapshot = restic.LoadSnapshot
var NewBlobBuffer = restic.NewBlobBuffer
var CiphertextLength = restic.CiphertextLength
var NewRepository = repository.New
var NewCache = cache.New
```

fd0:
This PR does two very different things:
 1. Add a new API to call restic from within another Go project
 2. Embed rclone as a new backend

Both need to be discussed separately. I've outline my thoughts for the first point a few times already: We (as a project) struggle already with maintaining one interface (the restic binary with the JSON API), even more since the last time we discussed this feature request. Adding another API (which even exposes internal things that we cannot change afterwards without breaking the API contract) will add to this burden. Therefore, my initial response would be to reject this feature request, at least for now.

Embedding rclone is an interesting thought. On the upside, people don't need to install and maintain rclone. On the downside, this also adds to our maintenance burden. Restic would inherit all dependencies from rclone, we would need to closely follow rclone's release cycle and we would be force do to a release whenever there's a security issue in rclone. I'm undecided if it's a good idea for us (as a project). Let's open an issue for this feature request and discuss it there before implementing it.

I'm closing this PR for now. Please feel free to add further comments, thanks!

PR #2855, opened 2020-07-25
Fix rclone subprocess handling

### LINKED ISSUE #2381: restic does not detect rclone termination

MichaelEischer:
restic currently does not detect when the rclone subprocess exits and hangs forever (see #2381 and #2797). We also see test failure from time to time which fail with "signal: broken pipe" while closing the rclone backend.

This PR is partially inspired by #2797 and reuses the new test.

restic currently holds on to file descriptors for both the read and write side of the stdin and stdout pipes to the rclone subprocess. That prevents the pipes from closing properly once rclone stops and is also the root cause of not noticing the subprocess exit. Calling `stdioConn.Close()` once the child process has exited is still necessary to free the file descriptors.

Failing rclone tests seems to be caused because for some reason rclone dies due to a SIGPIPE error which is then returned by `cmd.Wait()` and gets printed as `signal: broken pipe`. I have no idea what data rclone writes to the stdio_conn when `Close` is called, maybe some leftover data on the HTTP2 connection? The fix for this is to give rclone time to shut down after stdin was closed, such that stdout is not closed while rclone still write to it.

fixes #2381 
replaces #2797 

Checklist
---------

### DIFF
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -197,7 +197,7 @@ func newBackend(cfg Config, lim limiter.Limiter) (*Backend, error) {
 		debug.Log("Wait returned %v", err)
 		be.waitResult = err
 		// close our side of the pipes to rclone
-		stdioConn.Close()
+		stdioConn.CloseAll()
 		close(waitCh)
 	}()
 
@@ -314,7 +314,7 @@ func (be *Backend) Close() error {
 		debug.Log("rclone exited")
 	case <-time.After(waitForExit):
 		debug.Log("timeout, closing file descriptors")
-		err := be.conn.Close()
+		err := be.conn.CloseAll()
 		if err != nil {
 			return err
 		}
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -114,7 +114,7 @@ func wrapConn(c *StdioConn, lim limiter.Limiter) wrappedConn {
 }
 
 // New initializes a Backend and starts the process.
-func New(cfg Config, lim limiter.Limiter) (*Backend, error) {
+func newBackend(cfg Config, lim limiter.Limiter) (*Backend, error) {
 	var (
 		args []string
 		err  error
@@ -234,7 +234,7 @@ func New(cfg Config, lim limiter.Limiter) (*Backend, error) {
 
 // Open starts an rclone process with the given config.
 func Open(cfg Config, lim limiter.Limiter) (*Backend, error) {
-	be, err := New(cfg, lim)
+	be, err := newBackend(cfg, lim)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


internal/backend/rclone/stdio_conn.go
```diff
@@ -28,21 +29,25 @@ func (s *StdioConn) Write(p []byte) (int, error) {
 	return n, err
 }
 
-// Close closes both streams.
+// Close closes the stream to the child process.
```
greatroar, line 32: Maybe this is the right time to rename stdin/stdout to toRclone/fromRclone. I was about to say that the comment is wrong, until I checked the code and realized I wanted to make that change myself and decided on the smaller patch.

internal/backend/rclone/backend.go
```diff
@@ -73,6 +73,13 @@ func run(command string, args ...string) (*StdioConn, *exec.Cmd, *sync.WaitGroup
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
+	// close rclone side of pipes
+	if err := r.Close(); err != nil {
+		return nil, nil, nil, nil, err
```
greatroar, line 78: In this case, it looks like w is never closed.

greatroar:
>  I have no idea what data rclone writes to the stdio_conn when Close is called, maybe some leftover data on the HTTP2 connection?

I'm not familiar with HTTP2, but a quick glance at RFC 7540 suggests that it may still send control messages when the connection is in half-open state. Maybe it just acknowledges the close?

In any case, tests pass on my machine.

internal/backend/rclone/backend.go
MichaelEischer: I have to admit I was too lazy to improve the resource cleanup here after an error. I'm still not sure I've caught everything.

internal/backend/rclone/stdio_conn.go
MichaelEischer: I've renamed the stream to send and receive

MichaelEischer:
> I'm not familiar with HTTP2, but a quick glance at RFC 7540 suggests that it may still send control messages when the connection is in half-open state. Maybe it just acknowledges the close?

The usual behavior seems to be to explicitly close the substreams, but I haven't seen commands to close the whole connection. At least nothing shows up when logging reads and writes on the stream.

fd0 (approved):
Great catch! LGTM!

PR #2856, opened 2020-07-28
Simplify cache logic

aawsome:

Simplifies the cache logic: 
- Actually a "ReadAhead" closure is defined which basically keeps the information which pack is a "treepack".
This closure is used for one single purpose: If a pack file is loaded and not yet contained in the cache, it will be loaded and saved in the cache for future use.
- Actually when saving tree packs, the same pack is saved again in the cache in `packer_manager`.

I changed the logic and added `BlobType` to the handle used to access the backend. This is in fact only used by the cache to determine whether a pack file from the backend should be cached.
As a side effect no tree packs need to be determined and stored within the index implementation, but instead mixed packs are saved.

Note that this PR does change the behavior of restic with respect to these points:
- When a pack files is accessed without knowing what blob types are within, this file will not be cached if not already present in the cache. This is e.g. the case for `pack.List` which is used in some places.
- Also tests are not modified and will sometimes not use the cache

No.

Checklist
---------

### DIFF
--- a/internal/cache/backend.go
+++ b/internal/cache/backend.go
@@ -21,7 +21,7 @@ type Backend struct {
 	inProgress      map[restic.Handle]chan struct{}
 }
 
-// ensure cachedBackend implements restic.Backend
+// ensure Backend implements restic.Backend
 var _ restic.Backend = &Backend{}
 
 func newBackend(be restic.Backend, c *Cache) *Backend {
@@ -43,13 +43,19 @@ func (b *Backend) Remove(ctx context.Context, h restic.Handle) error {
 	return b.Cache.remove(h)
 }
 
-func autoCached(t restic.FileType) bool {
-	return t == restic.IndexFile || t == restic.SnapshotFile
+func autoCacheTypes(h restic.Handle) bool {
+	switch h.Type {
+	case restic.IndexFile, restic.SnapshotFile:
+		return true
+	case restic.PackFile:
+		return h.ContainedBlobType == restic.TreeBlob
+	}
+	return false
 }
 
 // Save stores a new file in the backend and the cache.
 func (b *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
-	if !autoCached(h.Type) {
+	if !autoCacheTypes(h) {
 		return b.Backend.Save(ctx, h, rd)
 	}
 
@@ -168,25 +174,8 @@ func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset 
 		debug.Log("error loading %v from cache: %v", h, err)
 	}
 
-	// partial file requested
-	if offset != 0 || length != 0 {
-		if b.Cache.PerformReadahead(h) {
-			debug.Log("performing readahead for %v", h)
-
-			err := b.cacheFile(ctx, h)
-			if err == nil {
-				return b.loadFromCacheOrDelegate(ctx, h, length, offset, consumer)
-			}
-
-			debug.Log("error caching %v: %v", h, err)
-		}
-
-		debug.Log("Load(%v, %v, %v): partial file requested, delegating to backend", h, length, offset)
-		return b.Backend.Load(ctx, h, length, offset, consumer)
-	}
-
 	// if we don't automatically cache this file type, fall back to the backend
-	if !autoCached(h.Type) {
+	if !autoCacheTypes(h) {
 		debug.Log("Load(%v, %v, %v): delegating to backend", h, length, offset)
 		return b.Backend.Load(ctx, h, length, offset, consumer)
 	}
--- a/internal/restic/file.go
+++ b/internal/restic/file.go
@@ -21,8 +21,9 @@ const (
 
 // Handle is used to store and access data in a backend.
 type Handle struct {
-	Type FileType
-	Name string
+	Type              FileType
+	ContainedBlobType BlobType
+	Name              string
 }
 
 func (h Handle) String() string {
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -425,10 +425,11 @@ func removePacksExcept(gopts GlobalOptions, t *testing.T, keep restic.IDSet, rem
 
 	// Get all tree packs
 	rtest.OK(t, r.LoadIndex(gopts.ctx))
+
 	treePacks := restic.NewIDSet()
-	for _, idx := range r.Index().(*repository.MasterIndex).All() {
-		for _, id := range idx.TreePacks() {
-			treePacks.Insert(id)
+	for pb := range r.Index().Each(context.TODO()) {
+		if pb.Type == restic.TreeBlob {
+			treePacks.Insert(pb.PackID)
 		}
 	}
 
@@ -487,11 +488,10 @@ func TestBackupTreeLoadError(t *testing.T) {
 	r, err := OpenRepository(env.gopts)
 	rtest.OK(t, err)
 	rtest.OK(t, r.LoadIndex(env.gopts.ctx))
-	// collect tree packs of subdirectory
-	subTreePacks := restic.NewIDSet()
-	for _, idx := range r.Index().(*repository.MasterIndex).All() {
-		for _, id := range idx.TreePacks() {
-			subTreePacks.Insert(id)
+	treePacks := restic.NewIDSet()
+	for pb := range r.Index().Each(context.TODO()) {
+		if pb.Type == restic.TreeBlob {
+			treePacks.Insert(pb.PackID)
 		}
 	}
 
@@ -499,7 +499,7 @@ func TestBackupTreeLoadError(t *testing.T) {
 	testRunCheck(t, env.gopts)
 
 	// delete the subdirectory pack first
-	for id := range subTreePacks {
+	for id := range treePacks {
 		rtest.OK(t, r.Backend().Remove(env.gopts.ctx, restic.Handle{Type: restic.PackFile, Name: id.String()}))
 	}
 	testRunRebuildIndex(t, env.gopts)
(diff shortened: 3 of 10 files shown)


internal/cache/backend.go
```diff
@@ -43,14 +43,27 @@ func (b *Backend) Remove(ctx context.Context, h restic.Handle) error {
...
 
-var autoCacheTypes = map[restic.FileType]struct{}{
-	restic.IndexFile:    {},
-	restic.SnapshotFile: {},
+func autoCacheTypes(ft restic.FileType, bt restic.BlobType) bool {
+	if ft == restic.IndexFile || ft == restic.SnapshotFile ||
+		(ft == restic.DataFile && bt != restic.DataBlob) {
+		return true
+	}
+	return false
```
greatroar, line 51:
```
return ft == restic.IndexFile || ft == restic.SnapshotFile ||
	ft == restic.DataFile && bt != restic.DataBlob
```

or

```
switch ft {
case restic.IndexFile, restic.SnapshotFile:
	return true
case restic.DataFile:
	return bt != restic.DataBlob
}
```

Also, this conflicts with #2838, but the changes are very similar.

internal/cache/backend.go
```diff
@@ -149,7 +157,19 @@ func (b *Backend) loadFromCacheOrDelegate(ctx context.Context, h restic.Handle,
 }
 
 // Load loads a file from the cache or the backend.
+// For loading data files, please use LoadData!
```
greatroar, line 160: I'm not sure if this really simplifies things. The cached backend still implements Backend, but is no longer supposed to be used that way. The comment isn't clear on the relation between the two: if client code does use Load, will it break?

internal/restic/cache.go
```diff
@@ -35,3 +38,10 @@ type Cache interface {
 	// Has returns true if the file is cached.
 	Has(Handle) bool
 }
+
+// CachedBackend enhances the Backend interfaces to work with caches
```
greatroar, line 42: There's no need for another [interface with only one implementation](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a). I know this is how restic is currently structured in some places, but I'd argue for "return structs, accept interfaces" in new code. Also, the methods aren't documented.

internal/cache/backend.go
```diff
@@ -43,14 +43,27 @@ func (b *Backend) Remove(ctx context.Context, h restic.Handle) error {
...
 
 // Save stores a new file in the backend and the cache.
+// Use SaveData when saving data files
 func (b *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
-	if _, ok := autoCacheTypes[h.Type]; !ok {
+	return b.save(ctx, h, restic.InvalidBlob, rd)
+}
+
+// Save stores a new data file in the backend and the cache.
+func (b *Backend) SaveData(ctx context.Context, h restic.Handle, tpe restic.BlobType, rd restic.RewindReader) error {
```
MichaelEischer, line 61: I'm not convinced by that change. It requires differentiating between using `Save` and `SaveData`. I agree that the manual cache management in the `packer_manager` is not really optimal. But replacing a single use method with another, just makes the problem look different but won't solve it. Now we have the `Cache`, the `Backend` and a `CacheBackend`. 

The main problem with `PerformReadahead` in regards to the `packer_manager` is that it can't be updated. So my suggestion would be to modify the `Cache` and add a `AddReadahead(h restic.Handle)` method to it which replaces the `PerformReadahead` method. Then the `packer_manager` would just have to register the pack file before uploading and we're done.

internal/cache/backend.go
```diff
@@ -149,7 +157,19 @@ func (b *Backend) loadFromCacheOrDelegate(ctx context.Context, h restic.Handle,
 }
 
 // Load loads a file from the cache or the backend.
+// For loading data files, please use LoadData!
 func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset int64, consumer func(rd io.Reader) error) error {
+	return b.load(ctx, h, restic.InvalidBlob, length, offset, consumer)
```
MichaelEischer, line 162: This prevents caching of tree packs when using the generic `Load` method.

internal/cache/backend.go
```diff
@@ -174,25 +194,8 @@ func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset
 		debug.Log("error loading %v from cache: %v", h, err)
 	}
 
-	// partial file requested
-	if offset != 0 || length != 0 {
```
MichaelEischer, line 178: I just noticed that the current behavior is equivalent to removing this if and merging it into the if in line 195:
`if _, ok := autoCacheFiles[h.Type]; (offset != 0 || length != 0) && !b.Cache.PerformReadahead(h) || !ok`

As the index and snapshot files are always read completely this simplifies to `!ok && !b.Cache.PerformReadahead(h)` similar to the `autoCacheTypes` change. It would be good to keep this logic simplification.

internal/cache/backend.go
```diff
@@ -43,14 +43,27 @@ func (b *Backend) Remove(ctx context.Context, h restic.Handle) error {
 	return b.Cache.Remove(h)
 }
 
-var autoCacheTypes = map[restic.FileType]struct{}{
-	restic.IndexFile:    {},
-	restic.SnapshotFile: {},
+func autoCacheTypes(ft restic.FileType, bt restic.BlobType) bool {
+	if ft == restic.IndexFile || ft == restic.SnapshotFile ||
+		(ft == restic.DataFile && bt != restic.DataBlob) {
```
MichaelEischer, line 48: Distinguishing between mixed and tree-only packs is still a bit relevant for repositories created using old restic versions. For these the new behavior would amount to caching the whole backup repository, which seems rather problematic to me.

MichaelEischer (changes requested):
My impression that this PR merely shift the complexity from one place to another: now there's a `Backend` and a `CachedBackend` which requires using the right one to get tree packs cached.

I'd also like to keep the explicit tracking of tree packs. My suggestion would be to replace the `PerformReadahead` closure with a method to register `Handle`s which should be cached. This would also fit in nicely with the other already existing methods for cache management.

aawsome:
Thank you guys for your feedback! I agree that the new `CachedBackend` isn't a good idea; I'll change this.

My main intention in this was to change how restic determines which files are to be cached. The `ReadAhead` solution is not only unflexible but it also needs to save the list of "tree packs" in memory. This is then actually saved multiple times which I wanted to prevent. Moreover, the already-cached-files can also be easily determined by just looking at the cache itself, so there is in priciple no need to save those filenames again somewhere..

Also I find it way more natural to determine whether a file is to to be cached by its attributes and not by saving its name somewhere. For all files except pack files this is the way the cache already works (just determining by the filetype whetere the file is to be cached). So I would say it is much more natural to extend this to all files - and this requires for pack files to give the cache not only the information about the filetype, but also about the blobtype.

So I propose to extend the function signatures in the `Backend` interface such that the blob type can also be given for `Save` and `Load`. The real backend implementation would not use it, but a backend that wraps a cache or implements #3202 in some way would use it. To still support mixed pack files, the `InvalidBlobType` could be given for that case.

aawsome:
I changed the logic (and the description of this PR). Now the `Handle` structure is extended by the BlobType which is set when saving/loading pack files and knowing whether it is a data/tree/mixed pack. This information is then only used by the cache backend to store tree packs in the cache.

internal/cache/backend.go
aawsome: is changed.

MichaelEischer (changes requested):
I like the idea of annotating which type of blobs in contained in a pack file in general. However, we could (accidentally) end up with a few places in the code which no longer cache a file, as the burden to determine the file type now has moved to the caller and out of the cache backend. On the other hand it should be easier for the caller to know whether a file will be cached or not.

I've found a few more places in the code whose behavior is affected by the cache changes:
- `CreateIndexFromPacks` in `internal/repository/repository` calls `PackList` which then has to read a pack file from the repository. In most cases there pack file is probably not yet part of the repository index and thus this shouldn't make much of a difference.
-  `pack.List` is called in `packsToBlobs` of `cmd_find`. This would no longer cache the pack file if it is a tree pack. Although, I'd expect that only a very small number of packs will be used (actually just one before #3427)
- `cat pack ...` in `cmd_cat` would no longer cache a tree pack when dumping its content. But I guess that's not much of a problem.
- `printPacks` in `cmd_debug` uses `pack.List` which would no longer be cached. But I guess that would be ok for a debug command.
- And finally `examinePack` in `cmd_debug` uses `LoadAll` and `be.Load` which would no longer cache an examined tree pack. But I think the same reasoning as with the previous debug command applies.

So in summary I think these small behavior changes should nevertheless be ok. I only have one nitpick left before we could merge this, see the comment below.

internal/restic/file.go
```diff
@@ -22,6 +22,7 @@ const (
 // Handle is used to store and access data in a backend.
 type Handle struct {
 	Type FileType
+	BT   BlobType
```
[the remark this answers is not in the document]
aawsome: is changed now.

aawsome:
@MichaelEischer  Thanks a lot for your review and big excuse for the very late reply. I was heavily busy with finding a new home for my family and finally relocating. And wow, that was really a *very* busy time...
While not completely finished I'm starting to see light at the end of the tunnel now ;-)

I also rebased this PR to solve the merge conflicts. Hope it is now in a good shape to be merged.

MichaelEischer (approved):
LGTM.

> @MichaelEischer Thanks a lot for your review and big excuse for the very late reply. I was heavily busy with finding a new home for my family and finally relocating. And wow, that was really a _very_ busy time...
> While not completely finished I'm starting to see light at the end of the tunnel now ;-)

No problem. That really sounds like a lot of work, and I guess it's always surprising how much stuff accumulates over the years.

PR #2859, opened 2020-07-31
Add filter by tag and path to stats command

### LINKED ISSUE #2858: restic stats latest --tag
Output of `restic version`
--------------------------

With restic stats latest there is the possibility to filter on --host. But there is no possibility to filter on --tag. This makes it difficult if different backups are in the same repo and you want to check the backups automatically using stats latest.

With restic snapshots latest you have the possibility to filter on --tag so I would like to have it also with restic stats.

Would be cool if someone could integrate --tag into restic stats :)

What are you trying to do?
--------------------------

I am trying to provide metics for prometheus using restic stats and restic snapshots as described in https://forum.restic.net/t/restic-grafana-dashboard/1662.

buschjost:

This PR adds options to `restic stats` to filter latest snapshot by `--tag` and by `--path` . 

Closes #2858 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2858
+++ b/changelog/unreleased/issue-2858
@@ -1,0 +1,9 @@
+Enhancement: Support filtering snapshots by tag and path in the stats command
+
+We've added filtering snapshots by `--tag tagList` and by `--path path` to
+the `stats` command. This includes filtering of only 'latest' snapshots or
+all snapshots in a repository.
+
+https://github.com/restic/restic/issues/2858
+https://github.com/restic/restic/pull/2859
+https://forum.restic.net/t/stats-for-a-host-and-filtered-snapshots/3020
--- a/cmd/restic/cmd_stats.go
+++ b/cmd/restic/cmd_stats.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/walker"
 
@@ -15,18 +14,19 @@ import (
 )
 
 var cmdStats = &cobra.Command{
-	Use:   "stats [flags] [snapshot-ID]",
+	Use:   "stats [flags] [snapshot ID] [...]",
 	Short: "Scan the repository and show basic statistics",
 	Long: `
-The "stats" command walks one or all snapshots in a repository and
-accumulates statistics about the data stored therein. It reports on
-the number of unique files and their sizes, according to one of
+The "stats" command walks one or multiple snapshots in a repository
+and accumulates statistics about the data stored therein. It reports 
+on the number of unique files and their sizes, according to one of
 the counting modes as given by the --mode flag.
 
-If no snapshot is specified, all snapshots will be considered. Some
-modes make more sense over just a single snapshot, while others
-are useful across all snapshots, depending on what you are trying
-to calculate.
+It operates on all snapshots matching the selection criteria or all
+snapshots if nothing is specified. The special snapshot ID "latest"
+is also supported. Some modes make more sense over 
+just a single snapshot, while others are useful across all snapshots,
+depending on what you are trying to calculate.
 
 The modes are:
 
@@ -50,11 +50,26 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 	},
 }
 
+// StatsOptions collects all options for the stats command.
+type StatsOptions struct {
+	// the mode of counting to perform (see consts for available modes)
+	countMode string
+
+	// filter snapshots by, if given by user
+	Hosts []string
+	Tags  restic.TagLists
+	Paths []string
+}
+
+var statsOptions StatsOptions
+
 func init() {
 	cmdRoot.AddCommand(cmdStats)
 	f := cmdStats.Flags()
-	f.StringVar(&countMode, "mode", countModeRestoreSize, "counting mode: restore-size (default), files-by-contents, blobs-per-file, or raw-data")
-	f.StringArrayVarP(&snapshotByHosts, "host", "H", nil, "filter latest snapshot by this hostname (can be specified multiple times)")
+	f.StringVar(&statsOptions.countMode, "mode", countModeRestoreSize, "counting mode: restore-size (default), files-by-contents, blobs-per-file or raw-data")
+	f.StringArrayVarP(&statsOptions.Hosts, "host", "H", nil, "only consider snapshots with the given `host` (can be specified multiple times)")
+	f.Var(&statsOptions.Tags, "tag", "only consider snapshots which include this `taglist` in the format `tag[,tag,...]` (can be specified multiple times)")
+	f.StringArrayVar(&statsOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path` (can be specified multiple times)")
 }
 
 func runStats(gopts GlobalOptions, args []string) error {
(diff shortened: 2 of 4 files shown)


faizshukri:
Hey @buschjost,

I'm really excited to see this feature get merged.

However, I have a question. Why do we need to specify `snapshotId` or `latest` when filtering snapshots? Even existing `stats --host hostname` requires it too. Without it, `restic stats` returns the size of all snapshots (the whole repo) instead.

So then, is there any other way to get Total size for a particular host, or by a set of snapshots filtered by tag/path?

Thanks

cmd/restic/cmd_stats.go
```diff
@@ -101,7 +103,7 @@ func runStats(gopts GlobalOptions, args []string) error {
 
 		var sID restic.ID
 		if snapshotIDString == "latest" {
```
aawsome, line 105: Instead of this whole determination of the snapshot (if given) by `snapshotIDString := arg[0]`, IMO we should use
```golang
	for sn := range FindFilteredSnapshots(ctx, repo, opts.Hosts, opts.Tags, opts.Paths, args) {
		err = statsWalkSnapshot(ctx, sn, repo, stats)
		if err != nil {
			return fmt.Errorf("error walking snapshot: %v", err)
		}
	}
```

This would make `stats` behave like forget w.r.t. the given command line flags and arguments and solve the issue of @faizshukri while also making the code much cleaner. Moreover, the actual behavior (i.e. the only the two supported cases "no arguments = all snapshots" and "latest") would work almost identically. The only thing to change would be that the flag `--host` without `latest` would now not ignore the `--host` anymore.

However this change would also need some small additional modifications where `snapshotIDString` is actually used. 
@buschjost Are you ok to change this or should I propose another PR to implement this?

buschjost: @aawsome I was not sure about the reason of the current implementation and wanted to avoid breaking something. 

I had a short look on your suggestion and it feels like a change cleaning up many things. FindFilteredSnapshots seems to handle all cases. 

I'll have a look at adjusting it.

MichaelEischer: As specifying the `--host` option without `latest` should have thrown an error in the first place it should be fine to change this. After all I'd expected that the `--host` option is taken into account when passed to the `stats` command.

The error handling of `FindFilteredSnapshots` is different from the current behavior: Currently the command just aborts if an error occurs. FindFilteredSnapshots on the other hand just prints a warning message for the user and continues. This could become a problem when using the JSON interfaces. But in the end it just exchanges one invalid JSON output with another one, so it should be fine I guess.

buschjost:
Thanks for suggestions. Please have a look - I think I changed it according to your ideas. 
But I feel like I discovered a bug in FindFilteredSnapshots. 
I am now getting warnings "Ignoring filters as there are explicit snapshot ids given" that should not be there. While all behaves as expected and targets the expected snapshot.

> restic-dev/restic stats latest --host whatever.existing
> repository 572530f7 opened successfully, password is correct
> scanning...
> Ignoring filters as there are explicit snapshot ids given
> Stats in restore-size mode:
>Snapshots processed:   1
>   Total File Count:   161356
>         Total Size:   203.075 GiB

For me it seems like find.go:28 where it sets `usedFilter = true` would need to be only or also set if there was no error.  
 
```go
			for _, s := range snapshotIDs {
				if s == "latest" {
					id, err = restic.FindLatestSnapshot(ctx, repo, paths, tags, hosts)
					if err != nil {
						Warnf("Ignoring %q, no snapshot matched given filter (Paths:%v Tags:%v Hosts:%v)\n", s, paths, tags, hosts)
						usedFilter = true
						continue
					}
				} else {
```

because later the warning is triggered like:

```go
			// Give the user some indication their filters are not used.
			if !usedFilter && (len(hosts) != 0 || len(tags) != 0 || len(paths) != 0) {
				Warnf("Ignoring filters as there are explicit snapshot ids given\n")
			}
```

I would suggest changing it to: 

```go
			for _, s := range snapshotIDs {
				if s == "latest" {
					usedFilter = true
					id, err = restic.FindLatestSnapshot(ctx, repo, paths, tags, hosts)
					if err != nil {
						Warnf("Ignoring %q, no snapshot matched given filter (Paths:%v Tags:%v Hosts:%v)\n", s, paths, tags, hosts)
						continue
					}
				} else {
```

What do you think about the change. Should I include it in this MR or open a small separate one?

cmd/restic/cmd_stats.go
```diff
@@ -154,22 +129,17 @@ func runStats(gopts GlobalOptions, args []string) error {
...
-	// inform the user what was scanned and how it was scanned
-	snapshotsScanned := snapshotIDString
-	if snapshotsScanned == "latest" {
-		snapshotsScanned = "the latest snapshot"
-	} else if snapshotsScanned == "" {
-		snapshotsScanned = "all snapshots"
-	}
-	Printf("Stats for %s in %s mode:\n", snapshotsScanned, countMode)
+	Printf("Stats in %s mode:\n", countMode)
+
```
MichaelEischer, line 133: Please remove that newline between the Printf commands

cmd/restic/cmd_stats.go
```diff
@@ -154,22 +129,17 @@ func runStats(gopts GlobalOptions, args []string) error {
...
+	Printf("Snapshots processed:   %d\n", stats.snapshotsCount)
 
 	if stats.TotalBlobCount > 0 {
-		Printf("  Total Blob Count:   %d\n", stats.TotalBlobCount)
+		Printf("   Total Blob Count:   %d\n", stats.TotalBlobCount)
```
MichaelEischer, line 137: Is there a reason to change these statistic output commands?

cmd/restic/cmd_stats.go
```diff
@@ -320,6 +283,9 @@ type statsContainer struct {
 	// blobs is used to count individual unique blobs,
 	// independent of references to files
 	blobs restic.BlobSet
+
+	// holds count of all considered snapshots
+	snapshotsCount uint32
```
MichaelEischer, line 288: please use `int` as the datatype

doc/manual_rest.rst
```diff
@@ -306,6 +306,10 @@ host by using the ``--host`` flag:
 There we see that it would take 482 GiB of disk space to restore the latest
 snapshot from "myserver".
 
+In case you have multiple backups running from the same host so can also use
+``--tag`` and ``-path`` to be more specific about which snapshots you
```
MichaelEischer, line 310: The `-path` flag is missing a dash `-`.

cmd/restic/cmd_stats.go
```diff
@@ -329,12 +295,17 @@ var (
 	// the mode of counting to perform
 	countMode string
 
-	// the snapshot to scan, as given by the user
-	snapshotIDString string
+	// snapshotsByHosts are the hosts to filter
```
MichaelEischer, line 298: I wouldn't complain if these global variables moved into a `StatOptions` struct just like with every other command.

cmd/restic/cmd_stats.go
```diff
@@ -15,18 +14,19 @@ import (
...
 The "stats" command walks one or all snapshots in a repository and
 accumulates statistics about the data stored therein. It reports on
 the number of unique files and their sizes, according to one of
 the counting modes as given by the --mode flag.
 
-If no snapshot is specified, all snapshots will be considered. Some
-modes make more sense over just a single snapshot, while others
-are useful across all snapshots, depending on what you are trying
-to calculate.
+It allows "latest" as snapshot id and operates on all snapshots
```
MichaelEischer, line 25: I suggest to change this to something like:
```
It operates on all snapshots matching the selection criteria or all
snapshots if nothing is specified. The special snapshot ID "latest"
is also supported. Some modes [...]
```

MichaelEischer:
> But I feel like I discovered a bug in FindFilteredSnapshots.

I agree that it is a bug and also with the suggested change to fix it. I'm ok with fixing the error message as part of this PR.

cmd/restic/cmd_stats.go
buschjost: It is only to align the output with the new line containing the number of snapshots that were considered during taking the stats.

buschjost:
I applied the suggestions, introduced the options type and applied the change in FindFilteredSnapshots (in a separate commit).

cmd/restic/cmd_stats.go
MichaelEischer: Ah, I didn't notice from looking at the code alone. Looks fine.

MichaelEischer (approved):
LGTM. I've made one small change and moved the `stats.snapshotsCount++` line to the start of the function. Otherwise that snapshots counter stays at zero for most modes.

changelog/unreleased/issue-2858
```diff
@@ -0,0 +1,7 @@
+Enhancement: Support filtering snapshots by tag and path in the stats command
+
+We've added filtering snapshots by `--tag tagList` and by `--path path`
+to the`stats` command.
```
rawtaz, line 4: Please change this to:

~~~
We've added support for filtering snapshots by `--tag tagList` and `--path path`
in the `stats` command.
~~~

cmd/restic/cmd_stats.go
```diff
@@ -15,18 +14,19 @@ import (
 )
 
 var cmdStats = &cobra.Command{
-	Use:   "stats [flags] [snapshot-ID]",
+	Use:   "stats [flags] [snapshot ID] [...]",
 	Short: "Scan the repository and show basic statistics",
 	Long: `
 The "stats" command walks one or all snapshots in a repository and
```
rawtaz, line 20: Shouldn't this line have "one or more snapshots" rather than "one or all snapshots"?

cmd/restic/cmd_stats.go
```diff
@@ -50,11 +50,26 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
+}
+
+var statsOptions StatsOptions
+
 func init() {
 	cmdRoot.AddCommand(cmdStats)
 	f := cmdStats.Flags()
-	f.StringVar(&countMode, "mode", countModeRestoreSize, "counting mode: restore-size (default), files-by-contents, blobs-per-file, or raw-data")
-	f.StringArrayVarP(&snapshotByHosts, "host", "H", nil, "filter latest snapshot by this hostname (can be specified multiple times)")
+	f.StringVar(&statsOptions.countMode, "mode", countModeRestoreSize, "counting mode: restore-size (default), files-by-contents, blobs-per-file, or raw-data")
```
rawtaz, line 69: Normally it's written without the last , (that is, "foo, bar, star or blah" instead of "foo, bar, star, or blah").

rawtaz (changes requested):
I added some comments to texts/descriptions that I think should be fixed before this is merged.

changelog/unreleased/issue-2858
aawsome: I would add in the changelog that this change now allows filtering for all snapshots and not only for "latest" snapshot.

Maybe also add here a reference to the forum:
https://forum.restic.net/t/stats-for-a-host-and-filtered-snapshots/3020

aawsome (approved):
LGTM

rawtaz (approved):
If the code works it looks good to me :)

faizshukri:
Awesome. Thanks you all, especially @buschjost for making this feature available. 🎉

PR #2860, opened 2020-07-31
Add webdav server

### LINKED ISSUE #485: Add FTP server as alternative for fuse
Original issue title: Add WebDAV server as alternative for fuse

Important: This is not about an WebDAV Backend.

I think the idea of using fuse to examine the backuped snapshots is great, but as some OS do not support fuse and somebody might not have enough rights to use fuse, it would be great to have an alternative.
My idea was to add a command to restic that instead of mounting the snapshots via fuse opens a local WebDAV server. As all major OS have a options to mount a WebDAV server, everybody would be able to mount the snapshots.

As I guess that it is easy to find a WebDAV serverlib for Go and you would use the same directory structure the fuse mount, it should not be too complicated to add this feature.

aawsome:
Note: This PR is still WIP, however IMO it can be savely tested; only reads the repository and does no writes.

Add a webdav server to enable browsing repository contents with any webdav client.
The implementation uses the already existing fuse implementation and just "maps" it to webdav.

There are still some open points which are not solved within this PR:

- no enhanced webdav support (TLS, authentication, etc.)
- no performance tuning
- treatment of special files like symlinks, etc.

Closes #485

Checklist
---------

### DIFF
--- a/internal/webdav/fs.go
+++ b/internal/webdav/fs.go
@@ -1,0 +1,102 @@
+package webdav
+
+import (
+	"context"
+	"os"
+	"path"
+	"strings"
+
+	"bazil.org/fuse"
+	fusefs "bazil.org/fuse/fs"
+	"github.com/restic/restic/internal/debug"
+	"golang.org/x/net/webdav"
+)
+
+// RepoFileSystem implements a read-only file system on top of a repositoy.
+type RepoFileSystem struct {
+	ctx  context.Context
+	root fuseDir
+}
+
+// Mkdir creates a new directory, it is not available for RepoFileSystem.
+func (fs *RepoFileSystem) Mkdir(ctx context.Context, name string, perm os.FileMode) error {
+	return webdav.ErrForbidden
+}
+
+// RemoveAll recursively removes files and directories, it is not available for RepoFileSystem.
+func (fs *RepoFileSystem) RemoveAll(ctx context.Context, name string) error {
+	return webdav.ErrForbidden
+}
+
+// Rename renames files or directories, it is not available for RepoFileSystem.
+func (fs *RepoFileSystem) Rename(ctx context.Context, oldName, newName string) error {
+	return webdav.ErrForbidden
+}
+
+// open opens a file.
+func (fs *RepoFileSystem) open(ctx context.Context, name string) (webdav.File, error) {
+	var err error
+
+	name = path.Clean(name)
+	parts := strings.Split(name, "/")
+
+	node := fs.root.(fusefs.Node)
+
+	for _, part := range parts {
+		if part == "." || part == "" {
+			continue
+		}
+
+		// if there is a part left, the actual node must be a dir
+		nodedir, ok := node.(fuseDir)
+		if !ok {
+			// didn't get a dir
+			return nil, os.ErrNotExist
+		}
+
+		node, err = nodedir.Lookup(fs.ctx, part)
+		if err != nil {
+			if err == fuse.ENOENT {
+				return nil, os.ErrNotExist
+			}
+			return nil, err
+		}
+	}
+
+	var attr fuse.Attr
+	err = node.Attr(fs.ctx, &attr)
+	if err != nil {
+		return nil, err
+	}
+
+	switch {
+	case attr.Mode&os.ModeDir != 0: // dir
+		return &RepoDir{ctx: fs.ctx, dir: node.(fuseDir), name: name}, nil
+	case attr.Mode&os.ModeType == 0: // file
+		return &RepoFile{ctx: fs.ctx, file: node.(fuseFile), name: name, size: int64(attr.Size)}, nil
+	}
+
+	return &RepoLink{name: name}, nil
+
+}
+
+// OpenFile opens a file for reading.
+func (fs *RepoFileSystem) OpenFile(ctx context.Context, name string, flag int, perm os.FileMode) (webdav.File, error) {
+	debug.Log("Open %v", name)
+	if flag != os.O_RDONLY {
+		return nil, webdav.ErrForbidden
+	}
+	return fs.open(ctx, name)
+}
+
+// Stat returns information on a file or directory.
+func (fs *RepoFileSystem) Stat(ctx context.Context, name string) (os.FileInfo, error) {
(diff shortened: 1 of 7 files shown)


rawtaz:
In what way does this close #60? That issue is about implementing a *web server*, not a WebDAV server.

Does this work on Windows?

In general I'm not entirely sure about WebDAV as it's a protocol which only a few clients actually support well (the others support it like crap). But that's for #485.

aawsome:
> In what way does this close #60? That issue is about implementing a _web server_, not a WebDAV server.

It was just a suggestion. I'm not insisting on closing #60 :smile: 
 
> Does this work on Windows?

Obviously there are still some issues. However, by design it should work on Windows as well. It just uses the interfaces defined by `bazil.org/fuse` which are implemented by `internal/fuse/` and that's it.
So in fact, it does not rely on FUSE, but just on the restic code to implement FUSE.

I propose this PR to have an answer for "FUSE-demands" on system that don't support it :smile: 

> In general I'm not entirely sure about WebDAV as it's a protocol which only a few clients actually support well (the others support it like crap). But that's for #485.

I agree that some early tests showed strange behavior in the sense what action causes which call on the webdav server. This may also be a performance issue, but these are anyway not tackled here.

However I see your point that user issues may arise which may be due to a crappy client and which still will bind maintainer's time.. Can you propose a better protocol as FUSE-replacement that is widely used? (I don't think we do us a favor when implementing ftp...)

internal/webdav/fs.go
```diff
@@ -0,0 +1,102 @@
...
+
+// RepoFileSystem implements a read-only file system on top of a repositoy.
+type RepoFileSystem struct {
+	ctx  context.Context
+	root fuseDir
+}
+
+// Mkdir creates a new directory, it is not available for RepoFileSystem.
+func (fs *RepoFileSystem) Mkdir(ctx context.Context, name string, perm os.FileMode) error {
+	return webdav.ErrForbidden
```
greatroar, line 23: I'm not sure this error is supposed to be used by client code of x/net/webdav: its description says something about webdav.LockSystem. Returning something like errors.New("modifying a repository is not possible") would be cleaner, I think.

internal/webdav/fs.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+
+	"bazil.org/fuse"
+	fusefs "bazil.org/fuse/fs"
+	"github.com/restic/restic/internal/debug"
+	"golang.org/x/net/webdav"
+)
+
+// RepoFileSystem implements a read-only file system on top of a repositoy.
+type RepoFileSystem struct {
```
greatroar, line 16: Doesn't need to be public.

greatroar:
>  It just uses the interfaces defined by bazil.org/fuse which are implemented by internal/fuse/ and that's it.

The problem is that the fuse lib doesn't even build on platforms that it doesn't support. If we want to factor out the common code, we'd need to separate the fuse-using parts into separate files.

aawsome:
> > It just uses the interfaces defined by bazil.org/fuse which are implemented by internal/fuse/ and that's it.
> 
> The problem is that the fuse lib doesn't even build on platforms that it doesn't support. If we want to factor out the common code, we'd need to separate the fuse-using parts into separate files.

Ouch! That explains why the CI tests failed.. I didn't have a Windows machine to test this but I felt that webdav might be also interesting for Linux users..

However, I think we should wait for maintainers direction first before deciding about webdav or alternatives. There is also #2862 which goes into a similar direction. Maybe we should first start to factor out common "restic-as-filesystem" code, as you suggested.

rawtaz:
I'm pretty certain that this is not a high priority at the moment, there are other more important work to focus on for the next release and the one before that. I wouldn't spend much time on it right now.

fgma:
@aawsome I had the same problem as you already noticed. I reimplemented most of the fuse layer myself for the windows port. In my case there is also a big difference between different fuse implementations. bazil needs inodes, cgofuse just works with paths and does its own magic in the background. A common abstraction would be a good idea. If someone would provide some guidelines how the abstraction layer should look like we could join forces and get this done first and afterwards port our PRs to the new layer.

I've also mentioned the idea of a common abstraction in my forum posting here: https://forum.restic.net/t/pr-for-mount-support-on-windows/2966. So far I received no feedback.

@rawtaz Is there any way to for external contributors whats currently considered a priority feature/bug?

rawtaz:
@fgma What's considered a priority feature/bug is pretty dynamic as time passes. If a critical bug would surface we'd of course focus more on that, if someone would have severe problems with their integrity or similar we'd of course focus more on figuring that out and making sure it's not a bug in restic, etc. We don't keep an up to date list of items with order of priority, as developer and maintainer time is spent where it makes sense in general, where the person doing the work feel good about it (after all, we do this because we enjoy doing it, so everyone has to be free to tackle the things they find interesting, as well as what they think is important for the project), and where focus is needed to move the project forward. I guess this answer doesn't help you much. Is there any particular context/reason you ask? Currently we're trying to focus on not dealing with every little PR and issue that's filed, so we can get some things ready for release. That's not always simple given how many small code contributions pop up, since even the small ones take time :)

rawtaz:
Given #3096 this PR is of more interest nowadays :) Doesn't mean we have much time to focus on it right now - the main focus for us right now is the prune refactoring and optimizations as well as going through the pile of issues and PRs - but the work is of interest.

aawsome:
Because of real-life issues (good ones: my third child was born just a couple of days ago :smile: ), I'll have to decrease my engagement into restic in the next months. I decided to focus on the prune-related topics.

So if anyone wants to take over here, feel free to do what ever you want with this PR. If the topic stays open for the next couple of months, I might be more available to work on this...

dimejo:
> Because of real-life issues (good ones: my third child was born just a couple of days ago smile )

Congrats! :smile:

fd0:
Congratulations! Please do spend all the time you like with your family, everything else is secondary! :)

fgma:
Will this be the solution for mounting on mac in future? My PR would only allow mounting on windows while this should work on linux, windows and mac.

Any chance to get this PR ready for a merge in near future? I could spend some time helping with this PR in case it is needed.

Oakwas:
Any updates on this?

PR #2863, opened 2020-08-02
Don't save exact duplicates in merged index

aawsome:

Don't save exact duplicates (i.e. same blob ID, same pack, same offset and same length) in index. This situation can happen in rare cases where superseded index files are not deleted. Saving exact duplicates doesn't help anywhere but may pose some problems (more memory usage; and possible trouble within #2842) 

see #2839 

Checklist
---------

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -464,7 +464,7 @@ func (idx *Index) TreePacks() restic.IDs {
 }
 
 // merge() merges indexes, i.e. idx.merge(idx2) merges the contents of idx2 into idx.
-// idx2 is not changed by this method.
+// During merging exact duplicates are removed;  idx2 is not changed by this method.
 func (idx *Index) merge(idx2 *Index) error {
 	idx.m.Lock()
 	defer idx.m.Unlock()
@@ -476,18 +476,35 @@ func (idx *Index) merge(idx2 *Index) error {
 	}
 
 	packlen := len(idx.packs)
+	// first append packs as they might be accessed when looking for duplicates below
+	idx.packs = append(idx.packs, idx2.packs...)
+
 	// copy all index entries of idx2 to idx
 	for typ := range idx2.byType {
 		m2 := &idx2.byType[typ]
 		m := &idx.byType[typ]
-		m2.foreach(func(entry *indexEntry) bool {
-			// packIndex is changed as idx2.pack is appended to idx.pack, see below
-			m.add(entry.id, entry.packIndex+packlen, entry.offset, entry.length)
+
+		// helper func to test if identical entry is contained in idx
+		hasIdenticalEntry := func(e2 *indexEntry) (found bool) {
+			m.foreachWithID(e2.id, func(e *indexEntry) {
+				b := idx.toPackedBlob(e, restic.BlobType(typ))
+				b2 := idx2.toPackedBlob(e2, restic.BlobType(typ))
+				if b.Length == b2.Length && b.Offset == b2.Offset && b.PackID == b2.PackID {
+					found = true
+				}
+			})
+			return found
+		}
+
+		m2.foreach(func(e2 *indexEntry) bool {
+			if !hasIdenticalEntry(e2) {
+				// packIndex needs to be changed as idx2.pack was appended to idx.pack, see above
+				m.add(e2.id, e2.packIndex+packlen, e2.offset, e2.length)
+			}
 			return true
 		})
 	}
 
-	idx.packs = append(idx.packs, idx2.packs...)
 	idx.treePacks = append(idx.treePacks, idx2.treePacks...)
 	idx.ids = append(idx.ids, idx2.ids...)
 	idx.supersedes = append(idx.supersedes, idx2.supersedes...)
--- a/internal/repository/master_index_test.go
+++ b/internal/repository/master_index_test.go
@@ -1,6 +1,7 @@
 package repository_test
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"testing"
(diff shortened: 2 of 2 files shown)


internal/repository/indexmap.go
```diff
@@ -34,7 +34,7 @@ const (
 )
 
 // add inserts an indexEntry for the given arguments into the map,
-// using id as the key.
+// does not insert duplicates (all values identical to an existing one).
```
greatroar, line 37: This is a new sentence and it contradicts the comment on type indexMap.

```
 // add inserts an indexEntry for the given arguments into the map.
// It allows duplicate ids, but not fully duplicate entries.
```

The comment on indexMap needs an update too.

greatroar:
I haven't investigated the problem, but the code looks good. Insertion will be more expensive, but is still O(1) expected. Merge reduces to add so that is covered as well (maybe add a test for that?).

internal/repository/indexmap_test.go
```diff
@@ -44,26 +44,38 @@ func TestIndexMapForeach(t *testing.T) {
...
+	testForEach := func() {
+		seen := make(map[int]struct{})
+		m.foreach(func(e *indexEntry) bool {
+			i := int(e.id[0])
+			rtest.Assert(t, i < N, "unknown id %v in indexMap", e.id)
+			rtest.Equals(t, i, e.packIndex)
+			rtest.Equals(t, i, int(e.length))
+			rtest.Equals(t, i, int(e.offset))
+
+			seen[i] = struct{}{}
```
MichaelEischer, line 56: This doesn't count duplicate ids. Please make sure that your tests actually _fail_ without the fixed code. So either write the test first or undo the change afterwards and check that the test fails.

aawsome: My bad :-1: 
I'll change this.

aawsome:
There is another issue I just came around when thinking about this change once again: The packIDs are not deduplicated so far.. 
I need to think about how to best do this and change this to WIP in meanwhile..

aawsome:
I changed this such that now the check is performed in `Index.merge()`. IndexMap` stays unchanged now.

internal/repository/index.go
```diff
@@ -476,18 +476,30 @@ func (idx *Index) merge(idx2 *Index) error {
...
+
 	// copy all index entries of idx2 to idx
 	for typ := range idx2.byType {
 		m2 := &idx2.byType[typ]
 		m := &idx.byType[typ]
-		m2.foreach(func(entry *indexEntry) bool {
-			// packIndex is changed as idx2.pack is appended to idx.pack, see below
-			m.add(entry.id, entry.packIndex+packlen, entry.offset, entry.length)
+		m2.foreach(func(e2 *indexEntry) bool {
+			found := false
```
MichaelEischer, line 487: Please add a comment the the `merge` method which mentions that duplicate entries are removed.

The next few lines also needs a comment. And I'd prefer these to be move into a method like `hasPackedBlob` or similar. At first glance it's not obvious at all what the foreach inside another foreach does.

internal/repository/master_index_test.go
```diff
@@ -167,6 +168,14 @@ func TestMasterMergeFinalIndexes(t *testing.T) {
 	rtest.Equals(t, []*repository.Index{idx1, idx2}, finalIndexes)
 
 	mIdx.MergeFinalIndexes()
+	allIndexes := mIdx.All()
+	rtest.Equals(t, 1, len(allIndexes))
+
+	count := 0
```
MichaelEischer, line 174: Please rename the variable to `blobCount`.

MichaelEischer (commented):
@aawsome Did you check the masterIndex benchmarks regarding the performance impact of the additional `foreach` loop?

internal/repository/index.go
aawsome: is changed.

internal/repository/master_index_test.go
aawsome: is changed

aawsome:
> @aawsome Did you check the masterIndex benchmarks regarding the performance impact of the additional `foreach` loop?

Note that the additional `foreachID` loop is `O(1)` itself (at least if the undelying hash is not degenerated). Hence the total complexity of the merge is still roughly (neglecting hash growing of `idx`) `O(n)` with `n` being the number of elements of the index to be merged, i.e. `idx2`. 

Moreover this change only affects the merge itself, and in real-world-scenarios this occurs during 

- loading the index from index files into memory (should be dominated by loading and extracting the files)
- saving a new finished index during backup (this should happen quite rarely as backup itself should be mainly dominated by 

So I do not expect any impact on real-life usage.

About benchmarking: Seems to me that none of the current benchmarks actually includes merging of indexes in the measurements.
I added the benchmark `MasterIndexAlloc` which creates Indexes and merges them under measurement.
Here are the results comparing 5e63294 with b112533:
```
name                old time/op    new time/op    delta
MasterIndexAlloc-4     334ms ± 1%     460ms ± 1%  +37.68%  (p=0.000 n=9+8)

name                old alloc/op   new alloc/op   delta
MasterIndexAlloc-4     258MB ± 0%     258MB ± 0%     ~     (p=0.101 n=10+8)

name                old allocs/op  new allocs/op  delta
MasterIndexAlloc-4      317k ± 0%      317k ± 0%     ~     (p=0.267 n=10+6)
```

MichaelEischer (approved):
LGTM.

> Note that the additional `foreachID` loop is `O(1)` itself (at least if the undelying hash is not degenerated). Hence the total complexity of the merge is still roughly (neglecting hash growing of `idx`) `O(n)` with `n` being the number of elements of the index to be merged, i.e. `idx2`.

`foreachID` could end up a rather large value if a single blob exists lots of times in different packs. But in that case the repository is probably really broken. So `amortized O(1)` would be slightly more precise (not that it matters).

The overhead of 40% seems fine for me, I was worried a bit that we could end up with an overhead of maybe 2x-3x.

PR #2866, opened 2020-08-03
Fix TestMetadataChanged on macOS

erikgeiser:

This pull request fixes a test (`TestMetadataChanged`) that fails on macOS due to sub-second differences in the access time of a node after a snapshot. During my investigation I found that the access time behaviour is a bit strange on macOS Catalina with APFS. It seems to be only updated upon modification with a slight delay causing the access time and the mod time to differ a little bit. This revealed an inconsistency in the archiver's metadata test:

The `want` node is created using [`restic.NodeFromFileInfo()`](https://github.com/restic/restic/blob/66d089e23970dd34e93e87f739740bf9702f4f5c/internal/archiver/archiver_test.go#L2040). This results in a node with separate `AccessTime` and `ModTime` values that might differ. Then, `snapshot()` is called, which does not use [`restic.NodeFromFileInfo()`](https://github.com/restic/restic/blob/66d089e23970dd34e93e87f739740bf9702f4f5c/internal/archiver/archiver.go#L185-L191) directly when saving, but rather uses the wrapper method `Archiver.nodeFromFileInfo()` on an `Archiver` with `WithAtime` set to `false`. Thus, the following code causes the `AccessTime` in the resulting node to be set to the `ModTime`:
```go
func (arch *Archiver) nodeFromFileInfo(filename string, fi os.FileInfo) (*restic.Node, error) {
	node, err := restic.NodeFromFileInfo(filename, fi)
	if !arch.WithAtime {
		node.AccessTime = node.ModTime
	}
	return node, errors.Wrap(err, "NodeFromFileInfo")
}
```

As the `AccessTime` is now changed from the value present in the `want` node, the test fails if `AccessTime` was previously (in the `want` node) not equal to `ModTime` (as seems to be the case on macOS).

I fixed the test by forcing the `want` node to have an equal `AccessTime` and `ModTime` because this node configuration matches the expected result if an `Archiver` with `WithAtime` set to `false` is used to create the snapshot.

Not as far as I'm aware. I could not find any related issues and the change is so minor that I did not want to split the discussion between this PR and a new separate issue.

Checklist
---------

- [x] I have added tests for all changes in this PR (not applicable, as this PR modifies an existing test)

Is a changelog file necessary in this case where the changes are not user facing and only fix an edge case? If yes, I will provide one.

### DIFF
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -2050,6 +2050,11 @@ func TestMetadataChanged(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// create want node the same way the archiver
+	// does when WithAtime is false by setting
+	// AccessTime to ModTime
+	want.AccessTime = want.ModTime
+
 	fs := &StatFS{
 		FS: fs.Local{},
 		OverrideLstat: map[string]os.FileInfo{


MichaelEischer:
This has already been fixed by #2465 . Please make sure that you use the _current_ master branch as starting point. Your PR is based on a commit which is more than a year old, see https://github.com/restic/restic/commits/23ff0cf7e8bba77143a8c4918b84ba7034acf267 .

PR #2871, opened 2020-08-04
Add content_hash to ls --json output

### LINKED ISSUE #2870: Add content hash to ls --json
Output of `restic version`
--------------------------

restic 0.9.6 (v0.9.6-337-g0b21ec44-dirty)

Add a content hash to the `ls --json` output.

When the file was stored in a single chunk, the real sha256 has is available and the reported hash will have the format `"sha256:..."`.

When the file was split into multiple chunks, it is not possible to show a real content hash, because restic does not currently store this information. We can however construct a hash out of the chunk hashes. To distinguish this from a real sha256 of the contents, this hash will have the format `"multi:..."`.

This 'multi' hash can only be compared within a single repo, as different repos will split files in different locations.

I have a PR ready that I add in a moment.

What are you trying to do?
--------------------------

I am trying to figure out if any original pictures have succumbed to bitrot since my initial restic backup of them.

Keeping the full backup history of my pictures in an efficient way makes it possible to recover from bitrot in the future.

wojas:

Add a new content_hash field to every file in the `ls --json` output.

For small files, this will usually be their sha256 hash.

For larger files, we only have the hashes of the content blocks, so we construct a single hash out of those hashes.

See #2870.

Closes #2870 

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2871
+++ b/changelog/unreleased/pull-2871
@@ -1,0 +1,8 @@
+Enhancement: Add content hash to `ls --json` output
+
+Added a new `content_hash` field to the `ls --json` output.
+
+These content hashes can only be compared within a single backup repository.
+
+https://github.com/restic/restic/issues/2870
+https://github.com/restic/restic/pull/2871
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -2,11 +2,13 @@ package main
 
 import (
 	"context"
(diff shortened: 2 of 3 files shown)


MichaelEischer:
I don't really understand the use case for this PR. Wouldn't `diff --json` be a much better solution to compare snapshots?

wojas:
I agree it's a clunky solution. I have dropped this idea and will close PR.

Issue #2004, opened 2018-09-22
Restic backup of root of a drive letter on windows refuses to restore properly


addos:
Output of `restic version`
--------------------------
restic 0.9.2 compiled with go1.10.3 on windows/amd64

How did you run restic exactly?
-------------------------------
restic -r D:\backups backup E:\
restic -r D:\backups restore <id> -t target_dir

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

Expected behavior
-----------------
It should be able to store/restore properly

Actual behavior
---------------
Restic apparently stores : in the paths, which make the repo unable to be restored?

When I run ls against my repo, it shows something like the following:
/E/E:./repo/.git/objects/info
/E/E:./repo/.git/objects/pack
/E/E:./repo/.git/refs
/E/E:./repo/.git/refs/heads
/E/E:./repo/.git/refs/heads/git-annex
/E/E:./repo/.git/refs/heads/synced
/E/E:./repo/.git/refs/heads/synced/git-annex
/E/E:./repo/.git/refs/heads/synced/master
/E/E:./repo/.git/refs/tags

When I try to restore, it shows something like the following:
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\objects\f6: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\objects\f6\65cfca35e159a0a47237041d1d20edd0218ba3: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\objects\info: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\objects\pack: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\refs: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
ignoring error for C:\Users\brad\a\E\E:.\repo\.git\refs\heads: mkdir C:\Users\brad\a\E\E:.: The filename, directory name, or volume label syntax is incorrect.
[...]

PR #2875, opened 2020-08-05
issue2699: restore symlinks on windows when run as admin user

### LINKED ISSUE #1078: Symlinks to a directory are not detected in Windows
## Output of `restic version`
    restic 0.7.0
    compiled with go1.8.3 on windows/amd64

    restic backup --repo R:\test_backup R:\test

Local.

## Expected behavior
The symlink to a directory should be detected just like symlinks to files are.

## Actual behavior
    warning for test\symlinkdir: invalid node type ""

    restic init --repo R:\test_backup
    restic backup --repo R:\test_backup R:\test

`R:\test` contains a directory called `realdir` and a symlink to it called `symlinkdir` created with `mklink /D symlinkdir realdir` (you'll need an elevated prompt in order to do that). A junction to the directory (`mklink /J`) behaves the same way (what's the difference? [see end of page 431](https://books.google.com/books?id=3rdCAwAAQBAJ&pg=PA431&lpg=PA431&dq=there+are+no+windows+function+for+creating+reparse+points&source=bl&ots=d7zxxodyqG&sig=D9qYu2h69YJwSkkSpU6GffoWxO4&hl=es&sa=X&redir_esc=y#v=onepage&q=there%20are%20no%20windows%20function%20for%20creating%20reparse%20points&f=false)).

Somehow `nodeTypeFromFileInfo()` is returning `""` for a symlink to a directory. It might be a bug in the `os` package and this is more like an upstream thing?

### LINKED ISSUE #2699: Windows: Symlinks are not backuped
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on windows/amd64

```
1. restic -r symlinktest init
2. restic -r symlinktest backup resticsymlinktest
```

This prints
```
repository 17418abd opened successfully, password is correct
created new cache in C:\Users\JsBerbau\AppData\Local\restic

Added to the repo: 343 B

processed 1 files, 31 B in 0:00
snapshot 5f283900 saved
```

The folder resticsymlinktest contained:
File.txt containing

```
Hallo,
I'm the "real" testfile
```

A symlink to that file, created via
`mklink link File.txt`

Tested restore via
`restic -r symlinktest restore -t resticsymlinkrestoretest latest`
Only 1 file was restored, missing the symlink.

Filesystem

Expected behavior
-----------------
The symlink file is included in the backup and restored when restoring.

Actual behavior
---------------
Symlinkfile is not added to backup and thus also not restored.

See above

No

Include symlink files into backup and restore them

Restic is a very great tool and I hope that symlinks are also restored. Then I'm calm that the backup contains all my files and I can recover everything in case of a disaster.

fgma:

Currently symlinks on windows are backed up but not restored. Because of windows restrictions this is only possible with admin privileges. This patch checks if the user running restic has sufficient privileges to restore symlinks and restores them if possible.

For testing the feature you may use the attached demo script: [issue2699.zip](https://github.com/restic/restic/files/5030502/issue2699.zip)

Running demo.bat will do the following:

1. call init.bat: create a demo repository and demo data in the current directory
2. call backup.bat: backup the created data
2. call restore.bat: restore the data to a new folder

Closes #2699
Closes #1078

Checklist
---------

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -14,7 +14,6 @@ import (
 	"github.com/restic/restic/internal/errors"
 
 	"bytes"
-	"runtime"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/fs"
@@ -295,10 +294,6 @@ func (node Node) writeNodeContent(ctx context.Context, repo Repository, f *os.Fi
 }
 
 func (node Node) createSymlinkAt(path string) error {
-	// Windows does not allow non-admins to create soft links.
-	if runtime.GOOS == "windows" {
-		return nil
-	}
 
 	if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return errors.Wrap(err, "Symlink")
--- a/changelog/unreleased/issue-2699
+++ b/changelog/unreleased/issue-2699
@@ -1,0 +1,9 @@
+Bugfix: Restore symbolic links on windows
+
+We've added support to restore symbolic links on windows.
+Because of windows specific restrictions this is only possible when running
+restic having SeCreateSymbolicLinkPrivilege privilege or when running as admin.
+
+https://github.com/restic/restic/issues/1078
+https://github.com/restic/restic/issues/2699
+https://github.com/restic/restic/pull/2875
--- a/doc/050_restore.rst
+++ b/doc/050_restore.rst
@@ -56,6 +56,10 @@ There are case insensitive variants of ``--exclude`` and ``--include`` called
 ``--iexclude`` and ``--iinclude``. These options will behave the same way but
 ignore the casing of paths.
 
+Restoring symbolic links on windows is only possible when the user has
+``SeCreateSymbolicLinkPrivilege`` privilege or is running as admin. This is a
+restriction of windows not restic.
+
 Restore using mount
 ===================
 
--- a/internal/archiver/testing.go
+++ b/internal/archiver/testing.go
@@ -78,10 +78,6 @@ func TestCreateFiles(t testing.TB, target string, dir TestDir) {
 				t.Fatal(err)
 			}
 		case TestSymlink:
-			if runtime.GOOS == "windows" {
-				continue
-			}
-
 			err := fs.Symlink(filepath.FromSlash(it.Target), targetPath)
 			if err != nil {
 				t.Fatal(err)
@@ -139,16 +135,6 @@ func TestEnsureFiles(t testing.TB, target string, dir TestDir) {
 
 	// first, test that all items are there
 	TestWalkFiles(t, target, dir, func(path string, item interface{}) error {
-		// ignore symlinks on Windows
-		if _, ok := item.(TestSymlink); ok && runtime.GOOS == "windows" {
-			// mark paths and parents as checked
-			pathsChecked[path] = struct{}{}
-			for parent := filepath.Dir(path); parent != target; parent = filepath.Dir(parent) {
-				pathsChecked[parent] = struct{}{}
-			}
-			return nil
-		}
-
 		fi, err := fs.Lstat(path)
 		if err != nil {
 			return err
@@ -298,10 +284,6 @@ func TestEnsureTree(ctx context.Context, t testing.TB, prefix string, repo resti
 			}
 			TestEnsureFileContent(ctx, t, repo, nodePrefix, node, e)
 		case TestSymlink:
-			// skip symlinks on windows
-			if runtime.GOOS == "windows" {
-				continue
-			}
 			if node.Type != "symlink" {
 				t.Errorf("tree node %v has wrong type %q, want %q", nodePrefix, node.Type, "file")
 			}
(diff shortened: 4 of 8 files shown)


internal/restic/util_windows.go
```diff
@@ -0,0 +1,39 @@
+package restic
+
+import (
+	"github.com/restic/restic/internal/debug"
+	"golang.org/x/sys/windows"
+)
+
+// IsRunningAsAdminOnWindows is true if the process is running with admin privileges
+var IsRunningAsAdminOnWindows = func() bool {
```
greatroar, line 9: This runs unconditionally for all commands, even though only recover needs the result. As these hacks accumulate, the startup cost of all commands increases. Also, errors are logged but then ignored. It's better written as a function that reads a variable initialized Once:

```
var windowsAdmin struct {
	sync.Once
	isAdmin bool
	err     error
}

func isWindowsAdmin() (bool, error) {
	windowsAdmin.Do(func() {
		windowsAdmin.isAdmin, windowsAdmin.err = /* whatever */
	})
	return windowsAdmin.isAdmin, windowsAdmin.err
}
```

Note that the function need not be public since it is used only inside this package.

internal/restic/node.go
```diff
@@ -297,11 +298,28 @@ func (node Node) writeNodeContent(ctx context.Context, repo Repository, f *os.Fi
...
+		if !IsRunningAsAdminOnWindows {
+			debug.Log(
+				"can't restore symlink to %v without admin privileges on windows",
+				path,
+			)
+			return nil
+		}
+
+		// fix path so a restore to a different location will produce
+		// consistent data
```
greatroar, line 315: Is this necessary? It introduces an incompatbility between Windows and Unix.

internal/restic/util_windows.go
fgma: Was already there in a uncommited version until I "simplified" it. Added it back in https://github.com/restic/restic/pull/2875/commits/a69924fb41920a49681cf4363e0a82e8a4e02ae5.

internal/restic/node.go
fgma: For me it was necessary to restore the symlinks properly, otherwise they would point to invalid locations. This might be related to https://github.com/restic/restic/issues/2004. If this one is done I hope to investigate https://github.com/restic/restic/issues/2004.

hgraeber:
Instead of checking for the Admin User it would be more general to check whether the user has the SeCreateSymbolicLinkPrivilege privilege.

Additional there is the possibility to create symlinks without elevation if developer mode is activated for Windows 10 and CreateSymbolicLink is called with the SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE flag.

internal/restic/node.go
greatroar: On Unix, restore may produce invalid symlinks as well, but I guess the drive letters make this more painful on Windows. I'm not sure what would be the proper behavior here.

fgma:
@hgraeber Implemented in my last commit. Please have a look.

hgraeber:
In principle the check for the check for the privilege looks good to me, but I am a go newbie coming from C/C++.

After thinking more about this, I have doubts whether it is worth making this check to this extent.The only thing it does is getting an error message before creating symlinks. Simply trying to create a symlink an having a error message if it fails should be sufficient. Having the privilege - eg. through an elevated command prompt - is needed in both cases.

So dropping

    // Windows does not allow non-admins to create soft links.
        return nil
    }

in `createSymlinkAt` may be enough to solve the issue.

fgma:
@hgraeber A valid option. I'd add some windows specific code to the error handling to give the user a more specific source of the problem inside the error handling block after calling fs.Symlink(). I'm going to wait for some official feedback on how to continue before putting more work into this PR.

internal/restic/node.go
werbes: > Is this necessary? It introduces an incompatbility between Windows and Unix.

Yes - When doing full system resore of Windows it needs the symbolic links to be restore - else user profiles will not be usable.

werbes (commented):
I have done several restore jobs with this new code on Windows and it gets the symbolic links right when restoring now.

parasiteoflife:
Please, implement this. It's been 2 years

MichaelEischer:
> It's pretty clear restic is just unsupported/abandonware on Windows. It's had data loss inducing bugs with both backup and restore containing symlinks/junctions being broken since release, and unmerged fixes sitting for years.

Windows support for restic has always been primarily community driven, AFAIK there's not much/no windows usage among the maintainers. Nevertheless, I try to also fix Windows-specific problems from time to time. But some pull requests get buried under the large pile of other PRs and then it takes quite some time to dig them out again.

Would you care to elaborate which specific bugs you are referring to?

fgma:
I will look into this in a few days. 

@MichaelEischer What about os.Symlink()? Will this be fixed or should this PR keep the workaround for now?

MichaelEischer:
> Will this be fixed or should this PR keep the workaround for now?

Please change `fs.Symlink()` in `internal/fs/file.go` to
```
return os.Symlink(oldname, fixpath(newname))
```
and remove the workaround afterwards.

MichaelEischer:
@fgma Any progress with cleaning up the PR?

fgma:
@MichaelEischer Looking into it right now.

MichaelEischer:
I've rebased the PR to remove the duplicate commits. The changelog now also contains links to all fixed issues and this PR.

MichaelEischer (approved):
LGTM.

Issue #628, opened 2016-09-19
Repository format v2


fd0:
I'd like to start the discussion on changing the repository format to version 2. This is needed in order to support compression (see #21).

The following list will be updated when new proposals come in.

Accepted:
- Pack files: Move the header to the start of the file. At the moment, the header is at the end. I thought that it'd be nice to just write the file and when that is done write the header. However it turned out that in order to be able to retry failed backend requests, we need to buffer the file locally anyway. So we can write the content (blobs) to a tempfile, and then write the header when uploading the pack file to the backend. This allows reading the header more easily, since we don't need to start from the end of the file.
- Pack files: At the moment the pack file header is a custom binary structure (see [the design  document](https://github.com/restic/restic/blob/master/doc/Design.md#pack-format)). This is inflexible, requires a custom parser and does not allow extension without changing the repository format. I'd like to rebuild the pack header as a JSON data structure, similar to the way the tree objects are stored in the repo. This allows extension without having to change the underlying data format.
- Pack files/Index: When the pack header is changed, add support for compression (algorithm, compressed/uncompressed length). Also add the compressed/uncompressed size to the index files.
- Snapshot files: Allow packed snapshots so that having a lot of snapshots becomes usable (cf #523)
- Add a `README` file into new repositories which describes what this directory contains.
- Remove username and hostname from key files (#2128)

To be discussed:
- Is there a way to add error-correcting codes to the files? Other ideas to recover from data errors?
- Change the Index format to improve memory usage
- Add an encryption indirection: Write down in the header which key is used for authentication/encryption of each blob (so we can implement #187 easier later on)

Postponed/rejected:
- Switch to a faster hash function (SHA3/Keccak/Blake2 instead of SHA256)
- Support asymmetric crypto

Anything else?

rfjakob:
I'm not sure about moving the header to the front. I know this is not currently implemented, but for a local repository, having the header on the end means that we can save a file copy.

fd0:
Interesting point, thanks. I'm not sure yet how to judge what's better. For remote backends we could also then (after a few changes to the backend interface) just pass an `io.Reader` and then maybe the stdlib can use `sendfile` to stream the file directly from disk. hm.

rfjakob:
[...]

PR #2876, opened 2020-08-05
Add repair command

### LINKED ISSUE #1714: file "data_0" has nil blob list
## Output of `restic version`
restic 0.8.3 (v0.8.3-148-g2c76e724) compiled with go1.9.4 on linux/amd64

./restic -r sftp:user@bla:/foo/bar -p pw check

SFTP

## Expected behavior
```
password is correct
load indexes
check all packs
no errors were found
```

## Actual behavior
```
using temporary cache in /tmp/restic-check-cache-157786933
password is correct
load indexes
check all packs
error for tree b75f44fc:
  tree b75f44fc: file "data_0" has nil blob list
```

Run the command.

No. I do not understand the error message.

### LINKED ISSUE #1759: prune: Handle tree not found correctly
Bug reported by a user in the forum:
https://forum.restic.net/t/snapshots-seem-to-be-created-but-i-always-get-a-strange-error/637

## Output of `restic version`

0.8.3

```
42 * * * * source /Users/gio/.restic/restic-env; /usr/local/bin/restic backup -q --exclude-file /Users/gio/.restic/exclude.txt /Users/gio; /usr/local/bin/restic forget -q --prune --keep-hourly 48 --keep-daily 30 --keep-monthly 12
```

GCS

## Expected behavior

Handle the error, complain about it, but complete the process.

## Actual behavior

```
tree 5cb11d839b361e406f30867802f34a7a26b99470b599da5ef36cd209cd4fdb86 not found in repository
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/repository/repository.go:620
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:11
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
        /tmp/restic-build-881482904/src/github.com/restic/restic/internal/restic/find.go:31
main.pruneRepository
        src/github.com/restic/restic/cmd/restic/cmd_prune.go:191
main.runForget
        src/github.com/restic/restic/cmd/restic/cmd_forget.go:235
main.glob..func9
        src/github.com/restic/restic/cmd/restic/cmd_forget.go:24
        /tmp/restic-build-881482904/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:698
        /tmp/restic-build-881482904/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:783
        /tmp/restic-build-881482904/src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:736
main.main
        src/github.com/restic/restic/cmd/restic/main.go:69
runtime.main
        /usr/local/Cellar/go/1.10/libexec/src/runtime/proc.go:198
runtime.goexit
        /usr/local/Cellar/go/1.10/libexec/src/runtime/asm_amd64.s:2361
```

I'm guessing here:

 * Initialize a repo
 * Backup some data
 * Delete a pack file which contains metadata (trees)
 * Run `restic prune`

Error is not correctly handled.

Handle the error, try to salvage as much data as possible.

### LINKED ISSUE #1798: Tree xxxxxx not found in repository and abort
## Output of `restic version`
`restic 0.9.0 compiled with go1.10.2 on linux/amd64`

`restic -r swift:backup:/ forget --group-by paths --keep-last 15 --prune`

OVH object storage

## Expected behavior
Warn about the missing object and continue the execution.

## Actual behavior
```
[...]
2 snapshots have been removed, running prune
counting files in repo
building new index for repo
[1:49] 100.00%  1605 / 1605 packs
repository contains 1605 packs (225480 blobs) with 6.871 GiB
processed 225480 blobs: 3485 duplicate blobs, 4.761 MiB duplicate
load all snapshots
find data that is still in use for 30 snapshots
tree c649f1f83e9be8d420fd29d21cd9fb03f0e0838c86e8bf8fc6877e25738b6bb5 not found in repository
	src/github.com/restic/restic/internal/repository/repository.go:654
	src/github.com/restic/restic/internal/restic/find.go:11
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
	src/github.com/restic/restic/internal/restic/find.go:31
main.pruneRepository
	src/github.com/restic/restic/cmd/restic/cmd_prune.go:191
main.runForget
	src/github.com/restic/restic/cmd/restic/cmd_forget.go:240
main.glob..func10
	src/github.com/restic/restic/cmd/restic/cmd_forget.go:24
	src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:756
	src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:846
	src/github.com/restic/restic/vendor/github.com/spf13/cobra/command.go:794
main.main
	src/github.com/restic/restic/cmd/restic/main.go:86
runtime.main
	/usr/local/go/src/runtime/proc.go:198
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:2361
```

Running `resting forget --prune` against a repo

Restic is really awesome! Thanks for developing this!

### LINKED ISSUE #2334: rebuild-index appears to break the repo, if a data file is missing
Output of `restic version`
--------------------------

Initialize a local repo
`restic init --repo /home/vagrant/repo`

Backup a test dir (clone of the openzfs repo, 85MB)
`restic -r /home/vagrant/repo backup /home/vagrant/test`

Remove a single data file to simulate data loss
`rm /home/vagrant/repo/data/62/62d83f9eba8b34947941ed281ef9131887fd499d39b566c4c57f729093b2e071`

`restic -r /home/vagrant/repo check`
Correctly complains about a missing file. But, I am still able to restore data from the repo, which is great (minus some of the files, obviously)

The problem starts here. Running rebuild-index appears to break the repo
`restic -r /home/vagrant/repo rebuild-index`
completes without error

A restore will now panic
`restic -r /home/vagrant/repo restore latest --target /home/vagrant/restore/`
repository 2fdae5c4 opened successfully, password is correct
restoring <Snapshot 765052de of [/recover] at 2019-07-12 12:56:59.426074427 +0000 UTC by vagrant@test> to /home/vagrant/restore/
Unknown blob 43dfd85e0eac10d828a4b596b27be76006d1c085838cdda2988276ee5e4dec1f
github.com/restic/restic/internal/restorer.(*filePackTraverser).forEachFilePack.func1
	/restic/internal/restorer/filepacktraverser.go:21
github.com/restic/restic/internal/restorer.(*filePackTraverser).forEachFilePack
	/restic/internal/restorer/filepacktraverser.go:32
github.com/restic/restic/internal/restorer.newPackQueue
	/restic/internal/restorer/packqueue.go:52
github.com/restic/restic/internal/restorer.(*fileRestorer).restoreFiles
	/restic/internal/restorer/filerestorer.go:118
github.com/restic/restic/internal/restorer.(*Restorer).RestoreTo
	/restic/internal/restorer/restorer.go:252
main.runRestore
[..]

prune fails
repository 2fdae5c4 opened successfully, password is correct
counting files in repo
building new index for repo
[0:00] 100.00%  17 / 17 packs
repository contains 17 packs (3390 blobs) with 71.737 MiB
processed 3390 blobs: 0 duplicate blobs, 0 B duplicate
load all snapshots
find data that is still in use for 1 snapshots
[0:00] 100.00%  1 / 1 snapshots
Fatal: number of used blobs is larger than number of available blobs!
Please report this error (along with the output of the 'prune' run) at
https://github.com/restic/restic/issues/new

recover completes succesfully, but attempting to restore from the snapshot it creates also fails. Running a new backup shows no error, but it will again panic if you attempt to restore from it.

Reproduced it on a local repository, although I found the same issue while testing against an openstack swift repo.

Expected behavior
-----------------
Be able to recover the rest of the data from the repository, and ideally, be able to re-upload only the missing data, not re-create the whole repo.

Actual behavior
---------------
Cannot restore data, or fix the repo (new backups will run without errors, but you're still unable to restore from these new snapshots)

As described in the "how did you run restic" section

Deleting a file within the "data" directory to simulate data loss.

No

Just testing so far, hopefully I did not miss anything obvious. I'll be happy to help troubleshoot this, if you think it is worth investigating.

### LINKED ISSUE #2683: doc: Add troubleshooting section
The current documentation at [restic.readthedocs.io](https://restic.readthedocs.io/) does not contain a proper troubleshooting section, where users can find common problems and how to (initially) handle them. This needs to be addressed.

I propose the following:

- Use [the FAQ](https://restic.readthedocs.io/en/latest/faq.html) as the starting point.
- Replace the items in the FAQ that are about specific error messages with a reference to the new troubleshooting section.
- Make sure to cover what's currently in the FAQ in the new troubleshooting section.
- Make sure that the troubleshooting section references the forum as a place to get further help if the troubleshooting section isn't sufficient, and the issue tracker as a place to report what the user believes is a bug.
- Once we have the troubleshooting section in place, in the various commands of restic, update the current references to creating a new issue/bug report, so that they instead point to the troubleshooting section in the documentation.

This issue supersedes issues #1813 and #2361.

aawsome:

Allow users to recover from broken repositories/snapshots while still salvaging the sane parts of the repository/snapshot.

For given snapshots (selection identical to, e.g., `forget`) the command tries to read all trees and checks if the needed blobs are contained in the index. If blobs are missing or trees cannot be read, it will create new trees and snapshots which only miss these "defect" parts.
Those newly generated snapshots can be used to recover needed data.
Also, after removing the "defect" snapshots, `prune` is able to clean up the repo again.

While this command is able to cause data loss, special care is taken such that the default flags won't
do any harm - in fact, users have to explicitly specify `--dry-run=false --delete` to loose data.

Output looks like:
```
./restic -r /home/thinkpad/repo.index-missingblob2/repo repair 

 note: --dry-run is set
-> repair will only show what it would do.

repository b270637f opened successfully, password is correct
check and repair 1 snapshots
<Snapshot f22c6d3a of [/home/thinkpad/data] at 2020-07-09 11:18:26.501071439 +0200 CEST by alex@thinkpad>:
removed defect file '/home/thinkpad/data/test'
would have modified tree 705adc0d
would have modified tree 1ee0d0e1
would have modified tree 98c948be
would have modified tree 9d89d7fe
would have repaired snpshot f22c6d3a.
[0:00] 100.00%  1 / 1 snapshots
```

Depends on #2878 for the troubleshooting docu update.

Closes #1759 
Closes #1798
Closes #2334 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_repair.go
+++ b/cmd/restic/cmd_repair.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var cmdRepair = &cobra.Command{
+	Use:   "repair",
+	Short: "Repair the repository",
+}
+
+func init() {
+	cmdRoot.AddCommand(cmdRepair)
+}
--- a/cmd/restic/cmd_repair.go
+++ b/cmd/restic/cmd_repair.go
@@ -84,8 +84,7 @@ func runRepair(opts RepairOptions, args []string) error {
 	}
 
 	lock, err := lockRepoExclusive(globalOptions.ctx, repo)
-	// to make linter happy, as unlockRepo returns an error (which is ignored)
-	defer func() { _ = unlockRepo(lock) }()
+	defer unlockRepo(lock)
 	if err != nil {
 		return err
 	}
@@ -115,9 +114,11 @@ func repairSnapshots(opts RepairOptions, repo restic.Repository, snapshots []*re
 	for _, sn := range snapshots {
 		debug.Log("process snapshot %v", sn.ID())
 		Printf("%v:\n", sn)
-		newID, changed, err := repairTree(opts, repo, "/", *sn.Tree, replaces, seen)
+		newID, changed, lErr, err := repairTree(opts, repo, "/", sn.Tree, replaces, seen)
 		switch {
 		case err != nil:
+			return err
+		case lErr:
 			Printf("the root tree is damaged -> delete snapshot.\n")
 			deleteSn.Insert(*sn.ID())
 		case changed:
@@ -153,13 +154,13 @@ func repairSnapshots(opts RepairOptions, repo restic.Repository, snapshots []*re
 // - add the rag opts.AddTag
 // - preserve original ID
 // if opts.DryRun is set, it doesn't change anything but only
-func changeSnapshot(opts RepairOptions, repo restic.Repository, sn *restic.Snapshot, newID restic.ID) error {
+func changeSnapshot(opts RepairOptions, repo restic.Repository, sn *restic.Snapshot, newID *restic.ID) error {
 	sn.AddTags([]string{opts.AddTag})
 	// Retain the original snapshot id over all tag changes.
 	if sn.Original == nil {
 		sn.Original = sn.ID()
 	}
-	sn.Tree = &newID
+	sn.Tree = newID
 	if !opts.DryRun {
 		newID, err := repo.SaveJSONUnpacked(globalOptions.ctx, restic.SnapshotFile, sn)
 		if err != nil {
@@ -175,29 +176,43 @@ func changeSnapshot(opts RepairOptions, repo restic.Repository, sn *restic.Snaps
 type idMap map[restic.ID]restic.ID
 
 // repairTree checks and repairs a tree and all its subtrees
-// Two error cases are checked:
+// Three error cases are checked:
+// - tree is a nil tree (-> will be replaced by an empty tree)
 // - trees which cannot be loaded (-> the tree contents will be removed)
 // - files whose contents are not fully available  (-> file will be modified)
 // In case of an error, the changes made depends on:
 // - opts.Append: string to append to "repared" names; if empty files will not repaired but deleted
 // - opts.DryRun: if set to true, only print out what to but don't change anything
-func repairTree(opts RepairOptions, repo restic.Repository, path string, treeID restic.ID, replaces idMap, seen restic.IDSet) (restic.ID, bool, error) {
+// Returns:
+// - the new ID
+// - whether the ID changed
+// - whether there was a load error when loading this tre
+// - error for other errors (these are errors when saving a tree)
+func repairTree(opts RepairOptions, repo restic.Repository, path string, treeID *restic.ID, replaces idMap, seen restic.IDSet) (*restic.ID, bool, bool, error) {
 	ctx := globalOptions.ctx
 
+	// handle and repair nil trees
+	if treeID == nil {
+		empty, err := emptyTree(opts.DryRun, repo)
+		Printf("repaired nil tree '%v'\n", path)
+		return &empty, true, false, err
+	}
+
 	// check if tree was already changed
-	newID, ok := replaces[treeID]
+	newID, ok := replaces[*treeID]
 	if ok {
-		return newID, true, nil
+		return &newID, true, false, nil
 	}
 
 	// check if tree was seen but not changed
-	if seen.Has(treeID) {
-		return treeID, false, nil
+	if seen.Has(*treeID) {
+		return treeID, false, false, nil
 	}
 
-	tree, err := repo.LoadTree(ctx, treeID)
+	tree, err := repo.LoadTree(ctx, *treeID)
 	if err != nil {
-		return newID, false, err
+		// mark as load error
+		return &newID, false, true, nil
 	}
 
 	var newNodes []*restic.Node
@@ -231,17 +246,23 @@ func repairTree(opts RepairOptions, repo restic.Repository, path string, treeID 
 			}
 		case "dir":
 			// rewrite if necessary
-			newID, c, err := repairTree(opts, repo, path+node.Name+"/", *node.Subtree, replaces, seen)
+			newID, c, lErr, err := repairTree(opts, repo, path+node.Name+"/", node.Subtree, replaces, seen)
 			switch {
 			case err != nil:
+				return newID, true, false, err
+			case lErr:
 				// If we get an error, we remove this subtree
 				changed = true
 				Printf("removed defect dir '%v'", path+node.Name)
 				node.Name = node.Name + opts.Append
 				Printf("(now emtpy '%v')\n", node.Name)
-				node.Subtree = nil
+				empty, err := emptyTree(opts.DryRun, repo)
+				if err != nil {
+					return newID, true, false, err
+				}
+				node.Subtree = &empty
 			case c:
-				node.Subtree = &newID
+				node.Subtree = newID
 				changed = true
 			}
 		}
@@ -249,8 +270,8 @@ func repairTree(opts RepairOptions, repo restic.Repository, path string, treeID 
 	}
 
 	if !changed {
-		seen.Insert(treeID)
-		return treeID, false, nil
+		seen.Insert(*treeID)
+		return treeID, false, false, nil
 	}
 
 	tree.Nodes = newNodes
@@ -258,13 +279,22 @@ func repairTree(opts RepairOptions, repo restic.Repository, path string, treeID 
 	if !opts.DryRun {
 		newID, err = repo.SaveTree(ctx, tree)
 		if err != nil {
-			return newID, false, err
+			return &newID, true, false, err
 		}
 		Printf("modified tree %v, new id: %v\n", treeID.Str(), newID.Str())
 	} else {
 		Printf("would have modified tree %v\n", treeID.Str())
 	}
 
-	replaces[treeID] = newID
-	return newID, true, nil
+	replaces[*treeID] = newID
+	return &newID, true, false, nil
+}
+
+func emptyTree(dryRun bool, repo restic.Repository) (restic.ID, error) {
+	ctx := globalOptions.ctx
+	var tree restic.Tree
+	if !dryRun {
+		return repo.SaveTree(ctx, &tree)
+	}
+	return restic.ID{}, nil
 }
--- a/cmd/restic/cmd_repair.go
+++ b/cmd/restic/cmd_repair.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"context"
+
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/restic"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/spf13/cobra"
 )
@@ -41,15 +44,14 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runRepair(repairOptions, args)
+		return runRepair(cmd.Context(), globalOptions, repairOptions, args)
 	},
 }
 
 // RepairOptions collects all options for the repair command.
 type RepairOptions struct {
-	Hosts           []string
-	Paths           []string
-	Tags            restic.TagLists
+	restic.SnapshotFilter
+
 	AddTag          string
 	Append          string
 	DryRun          bool
@@ -61,16 +63,16 @@ var repairOptions RepairOptions
 func init() {
 	cmdRoot.AddCommand(cmdRepair)
 	flags := cmdRepair.Flags()
-	flags.StringArrayVarP(&repairOptions.Hosts, "host", "H", nil, `only consider snapshots for this host (can be specified multiple times)`)
-	flags.Var(&repairOptions.Tags, "tag", "only consider snapshots which include this `taglist`")
-	flags.StringArrayVar(&repairOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`")
+
+	initMultiSnapshotFilter(flags, &repairOptions.SnapshotFilter, true)
+
 	flags.StringVar(&repairOptions.AddTag, "add-tag", "repaired", "tag to add to repaired snapshots")
 	flags.StringVar(&repairOptions.Append, "append", ".repaired", "string to append to repaired dirs/files; remove files if empty or impossible to repair")
 	flags.BoolVarP(&repairOptions.DryRun, "dry-run", "n", true, "don't do anything, only show what would be done")
 	flags.BoolVar(&repairOptions.DeleteSnapshots, "delete-snapshots", false, "delete original snapshots")
 }
 
-func runRepair(opts RepairOptions, args []string) error {
+func runRepair(ctx context.Context, gopts GlobalOptions, opts RepairOptions, args []string) error {
 	switch {
 	case opts.DryRun:
 		Printf("\n note: --dry-run is set\n-> repair will only show what it would do.\n\n")
@@ -78,64 +80,67 @@ func runRepair(opts RepairOptions, args []string) error {
 		Printf("\n note: --dry-run is not set and --delete-snapshots is set\n-> this may result in data loss!\n\n")
 	}
 
-	repo, err := OpenRepository(globalOptions)
+	repo, err := OpenRepository(ctx, globalOptions)
 	if err != nil {
 		return err
 	}
 
-	lock, err := lockRepoExclusive(globalOptions.ctx, repo)
+	lock, ctx, err := lockRepoExclusive(ctx, repo, gopts.RetryLock, gopts.JSON)
 	defer unlockRepo(lock)
 	if err != nil {
 		return err
 	}
 
-	if err := repo.LoadIndex(globalOptions.ctx); err != nil {
+	if err := repo.LoadIndex(ctx); err != nil {
 		return err
 	}
 
 	// get snapshots to check & repair
 	var snapshots []*restic.Snapshot
-	for sn := range FindFilteredSnapshots(globalOptions.ctx, repo.Backend(), repo, opts.Hosts, opts.Tags, opts.Paths, args) {
+	for sn := range FindFilteredSnapshots(ctx, repo.Backend(), repo, &opts.SnapshotFilter, args) {
 		snapshots = append(snapshots, sn)
 	}
 
-	return repairSnapshots(opts, repo, snapshots)
+	return repairSnapshots(ctx, opts, repo, snapshots)
 }
 
-func repairSnapshots(opts RepairOptions, repo restic.Repository, snapshots []*restic.Snapshot) error {
-	ctx := globalOptions.ctx
-
+func repairSnapshots(ctx context.Context, opts RepairOptions, repo restic.Repository, snapshots []*restic.Snapshot) error {
 	replaces := make(idMap)
 	seen := restic.NewIDSet()
 	deleteSn := restic.NewIDSet()
 
 	Verbosef("check and repair %d snapshots\n", len(snapshots))
 	bar := newProgressMax(!globalOptions.Quiet, uint64(len(snapshots)), "snapshots")
-	for _, sn := range snapshots {
-		debug.Log("process snapshot %v", sn.ID())
-		Printf("%v:\n", sn)
-		newID, changed, lErr, err := repairTree(opts, repo, "/", sn.Tree, replaces, seen)
-		switch {
-		case err != nil:
-			return err
-		case lErr:
-			Printf("the root tree is damaged -> delete snapshot.\n")
-			deleteSn.Insert(*sn.ID())
-		case changed:
-			err = changeSnapshot(opts, repo, sn, newID)
-			if err != nil {
+	wg, ctx := errgroup.WithContext(ctx)
+	repo.StartPackUploader(ctx, wg)
+	wg.Go(func() error {
+		for _, sn := range snapshots {
+			debug.Log("process snapshot %v", sn.ID())
+			Printf("%v:\n", sn)
+			newID, changed, lErr, err := repairTree(ctx, opts, repo, "/", sn.Tree, replaces, seen)
+			switch {
+			case err != nil:
 				return err
+			case lErr:
+				Printf("the root tree is damaged -> delete snapshot.\n")
+				deleteSn.Insert(*sn.ID())
+			case changed:
+				err = changeSnapshot(ctx, opts, repo, sn, newID)
+				if err != nil {
+					return err
+				}
+				deleteSn.Insert(*sn.ID())
+			default:
+				Printf("is ok.\n")
 			}
-			deleteSn.Insert(*sn.ID())
-		default:
-			Printf("is ok.\n")
+			debug.Log("processed snapshot %v", sn.ID())
+			bar.Add(1)
 		}
-		debug.Log("processed snapshot %v", sn.ID())
-		bar.Add(1)
-	}
-	bar.Done()
+		bar.Done()
+		return repo.Flush(ctx)
+	})
 
-	err := repo.Flush(ctx)
+	err := wg.Wait()
 	if err != nil {
 		return err
 	}
@@ -143,7 +148,7 @@ func repairSnapshots(opts RepairOptions, repo restic.Repository, snapshots []*re
 	if len(deleteSn) > 0 && opts.DeleteSnapshots {
 		Verbosef("delete %d snapshots...\n", len(deleteSn))
 		if !opts.DryRun {
-			DeleteFiles(globalOptions, repo, deleteSn, restic.SnapshotFile)
+			DeleteFiles(ctx, globalOptions, repo, deleteSn, restic.SnapshotFile)
 		}
 	}
 	return nil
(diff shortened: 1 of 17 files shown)


MichaelEischer:
Could we please discuss _first_ how such a command should look like? How does this relate to the rewrite command in #2731 ? What if the prune command could handle certain types of repository damages? Which types of repair commands for a repository do we need and how to avoid cluttering the CLI?

Btw, with the current flood of new pull requests and the still huge pile of older ones, it will quite likely take some time until anyone of the maintainers has time to participate in that discussion.

This PR does not fix the underlying issue of #2334, it merely removes the symptoms (the broken file). But it's still not possible to restore at least the remaining parts of a file.

aawsome:
I thought there would be some changes appearing in near future on either `rewrite` or `prune`, so I waited with an answer.

So, basically the point is that restic is missing such kind of functionality and IMO this is critical as cases has already been reported where users are struggling with broken repos which can only be fixed by forgetting and pruning whole snapshots. This is why I proposed this PR.

About the name and where to put such functionality we can and should of course discuss and I'm fine to have the discussion here. I agree that the functionality is pretty similar to the `rewrite` command proposed in #2720 and maybe adding some flags there would solve the issue similarly. I however don't think that this functionality should be included into `prune`: `prune` removes data and should start only if the repository is sane. This functionality is proposed to help users the possibly best way if their repository is no sane to put it back into a sane state.
For now, we already do have a `recover` command which somehow stands beside the other restic commands and seems to be a "quick" solution for a problem that occured. So I used a new command in this "tradition".

I must admit that there is another reason why I used a new command `repair` and it is the situation you depicted, namely that I must take into account that it might take a long time before this PR is getting discussed or merged. In this situation, I really do want to have a patch that does not depend on another WIP patch (i.e. #2720) or is expected to need rebasing after some other PRs get merged into master. This is a patch that is able to help users who need a quick solution for a broken repo and therefore should stay easily mergeable. This can be best guaranteed by only adding new files to the source code.

> This PR does not fix the underlying issue of #2334, it merely removes the symptoms (the broken file). But it's > still not possible to restore at least the remaining parts of a file.

If `--append` is given (defaults to ".repaired") this `repair` command will generate a " file.repaired" which contains the remaining parts of that file. Only if no parts are left to repair the file, it will delete the file (as shown in the example output where `/home/thinkpad/data/test` was a 1-blob file and that blob was missing in the repo).

MichaelEischer:
> So, basically the point is that restic is missing such kind of functionality and IMO this is critical as cases has already been reported where users are struggling with broken repos which can only be fixed by forgetting and pruning whole snapshots. This is why I proposed this PR.

It's definitely useful to have a WIP PR to point people to to repair damaged repositories.

> I agree that the functionality is pretty similar to the rewrite command proposed in #2731 and maybe adding some flags there would solve the issue similarly.

From an implementation perspective both commands are pretty similar so it would be useful to make the recursive tree rewriting part reusable for both commands. (Assuming we end up with two separate commands)

> I however don't think that this functionality should be included into prune

With "handle certain types of repository damages" I had in mind that it should be possible to let the prune command tolerate (= be able to remove unused blobs/packs) repositories which e.g. miss a few data blobs. Of course, the prune command shouldn't start to modify snapshots.

> If --append is given (defaults to ".repaired") this repair command will generate a " file.repaired" which contains the remaining parts of that file. Only if no parts are left to repair the file, it will delete the file (as shown in the example output where /home/thinkpad/data/test was a 1-blob file and that blob was missing in the repo).

I've missed the `--append` option and yes it helps a lot if blob sizes are missing, but I'd still consider this as fixing the symptoms. The underlying problem I had in mind is the repository format itself: it can't store the size of missing blobs without risking further damaged snapshots, but when rebuilding the index to fix this we loose the information about the position of the data blobs.

aawsome:
> With "handle certain types of repository damages" I had in mind that it should be possible to let the prune command tolerate (= be able to remove unused blobs/packs) repositories which e.g. miss a few data blobs. Of course, the prune command shouldn't start to modify snapshots.

You are right - in general missing data blobs are not a that big issue when thinking about the sanity of the repository. Of course, data might be lost which might be a bad thing for users. But starting with a repository where you have missing data blobs, probably the best is actually to remove them from the index but let the trees still reference them. If another backup run will stumble over the missing data, the repository will be healed. And if not, all functionality is working perfectly except if you need to access that specific data, i.e. a file which contains a missing blob. (and even restore and mount could work around this and include the functionality I propose here for missing blobs)

Another issue is missing tree blobs. This severely damages the repository as most commands like `check` and `prune` will no longer work. For this kind of damage IMO it is best to modify trees and snapshots to remove these damaged subtrees which is a much better solution than complete delete the snapshot.

> I've missed the `--append` option and yes it helps a lot if blob sizes are missing, but I'd still consider this as fixing the symptoms. The underlying problem I had in mind is the repository format itself: it can't store the size of missing blobs without risking further damaged snapshots, but when rebuilding the index to fix this we loose the information about the position of the data blobs.

I recently thought about a check when loading the index whether the pack files referenced in the index still exist in the backend. This would allow users to just remove damaged packs (missing packs would be already detected) and then an information in the index could be saved whether the blob is present or only the length saved in the index may be used. I rejected to implement this as a regular check as it would mean that every 'LoadIndex' would need to 'List' the pack files which could be expensive. Maybe that might be a direction to think of?

MichaelEischer:
> I recently thought about a check when loading the index whether the pack files referenced in the index still exist in the backend. This would allow users to just remove damaged packs (missing packs would be already detected) and then an information in the index could be saved whether the blob is present or only the length saved in the index may be used. I rejected to implement this as a regular check as it would mean that every 'LoadIndex' would need to 'List' the pack files which could be expensive. Maybe that might be a direction to think of?

I don't think we want to list all pack files every time the repository index is loaded, it really sounds (needlessly) expensive. It would also only be a partial workaround, as having to rebuild the repository index would still loose the information. And we'd also get into a lot of trouble when mixing older and newer restic version which have differing opinions on which blobs actually exist and which don't. The only way to properly fix this is to handle this in a repository format v2 #628 .

pddp:
For the second time I have a corrupted repo - too bad. Probably the reason was that the backup disk was full.
I tried ./restic -r$REPO repair and it proposes fixes, as expected.

However, when I run 'make test' I get 
?   	github.com/restic/restic/internal/limiter	[no test files]
?   	github.com/restic/restic/internal/migrations	[no test files]
?   	github.com/restic/restic/internal/mock	[no test files]
--- FAIL: TestOptionsApplyInvalid (0.00s)
    --- FAIL: TestOptionsApplyInvalid/test-2 (0.00s)
        options_test.go:216: expected error "time: missing unit in duration 2134", got "time: missing unit in duration \"2134\""
FAIL
FAIL	github.com/restic/restic/internal/options	0.014s
ok  	github.com/restic/restic/internal/pack	0.018s

other tests are o.k.

Tested on ARCH 5.8.7-arch1-1

aawsome:
Hi @pddp 
thanks for reporting the issue!

> FAIL github.com/restic/restic/internal/options 0.014s

Actually, this PR doesn't change anything on `/internal/options`. Did you redo your test with master? Seems to me that this might be an uncorrelated issue..

pddp:
Hi @aawsome 

your guess was correct - this was uncorrelated. Also happens on master.

pddp (changes requested):
Just a minor thing: '--delete' vs '--delete-snapshots'.
Options should be mentioned in the --help text.

Other than that: works as expected.

huyz:
My repo seems to be corrupted after I ran `rebuild-index`. Is it a bad idea to check out this branch and invoke the `repair` command?

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	sn.Tree = &newID
+	if !opts.DryRun {
+		newID, err := repo.SaveJSONUnpacked(globalOptions.ctx, restic.SnapshotFile, sn)
+		if err != nil {
+			return err
+		}
+		Printf("snapshot repaired -> %v created.\n", newID.Str())
+	} else {
+		Printf("would have repaired snpshot %v.\n", sn.ID().Str())
```
huyz, line 170: Typo
```suggestion
		Printf("would have repaired snapshot %v.\n", sn.ID().Str())
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+	flags.BoolVarP(&repairOptions.DryRun, "dry-run", "n", true, "don't do anything, only show what would be done")
+	flags.BoolVar(&repairOptions.DeleteSnapshots, "delete-snapshots", false, "delete original snapshots")
+}
+
+func runRepair(opts RepairOptions, args []string) error {
+	switch {
+	case opts.DryRun:
+		Printf("\n note: --dry-run is set\n-> repair will only show what it would do.\n\n")
+	case opts.DeleteSnapshots:
+		Printf("\n note: --dry-run is not set and --delete is set\n-> this may result in data loss!\n\n")
```
huyz, line 78:
```suggestion
		Printf("\n note: --dry-run is not set and --delete-snapshots is set\n-> this may result in data loss!\n\n")
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+					newSize += uint64(size)
+				}
+			}
+			if !ok {
+				changed = true
+				if opts.Append == "" || newSize == 0 {
+					Printf("removed defect file '%v'\n", path+node.Name)
+					continue
+				}
+				Printf("repaired defect file '%v'", path+node.Name)
```
huyz, line 226:
```suggestion
				Printf("repaired defective file '%v'", path+node.Name)
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+				node.Size = newSize
+			}
+		case "dir":
+			// rewrite if necessary
+			newID, c, err := repairTree(opts, repo, path+node.Name+"/", *node.Subtree, replaces, seen)
+			switch {
+			case err != nil:
+				// If we get an error, we remove this subtree
+				changed = true
+				Printf("removed defect dir '%v'", path+node.Name)
```
huyz, line 239:
```suggestion
				Printf("removed defective dir '%v'", path+node.Name)
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+					ok = false
+				} else {
+					newContent = append(newContent, id)
+					newSize += uint64(size)
+				}
+			}
+			if !ok {
+				changed = true
+				if opts.Append == "" || newSize == 0 {
+					Printf("removed defect file '%v'\n", path+node.Name)
```
huyz, line 223:
```suggestion
					Printf("removed defective file '%v'\n", path+node.Name)
```

doc/077_troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+
+You might even need to manually remove corrupted pack files. In this case make sure, you run 
+`restic rebuild-index` after.
+
+Also if you encounter problems with the index files itselves, `rebuild-index` will solve these problems
+immediately.
+
+However, rebuilding the index does not solve every problem, e.g. lost pack files.
+
+4. Delete unneeded defect snapshots -> `forget`
```
huyz, line 76:
```suggestion
4. Delete unneeded defective snapshots -> `forget`
```

doc/077_troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+********************************************
+
+If you encounter defect snapshots but realize you can spare them, it is often a good idea to simply
```
huyz, line 79:
```suggestion
If you encounter defective snapshots but realize you can spare them, it is often a good idea to simply
```

doc/077_troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+delete them using `forget`. In case that your repository remains with just sane snapshots (including
+all trees and files) the next `prune` run will put your repository in a sane state.
+
+This can be also used if you manage to create new snapshots which can replace the defect ones, see
```
huyz, line 83:
```suggestion
This can be also used if you manage to create new snapshots which can replace the defective ones, see
```

doc/077_troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+********************************************
+
+If for some reason you have unreferenced trees in your repository but you actually need them, run
+`recover` it will generate a new snapshot which allows access to all trees that you have in your
+repository. 
+
+Note that `recover` relies on a correct index state, so make sure your index is fine or run `rebuild-index`
+before running `recover`.
+
+7. Repair defect snapshots using `repair`
```
huyz, line 107:
```suggestion
7. Repair defective snapshots using `repair`
```

doc/077_troubleshooting.rst
```diff
@@ -0,0 +1,111 @@
...
+********************************************
+
+If all other things did not help, you can repair defect snapshots with `repair`. Note that the repaired
+snapshots will miss data which was referenced in the defect snapshot.
```
huyz, line 111:
```suggestion
If all other things did not help, you can repair defective snapshots with `repair`. Note that the repaired
snapshots will miss data which was referenced in the defective snapshot.
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+var repairOptions RepairOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdRepair)
+	flags := cmdRepair.Flags()
+	flags.StringArrayVarP(&repairOptions.Hosts, "host", "H", nil, `only consider snapshots for this host (can be specified multiple times)`)
+	flags.Var(&repairOptions.Tags, "tag", "only consider snapshots which include this `taglist`")
+	flags.StringArrayVar(&repairOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`")
+	flags.StringVar(&repairOptions.AddTag, "add-tag", "repaired", "tag to add to repaired snapshots")
+	flags.StringVar(&repairOptions.Append, "append", ".repaired", "string to append to repaired dirs/files; remove files if emtpy or impossible to repair")
```
huyz, line 68:
```suggestion
	flags.StringVar(&repairOptions.Append, "append", ".repaired", "string to append to repaired dirs/files; remove files if empty or impossible to repair")
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+		case "dir":
+			// rewrite if necessary
+			newID, c, err := repairTree(opts, repo, path+node.Name+"/", *node.Subtree, replaces, seen)
+			switch {
+			case err != nil:
+				// If we get an error, we remove this subtree
+				changed = true
+				Printf("removed defect dir '%v'", path+node.Name)
+				node.Name = node.Name + opts.Append
+				Printf("(now emtpy '%v')\n", node.Name)
```
huyz, line 241:
```suggestion
				Printf("(now empty '%v')\n", node.Name)
```

cmd/restic/cmd_repair.go
```diff
@@ -0,0 +1,270 @@
...
+				node.Subtree = nil
```
MichaelEischer, line 242: While looking at #3288 I've noticed that the crash there is actually caused by this line. A directory must always refer to a valid subtree, at least that's assumed by most code parts. The correct subtree here would be an empty one `{}`.

huyz: Ah that would explain my remaining errors https://forum.restic.net/t/getting-rid-of-nil-subtrees/3602

aawsome: Is corrected now

aawsome:
I also rebased to current master and resolved all typos in `cmd_repair.go` - thanks @huyz!

I'll keep the `troubleshooting.rst` unchanged, is is mainly a topic of #2878 (I think this will be removed from this PR once this is reviewed)

cmd/restic/cmd_repair.go
huyz: Great, thanks!

huyz:
I re-rain the `repair` command and it cleared up all my errors. Thanks!

kradalby:
Any updates on this? the command has saved me a couple of times and it would be neat to have it mainlined :)

noduck:
This works for me too, I repaired three B2 repos.
Unfortunately, I had some instances of "the root tree is damaged -> delete snapshot". A cursory investigation showed that these snapshots where not recoverable. I didn't do a deep-dive (I have alternative backups, and these were older snapshots), and just got rid of the errand snapshots.

seqizz:
Just another success story to add. Deleted some files by mistake from a repository with 27 snapshots. Neither prune nor rebuild-index was able to do any help, also I couldn't find anything convenient in the options. Didn't try `recover` since my snapshots folder did not have any losses. Without `repair`, I had to trace all missing packs and forget related snapshots I assume. I tried this one as:
- Ran backup with `--force` option on repo's 3 clients
- Ran `rebuild-index` on the repo
- Ran `repair --dry-run=false --delete-snapshots`
- Ran `check --read-data` (succeeded)
Ended up with 22 snapshots, 20 of them has the tag `repaired`

Thanks for this @aawsome

patrickuhlmann:
I also used the command successfully. In my setup I used OneDrive over rclone as target and I had about 160GB and 14 snapshots. Before I executed the repair I was not able to run prune because it was missing a blob. I already found out that the file contained was just some cache file I was not interested but it was present in all snapshots.

The command took a long time to execute (about 45 hours) but was able to keep all snapshots and prune ran fine afterwards.

I really think this should be merged. There is no other reasonable possibility to repair in such cases. That doesn't involve command line hacking and examining and running debug commands with your own build etc.

Here the steps I took to run it:
```
git clone https://github.com/aawsome/restic.git
cd restic
git checkout new-repair-command
make
./restic -r /myrepo repair 
./restic -r /myrepo repair --dry-run=false --delete-snapshots
```

MichaelEischer:
I've rebased the PR such that it includes support for repository format version 2 repositories.

jim3ma:
Very useful PR. It help me repair my 10Ti+ restic backup with 100+ snapshots.
After repaired, I can force backup to corrupted files.

JaCoB1123:
I recently tried this branch as well and had a special case:
When pruning I got the error that a packfiles hash does not match the expected hash. So I ran the `repair` command, but it didn't show anything to repair.
So I just fell back to the alternative route: Deleting the problematic packfiles. Which - to my surprise - yielded error-free `check` and `prune` runs.
I'm not sure if it's worth handling that case and if so, if the `repair` command should be extended or the `prune` command should be smarter (not erroring on unneeded packs).

jim3ma:
@aawsome Any updates ?
Some actions failed, can we fix it ?

aawsome:
> @aawsome Any updates ? Some actions failed, can we fix it ?

Yes, seems that some Windows tests are failing. I am speculating as there are no logs available any longer, but there must be some incompatible changes in master.

My priority, however, ATM is to improve [rustic](https://github.com/rustic-rs/rustic), where this functionality is currently in development (see https://github.com/rustic-rs/rustic/pull/258) and I'm pretty sure it will be released in near future. Note that you can use rustic to access and modify a restic repository.

akrabu:
Just wanted to point out that the --no-lock switch seems to not work with repair.  I have a read-only repository and I was just curious what a dry run of repair would do, but it insisted on creating the lock despite --no-lock and wouldn't work.  Might be by design, and I'd understand if so - but I thought I'd mention it!

aawsome:
@akrabu You are right. There is no special treatment for `--dry-run` with respect to locking and the standard is to lock the repo in restic.

But you can try out `rustic repair snapshots --dry-run` (from rustic main branch) on your repository - [rustic](https://github.com/rustic-rs/rustic) works completely lock-free and never creates a lock file.

MichaelEischer:
I've updated and massively refactored the PR:
- the command has moved to `repair snapshots`
- `rebuild-index` has moved to `repair index`
- `repair snapshots` now has largely the same options as `rewrite`. The CLI output is also rather similar. In fact, both commands now share significant parts of their implementation.
- rewritten troubleshooting section. It is now more of a step by step guide. I might have overdone it with encouraging users to ask for help if necessary...

Repaired files are currently not suffixed as this could break the invariant the tree nodes must be sorted in lexicographical order. Like the `rewrite` command, `--dry-run` also works without requesting a lock.

Current todos:
- [ ] Check the code / documentation again
- [ ] Update changelog
- [ ] Add basic tests for the `repair snapshots` command

MichaelEischer:
LGTM.

PR #2880, opened 2020-08-07
Improve recover command

aawsome:

Enhance the `recover` command to make it useful in more recovery situations:

- only save data not referenced by any snapshot
- dont't write an empty snapshot
- use progress bar for scanning trees

OLD AND REMOVED:

> 
> - allow to combine all found directories in one snapshot (equals the current behavior; default) or make one snapshot per "hanging" root tree
> - allow to add user-defined tags to newly generated snapshots
> - add dry-run possibility
> 
> Especially the flags `--split=true` seem very useful for me:  If you `forget` a couple of snapshots and run `restic recover --split=true`, you'll get the removed snapshots back (but deduplicated in the sense that multiple completely unchanged snapshots will resurrect only once and without snapshot metadata).
> 

no.

Checklist
---------

- I have not added documentation for the changes (in the manual) - there is no docu for `recover` so far

### DIFF
--- a/cmd/restic/cmd_recover.go
+++ b/cmd/restic/cmd_recover.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"os"
 	"time"
 
@@ -11,11 +12,11 @@ import (
 
 var cmdRecover = &cobra.Command{
 	Use:   "recover [flags]",
-	Short: "Recover data from the repository",
+	Short: "Recover data from the repository not referenced by snapshots",
 	Long: `
 The "recover" command builds a new snapshot from all directories it can find in
-the raw data of the repository. It can be used if, for example, a snapshot has
-been removed by accident with "forget".
+the raw data of the repository which are not referenced in an existing snapshot.
+It can be used if, for example, a snapshot has been removed by accident with "forget".
 
 EXIT STATUS
 ===========
@@ -59,24 +60,14 @@ func runRecover(gopts GlobalOptions) error {
 	trees := make(map[restic.ID]bool)
 
 	for blob := range repo.Index().Each(gopts.ctx) {
-		if blob.Blob.Type != restic.TreeBlob {
-			continue
+		if blob.Type == restic.TreeBlob {
+			trees[blob.Blob.ID] = false
 		}
-		trees[blob.Blob.ID] = false
 	}
 
-	cur := 0
-	max := len(trees)
-	Verbosef("load %d trees\n\n", len(trees))
-
+	Verbosef("load %d trees\n", len(trees))
+	bar := newProgressMax(!gopts.Quiet, uint64(len(trees)), "trees loaded")
 	for id := range trees {
-		cur++
-		Verbosef("\rtree (%v/%v)", cur, max)
-
-		if !trees[id] {
-			trees[id] = false
-		}
-
 		tree, err := repo.LoadTree(gopts.ctx, id)
 		if err != nil {
 			Warnf("unable to load tree %v: %v\n", id.Str(), err)
@@ -84,26 +75,37 @@ func runRecover(gopts GlobalOptions) error {
 		}
 
 		for _, node := range tree.Nodes {
-			if node.Type != "dir" || node.Subtree == nil {
-				continue
+			if node.Type == "dir" && node.Subtree != nil {
+				trees[*node.Subtree] = true
 			}
-
-			subtree := *node.Subtree
-			trees[subtree] = true
 		}
+		bar.Add(1)
 	}
-	Verbosef("\ndone\n")
+	bar.Done()
+
+	Verbosef("load snapshots\n")
+	err = restic.ForAllSnapshots(gopts.ctx, repo, nil, func(id restic.ID, sn *restic.Snapshot, err error) error {
+		trees[*sn.Tree] = true
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	Verbosef("done\n")
 
 	roots := restic.NewIDSet()
 	for id, seen := range trees {
-		if seen {
-			continue
+		if !seen {
+			Verboseff("found root tree %v\n", id.Str())
+			roots.Insert(id)
 		}
-
-		roots.Insert(id)
 	}
+	Printf("\nfound %d unreferenced roots\n", len(roots))
 
-	Verbosef("found %d roots\n", len(roots))
+	if len(roots) == 0 {
+		Verbosef("no snapshot to write.\n")
+		return nil
+	}
 
 	tree := restic.NewTree()
 	for id := range roots {
@@ -117,7 +119,7 @@ func runRecover(gopts GlobalOptions) error {
 			ModTime:    time.Now(),
 			ChangeTime: time.Now(),
 		}
-		err = tree.Insert(&node)
+		err := tree.Insert(&node)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


aawsome:
rebased this to resolve the merge conflicts.
I have to add a changelog entry and maybe docu for the new recover command (there is no docu for the current `recover`, though), but besides that, this PR is ready for review.

cmd/restic/cmd_recover.go
```diff
@@ -24,15 +25,29 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
+	Tags    []string
+}
+
+var recoverOptions RecoverOptions
+
 func init() {
 	cmdRoot.AddCommand(cmdRecover)
+	recoverFlags := cmdRecover.Flags()
+	recoverFlags.BoolVar(&recoverOptions.All, "all", false, "add all directories, not only unreferenced ones")
+	recoverFlags.BoolVar(&recoverOptions.Combine, "combine", true, "if true, combine all directories into one snapshot, else generate multple snapshots")
```
MichaelEischer, line 45: With the new prune implementation I'd expect there to be thousands of unreferenced trees, so I doubt that this option would still be useful. A option which must be passed as `--combine=false` is really confusing.

cmd/restic/cmd_recover.go
```diff
@@ -24,15 +25,29 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
+	DryRun  bool
+	Tags    []string
+}
+
+var recoverOptions RecoverOptions
+
 func init() {
 	cmdRoot.AddCommand(cmdRecover)
+	recoverFlags := cmdRecover.Flags()
+	recoverFlags.BoolVar(&recoverOptions.All, "all", false, "add all directories, not only unreferenced ones")
```
MichaelEischer, line 44: Do you see a use case where one would want to add root directories already covered by existing snapshots?

cmd/restic/cmd_recover.go
```diff
@@ -24,15 +25,29 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
+	recoverFlags.BoolVar(&recoverOptions.Combine, "combine", true, "if true, combine all directories into one snapshot, else generate multple snapshots")
+	recoverFlags.BoolVarP(&recoverOptions.DryRun, "dry-run", "n", false, "only show what would be done")
```
MichaelEischer, line 46: `--dry-run` only seems to be useful in combination with `--combine=false`?

cmd/restic/cmd_recover.go
```diff
@@ -59,90 +74,113 @@ func runRecover(gopts GlobalOptions) error {
...
+		}
+		bar.Add(1)
+	}
+	bar.Done()
 
-			subtree := *node.Subtree
-			trees[subtree] = true
+	if !opts.All {
+		Verbosef("load snapshots\n")
+		snapshots, err := restic.LoadAllSnapshots(gopts.ctx, repo, nil)
```
MichaelEischer, line 102: That should now use `ForAllSnapshots`.

cmd/restic/cmd_recover.go
```diff
@@ -59,90 +74,113 @@ func runRecover(gopts GlobalOptions) error {
...
+		Verbosef("done\n")
 	}
-	Verbosef("\ndone\n")
 
 	roots := restic.NewIDSet()
 	for id, seen := range trees {
-		if seen {
-			continue
+		if !seen {
+			Verbosef("found root tree %v\n", id.Str())
```
MichaelEischer, line 115: `s/Verbosef/Verboseff`. Printing all root tree ids is too much for normal verbosity.

MichaelEischer (changes requested):
I don't think that the `recover` command needs that many flags. For me it looks like `--all=false` is the behavior everyone would. `--all=true` just clutters the recovered snapshot(s). Now with the new prune implementation, there will be hundred or thousands of "root" trees in the repository, which makes `--combine=false` essentially useless in my opinion. When `recover` only creates a single snapshot, I don't see much use for a `--dry-run` option either, in the worst case `recover` will create a single snapshot which is also tagged accordingly.

`--all=false --combine=false` will not really recover deleted snapshots, these still miss all metadata.

cmd/restic/cmd_recover.go
aawsome: Thanks for the hint about the pruning. This applies to `recover` with and without combine. The only difference is whether the unreferenced trees are combined in one new snapshot (as current `recover` does) or in subtrees within one single new snapshot. I would always prefer to use `--combine=false` as this allows to clean up the unneeded snapshots.
But as you can specify tags for the "uncombined" snapshots, even lots of snapshots could be handled or easily removed if not needed.

About the option name: What about inverting it: `--split` with default "false"?

cmd/restic/cmd_recover.go
aawsome: No, actually not :wink: 
The current version effectively is `--all=true` and therefore I added this option to preserve the current behavior. I can remove this option and always remove directories covered by existing snapshots.

cmd/restic/cmd_recover.go
aawsome: Another use case would be to use it for debugging/troubleshooting which would also work on read-only repositories.

cmd/restic/cmd_recover.go
```diff
@@ -24,15 +25,29 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
-		return runRecover(globalOptions)
+		return runRecover(globalOptions, recoverOptions)
 	},
 }
 
+type RecoverOptions struct {
+	All     bool
+	Combine bool
+	DryRun  bool
+	Tags    []string
```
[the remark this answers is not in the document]
aawsome: is changed.

cmd/restic/cmd_recover.go
aawsome: I removed the `--all` option.

cmd/restic/cmd_recover.go
aawsome: is changed.

cmd/restic/cmd_recover.go
aawsome: changed.

cmd/restic/cmd_recover.go
aawsome: About the unreferenced trees: I made #3228 and think we should merge this in order to save the `recover` command (independent from this PR).

I already changed the name to `--split`.

rawtaz:
What's the use case of tagging a snapshot created by `recover`? This command is something you'd run manually when needed, so  why would you need to tag it? Even if you do, you could just use the existing `tag` command so there's little need for adding code to tag these manually produced snapshots IMO.

aawsome:
> What's the use case of tagging a snapshot created by `recover`? This command is something you'd run manually when needed, so why would you need to tag it? Even if you do, you could just use the existing `tag` command so there's little need for adding code to tag these manually produced snapshots IMO.

With `--split` you can get quite a lot of new snapshots. I thought it might be handy to let `recover` tag them all.
And it is exactly 3 extra lines of code... 

> Also, I don't understand the point of having a boolean argument to `--split`. If the default is to not split, then just `--split` would be enough, no?

You are right. Of course `--split` also works.

aawsome:
rebased to solve merge conflicts.

MichaelEischer (changes requested):
I'm still not convinced about the `--split` option as it just feels like a crutch to me. After all, the whole metadata of a snapshot is still lost in the process. So, I'd like to limit this PR to the code cleanups and tweaks, but defer new options to a discussion on how "un-deleting" a snapshot could/should work.

Please remove the `--split` and the then obsolete `--dry-run` and `--tags` options. The factored out `createSnapshot` function can stay though.

cmd/restic/cmd_recover.go
```diff
@@ -59,90 +73,111 @@ func runRecover(gopts GlobalOptions) error {
...
 		tree, err := repo.LoadTree(gopts.ctx, id)
 		if err != nil {
 			Warnf("unable to load tree %v: %v\n", id.Str(), err)
 			continue
 		}
 
 		for _, node := range tree.Nodes {
-			if node.Type != "dir" || node.Subtree == nil {
-				continue
+			if node.Type == "dir" && node.Subtree != nil {
```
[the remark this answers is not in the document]
aawsome: Very true :+1: 
Let's see when I'll find some time for it :wink:

aawsome:
I'm fine with postponing the '--split' or something else to a future PR.

> Please remove the `--split` and the then obsolete `--dry-run` and `--tags` options. The factored out `createSnapshot` function can stay though.

Is removed. Hope these small improvements are nevertheless helpful!

MichaelEischer (approved):
LGTM.

Excluding trees belonging to snapshot might be quite helpful, especially for a repository containing lots of snapshots. And the code cleanup is also a plus :-) .

Issue #2796, opened 2020-06-18
Restore --dry-run option to enable cold storage based repositories


beer4duke:
Output of `restic version`
--------------------------

Running `restic 0.9.6 compiled with go1.13.6 on linux/amd64`

Which functionality do you think we should add?
----------------------------------------------------------------------------------

`restic restore` command should have a dry-run option that lists all the pack files needed to satisfy the restore.
Other commands that require the presence of pack files on the restic backend could be complemented as well with the same dry-run option in a second time (`dump` for example).
Json output as a list of packs would be perfect when used in conjunction with `--json`.
In addition the json format could contain the restic command that generated this pack list and a timestamp that indicates when the command was issued. 

What are you trying to do?
--------------------------

I am trying to put restic in front of a tape based cold storage.

The basic properties of a cold storage restic repository are the following with regard to the folder structure and files:
- `config` file, `index`, `keys, `snapshots` directories both on hot and cold storage to allow immediate metadata operations on the repository
- `locks` directory is a hot storage only directory as the repository can be recreated from cold storage without its content.
- `data` directory cold storage

The idea users can perform backups that go straight to cold storage.
They can query restic repo from hot storage to look into metadata as usual.

And then they could issue a `restic restore --dry-run --json` query. The json result is then submitted to the cold storage that asynchronously brings back the cold pack files in the hot storage area.
When all the pack files are back in hot storage, the user can proceed with the actual restore.

This could be extended in a second time to other commands that require the presence of pack files in hot storage.

Adding the command and timestamp that generated the pack list would allow cold storage manager to understand usage and refine various parameters. For example if users are requesting to bring back 2 days old snapshots regularly, the pack files could stay a bit longer than 2 days in the hot storage area. 

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

I am very happy with the nice specification work and the clean design of restic that nicely separates content, metadata and encryption. Very well done!
I was able to test the backup workflow to cold storage and understand the blobls/subtrees/indexes... structures in only a few hours.
I tried to structure this proposal to avoid any conceptual issue but do not hesitate to complement it as I have only been using restic for few days.

I am proposing this feature request as a follow up to my initial [forum question](https://forum.restic.net/t/restore-dry-run/2840). As it looks like I am not the only one with this cold storage interest.

beer4duke:
I have started something not JSON yet but I have added the --dry-run option to the restore command.
I will publish this in a fork as a base branch for this issue.

[...]

PR #2881, opened 2020-08-07
prune/rebuild-index: Add warmup possibilities

aawsome:

Adds the possibility to access all packs which need a repack during `prune` and `rebuild.index`
This allows some cold storages to warm up these packs such that they are all accessable. This warm-up can also be done within a dry run such that the needed packs are available for the next `prune` run.
Also the `--json` option is now implemented for prune dry-run. This allows to build custom warm-up processes if the pure access to files doesn't do the warmup.

This can be easily extended to `restore`, see #2796.

The idea comes from #2796.
For cold storage discussions see  also #2504

I didn't see a discussion about pruning cold storage, but this PR proved to be very useful for pruning my OVH Cloud Archive repositories.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2881
+++ b/changelog/unreleased/pull-2881
@@ -1,10 +1,11 @@
-Enhancement: Add options for warmup-possibilities to prune
+Enhancement: Add options for warmup-possibilities to prune and reabuild-index
 
 Restic `prune` did not have a good possibility to warm up pack files that need
 repacking but are stored in a cold storage backend.
-We have added the option `--warmup`` which just tries to access all needed
+We have added the option `--warm-up`` which just tries to access all needed
 files. This triggers a warmup for some cold storages.
 Also the option `--json` in combination with `--dry-run` now prints the pack
 file names in JSON format. This can be used to implement a custom warm-up.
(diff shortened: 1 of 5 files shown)


aawsome:
I rebased to master and added to `--warm-up` option to `rebuild-index`.

rawtaz:
The more options there are, the more confusing it is for users. We want to keep the number of options as small as possible. Please try to not introduce new options unless really necessary. Is it really necessary in this case, and if so why can't it be the default behavior instead?

aawsome:
> Is it really necessary in this case, and if so why can't it be the default behavior instead?

See #3202. This is only relevant if "data pack files" cannot be accessed immediately. If this is the case, we need to distinguish the "warm-up" run and the real run, where the needed files are accessible (which can be hours later that the warm-up run).

Of course, we could decide not to add the `--warm-up` option at all and instead only give a verbose JSON output for `--dry-run --json`. This would mean that we only support the workaround where the users create scripts to do the warmup themselves and the option for warmup would be effectively the `--dry-run` option.

aawsome:
see also https://github.com/rustic-rs/rustic/pull/121

PR #2882, opened 2020-08-08
Add config file support

aawsome:

Adds support for config files.
This PR uses viper to allow reading all command flags from a config file. Supports all config file formats that are supported by viper. 
Also all future to-be-added flags and commands will be automatically covered.

Flags can be either specified globally, or only for given subcommands, e.g. this `restic.yaml`
```
repo: /path/to/repo
exclude-file: exludes
backup:
   password-file: password
```
would set `repo` and `exclude-files` for all commands (where the flag is used), but set `password-file` only for the backup command. 

While this change only adds a couple of lines of code to `cmd/restic/main.go`, it introduces quite some dependencies and increases the binary by about 1.4 MB. 
@maintainers: Can you please give me feedback if this is acceptable?

see, e.g. https://forum.restic.net/t/configuration-file-for-restic-crestic/2743
closes #16 

Checklist
---------


rawtaz:
With all due respect, this isn't helping right now. If you read the issue thread, 1) there is no consensus on which format to use, 2) viper has been mentioned several times, 3) the author has stated several times he's not convinced about viper and would prefer to use another tool, 4) there are already several other branches referenced that attempts to solve the configuration file feature, 5) given that there are so many different suggestions and ways to do this it's pretty reasonable that whatever the final decision is it should be @fd0's.

So, with that in mind, I fail to see how filing a PR that "solves" the problem in a way that others have already suggested before, and that is certainly not the decided way forward, especially when there are already other work of similar type, will help the project. Especially when the maintainers are busy trying to focus on things that leads up to an actual release, it's not like we're going to (or at least should) sit down and tackle this thing all of a sudden. Not to mention posting about it in two places gets so much attention that we do have to sit down and spend time responding to it, time that would otherwise have been spent on things leading up to a release.

My point is that in this particular case it should be very clear that the issue being discussed is not ready for yet another PR. If I'm misinterpreting something in that assessment, please feel perfectly free to correct me, I am very open to being wrong or misunderstanding something, and I don't want to come across as being ungrateful for your work, you have done *a lot* of great contributions to restic lately!

That said, I'd of course be happy to keep the PR around, but for now let us focus on other issues and PRs so we can get the next release out the door. We can't just attend new issues and PRs, we have to attend primarily those that we think are relevant for the next release and after that older PRs should have precedence. Things like this one can wait (yes, the issue is very old, but that is how it is and honestly it's not critical anyway since there are several alternatives to configure and run restic already), it won't be in the next release anyway.

Peace!

aawsome:
Ok, seems I happened to steal precious maintainers' time with this. Sorry, this was not my intend.

I also did not expect at all that this PR was to be tackled with high priority. You maintainers are perfectly right to prioritize and I really do appreciate a lot how much work you are doing. Especially given the fact that reviewing code is so much harder than writing code and work is distributed over so few shoulders (I just experience maintaining work being done by @fd0, @rawtaz and @MichaelEischer)

So, why did I propose this PR?

- I do have a need for it myself (running restic in an environment where I did not manage to get a wrapper or something else working nicely and the best solution actually is to patch restic)
- The issue is old as the hills (I have to admit I didn't read it all in detail and especially missed the viper discussion part) and seems to end with the proposition of a wrapper solution which did not fully satisfy users and maintainers - you could have easily closed the issue stating that config file support is not scope of the project.
- There is the branch `add-config-files` in the upstream repository which is already two years old and a hundreds-of-lines heavily "wip" code change - obviously nothing to build on...
- When thinking about cobra and viper within another project, I realized that cobra supports some kind of self-inspection for the used flags allowing to build a generic config-file-reading for the used parameters. This leads to this very local and small patch. The ~20 lines involving cobra can be easily changed to any config files reading module.
- The changes work really fine for me, I'm using them in the above mentioned environment. As open-source developer I like to share things :smile: - especially if I feel others could benefit from my work.

So sorry again, if this caused just confusion. I'm closing this now; feel free to come up to me when you feel time is ready for this topic.

rawtaz:
Thanks for explaining your thought process and there's really no need to apologize, you're just trying to help and that is admirable. You are right that we could close issues that just linger, but at the same time we think they have value in their discussions and many of them (like the config file one) is something that we do envision in the long run. If we close them we'd get other new ones about the same thing as time passes instead of people finding that central one. Not sure which is the best way to manage that type of issues :)

I think it's never wrong that there is work done by others that work fine and that people can find and try. For example a lot of the improving PRs like Michael's improved check command, the copy command, the VSS for Windows support and a lot of yours and greatroar's improvements I'm very happy that people try as time goes along, because at the end of the day we really want as many to have tried the code we're merging as possible. This is very important, just as an example even with the VSS support which at glance seem fine and many reported it working there's been the occasional issue popping up as more people tried it, and this is something we should expect with all PRs. Obviously it's even more important with PRs that have the potential to destroy data if they don't work properly.

Anyway, thanks again and I hope we can revisit this some day in the not too distant future. But for now there's other more important things to work on. Regardless, people can find your PR through the forum post and the issue now and use it as much as they want :)

creativeprojects:
@aawsome have you tried https://github.com/creativeprojects/resticprofile ? for what I can see in your PR it's the same kind of configuration file in yaml (it's using viper internally)

PR #2883, opened 2020-08-11
Clarify: message applies to storage encryption (not transport mechanism)

tim-seoss:
Previous message "read password from stdin" is ambiguous with respect to
whether the password being read refers to the respository encryption
password (passphrase), or to a password required for a non-local access
method for a remote repository.

Aide failure debugging by clarifying that at this point restic is attempting to aquire the repository encryption passphrase, as opposed to a transport (e.g. https / sftp) password.

Would have saved me ~30 mins wondering why an ssh auth key wasn't being used (when in fact it was, but the script was failing to inherit the RESTIC_PASSWORD_FILE env var under some execution paths).

Checklist
---------

- [N/A] I have added tests for all changes in this PR
- [N/A] I have added documentation for the changes (in the manual)
- [N/A] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -343,7 +343,7 @@ func ReadPassword(opts GlobalOptions, prompt string) (string, error) {
 		password, err = readPasswordTerminal(os.Stdin, os.Stderr, prompt)
 	} else {
 		password, err = readPassword(os.Stdin)
-		Verbosef("read password from stdin\n")
+		Verbosef("reading repository encryption password from stdin\n")
 	}
 
 	if err != nil {


rawtaz:
Sure doesn't hurt to clarify this a bit, and due to this PR I found an inconsistency in the documentation as wel, so thanks!

That said, "repository encryption password" is a bit too verbose, please change it to just "repository password" to make it more in line with the current documentation. It should still be clear enough that it's the repository this password is referring to, not a backend or storage or something else.

tim-seoss:
> Sure doesn't hurt to clarify this a bit, and due to this PR I found an inconsistency in the documentation as well, so thanks!
> 
> That said, "repository encryption password" is a bit too verbose, please change it to just "repository password"

Hi,

I did think about that when I wrote the message, but I though "encryption password" was the clearest.  i.e.

When trying to differentiate between:

- "the password needed to gain access to the repository cipher text files"
- "the password needed to encrypt/decrypt the data contained in the repository cipher text files"

... "repository encryption password" seemed like the shortest unambiguous phrase I could come up with, which answered the question:

"what could this have said, which would have definitely stopped me going on that wild goose chase?"

Whilst "repository password" steers the user in the right direction somewhat, I think it still leaves room for confusion.

This is both an error message and one logged with `Verbosef()`, and won't be seen during normal correct operation - so it seems to me like the most terse option that it's worth changing to.

Up to you of course!

HTH,

Tim.

tim-seoss:
On further consideration perhaps one of:

"repository password (encryption)", or "repository (encryption) password" might be equally clear, but also more consistent with the docs?

Cheers,

Tim.

rawtaz:
I'm sorry, but I think that "repository password" should be clear enough to be understood that it is the password for the actual repository. If the user then has other passwords involved, for the storage or transport of that repository, this is something they have to understand. We can't clutter simple output like this. I hope you don't mind changing it to just "repository password".

tim-seoss:
My personal opinion is that the form of the message you propose is not sufficiently clear.

"Error messages should be as simple as possible, but not simpler".

That having been said, if you would prefer to merge your form of the message, please go ahead, and I'll close this PR/issue.

rawtaz:
What other repository passwords are there, besides the actual repository password?

The only other passwords there are are for transport or storage, no?

tim-seoss:
Yes, but in the cases where there is a transport, then it is usually specific to each repository, hence the need to differentiate them clearly in the error message I think.

rawtaz:
@tim-seoss Thanks for your thorough concerns, I totally respect wanting to get it done right.

I touched base with others on this so I don't make decisions prematurely, and we agree that we should keep it at just "repository password". I'm sorry we disagree, but that happens sometimes.

I don't want to step on your toes here so feel free to close the PR if you prefer to do so and I can make the change separately.

rawtaz:
I'm closing this PR now since you suggested to close it if we didn't merge it, which I understand. Thanks regardless for your contribution, it's appreciated even though we couldn't agree on the text. Cheers!

PR #2884, opened 2020-08-11
doc: Correct wording for password

rawtaz:
Use "repository" instead of "backend" in a couple of places, to match documentation with restic output.


MichaelEischer (approved):
LGTM.

PR #2543, opened 2020-01-12
Add new command `repoinfo`

aawsome:
What is the purpose of this change? What does it change?
--------------------------------------------------------

Adds the possibility to show various repository statistics without walking trees, i.e. should be very fast.

The output looks like:
```
restic -r /path/to/repo repoinfo
enter password for repository: 
repository f1a1cc2c opened successfully, password is correct
scanning repo..

Repository content:
==================
                      count |    raw size |      crypto |   encr size
---------------------------------------------------------------------
lock files:               1 |       137 B |        32 B |       169 B
snapshot files:           1 |       222 B |        32 B |       254 B
index files:              1 |  18.270 KiB |        32 B |  18.301 KiB
data files:              39 | 190.754 MiB |         0 B | 190.754 MiB
key files:                1 |         0 B |       449 B |       449 B
---------------------------------------------------------------------
all files:               43 | 190.772 MiB |       545 B | 190.773 MiB

scanning index..

Index content:
==============
                 # packs |     # blobs |   raw blobs | pack header |      crypto |       total
----------------------------------------------------------------------------------------------
data blobs:           38 |         125 | 190.735 MiB |   4.665 KiB |   5.094 KiB | 190.744 MiB
tree blobs:            1 |           4 |   9.577 KiB |       152 B |       160 B |   9.882 KiB
----------------------------------------------------------------------------------------------
                      39 |         129 | 190.744 MiB |   4.813 KiB |   5.250 KiB | 190.754 MiB

Overhead:
=========
index:                  18.270 KiB (  0.01%)
snapshots:                   222 B (  0.00%)
locks:                       137 B (  0.00%)
pack header:             4.813 KiB (  0.00%)
crypto:                  5.782 KiB (  0.00%)
--------------------------------------------
total:                  29.082 KiB (  0.01%)

Total:
======
        129 blobs
         43 files
190.773 MiB total repository size
```

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------

No, but I needed some statistics while developing and thought it might be a good extension.
There is an issue with `restic stat` because it walks all trees for all snapshots and hence is pretty slow, see #2126. This extension does not solve this issue but is able to give repository statistics very fast. 

Please give me feedback if this is useful or should be complemented by something. I will then finish this PR (and add docu, etc).
If it is not useful, feel free to close this PR.

closes #1047

Checklist
---------

[...]

PR #2885, opened 2020-08-12
Use "pack file" instead of "data file"

### LINKED ISSUE #2877: Pack(File) or DataFile?
What are you trying to do?
--------------------------

When working on implementations, I always get a bit confused whether to use the name "pack"/"pack file" or "data file".

In the users documentation part "Terminology", it is referred as "Packs", but the dir in the repository layout is `/data`.

Also, in the implementations, the constant is called `DataFile` which may be mixed up with `DataBlob` but I've seen may places where variables/comments or user outputs refer to "packs".

I suggest to:

- change `DataFile` into `PackFile`
- look for "data file" et al in code/comments and change if necessary
- add an explanation in `resign.rst` that packs are stored under `/data` for historic reasons

Is this a common view? I can prepare an PR if we agree that this clarifies things.

aawsome:

Clarify code and documentation by using "pack file" or "pack" everywhere. In some places "data file" was used. This was a bit irritating as there are also data and tree blobs and pack which contain only data blobs or only tree blobs, respectively.

closes #2877 

Checklist
---------

- [x] I have *not* added tests for all changes in this PR, but tests are changed accordingly
- [x] I have *not* added documentation for the changes (in the manual) but changed the docu accordingly

### DIFF
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -149,7 +149,7 @@ func runCat(gopts GlobalOptions, args []string) error {
 
 	switch tpe {
 	case "pack":
-		h := restic.Handle{Type: restic.DataFile, Name: id.String()}
+		h := restic.Handle{Type: restic.PackFile, Name: id.String()}
 		buf, err := backend.LoadAll(gopts.ctx, nil, repo.Backend(), h)
 		if err != nil {
 			return err
--- a/cmd/restic/cmd_debug.go
+++ b/cmd/restic/cmd_debug.go
@@ -84,8 +84,8 @@ type Blob struct {
 
 func printPacks(repo *repository.Repository, wr io.Writer) error {
 
-	return repo.List(context.TODO(), restic.DataFile, func(id restic.ID, size int64) error {
-		h := restic.Handle{Type: restic.DataFile, Name: id.String()}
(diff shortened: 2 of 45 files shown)


aawsome:
The change is also needed within #2718 - I used "data files" there and were wondering if that was confusing...

rawtaz:
I wonder if there are any or many PRs that will have to be updated if we merge this one first? If so, it might be best to put this on ice until after merging the other ones.

aawsome:
I assume that this doesn't affect many PRs directly, but may produce easy-to-solve merge conflicts if "nearby" lines are changed. From my PRs - as far as I can see - only #2718 and #2543 should be affected.

I would argue that this PR could be used immediately as potential merge conflicts of other PRs will not arise or be easy to solve.
On the other hand, it won't be much to fix for this PR if this really doesn't interfere with other PRs - and if there are merge conflicts, they will be also easy to fix here. When postponing this, we should just make sure that all tests still pass and we do not forget a "data file" sneaked in by another PR :wink:

MichaelEischer (approved):
LGTM. I've thought a bit about whether to merge this now or later on. Besides a few comment and doc changes everything else will lead to a compiler error if the name change is lost during a merge. The few merge conflicts that could arise should also be rather simple to resolve. So I think it's preferential to merge this PR just now rather than wait.

PR #2898, opened 2020-08-25
backup: Add --iexclude-file option

### LINKED ISSUE #2427: Case insensitive exclude file
Output of `restic version`
--------------------------

Add case insensitive versions of the `--exclude-file`.

What are you trying to do?
--------------------------
Exclude files properly and easily on Windows.
Actually, not me personally, but I've seen some Windows users very confused about file path letter cases. For example:
https://forum.restic.net/t/exclude-file-to-block-folders-from-windows-drives/2070
Even at the end when the problem was resolved, I don't think they really understood the reason properly.
restic already has case insensitive include/exclude. I think it makes sense to extend it to `--exclude-file`.

Yes, every day.

aawsome:

closes #2427

see #2427

Checklist
---------

-  I have not added tests for all changes in this PR; did not find tests for `--exclude-file` nor `--iexclude`

### DIFF
--- a/changelog/unreleased/issue-2427
+++ b/changelog/unreleased/issue-2427
@@ -1,0 +1,7 @@
+Enhancement: Add flag `--iexclude-file` to backup command
+
+The backup command now supports the flag `--iexclude-file` which is a
+case-insensitive version of `--exclude-file`.
+
+https://github.com/restic/restic/issues/2427
+https://github.com/restic/restic/pull/2898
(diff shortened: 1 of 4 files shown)


MichaelEischer (approved):
LGTM. Thanks!

I've made a few small changes: The help text of the backup command now prints `file` as flag-value name instead of `stringArray`. I've also updated the help text in the documentation along with rewording the changelog message a bit.

PR #2904, opened 2020-08-27
Update usage strings to put flags before args

### LINKED ISSUE #2830: Optional/Positional Commandline Argument Ordering Inconsistent
While investigating [this issue in `crestic`](https://github.com/nils-werner/crestic/issues/14), I stumbled across a peculiarity in `restic`: The ordering of positional arguments vs optional arguments is a little inconsistent and in some cases not following the "UNIX getopt standard":

The "standard" way of doing it is

```
command [subcommand] [--optional-arguments ...] [positional-arguments ...]
```

where optional arguments come first, and positional arguments come second. Most (19 out of 26) `restic` commands follow this scheme, i.e. `backup`:

```
Usage:
  restic backup [flags] FILE/DIR [FILE/DIR] ...
```

However, there are 7 subcommands that follow a different, non-standard scheme that swaps optional and positional arguments:

```
command [subcommand] [positional-arguments ...] [--optional-arguments ...]
```

they are

```
Usage:
  restic diff snapshot-ID snapshot-ID [flags]

Usage:
  restic generate [command] [flags]

Usage:
  restic help [command] [flags]

Usage:
  restic key [list|add|remove|passwd] [ID] [flags]

Usage:
  restic list [blobs|packs|index|snapshots|keys|locks] [flags]

Usage:
  restic migrate [name] [flags]

Usage:
    restic snapshots [snapshotID ...] [flags]
```

I believe this is just a minor inconsistency in the docstrings, Go/Cobra don't seem to care about the ordering and therefore it might only be a little confusing to users.

However it makes it hard to interface with the CLI the way I am doing with `crestic`: `crestic` uses the same commands and arguments as `restic`, but it is written in Python. But the Python `argparse` implementation is much more strict about the ordering of optional arguments and positional arguments, and therefore, the `restic` docstring are sometimes not valid for `crestic` calls. This causes errors when users seemingly do everything right.

Output of `restic version`
--------------------------

```
restic 0.9.6 compiled with go1.14.4 on linux/amd64
```

```
restic backup --help
restic cache --help
restic cat --help
restic check --help
restic diff --help
restic dump --help
restic find --help
restic forget --help
restic generate --help
restic help --help
restic init --help
restic key --help
restic list --help
restic ls --help
restic migrate --help
restic mount --help
restic prune --help
restic rebuild-index --help
restic recover --help
restic restore --help
restic self-update --help
restic snapshots --help
restic stats --help
restic tag --help
restic unlock --help
restic version --help
```

None

Expected behavior
-----------------
All commands follow the same "standard" convention of optional arguments first.

Actual behavior
---------------
Some commands follow the same "non-standard" convention of positional arguments first.

Have a look at the output of the commands above.

This is likely just a docstring inconsistency.

Just reorder the arguments in the docstring for those 7 commands.

I use it every day and I sleep better because of it ;-)

jasonccox:

The standard UNIX-style ordering of command-line arguments places
optional flags before other positional arguments. All of restic's
commands support this ordering, but some of the usage strings showed the
flags after the positional arguments (which restic also parses just
fine). This change updates the doc strings to reflect the standard
ordering.

Because the `restic help` command comes directly from Cobra, there does
not appear to be a way to update the argument ordering in its usage
string, so it maintains the non-standard ordering (positional arguments
before optional flags).

Closes #2830 

Checklist
---------

Seeing as this is just a documentation change and restic continues to work exactly the same as before, I don't think new tests or changelog entries are necessary.

### DIFF
--- a/cmd/restic/cmd_generate.go
+++ b/cmd/restic/cmd_generate.go
@@ -9,7 +9,7 @@ import (
 )
 
 var cmdGenerate = &cobra.Command{
-	Use:   "generate [command]",
+	Use:   "generate [flags]",
 	Short: "Generate manual pages and auto-completion files (bash, zsh)",
 	Long: `
 The "generate" command writes automatically generated files (like the man pages
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -14,7 +14,7 @@ import (
 )
 
 var cmdDiff = &cobra.Command{
-	Use:   "diff snapshot-ID snapshot-ID",
+	Use:   "diff [flags] snapshot-ID snapshot-ID",
 	Short: "Show differences between two snapshots",
 	Long: `
 The "diff" command shows differences from the first to the second snapshot. The
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -16,7 +16,7 @@ import (
 )
 
 var cmdKey = &cobra.Command{
-	Use:   "key [list|add|remove|passwd] [ID]",
+	Use:   "key [flags] [list|add|remove|passwd] [ID]",
 	Short: "Manage keys (passwords)",
 	Long: `
 The "key" command manages keys (passwords) for accessing the repository.
--- a/cmd/restic/cmd_list.go
+++ b/cmd/restic/cmd_list.go
@@ -9,7 +9,7 @@ import (
 )
 
 var cmdList = &cobra.Command{
-	Use:   "list [blobs|packs|index|snapshots|keys|locks]",
+	Use:   "list [flags] [blobs|packs|index|snapshots|keys|locks]",
 	Short: "List objects in the repository",
 	Long: `
 The "list" command allows listing objects in the repository based on type.
--- a/cmd/restic/cmd_migrate.go
+++ b/cmd/restic/cmd_migrate.go
@@ -8,7 +8,7 @@ import (
 )
 
 var cmdMigrate = &cobra.Command{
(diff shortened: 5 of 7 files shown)


cmd/restic/cmd_generate.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var cmdGenerate = &cobra.Command{
-	Use:   "generate [command]",
+	Use:   "generate [flags] [command]",
```
MichaelEischer, line 12: I don't see any commands which are accepted by `generate`. Please remove the `[command]` part.

doc/man/restic-diff.1
```diff
@@ -10,7 +10,7 @@ restic\-diff \- Show differences between two snapshots
 
 .SH SYNOPSIS
 .PP
-\fBrestic diff snapshot\-ID snapshot\-ID [flags]\fP
+\fBrestic diff [flags] snapshot\-ID snapshot\-ID\fP
```
MichaelEischer, line 13: Please remove all man-page changes, these are generated automatically on release.

MichaelEischer (changes requested):
I've just two minor remarks. Once these are fixed the PR is ready for merging.

cmd/restic/cmd_generate.go
jasonccox: Done!

doc/man/restic-diff.1
jasonccox: Okay, that makes sense.

MichaelEischer (approved):
LGTM. Thanks for the cleanup!

PR #2905, opened 2020-08-29
ls: Explicitly enforce that the user specifies a snapshot ID

### LINKED ISSUE #2299: can't use --path with ls
NixOS
bash shell

Output of `restic version`
--------------------------
restic 0.9.5 compiled with go1.12.1 on linux/amd64

`export RCLONE_CONFIG_PASS=somepass`
`restic -r rclone:onedrive:somename ls --path=/home/bbigras/something`

onedrive with rclone

Expected behavior
-----------------
Seing the files from the last snapshot containing the specified path.

Actual behavior
---------------
I get:
```
repository <something> opened successfully, password is correct
Ignoring "", it is not a snapshot id
Ignoring filters as there are explicit snapshot ids given
```

do the steps from "How did you run restic exactly". Probably don't even need a real restic repo.

I guess restic thinks a snapshot id was specified when it wasn't and then ignored the filter.

no

restic is awesome.

MichaelEischer:
As reported in #2299 the help message of the `ls` command suggests that the snapshotID is optional but falls when call like this. This PR changes the help message and parameter check to always require a snapshotID.

Fixes #2299

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -16,7 +16,7 @@ import (
 )
 
 var cmdLs = &cobra.Command{
-	Use:   "ls [flags] [snapshotID] [dir...]",
+	Use:   "ls [flags] snapshotID [dir...]",
 	Short: "List files in a snapshot",
 	Long: `
 The "ls" command lists files and directories in a snapshot.
@@ -89,8 +89,8 @@ type lsNode struct {
 }
 
 func runLs(opts LsOptions, gopts GlobalOptions, args []string) error {
-	if len(args) == 0 && len(opts.Hosts) == 0 && len(opts.Tags) == 0 && len(opts.Paths) == 0 {
-		return errors.Fatal("Invalid arguments, either give one or more snapshot IDs or set filters.")
+	if len(args) == 0 {
+		return errors.Fatal("no snapshot ID specified")
 	}
 
 	// extract any specific directories to walk


rawtaz:
Do we really want this command to require a snapshot ID? What it you want to list all snapshots for a specific host, tag, etc?

MichaelEischer:
It would be possible to _extend_ the ls command to handle multiple snapshots in one run. Right now all this PR does is to ensure that `ls --host abc` no longer returns `Ignoring "", it is not a snapshot id` as a warning, but rather the more helpful message `no snapshot ID specified`.

I'm not sure how the CLI params for `ls` would generalize to multiple snapshots:
`ls [flags] [snapshotID ...] [dir ...]` has the problem that as both `snapshotID`s and `dir`s can be present 0 to many times (once #2859 is merged, all commands which handle multiple snapshots at once, support between zero and n specified snapshot IDs). It would be possible to recognize the dir parameters based on their `/` prefix, but that's not obvious for a user at first.

skarekrow:
Doesn't `restic snapshots` already cover that usecase @rawtaz?

rawtaz:
> Doesn't `restic snapshots` already cover that usecase @rawtaz?

It does, I meant what `restic ls` does but for all snapshots. Badly worded by me.

MichaelEischer:
Any suggestions for how the CLI should look for an `ls` command that can process multiple snapshots? Otherwise I'd suggest to defer handling multiple snapshots and create an issue for it.

rawtaz:
> I'd suggest to defer handling multiple snapshots

I completely agree. Let's fix the current issue. If anyone really wants `ls` to handle multiple snapshots, make a feature request. If you don't have a real need for it, don't make a feature request so that we can focus on other work instead :)

rawtaz (approved):
<3

PR #2906, opened 2020-08-29
Restore inconsistent timestamps permissions

### LINKED ISSUE #1212: Restore code produces inconsistent timestamps/permissions
Follow-up to #982 and #1044, relevant info excerpted below.

## Output of `restic version`

```
> restic version
restic 0.6.0-rc.1 (v0.6.0-rc.1-0-g17ff41af)
```

See repro steps below.

Local + S3.

## Expected behavior

- timestamps should be restored consistently for intermediate directories
- rights should be restored consistently for intermediate directories

## Actual behavior

- timestamps and rights are not restored correctly depending on the pattern and the fact that the path matches or not.

```
mkdir test/{foo,bar}
chmod 700 test/foo
touch test/{foo,bar}/{what,wut}
restic -r repo init
restic -r repo backup test
restic -r repo restore f608e300 --include 'wut' --target restored
rm -rf restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/bar  # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/foo  # timestamp is at restore date, rights are 0700
restic -r repo restore f608e300 --include 'foo' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
rm -rf restored
restic -r repo restore f608e300 --include 'test' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0755
ls -ld restored/test/bar  # timestamp is at backup date, rights are 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
restic -r repo restore f608e300 --include '/test/**' --target restored
ls -ld restored/test      # timestamp is at restore date, rights are 0700, should be 0755
ls -ld restored/test/bar  # timestamp is at backup date, rights are 0755
ls -ld restored/test/foo  # timestamp is at backup date, rights are 0700
rm -rf restored
```

Yes.

```diff
--- a/src/restic/restorer.go
+++ b/src/restic/restorer.go 
+			// FIXME: timestamp is restored only for final directory matching the pattern, e.g only for baz in foo/bar/baz, if baz is a dir, and in no other case
+			// instead, restoreTo could return a `hasRestorable` flag (along with the existing err) set to true if something was indeed found to be restored during recursion
 			if selectedForRestore {
 				// Restore directory timestamp at the end. If we would do it earlier, restoring files within
 				// the directory would overwrite the timestamp of the directory they are in.
@@ -96,6 +102,9 @@ func (res *Restorer) restoreNodeTo(node *Node, dir string, dst string, idx *Hard
 	if err != nil && os.IsNotExist(errors.Cause(err)) {
 		debug.Log("create intermediate paths")
 
+		// FIXME: not mkdir str, restore node for each path component (which requires passing the current node stack), else attrs are incorrect
+		// also, one should take care of not doing restore ops twice or more if the node path component is already there.
 		// Create parent directories and retry
 		err = fs.MkdirAll(filepath.Dir(dstPath), 0700)
 		if err == nil || os.IsExist(errors.Cause(err)) { 
```

### LINKED ISSUE #1402: Restoring with --include creates directories with wrong permissions on the way down.
## Output of `restic version`
restic 0.7.3

See details below

Client: Debian Stable (9.2) 
Server: FreeBSD 11.1 server in the LAN using OpenSSH (OpenSSH_7.2p2, OpenSSL 1.0.2k-freebsd  26 Jan 2017)

## Expected behavior
The all directories should be created with the original permissions and ownership, even if you only restore files being deep in the hierarchy.

## Actual behavior
When restoring files lying deep in the hierarchy the directories on the way down are restored with a different user and permission as they had on the original.

testrestore/:
insgesamt 4
drwx------ 2 root root 4096 Okt 31 21:30 testdir
>
>testrestore/testdir:
insgesamt 0
-rwxr-xr-x 1 rakor rakor 0 Okt 31 21:25 testfile

I would think that directories on the way down are not "restored" but just "created" with "secure defaults".

restore the empty directories instead of just creating them.

kitone:

Fix inconsistent timestamps/permissions during restore in some case

https://github.com/restic/restic/issues/1212

Checklist
---------

- ~~I have added documentation for the changes (in the manual)~~

Information
-----------

The current implementation of restore code using two pass does not restore the metadata of all elements.

I will use a simplified reproductible script to reduce the log/debug
you can find the complet reproductible script (thanks to @deviantintegral) there https://github.com/restic/restic/issues/1212#issuecomment-671553782

About the change :
- The first commit extend and add test case to check Mode and ModTime
- The second add log information because it was pretty hard to debug treeVisitor proc functions and understand what happenning.
- The fix make the simplified and complet reproductible script work, ~~but introduce a bad behaviour, trying to restore metadata on a file/dir not selected for restore.
```ignoring error for /test/bar: UtimesNano: no such file or directory```~~

The simplified script, show that metadata on the root directory isn't restored

```bash
#!/bin/bash

export DEBUG_LOG=restic-debug.log

check_equal_permission()
{
  echo "Comparing permissions on $1 to $2..."
  (test $(stat -f %p $1) -eq $(stat -f %p $2) && echo "Passed!") || (echo "#### Comparison failed." && ls -ld $1 && ls -ld $2)
  echo ""
}

check_equal_modified()
{
  echo "Comparing modified dates on $1 to $2..."
  (test $(stat -f %m $1) -eq $(stat -f %m $2) && echo "Passed!") || (echo "#### Comparison failed." && ls -ld $1 && ls -ld $2)
  echo ""
}

check_file()
{
  check_equal_permission $1 $2
  check_equal_modified $1 $2
}

rm restic-debug.log
mkdir -p test/{dir1,dir2}
mkdir -p test/dir1/dir3/
chmod 700 test/dir1
chmod 700 test/dir1/dir3
touch -t 201901010000 test/{dir1,dir2}/{file1,file2} test/{dir1,dir2} test test/dir1/dir3/file3
restic --password-command='echo pass' -r repo init
restic --password-command='echo pass' -r repo backup test

rm restic-debug.log

restic --password-command='echo pass' -r repo restore latest --include 'file1' --target restored
check_file test restored/test

rm -rf test repo restored
```

The logs show that leaveDir is never executed, this is why the metadata of the root directory and in some case isn't restored.

```go
// first tree pass: create directories and collect all files to restore
err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
  enterDir: func(node *restic.Node, target, location string) error {
    debug.Log("first pass, enterDir: mkdir %q, leaveDir should restore metadata", location)
    // create dir with default permissions
    // #leaveDir restores dir metadata after visiting all children
    return fs.MkdirAll(target, 0700)
  },

  visitNode: func(node *restic.Node, target, location string) error {
    debug.Log("first pass, visitNode: mkdir %q, leaveDir on second pass should restore metadata", location)
    // create parent dir with default permissions
    // second pass #leaveDir restores dir metadata after visiting/restoring all children
    err := fs.MkdirAll(filepath.Dir(target), 0700)
```

I guess, also because we use MkdirAll, so when we pass on ```"/test/dir1"``` but restoring with ```---include 'file1'``` ```childMayBeSelected``` is equal to true, but the leaveDir is only executed on ```selectedForRestore```

```bash
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test"
```

```go
if selectedForRestore {
  err = sanitizeError(visitor.leaveDir(node, nodeTarget, nodeLocation))
  if err != nil {
    return err
  }
}
```

```bash
restorer/restorer.go:208	restorer.(*Restorer).RestoreTo	first pass for "/Users/kitone/Downloads/restic_kitone/issues_1212/restored"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored / cfd11f1ffa1ce50693db112af06c4df85ba92ec5dd3a5a18247c81e7a067c774
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test /test d6c3f48a1e21c0032e57790081a46f6aa25de05e307fae94c2859d3bc1670c33
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir1"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir1 /test/dir1 5adbd0033dc689a2195d61fb22b8ef6e8be43ba473528bb7e8c3fd8e0e4e2083
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir1/dir3"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir1/dir3 /test/dir1/dir3 ab6863baeb85cbe416e25baefd0f11622a69712323c986a027dfe4a85cf1852d
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir1/dir3/file3"
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned true false for "/test/dir1/file1"
restorer/restorer.go:220	restorer.(*Restorer).RestoreTo.func2	first pass, visitNode: mkdir "/test/dir1/file1", leaveDir on second pass should restore metadata
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir1/file2"
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir2"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir2 /test/dir2 81891851485d83cf6cf8e605bf3ea3487a9951cbfb2e58696db5fb326373ddd5
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned true false for "/test/dir2/file1"
restorer/restorer.go:220	restorer.(*Restorer).RestoreTo.func2	first pass, visitNode: mkdir "/test/dir2/file1", leaveDir on second pass should restore metadata
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir2/file2"

restorer/restorer.go:261	restorer.(*Restorer).RestoreTo	second pass for "/Users/kitone/Downloads/restic_kitone/issues_1212/restored"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored / cfd11f1ffa1ce50693db112af06c4df85ba92ec5dd3a5a18247c81e7a067c774
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test /test d6c3f48a1e21c0032e57790081a46f6aa25de05e307fae94c2859d3bc1670c33
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir1"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir1 /test/dir1 5adbd0033dc689a2195d61fb22b8ef6e8be43ba473528bb7e8c3fd8e0e4e2083
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir1/dir3"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir1/dir3 /test/dir1/dir3 ab6863baeb85cbe416e25baefd0f11622a69712323c986a027dfe4a85cf1852d
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir1/dir3/file3"
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned true false for "/test/dir1/file1"
restorer/restorer.go:270	restorer.(*Restorer).RestoreTo.func5	second pass, visitNode: restore node "/test/dir1/file1"
restorer/restorer.go:158	restorer.(*Restorer).restoreNodeMetadataTo	restoreNodeMetadata file1 /Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir1/file1 /test/dir1/file1
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir1/file2"
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false true for "/test/dir2"
restorer/restorer.go:53	  restorer.(*Restorer).traverseTree	/Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir2 /test/dir2 81891851485d83cf6cf8e605bf3ea3487a9951cbfb2e58696db5fb326373ddd5
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned true false for "/test/dir2/file1"
restorer/restorer.go:270	restorer.(*Restorer).RestoreTo.func5	second pass, visitNode: restore node "/test/dir2/file1"
restorer/restorer.go:158	restorer.(*Restorer).restoreNodeMetadataTo	restoreNodeMetadata file1 /Users/kitone/Downloads/restic_kitone/issues_1212/restored/test/dir2/file1 /test/dir2/file1
restorer/restorer.go:93	  restorer.(*Restorer).traverseTree	SelectFilter returned false false for "/test/dir2/file2"
```

The solution is to keep track of restored child status so parent and root directory not selected by filter will also restore metadata when traversing the tree.

### DIFF
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -90,7 +90,7 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 		}
 
 		selectedForRestore, childMayBeSelected := res.SelectFilter(nodeLocation, nodeTarget, node)
-		debug.Log("SelectFilter returned %v %v", selectedForRestore, childMayBeSelected)
+		debug.Log("SelectFilter returned %v %v for %q", selectedForRestore, childMayBeSelected, nodeLocation)
 
 		sanitizeError := func(err error) error {
 			if err != nil {
@@ -198,24 +198,23 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 		}
 	}
 
-	restoreNodeMetadata := func(node *restic.Node, target, location string) error {
-		return res.restoreNodeMetadataTo(node, target, location)
-	}
-	noop := func(node *restic.Node, target, location string) error { return nil }
-
 	idx := restic.NewHardlinkIndex()
 
 	filerestorer := newFileRestorer(dst, res.repo.Backend().Load, res.repo.Key(), res.repo.Index().Lookup)
 
+	debug.Log("first pass for %q", dst)
+
 	// first tree pass: create directories and collect all files to restore
 	err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
 		enterDir: func(node *restic.Node, target, location string) error {
+			debug.Log("first pass, enterDir: mkdir %q, leaveDir should restore metadata", location)
 			// create dir with default permissions
 			// #leaveDir restores dir metadata after visiting all children
 			return fs.MkdirAll(target, 0700)
 		},
 
 		visitNode: func(node *restic.Node, target, location string) error {
+			debug.Log("first pass, visitNode: mkdir %q, leaveDir on second pass should restore metadata", location)
 			// create parent dir with default permissions
 			// second pass #leaveDir restores dir metadata after visiting/restoring all children
 			err := fs.MkdirAll(filepath.Dir(target), 0700)
@@ -242,7 +241,9 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 
 			return nil
 		},
-		leaveDir: noop,
+		leaveDir: func(node *restic.Node, target, location string) error {
+			return nil
+		},
 	})
 	if err != nil {
 		return err
@@ -253,10 +254,15 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 		return err
 	}
 
+	debug.Log("second pass for %q", dst)
+
 	// second tree pass: restore special files and filesystem metadata
 	return res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
-		enterDir: noop,
+		enterDir: func(node *restic.Node, target, location string) error {
+			return nil
+		},
 		visitNode: func(node *restic.Node, target, location string) error {
+			debug.Log("second pass, visitNode: restore node %q", location)
 			if node.Type != "file" {
 				return res.restoreNodeTo(ctx, node, target, location)
 			}
@@ -275,7 +281,10 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 
 			return res.restoreNodeMetadataTo(node, target, location)
 		},
-		leaveDir: restoreNodeMetadata,
+		leaveDir: func(node *restic.Node, target, location string) error {
+			debug.Log("second pass, leaveDir restore metadata %q", location)
+			return res.restoreNodeMetadataTo(node, target, location)
+		},
 	})
 }
 
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -49,12 +49,12 @@ type treeVisitor struct {
 
 // traverseTree traverses a tree from the repo and calls treeVisitor.
 // target is the path in the file system, location within the snapshot.
-func (res *Restorer) traverseTree(ctx context.Context, target, location string, treeID restic.ID, visitor treeVisitor) error {
+func (res *Restorer) traverseTree(ctx context.Context, target, location string, treeID restic.ID, visitor treeVisitor) (hasRestored bool, err error) {
 	debug.Log("%v %v %v", target, location, treeID)
 	tree, err := res.repo.LoadTree(ctx, treeID)
 	if err != nil {
 		debug.Log("error loading tree %v: %v", treeID, err)
-		return res.Error(location, err)
+		return hasRestored, res.Error(location, err)
 	}
 
 	for _, node := range tree.Nodes {
@@ -66,7 +66,7 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 			debug.Log("node %q has invalid name %q", node.Name, nodeName)
 			err := res.Error(location, errors.Errorf("invalid child node name %s", node.Name))
 			if err != nil {
-				return err
+				return hasRestored, err
 			}
 			continue
 		}
@@ -79,7 +79,7 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 			debug.Log("node %q has invalid target path %q", node.Name, nodeTarget)
 			err := res.Error(nodeLocation, errors.New("node has invalid path"))
 			if err != nil {
-				return err
+				return hasRestored, err
 			}
 			continue
 		}
@@ -92,6 +92,10 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 		selectedForRestore, childMayBeSelected := res.SelectFilter(nodeLocation, nodeTarget, node)
 		debug.Log("SelectFilter returned %v %v for %q", selectedForRestore, childMayBeSelected, nodeLocation)
 
+		if selectedForRestore {
+			hasRestored = true
+		}
+
 		sanitizeError := func(err error) error {
 			if err != nil {
 				err = res.Error(nodeLocation, err)
@@ -101,27 +105,38 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 
 		if node.Type == "dir" {
 			if node.Subtree == nil {
-				return errors.Errorf("Dir without subtree in tree %v", treeID.Str())
+				return hasRestored, errors.Errorf("Dir without subtree in tree %v", treeID.Str())
 			}
 
 			if selectedForRestore {
 				err = sanitizeError(visitor.enterDir(node, nodeTarget, nodeLocation))
 				if err != nil {
-					return err
+					return hasRestored, err
 				}
 			}
+
+			// keep track of restored child status
+			// so metadata of the current directory are restored on leaveDir
+			childHasRestored := false
 
 			if childMayBeSelected {
-				err = sanitizeError(res.traverseTree(ctx, nodeTarget, nodeLocation, *node.Subtree, visitor))
+				childHasRestored, err = res.traverseTree(ctx, nodeTarget, nodeLocation, *node.Subtree, visitor)
+				err = sanitizeError(err)
 				if err != nil {
-					return err
+					return hasRestored, err
+				}
+				// inform the parent directory to restore parent metadata on leaveDir if needed
+				if childHasRestored {
+					hasRestored = true
 				}
 			}
 
-			if selectedForRestore {
+			// metadata need to be restore when leaving the directory in both cases
+			// selected for restore or any child of any subtree have been restored
+			if selectedForRestore || childHasRestored {
 				err = sanitizeError(visitor.leaveDir(node, nodeTarget, nodeLocation))
 				if err != nil {
-					return err
+					return hasRestored, err
 				}
 			}
 
@@ -131,12 +146,12 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 		if selectedForRestore {
 			err = sanitizeError(visitor.visitNode(node, nodeTarget, nodeLocation))
 			if err != nil {
-				return err
+				return hasRestored, err
 			}
 		}
 	}
 
-	return nil
+	return hasRestored, nil
 }
 
 func (res *Restorer) restoreNodeTo(ctx context.Context, node *restic.Node, target, location string) error {
@@ -205,7 +220,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 	debug.Log("first pass for %q", dst)
 
 	// first tree pass: create directories and collect all files to restore
-	err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
+	_, err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
 		enterDir: func(node *restic.Node, target, location string) error {
 			debug.Log("first pass, enterDir: mkdir %q, leaveDir should restore metadata", location)
 			// create dir with default permissions
@@ -257,7 +272,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 	debug.Log("second pass for %q", dst)
 
 	// second tree pass: restore special files and filesystem metadata
-	return res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
+	_, err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
 		enterDir: func(node *restic.Node, target, location string) error {
 			return nil
 		},
@@ -286,6 +301,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 			return res.restoreNodeMetadataTo(node, target, location)
 		},
 	})
+	return err
 }
 
 // Snapshot returns the snapshot this restorer is configured to use.
@@ -298,7 +314,7 @@ func (res *Restorer) VerifyFiles(ctx context.Context, dst string) (int, error) {
 	// TODO multithreaded?
 
 	count := 0
-	err := res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
+	_, err := res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
 		enterDir: func(node *restic.Node, target, location string) error { return nil },
 		visitNode: func(node *restic.Node, target, location string) error {
 			if node.Type != "file" {
--- a/internal/restorer/restorer_test.go
+++ b/internal/restorer/restorer_test.go
@@ -692,7 +692,7 @@ func TestRestorerTraverseTree(t *testing.T) {
 			// make sure we're creating a new subdir of the tempdir
 			target := filepath.Join(tempdir, "target")
 
-			err = res.traverseTree(ctx, target, string(filepath.Separator), *sn.Tree, test.Visitor(t))
+			_, err = res.traverseTree(ctx, target, string(filepath.Separator), *sn.Tree, test.Visitor(t))
 			if err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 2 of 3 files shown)


kitone:
I have push a proper solution that doesn't have side effect of trying to restore unwanted directory.

This broke a test case, causing leaveDir call more that one time, but I'm not sure to understand this test...

```
--- FAIL: TestRestorerTraverseTree (0.01s)
    --- FAIL: TestRestorerTraverseTree/#04 (0.00s)
        restorer_test.go:506: step 1, leaveDir(/dir): expected no more than 1 function calls
```

kitone:
Continuous integration fail on windows, I have try to fix it without success. Would love little help, I don't understand why this fail on windows.

MichaelEischer:
@kitone My guess would be that `res.SelectFilter` doesn't work on Windows. item and dstpath are assembled using `filepath.Join` which uses `\` as path separator on Windows. You could try whether `filepath.ToSlash` solves the problem.

kitone:
@MichaelEischer thank you! You right, ToSlash fix the select filter on windows.
The test was still falling to set the permission on windows but after checking the doc https://golang.org/pkg/os/#Chmod I think my test was to specific to unix, so I move the test into restorer_unix_test.go file instead.

I will try to add a simple test case for windows.

kitone:
integration for unix and windows are green.

I guess we could merge both new case into ```restorer_test.go``` using ```filepath.ToSlash(item)``` in ```SelectFilter``` and use ```0555 | os.ModeDir``` and ```os.FileMode(0444)``` also for unix, test should work.

Let me know if you think it's better to do that or let this separate for more specific test for the future.

internal/restorer/restorer.go
```diff
@@ -242,7 +256,10 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 
 			return nil
 		},
-		leaveDir: noop,
+		leaveDir: func(node *restic.Node, target, location string) error {
+			debug.Log("first pass, leaveDir noop %q", location)
```
MichaelEischer, line 260: I think it's enough to log the visitor methods which actually do something. Please remove this log call and the one for enterDir in the second pass.

internal/restorer/restorer_unix_test.go
```diff
@@ -59,3 +60,97 @@ func TestRestorerRestoreEmptyHardlinkedFileds(t *testing.T) {
...
+		t.Errorf("checking %q, Mode() returned wrong value, want 0%o, got 0%o", file, mode, fi.Mode())
+	}
+
+	if !fi.ModTime().Equal(modtime) {
+		t.Errorf("checking %s, ModTime() returned wrong value, want %v, got %v", file, modtime, fi.ModTime())
+	}
+}
+
+// test inspired from test case https://github.com/restic/restic/issues/1212
+func TestRestorerConsistentTimestampsAndPermissions(t *testing.T) {
```
MichaelEischer, line 75: Please move this test back to `restorer_test.go`, see my suggestions below on how to unify the code between Unix and Windows.

internal/restorer/restorer_unix_test.go
```diff
@@ -59,3 +60,97 @@ func TestRestorerRestoreEmptyHardlinkedFileds(t *testing.T) {
...
+				},
+			},
+		},
+	})
+
+	res, err := NewRestorer(repo, id)
+	rtest.OK(t, err)
+
+	res.SelectFilter = func(item string, dstpath string, node *restic.Node) (selectedForRestore bool, childMayBeSelected bool) {
+		switch item {
```
MichaelEischer, line 116: please use `filepath.ToSlash(item)` here, then the switch statement can be identical across operating systems.

internal/restorer/restorer_unix_test.go
```diff
@@ -59,3 +60,97 @@ func TestRestorerRestoreEmptyHardlinkedFileds(t *testing.T) {
...
+func TestRestorerConsistentTimestampsAndPermissions(t *testing.T) {
+	timeForTest := time.Date(2019, time.January, 9, 1, 46, 40, 0, time.UTC)
+
+	repo, cleanup := repository.TestRepository(t)
+	defer cleanup()
+
+	_, id := saveSnapshot(t, repo, Snapshot{
+		Nodes: map[string]Node{
+			"dir": Dir{
+				Mode:    0750 | os.ModeDir,
```
MichaelEischer, line 84: Please create a wrapper function e.g. `normalizeFileMode` which on Windows returns `0555` for directories (`os.ModeDir`) and `0444` otherwise. Then also use it in `checkConsistentInfo`. I'm a bit surprised that filemode lies 0700 shouldn't work on Windows, at least NTFS is perfectly capable of emulating the standard unix permissions.

internal/restorer/restorer_unix_test.go
```diff
@@ -59,3 +60,97 @@ func TestRestorerRestoreEmptyHardlinkedFileds(t *testing.T) {
...
+	tempdir, cleanup := rtest.TempDir(t)
+	defer cleanup()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err = res.RestoreTo(ctx, tempdir)
+	rtest.OK(t, err)
+
+	d1, err := os.Stat(filepath.Join(tempdir, "dir"))
```
MichaelEischer, line 141: Please use a loop to check the expected file metadata. See for example https://github.com/restic/restic/blob/ab769abeaff58099449f8d36735845d89e9d1941/internal/filter/filter_test.go#L305-L308 on how to store the expected results in an array. For the path which get's passed to `checkConsistentInfo` you'll probably have to call `filepath.ToSlash`.

MichaelEischer (changes requested):
The fix itself looks good to me. The debug logging could be a bit less verbose (see comment) and the test implementation has to be unified between Unix and Windows. And this PR also needs a changelog entry.

internal/restorer/restorer.go
kitone: Done.

internal/restorer/restorer_unix_test.go
kitone: Done. The windows test file was removed.

internal/restorer/restorer_unix_test.go
kitone: Done.

internal/restorer/restorer_unix_test.go
kitone: Done, please tell me if that's what you had in mind.

internal/restorer/restorer_unix_test.go
kitone: Done, please tell me if this is right.

internal/restorer/restorer_test.go
```diff
@@ -682,10 +693,116 @@ func TestRestorerTraverseTree(t *testing.T) {
...
+	err = res.RestoreTo(ctx, tempdir)
+	rtest.OK(t, err)
+
+	var testPatterns = []struct {
+		path    string
+		modtime time.Time
+		mode    os.FileMode
+	}{
+		{"dir", timeForTest, normalizeFileMode(0750 | os.ModeDir)},
+		{"dir/file1", timeForTest, normalizeFileMode(os.FileMode(0700))},
```
MichaelEischer, line 797: I had in mind to use `filepath.Join("dir", "file1")` here instead of `dir/file1`. Currently the `filepath.ToSlash` call later on has no effect.

changelog/unreleased/pull-2906
```diff
@@ -0,0 +1,9 @@
+Bugfix: Fix inconsistent timestamps and permissions on directories restored
```
MichaelEischer, line 1: I'd suggest something like `Restore timestamps and permissions on intermediate directories`. However, I'm not really sure whether `intermediate` is the correct word here: I want to refer to the directories which have to be created to restore a file located in a subdirectory.

changelog/unreleased/pull-2906
```diff
@@ -0,0 +1,9 @@
...
+
+We've fixed the inconsistent timestamps and permissions on directories not
```
MichaelEischer, line 3: I'd include a more explicit description of the solved problem, maybe something like this:
```
When using the `--include` option with the restore command, it only restored
timestamps and permissions on directories selected by the include pattern.
Intermediate directories which are necessary to restore files located in
subdirectories, were created with default permissions. We've fixed the
`restore` command to also restore permissions for these directories.
```

changelog/unreleased/pull-2906
kitone: done

changelog/unreleased/pull-2906
kitone: done, replace with this one

internal/restorer/restorer_test.go
kitone: done

changelog/unreleased/pull-2906
```diff
@@ -0,0 +1,11 @@
+Bugfix: Restore timestamps and permissions on intermediate directories
+
+When using the `--include` option with the restore command, it only restored
+timestamps and permissions on directories selected by the include pattern.
+Intermediate directories which are necessary to restore files located in
+subdirectories, were created with default permissions. We've fixed the
+`restore` command to also restore permissions for these directories.
```
rawtaz, line 7: I suggest a small change to this:

~~~
When using the `--include` option of the restore command, restic restored
timestamps and permissions only on directories selected by the include pattern.
Intermediate directories, which are necessary to restore files located in sub-
directories, were created with default permissions. We've fixed the `restore`
command to restore timestamps and permissions for these directories as well.
~~~

kitone: Modified

MichaelEischer (approved):
LGTM. Thanks for quickly fixing all our comments :-) .
I've rebased the branch to the current master and renamed the changelog file to `issue-1212`.

PR #2913, opened 2020-09-02
mount: Make snapshots dir structure customizable

### LINKED ISSUE #2907: show snapshots in datetime-tree in fuse-mount
Output of `restic version`
--------------------------

`restic 0.9.6 (v0.9.6-0-gb723ca3) compiled with go1.13.8 on linux/386`

Restic should either represent the snapshots in a directory tree with date and time as separate path elements or show tags with slash as subdirectories under tags directory in fuse mount.
By separate date and time path elements I mean eg. `snapshots-by-date/2020/08/31/08/15` for a snapshot which was created at 2020-08-31 08:15:ss. It may be configured, because everyone has different snapshotting frequency: who does snapshots hourly he may want this 5 level depth, who does it daily he may want only 3 levels, etc. The last directory (before the snapshot's content) can be the snapshot ID to distinguish snapshots with the same date/time up to the indicated precision: `snapshots-by-date/2020/08/31/SNAPSHOTID/home/joe/backup/...`. 
But IMO it'd be easier to show tags with slash in their name as directory structure. Currently tags with slash are not shown in fuse mount. In this way users could organize huge number of snapshots simply by tags.

I have many snapshots (eg. for every hour in every day kept for years) and directory listing in fuse mount loads slowly and navigating through it by file managers are also challenging. I like to have directory structure which I can browse easier, ie. less entries per directory. IMO restic should show tags in dir structure (eg. `tags/2020/aug-31/...` for tag `"2020/aug-31"`) in fuse mount.

yes, absolutely, i can not emphasize more.

aawsome:

Makes the directory structure for displaying/grouping/sorting snapshots customizable for users.
Allows to give arbitrary directory templates containing `%i` (will be replaced by short ID), `%I` (replaced by long ID), `%u` (replaced by username), `%h` (replaced by hostname), `%t` (replaced by tag) and `%T` replaced by the time format.

Also allows to give arbitrary paths within the the directory template and the time format template to make user-defined subdirs possible.

The default behavior can be obtained by specifying the following directory templates:
```
"ids/%i"
"snapshots/%T"
"hosts/%h/%T"
"tags/%t/%T"
```
(which is the default if not explicitly specifying `--path-template`)

Have a look at the unit tests to see what is possible with the new templates.

The idea comes from #2907.

closes #2907

Checklist
---------

Note to reviewers:  `internal/fuse/snapshots_dir.go` is completely reimplemented; so better not look at the diffs when checking this file.


aawsome:
rebased after #3035 has been merged

internal/fuse/snapshots_dir.go
```diff
@@ -15,186 +17,183 @@ import (
...
+	lastCheck  time.Time
 }
 
-// SnapshotsIDSDir is a fuse directory which contains snapshots named by ids.
-type SnapshotsIDSDir struct {
-	inode   uint64
-	root    *Root
-	names   map[string]*restic.Snapshot
-	snCount int
+// NewSnapshotsDir returns a new directory containing snapshots.
```
greatroar, line 36: Stale comment.

greatroar:
The reduction in code size is nice, but I don't think a PR that introduces a new formatting language is ready for review without unit tests.

internal/fuse/snapshots_dir.go
```diff
@@ -15,186 +17,183 @@ import (
...
-type TagsDir struct {
-	inode   uint64
-	root    *Root
-	tags    map[string]bool
-	snCount int
+// replaceMultiple replaces pattern in all paths strings by repl strings.
+// If pattern does not exist in a string, the string remains unchanged.
+// If pattern exists, for each repl string a new path with replaced patterns will be added
+// Note that if the pattern exists in a string and repl is empty, the string will be removed.
+func replaceMultiple(paths []string, pattern string, repl []string) (replacedPaths []string) {
```
greatroar, line 61: I find the logic that uses this function hard to follow. Why not have a function that formats a single filename, by just looping over the content?

```go
func formatPath(template string, sn *restic.Snapshot) string {
	var (
		inVerb = false
		out    strings.Builder
	)
	for _, c := range template {
		if !inVerb {
			if c == '%' {
				inVerb = true
			} else {
				out.WriteRune(c)
			}
			continue
		}

		switch c {
		// handle the various formatting verbs
		}
		inVerb = false
	}

	return out.String()
}
```

internal/fuse/snapshots_dir.go
aawsome: is fixed.

internal/fuse/snapshots_dir.go
aawsome: Thanks for the suggestion! It is however a bit more complicated: There can be multiple tags resulting in more than one path and the "latest" link should be handled. I hence extended your code example a bit...
Hope it is now much clearer!

aawsome:
@greatroar Thanks for your feedback and suggestions! And sorry for the long delay to improve this. I changed the things as you proposed and added unit tests.

MichaelEischer (changes requested):
I like the simpler and much more flexible structure provided by this PR. I'm currently a bit worried whether the current implementation could cause some performance problems when navigating repositories with lots of snapshots.

MichaelEischer:
I've extended SnapshotsDirStructure to assemble a full fledged tree of meta directories. This simplifies directory accesses to a simple lookup and also moves the logic to handle a `latest` directory out of the `SnapshotsDir`. In total the changes address all my review comments.

MichaelEischer (approved):
LGTM. Thanks for the idea to clean up the snapshot dir structure!

PR #2927, opened 2020-09-19
Update dependencies

fd0:

Update all dependencies.


MichaelEischer:
We could probably also upgrade to https://github.com/cenkalti/backoff/releases/tag/v4.0.2 . Minio now also has a v7 release, to which we could switch in some future PR.

changelog/unreleased/pull-2927
```diff
@@ -0,0 +1,9 @@
+Change: Update dependencies, require Go >= 1.13
```
MichaelEischer, line 1: I think we should merge that with `pull-2600` which updates Go to >= 1.12

fd0: Ah, okay. i'll have a look

fd0:
> We could probably also upgrade to https://github.com/cenkalti/backoff/releases/tag/v4.0.2 

That doesn't work without changing the code, and I don't have time to look into it right now. We'll leave it at v2 for now.

changelog/unreleased/pull-2927
fd0: I've merged the changelog files.

changelog/unreleased/pull-2600
```diff
@@ -1,7 +1,11 @@
-Change: Require Go >= 1.12
+Change: Update dependencies, require Go >= 1.13
 
-Restic now requires Go to be at least 1.12. This allows simplifications in the
+Restic now requires Go to be at least 1.13. This allows simplifications in the
 build process and removing workarounds.
 
+This is also probably the last version of restic still supporting mounting
+repositories via fuse on macOS.
```
rawtaz, line 7: Perhaps add a note about why this is?

fd0: Good idea

PR #2933, opened 2020-09-19
Replace most usages of context.TODO()

### LINKED ISSUE #1434: Restic does not stop on SIGINT
## Output of `restic version`
```
restic 0.7.3 (v0.7.3-237-g06bd606)
```

```
restic backup --repo b2:my-bucket: my-directory --limit-upload 200 -o b2.connections=1
```

Backblaze B2.

## Expected behavior

Stop shortly after I press `Ctrl + C`.

## Actual behavior

Restic does not stop (or takes very long time to stop) after I press `Ctrl + C`.

The standard `Interrupt received, cleaning up` message gets printed, but then Restic carries on with the upload. Waiting/pressing `Ctrl + C` repeatedly does not help. The `-o b2.connections=...` setting does not seem to affect anything.

## Steps to reproduce the behaviour

1. Create a repo in a Backblaze B2 bucket.
2. Run `restic backup` to upload your files.
3. Press `Ctrl + C`.

I could not reproduce this with "local" repositories.

No.

No.

MichaelEischer:
`context.TODO()` is currently used in quite a few places instead of a proper context. This delays interruption of restic in some cases. In most places we already have a proper context at hand or nearby, so use these.

The only remaining usages are in `cmd/restic/lock.go` and `internal/restic/lock.go` to ensure that lock refresh and cleanup still work while the rest of restic prepares for shutdown. And for `internal/restic/readerat.go` there's no clean way to pass down a proper context.

This fixes the remaining parts to resolve #1434.

Checklist
---------

- [ ] **I don't think that there are context cancellation tests in most places. However, the existing tests should at least ensure that the context doesn't break normal executions** I have added tests for all changes in this PR
- [ ] **What someone might notice is that some commands could react a bit faster to being interrupted** There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -496,7 +496,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	if !gopts.JSON {
 		p.V("lock repository")
 	}
-	lock, err := lockRepo(repo)
+	lock, err := lockRepo(gopts.ctx, repo)
 	defer unlockRepo(lock)
 	if err != nil {
 		return err
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -394,7 +394,7 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
 func findParentSnapshot(ctx context.Context, repo restic.Repository, opts BackupOptions, targets []string) (parentID *restic.ID, err error) {
 	// Force using a parent
 	if !opts.Force && opts.Parent != "" {
-		id, err := restic.FindSnapshot(repo, opts.Parent)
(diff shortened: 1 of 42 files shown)


greatroar:
LGTM. For backendReaderAt, we could just pass a Context to the constructor, store it and explain in the docs that the ReaderAt should not be reused across contexts. It's ugly, but I've done it before in custom Readers.

MichaelEischer:
> For backendReaderAt, we could just pass a Context to the constructor, store it and explain in the docs that the ReaderAt should not be reused across contexts. It's ugly, but I've done it before in custom Readers.

I've embedded the context and added a comment that the created reader should not escape the calling function.

fd0 (approved):
Awesome, thanks! LGTM!

PR #2937, opened 2020-09-20
Don't require `self-update --output` placeholder file

### LINKED ISSUE #2491: "self-update --output" fails to create a new file on Windows
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on windows/amd64

How did you run restic exactly? (/Actual behavior/Steps to reproduce the behavior)
Got the following error when running restic self-update:
```
$ restic self-update --output restic-new.exe
CreateFile restic-new.exe: The system cannot find the file specified.
```

If I create an empty file named `restic-new.exe` on the current directory, the command runs successfully:

```
$ touch restic-new.exe

$ restic self-update --output restic-new.exe
writing restic to restic-new.exe
restic is up to date
```

I was able to update from 0.9.4 by creating such empty file.

Expected behavior
-----------------
Restic should check for new version even when the output file doesn't exist.

Related Issues
---------------
On #2248 (self-update doesn't work on Windows), the suggested workaround is to use "--output". 

It's the best backup solution I could find!

andreaso:

This removes the requirement on `restic self-update --output` to point
to a path of an existing file, to overwrite. In case the specified
path does exist we still want to verify that it's a regular file,
rather than a directory or a device, which gets overwritten.

We also want to verify that a path to a new file exists within an
existing directory. The alternative being running into that issue
after the actual download, etc has completed.

Resolves #2491

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2491
+++ b/changelog/unreleased/issue-2491
@@ -1,0 +1,9 @@
+Bugfix: Don't require `self-update --output` placeholder file
+
+`restic self-update --output /path/to/new-restic` used to require that
+new-restic was an existing file, to be overwritten. Now it's possible
+to download an updated restic binary to a new path, without first
+having to create a placeholder file.
+
+https://github.com/restic/restic/issues/2491
+https://github.com/restic/restic/pull/2937
--- a/cmd/restic/cmd_self_update.go
+++ b/cmd/restic/cmd_self_update.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"os"
+	"path/filepath"
 
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/selfupdate"
@@ -56,11 +57,18 @@ func runSelfUpdate(opts SelfUpdateOptions, gopts GlobalOptions, args []string) e
 
 	fi, err := os.Lstat(opts.Output)
 	if err != nil {
-		return err
-	}
-
-	if !fi.Mode().IsRegular() {
-		return errors.Errorf("output file %v is not a normal file, use --output to specify a different file", opts.Output)
+		dirname := filepath.Dir(opts.Output)
+		di, err := os.Lstat(dirname)
+		if err != nil {
+			return err
+		}
+		if !di.Mode().IsDir() {
+			return errors.Fatalf("output parent path %v is not a directory, use --output to specify a different file path", dirname)
+		}
+	} else {
+		if !fi.Mode().IsRegular() {
+			return errors.Fatalf("output path %v is not a normal file, use --output to specify a different file path", opts.Output)
+		}
 	}
 
 	Printf("writing restic to %v\n", opts.Output)


cmd/restic/cmd_self_update.go
```diff
@@ -56,11 +57,18 @@ func runSelfUpdate(opts SelfUpdateOptions, gopts GlobalOptions, args []string) e
...
-
-	if !fi.Mode().IsRegular() {
-		return errors.Errorf("output file %v is not a normal file, use --output to specify a different file", opts.Output)
+		dirname := filepath.Dir(opts.Output)
+		di, err := os.Lstat(dirname)
+		if err != nil {
+			return err
+		}
+		if !di.Mode().IsDir() {
+			return errors.Errorf("output parent path %v is not a directory, use --output to specify a different file path", dirname)
```
MichaelEischer, line 66: Please use `errors.Fatalf` to create the error, otherwise the error looks like restic just exploded ;-)

MichaelEischer (dismissed):
The implementation works for me, only the error reporting needs a small change.

cmd/restic/cmd_self_update.go
andreaso: What about the fact that the corresponding error message I'm replacing uses `errors.Errorf`?

MichaelEischer: An error message should not look like this:
```
output path tmp is not a normal file, use --output to specify a different file path
main.runSelfUpdate
	restic/restic/cmd/restic/cmd_self_update.go:70
main.glob..func23
	restic/restic/cmd/restic/cmd_self_update.go:30
	go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842
	go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950
	go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main
	restic/restic/cmd/restic/main.go:86
runtime.main
	go/1.15.2/libexec/src/runtime/proc.go:204
runtime.goexit
	go/1.15.2/libexec/src/runtime/asm_amd64.s:1374
```
The following is much nicer:
`Fatal: output path tmp is not a normal file, use --output to specify a different file path`
So the current behavior needs fixing. If you take a look at line 78, there `errors.Fatalf` is already used.

andreaso: Ack.

I have now (force) pushed a new commit, replacing `errors.Errorf` with `errors.Fatalf`.

fd0 (approved):
Awesome, thanks!

changelog/unreleased/issue-2491
```diff
@@ -0,0 +1,9 @@
+Bugfix: Don't require `self-update --output` placeholder file
+
+`restic self-update --output /path/to/new-restic` used to require that
+_new-restic_ was an existing file, to be overwritten. Now it's
```
rawtaz, line 4: Please remove the _ around new-restic as the changelog isn't Markdown. It's perfectly fine to put commands in some kind of delimiter though, so we don't need to change that.

MichaelEischer (commented):
I'm fine with the code changes. Please adjust the changelog, then this PR is ready for merging.

changelog/unreleased/issue-2491
andreaso: Done.

rawtaz (approved):
Awesome, thanks for the fix!

Issue #2696, opened 2020-04-18
restic is very slow to backup small files


blastrock:
Output of `restic version`
--------------------------

restic 0.9.4 compiled with go1.11.6 on linux/amd64
Actually 5a7c27ddb62d86440e60764fab4725e520f78e54

How did you run restic exactly?
-------------------------------

`~/go/bin/restic init && time ~/go/bin/restic backup . -o b2.connections=200`

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

I upload on b2 in EU region.

Benchmarks
----------

I have made a benchmark and reported the results on https://github.com/restic/restic/issues/1383#issuecomment-614834222 .

I have made a few more benchmarks and opened this new issue, as requested by @MichaelEischer. The issue is indeed specific to backing up small files (and I don't know if it's related to b2).

I have found in restic's code that the parallelism is controlled by the non-exposed variables there: https://github.com/restic/restic/blob/95da6c1c1d6fc90a52ffcd5fc69a35d103b82e51/internal/archiver/archiver.go#L104 . My initial benchmarks have shown that tweaking the FileReadConcurrency had a great influence on the backup time, so these benchmarks will focus on that variable.

- fast.com measured my upload speed at 300Mbps
- The hard drive spins at 7200RPM. hdparm reports 175MB/s for non-cached reads.
- The linux cache is cleared before each test with `sync; echo 3 | sudo tee /proc/sys/vm/drop_caches`
- I live in France and upload to b2 in their EU region
- I used -o b2.connections=200 because tests with the default value give very
  low numbers...
- Each test is run on a bare repository
- I compiled commit 5a7c27d from master
- I am using a debian unstable with linux 5.4.0

### Test #1

- 1 file of 1GiB full of random

| Drive type | FileReadConcurrency | Total time | Speed |
| --- | --- | --- | --- |
| HDD | 2 | 53s | 19MiB/s |
| HDD | 16 | 49s | 21MiB/s |

### Test #2

- 1024 files of 1MiB full of random

| Drive type | FileReadConcurrency | Total time | Speed |
| --- | --- | --- | --- |
| HDD | 2 | 5min19s | 3.2MiB/s |
| tmpfs | 2 | 5min11s | 3.3MiB/s |
| HDD | 16 | 54s | 19MiB/s |
| tmpfs | 16 | 55s | 19MiB/s |

So regardless of the storage device (HDD or RAM), the current value of FileRead seems to slow down the backup of small files significantly.

[...]

PR #2941, opened 2020-09-20
prune: Parallelize repack step

MichaelEischer:
The PR parallelizes the repack step of the prune command.

This change is basically a simpler reimplementation of the corresponding part of #2340. There are a few notable differences though:
- The repack pipeline in the PR only has 3 steps instead of 4 in #2340. The steps are: send packIDs into the pipeline, download packs, process pack + store blobs. In #2340 storing blobs is a separate pipeline step. However, this requires an additional buffer pool. Parallelizing the actual upload should be handled via #2696.
- Repacked blobs are still removed from `keepBlobs` and not tracked in an additional sync.Map. The implementation follows #2340 in that it  is prepared to support using one of multiple duplicates of a blob in the future. (This is the reason for the duplicate keepBlobs check)
- The number of workers is currently fixed at 8, adapting this depending on the backend and CPU core available should happen in another PR which adjusts other parts of restic as well.
- Usage of sync.errgroup instead of tomb.

Note: Most of the changes are indentation only, so the actual changes are best viewed when ignoring whitespace changes.

Probably in one of the issues regarding prune performance.

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ **Repack already has tests**

### DIFF
--- a/internal/repository/repack.go
+++ b/internal/repository/repack.go
@@ -19,6 +19,9 @@ const numRepackWorkers = 8
 // these packs. Each pack is loaded and the blobs listed in keepBlobs is saved
 // into a new pack. Returned is the list of obsolete packs which can then
 // be removed.
+//
+// The map keepBlobs is modified by Repack, it is used to keep track of which
+// blobs have been processed.
 func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, keepBlobs restic.BlobSet, p *restic.Progress) (obsoletePacks restic.IDSet, err error) {
 	if p != nil {
 		p.Start()
--- a/internal/repository/repack.go
+++ b/internal/repository/repack.go
@@ -2,13 +2,18 @@ package repository
 
 import (
 	"context"
+	"os"
+	"sync"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/pack"
 	"github.com/restic/restic/internal/restic"
+	"golang.org/x/sync/errgroup"
 )
+
+const numRepackWorkers = 8
 
 // Repack takes a list of packs together with a list of blobs contained in
 // these packs. Each pack is loaded and the blobs listed in keepBlobs is saved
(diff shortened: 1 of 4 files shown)


J0WI:
IMHO it would make sense to review and merge https://github.com/restic/restic/pull/2844 first before changing prune.

MichaelEischer:
> IMHO it would make sense to review and merge #2844 first before changing prune.

That's the plan.

I've primarily opened the PR to avoid duplicate work. (Well, actually it was quicker to code this and run prune afterwards than wait for the four hour unparallelized repack step to complete ;-) )

greatroar:
I only took a cursory glance at the code, but could you explain why there are still two worker pools connected by a channel? It looks like the downloader goroutines could just call the processing code synchronously. Then numRepackWorkers would be the actual number of workers, rather than half of that number.

MichaelEischer:
@greatroar The idea is to split the workers in the future into CPU- and IO-bound (to some extent). That is the number of download workers would scale with number of connections the backend can handle, whereas the repacking itself is more CPU-bound and would be limited to `GOMAXPROCS` (more workers would just unnecessarily gobble up RAM, but provide no speed-up). Right now the second part of the pipeline also handles the uploads itself and thus alternates between CPU/IO-bound.

I'm aware that `DownloadAndHash` will also require a lot CPU to verify the pack hash, but I want to keep the check of the downloaded pack file near the download itself to e.g. retry the download of damaged packs in the future. Repacking blobs is probably more expensive as it has to decrypt and hash each blob, then encrypt the blob and has the new pack file.

Right now, the number of repack workers is just an arbitrary number (I just used the same value as in `cmd/restic/delete.go`).

internal/repository/repack.go
```diff
@@ -22,91 +27,166 @@ func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, kee
...
+						h, tempfile.Name(), id)
+				}
+
+				keepMutex.Lock()
+				// recheck whether some other worker was faster
+				shouldKeep = keepBlobs.Has(h)
+				if shouldKeep {
+					keepBlobs.Delete(h)
+				}
+				keepMutex.Unlock()
```
fd0, line 150: This modifies the map passed from the caller (maps are call-by-reference). How about we use a new (local) map to keep track of which blobs were already processed?

fd0 (changes requested):
Great work! I'd love to refactor this into smaller functions (and maybe a `Repacker` struct which bundles all things we need nicely), but this will do for now.

We just need to clarify what we're going to do with the map, then it's ready to be merged.

internal/repository/repack.go
MichaelEischer: I'd actually prefer to stick to the current way of modifying the map: the list of blobs to keep can be one the main contributors to the memory usage of the prune command. When restic has to rewrite a large fraction of the packs in a repository, then your suggested `rewrittenBlobs` map would in the worst case contain nearly all existing blobs. That could increase the memory usage of prune by up to 50% (compared to restic with merged #2718 and #2842. In that constellation the repository index and the usedBlobs map should consume roughly similar amounts of memory).

fd0: Fine with me, I'll add a note to the function's documentation then we're good to go.

fd0 (approved):

Issue #2942, opened 2020-09-20
Panic in the –-exclude-larger-than filter function


DRON-666:
Output of `restic version`
--------------------------
restic 0.10.0 (v0.10.0-0-g40832b2-dirty) compiled with go1.15.2 on windows/amd64
It's custom windows build with VSS and rewrite support, so I decided to write on [forum](https://forum.restic.net/t/panic-in-the-exclude-larger-than-filter-function/3074) first.

How did you run restic exactly?
-------------------------------
restic backup --use-windows-vss **--exclude-larger-than=1g** C:\SomeBigFolder

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
local filesystem repository

Expected behavior
-----------------
restic create snapshot without error

Actual behavior
---------------
~~~
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x18 pc=0xce17f6]

goroutine 1 [running]:
main.rejectBySize.func1(0xc008c641e0, 0x3e, 0x0, 0x0, 0xc000b64040)
        **github.com/restic/restic/cmd/restic/exclude.go:305** +0x36
main.runBackup.func4(0xc008c641e0, 0x3e, 0x0, 0x0, 0x0)
        github.com/restic/restic/cmd/restic/cmd_backup.go:550 +0x8f
github.com/restic/restic/internal/archiver.(*Archiver).Save(0xc0002307e0, 0x1029180, 0xc000220ac0, 0xc008c6edc0, 0x3e, 0xc000233680, 0x3e, 0x0, 0xc008c6edc0, 0x3e, ...)
        **github.com/restic/restic/internal/archiver/archiver.go:348** +0x23a
github.com/restic/restic/internal/archiver.(*Archiver).SaveTree(0xc0002307e0, 0x1029180, 0xc000220ac0, 0xc008c4b3b0, 0x28, 0xc008bf1700, 0x0, 0x0, 0x4a, 0xc008c64190)
        github.com/restic/restic/internal/archiver/archiver.go:551 +0x4d4
github.com/restic/restic/internal/archiver.(*Archiver).SaveTree(0xc0002307e0, 0x1029180, 0xc000220ac0, 0xc008c2c820, 0x1d, 0xc008bf16c0, 0x0, 0x5bc, 0x8a877424, 0x0)
        github.com/restic/restic/internal/archiver/archiver.go:579 +0xc11
[...]

PR #2945, opened 2020-09-21
Fix nil check in rejectBySize

YoshieraHuang:

`rejectBySize` function did not check if the `fileInfo` is nil, which may lead to panic if already scanned file quickly disappearing.
This PR fixes this bug like `rejectByDevice` did.

#2942 and [forum](https://forum.restic.net/t/panic-in-the-exclude-larger-than-filter-function/3074)

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2942
+++ b/changelog/unreleased/issue-2942
@@ -1,0 +1,7 @@
+Bugfix: Make --exclude-larger-than handle disappearing files
+
+There was a small bug in the backup command's --exclude-larger-than
+option where files that disappeared between scanning and actually
+backing them up to the repository caused a panic. This is now fixed.
+
+https://github.com/restic/restic/issues/2942
--- a/cmd/restic/exclude.go
+++ b/cmd/restic/exclude.go
@@ -301,6 +301,10 @@ func rejectBySize(maxSizeStr string) (RejectFunc, error) {
 	}
 
 	return func(item string, fi os.FileInfo) bool {
+		if fi == nil {
+			return false
+		}
+
 		// directory will be ignored
 		if fi.IsDir() {
 			return false


rawtaz:
Thanks for the patch. But the "There's a new file in changelog/unreleased/ that describes the changes for our users" checkbox is checked, but I don't see a changelog file?

YoshieraHuang:
> Thanks for the patch. But the "There's a new file in changelog/unreleased/ that describes the changes for our users" checkbox is checked, but I don't see a changelog file

Now new commit has a new changelog file.

changelog/unreleased/issue-2942
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix nil check in rejectBySize
```
rawtaz, line 1: Please change to: `Bugfix: Make --exclude-larger-than handle disappearing files`.

fd0 (approved):
Great catch, thanks for fixing it! The only thing left before we can merge is that the text in the changelog needs to be reworked so that users can understand what's going on. My suggestion:

> If files restic encounters during backup are deleted before restic has a chance to process them and check the size, a panic may occur. This is fixed now.

fd0 (dismissed):
Damn, clicked the wrong button. We need to change something here :)

changelog/unreleased/issue-2942
```diff
@@ -0,0 +1,6 @@
+Bugfix: Make --exclude-larger-than handle disappearing files
+
+If files restic encounters during backup are deleted before restic has a chance to
+process them and check the size, a panic may occur. This is fixed now.
```
rawtaz, line 4: I suggest changing this to the following, to make it more in the past:

~~~
There was a small bug in the backup command's --exclude-larger-than
option where files that disappeared between scanning and actually
backing them up to the repository caused a panic. This is now fixed.
~~~

rawtaz (changes requested):

rawtaz (approved):
Thanks!

PR #2957, opened 2020-09-27
Don’t print excessively detailed debug messages on object deletion when —verbose is used

plumbeo:

PR #2840 added a new debug message printed for each deleted object to the --verbose output level. This means that the output for any operation with -v enabled, which was detailed but manageable, now can include thousands of not very informative lines (removed <index/41df4139f1>, removed <data/b62da41602>, removed <snapshot/54cda69f1a>, etc) obscuring the important information.

This PR hides the new debug messages under --verbose=2, which I think is a more appropriate level.

No.

Checklist
---------

### DIFF
--- a/cmd/restic/delete.go
+++ b/cmd/restic/delete.go
@@ -48,7 +48,7 @@ func deleteFiles(gopts GlobalOptions, ignoreError bool, repo restic.Repository, 
 						return err
 					}
 				}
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. I agree that the output might be a bit excessive. With this change deleting files also behaves similar to backing up which only lists every file with `-vv`.

PR #2961, opened 2020-09-28
Interrupt backup

### LINKED ISSUE #2960: Save an "interrupted" snapshot when backup is interrupted by SIGINT
Output of `restic version`
--------------------------
restic 0.10.0 (v0.10.0-23-g506fcfd4-dirty) compiled with go1.14.7 on linux/amd64

When `restic backup` is interrupted by SIGINT, it does not save any snapshot. The only thing that can be used in following backups is the blobs that have been saved and included in an index file that also has been already saved.

This has many drawbacks:

- Users may wonder why snapshots are missing (e.g. for automated backup runs)
- The data already backuped is not accessible (can be only accessed after a `recover` run)
- pack files may have been written which didn't make it in an written index and are effectively "lost"
- a `prune` run will completely remove the interrupted backup which may not be what users wish.

IMO, we can do better! I propose to do the following after getting an SIGINT:

- immediately stop saving of data blobs
- BUT continue saving trees (which should now finish rather quickly)
- finish the index and save it
- also save the snapshot, but with added tag "interrupted"
- if everything of the above does not finish within a given time - e.g. 15s - cancel everything to let restic terminate.

Of course, this approach also involves ignoring snapshots with the tag "interrupted" when search for the parent snapshot.

While I already have a draft implementation which does exactly the steps I pointed out, I would like to use this issue to discuss the solution:

- Is the general way to follow ok?
- Are there some pitfalls I didn't see?
- Should we make the tag name and/or time to hard finish configurable for users?

See the problems above.

A user in the forum tried to find a workaround and didn't really suceed, see:
https://forum.restic.net/t/forget-instead-of-rebuild-index-then-recover-then-prune/3088

Not today so far, but makes me happy on several other days :wink:

aawsome:

Add a handler to interrupt `restic backup` with saving an "interrupted" snapshot

closes #2960 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -404,7 +404,7 @@ func findParentSnapshot(ctx context.Context, repo restic.Repository, opts Backup
 
 	// Find last snapshot to set it as parent, if not already set
 	if !opts.Force && parentID == nil {
-		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, []string{opts.Host})
+		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, []restic.TagList{}, []string{opts.Host})
 		if err == nil {
 			parentID = &id
 		} else if err != restic.ErrNoSnapshotFound {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -404,7 +404,7 @@ func findParentSnapshot(ctx context.Context, repo restic.Repository, opts Backup
 
 	// Find last snapshot to set it as parent, if not already set
 	if !opts.Force && parentID == nil {
-		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, []restic.TagList{}, []string{opts.Host})
+		id, err := restic.FindLatestSnapshot(ctx, repo, targets, []restic.TagList{}, []restic.TagList{[]string{"interrupted"}}, []string{opts.Host})
 		if err == nil {
 			parentID = &id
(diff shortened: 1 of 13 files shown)


rawtaz:
> While I already have a draft implementation which does exactly the steps I pointed out, I would like to use this issue to discuss the solution.

Why not let this be discussed before opening another PR?

aawsome:
> Why not let this be discussed before opening another PR?

Hey, come on, it's only a draft...

I had to make sure that my idea works in principle before I wanted to propose it. And, once I had it coded, what is the sense in hiding it? Open source is sharing source code, IMHO.

rawtaz:
I know you have good intentions, but when you wrote such a nice and clear issue describing the proposed feature, I'd start with that first. Otherwise there's two numbers (issues and PRs) going up instead of one (issues), taking ever so slightly more of maintainer focus. That's the reason I asked.

cmd/restic/cmd_backup.go
```diff
@@ -601,10 +601,33 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
...
+		case <-backupFinish:
+		}
+		return nil
+	})
+
 	if !gopts.JSON {
 		p.V("start backup on %v", targets)
 	}
-	_, id, err := arch.Snapshot(gopts.ctx, targets, snapshotOpts)
+	_, id, err := arch.Snapshot(gopts.ctx, fileCtx, targets, snapshotOpts)
```
greatroar, line 630: I shudder when seeing two contexts passed around all over the codebase. Surely we can get by with a single one, if need be a custom type (Context being an interface)?

aawsome:
Closing this as IMO #3230 is much more promising.

Issue #2946, opened 2020-09-23
Windows 10 x32 report Restic as a Trojan:Win32/Zpevdo.B 


mariusda:
Hello, 

I try to download restic on a clean and updated to latest version of Microsoft Windows 10 x32 , when I download from (https://github.com/restic/restic/releases/download/v0.10.0/restic_0.10.0_windows_386.zip) after download it give the message that Windows Defender has detected the Trojan:Win32/Zpevdo.B  threath in the zip file and don't allow me to extract the executable from the zip file. 

Steps to reproduce the behavior
-------------------------------
Try to download restic on a Windows 10 x32 machine from the official repository - https://github.com/restic/restic/releases/download/v0.10.0/restic_0.10.0_windows_386.zip

Do you have any idea what may have caused this?
-----------------------------------------------
I hope that is a false positive from windows defender. 

I have checked the sha254 signature and the signature is correct.
sha256sum restic_0.10.0_windows_386.zip 
ff1a32145246a5c3e38142aa015cfbcd5dc046674d0a3f16979ff6c4eb1cfe6a  restic_0.10.0_windows_386.zip

I checked the file on virustotal.com and it give that 
 BitDefenderTheta  detected Gen:NN.ZexaE.34254.@BW@aWGQDPh and
McAfee detected GenericRXLY-CB!8AD4E3F2816C

Do you have an idea how to solve the issue?
-------------------------------------------
Recompile the program maybe ? 

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Yes it make me happy when I run it from linux, not from windows, but I need a solution for a client and I had this problem. 

Thank you for this utility that make the backup so easy. 

![Restic-Windows10x32-Virus](https://user-images.githubusercontent.com/16689517/94022604-f3412300-fdbd-11ea-967d-94bc563b3ed4.png)

ProactiveServices:
Please report this false positive to your anti-virus vendor.

mariusda:
The "vendor" is Microsoft - I will try to report this to Microsoft. Thank you !

rawtaz:
@mariusda Thanks for being careful :)

PR #2964, opened 2020-09-30
Fix instructions to reproduce the release binaries

MichaelEischer:
After the false positive virus report in #2946, I've tested the reproduction instructions and noticed that these require a few small tweaks to actually arrive at the official (compressed) binaries.

The newer instructions don't work for restic 0.10.0, and therefore I've added a hint to look at the documentation of the respective version. To reproduce the zip archives of the windows binaries, it is also necessary for the filename to exactly match, which requires passing the `--version` flag to the build helper script.

No.

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ But I've tested the instructions manually
- [x] I have added documentation for the changes (in the manual) _(sort of)_


fd0 (approved):
Thanks!

PR #2970, opened 2020-10-02
Skip fopen for file change check to avoid network penalty

### LINKED ISSUE #2969: Skip fopen on archiver for file change check
Output of `restic version`
--------------------------

restic 0.10.0 (v0.10.0-37-g5c8a350-dirty) compiled with go1.13.3 on linux/amd64

The archiver module performs an extra fopen check to increase the probability that the file information has not changed from the previous stat. An fopen operation is more expensive that just a stat, and such penalty is magnified when doing the backup of network filesystems.

A WIP pull request https://github.com/restic/restic/issues/2969 will illustrate the desired behaviour.

labkode:
Closes https://github.com/restic/restic/issues/2969

This PR reduces drastically the time to perform a backup on top of a network filesystem. It does so by reducing the number of `fopen` calls by checking for file changes before.

On a 1 million files tree, the time to backup is reduced from 60 minutes to 22 minutes. Network filesystems are sensitive to `fopen` operations, as it often requires to open the file in the data server, while a `stat` operation does not.

Checklist
---------

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -364,6 +364,30 @@ func (arch *Archiver) Save(ctx context.Context, snPath, target string, previous 
 		debug.Log("  %v regular file", target)
 		start := time.Now()
 
+		// check if the file has not changed before performing a fopen operation (more expensive, specially
+		// in network filesystems)
+		if previous != nil && !fileChanged(fi, previous, arch.IgnoreInode) {
+			if arch.allBlobsPresent(previous) {
+				debug.Log("%v hasn't changed, using old list of blobs", target)
+				arch.CompleteItem(snPath, previous, previous, ItemStats{}, time.Since(start))
+				arch.CompleteBlob(snPath, previous.Size)
+				fn.node, err = arch.nodeFromFileInfo(target, fi)
+				if err != nil {
+					return FutureNode{}, false, err
+				}
+
+				// copy list of blobs
+				fn.node.Content = previous.Content
+
+				return fn, false, nil
+			}
+
+			debug.Log("%v hasn't changed, but contents are missing!", target)
+			// There are contents missing - inform user!
+			err := errors.Errorf("parts of %v not found in the repository index; storing the file again", target)
+			arch.error(abstarget, fi, err)
+		}
+
 		// reopen file and do an fstat() on the open file to check it is still
 		// a file (and has not been exchanged for e.g. a symlink)
 		file, err := arch.FS.OpenFile(target, fs.O_RDONLY|fs.O_NOFOLLOW, 0)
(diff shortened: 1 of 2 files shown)


internal/archiver/archiver.go
```diff
@@ -364,6 +364,30 @@ func (arch *Archiver) Save(ctx context.Context, snPath, target string, previous
 		debug.Log("  %v regular file", target)
 		start := time.Now()
 
+		// check if the file has not changed before performing a fopen operation (more expensive, specially
+		// in network filesystems)
+		if previous != nil && !fileChanged(fi, previous, arch.IgnoreInode) {
```
greatroar, line 369: This is copy-pasted from below, right? Why do we need to do it twice?

labkode: Yes. The PR is work in progress. I pushed early to obtain feedback on it. To refactor that logic one has to take extra care of proper error values inside this function: https://github.com/restic/restic/blob/5c8a3505641300939ac134ae7163006b44b31ab4/internal/archiver/archiver.go#L426

MichaelEischer: > To refactor that logic one has to take extra care of proper error values inside this function: 

I'm not sure I understand what you mean by that. If `fileChanged` is checked before `OpenFile` then there's no reason to check it again afterwards. (Unless we assume that a modified file might be restored to its previous state just between the two previously mentioned methods. And that is very unlikely to happen.)

labkode: > > To refactor that logic one has to take extra care of proper error values inside this function:
> 
> I'm not sure I understand what you mean by that. If `fileChanged` is checked before `OpenFile` then there's no reason to check it again afterwards. (Unless we assume that a modified file might be restored to its previous state just between the two previously mentioned methods. And that is very unlikely to happen.)


That unlikeness is the reason there is the fopen, to ensure that between those two methods, if a file was modified, you ensure that is still a regular file (a not a symlink).

greatroar: The file is opened primarily because FileSaver.Save needs to read it if it has changed.

The possibility that @MichaelEischer describes is that Lstat in line 347 says regular file, then fileChanged in 369 detects a change, but then fileChanged in line 426 detects that the old file was put back in its place. That is so unlikely that it's hardly worth doing two fileChanged+allBlobsPresent checks for. If the file was restored, FileSaver.Save will detect that by chunking and hashing it.

labkode: @greatroar in that case I will move the fopen functionality only when the file needs to be read, currently is read for obtaining a newer file info. Let me know if you have other suggestions.

MichaelEischer: `fopen` and the `lstat` afterwards are only necessary when the file should be read again. The additional call to `lstat` is there to ensure that the opened file is still a file. So just remove the second check for `fileChanged`. If the file changes while reading it then there's a race condition anyways and it doesn't matter whether the file info used by `fileChanged` is one millisecond older or newer.

labkode: I've updated the PR simplifying the logic and removing the second fileChanged check.

MichaelEischer: Please revert the logic simplification. As previously mentioned, that way it is no longer guaranteed that the file info actually matches the file we've opened.

labkode: @MichaelEischer thanks for the quick review. I applied your suggestion, let me know if there is something else you will like to add.

MichaelEischer:
Please squash the commits and add a changelog entry (see changelog/TEMPLATE for the template).

changelog/unreleased/2970
```diff
@@ -0,0 +1,9 @@
+Enhancement: optimize fileChanged check
```
MichaelEischer, line 1: I'm afraid that's a bit too technical for most users to understand. The changelog is intended for restic users, who are not familiar with restic internals. What about `Enhancement: Optimize check for unchanged files during backup`?

labkode:
@MichaelEischer PR is ready.

changelog/unreleased/2970
```diff
@@ -0,0 +1,9 @@
...
+
+We've moved the fileChanged check up in the code to
```
MichaelEischer, line 3: I'd suggest something like this:
```
During a backup restic skips processing files which have not changed since the last backup run.
Previously this required opening each file once which can be slow on network filesystems. The
backup command now checks for file changes before opening a file. This considerably reduces
the time to create a backup on network filesystems.
```

changelog/unreleased/2970
MichaelEischer, line 1: The filename of the changelog entry should be `issue-2969`.

labkode:
@MichaelEischer I've changed the wording with your suggestions, is much clear now.

MichaelEischer (approved):
LGTM. Thanks a lot

labkode:
@MichaelEischer thanks to you for such an amazing tool!
These are some stats with this PR merged, the time to perform the backup is drastically lower.

files | 0.9.6 | pull-2970
-- | -- | --
178K | 7m | 4m
420K | 59m | 16m
850K | 1h51m | 23m

PR #2974, opened 2020-10-03
Swap deprecated GCS lib with replacement

gottwald:
It swaps the deprecated GCS library (google.golang.org/api/storage/v1) with the replacement (cloud.google.com/go/storage).

no

Checklist
---------

### DIFF
--- a/internal/backend/gs/gs.go
+++ b/internal/backend/gs/gs.go
@@ -3,13 +3,13 @@ package gs
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"path"
 	"strings"
 
+	"cloud.google.com/go/storage"
 	"github.com/pkg/errors"
 	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/debug"
@@ -18,8 +18,8 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/googleapi"
+	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
-	storage "google.golang.org/api/storage/v1"
 )
 
 // Backend stores data in a GCS bucket.
@@ -30,10 +30,11 @@ import (
 //  * storage.objects.get
 //  * storage.objects.list
 type Backend struct {
-	service      *storage.Service
+	gcsClient    *storage.Client
 	projectID    string
 	sem          *backend.Semaphore
 	bucketName   string
+	bucket       *storage.BucketHandle
 	prefix       string
 	listMaxItems int
 	backend.Layout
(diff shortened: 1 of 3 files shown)


internal/backend/gs/gs.go
```diff
@@ -59,30 +60,41 @@ func getStorageService(rt http.RoundTripper) (*storage.Service, error) {
...
+
+func (be *Backend) bucketExists(ctx context.Context, bucket *storage.BucketHandle) (bool, error) {
+	_, err := bucket.Attrs(ctx)
+	if err == storage.ErrBucketNotExist {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
```
greatroar, line 87: Last four lines can be ```return err == nil, err```.

internal/backend/gs/gs.go
```diff
@@ -394,47 +368,36 @@ func (be *Backend) List(ctx context.Context, t restic.FileType, fn func(restic.F
...
 	defer cancel()
 
-	listReq := be.service.Objects.List(be.bucketName).Context(ctx).Prefix(prefix).MaxResults(int64(be.listMaxItems))
-	for {
-		be.sem.GetToken()
-		obj, err := listReq.Do()
-		be.sem.ReleaseToken()
+	itr := be.bucket.Objects(ctx, &storage.Query{Prefix: prefix})
 
+	be.sem.GetToken()
```
MichaelEischer, line 373: The semaphore should only protect the invocation of `itr.Next()` otherwise restic will deadlock when using `gcs.connections=1`. See the `List` method of `internal/backend/s3/s3.go` for a similar case.

internal/backend/gs/gs.go
gottwald: done, thx!

internal/backend/gs/gs.go
gottwald: Thx! I moved it.

MichaelEischer:
@gottwald Please squash the commits. Did you manually test that the GCS backend works with these changes? If yes, please also test `-o gcs.connections=1`. Does the GCS library upgrade provide new noteworthy features? Or is it just updating to the current library?

gottwald:
@MichaelEischer Squashed and rebased on master.
I ran all gs tests and benchmarks with the env vars to make it test against a real bucket in my cloud account.
I also did manual tests of a backup, restore, listing snapshots and deleting snapshots.
And now also the same manual testing set with the `-o gcs.connections=1` option.

Regarding noteworthy features: I'm not aware of any.
The main difference AFAIK is that the old one is code generated and now deprecated and the new one is written by hand and a bit nicer to use.
Also did not see any significant difference in performance according to the benchmarks.

MichaelEischer (approved):
LGTM. Thanks for the library update.

PR #2975, opened 2020-10-03
Add FAQ on whether restic can resume backups

rawtaz:

This new FAQ entry explains that restic can resume interrupted backups, which previously was not clearly mentioned in the documentation.

Clarification needed as evident by discussion in #2960.

Checklist
---------

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -3,6 +3,32 @@ FAQ
 
 This is the list of Frequently Asked Questions for restic.
 
+Will restic resume an interrupted backup?
+-----------------------------------------
+
+Yes, restic will resume interrupted backups when they are re-run.
+
+When backing up, restic periodically writes index files to keep a record of
+the uploaded data. Even if there's no snapshot created in the end (due to the
+backup being interrupted), these indexes are stored in the repository for the
+data that has been uploaded so far. Next time restic runs, it is then able to
+find the uploaded data through these indexes, and thereby reference it again
+without having to upload it a second time. This effectively makes it continue
+from where it saved the last index, which should be up to a few minutes ago.
+
+It does not matter if the backup was interrupted by the user or if it was due
+to unforeseen circumstances such as connectivity issues, power loss, etc.
+Simply re-run the backup again and restic should only upload what it needs to
+in order to complete the interrupted backup and create a snapshot.
+
+Note however that during the initial backup run and any re-tries, until there
+has been a first snapshot created for the backup set (list of files and
+directories to be backed up), restic will need to re-scan the files on disk as
+there is no parent snapshot to compare the filesystem with to determine which
+files have changed. This process should however be far quicker than the
+uploading, and it's normal to see restic scan the files again when re-running
+the backup.
+
 ``restic check`` reports packs that aren't referenced in any index, is my repository broken?
 --------------------------------------------------------------------------------------------
 


doc/faq.rst
```diff
@@ -3,6 +3,19 @@ FAQ
...
+Can restic resume an interrupted backup?
+----------------------------------------
+
+Yes it can. When backing up, restic periodically writes index files to keep a record of the uploaded
+data. Even if there's no snapshot created in the end (due to the backup being interrupted), these
+indexes are stored in the database for the data that has been uploaded so far. Next time restic runs,
+it is then able to find the uploaded data through these indexes, and thereby reference it again without
+having to upload it a second time. This effectively makes it continue from where it saved the last index.
+
+It does not matter if the backup was interrupted by the user (e.g. for having to pause it temporarily)
```
greatroar, line 15: I'd leave out the parenthesized part. The user's intention is irrelevant.

greatroar:
I left a nitpicky comment, and I'll leave another: it might be good to explain that restic may rescan files when the snapshot was not completed. When doing the first (presumably interactive) backup of a directory and looking at restic's output, it's immediately clear that it is doing *some* repeated work, but not how much.

Otherwise, good FAQ.

doc/faq.rst
rawtaz: You're right, less is more.

rawtaz:
I've updated the text, a large part of it actually. Please check again!

doc/faq.rst
```diff
@@ -3,6 +3,25 @@ FAQ
 
 This is the list of Frequently Asked Questions for restic.
 
+Can restic resume an interrupted backup?
+----------------------------------------
+
+Yes it can. When backing up, restic periodically writes index files to keep a record of the uploaded data.
```
greatroar, line 9: Not sure if this was there when I left my previous comment, but I'd rephrase "Yes it can" by something like "It does by default:" (note colon). This is the first bullet point in the FAQ (deservedly so), and a three-word answer comes over as a little snug.

doc/faq.rst
```diff
@@ -3,6 +3,25 @@ FAQ
...
+Even if there's no snapshot created in the end (due to the backup being interrupted), these indexes are
+stored in the repository for the data that has been uploaded so far. Next time restic runs, it is then able
+to find the uploaded data through these indexes, and thereby reference it again without having to upload
+it a second time. This effectively makes it continue from where it saved the last index, which is recently.
```
greatroar, line 13: I'd remove "which is recently".

rawtaz: If we remove this, it becomes less clear to the user/reader that it's not a long time and thereby won't be a lot of data that's not re-used, I think. At first I wrote "up to 10 minutes", but don't want to put fixed numbers in the documentation. At this point there's nothing in the text that explains how regularly the indexes are created, and with that I feel there is room for them thinking it might be an hour or whatever, it's really undefined. Thoughts?

doc/faq.rst
rawtaz: I didn't fully agree with your suggestion, but I changed the title from "Can" to "Will" and changes the first sentence.

doc/faq.rst
rawtaz: Changed the text, see what you think.

greatroar: How about replacing "periodically" (first sentence) by "every 10 minutes" and adding a comment in the code (# promised in the FAQ)?

rawtaz: The word "periodically" was chosen because it's what fd0 used in the PR adding this functionality to restic, and it's a perfectly valid word. Could be replaced with "regularly", doesn't matter to me.

Replacing it with a fixed number is not good because it adds another data point we have to maintain in case those 10 minutes ever change. We should try to keep documentation as neutral and maintenance-free as possible.

doc/faq.rst
```diff
@@ -3,6 +3,32 @@ FAQ
...
+It does not matter if the backup was interrupted by the user or if it was due
+to unforeseen circumstances such as connectivity issues, power loss, etc.
+Simply re-run the backup again and restic should only upload what it needs to
+in order to complete the interrupted backup and create a snapshot.
+
+Note however that during the initial backup run and any re-tries, until there
+has been a first snapshot created for the backup set (list of files and
+directories to be backed up), restic will need to re-scan the files on disk as
+there is no parent snapshot to compare the filesystem with to determine which
+files has changed. This process should however be far quicker than the
```
MichaelEischer, line 28: s/has/have/

rawtaz: Fixed.

MichaelEischer (approved):
LGTM

doc/faq.rst
greatroar: As a user, I'd like to have a definite amount of time to wait. As a regular contributor of patches, I'd heed a comment above `repository.indexMagAge = 10 * time.Minute` telling me to `Update the FAQ before changing this constant`. That's all I was trying to say.

rawtaz: Yeah, I see. But it makes for a mess. Even if contributors see that, it's still a mess. I'm sorry but it should be enough with what it says now, since "a few minutes" isn't expected to mean something like half an hour or more. But if it's better to remove it that's fine too. I don't want to depend on cross-referencing comments in code though.

rawtaz: @greatroar I will go ahead and merge it as it is after having discussed the wording with @MichaelEischer as well. Thanks for your input though, it's always appreciated even if we don't agree fully on specifics all the time :)

doc/faq.rst
```diff
@@ -3,6 +3,32 @@ FAQ
...
+backup being interrupted), these indexes are stored in the repository for the
+data that has been uploaded so far. Next time restic runs, it is then able to
+find the uploaded data through these indexes, and thereby reference it again
+without having to upload it a second time. This effectively makes it continue
+from where it saved the last index, which should be up to a few minutes ago.
+
+It does not matter if the backup was interrupted by the user or if it was due
+to unforeseen circumstances such as connectivity issues, power loss, etc.
+Simply re-run the backup again and restic should only upload what it needs to
+in order to complete the interrupted backup and create a snapshot.
```
dimejo, line 22: IMHO adding a note about `restic recover` would be a good idea in case the user needs to restore files but no snapshot has been created so far.

PR #2980, opened 2020-10-05
Remove repetitive error wrapping from internal/cache

greatroar:

Many instances of errors.Wrap in this package would produce messages like "Open: open <filename>: no such file or directory"; those now omit the first "Open:" (or "Stat:", or "MkdirAll"). The function readVersion now appends its own name to the error message, rather than the function that failed, to make it easier to spot. Other function names (e.g., Load) are already added further up in the call chain.

I tested this by writing small test programs that trigger errors in os functions.

I considered opening an issue for this, as this pattern occurs in several places throughout the codebase, but I decided against it because the issue would be very open-ended. I'll submit separate PRs if and when I find more occurrences.

Checklist
---------

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -33,12 +33,12 @@ func readVersion(dir string) (v uint, err error) {
 	}
 
 	if err != nil {
-		return 0, errors.Wrap(err, "ReadFile")
+		return 0, errors.Wrap(err, "readVersion")
 	}
 
 	ver, err := strconv.ParseUint(string(buf), 10, 32)
 	if err != nil {
-		return 0, errors.Wrap(err, "ParseUint")
+		return 0, errors.Wrap(err, "readVersion")
 	}
 
 	return uint(ver), nil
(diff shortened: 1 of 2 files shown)


internal/cache/cache.go
```diff
@@ -27,21 +27,23 @@ const dirMode = 0700
 const fileMode = 0644
 
 func readVersion(dir string) (v uint, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Wrap(err, "readVersion")
```
MichaelEischer, line 32: Do we still get an accurate stacktrace which shows which return was used? `errors.Wrap` already ignores `nil` errors.

MichaelEischer (commented):
I'm a bit worried that we loose stack trace information: AFAIK the errors returned from the go standard library don't include a stacktrace and we therefore have to add it ourselves. To avoid the duplicate method name, just using `errors.WithStack` might be the best solution.

internal/cache/cache.go
greatroar: It still shows the line of the return at line 40, but it also adds the anonymous function to the trace.

greatroar:
I hadn't considered the stack traces. I pushed a new version that uses WithStack and added a few more calls to that, thought not where errors come from functions in the same package. The error handling is still not entirely consistent in how it works.

MichaelEischer (approved):
LGTM. Thanks for the cleanup.

PR #2981, opened 2020-10-05
Simplify cache directory creation

greatroar:

The package internal/cache creates a cache directory, in mkdirCacheDir, by doing a Stat to check for existence, a MkdirAll, then another Stat to check the result. Both Stats are unncessary, because MkdirAll succeeds when the directory already existed and fails (with an appropriate error message) if a non-directory existed at the given path. The only added value from mkdirCacheDir was a return value indicating whether a new directory was created, but this value wasn't always set correctly and wasn't being used anywhere in the code either, so I removed the function.

I think the only potential change in behavior might occur when symlink to an existing directory is passed to cache.New: previously, that should have failed, while now, it should succeed. I don't care enough about this corner case to write a test for it.

No.

Checklist
---------

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -96,7 +96,7 @@ func New(id string, basedir string) (c *Cache, err error) {
 		}
 	}
 
-	created, err := mkdirCacheDir(basedir)
+	err = fs.MkdirAll(basedir, 0700)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
> I think the only potential change in behavior might occur when symlink to an existing directory is passed to cache.New: previously, that should have failed, while now, it should succeed. I don't care enough about this corner case to write a test for it.

The first call to `fs.Stat` in `mkdirCacheDir` would have followed the symlink and then happily accepted the target directory. So I don't think this is a change at all.

MichaelEischer (approved):
LGTM.

PR #2984, opened 2020-10-05
snapshots: Make --json output [] instead of null when no snapshots

### LINKED ISSUE #2979: If there are no snapshots in a repository, return `[]` from restic snapshots --json
Output of `restic version`
--------------------------

restic 0.9.6 (v0.9.6-204-gf0b2dee7) compiled with go1.12.9 on linux/amd64

I ran this

```
cd "$(mktemp -d)"
mkdir repo
RESTIC_REPOSITORY="local://$PWD/repo"
RESTIC_PASSWORD=a
restic init
restic --json snapshots
```

`restic --json snapshots` printed `null`, but I would have expected `[]`. I think using `[]` would make the output easier to process correctly.

I use restic to back up an explicit list of files. When this list of files changes, restic is not able to locate the parent snapshot. This is documented in #2246.

To work around this I am using a variation of [#2246 (comment)](https://github.com/restic/restic/issues/2246#issuecomment-628022549):

```
restic backup --parent "$(restic  --json snapshots | jq -r 'max_by(.time) | .short_id // empty')"
```

This works, but when the repository does not have a snapshot `jq` traces an error message: `jq: error (at <stdin>:1): Cannot iterate over null (null)`. Printing `[]` when there are no snapshots would make this obvious approach the best approach.

Sorting output is another case when the straightforward approach is wrong by default. Normally I would do

```
restic snapshots --json | jq 'sort_by(.time) | reverse'
```

but this fails for new repositories for the same reason as above.

restic considered all my files new whenever any one was added, but the impact was just extra CPU consumed on my local PC instead of reuploading files to the cloud storage provider. That's a very safe way to fallback, and I'm happy I switched. :)

rawtaz:

Makes `--json snapshots` (both with and without `--group-by`) output `[]` instead of `null` when the list of snapshots is empty. This solves some minor parsing issues as in #2979.

Closes #2979.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2979
+++ b/changelog/unreleased/issue-2979
@@ -1,0 +1,9 @@
+Bugfix: Make snapshots --json output [] instead of null when no snapshots
+
+Restic previously output `null` instead of `[]` for the `--json snapshots`
+command, when there were no snapshots in the repository. This caused some
+minor problems when parsing the output, but is now fixed such that `[]` is
+output when the list of snapshots is empty.
+
+https://github.com/restic/restic/issues/2979
+https://github.com/restic/restic/pull/2984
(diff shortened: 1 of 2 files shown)


greatroar:
This is so subtle that I really think it needs a test. See #2985.

MichaelEischer (approved):
LGTM.

PR #2985, opened 2020-10-05
Regression test for #2979

greatroar:
This is a PR against #2984.

### DIFF
--- a/cmd/restic/cmd_snapshots_test.go
+++ b/cmd/restic/cmd_snapshots_test.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


MichaelEischer:
@greatroar Could you recreate this PR against master? I've only noticed that this PR wanted to merge into #2984 after merging that one, sorry.

rawtaz:
Yeah, please open it and rebase it so we can merge it, nice addition @greatroar!

greatroar:
Could you reopen it so I can try to change the target branch? I'm not allowed to reopen, apparently.

rawtaz:
Seems like this is because the restic:snapshots-json branch was deleted. Restic tells me that when I hover the "Reopen and comment" button.

PR #2987, opened 2020-10-05
Don't read password for generate, help or self-update

### LINKED ISSUE #2951: non-repo subcommands that don't need passwords still try to run RESTIC_PASSWORD_COMMAND
Output of `restic version`
--------------------------
restic 0.10.0 compiled with go1.15.2 on linux/amd64

This was my initial scenario -- I store my restic password in a password-store file which requires a hardware token to decrypt.  The token was not plugged in at the time that I was running the help command because I had no intention of performing any backup/restore operations at the time.  Had just updated to v0.10.0 and thought I'd look at the help output and update bash completions.
```
$ export RESTIC_PASSWORD_COMMAND='pass show my-restic-password'
$ restic help
Resolving password failed: exit status 1
```
This is easily reproduced with any password-command that fails, not specifically related to pass
```
$ export RESTIC_PASSWORD_COMMAND=/bin/false
$ restic help
Resolving password failed: exit status 1
```
Same behaviour for other commands that are unrelated to using a repository that might need a password for decryption
```
$ export RESTIC_PASSWORD_COMMAND=/bin/false
$ restic generate --bash-completion ~/.local/etc/bash_completion.d/restic.bash-completion
Resolving password failed: exit status 1
```

Not storing a repository at all yet, just trying to run `help` or `generate`.

Expected behavior
-----------------
I expected to be able to see the help output in the first command and to generate the tab-completions for bash in the second command.

The RESTIC_PASSWORD_COMMAND feature is really nice and works really well with password-store but it doesn't need to be evaluated for commands that don't require the password.

Actual behavior
---------------
I get an error about resolving passwords instead of seeing the output from `help` or `generate` -- non repo commands.

```
$ RESTIC_PASSWORD_COMMAND=/bin/false restic help
Resolving password failed: exit status 1
```

Even though I was running subcommands that have no need for a password, code inspection shows that [main.go](https://github.com/restic/restic/blob/master/cmd/restic/main.go#L57) evaluates resolvePassword() for all subcommands (except version).  resolvePassword() tries to execute the RESTIC_PASSWORD_COMMAND.  This command fails due to a no hardware token being plugged in.

One of these will solve the issue
* Move execution of resolvePassword() into each subcommand that needs a password
* Add a list of subcommands to main.go for which you skip running resolvePassword() 

It's easy enough to unset the env var as a workaround but it's not that initially obvious what's failing or why it even needs a password for `help`.

Restic is awesome.  It works very well and is an important part of my backup strategy.  Thanks for all the work you put into this!

greatroar:

restic generate, help and self-update don't touch any repo, so they don't need to check passwords (let alone run $RESTIC_PASSWORD_COMMAND). This extends the special handling of restic version to also work for those commands.

Fixes #2951.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2951
+++ b/changelog/unreleased/issue-2951
@@ -1,0 +1,9 @@
+Bugfix: restic generate, help and self-update no longer check passwords
+
+The commands `restic cache`, `generate`, `help` and `self-update` don't need
+passwords, but they previously did run the RESTIC_PASSWORD_COMMAND (if set in
+the environment), prompting users to authenticate for no reason. They now skip
+running the password command.
+
+https://github.com/restic/restic/issues/2951
+https://github.com/restic/restic/pull/2987
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -51,7 +51,7 @@ directories in an encrypted repository stored on different backends.
 			return err
 		}
 		globalOptions.extended = opts
-		if c.Name() == "version" {
+		if !needsPassword(c.Name()) {
 			return nil
 		}
 		pwd, err := resolvePassword(globalOptions, "RESTIC_PASSWORD")
@@ -69,6 +69,18 @@ directories in an encrypted repository stored on different backends.
 
 		return nil
 	},
+}
+
+// Distinguish commands that need the password from those that work without,
+// so we don't run $RESTIC_PASSWORD_COMMAND for no reason (it might prompt the
+// user for authentication).
+func needsPassword(cmd string) bool {
+	switch cmd {
+	case "cache", "generate", "help", "options", "self-update", "version":
+		return false
+	default:
+		return true
+	}
 }
 
 var logBuffer = bytes.NewBuffer(nil)


cmd/restic/main.go
```diff
@@ -71,6 +71,17 @@ directories in an encrypted repository stored on different backends.
 	},
 }
 
+// Distinguish commands that need the password from those that work without
+// (#2951).
```
MichaelEischer, line 75: I don't think we need a reference to the original issue here.

cmd/restic/main.go
```diff
@@ -71,6 +71,17 @@ directories in an encrypted repository stored on different backends.
...
+func needsPassword(cmd string) bool {
+	switch cmd {
+	case "generate", "help", "self-update", "version":
```
MichaelEischer, line 78: Please also add the "cache" and the (hidden) "options" command here.

changelog/unreleased/issue-2951
```diff
@@ -0,0 +1,9 @@
+Bugfix: restic generate, help and self-update no longer check passwords
+
+The commands restic generate, restic help and restic self-update don't need
```
MichaelEischer, line 3: I think we don't need to repeat `restic` here multiple times. I'd also enclose the command names in backticks \` to highlight them in the changelog.

changelog/unreleased/issue-2951
```diff
@@ -0,0 +1,9 @@
...
+passwords, but they previously did run the RESTIC_PASSWORD_COMMAND (if set in
+the environment), prompting users to authenticate for no reason. They now skip
+the password command.
```
MichaelEischer, line 6: s/the/running the/

MichaelEischer (changes requested):

cmd/restic/main.go
greatroar: Replaced by something more of an explanation.

changelog/unreleased/issue-2951
greatroar: Done.

changelog/unreleased/issue-2951
greatroar: Done.

MichaelEischer (approved):
LGTM. Thanks!

Issue #2694, opened 2020-04-18
exclude performance on large exclude files extremely poor


RubenKelevra:
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.14 on linux/amd64

How did you run restic exactly?
-------------------------------
`restic -r sftp:server:/home/user/backups/reponame --verbose --cache-dir="/var/cache/restic/" backup / --exclude-file=/tmp/blacklist`

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Own CentOS 7 Server

Expected behavior
-----------------
Fast operation even with large number of entries in the blacklist

Actual behavior
---------------
A large number of non-wildcard entries will reduce the performance of restic to nearly a busy loop. Usually, a complete system backup is completed in like 15 minutes with around 1 GB changes and around 30 GB of data...

I have the need to specify a large number of non-wildcard entries that exclude files that don't need to be backuped.

The full system backup runs on the same system now for 30 minutes and is still in `/etc/` - so far from complete.

Steps to reproduce the behavior
-------------------------------
Add a large number of non-wildcard entries in the exclude list and run a backup.

Do you have any idea what may have caused this?
-----------------------------------------------
Probably the non-wildcard entries are looped and compared to the full path.

Do you have an idea how to solve the issue?
-------------------------------------------
Use a hashtable for this operation.

MichaelEischer:
How many exclude entries exactly do you have? I've noticed that the exclude checks take some time when you dozens of them, but that was only in the range of a few percent slowdown.

A hashtable might work for the special case of non-wildcard entries with absolute paths. It would however conflict with #2311 which requires that the exclude entries appear to be processed in order.

Juding from my last experiments with the exclude performance, there are several low-hanging fruits in the exclude code, like preprocessing the excludes once, avoid checks whose results are not used, ... . I've drafted some of the optimizations in https://github.com/MichaelEischer/restic/tree/fast-excludes , but the code in that branch is still extremely hacky and nowhere near ready for a PR. Just those changes gave me a speed-up of 10-20 times for the exclude checks.

RubenKelevra:
> How many exclude entries exactly do you have? I've noticed that the exclude checks take some time when you dozens of them, but that was only in the range of a few percent slowdown.

Well, they are dynamically generated, so the number is not steady. But here's an example.

20 or so are wildcards, everything else are just non-wildcards

```
$ wc -l /tmp/blacklist
350802 /tmp/blacklist
```

[...]

PR #2997, opened 2020-10-07
Speedup exclude/include checking

MichaelEischer:
Backup (and restore) slow down a lot if a large number of excludes is used, see e.g. #2694. This PR optimizes the pattern processing to reduce the number and size of memory allocations. In total this speeds up the benchmarks by a factor between 5 and 100:

```
name                          old time/op    new time/op    delta
FilterPatterns/Relative-4       62.1ms ±15%    11.0ms ± 2%  -82.25%  (p=0.000 n=9+10)
FilterPatterns/Absolute-4        111ms ±10%       9ms ± 3%  -92.24%  (p=0.000 n=10+10)
FilterPatterns/Wildcard-4        393ms ±15%      36ms ± 2%  -90.73%  (p=0.000 n=10+9)
FilterPatterns/ManyNoMatch-4     10.0s ± 3%      0.1s ± 6%  -98.98%  (p=0.000 n=10+9)

name                          old alloc/op   new alloc/op   delta
FilterPatterns/Relative-4       16.4MB ± 0%     3.6MB ± 0%  -78.25%  (p=0.000 n=10+9)
FilterPatterns/Absolute-4       31.4MB ± 0%     3.6MB ± 0%  -88.62%  (p=0.000 n=9+10)
FilterPatterns/Wildcard-4        168MB ± 0%      20MB ± 0%  -88.25%  (p=0.000 n=10+10)
FilterPatterns/ManyNoMatch-4    3.23GB ± 0%    0.00GB ± 0%  -99.89%  (p=0.000 n=10+10)

name                          old allocs/op  new allocs/op  delta
FilterPatterns/Relative-4         178k ± 0%       22k ± 0%  -87.50%  (p=0.000 n=10+10)
FilterPatterns/Absolute-4         266k ± 0%       22k ± 0%  -91.67%  (p=0.000 n=10+10)
FilterPatterns/Wildcard-4        1.87M ± 0%     0.09M ± 0%  -95.27%  (p=0.000 n=10+10)
FilterPatterns/ManyNoMatch-4     17.7M ± 0%      0.0M ± 0%  -99.88%  (p=0.000 n=9+10)
```

The`ManyNoMatch` test uses 200 exclude patterns which are suffixed such that these do not match any filenames, but require matching against the full pattern which is the worst-case performance wise. Originally I've tried to add all 22k tested filepaths to the filter patterns, however that was way too slow.

Assuming restic should be able to check 20k excludes per second (i.e., 1.2 million excludes per minute), one "op" of the `ManyNoMatch` benchmark shouldn't take much more than one seconds. With the optimizations I'm currently able to increase the number of exclude pattern to 2000.

The exclude performance problems were in parts discussed in #2694.

Checklist
---------

- [x] I have added tests for all changes in this PR **The code already had a lot of tests, I've added some special case tests to increase coverage to 95%**

### DIFF
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -149,8 +149,13 @@ func match(patterns Pattern, strs []string) (matched bool, err error) {
 	}
 
 	if len(patterns) <= len(strs) {
+		maxOffset := len(strs) - len(patterns)
+		// special case absolute patterns
+		if patterns[0] == "" {
+			maxOffset = 0
+		}
 	outer:
-		for offset := len(strs) - len(patterns); offset >= 0; offset-- {
+		for offset := maxOffset; offset >= 0; offset-- {
 
 			for i := len(patterns) - 1; i >= 0; i-- {
 				ok, err := filepath.Match(patterns[i], strs[offset+i])
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -11,8 +11,13 @@ import (
 // second argument.
 var ErrBadString = errors.New("filter.Match: string is empty")
 
+type patternPart struct {
+	pattern  string
+	isSimple bool
+}
+
 // Pattern represents a preparsed filter pattern
-type Pattern []string
+type Pattern []patternPart
 
 func prepareStr(str string) ([]string, error) {
 	if str == "" {
@@ -35,7 +40,14 @@ func preparePattern(pattern string) Pattern {
 		pattern = strings.Replace(pattern, string(filepath.Separator), "/", -1)
 	}
 
-	return strings.Split(pattern, "/")
+	parts := strings.Split(pattern, "/")
+	patterns := make([]patternPart, len(parts))
+	for i, part := range parts {
+		isSimple := !strings.ContainsAny(part, "\\[]*?")
+		patterns[i] = patternPart{part, isSimple}
+	}
+
+	return patterns
 }
 
 // Match returns true if str matches the pattern. When the pattern is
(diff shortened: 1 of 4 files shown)


MichaelEischer:
I've fixed the "ManyNoMatch" (previously "Many") benchmark to actually work as intended and to cause the worst-case performance.

internal/filter/filter.go
```diff
@@ -11,6 +11,45 @@ import (
...
+
+func prepareStr(str string) ([]string, error) {
+	if str == "" {
+		return nil, ErrBadString
+	}
+
+	// convert file path separator to '/'
+	if filepath.Separator != '/' {
+		str = strings.Replace(str, string(filepath.Separator), "/", -1)
+	}
```
greatroar, line 30: str = filepath.ToSlash(str) is equivalent but can be compiled away entirely on Unix.

MichaelEischer: Thanks for the hint. Now the code coverage is at 98% :-)

MichaelEischer:
Profiling the optimized code now reports that more than half of the time is spent in `hasDoubleWildcard`. However, optimizing that function away just shifts the costs to some other part of the `match` function. So it looks like the string comparisons begin to be memory-bound.

greatroar:
Impressive result. I implemented my glob-to-regexp translation idea, and while it gets rid of the allocations, it's almost twice as slow as the current code.

greatroar:
MichaelEischer/restic#1 gives a further 25% speedup on the hardest benchmark.

MichaelEischer:
@greatroar Wow, thanks for the extra speed-up. I've updated the comparison in the PR description, now we're at a speed-up by a smooth factor of 100. 🥳

fd0 (approved):

PR #2998, opened 2020-10-07
Systematize documentation of environment variables

greatroar:

Cleans up the documentation of restic backup's environment variables.

Cache locations were documented inconsistently in three places.

The backup docs mentioned PATH being used to find fusermount, which is never run by restic backup, while not mentioning rclone and ssh. 

The notion of a "system-wide" environment variable makes no sense, so I 
put everything in one table.

Follow-up to #2769.

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -416,6 +416,8 @@ environment variables. The following lists these environment variables:
     RESTIC_CACHE_DIR                    Location of the cache directory
     RESTIC_PROGRESS_FPS                 Frames per second by which the progress bar is updated
 
+    TMPDIR                              Location for temporary files
+
     AWS_ACCESS_KEY_ID                   Amazon S3 access key ID
     AWS_SECRET_ACCESS_KEY               Amazon S3 secret access key
     AWS_DEFAULT_REGION                  Amazon S3 default region
@@ -453,12 +455,12 @@ environment variables. The following lists these environment variables:
 
     RCLONE_BWLIMIT                      rclone bandwidth limit
 
-In addition to restic-specific environment variables, the following system-wide environment variables
-are taken into account for various operations:
+See :ref:`caching` for the rules concerning cache locations when
+``RESTIC_CACHE_DIR`` is not set.
 
- * ``$XDG_CACHE_HOME/restic``, ``$HOME/.cache/restic``: :ref:`caching`.
- * ``$TMPDIR``: :ref:`temporary_files`.
- * ``$PATH/fusermount``: Binary for ``restic mount``.
+The external programs that restic may execute include ``rclone`` (for rclone
+backends) and ``ssh`` (for the SFTP backend). These may respond to further
+environment variables and configuration files; see their respective manuals.
 
 
 Exit status codes
--- a/doc/cache.rst
+++ b/doc/cache.rst
@@ -3,22 +3,12 @@ Local Cache
 ***********
 
 In order to speed up certain operations, restic manages a local cache of data.
-This document describes the data structures for the local cache with version 1.
+The location of the cache directory depends on the operating system and the
+environment; see :ref:`caching`.
 
-Versions
-========
-
-The cache directory is selected according to the `XDG base dir specification
-<https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html>`__.
 Each repository has its own cache sub-directory, consisting of the repository ID
 which is chosen at ``init``. All cache directories for different repos are
 independent of each other.
-
-The cache dir for a repo contains a file named ``version``, which contains a
-single ASCII integer line that stands for the current version of the cache. If
-a lower version number is found the cache is recreated with the current
-version. If a higher version number is found the cache is ignored and left as
-is.
 
 Snapshots, Data and Indexes
 ===========================
--- a/doc/manual_rest.rst
+++ b/doc/manual_rest.rst
@@ -398,9 +398,13 @@ This allows faster operations, since meta data does not need to be loaded from
 a remote repository. The cache is automatically created, usually in an
 OS-specific cache folder:
 
- * Linux/other: ``~/.cache/restic`` (or ``$XDG_CACHE_HOME/restic``)
+ * Linux/other: ``$XDG_CACHE_HOME/restic``, or ``~/.cache/restic`` if
+   ``XDG_CACHE_HOME`` is not set
  * macOS: ``~/Library/Caches/restic``
  * Windows: ``%LOCALAPPDATA%/restic``
+
+If the relevant environment variables are not set, restic exits with an error
+message.
 
 The command line parameter ``--cache-dir`` or the environment variable
 ``$RESTIC_CACHE_DIR`` can be used to override the default cache location.  The
--- a/doc/050_restore.rst
+++ b/doc/050_restore.rst
@@ -71,10 +71,11 @@ command to serve the repository with FUSE:
     Now serving /srv/restic-repo at /mnt/restic
     When finished, quit with Ctrl-c or umount the mountpoint.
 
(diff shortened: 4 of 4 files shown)


doc/040_backup.rst
```diff
@@ -453,12 +459,11 @@ environment variables. The following lists these environment variables:
...
-In addition to restic-specific environment variables, the following system-wide environment variables
-are taken into account for various operations:
+See :ref:`caching` for the rules concerning cache locations.
 
- * ``$XDG_CACHE_HOME/restic``, ``$HOME/.cache/restic``: :ref:`caching`.
- * ``$TMPDIR``: :ref:`temporary_files`.
- * ``$PATH/fusermount``: Binary for ``restic mount``.
+The external programs that restic may execute include ``rclone`` (for rclone
+backends) and ``ssh`` (for the SFTP backend). These may respond to further
+environment variables; see their respective manuals.
```
rawtaz, line 466: Really good addition!

doc/040_backup.rst
```diff
@@ -416,6 +416,12 @@ environment variables. The following lists these environment variables:
     RESTIC_CACHE_DIR                    Location of the cache directory
     RESTIC_PROGRESS_FPS                 Frames per second by which the progress bar is updated
 
+    HOME                                Backup location for caches
+    LocalAppData                        Backup location for caches
+    PATH                                Lookup path for external programs
+    TMPDIR                              Location for temporary files
+    XDG_CACHE_HOME                      Backup location for caches
```
rawtaz, line 423: I think that this is out of scope for this documentation. These are well established standard variables/paths, that doesn't belong in a list of environment variables that is more specific to restic's workings. Also, most of them are already mentioned/described in the caching section. I suggest we don't include this change, what do you think @fd0 and @MichaelEischer?

rawtaz (changes requested):
I added some comments :)

doc/cache.rst
```diff
@@ -20,6 +19,9 @@ a lower version number is found the cache is recreated with the current
 version. If a higher version number is found the cache is ignored and left as
 is.
 
+The rest of this chapter describes the data structures for the local cache with
+version 1.
```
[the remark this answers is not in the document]
greatroar: This was already here, I mostly moved it around (and rephrased it a tiny bit).

doc/manual_rest.rst
```diff
@@ -398,9 +398,13 @@ This allows faster operations, since meta data does not need to be loaded from
 a remote repository. The cache is automatically created, usually in an
 OS-specific cache folder:
 
- * Linux/other: ``~/.cache/restic`` (or ``$XDG_CACHE_HOME/restic``)
- * macOS: ``~/Library/Caches/restic``
+ * macOS: ``$HOME/Library/Caches/restic``
  * Windows: ``%LOCALAPPDATA%/restic``
+ * Linux/other: ``$XDG_CACHE_HOME/restic``, or ``$HOME/.cache/restic`` if
+   ``XDG_CACHE_HOME`` is not set
```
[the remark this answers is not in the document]
greatroar: I found it rather odd to have "other" as the first bullet point. I get restore the ~ if you feel strongly about it :)

doc/040_backup.rst
greatroar: Knowing where data goes is crucial when operating a security-focused program. The fact that these are all customary doesn't imply that restic uses them, rather than using a custom RESTIC_TEMPDIR, or /dev/shm, or /run/$uid/restic, or anonymous files in its working directory.

As a practical example, if your homedir is encrypted but the rest of the disk is not, it's important to know that and to know exactly how to keep restic from writing outside the homedir. (I think restic only writes encrypted data to TMPDIR, but I'm not 100% sure.)

Finally, restic has actually used different combinations of variables for locating the cache in previous releases, so the variables aren't that well-established...

rawtaz: I'm certainly not convinced. Please consider the following:

- The variables that are relevant for caches are mentioned in the caching section, and shouldn't be in here because that is redundant and these types of variables are not the goal of this list. If there's a need to describe that e.g. $HOME is a "backup location" for caches, then that should be described in the caching section which already talks about this type of information.

- $PATH is just too common to justify mentioning here. If we're going to mention such common variables, there are others to mention as well, and it's not a good idea. Same for $TMPDIR which is a common one too.

doc/040_backup.rst
```diff
@@ -453,12 +459,11 @@ environment variables. The following lists these environment variables:
 
     RCLONE_BWLIMIT                      rclone bandwidth limit
 
-In addition to restic-specific environment variables, the following system-wide environment variables
-are taken into account for various operations:
+See :ref:`caching` for the rules concerning cache locations.
```
MichaelEischer, line 462: This should mention, that this is about the default cache location. If `RESTIC_CACHE_DIR` is set, then that has precedence.

doc/040_backup.rst
MichaelEischer: Listing 3 (or actually 4) variables related to the cache location will probably confuse more that it helps. Regarding the `PATH` variable I think that it is that common that the documentation should only mention when it is _not_ used (e.g. if restic had hardcoded absolute paths).

`TMPDIR` on the other hand is useful to know as that way a user can e.g. direct it to a ramdisk.

doc/manual_rest.rst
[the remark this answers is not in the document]
greatroar: Ok.

doc/040_backup.rst
greatroar: I removed everything except TMPDIR.

doc/040_backup.rst
greatroar: Changed to:

    See :ref:`caching` for the rules concerning cache locations when
    ``RESTIC_CACHE_DIR`` is not set.

greatroar:
The test failure is unrelated, but maybe worth the trouble of tracking down:

```
--- FAIL: TestArchiverAbortEarlyOnError (0.02s)
    --- FAIL: TestArchiverAbortEarlyOnError/#01 (0.01s)
        archiver_test.go:1918: chdir to /tmp/restic-test-143559427
        archiver_test.go:1946: Snapshot return error: test error
        archiver_test.go:1948: track fs: map[.:1 dir:1 dir/file1:1 dir/file2:1 dir/file3:1 dir/file4:1 dir/file5:1 dir/file6:1 dir/file7:1]
        archiver_test.go:1952: opened dir/file7 1 times, want 0
        archiver_test.go:1955: chdir back to /home/travis/gopath/src/github.com/restic/restic/internal/archiver
FAIL
```

doc/cache.rst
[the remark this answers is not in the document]
rawtaz: @greatroar We discussed this a bit and would like to remove the entire Versions heading and its paragraphs, except the "Each repository has its own cache sub-directory, consisting of the repository ID
which is chosen at ``init``. All cache directories for different repos are
independent of each other." one which should be kept (instead of the Versions heading).

doc/040_backup.rst
```diff
@@ -453,12 +459,11 @@ environment variables. The following lists these environment variables:
...
 
- * ``$XDG_CACHE_HOME/restic``, ``$HOME/.cache/restic``: :ref:`caching`.
- * ``$TMPDIR``: :ref:`temporary_files`.
- * ``$PATH/fusermount``: Binary for ``restic mount``.
```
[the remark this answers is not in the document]
greatroar: Done by cherry-picking a commit from #2833 and adding a mention of fusermount.

doc/cache.rst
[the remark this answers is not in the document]
greatroar: Ok, done.

rawtaz (approved):
Bewm!

PR #2999, opened 2020-10-08
backup: Improve wording for --one-file-system description

### LINKED ISSUE #2109: BTRFS subvolumes and one-file-system
### I'm going to preface this by saying that I think the current behavior is arguably **correct**, it was just surprising.

Output of `restic version`
--------------------------
restic 0.9.3 compiled with go1.9.4 on linux/amd64

```
nice -n 10 "${SCRIPTHOME}/restic" ${SFTPOPT} -r "${REPOBASE}/${SET}" \ 
  --password-file="${SCRIPTHOME}/${SET}/password.txt" \ 
  backup **--one-file-system** \ 
  --exclude-file="${TMPEXCLUDE}" \ 
  --files-from="${SCRIPTHOME}/${SET}/include.txt" 2>&1 | tee "${LOG}"
```

The include.txt contents in this case are:
/boot/
/etc/
/root/
/var/
/files/

Those paths cover three filesystems, / (xfs), /boot (xfs), and /files (btrfs).

This backup set writes to a locally-mounted directory.

Expected behavior
-----------------
I expected to get the full contents of the /files mount, minus one directory that was listed in the exclusion file.

Actual behavior
---------------
In /files, only the directories that were **not** a subvolume were backed up.  

Create a btrfs filesystem, then within it create both normal subdirectories and subvolumes.
```
dd if=/dev/zero of=testdevice.dd bs=1M count=100  
mkfs.btrfs testdevice.dd  
mkdir test  
mount -o loop testdevice.dd test  
btrfs subvolume create test/subvol  
mkdir test/directory  
echo "Foo" > test/subvol/file1.txt  
echo "Bar" > test/directory/file2.txt
```
Once that is done run a backup against the mount point with "--one-file-system".  Restic will backup "test/directory/file2.txt", but not "test/subvol/file1.txt".

A BTRFS subvolume has some characteristics that make it look like a different filesystem.  For instance, moves between subvolumes are not atomic.  All BTRFS volumes contain an automatically created subvolume called "top-level".
[https://btrfs.wiki.kernel.org/index.php/Manpage/btrfs-subvolume](https://btrfs.wiki.kernel.org/index.php/Manpage/btrfs-subvolume)

I suspect that restic's "--one-file-system" logic is aware of the distinctions and correctly (from its view) treats them as separate file systems.

If the code is indeed correct, I suggest adding a caution statement to the documentation.  Perhaps something like this:

> Note that some file systems (such as BTRFS) support subvolumes, and these may be treated by restic as separate file systems.

One could argue that "--one-file-system" should not behave this way, but I think that starts to get into the particulars of individual expectations and goals.  Rather than adding complexity to the code base to handle these possibly conflicting goals, I think it is simplest to document the behavior and let people work around it as desired (specify each subvolume path, or just don't use that flag at all).

Restic provides peace of mind on a budget.  :)

rawtaz:

Improve wording for --one-file-system description to make it clearer that it also applies to mountpoints and subvolumes.

Closes #2109.

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -275,7 +275,7 @@ On most Unixy shells, you can either quote or use backslashes. For example:
 By specifying the option ``--one-file-system`` you can instruct restic
 to only backup files from the file systems the initially specified files
 or directories reside on. In other words, it will prevent restic from crossing
-filesystem boundaries when performing a backup.
+filesystem boundaries and subvolumes when performing a backup.
 
 For example, if you backup ``/`` with this option and you have external
 media mounted under ``/media/usb`` then restic will not back up ``/media/usb``
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -109,7 +109,7 @@ func init() {
 	f.StringArrayVar(&backupOptions.InsensitiveExcludes, "iexclude", nil, "same as --exclude `pattern` but ignores the casing of filenames")
 	f.StringArrayVar(&backupOptions.ExcludeFiles, "exclude-file", nil, "read exclude patterns from a `file` (can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.InsensitiveExcludeFiles, "iexclude-file", nil, "same as --exclude-file but ignores casing of `file`names in patterns")
-	f.BoolVarP(&backupOptions.ExcludeOtherFS, "one-file-system", "x", false, "exclude other file systems")
+	f.BoolVarP(&backupOptions.ExcludeOtherFS, "one-file-system", "x", false, "exclude other file systems, don't cross filesystem boundaries and subvolumes")
 	f.StringArrayVar(&backupOptions.ExcludeIfPresent, "exclude-if-present", nil, "takes `filename[:header]`, exclude contents of directories containing filename (except filename itself) if header of that file is as provided (can be specified multiple times)")
 	f.BoolVar(&backupOptions.ExcludeCaches, "exclude-caches", false, `excludes cache directories that are marked with a CACHEDIR.TAG file. See https://bford.info/cachedir/ for the Cache Directory Tagging Standard`)
 	f.StringVar(&backupOptions.ExcludeLargerThan, "exclude-larger-than", "", "max `size` of the files to be backed up (allowed suffixes: k/K, m/M, g/G, t/T)")
--- a/doc/manual_rest.rst
+++ b/doc/manual_rest.rst
@@ -105,7 +105,7 @@ command:
           --iexclude pattern                       same as --exclude pattern but ignores the casing of filenames
           --iexclude-file file                     same as --exclude-file but ignores casing of filenames in patterns
           --ignore-inode                           ignore inode number changes when checking for modified files
-      -x, --one-file-system                        exclude other file systems
+      -x, --one-file-system                        exclude other file systems, don't cross filesystem boundaries and subvolumes
           --parent snapshot                        use this parent snapshot (default: last snapshot in the repo that has the same target files/directories)
           --stdin                                  read backup from stdin
           --stdin-filename filename                filename to use when reading from stdin (default "stdin")


rawtaz:
I'm wondering if this is too specific, e.g. can there be subvolumes that *will* be crossed into by restic? If so, the description will be misguiding for those cases. Otherwise I think it should be fine even if it doesn't explicitly mention BTRFS.

MichaelEischer:
What happens when using bind mounts? These probably still count as the same filesystem.

MichaelEischer:
I just tested bind mounts, as I presumed the `--one-filesystem` option crosses bind mounts within the same filesystem. Maybe just state that the option `exclude[s] other file systems and subvolumes`?

rawtaz:
Does anyone have an ZFS setup they can test with to see how traversing is done with that type of filesystem?

doc/040_backup.rst
```diff
@@ -216,7 +216,7 @@ On most Unixy shells, you can either quote or use backslashes. For example:
 By specifying the option ``--one-file-system`` you can instruct restic
 to only backup files from the file systems the initially specified files
 or directories reside on. In other words, it will prevent restic from crossing
-filesystem boundaries when performing a backup.
+filesystem boundaries, mountpoints and subvolumes when performing a backup.
```
MichaelEischer, line 219: I'd prefer to just explicitly add `subvolumes` to the previously mentioned `filesystem`. mountpoints on their own are only a necessary, but no sufficient condition. For example bind mounts within a filesystem are crossed even with the `--one-file-system` option.

rawtaz:
> Does anyone have an ZFS setup they can test with to see how traversing is done with that type of filesystem?

I just tested this, and `--one-file-system` does not have any effect on a ZFS filesystem being mounted, despite ZFS calling it a "filesystem":

~~~
# pwd
/root


# ./restic -r temp backup /data/foo/apa.txt
repository 945fd9d4 opened successfully, password is correct
created new cache in /root/.cache/restic

Added to the repo: 1.073 KiB

processed 1 files, 5 B in 0:00
snapshot 9a06d7c1 saved


# ./restic -r temp backup --one-file-system /data/foo/apa.txt
repository 945fd9d4 opened successfully, password is correct
using parent snapshot 9a06d7c1

Files:           0 new,     0 changed,     1 unmodified
Dirs:            0 new,     0 changed,     2 unmodified
Added to the repo: 0 B

processed 1 files, 5 B in 0:00
snapshot f44eb4da saved


# vi /data/foo/apa.txt


# ./restic -r temp backup --one-file-system /data/foo/apa.txt
repository 945fd9d4 opened successfully, password is correct
using parent snapshot f44eb4da

Files:           0 new,     1 changed,     0 unmodified
Dirs:            0 new,     2 changed,     0 unmodified
Added to the repo: 1.085 KiB

processed 1 files, 16 B in 0:00
snapshot 1cc4af11 saved


# vi /data/foo/apa.txt


# ./restic -r temp backup -x /data/foo/apa.txt
repository 945fd9d4 opened successfully, password is correct
using parent snapshot 1cc4af11

Files:           0 new,     1 changed,     0 unmodified
Dirs:            0 new,     2 changed,     0 unmodified
Added to the repo: 1.096 KiB

processed 1 files, 27 B in 0:00
snapshot 8be5ba5b saved
~~~

rawtaz:
So, is the situation with `--one-file-system` as follows?:

1. Regular filesystem-on-other-disk/device are not traversed into.
1. Bind mounts are traveresed into.
1. ZFS filesystems, even though they are on a completely separate pool on separate disks, are traversed into (this is quite contrary to # 1 above).
1. BTRFS need to be tested as well I think.
1. I would expect that LVM volumes that are mounted are *not* traversed into.
1. How it behaves on Windows I haven't tested.

MichaelEischer:
Which filesystem is mounted where? If guess `/data/foo` is a mountpoint?

`--one-file-system` doesn't have any effect when just backing up a single file. The current implementation gathers a list of filesystem IDs for all specified backup target paths (`/data/foo/apa.txt` in this case) and then while walking the filesystem ignores all files/folders outside the whitelisted filesystems. 

So a test setup would have to be something like this:
```
/data (on root partition)
/data/foo (on subvolume)
/data/foo/something
```
then run `backup -x /data`. This should normally skip everything in `/data/foo`.

shd2h:
I ran through some quick tests for both zfs and btrfs, which hopefully are some help. I set both up with a second volume mounted underneath the filesystem mountpoint.

```
root@ubu2010:~# zfs list
NAME            USED  AVAIL     REFER  MOUNTPOINT
zfspool         190K  9.20G       24K  /zfspool
zfspool/alpha    24K  9.20G       24K  /fs_test/zfs
zfspool/beta     24K  9.20G       24K  /fs_test/zfs/anothervol

root@ubu2010:~# btrfs fi show /fs_test/btrfs
Label: 'antitank'  uuid: 0151d3ba-84b9-4546-b2c2-d1850484322f
        Total devices 1 FS bytes used 216.00KiB
        devid    1 size 10.00GiB used 536.00MiB path /dev/sdb
root@ubu2010:~# btrfs subvol list /fs_test/btrfs
ID 257 gen 9 top level 5 path subvolume1
```


Complete filesystem layout was as below:

```
root@ubu2010:~# ls -lhR /fs_test/btrfs
/fs_test/btrfs:
total 4.0K
-rw-r--r-- 1 root root 2.9K Apr  2 14:35 normal_file
drwxr-xr-x 1 root root   28 Apr  2 14:35 subvolume1

/fs_test/btrfs/subvolume1:
total 4.0K
-rw-r--r-- 1 root root 3.0K Apr  2 14:35 subvolume_file

root@ubu2010:~# ls -lhR /fs_test/zfs
/fs_test/zfs:
total 4.0K
drwxr-xr-x 2 root root    3 Apr  2 14:44 anothervol
-rw-r--r-- 1 root root 2.9K Apr  2 14:35 normal_file

/fs_test/zfs/anothervol:
total 3.5K
-rw-r--r-- 1 root root 3.0K Apr  2 14:35 subvolume_file
```


backing up the parent dir containing both filesystem mountpoints, using the --one-file-system flag, nothing is backed up:

```
root@ubu2010:~# restic backup --one-file-system /fs_test
repository 4739356d opened successfully, password is correct

Dirs:            3 new,     0 changed,     0 unmodified
Added to the repo: 1.023 KiB

snapshot d6b1b075 saved
root@ubu2010:~# restic ls d6b1b075
repository 4739356d opened successfully, password is correct
snapshot d6b1b075 of [/fs_test] filtered by [] at 2021-04-02 14:55:24.212439704 +0000 UTC):
/fs_test
/fs_test/btrfs
/fs_test/zfs
```


backing up the btrfs mountpoint, the contents of the subvolume are excluded:

```
root@ubu2010:~# restic backup --one-file-system /fs_test/btrfs
repository 4739356d opened successfully, password is correct

Dirs:            3 new,     0 changed,     0 unmodified
Added to the repo: 4.224 KiB

processed 1 files, 2.857 KiB in 0:00
snapshot 9ae34f42 saved
root@ubu2010:~# restic ls 9ae34f42
repository 4739356d opened successfully, password is correct
snapshot 9ae34f42 of [/fs_test/btrfs] filtered by [] at 2021-04-02 14:55:29.844439118 +0000 UTC):
/fs_test
/fs_test/btrfs
/fs_test/btrfs/normal_file
/fs_test/btrfs/subvolume1
```


doing the same for zfs, the contents of the second "nested" volume are excluded:

```
root@ubu2010:~# restic backup --one-file-system /fs_test/zfs
repository 4739356d opened successfully, password is correct

Dirs:            3 new,     0 changed,     0 unmodified
Added to the repo: 1.360 KiB

processed 1 files, 2.857 KiB in 0:00
snapshot ad22eaff saved
root@ubu2010:~# restic ls ad22eaff
repository 4739356d opened successfully, password is correct
snapshot ad22eaff of [/fs_test/zfs] filtered by [] at 2021-04-02 14:55:32.980258298 +0000 UTC):
/fs_test
/fs_test/zfs
/fs_test/zfs/anothervol
/fs_test/zfs/normal_file
```

MichaelEischer:
@shd2h Thanks for the tests.

So this confirms that subvolumes are just treated like filesystems and thus `--one-file-system` does not traverse mountpoints to other filesystems or subvolumes. 

@rawtaz The documentation should then read:
`[...] it will prevent restic from crossing filesystem boundaries and subvolumes when performing a backup [...]`

MichaelEischer (approved):
LGTM.

PR #3000, opened 2020-10-08
Call enterDir/leaveDir if a directory's child may be selected

### LINKED ISSUE #1402: Restoring with --include creates directories with wrong permissions on the way down.

### LINKED ISSUE #2563: Wrong Permissions when restoring backup
Output of `restic version`

I executed the following command as a root user.
time restic --password-file=restic_password.txt -r sftp:desktop_sftp@southbridge:/desktop_sftp/.restic restore latest --target=/home/florian/Daten/ --include="/home/florian/Daten"

A SFTP Server using the OPENSSH included SSH Server with a ZFS Pool as backend.

Expected behavior
Restore the Data with the same permissions that they had before.

Actual behavior
Restoring the Backup but with root as owner and group.

Backup Data as root user from another user and Restore it as a root user.

Since the Permissions are correct in the Backup when i mount the Backup i guess there is something wrong with the restore.

No

It made my backup way easier.

hoelzro:
This allows leaveDir to set metadata properly for directories
with descendants that are selected by an include filter, even
if the directory in question isn't selected by that filter.

For example, let's say you have the following file in your repo:

    /home/user/nested/data/values.txt

...and / and /home are owned by root, but /home/user and below are owned
by user.

If you run "restic restore -i /home/user/nested/data -t /tmp/restore $SNAPSHOT" as root,
/home/user/nested/ will currently be restored with root as the owner,
even though it's supposed to be owned by user.

This change makes sure that leaveDir is called to restore metadata to
nodes like /home/user/nested - enterDir is also called under the same
circumstances for purposes of symmetry.

This fixes GH #1402 and (I think) GH #2563

This PR changes directory traversal logic when restoring a snapshot, and it fixes a bug around restoration of metadata.

This PR closes #1402 and closes #2563 

Checklist
---------

### DIFF
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -104,7 +104,7 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 				return errors.Errorf("Dir without subtree in tree %v", treeID.Str())
 			}
 
-			if selectedForRestore {
+			if selectedForRestore || childMayBeSelected {
 				err = sanitizeError(visitor.enterDir(node, nodeTarget, nodeLocation))
 				if err != nil {
 					return err
@@ -118,7 +118,7 @@ func (res *Restorer) traverseTree(ctx context.Context, target, location string, 
 				}
 			}
 
-			if selectedForRestore {
+			if selectedForRestore || childMayBeSelected {
 				err = sanitizeError(visitor.leaveDir(node, nodeTarget, nodeLocation))
 				if err != nil {
 					return err
@@ -209,11 +209,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 
 	// first tree pass: create directories and collect all files to restore
 	err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
-		enterDir: func(node *restic.Node, target, location string) error {
-			// create dir with default permissions
-			// #leaveDir restores dir metadata after visiting all children
-			return fs.MkdirAll(target, 0700)
-		},
+		enterDir: noop,
 
 		visitNode: func(node *restic.Node, target, location string) error {
 			// create parent dir with default permissions


hoelzro:
I haven't added tests for this yet, since the issue can only manifest when restic is run as root.  I'm open to to ideas on how I could write a test for this, though!

One behavior I'm uncertain of is `leaveDir` restoring metadata - if a directory already exists in the restoration target directory but no new files are actually created under it, that directory will still have its metadata restored.  I'm not sure if that's the behavior users would expect or not, so I'm open to feedback on that!

greatroar:
Does the issue also manifest when the owner is the same, but the mode is different? In that case, backing up a directory with mode 710 (or some other unlikely value) would trigger the bug with needing root for the test.

MichaelEischer:
This PR spews out a large number of errors, when using a relative include pattern: `restic restore latest --target test --include command_win.go`:
```
ignoring error for /restic/.git/hooks: UtimesNano: no such file or directory
ignoring error for /restic/.git/info: UtimesNano: no such file or directory
ignoring error for /restic/.git/logs/refs/heads: UtimesNano: no such file or directory
```

How does this PR relate to #2906 which AFAIK fixes the exact same problem?

internal/restorer/restorer.go
```diff
@@ -209,11 +209,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 
 	// first tree pass: create directories and collect all files to restore
 	err = res.traverseTree(ctx, dst, string(filepath.Separator), *res.sn.Tree, treeVisitor{
-		enterDir: func(node *restic.Node, target, location string) error {
-			// create dir with default permissions
-			// #leaveDir restores dir metadata after visiting all children
-			return fs.MkdirAll(target, 0700)
-		},
+		enterDir: noop,
```
MichaelEischer, line 212: This change breaks restoring empty directories.

hoelzro:
@MichaelEischer Ah, I neglected to check if the target directory was created by restic before trying to restore the metadata - I fixed this, but that PR you linked to looks to be more comprehensive

PR #3006, opened 2020-10-10
Reimplement rebuild-index and remove /internal/index

### LINKED ISSUE #2547: Discussion: Future of prune and rebuild-index

aawsome:

After #2842 we can as well reimplement rebuild-index to use the same functionality.
This allows to get rid of `internal/index` which is now no longer used.

Moreover, now existing index entries are used to reduce the number of packs that need to be read.
This improves speed a lot.
The newly added option `--read-all-packs` implements the previous behavior.

This PR depends on #2842; only the last 6 commits are relevant for this PR.

closes #2547 

Checklist
---------

- [x] I have not added tests for all changes in this PR, existing tests still work.


MichaelEischer:
> Also in an future improvement, we could add more functionality to this new rebuild-index: It could load the existing index files and just check which pack files are not (fully) covered. Only those would then need to be read; for all other packs, the existing index can be used.

We should also keep the option to rebuild the index from scratch.

aawsome:
> > Also in an future improvement, we could add more functionality to this new rebuild-index: It could load the existing index files and just check which pack files are not (fully) covered. Only those would then need to be read; for all other packs, the existing index can be used.
> 
> We should also keep the option to rebuild the index from scratch.

I just implemented the improved algorithm to take existing index entries into account. I named the new option `--read-all-packs` to read all pack headers (and ignore the existing index).

doc/060_forget.rst
```diff
@@ -282,3 +303,45 @@ last-day-of-the-months (11 or 12 depends if the 5 weeklies cross a month).
...
+*****************
+
+To understand the custom options, we first explain how the pruning process works:
+
+- First all snapshots and directories within snapshots are scanned to determine
+  which data is still in use.
+- Then for all pack files ``prune`` finds out if the file is fully used, partly
+  used or completely unused.
+- Completely unused packs are marked for deletion. Fully used packs are kept.
+  A partiallly used pack is either kept or marked for repacking depending on user
```
greatroar, line 316: Typo: partially

MichaelEischer:
What about calling the option `--full-rebuild`? That would avoid encoding implementation details into option names.

I still have to think about whether a full or a partial rebuild should be the default. Right now I slightly prefer the former. That way we wouldn't change the current behavior and after all it's the safer option when a repository was damaged.

Apropos, under which conditions would it be safe to use the faster variant? That is probably the case when a pack file (or multiple ones) is for some reason missing from the index. When a pack file does not match the size it should have according to the index, then I'd be very careful, after all that pack file is probably damaged. If you already know which pack file is damaged, then the fast path is definitely useful as it allows quickly removing the damaged file from the index.

aawsome:
> Apropos, under which conditions would it be safe to use the faster variant? That is probably the case when a pack file (or multiple ones) is for some reason missing from the index. 

I would say the faster variant will give the identical result as the "full rebuild" iff all pack files that are omitted are completely and correctly contained in the index.
So, the questions is under what condition is a pack file omitted: This is the case iff the pack is referenced in the index and the size calculated from the index does match the actual file size.

So, for the faster variant to be unsafe, we basically need to have a pack file with name and size of a correctly indext pack file, but with different contents; i.e. a modified packfile due to manipulation or bit rot.
In that case the faster variant will just keep the index entries while the "full rebuild" variant will read the header. If the modification is in the header, the "full rebuild" will handle the error. If the modification is, however, not in the header, both variants will again give the same result.
I don't think that modification detection should be regarded, that is the job of `restic check --read-data`..

But actually I remember, that I wanted to work on the error handling when reading a pack file. The `rebuild-index` of master just ignores errors and builds an index without the pack file contents, while this PR aborts. I wanted to add this and almost forgot it - I'll work on this.

> When a pack file does not match the size it should have according to the index, then I'd be very careful, after all that pack file is probably damaged. If you already know which pack file is damaged, then the fast path is definitely useful as it allows quickly removing the damaged file from the index.

When a pack file does not match the size it should have, it will be treated identically for both variants: It will be read and if this succeeds, the contents are added to the newly generated index.

MichaelEischer:
> I don't think that modification detection should be regarded, that is the job of restic check --read-data..

There's one special case here: For backends which charge for traffic it's far less expensive to let `rebuild-index` sort of check that all pack files are still accessible than running `check --read-data`. And if `rebuild-index` notices data corruption that it should also report it, but of course actively searching for modifications is the job of `check`.

> (EDIT: one could argue that the faster variant is even more safe in this case; if only the header is corrupted it should be better to just use the correct index entries that are present 😉 )

That depends on what you intend to do: Such a damaged pack file should not at all be used in a backup, but it may still be valuable when restoring files. When rebuild-index notices such a damaged pack header, it could keep the old data for that pack file, finish the rebuild otherwise and then at the end return an error. The exact steps to recover from a error would be slightly different than currently, but that should still be manageable.

> But actually I remember, that I wanted to work on the error handling when reading a pack file. The rebuild-index of master just ignores errors and builds an index without the pack file contents, while this PR aborts. I wanted to add this and almost forgot it - I'll work on this soon.

Ideally recovery from repository damages works automatically in large parts (at least in the future). It's not really ideal to require users to manually delete this or that file from the repository. How would aborting the index rebuild on a damaged pack file fit in with that? It's also quite common to have one or another incomplete pack file stored (at least for some backends).

MichaelEischer:
So the faster variant wouldn't detect inaccessible pack files or corrupted pack file headers. And it wouldn't recover from bit flips in the index. (`check --read-data` currently also does not check that the pack blob list matches the data stored in the index). That leaves e.g. accidental pack deletions, packs deleted during repository recovery and packs which for some mysterious reason are missing from the index (most likely an interrupted backup run. Could be useful when planing to use `recover`) as use cases for the fast variant.

The full rebuild would recover bit flips and remove damaged/inaccessible packs from the index. That latter step has the benefit that it simplifies recovery: after the damaged pack files are removed from the index, then its safe again to run backups using the repository. It would also allow prune to work again.

aawsome:
> There's one special case here: For backends which charge for traffic it's far less expensive to let `rebuild-index` sort of check that all pack files are still accessible than running `check --read-data`. And if `rebuild-index` notices data corruption that it should also report it, but of course actively searching for modifications is the job of `check`.

Right. Of course, as the faster variant doesn't read as much, there are error cases which can't be detected.

> > (EDIT: one could argue that the faster variant is even more safe in this case; if only the header is corrupted it should be better to just use the correct index entries that are present wink )
> That depends on what you intend to do: Such a damaged pack file should not at all be used in a backup, but it may still be valuable when restoring files. When rebuild-index notices such a damaged pack header, it could keep the old data for that pack file, finish the rebuild otherwise and then at the end return an error. The exact steps to recover from a error would be slightly different than currently, but that should still be manageable.

I agree. The faster variant will just not be able to detect a damaged pack header, the "full rebuild" will be, of course.

> So the faster variant wouldn't detect inaccessible pack files or corrupted pack file headers. And it wouldn't recover from bit flips in the index. 

Well, it wouldn't detect pack files that are listed in the backend but are not accessible. 

About bit flips in the index: `Repository.LoadIndex` currently aborts if there is a bitflip in the index. Hence the fast variant would simply abort as well as all other command that use `LoadIndex`. And yes, the only way for users to recover from this would be to use a "full rebuild".  If we would change `Repository.LoadIndex` to just ignore corrupt index files, the faster variant would also work. To increase resiliance, I would recommend to give a warning, ignore corrupt index files and continue with all operations (and finally exit with an error code at the end). Then both variants can be used to remove broken index files.

Or did you mean bit flips in the pack header here? 

> (check --read-data currently also does not check that the pack blob list matches the data stored in the index).

You are right - actually we should add that check to `check --read-data` ...

>  That leaves e.g. accidental pack deletions, packs deleted during repository recovery and packs which for some mysterious reason are missing from the index (most likely an interrupted backup run. Could be useful when planing to use recover) as use cases for the fast variant.

I would say the use cases of the fast variant are basically all cases where the user wants to rebuild the index as it is not in ideal state. That might be due to an aborted `prune` (or other aborted commands) or due to (accidentally or intentionally) deleted/added pack or index files directly in the repository backend.

> The full rebuild would recover bit flips and remove damaged/inaccessible packs from the index.

Note that the "full rebuild" is not able to remove all kind of damaged packs from the index! As it only reads the header, it is not able to checksum the whole pack file. It is not even able to remove a pack from the index if `check --read-data` reported an corrupt pack file, but that corruption is a modification in the non-header part. If we are talking about bit rots, those will be most likely not in the pack header as this usually is a very small part of the pack file.
You are right about inaccessible pack files.

>  That latter step has the benefit that it simplifies recovery: after the damaged pack files are removed from the index, then its safe again to run backups using the repository. 

To be really safe, I would say you must run `check --read-data` before or after `rebuild-index`. But if `check --read-data` doesn't give an error, both variants should be equal.

> It would also allow prune to work again.
Can you give an example where the "full rebuild" will allow prune to work after whereas the fast variant won't?

MichaelEischer:
> About bit flips in the index: Repository.LoadIndex currently aborts if there is a bitflip in the index.

I had bitflips in mind which happened before the index was hashed and written to the backend. For these indexes there would only be some flipped bits in the index entries such that the blob size (easy to detect), offset (not that simple) or hash (the blob you just appear as missing) would differ. But these are probably very, very rare (even though we've had a few reports about bitflips in packfiles).

Just printing a warning when an index can't be loaded should be fine. It won't break backups and prune checks beforehand that it finds all necessary blobs.

> > It would also allow prune to work again.
> 
> Can you give an example where the "full rebuild" will allow prune to work after whereas the fast variant won't?

It depends on the exact behavior of the fast variant when a pack file suddenly is shorter than expected and has now a damaged pack header. If the fast repack keeps the old blob list for that pack, then prune will also detect the discrepancy and abort. If it doesn't, then prune will work. Your earlier comments sounded like this PR uses the former variant, I haven't looked at the implementation itself yet.

doc/060_forget.rst
aawsome: Is corrected in #2718

aawsome:
I rebased this PR and changed it according to #3022 
Note that the tests fail until #3025 has been merged.

aawsome:
> It depends on the exact behavior of the fast variant when a pack file suddenly is shorter than expected and has now a damaged pack header. If the fast repack keeps the old blob list for that pack, then prune will also detect the discrepancy and abort. If it doesn't, then prune will work. Your earlier comments sounded like this PR uses the former variant, I haven't looked at the implementation itself yet.

The faster variant will remove all contents of the current index for packs that are either read or exist in the index but not in the repository and add all contents of the pack headers which are read.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+			}
+			delete(packInfo, id)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+		// process pack files that are referenced in the index but do not exist
+		for id := range packInfo {
+			packsWronglyIndexed.Insert(id)
```
MichaelEischer, line 125: `missingPacks` would be a more accurate description.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+		Verbosef("finding pack files in repo\n")
+		err = repo.List(ctx, restic.PackFile, func(id restic.ID, size int64) error {
+			packs.Insert(id)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+	} else {
+		Verbosef("load indexes\n")
```
MichaelEischer, line 84: s/load/loading/

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+		}
+		// process pack files that are referenced in the index but do not exist
+		for id := range packInfo {
+			packsWronglyIndexed.Insert(id)
 		}
 	}
 
-	Verbosef("finding old index files\n")
+	if len(packs) > 0 {
+		Verbosef("reading pack files...\n")
```
MichaelEischer, line 130: Please remove the '...'

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+		//		for _, id := range invalidFiles {
+		//			Printf("skipped incomplete pack file: %v\n", id)
+		//		}
+		//	}
 
-	ids, err := idx.Save(ctx, repo, supersedes)
-	if err != nil {
-		return errors.Fatalf("unable to save index, last error was: %v", err)
+	} else {
+		Verbosef("no need to read any pack file.\n")
```
MichaelEischer, line 147: Please remove the "." . The other output lines also don't have one.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
 
-	var supersedes restic.IDs
-	err = repo.List(ctx, restic.IndexFile, func(id restic.ID, size int64) error {
-		supersedes = append(supersedes, id)
-		return nil
-	})
-	if err != nil {
-		return err
-	}
+		//	if globalOptions.verbosity >= 2 {
```
MichaelEischer, line 140: That output was/is pretty useful, when you wonder why a pack file is missing from the index.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,58 +52,109 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+	packs.Merge(packsWronglyIndexed)
 
-	Verbosef("remove %d old index files\n", len(supersedes))
-	err = DeleteFilesChecked(globalOptions, repo, restic.NewIDSet(supersedes...), restic.IndexFile)
+	err = rebuildIndexFiles(gopts, repo, packs, uint64(totalBlobs))
 	if err != nil {
-		return errors.Fatalf("unable to remove an old index: %v\n", err)
+		return err
 	}
+	Verbosef("done.\n")
```
MichaelEischer, line 157: Please remove the trailing dot

internal/repository/repository.go
```diff
@@ -515,6 +515,92 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
 	return nil
 }
 
+// LoadIndexFromPacks loads a new index by reading all pack files
+// This will create one Index data structure per pack file readed that is not
```
MichaelEischer, line 519: s/readed/read/

internal/repository/repository.go
```diff
@@ -515,6 +515,92 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
...
+			}
+		}
+
+		return nil
+	}
+
+	// run workers on ch
+	wg.Go(func() error {
+		defer close(indexCh)
+		return RunWorkers(loadIndexParallelism, worker)
```
MichaelEischer, line 577: Let's stick with the higher `const listPackWorkers = 10` for now. Listing pack file contents is completely IO bound.

internal/repository/repository.go
```diff
@@ -515,6 +515,92 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
...
+	})
+
+	// receive decoded indexes
+	wg.Go(func() error {
+		for idx := range indexCh {
+			r.idx.Insert(idx)
+			p.Report(restic.Stat{Blobs: 1})
+		}
+		// Add new finalized index to set given index IDs
+		for _, id := range ids {
```
MichaelEischer, line 587: I don't like this hack to inject the index files list into the index. Wouldn't it be cleaner to pass the list of obsolete index files directly to `rebuildIndexFiles` / `MasterIndex.Save`? Or add a new constructor to `Index` which allows directly creating a finalized index with a list of superseded packs?

MichaelEischer (changes requested):
The code looks fine mostly, there are just a few very subtle things happening, see my comments.

internal/repository/index.go
```diff
@@ -295,7 +296,7 @@ func (idx *Index) EachByPack(ctx context.Context, packBlacklist restic.IDSet) <-
 			m := &idx.byType[typ]
 			m.foreach(func(e *indexEntry) bool {
 				packID := idx.packs[e.packIndex]
-				if !packBlacklist.Has(packID) {
+				if idx.final == false || !packBlacklist.Has(packID) {
```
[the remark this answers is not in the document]
aawsome: changed.

internal/repository/index.go
```diff
@@ -276,7 +276,8 @@ func (idx *Index) Each(ctx context.Context) <-chan restic.PackedBlob {
 }
 
 // EachByPack returns a channel that yields all blobs known to the index
-// grouped by packID but ignoring blobs with a packID in packPlacklist.
+// grouped by packID but ignoring blobs with a packID in packPlacklist for
+// finalized indexes.
```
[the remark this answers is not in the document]
aawsome: I added an explanation.

cmd/restic/cmd_rebuild_index.go
aawsome: You are right, that is the better name - is changed.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

cmd/restic/cmd_rebuild_index.go
aawsome: Actually this part was missing as `LoadIndexFromPacks` just aborted. This is now changed and this code is activated.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

internal/repository/repository.go
aawsome: changed.

internal/repository/repository.go
aawsome: I used `listPackParellelism` to stay consistent with `loadIndexParallelism`.

aawsome:
I rebased this. As #2842 now counts pack files when building the new index, I now determine the number of packs. This simplified the code a bit.

internal/repository/repository.go
aawsome: You are right, this is a bad hack :wink: 
I added `extraObsolete` to `MasterIndex.Save` to solve this.

cmd/restic/cmd_rebuild_index.go
aawsome: done.

aawsome:
After the discussion about calling `Backend.List()` twice, I realized that this might be also an issue here. So I changed this PR such that now at most one `List()` per filetype is needed. Also added this to the `ListOnce` tests.

aawsome:
rebased after #2718 has been merged.

aawsome:
rebased after #2842 has been merged

aawsome:
For this PR, only documentation is open. I just noticed that there yet exists no documentation for the `rebuild-index` command and it might anyway better fit in an troubleshooting documentation (see e.g. #2683).
So from my side, this PR is ready to merge.

aawsome:
Rebased this after #3058 has been merged.

Also made index saving parallel (see last added commit).

rawtaz:
@MichaelEischer Were your concerns about which variant should be the default resolved?

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+		Verbosef("getting pack files to read...\n")
+		// iterate over all blobs in index
+		for blob := range repo.Index().Each(ctx) {
+			size, ok := packSizeFromIndex[blob.PackID]
+			if !ok {
+				size = pack.HeaderSize
+			}
+			size += int64(pack.PackedSizeOfBlob(blob.Length))
+			// update packSizeFromIndex
+			packSizeFromIndex[blob.PackID] = size
```
MichaelEischer, line 104: I'd prefer to merge this with the previous line.
`packSizeFromIndex[blob.PackID] = size + int64(pack.PackedSizeOfBlob(blob.Length))`

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+			return err
+		}
 
-	bar := newProgressMax(!globalOptions.Quiet, packs-uint64(len(ignorePacks)), "packs")
-	idx, invalidFiles, err := index.New(ctx, repo, ignorePacks, bar)
-	bar.Done()
-	if err != nil {
-		return err
-	}
+		packSizeFromIndex := make(map[restic.ID]int64)
```
MichaelEischer, line 93: Please extract the `packSizeFromIndex` calculation into a separate method. #3082 also adds the exact same loop.

internal/repository/master_index.go
```diff
@@ -262,7 +262,8 @@ func (mi *MasterIndex) MergeFinalIndexes() {
 }
 
 // Save saves all known indexes to index files, leaving out any
-// packs whose ID is contained in packBlacklist. The new index contains the IDs
+// packs whose ID is contained in packBlacklist from finalized indexes.
+// The new index contains the IDs
```
MichaelEischer, line 269: please reflow the next two comment lines.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
 }
 
-func rebuildIndex(ctx context.Context, repo restic.Repository, ignorePacks restic.IDSet) error {
-	Verbosef("counting files in repo\n")
+func rebuildIndex(opts RebuildIndexOptions, gopts GlobalOptions, repo *repository.Repository, ignorePacks restic.IDSet) error {
+	ctx := gopts.ctx
+
+	var obsolete restic.IDs
+	packSizeFromList := make(map[restic.ID]int64)
+	packs := restic.NewIDSet()
```
MichaelEischer, line 63: Please call this variable `removePacks`. At first I thought that this is the list of packs to keep in the index, and then the code did unexpected things...

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
-	return rebuildIndex(ctx, repo, restic.NewIDSet())
+	return rebuildIndex(opts, gopts, repo, restic.NewIDSet())
 }
 
-func rebuildIndex(ctx context.Context, repo restic.Repository, ignorePacks restic.IDSet) error {
-	Verbosef("counting files in repo\n")
+func rebuildIndex(opts RebuildIndexOptions, gopts GlobalOptions, repo *repository.Repository, ignorePacks restic.IDSet) error {
+	ctx := gopts.ctx
+
+	var obsolete restic.IDs
```
MichaelEischer, line 61: please rename this to `obsoleteIndexes`.  Otherwise it's just too confusing to have two different kinds of IDs next to each other.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
 
-	ids, err := idx.Save(ctx, repo, supersedes)
-	if err != nil {
-		return errors.Fatalf("unable to save index, last error was: %v", err)
+		for _, id := range invalidFiles {
+			Verboseff("skipped incomplete pack file: %v\n", id)
+			totalPacks--
+		}
+	} else {
+		Verbosef("no need to read any pack file\n")
```
MichaelEischer, line 141: I think it's better to not print anything when no packs have to be indexed.

internal/repository/master_index.go
```diff
@@ -261,6 +262,8 @@ func (mi *MasterIndex) MergeFinalIndexes() {
 	mi.idx = newIdx
 }
 
+const SaveIndexParallelism = 4
```
MichaelEischer, line 265: That constant should start with a lower-case letter

internal/repository/master_index.go
```diff
@@ -276,50 +279,76 @@ func (mi *MasterIndex) Save(ctx context.Context, repo restic.Repository, packBla
...
-		for pbs := range idx.EachByPack(ctx, packBlacklist) {
-			newIndex.StorePack(pbs.packID, pbs.blobs)
-			p.Add(1)
-			if IndexFull(newIndex) {
-				newIndex.Finalize()
-				if _, err := SaveIndex(ctx, repo, newIndex); err != nil {
-					return nil, err
-				}
-				newIndex = NewIndex()
+		ch <- newIndex
```
MichaelEischer, line 331: Please add a `ctx.Done()` escape route.

cmd/restic/cmd_rebuild_index.go
```diff
@@ -44,59 +52,100 @@ func runRebuildIndex(gopts GlobalOptions) error {
...
+			}
+			totalPacks++
+			delete(packSizeFromIndex, id)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+		for id := range packSizeFromIndex {
+			// ignore pack files that are referenced in the index but do not exist
```
MichaelEischer, line 122: s/ignore/forget/

Ignore is a bit ambiguous here.

internal/repository/repository.go
```diff
@@ -515,6 +517,85 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
 	return nil
 }
 
+const listPackParallelism = 10
+
+// LoadIndexFromPacks loads a new index by reading all given pack files
+// This will create one Index data structure per pack file read that is not
+// marked as finalized.
+func (r *Repository) LoadIndexFromPacks(ctx context.Context, packs map[restic.ID]int64, p *progress.Counter) (invalid restic.IDs, err error) {
```
MichaelEischer, line 525: What about `ReindexPacks`?

internal/repository/repository.go
```diff
@@ -515,6 +517,85 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
...
+		for fi := range ch {
+			entries, _, err := r.ListPack(ctx, fi.ID, fi.Size)
+			if err != nil {
+				debug.Log("unable to list pack file %v", fi.ID.Str())
+				m.Lock()
+				invalid = append(invalid, fi.ID)
+				m.Unlock()
+			}
+			idx := NewIndex()
+			idx.StorePack(fi.ID, entries)
```
MichaelEischer, line 565: Is there a specific reason to use individual indexes and not just one? Or one per worker thread? I'd expect a single index to use quite a bit less memory.

cmd/restic/cmd_rebuild_index.go
aawsome: Fully agree. In fact I would add this to `repository.MasterIndex`.

There is one thing we should consider about this pack size computation used here, in #3082 and in the new `prune` implementations: In the discussions about a repo format V2 there is the proposal to save the pack header in JSON format and I would strongly advice to do so (e.g. to support compression :wink: ). However, this would make the computation of the pack header size much more difficult and even impossible if old and new pack format formats will be supported simultaneously. So IMHO saving the pack header in JSON would also imply that we overwork the format of the index files and save some additional pack information there - like the pack header size or whether the pack header is in JSON or not. Therefore this information anyway is much better suited to be saved within the in-memory index (and if we enlarge the pack files, there won't be much of an memory issue).

Do you mind if I keep this comparably small piece of code as it is in this PR and prepare a follow-up PR which extracts this calculation and prepares for a maybe-coming additional pack file format?

MichaelEischer: To handle compressed blobs the index must be changed. We'd need both the compressed and uncompressed blob sizes in the index. And along with that we can simply add enough information to allow check and the other commands to deduce the pack header size. I'm not too much a fan of JSON in particular as parsing it is slow as hell. (And without additional header compression that JSON would be two or three times larger than the current binary format, for the same information!)

Deduplicating that pack size loop in a separate PR is fine for me.

internal/repository/repository.go
aawsome: I agree that `Load` is not a good part of the name. But I also disliked `ReindexPacks` as this might indicate that the packs must be already part of the index. I changed the name to `CreateIndexFromPacks`.

internal/repository/repository.go
aawsome: Thanks for giving the hint. In fact, the only reason was that I copied the code from `LoadIndex`. Is now changed such that only one index is created.

internal/repository/master_index.go
aawsome: Thanks for finding this! Is added.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

internal/repository/master_index.go
aawsome: changed.

cmd/restic/cmd_rebuild_index.go
aawsome: the `else` branch is now removed.

cmd/restic/cmd_rebuild_index.go
aawsome: is renamed.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

cmd/restic/cmd_rebuild_index.go
aawsome: changed.

internal/repository/repository.go
```diff
@@ -515,6 +516,88 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
...
+	wg.Go(func() error {
+		defer close(indexCh)
+		return RunWorkers(listPackParallelism, worker)
+	})
+
+	// receive decoded indexes
+	wg.Go(func() error {
+		for idx := range indexCh {
+			r.idx.Insert(idx)
+			p.Report(restic.Stat{Blobs: 1})
```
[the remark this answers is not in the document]
aawsome: Oh, that was a rebase mistake. Should be changed now.

fd0 (approved):
Great work, thanks!

MichaelEischer (approved):
LGTM. Nice to once again only have a single index implementation :-) .

cmd/restic/cmd_rebuild_index.go
aawsome: >  We'd need both the compressed and uncompressed blob sizes in the index.

@MichaelEischer Just as a side note: I think we need to save the (uncompressed, unencrypted) blob size and the size the blob occupies within the pack (i.e. maybe compressed and including crypto overhead) in the index.

Issue #1724, opened 2018-04-17
Restic Mount - Be More Vocal Regarding Permission Issues


stevesbrain:
## Output of `restic version`
restic 0.8.3
compiled with go1.10.1 on linux/amd64

## How did you run restic exactly?
```bash
sudo mkdir /mnt/restic-b2
--------
:: [stephen@waddleduck] ~
$ ls -ls /mnt
total 12
4 drwxr-xr-x  2 root root 4096 Apr 17 22:13 restic-b2
----------
export B2_ACCOUNT_KEY=XXXXX
export B2_ACCOUNT_ID=YYYYYYY
export RESTIC_PASSWORD=ZZZZZZZz
cd ~/
restic -r b2:bucket-name:/ mount /mnt/restic-b2
------------------
password is correct
*hit CTRL+C at this point after waiting about 7 minutes*
  signal interrupt received, cleaning up
context canceled, ignoring
fusermount: exit status 1
also, the following messages were logged by a library:
2018/04/17 22:18:21 mount helper error: fusermount: user has no write access to mountpoint /mnt/restic-b2
unable to umount (maybe already umounted?): exit status 1: fusermount: entry for /mnt/restic-b2 not found in /etc/mtab
-----------------------
:: [stephen@waddleduck] ~
$ sudo chown stephen:stephen /mnt/restic-b2
-------------
[sudo] password for stephen:
:: [stephen@waddleduck] ~
$ restic -r b2:bucket-name:/ mount /mnt/restic-b2
------------
password is correct
Now serving the repository at /mnt/restic-b2
Don't forget to umount after quitting!
```

## What backend/server/service did you use to store the repository?
Backblaze B2

## Expected behavior
Restic should warn you on the command line that the repository couldn't be mounted due to permissions issues.

## Actual behavior
Restic does not inform you of this until you "cancel" after waiting for a confirmation of mounting (which is not received). Then, the user is informed as a "mount helper error". 

## Steps to reproduce the behavior
Make a new directory which your user account does not have write access to, and try mounting to it. Then, after waiting some time (to confirm it won't mount), Ctrl+c, and you will see what happens.

## Do you have any idea what may have caused this?
Presumably just that nobody thought it would be necessary to generate such a message at this point :)

## Do you have an idea how to solve the issue?
Catch the error and display relevant info to the user.

## Did restic help you or made you happy in any way?
Absolutely; every day :D Restic is life!

fd0:
[...]

PR #3008, opened 2020-10-11
Fix #1681 mount: change behavior of `restic mount` when path doesn't exist

### LINKED ISSUE #1681: Restic fails to create the mount point directory if the parent directory does not exist
## Output of `restic version`
`restic 0.8.3
compiled with go1.10 on darwin/amd64`

`sudo restic -r /Volumes/ex mount /mnt/bak `

Since this is a Mac `/mnt` does not exist yet.

Local

## Expected behavior
If the mount point is not accessible to restic, one of:
- Restic would not try to create any directory and just fail if the mount point does not exist
- Restic would try and create the mount point and all parent directories

## Actual behavior
Restic wants to be nice and tries to create the mount point for you, if it notices that it's missing.
However restic does not check if the parent directory of the mount point exists. This leads to a situation where restic tries to create the mount point, but the operation fails due to the parent directory not existing.

```
# restic -r /tmp/test init
# restic -r /tmp/test mount $HOME/path/to/your/very/deep/mount/directory/thats/without/a/parent

password is correct
Mountpoint /Users/nobody/path/to/your/very/deep/mount/directory/thats/without/a/parent doesn't exist, creating it
mkdir /Users/nobody/path/to/your/very/deep/mount/directory/thats/without/a/parent: no such file or directory
unable to umount (maybe already umounted?): unmount /Users/nobody/path/to/your/very/deep/mount/directory/thats/without/a/parent: no such file or directory
```

See Line: https://github.com/restic/restic/blob/be0a5b7f065a0b80f82c3ba1aa5abd9253b150ee/cmd/restic/cmd_mount.go#L103:18

In this case I don't think restic has to be extra nice to the end user. The program should just fail if the mount point does not exist.

However if restic wants to be nice to the end user, we would need to add `MkDirAll` to the FS functions and use that to create the mount point.

kitone:

The `restic mount` command will not try to create any directory and just fail if the mount point does not exist with explicit messages. The checks of the validity of the mount point (exist and permission) will be done sooner.

#1681
~~#1724~~

Checklist
---------

- ~~I have added tests for all changes in this PR~~
- ~~I have added documentation for the changes (in the manual)~~

### DIFF
--- a/changelog/unreleased/issue-1681
+++ b/changelog/unreleased/issue-1681
@@ -1,0 +1,9 @@
+Bugfix: Make mount not create missing mount point
+
+When specifying a non-existent directory as mount point for the mount
+command, restic used to create the directory automatically. This has now
+changed such that restic instead gives an error when the directory for the
+mount point doesn't exist.
+
+https://github.com/restic/restic/issues/1681
+https://github.com/restic/restic/pull/3008
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -104,11 +104,8 @@ func mount(opts MountOptions, gopts GlobalOptions, mountpoint string) error {
 	}
 
 	if _, err := resticfs.Stat(mountpoint); os.IsNotExist(errors.Cause(err)) {
-		Verbosef("Mountpoint %s doesn't exist, creating it\n", mountpoint)
-		err = resticfs.Mkdir(mountpoint, os.ModeDir|0700)
-		if err != nil {
-			return err
-		}
+		Verbosef("Mountpoint %s doesn't exist\n", mountpoint)
+		return err
 	}
 
 	mountOptions := []systemFuse.MountOption{
--- a/cmd/restic/integration_fuse_test.go
+++ b/cmd/restic/integration_fuse_test.go
@@ -163,9 +163,6 @@ func TestMount(t *testing.T) {
 	repo, err := OpenRepository(env.gopts)
 	rtest.OK(t, err)
 
-	// We remove the mountpoint now to check that cmdMount creates it
-	rtest.RemoveAll(t, env.mountpoint)
-
 	checkSnapshots(t, env.gopts, repo, env.mountpoint, env.repo, []restic.ID{}, 0)
 
 	rtest.SetupTarTestFixture(t, env.testdata, filepath.Join("testdata", "backup-data.tar.gz"))


rawtaz:
This changes the behavior of the `mount` command in that it no longer creates the mount point folder if it doesn't exist. If we decide to do this, which would break BC but is [suggested by @fd0 in #1681](https://github.com/restic/restic/issues/1681#issuecomment-375785383), we must be very clear about it in the changelog entry (which is needed regardless).

changelog/unreleased/issue-1681
```diff
@@ -0,0 +1,11 @@
+Bugfix: mount: change behavior of `restic mount` when path doesn't exist
+
+We've change the behavior of `restic mount` command when the directory doesn't
+exist, prevously restic was trying to create the mount point path in that case,
+this is not the current behavior anymore, we now return an explicit error.
+In addition to this, we check that we have permission to write in the directory.
+
+
```
rawtaz, line 8: Thanks for adding the changelog! Please update it to the following:

~~~
Bugfix: Make mount check permissions and not create missing mount point

When specifying a non-existent directory as mount point for the mount
command, restic used to create the directory automatically. This has now
changed such that restic instead gives an error when the directory for the
mount point doesn't exist.

Restic now also checks the permissions of the specified mount point to
make sure that they are sufficient for mounting the repository in it.
~~~

Also please remove the excessive empty line before the URLs at the end :)

kitone: Thanks, done!

cmd/restic/cmd_mount.go
```diff
@@ -160,6 +159,32 @@ func umount(mountpoint string) error {
...
+	if err != nil {
+		// check with IsPermission(err) ?
+		return errors.Fatalf("the mount path is not writable directory: %q", mountpoint)
+	}
+	// fixme: check for error ?
+	f.Close()
+	os.Remove(tmpFile)
+
+	return nil
+}
```
greatroar, line 186: I don't see why any of this is necessary. When the mountpoint doesn't exist, bazil.org/fuse reports that:

    mountpoint does not exist: /tmp/foo

When the permission are wrong, it reports that:

    2020/10/11 16:21:43 mount helper error: fusermount: user has no write access to mountpoint /tmp/foo

cmd/restic/cmd_mount.go
```diff
@@ -104,11 +106,8 @@ func mount(opts MountOptions, gopts GlobalOptions, mountpoint string) error {
 	}
 
 	if _, err := resticfs.Stat(mountpoint); os.IsNotExist(errors.Cause(err)) {
-		Verbosef("Mountpoint %s doesn't exist, creating it\n", mountpoint)
-		err = resticfs.Mkdir(mountpoint, os.ModeDir|0700)
-		if err != nil {
-			return err
-		}
+		Verbosef("Mountpoint %s doesn't exist\n", mountpoint)
+		return err
```
greatroar, line 110: This check is redundant.

kitone: Thank, forget to remove it

cmd/restic/cmd_mount.go
kitone: Checking about this earlier make it fail faster and remove unnecessary work and terminal output like opening repo, ...
For the end user, this is a more friendly error message that reading.

But I guess we can remove this commit and simply close the #1724, on my last test the message print without blocking.

MichaelEischer: Let's assume for now that #1724 was caused by something external to restic. So please remove this commit for now.

kitone: Hi, the commit was removed and the changelog updated.

MichaelEischer (approved):
LGTM. Thanks!

PR #3009, opened 2020-10-11
Rework print functions to better match verbosity levels 

plumbeo:

This PR reworks the current print functions to better match the verbosity levels. It adds a `Defaultf` function (I'm open to using a different name, I don't really like it that much) that prints output when `verbosity` is greater than or equal to 1 (default) and a `Verboseff` function, that prints output when `verbosity` is greater than or equal to 3 (`--verbose=2` or `-vv`), and modifies `Verbosef` to print output when `verbosity` is greater than or equal to 2 (`--verbose`).

It then converts all current `Verbosef` calls to the appropriate `Defaultf`/`Verbosef`/`Verboseff` calls, in some cases removing some duplicated checks of the `verbosity` level, and changes some non-error `Printf` to `Defaultf` to suppress the output when `--quiet` is used.

It came up in https://github.com/restic/restic/pull/2718#discussion_r502906278 where a new function Verboseff was first added.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_forget.go
+++ b/cmd/restic/cmd_forget.go
@@ -130,13 +130,13 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
 
 		if policy.Empty() && len(args) == 0 {
 			if !gopts.JSON {
-				Verbosef("no policy was specified, no snapshots will be removed\n")
+				PrintDef("no policy was specified, no snapshots will be removed\n")
 			}
 		}
 
 		if !policy.Empty() {
 			if !gopts.JSON {
-				Verbosef("Applying Policy: %v\n", policy)
+				PrintDef("Applying Policy: %v\n", policy)
 			}
 
 			for k, snapshotGroup := range snapshotGroups {
--- a/cmd/restic/cmd_forget.go
+++ b/cmd/restic/cmd_forget.go
@@ -129,18 +129,14 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
 		}
 
 		if policy.Empty() && len(args) == 0 {
-			if !gopts.JSON {
-				PrintDef("no policy was specified, no snapshots will be removed\n")
-			}
+			PrintDef("no policy was specified, no snapshots will be removed\n")
 		}
 
 		if !policy.Empty() {
-			if !gopts.JSON {
-				PrintDef("Applying Policy: %v\n", policy)
-			}
+			PrintDef("Applying Policy: %v\n", policy)
 
 			for k, snapshotGroup := range snapshotGroups {
-				if gopts.Verbose >= 1 && !gopts.JSON {
+				if !gopts.Quiet && !gopts.JSON {
 					err = PrintSnapshotGroupHeader(gopts.stdout, k)
 					if err != nil {
 						return err
@@ -191,9 +187,7 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
 				return err
 			}
 		} else {
-			if !gopts.JSON {
-				Printf("Would have removed the following snapshots:\n%v\n\n", removeSnIDs)
-			}
+			Printf("Would have removed the following snapshots:\n%v\n\n", removeSnIDs)
 		}
 	}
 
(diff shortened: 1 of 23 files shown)


aawsome:
I like the idea, but I'm not totally satisfied with the name `Defaultf`. However, I can't give a better name ATM.

I would like to add another aspect: For commands which support `--json` it is usually not wanted to print out some verbose stuff and we get these extra checks for `opts.JSON`, see e.g. `delete.go` . Maybe we can incorporate this check in the functions or we set verbosity automatically to 0 if `--json` is set?

rawtaz:
For someone who isn't accustomed to these functions, names like `Defaultf()` make absolutely no sense - they convey nothing about what they actually do. They're terrible function names :)

plumbeo:
I renamed `Defaultf` to `PrintDef` which is marginally more obvious. I also added a commit that forces `Quiet` when JSON output is on. Looking to the code there shouldn’t be any unintended consequences but I’ve never used the JSON output.

dimejo:
Personally I really dislike testing different `-v` options just to find the right amount of logging. What about using different log levels like `error`, `warning`, `info` and `debug`?

cmd/restic/cmd_forget.go
```diff
@@ -129,18 +129,14 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
...
 
 		if !policy.Empty() {
-			if !gopts.JSON {
-				Verbosef("Applying Policy: %v\n", policy)
-			}
+			PrintDef("Applying Policy: %v\n", policy)
 
 			for k, snapshotGroup := range snapshotGroups {
-				if gopts.Verbose >= 1 && !gopts.JSON {
+				if !gopts.Quiet && !gopts.JSON {
```
MichaelEischer, line 139: `gopts.Verbose` != `gopts.verbosity`. `PrintSnapshotGroupHeader` should only be printed if `-v` is passed at least once. However, the check should be `gopts.verbosity >= 2` instead of the current `gopts.Verbose >= 1`.

cmd/restic/cmd_mount.go
```diff
@@ -143,8 +143,8 @@ func mount(opts MountOptions, gopts GlobalOptions, mountpoint string) error {
 	}
 	root := fuse.NewRoot(repo, cfg)
 
-	Printf("Now serving the repository at %s\n", mountpoint)
-	Printf("When finished, quit with Ctrl-c or umount the mountpoint.\n")
+	PrintDef("Now serving the repository at %s\n", mountpoint)
```
MichaelEischer, line 146: Why is this changed to `PrintDef`?

cmd/restic/cmd_recover.go
```diff
@@ -142,7 +142,7 @@ func runRecover(gopts GlobalOptions) error {
 		return errors.Fatalf("unable to save snapshot: %v", err)
 	}
 
-	Printf("saved new snapshot %v\n", id.Str())
+	PrintDef("saved new snapshot %v\n", id.Str())
```
MichaelEischer, line 145: Why is this changed to `PrintDef`?

cmd/restic/cmd_self_update.go
```diff
@@ -71,15 +71,15 @@ func runSelfUpdate(opts SelfUpdateOptions, gopts GlobalOptions, args []string) e
 		}
 	}
 
-	Printf("writing restic to %v\n", opts.Output)
+	PrintDef("writing restic to %v\n", opts.Output)
```
MichaelEischer, line 74: Why is this changed to `PrintDef`?

MichaelEischer (changes requested):
Although `--json` sort of implies `--quiet` for most commands, this is not true for the `backup` command. The alternative would be to check for `globalOptions.JSON` inside the `Verbosef`, etc. functions. However, that leads to a rather bad user experience for commands which don't support JSON output, as the command will do something without providing any feedback to the user. So we either should let all command which don't support JSON fail with an error or wait with that change until _all_ commands support JSON.

I generally agree that the current `Verbosef` function name is rather confusing. `PrintDef` sounds sort of ok. However, I'm wondering whether there's a nice way to name `Verbosef` and `Verboseff` to ensure that the compiler will will throw errors when PRs add new output using the old function names?

MichaelEischer:
@plumbeo Are you still interested in working on this PR?

plumbeo:
> @plumbeo Are you still interested in working on this PR?

Hi, no, I will not have the time to work on this any time soon so I’m closing the PR, one less thing to keep track of. 

Thanks for the comments, I’m sorry I wasn’t able to spend more time on it.

PR #3014, opened 2020-10-11
Fix sporadic stream reset between rclone and restic

ivandeex:

When restic is run with rclone backend, it sporadically prints ERROR messages similar to this:
```
rclone: 2020/10/05 11:40:23 ERROR : data/xx/xxxx: Didn't finish writing GET request (wrote 0/xxx): http2: stream closed
```

This happens because restic's `Load()` does not wait for EOF on the body stream from rclone after
reading full body of a partial object. Since restic and rclone internally talk HTTP/2 on their stdio pipe, this sometimes makes Golang's HTTP/2 client send back a `stream reset` frame to rclone. This was investigated here https://github.com/ivandeex/rclone/tree/stream-reset/stream-reset#the-gory-details

This patch makes restic drain rclone response body till EOF before calling `Close`.
The drain is done on a separate goroutine so that main restic process is not delayed.
The distribution of delay time till EOF was measured in https://github.com/ivandeex/rclone/tree/stream-reset/stream-reset#distribution-of-delays-till-http2-eof

The drain approach was once discussed in https://github.com/restic/restic/issues/1561#issuecomment-373118890 but was not implemented.

This bug is 2 year old, it was first reported in https://github.com/rclone/rclone/issues/2598
and can be reproduced as decribed here https://github.com/ivandeex/rclone/tree/stream-reset/stream-reset#running-the-test

The issue was also investigated here: https://github.com/ivandeex/rclone/tree/stream-reset/stream-reset#premature-stream-reset-between-restic-and-rclone

A similar issue (*RST_STREAM + CANCEL sent by HTTP2 client to server after reading the DATA frame of a response*) was reported against Golang HTTP/2 library in https://github.com/golang/go/issues/41570
and can be reproduced by running a simple test https://github.com/ivandeex/rclone/blob/stream-reset/stream-reset/http2test.go which does not require restic or rclone:
```
go run http2test.go http1 nodrain 10
```
runs successfully
```
go run http2test.go http2 nodrain 10
```
prints error messages
```
go run http2test.go http2 drain 10
```
is successful again.

Checklist
---------

### DIFF
--- a/internal/backend/rest/rest.go
+++ b/internal/backend/rest/rest.go
@@ -173,7 +173,32 @@ func (b *Backend) IsNotExist(err error) bool {
 // Load runs fn with a reader that yields the contents of the file at h at the
 // given offset.
 func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error {
-	return backend.DefaultLoad(ctx, h, length, offset, b.openReader, fn)
+	r, err := b.openReader(ctx, h, length, offset)
+	if err != nil {
+		return err
+	}
+	err = fn(r)
+	if err != nil {
+		_ = r.Close() // ignore error here
+		return err
+	}
+
+	// Note: readerat.ReadAt() (the fn) uses io.ReadFull() that doesn't
+	// wait for EOF after reading body. Due to HTTP/2 stream multiplexing
+	// and goroutine timings the EOF frame arrives from server (eg. rclone)
+	// with a delay after reading body. Immediate close might trigger
+	// HTTP/2 stream reset resulting in the *stream closed* error on server,
+	// so we wait for EOF before closing body.
+	var buf [1]byte
+	_, err = r.Read(buf[:])
+	if err == io.EOF {
+		err = nil
+	}
+
+	if e := r.Close(); err == nil {
+		err = e
+	}
+	return err
 }
 
 func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error) {


ncw:
That is great debugging @ivandeex :-) I've been staring at that code for a long time without figuring out the bug.

Looking at your fix, I think it is relevant to the rest backend which the rclone backend is based on - maybe the fix should go there?

BTW This is incredibly similar to fixes that we had to do to ncw/swift with [drainAndClose](https://github.com/ncw/swift/blob/31c380e4df88654745e636a5e90f7601e6c64f2f/swift.go#L353)

ivandeex:
I personally don't have a setup with restic's rest server with http/2 activated but I use restic over rclone on a daily basis. If reviewer from restic team suggests extending the patch I'll do that readily. BTW I am long aware of this issue. Before I had time to investigate it I just... [hid it](https://github.com/ivansible/lin-restic/commit/8596287144e03f53a30f8f3c3a045066b0fa711b#diff-439fe0ae598155b99f8551e92dc7a9cbac860e9897cb566db5f7b684b2fe2ee8R108) ;-) :D

MichaelEischer:
@ivandeex I'd prefer to fix this for both the REST and rclone backends. If I understood your write-up correctly, the resp.Body for a HTTP/1.1 connection uses a LimitReader which will also generate a EOF once all bytes were read. So the drainAndClose code should also work for non HTTP/2 connections.

Regarding the question whether the stream close should happen synchronously or asynchronously, did you take a look at the timings for the whole `Load` method with synchronous / asynchronous read until eof + close?

ncw:
> I personally don't have a setup with restic's rest server with http/2 activated but I use restic over rclone on daily basis. If reviewer from restic team suggests extending the patch I'll do that readily.

I think it should just be a question of moving your patch into here and removing the change from the rclone backend - the rclone backend will call this anyway.

https://github.com/restic/restic/blob/6003dada14a188fe158a5f5b33ca9b5b59436d6b/internal/backend/rest/rest.go#L173-L177

> BTW I am long aware of this issue. Before I had time to investigate it I just... [hid it](https://github.com/ivansible/lin-restic/commit/8596287144e03f53a30f8f3c3a045066b0fa711b#diff-a2c515e8a33ecbb7e28ff6c6f7d841072ddaf045de649e76d584f40354e7c3b2L15) ;-) :D

:-)

ivandeex:
> Regarding the question whether the stream close should happen synchronously or asynchronously, did you take a look at the timings for the whole `Load` method with synchronous / asynchronous read until eof + close?

My tests show (https://github.com/ivandeex/rclone/tree/stream-reset/stream-reset#comparing-synchronous-vs-async-drain-methods) that using _goroutine_ for drain and close gives on average _2% improvement_ of load time. Goroutines are _very_ cheap, but margin's too narrow for me to decide. Waiting for conclusion from @MichaelEischer

fd0 (changes requested):
Thank you very much for all the time you spent on debugging and fixing it!

In order to merge it, we'll need to do the following:
 * As @MichaelEischer already mentioned, the fix should go into the `rest` backend so we don't have to maintain a separate instance of `Load()` just for the rclone backend. The fix works for both HTTP1 and HTTP2 anyway, so we can just use it for all requests (that way, we'll also discover any issues with it much faster).
 * Waiting for EOF should not start a separate Goroutine, I can imagine many things going wrong when there's no limit on the number of Goroutines started. If we do it synchronously, the usual limitations for backend connections work. Before starting a new Goroutine, we need to think about [how it is stopped](https://dave.cheney.net/2020/02/23/the-zen-of-go#_goroutines).

Please let me know if you need any help with this :)

ivandeex:
Done. Please review.

internal/backend/rest/rest.go
```diff
@@ -173,7 +174,29 @@ func (b *Backend) IsNotExist(err error) bool {
 // Load runs fn with a reader that yields the contents of the file at h at the
 // given offset.
 func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error {
-	return backend.DefaultLoad(ctx, h, length, offset, b.openReader, fn)
+	r, err := b.openReader(ctx, h, length, offset)
+	if err != nil {
+		return err
+	}
+	err = fn(r)
```
fd0, line 181: If `fn` returns an error, something else went wrong (and maybe the body wasn't even read completely), so we can skip waiting for EOF and just close the reader. Waiting for EOF only makes sense in the happy case when no error occurred:

```suggestion
	err = fn(r)
	if err != nil {
		_ = r.Close() // ignore error here
		return err
	}
```

This will also decouple the `err` variable from whatever happens below, which makes the code much easier to read.

internal/backend/rest/rest.go
```diff
@@ -173,7 +174,29 @@ func (b *Backend) IsNotExist(err error) bool {
...
+
+	// Note: readerat.ReadAt() (the fn) uses io.ReadFull() that doesn't
+	// wait for EOF after reading body. Due to HTTP/2 stream multiplexing
+	// and goroutine timings the EOF frame arrives from server (eg. rclone)
+	// with a delay after reading body. Immediate close might trigger
+	// HTTP/2 stream reset resulting in the *stream closed* error on server,
+	// so we wait for EOF before closing body.
+	var buf [1]byte
+	if n, e := r.Read(buf[:]); n != 0 || e != io.EOF {
+		fmt.Fprintf(os.Stderr, "%v: error waiting for EOF from server\n", h)
```
fd0, line 191: Just logging it to `stderr` does not handle the error and will mess with other programs calling restic.

We need to figure out what to do with the error returned by `Read()`:
 * Either return it from `Load()` and pass it up the stack
 * Or ignore it

IMHO, we can drop the check for `n` and just call `Read()` and inspect the error: If it's `nil` or `io.EOF` then ignore it, otherwise return it and pass it to the caller. Thoughts?

fd0 (changes requested):
Wow, that was fast. I've reviewed the code, please have a look!

internal/backend/rest/rest.go
ivandeex: done

internal/backend/rest/rest.go
ivandeex: done

ivandeex:
review addressed

fd0 (approved):
Awesome, thanks for the fast response!

PR #3015, opened 2020-10-12
Improved changelog template

### LINKED ISSUE #3001: Describe target audience in changelog template
What should be changed?
--------------------------

The changelog template currently does not describe the target audience of a changelog entry. Some of the details for writing a changelog entry are only described in point seven of [CONTRIBUTING.md](https://github.com/restic/restic/blob/88cc4447791abc4bf030f3099f74ad22812ec631/CONTRIBUTING.md#providing-patches):
> If your pull request changes anything that users should be aware of (a bugfix, a new feature, ...) please add an entry as a new file in changelog/unreleased including the issue number in the filename (e.g. issue-8756). Use the template in changelog/TEMPLATE for the content. It will be used in the announcement of the next stable release. While writing, ask yourself: If I were the user, what would I need to be aware of with this change?

I think it would be useful if the changelog template describes all relevant aspects for adding/writing a changelog entry (especially as the PR template directly links to the changelog template!). It currently IMHO misses the following aspects:
- How to decide what the changelog should describe?
- Who is the target audience?
- Which filename should the changelog entry get?

greatroar:

This is an edited version of the changelog template is suggested at #3001, taking @rawtaz's comments into account.

Fixes #3001, I hope.

Checklist
---------

### DIFF
--- a/changelog/TEMPLATE
+++ b/changelog/TEMPLATE
@@ -1,11 +1,21 @@
-Bugfix: Fix behavior for foobar (in present tense)
+# The first line must start with Bugfix:, Enhancement: or Change:,
+# including the colon. Use present use. Remove lines starting with '#'
+# from this template.
+Enhancement: Allow custom bar in the foo command
 
-We've fixed the behavior for foobar, a long-standing annoyance for restic
-users.
+# Describe the problem in the past tense, the new behavior in the present
+# tense. Mention the affected commands, backends, operating systems, etc.
+# Focus on user-facing behavior, not the implementation.
 
-The text in the paragraphs is written in past tense. The last section is a list
-of issue URLs, PR URLs and other URLs. The first issue ID (or the first PR ID,
-in case there aren't any issue links) is used as the primary ID.
+Restic foo always used the system-wide bar when deciding how to frob an
+item in the baz backend. It now permits selecting the bar with --bar or
+the environment variable RESTIC_BAR. The system-wide bar is still the
+default.
+
+# The last section is a list of issue, PR and forum URLs.
+# The first issue ID determines the filename for the changelog entry:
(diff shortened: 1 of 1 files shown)


changelog/TEMPLATE
```diff
@@ -1,11 +1,21 @@
...
+# The first line must start with Bugfix:, Enhancement: or Change:,
+# including the colon. Use present use. Remove lines starting with '#'
+# from this template.
+Enhancement: Allow custom bar in the foo command
 
-We've fixed the behavior for foobar, a long-standing annoyance for restic
-users.
+# Describe the problem in the past tense, the new behavior in the present
+# tense. Mention the affected commands, backends, operating systems, etc.
+# Focus on user-facing behavior, not the implementation.
```
rawtaz, line 8: I think we should probably add this sentence at the end of this paragraph: `If the text becomes long, split the past/old and the present/new parts into two paragraphs.`

rawtaz:
About time this is improved!

changelog/TEMPLATE
greatroar: That enforces a rule that I already know I'm going to try and break. When the first paragraph is too long, I would rather summarize both the problem and the solution in the first paragraph, then add more paragraphs with details. Maybe leave this to the contributor? Your call, obviously.

rawtaz: I don't agree with how you'd do it, but if it doesn't work out in the end we'll just comment on it at that point, so I don't think this is important enough to argue about.

rawtaz (approved):
LGTM!

rawtaz:
@MichaelEischer Is this new template fine with you, do you feel it resolves the concerns in your original issue?

MichaelEischer (approved):
LGTM.

PR #3016, opened 2020-10-12
Simplify/optimize cached UID/GID lookups

greatroar:

LookupUsername and lookupGroup in internal/restic/node.go returned an error value that was always nil. When a lookup fails, the user/group name is simply empty and doesn't show up in the JSON.

The functions also cached UIDs and GIDs as strings, meaning that every call would still need a strconv.Itoa call. This PR removes the error return, adds comments to explain the behavior, merges some variables with the equal meanings and pushes the strconv.Itoa calls down into the cache miss path.

Benchmark results:

```name                  old time/op    new time/op    delta
NodeFillUser-8          1.92µs ± 5%    1.75µs ± 3%   -8.89%  (p=0.000 n=10+10)
NodeFromFileInfo-8      1.89µs ± 7%    1.76µs ± 4%   -6.69%  (p=0.001 n=10+10)

name                  old alloc/op   new alloc/op   delta
NodeFillUser-8            504B ± 0%      496B ± 0%   -1.59%  (p=0.000 n=10+10)
NodeFromFileInfo-8        504B ± 0%      496B ± 0%   -1.59%  (p=0.000 n=10+10)

name                  old allocs/op  new allocs/op  delta
NodeFillUser-8            5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.000 n=10+10)
NodeFromFileInfo-8        5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.000 n=10+10)
```

No.

Checklist
---------

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -506,24 +506,12 @@ func (node Node) sameExtendedAttributes(other Node) bool {
 	return true
 }
 
-func (node *Node) fillUser(stat statT) error {
-	node.UID = stat.uid()
-	node.GID = stat.gid()
+func (node *Node) fillUser(stat statT) {
+	uid, gid := stat.uid(), stat.gid()
+	node.UID, node.GID = uid, gid
 
-	username, err := lookupUsername(strconv.Itoa(int(stat.uid())))
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. Thanks for the cleanup

Issue #3013, opened 2020-10-11
Add '--files-from-raw' flag that treats every line literally, processes in exactly that order, and with no comments or globbing


jim-collier:
What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

`restic --files-from <file>` does not work like most other similar flags in other programs - notably, "industry-standard" rsync, or rclone.

(That "problem" is already documented in other issues, including #3005 which I opened. But that issue is contentious, not going to happen, is a losing battle, and can/should be closed.)

`restic --files-from <file>` does not treat `<file>` as a raw dump of file specifications as typically generated by other utilities (e.g. chaining in more powerful `find`, `sed`, `awk`, and/or `grep` into a scripted workflow).

Instead `restic --files-from <file>`, treats each line as potentially hand-crafted globbing specifications. In other words, it's literally the functionality of what many other programs provide a `--patterns-from` or `--includes-from` option for. (Including rsync, rclone, and borg.)

This has already caused numerous issues to be opened or tengentially mentioned (enumerated below), e.g. because brackets in filenames causing problems. (The solution being, ironically, to chain together a scripted workflow which first escapes all globbing characters from the filenames in `<file>`.)

However, as others have pointed out, changing this logic now [might horribly break existing user scripts](https://xkcd.com/1172/) that depend on the globbing functionality of `--files-from`.

**So it seems that the only solution is to build a `--files-from-raw <file>` option, that operates _exactly_ as `[rsync](https://linux.die.net/man/1/rsync) --files-from` does.**

Rclone provides both `--files-from` and `--files-from-raw` flags, so there is some industry precedence. 

(In rclone's case though, `--files-from` behaves much more like rsync's `--files-from`, than restic's. It just allows comments and non-filename spaces, which their `--files-from-raw` option does not.)

Rclone's `--files-from-raw` behaves exactly as rsync's `--files-from` does, **as `restic --files-from-raw` should also behave**:

- Every line is _an exact 1:1 file specification_. No globbing.
- Leading and trailing spaces, tabs, etc. are considered part of the filename.
- Files are processed (backed up) exactly in the order they appear in `<file>`
    - This is a very important part of this feature, as the file order may be very particular and exact, e.g. already sorted by the user on date, size, etc.
- Comments or empty lines aren't "allowed".
    - More specifically, it just might not exist as an exact, literal file path - in which case it would be ignored, with a warning emitted.
[...]

Issue #2276, opened 2019-05-19
--files-from: "pattern ... does not match any files, skipping" / "syntax error in pattern"


kiuuik:
Output of `restic version`
--------------------------

I tried with a binary you provide 

     restic 0.9.5 compiled with go1.12.4 on linux/amd64

and a version I built today ("go get" and then "go build.go")

     restic 0.9.5 (v0.9.5-30-g5bd5db42) compiled with go1.12.5 on linux/amd64

summary
--------------------------

`--files-from` doesn't work for me as expected. If I have a file `0 [1].docx` and its absolute paths `/some/path/0 [1].docx` in a line in a textfile and when I run `restic backup --files-from textfile` I get this error: `pattern "/some/path/0 [1].docx" does not match any files, skipping`.

After reading the example with `find ... | grep ...` in the manual under [including-and-excluding-files](https://restic.readthedocs.io/en/stable/040_backup.html#including-and-excluding-files) my impression is that I can use one file per line with any name?

How did you run restic exactly?
-------------------------------

    $ flist="/tmp/flist.txt"
    $ find "$HOME/Desktop/testsource/" -type f > "${flist}"
    $ restic --repo "$HOME/repo" -p "$HOME/password.txt" backup --files-from "${flist}"

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

local folder

Expected behavior
-----------------

backup those files when they are in a text file that I use as an argument for `--files-from`. The textfile has these lines:

    /home/user/Desktop/testsource/1 [1] text
    /home/user/Desktop/testsource/0 [1].docx
    /home/user/Desktop/testsource/01 [02] text.docx
    /home/user/Desktop/testsource/0 [1]
    /home/user/Desktop/testsource/2 [3]

Actual behavior
---------------

    $ restic --repo "$HOME/repo" -p "$HOME/password.txt" backup --files-from "${flist}"
    pattern "/home/user/Desktop/testsource/1 [1] text" does not match any files, skipping
    pattern "/home/user/Desktop/testsource/0 [1].docx" does not match any files, skipping
    pattern "/home/user/Desktop/testsource/01 [02] text.docx" does not match any files, skipping
    pattern "/home/user/Desktop/testsource/0 [1]" does not match any files, skipping
    pattern "/home/user/Desktop/testsource/2 [3]" does not match any files, skipping
    Fatal: nothing to backup, please specify target files/dirs

If I have a file with a name like `2 [3` and I run the commands from above again I get

[...]

PR #1748, opened 2018-05-01
Respect Encoding and Byte Order Mark when reading text files

fd0:
This PR adds code to support files with a Byte Order Mark. Windows programs seem to commonly leave one at the start of a file, indicating the endianness and the Encoding. The code now supports UTF-8 and UTF-16 (in big and little endian).

Closes #1738
Closes #1433

Issue #1738, opened 2018-04-29
windows: passwords with a paragraph character have a problem


cd1188:
## Output of `restic version`
restic 0.8.3
compiled with go1.10 on windows/amd64

## How did you run restic exactly?
on windows 10 in german and windows cmd
password in the file pwd.txt is test§§test and saved as ANSI
[pwd.txt](https://github.com/restic/restic/files/1958736/pwd.txt)

C:\restic>restic -r c:\backup_test init -p c:\restic\pwd.txt
restic -r c:\backup_test snapshots

## What backend/server/service did you use to store the repository?
test on local drives

## Expected behavior

C:\restic>restic -r c:\backup_test snapshots
password is correct

## Actual behavior

C:\restic>restic -r c:\backup_test snapshots
enter password for repository:
Fatal: wrong password or no key found

## Steps to reproduce the behavior

only passwords with a paragraph character in it have problems on the following situation
-> password is here always test§§test

C:\restic>restic -r c:\backup_test init -p c:\restic\pwd.txt
restic -r c:\backup_test snapshots
ERROR

C:\restic>restic -r c:\backup_test init
restic -r c:\backup_test snapshots -p c:\restic\pwd.txt
ERROR

C:\restic>restic -r c:\backup_test init
restic -r c:\backup_test snapshots
WORK

C:\restic>restic -r c:\backup_test init -p c:\restic\pwd.txt
restic -r c:\backup_test snapshots -p c:\restic\pwd.txt
WORK

## Do you have any idea what may have caused this?

ascii-ansi / windows codepage problem?
german-problem?

## Do you have an idea how to solve the issue?

no

fd0:
Thanks for the report, I'll summarize what I understood so far:

 * You initialize a repository with the password `test§§test`, the password is read from a file encoded as `latin1`/`ISO-8859-1` (I've verified that on the file you've uploaded)
 * When you try to access the repository, typing the password manually, it does not work
 * When you try to access the repository while reading the password from the file again, it works

Is that correct?

If so, the reason is that the string `test§§test` has different bytes, depending on the encoding:

`latin1`/`ISO-8859-1`:

```
00000000: 7465 7374 a7a7 7465 7374              test..test
```

`UTF-8`:

```
00000000: 7465 7374 c2a7 c2a7 7465 7374         test....test
```

[...]

PR #3017, opened 2020-10-12
Add backup options --files-from-verbatim and --files-from-raw

### LINKED ISSUE #2944: Support backing up user-provided list of files
Output of `restic version`
--------------------------

```
restic 0.9.6 (v0.9.6-204-gf0b2dee7) compiled with go1.12.9 on linux/amd64
```

Support a well-specified and simple way for users to tell restic to backup a specific set of files.

For example, consider this JSON list of files and assume that there are files with these exact names on the filesystem

```
[
  "/src/**",
  "/src/\n",
  "/src/\\"
]
```

restic could accept them as a JSON array, like

```
echo '["/src/**","/src/\n","/src/\\"]' | restic backup --files-from-json -
```

Alternatively, restic could accept them with NUL-delimiters, like

```shell
printf '%s\0' '/src/**' '/src/\n' "/src/\\" | restic backup --files-from-z -
```

which aligns with similar features in several other tools for producing output and receiving input in a consistent way: `find` (`-print0`), `xargs` (`-0`), `git ls-files` (`-z`).

Personally, I prefer the JSON-based approach since we can provide more meaningful errors, and there are fewer edge cases or decisions to make around the parsing of it (like how to treat an empty field or whether to consider a trailing newline as extraneous or as a path), but it may be a little less conventional.

To be forward-compatible, it may be worth putting the list of files under a key, like `{"files": []}`. I really do not want restic to do any processing at all here, but later someone may want to be able to provide globs in a similar way, without the caveats of `--files-from`. `{"patterns": []}` could be simpler than another argument in that case.

I have a specific list of files to back up, and want to use restic to do it. I want to back up only these specific files, and the paths may contain any character the OS/filesystem considers valid.

Currently, to be understood by restic in a compatible way, I have to carefully consider each path. The logic looks like:

1. Leave `$` as-is, since `$` does not need to be escaped in the paths provided via `--files-from` (as opposed to exclude files)
2. Make every path absolute
   - avoids needing to escape leading whitespace, which would be trimmed if in a `--files-from` file
   - avoids needing to escape a leading `#`, which would cause the line to be skipped if in a `--files-from` file
3. Pass paths that contain newlines via positional command-line arguments
   - files passed to `--files-from` are split by newline, so newline-having paths would get split incorrectly
   - currently, no other processing is done on paths provided this way
   - only passing newline-having paths as arguments, as opposed to all paths, reduces the chance of hitting an OS-specific limit (for example, `ARG_MAX` on Linux)
4. For paths without newlines, escape as described below and save them one-per-line to a file and pass the file path to `--files-from`
   1. Surround each `[`, `*`, and `?` with `[]`
      - necessary since all paths derived from `--files-from` are passed through `filepath.Glob`
      - putting special characters in a character class causes `filepath.Glob` to treat them as normal characters, without relying on `\`-escaping, which is not supported in `filepath.Glob` on Windows
   2. If there is any trailing whitespace, surround the last character with `[]`
      - currently, restic strips any leading and trailing whitespace from each line derived from a `--files-from` argument
      - preserves the trailing whitespace by eliminating it in a way that will be interpreted correctly by `filepath.Glob`
   3. On non-Windows OSes, replace `\` with `\\`
      - `\\` will be interpreted correctly as `\` by `filepath.Glob`

Which has these downsides:

1. Not obvious at all
2. Requires additional code to integrate with restic in a generic way
3. Minor performance drawback since we may be globbing for what are essentially plain paths to individual files, and taking the slower path in the Go std library
4. Very fragile. Any of the above could easily change if untested, and I wouldn't hold it against anyone that broke it either

If restic supports this feature, then none of the above is necessary and it gets a lot easier to use restic as a building block for different backup workflows.

Yes! I like the content-hash-based de-duplication a lot, especially where the modified time of a file may change, but not its contents. Other approaches like rclone with a crypt backend would unfortunately reupload the file.

greatroar:

restic backup now has a flag --files-from-raw0 that reads a list of NUL-terminated, uninterpreted filenames from a file. This makes it easier to compile lists of files to back up; previously, users had to write scripts to work around the globbing behavior of --files-from.

Closes #2944, #3013. Updates #2276.

Checklist
---------


rawtaz:
@greatroar I've started reviewing the documentation and have some changes I'd like to make.

But I wish to first ask, is there a limitation on Windows such that the files cannot be UTF-16? The docs mention that on Windows they have to be UTF-8.

Also, what would happen if you gave these options a file that isn't UTF-encoded? I presume you added the note about the encoding for a reason, otherwise it might not be worth mentioning since pretty much everyone and their cat will produce plain ASCII or UTF files.

greatroar:
UTF-16 is the native encoding of filenames on Windows, and other parts of restic (--files-from, RESTIC_PASSWORD_FILE) do accept it. --files-from-raw0 is the exception.

If you enter a filename that is not UTF-8, then what happens depends on the encoding. UTF-16 contains zero bytes in the encodings for all ASCII characters, so you'll get bogus filenames that may or may not refer to existing files. Common 8-bit encodings should cause failures in the syscall or x/sys/windows packages, which try to encode filenames in UTF-16 before passing them to system calls.

rawtaz:
> UTF-16 is the native encoding of filenames on Windows, and other parts of restic (--files-from, RESTIC_PASSWORD_FILE) do accept it. --files-from-raw{,0} is the exception.

This is why I was confused, because the docs in the PR state "On Windows, the arguments to `--files-from-raw` and `--files-from-raw0` must contain filenames encoded in UTF-8." - that is, it's saying that both of the -raw variants must get UTF-8 encoded filenames. But given what you wrote above, it seems that only -raw0 must get UTF-8, and -raw is fine with either UTF-8 and UTF-16 just like the rest of restic.

Shall I (with the rest of my document changes) change this to a note that informs the user that the `--files-from-raw0` *must* get a file that's UTF-16 encoded, but that doesn't mention any limitation about in regards to encoding for `--files-from` and `--files-from-raw`?

greatroar:
Actually, --files-from-raw also doesn't accept UTF-16, but I forgot to mention that initially.

rawtaz:
How come it doesn't? If it isn't looking for NUL bytes, then what you wrote about that and UTf-16 shouldn't apply. That leaves the question why --files-from-raw doesn't support UTF-16 when --files-from does? :)

greatroar:
Because that violates the requirement that --files-from-raw should not interpret filenames. As soon you allow UTF-16, you have to strip a byte-order mark off the first filename and decode every single name. That also means limiting the set of allowed filenames further, because now they have to be valid Unicode.

I know this is inconsistent, but adding UTF-16 support was [a bad idea to begin with](https://utf8everywhere.org/). Let's not repeat the mistake. (In fact, I would argue for dropping UTF-16 support in --files-from and password files, but I'm not sure how important backward compatibility is at this point.)

rawtaz:
I agree that UTF-8 is the way to go in general. You mention UTF-16 support, is there explicit support for that somewhere in restic other than restic just using libraries that parses some text regardless of whether it's UTF-8 or UTF-16 or some other encoding (in such a way that restic isn't explicitly asking it to parse UTF-16)? In other words, is there some part of restic where we *actively* support UTF-16?

greatroar:
Yes, in internal/textfile. That's used for RESTIC_PASSWORD, RESTIC_REPOSITORY_FILE, exclude files and --files-from input.

MichaelEischer (changes requested):

rawtaz:
Thanks for the pointer to internal/textfile/read.go that specifically handles UTF16. It seems that this code was introduced in #1748 to fix problems like #1433 and #1738. While I don't like UTF16 and never use it myself, the two issues and the PR seems like a valid use case for the code. It's an unfortunate fact that people will have UTF16 encoded stuff every now and then, as well as BOMs in UTF8, and it will probably overall cause more problems to not handle that, than it's worth not handling it.

> Because that violates the requirement that --files-from-raw should not interpret filenames. As soon you allow UTF-16, you have to strip a byte-order mark off the first filename and decode every single name. That also means limiting the set of allowed filenames further, because now they have to be valid Unicode.

There's a difference between dealing with encodings and parsing data or filenames though. Just because you properly deal with different encodings that the data is written in, doesn't mean that you are interpreting the data differently. The interpretation/reading of filenames is something that happens *after* properly establishing and adhering to the encoding used in the file - once you have done that, the filenames should be read equally regardless of whether the file is encoded in UTF8 or UTF16. In this case, if `--files-from-raw` were to deal with both UTF8 and UTF16, it wouldn't change how it reads the filenames, because the filenames should be interpreted the same regardless of the "underlying" encoding.

> I know this is inconsistent, but adding UTF-16 support was a bad idea to begin with. Let's not repeat the mistake. (In fact, I would argue for dropping UTF-16 support in --files-from and password files, but I'm not sure how important backward compatibility is at this point.)

Unfortunately I think the use case for dropping the code introduced in #1748 would cause more harm than it's worth given that the code was implemented to solve actual cases of encountering UTF16 and BOM in UTF8 as well.

---

If we sum up the intentions of the three `--files-from*` options they are:

- `--files-from is` about allowing for patterns that are globbed
- `--files-from-raw` is about allowing for filenames that are not globbed
- `--files-from-raw0` is about allowing for filenames that are not globbed and that *also* may contain newlines in their filenames

I'd say that neither of these has anything to do with supporting or not supporting specific encodings.

Is there a technical hurdle in supporting both ISO-8859-1, UTF8 and UTF16 in all three of these options? I don't see how it would affect the interpretation of the actual patterns or filenames in the files, as we don't expect anyone to have BOMs as part of their filenames. But I'm open to misunderstanding something.

cmd/restic/cmd_backup_test.go
```diff
@@ -0,0 +1,86 @@
...
+	rtest.OK(t, err)
+	rtest.Equals(t, 0, len(got))
+
+	// Empty lines are ok, as is a missing newline at the end.
+	got, err = readFilenamesRaw(strings.NewReader("foo\n\nbar"))
+	rtest.OK(t, err)
+	rtest.Equals(t, got, []string{"foo", "bar"})
+
+	_, err = readFilenamesRaw(strings.NewReader("zero-terminated\x00"))
+	rtest.Assert(t, strings.Contains(err.Error(), "zero byte"),
```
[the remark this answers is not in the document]
greatroar: That was intentional, but kind of lazy. Added an assert.

cmd/restic/cmd_backup_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// An empty filename is an error.
+	_, err = readFilenamesRaw0(bufio.NewReader(strings.NewReader("foo\x00\x00")))
+	rtest.Assert(t, strings.Contains(err.Error(), "empty filename"),
+		"wrong error message: %v", err.Error())
+
+	// No trailing NUL byte is an error, because it likely means we're
+	// reading a line-oriented text file (someone forgot -print0).
+	_, err = readFilenamesRaw0(bufio.NewReader(strings.NewReader("simple.txt")))
+	rtest.Assert(t, strings.Contains(err.Error(), "zero byte"),
```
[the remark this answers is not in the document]
greatroar: ... and one more of these above. Fixed.

cmd/restic/cmd_backup.go
```diff
@@ -198,12 +197,99 @@ func readLinesFromFile(filename string) ([]string, error) {
...
 func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 	if gopts.password == "" {
+		err := errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
 		for _, filename := range opts.FilesFrom {
 			if filename == "-" {
-				return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
+				return err
+			}
+		}
+		for _, filename := range opts.FilesFromRaw0 {
```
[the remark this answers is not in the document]
greatroar: I've appended all three lists together instead.

cmd/restic/cmd_backup.go
```diff
@@ -198,12 +197,99 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+// or stdin if filename is "-". Each filename is terminated by a zero byte,
+// which is stripped off.
+func readFilenamesFromFileRaw0(filename string) ([]string, error) {
+	f, err := openOrStdin(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return readFilenamesRaw0(bufio.NewReader(f))
```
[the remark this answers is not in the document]
greatroar: Done and removed the bufio.Readers from the test.

cmd/restic/cmd_backup.go
```diff
@@ -368,14 +460,30 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
...
+	for _, file := range opts.FilesFromRaw {
+		fromfile, err := readFilenamesFromFileRaw(file)
+		if err != nil {
+			return nil, err
+		}
+		names = append(names, fromfile...)
+	}
+
+	for _, file := range opts.FilesFromRaw0 {
+		fromfile, err := readFilenamesFromFileRaw(file)
```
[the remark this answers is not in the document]
greatroar: Done.

greatroar:
> There's a difference between dealing with encodings and parsing data or filenames though. Just because you properly deal with different encodings that the data is written in, doesn't mean that you are interpreting the data differently. The interpretation/reading of filenames is something that happens after properly establishing and adhering to the encoding used in the file - once you have done that, the filenames should be read equally regardless of whether the file is encoded in UTF8 or UTF16.

You're assuming filenames must be Unicode. That's not true on Linux or *BSD (and presumably Solaris), where filenames are arbitrary binary strings except that they can't contain a slash or zero byte and must be at least one byte long. There is no "underlying encoding".

rawtaz:
This PR was discussed earlier today, and we reasoned that:

- The point of --files-from-raw in the first place was to ditch the globbing in --files-from.
- It should therefore behave like --files-from in all other relevant aspects.
- Hence it should accept both UTF-8 and UTF-16 the same way as --files-from.
- This however makes it even more separate from -raw0 which is already very different from -raw.
- Therefore we decided to change the option names and make them behave like the following (effectively a three step ladder from flexible to strict):

| - | --files-from | --files-from-verbatim | --files-from-raw |
| - | ------------ | --------------------- | ---------------- |
| Previous name | - | --files-from-raw | --files-from-raw0 |
| Intended for list type | User-crafted (tolerant) | User-crafted (tolerant) | Generated (strict) |
| Parsing of lines | Globbed *patterns* | Verbatim *paths* | Verbatim *paths* |
| Separator/terminator | \n (between lines) | \n (between lines) | \0 (every line) |
| Whitespace at start/end of line | Ignored | Kept | Kept |
| # character at start of line | Line ignored | Kept | Kept |
| Empty lines | Ignored | Ignored | Error |
| Supports UTF-8 (*without* BOM) | Yes | Yes | Yes |
| Supports UTF-16 (*with* BOM) | Yes | Yes | No |

The reasoning to skip the 0 in -raw0, so it's named just -raw, was that we already deviated from how some other software does things (little point in trying to be "them"), not all software uses 0 in the name anyway, it's arguably an implementation detail (we don't name the other options -fromN or -verbatimN), etc.

@greatroar I'm sorry if this isn't what you intended when first opening the PR, but I hope you can find some comfort in that the strictest version is really strict and that a user should then be able to feed *any* path to it successfully :) Do you think the above specification for how the options should work is technically fine? In particular, did I get the UTF8 without BOM in the -raw option right?

Hopefully once this is done we can merge this, just needs a little final adjusting. If you're tired of editing I can fix the docs if you fix the code, just let me know.

greatroar:
I'm fine with any solution that lets me write a zero-terminated list of files. As for BOMs, there's a missing option, "supports UTF-8 with BOM". --files-from currently supports that. --fles-from-raw should not.

rawtaz:
@greatroar Thanks for correcting me. Did I get the table right now (I just edited it, second-to-last line)?

greatroar:
Yes! I just pushed updated docs.

changelog/unreleased/issue-2944
```diff
@@ -0,0 +1,12 @@
+Enhancement: Backup options --files-from-verbatim and --files-from-raw
+
+The new backup flags ``--files-from-verbatim`` and ``--files-from-raw``
+read a list of files to save from a file. Unlike the existing ``--files-from``,
+these options cause the filenames to not be interpreted  as glob patterns.
+whitespace in filenames is preserved as-is and no pattern expansion is done.
+
+These flags are recommended over ``--files-from`` when generating the list of
+files to back up from a script.
```
rawtaz, line 9: Syntax and grammar correction:

~~~
The new backup options `--files-from-verbatim` and `--files-from-raw`
reads a list of files to back up from a file. Unlike the existing `--files-from`,
these options does not interpret the listed filenames as glob patterns;
whitespace in filenames is preserved as-is and no pattern expansion is done.

These new flags are recommended over `--files-from` when generating the
list of files to back up from a script. Please see the documentation for specifics.
~~~

cmd/restic/cmd_backup.go
```diff
@@ -128,6 +122,8 @@ func init() {
 	f.MarkDeprecated("hostname", "use --host")
 
 	f.StringArrayVar(&backupOptions.FilesFrom, "files-from", nil, "read the files to backup from `file` (can be combined with file args/can be specified multiple times)")
+	f.StringArrayVar(&backupOptions.FilesFromVerbatim, "files-from-verbatim", nil, "read the files to backup from `file` (can be combined with file args/can be specified multiple times)")
+	f.StringArrayVar(&backupOptions.FilesFromRaw, "files-from-raw", nil, "read the files to backup from `file` (can be combined with file args/can be specified multiple times)")
```
rawtaz, line 126: Can we take the opportunity to change the `/` to `, ` inside the parenthesis for these two lines and the one above them?

cmd/restic/cmd_backup.go
```diff
@@ -203,12 +202,55 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+		}
+		names = append(names, name)
+	}
+}
+
 // Check returns an error when an invalid combination of options was set.
 func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 	if gopts.password == "" {
-		for _, filename := range opts.FilesFrom {
+		err := errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
```
rawtaz, line 249: Forgive my ignorance, but what is the point of saving this error in a variable and placing it here, instead of just returning it below like before?

cmd/restic/cmd_backup.go
```diff
@@ -373,19 +420,34 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
 			if len(expanded) == 0 {
 				Warnf("pattern %q does not match any files, skipping\n", line)
 			}
-			lines = append(lines, expanded...)
+			targets = append(targets, expanded...)
+		}
+	}
+
+	for _, file := range opts.FilesFromVerbatim {
+		fromfile, err := readPatternsFromFile(file, false)
```
rawtaz, line 428: `readPatternsFromFile()` always strips surrounding whitespace, but `-verbatim` should be just that, and not have this stripped. I guess add another boolean for that?

cmd/restic/cmd_backup.go
```diff
@@ -373,19 +420,34 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
...
+		}
+		targets = append(targets, fromfile...)
+	}
+
+	for _, file := range opts.FilesFromRaw {
+		fromfile, err := readFilenamesFromFileRaw(file)
+		if err != nil {
+			return nil, err
 		}
+		fmt.Println(file, fromfile)
```
rawtaz, line 440: Is this a left-over from development?

cmd/restic/cmd_backup.go
```diff
@@ -203,12 +202,55 @@ func readLinesFromFile(filename string) ([]string, error) {
 	return lines, nil
 }
 
+// readFilenamesFromFileRaw reads a list of filenames from the given file,
+// or stdin if filename is "-". Each filename is terminated by a zero byte,
+// which is stripped off.
+func readFilenamesFromFileRaw(filename string) (names []string, err error) {
```
rawtaz, line 208: I suggest we rename this (and the corresponding one below) to `readPathsFromFileRaw`, as it's really paths we're reading, not just filenames.

cmd/restic/cmd_backup.go
```diff
@@ -373,19 +420,34 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
...
+		targets = append(targets, fromfile...)
 	}
 
-	// merge files from files-from into normal args so we can reuse the normal
-	// args checks and have the ability to use both files-from and args at the
-	// same time
-	args = append(args, lines...)
+	// Merge args into files-from so we can reuse the normal args checks
+	// and have the ability to use both files-from and args at the same time.
+	targets = append(targets, args...)
```
rawtaz, line 446: The line *below this one I'm commenting on* should probably be checking `len(targets)` instead?

cmd/restic/cmd_backup_test.go
```diff
@@ -0,0 +1,112 @@
...
+		FilesFromRaw:      []string{f3.Name()},
+	}
+
+	targets, err := collectTargets(opts, []string{filepath.Join(dir, "cmdline arg")})
+	rtest.OK(t, err)
+	sort.Strings(targets)
+	rtest.Equals(t, expect, targets)
+}
+
+func TestReadFilenamesRaw0(t *testing.T) {
```
rawtaz, line 73: s/0//

rawtaz: Perhaps also s/Filenames/Paths/ as per earlier comment on function name in cmd_backup.go.

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
 Including Files
 ***************
 
-By using the ``--files-from`` option you can read the files you want to back
-up from one or more folders. This is especially useful if a lot of files have
-to be backed up that are not in the same folder or are maybe pre-filtered by
-other software.
+The options ``--files-from``, ``--files-from-verbatim`` and ``--files-from-raw``
+allow you to list the files to back up in one or more files. This is useful
+when a lot of files have to be backed up that are not in the same folder.
```
rawtaz, line 281: Every time I've read through this documentation I've found this paragraph to be diffuse. I suggest we change it to:

~~~
The options ``--files-from``, ``--files-from-verbatim`` and ``--files-from-raw``
allow you to list files that should be backed up, using one or more files. This is useful
when a lot of files have to be backed up, or when they are not all in the same folder.
~~~

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
 
-For example maybe you want to backup files which have a name that matches a
-certain pattern:
+The argument passed to ``--files-from`` must be the name of a text file that
+contains one pattern per line. The file must be encoded as UTF-8, or UTF-16
+with a byte-order mark. Leading and trailing whitespace are removed from the
```
rawtaz, line 285: s/are/is/

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+patterns. Empty lines and lines starting with a ``#`` are considered comments.
```
rawtaz, line 286: s/considered comments/ignored/

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+The patterns are expanded, when the file is read, by the Go function
+`filepath.Glob <https://golang.org/pkg/path/filepath/#Glob>`__.
+
+The option ``--files-from-verbatim`` has the same behavior as ``--files-from``,
+except that it contains literal filenames. It does not cause pattern expansion,
+comments are not allowed and whitespace is not trimmed off the filenames.
+Empty lines are still allowed, so that files can be grouped.
```
rawtaz, line 293: Just slight adjustments to this paragraph:

~~~
The option ``--files-from-verbatim`` has the same behavior as ``--files-from``,
except that it contains literal filenames. It does not expand any patterns, filenames
are parsed verbatim. Lines starting with a ``#`` are not ignored, leading and trailing
whitespace is not trimmed. Empty lines are still allowed, so that files can be grouped.
~~~

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+
+``--files-from-raw`` is a third variant that requires filenames to be terminated
+by a zero byte (the NUL character), so that it can even handle filenames
+that contain newlines or are not encoded as UTF-8 (except that on Windows,
+the file must must still contain filenames encoded in UTF-8).
```
rawtaz, line 298: Suggest:

~~~
containing newlines or that are not encoded as UTF-8 (except on Windows, where
the listed filenames must still be encoded in UTF-8).
~~~

doc/040_backup.rst
```diff
@@ -276,36 +276,53 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+This option is the safest choice when generating filenames from a script.
```
rawtaz, line 299: s/filenames/filename lists/

changelog/unreleased/issue-2944
greatroar: Done, except that options really is plural (reads=>read, does=>do).

cmd/restic/cmd_backup.go
greatroar: Made it a semicolon.

cmd/restic/cmd_backup.go
greatroar: I think I wanted shorter lines, but they don't get to 80 cols anyway, so reverted.

cmd/restic/cmd_backup.go
greatroar: Yes. Removed.

cmd/restic/cmd_backup.go
greatroar: Moved the whitespace trimming outside the function and extended the unit test.

cmd/restic/cmd_backup.go
greatroar: Done.

doc/040_backup.rst
greatroar: Done.

doc/040_backup.rst
greatroar: Ok.

doc/040_backup.rst
greatroar: Ok.

doc/040_backup.rst
greatroar: Ok.

doc/040_backup.rst
greatroar: Ok.

doc/040_backup.rst
greatroar: I don't think "using" clarifies much. How about:

> The options ``--files-from``, ``--files-from-verbatim`` and ``--files-from-raw`` allow you to list files that should be backed up in a file, rather than on the command line.

cmd/restic/cmd_backup.go
greatroar: The convention in the Go stdlib is to call paths filenames (or just names) as well. See the os pkg docs, for example.

rawtaz: I see, fair enough!

cmd/restic/cmd_backup.go
```diff
@@ -203,10 +202,52 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+		default:
+			return nil, err
+		}
+
+		name = name[:len(name)-1]
+		if name == "" {
+			// The empty filename is never valid. Handle this now to
+			// prevent downstream code from erroneously backing up
+			// filepath.Clean("") == ".".
+			return nil, errors.New("--from-files-raw0: empty filename in listing")
```
rawtaz, line 240: s/0//

cmd/restic/cmd_backup.go
```diff
@@ -156,11 +152,14 @@ func filterExisting(items []string) (result []string, err error) {
 	return
 }
 
-// readFromFile will read all lines from the given filename and return them as
-// a string array, if filename is empty readFromFile returns and empty string
-// array. If filename is a dash (-), readFromFile will read the lines from the
+// readPatternsFromFile will read all lines from the given filename and return
+// them as a string slice, after stripping off leading and trailing whitespace
+// and removing empty lines and comments.
```
rawtaz, line 157: Comment is a bit outdated due to the changes to whitespace, suggest:

~~~
// readPatternsFromFile will read all lines from the given filename and return
// them as a string slice, after removing empty lines and possibly comments.
~~~

doc/040_backup.rst
rawtaz: That's probably a mich clearer description, thank you. I felt that the previous sentence could be read in at least two different ways. Let's change it to your suggestion!

doc/040_backup.rst
```diff
@@ -276,36 +276,55 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+contains one pattern per line. The file must be encoded as UTF-8, or UTF-16
+with a byte-order mark. Leading and trailing whitespace is removed from the
+patterns. Empty lines and lines starting with a ``#`` are ignored.
+The patterns are expanded, when the file is read, by the Go function
+`filepath.Glob <https://golang.org/pkg/path/filepath/#Glob>`__.
+
+The option ``--files-from-verbatim`` has the same behavior as ``--files-from``,
+except that it contains literal filenames. It does expand patterns; filenames
+are listed verbatim.  Lines starting with a ``#`` are not ignored; leading and trailing
+whitespace is not trimmed off.
```
rawtaz, line 293: Now I'm just being picky, but should probably move "trailing" to the next line.

doc/040_backup.rst
```diff
@@ -276,36 +276,55 @@ suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for meg
...
+Empty lines are still allowed, so that files can be grouped.
+
+``--files-from-raw`` is a third variant that requires filenames to be terminated
+by a zero byte (the NUL character), so that it can even handle filenames
+that contain newlines or are not encoded as UTF-8
+(except on Windows, where the listed filenames must still be encoded in UTF-8).
```
rawtaz, line 299: Just nitpicking again, can you condense these lines so they're don't waste space?

rawtaz:
Made a few more comments but I think this is pretty much done in general. I have reviewed everything but the tests.

cmd/restic/cmd_backup.go
```diff
@@ -203,10 +202,52 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+	br := bufio.NewReader(r)
+	for {
+		name, err := br.ReadString(0)
+		switch err {
+		case nil:
+		case io.EOF:
+			if name == "" {
+				return names, nil
+			}
+			return nil, errors.New("--from-files-raw: trailing zero byte missing")
```
chrahunt, line 230: `--files-from-raw` instead of `--from-files-raw`

cmd/restic/cmd_backup.go
chrahunt: Also `--files-from-raw` instead of `--from-files-raw`

doc/040_backup.rst
greatroar: Ok, done.

doc/040_backup.rst
greatroar: I'm not going to manually format text that is input to an automatic formatter, except to make it easier to diff.

cmd/restic/cmd_backup.go
greatroar: I just realized that lines starting with whitespace, then a '#' are no longer comments. I fixed this by renaming this function readLines and restricting its functionality.

cmd/restic/cmd_backup.go
```diff
@@ -184,29 +182,61 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+	br := bufio.NewReader(r)
+	for {
+		name, err := br.ReadString(0)
+		switch err {
+		case nil:
+		case io.EOF:
+			if name == "" {
+				return names, nil
+			}
+			return nil, errors.New("--files-from-raw: trailing zero byte missing")
```
rawtaz, line 219: This causes a stack trace to be shown when the error occurs, I suggest replacing with `errors.Fatal()` or something similar.

cmd/restic/cmd_backup.go
```diff
@@ -184,29 +182,61 @@ func readLinesFromFile(filename string) ([]string, error) {
...
+		default:
+			return nil, err
+		}
+
+		name = name[:len(name)-1]
+		if name == "" {
+			// The empty filename is never valid. Handle this now to
+			// prevent downstream code from erroneously backing up
+			// filepath.Clean("") == ".".
+			return nil, errors.New("--files-from-raw: empty filename in listing")
```
rawtaz, line 229: This causes a stack trace to be shown when the error occurs, I suggest replacing with `errors.Fatal()` or something similar.

doc/040_backup.rst
rawtaz: I respect that. For me it's not about the generated content, it's about readability in the documentation source, which is equivalent to source code readability :) But never mind, I'll fix it!

rawtaz:
I did some manual testing of all three options and it seems to work in general. There were two error messages with stack traces though that I added a couple of comments on. Once that's fixed (assuming you agree with my suggestions) I think we can merge this. I haven't looked into the test though, just the rest of the code and documentation.

rawtaz (approved):
I think this works, I fixed the two error messages, rebased on master and squashed it all.

@MichaelEischer need to approve this PR before I can hit the merge button though, due to GitHub :/

MichaelEischer (approved):
LGTM.

rawtaz:
@greatroar Thanks for this one, I think a lot of users will find great value in it!

PR #3018, opened 2020-10-13
Fix IsProcessBackground on Linux with stdin redirection

greatroar:

ui/termstatus.IsProcessBackground previously assumed that stdin was the terminal, causing it to report false when stdin was redirected. It now checks the actual terminal's fd.

To see the faulty behavior, run `restic backup --stdin < /dev/zero &` (or omit the & and hit Ctrl+Z, then bg).

Not that I know.

Checklist
---------

### DIFF
--- a/internal/ui/termstatus/background_linux.go
+++ b/internal/ui/termstatus/background_linux.go
@@ -1,21 +1,23 @@
 package termstatus
 
 import (
-	"syscall"
-	"unsafe"
-
 	"github.com/restic/restic/internal/debug"
+
+	"golang.org/x/sys/unix"
 )
 
-// IsProcessBackground reports whether the current process is running in the background.
-func IsProcessBackground() bool {
-	var pid int
-	_, _, err := syscall.Syscall(syscall.SYS_IOCTL, uintptr(syscall.Stdin), syscall.TIOCGPGRP, uintptr(unsafe.Pointer(&pid)))
-
-	if err != 0 {
+// IsProcessBackground reports whether the current process is running in the
+// background. fd must be a file descriptor for the terminal.
+func IsProcessBackground(fd uintptr) bool {
+	bg, err := isProcessBackground(fd)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/ui/termstatus/background_linux.go
```diff
@@ -1,21 +1,23 @@
...
+	bg, err := isProcessBackground(fd)
+	if err != nil {
 		debug.Log("Can't check if we are in the background. Using default behaviour. Error: %s\n", err.Error())
 		return false
 	}
+	return bg
+}
 
-	return pid != syscall.Getpgrp()
+func isProcessBackground(fd uintptr) (bool, error) {
```
MichaelEischer, line 20: Is the split into `IsProcessBackground` and `isProcessBackground` necessary?

internal/ui/termstatus/status.go
```diff
@@ -95,7 +95,7 @@ func (t *Terminal) run(ctx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
-			if IsProcessBackground() {
+			if IsProcessBackground(t.fd) {
```
MichaelEischer, line 98: This is equivalent to stdout. Why not just replace stdin with stdout in `IsProcessBackground`?

MichaelEischer (changes requested):
Shouldn't the command be `restic backup --stdin < /dev/zero &`? At least that's what works for me to reproduce the reported problem.

internal/ui/termstatus/background_linux.go
greatroar: This is a holdover from an earlier attempt. I'll merge the two.

internal/ui/termstatus/status.go
greatroar: The Terminal type is designed to work with any Writer that has an Fd method and never mentions stdout. If for whatever reason it's changed to use a different fd, IsProcessBackground would break.

greatroar:
Yes, I meant `< /dev/zero`.

MichaelEischer (approved):
LGTM

PR #3019, opened 2020-10-13
Refactor index decoding

greatroar:

There were three places in the codebase that did

decode new-style index -> check error -> decode old-style index

via a system of callbacks. I've replaced all three by a single DecodeIndex that can handle both formats, saving some 20 lines of code, two public functions in internal/repository and one public error type.

There's still some repetition in the code, but I didn't see a clean way of getting rid it. Two of the call sites pass a buffer around that would have to become an argument and a fourth return value on DecodeIndex.

New-style decoding is as fast as it was before:

```
name                   old time/op  new time/op  delta
DecodeIndex-8           4.19s ± 2%   4.16s ± 1%   ~     (p=0.105 n=10+10)
DecodeIndexParallel-8   4.18s ± 1%   4.17s ± 1%   ~     (p=0.360 n=10+8)
```

Decoding old-format indices no longer requires loading and decrypting twice, so that may be faster, but there's no benchmark for that.

No.

Checklist
---------

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -117,12 +117,16 @@ func (c *Checker) LoadIndex(ctx context.Context) (hints []error, errs []error) {
 			debug.Log("worker got file %v", fi.ID.Str())
 			var err error
 			var idx *repository.Index
-			idx, buf, err = repository.LoadIndexWithDecoder(ctx, c.repo, buf[:0], fi.ID, repository.DecodeIndex)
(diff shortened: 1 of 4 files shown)


MichaelEischer (approved):
LGTM. There is a minimal change in behavior such that a index is now only reported to have the old format when it can actually be loaded as such, which is actually more accurate than before.

Issue #322, opened 2015-10-24
Make IO-parallelism configurable


episource:
**Restic version:**  5de36df 
**Observed behavior:**
IO-parallelism is hard coded. E. g. [defaultParallelism = 40 (checker.go:50)](https://github.com/restic/restic/blob/5de36dfdf09726287521f47b20b0c80a682c2a37/checker/checker.go#L50).

The hard coded parallelism might be well chosen for solid state disks, but is not always benefical and could slow down IO-operations or cause other issues. In practice I'm observing issues checking a repository on a samba/cifs mounted network location: IO errors due to NT_STATUS_INSUFF_SERVER_RESOURCES. Reducing [defaultParallelism (checker.go:50)](https://github.com/restic/restic/blob/5de36dfdf09726287521f47b20b0c80a682c2a37/checker/checker.go#L50) solves these issues but requires code changes.

**Expected behavior:**
IO-Parallelism should be configurable with a command line parameter. E. g. `restic --max-parallel-io 10 [...]`. This option should be available for backup, restore and check operations.

**Related:** Reducing IO parallelism might also resolve #313. Exhausting resource consumption as reported in #300, #316, #298 might also be related to high IO-parallelism.

fw42:
Instead of exposing this to the user with a configuration parameter, I think it would be more useful to try to come up with an (automatic) way of determining an ideal value. This needs to be a sane (potentially automatic) default, but most users can't make an educated choice for this value by themselves and it will just make the command line user interface more confusing, imho.

episource:
Consider splitting help output into a reduced basic and an advanced section: `restic help` shows a reduced basic help text where as `restic help --advanced` shows all available options. Any user happy with restic's defaults reads the reduced help section, only.

fw42:
"Advanced settings" are a total anti-pattern, in my opinion. Sane defaults and convention over configuration are much more useful and lead to better usability in the long-term, in my experience.

fd0:
I also don't like "advanced settings", however tempting that may be. I see you point @episource that there should be a way to either manually or automatically select the IO parallelism. #313 is not affected because the fuse backend doesn't limit the amount of parallel IO operations, that comes from the kernel.

julianbrost:
We recently into the same issue on a VM host which doesn't have too much spare I/O performance. Running restic with its default parallelism (which now is 10) results in a significant overall performance degradation.

[...]

PR #3028, opened 2020-10-18
Reveal archiver's concurrency options to backend command

### LINKED ISSUE #2204: rclone parallel uploads
Output of `restic version`
--------------------------
restic 0.9.4 (v0.9.4-37-g7b8d1dc0) compiled with go1.11.4 on linux/arm64

uploading more files in parallel through rclone.

What are you trying to do?
--------------------------

restic uploads 4 files in parallel through rclone. Due to some weird limitations on the remote storage, they only allow 500kb speed or so per file uploaded, however, if we increase the number of parallels connections we can use more bandwidth and reduce backup time (tested and confirmed with rclone sync).

Example of current rclone uploads to Jotta Cloud (just as an example, it can be with oneDrive, hubic, opendrive, b2, etc...) command used:

`restic  backup /srv/dev-disk-by-label-DATAdroid  --exclude-file=/root/filter.restic -v -v  --repo rclone:oDroid-Jotta:/restic/dataDroid -o rclone.connections="10"   -o rclone.args="serve restic --stdio --rc --rc-addr localhost:5582 --rc-no-auth --config /root/.config/rclone/rclone.conf --log-file /mnt/cache/log/restic.jotta -v --transfers 10 --stats=20s"`

achieving the following output...

```
Transferred:        1.307G / 1.316 GBytes, 99%, 1.616 MBytes/s, ETA 5s
Errors:                 0
Checks:                 0 / 0, -
Transferred:          275 / 279, 99%
Elapsed time:    13m48.1s
Transferring:
 *     data/45/45a7f45c3de3…0be968fb3887a4f3316f:  7% /4.005M, 163.955k/s, 23s
 *     data/4a/4a9a4b16f1a9…668a9089f9602d2c731e: 27% /4.593M, 451.721k/s, 6s
 *     data/97/97965e0121dc…c76f0b070a9b9876f7d4: 97% /4.146M, 565.974k/s, 0s
 *     data/e4/e4562dd9205e…1dfd2c0b293fa66dfc4b: 54% /4.599M, 520.954k/s, 4s
```

how it should be...

as many files as the parameter rclone.connections (10 in this command), or a new parameter, is set up, so then we could be uploading 10 files (instead of our current 4, the default by rclone with no modifiers) reducing the time of the backups.

I have seen in the code (restic/internal/backend/rclone/config.go), we can configure the number of connections, however, that does not change the number of uploads through rclone.

`Connections uint option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`

and in 

```
func NewConfig() Config {
        return Config{
                Connections: 5,
        }
}
```
Nevertheless, I guess above configurations are more about the number of connections to rclone rather than the amount of files to be uploaded in parallel. am I right?

I am starting to use this great piece of software and I guess that can help me to keep my data safe, I would really like the idea of having a software fully customizable. Please, allow setting up a new parameter that will be the amount of parallels files to be uploaded. We can then extensively use more bandwidth uploading more files.

I have a lot of experience uploading backup to cloud storage providers, and is very annoying having enough bandwidth to upload 1Tb in less than 12hours and spend 3 days or more in doing so....
 
(1Tb of data to be uploaded to jotta through restic/rclone is taking 170hours).

Keep doing such a good stuff!!, you rock!

aspcartman:
This PR adds an ability to set the `archiver.Options` values in `backup` command. There are three options:

```go
// Options is used to configure the archiver.
type Options struct {
	// FileReadConcurrency sets how many files are read in concurrently. If
	// it's set to zero, at most two files are read in concurrently (which
	// turned out to be a good default for most situations).
	FileReadConcurrency uint

	// SaveBlobConcurrency sets how many blobs are hashed and saved
	// concurrently. If it's set to zero, the default is the number of CPUs
	// available in the system.
	SaveBlobConcurrency uint

	// SaveTreeConcurrency sets how many trees are marshalled and saved to the
	// repo concurrently.
	SaveTreeConcurrency uint
}
```

And there were no ability to alter them any how. They were always set to defaults. I added those three fields to the `backupOptions` struct. Now you can set them as `--file-read-concurrency=20`, `--save-blob-concurrency=20` and `--save-tree-concurrency=20` in `backup` command.

The purpose of those changes is to allow parallel uploads of more than 2 files when using rclone (#2204). Increasing the parallelism dramatically increases the speeds of the upload for unstable and rate limited connections and rclone backends that tend to hang.

For example YandexDisk backend hangs after the upload completes to calculate the hash sum of the file, and only then responds to the POST request. The larger the file - the longer it hangs. Uploading 5mb photos in 10 threads leads to 8 threads stuck on 100%, waiting for the response, and only 2 threads actually working most of the time.

Restic provides us with the `rclone.connections` option, yet increasing it doesn't change the fact that the archiver works with 2 file uploads at a time by default. 

The following command works ten times faster now:
```
restic -r rclone:yd:Restic -o rclone.connections=20 -o rclone.args="serve restic --stdio --b2-hard-delete --drive-use-trash=false --transfers=20" backup --file-read-concurrency=20 --save-blob-concurrency=20 ~/Music
```

Closes #2204 

Checklist
---------

- [ ] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -96,6 +96,10 @@ type BackupOptions struct {
 	TimeStamp               string
 	WithAtime               bool
 	IgnoreInode             bool
+
+	FileReadConcurrency uint
+	SaveBlobConcurrency uint
+	SaveTreeConcurrency uint
 }
 
 var backupOptions BackupOptions
@@ -129,6 +133,10 @@ func init() {
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "`time` of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
 	f.BoolVar(&backupOptions.IgnoreInode, "ignore-inode", false, "ignore inode number changes when checking for modified files")
+
+	f.UintVar(&backupOptions.FileReadConcurrency, "file-read-concurrency", 0, "sets how many files are read in concurrently by archiver (default: 2)")
+	f.UintVar(&backupOptions.SaveBlobConcurrency, "save-blob-concurrency", 0, "sets how many blobs are hashed and saved concurrently by archiver (default: runtime.NumCPU())")
+	f.UintVar(&backupOptions.SaveTreeConcurrency, "save-tree-concurrency", 0, "sets how many trees are marshalled and saved to the repo concurrently by archiver (default: save-blob-concurrency * 20)")
 }
 
 // filterExisting returns a slice of all existing items, or an error if no
@@ -575,7 +583,11 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	}
 	t.Go(func() error { return sc.Scan(t.Context(gopts.ctx), targets) })
 
-	arch := archiver.New(repo, targetFS, archiver.Options{})
+	arch := archiver.New(repo, targetFS, archiver.Options{
+		FileReadConcurrency: backupOptions.FileReadConcurrency,
+		SaveBlobConcurrency: backupOptions.SaveBlobConcurrency,
+		SaveTreeConcurrency: backupOptions.SaveTreeConcurrency,
+	})
 	arch.SelectByName = selectByNameFilter
 	arch.Select = selectFilter
 	arch.WithAtime = opts.WithAtime


MichaelEischer:
I think this PR solves the wrong problem in the wrong way: There's no point in manually setting SaveTreeConcurrency, once #2696 is fixed, it will be fixed at `GOMAXPROCS`. #2696 will also allow choosing the number of concurrent uploads independent of the file read and blob save concurrency. See #2750 for further discussions on that topic.

In addition restic should automatically adjust the number of uploads to the backend concurrency limit set via `-o rclone.connections=20`, there's no need to specify that manually.

aspcartman:
@MichaelEischer I don't seem to understand your point : ) Mentioned #2696 is an issue, not a PR. I don't see any fixes anywhere.

My point is dramatically simple: 
FileReadConcurrency=2 - backup takes 1.5 hour. 
FileReadConcurrency=20 - 8 minutes. 
A user needs a command line option to change this value.

> restic should automatically adjust the number of uploads to the backend concurrency limit set via -o rclone.connections=20, there's no need to specify that manually.

I don't see such code in the repo, restic doesn't do that. And the number of concurrent connections != number of concurrent uploads.

MichaelEischer:
@aspcartman I was mostly speaking how the backup should work and not how it does right now. We can't just add cli options which aren't useful for much longer than a few month (hopefully). There is no PR to fix #2696 right now; currently I'm busy with reviewing the prune optimizations.

I don't doubt that manually increasing the number of file reader and blob savers improves performance in your case, but that is merely a hack not a proper solution. `-o rclone.connections=20` should be the only option that has to be set, the rest should automatically do the right thing. Please take a look at #2750 for a discussion of some of the concurrency options and also at #322.

MichaelEischer:
Closing as sort of duplicate of #2750.

Issue #1800, opened 2018-05-22
Using source-directory w/o xattr fails


EleRas:
## Output of `restic version`
restic 0.9.0 compiled with go1.10.2 on linux/amd64

## How did you run restic exactly?

> restic -r rclone:gdrive:restic --verbose --verbose init
```
enter password for new repository: 
enter password again: 
created restic repository 86c6846e1f at rclone:gdrive:restic

Please note that knowledge of your password is required to access
the repository. Losing your password means that your data is
irrecoverably lost.
```
> restic -r rclone:gdrive:restic --verbose --verbose backup /mnt/photos

/mnt/photos is a mount:
> //<server>/photo on /mnt/photos type cifs (rw,relatime,vers=default,cache=strict,username=<user>,domain=,uid=1000,forceuid,gid=100,forcegid,addr=192.168.1.17,file_mode=0775,dir_mode=0775,soft,nounix,serverino,mapposix,noperm,rsize=1048576,wsize=1048576,echo_interval=60,actimeo=1,user=<user>,x-systemd.automount)

This mount does NOT allow xattr.

restic output:
```
open repository
enter password for repository: 
repository 86c6846e opened successfully, password is correct
lock repository
rclone: 2018/05/22 19:25:20 ERROR : locks: error listing: directory not found
load index files
rclone: 2018/05/22 19:25:23 ERROR : index: error listing: directory not found
rclone: 2018/05/22 19:25:23 ERROR : snapshots: error listing: directory not found
start scan on [/mnt/photos]
start backup on [/mnt/photos]
error: NodeFromFileInfo: Listxattr: xattr.List /mnt/photos : no data available
new       /mnt/, saved in 0.022s (0 B added, 13 B metadata)

Files:           0 new,     0 changed,     0 unmodified
Dirs:            1 new,     0 changed,     0 unmodified
Data Blobs:      0 new
Tree Blobs:      2 new
Added:      377 B

processed 0 files, 0 B in 0:16
snapshot 99ae013a saved
context canceled B, total 32856 files 13.350 GiB, 1 errors
```
Notice how restic does not process any files, even if it tells it has a total of 32856 files. After this restic quits and no data is in snapshot

## What backend/server/service did you use to store the repository?
google drive via rclone

## Expected behavior
restic creates a snapshot containing data

## Actual behavior
restic creates an empty snapshot which can't be restored:
>restic -r rclone:gdrive:restic --verbose --verbose restore 99ae013a --target /tmp/tmporestore
```
[...]

PR #3034, opened 2020-10-21
bugfix: omit ENOTDATA for extended attributes

hoyho:
Signed-off-by: hoyho <luohaihao@gmail.com>

node_xattr.go have some operation on file extended attributes like xattr.Get xattr.List .
However, some file share do not support such operations. 
For example, backup source files from a cifs share may cause restic reporting an ENODATA error
and mount the cifs share with option `nouser_xattr` can omit this error because fs will report `ENOTSUP` instead and restic will absorb this error.
As mentioned in https://github.com/restic/restic/issues/1800#issuecomment-706516649 , we can whitelist ENOTDATA error

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1800
+++ b/changelog/unreleased/issue-1800
@@ -1,0 +1,8 @@
+Bugfix: Ignore 'no data available' error during backup
+
+restic failed to backup files on some filesystems, for example certain configurations
+of CIFS on Linux, which return a "no data available" error when reading extended
+attributes. These errors are now ignored.
+
+https://github.com/restic/restic/issues/1800
+https://github.com/restic/restic/pull/3034
--- a/internal/restic/node_xattr.go
+++ b/internal/restic/node_xattr.go
@@ -16,7 +16,8 @@ import (
 // Getxattr retrieves extended attribute data associated with path.
 func Getxattr(path, name string) ([]byte, error) {
 	b, e := xattr.Get(path, name)
-	if err, ok := e.(*xattr.Error); ok && err.Err == syscall.ENOTSUP {
+	if err, ok := e.(*xattr.Error); ok &&
+		(err.Err == syscall.ENOTSUP || err.Err == xattr.ENOATTR) {
 		return nil, nil
 	}
 	return b, errors.Wrap(e, "Getxattr")
@@ -26,7 +27,8 @@ func Getxattr(path, name string) ([]byte, error) {
 // given path in the file system.
 func Listxattr(path string) ([]string, error) {
 	s, e := xattr.List(path)
-	if err, ok := e.(*xattr.Error); ok && err.Err == syscall.ENOTSUP {
+	if err, ok := e.(*xattr.Error); ok &&
+		(err.Err == syscall.ENOTSUP || err.Err == xattr.ENOATTR) {
 		return nil, nil
 	}
 	return s, errors.Wrap(e, "Listxattr")
@@ -35,7 +37,8 @@ func Listxattr(path string) ([]string, error) {
 // Setxattr associates name and data together as an attribute of path.
 func Setxattr(path, name string, data []byte) error {
 	e := xattr.Set(path, name, data)
-	if err, ok := e.(*xattr.Error); ok && err.Err == syscall.ENOTSUP {
+	if err, ok := e.(*xattr.Error); ok &&
+		(err.Err == syscall.ENOTSUP || err.Err == xattr.ENOATTR) {
 		return nil
 	}
 	return errors.Wrap(e, "Setxattr")


internal/restic/node_xattr.go
```diff
@@ -16,6 +17,10 @@ import (
 // Getxattr retrieves extended attribute data associated with path.
 func Getxattr(path, name string) ([]byte, error) {
 	b, e := xattr.Get(path, name)
+	if err, ok := e.(*xattr.Error); ok && err.Err == syscall.ENODATA {
```
MichaelEischer, line 20: Please merge the if condition to check for `ENODATA` and `ENOTSUP`.

internal/restic/node_xattr.go
```diff
@@ -16,6 +17,10 @@ import (
...
+		debug.Log("Getxattr error:ENODATA, path:%s", path)
```
MichaelEischer, line 21: I don't think the `debug.Log` is necessary.

MichaelEischer (changes requested):
Thanks for the patch. I've just two short comments on the code, see below. Please also add a short changelog entry, please take a look at the [changelog entry template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE).

internal/restic/node_xattr.go
hoyho: thanks for the feedback. 
I have updated the commit and add a issue template

internal/restic/node_xattr.go
hoyho: done

internal/restic/node_xattr.go
```diff
@@ -16,7 +16,8 @@ import (
 // Getxattr retrieves extended attribute data associated with path.
 func Getxattr(path, name string) ([]byte, error) {
 	b, e := xattr.Get(path, name)
-	if err, ok := e.(*xattr.Error); ok && err.Err == syscall.ENOTSUP {
+	if err, ok := e.(*xattr.Error); ok &&
+		(err.Err == syscall.ENOTSUP || err.Err == syscall.ENODATA) {
```
MichaelEischer, line 20: Apparently FreeBSD (see Travis failures) does not have `syscall.ENODATA`. As the error has only been reported on linux, it's probably best to just ignore ENODATA for Linux.

So we'll need something like a `isXattrNoData` function. With one version in `node_linux.go` which returns true if the error is ENODATA. And a `node_xattr_bsd.go` which is not build on Linux and all platforms already excluded in `node_xattr.go` (see the `// +build !netbsd` lines).

changelog/unreleased/issue-1800
```diff
@@ -0,0 +1,10 @@
+Bugfix: omit ENOTDATA for extended attributes
```
MichaelEischer, line 1: That sounds rather technical and it is not clear which command is affected. What about
`Bugfix: Ignore 'no data available' error during backup`

changelog/unreleased/issue-1800
```diff
@@ -0,0 +1,10 @@
...
+
+Restic had some operations on file extended attributes like xattr.Get xattr.List.
```
MichaelEischer, line 3: I'd suggest to word this as follows:
```
restic failed to backup files on some filesystems, for example certain configurations
of CIFS on Linux, which return a "no data available" error when reading extended
attributes. These errors are now ignored.
```

MichaelEischer (changes requested):
I just noticed that the code changes do not build on FreeBSD. Please change the code according to my comments below.

internal/restic/node_xattr.go
hoyho: Thanks for the instruction. I don't have much experience about cross-platform developing But I will try it later.

hoyho: I'm using macOS and It also have definition of syscall.ENODATA. So I add a file node_xattr_darwin.go

internal/restic/node_xattr_darwin.go
```diff
@@ -0,0 +1,15 @@
+package restic
```
greatroar, line 1: This file is the exact same as the Linux version, but we haven't established that the problem occurs on MacOS.

internal/restic/node_xattr_linux.go
```diff
@@ -0,0 +1,15 @@
+package restic
+
+import (
+	"syscall"
+
+	"github.com/pkg/xattr"
+)
+
+//isXattrNoData indicates if an error is NODATA
```
greatroar, line 9: The comment should explain why we're checking.

internal/restic/node_xattr_bsd.go
```diff
@@ -0,0 +1,7 @@
+// +build freebsd netbsd openbsd
+
+package restic
+
+func isXattrNoData(err error) bool {
+	return false //no such error
```
greatroar, line 6: NetBSD has ENODATA, so the comment is incorrect. You can check this by compiling

```
package foo

import "syscall"

var _ = syscall.ENODATA
```

with GOOS=netbsd, etc.

A better comment would be "xattr should never return ENODATA."

internal/restic/node_xattr_darwin.go
hoyho: ok, node_xattr_darwin.go have been removed

internal/restic/node_xattr_linux.go
hoyho: updated

internal/restic/node_xattr_bsd.go
hoyho: updated

internal/restic/node_xattr.go
hoyho: now it removed 😅

internal/restic/node_xattr_linux.go
```diff
@@ -0,0 +1,16 @@
+package restic
+
+import (
+	"syscall"
+
+	"github.com/pkg/xattr"
+)
+
+//isXattrNoData check if the error is NODATA('no data available')
+//so the caller can decide whether to ignore this error or not
```
greatroar, line 10: There should be a space after "//". The compiler interprets some comments without that space specially.

changelog/unreleased/issue-1800
```diff
@@ -0,0 +1,9 @@
+Bugfix: Ignore 'no data available' error during backup
+
+restic failed to backup files on some filesystems, for example certain configurations
+of CIFS on Linux, which return a "no data available" error when reading extended
+attributes. These errors are now ignored.
+
+
```
greatroar, line 7: One blank line too many.

greatroar:
I just found out that the xattr package has an ENOATTR constant, which is ENODATA on Linux and the real ENOATTR on BSD/Darwin. I think we should check for that instead, so we don't need per-platform files.

MichaelEischer:
I agree with @greatroar that using `ENOATTR` seems to be the cleanest solution, please use that instead of the per-platform files. @hoyho Sorry for the back and forth.

hoyho:
good to hear that 😅

changelog/unreleased/issue-1800
hoyho: done

hoyho:
Please take a look @greatroar @MichaelEischer

changelog/unreleased/issue-1800
```diff
@@ -0,0 +1,8 @@
+Bugfix: Ignore 'no data available' error during backup
+
+restic failed to backup files on some filesystems, for example certain configurations
+of CIFS on Linux, which return a "no data available" error when reading extended
+attributes. These errors are now ignored.
+
+https://github.com/restic/restic/issues/1800
+https://github.com/restic/restic/pull/3034
```
rawtaz, line 8: Thanks for writing a nice changelog entry in line with the example and how it should be written!

MichaelEischer (approved):
LGTM. This patch now really looks nice and clean and simple :-) .

PR #3035, opened 2020-10-21
restic mount: don't umount unless we actually tried to mount

greatroar:

Failure to open a repository in restic mount (wrong password, wrong path, ...) would previously result in an "unable to umount" message, because the umount cleanup handler was installed before the actual mount was attempted. This PR fixes that by installing the handler later.

No.

Checklist
---------


MichaelEischer (approved):
Good catch, thanks! LGTM.

PR #3036, opened 2020-10-21
Clean up internal/fuse

greatroar:

Cleans up the internal/fuse package. Mainly:

* Generate inodes from parent directories instead of the root. Using the root gives the same inodes for distinct directories.
* Don't generate "." and ".." entries. bazil.org/fuse works fine without these, its examples don't generate them, and the ".." entries were incorrect anyway: SnapshotsDir was using the root for "..", regardless of its actual parent.
* Merged MetaDir back into Root. It was introduced in 2c02efd1fe1d35239e0bbacabebc6c2ef9140cc4 as a refactor, but all of its users have since dropped it except Root.

No.

Checklist
---------

### DIFF
--- a/internal/fuse/root.go
+++ b/internal/fuse/root.go
@@ -68,7 +68,7 @@ func NewRoot(repo restic.Repository, cfg Config) *Root {
 		}
 	}
 
-	root.SnapshotsDir = NewSnapshotsDir(root, rootInode, NewSnapshotsDirStructure(root, cfg.PathTemplates, cfg.TimeTemplate), "")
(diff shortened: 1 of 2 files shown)


aawsome:
Maybe consider #2913 first, which also refactors and simplifies this... (+ customizes the snapshots dirs)

greatroar:
I can restrict this PR to the pieces that don't conflict with #2913 if desired.

MichaelEischer:
I've rebased to PR on top of #2913.

MichaelEischer (approved):
LGTM.

MichaelEischer:
I've decided to pass the correct parentInode along the snapshot directories. If `.` and `..` are not generated, the directory navigation still works, but `.` and `..` will be missing from the output of e.g. `ls -la .` which is rather weird.

PR #3038, opened 2020-10-24
Check data subset: check random percentage subset

### LINKED ISSUE #2186: Increase the effective granularity of `--read-data-subset`

fgma:

Add the option to check a random subset of pack files. Currently one can only check data using ``--read-data-subset=3/5`` which checks a specific subset of data or the whole available data. To partially check different parts of pack files this needs some form of automation.

This PR adds the possibility to specify a percentage of pack files to check using ``--read-data-subset=2.5%``. This will check 2.5% of the pack files in the repository for checking.

The old variant is still supported as before.

Closes #2186

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3038
+++ b/changelog/unreleased/pull-3038
@@ -1,0 +1,8 @@
+Enhancement: Allow specifying percentage in check's --read-data-subset
+
+We've enhanced the command-line option --read-data-subset to also accept a
+percentage. This will check the given percentage of pack files which are
+randomly selected on each run.
+
+https://github.com/restic/restic/pull/3038
+https://github.com/restic/restic/issues/2186
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -83,7 +83,7 @@ func checkFlags(opts CheckOptions) error {
 
 			if percentage <= 0.0 || percentage > 100.0 {
 				return errors.Fatal(
-					"check flag --read-data-subset=n% n must be above 0.0% and 100.0%")
+					"check flag --read-data-subset=n% n must be above 0.0% and at most 100.0%")
 			}
 		}
 	}
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -1,8 +1,8 @@
 package main
 
 import (
-	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"strconv"
 	"strings"
 
@@ -53,25 +53,38 @@ func init() {
 
 	f := cmdCheck.Flags()
 	f.BoolVar(&checkOptions.ReadData, "read-data", false, "read all data blobs")
-	f.StringVar(&checkOptions.ReadDataSubset, "read-data-subset", "", "read subset n of m data packs (format: `n/m`)")
+	f.StringVar(&checkOptions.ReadDataSubset, "read-data-subset", "", "read a `subset` of data packs, specified as 'n/t' for specific subset or either 'x%' or 'x.y%' for random subset")
 	f.BoolVar(&checkOptions.CheckUnused, "check-unused", false, "find unused blobs")
 	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use the cache")
 }
 
 func checkFlags(opts CheckOptions) error {
 	if opts.ReadData && opts.ReadDataSubset != "" {
-		return errors.Fatalf("check flags --read-data and --read-data-subset cannot be used together")
+		return errors.Fatal("check flags --read-data and --read-data-subset cannot be used together")
 	}
 	if opts.ReadDataSubset != "" {
 		dataSubset, err := stringToIntSlice(opts.ReadDataSubset)
-		if err != nil || len(dataSubset) != 2 {
-			return errors.Fatalf("check flag --read-data-subset must have two positive integer values, e.g. --read-data-subset=1/2")
-		}
-		if dataSubset[0] == 0 || dataSubset[1] == 0 || dataSubset[0] > dataSubset[1] {
-			return errors.Fatalf("check flag --read-data-subset=n/t values must be positive integers, and n <= t, e.g. --read-data-subset=1/2")
-		}
-		if dataSubset[1] > totalBucketsMax {
-			return errors.Fatalf("check flag --read-data-subset=n/t t must be at most %d", totalBucketsMax)
+		argumentError := errors.Fatal("check flag --read-data-subset must have two positive integer values or a percentage, e.g. --read-data-subset=1/2 or --read-data-subset=2.5%%")
+		if err == nil {
+			if len(dataSubset) != 2 {
+				return argumentError
+			}
+			if dataSubset[0] == 0 || dataSubset[1] == 0 || dataSubset[0] > dataSubset[1] {
+				return errors.Fatal("check flag --read-data-subset=n/t values must be positive integers, and n <= t, e.g. --read-data-subset=1/2")
+			}
+			if dataSubset[1] > totalBucketsMax {
+				return errors.Fatalf("check flag --read-data-subset=n/t t must be at most %d", totalBucketsMax)
+			}
+		} else {
+			percentage, err := parsePercentage(opts.ReadDataSubset)
+			if err != nil {
+				return argumentError
+			}
+
+			if percentage <= 0.0 || percentage > 100.0 {
+				return errors.Fatal(
+					"check flag --read-data-subset=n% n must be above 0.0% and 100.0%")
+			}
 		}
 	}
 
@@ -96,6 +109,21 @@ func stringToIntSlice(param string) (split []uint, err error) {
 		result[idx] = uint(uintval)
 	}
 	return result, nil
+}
+
+// ParsePercentage parses a percentage string of the form "X%" where X is a float constant,
+// and returns the value of that constant. It does not check the range of the value.
+func parsePercentage(s string) (float64, error) {
+	if !strings.HasSuffix(s, "%") {
+		return 0, errors.Errorf(`parsePercentage: %q does not end in "%%"`, s)
+	}
+	s = s[:len(s)-1]
+
+	p, err := strconv.ParseFloat(s, 64)
+	if err != nil {
+		return 0, errors.Errorf("parsePercentage: %v", err)
+	}
+	return p, nil
 }
 
 // prepareCheckCache configures a special cache directory for check.
@@ -233,22 +261,8 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
 		}
 	}
 
-	doReadData := func(bucket, totalBuckets uint) {
-		packs := make(map[restic.ID]int64)
-		for pack, size := range chkr.GetPacks() {
-			// If we ever check more than the first byte
-			// of pack, update totalBucketsMax.
-			if (uint(pack[0]) % totalBuckets) == (bucket - 1) {
-				packs[pack] = size
-			}
-		}
+	doReadData := func(packs map[restic.ID]int64) {
 		packCount := uint64(len(packs))
-
-		if packCount < chkr.CountPacks() {
-			Verbosef(fmt.Sprintf("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, packCount, chkr.CountPacks(), totalBuckets))
-		} else {
-			Verbosef("read all data\n")
-		}
 
 		p := newProgressMax(!gopts.Quiet, packCount, "packs")
 		errChan := make(chan error)
@@ -264,10 +278,26 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
 
 	switch {
 	case opts.ReadData:
-		doReadData(1, 1)
+		Verbosef("read all data\n")
+		doReadData(selectPacksByBucket(chkr.GetPacks(), 1, 1))
 	case opts.ReadDataSubset != "":
-		dataSubset, _ := stringToIntSlice(opts.ReadDataSubset)
-		doReadData(dataSubset[0], dataSubset[1])
+		var packs map[restic.ID]int64
+		dataSubset, err := stringToIntSlice(opts.ReadDataSubset)
+		if err == nil {
+			bucket := dataSubset[0]
+			totalBuckets := dataSubset[1]
+			packs = selectPacksByBucket(chkr.GetPacks(), bucket, totalBuckets)
+			packCount := uint64(len(packs))
+			Verbosef("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, packCount, chkr.CountPacks(), totalBuckets)
+		} else {
+			percentage, _ := parsePercentage(opts.ReadDataSubset)
+			packs = selectRandomPacksByPercentage(chkr.GetPacks(), percentage)
+			Verbosef("read %.1f%% of data packs\n", percentage)
+		}
+		if packs == nil {
+			return errors.Fatal("internal error: failed to select packs to check")
+		}
+		doReadData(packs)
 	}
 
 	if errorsFound {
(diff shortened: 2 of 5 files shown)


changelog/unreleased/pull-3038
```diff
@@ -0,0 +1,7 @@
+Enhancement: Specify a percentage when using --read-data-subset
```
rawtaz, line 1: Suggestion: `Enhancement: Allow specifying percentage in check's --read-data-subset`

cmd/restic/cmd_check.go
```diff
@@ -233,27 +264,44 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
...
+		return packs
+	}
 
-		if packCount < chkr.CountPacks() {
-			Verbosef(fmt.Sprintf("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, packCount, chkr.CountPacks(), totalBuckets))
-		} else {
-			Verbosef("read all data\n")
+	doSelectPacksByPercentage := func(percentage float64) *restic.IDSet {
+		packCount := len(chkr.GetPacks())
+		packsToCheck := int(math.Max((((float64(packCount)*percentage)/100.0)+0.5)-1.0, 0.0))
```
fgma, line 281: Consider enforcing to check at least one pack if given percentage is too low but above 0.0.

MichaelEischer:
@fgma How do this PR relate to #2801?

rawtaz:
Would it be possible to make the percentage parser allow specifying an integer percentage instead of always requiring a float? I imagine a lot of people don't need fine-grained and would like to just say `5%` or so.

rawtaz:
Since I cannot add a comment in a place where you haven't made any changes, I'll add this one here:

I suggest changing the description of the `--read-data-subset` option to:

~~~
read a `subset` of data packs, specified as 'n/t' for specific subset or either 'x%' or 'x.y%' for random subset"
~~~

doc/045_working_with_repos.rst
```diff
@@ -252,3 +257,13 @@ check all repository pack files over 5 separate invocations:
     $ restic -r /srv/restic-repo check --read-data-subset=3/5
     $ restic -r /srv/restic-repo check --read-data-subset=4/5
     $ restic -r /srv/restic-repo check --read-data-subset=5/5
+
+Use ``--read-data-subset=n%`` to check a randomly choosen subset of the
+repository pack files. It takes one parameter, ``n``, the percentage of pack
+files to check as a floating point number. This will not guarantee to cover all
```
rawtaz, line 263: Should probably write "as an integer or floating point number" to make it clear that both x% and x.y% are accepted.

doc/045_working_with_repos.rst
```diff
@@ -252,3 +257,13 @@ check all repository pack files over 5 separate invocations:
...
+available pack files after sufficient runs, but it is easy to automate checking
+a small subset of data after each backup:
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo check --read-data-subset=2.5%
```
rawtaz, line 269: Should probably add an example of integer-only percentage as well, just for clarity.

fgma:
> 
> 
> @fgma How do this PR relate to #2801?

I'm sorry, I did not see #2801 before.

The only remark I have considering #2801 is that it is quite hard for the user to decide about the number of packs to check compared to a percentage which relates to the size of the repository and will automatically scale with bigger repositories. Easily estimating the amount of bandwidth necessary for checks is somewhat critical for me because I'm on a metered connection for  my backup hosts.

rawtaz:
I think we should use this percentage one instead. I completely agree with @fgma in that no normal user will be able to determine what a good value for the proposed option in #2801 is, while a percentage is something everyone can relate to. Also, we *really* don't need more flags, it's better to reuse the one we have. And finally, if we truly want to add the feature that's in #2801, we can add it to the existing option instead of adding a new one.

cmd/restic/cmd_check.go
```diff
@@ -98,6 +112,23 @@ func stringToIntSlice(param string) (split []uint, err error) {
 	return result, nil
 }
 
+// parsePercentage parses the given percentage value to a float32
+func parsePercentage(param string) (float64, error) {
```
greatroar, line 116: This comment mainly repeats the function name and the result type (which it gets wrong). It's also more complicated than it could be with the regex. How about:

```
// ParsePercentage parses a percentage string of the form "X%" where X is a float constant,
// and returns the value of that constant. It does not check the range of the value.
func parsePercentage(s string) (float64, error) {
	if !strings.HasSuffix(s, "%") {
		return 0, errors.Errorf(`parsePercentage: %q does not end in "%%"`, s)
	}
	s = s[:len(s)-1]

	p, err := strconv.ParseFloat(s, 64)
	if err != nil {
		return 0, errors.Errorf("parsePercentage: %v", err)
	}
	return p, nil
}
```

This passes all the tests except the one for -100%, but that's already covered above.

cmd/restic/cmd_check.go
```diff
@@ -233,27 +264,44 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
...
 		}
-		packCount := uint64(len(packs))
+		return packs
+	}
 
-		if packCount < chkr.CountPacks() {
-			Verbosef(fmt.Sprintf("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, packCount, chkr.CountPacks(), totalBuckets))
-		} else {
-			Verbosef("read all data\n")
+	doSelectPacksByPercentage := func(percentage float64) *restic.IDSet {
```
greatroar, line 279: If this were a top-level function, it could be tested independently and runCheck would be shorter.

cmd/restic/cmd_check.go
```diff
@@ -263,10 +311,28 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
...
+		dataSubset, err := stringToIntSlice(opts.ReadDataSubset)
+		if err == nil {
+			bucket := dataSubset[0]
+			totalBuckets := dataSubset[1]
+			packs = doSelectPacksByNumbers(bucket, totalBuckets)
+			packCount := uint64(len(*packs))
+			Verbosef("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, packCount, chkr.CountPacks(), totalBuckets)
+		} else {
+			percentage, err := parsePercentage(opts.ReadDataSubset)
+			if err == nil {
```
greatroar, line 327: This is the expected case, so it should be unindented. Just store the error in _, or panic if it is non-nil.

cmd/restic/cmd_check.go
greatroar: I think this should be either

```
packsToCheck := int(float64(packCount)*(percentage/100))
```

(current behavior), or

```
packsToCheck := int(math.Round(float64(packCount)*(percentage/100)))
```

I'm not sure why you're rounding down. The math.Max shouldn't do anything, if the rest of the code is correct. It can only obscure bugs elsewhere.

changelog/unreleased/pull-3038
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

greatroar:
Also, the sampling code here is much simpler than the one in #2801.

cmd/restic/cmd_check.go
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

I was thinking about array indices which is wrong in this case.

doc/045_working_with_repos.rst
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

doc/045_working_with_repos.rst
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

cmd/restic/cmd_check.go
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

cmd/restic/cmd_check.go
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/4943b1ebda8ad41c5cd485e202336e780a1e8c77

cmd/restic/cmd_check.go
fgma: I'm working on this. Just lacking the time to finish it right now.

cmd/restic/cmd_check.go
greatroar: But packsToCheck must be <= the number of packs, not <, because rand.Perm(n) generates integers < n. Without math.Round, if you ask for 99.99% of 2 packs, only 50% gets checked.

cmd/restic/cmd_check.go
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/249e776e37f9d63d3d18683722ad97008de81122

aawsome:
I don't remember the statistical test exactly, but I think that one may want to test a fixed number of packs to get a statistical result about "repository sanity". (If one wants a definitive answer, we have always have to check all packs).

About the sampling code: The only intention I had in #2801 was to save some memory by *not* using the slice produced `rand.Perm`. But its length is the number of packs, so it might not matter too much.

As I'm a bit short on time ATM and want to focus on the pruning topics, I'll close #2801. You may want to check if this closes #2186.

cmd/restic/cmd_check.go
greatroar: LGTM, feel free to mark as resolved (I can't do that, apparently).

greatroar:
Re: statistics, isn't the rule of thumb for sample sizes that they should be proportional to sqrt N?

(PS.: No offense meant, I didn't mean to disparage your efforts.)

aawsome:
> Re: statistics, isn't the rule of thumb for sample sizes that they should be proportional to sqrt N?

You are right, that most statistical tests are with sample sizes proportional to sqrt. I simply didn't remember the right test, and all I wanted to say is that maybe a random subset which size is a fixed percentage of the sample size might not be a wise choice :wink:

fgma:
I think this PR implements what @cfbao requested. I've added a reference to the issue.

fgma:
For me this PR is ready to merge.

cmd/restic/cmd_check_test.go
```diff
@@ -0,0 +1,124 @@
...
+		id := restic.NewRandomID()
+		// ensure relevant part of generated id is reproducable
+		id[0] = byte(i)
+		testPacks.Insert(id)
+	}
+
+	selectedPacks := selectPacksByBucket(testPacks, 0, 10)
+	rtest.Assert(t, len(selectedPacks.List()) == 0, "Expected 0 selected packs")
+
+	selectedPacks = selectPacksByBucket(testPacks, 1, 5)
```
MichaelEischer, line 82: Please add a loop here.

MichaelEischer (commented):
Thanks for the really clean implementation. I've notice a few "cosmetic" issues and one more important point: The random subset should always contain at least one pack file. Otherwise for a small repository with a small percentage, nothing might ever be checked.

After merging #3082 this PR unfortunately has a merge conflict. See https://github.com/restic/restic/pull/3082/files#diff-efd94e38c190ad55029b381d903f2758124d8ccec01a8f3892d439afd835eec3 for the relevant change. The selected packs must now be a `map[restic.ID]int64` instead of an `restic.IDSet`.

changelog/unreleased/pull-3038
```diff
@@ -0,0 +1,7 @@
+Enhancement: Allow specifying percentage in check's --read-data-subset
+
+We've enhanced the command-line option --read-data-subset to also accept a
+percentage. This will check the given percentage of pack files which are
+randomly selected on each run.
+
+https://github.com/restic/restic/pull/3038
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check.go
```diff
@@ -64,14 +64,26 @@ func checkFlags(opts CheckOptions) error {
...
 		dataSubset, err := stringToIntSlice(opts.ReadDataSubset)
-		if err != nil || len(dataSubset) != 2 {
-			return errors.Fatalf("check flag --read-data-subset must have two positive integer values, e.g. --read-data-subset=1/2")
-		}
-		if dataSubset[0] == 0 || dataSubset[1] == 0 || dataSubset[0] > dataSubset[1] {
-			return errors.Fatalf("check flag --read-data-subset=n/t values must be positive integers, and n <= t, e.g. --read-data-subset=1/2")
-		}
-		if dataSubset[1] > totalBucketsMax {
-			return errors.Fatalf("check flag --read-data-subset=n/t t must be at most %d", totalBucketsMax)
+		argumentError := errors.Fatalf("check flag --read-data-subset must have two positive integer values or a percentage, e.g. --read-data-subset=1/2 or --read-data-subset=2.5%%")
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check.go
```diff
@@ -64,14 +64,26 @@ func checkFlags(opts CheckOptions) error {
...
+			if dataSubset[1] > totalBucketsMax {
+				return errors.Fatalf("check flag --read-data-subset=n/t t must be at most %d", totalBucketsMax)
+			}
+		} else {
+			percentage, err := parsePercentage(opts.ReadDataSubset)
+			if err != nil {
+				return argumentError
+			}
+
+			if percentage < 0.0 || percentage > 100.0 {
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check.go
```diff
@@ -277,3 +306,36 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
...
+			packs.Insert(pack)
+		}
+	}
+	return packs
+}
+
+// selectRandomPacksByPercentage selects the given percentage of packs which are randomly choosen.
+func selectRandomPacksByPercentage(allPacks restic.IDSet, percentage float64) *restic.IDSet {
+	packCount := len(allPacks)
+	packsToCheck := int(float64(packCount) * (percentage / 100.0))
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check.go
```diff
@@ -277,3 +306,36 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
 
 	return nil
 }
+
+// selectPacksByBucket selects subsets of packs by ranges of buckets.
+func selectPacksByBucket(allPacks restic.IDSet, bucket, totalBuckets uint) *restic.IDSet {
+	packs := &restic.IDSet{}
```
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check_test.go
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

cmd/restic/cmd_check.go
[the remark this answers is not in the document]
fgma: Fixed in https://github.com/restic/restic/pull/3038/commits/a3e5c707165252ab446770dd15077b49f446537d

MichaelEischer (approved):
LGTM. I've rebased and squashed the commits, and cleaned up to more cosmetic issues.

PR #2274, opened 2019-05-13
Support for Volume Shadow Copy Service (VSS) on windows

fgma:
What is the purpose of this change? What does it change?
--------------------------------------------------------

Add transparent support for Volume Shadow Copy Service (VSS) on windows.

Use of VSS needs to be activated via the new flag `--use-windows-vss`  for the backup command e.g.

`restic backup --verbose --use-windows-vss data`

To easily test the new feature I prepared some bat/powershell files that will run restic with the new flag and also get an exclusive lock on a file via the included powershell script to test VSS allows reading the file:
[vss_test.zip](https://github.com/restic/restic/files/3174318/vss_test.zip)

Right now it is not a finished pull request. It is missing many details:

- proper logging
- code cleanup
- testing

But basic VSS functionality should work! 

Right now I've tested it only on Windows 10 Professional 64 bit.

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------

closes #340

Checklist
---------

- [x] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [ ] I have added tests for all changes in this PR
- [x] I have added documentation for the changes (in the manual)
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))
- [x] I have run `gofmt` on the code in all commits
- [ ] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits)
- [x] I'm done, this Pull Request is ready for review

PR #3045, opened 2020-10-29
vss: fix potential crash (not reachable in restic)

fgma:

This fixes a potential crash in the VSS code. This is currently not reachable because restic checks for sufficient permissions before calling this code. If used somewhere this ends up dereferencing nil.

Mentioned in https://github.com/restic/restic/pull/2274 by @tigerwill90

Checklist
---------

### DIFF
--- a/internal/fs/vss_windows.go
+++ b/internal/fs/vss_windows.go
@@ -697,9 +697,20 @@ func initializeVssCOMInterface() (*ole.IUnknown, uintptr, error) {
 
 	var oleIUnknown *ole.IUnknown
 	result, _, _ := vssInstance.Call(uintptr(unsafe.Pointer(&oleIUnknown)))
+	hresult := HRESULT(result)
+
+	switch hresult {
+	case S_OK:
+	case E_ACCESSDENIED:
+		return oleIUnknown, result, newVssError(
+			"The caller does not have sufficient backup privileges or is not an administrator",
+			hresult)
+	default:
+		return oleIUnknown, result, newVssError("Failed to create VSS instance", hresult)
+	}
 
 	if oleIUnknown == nil {
-		return nil, result, newVssError("Failed to initialize COM interface", HRESULT(result))
+		return nil, result, newVssError("Failed to initialize COM interface", hresult)
 	}
 
 	return oleIUnknown, result, nil
@@ -716,7 +727,7 @@ func HasSufficientPrivilegesForVSS() bool {
 		return false
 	}
 
-	return !(HRESULT(result) == E_ACCESSDENIED)
+	return HRESULT(result) != E_ACCESSDENIED
 }
 
 // NewVssSnapshot creates a new vss snapshot. If creating the snapshots doesn't
@@ -738,23 +749,12 @@ func NewVssSnapshot(
 
 	timeoutInMillis := uint32(timeoutInSeconds * 1000)
 
-	oleIUnknown, result, err := initializeVssCOMInterface()
+	oleIUnknown, _, err := initializeVssCOMInterface()
 	if oleIUnknown != nil {
 		defer oleIUnknown.Release()
 	}
 	if err != nil {
 		return VssSnapshot{}, err
-	}
-
-	switch HRESULT(result) {
-	case S_OK:
-	case E_ACCESSDENIED:
-		return VssSnapshot{}, newVssTextError(fmt.Sprintf("%s (%#x) The caller does not have "+
-			"sufficient backup privileges or is not an administrator.", HRESULT(result).Str(),
-			result))
-	default:
-		return VssSnapshot{}, newVssTextError(fmt.Sprintf("Failed to create VSS instance: %s (%#x)",
-			HRESULT(result).Str(), result))
 	}
 
 	comInterface, err := queryInterface(oleIUnknown, UUID_IVSS)
--- a/internal/fs/vss_windows.go
+++ b/internal/fs/vss_windows.go
@@ -686,10 +686,10 @@ func (p *VssSnapshot) GetSnapshotDeviceObject() string {
 }
 
 // initializeCOMInterface initialize an instance of the VSS COM api
-func initializeVssCOMInterface() (*ole.IUnknown, uintptr, error) {
+func initializeVssCOMInterface() (*ole.IUnknown, error) {
 	vssInstance, err := loadIVssBackupComponentsConstructor()
 	if err != nil {
-		return nil, 0, err
+		return nil, err
 	}
 
 	// ensure COM is initialized before use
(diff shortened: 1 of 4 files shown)


rawtaz (approved):

internal/fs/vss_windows.go
```diff
@@ -741,7 +741,9 @@ func NewVssSnapshot(
 		}
 		return VssSnapshot{}, err
 	}
-	defer oleIUnknown.Release()
+	if oleIUnknown != nil {
```
MichaelEischer, line 744: Please move this if block before the `if err != nil` block and remove the then redundant call to `Release` inside the latter if block.

greatroar:
Wouldn't it be cleaner to return an error from initializeVssCOMInterface when it gets a nil? Just asking, I've no idea what the effect on HasSufficientPrivilegesForVSS would be.

tigerwill90:
> Wouldn't it be cleaner to return an error from initializeVssCOMInterface when it gets a nil? Just asking, I've no idea what the effect on HasSufficientPrivilegesForVSS would be.

Yes. Actually the HRESULT code is correct and fall back to `E_ACCESSDENIED` case before crashing. It seems possible to treat all result code and error case in `initializeVssCOMInterface()` and returning only `oleIUnknown` or an error.

greatroar:
I'm not sure what you mean. Is there a case where oleIUnknown is nil but still usable? I.e., is the following incorrect?

```go
// initializeCOMInterface initialize an instance of the VSS COM api
func initializeVssCOMInterface() (*ole.IUnknown, uintptr, error) {
	vssInstance, err := loadIVssBackupComponentsConstructor()
	if err != nil {
		return nil, 0, err
	}

	// ensure COM is initialized before use
	ole.CoInitializeEx(0, ole.COINIT_MULTITHREADED)

	var oleIUnknown *ole.IUnknown
	result, _, _ := vssInstance.Call(uintptr(unsafe.Pointer(&oleIUnknown)))
	if oleIUnknown == nil {
		return nil, 0, errors.Errorf("VSS: got nil OLE object") // or whatever
	}

	return oleIUnknown, result, nil
}
```

DRON-666:
Ideally `HRESULT` should be returned only as part of `vssError` (functions like `Cancel`, `Wait`, `QueryStatus` must be changed) and all function such as `DoSnapshotSet`, `GatherWriterMetadata`, `PrepareForBackup`, `BackupComplete`, `initializeVssCOMInterface`, `queryInterface` should call `Release` itself and `return nil, err`. Also situation where returned COMinterface!=nil and err==nil or COMinterface==nil and err!=nil should be impossible.

greatroar:
@DRON-666 PR?

DRON-666:
Yes @greatroar, I should write PR, but problem is that my developer skills predates git, github and golang. I want to contribute to this project, but definitely not today.

internal/fs/vss_windows.go
fgma: Fixed in https://github.com/restic/restic/pull/3045/commits/767a686f4032f25d06f81f165bda3f6fee80329c

fgma:
I've cleaned it up. It now handles errors in initializeVssCOMInterface() properly. I've tested it by removing the call to HasSufficientPrivilegesForVSS(). This leads to a to expected error:

```
VSS error: Failed to initialize COM interface: E_ACCESSDENIED (0x80070005)
```

internal/fs/vss_windows.go
greatroar: Now that initializeVssCOMInterface check for nil, this function no longer needs to.

```
oleIUnknown, result, err := initializeVssCOMInterface()
if err != nil {
	// handle error
}
defer oleIUnknown.Release()
```

HasSufficientPrivilegesForVSS can do the same without the defer.

DRON-666:
>I've cleaned it up. 

After that `switch HRESULT(result)` cases become mostly unreachable.

fgma:
@DRON-666 To make it clean we should also move the error handling in the switch statement into initializeVssCOMInterface(). I'd like to keep the special error handling telling the user about necessary privileges or being an administrator for easy troubleshooting.

fgma:
I've move error handling to initializeVssCOMInterface(). It will still return a the result because it is needed in HasSufficientPrivilegesForVSS().

internal/fs/vss_windows.go
```diff
@@ -716,7 +727,7 @@ func HasSufficientPrivilegesForVSS() bool {
 		return false
 	}
 
-	return !(HRESULT(result) == E_ACCESSDENIED)
+	return HRESULT(result) != E_ACCESSDENIED
```
DRON-666, line 730: This code unreachable if `result` == `E_ACCESSDENIED` because `err` always != `nil` if `hresult` != `S_OK`.
`initializeVssCOMInterface` should not return `HRESULT`. If you still need it in other places you may use something like this:
```
if e, ok := err.(*vssError); ok {
    // Use e.hresult
}
```
And if you want your custom message from `initializeVssCOMInterface` to be seen by user, `HasSufficientPrivilegesForVSS` should return `error`, but not `bool`.

internal/fs/vss_windows.go
```diff
@@ -697,9 +697,20 @@ func initializeVssCOMInterface() (*ole.IUnknown, uintptr, error) {
 
 	var oleIUnknown *ole.IUnknown
 	result, _, _ := vssInstance.Call(uintptr(unsafe.Pointer(&oleIUnknown)))
+	hresult := HRESULT(result)
```
DRON-666, line 700: Please add here
```
if hresult != S_OK &&  oleIUnknown != nil {
    oleIUnknown.Release()
}
```
Callers don't need to release in case of errors.

internal/fs/vss_windows.go
```diff
@@ -738,25 +749,14 @@ func NewVssSnapshot(
 
 	timeoutInMillis := uint32(timeoutInSeconds * 1000)
 
-	oleIUnknown, result, err := initializeVssCOMInterface()
+	oleIUnknown, _, err := initializeVssCOMInterface()
 	if oleIUnknown != nil {
 		defer oleIUnknown.Release()
 	}
```
DRON-666, line 755: And now you can call `defer oleIUnknown.Release()` unconditionally after next `if err != nil` statement.

internal/fs/vss_windows.go
fgma: I'd like to keep it like this because they need co release on success anyway. Adding a special case will just make the code more complex.

rawtaz:
Is there much left to do on this or is it ready for merging soon?

fgma:
I'm right now thinking about @DRON-666's remarks.

internal/fs/vss_windows.go
fgma: Fixed in https://github.com/restic/restic/pull/3045/commits/947f00e9ec9237ae07d04072721d2c67fce7adf5

internal/fs/vss_windows.go
fgma: I prefer to keep it conditional here so it gets cleaned up even if I messed up somewhere else.

fd0:
Let us know when you're done :)

fgma:
Just waiting for @DRON-666 approve my changes.

internal/fs/vss_windows.go
greatroar: The current setup makes the code complex. Idiomatic Go is, e.g.,

```go
f, err := os.Open(filename)
if err != nil {
	return err
}
defer f.Close()
```

with the understanding that the caller should not use f when err == nil, but isn't responsible for disposing of it either. os.Open is responsible for making that possible.

fgma:
I also thought about it but I can't easily get around it. Its essentially an implementation detail. In case multiple other things might fail before permissions are checked the API might return a different error. All other possible errors won't give any information about permissions. Actually checking for correct permissions is a bit more involved on windows. I implemented this in https://github.com/restic/restic/pull/2875/ which is still open for discussion.

MichaelEischer:
> I also thought about it but I can't easily get around it. Its essentially an implementation detail. In case multiple other things might fail before permissions are checked the API might return a different error. All other possible errors won't give any information about permissions. Actually checking for correct permissions is a bit more involved on windows. I implemented this in #2875 which is still open for discussion.

The cleanest fix would be to let `HasSufficientPrivilegesForVSS` return true/false/"something else went wrong". That is return `(bool, err)`. That way the initial permissions check could already return allowed/denied/some other error to the user.

greatroar:
The only call site turns a false result into a fatal error anyway, so even just returning the error would be enough :)

internal/fs/vss_windows.go
DRON-666: I totally agree with @greatroar here and it's common for other languages too: errors and values mutually exclusive. In languages with exception support you simple can't get value if exception occurred. Go goes the same way, but you have to do everything yourself.

DRON-666:
>If err != nil but of a different type, the function returns true. That doesn't seem right.

Yes and code from cmd_backup.go should be
```
if err := fs.HasSufficientPrivilegesForVSS(); err != nil {
    return err
}
```
Now it's convert `HRESULT` to `error`, then to `bool` and then to different `error`.

fgma:
HasSufficientPrivilegesForVSS() now returns an only an error

rawtaz:
@DRON-666 and @greatroar Does your thumbs-up clickety-clicks mean that you have no further comments on this PR, that you think it's fine to merge?

MichaelEischer (approved):
LGTM.

PR #3048, opened 2020-11-01
check: check index for packs that are read

aawsome:

Adds a check for `restic check --read-data*` that the blobs in the read pack are also contained in the index.

Discussed this when talking about #3006; this PR is however completely independent. 

Checklist
---------

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -727,6 +727,7 @@ func checkPack(ctx context.Context, r restic.Repository, id restic.ID) error {
 
 	var errs []error
 	var buf []byte
+	idx := r.Index()
 	for i, blob := range blobs {
 		debug.Log("  check blob %d: %v", i, blob)
 
@@ -760,6 +761,19 @@ func checkPack(ctx context.Context, r restic.Repository, id restic.ID) error {
 		if !hash.Equal(blob.ID) {
 			debug.Log("  Blob ID does not match, want %v, got %v", blob.ID, hash)
 			errs = append(errs, errors.Errorf("Blob ID does not match, want %v, got %v", blob.ID.Str(), hash.Str()))
+			continue
+		}
+
+		// Check if blob is contained in index
+		idxHas := false
+		for _, pb := range idx.Lookup(blob.ID, blob.Type) {
+			if pb.PackID == id {
+				idxHas = true
+				break
+			}
+		}
+		if !idxHas {
+			errs = append(errs, errors.Errorf("Blob ID %v is not contained in index", blob.ID.Str()))
 			continue
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/checker/checker.go
```diff
@@ -762,6 +763,18 @@ func checkPack(ctx context.Context, r restic.Repository, id restic.ID) error {
 			errs = append(errs, errors.Errorf("Blob ID does not match, want %v, got %v", blob.ID.Str(), hash.Str()))
 			continue
 		}
+
+		// Check if blob is contained in index
+		idxHas := false
+		for _, pb := range idx.Lookup(blob.ID, blob.Type) {
+			if pb.PackID == id {
+				idxHas = true
```
greatroar, line 771: Break statement here? Otherwise, LGTM.

aawsome: is added.

aawsome:
Also added a changelog entry as this error may be new to users.

fd0 (approved):
Nice, great idea!

MichaelEischer:
It would actually possible to check a few additional things here: Is the blob in the right place? Does it's length match the index? I also wonder whether there's an efficient way to check whether the index contains entries which do not exist in a pack file. The first two cases could occur with simple bitflips in memory. For the last case I'm struggling a bit to come up with a scenario which isn't already detected the the check that all blobs are indexed and does not involve active manipulations of the repository index or a serious bug in restic.

aawsome:
> It would actually possible to check a few additional things here: Is the blob in the right place? Does it's length match the index?

Thanks for the hint! I can add this to this test in another PR. This test will also verify that there are no duplicate blobs within a pack file which is needed by my argumentation below :+1: 

> I also wonder whether there's an efficient way to check whether the index contains entries which do not exist in a pack file.

This is another check I was also thinking of but I wanted to postpone it until the `prune` changes and the new `rebuild-index` got seasoned.

My idea was the following: Similar to the above changes, we can compute the size of each pack completely from the data in the index (= header length + length of all blobs + header entry length * number of blobs) and additional to the already-present check if the pack file exists also check whether the length given by 'List' fits this computed pack size.
Maybe also check if this equals the actual read pack size after reading the pack.

Now assume, we have an index entry which does not exist in a pack file.
- If the index entry refers to a pack that is not existing the already implemented check will throw an error, so we can assume this entry refers to an existing pack.
- Moreover the check of this PR ensures (when reading that pack) that all blobs in this pack are contained in the index.
- Moreover, we can assume that there are no duplicate blobs within this pack.
- Hence, all blobs that actually exist in the pack exist in the index and are counted. As the non-existing index entry is also counted, the computed pack size would be larger than the actual pack size which would be detected by that test. :wink:

PR #3050, opened 2020-11-02
Fix string truncation in ui/termstatus

### LINKED ISSUE #3046: Restic outputs invalid UTF-8 when printing long UTF-8 file names to stdout when the terminal is small
Output of `restic version`
--------------------------
restic 0.10.0 compiled with go1.15.3 on linux/amd64

Local/SFTP doesn't matter it's a frontend issue

Expected behavior
-----------------
filenames are truncated without cutting utf8 characters in half

Actual behavior
---------------
filenames are truncated and displayed with a replacement character at the end as if a utf8 character was cut in half

The following shell script, if run in a shell that is very narrow (for example termux on android), will output the partial filename as intended, but depending on width will output the file name of two of these directories will also show an �, indicating that invalid utf8 was printed out

```sh
export RESTIC_PASSWORD=1
mkdir 'ああああああああああああああああああああああああああああああああああ'
truncate -s 2G ああああああああああああああああああああああああああああああああああ/data
cp -r ああああああああああああああああああああああああああああああああああ aああああああああああああああああああああああああああああああああああ
cp -r ああああああああああああああああああああああああああああああああああ aaああああああああああああああああああああああああああああああああああ
rm -rf repo
restic -r repo init
restic -r repo backup *ああああああああああああああああああああああああああああああああああ
```

something about filenames being handled as byte strings instead of regular strings (like they should except ideally not in the frontend)?

try to decode the string as utf8 before trimming it

restic imo just works :tm: and has a binary release on basically every platform i actively use. Almost exactly maps to how i want to do backups, especially since #2606 was merged. Am using it to manage around 500GB of data on two repositories (one on a local external hard disk and one on backblaze, uploaded regularly) and it already has come in handy both after mistakes and reinstalling the OS

greatroar:

It handles Unicode properly when truncating strings in ui/termstatus, and tries to detect the display width of characters. It does so in a conservative, locale-neutral way.

The obvious alternative to this implementation is to use the Truncate function from github.com/mattn/go-runewidth. I decided against that package when I noticed it uses behaves differently depending on the locale and on an undocumented environment variable.

The upgrade to golang.org/x/text isn't strictly necessary, but the new version does add Unicode 13 tables for the width package that will be picked up by Go 1.16 when that's released.

Fixes #3046.

Checklist
---------

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,7 +41,7 @@ require (
 	golang.org/x/oauth2 v0.0.0-20200902213428-5d25da1a8d43
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	golang.org/x/sys v0.0.0-20201007082116-8445cc04cbdf
-	golang.org/x/text v0.3.3
+	golang.org/x/text v0.3.4
 	google.golang.org/api v0.32.0
 	gopkg.in/check.v1 v1.0.0-20200902074654-038fdea0a05b // indirect
 	gopkg.in/ini.v1 v1.61.0 // indirect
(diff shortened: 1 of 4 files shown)


fd0 (approved):
Awesome, thanks!

PR #3058, opened 2020-11-04
Replace restic.Progress with new progress.Counter (fixes two race conditions)

greatroar:

This PR replaces restic.Progress with a new type, ui/progress.Counter. It fixes several issues:

* Progress.Done races with Progress.reporter for access to Progress.fnM. As a result, a progress report can be printed after the Progress has been canceled.
* Progress.Done does not wait for termination, so it races with other routines for stdout, or the final message may not be printed before the process exits.
* Initialization is spread out over the constructor and various init functions, that are called even for processes that do not use a Progress.
* Progress is defined in internal/restic, overloading the meaning of that package.
* Callbacks may be called either from a separate goroutine or from Progress.Report's caller's goroutine. This was probably meant as an optimization, but it means that the caller may have to wait on stdout.
* Progress is designed to handle restic.Stat objects, which carry counts of various object types. Its callers (ab)use only its blobs count for various purposes, most of which are not counting blobs.
* The code has grown organically over time, making it complicated and leaving dead code in various places.

Instead of fixing these issues one by one, I've decided to write a replacement from scratch. It's >100 lines shorter, leaving ample space for a unit test that covers 92% of its statements.

No.

Checklist
---------


internal/index/index.go
```diff
@@ -187,10 +188,9 @@ func loadIndexJSON(ctx context.Context, repo ListLoader, id restic.ID) (*indexJS
 }
 
 // Load creates an index by loading all index files from the repo.
-func Load(ctx context.Context, repo ListLoader, p *restic.Progress) (*Index, error) {
+func Load(ctx context.Context, repo ListLoader, p *progress.Counter) (*Index, error) {
 	debug.Log("loading indexes")
 
-	p.Start()
 	defer p.Done()
```
MichaelEischer, line 194: As the progress bar is now started outside`Load` I'd also like to move `p.Done()` to the calling site. The way the rule for stopping a progress bar stays the same as before: Whoever calls `Start()` (now done implicitly) also has to call `Done()`.

internal/repository/repack.go
```diff
@@ -22,11 +24,8 @@ const numRepackWorkers = 8
 //
 // The map keepBlobs is modified by Repack, it is used to keep track of which
 // blobs have been processed.
-func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, keepBlobs restic.BlobSet, p *restic.Progress) (obsoletePacks restic.IDSet, err error) {
-	if p != nil {
-		p.Start()
-		defer p.Done()
-	}
+func Repack(ctx context.Context, repo restic.Repository, packs restic.IDSet, keepBlobs restic.BlobSet, p *progress.Counter) (obsoletePacks restic.IDSet, err error) {
+	defer p.Done()
```
MichaelEischer, line 28: Please move this call to `Done()` to the respective caller(s).

MichaelEischer (changes requested):
Really nice cleanup! I've just a small remark:
I'd like to keep the progress bar creation and the calls to `Done()` together on the same level to consistently let the progress bar starter stop it. Besides that the code changes are fine.

greatroar:
Moved Done() to the call sites of newProgressMax. Index.Load did not have a call site outside of tests. Also removed the Counter argument from checker.ReadData, since that was always called with a nil Counter.

greatroar:
Btw., is it ok if I change the default FPS from 60 to, say, 30, or even 20? It's not like we're trying to create a fluent animation here.

MichaelEischer (approved):
LGTM. Thanks for cleaning this up. I'd say we just stick with the 60 fps for now. After all it shouldn't cost too much performance.

PR #3059, opened 2020-11-05
Remove 0.11.0 items from changelog/unreleased

greatroar:
These files were copied instead of moved. It looks like the prepare-release change in 79a50e3b1f9b5672388107eb5826229f3b7c367b didn't work entirely as intended. Maybe use git mv instead of os.Rename+git add+git add -u?


greatroar:
Fixed by 4e3ad8b3b1a92064ec173a68080aeaa0cf14b865.

fd0:
Hah, you're fast! I'm feeling watched 😁

fd0:
Fixed for the future in https://github.com/restic/restic/commit/8a0dbe7c1ae0255d6b9a4a864f0d6f59853600e0

PR #3065, opened 2020-11-06
Don't recurse in local backend's List if not required

greatroar:

The local backend's List method would recursively read subdirectories even for layouts that do not use subdirectories. Due to the return if !isFile, the IsDir branch was never taken.

Also replaced isFile by an IsRegular check, which has been equivalent since Go 1.12 (golang/go@a2a3dd00c934fa15ad880ee5fe1f64308cbc73a7).

No.

Checklist
---------

### DIFF
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -220,32 +220,27 @@ func (b *Local) Remove(ctx context.Context, h restic.Handle) error {
 	return fs.Remove(fn)
 }
 
-func isFile(fi os.FileInfo) bool {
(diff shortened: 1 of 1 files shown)


MichaelEischer (commented):
I've noticed that restic picks up files in the root directory even if the layout for that filetype uses subdirectories. Maybe we should change `List()` to skip these files?
For example move `data/ab/abcdef...` to `data/abcdef...`  and watch the error message.

greatroar:
I've replaced the fs.Walk with custom directory traversal that breaks when there are files or directories in the wrong place.

MichaelEischer (approved):
LGTM. Thanks!

PR #3075, opened 2020-11-08
UI: Add several hints

fd0:

It adds two messages:
 * During prune, warn if no cache is used, because this will be very slow
 * During backup, always print whether or not a parent snapshot is used, so we can see what's going on when people report restic as being slow...

Nope

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -532,8 +532,12 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		return err
 	}
 
-	if !gopts.JSON && parentSnapshotID != nil {
-		p.V("using parent snapshot %v\n", parentSnapshotID.Str())
+	if !gopts.JSON {
+		if parentSnapshotID != nil {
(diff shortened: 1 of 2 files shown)


cmd/restic/cmd_backup.go
```diff
@@ -532,8 +532,12 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		return err
 	}
 
-	if !gopts.JSON && parentSnapshotID != nil {
-		p.V("using parent snapshot %v\n", parentSnapshotID.Str())
+	if !gopts.JSON {
+		if parentSnapshotID != nil {
+			p.P("using parent snapshot %v\n", parentSnapshotID.Str())
+		} else {
+			p.P("no parent snapshot, need to read all data\n")
```
rawtaz, line 539: I suggest "no parent snapshot found, will read all data\n".

rawtaz: Or maybe "I suggest "no parent snapshot found, will read all files\n" is more relatable to users.

rawtaz (approved):
Moo

PR #3085, opened 2020-11-11
Extended option to select V1 API for ListObjects on S3 backend

### LINKED ISSUE #3083: restic 0.11.0 breaks with S3 compatible backends that don't (fully) support ListObjectsV2
Output of `restic version`
--------------------------
Broken on: restic 0.11.0 compiled with go1.15.3 on linux/amd64
Works on: restic 0.10.0 compiled with go1.15.2 on linux/amd64 (and earlier)

Nightly backups are made to a S3 repository. When upgrading to the latest restic version, and trying to list snapshots, an error occurs.

```
# restic snapshots 
repository beb6e91b opened successfully, password is correct
List(snapshot) returned error, retrying after 462.318748ms: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 890.117305ms: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 636.341646ms: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 1.107876242s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 1.511079094s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 3.04096222s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 5.781953173s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 11.222385136s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 9.152900914s: Truncated response should have continuation token set
List(snapshot) returned error, retrying after 16.927710407s: Truncated response should have continuation token set
could not load snapshots: Truncated response should have continuation token set
```

This also happens during a restore command, which produces similar output.

S3 backend, backed by Ceph on an OpenStack platform.

Expected behavior
-----------------
Listing snapshots and restoring should work normally.

Actual behavior
---------------
Listing snapshots and restoring does not work.

See above.

This issue seems to be caused by Ceph's S3 compatible gateway. It advertises V2 API compatible endpoints, but doesn't implement them properly, specifically ListObjectsV2. Switching to the V1 API 'solves' this issue, in this case done by downgrading restic.

The issue is introduced in minio-go/v7 , see the bug report here: https://github.com/minio/minio-go/issues/1130 . There is also a bug report in the Thanos project, a workaround there is providing the option to use the V1 api: https://github.com/thanos-io/thanos/issues/3291 . There is also an issue tracking this by Ceph itself: https://tracker.ceph.com/issues/23908 .

An option to select the V1 API would help, and would allow a user to use restic + S3 + Ceph on versions of Ceph that don't implement the V2 APIs (properly). May also help other S3 implementations.

The 0.10 and 0.11 releases have a lot of nice features, and using restic is very easy.

LordGaav:

Provides an extended option to select the V1 ListObjects API for S3 backends that do not properly
implement the ListObjectsV2 API (versions of Ceph before v14.2.5) .

Closes #3083 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3083
+++ b/changelog/unreleased/issue-3083
@@ -1,0 +1,13 @@
+Enhancement: Allow usage of deprecated S3 ListObjectsV1 API
+
+Restic didn't work with some S3 API implementations that don't have a
+working ListObjectsV2 API endpoint. This concerns mainly Ceph versions
+before v14.2.5 , but may impact other S3 API implementations.
+
+Restic now allows selection of the older ListObjectsV1 endpoint by using
+the 's3.list-objects-v1' extended option, for instance:
+
+`restic -o s3.list-objects-v1=true snapshots`
+
+https://github.com/restic/restic/issues/3083
+https://github.com/restic/restic/pull/3085
--- a/changelog/unreleased/issue-3083
+++ b/changelog/unreleased/issue-3083
@@ -1,13 +1,19 @@
 Enhancement: Allow usage of deprecated S3 ListObjectsV1 API
 
 Restic didn't work with some S3 API implementations that don't have a
-working ListObjectsV2 API endpoint. This concerns mainly Ceph versions
-before v14.2.5 , but may impact other S3 API implementations.
+broken `ListObjectsV2` API endpoint. This concerns mainly Ceph versions
+before v14.2.5 , but may impact other S3 API implementations. When a broken
+server implementation is used, restic prints errors similar to the following:
 
-Restic now allows selection of the older ListObjectsV1 endpoint by using
-the 's3.list-objects-v1' extended option, for instance:
+    List() returned error: Truncated response should have continuation token set
 
-`restic -o s3.list-objects-v1=true snapshots`
+As a temporary workaround, restic now allows selection of the older
+`ListObjects` endpoint by using the `s3.list-objects-v1` extended option, for
+instance:
+
+    restic -o s3.list-objects-v1=true snapshots
+
+This option may be removed in future versions of restic.
 
 https://github.com/restic/restic/issues/3083
 https://github.com/restic/restic/pull/3085
--- a/changelog/unreleased/issue-3083
+++ b/changelog/unreleased/issue-3083
@@ -1,15 +1,14 @@
 Enhancement: Allow usage of deprecated S3 ListObjectsV1 API
 
-Restic didn't work with some S3 API implementations that don't have a
-broken `ListObjectsV2` API endpoint. This concerns mainly Ceph versions
-before v14.2.5 , but may impact other S3 API implementations. When a broken
-server implementation is used, restic prints errors similar to the following:
+Some S3 API implementations, e.g. Ceph before version 14.2.5, have a broken
+`ListObjectsV2` implementation which cause problems for restic when using their
+API endpoints. When a broken server implementation is used, restic prints
+errors similar to the following:
 
     List() returned error: Truncated response should have continuation token set
 
-As a temporary workaround, restic now allows selection of the older
-`ListObjects` endpoint by using the `s3.list-objects-v1` extended option, for
-instance:
+As a temporary workaround, restic now allows using the older `ListObjects`
+endpoint by setting the `s3.list-objects-v1` extended option, for instance:
 
     restic -o s3.list-objects-v1=true snapshots
 
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -339,6 +339,39 @@ For example with an actual endpoint:
 
     $ restic -o s3.bucket-lookup=dns -o s3.region=oss-eu-west-1 -r s3:https://oss-eu-west-1.aliyuncs.com/bucketname init
 
+Ceph
+****
+
+`Ceph Object Gateway <https://ceph.io/ceph-storage/object-storage/>`__ is an object
+storage interface built on top of librados to provide applications with a RESTful
+gateway to Ceph Storage Clusters.
+
+Ceph Rados Gateway (RGW) provides a S3 interface that is compatible with a large
+subset of the Amazon S3 RESTful API.
+
+You will need to setup the following environment variables with the proper credentials:
+
+.. code-block:: console
+
+    $ export AWS_ACCESS_KEY_ID=<YOUR-ACCESS-KEY-ID>
+    $ export AWS_SECRET_ACCESS_KEY=<YOUR-SECRET-ACCESS-KEY>
+
+Now you can easily initialize restic to use Ceph RGW as a backend with
+this command.
+
+.. code-block:: console
+
+    $ restic -r s3://<RGW-S3-ENDPOINT>/<RGW-S3-BUCKET> init
+    enter password for new backend:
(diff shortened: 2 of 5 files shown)


rawtaz:
I'm not sure that we should add yet another S3-compatible backend to the list of instructions in the documentation. It's one thing if what we add differs in how you configure it, but this Ceph backend looks like it's configured the same way as the regular S3 one. Maybe we should just keep the header and the note about the version, and skip the rest? It doesn't make sense to have redundant instructions just because services have different names, there has to be something to actually document in terms of instructions.

doc/030_preparing_a_new_repo.rst
```diff
@@ -339,6 +339,39 @@ For example with an actual endpoint:
...
+.. code-block:: console
+
+    $ restic -r s3://<RGW-S3-ENDPOINT>/<RGW-S3-BUCKET> init
+    enter password for new backend:
+    enter password again:
+    created restic backend xxxxxxxxxx at s3://<RGW-S3-ENDPOINT>/<RGW-S3-BUCKET>
+    Please note that knowledge of your password is required to access
+    the repository. Losing your password means that your data is irrecoverably lost.
+
+.. note:: Version of Ceph before v14.2.5 do not implement the ListObjectsV2 API properly.
```
fd0, line 371: I agree with @rawtaz, the code looks good to me, but the documentation is too much. Let's just keep this note and remove the rest. :)

fd0 (changes requested):

doc/030_preparing_a_new_repo.rst
LordGaav: It looks a bit strange to add just the note to the S3 section, because out of nowhere we're talking about Ceph. I'll see if I can make it look logical.

fd0: Great, thanks!

fd0 (approved):
Thanks for the PR! I've slightly amended the docs, now it's ready to be merged.

doc/030_preparing_a_new_repo.rst
```diff
@@ -239,6 +239,14 @@ For an S3-compatible server that is not Amazon (like Minio, see below),
 or is only available via HTTP, you can specify the URL to the server
 like this: ``s3:http://server:port/bucket_name``.
 
+
+.. note:: Certain S3-compatible servers do not properly implement the
+          ``ListObjectsV2`` API, most notably Ceph versions before v14.2.5. On these
+          backends, as a temporary workaround, you can provide the
+          ``-o s3.list-objects-v1=true`` option to use the older
+          ``ListObjects`` API instead.
```
rawtaz, line 247: I suggest adding "This option may be removed in future versions of restic." to the end.

changelog/unreleased/issue-3083
```diff
@@ -0,0 +1,17 @@
...
+before v14.2.5 , but may impact other S3 API implementations. When a broken
+server implementation is used, restic prints errors similar to the following:
+
+    List() returned error: Truncated response should have continuation token set
+
+As a temporary workaround, restic now allows selection of the older
+`ListObjects` endpoint by using the `s3.list-objects-v1` extended option, for
+instance:
+
+    restic -o s3.list-objects-v1=true snapshots
```
rawtaz, line 14: I suggest adding "This option may be removed in future versions of restic." with an empty line above and below it.

internal/backend/s3/config.go
```diff
@@ -20,16 +20,18 @@ type Config struct {
...
 
-	Connections  uint   `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
-	MaxRetries   uint   `option:"retries" help:"set the number of retries attempted"`
-	Region       string `option:"region" help:"set region"`
-	BucketLookup string `option:"bucket-lookup" help:"bucket lookup style: 'auto', 'dns', or 'path'."`
+	Connections   uint   `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+	MaxRetries    uint   `option:"retries" help:"set the number of retries attempted"`
+	Region        string `option:"region" help:"set region"`
+	BucketLookup  string `option:"bucket-lookup" help:"bucket lookup style: 'auto', 'dns', or 'path'."`
+	ListObjectsV1 bool   `option:"list-objects-v1" help:"use deprecated V1 api for ListObjects calls."`
```
rawtaz, line 27: Please remove the dots in this new and other option above it.

changelog/unreleased/issue-3083
```diff
@@ -0,0 +1,19 @@
+Enhancement: Allow usage of deprecated S3 ListObjectsV1 API
+
+Restic didn't work with some S3 API implementations that don't have a
+broken `ListObjectsV2` API endpoint. This concerns mainly Ceph versions
+before v14.2.5 , but may impact other S3 API implementations. When a broken
+server implementation is used, restic prints errors similar to the following:
```
rawtaz, line 6: Please change to the following to fix grammar and meaning errors:

~~~
Some S3 API implementations, e.g. Ceph before version 14.2.5, has a broken
`ListObjectsV2` implementation which cause problems for restic when using
their API endpoints. When a broken server implementation is used, restic prints
errors similar to the following:
~~~

changelog/unreleased/issue-3083
```diff
@@ -0,0 +1,19 @@
...
+
+    List() returned error: Truncated response should have continuation token set
+
+As a temporary workaround, restic now allows selection of the older
+`ListObjects` endpoint by using the `s3.list-objects-v1` extended option, for
+instance:
```
rawtaz, line 12:
~~~
As a temporary workaround, restic now allows using the older `ListObjects`
endpoint by setting the `s3.list-objects-v1` extended option, for instance:
~~~

rawtaz (approved):
I feel much better now.

LordGaav:
:ok_hand:

robbat2:
:wave: Ceph RGW S3 developer here. I'd like to specifically request that your implementation is backwards compatible and accepts a v1 ListObjects response in all cases. The S3 specification for ListObjectsV2 was specifically designed such that if the query-parameter argument to select a V2 response is not known, that it would issue a V1 response regardless.

As such, an ideal client should issue the v2 query for the initial request, and check the style of the response, if it's V2 response, continue to use it as V2; If it's V1 response, continue to issue V1 requests.
The arguments: prefix, max-keys, delimiter, encoding-type are common between them and portable.

The exception being if you wanted specific V2 parameters/fields: `start-after`, `fetch-owner`

fd0:
@robbat2 thanks for chiming in! We're using https://github.com/minio/minio-go for accessing S3 backends. I think that'd be something the library needs to do, right?

@harshavardhana what's your take on this?

harshavardhana:
> As such, an ideal client should issue the v2 query for the initial request, and check the style of the response, if it's V2 response, continue to use it as V2; If it's V1 response, continue to issue V1 requests.
> The arguments: prefix, max-keys, delimiter, encoding-type are common between them and portable.
> 
> The exception being if you wanted specific V2 parameters/fields: `start-after`, `fetch-owner`

AFAIK no AWS SDKs actually do such fallback - I am not sure why any SDK would do this. minio-go provides a way to chose which version of the API to use that is perhaps what `restic` did here.

PR #3101, opened 2020-11-16
Compute packsizes in MasterIndex

aawsome:

Cleans up code where pack (header) sizes are computed from index entries.
Now all code which takes assumptions on the pack format is either in `internal/pack` or in `MasterIndex.PackSize()`. This makes modification to the pack format more easy in the future.

This PR introduces a small change in the new pruning algorithm (see #2718): Now the complete header is treated as "used space" within a pack, whereas before only the header length and header crypto were treated as "used space" + only the header entries of really used blobs. The difference however should be marginal and may only slightly change the decision which pack to repack and which to keep.

see #3006 

Checklist
---------

- There's no new file in `changelog/unreleased/` that describes the changes for our users - this is too technical for users to notice.

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -740,16 +740,11 @@ func (r *Repository) List(ctx context.Context, t restic.FileType, fn func(restic
 }
 
 // ListPack returns the list of blobs saved in the pack id and the length of
-// the file as stored in the backend.
-func (r *Repository) ListPack(ctx context.Context, id restic.ID, size int64) ([]restic.Blob, int64, error) {
+// the the pack header.
+func (r *Repository) ListPack(ctx context.Context, id restic.ID, size int64) ([]restic.Blob, uint32, error) {
 	h := restic.Handle{Type: restic.PackFile, Name: id.String()}
 
-	blobs, err := pack.List(r.Key(), restic.ReaderAt(ctx, r.Backend(), h), size)
-	if err != nil {
-		return nil, 0, err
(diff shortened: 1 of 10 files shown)


internal/repository/repository.go
```diff
@@ -741,15 +741,10 @@ func (r *Repository) List(ctx context.Context, t restic.FileType, fn func(restic
 
 // ListPack returns the list of blobs saved in the pack id and the length of
 // the file as stored in the backend.
-func (r *Repository) ListPack(ctx context.Context, id restic.ID, size int64) ([]restic.Blob, int64, error) {
+func (r *Repository) ListPack(ctx context.Context, id restic.ID, size int64) ([]restic.Blob, uint32, error) {
```
fd0, line 744: What's going on here? Before, `ListPack()` returned its parameter `size` (which is a bit pointless), afterwards the size of the header is returned? If that's correct, we should amend the docstring here and in the `restic.Repository` interface.

fd0 (commented):
LGTM, apart from the docs for ListPack() :)

internal/repository/repository.go
aawsome: Right - I chose to exchange this as this `size` was - no wonder - not used.
The comment is changed, sorry for the confusion!

aawsome:
Also rebased this to master as #3113 was merged.

fd0 (approved):
Great, thanks!

PR #3106, opened 2020-11-17
Parallelize tree walk in prune and copy and add progress bar to check

MichaelEischer:
This PR parallelizes the tree walking in `prune` to find used blobs and in `copy` to find tree blobs for copying. For that the parallel tree loading used by the `check` is extracted as `StreamTrees` and reused. Afterwards `StreamTrees` is enhanced with progress tracking which reports how many snapshots are already fully processed. This replaces the current progress tracking in prune and adds a progress bar to the `Structure` step of the `check` command.

The signature of the StreamTrees function is composed as follows:
```
func StreamTrees(
    // ctx must be tied to wg. These four parameters and the result channel are the result
    // of extracting the corresponding implementation from the checker into a function
    ctx context.Context, wg *errgroup.Group, repo restic.TreeLoader, trees restic.IDs,
    // allow the caller to track visited trees as necessary. This e.g. allows subtree skipping
    // across StreamTrees calls (used by copy) and caller specific tracking of visited trees
    // The latter is used to prevent duplicate bookkeeping in the checker
    visit func(tree restic.ID) bool,
    // progress bar
    p *progress.Counter
) <-chan TreeItem
```

There are currently a few open points:
- [x] I've put the parallel implementations into a `parallel` package. They might fit better into e.g. the `repository` package. I also have an implementation of a `StreamIndex` function to unify index loading, which however must reside in the `repository` package as everything else currently causes import loops.
- [x] I've split the `Structure` method of the checker into `LoadSnapshots` & `Structure` as a progress bar currently requires knowing the upper limit up front. This add quite a bit of boilerplate code at different locations. It might be to enhance the progress bar to allow setting the limit later on.

Tests and changelog are still missing. Most of the code should be tested indirectly, especially by the `FindUsedBlobs` test or the integration tests, but I haven't looked at the test coverage yet.

The changes made in this PR were mentioned in https://github.com/restic/restic/issues/2162#issuecomment-723562425 .

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -104,12 +104,8 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		dstSnapshotByOriginal[*sn.ID()] = append(dstSnapshotByOriginal[*sn.ID()], sn)
 	}
 
-	cloner := &treeCloner{
-		srcRepo:      srcRepo,
-		dstRepo:      dstRepo,
-		visitedTrees: restic.NewIDSet(),
-		buf:          nil,
-	}
+	// remember already processed trees across all snapshots
+	visitedTrees := restic.NewIDSet()
 
 	for sn := range FindFilteredSnapshots(ctx, srcRepo, opts.Hosts, opts.Tags, opts.Paths, args) {
 		Verbosef("\nsnapshot %s of %v at %s)\n", sn.ID().Str(), sn.Paths, sn.Time)
@@ -134,7 +130,7 @@ func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
 		}
 		Verbosef("  copy started, this may take a while...\n")
 
-		if err := cloner.copyTree(ctx, *sn.Tree); err != nil {
+		if err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree); err != nil {
 			return err
 		}
 		debug.Log("tree copied")
@@ -178,31 +174,29 @@ func similarSnapshots(sna *restic.Snapshot, snb *restic.Snapshot) bool {
 	return true
 }
 
-type treeCloner struct {
-	srcRepo      restic.Repository
-	dstRepo      restic.Repository
-	visitedTrees restic.IDSet
-	buf          []byte
-}
+func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
+	visitedTrees restic.IDSet, rootTreeID restic.ID) error {
 
-func (t *treeCloner) copyTree(ctx context.Context, rootTreeID restic.ID) error {
 	wg, ctx := errgroup.WithContext(ctx)
 
-	treeStream := restic.StreamTrees(ctx, wg, t.srcRepo, restic.IDs{rootTreeID}, func(treeID restic.ID) bool {
-		visited := t.visitedTrees.Has(treeID)
-		t.visitedTrees.Insert(treeID)
+	treeStream := restic.StreamTrees(ctx, wg, srcRepo, restic.IDs{rootTreeID}, func(treeID restic.ID) bool {
+		visited := visitedTrees.Has(treeID)
+		visitedTrees.Insert(treeID)
 		return visited
 	})
 
 	wg.Go(func() error {
+		// reused buffer
+		var buf []byte
+
 		for tree := range treeStream {
 			if tree.Error != nil {
 				return fmt.Errorf("LoadTree(%v) returned error %v", tree.ID.Str(), tree.Error)
 			}
 
 			// Do we already have this tree blob?
-			if !t.dstRepo.Index().Has(restic.BlobHandle{ID: tree.ID, Type: restic.TreeBlob}) {
-				newTreeID, err := t.dstRepo.SaveTree(ctx, tree.Tree)
+			if !dstRepo.Index().Has(restic.BlobHandle{ID: tree.ID, Type: restic.TreeBlob}) {
+				newTreeID, err := dstRepo.SaveTree(ctx, tree.Tree)
 				if err != nil {
 					return fmt.Errorf("SaveTree(%v) returned error %v", tree.ID.Str(), err)
 				}
@@ -219,17 +213,17 @@ func (t *treeCloner) copyTree(ctx context.Context, rootTreeID restic.ID) error {
 				// Copy the blobs for this file.
 				for _, blobID := range entry.Content {
 					// Do we already have this data blob?
-					if t.dstRepo.Index().Has(restic.BlobHandle{ID: blobID, Type: restic.DataBlob}) {
+					if dstRepo.Index().Has(restic.BlobHandle{ID: blobID, Type: restic.DataBlob}) {
 						continue
 					}
 					debug.Log("Copying blob %s\n", blobID.Str())
 					var err error
-					t.buf, err = t.srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, t.buf)
+					buf, err = srcRepo.LoadBlob(ctx, restic.DataBlob, blobID, buf)
 					if err != nil {
 						return fmt.Errorf("LoadBlob(%v) returned error %v", blobID, err)
 					}
 
-					_, _, err = t.dstRepo.SaveBlob(ctx, restic.DataBlob, t.buf, blobID, false)
+					_, _, err = dstRepo.SaveBlob(ctx, restic.DataBlob, buf, blobID, false)
 					if err != nil {
 						return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
 					}
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/restic"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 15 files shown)


internal/parallel/trees.go
```diff
@@ -0,0 +1,191 @@
...
+			debug.Log("tree sent to process: %v", job.ID)
+			outCh = nil
+			inCh = in
+		}
+	}
+}
+
+// StreamTrees iteratively loads the given trees and their subtrees. The visit method
+// is guaranteed to always be called from the same goroutine.
+func StreamTrees(ctx context.Context, wg *errgroup.Group, repo restic.TreeLoader, trees restic.IDs, visit func(tree restic.ID) bool, p *progress.Counter) <-chan TreeItem {
```
greatroar, line 161: I don't understand why the wg argument is there. All workers inside the function are added to it, but they never return non-nil. Only the associated context is needed to cancel them. If callers always either drain the channel or cancel the context, there should not be any leaks, right?

Also, I suggest renaming visit to skip and swapping its return value's meaning. The existing visit funcs all return !someCondition and visit suggests the visitor pattern, but they should really only do a short (and idempotent) check.

aawsome, line 161: I suggest to add a parameter `rootTreeIdx` to the `visit` func which is the index of the root tree in `trees`.
That way the StreamTress can be also used in future to speed up the `restic find` command.

MichaelEischer: > If callers always either drain the channel or cancel the context, there should not be any leaks, right?

You always should drain the channel, even if the context is canceled to ensure that all go routines are stopped before continuing. With the errgroup one just has to stick to the usual pattern of calling `wg.Wait()`.

I've renamed the callback method to `skip`.

MichaelEischer: I like the suggestion. However, I'd suggest to defer it until there's an actual usecase (e.g. when an optimized find gets implemented) such that we know how exactly the interface has to look like.

aawsome: Alright :+1: , just wanted to mention it in case I'll forget the idea :wink:

internal/ui/progress/counter.go
```diff
@@ -66,13 +76,17 @@ func (c *Counter) Done() {
 	*c = Counter{} // Prevent reuse.
 }
 
-func (c *Counter) get() uint64 { return atomic.LoadUint64(&c.value) }
+// Get the current Counter value. This method is concurrency-safe.
+func (c *Counter) Get() uint64 { return atomic.LoadUint64(&c.value) }
+
+// GetMax returns the expected maximum. This method is concurrency-safe.
+func (c *Counter) GetMax() uint64 { return atomic.LoadUint64(&c.max) }
```
greatroar, line 83: This doesn't seem to get used outside the package.

MichaelEischer: `getMax` is now unexported again. Looks like I've forgotten to clean it up along with the broken test code that used that method.

`Get` is still exported as it's used in testing that the progress bar behaves somewhat correctly. I wonder whether it would be better to add a testing helper method that makes it more obvious that `Get` should only be called by tests.

MichaelEischer:
I've moved the `SteamTrees` function to the restic package which already contains related `ForAllSnapshots` function.

internal/restic/find.go
```diff
@@ -9,31 +15,38 @@ type TreeLoader interface {
...
+func FindUsedBlobs(ctx context.Context, repo TreeLoader, treeIDs IDs, blobs BlobSet, p *progress.Counter) error {
+	var lock sync.Mutex
 
-	tree, err := repo.LoadTree(ctx, treeID)
-	if err != nil {
-		return err
-	}
+	wg, ctx := errgroup.WithContext(ctx)
+	treeStream := StreamTrees(ctx, wg, repo, treeIDs, func(treeID ID) bool {
+		lock.Lock()
```
aawsome, line 23: maybe add a comment that this lock is needed as data blobs are added concurrenly..

cmd/restic/cmd_copy.go
```diff
@@ -184,57 +185,64 @@ type treeCloner struct {
 	buf          []byte
 }
 
-func (t *treeCloner) copyTree(ctx context.Context, treeID restic.ID) error {
+func (t *treeCloner) copyTree(ctx context.Context, rootTreeID restic.ID) error {
```
aawsome, line 188: Now that `copyTree` is no longer recursively called, we could think of calling this with all root tree IDs as argument. But this would need a larger change in the `copy` code and can also be postponed...

cmd/restic/cmd_copy.go
```diff
@@ -184,57 +185,64 @@ type treeCloner struct {
...
 	// We have already processed this tree
-	if t.visitedTrees.Has(treeID) {
+	if t.visitedTrees.Has(rootTreeID) {
```
aawsome, line 190: isn't this check already done by the `skip` function?

internal/checker/checker.go
```diff
@@ -308,200 +308,27 @@ func (e TreeError) Error() string {
...
-				treeError = TreeError{ID: job.ID, Errors: errs}
-				outCh = out
-				inCh = nil
-			}
-
-		case outCh <- treeError:
-			debug.Log("tree %v: sent %d errors", treeError.ID, len(treeError.Errors))
-			outCh = nil
-			inCh = in
+func (c *Checker) checkTreeWorker(ctx context.Context, in <-chan restic.TreeItem, out chan<- error) {
```
aawsome, line 312: I found the name `in` irritating. Maybe `trees`?

aawsome (commented):
I did not look at `tree_stream.go` today (but I think I already had a quick look some time and it doesn't differ much from the former check code).

For the other changes, I have just a few minor suggestions. Besides this these changes are IMO fine!

internal/checker/checker.go
MichaelEischer: Is renamed.

cmd/restic/cmd_copy.go
MichaelEischer: I've removed the check. Setting up `StreamTrees` to do nothing should require basically no time at all.

cmd/restic/cmd_copy.go
MichaelEischer: I've added a commit to remove the `treeCloner` struct for now. Copying all snapshots in one go would be the task of another PR. That would also require some though on the user interface to possibly provide more feedback to the user than is done currently.

internal/restic/find.go
MichaelEischer: Is added.

internal/checker/checker.go
```diff
@@ -541,19 +367,25 @@ func (c *Checker) Structure(ctx context.Context, errChan chan<- error) {
 		}
 	}
 
-	treeIDChan := make(chan restic.ID)
-	treeJobChan1 := make(chan treeJob)
-	treeJobChan2 := make(chan treeJob)
+	wg, ctx := errgroup.WithContext(ctx)
+	treeStream := restic.StreamTrees(ctx, wg, c.repo, trees, func(treeID restic.ID) bool {
+		c.blobRefs.Lock()
```
aawsome, line 372: Here we have the same situation that it may be hard to see that this lock is needed because of the concurrent use in `checkTree` - maybe also add a comment here?

internal/restic/tree_stream.go
```diff
@@ -0,0 +1,190 @@
...
+			debug.Log("tree sent to process: %v", job.ID)
+			outCh = nil
+			inCh = in
+		}
+	}
+}
+
+// StreamTrees iteratively loads the given trees and their subtrees. The skip method
+// is guaranteed to always be called from the same goroutine.
+func StreamTrees(ctx context.Context, wg *errgroup.Group, repo TreeLoader, trees IDs, skip func(tree ID) bool, p *progress.Counter) <-chan TreeItem {
```
aawsome, line 160: I'm coming back to @greatroar 's comment. I also find it irritating and somewhat artificial that we should pass a `wg` to the StreamTrees.

I would say, we could use:

```golang
func loadTreeWorker(ctx context.Context, repo TreeLoader,
	in <-chan trackedID, out chan<- trackedTreeItem) {

	for treeID := range in {
		tree, err := repo.LoadTree(ctx, treeID.ID)
		debug.Log("load tree %v (%v) returned err: %v", tree, treeID, err)
		job = trackedTreeItem{TreeItem: TreeItem{ID: treeID.ID, Error: err, Tree: tree}, rootIdx: treeID.rootIdx}

		select {
		case <-ctx.Done():
			return
		case out <- job:
		}
	}
}

func StreamTrees(ctx context.Context, wg *errgroup.Group, repo TreeLoader, trees IDs, skip func(tree ID) bool, p *progress.Counter) <-chan TreeItem {
	loaderChan := make(chan trackedID)
	loadedTreeChan := make(chan trackedTreeItem)
	treeStream := make(chan TreeItem)

	for i := 0; i < streamTreeParallelism; i++ {
		go loadTreeWorker(ctx, repo, loaderChan, loadedTreeChan)
	}

	go func() {
		defer close(loaderChan)
		defer close(treeStream)
		filterTrees(ctx, trees, loaderChan, loadedTreeChan, treeStream, skip, p)
	}()
	return treeStream
}
```

Or am I missing something?

MichaelEischer: The problem here is that when the context is cancelled, there are still several `loadTreeWorker`s running. We should wait for these to have stopped before continuing. And that is the main use case for the `errgroup` (now that I think of it, a `WaitGroup` would also be sufficient for that). 

I've changed the PR to use your simplified `loadTreeWorker`. The idea there was to break the cycle between `filterTrees` and `loadTreeWorker` on both sides, but doing it only in `filterTrees` is enough.

internal/checker/checker.go
MichaelEischer: I've added a small comment

cmd/restic/cmd_copy.go
```diff
@@ -177,64 +174,64 @@ func similarSnapshots(sna *restic.Snapshot, snb *restic.Snapshot) bool {
...
+				// Assurance only.
+				if newTreeID != tree.ID {
+					return fmt.Errorf("SaveTree(%v) returned unexpected id %s", tree.ID.Str(), newTreeID.Str())
+				}
 			}
 
-			_, _, err = t.dstRepo.SaveBlob(ctx, restic.DataBlob, t.buf, blobID, false)
-			if err != nil {
-				return fmt.Errorf("SaveBlob(%v) returned error %v", blobID, err)
+			// TODO: parellize blob down/upload
```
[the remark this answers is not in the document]
MichaelEischer: Thanks.

internal/restic/tree_stream.go
aawsome: Ok, thanks for the explanaition. I suggest to use

```golang
// If the processing is trees is cancelled, the calling function *must* cancel the context ctx!
func StreamTrees(ctx context.Context, repo TreeLoader, trees IDs, skip func(tree ID) bool, p *progress.Counter) <-chan TreeItem {
	loaderChan := make(chan trackedID)
	loadedTreeChan := make(chan trackedTreeItem)
	treeStream := make(chan TreeItem)
	
	var loadTreeWg sync.WaitGroup
		
	for i := 0; i < streamTreeParallelism; i++ {
		loadTreeWg.Add(1)
		go func() {
			loadTreeWorker(ctx, repo, loaderChan, loadedTreeChan)
			loadTreeWg.Done()
		}()
	}

	// close once all loadTreeWorkers have completed
	go func() {
		loadTreeWg.Wait()
		close(loadedTreeChan)
	} ()

	go func() {
		defer close(loaderChan)
		defer close(treeStream)
		filterTrees(ctx, trees, loaderChan, loadedTreeChan, treeStream, skip, p)
	}()
	return treeStream
}
```

There is another thing: if for some reason the processing of the trees should be aborted, these goroutines must be cancelled. This is done by the errgroup in the case of an error. But it still does not prevent the caller to directly consume the channel and not completely process it. So I'd prefer the solution with a comment on how to use this function and make sure that the consumers cancel the given context.

MichaelEischer: When using the errgroup one will notice rather quickly when the channel was not processed completely. In that case `Wait()` will just hang.

The more I think about the error handling for `StreamTrees` the more I'm convinced that using an `errgroup` is by far the best option. But the reasoning behind that requires some explanation. So let's first take a step back and take a look at the underlying question: What should happen to the goroutines started by StreamTrees when the latter completes or there's an error of some kind? My opinion here is that the proper answer is always the same: make sure that _all_ goroutines have been wound down properly before continuing / passing the error up the call chain. However, the step from stopping all goroutines when there was not error to always doing so introduces quite a few complications.

In the normal case when everything works properly stopping everything is rather easy to guarantee. Just read all trees from the channel and you're done. The crux of the problem shows up when taking a look at how to properly handle errors when using `StreamTrees`. For example when using the context-only variant of `StreamTrees` a snippet of `FindUsedBlobs` would look as follows:
```go
ctx, cancel := context.WithCancel(ctx)
treeStream := StreamTrees(ctx, repo, treeIDs, func(treeID ID) bool { [...] }, p)
err := func() error {
	// Explicit cancel is necessary as the error return doesn't cancel the context
	defer cancel()
	for tree := range treeStream {
		if tree.Error != nil {
			return tree.Error
		}
		[...]
	}
}()
for tree := range treeStream {
	// catch-all loop to drain stream
}
return err
```

The only means of synchronization that (partially) ensures that all goroutines have stopped is to drain the `treeStream`. (With the last proposed variant the would still be no synchronization against the loadTreeWorkers). However, this complicates error handling, so I've decided to use a catch-all solution using a second loop here. Avoiding the use of a second loop only makes things worse:
```go
ctx, cancel := context.WithCancel(ctx)
treeStream := StreamTrees(ctx, repo, treeIDs, func(treeID ID) bool { [...] }, p)
// between here and the final `return err`, no other return calls are allowed
var err error
for tree := range treeStream {
	if err != nil {
		cancel()
		// we must still drain the stream
		continue
	}
	if tree.Error != nil {
		// this pattern is required for every error, not following
		// it can cause some ugly race conditions.
		err = tree.Error
		continue
	}
	[...]
}
return err
```

Neither of the last two variants is enforced by the method interface, but instead would only be documented in a comment. Whereas when using an errgroup, it's a simple standard implementation:
```go
wg, ctx := errgroup.WithContext(ctx)
// StreamTrees could also be called from within wg.Go
treeStream := StreamTrees(ctx, wg, repo, treeIDs, func(treeID ID) bool { [...] }, p)
wg.Go(func() error {
	for tree := range treeStream {
		if tree.Error != nil {
			return tree.Error
		}
		[...]
	}
})
return wg.Wait()
```
There's no need here to drain the `treeStream` as the shutdown synchronization doesn't use it as side-channel but is instead done explicitly via the errgroup.

The first context-only example could be modified to use an `errgroup` but not passing that errgroup on to `StreamTrees`  would be rather pointless.

aawsome: I think that 
```golang
ctx, cancel := context.WithCancel(ctx)
defer cancel()
treeStream := StreamTrees(ctx, repo, treeIDs, func(treeID ID) bool { [...] }, p)
for tree := range treeStream
	if tree.Error != nil {
		return tree.Error
	}
	[...]
}
```
should work and for me also seems to be golang way to use such a streaming interator, see e.g. https://golang.org/pkg/context/#example_WithCancel
Can you explain why you think that this would not stop the `loadTreeWorkers` if the loop is aborted?

In your last example, noone prevents the callers to use
```golang
// StreamTrees needs an errgroup, so create one
wg, ctx := errgroup.WithContext(ctx)
treeStream := StreamTrees(ctx, wg, repo, treeIDs, func(treeID ID) bool { [...] }, p)
for tree := range treeStream {
	if tree.Error != nil {
		return tree.Error
	}
	[...]
}
```
which would leak some goroutines...

So, from the perspective of the function signature, I'm fine with either your proposed `errgroup` variant or my proposed `ctx`-only variant, but I think that both would require some comment on how to use it properly :wink:

MichaelEischer: > Can you explain why you think that this would not stop the loadTreeWorkers if the loop is aborted?

The goroutines will be stopped in either case. The aspect I was looking at was _when_ that happens. Will the goroutines eventually be stopped as in the golang streaming iterator example, or do we know that all goroutines are already stopped before continuing? The context-only variant always ensures the eventual stopping, but getting the goroutines to stop at a defined point in time requires additional constructs as in my code examples. For the example from the context documentation it doesn't matter much when the goroutine is stopped as it has no sideeffects. However, that is not the case for StreamTrees as a loaded tree might e.g. cause a pack file to be added to the cache.

Forgetting the `Wait` call is of course possible, but a reviewer can follow a simple rule here: there are no `return`s for the current function allowed between `errgroup.WithContext` and `wg.Wait`. The main benefit here is that this rule is only specific to the generic `errgroup` construct and does not depend on the specific semantics of a StreamTrees or something else.

aawsome: Ok, I agree it might be preferable here to wait for the goroutines to cancel before continuing. But this could be also achieved by the following construct:
```golang
treeStream, wait := StreamTrees(ctx, repo, treeIDs, func(treeID ID) bool { [...] }, p)
defer wait()
for tree := range treeStream
	if tree.Error != nil {
		return tree.Error
	}
	[...]
}
```
where `StreamTree` returns the channel and a `func wait()` (could also be named `finish` or `cancel`) which (if needed cancels the context used within and) effectively waits for all goroutines to cancel. 

I would like that solution more as it shows by the function signature what is actually important...

But of course, your solution is IMO also correctly working. I also think this very minor point should not prevent this PR from being finished. So either way, if you could add a comment about how to handle the cancellation (maybe just referencing to our discussion here), I would be very satisfied :+1:

MichaelEischer: I've added a short comment on how to properly shutdown `StreamTrees`. This discussion is IMO too long if one just wants to know how to properly use the function.

The `defer wait()` solution would also work, but at that point the only difference to using an `errgroup` is whether the object controlling the goroutine lifetime is passed to `StreamTrees` or is returned by it. I prefer the variant using `errgroup` as it provides a clear hierarchy of "resource managers" to control concurrency.

But let's stop the discussion here.

aawsome (approved):
LGTM!

fd0 (approved):
LGTM! I've rebased the code on master and resolved a merge conflict.

PR #3107, opened 2020-11-18
do not require gs bucket permissions to init repository

### LINKED ISSUE #3100: error on init repo for google cloud storage on restic 11 but works in restic 10
restic version
--------------------------

`restic -r gs:bucket/folder init`

Expected behavior
-----------------

repo is initialized on restic 11.0 and 10.0

Actual behavior
---------------

Fatal: create repository at gs:bucket:/folder failed: service.Buckets.Get: googleapi: Error 403: xyz@cloud-id.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket., forbidden 

but on 10.0 it works

eleith:

a gs service account may only have object permissions on an existing
bucket but no bucket create/get permissions.

these service accounts currently are blocked from initializing a
restic repository because restic can not determine if the bucket exists.

this PR updates the logic to assume the bucket exists when the bucket
attribute request results in a permissions denied error.

this way, restic can still initialize a repository if the service
account does have object permissions

fixes: https://github.com/restic/restic/issues/3100

Checklist
---------

### DIFF
--- a/internal/backend/gs/gs.go
+++ b/internal/backend/gs/gs.go
@@ -136,6 +136,11 @@ func Create(cfg Config, rt http.RoundTripper) (restic.Backend, error) {
 	ctx := context.Background()
 	exists, err := be.bucketExists(ctx, be.bucket)
 	if err != nil {
+		if e, ok := err.(*googleapi.Error); ok && e.Code == http.StatusForbidden {
+			// the bucket might exist!
+			// however, the client doesn't have storage.bucket.get permission
+			return be, nil
+		}
 		return nil, errors.Wrap(err, "service.Buckets.Get")
 	}
 
--- a/changelog/unreleased/issue-3100
+++ b/changelog/unreleased/issue-3100
@@ -1,0 +1,10 @@
+Bugfix: Do not require gs bucket permissions when running init 
+
+Restic used to require bucket level permissions for the gs backend 
+in order to initialize a restic repository.
+
+It now allows a gs service account to initialize a repository if the 
+bucket does exist and the service account has permissions to write/read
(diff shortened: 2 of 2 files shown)


eleith:
can i get some direction on how to add tests for this specific backend logic?

in my quick review of the backend tests, they seem to be a generic test suite meant to test all backends, but this PR adds specific logic for one backend, and it's unclear to me how to add unit tests for this.

cc @fd0 / @MichaelEischer

fd0:
Thanks for giving this a try! I don't think we can even add such a test. The backend tests for Google Cloud Storage run with a service account provided by a sponsor, I don't have an easy way to create a service account necessary to produce the error.

So we'll just go without.

internal/backend/gs/gs.go
```diff
@@ -136,7 +136,14 @@ func Create(cfg Config, rt http.RoundTripper) (restic.Backend, error) {
 	ctx := context.Background()
 	exists, err := be.bucketExists(ctx, be.bucket)
 	if err != nil {
-		return nil, errors.Wrap(err, "service.Buckets.Get")
+		switch err.(type) {
```
fd0, line 139: You almost got it! You could use a check similar to the following, this checks the type as well as the status code:

https://github.com/googleapis/google-cloud-go/blob/91b992b32bb3488a927338e7fd3203d9b43036eb/storage/bucket.go#L168-L170

eleith: doh. forgot the status check. done.

fd0 (approved):
LGTM, t hanks!

PR #3109, opened 2020-11-18
restore: Don't save pack content in memory 

aawsome:

`filerestorer.go` (which is used in restore) used to save packs (or at least large parts) in memory in order to process the blobs by a `ReaderAt`. As this is not supplied by the `Backend` implementation, the contents were loaded into a memory buffer.

This PR first sorts the blobs by the offset before processing them and thus allows to only use a `Reader` to read all blobs.
This remove the necessity to store the pack content in memory.

See #2750 
This PR reduces memory usage a lot when using large pack sizes.

Checklist
---------

- [x] I have added tests for all changes in this PR.

### DIFF
--- a/internal/restorer/filerestorer.go
+++ b/internal/restorer/filerestorer.go
@@ -1,11 +1,12 @@
 package restorer
 
 import (
-	"bytes"
+	"bufio"
 	"context"
 	"io"
 	"math"
 	"path/filepath"
+	"sort"
 	"sync"
 
 	"github.com/restic/restic/internal/crypto"
@@ -179,6 +180,8 @@ func (r *fileRestorer) restoreFiles(ctx context.Context) error {
 	return nil
 }
 
+const maxBufferSize = 4 * 1024 * 1024
+
 func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) {
 
 	// calculate pack byte range and blob->[]files->[]offsets mappings
@@ -226,18 +229,12 @@ func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) {
 		}
 	}
 
-	packData := make([]byte, int(end-start))
-
-	h := restic.Handle{Type: restic.PackFile, Name: pack.id.String()}
-	err := r.packLoader(ctx, h, int(end-start), start, func(rd io.Reader) error {
-		l, err := io.ReadFull(rd, packData)
-		if err != nil {
-			return err
-		}
-		if l != len(packData) {
-			return errors.Errorf("unexpected pack size: expected %d but got %d", len(packData), l)
-		}
-		return nil
+	sortedBlobs := make([]restic.ID, 0, len(blobs))
+	for blobID := range blobs {
+		sortedBlobs = append(sortedBlobs, blobID)
+	}
+	sort.Slice(sortedBlobs, func(i, j int) bool {
+		return blobs[sortedBlobs[i]].offset < blobs[sortedBlobs[j]].offset
 	})
 
 	markFileError := func(file *fileInfo, err error) {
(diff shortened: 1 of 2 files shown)


aawsome:
Also noticed, that we could reuse the buffer in `filerestorer.go#loadBlob`. Should I add this here?

metalsp0rk:
We could re-focus this as general memory optimizations if you'd like.

aawsome:
I had a bug in the calculation of bytes to skip. This was only detected randomly by the (randomized) integration tests. I added a new unit test for `filerestorer` which would have caught this bug.

aawsome:
@ifedorenko: As the restorer has been mainly implemented by you - do you think you can have a look at this PR?

internal/restorer/filerestorer.go
```diff
@@ -327,25 +333,33 @@ func (r *fileRestorer) loadBlob(rd io.Reader, blobID restic.ID, length int) ([]b
...
-	return plaintext, nil
+	// move decrypted data to the start of the buffer
+	copy(buf, plaintext)
+	return buf[:len(plaintext)], nil
 }
 
-func skipReader(rd io.Reader, skip int) error {
+// skipReader reads exactly skip bytes and ignores them, reusing a buffer if given.
+func skipReader(rd io.Reader, skip int, buf []byte) ([]byte, error) {
 	const maxReadLen = 1024
```
MichaelEischer, line 343: This is way too small. `rd.Read` can now trigger IO and is not just an in-memory copy.

internal/restorer/filerestorer.go
```diff
@@ -301,10 +302,15 @@ func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) {
...
-func (r *fileRestorer) loadBlob(rd io.Reader, blobID restic.ID, length int) ([]byte, error) {
+func (r *fileRestorer) loadBlob(rd io.Reader, blobID restic.ID, length int, buf []byte) ([]byte, error) {
 	// TODO reconcile with Repository#loadBlob implementation
 
-	buf := make([]byte, length)
+	switch {
+	case cap(buf) < length:
+		buf = make([]byte, length)
+	case len(buf) != length:
+		buf = buf[:length]
```
MichaelEischer, line 312: I'd call `buf = buf[:length]` unconditionally. reslicing is probably cheaper than / just as expensive as checking the buffer length.

internal/restorer/filerestorer.go
```diff
@@ -327,25 +333,33 @@ func (r *fileRestorer) loadBlob(rd io.Reader, blobID restic.ID, length int) ([]b
 		return nil, errors.Errorf("blob %v returned invalid hash", blobID)
 	}
 
-	return plaintext, nil
+	// move decrypted data to the start of the buffer
+	copy(buf, plaintext)
```
MichaelEischer, line 337: It would be possible to avoid the data relocation by returning both `plaintext` and `buf`.

internal/restorer/filerestorer.go
```diff
@@ -248,60 +242,77 @@ func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) {
 		}
 	}
 
+	h := restic.Handle{Type: restic.PackFile, Name: pack.id.String()}
+	err := r.packLoader(ctx, h, int(end-start), start, func(rd io.Reader) error {
+		var blobData []byte
+		lastBlobEnd := start
```
MichaelEischer, line 248: I'd prefer `currentBlobEnd` or `latestBlobEnd`.

internal/restorer/filerestorer_test.go
```diff
@@ -189,8 +198,7 @@ func TestFileRestorerBasic(t *testing.T) {
 		{
 			name: "file2",
 			blobs: []TestBlob{
-				{"data2-1", "pack2-1"},
-				{"data2-2", "pack2-2"},
+				{"data2-1", "pack2-1"}, {"data2-2", "pack2-2"},
```
MichaelEischer, line 201: What's the reason for this change?

MichaelEischer (changes requested):
Have you measured whether / which performance impact the removal of the in-memory buffer has? My current expectation is that the current changes lead to a large regression for the SFTP backend and probably also several others. For SFTP each `rd.Read` call probably triggers a read request. The local backend will issue lots of syscalls which might also apply in similar fashion to other backends.

The simplest solution would probably to introduce a `bufio.NewReaderSize(rd io.Reader, size int)` with a buffer of a few megabytes or `int(end-start)`, whichever is lower. That way the memory usage is still bounded, but we should be able to avoid most of the performance impact. `Reader.Discard` would also allow replacing the `skipReader` function.

internal/restorer/filerestorer.go
aawsome: I now use `bufio` and the `Discard` method which is already defined there.

internal/restorer/filerestorer.go
aawsome: is changed.

internal/restorer/filerestorer.go
aawsome: Good idea! is changed.

internal/restorer/filerestorer.go
aawsome: changed.

internal/restorer/filerestorer_test.go
aawsome: Sorry, was a mistake. Is undone.

aawsome:
> The simplest solution would probably to introduce a `bufio.NewReaderSize(rd io.Reader, size int)` with a buffer of a few megabytes or `int(end-start)`, whichever is lower. That way the memory usage is still bounded, but we should be able to avoid most of the performance impact. `Reader.Discard` would also allow replacing the `skipReader` function.

Good idea! Is changed now. I also didn't notice there is a Discard method already (only looked in the `io` package...)

I didn't make any speed tests. But now I use a maximum buffer of 4MB. If there is anyone, who could help out with performance tests, I would be very happy!

aawsome:
I was wondering why the error in `loadBlob` (`Read` instead of `io.ReadAll`) did cause the integration tests to fail but not bring up the error which should be set in `loadBlob`. Turns out the errors are set in `filereader.go` but never checked.. I'll make a separate PR with a fix.

dimejo:
> I didn't make any speed tests. But now I use a maximum buffer of 4MB. If there is anyone, who could help out with performance tests, I would be very happy!

I ran some tests and restored a snapshot with ~9GiB and ~24k files from my NAS.

```
restic_stable version
```

```
restic_master version
restic 0.11.0 (v0.11.0-249-ge8abc79ce) compiled with go1.15.6 on linux/amd64
```

```
restic_pr-3109 version
restic 0.11.0 (v0.11.0-252-g34a33565c) compiled with go1.15.6 on linux/amd64
```


restic version | Backend | time (mm:ss) | memory (MiB)
-- | -- | -- | --
stable | SFTP | 01:46 | 404
stable | rest-server | 07:52 | 381
master | SFTP | 01:46 | 405
master | rest-server | 07:51 | 373
PR 3109 | SFTP | 01:45 | 355
PR 3109 | rest-server | 07:47 | 315

_Posted values are average for 3 runs (see attached logs from `time` for details)._

[restic_pr-3109_logs.zip](https://github.com/restic/restic/files/5760816/restic_pr-3109_logs.zip)

MichaelEischer (approved):
LGTM. I've added a commit to lower-case the bufRd variable name.

Issue #2706, opened 2020-04-23
Option for newlines after each progress update


FrederikNS:
Output of `restic version`
--------------------------

restic 0.9.6 compiled with go1.13.4 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I would like the ability to print a newline at each progress log update.

What are you trying to do?
--------------------------

I'm running restic in Kubernetes, where I'm using the native methods for reading the logs (`kubectl logs <restic pod>`). Unfortunately in Kubernetes a log line is only presented when the line is terminated with a newline. That means that when I run any restic command, and progress output is continually printed, I cannot see it. I only see the final progress line written.

I ran a `prune` to clean out some cruft from my backup repository, and I even ran it in verbose mode, but I couldn't see anything but the 3 first lines of log for a while:

    created new cache in /root/.cache/restic
    counting files in repo
    building new index for repo

I waited and waited, and was concerned if anything was actually happening, until ~49 hours later I finally got the next log line:

    [48:58:16] 100.00%  581951/581951 packs

It's pretty clear to me that restic has continually printed progress log, but as there never was a newline, I just couldn't see it until it was done.

I haven't tried yet, as I'm still waiting for the prune to finish, but I suspect that the `--json` parameter would actually give me the progress report in separate lines, but it would also be very nice to have a non-json "print newline for each progress update" mode.

I haven't tried, but I suspect that the same behaviour would happen for plain Docker without Kubernetes.

Bonus request: If this log output method could be controlled via an Environment Variable, it would be even easier to handle in my setup.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yes, restic seems to be a brilliant backup solution, and fits my usecase very well.

rawtaz:
Is restic doing something that is unconventional in how it outputs its progress?

If not, wouldn't it be better if Kubernetes is fixed to handle this type of (from what I can tell common) progress bar? I have a hard time thinking we should add a switch to work around a problem in Kubernetes. I mean, adding a flag just to control one little character in the output seems pretty crufty/dirty to me.

At the very least, let's see if the `--json` output works for you, if you don't mind.

FrederikNS:
Well, it's not exactly "unconventional"... Plenty of CLI tools does it, however it sort of goes against the [12-factor application guidelines](https://12factor.net/logs), which specified "one event per line". It also makes it impossible to collect that kind of log in a unified log collection system like EFK/ELK.

[...]

PR #3112, opened 2020-11-18
Limit progress bar updates to once per second on non-terminal outputs

MichaelEischer:
PR #3058 rewrote the progress bar implementation and in that progress reenabled it for non-terminal output.

However, the code accidentally checked whether stdin is a terminal instead of stdout, the former is not relevant here as the output is printed on stdout.

Reported at https://forum.restic.net/t/new-prune-log-spam/3284 .

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ There's no nice way right now to test this.
- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)). ~~ I haven't added a changelog entry as this PR is a fixup for #3058, which (currently) does not have a changelog entry.

### DIFF
--- a/cmd/restic/progress.go
+++ b/cmd/restic/progress.go
@@ -16,7 +16,7 @@ func newProgressMax(show bool, max uint64, description string) *progress.Counter
 	}
 
 	interval := time.Second / 60
-	if !stdinIsTerminal() {
+	if !stdoutIsTerminal() {
 		interval = time.Second
 	} else {
 		fps, err := strconv.ParseInt(os.Getenv("RESTIC_PROGRESS_FPS"), 10, 64)


plumbeo:
Is there really a need to print progress bars’ updates on non-terminal output? If I’m understanding the change correctly even if limited to 1 line per second it will still mean hundreds or thousands of useless lines in every log file or email message.

The current behaviour (only one line printed at the end) looks a lot more sensible to me.

fd0:
> The current behaviour (only one line printed at the end) looks a lot more sensible to me.

It's a feature request to have status updates even when output is written to a log file. In your case, passing `--quiet` should do what you want. :)

fd0 (approved):
LGTM!

plumbeo:
> > The current behaviour (only one line printed at the end) looks a lot more sensible to me.
> 
> It's a feature request to have status updates even when output is written to a log file. In your case, passing `--quiet` should do what you want. :)

But `--quiet` silences other output that one generally may want in a log file so it’s not a solution. 

My point is that interim updates are something that is necessary when using a software interactively (because they show that the software is still working as expected and allow the user to roughly estimate how much work is left to do) but are useless when one is just interested in receiving a final status report (say an email, that’s my case) or is logging information to eventually be able to check it later.

In that case a user may very well want to have the most output possible (increasing verbosity) but 900 almost identical lines, one per second on a simple 15 minutes run, don’t tell anything useful and only drown relevant information in a sea of noise.

As far as I know this is the behaviour of most software who has progress bars, I may be wrong but I can’t remember seeing interim updates in non-interactive usage.

chrestomanci:
When you get that email with 900 almost identical lines you can scroll past them with a quick flick of the mouse. Those lines will also compress very well, so will take an insignificant amount of email storage capacity.

IMHO the progress bars when the output is not a TTY is useful because for long running jobs it is common practice to direct output to a logfile, and then tail the logfile. Personally when a cron job is taking a long time I will examine /proc to find the process and it's logfile & tail that.

I would agree that once per second could be a bit too frequent. I would suggest once per 10 or 30 seconds would probably be better (or make it configurable).

plumbeo:
What I care about is not disk space, it's attention, which is a quite more valuable commodity: since every task with a progress bar outputs a share of those lines what you get is a lot of almost identical lines interspersed with the few lines you were actually interested in, and what would have been a short email visible in its entirety even on a smartphone display and checked in a couple seconds becomes a long email that must be browsed carefully.

Multiply that for N clients sending M emails every day every time restic runs, or exporting the logs to a central server, and suddenly this is not something you can just work around by scrolling past it with a quick flick of the mouse anymore. I may be wrong but I think that this use case is more common than the occasional time when one has to check the log file live when restic is running.

IMHO the default for non-terminal outputs should go back to not printing the progress bar interim updates, as it was before, and the feature should be explicitly enabled with a specific option, but I suppose the opposite would be good too as long as it's possible to completely disable the extra output.

Enabling it only for higher verbosity levels would be a possibility too, but in that case it should happen only with `--verbose=2` or `-vv` and it must be done carefully because right now the verbosity levels are treated a bit inconsistently (basically `--verbose` and default are the same thing, see https://github.com/restic/restic/pull/3009 for more details).

MichaelEischer:
@plumbeo Please add your use case to #2706. I agree that it's probably best to give user the choice between progress bar updates or not.

Olen:
How about limiting it even more than once per second on non-tty.  I would love to have some periodic feedback about progress of long running commands in my logs, but don't need one log line every second.  Maybe one every 10 or 30 minutes (or even better -every estimated 10% of the remaining time or something), just to know the job is actually progressing.

PR #3119, opened 2020-11-21
Keep mountpoints as empty directories for --one-file-system

### LINKED ISSUE #909: Empty Mountpoints Are Not Backed Up
When backing up a system with, say, `restic backup -x /`, I expect for `/dev/*` or `/proc/*` to not be backed up, since it's a different filesystem. However, I _do_ expect the base directory, `/dev`, to be backed up as an empty directory, since it resides on the parent filesystem. I realize you can't descend into covered mounts, in case an unmounted `/dev` isn't empty, somehow, but at the very least you can back it up as empty?

fd0:

Before, when `restic backup` was run with `--one-file-system`, restic completely ignored mount points. For example, `restic backup --one-file-system /` would create a snapshot which does not include the directory `/proc` (where the `procfs` pseudo file system is usually mounted) or `/tmp/`. This was not so great because when a backup is restored, users needed to create the folders manually.

A workaround was specifying `--exclude=/proc/*` so that `/proc` itself is included, but everything in it was excluded from the backup. This approach is racy, items in `/proc` may appear between the shell expanding the glob and restic looking at the file system.

I had a look what `tar` does and implemented the same behavior: the mountpoints are automatically included in the snapshot as empty directories.

Closes #909

Checklist
---------

### DIFF
--- a/cmd/restic/exclude.go
+++ b/cmd/restic/exclude.go
@@ -199,12 +199,17 @@ func isDirExcludedByFile(dir, tagFilename, header string) bool {
 	return true
 }
 
-// gatherDevices returns the set of unique device ids of the files and/or
-// directory paths listed in "items".
-func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
-	deviceMap = make(map[string]uint64)
-	for _, item := range items {
-		item, err = filepath.Abs(filepath.Clean(item))
+// DeviceMap is used to track allowed source devices for backup. This is used to
+// check for crossing mount points during backup (for --one-file-system). It
+// maps the name of a source path to its device ID.
+type DeviceMap map[string]uint64
+
+// NewDeviceMap creates a new device map from the list of source paths.
+func NewDeviceMap(allowedSourcePaths []string) (DeviceMap, error) {
+	deviceMap := make(map[string]uint64)
+
+	for _, item := range allowedSourcePaths {
+		item, err := filepath.Abs(filepath.Clean(item))
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


fd0:
@aawsome would you mind reviewing this PR? :)

cmd/restic/exclude.go
```diff
@@ -213,57 +218,117 @@ func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
...
 	if len(deviceMap) == 0 {
 		return nil, errors.New("zero allowed devices")
 	}
+
 	return deviceMap, nil
 }
 
+// IsAllowed returns true if the path is located on an allowed device.
+func (m DeviceMap) IsAllowed(item string, deviceID uint64) (bool, error) {
+	for dir := item; ; dir = filepath.Dir(dir) {
```
aawsome, line 239: Here we loop over parent dirs until one of those exactly matches one of the paths given to backup (and check if the device id matches).
Maybe it is clearer to loop over the paths to backup and check if `item` is a subdir (and then check for matching device ids). To test whether it's a subdir I used `filepath.Match` in #3121.

But I don't know which is more performant and as this will be called quite often, we should choose the more performant solution.

cmd/restic/exclude.go
```diff
@@ -213,57 +218,117 @@ func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
...
+
+		// special case: make sure we keep mountpoints (directories which
+		// contain a mounted file system). Test this by checking if the parent
+		// directory would be included.
+		parentDir := filepath.Dir(filepath.Clean(item))
+
+		parentFI, err := fs.Lstat(parentDir)
+		if err != nil {
+			debug.Log("item %v: error running lstat() on parent directory: %v", item, err)
+			// if it doubt, reject
```
aawsome, line 309: typo: if in doubt

cmd/restic/exclude.go
```diff
@@ -213,57 +218,117 @@ func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
...
+			return true
+		}
+
+		parentDeviceID, err := fs.DeviceID(parentFI)
+		if err != nil {
+			debug.Log("item %v: getting device ID of parent directory: %v", item, err)
+			// if it doubt, reject
```
aawsome, line 316: see above.

cmd/restic/exclude.go
```diff
@@ -213,57 +218,117 @@ func gatherDevices(items []string) (deviceMap map[string]uint64, err error) {
...
+		parentDir := filepath.Dir(filepath.Clean(item))
+
+		parentFI, err := fs.Lstat(parentDir)
+		if err != nil {
+			debug.Log("item %v: error running lstat() on parent directory: %v", item, err)
+			// if it doubt, reject
+			return true
+		}
+
+		parentDeviceID, err := fs.DeviceID(parentFI)
```
aawsome, line 313: I think we can move the `fs.DeviceID() call` to `IsAllowed` which should then accept a `os.FileInfo` (or an `DeviceIDer` interface)

This however would make the test settings a bit more involved.

cmd/restic/exclude.go
fd0: I'd rather not change this code in this PR and postpone it for a later PR because I haven't changed it, that's the code from master:

https://github.com/restic/restic/blob/2f9346a5afa95a2a65c08a44117ace8808b62c1f/cmd/restic/exclude.go#L247-L264

cmd/restic/exclude.go
fd0: Hm, I don't understand what the benefit would be?

cmd/restic/exclude.go
aawsome: Yes, I saw that you just moved the code. Took me quite some time to understand what's going on, this is why I thought whether there is a clearer way. But the code is ok, of course! :+1:

cmd/restic/exclude.go
aawsome: The benefit would be less code - especially the doubled error handling for the parent dir.
But I'm also fine with just letting it as it is!

aawsome (approved):
LGTM

fd0:
Thanks!

PR #3125, opened 2020-11-26
Make restic cat respect --no-lock

### LINKED ISSUE #2739: `restic cat` does not respect `--no-lock` flag
Output of `restic version`
--------------------------

`restic --no-lock --json cat lock <LOCK_ID>`

S3 (minio)

Expected behavior
-----------------
restic should ignore the exclusive lock in the repo (due to an ongoing check operation) and output the details of the lock.

Actual behavior
---------------
restic fails with an error message:  
`Fatal: unable to create lock in backend: repository is already locked exclusively by PID XXX on YYY by root (UID 0, GID 0)`

1. Start an operation that requires an exclusive lock (e.g. check, forget, prune)
1. List the locks:
   `$ restic --no-lock --json list locks`
   `<LOCK_ID>`
1. Attempt to cat a lock:
   `$ restic --no-lock --json cat lock <LOCK_ID>`
   `Fatal: unable to create lock in backend: repository is already locked exclusively by PID XXX on YYY by root (UID 0, GID 0)`

The code in `restic/cmd/restic/cmd_cat.go` does not check the `opts.NoLock` before acquiring a lock:

```golang
	lock, err := lockRepo(repo)
	defer unlockRepo(lock)
	if err != nil {
		return err
	}
```

Adding an if around the lock code in `restic/cmd/restic/cmd_cat.go` should solve the issue:

```golang
	if !opts.NoLock {
		lock, err := lockRepo(repo)
		defer unlockRepo(lock)
		if err != nil {
			return err
		}
	}
```

In general I'm very happy with restic and especially the speed and space savings it provides!

metalsp0rk:

Makes the cat command respect the --no-lock global flag

Closes #2739 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2739
+++ b/changelog/unreleased/issue-2739
@@ -1,0 +1,5 @@
+Bugfix: Make cat command respect --no-lock
+
+Restic cat would not respect the --no-lock flag, it now does.
+
+https://github.com/restic/restic/issues/2739
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -42,10 +42,13 @@ func runCat(gopts GlobalOptions, args []string) error {
 		return err
 	}
 
-	lock, err := lockRepo(gopts.ctx, repo)
-	defer unlockRepo(lock)
-	if err != nil {
-		return err
+	if !gopts.NoLock {
+		lock, err := lockRepo(gopts.ctx, repo)
+		// Make the linter happy
+		if err != nil {
+			return err
+		}
+		defer func() { _ = unlockRepo(lock) }()
 	}
 
 	tpe := args[0]
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -44,11 +44,16 @@ func runCat(gopts GlobalOptions, args []string) error {
 
 	if !gopts.NoLock {
 		lock, err := lockRepo(gopts.ctx, repo)
-		// Make the linter happy
 		if err != nil {
 			return err
 		}
-		defer func() { _ = unlockRepo(lock) }()
+
+		defer func() {
+			err := unlockRepo(lock)
+			if err != nil {
+				Warnf("unlock repo failed: %v", err)
+			}
+		}()
 	}
 
 	tpe := args[0]


metalsp0rk:
Briefly tested, no issues.  This is a tiny change, so I don't expect it to do anything particularly crazy.

metalsp0rk:
Lint failure:

> Error: Error return value of `unlockRepo` is not checked (errcheck)

cmd/restic/cmd_cat.go
```diff
@@ -42,10 +42,12 @@ func runCat(gopts GlobalOptions, args []string) error {
 		return err
 	}
 
-	lock, err := lockRepo(gopts.ctx, repo)
-	defer unlockRepo(lock)
-	if err != nil {
-		return err
+	if !gopts.NoLock {
+		lock, err := lockRepo(gopts.ctx, repo)
+		defer unlockRepo(lock)
```
greatroar, line 47: The lint failure can be solved by doing

    defer func() { _ = unlockRepo(lock) }()

Though honestly, I think this check should be disabled. It leads to lots of "explicit error ignoring", which IMHO just hurts readability rather than improving the codebase.

changelog/unreleased/issue-2739
```diff
@@ -0,0 +1,5 @@
+Bugfix: Make cat command respect --no-lock
+
+Restic cat would not respect the --no-lock flag, this change fixes that issue.
```
greatroar, line 3: ", this change fixes that issue" => ". It now does". The reader of the changelog is not looking at the commit log.

cmd/restic/cmd_cat.go
metalsp0rk: I looked around and the way I have it above is how it's replicated in all other instances of repo locking.

I agree that this check is not helpful.

changelog/unreleased/issue-2739
metalsp0rk: Fixed.

cmd/restic/cmd_cat.go
greatroar: Alternatively, unlockRepo could print the error (with Warnf) instead of returning it.

aawsome: I also got the same error in #2876. I also vote for the `Warnf` solution!

metalsp0rk: I think warnf would be a preferable solution.

cmd/restic/cmd_cat.go
```diff
@@ -42,10 +42,13 @@ func runCat(gopts GlobalOptions, args []string) error {
...
 
-	lock, err := lockRepo(gopts.ctx, repo)
-	defer unlockRepo(lock)
-	if err != nil {
-		return err
+	if !gopts.NoLock {
+		lock, err := lockRepo(gopts.ctx, repo)
+		// Make the linter happy
+		defer func() { _ = unlockRepo(lock) }()
+		if err != nil {
```
fd0, line 49: The error check should happen before the `defer unlock()`, apart from that it's fine.

fd0 (changes requested):

fd0:
By the way: we've only recently started running the linter which complained about the error check. With a non-trivial amount of code it's not possible to fix every issue at once, so for now it'll only complain about code changed in a PR. I hope that over time the codebase improves :)

cmd/restic/cmd_cat.go
metalsp0rk: Done.

metalsp0rk:
Yeah, I assumed that's what happened there.

From what I've seen so far, the codebase is fairly solid.  I'll echo that hope for improvement though.  :+1:

fd0:
I've squash the commits and added code which warns when the unlock fails. We should fix that everywhere, but I don't have the time to do that right now.

fd0 (approved):

Issue #3114, opened 2020-11-18
prune: Enhance support for duplicate blobs


aawsome:
Output of `restic version`
--------------------------

restic 0.11.0 (v0.11.0-120-gcb5ec7ea) compiled with go1.14.7 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

In the reimplementation of `prune` (#2718) the handling of duplicate blobs was implemented, but in a very poor way: All packs that contain duplicates are marked for repacking when using the prune default options.

I think there are quite some enhancements possible:

- Keep some packs with duplicates if these duplicates only contribute to the limit allowed by `--max-unused`
- If prune knows that a pack only contains unused blobs and duplicates and all the duplicates are also contained in a pack that is either kept or marked for repacking, then this pack can be simply deleted without trying to "repack" (which would download it, read every blob and realize that none of the blobs are needed) it.
- If duplicate blobs exists and one of the duplicates is broken, use another duplicate.
- Make correct statistics even if not all duplicates are repacked, see #3113 

What are you trying to do? What problem would this solve?
---------------------------------------------------------

Speed up prune even more in the case of duplicates.
IMHO this is also a prerequisite for lock-free pruning, like #1141 

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Makes me happy every day :smile:

PR #3139, opened 2020-11-30
prune: Add healing of repository in some situations

aawsome:

Pruning fails if the index does not match the pack files in the repository.
There are however two situations, that `prune` can easily detect and fix:

- missing pack file whose contents are not needed
- corrupted pack file whose contents are not needed

Actually users need to try around with `check`, `rebuild-index` and `prune` in case of a broken repository which resembles a bit of [Find Permit 38 A](https://www.youtube.com/watch?v=GI5kwSap9Ug).

With this PR, `prune` just heals the above situations.

Note that all duplicate blobs are still treated as "needed". This could be further improved when #3114 is fixed.

Not directly, but we are seeing lots of broken repo/prune issues.
I hope that with this PR, there will be more users able to tackle those problems by themselves.

Checklist
---------


cmd/restic/cmd_prune.go
```diff
@@ -316,7 +316,9 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 			return nil
 		}
 
-		if p.unusedSize+p.usedSize != uint64(packSize) {
+		if p.unusedSize+p.usedSize != uint64(packSize) &&
+			(p.usedBlobs != 0 || p.duplicateBlobs != 0) {
```
MichaelEischer, line 320: I'd prefer `!(p.usedBlobs == 0 && p.duplicateBlobs == 0)`. That way it's more obvious that this corresponds to the "remove pack"-case later on. Please also add a comment, which states that it is permitted to remove a damaged but unnecessary pack file. (The main difference to the current comment is that I'm asking for why a certain behavior is _correct_ and not whether a certain case is an error.)

cmd/restic/cmd_prune.go
```diff
@@ -362,10 +364,24 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 		return err
 	}
 
+	// packs that are referenced in the index but are not needed can be ignored
```
MichaelEischer, line 367: I think it would be helpful to add a remark to the comment that at this point `indexPack` only contains pack files missing from the repository.

cmd/restic/cmd_prune.go
```diff
@@ -362,10 +364,24 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
...
+		}
+	}
+
 	if len(indexPack) != 0 {
-		Warnf("The index references pack files which are missing from the repository: %v\n", indexPack)
+		Warnf("The index references needed pack files which are missing from the repository: %v\n", indexPack)
 		return errorPacksMissing
 	}
+	if len(ignorePacks) != 0 {
+		Verbosef("more pack files referenced in the index than present in the repository\n")
```
MichaelEischer, line 383: What is a user supposed to do when reading that warning? Maybe tell the user that prune will repair the damage as these packs are not necessary.

MichaelEischer (changes requested):
AFAICT this type of repository repair was also implicitly in the old prune implementation? There the initial `rebuild-index` step should also have removed missing/broken pack files.

I've only found a few cosmetic issues.

aawsome:
@MichaelEischer thanks for your feedback! I changed everything. Also this needed a rebase after #3148 was merged.

> AFAICT this type of repository repair was also implicitly in the old prune implementation? There the initial `rebuild-index` step should also have removed missing/broken pack files.

Yes, exactly. There is the case that the index does not match the actual pack contents which was also "cured" in some situations (like completely missing index) by previous prune. The new prune now prints an error and tells users to run `rebuild-index` manually in this case.

MichaelEischer (approved):
LGTM.

PR #3141, opened 2020-11-30
fuse: Properly set uid/gid for directories

### LINKED ISSUE #2563: Wrong Permissions when restoring backup

MichaelEischer:
In #2584 the uid/gid of a directory was changed to use the uid/gid of the root node. This
would be okay for the top-level directory of a snapshot, however, this
change also applied to normal directories within a snapshot. This
change reverts the problematic part and adds a test that directory
attributes are represented correctly.

I'm not entirely sure that the top-level node of a snapshot has a reasonable uid/gid set, but it should work as this simply reverts to the old behavior.

See https://github.com/restic/restic/issues/2563#issuecomment-731863302 .
Fixes #2563.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2563
+++ b/changelog/unreleased/issue-2563
@@ -1,0 +1,8 @@
+Bugfix: Use correct directory owner in fuse mount
+
+Restic 0.10.0 changed the fuse mount to always report the current user
+as the owner of directories within the fuse mount, which is incorrect.
+We have changed this back to use the correct owner of a directory.
+
+https://github.com/restic/restic/issues/2563
+https://github.com/restic/restic/pull/3141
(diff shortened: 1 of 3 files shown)


greatroar:
I tested this and it seems to work fine: when I mount a repo that contains directories owned by root with mode 700, I can still access those (because FUSE [leaves permission checking to restic](https://www.kernel.org/doc/html/latest/filesystems/fuse.html)).

fd0 (approved):
LGTM, thanks!

PR #3148, opened 2020-12-05
rebuild-index: code simplification

aawsome:
Simplifies the code a bit, the two cases with and without `--read-all-packs` now use almost the same code.
Also simplifies the counting of packs for the index generation both in `rebuild-index` and in `prune`.

This PR conflicts with #3140 and should preferably be merged first as it allows easier code separation.

No.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_rebuild_index.go
+++ b/cmd/restic/cmd_rebuild_index.go
@@ -60,7 +60,6 @@ func rebuildIndex(opts RebuildIndexOptions, gopts GlobalOptions, repo *repositor
 	var obsoleteIndexes restic.IDs
 	packSizeFromList := make(map[restic.ID]int64)
 	removePacks := restic.NewIDSet()
-	totalPacks := 0
 
 	if opts.ReadAllPacks {
 		// get old index files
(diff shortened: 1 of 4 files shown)


aawsome:
I introduced the very subtile bug in #3006. Was actually a bad idea to use different variables in the two `List()` calls and the copy-paste code...

cmd/restic/cmd_rebuild_index.go
```diff
@@ -97,7 +96,6 @@ func rebuildIndex(opts RebuildIndexOptions, gopts GlobalOptions, repo *repositor
 			if !ok || size != packSize {
 				// Pack was not referenced in index or size does not match
 				packSizeFromList[id] = packSize
-				removePacks.Insert(id)
```
MichaelEischer, line 100: Shouldn't rebuild-index completely remove a broken index entry for a pack file?

aawsome: You are right - I had `!ok` case in mind.
I make a new PR for the bugfix, so I can take my time on the code simplifications...

aawsome: is corrected now.

aawsome:
I think this PR is ready for review. However, I also noticed that we need better tests for `rebuild-index`. I'll work on this but so far didn't find enough time.

We may either wait for me to add more tests here or - if the changes in this PR are ready for merging - I can add tests in a new PR.

MichaelEischer (approved):
LGTM. I've slightly tweaked the comment for the `Packs` method and rebased the commits to the current main branch.

PR #3162, opened 2020-12-13
Stop retry if disk is full

### LINKED ISSUE #2453: Restic check: no space left on device

**Log**
```
Load(<index/7e8936465a>, 0, 0) returned error, retrying after 552.330144ms: Copy: write /tmp/restic-check-cache-311239722/30afc1b0f86ae2859baa4dbc3aaaa661198
d492901999ec22bb1784be441da90/index/7e/7e8936465ae6ad8b7ac488324f88688ff504d0d9d7448d40aba3f6f932345759: no space left on device
Load(<index/7e8936465a>, 0, 0) returned error, retrying after 1.080381816s: Copy: write /tmp/restic-check-cache-311239722/30afc1b0f86ae2859baa4dbc3aaaa661198
d492901999ec22bb1784be441da90/index/7e/7e8936465ae6ad8b7ac488324f88688ff504d0d9d7448d40aba3f6f932345759: no space left on device
Load(<index/7e8936465a>, 0, 0) returned error, retrying after 1.31013006s: Copy: write /tmp/restic-check-cache-311239722/30afc1b0f86ae2859baa4dbc3aaaa661198d
492901999ec22bb1784be441da90/index/7e/7e8936465ae6ad8b7ac488324f88688ff504d0d9d7448d40aba3f6f932345759: no space left on device
Load(<index/7e8936465a>, 0, 0) returned error, retrying after 1.582392691s: Copy: write /tmp/restic-check-cache-311239722/30afc1b0f86ae2859baa4dbc3aaaa661198
d492901999ec22bb1784be441da90/index/7e/7e8936465ae6ad8b7ac488324f88688ff504d0d9d7448d40aba3f6f932345759: no space left on device
```

**Describe what you'd like restic to do differently.**
I understand how check works and that it needs tmp space to do that. I also know that it can be fixed with chaning tmp dir.
But it would be great - if possible - restic checks first if there is enough space to do that or exits with error. Or that it stops during check with an error if no space is left instead of retrying (forever?)

**Version**
restic 0.9.5 compiled with go1.12.4 on linux/386

**Did restic help you or made you happy in any way?**
Yes, this is a great piece of software with friendly developers and helpful community.
Please continue being awesome!

tWido:
Stop retry cycle in RetryBackend if error was caused by low space on disk.

Fixes #2453

Function given to RetryBackend is wrapped with extra check that causes retry cycle to stop if error is classified as unrecoverable.  But checking error message doesn't seem right. Creating specialized error types for filesystem, network, permissions, etc. seems like good idea, but requires changes throughout whole project. 
Any advice or ideas?

Checklist
---------

### DIFF
--- a/internal/backend/backend_retry.go
+++ b/internal/backend/backend_retry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -32,8 +33,21 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 	}
 }
 
+func isRecoverable(err error) bool {
+	return !strings.HasSuffix(err.Error(), "no space left on device")
+}
+
 func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) error {
-	err := backoff.RetryNotify(f,
+	nf := func() error {
+		err := f()
+		if err != nil && !isRecoverable(err) {
+			be.Report(msg, err, 0)
+			return backoff.Permanent(err)
+		}
+		return err
+	}
+
+	err := backoff.RetryNotify(nf,
 		backoff.WithContext(backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(be.MaxTries)), ctx),
 		func(err error, d time.Duration) {
 			if be.Report != nil {
--- a/internal/backend/backend_retry.go
+++ b/internal/backend/backend_retry.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 )
 
@@ -33,14 +33,10 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 	}
 }
 
-func isRecoverable(err error) bool {
-	return !strings.HasSuffix(err.Error(), "no space left on device")
-}
-
 func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) error {
 	nf := func() error {
 		err := f()
-		if err != nil && !isRecoverable(err) {
+		if err != nil && errors.IsFatal(errors.Cause(err)) {
 			be.Report(msg, err, 0)
 			return backoff.Permanent(err)
 		}
--- a/internal/cache/file.go
+++ b/internal/cache/file.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/restic/restic/internal/crypto"
 	"github.com/restic/restic/internal/debug"
+	resticErrors "github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/restic"
 )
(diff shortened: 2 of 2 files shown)


internal/backend/backend_retry.go
```diff
@@ -32,8 +33,21 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 	}
 }
 
+func isRecoverable(err error) bool {
+	return !strings.HasSuffix(err.Error(), "no space left on device")
```
greatroar, line 37: I think this is a good idea in general, but I'm not sure the check belongs here, rather than in the individual backends. I know that would tie knowledge of the retrying logic into the backends, but there are many conditions that are not recoverable (dropped connections being the foremost) and enumerating them here would become messy. It would also allow a more reliable check than formatting the error and checking for string patterns.

fd0: I agree. The backend should wrap the error (e.g. within a struct) so that the returned error has a new method (e.g. `IsFatal()` or so), so we can test that in the retry logic. Unfortunately, it's not so simple :)

greatroar: The backoff package actually has an error type for this (backoff.PermanentError), but I found that installing that type is a non-trivial undertaking.

tWido: Used Fatal error to break cycle instead of custom function. Errors in mkdir and io.Copy are considered fatal.

internal/backend/backend_retry.go
```diff
@@ -33,7 +34,16 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 }
 
 func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) error {
-	err := backoff.RetryNotify(f,
+	nf := func() error {
```
MichaelEischer, line 37: This doesn't work at all. The `RetryBackend` is wrapped inside the `Cache`.

internal/backend/backend_retry.go
```diff
@@ -33,7 +34,16 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
...
+		err := f()
+		if err != nil && errors.IsFatal(errors.Cause(err)) {
+			be.Report(msg, err, 0)
+			return backoff.Permanent(err)
```
MichaelEischer, line 41: Why not simply use `backoff.Permanent(err)` when initially returning an error, instead of hot-patching it here?

internal/cache/file.go
```diff
@@ -94,7 +95,7 @@ func (c *Cache) saveWriter(h restic.Handle) (io.WriteCloser, error) {
 	p := c.filename(h)
 	err := fs.MkdirAll(filepath.Dir(p), 0700)
 	if err != nil {
-		return nil, errors.WithStack(err)
+		return nil, resticErrors.Fatal(errors.WithStack(err).Error())
```
MichaelEischer, line 98: The save method of the `Cache` itself is never retried.

MichaelEischer (changes requested):

internal/backend/backend_retry.go
tWido: Didn't want to use package specific error throughout project, when there already is project specific error Fatal.

tWido:
Closing PR since #3170 is solving issue in more general way.

dbrand666:
Did it though? I'm still seeing the same problem in restic 0.16.0.

MichaelEischer:
@dbrand666 Please open a new issue and describe the issue you're seeing there.

dbrand666:
https://github.com/restic/restic/issues/4558

PR #3164, opened 2020-12-13
Improve context cancel handling in archiver and backends

### LINKED ISSUE #3151: Receiving error "error for tree [null]
Output of `restic version`

❯ restic version
`restic 0.11.0 compiled with go1.15.3 on linux/amd64`

From my script, backup works but stats and prune don't.
```
export RESTIC_REPOSITORY="/run/media/marcelo/THEGREY/BACKUPS/tars_linux/" && export RESTIC_PASSWORD_COMMAND="\gpg --quiet --decrypt /home/marcelo/Projects/rest-o/pwd.restic.txt.gpg" 
\restic backup --repo "${RESTIC_REPOSITORY}" --force --verbose --exclude-file="${RESTIC_EXCLUDED:-}" --files-from "${RESTIC_INCLUDED}" "${RESTIC_USER_HOME:-}"
restic prune [breaks]
restic stats [breaks]
```

`export RESTIC_REPOSITORY="/run/media/marcelo/THEGREY/BACKUPS/tars_linux/" && export RESTIC_PASSWORD_COMMAND="\gpg --quiet --decrypt /home/marcelo/Projects/rest-o/pwd.restic.txt.gpg" `

Expected behaviour
Removal of unneeded data with the usage of prune command and get statistics from the repository with the stats command.
The check command is returning various blobs errors

Actual behaviour
```
❯ restic prune --verbose 3
repository 6d9f087d opened successfully, password is correct
counting files in repo
building new index for repo
[0:00] 100.00%  11878 / 11878 packs
repository contains 11878 packs (408993 blobs) with 58.120 GiB
processed 408993 blobs: 0 duplicate blobs, 0 B duplicate
load all snapshots
find data that is still in use for 51 snapshots
[0:21] 64.71%  33 / 51 snapshots
id 0000000000000000000000000000000000000000000000000000000000000000 not found in repository
github.com/restic/restic/internal/repository.(*Repository).LoadBlob
	/restic/internal/repository/repository.go:156
	/restic/internal/repository/repository.go:723
	/restic/internal/restic/find.go:19
main.getUsedBlobs
	/restic/cmd/restic/cmd_prune.go:276
main.pruneRepository
	/restic/cmd/restic/cmd_prune.go:158
main.runPrune
	/restic/cmd/restic/cmd_prune.go:62
main.glob..func19
	/restic/cmd/restic/cmd_prune.go:27
	/home/build/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826
	/home/build/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914
	/home/build/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main
	/restic/cmd/restic/main.go:98
runtime.main
	/usr/local/go/src/runtime/proc.go:204
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1374
```

```
❯ restic stats
repository 6d9f087d opened successfully, password is correct
scanning...
error walking snapshot: walking tree 0000000000000000000000000000000000000000000000000000000000000000: id 0000000000000000000000000000000000000000000000000000000000000000 not found in repository

```

```
❯ restic check
using temporary cache in /tmp/restic-check-cache-801153215
repository 6d9f087d opened successfully, password is correct
created new cache in /tmp/restic-check-cache-801153215
load indexes
check all packs
error for tree [null]:
  id 0000000000000000000000000000000000000000000000000000000000000000 not found in repository
```
_restic check --check-unused_ finds various blobs, more than my terminal can show me without a less/more piped. And issues an error at the end:
...
```
unused blob <data/63a863fd>
unused blob <data/66d14b00>
unused blob <data/dcc78d75>
unused blob <tree/20c594fa>
unused blob <data/158204f3>
unused blob <data/3ff78ad6>
unused blob <data/7e6034c5>
unused blob <data/d3110c55>
unused blob <data/9a1e75c9>
unused blob <tree/49b7fcff>
unused blob <data/3c41fe3e>
unused blob <data/5420f020>
unused blob <data/fab494e9>
unused blob <data/39e39119>
unused blob <data/feb2cec1>
unused blob <data/bcd2e562>
unused blob <data/cdc2d274>
unused blob <data/36e113dd>
unused blob <data/a10905b9>
unused blob <tree/9331a7b5>
unused blob <data/3ecaa4e5>
unused blob <data/56769a41>
unused blob <data/685b08db>
unused blob <data/c34fdde0>
unused blob <data/06dc8a71>
unused blob <tree/ef06d66d>
```

Steps to reproduce the behaviour
Do a backup, run stats, run prune, run check
Running from my script or manually the backup command works. The mount, find, rebuild-index commands also work.
Mounting and copying some random files from different snapshots from different months worked.

Not really. Just noticed it after updating to 0.11.0. I use a script to back up, prune, check and get the stats before and after the backup, but I do not check the report every time. I think I checked once when I updated to 0.10. Running it manually also shows the errors. I checked both my system filesystem and the external USB 3 HDD, both with BTRFS. Check with _btrfs scrub_ and _btrfs filesystem usage_

No

...

MichaelEischer:
This PR is a more comprehensive variant of #3155. It consists of two parts:
The changes to the archiver ensure that a 'context canceled' error triggered by a canceled background context won't be lost. In addition `saveTree` which is also responsible for the final tree root upload now properly reports a canceled context as error.

The second part of this PR can be seen as a "defense in depth" measure: It immediately fails all backend operations that use a canceled context. This would also prevent a snapshot to be saved if the background context was previously canceled. Currently for example most operations of the local backend still work when run with a canceled context. The rest backend would fail in that constellation. However, retries work for neither backend. 

[Edit]Checking for a canceled context as first step in the RetryBackend also has the benefit that this avoid playing whack-a-mole with possibly missing context checks in all the different backends and their methods which can have multiple code path depending on the object size.

The current inconsistent context handling could actually cause further problems down the line. My assumption (before this PR) was that a canceled context prevents further commands which take a context from making changes. Or put differently: Canceling a context should work as a sort of barrier, such that a canceled context behaves similar to a crash of restic in that there's a clear point in time which splits operations in before (have been completed by now), currently active (might or might not succeed) and after (always fail; in case of a crash: are never issued). Properly checking all errors is currently necessary to give that guarantee, however, a single missing error check / handling can cause problems. Preventing further repository modifications provides a generic solution to stop such bugs from causing permanent repository damage.[/Edit]

Either part of this PR is enough to fix the problem from #3151. There are currently no new test / changelog as I want to get feedback on the approach first.

Fixes #3151.
The approach itself hasn't been discussed yet.

Checklist
---------

### DIFF
--- a/internal/backend/backend_retry.go
+++ b/internal/backend/backend_retry.go
@@ -33,6 +33,16 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 }
 
 func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) error {
+	// Don't do anything when called with an already cancelled context. There would be
+	// no retries in that case either, so be consistent and abort always.
+	// This enforces a strict contract for backend methods: Using a cancelled context
+	// will prevent any backup repository modifications. This simplifies ensuring that
+	// a backup repository is not modified any further after a context was cancelled.
+	// The 'local' backend for example does not provide this guarantee on its own.
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
 	err := backoff.RetryNotify(f,
 		backoff.WithContext(backoff.WithMaxRetries(backoff.NewExponentialBackOff(), uint64(be.MaxTries)), ctx),
 		func(err error, d time.Duration) {
--- a/changelog/unreleased/issue-3151
+++ b/changelog/unreleased/issue-3151
@@ -1,0 +1,9 @@
+Bugfix: Never create invalid snapshots on backup interruption
+
+When canceling a backup run in the wrong moment it was possible that
+restic created a snapshot with an invalid "null" tree. This caused
+check and other operations to fail. The backup command now properly
+handles interruptions and never saves a snapshot in that case.
+
+https://github.com/restic/restic/issues/3151
+https://github.com/restic/restic/pull/3164
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -803,7 +803,8 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 	t.Kill(nil)
 	werr := t.Wait()
 	debug.Log("err is %v, werr is %v", err, werr)
-	if err == nil || errors.Cause(err) == context.Canceled {
+	// Use werr when it might contain a more specific error than "context canceled"
+	if err == nil || (errors.Cause(err) == context.Canceled && werr != nil) {
 		err = werr
 	}
 
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -178,6 +178,10 @@ func (arch *Archiver) saveTree(ctx context.Context, t *restic.Tree) (restic.ID, 
 		s.TreeBlobs++
 		s.TreeSize += uint64(len(buf))
 	}
+	// The context was canceled in the meantime, res.ID() might be invalid
+	if ctx.Err() != nil {
+		return restic.ID{}, s, ctx.Err()
+	}
 	return res.ID(), s, nil
 }
 
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -3,6 +3,7 @@ package archiver
 import (
 	"bytes"
 	"context"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -15,6 +16,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/restic/restic/internal/backend/mem"
 	"github.com/restic/restic/internal/checker"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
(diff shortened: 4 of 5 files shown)


internal/backend/backend_retry.go
```diff
@@ -33,6 +33,10 @@ func NewRetryBackend(be restic.Backend, maxTries int, report func(string, error,
 }
 
 func (be *RetryBackend) retry(ctx context.Context, msg string, f func() error) error {
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
```
greatroar, line 38: Since this check isn't strictly necessary (assuming backends work properly; backoff.RetryNotify also checks the context), I think it should have a comment.

MichaelEischer: I've added a few lines of comments. However, I'm not sure whether that covers the `assuming backends work properly` part of your comment. Does "working properly" necessaryly entail checking whether a context is cancelled, even if the backend uses no cancelable functions? The context check by `RetryNotify` only happens between retries, so the first try is not covered.

greatroar: That is what I meant, but thinking about it more, I'm no longer sure it's the right definition of "working"...

MichaelEischer: I've added a sentence which mentions that the local backend doesn't prevent operations from working once a context is canceled.

> That is what I meant, but thinking about it more, I'm no longer sure it's the right definition of "working"...

My expectation was/is that the operation fails when the context was already canceled. After all, why would we pass a context to the method otherwise?

fd0 (approved):
This is great work, thanks a lot for digging in!

Do you have an idea how to test this? How deep should we go?

I'll already approve the PR now, from my point, the only thing we're missing right now is a changelog entry.

MichaelEischer:
@fd0 I've added a short changelog and tests for the fixed error handling in `arch.Snapshot` and the context cancellation checks in the `RetryBackend`. I haven't added a test for `arch.saveTree`, as it would require messing with the internals of the archiver.

From what I can tell the snapshots with null treeIDs could be created for the following backends: Azure, Local, SFTP and Swift. These don't (always) check for context cancellations.

fd0 (approved):
LGTM!

PR #3165, opened 2020-12-15
doc: Clarify calendar boundaries for --keep-* options

### LINKED ISSUE #2747: forget: Document that time spans (week, month) are absolute
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.12.12 on linux/amd64

    restic -r /testrepo forget --keep-yearly 1 --keep-monthly 1 --keep-weekly 1 --keep-daily 1 --keep-last 1

local

Expected behavior
-----------------
Keep 1 year of snapshots.

Actual behavior
---------------
Only the last snapshot was retained.

See _How did you run restic exactly?_

Design decision of using always the _last_ snapshot. This way, specifying an option only makes sense if the value ist >=2. It would be more intuitive to have values >=1 if an option should have an effect.

For every interval, use earliest (instead of latest) snapshot.
`--keep-last` should default to 1 (this way the last snapshot will always be preserved when not overwritten). This also maintains backwards compatibility such that only _more_ and not _less_ snapshots are preserved for existing scripts after the change.

It looks like restic will be a suitable tool for my backups (open source, cloud storage backends, secure).

rawtaz:

Clarifies how weeks, days, hours, etc are bound to natural time boundaries rather than relative to when you run the `forget` command, to close #2747.

Closes #2747.

Checklist
---------

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -191,6 +191,10 @@ The ``forget`` command accepts the following parameters:
    made in the two years, five months, seven days, and three hours before the
    latest snapshot.
 
+.. note:: All calendar related ``--keep-*`` options work on the natural time
+    boundaries and not relative to when you run the ``forget`` command. Weeks
+    are Monday 00:00 -> Sunday 23:59, days 00:00 to 23:59, hours :00 to :59, etc.
+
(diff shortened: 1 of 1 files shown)


fd0 (approved):
LGTM, thanks!

PR #3168, opened 2020-12-16
Ensure handle is valid before passing it to a Backend

greatroar:

It validates handles before passing them to backends, rather than repeating the validation code inside every backend (except the wrapper types).

This is a first step toward not retrying fatal errors, which needs error inspection inside the backend (see comments at #3162). The alternative to this approach is wrapping the error from Handle.Valid in a backoff.PermanentError everywhere, which I felt was too repetitive and error-prone.

Checklist
---------

### DIFF
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -118,10 +118,6 @@ func (be *Backend) Path() string {
 
 // Save stores data in the backend at the handle.
 func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
-	if err := h.Valid(); err != nil {
-		return err
-	}
-
 	objName := be.Filename(h)
 
 	debug.Log("Save %v at %v", h, objName)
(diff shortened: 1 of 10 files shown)


MichaelEischer:
This PR is broken: The code change removes calls to `Valid()` from `Save(...)` or `openReader(...)` (one of both for each backend) and then adds one validity check to `SaveUnpacked`, however, this isn't the only caller of the `Save` method of backends (see `internal/repository/{key,packer_manager}.go`).

I also think that the caller is the wrong place to check for validity. After all why would a correctly implemented calling function ever knowingly create an invalid handle.

greatroar:
By the same token, the check is unnecessary, because there are no callers that generate invalid handles...

MichaelEischer:
> By the same token, the check is unnecessary, because there are no callers that generate invalid handles...

It's usually the called function (and not the caller!) which checks that the caller has passed valid parameters.

To deduplicate the check for a valid handle, it should rather be moved to a central place where all calls have to pass through. And that would currently be the `RetryBackend`. However, I'm not sure whether the RetryBackend should also handle that check, or whether a sanity check wrapper would be better... @fd0 Do you have a preference?

greatroar:
Closing as this is of too small importance to fix the merge conflicts.

PR #3170, opened 2020-12-16
Don't retry permanent errors in backends

### LINKED ISSUE #2453: Restic check: no space left on device

greatroar:

This PR does two things: it marks a few errors as don't-retry in the backends, and, perhaps more importantly, it introduces a convention to handle these errors in general.

To do the former, the PR has to introduce some new-style (Go 1.13) error handling with errors.Is, which is rather different from the pkg/errors style (errors.Cause) handling that restic has done so far. I considered extending internal/errors.Cause further to handle this, but I didn't, because changes in that function's behavior have consequences in the rest of the codebase that I found hard to fathom (if the Cause of a *os.PathError is no longer the error itself, do existing checks for specific errors continue to work? are they all tested?).

Fixes #2453. Closes #3162.

Checklist
---------

### DIFF
--- a/internal/restic/backend.go
+++ b/internal/restic/backend.go
@@ -6,6 +6,12 @@ import (
 )
 
 // Backend is used to store and access data.
+//
+// Backend operations that return an error will be retried when a Backend is
+// wrapped in a RetryBackend. To prevent that from happening, the operations
+// should return a github.com/cenkalti/backoff/v4.PermanentError. Errors from
+// the context package need not be wrapped, as context cancellation is checked
+// separately by the retrying logic.
 type Backend interface {
 	// Location returns a string that describes the type and location of the
 	// repository.
--- a/changelog/unreleased/issue-2453
+++ b/changelog/unreleased/issue-2453
@@ -1,0 +1,12 @@
+Enhancement: Improve handling of permanent backend errors
+
+When encountering errors in reading from or writing to storage backends,
+restic retries the failing operation up to nine times (for a total of ten
+attempts). It used to retry all backend operations, but now detects some
+permanent error conditions so it can report fatal errors earlier.
+
+Permanent failures include local disks being full and SSH connections
+dropping.
+
+https://github.com/restic/restic/issues/2453
+https://github.com/restic/restic/pull/3170
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -10,11 +10,13 @@ import (
 	"path"
 	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/cenkalti/backoff/v4"
 )
 
 // Backend stores data on an azure endpoint.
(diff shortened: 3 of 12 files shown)


internal/restic/backend.go
```diff
@@ -6,6 +6,11 @@ import (
 )
 
 // Backend is used to store and access data.
+//
+// Backend operations that return an error will be retried when a Backend is
+// wrapped in a RetryBackend. To prevent that from happening, the operations
+// should return a backoff.PermanentError. Errors from the context package
+// are always considered permanent.
```
MichaelEischer, line 13: I don't think that's the case:

According to https://github.com/cenkalti/backoff/blob/18fe4ce5a8550e0d0919b680ad3c080a5455bddf/retry.go#L56
```
if permanent, ok := err.(*PermanentError); ok {
```

As `PermanentError` is a named struct from the `backoff` package, it won't consider a context cancelled error or similar as permanent. (If the used context itself is cancelled then the retries will be aborted, however that is independent of the context error itself).

MichaelEischer (commented):
The overall approach looks good.

Is the `Ensure handle is valid before passing it to a Backend ` commit necessary for this PR? I'm not really convinced that the changes it introduces are a good idea.

internal/restic/backend.go
greatroar: You're right. I changed the comment.

greatroar:
We don't need the validation changes, but they do greatly simplify the code here. Without that patch, the validation checks would all need to wrap their errors using backoff.Permanent.

greatroar:
I've changed the PR to do the wrapping to not tie together two discussions.

MichaelEischer:
I think this PR needs a changelog, The "missing" retries are quite user visible, especially in the case when the sftp connection gets interrupted. Besides that it looks fine.

> Without that patch, the validation checks would all need to wrap their errors using backoff.Permanent.

It would also have been ok to not convert these errors into permanent ones. After all `Valid()` shouldn't fail that often. But as the code is already there, let's keep it.

greatroar:
I suppose that would have been an option too. I just couldn't resist :)

Added a changelog.

MichaelEischer (approved):
LGTM. I've tweaked the changelog summary line to mention that these errors originate from the repository storage backends.

PR #3176, opened 2020-12-18
Pass upload size to backends and sanity check it

MichaelEischer:
restic did not pass the expected file size to the Swift backend during a save operation. The Azure backend had to read the upload into memory to determine it's size. The latter is replaced by explicitly providing the file size to the backend.

The second part of the PR adds sanity checks that a file was actually uploaded completely. This could help retry some situations which would otherwise cause incomplete uploads. I've also noticed that the go had a bug where io.Copy produced empty files https://github.com/golang/go/issues/42400 .

No.

Checklist
---------

### DIFF
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/base64"
 	"io"
-	"io/ioutil"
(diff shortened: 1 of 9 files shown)


fd0 (approved):
Great idea! I've rebased it on master.

PR #3178, opened 2020-12-19
Calculate content hashes during upload

### LINKED ISSUE #2202: Cannot create/use repository in S3 with S3 Object Lock enabled
Output of `restic version`
--------------------------

    RESTIC_REPOSITORY=s3:s3.amazonaws.com/[myBucketName]/[myPrefixInBucket] \
    RESTIC_PASSWORD_FILE=/home/me/.resticpassword \
    ./restic init --json

S3, with [S3 Object Lock](https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lock.html) enabled and configured to apply to all objects.

Expected behavior
-----------------

Restic should have initialized the repository.

Actual behavior
---------------

    Fatal: create key in repository at s3:s3.amazonaws.com/[myBucketName]/[myPrefixInBucket] failed: client.PutObject: Content-MD5 HTTP header is required for Put Object requests with Object Lock parameters

Create a bucket with s3 object lock enabled and configured to apply to all objects. Here's some (partial) terraform code:

    resource "aws_s3_bucket" "bucket" {
      bucket = "${var.name}"

      versioning {
        enabled = true
      }

      lifecycle_rule {
        enabled = true
        prefix  = ""

        abort_incomplete_multipart_upload_days = "7"

        noncurrent_version_expiration {
          days = "${var.noncurrent_version_expiration_days}"
        }
      }

      object_lock_configuration {
        object_lock_enabled = "Enabled"

        rule {
          default_retention {
            mode = "COMPLIANCE"
            days = "${var.object_lock_days}"
          }
        }
      }
    }

    variable "name" {
      type    = "string"
      default = "myBucketName"
    }

    variable "noncurrent_version_expiration_days" {
      type        = "string"
      description = "In reality, this would be large, like 30-90."
      default     = "1"
    }

    variable "object_lock_days" {
      type        = "string"
      description = "In reality this would be large, like 180. Don't make this big when testing!"
      default     = "3"
    }

Restic is attempting to PUT to S3 without computing Content-MD5 and including that in the request.

Compute Content-MD5 and set the header when making requests to s3.

😐

### LINKED ISSUE #2700: restic prune failing - tree XXX not found in repository
Output of `restic version`
--------------------------
`restic 0.9.6 compiled with go1.12.12 on linux/amd64
`

```
$ sudo -E restic -r /media/auto/nas_robertfoss/restic prune
repository 237fdd10 opened successfully, password is correct
counting files in repo
building new index for repo
[3:52:27] 100.00%  440951 / 440951 packs
incomplete pack file (will be removed): 01e37cc714147b7ec051e8eb4cf045e71c69b1b10e18eb83ed9fe9338c2d24d6
incomplete pack file (will be removed): 02166c70e25a51cb5b5b44d3b9702c0a540f6ca842c19e4f78bd1774ca782957
incomplete pack file (will be removed): 02b80c6de7f81142a5e096891069209b88a4da91e1c84ae316a40f2b5238bcd4
incomplete pack file (will be removed): 0be78f6a34de32f800ad05cd49347b27ca16ccc1185e7858481faf785aa9992a
incomplete pack file (will be removed): 0d3bddd90c2748e6f1acd70ed286283d346f17d8c012c2ed606dd1aa6bac2f06
incomplete pack file (will be removed): 11a70cf10e5fc08fe6875b8c988c17af3447c99c6082f82006e9dd14a26b635e
incomplete pack file (will be removed): 11c8c8ef5e80cd654115769b87a3bf7c1d4fefaf5a4ec226209f3627b7159969
incomplete pack file (will be removed): 1efc58add323f3e1e2aae1d751d1be6f26b90d1c5a408cee3edffe1de02e4b4b
incomplete pack file (will be removed): 2d1763064c0a39abf21377bcdf056ad6f4e12a7b93cd28b3dbb98002dd91813d
incomplete pack file (will be removed): 31c299e7f80a3836a394da0795e8ed65d7277dfed9e3863db013bc0b4d99d9c7
incomplete pack file (will be removed): 3557c70b0a76e3303ad51ac755a2ab577d6c041640cd015da345308f492b563b
incomplete pack file (will be removed): 36745b2ec4fffe49ecf6a9483d74a7a7c352e077ec7b830d4d162b0d16739a6c
incomplete pack file (will be removed): 44090b5084dbe0746a68151a5980d5078666b8cd92b889032a99a4297ceca7c3
incomplete pack file (will be removed): 4433545485727ef6406307e1d18ded514ea59e6230fba5deeba28f82436cbf71
incomplete pack file (will be removed): 4b0be68275743cdf4e69dd795840050d038eb69d3f6d0f3c36cfc5d7d01d991b
incomplete pack file (will be removed): 4c4dd542a0300096250e01b86f71dd2f5a2e9e7e4b2cb9242df953ab432814cf
incomplete pack file (will be removed): 4df36fec011b6add8541e9c20064322b2292297f931fe525f46c88c69db8d752
incomplete pack file (will be removed): 4f654a34453ba4ff32aabd25013081c660ee9895cfe1482d764fa53d31f392e3
incomplete pack file (will be removed): 4fa5793656a1514e2846a4ff57aa2357333f92eb29617682760d42b72771cd8b
incomplete pack file (will be removed): 512ba62d659a52d26ce0db2e4bf09da509cac1f418fe9db151bf2a5ac830cd5d
incomplete pack file (will be removed): 59a07eafe7a187400ab744521edd9800dede1064260dad1b3f32080d090c2b5a
incomplete pack file (will be removed): 60719e8a0247033dcd8bc1c89dfd34cdf67499e423145ab568b32ac9b209e00d
incomplete pack file (will be removed): 625243d5fee5a4bdd8dbb33d9864a375b424a15c9fd7d190d244f2f1eb4e8ec3
incomplete pack file (will be removed): 64ecbbf7d6365c73a1e431b64697eed338fc8e130800dc261f268611b07b8447
incomplete pack file (will be removed): 684040807b3515b1336cc22d5ebe750e3dd0fa7a12e99d5808b1ee834a784f81
incomplete pack file (will be removed): 6a4e046d6de8f89f4e9a7a87dba5bcbf43e1b150148646a20f67130e9421ade4
incomplete pack file (will be removed): 6b6d7766ac57c254e8ed539fcb1f74d2722d024fc231abe7e5b8d26e3367ef01
incomplete pack file (will be removed): 6e05c04f66e9e8335dd4b3f5217c0edf389ce3f2cd53362273edddc7b2026c76
incomplete pack file (will be removed): 6e3407887327ac22b1271d22096b0fa817701166da8a0d7b1200b6795fd0ca3d
incomplete pack file (will be removed): 70c4133e9a9262d537e1c46eb6b09df5c9c04b8028108f3ca7d1a4b9e614f293
incomplete pack file (will be removed): 70ce970d142d08dd86bf4f4e9e1f7c2b6d5dca3655e72bea9a74c56de0608d69
incomplete pack file (will be removed): 74dac8ca2412961d27289c47562af19b62251ba83db41655a809a2aa3e04d1c4
incomplete pack file (will be removed): 780de5f377e4db9d7b5d383b530bd8e87768268f4e36d8182e9b8dcfd62f1292
incomplete pack file (will be removed): 784ace35191fc681b3a90fba99654d6b3983dc9e224f5ee96ca50101e1cc3c57
incomplete pack file (will be removed): 7e6fd087f8bcb9d77d97284c05568f0bb79db6425129a4b6796b27c115196629
incomplete pack file (will be removed): 7fe7e27c4aa423d015e85a9eb8abe7939af51ef24a0fb0c39c46d7d1f4260b8a
incomplete pack file (will be removed): 84cac8089d827106faf06a5c8bc9446573c75f0ac233749130c81b7e3538b0b6
incomplete pack file (will be removed): 899ff4b1a0d24b818b3b106d950804a747ba2a7f72676b7bd9b81c48306eaadd
incomplete pack file (will be removed): 9397dbad057de654ed96524e16db2a7101811352ebcb08d153ab96a2628720fe
incomplete pack file (will be removed): 946ad13a3cec6953448fc8947bd3f418ea1be1a9d38ac39b80dd108ad7a797f9
incomplete pack file (will be removed): 9ae6b1f02d3e4f2008bf9411213b51e697a703bfddfb5383165b1b17fceaf3bc
incomplete pack file (will be removed): 9b7662aeab336fef8dfba60399631c92069fd8514aea97d1175d92f7c8aeade8
incomplete pack file (will be removed): a0b52520a5c2b9f4f477e8c4155e269a24aaad5ac1c12c36fbf56072b28f2036
incomplete pack file (will be removed): a15b2d9db7b0955b1a0ebf4f76364b4a87ed0c830adf539be8aa592922e19ba7
incomplete pack file (will be removed): a6f12ba5da6420a4171b866a5c114ab7e558820331dfa82c5fbc909371a1c96c
incomplete pack file (will be removed): a889ced3aaba4a3b6c7536a4262b703c90244d842e0808f942819443fdd8049d
incomplete pack file (will be removed): b574b68758b3ad8f3eaf16f26c1af4e127b1a93af17e6f9b1ec5a17ad322452c
incomplete pack file (will be removed): bc36699cfcc6babb8670c6a27eef2ed23acc7f091e82bbc7102017ffab792dc6
incomplete pack file (will be removed): bd7b650d802217450570976e56e43185b36910b0a12ff5b73bf04a07a95c3c50
incomplete pack file (will be removed): bf384616635115743e79b6c25ebd0e65bc7c8b3fa27090e3719837811b753f9b
incomplete pack file (will be removed): c55c9b36a53c3953306fd4f003750b9c164c3c95d634becdd68b36a9343c45d7
incomplete pack file (will be removed): c5d74ba19f733815d2f26cf21a23cde22328d20d162bf16f8a9ff3cfa3f7a1d4
incomplete pack file (will be removed): c6d47d52ac7cb2f67c59226da1f488bcd77cb4e539be0069c9131585b2da8d64
incomplete pack file (will be removed): ce6f710e03a54ce10f747531efcdf3d6e54edbab42de62ddfb8eb955b3b0caaf
incomplete pack file (will be removed): d44fa38cf9ffb990049db2bbf233e7dee56678304326b834bd67259f71989112
incomplete pack file (will be removed): d8832cf344b82f26c10a68c65694ebd1a21e51ac117d874cdc82c84f1eb1dfd2
incomplete pack file (will be removed): dbe65af86c8eea1ec694c25180e9a2ca46f0047d7c65f4b48789eb48cafd0934
incomplete pack file (will be removed): dc6adc442ffd91c8841ef40133b45d0e27ebe2f45eda0a5ca5b698daf828261d
incomplete pack file (will be removed): de6d33643239da4809e9ff6ccdbb66f49ca31df8ea33e8378ba60fee00c751f7
incomplete pack file (will be removed): e4015a03a9b69818f2ba631f37d0cc713a3a4c8e73c1aba92327cf0bbead69b9
incomplete pack file (will be removed): ea3bf0acf2326a2c730e01fc233916686e4c1243ea327110bce1b0639b78238a
incomplete pack file (will be removed): ea896919de15181889c941a093021f90ef5b8340bc9cd894b9a91ca93b75751f
incomplete pack file (will be removed): ee0af7c8d44a05b74083bf20d9116ddb80f93f3f907da8a7132c6d5861df03b2
incomplete pack file (will be removed): ef6c72c5c9e34bb96456e53e8f646376416a6194db2260200c6764a1e9626a9f
incomplete pack file (will be removed): ef6e0530aec863e408e09860ff91121df08f3870d0a251c78b2b2cbff05a6f9b
incomplete pack file (will be removed): f003cd6ebfa976771f8d7e0a76eb41d14c92d62d962c3c74fcc67dc87769d209
incomplete pack file (will be removed): f095ad5213e7b4bae17fb627db806f41f644f8412f9ae89b8dddb1e2c57cc69c
incomplete pack file (will be removed): f460b877db2fe4148cb67826c70f757dedbd0f4be44da59536e17ca2863f82fe
incomplete pack file (will be removed): fcfc1622fe2967de167c5a6b2f1838117daf1b26dc3d11a9731d16ef04cb5188
repository contains 440882 packs (21714968 blobs) with 2.009 TiB
processed 21714968 blobs: 1546276 duplicate blobs, 53.897 GiB duplicate
load all snapshots
find data that is still in use for 42 snapshots
tree 51a481d9674fe804a2900e71f9a95ea8c1083d2167629db4549e0c9693ecf79b not found in repository
	github.com/restic/restic/internal/repository/repository.go:713
	github.com/restic/restic/internal/restic/find.go:11
main.pruneRepository
	github.com/restic/restic/cmd/restic/cmd_prune.go:191
main.runPrune
	github.com/restic/restic/cmd/restic/cmd_prune.go:85
main.glob..func18
	github.com/restic/restic/cmd/restic/cmd_prune.go:25
	github.com/spf13/cobra/command.go:762
	github.com/spf13/cobra/command.go:850
	github.com/spf13/cobra/command.go:800
main.main
	github.com/restic/restic/cmd/restic/main.go:86
runtime.main
	/usr/lib/go-1.12/src/runtime/proc.go:200
runtime.goexit
	/usr/lib/go-1.12/src/runtime/asm_amd64.s:1337
```
Local filesystem (backed by CIFS)

Expected behavior
-----------------

Prune not failing

Actual behavior
---------------

Prune failing

```
restic -r /media/auto/nas_robertfoss/restic prune
```

Likely some corruption in the repository. Not that I have any particular reasons to think it is corrupt.

No

Not today, but most other days :)

### LINKED ISSUE #3023: Data blobs seem to be missing, aborting prune to prevent further data loss!
# [Edit by @MichaelEischer] **See [this comment](https://github.com/restic/restic/issues/3023#issuecomment-724978137) for instructions to recover damaged packs.**

repository 575e9064 opened successfully, password is correct
counting files in repo
building new index for repo
[0:01] 100.00%  101845 / 101845 packs
repository contains 101845 packs (1070217 blobs) with 424.769 GiB
processed 1070217 blobs: 0 duplicate blobs, 0 B duplicate
load all snapshots
find data that is still in use for 29 snapshots
[1:17] 100.00%  29 / 29 snapshots
Fatal: [<data/a5c66b46> <data/0263dd34> <data/c30d23a3> <data/b6219ef4> <data/5d320c9c> <data/47effa2a> <data/9d6ee6a3> <data/0d681206> <data/43902c2b> <data/a0057fa2> <data/f91d6a50> <data/7cba5fe6> <data/355d43a3> <data/2e9feb61> <data/c87cc3ed> <data/8104ef95> <data/128ee5a4> <data/5b75af0b> <data/94633e1c> <data/ec6aa989> <data/f876e05c> <data/be58954b> <data/c9ffe20e> <data/991c4a30> <data/48594d83> <data/2dc03dbc> <data/6d8fe121> <data/09840332> <data/33b40ab4> <data/222e85bd> <data/dd3bd177> <data/c6bc4445> <data/e5cf426d> <data/812e4cee> <data/4c99865e> <data/edeaff56> <data/d81d207c> <data/674fd0cd> <data/2a8e8c07> <data/22bebce6> <data/0ac47ee6> <data/5c0cd1a2> <data/6aa4f020> <data/868c7214> <data/be560cf2> <data/7330f83c> <data/88a3ff3b> <data/41f8a329> <data/f2dc6ab9> <data/13f65cd0> <data/cb9eadb7> <data/0a733e91> <data/e66af3e4> <data/b90c76b6> <data/2bcb6e6f> <data/f273499b> <data/294cc8f5> <data/301f2084> <data/7adf05af> <data/e3a60adb> <data/58ce1d25> <data/9fe36716> <data/020dbd1d> <data/9ac14d3a> <data/1d70f084> <data/90752236> <data/ad3a6190> <data/91da8cc7> <data/c00ebb0e> <data/cac4a750> <data/04d91469> <data/67eaea1b> <data/5e9d64b5> <data/bf86fc21> <data/e3c6e99b> <data/242ec3b3> <data/35354808> <data/9caa443c> <data/daa96050> <data/767638bc> <data/d5b17efd> <data/1cad7845> <data/92e59720> <data/cbb2db03> <data/dd91ca2d> <data/44d44449> <data/acb667c2> <data/025b7630> <data/acecc383> <data/eea62998> <data/3031267e> <data/b841096c> <data/c9000183> <data/506e076d> <data/a2eb0732> <data/bf0b9874> <data/3b1c6115> <data/f2984878> <data/026282d0> <data/429e1cf9> <data/027f0e01> <data/d4bf6073> <data/1877465e> <data/123aad11> <data/a57a748e> <data/b46dc499> <data/2c266d43> <data/cf08e3f6> <data/78650f91> <data/a691b49b> <data/32e22aed> <data/21dc4ae2> <data/dce17c97> <data/03bd6674> <data/ce5c53d5> <data/5fe04c53> <data/b06eafd3> <data/b666ffd6> <data/916bb531> <data/ce658f67> <data/f6ab04f9> <data/84e43d89>] not found in the new index
Data blobs seem to be missing, aborting prune to prevent further data loss!
Please report this error (along with the output of the 'prune' run) at
https://github.com/restic/restic/issues/new/choose

MichaelEischer:
Up to now only the B2 backend and to a lesser extend the Swift backend have checked the integrity of uploaded files. For B2 the blazer library calculates SHA1 hashes for the uploaded files and provides that information to B2. The server-side of B2 then verifies that the uploaded data matches the hashes and rejects the upload in case of a mismatch. This effectively rules out any data corruption during upload _and_ storage. The latter point probably needs some explanation. As the backend now knows the correct hash of a file it is able to check the integrity of the stored data from time to time. The hash also allows detecting hardware problems causing bitflips within the storage system or data corruption before it is stored.

For Swift the library was calculating the MD5 hash of the data during the upload and compared it with the value returned by the server. Compared to the variant used for B2 it has the benefit of being simpler to calculate on-the-fly but has several drawbacks: It is necessary to delete the broken upload after detecting the problem. This could be impossible when e.g. using an append-only backend or when the delete operation fails due to network problems. Depending on the consistency guarantees of a backend (and bugs) it could be possible that the broken upload can show up again for some time. This cannot happen, if the broken upload is rejected right from the start.

This PR enables (among others) the calculation of the content hash for S3 in the minio library. The calculation happens inside the library which could increase the memory usage a bit compared to before. As the content hash is included in the request authenticator, this also prevents any intermediate load balancers etc. from damaging the upload.

The main change of this PR is that it enhances the `RewindReader` to also return the hash for the uploaded data when requested by the backend. For that the backends can return a `hash.Hash` object via a new `Hasher()` method which is then used during pack assembly to calculate the content hash. Calculating the hash during pack assembly has the big benefit that it calculates the hash at the earliest point in time possible, and thus would also detect a damaged temporary pack file.

The azure, gs, swift and mem (for testing) backends now use that mechanism with the MD5 hash to ensure the integrity of the uploaded data.

The content hash also effectively prevents the creation of incomplete pack files due to interrupted upload for all supported backends.

The following backends are not affected by this PR:
- REST / rclone: The REST server could easily check the sha256 hash of the uploaded data. It would also be possible to provide the md5 content hash, but that probably not too useful. For rclone that could be useful however. @ncw Would rclone be able to use the sha256/md5 sum to help the backends verify the upload (md5 would cover quite a lot of storage backends according to https://rclone.org/overview/)?
- local / sftp: As there is no "smart" storage system on the other side there's not much that can be done here. I mean it would be possible to check that the generated pack file data matches the data which is later on written to the repository. However, I'm not sure whether that's worth the trouble.

Fixes #2202.
Fixes #2700.
Fixes #3023.

Checklist
---------

### DIFF
--- a/internal/backend/mem/mem_backend.go
+++ b/internal/backend/mem/mem_backend.go
@@ -3,6 +3,7 @@ package mem
 import (
 	"bytes"
 	"context"
+	"hash"
 	"io"
 	"io/ioutil"
 	"sync"
@@ -212,6 +213,11 @@ func (be *MemoryBackend) List(ctx context.Context, t restic.FileType, fn func(re
 // Location returns the location of the backend (RAM).
 func (be *MemoryBackend) Location() string {
 	return "RAM"
+}
+
+// Hasher may return a hash function for calculating a content hash for the backend
+func (be *MemoryBackend) Hasher() hash.Hash {
+	return nil
 }
 
 // Delete removes all data in the backend.
--- a/internal/backend/mem/mem_backend.go
+++ b/internal/backend/mem/mem_backend.go
@@ -86,7 +86,10 @@ func (be *MemoryBackend) Save(ctx context.Context, h restic.Handle, rd restic.Re
 
 	beHash := be.Hasher()
 	// must never fail according to interface
-	_, _ = beHash.Write(buf)
+	_, err = beHash.Write(buf)
+	if err != nil {
+		panic(err)
+	}
 	if !bytes.Equal(beHash.Sum(nil), rd.Hash()) {
 		return errors.Errorf("invalid file hash or content, got %s expected %s",
 			base64.RawStdEncoding.EncodeToString(beHash.Sum(nil)),
--- a/internal/backend/mem/mem_backend.go
+++ b/internal/backend/mem/mem_backend.go
@@ -3,6 +3,8 @@ package mem
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
+	"encoding/base64"
 	"hash"
 	"io"
 	"io/ioutil"
@@ -80,6 +82,16 @@ func (be *MemoryBackend) Save(ctx context.Context, h restic.Handle, rd restic.Re
 	buf, err := ioutil.ReadAll(rd)
 	if err != nil {
 		return err
+	}
+
+	beHash := be.Hasher()
+	// must never fail according to interface
+	_, _ = beHash.Write(buf)
+	if !bytes.Equal(beHash.Sum(nil), rd.Hash()) {
+		return errors.Errorf("invalid file hash or content, got %s expected %s",
+			base64.RawStdEncoding.EncodeToString(beHash.Sum(nil)),
+			base64.RawStdEncoding.EncodeToString(rd.Hash()),
+		)
 	}
 
 	be.data[h] = buf
@@ -217,7 +229,7 @@ func (be *MemoryBackend) Location() string {
 
 // Hasher may return a hash function for calculating a content hash for the backend
 func (be *MemoryBackend) Hasher() hash.Hash {
-	return nil
+	return md5.New()
 }
 
 // Delete removes all data in the backend.
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -605,6 +605,10 @@ func (s *Suite) TestSaveError(t *testing.T) {
 	h := restic.Handle{Type: restic.PackFile, Name: id.String()}
 	err := b.Save(context.TODO(), h, &incompleteByteReader{ByteReader: *restic.NewByteReader(data, b.Hasher())})
 	if err == nil {
+		err = s.delayedRemove(t, b, h)
+		if err != nil {
+			t.Fatalf("error removing item: %+v", err)
+		}
 		t.Fatal("incomplete upload did not fail")
 	}
 }
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -609,6 +609,47 @@ func (s *Suite) TestSaveError(t *testing.T) {
 	}
 }
 
+type wrongByteReader struct {
+	restic.ByteReader
+}
+
+func (b *wrongByteReader) Hash() []byte {
+	h := b.ByteReader.Hash()
+	modHash := make([]byte, len(h))
+	copy(modHash, h)
+	// flip a bit in the hash
+	modHash[0] ^= 0x01
+	return modHash
+}
+
+// TestSaveWrongHash tests that uploads with a wrong hash fail
+func (s *Suite) TestSaveWrongHash(t *testing.T) {
+	seedRand(t)
+
+	b := s.open(t)
+	defer s.close(t, b)
+	// nothing to do if the backend doesn't support external hashes
+	if b.Hasher() == nil {
+		return
+	}
+
+	length := rand.Intn(1<<23) + 200000
+	data := test.Random(23, length)
+	var id restic.ID
+	copy(id[:], data)
+
+	// test that upload with hash mismatch fails
+	h := restic.Handle{Type: restic.PackFile, Name: id.String()}
+	err := b.Save(context.TODO(), h, &wrongByteReader{ByteReader: *restic.NewByteReader(data, b.Hasher())})
+	if err == nil {
+		err = s.delayedRemove(t, b, h)
+		if err != nil {
+			t.Fatalf("error removing item: %+v", err)
+		}
+		t.Fatal("upload with wrong hash did not fail")
+	}
+}
+
 var filenameTests = []struct {
 	name string
 	data string
--- a/internal/backend/test/tests.go
+++ b/internal/backend/test/tests.go
@@ -84,7 +84,7 @@ func (s *Suite) TestConfig(t *testing.T) {
 		t.Fatalf("did not get expected error for non-existing config")
 	}
 
-	err = b.Save(context.TODO(), restic.Handle{Type: restic.ConfigFile}, restic.NewByteReader([]byte(testString)))
+	err = b.Save(context.TODO(), restic.Handle{Type: restic.ConfigFile}, restic.NewByteReader([]byte(testString), b.Hasher()))
 	if err != nil {
 		t.Fatalf("Save() error: %+v", err)
 	}
@@ -134,7 +134,7 @@ func (s *Suite) TestLoad(t *testing.T) {
 	id := restic.Hash(data)
 
 	handle := restic.Handle{Type: restic.PackFile, Name: id.String()}
-	err = b.Save(context.TODO(), handle, restic.NewByteReader(data))
+	err = b.Save(context.TODO(), handle, restic.NewByteReader(data, b.Hasher()))
 	if err != nil {
 		t.Fatalf("Save() error: %+v", err)
 	}
@@ -253,7 +253,7 @@ func (s *Suite) TestList(t *testing.T) {
 		data := test.Random(rand.Int(), rand.Intn(100)+55)
 		id := restic.Hash(data)
 		h := restic.Handle{Type: restic.PackFile, Name: id.String()}
-		err := b.Save(context.TODO(), h, restic.NewByteReader(data))
+		err := b.Save(context.TODO(), h, restic.NewByteReader(data, b.Hasher()))
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -343,7 +343,7 @@ func (s *Suite) TestListCancel(t *testing.T) {
 		data := []byte(fmt.Sprintf("random test blob %v", i))
 		id := restic.Hash(data)
 		h := restic.Handle{Type: restic.PackFile, Name: id.String()}
-		err := b.Save(context.TODO(), h, restic.NewByteReader(data))
+		err := b.Save(context.TODO(), h, restic.NewByteReader(data, b.Hasher()))
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -447,6 +447,7 @@ type errorCloser struct {
 	io.ReadSeeker
 	l int64
 	t testing.TB
+	h []byte
 }
 
 func (ec errorCloser) Close() error {
@@ -456,6 +457,10 @@ func (ec errorCloser) Close() error {
 
 func (ec errorCloser) Length() int64 {
 	return ec.l
+}
+
+func (ec errorCloser) Hash() []byte {
+	return ec.h
 }
 
 func (ec errorCloser) Rewind() error {
@@ -486,7 +491,7 @@ func (s *Suite) TestSave(t *testing.T) {
 			Type: restic.PackFile,
 			Name: fmt.Sprintf("%s-%d", id, i),
 		}
-		err := b.Save(context.TODO(), h, restic.NewByteReader(data))
+		err := b.Save(context.TODO(), h, restic.NewByteReader(data, b.Hasher()))
 		test.OK(t, err)
 
 		buf, err := backend.LoadAll(context.TODO(), nil, b, h)
@@ -538,7 +543,19 @@ func (s *Suite) TestSave(t *testing.T) {
 
 	// wrap the tempfile in an errorCloser, so we can detect if the backend
 	// closes the reader
-	err = b.Save(context.TODO(), h, errorCloser{t: t, l: int64(length), ReadSeeker: tmpfile})
+	var beHash []byte
+	if b.Hasher() != nil {
+		beHasher := b.Hasher()
+		// must never fail according to interface
+		_, _ = beHasher.Write(data)
+		beHash = beHasher.Sum(nil)
+	}
+	err = b.Save(context.TODO(), h, errorCloser{
+		t:          t,
+		l:          int64(length),
+		ReadSeeker: tmpfile,
+		h:          beHash,
+	})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -583,7 +600,7 @@ func (s *Suite) TestSaveError(t *testing.T) {
 
 	// test that incomplete uploads fail
 	h := restic.Handle{Type: restic.PackFile, Name: id.String()}
-	err := b.Save(context.TODO(), h, &incompleteByteReader{ByteReader: *restic.NewByteReader(data)})
+	err := b.Save(context.TODO(), h, &incompleteByteReader{ByteReader: *restic.NewByteReader(data, b.Hasher())})
 	if err == nil {
 		t.Fatal("incomplete upload did not fail")
 	}
@@ -608,7 +625,7 @@ func (s *Suite) TestSaveFilenames(t *testing.T) {
 
 	for i, test := range filenameTests {
 		h := restic.Handle{Name: test.name, Type: restic.PackFile}
-		err := b.Save(context.TODO(), h, restic.NewByteReader([]byte(test.data)))
+		err := b.Save(context.TODO(), h, restic.NewByteReader([]byte(test.data), b.Hasher()))
 		if err != nil {
 			t.Errorf("test %d failed: Save() returned %+v", i, err)
 			continue
@@ -645,7 +662,7 @@ var testStrings = []struct {
 func store(t testing.TB, b restic.Backend, tpe restic.FileType, data []byte) restic.Handle {
 	id := restic.Hash(data)
 	h := restic.Handle{Name: id.String(), Type: tpe}
-	err := b.Save(context.TODO(), h, restic.NewByteReader([]byte(data)))
+	err := b.Save(context.TODO(), h, restic.NewByteReader([]byte(data), b.Hasher()))
 	test.OK(t, err)
 	return h
 }
@@ -799,7 +816,7 @@ func (s *Suite) TestBackend(t *testing.T) {
 		test.Assert(t, !ok, "removed blob still present")
 
 		// create blob
-		err = b.Save(context.TODO(), h, restic.NewByteReader([]byte(ts.data)))
+		err = b.Save(context.TODO(), h, restic.NewByteReader([]byte(ts.data), b.Hasher()))
 		test.OK(t, err)
 
 		// list items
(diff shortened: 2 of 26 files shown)


ncw:
> * REST / rclone: The REST server could easily check the sha256 hash of the uploaded data. It would also be possible to provide the md5 content hash, but that probably not too useful. For rclone that could be useful however. @ncw Would rclone be able to use the sha256/md5 sum to help the backends verify the upload (md5 would cover quite a lot of storage backends according to https://rclone.org/overview/)?

Rclone will be doing end to end checksum tests internally. It would be possible for rclone to return the checksum of the object from the remote cloud storage system after it was uploaded very easily - this could go in a header in the response to the POST request. Maybe in a  [Digest:](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest) header which can return multiple checksum types.

Which hash to return would be up go the rclone backend, but if just asking for a single checksum, rclone will tend to choose MD5 if at all possible.

> * local / sftp: As there is no "smart" storage system on the other side there's not much that can be done here. I mean it would be possible to check that the generated pack file data matches the data which is later on written to the repository. However, I'm not sure whether that's worth the trouble.

Rclone supports hashes on SFTP by running the `md5sum` or `sha1sum` commands remotely.

MichaelEischer:
Would it be possible to provide rclone with the MD5sum beforehand, such that it can pass that information on to the cloud storage systems? At least for the storage systems which support that it would be preferable to immediately fail the upload if it doesn't match the hash. Check the hash afterwards as a fallback would of course also be better than nothing.

ncw:
> Would it be possible to provide rclone with the MD5sum beforehand, such that it can pass that information on to the cloud storage systems? At least for the storage systems which support that it would be preferable to immediately fail the upload if it doesn't match the hash. Check the hash afterwards as a fallback would of course also be better than nothing.

Yes, if rclone can have the MD5SUM first then that would be the best of all worlds.

internal/hashing/writer.go
```diff
@@ -28,5 +28,8 @@ func (h *Writer) Write(p []byte) (int, error) {
 
 // Sum returns the hash of all data written so far.
 func (h *Writer) Sum(d []byte) []byte {
+	if h == nil {
```
fd0, line 31: This feels wrong to me. I can see that it's used in `packer_manager.go` so that when a backend does return `nil` for the hash, the `hashing.Writer` is set to nil. I'd be more comfortable to handle this special case in the `PackerManager` instead of here, what do you think?

internal/backend/test/tests.go
```diff
@@ -538,7 +543,19 @@ func (s *Suite) TestSave(t *testing.T) {
 
 	// wrap the tempfile in an errorCloser, so we can detect if the backend
 	// closes the reader
-	err = b.Save(context.TODO(), h, errorCloser{t: t, l: int64(length), ReadSeeker: tmpfile})
+	var beHash []byte
+	if b.Hasher() != nil {
+		beHasher := b.Hasher()
+		// must never fail according to interface
+		_, _ = beHasher.Write(data)
```
fd0, line 550: I'd rather panic here in the unlikely case that `Write()` returns a non-nil error:
```go
_, err := beHasher.Write(data)
if err != nil {
    panic(err)
}
```

internal/restic/rewind_reader.go
```diff
@@ -38,14 +43,26 @@ func (b *ByteReader) Length() int64 {
...
 // statically ensure that *ByteReader implements RewindReader.
 var _ RewindReader = &ByteReader{}
 
 // NewByteReader prepares a ByteReader that can then be used to read buf.
-func NewByteReader(buf []byte) *ByteReader {
+func NewByteReader(buf []byte, hasher hash.Hash) *ByteReader {
+	var hash []byte
+	if hasher != nil {
+		// must never fail according to interface
+		_, _ = hasher.Write(buf)
```
fd0, line 59: Same here, panic?

internal/restic/rewind_reader_test.go
```diff
@@ -45,15 +49,23 @@ func TestFileReader(t *testing.T) {
...
-	fn := func() RewindReader {
-		rd, err := NewFileReader(f)
-		if err != nil {
-			t.Fatal(err)
+	for _, hasher := range []hash.Hash{nil, md5.New()} {
+		fn := func() RewindReader {
+			var hash []byte
+			if hasher != nil {
+				// must never fail according to interface
+				_, _ = hasher.Write(buf)
```
fd0, line 57: Also panic on non-nil err?

internal/backend/mem/mem_backend.go
```diff
@@ -82,6 +84,16 @@ func (be *MemoryBackend) Save(ctx context.Context, h restic.Handle, rd restic.Re
 		return err
 	}
 
+	beHash := be.Hasher()
+	// must never fail according to interface
+	_, _ = beHash.Write(buf)
```
fd0, line 89: I'd also panic here just to be consistent :)

fd0 (changes requested):
In general I agree with what you're proposing, great stuff! Just a few minor things I've found need to be clarified.

fd0:
Btw, did you run the integration tests against the live backends? Do we maybe need a test that the backends (especially the cloud-based ones) indeed fail when a wrong hash is supplied?

internal/hashing/writer.go
MichaelEischer: I've moved the check nil check to the packer_manager and its test.

internal/backend/test/tests.go
MichaelEischer: I think it's never possible to receive an error here. But a sanity check won't hurt. -> I've added the check.

MichaelEischer:
> Btw, did you run the integration tests against the live backends? Do we maybe need a test that the backends (especially the cloud-based ones) indeed fail when a wrong hash is supplied?

I haven't run the integration tests against the live backends yet. A test that the backends actually reject files with wrong hashes sounds like a really important addition here, I'll add one.

MichaelEischer:
Superseded by #3246.

PR #3179, opened 2020-12-20
check: Remove filesize counter

aawsome:

Remove file size counter in `check` and adds a comment referring to #1887.

No, I found it while reviewing #3106 

Checklist
---------

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -567,18 +567,21 @@ func (c *Checker) checkTree(id restic.ID, tree *restic.Tree) (errs []error) {
 				errs = append(errs, Error{TreeID: id, Err: errors.Errorf("file %q has nil blob list", node.Name)})
 			}
 
-			var size uint64
 			for b, blobID := range node.Content {
 				if blobID.IsNull() {
 					errs = append(errs, Error{TreeID: id, Err: errors.Errorf("file %q blob %d has null ID", node.Name, b)})
 					continue
 				}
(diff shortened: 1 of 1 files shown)


MichaelEischer:
If I understand the old archiver implementation (restic before 0.9.0) correctly, then the `Size` field of a node and its chunk list might be inconsistent:

https://github.com/restic/restic/blob/272ccec7e16f74aaad771ace8a1774ccfdfe249b/internal/archiver/archiver.go#L211

aawsome:
> If I understand the old archiver implementation (restic before 0.9.0) correctly, then the `Size` field of a node and its chunk list might be inconsistent:
> 
> [](https://github.com/restic/restic/blob/272ccec7e16f74aaad771ace8a1774ccfdfe249b/internal/archiver/archiver.go#L211)

Ok, but it might still be worth to show a warning in that case.. So what should we do?

- print a warning in that case (compared to the orphaned pack warning)?
- or remove the code that calculates the size from the contents and add a comment that this is not checked on purpose?

MichaelEischer:
The size check was removed in #1887 after throwing errors which can't be fixed at users. And I don't see any reliable way to distinguish pre and post 0.9.0 snapshots. So it's probably best to remove the dead counter.

aawsome:
Changed this PR such that it removes the counter and adds a comment referring to #1887.

MichaelEischer (approved):
LGTM. Thanks!

PR #3199, opened 2020-12-29
Don't print progress on non-interactive terminals

### LINKED ISSUE #2706: Option for newlines after each progress update

### LINKED ISSUE #3194: Improve progress information for non-interactive usage
Output of `restic version`
--------------------------
```
$ /home/restic/bin/restic version
restic 0.11.0 (v0.11.0-215-g8efb874f) compiled with go1.15.6 on linux/amd64
```

I run restic backups, checks, forgets and prunes through systemd and I use templates for running it with different configurations. In this case: `bash -c '/home/restic/bin/restic $(</home/restic/restic.s3.check)'` will be executed by systemd which expands to `bash -c '/home/restic/bin/restic check --read-data -o s3.list-objects-v1=true'`. This leads to lot of stdout output which then "spams" the journal as you can see with `journalctl`.

```
❯ systemctl cat restic-run-s3@.service
# /usr/lib/systemd/system/restic-run-s3@.service
[Unit]
Description=Run restic job with parameters from /home/restic/restic.s3.%i
OnFailure=status-mail@%n.service

[Service]
Type=simple
Nice=10
IOSchedulingClass=best-effort
IOSchedulingPriority=7
User=restic
Environment="HOME=/home/restic"
EnvironmentFile=/home/restic/restic.s3.env
ExecStart=bash -c '/home/restic/bin/restic $(</home/restic/restic.s3.%i)'
❯ sudo systemctl start restic-run-s3@check.service && sudo journalctl -fu $_
-- Journal begins at Sun 2020-07-12 14:07:29 CEST. --
[...]
Dec 26 01:15:47 machine systemd[1]: Started Run restic job with parameters from /home/restic/restic.s3.check.
Dec 26 01:15:48 machine bash[77720]: using temporary cache in /tmp/restic-check-cache-618104479
Dec 26 01:15:49 machine bash[77720]: created new cache in /tmp/restic-check-cache-618104479
Dec 26 01:15:49 machine bash[77720]: create exclusive lock for repository
Dec 26 01:15:49 machine bash[77720]: load indexes
Dec 26 01:16:03 machine bash[77720]: check all packs
Dec 26 01:16:07 machine bash[77720]: check snapshots, trees and blobs
Dec 26 01:17:32 machine bash[77720]: read all data
Dec 26 01:17:33 machine bash[77720]: [0:01] 0.00%  0 / 23628 packs
Dec 26 01:17:34 machine bash[77720]: [0:02] 0.01%  2 / 23628 packs
Dec 26 01:17:35 machine bash[77720]: [0:03] 0.02%  4 / 23628 packs
Dec 26 01:17:36 machine bash[77720]: [0:04] 0.02%  5 / 23628 packs
Dec 26 01:17:37 machine bash[77720]: [0:05] 0.03%  7 / 23628 packs
Dec 26 01:17:38 machine bash[77720]: [0:06] 0.04%  9 / 23628 packs
Dec 26 01:17:39 machine bash[77720]: [0:07] 0.04%  10 / 23628 packs
Dec 26 01:17:40 machine bash[77720]: [0:08] 0.06%  14 / 23628 packs
Dec 26 01:17:41 machine bash[77720]: [0:09] 0.06%  15 / 23628 packs
Dec 26 01:17:42 machine bash[77720]: [0:10] 0.07%  16 / 23628 packs
Dec 26 01:17:43 machine bash[77720]: [0:11] 0.08%  19 / 23628 packs
Dec 26 01:17:44 machine bash[77720]: [0:12] 0.08%  20 / 23628 packs
[...]
```

S3

Note: I compiled it myself because I needed ListObjects v1. I could not test this, but I would except it to be unrelated to this flag.

Expected behavior
-----------------
The expected behavior would be to not have this output on stdout but on the tty so it is displayed in a interactive shell but not in the journal. I think the `restic backup` command does it like this or it registers the absence of a interactive session and behaves differently.

Actual behavior
---------------
restic is unaware of context and writes to much stdout logs (to the systemd journal).

Run `restic check --read-data` through a systemd service `restic-unit.service` and check the progress with `journalctl -fu restic-unit.service`.

I think this can also reproduced by running: `restic check --read-data -o s3.list-objects-v1=true | cat`

The best guesses are that `restic check` is unaware that the context it is running in is non-interactive, for example through a pipe, or is writing progress to stdout instead to tty (unlike other commands that display progress).

If my guesses are correct, I would check the other progress display commands and adjust the one for `restic check` accordingly.

I went from manual backup to 5min snapshots for my home directory and it is the best!

MichaelEischer:
The PR revert the progress bar updates to the old behavior of not printing progress updates on non-interactive terminals.

As an additional change, the `RESTIC_PROGRESS_FPS` environment variable is now also interpreted for non-interactive terminals. This makes it possible to use the environment variable to also get regular progress updates on non-interactive terminals.

Fixes #3194.
See #2706 for further discussions.

Checklist
---------

- [x] I have added tests for all changes in this PR **We currently have no means for testing the `progress.go` change, however, I've verified that it works manually.**

### DIFF
--- a/changelog/unreleased/issue-2706
+++ b/changelog/unreleased/issue-2706
@@ -1,0 +1,16 @@
+Enhancement: Configurable progress reports for non-interactive terminals
+
+The `backup`, `check` and `prune` commands never printed any progress
+reports on non-interactive terminals. This behavior is now configurable
+using the `RESTIC_PROGRESS_FPS` environment variable. Use for example a
+value of `1` for an update per second or `0.01666` for an update per minute.
+
+The `backup` command now also prints the current progress when restic
+receives a `SIGUSR1` signal.
+
+Setting the `RESTIC_PROGRESS_FPS` environment variable or sending a `SIGUSR1`
+signal prints a status report even when `--quiet` was specified.
+
+https://github.com/restic/restic/issues/2706
+https://github.com/restic/restic/issues/3194
+https://github.com/restic/restic/pull/3199
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -11,7 +11,6 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"strings"
 	"time"
 
@@ -545,15 +544,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	}()
 	gopts.stdout, gopts.stderr = p.Stdout(), p.Stderr()
 
-	if s, ok := os.LookupEnv("RESTIC_PROGRESS_FPS"); ok {
-		fps, err := strconv.Atoi(s)
-		if err == nil && fps >= 1 {
-			if fps > 60 {
-				fps = 60
-			}
-			p.SetMinUpdatePause(time.Second / time.Duration(fps))
-		}
-	}
+	p.SetMinUpdatePause(calculateProgressInterval())
 
 	t.Go(func() error { return p.Run(t.Context(gopts.ctx)) })
 
--- a/cmd/restic/progress.go
+++ b/cmd/restic/progress.go
@@ -9,12 +9,10 @@ import (
 	"github.com/restic/restic/internal/ui/progress"
 )
 
-// newProgressMax returns a progress.Counter that prints to stdout.
-func newProgressMax(show bool, max uint64, description string) *progress.Counter {
-	if !show {
-		return nil
-	}
-
+// calculateProgressInterval returns the interval configured via RESTIC_PROGRESS_FPS
+// or if unset returns an interval for 60fps on interactive terminals and 0 (=disabled)
+// for non-interactive terminals
+func calculateProgressInterval() time.Duration {
 	interval := time.Second / 60
 	fps, err := strconv.ParseInt(os.Getenv("RESTIC_PROGRESS_FPS"), 10, 64)
 	if err == nil && fps >= 1 {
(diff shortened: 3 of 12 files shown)


fd0:
Code looks good already!

DRON-666:
If I understand correctly, this PR is primary intended for Unix users, but it would be nice to add Windows support.
Now on Windows output redirected to a pipe always contains progress lines and redirected to a file never contains it (this PR doesn't change anything in this behavior). I think it's better to disable progress on redirected output by default and enable it via `RESTIC_PROGRESS_FPS`.

MichaelEischer:
@DRON-666 That behavior on Windows is currently intentional. Please have a look at https://github.com/restic/restic/blob/a488d4c847ed9631bab537d0e3b9ec654bed978a/internal/ui/termstatus/terminal_windows.go#L97 . The problem is that the detection of whether a terminal is interactive or not is not always reliable. However, that problem is out of scope for this PR.

I've updated the PR to also adapt the progress output of the `backup` command which uses its own implementation. The progress report frequency is now also controlled in a similar manner to the generic progress bar implementation.

`RESTIC_PROGRESS_FPS` now uses floating-point numbers to allow reducing the progress report frequency to less than once per second.

DRON-666:
>However, that problem is out of scope for this PR.

You right. I will write a new PR after merging of this.

fd0 (approved):
Great work, thanks! I've added a tiny improvement on top, but then it's good to go!

Issue #1514, opened 2017-12-30
Feature-request: Source .gitignore


da2x:
So this is just an idea. I’ve got a lot of source code that I want backed up. There is a ton of non-essential reproducible build related files in between all the valuable source code. Most of it is in git, and most of the repositories knows what files are important to backup already; everything else is in `.gitignore`. I’d like to see an optional flag for `--exclude-from-gitignores` that would source `.gitignore` files for exclusions (relative to currently scanned directory) as they’re encountered by the scanner.

rawtaz:
`--exclude-gitignores` is more coherent with the existing exclude options.

JackBlackmore:
Just an idea but if this was going to be implemented I wouldn't limit it to just `.gitignore` files as this feature could be extremely useful for non git directory trees. 

An option for something like `--exclude-path-file=.gitignore` (really can't think of a better name) and these files would get picked up and used to generate the exclude list for everything from that directory down. 

Is the current `--exclude-file=` format the same as `.gitignore` files as otherwise it's going to need another parser to be implemented?

I've got a feeling something similar has been asked for before so I'll have a quick look through the issues log.

alphapapa:
@JackBlackmore As you said, the gitignore glob patterns and Go match/glob patterns would have to be exactly the same.  It would be very convenient if they are.

What this is essentially asking for is a file contents-based version of `--exclude-if-present`, so maybe something like `--exclude-file-if-present`.

What I'd really like to see is a more universal standard for something like a `.backupignore` file format.  I use a few different backup programs, and it would save a lot of manual work if they could all read the same ignore/exclude files, so I could drop such a file in a directory and have all my backup programs ignore the matching paths.  Maybe Restic could lead the way here...?  :)

dionorgua:
I also like idea for automatic exclude of `.gitignore` / `.hgignore` or similar files. Just because they are usually already provided for free.

thebrave:
`tar` already has `--exclude-vcs-ignores`. From the [manual](https://www.gnu.org/software/tar/manual/html_node/exclude.html):

> When archiving directories that are under some version control system (VCS), it is often convenient to read exclusion patterns from this VCS ignore files (e.g. `.cvsignore`, `.gitignore`, etc.) The following options provide such possibility:
> 
> `--exclude-vcs-ignores`
> 
> Before archiving a directory, see if it contains any of the following 
[...]

Issue #2376, opened 2019-08-16
Option to add description to snapshot


erlanger:
Output of `restic version`
--------------------------
restic 0.9.5 (v0.9.5-46-g604b18aa) compiled with go1.12.7 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
Ability to add a description to a snapshot (i.e. "after upgrading OS to version 5.4.2")

What are you trying to do?
--------------------------
When I list snapshots, I would like an optional description describing the event that
caused us to make the snapshot (i.e. after upgrade to version 3.4.5)

Did restic help you or made you happy in any way?
-------------------------------------------------
Very good things:
* Static binary (which can live with the backups)
* Ability to mount the snapshots
* Save a good amount of space by deduping

rawtaz:
Just add a tag to the snapshot. See `restic help backup`, in particular the `-t tag` argument. It's not free-form text, but good enough to do what you need if you just don't write such plain english. An example tag could be `postUpgradeOsToV542`.

erlanger:
I knew about tags, but sometimes I want to include 'bullet points' in my description, such as:

"After running tronscript 10.8.2:
   * with mb disabled
   * and cleaners disabled"

I find it improper and difficult to put this in a tag.

rawtaz:
Heh, seems terribly overkill to me. I can't even relate to wanting to do that. But I guess everyone has their use cases :-)

moritzdietz:
I can see the usecase though. It's like a btrfs snapshot where you can add a comment or when you do a snapshot in VMWare; you can add a comment to it. I just checked and Borg has a feature where you can add a comment to a snapshot. Not to say that just because borg has it, restic must have it also, but just to highlight that there can be usecases.

I could see that one would do a `restic snapshots` and maybe filter for specific words in a comment field to find the right snapshot. But then again this could also be solved via tags - but maybe this is abusing the tag feature for something it shouldn't do.

AlBundy33:
because restic uses principles from git i would say that the request sounds similar to a commit-message

rehanone:
[...]

PR #3200, opened 2020-12-29
backup: Add options --set-path and --set-path-from

### LINKED ISSUE #555: Ability to override a target

### LINKED ISSUE #1376: Ability to overwrite Drive/Path on backups
Restic has the ability to override the machine name, but not the actual path name.  This would be useful for backing up USB drives from different machines as they may not always mount to the same drive letter.

This is more relevant in windows than in linux, but the same issue might arise for temporarily mounted volumes.

Reference: https://forum.restic.net/t/ability-to-override-drive-path-name/201

### LINKED ISSUE #2092: Backup option to remove a leading path prefix
Output of `restic version`
--------------------------
```
```

It would be helpful to have an option to `restic backup` that says "strip this leading path from the paths of all backed-up files."  For example, `--backup-root /some/path`.  This would have the following effects:

* A file `/some/path/to/file` would be stored in the snapshot as `/to/file`.
* This file would also have its metadata check performed against `/to/file` in the parent snapshot.
* You are not permitted to specify files/directories to `restic backup` that do not begin with this prefix.

(I think this may be related to #1376.)

What are you trying to do?
--------------------------
One of my backup scripts is run on a system with many running services that a cannot stop.  These services guarantee that recovery is possible from a point in time (e.g. they do enough journaling to get their data back in a consistent state following a power cut).  However, restic backups are not point-in-time.

To fix this, we:

1. Take an LVM snapshot of `/`.
2. Mount the LVM snapshot under `/mnt/backup-snapshot`.
3. Run the restic backup against `/mnt/backup-snapshot`.
4. Unmount the LVM snapshot.
5. Delete the LVM snapshot.

Unfortunately, this causes files to be stored in our restic repository with the (useless) prefix `/mnt/backup-snapshot`.  The only feasible workaround I can think of is to run the backup within a chroot.  While not the end of the world, it might be nicer for restic to provide an option to remove some leading prefix from files.

### LINKED ISSUE #2246: Parent-backup detection fails with changing --files-from
Output of `restic version`
--------------------------

`restic 0.9.4 compiled with go1.11.5 on linux/amd64`

(debian testing, from the debian repository)

In short:  With an invocation like `restic backup --files-from my-files-and-dirs.lst`, restic could be more efficient about choosing a parent snapshot.

In long:

Whenever `my-files-and-dirs.lst` changes, no matter how slightly, restic apparently sees a different set of paths.  Only slightly different, but still different.  When searching for a parent snapshot during `backup`, this leads to unexpected behavior:

On the one hand, restic sees that the path set is different from anything seen ever before, and assumes a completely new backup.  All data is re-uploaded:

    repository 7b6b235d opened successfully, password is correct
    Files:        1905 new,     0 changed,     0 unmodified
    Added to the repo: 44.372 KiB
    processed 1905 files, 346.776 MiB in 0:02
    snapshot e50ef85d saved

On the other hand, I made only a small change in `my-files-and-dirs.lst`, so I expected that only the new files need to be uploaded.

I'm new to restic, so maybe I'm using it wrong.  However, using tags does not seem to change automatic parent detection, and `--parent latest` does not seem to be supported.  And I don't want to specify `--parent 12345678` manually all the time, and would like to avoid fiddling with `restic snapshots` on my own.

I'm not sure which feature to propose.  There are multiple things that might help:

1. Allow `--parent latest` to use the latest snapshot, no matter what.  This would be helpful for people like me, who only have one guest per repository anyway, but might result in other scenarios.
2. Instead of 1., allow `--parent latest-sametags` to use the latest snapshot of the same tag set.  This would avoid potential problems, and still cover most use cases.
3. Automatic parent detection could try to find a close match in the previous few snapshots, and if it finds one, use that.  As far as I can see, a false positive cannot have a bad impact, can it?

What are you trying to do?
--------------------------

I'm making snapshots of parts of my home directory, and have restrictions on the target repository size.  So I only want to include specific things, like `~/workspace/`, `~/bin/`, `~/.bashrc`, and so on.  But not other things, like the gigantic folder of virtual machines, as the restic host has not enough space for that.  Obviously, this list is subject to change.

With the current behavior, a simple backup without `--parent 5eab0a7` makes the backup run a bit longer than necessary.  Deduplication does its job perfectly, and no excess data is stored.

With the proposed behavior, no such delay would happen, or only when detection fails.

I'm currently using rsnapshot, and it works great for its usecase.  However, with so many small files, and some large files constantly changing (thunderbird is a strong offender), my home directory falls outside rsnapshot's use case.  It seems restic is fast and space-efficient enough to cope with that much better.  Hooray!

### LINKED ISSUE #2714: Add an option to restic backup to override the snapshot's "paths" field
Output of `restic version`
--------------------------

Add an option to `restic backup` to override the snapshot's "paths" field.

What are you trying to do?
--------------------------

My restic experience went like this:

* Looks like it has some features for excluding patterns etc., but I want more control over which files are backed up.
* Ah, no problem, it has an option for that: `restic backup --files-from` lets me just tell it exactly which files to backup.  Great!
* Hmm, it's actually treating the contents not as files, but as patterns.  Never mind, I'll just escape any glob characters in the file paths, and...
* Ok, it worked!  Uploading the snapshot manifest at the end took a strangely long time, but :shrug:.  Now let's take a look at `restic snapshots`, and OOOH sweet jesus!

As you can probably guess, I had created a snapshot with 140k items in its `paths` field :grimacing:. As far as I can tell, this is not actually causing any real problems.  I'd just like to be able to override the "path" field when I make a backup, so that the output of `restic snapshots` is more usable.

Restic seems really good.  I'm a long-time bup user, and bup is great too - but I wanted encryption so I could use a cloud storage provider.  Restic seems to follow mostly the same design principles as bup, so I thought I'd give it a try.  I'm liking it so far!

### LINKED ISSUE #3198: `--files-from-raw` breaks parent detection and snapshots output
Output of `restic version`
--------------------------
restic 0.11.0 (v0.11.0-221-g9c41e4a3) compiled with go1.14.5 on darwin/amd64

```sh
  restic backup \
    --repo "$repo" \
    --password-command "$pw_cmd" \
    --files-from-raw <( fd -0Ht f --ignore-file "$ignore_file" . )
```

Then something like: 
```sh
restic snapshots
```

Or run the backup command again with `--verbose`.

Local directory.

Expected behavior
-----------------
Restic should have useful output for commands like `restic snapshots` even if some snapshots were made using `--files-from-raw` (or similar flags).

Restic should recognize a parent snapshot even when a reasonably small amount of files is added or deleted from the backup path.

Actual behavior
---------------
After creating a backup using `--files-from-raw`, running `restic snapshots` outputs thousands of lines per snapshot, making the listing scroll by too quickly too see, and making the output way too large to be useful.

After creating a backup using `--files-from-raw`, running another subsequent backup with even a single file added or deleted makes restic output `no parent snapshot found, will read all files` and redo a scan (which slows down the backup).

See command above. Run that (multiple times if needed), then run other restic commands that rely on paths (listing snapshots, etc).

I'm guessing that restic is taking the list from `--files-from-raw` very literally. If a file is added or deleted, then the new backup is technically using a new path. This isn't the same behavior as other backups, so it's a bit jarring.

This approach makes a lot of sense if paths are user provided (restic up until now), but has some pretty major drawbacks when used with a potentially very large, computer-generated list of paths (like what `--files-from-raw` is intended for).

I'm not familiar with restic's codebase, so I'm making a lot of guesses, but I think maybe grouping paths in some way could help here. If the paths are grouped into a smaller, more manageable size, then this problem pretty much goes away.

I think there are three possible ways to approach this:

1. Auto generate the shorter path list.
2. Allow the user to provide a shorter path list.
3. Hybrid of 1 and 2.

If auto generated, maybe some heuristic involving directory depth, max results, or some combination of the two could be used to merge the provided paths. This is probably easiest for the user, but has the potential to make the list too short and possibly useless if the shortener is too aggressive.

If the user provides the list, then that gives the most control to the users. It's probably fairly easy to implement (again guessing, so please correct me if I'm wrong) by taking an approach that ignores other paths if the flag is used (making `--files-from-raw` mean "backup these--and only these--files"). This does have the downside of limiting the new behavior so that it can't be used in conjunction with other lists for a snapshot, so maybe another flag? If this approach is used, maybe something like `restic backup --files-from-raw somefile.txt ~/mypath ~/another-path` would make sense, although that could lead to some weird situations wher the provided path list and the given paths are unrelated to each other (not sure what impact that would have).

If the hybrid approach is taken, maybe it could be something like restic auto-generates the shorter path list if no path is provided, or uses whatever the user provides if it exists. This leads to the most control for users while still providing a sensible default for users that don't know to provide more input. The downside would be that it's more moving parts, which means more complexity. As I don't know your codebase, I can't accurately estimate effort for any of these approaches.

I realize that this isn't probably an easy fix, but I think it's worth a look because `--files-from-raw` is a very useful new feature, but it also causes some of the other functionality already in restic to take a big step backward.

Yes! Restic is great! I've been testing out about a dozen or so different backup methods over the last few months, and restic was by the far the best one for my needs. It's super easy to get going (it took almost as little time as the rsync script), and has a great balance of features and usability. It's also way faster for my setup than other programs with similar functionality. So, thank you!

aawsome:

Adds an option `--set-path` to `backup` which allows to manually set the path(s) saved in the snapshot and used for finding the parent snapshot. Also the option `--set-path-from` is added to read the paths from a file.

Both options are useful e.g. if the files to backup are selected by an external tool.

As `set-path` functionally replaces `--stdin-filename`, the latter is now marked as deprecated.

closes #2714 
closes #3198 
allows users to use an easy workaround for #1514 by using `--files-from-raw` in combination with `fd` (or similar find tools) and `--set-path`
maybe also closes #2246 
closes #1376
closes #2092

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -117,11 +117,17 @@ func init() {
 	f.StringVar(&backupOptions.ExcludeLargerThan, "exclude-larger-than", "", "max `size` of the files to be backed up (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&backupOptions.Stdin, "stdin", false, "read backup from stdin")
 	f.StringVar(&backupOptions.StdinFilename, "stdin-filename", "stdin", "`filename` to use when reading from stdin")
+	err := f.MarkDeprecated("stdin-filename", "use --set-path")
+	if err != nil {
+		// MarkDeprecated only returns an error when the flag could not be found
+		panic(err)
+	}
+
 	f.Var(&backupOptions.Tags, "tag", "add `tags` for the new snapshot in the format `tag[,tag,...]` (can be specified multiple times)")
 
 	f.StringVarP(&backupOptions.Host, "host", "H", "", "set the `hostname` for the snapshot manually. To prevent an expensive rescan use the \"parent\" flag")
 	f.StringVar(&backupOptions.Host, "hostname", "", "set the `hostname` for the snapshot manually")
-	err := f.MarkDeprecated("hostname", "use --host")
+	err = f.MarkDeprecated("hostname", "use --host")
 	if err != nil {
 		// MarkDeprecated only returns an error when the flag could not be found
 		panic(err)
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -86,6 +86,8 @@ type BackupOptions struct {
 	FilesFrom               []string
 	FilesFromVerbatim       []string
 	FilesFromRaw            []string
+	Paths                   []string
+	PathsFrom               []string
 	TimeStamp               string
 	WithAtime               bool
 	IgnoreInode             bool
@@ -128,6 +130,8 @@ func init() {
 	f.StringArrayVar(&backupOptions.FilesFrom, "files-from", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.FilesFromVerbatim, "files-from-verbatim", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.FilesFromRaw, "files-from-raw", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
+	f.StringArrayVar(&backupOptions.Paths, "set-path", nil, "manually set the `path` of snapshot (default: paths to backup; can be specified multiple times)")
+	f.StringArrayVar(&backupOptions.PathsFrom, "set-path-from", nil, "read the path(s) to set from `file` (can be combined with --set-path; can be specified multiple times)")
 	f.StringVar(&backupOptions.TimeStamp, "time", "", "`time` of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&backupOptions.WithAtime, "with-atime", false, "store the atime for all files and directories")
 	f.BoolVar(&backupOptions.IgnoreInode, "ignore-inode", false, "ignore inode number changes when checking for modified files")
@@ -252,7 +256,7 @@ func readFilenamesRaw(r io.Reader) (names []string, err error) {
 // Check returns an error when an invalid combination of options was set.
 func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 	if gopts.password == "" {
-		filesFrom := append(append(opts.FilesFrom, opts.FilesFromVerbatim...), opts.FilesFromRaw...)
+		filesFrom := append(append(append(opts.FilesFrom, opts.FilesFromVerbatim...), opts.FilesFromRaw...), opts.PathsFrom...)
 		for _, filename := range filesFrom {
 			if filename == "-" {
 				return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
@@ -269,6 +273,12 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 		}
 		if len(opts.FilesFromRaw) > 0 {
 			return errors.Fatal("--stdin and --files-from-raw cannot be used together")
+		}
+		if len(opts.Paths) > 0 {
+			return errors.Fatal("--stdin and --set-path cannot be used together")
+		}
+		if len(opts.PathsFrom) > 0 {
+			return errors.Fatal("--stdin and --set-paths-from cannot be used together")
 		}
 
 		if len(args) > 0 {
@@ -507,6 +517,19 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		return err
 	}
 
+	for _, file := range opts.PathsFrom {
+		fromfile, err := readLines(file)
+		if err != nil {
+			return err
+		}
+		opts.Paths = append(opts.Paths, fromfile...)
+	}
+
+	// set snapshot paths to targets if not given explicitely
+	if len(opts.Paths) == 0 {
+		opts.Paths = targets
+	}
+
 	timeStamp := time.Now()
 	if opts.TimeStamp != "" {
 		timeStamp, err = time.ParseInLocation(TimeFormat, opts.TimeStamp, time.Local)
@@ -579,7 +602,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		return err
 	}
 
-	parentSnapshotID, err := findParentSnapshot(gopts.ctx, repo, opts, targets)
+	parentSnapshotID, err := findParentSnapshot(gopts.ctx, repo, opts, opts.Paths)
 	if err != nil {
 		return err
 	}
@@ -687,6 +710,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		Time:           timeStamp,
 		Hostname:       opts.Host,
 		ParentSnapshot: *parentSnapshotID,
+		Paths:          opts.Paths,
 	}
 
 	if !gopts.JSON {
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -372,6 +372,11 @@ the list of files/patterns from there instead of a text file.
 In all cases, paths may be absolute or relative to ``restic backup``'s working
 directory.
 
+Note that if you use restic in combination with include files, by default all
+entries of the include files are added to the snapshots paths. This may break
+the finding of parent snapshots. Please consider setting the paths manually
+using ``--set-path`` or ``--set-path-from`` to reflect what is actually backuped.
+
 For example, maybe you want to backup files which have a name that matches a
 certain regular expression pattern (uses GNU ``find``):
 
@@ -383,14 +388,17 @@ You can then use restic to backup the filtered files:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo backup --files-from-raw /tmp/files_to_backup
+    $ restic -r /srv/restic-repo backup --files-from-raw /tmp/files_to_backup --set-path /tmp/somefiles
 
 You can combine all three options with each other and with the normal file arguments:
 
 .. code-block:: console
 
-    $ restic backup --files-from /tmp/files_to_backup /tmp/some_additional_file
-    $ restic backup --files-from /tmp/glob-pattern --files-from-raw /tmp/generated-list /tmp/some_additional_file
+    $ restic backup --files-from /tmp/files_to_backup /tmp/some_additional_file \
+                    --set-path /tmp/somefiles --set-path /tmp/some_additional_file
+    $ restic backup --files-from /tmp/glob-pattern --files-from-raw /tmp/generated-list /tmp/some_additional_file \
+                    --set-path-from /tmp/glob-pattern --set-path /tmp/somefiles --set-path /tmp/some_additional_file
+
 
 Comparing Snapshots
 *******************
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -471,11 +471,11 @@ use e.g. the fuse mounting option (see below) to mount the repository
 and read the file.
 
 By default, the file name ``stdin`` is used, a different name can be
-specified with ``--stdin-filename``, e.g. like this:
+specified with ``--set-path``, e.g. like this:
 
 .. code-block:: console
 
-    $ mysqldump [...] | restic -r /srv/restic-repo backup --stdin --stdin-filename production.sql
+    $ mysqldump [...] | restic -r /srv/restic-repo backup --stdin --set-path production.sql
 
 The option ``pipefail`` is highly recommended so that a non-zero exit code from
 one of the programs in the pipe (e.g. ``mysqldump`` here) makes the whole chain
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -727,6 +727,7 @@ func resolveRelativeTargets(filesys fs.FS, targets []string) ([]string, error) {
 // SnapshotOptions collect attributes for a new snapshot.
 type SnapshotOptions struct {
 	Tags           restic.TagList
+	Paths          []string
 	Hostname       string
 	Excludes       []string
 	Time           time.Time
@@ -827,7 +828,7 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 		return nil, restic.ID{}, err
 	}
 
-	sn, err := restic.NewSnapshot(targets, opts.Tags, opts.Hostname, opts.Time)
+	sn, err := restic.NewSnapshot(opts.Paths, opts.Tags, opts.Hostname, opts.Time)
 	if err != nil {
 		return nil, restic.ID{}, err
 	}
--- a/changelog/unreleased/issue-2714
+++ b/changelog/unreleased/issue-2714
@@ -1,0 +1,11 @@
+Enhancement: Allow to set paths saved in snapshots manually
+
+Restic backup always used to save the paths specified for the backup
+in the snapshot. We've added new options to set this directly. This
+allows to have clearer snapshots and use parent snapshots e.g. if
+the files to backup comes from a external file find tool. 
+
+https://github.com/restic/restic/issues/2246
+https://github.com/restic/restic/issues/2714
+https://github.com/restic/restic/issues/3198
+https://github.com/restic/restic/pull/3200


MichaelEischer:
There has already been some discussion on such an option in #2714.

aawsome:
> There has already been some discussion on such an option in #2714.

Thanks for the hint! I added this to in the description and adapted the changelog file.

cmd/restic/cmd_backup.go
```diff
@@ -124,6 +125,7 @@ func init() {
 	f.StringArrayVar(&backupOptions.FilesFrom, "files-from", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.FilesFromVerbatim, "files-from-verbatim", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
 	f.StringArrayVar(&backupOptions.FilesFromRaw, "files-from-raw", nil, "read the files to backup from `file` (can be combined with file args; can be specified multiple times)")
+	f.StringArrayVar(&backupOptions.Paths, "set-path", nil, "manually set the `path` of snapshot (default: paths to backp; can be specified multiple times)")
```
chrahunt, line 128:
```suggestion
	f.StringArrayVar(&backupOptions.Paths, "set-path", nil, "manually set the `path` of snapshot (default: paths to backup; can be specified multiple times)")
```

internal/archiver/archiver.go
```diff
@@ -819,7 +825,7 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 		return nil, restic.ID{}, err
 	}
 
-	sn, err := restic.NewSnapshot(targets, opts.Tags, opts.Hostname, opts.Time)
+	sn, err := restic.NewSnapshot(cleanPaths, opts.Tags, opts.Hostname, opts.Time)
```
chrahunt, line 828: I think this could have an impact on parent snapshot detection.

[Here](https://github.com/restic/restic/pull/3200/files#diff-30d8672eafa9c0585766b6f437530377563cd1aac5606674c366f3ede4abea9eR596) we use `targets` provided by the user to locate the parent snapshot. On the current line we use `cleanPaths` for the snapshot paths. If `resolveRelativeTargets` encounters a relative target (like `.`) then these two values will be different, and a subsequent backup would fail to locate the parent snapshot.

Is there any reason not to use `opts.Paths` directly?

chrahunt (commented):
Some hints in the docs for users would be helpful. I think the [including files](https://restic.readthedocs.io/en/latest/040_backup.html#including-files) section would be a good place to mention it, along with an explanation that if it is not set then the individual paths will be included in the snapshot metadata, which can have negative impacts for users specifying large numbers of files.

cmd/restic/cmd_backup.go
aawsome: Thanks for finding that typo! Is fixed.

internal/archiver/archiver.go
aawsome: Good catch! Is changed now!

aawsome:
I added some hints in the docu and realized that an extra options `--set-paths-from` would be also quite handy. This is now also added.

aawsome:
Also added some checks for the new flags in combination with stdin, analog to the checks of `--files-from*`.

aawsome:
About `--stdin`, we might think of making `--stdin-filename` an alias of `--set-path` and mark it deprecated. Then, for `--stdin` the only check we need is that only one path is given.
Should I add this to this PR?

robvalca:
Thanks a lot for this ! I was playing with this PR today and although `--set-path` works as expected, I'm see some inconsistencies compare to the normal backup command. This was always like that, but I don't know if this is expected also after this PR. I'll detail here what I'm doing:

1. I do a normal initial backup:

```
restic_dev -r local:/tmp/test backup /opt
```
2.  I add a new file `/opt/this_is_a_test` and I add this path in a `list.txt` file which I source to restic using `--files-from`. I've also pass `--set-path /opt`. (Tried with and without the `--parent` flag but the result is the same)
```
restic_dev -r local:/tmp/test backup --files-from /list.txt  --set-path /opt --parent 5976257d
```
3.  The backup ends correctly, and in the snapshots view I see both snapshots with the same path, so this is OK. 
```
ID        Time                 Host                 Tags        Paths
---------------------------------------------------------------------
5976257d  2021-01-15 16:29:35  cbox-restic.cern.ch              /opt
e5fa3044  2021-01-15 16:38:56  cbox-restic.cern.ch              /opt
---------------------------------------------------------------------
```
4. Now, if I mount the repo and traverse to snapshots/latests I see only the file added in the last snapshot:
```
[16:39][root@cbox-restic (qa:box/restic/dev) restic]# cd snapshots/latest/
[16:39][root@cbox-restic (qa:box/restic/dev) latest]# ls
opt
[16:39][root@cbox-restic (qa:box/restic/dev) latest]# cd opt/
[16:39][root@cbox-restic (qa:box/restic/dev) opt]# ls -la
total 1
drwxr-xr-x. 2 root root  0 Jan 15 16:31 .
dr-xr-xr-x. 2 root root  0 Jan 15 16:38 ..
-rw-r--r--. 1 root root 13 Jan 15 16:31 this_is_a_test
```
5. If I try restore the snapshot, only that file is recovered.
6. If instead of doing the combination of `--files-from`,`--set-path` and `--parent` I trigger a normal backup after adding the `this_is_a_test` file, the contents of `snapshots/latests` I see the full /opt view with the recently added file. 

I think that we should expect the same final state either doing it with normal backup or using the combination of `--files-from` and `--set-path`, what do you think?

Roberto.

aawsome:
@robvalca Thanks for your feedback! 
Is in your `/list.txt` only the line `/opt/this_is_a_test` or are there also the other files?
As you write "This was always like this", I guess so. In this case, yes then the new snapshot should only contain this one file. 
Note that in this case, there is no sense in using a parent as there are no "old" files which should be in the new snapshot.

robvalca:
Hi @aawsome , thanks for the quick reply! I only have that single path in the txt. I thought that there would be a merge with the files on the list  and the previous snapshot of the path specified by `--set-path`.  In any case, with this PR the internal view of the snapshots makes more sense.  I'm looking forward to have this merged. Thanks!

amuckart:
I very much look forward to seeing this merged! 

What would be extra useful would be the ability to also change the path of existing snapshots so as not to have to do a full backup of everything after this option is in use.

My use case is backing up a series of backups created with dirvish (for various reasons we can't use restic to directly back up the hosts), so they all end up in /var/backups/dirvish/<hostname>/<datestamp>/tree. I'm making the path consistent between backups by bind mounting the .../tree directory but this feature would completely remove the need for that complexity and allow me to directly back up the directory without having to bind mount it, as well as making restoration on to the server a whole lot easier, but then the path would be inconsistent with all of the existing months of backups I've got in the restic repository.

hamishcoleman:
Another use case not described above is when using filesystem snapshots (eg, with btrfs or zfs) as the source for the restic backup but wanting the restic backup path to clearly represent the actual data source.  Using the filesystem snapshot makes it much more likely to have a consistent backup of things like database files.

I also envision using the --set-path feature to write a utility that can take a list of local filesystem snapshots and ensure that they are all replicated in restic (including correct creation timestamps and parent relationship) as that would allow simple interfacing of an existing local backup system with its creation and expiry schedules and the more flexible off-site restic backup.

wren:
@amuckart I agree that changing the path would be a useful feature. But maybe that could be left for another PR (so this one doesn't get held up)?

aawsome:
I would also like to leave the scope of this PR as it is. Can we open a new issue to discuss this changing of trees in the repo? I think it's not just changing the tree names (which should be pretty simple) but the syntax how to specify which path maps to which tree. I think it there are many pitfalls which should be discussed first.

lalmeras:
:+1: for this proposal.

My use-case is for a selective backup that I cannot perform with an exclusion-list :

* >100.000 files to exclude and growing, even with optimization done in 0.11.0 and 0.12.0, scanning file takes 2h30
* no general glob / regexp pattern applicable to reduce exclusion list

With an inclusion list (files-from) :
* I don't detect myself the incorrect parent lookup behavior ; but I encounter a x2 time for my first snapshot with files-from
* I detect that snapshot list would be an issue, with the growing paths attribute

I read the patch and it seems sane, but I surely does not have enough experience on restic codebase to give an informed opinion.

Is there anything I can do to help ? (I'm building and trying this patch on a working repository ; I'll give some feedback on this).

lalmeras:
Working with my repository:
* parent snapshot is correctly detected; cpu and io seems to confirm that useless re-read are not done
* backup elapsed time is consistent with timings on a similar tree that does not need include/exclude tricks
* snapshots list display the expected paths
* backup restore give me back the expected files

lpulley:
> using filesystem snapshots (eg, with btrfs or zfs) as the source for the restic backup but wanting the restic backup path to clearly represent the actual data source

@hamishcoleman This is *exactly* what I was trying to do yesterday and realized it didn't exist. Anxiously awaiting the approval of this PR. I'll probably build 0.12.0 from source with this patch applied just so I can use this functionality!

This is going to make backing up ZFS snapshots clean.

hamishcoleman:
@lpulley I have worked around it for the moment by making my backup script take a temporary snap of the real snap but using a single stable name - then destroying the temporary snap after restic has backed things up - which is working well.

cmd/restic/cmd_backup.go
```diff
@@ -490,6 +500,19 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
...
 
+	for _, file := range opts.PathsFrom {
+		fromfile, err := readLines(file)
+		if err != nil {
+			return err
+		}
+		opts.Paths = append(opts.Paths, fromfile...)
+	}
+
+	// set snapshot paths to targets if not given explicitely
```
chrahunt, line 511:
```suggestion
	// set snapshot paths to targets if not given explicitly
```

doc/040_backup.rst
```diff
@@ -307,6 +307,11 @@ All three arguments interpret the argument ``-`` as standard input.
 In all cases, paths may be absolute or relative to ``restic backup``'s
 working directory.
 
+Note that if you use restic in combination with include files, by default all
+entries of the include files are added to the snapshots paths. This may break
+the finding of parent snapshots. Please consider setting the paths manually
+using ``--set-path`` or ``--set-paths-from`` to reflect what is actually backuped.
```
chrahunt, line 313: A few suggestions:

```suggestion
Note that if you use restic in combination with ``--files-from`` arguments, by default all
entries from the files are added to the snapshot's "paths" field. This may break automatic
finding of parent snapshots if the entries change between multiple runs of restic. Please consider setting the "paths" field manually
using ``--set-path`` or ``--set-paths-from`` to reflect a consistent description of what is being backed up.
```

chrahunt (commented):
2 minor comments, otherwise this LGTM.

chrahunt (approved):

lpulley:
Is this just awaiting review?

aawsome:
> Is this just awaiting review?

From my side: yes

juanrgm:
I wanted to use this option to alter the snapshot paths and use the same paths between platforms (win/unix), but this patch keeps transforming the paths into [absolute](https://github.com/restic/restic/blob/fdbd65485e1d8b4f56159f7120a3277979921f4e/internal/restic/snapshot.go#L38).

```sh
restic init --repo repo
restic backup --repo repo --set-path /custom-data data
restic snapshots --repo repo
```

```
---------------------------------------------------------------------------------------
e69de59e  2021-06-03 18:16:43  ?????????               C:\Program Files\Git\custom-data
---------------------------------------------------------------------------------------
1 snapshots
```

You can bypass this feature specifying the driver letter (ex: `X:\custom-path`), but this workaround is not cross-platform.

sydney-d:
Hello,

What's preventing this pull request from being merged ? 

Thanks

aawsome:
> I wanted to use this option to alter the snapshot paths and use the same paths between platforms (win/unix), but this patch keeps transforming the paths into [absolute](https://github.com/restic/restic/blob/fdbd65485e1d8b4f56159f7120a3277979921f4e/internal/restic/snapshot.go#L38).

> You can bypass this feature specifying the driver letter (ex: X:\custom-path), but this workaround is not cross-platform.

You are right. In the snapshots, only absolute paths are saved. The code that turns a path into an absolute one is platform-dependent and therefore prevents you from using this PR to simulate on Windows that the snapshot was created under Linux.

The problem is, if we omit this, then the users always have to specify absolute paths with this new option. Should I change the behavior? Opinions?

nicnab:
I have a similar situation where the data was backed up via ssh so far and now I want to use sshfs so the path changes but the data is all the same. I think it'd be nice if the path was _exactly_ as specified in the --set-path option.

juanrgm:
It seems that to skip the path conversion with `--set-path` is not sufficient because the conversion is used at [FindLatestSnapshot](https://github.com/restic/restic/blob/74c0607c9222edec3b0c140bb6fee962d6d2e82d/internal/restic/snapshot_find.go#L22) too, and that function is used in the next commands: backup, copy, dump, find, forget, ls, recover, restore, snapshots, stats and tag. 

Implementing `--set-path` in all those commands would not be correct.

Maybe a new option (`--raw-paths`) cover all this needs so... new PR?

https://github.com/juanrgm/restic/compare/backup-set-path...backup-raw-paths

wren:
@juanrgm It does seem like there are already a few issues in the backlog about how paths are handled. And while I would love to see restic support for relative paths, I completely agree it's out of scope for this PR.

In my opinion, `--set-path` should resolve a path the same way the rest of restic does so. And it sounds like that's already what you're doing here. Am I reading that correctly?

juanrgm:
> @juanrgm It does seem like there are already a few issues in the backlog about how paths are handled. And while I would love to see restic support for relative paths, I completely agree it's out of scope for this PR.

I will create formally a PR  for this issue when I have time.

> In my opinion, `--set-path` should resolve a path the same way the rest of restic does so. And it sounds like that's already what you're doing here. Am I reading that correctly?

Yep, @aawsome added `--set-path` with the same behaviour than rest of restic.

wren:
It seems like this PR should be ready, then, doesn't it?

I'm not familiar with the process here at restic. Does someone need to be notified for a review or anything like that?

rawtaz:
@wren There is no further action needed, this PR is in the queue along with the others. It needs review and consideration, which will happen with time.

For anyone wanting to try out this PR's feature, it's very simple to check out the code in this PR's branch and build restic by `go run build.go`. Give it a spin if you want :)

haslersn:
How does this address #2092?

rawtaz:
@aawsome I'm thinking it would be nice to use the same form of the noun `path` in both arguments, e.g. `--set-path` and `--set-path-from`, considering that (according to your first sentence) both of them take one or more path(s).

For example, there's not really anything suggesting that just because one feeds path(s) from a file it's going to be multiple paths - might as well be one single path generated by some external software :) Also, it would be less confusing and more coherent.

What do you think, does that make sense?

aawsome:
> @aawsome I'm thinking it would be nice to use the same form of the noun `path` in both arguments, e.g. `--set-path` and `--set-path-from`, considering that (according to your first sentence) both of them take one or more path(s).

Fully agree! Thanks for the suggestion :+1: . I'll change it immediately.

aawsome:
is updated and rebast to current master.
I also marked the flag `--stdin-filename` as deprecated as this can be now done by `--set-path`.

cmd/restic/cmd_backup.go
```diff
@@ -270,6 +280,12 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 		if len(opts.FilesFromRaw) > 0 {
 			return errors.Fatal("--stdin and --files-from-raw cannot be used together")
 		}
+		if len(opts.Paths) > 0 {
+			return errors.Fatal("--stdin and --set-path cannot be used together")
+		}
+		if len(opts.PathsFrom) > 0 {
+			return errors.Fatal("--stdin and --set-paths-from cannot be used together")
+		}
```
rsnitsch, line 288: This is at conflict with the deprecation of the `--stdin-filename` option at [R120](https://github.com/restic/restic/pull/3200/files#diff-30d8672eafa9c0585766b6f437530377563cd1aac5606674c366f3ede4abea9eR120).

If ``--set-path``/``--set-paths-from`` is supposed to replace ``--stdin-filename``, then using ``--stdin`` together with them must be possible.

cmd/restic/cmd_backup.go
```diff
@@ -115,11 +117,17 @@ func init() {
...
+	if err != nil {
+		// MarkDeprecated only returns an error when the flag could not be found
+		panic(err)
+	}
+
 	f.Var(&backupOptions.Tags, "tag", "add `tags` for the new snapshot in the format `tag[,tag,...]` (can be specified multiple times)")
 
 	f.StringVarP(&backupOptions.Host, "host", "H", "", "set the `hostname` for the snapshot manually. To prevent an expensive rescan use the \"parent\" flag")
 	f.StringVar(&backupOptions.Host, "hostname", "", "set the `hostname` for the snapshot manually")
-	err := f.MarkDeprecated("hostname", "use --host")
```
damoclark, line 122: Hi @aawsome 

Just noticed this line, where assignment was changed to equality.  Was this intended?

FallenWarrior2k: Isn't that just a change from declaration assignment to plain assignment? Since `err` is now declared further up in line 120.

aawsome: exactly!

damoclark: > Isn't that just a change from declaration assignment to plain assignment? Since err is now declared further up in line 120.
Thanks for the explanation.  Indeed my mistake - I assumed Go used Pascal syntax.

flokli:
I've been using this PR on top of https://github.com/flokli/borg2restic to migrate from a mounted borg repo to restic. I can confirm it still does what it's supposed to do on `0.14.0` just fine (only needs a minimal rebase).

Hurricos:
Reading through this pull request, the two remaining blockers appear to be:

- [Option cleanup, given the deprecation of `--stdin-filename`](https://github.com/restic/restic/pull/3200#discussion_r825264537), and
- the inclusion of tests. For guidance on this one, [here](https://github.com/restic/restic/pull/3850) is a merged PR which manipulates tests quite a bit to bring them to Go 1.19.

MichaelEischer:
> Reading through this pull request, the two remaining blockers appear to be:

There has been _no_ feedback regarding the actual design of this PR yet from the core devs. The PR is by now relatively far up in my review queue, but it will still take some more time to get there. From a quick look, my main criticism is that it allows setting arbitrary paths which don't reflect the structure within a snapshot in any way. What's the use case for that? (To remove a prefix from all paths, I'd rather prefer something like `--strip-prefix /prefix/path`)

ArsenArsen:
I use that behavior from switching prefixes from `/home/.zfs/snapshot/restic2022-08-20` to simply `/home`.

Resulting snapshots:

```
[c] ~$ restic snapshots 
repository f84e9c1d opened successfully, password is correct
found 3 old cache directories in /home/arsen/.cache/restic, run `restic cache --cleanup` to remove them
ID        Time                 Host              Tags        Paths
------------------------------------------------------------------
758dbc14  2022-02-05 13:41:43  bstg.arsen.local              /home
010297d2  2022-02-07 03:00:01  bstg.arsen.local              /home
...
```

DRON-666:
> What's the use case for that?

I create lists of files to backup mostly dynamically and use `--set-path` with fake paths like `A:\SOMETHING`.
This solves two problems for me:
 - no need to bother with getting the snapshot ID for the `--parent` flag
 - logs no longer overflows with multi-megabyte lists of unnecessary paths (not all commands support the `--compact` flag).

ArsenArsen:
looks like findParentSnapshot gets called with the unoverriden path, though; not sure if that's intentional

aawsome:
I thought about this a bit. Actually I don't like this idea not that much as it allows to set paths which do not correspond to the trees within the snapshot.
Actually a much better idea would be to exchange paths as well in the snapshot entry as well as in the tree structure, see e.g. https://github.com/rustic-rs/rustic/pull/280.

So, I'm closing this PR hoping someone will find time to implement a better solution.

damoclark:
I think there is a clear need for this type of functionality.  

Unfortunately, I suspect that currently proposed solutions don't really address the underlying issue.  They will provide solutions to many existing use cases, but as @aawsome and @MichaelEischer have noted, it does introduce other problems.  

I'd like to propose a radically different approach that I think will address the underlying issue itself.  Unfortunately what I propose will almost certainly require changes to the repository format, necessitating a new repo version.  Thus, I understand adoption of this approach will need to meet a high bar in terms of its value proposition.  What follows is an explanation of the underlying issue as I understand it, an explanation of the new approach and it's benefits, examples of how it would be used, and my limited understanding of potential obstacles or issues, along with an invitation for critique and alternate suggestions.  I accept that I may have some things wrong - happy to be (respectfully) corrected.

But first, I would like to say that Restic is truly remarkable technology.  My sincere congrats to the creators on an amazing tool.  And more broadly, well done to all the contributors who have made Restic what it is today.  My contribution in this post, is well-meaning and constructive.  

Restic provides a repository that can be shared among multiple hosts.  Which of course is really neat.  To identify which backup snapshots belong to which hosts, the hostname, by default is attached to each backup snapshot.  What is backed up in a snapshot on that host, is recorded by one or more absolute paths.  Unlike the host, which is but a simple text label, paths have intrinsic meaning.  And a snapshot can comprise multiple paths.  

A backup is a snapshot in time of data stored on a particular host, at one or more particular absolute paths on that host.  For the purpose of this writing, I am defining a backup 'set' as all snapshots in the repository for the combination of absolute paths and host.  This unit is important for how we identify files we wish to restore in the dimensions of host, path location and time.  It is also important for subsequent backups that seek to identify their parent backup snapshot (i.e. --parent).  This triangulated identification of backup snapshot sets is where I see a problem.  

The flaw in this strategy I think is that Restic tightly couples the identification of backup snapshot sets to absolute paths on the host virtual filesystem.  Or on windows, the drive letters.  The core issue is that these absolute paths have their own meaning, but they can also change.  

The virtual filesystem comprises separate logical and/or physical storage devices attached at different path locations.  In the simple case of windows, separate storage attached to different drive letters.  Filesystem boundaries are significant.  Restic already has a --one-file-system option that recognises this importance.  Filesystem snapshots are a clear example.  And while the virtual filesystem,  using fstab, has pre-determined mount points, there are exceptions.  Notably attachment of external storage devices.  Linux and macOS use filesystem volume labels often mounting under /var/media and /Volumes respectively, which can obviously collide.  While Windows can assign arbitrary drive letters to attached storage devices, influenced by the order of attachment.  

Working around this variability by fudging absolute paths to correctly align with backup snapshot sets is complicated and risky for mere humans - it's easy to bugger it up!  I suspect it is also making it difficult to implement some requested features that relate to paths, such as path translations because of how it is tethered to identifying backup snapshot sets.  Hope I have understood things correctly here.

What I propose is to decouple the paths as part of the identifier for a backup snapshot set.  Instead, uniquely identify backup sets by a user-defined label alongside the host, but still separate to tags.  So the host/label combination is a unique identifier for backup snapshot sets instead of host/paths.  The labels themselves could be arbitrary, and can be chosen meaningfully by the data owner.  Much like a volume/filesystem label.  But they would be abstract - so they aren't tied to diverse semantics associated with paths.  Like arbitrary drive letters selected by the OS, or absolute virtual filesystem paths.  This concept makes it portable.

From a technical standpoint, which paths are given for the backup at that host/label combination doesn't matter as Restic clearly knows which previous --parent to use.  As Restic currently stands, relative paths can be given for a backup, but Restic still resolves them back to an absolute path for the backup to identify its parent.  So this change would also mean that all backup paths can be considered relative, without conflation of the backup snapshot set identity itself.  I anticipate this opens up much easier implementation strategies for path translations during backup, but also retrospective changes to existing backup snapshot sets, because the paths are no longer tightly coupled to the backup snapshot set identity.  As a concrete example, the tar command's --strip-components --transform --absolute-names I suspect would be much easier to implement.  It also simplifies the --stdin option because you identify the backup with the label, rather than an additional and specific --stdin-filename parameter.  So it removes another special case.

From a usability standpoint, this change also allows Restic to behave with similar semantics to other well established backup tools, such as tar, cpio, zip, etc.  One of things I struggled with when first using Restic was the effective reliance on absolute paths.  Backup utilities that I am familiar with all use semantics that allow for relative paths.  

Another consideration is that in many circumstances, restoring from backups can be a very stressful event.  Turning to backups generally means Plan A & B failed - or there was no Plan A. :) So investing time and effort into the backup strategy in exchange for keeping the restore procedure simple can reduce cognitive load during an emergency situation, limit potential human error, reduce stress, and expedite the recovery.  So while you can rename paths after restoring from backups, I think it is better to design your backups so that restores can be performed with minimal post processing.  This means having paths appropriately captured during the backup phase.  And performing backups per filesystem is a common strategy, because that is how restores are often reconstructed from a hardware failure.  This is my experience anyway.  But I am old. :) 

So how might this look from a usage perspective?

If you perform a tar backup for instance, the unique identifier for the destination backup is the filename (-f option) or stdout.  The selector for the source can be provided using relative or absolute paths.

For the existing traditional case as an example, a backup of /home and /data (i.e. all data files) on a host, could be:

```
tar cf filename.tar /home /data
# or
cd /
tar cf filename.tar home data
```

Restic, could look like the following, where 'data' is the backup label:

```
restic backup -r repo --label data /home /data
# or
cd /
restic backup -r repo --label data home data
```

These commands would store home and data has part of the backup file paths.  Alternatively, you could backup filesystems separately and relative to the mount points.  So for just /home

```
tar cf filename.tar -C /home .
# or
cd /home
tar cf filename.tar .
```

And the equivalent for Restic could be:

```
restic backup -r repo --label home -C /home .
# or
cd /home
restic backup -r repo --label home .
```

If you snapshot /home and mount the snapshot as /mnt/home then tar and Restic could be:

```
tar cf filename.tar -C /mnt/home .
# or
restic backup -r repo --label home -C /mnt/home .

# or if you want to retain the /home prefix

tar cf filename.tar -C /mnt home
restic backup -r repo --label home -C /mnt home
```

If you snapshot multiple filesystems and mount the snapshots for a full system backup (without mucking around with chroot):

```
mount /dev/vg00/root_snap /mnt/backup/root
mount /dev/vg00/home_snap /mnt/backup/root/home
mount /dev/vg00/var_snap /mnt/backup/root/var
restic backup -r repo --label fullbackup -C /mnt/backup/root .
umount -R /mnt/backup/root
```

For @ArsenArsen use case with zfs snapshots e.g. `/home/.zfs/snapshot/restic2022-08-20` :

```
restic backup -r repo --label home -C /home/.zfs/snapshot/restic2022-08-20 .

# if you wish to retain the home prefix, using syntax from tar

restic backup -r repo --label home --transform='s/^home\//' -C /home/.zfs/snapshot/restic2022-08-20 .
```

For @aawsome use case of backups performed on the same physical volume, being attached to cross-platform hosts:

```
# Unix-like
restic backup -r repo --label myfiles -C /var/media/aawsome/myfiles .
# Windows
restic backup -r repo --label myfiles -C X:\ .
```

Example usage with stdin - Backup for a database dump:

```
pg_dump -Fc -Z 0 -f - mydatabase | restic backup -r repo --label mydatabase --stdin
```

So what are the issues?  Here, I can contribute little as I don't have enough skin in the game.  So I am going to assume there are plenty more issues, some substantial with what I propose.  There is also a pretty good chance I've made mistakes with some of the facts I have used to underpin my arguments.  And to this extent, I accept that what I propose may not be viable because the benefits don't outweigh the obstacles (combined).

Potential obstacles include (but not limited to):

- With a new mandatory parameter, script/automation changes will be necessary to use new repo format
- could it make some existing use cases more difficult or impossible;
- it breaks future plans of the core Restic team; or,
- too difficult a change without breaking core philosophy of the technology (i.e. guaranteed backwards compatibility)

So what issues have I missed?  And what mistakes have I made or factors I have overlooked?

Do you agree that this would make Restic easier to use?  Does it create other issues?  Or does it not help or make things worse?

If this is worth exploring further, what other ideas might strengthen this concept?

If you made it this far, thanks for reading. :)

Damo.

aawsome:
@damoclark I agree that using the path or paths to identify backup source location can be very misleading. For example, if you have `/path/a`, `/path/b` and `path/c` and want to backup the first two but omit the third, you can
- run `backup /path/a /path/b`
- run `backup /path` and provide exclude options to exclude `/path/c`
- run `backup --files-from*` with a suitable file source (actually many possibilities here...)
- may be able to run `backup /path --one-file-system` if `/path/c` is located on a different file system

I however think that the discussion about how to identify what we have backup'ed is the second prio, whereas the first prio is about what intention we have to run that backup. And this may vary a lot which becomes mostly clear if we ask the questions: If things change, *what* should be backuped then. For example, what should be backup'ed if `/path/d` is added but `/path/a` is removed and `/path/b` is renamed to `/path/e`? And under what circumstances do we want to (respectively have to) change the backup command (or exclude list or file list)? And when does this impact backup runs like needing to re-read the source if no parent is found?

Now, I fully agree that saving the path(s) cannot fully reflect all situations. But I don't think that adding an additional identifier also simplifies the problem. There are already tags which can be used. And moreover there are simple settings, where the path is able to identify the backup. Moreover, the path (knowing the problems above) still is a very relevant information which helping to find candidates for the snapshots you might need for a given restore duty. Plus, I would reckon a needed additional identifier might irritate new users as they simply wouldn't know what to use and what it is good for - except if they simply use the path they provide to backup.

About the absolute vs. relative paths: You are right here, but from another point of view, relative paths could be always seen as absolute paths (if you take them relative to `/`). In fact the tree stored within a snapshot would always look the same, so it's only about semantics how to create and handle them.


So, I'd vote for keeping the path(s) but trying to rework where there are problems with its usage.

For instance this is what I did in [rustic](https://github.com/rustic-rs/rustic):
- Added more powerful include/exclude orptions which allows users to get results you would get by the various `-from-file` options while still being in the setting of specifying just the "base" path as source.
- The mentioned `--as-path` option which IMO fully allows to tackle the absolute/

aawsome:
Reopened by accident (mouse slip), hence closing this again.

damoclark:
Hey @aawsome I appreciate you engaging with my idea.  

A few clarifications and counter-arguments follow. :) 

> But I don't think that adding an additional identifier also simplifies the problem. There are already tags which can be used. 

The idea of a label isn't to add an identifier for a snapshot set.  It is to replace the existing identifier, which historically is the path, or a composite of multiple paths.  And for which is implicitly stated.  A label would be explicit, and thus, more reliably authoritative.  If it needs to change, that could be done explicitly too, and with far less complexity than a path or combination of paths.

The tags are for a different purpose, in terms of classifying backup snapshots, especially in ways that could encompass different combinations of backup snapshot sets.  And chiefly, a tag internally, or semantically, does not uniquely identify a backup snapshot set.

> And moreover there are simple settings, where the path is able to identify the backup. Moreover, the path (knowing the problems above) still is a very relevant information which helping to find candidates for the snapshots you might need for a given restore duty. 

Information about what paths given to backup with the label would still need to be stored.  They just wouldn't be used as the identifier for that backup set for the purposes of subsequent backups, nor identifying the `--parent`.  At restore time, you could still search based on path.  But you could also more meaningfully specify the label as well, triangulating the exact files you want.  

> Plus, I would reckon a needed additional identifier might irritate new users as they simply wouldn't know what to use and what it is good for - except if they simply use the path they provide to backup.

This is tricky to argue either way, because use cases and contexts can vary considerably for a backup utility.  Ranging from data centres with overworked sys admins, through to home server warriors (old sys admins like me), through to hobbyists backing up their personal computer data.  

Still with that diversity, I'm not convinced the requirement for an identifier would be that much of an irritant, and in many cases, might instead be valued.  Labels are commonplace in technology.  We label just about everything. :) So I just don't see it as a barrier.  But I invite further debate if I have overlooked something.

Also, an explicit label for a backup, in my mind is far more meaningful than the implicit path/s alone upon which the files were mounted when the backup was taken.  With the many cases where these paths can and do change, I just don't think paths are sufficiently authoritative to meaningfully identify backup snapshot sets in anything other than the most trivial of use cases.  And in those cases, the label could simply be the path anyway.  I'm reminded of filesystem labels being used in `/etc/fstab` to map mount locations.  

I didn't mention in my original message (which was already way too long), that I think Restic is also missing opportunities for further user-defined metadata.  The obvious example that came to my mind was a 'notes' or 'description' field in the snapshot records.  The use case for me recently was a manual backup snapshot I took of a host before I performed a substantial change on it.  I wanted to write comments about why I took that backup, what it was for, and when it might be safe to forget it.  

I think the --as-path option is a good stop-gap solution for the meantime.  But long term, by introducing the --as-path option, you allow arbitrary fudging of the paths, that remain the authoritative identifier for backup snapshot sets.  I think this will create far more confusion than a label.  For instance, by fudging the paths, they may collide with other existing path locations within the VFS on the same host.  There is also the complexity of when paths are added or removed for a backup snapshot.  Which parent do they belong to?  It's all determined implicitly.  These examples  too me seems far more likely a cause for confusion with users than using a label.  

What I am proposing is a substantial change, and likely a great deal of work.  So I get that it has to be worth it.

MichaelEischer:
@damoclark Let's move the discussion into a new issue. An already closed pull request isn't the right place.

So far the suggestion sounds like slightly beefed up tags in addition to a `--strip-prefix` or `-C` option. But I'm not exactly sure which problem exactly you're trying to solve: identifying the parent snapshot, efficiently handling renamed mountpoints, or something else? Identifying the parent snapshot could just look at the tags instead of the paths (which would solve the parent snapshot identification problem). Removing a prefix from all backup paths would also be relatively simple to implement (could be used to completely remove the changing part of the filename). 

Regarding snapshot descriptions: have a look at https://github.com/restic/restic/issues/2376.

aawsome:
@damoclark For [rustic](https://github.com/rustic-rs/rustic) I opened three issues to discuss backup label, snapshot descriptions and the relative/absolute path discussion, see the links above. Feel free to contribute to the discussion if you want to support the development of rustic.

PR #3204, opened 2020-12-30
archiver: fix race condition during worker startup

MichaelEischer:
I've noticed a test failure (https://github.com/restic/restic/runs/1623490049) after merging #3164.

When the tomb is created with a canceled context, then the workers started via `t.Go` exist nearly immediately. Once for the first time all started goroutines have been stopped, it is not allowed to issue further calls to `t.Go`. This is a problem when the started goroutines exit immediately, as for example the first goroutine might already have stopped before starting the second one, which is not allowed as once the first goroutines has stopped no goroutines were running.

To fix this race condition the startup and main task of the archiver now also run within a `t.Go` function. This also allows unifying the error handling as it is no longer necessary to distinguish between errors returned by the workers or the saveTree processing. The tomb now just returns the first error encountered, which should also be the most descriptive one.

No.

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ **The existing tests still work and I have no idea how to reliably test for this problem.**

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -784,33 +784,31 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 
 	var t tomb.Tomb
 	wctx := t.Context(ctx)
-
-	arch.runWorkers(wctx, &t)
-
 	start := time.Now()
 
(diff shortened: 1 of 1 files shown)


fd0 (approved):
Nice, that's a great improvement!

PR #3208, opened 2021-01-02
Add mips* architectures to CI and release

### LINKED ISSUE #3191: Add MIPS architecture to releases

offer mips architecture releases

i'm running a couple of mips devices all of which do have some state i'd like to backup with restic. they are usually so low on compute power and memory that compiling on them is wishful thinking. crosscompiling on my laptop is something i just haven't figured out, and i thought it might be an easy thing to add to the release process right here :)

srsly :* i've just automated all my backups with ansible and restic :D

fd0:

Add the following four architectures to the CI and release helper scripts:
 * `mips`
 * `mipsle`
 * `mips64`
 * `mips64le`

Closes #3191

Checklist
---------

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -210,7 +210,7 @@ jobs:
           GOX_ARCHS: "aix/ppc64 \
             darwin/amd64 \
             freebsd/386 freebsd/amd64 freebsd/arm \
-            linux/386 linux/amd64 linux/arm linux/arm64 linux/ppc64le \
+            linux/386 linux/amd64 linux/arm linux/arm64 linux/ppc64le linux/mips linux/mipsle linux/mips64 linux/mips64le \
             netbsd/386 netbsd/amd64 \
             openbsd/386 openbsd/amd64 \
             windows/386 windows/amd64 \
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -184,11 +184,29 @@ jobs:
         if: matrix.check_changelog
 
   cross_compile:
+    strategy:
+
+      # ATTENTION: the list of architectures must be in sync with helpers/build-release-binaries/main.go!
+      matrix:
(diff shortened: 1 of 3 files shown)


MichaelEischer:
The previous CI runtime seems to be around 8.5 to 10.5 minutes. (total duration of CI run)

I've started the tests for this PR two more times. The first run took 11m 48s, The second one took ca. 10m30s (the cloud backend tests failed). And the third run took 11m23s. So the additional MIPS targets add 1-2 minutes build time.

Taking a closer look a the build times shows the following pattern: The lint task takes roughly half a minute, The linux and macOS tests less than 3 minutes most of the times. The windows tests take 5-7 minutes and the cross build takes 8.5-10.5 minutes.

It would probably make sense to split the cross building step into two smaller parts. And I wonder why the windows tests are that much slower.

@fd0 There's another interesting thing to note: The "cloud backends" test step isn't run for merges into the main branch, which is probably a bug? For example https://github.com/restic/restic/runs/1633130748 , https://github.com/restic/restic/runs/1627435892 and https://github.com/restic/restic/runs/1626781559 are all recent merges into master, but don't include the "cloud backends" test. The CI run for this PR on the other hand includes it: https://github.com/restic/restic/runs/1637912469 . (The B2 and Swift backends are apparently exceptionally slow in running the test cases.)

fd0:
> So the additional MIPS targets add 1-2 minutes build time.

I think that's okay for four additional architectures, agreed?

> It would probably make sense to split the cross building step into two smaller parts.

Hm, good idea! I'll try in this PR :)

> The "cloud backends" test step isn't run for merges into the main branch

That's odd, I'll have a look.

fd0:
Done :)

MichaelEischer (approved):
LGTM. Do we want a changelog entry like in https://github.com/restic/restic/pull/2342 ?

fd0:
Yes, I'll add one.

whysthatso:
great stuff, everybody, thank you so much :heart:

PR #3211, opened 2021-01-03
Speed-up caching/pack download via SFTP

MichaelEischer:
Running the `check` command using SFTP took longer (12 minutes) than e.g. using rclone via ssh as backend (8 minutes) with the backup repository located on a server in the local network. This shows that the SFTP communication pattern bottlenecks the performance. This PR makes sure that the latency optimized `WriteTo` method of the SFTP reader is used which reduces the run time of my check example to 8 minutes.

The main difference between the REST interface used for rclone and the reader provided by the SFTP library is that the former requests the relevant file part in a single request, whereas the latter issues requests for small parts to read on demand. The on-demand part becomes a bottleneck when the latency to the SFTP server exceeds a few hundred microseconds (**micro** and not milli!). The SFTP reader provides a `WriteTo` method which allows it to request and stream larger parts of a file at once, which fixes the latency problem. `io.Copy` which is used when downloading a pack file to the cache or downloading it to check the pack hash, tries to use `WriteTo` by default (then `ReadFrom`) and then falls back to reading and writing a 32kB buffer. However, neither the `hashing.Reader` (only used for the `check` command and the repack step of the `prune` command) and the `limiter.limitedReadCloser` (is always used even without bandwidth limits) supported `WriteTo` such that `io.Copy` had to fallback to read and write small file chunks. As the read method has to request 32kB data and wait for the reply it is very sensitive to communication latency to the SFTP server.

This PR adds support for `WriteTo` to the `hashing.Reader` and the `Load` method of the `limiter.LimitBackend`. This allows `io.Copy` to use the optimized `WriteTo` code path if supported by the underlying reader. The `WriteTo` method is only added if the underlying reader support it, thus providing an automatic fallback to the normal `io.Copy` behavior if possible.

The upload path is adapted to always use `ReadFrom` provided by the SFTP library for best performance. This is a one-line change as the `io.Copy` call inside the SFTP backend has access to the writer returned by the library.

The core part of this PR is only about 80 lines long, the remainder are tests. A significant part of the tests were necessary as `./internal/limiter` previously had no test coverage at all.

No.

Checklist
---------

- [x] I have added tests for all changes in this PR.

### DIFF
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -1829,3 +1829,53 @@ func TestDiff(t *testing.T) {
 		rtest.Assert(t, r.MatchString(out), "expected pattern %v in output, got\n%v", pattern, out)
 	}
 }
+
+type writeToOnly struct {
+	rd io.Reader
+}
+
+func (r *writeToOnly) Read(p []byte) (n int, err error) {
+	return 0, fmt.Errorf("should have called WriteTo instead")
+}
+
+func (r *writeToOnly) WriteTo(w io.Writer) (int64, error) {
+	return io.Copy(w, r.rd)
+}
+
+type onlyLoadWithWriteToBackend struct {
+	restic.Backend
+}
+
+func (be *onlyLoadWithWriteToBackend) Load(ctx context.Context, h restic.Handle,
+	length int, offset int64, fn func(rd io.Reader) error) error {
+
+	return be.Backend.Load(ctx, h, length, offset, func(rd io.Reader) error {
+		return fn(&writeToOnly{rd: rd})
+	})
+}
+
+func TestBackendLoadWriteTo(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	// setup backend which only works if it's WriteTo method is correctly propagated upwards
+	env.gopts.backendInnerTestHook = func(r restic.Backend) (restic.Backend, error) {
+		return &onlyLoadWithWriteToBackend{Backend: r}, nil
+	}
+
+	testSetupBackupData(t, env)
+
+	// add some data, but make sure that it isn't cached during upload
+	opts := BackupOptions{}
+	env.gopts.NoCache = true
+	testRunBackup(t, "", []string{filepath.Join(env.testdata, "0", "0", "9")}, opts, env.gopts)
+
+	// loading snapshots must still work
+	env.gopts.NoCache = false
+	firstSnapshot := testRunList(t, "snapshots", env.gopts)
+	rtest.Assert(t, len(firstSnapshot) == 1,
+		"expected one snapshot, got %v", firstSnapshot)
+
+	// test readData using the hashing.Reader
+	testRunCheck(t, env.gopts)
+}
--- a/internal/hashing/reader.go
+++ b/internal/hashing/reader.go
@@ -5,25 +5,47 @@ import (
 	"io"
 )
 
-// Reader hashes all data read from the underlying reader.
-type Reader struct {
-	r io.Reader
+// ReadSumer hashes all data read from the underlying reader.
+type ReadSumer interface {
+	io.Reader
+	// Sum returns the hash of the data read so far.
+	Sum(d []byte) []byte
+}
+
+type reader struct {
+	io.Reader
 	h hash.Hash
 }
 
-// NewReader returns a new Reader that uses the hash h.
-func NewReader(r io.Reader, h hash.Hash) *Reader {
-	return &Reader{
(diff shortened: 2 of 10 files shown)


internal/hashing/reader_test.go
```diff
@@ -7,8 +7,29 @@ import (
 	"io"
 	"io/ioutil"
 	"testing"
+
+	rtest "github.com/restic/restic/internal/test"
 )
 
+type onlyReader struct {
+	rd io.Reader
```
aawsome, line 15: Here just a struct with `io.Reader` should be enough.

internal/hashing/reader.go
```diff
@@ -5,25 +5,51 @@ import (
...
 	r io.Reader
 	h hash.Hash
 }
 
-// NewReader returns a new Reader that uses the hash h.
-func NewReader(r io.Reader, h hash.Hash) *Reader {
-	return &Reader{
+type readWriterTo struct {
+	reader
+	writerTo io.WriterTo
```
aawsome, line 22: There is no need to name the `io.WriterTo` here..

cmd/restic/integration_test.go
```diff
@@ -1829,3 +1829,53 @@ func TestDiff(t *testing.T) {
 		rtest.Assert(t, r.MatchString(out), "expected pattern %v in output, got\n%v", pattern, out)
 	}
 }
+
+type writeToOnly struct {
+	rd io.Reader
```
aawsome, line 1834: `io.Reader` would suffice...

aawsome (commented):
just a few comments about the structs and interfaces you are using...

internal/hashing/reader.go
```diff
@@ -5,25 +5,51 @@ import (
 	"io"
 )
 
-// Reader hashes all data read from the underlying reader.
-type Reader struct {
+// ReadSumer hashes all data read from the underlying reader.
+type ReadSumer interface {
+	Read(p []byte) (int, error)
```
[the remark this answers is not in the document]
MichaelEischer: I've switched to using `io.Reader`.

internal/limiter/limiter_backend_test.go
```diff
@@ -0,0 +1,115 @@
...
+	limbe := LimitBackend(be, limiter)
+
+	rd := restic.NewByteReader(data)
+	err := limbe.Save(context.TODO(), testHandle, rd)
+	rtest.OK(t, err)
+}
+
+type tracedReadWriteToCloser struct {
+	rd     io.Reader
+	wt     io.WriterTo
```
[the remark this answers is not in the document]
MichaelEischer: is changed.

internal/hashing/reader.go
MichaelEischer: I'd prefer to keep the name, as the goal here is not to directly expose the `io.WriterTo` object but instead to always wrap it.

internal/limiter/limiter_backend.go
```diff
@@ -42,18 +42,36 @@ func (l limitedRewindReader) Read(b []byte) (int, error) {
...
-			limited: r.limiter.Downstream(rd),
-		}
-		return consumer(lrd)
+		return consumer(newDownstreamLimitedReadCloser(rd, r.limiter, nil))
 	})
 }
 
+type limitedReadWriteToCloser struct {
+	limitedReadCloser
+	writerTo io.WriterTo
```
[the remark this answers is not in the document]
MichaelEischer: `writerTo` is not intended to be exposed directly.

cmd/restic/integration_test.go
MichaelEischer: The `io.Reader` should only be accessed via `WriteTo`. The `Read` method of `rd` must not be accessible.

internal/limiter/static_limiter_test.go
```diff
@@ -0,0 +1,112 @@
...
+		test.Equals(t, limiter.UpstreamWriter(writer) != writer, mustWrapUpstream)
+
+		mustWrapDownstream := limits.downstream > 0
+		test.Equals(t, limiter.Downstream(reader) != reader, mustWrapDownstream)
+		test.Equals(t, limiter.DownstreamWriter(writer) != writer, mustWrapDownstream)
+	}
+}
+
+type tracedReadCloser struct {
+	rd     io.Reader
```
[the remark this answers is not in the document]
MichaelEischer: is changed

internal/hashing/reader_test.go
MichaelEischer: is changed

fd0 (approved):
LGTM!

PR #3217, opened 2021-01-05
Fix missing rand seed for restic check --read-data-subset=x%

M4a1x:
`rand.Perm()` in `restic/cmd/restic/cmd_check.go:332` always returns the same permutation. Thus this PR seeds rand with the current time in nanoseconds. 

The code worked before because since Go 1 the map iteration is in random order. I would still keep the permutation because the language specification only says "not specified", instead of specifically random.

No, since I only noticed it while trying to figure out which kind of random selection to expect when using the new --read-data-subset=x%. Since this is quite a simple "fix" I thought I'd just create this PR and see what you think.

rand.Perm() needs to be seeded https://github.com/golang/go/issues/20971
Golang Specs: https://golang.org/ref/spec#For_statements (Point 3 under "For statements with range clause")
Go 1 Release Notes: https://golang.org/doc/go1#iteration

I don't know a lot of Go, I found most of this through Google.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -5,6 +5,7 @@ import (
 	"math/rand"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


cmd/restic/cmd_check.go
```diff
@@ -329,7 +330,9 @@ func selectRandomPacksByPercentage(allPacks map[restic.ID]int64, percentage floa
 	if packsToCheck < 1 {
 		packsToCheck = 1
 	}
-	idx := rand.Perm(packCount)
+	timeNs := time.Now().UTC().UnixNano()
```
MichaelEischer, line 333: I'd prefer to use `time.Now().UnixNano()` here, which is also already used in a few test. The conversion to UTC doesn't have any effect here.

MichaelEischer (approved):
LGTM. I've rebased the PR and removed the UTC conversion.

Issue #3216, opened 2021-01-04
Prevent Windows going to sleep while restic operations in progress


JsBergbau:
Output of `restic version`
--------------------------
restic 0.11.0 compiled with go1.15.3 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I'd like that restic informs Windows not to go into sleepstates when restic is performing operations.

The way is documented here https://docs.microsoft.com/de-de/windows/win32/api/winbase/nf-winbase-setthreadexecutionstate?redirectedfrom=MSDN
I think "ES_SYSTEM_REQUIRED" is the right value.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

To save power my PC goes into sleep state after an hour of inactivity. However opertions like restic prune could last a lot longer.

Currently `powercfg /requests` does not show any request when restic is running.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yes, restic made me very happy today. Upgraded from 0.9.6 to 0.11.0 and noticed that backup is much faster now.

DRON-666:
You can switch power schemes to 'Always on' and back by adding [`powercfg -setactive`](https://docs.microsoft.com/en-us/windows-hardware/design/device-experiences/powercfg-command-line-options#setactive-or-s) to you scripts before and after `restic ...` command.
Or use some surrogates like [Caffeine](https://www.zhornsoftware.co.uk/caffeine/).

JsBergbau:
Thanks for the hint. These tools could be an option. Still I'd prefer integrating the API call directly into restic.

kmwoley:
If this were to be added, please do not make it on by default. It would result in an unexpected behavior for most people. For me, I *want* Windows to go to sleep during restic operations. Sleeping / saving power is more important to me than restic finishing a background operation.

I personally think this not be a feature of restic, but instead should be left to the person to implement something in a script to prevent Windows from sleeping.

FYI - Windows going to sleep during a backup (or other operation) doesn't cause any data loss, corruption or other errors in my experience.

MichaelEischer:
Closing, see https://github.com/restic/restic/pull/3218#issuecomment-1220958504

PR #3218, opened 2021-01-05
Prevent Windows going to sleep while restic operations in progress

DRON-666:

From #3216:
>To save power my PC goes into sleep state after an hour of inactivity. However opertions like restic prune could last a lot longer.
>Currently powercfg /requests does not show any request when restic is running.
>I'd like that restic informs Windows not to go into sleepstates when restic is performing operations.

Some popular console apps like [Wget](https://github.com/mirror/wget/blob/9f3df123bbe1cbf5873cea25169da4568122f026/src/mswindows.c#L433) and [Rar](https://github.com/pmachapman/unrar/blob/d559fa4c2d19d442a8109ad435611c735efa2cfe/system.cpp#L96) already do this.

In this PR I prefer to use [`Power*Request`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-powercreaterequest) API instead of XP-era `SetThreadExecutionState`: new API has a clear scope (owned kernel object instead of a global flag) and allows describing the reason of sleep blocking.

Issue #3216

Checklist
---------

~~- [ ] I have added tests for all changes in this PR~~
~~- [ ] I have added documentation for the changes (in the manual)~~

### DIFF
--- a/changelog/unreleased/pull-3218
+++ b/changelog/unreleased/pull-3218
@@ -1,0 +1,7 @@
+Enhancement: Prevent Windows entering sleep/hibernation-mode
+
+Previously, restic didn't prevent Windows from going to sleep 
+during some lengthy operations. This is fixed now.
+
+https://github.com/restic/restic/pull/3218
+https://github.com/restic/restic/issues/3216
--- a/cmd/restic/main_windows.go
+++ b/cmd/restic/main_windows.go
@@ -1,0 +1,88 @@
+// +build windows
+
+package main
+
+import (
+	"fmt"
+	"syscall"
+	"unsafe"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/sys/windows"
+)
+
+func init() {
+	f := cmdRoot.PersistentPreRunE
+	cmdRoot.PersistentPreRunE = func(c *cobra.Command, args []string) error {
+		if h, err := PowerCreateRequest(fmt.Sprintf("Restic %s in progress.", c.Name())); err == nil {
+			_ = PowerSetRequest(h, PowerRequestSystemRequired)
+			_ = PowerSetRequest(h, PowerRequestExecutionRequired)
+		}
+
+		return f(c, args)
+	}
+}
+
+type PowerRequestType int
+
+//nolint:deadcode
+const (
+	PowerRequestDisplayRequired PowerRequestType = iota
+	PowerRequestSystemRequired
+	PowerRequestAwayModeRequired
+	PowerRequestExecutionRequired
+)
+
+func PowerCreateRequest(reason string) (handle windows.Handle, err error) {
+	if err := procPowerCreateRequest.Find(); err != nil {
+		return windows.InvalidHandle, err
+	}
+	var reasonPtr *uint16
+	if reasonPtr, err = windows.UTF16PtrFromString(reason); err != nil {
+		return windows.InvalidHandle, err
+	}
+	ctx := powerReasonContext{
+		Version:            0, // POWER_REQUEST_CONTEXT_VERSION
+		Flags:              1, // POWER_REQUEST_CONTEXT_SIMPLE_STRING
+		SimpleReasonString: reasonPtr,
+	}
+	r0, _, e1 := syscall.Syscall(procPowerCreateRequest.Addr(), 1, uintptr(unsafe.Pointer(&ctx)), 0, 0)
+	handle = windows.Handle(r0)
+	if handle == windows.InvalidHandle {
+		if e1 != 0 {
+			err = e1
+		} else {
+			err = syscall.EINVAL
+		}
+	}
+
+	return
+}
+
+func PowerSetRequest(request windows.Handle, requestType PowerRequestType) (err error) {
+	r1, _, e1 := syscall.Syscall(procPowerSetRequest.Addr(), 2, uintptr(request), uintptr(requestType), 0)
+	if r1 == 0 {
+		if e1 != 0 {
+			err = e1
+		} else {
+			err = syscall.EINVAL
+		}
+	}
+
+	return
+}
+
+type powerReasonContext struct {
+	Version            uint32
+	Flags              uint32
+	SimpleReasonString *uint16
+	_                  uint32
+	_                  uint32
+	_                  **uint16
+}
+
+var (
+	kernel32Dll            = windows.NewLazySystemDLL("kernel32.dll")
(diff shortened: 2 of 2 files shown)


JsBergbau:
Hi DRON-666 thanks for creating a pull request for my suggestion for preventing windows going to sleep while restic operations run.

Sorry I don't understand the code fully, so I have to questions:

```
const (
	PowerRequestDisplayRequired PowerRequestType = iota
	PowerRequestSystemRequired
	PowerRequestAwayModeRequired
	PowerRequestExecutionRequired
)
```
I see `PowerRequestDisplayRequired` in changed code. Does that mean display would stay on? If so: Display is not needed, I prefer turning it off. 

If I understand correctly, the pull request just turns it on for prune operation. However a restic backup or restore could also take longer and so these operations should also be covered. So would be great if you could fix this. My suggestion. Just issue wake lock when starting restic and releasing when restic closes. Overhead should be minimal for operations like just printing help or version.

DRON-666:
>I see PowerRequestDisplayRequired in changed code. Does that mean display would stay on?

This is just a declaration of [constants](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-powersetrequest#powerrequestdisplayrequired), see [lines 18-19](https://github.com/restic/restic/pull/3218/commits/87eedfc90a8ba5c5d58aa45707cea4174c8ef508#diff-5699b47111c5c799cd60dd6d1c90669684ac69341f7849da906749d3108b3b69R18-R19) for their use.

>Just issue wake lock when starting restic and releasing when restic closes.

That's how it's already done in this PR.

kmwoley:
If this were to be added, please do not make it on by default. Please give it a command line option to enable. On-by-default result in an unexpected behavior people IMHO. 

For me, I want Windows to go to sleep during restic operations. Sleeping / saving power is more important to me than restic finishing a background operation. When I close my laptop lid, I expect my machine to go to sleep. If it doesn't, I can end up with unexpected situations such as my machine burning power while enclosed in a bag, overheating the machine, coming back to a laptop that has drained it's battery when I need it most, etc. Windows has enough problems preventing it from suspending properly, I'd like not to add another.

I personally think this should not be a feature of restic, but instead should be left to the person to implement something in a script to prevent Windows from sleeping. I understand that others may have a different viewpoint there, however.

FYI - Windows going to sleep during a backup (or other operation) doesn't cause any data loss, corruption or other errors in my experience.

DRON-666:
>When I close my laptop lid, I expect my machine to go to sleep.

This PR does not interfere with this behavior. Please read [this](https://www.microsoftpressstore.com/articles/article.aspx?p=2201309&seqNum=7):
>There are three kinds of requests that can be used through the Power Request API: a system request, a display request, and an “away-mode” request. The first type requests that the system not automatically go to sleep due to the idle timer (**although the user can still close the lid to enter sleep, for example**), while the second does the same for the display.

JsBergbau:
> For me, I want Windows to go to sleep during restic operations. Sleeping / saving power is more important to me than restic finishing a background operation. When I close my laptop lid, I expect my machine to go to sleep. 

Thats a very good point and I'm quite sure, that closing the lid will still get windows into sleep mode.
For example I had the problem that firefox prevented automatic sleep after 60 minutes. 
grafik

However when going to Start and selecting powersave PC immediately went to sleep mode, as expected. When closing the lid behaviour should be identical. So this PR should only change the requests. When users explicitly put PC into sleep mode like via closing the lid, then IHMO this is still respected.

PS: I prevented firefox from preventing my PC going into sleepmode via `powercfg -requestsoverride process firefox.exe display execution system awaymode`

kmwoley:
Thanks for the explanation. I understand how it should behave. My worry around this issue is mostly that Windows power management sleep behavior has been pretty bad in the past. So much so that I've had supposedly well-behaved applications prevent multiple laptops that I manage from going to sleep when the lids are closed. This is has led me not to trust Windows power management. Given that, I'm generally opposed to making this change without giving the user the option to enable/disable this behavior through a command-line switch or environment variable. And my preference would be off by default.

DRON-666:
>So much so that I've had supposedly well-behaved applications prevent multiple laptops that I manage from going to sleep when the lids are closed.

Sorry, but I don't think this is very valuable information without a proof. And you can easily test this PR: download [sources](https://github.com/DRON-666/restic/archive/no-sleep.zip), compile `go run build.go`, execute dummy command `restic init -r nul`, check (from another console window) power requests `powercfg -requests` (you will see something like `SYSTEM: [PROCESS] \Device\HarddiskVolume5\Restic\BIN\restic.exe Restic init in progress.`) and finally close the lid.

>Given that, I'm generally opposed to making this change without giving the user the option to enable/disable this behavior through a command-line switch or environment variable. And my preference would be off by default.

It makes no sense to add this feature disabled by default, and no one does it. I've given Rar and Wget as examples, but only because a direct link to the source code is available. Many other programs like players, network clients, media converters, installers, etc., use this feature, and I doubt they have a switch for that.

MichaelEischer:
I've been pondering quite a bit about what to do with this PR (and the corresponding issue). In some cases it may be quite useful to let restic ensure that the system stays awake. But I also see the point that this behavior may be unexpected or even unwanted. Another such example I can think of would be the `mount` command (once it exists on windows). I don't restic should prevent the system from sleeping in that case.

In general my feeling is that whatever component is scheduling restic should also manage the wake lock and not restic. So, I'm inclined to close this PR.

rawtaz:
Also, the performance/speed of `prune` has increased since the issue of this PR was opened, might have made the initial problem a bit smaller.

JsBergbau:
> Also, the performance/speed of `prune` has increased since the issue of this PR was opened, might have made the initial problem a bit smaller.

Actually with compression performance of prune is quite slow, see https://github.com/restic/restic/issues/3869

However I'm not so sure anymore if restic really should prevent OS from sleeping. I have seen this behaviour with a lot of other programs and to be honest, I'm now quite annoyed of it. Especially Firefox with a lot of open tabs prevents system from sleeping which is not what I want and you need `powercfg /requestsoverride` to save energy, especially I'm using

```
[SERVICE]

[PROCESS]
firefox.exe DISPLAY SYSTEM AWAYMODE EXECUTION
MoUsoCoreWorker.exe DISPLAY SYSTEM AWAYMODE EXECUTION
JDownloader2.exe DISPLAY SYSTEM AWAYMODE EXECUTION

[DRIVER]
High Definition Audio Device SYSTEM
Realtek High Definition Audio SYSTEM
```

So I'm fine with closing this pullrequest.

PR #3228, opened 2021-01-15
prune: Remove all unused trees

aawsome:

With this PR `prune` removes all unused trees. It does so by marking all packs containing trees for repacking (except when the `--max-repack-size` option is given).

Before, tree packs were priorized as data packs when it came to choosing packs for repacking.
I however think that this wasn't a good decision because:
- Tree packs are usually available in the cache, so repacking those is cheap. It is better to use the tolerated `--max-unused` for data blobs which are more expensive to access
- Tree blobs are usually comparably small and this often leads to small pack files which should be preferably not kept but repacked.
- It is very unlikely that a non-referenced tree can ever be re-used, so there is no sense to keep it.
- Even in the case that a tree would be re-used by a new backup, it is much cheaper to resave it (as trees blobs are usually small)
- Finally: Keeping unreferenced tree blobs may lead to unintended effect, e.g. in `recover`, see the discussion in #2880 

No.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -390,7 +390,7 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 	// This is equivalent to sorting by unused / total space.
 	// Instead of unused[i] / used[i] > unused[j] / used[j] we use
 	// unused[i] * used[j] > unused[j] * used[i] as uint32*uint32 < uint64
-	// Morover duplicates and mixed are sorted to the beginning
+	// Morover duplicates and packs containing trees are sorted to the beginning
 	sort.Slice(repackCandidates, func(i, j int) bool {
 		pi := repackCandidates[i].packInfo
 		pj := repackCandidates[j].packInfo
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
For my last prune run this only slightly increased the amount of data which had to be reuploaded (I've lost the exact numbers :-( ). So that looks ok to me.

I'd also like to add one more reason why trees should always be pruned:
As tree packs are kept in the cache, a lot of garbage data could accumulate there. In face the cache could grow to the actual size of all trees still in use + 5% of the total repository size.

aawsome:
> As tree packs are kept in the cache, a lot of garbage data could accumulate there.

Just had a look at some of my repos. In my cases, actually a lot of small tree pack files accumulate in the repo. 
The current logic leaves tree pack files if they are completely used - even if they just contain a single used tree blob. I think we should also always repack small tree pack files. 
@MichaelEischer what do you think? Should I add a PR to change this? If yes, what do you think would be a good threshold?

MichaelEischer:
Hmm, that's kind of similar to the `--repack-small` option. For already cached tree packs it probably makes no relevant difference whether the tree packs are repacked or not. However, such repacking would increase the amount of changes made to the repository by prune. I'm also not sure whether it's worth to increase the complexity of the prune code when we don't know what the benefit would be.

PR #3121, opened 2020-11-23
backup: Support multiple parents

aawsome:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Support multiple parents in the `backup` command and implements an algorithm to automatically find those.

Some examples: 
With this PR, If you previously backed up `/foo/bar1` and  `/foo/bar2` and now want to backup `/foo`, both previous snapshots are taken as parent. A side effect of the new parent selection algorithm is: If you previously backed up `/foo` previously and now backup `/foo/bar1`, the existing snapshot will be taken as parent.

Note that this PR modifies the `Snapshot` data structure used in the snapshots files by adding a `Parents` field.
The current `Parent` field is only set but never used by restic.
If there is only one (or zero) parent, the `Parent` field is still used, if there are more than one parents, the `Parents` field is now used.

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------

closes #3118 

Checklist
---------

- [x] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [x] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
- [x] I have added tests for all changes in this PR
- [x] I have added documentation for the changes (in the manual)
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))
- [x] I have run `gofmt` on the code in all commits
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits)
- [x] I'm done, this Pull Request is ready for review

PR #3229, opened 2021-01-15
backup: Add --resume option

### LINKED ISSUE #2280: Re-Use information about partial backup to reduce I/O
Output of `restic version`
--------------------------

When a backup is interrupted (especially an initial backup of a large dataset), restic is smart enough not to upload everything again. This allows to perform a large backup on a slow link in "chunks" (backing up as much as possible until it's interrupted, then restart untill everything is initially backed up).

Unfortunately, restic does not seem to re-use the cached checksum information of this partial backup, so every singe file is fully read. That creates unnecessarily high amount of I/O.

What are you trying to do?
--------------------------

I'm backing up some storage systems, having 5-20 TB data each. The initial backup takes >24h. I need do interrupt the backup during business hours. Basically it works but when I resume the backup, restic is reading everything again from disk, saturating I/O of the system until it has read up to the point where the backup was interrupted.

This happens only on incomplete backups, once a backup is completed, I/O is only performed on changed files.

Restic is a great backup tool. We did replace most of our home-brew backup scripts with it and are very happy.

aawsome:

Adds an option `--resume` to the backup command.

When specified, this option does:
- search for unreferences trees, like `recover`. These trees may exist due to an aborted backup
- use these tree as additional "parent trees" for each dir to backup, i.e. do a brute-force check whether these unreferenced trees may contain a suitable parent tree for the dirs to backup

=> This should effectively speed up "resumed backups", i.e. repetitions of backups that aborted and therefore could not write snapshots. It should especially work well if already big parts were processed during that aborted backup run.

There is however the trade-off that searching for unreferenced trees needs to read all trees and all snapshots in the repo.

Note that this PR relies on #2880 and on the first commits of #3121 and is therefore marked as draft. 
Moreover, I strongly recommend to merge #3228 to not have too many unreferenced trees.

The code should work already but I did not do any performance tests yet.

closes #2280 
alternative to #2960 

https://forum.restic.net/t/quicker-interrupted-backup-resumption/3470/6

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-2880
+++ b/changelog/unreleased/pull-2880
@@ -1,0 +1,10 @@
+Enhancement: Improve recover command
+
+Restic recover used to generate a snapshot that contains all root trees
+even those which are already referenced by a snapshot.
+It now only processes trees not referenced at all.
+Also now `recover` has new options that allow to generate one snapshot
+per unreferenced tree and specify which tags should be added to the
+newly generated snapshot(s).
+
+https://github.com/restic/restic/pull/2880
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -92,6 +92,7 @@ type BackupOptions struct {
 	WithAtime               bool
 	IgnoreInode             bool
 	UseFsSnapshot           bool
+	Resume                  bool
 }
 
 var backupOptions BackupOptions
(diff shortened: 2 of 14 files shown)


rawtaz:
Can you tell us more about the actual gains here? How much quicker is it in some example cases, and at what cost? I'm basically asking if it's worth it (after all, the user has to be educated to use --resume *when it makes sense*, and this may be more confusing than it's worth).

MichaelEischer:
To me this looks like `--resume` could take at least as much time as a full `check` run before it can "resume" a backup. And that can take several hours. For larger backups I'd expect that a single interrupted backup can leave hundreds of unreferenced trees behind which will only accumulate of time until `prune` is run. Checking thousands of trees for each file to backup could regress backup performance a lot. After all, not too long ago one could turn `backup` into a crawl by specifying a few thousand exclude patterns.

aawsome:
> To me this looks like `--resume` could take at least as much time as a full `check` run before it can "resume" a backup

I did some first tests, and you are right. Scanning for unreferenced trees may take a *long* time making this only work for aborted initial (or initial-like) backups, which I agree is a too narrow use case. 

That's a pitty - I was hoping it would be possible to gather the needed information by only what is already stored in restic...

So I close this PR.

PR #3248, opened 2021-01-30
Backend code and test cleanups

MichaelEischer:
This is in parts a follow-up to #3176. TestSaveError now uses a separate filename and tries to clean after itself. This should prevent accidentally triggering a "file already exists"-error instead of the error due to an incomplete upload.

The mem backend now checks the upload size before verifying its size. This prevents leftover files. The local / sftp backend still keep failed uploads and only return an error. That way the three backends together offer both ways how incomplete uploads could be handled by backends.

No. I've noticed these problems while testing #3246.

Checklist
---------

- ~~[ ] I have added tests for all changes in this PR~~ Already covered by tests
- ~~[ ] I have added documentation for the changes (in the manual)~~ Not user visible
- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))~~ Not user visible

### DIFF
--- a/internal/backend/mem/mem_backend.go
+++ b/internal/backend/mem/mem_backend.go
@@ -81,13 +81,13 @@ func (be *MemoryBackend) Save(ctx context.Context, h restic.Handle, rd restic.Re
 		return err
 	}
 
-	be.data[h] = buf
-	debug.Log("saved %v bytes at %v", len(buf), h)
-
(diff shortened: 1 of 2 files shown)


fd0 (approved):
LGTM

PR #3249, opened 2021-01-30
gs: Don't drop error when finishing upload

MichaelEischer:
The error returned when finishing the upload of an object was dropped. This could cause silent upload failures and thus data loss in certain cases. When a MD5 hash for the uploaded blob is specified, a wrong hash/damaged upload would return its error via the Close() whose error was dropped.

The problem showed up while testing #3246. I've split the fix into a separate PR to allow merging it independently of the other changes in #3246 as this bug might possibly cause data loss.

Checklist
---------

- [ ] I have added tests for all changes in this PR -> The failing test is in #3246, I'm not sure how to reliably test this otherwise.

### DIFF
--- a/changelog/unreleased/pull-3249
+++ b/changelog/unreleased/pull-3249
@@ -1,0 +1,7 @@
+Bugfix: Better error handling for gs backend
+
+The gs backend did not notice when the last steep of completing a file upload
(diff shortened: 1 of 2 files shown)


fd0 (approved):
LGTM!

PR #3250, opened 2021-01-30
Add golangci lint config, add many error checks

fd0:

Add a configuration file for `golangci-lint` so that we get good results from all the linters. Also, I tried to add all the error checks mentioned by `errcheck`, in order to prevent things like #3249.

No.

Checklist
---------

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,6 +141,14 @@ Installing the script `fmt-check` from https://github.com/edsrzf/gofmt-git-hook
 locally as a pre-commit hook checks formatting before committing automatically,
 just copy this script to `.git/hooks/pre-commit`.
 
+The project is using the program
+[`golangci-lint`](https://github.com/golangci/golangci-lint) to run a list of
+linters and checkers. It will be run on the code when you submit a PR. In order
+to check your code beforehand, you can run `golangci-lint run` manually.
+Eventually, we will enable `golangci-lint` for the whole code base. For now,
+you can ignore warnings printed for lines you did not modify, those will be
+ignored by the CI.
+
 For each pull request, several different systems run the integration tests on
 Linux, macOS and Windows. We won't merge any code that does not pass all tests
 for all systems, so when a tests fails, try to find out what's wrong and fix
--- a/cmd/restic/exclude.go
+++ b/cmd/restic/exclude.go
@@ -180,7 +180,9 @@ func isDirExcludedByFile(dir, tagFilename, header string) bool {
 		Warnf("could not open exclusion tagfile: %v", err)
 		return false
 	}
-	defer f.Close()
+	defer func() {
+		_ = f.Close()
+	}()
 	buf := make([]byte, len(header))
 	_, err = io.ReadFull(f, buf)
 	// EOF is handled with a dedicated message, otherwise the warning were too cryptic
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -566,9 +566,9 @@ func TestBackupErrors(t *testing.T) {
 
 	// Assume failure
 	inaccessibleFile := filepath.Join(env.testdata, "0", "0", "9", "0")
-	os.Chmod(inaccessibleFile, 0000)
+	rtest.OK(t, os.Chmod(inaccessibleFile, 0000))
 	defer func() {
-		os.Chmod(inaccessibleFile, 0644)
+		rtest.OK(t, os.Chmod(inaccessibleFile, 0644))
 	}()
 	opts := BackupOptions{}
 	gopts := env.gopts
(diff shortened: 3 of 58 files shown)


cmd/restic/integration_test.go
```diff
@@ -1747,16 +1747,32 @@ func copyFile(dst string, src string) error {
...
 		return err
 	}
-	defer dstFile.Close()
 
 	_, err = io.Copy(dstFile, srcFile)
-	return err
+	if err != nil {
+		// ignore subsequent errors
+		_ = srcFile.Close()
+		_ = dstFile.Close()
```
MichaelEischer, line 1762: That misses a `return err`.

cmd/restic/integration_test.go
```diff
@@ -1747,16 +1747,32 @@ func copyFile(dst string, src string) error {
...
+	}
+
+	err = srcFile.Close()
+	if err != nil {
+		return err
```
MichaelEischer, line 1767: `dstFile` is not closed. The code might be more readable by using a name return value `err error` and then use a function as follows:
```go
defer func() {
   cerr := srcFile.Close()
   if err == nil {
       err = cerr
   }
}()

File: internal/repository/repository.go [hunk 3]
```diff
@@ -419,7 +419,11 @@ func (r *Repository) saveIndex(ctx context.Context, indexes ...*Index) error {
 
 		debug.Log("Saved index %d as %v", i, sid)
 	}
-	r.idx.MergeFinalIndexes()
+
+	err := r.idx.MergeFinalIndexes()
```
MichaelEischer, line 423: `return r.idx.MergeFinalIndexes()` would be simpler

internal/backend/rclone/internal_test.go
```diff
@@ -23,7 +23,9 @@ func TestRcloneExit(t *testing.T) {
 		return
 	}
 	rtest.OK(t, err)
-	defer be.Close()
+	defer func() {
+		rtest.OK(t, be.Close())
```
MichaelEischer, line 27: That error should be ignored as rclone is killed in line 30

cmd/restic/integration_test.go
fd0: indeed

cmd/restic/integration_test.go
fd0: you're right, I've added `_ = dstFile.Close()` in the `if`. I don't like using named returns as it is very easy to accidentally shadow the named return and thereby dropping errors :confused:

internal/repository/repository.go
fd0: Oh, good catch.

internal/backend/rclone/internal_test.go
fd0: Agreed.

CONTRIBUTING.md
```diff
@@ -141,6 +141,14 @@ Installing the script `fmt-check` from https://github.com/edsrzf/gofmt-git-hook
 locally as a pre-commit hook checks formatting before committing automatically,
 just copy this script to `.git/hooks/pre-commit`.
 
+The project is using the program
+[`golangci-lint`](https://github.com/golangci/golangci-lint) to run a list of
+linters and checkers. It will be run on the code when you submit a PR. In order
+to check your code beforehand, you can run `golangci-lint run` manually.
+Eventually, we will enable `golangci-lint` for the whole code base. For now,
+you can ignore warnings printed for lines you did not modify, those will be
```
DRON-666, line 149: It may be better to use the `--new-from-rev upstream/master` flag than to manually ignore errors.

cmd/restic/exclude.go
```diff
@@ -180,7 +180,9 @@ func isDirExcludedByFile(dir, tagFilename, header string) bool {
 		Warnf("could not open exclusion tagfile: %v", err)
 		return false
 	}
-	defer f.Close()
+	defer func() {
+		_ = f.Close()
+	}()
```
DRON-666, line 185: `gofmt` allows a more compact form `defer func() { _ = f.Close() }()` or just add `// nolint:errcheck` before `defer`.

CONTRIBUTING.md
fd0: Ah, that's a good hint, thanks!

fd0: Hm, it does not work, I've just tested it locally. I'll leave the hint as it is.

cmd/restic/exclude.go
fd0: Thanks for the hint. I find the one line version hard to read, I prefer the version in three lines :)

CONTRIBUTING.md
DRON-666: Strange. It's work on Windows very well.

MichaelEischer (approved):
LGTM.

PR #3254, opened 2021-01-31
Allow HTTP/2

lorenz:

Allows Go to use HTTP/2 for backend conections. This is automatically disabled when using DialContext for compatibility with older versions of Go. Setting `ForceAttemptHTTP2` opts in to the new behavior.
This makes restic much faster with certain backends supporting HTTP/2.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3254
+++ b/changelog/unreleased/pull-3254
@@ -1,0 +1,8 @@
+Enhancement: Enable HTTP/2 for backend connections
+
+Go's HTTP library usually automatically chooses between HTTP/1.x and HTTP/2
+depending on what the server supports. But for compatibility this mechanism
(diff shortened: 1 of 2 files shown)


MichaelEischer:
Good catch :-) . Could you add a short changelog entry according to the template at https://github.com/restic/restic/blob/master/changelog/TEMPLATE ?

lorenz:
I've added a changelog entry. I'm using the REST backend, but terminated behind an ingress (so the REST backend itself doesn't see HTTP/2). On low-latency fiber connections it won't make much of a difference, but HTTP/2 being a single stream helps with bufferbloat-related latency issues on older connections.

fd0 (approved):
Great catch, thanks a lot!

PR #3257, opened 2021-02-01
ls: print "size":0 for empty files in JSON

### LINKED ISSUE #3247: "ls --json " does not output a size for all Windows files
Output of `restic version`
--------------------------
restic 0.11.0 compiled with go1.15.3 on windows/amd64

for backup
powershell script which runs restic with the following options
 -r $RESTIC_REPOSITORY --exclude-file $RESTIC_EXCLUDE_FILE  --use-fs-snapshot --verbose --verbose --cache-dir $CACHEDIR --json ${RESTIC_SOURCE_DRIVE}:\ 2>$RESTIC_STDERR_FILE | Out-File -FilePath .\restic_temp_stdout.log -encoding ascii
for ls
$snapid = 'latest'
ls $snapid -r $RESTIC_REPOSITORY --cache-dir $CACHEDIR | Out-File -FilePath $RESTIC_LS_FILE -encoding ascii

usb connect external harddrive

Expected behavior
-----------------
all files listed with ls should have a size in the json output.

Actual behavior
---------------
Note: there was no error for these files when the backup was run!
size was missing

On a Windows machine backup the whole C: drive. Do a "ls --json". Capture output and look for files which do not have the attribute size. I did this in python pandas.

1) The files may have actually been of zero bytes. 
2) I run the backup program using an Administrator id but the directory
c:\Program Files\WindowsApps is so tightly constrained by windows security that some of the files are not readable.  
3) Are the files actually just links to other files meaning there is no bytes being used in a file at that directory location.

Force the ls command to output a size of zero if there are no bytes being used. 

Yes. Both the program and the community are wonderful.

greatroar:

It causes restic ls --json to print empty files as having "size":0 instead of omitting the size, as was done previously.

It also removes unconditional panics introduced in 200f09522dfe034ced28075705c41de9a5fefebf, which broke JSON printing entirely.

Fixes #3247.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3247
+++ b/changelog/unreleased/issue-3247
@@ -1,0 +1,8 @@
+Change: Empty files now have size of 0 in restic ls --json output
+
+Restic ls --json used to omit the sizes of empty files in its output. It now
+reports "size":0 explicitly for regular files, while omitting the size field
+for all other types.
+
+https://github.com/restic/restic/issues/3247
+https://github.com/restic/restic/pull/3257
--- a/cmd/restic/cmd_ls.go
+++ b/cmd/restic/cmd_ls.go
@@ -74,18 +74,42 @@ type lsSnapshot struct {
 	StructType string     `json:"struct_type"` // "snapshot"
 }
 
-type lsNode struct {
-	Name       string      `json:"name"`
-	Type       string      `json:"type"`
-	Path       string      `json:"path"`
-	UID        uint32      `json:"uid"`
-	GID        uint32      `json:"gid"`
-	Size       uint64      `json:"size,omitempty"`
-	Mode       os.FileMode `json:"mode,omitempty"`
-	ModTime    time.Time   `json:"mtime,omitempty"`
-	AccessTime time.Time   `json:"atime,omitempty"`
-	ChangeTime time.Time   `json:"ctime,omitempty"`
-	StructType string      `json:"struct_type"` // "node"
+// Print node in our custom JSON format, followed by a newline.
+func lsNodeJSON(enc *json.Encoder, path string, node *restic.Node) error {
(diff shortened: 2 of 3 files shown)


fd0 (dismissed):
There's a better way (IMHO): Make `Size` a pointer to `uint64`, and set it to a non-nil value when the size should be printed. The code below misses error handling, but passes the tests you added:

```go
type lsNode struct {
	Name       string      `json:"name"`
	Type       string      `json:"type"`
	Path       string      `json:"path"`
	UID        uint32      `json:"uid"`
	GID        uint32      `json:"gid"`
	Size       *uint64     `json:"size,omitempty"`
	Mode       os.FileMode `json:"mode,omitempty"`
	ModTime    time.Time   `json:"mtime,omitempty"`
	AccessTime time.Time   `json:"atime,omitempty"`
	ChangeTime time.Time   `json:"ctime,omitempty"`
	StructType string      `json:"struct_type"` // "node"
}

// Print node in our custom JSON format, followed by a newline.
func lsNodeJSON(w io.Writer, path string, node *restic.Node) {
	var size uint64

	n := lsNode{
		Name:       node.Name,
		Type:       node.Type,
		Path:       path,
		UID:        node.UID,
		GID:        node.GID,
		Mode:       node.Mode,
		ModTime:    node.ModTime,
		AccessTime: node.AccessTime,
		ChangeTime: node.ChangeTime,
		StructType: "node",
	}

	if node.Type == "file" {
		size = node.Size
		n.Size = &size
	}

	json.NewEncoder(w).Encode(n)
}
```

What do you think?

greatroar:
Good idea, implemented that.

MichaelEischer (approved):
LGTM.

PR #3264, opened 2021-02-03
Refactor backup progress

### LINKED ISSUE #2738: restic backup --json --quiet is not quiet at all
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on linux/arm

`restic backup --json --quiet`
called from a python3 script.

rclone/local

Expected behavior
-----------------
output only one line with `"message_type":"summary"`

Actual behavior
---------------
MANY lines with `"message_type":"status"`

According to `restic help` these should not be there:
```
-q, --quiet                     do not output comprehensive progress report
```

see above

no

probably add an `if` somewhere?

Did restic help you today?
Every day!

amozoss:
`ui/backup.go` and `ui/json/backup.go` are almost identical to one another (copy paste). 

This change refactors the shared logic into `progress.go`. 

`ui/backup.go` and `ui/json/backup.go` have already begun to diverge, for example
```
// for the last item "/", current is nil
if current != nil { 
   p.summary.ProcessedBytes += current.Size
}
```
`p.summary.ProcessedBytes += current.Size` was done outside of the lock in json/backup.go , but ui/backup.go had the above

This will continue to be a "hope we don't forget to update the other place" problem..

This change also allows the progress reporting/printing to be shared with restore status (forth coming after this is merged).

No it wasn't

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2738
+++ b/changelog/unreleased/issue-2738
@@ -1,0 +1,8 @@
+Bugfix: Don't print progress for `backup --json --quiet`
+
+Unlike the text output, the json output format still printed progress
+information even in quiet mode. This has been fixed by always disabling the
+progress output in quiet mode.
+
+https://github.com/restic/restic/issues/2738
+https://github.com/restic/restic/pull/3264
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -537,7 +537,7 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 
 	if opts.DryRun {
 		repo.SetDryRun()
-		progressPrinter.SetDryRun()
+		progressReporter.SetDryRun()
 	}
 
 	// use the terminal for stdout/stderr
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -527,39 +527,17 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 		return err
 	}
 
-	type ArchiveProgressReporter interface {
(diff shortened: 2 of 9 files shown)


greatroar:
I don't agree with omitting "archive" or "backup" from the names of the moved types, which are all specific to the backup command. The generic names ui.Counter and ui.Progress mentally conflict with ui/progress.Counter. If this must be moved out of cmd/restic, I suggest it be moved to a new package, maybe archiver/ui, ui/archiver or ui/backup.

amozoss:
I'm not sure I follow what you're saying. I renamed the interface so that I could re-use the logic of the progress reporting with restores (I haven't submitted that change yet, it depends on this refactor).

MichaelEischer (commented):
@amozoss
I've rebased the commit onto the current master. The main change during the rebase is that there's now a `SetDryRun()` method.

Besides that I've made a few other noteworthy changes:
- The JSON statistics part of the summary output is restored, which was lost during the refactoring.
- I've removed the unused `TotalErrors` field from the `Summary` struct.
- `backup --json | cat -` no longer printed status updates. I've fixed that regression. Now only `backup --json --quiet` disabled the status updates. The latter actually is a change compared to the current behavior where `backup --json --quiet` and `backup --json` behave similarly.

The `Counter` struct had also diverged between the text and JSON output.

Regarding the name of the `Counter` and the `Summary` struct I agree with greatroar that these are pretty generic and could be confusing. The cleanest way would probably to move everything, that is the backup.go files and progress.go, into a package `internal/ui/backup`. However, as you've mentioned that you want to reuse some part of the logic for `restore`, I'm wondering whether some other structure would be preferable.

MichaelEischer:
I've added a changelog that his PR also fixes #2738.

@greatroar The backup ui code is for now consolidated in `ui/backup`.

MichaelEischer (approved):
LGTM.

PR #3270, opened 2021-02-07
ls: Check for non-nil error before calling panic()

fd0:

Don't panic when `restic ls --json` is called. It was accidentally included in the error checking cleanup PR #3250.

No.

Checklist
---------


MichaelEischer (approved):
LGTM.

PR #3290, opened 2021-02-19
prune: Handle duplicate blobs more efficiently

### LINKED ISSUE #3114: prune: Enhance support for duplicate blobs

aawsome:

Enhances treatment of duplicates during `prune`. Now an algorithm is implemented that marks each duplicate either as "used" or as "unused" (such that pack files containing only duplicates should be marked as completely unused and hence can be removed).
As a side effect, `prune` now can keep duplicates if unused space is allowed by `--max-unused` and the statistics are accurate even if not all packs with duplicates are repacked.

closes #3114 (at least all points except the treatment of duplicates where one of the duplicates is damaged)

Checklist
---------

-  I have not added tests for all changes in this PR

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -258,12 +258,11 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 			count, ok := duplicateBlobs[bh]
 			if !ok {
 				count = 2 // this one is already the second blob!
-			} else {
+			} else if count < math.MaxUint8 {
+				// don't overflow, but saturate count at 255
+				// this can lead to a non-optimal pack selection, but won't cause
+				// problems otherwise
 				count++
-				if count == 0 {
-					// catch uint8 overflow
-					panic("too many duplicates, prune can only handly up to 255!")
-				}
 			}
 			duplicateBlobs[bh] = count
 			stats.size.duplicate += size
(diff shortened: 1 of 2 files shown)


aawsome:
I changed the comments of the algorithm to make it more clear which of the duplicates is chosen as "used"

aawsome:
also rebased to current master

cmd/restic/cmd_prune.go
```diff
@@ -234,7 +233,17 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
...
 		case keepBlobs.Has(bh): // duplicate blob
-			duplicateBlobs.Insert(bh)
+			count, ok := duplicateBlobs[bh]
+			if !ok {
+				count = 2 // this one is already the second blob!
+			} else {
+				count++
+				if count == 0 {
+					// catch uint8 overflow
+					panic("too many duplicates, prune can only handly up to 255!")
```
aawsome, line 243: Actually, it is ok to just let it 255 here, see the implementation in rustic

aawsome: see https://github.com/rustic-rs/rustic/blob/main/src/commands/prune.rs#L387

MichaelEischer (approved):
LGTM. Thanks for improving this sort of special case.

I've rebased the PR to the current master, added the duplicated counter saturation at 255 and made a few minor cleanups.

PR #3298, opened 2021-02-23
doc: Add note about bash completion path

jniggemann:
Adds bash-completion example for Debian and derivatives

Documentation had an example that was not applicable to Debian/Ubuntu/Devuan/..., this PR adds an additional example for those systems.

No

Checklist
---------

### DIFF
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -316,3 +316,8 @@ Example for using sudo to write a bash completion script directly to the system-
 
     $ sudo ./restic generate --bash-completion /etc/bash_completion.d/restic
     writing bash completion file to /etc/bash_completion.d/restic
+
+.. note:: The path for the ``--bash-completion`` option may vary depending on
+   the operating system used, e.g. ``/usr/share/bash-completion/completions/restic``
+   in Debian and derivatives. Please look up the correct path in the appropriate
+   documentation.


rawtaz:
Thanks for the PR. With risk of sound really boring and so on, I am not sure we should make this change.

As it is now we already have a long list of distribution-specific instructions in the packages section under the Install chapter. Time has shown that as soon as there was one of them, other distros wanted to be on the list as well, so that noone missed out (which is reasonable).

What I'm getting at is that we add this change, then we will start seeing the same with the section for autocompletion. More or less all of the common distros/systems that doesn't match the current example because they have their autocomplete scripts in some other path, will want to be added to that list with an example of a command that works for them. We'll end up with yet another platform-specific list, and I'm hesitant that there's enough value in that.

I'd argue that the users who use packages to install restic will not have to manually run the autocomplete generation command at all (since the package does that), and those that install it manually have the ability to find the proper path for running this script.

Perhaps we should instead add a note that informs the user that the path in the example is just an example, and that they should make sure to look up the right one for their specific OS. But at this point in time I don't think we should start yet another platform specific list.

jniggemann:
> Thanks for the PR. With risk of sound really boring and so on, I am not sure we should make this change.
No problem, IMHO a PR is just a proposal. It can be accepted or rejected.

> What I'm getting at is that we add this change, then we will start seeing the same with the section for autocompletion. More or less all of the common distros/systems that doesn't match the current example because they have their autocomplete scripts in some other path, will want to be added to that list with an example of a command that works for them. We'll end up with yet another platform-specific list, and I'm hesitant that there's enough value in that.
True, although IMHO the vast majority of linux distros are either RedHat, Debian or Slackware (and their derivatives).

Let's let this sit for a couple of days and if no one is interested, feel free to close this.
After all, it's just a couple of lines in the docs :-)

rawtaz:
@jniggemann What do you say about changing your PR to add the following notice "box" in the documentation, below the current paragraph and code example:

~~~
.. note:: The path for the ``--bash-completion`` option can vary depending on
   operating system used, e.g. ``/usr/share/bash- completion/completions/restic``
   in Debian and derivatives. Please look up the correct path in the appropriate
   documentation.
~~~

That way it's more agnostic and general. If this sounds good, can you update your PR (squash into one commit and rebase on master)?

rawtaz:
Thank you! I realize maybe "may" is a better word than "can", would you mind fixing that if it's relevant and then squash the commits before we merge this? Sorry about the extra work.

jniggemann:
This is way better, good idea!
Never used squash and / or rebase, so I added another commit to the PR.

jniggemann:
> Thank you! I realize maybe "may" is a better word than "can", would you mind fixing that if it's relevant and then squash and rebase (on master) the commits before we merge this? Sorry about the extra work.
"may" is indeed better, I also added "the" before OS. I'll promise to read up on squashing :-)

rawtaz:
That's fine, I can polish that last thing and squash and rebase :)

rawtaz (approved):

PR #3210, opened 2021-01-03
backup: add --dry-run/-n flag to show only proposed changes

erincerys:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Implements `backup --dry-run` (#1542), continuing where @rmmh left
off in #2308 

This can be used to check how large a backup is or validate exclusions.
It does not actually write any data to the underlying backend. This is
implemented as a simple overlay backend that accepts writes without
forwarding them, passes through reads, and generally does the minimal
necessary to pretend that progress is actually happening.

Output of the backup operation is mostly unchanged. Omitting or passing
verbosity options results in showing the same information, but for emitting
'Would add' rather than 'Added,' as shown below, in example usage:

```
$ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep "added)$"
modified  /plan.txt, saved in 0.000s (9.110 KiB added)
modified  /archive.tar.gz, saved in 0.140s (25.542 MiB added)
Would be added to the repo: 25.551 MiB
```

Therefore, to see a list of paths, `-vv` must be passed to `backup`, exactly 
as if a dry run were not being done.

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------

Yes, #1542

Checklist
---------

- [x] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [x] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
- [x] I have added tests for all changes in this PR
- [x] I have added documentation for the changes (in the manual)
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))
- [ ] I have run `gofmt` on the code in all commits -- **ran gofmt in final commit of branch**
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits)
- [x] I'm done, this Pull Request is ready for review

PR #3300, opened 2021-02-23
backup: add --dry-run/-n flag

### LINKED ISSUE #1542: Dry-Run Backup
Related to #10

It would be nice to be able to see what all will be changed via a snapshot before actually creating one. In essence, a way to show a diff between the files stored in restic, and a given path on the local filesystem. This would especially help when making changes to backup configurations so as to be able to catch anything that might be missing/errantly included before hand.

I assume given the architecture, this would be impossible to do without going through the steps of an actual backup. If so, I think it makes more sense to position it as a dry-run backup (where nothing is actually committed) rather than a diff a la `git status`.

aawsome:

This PR is a follow-up of #2308 and #3210 as both PRs seem to be stale.
I used the changes by @rmmh and @erincerys and did the following changes:
- simplified the dry run backend quite a lot (but is now only working for backup dry runs)
- changed the tests accordingly and separated the integration test for the dry run `backup` run

closes #1542

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -193,13 +193,13 @@ Dry Runs
 You can perform a backup in dry run mode to see what would happen without
 modifying the repo.
 
--  ``--dry-run``/``-n`` do not write anything, just print what would be done
+-  ``--dry-run``/``-n`` Report what would be done, without writing to the repository
 
 Combined with ``--verbose``, you can see a list of changes:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep "added)$"
+    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep "added"
     modified  /plan.txt, saved in 0.000s (9.110 KiB added)
     modified  /archive.tar.gz, saved in 0.140s (25.542 MiB added)
     Would be added to the repo: 25.551 MiB
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -187,6 +187,23 @@ On **Windows**, a file is considered unchanged when its path, size
 and modification time match, and only ``--force`` has any effect.
 The other options are recognized but ignored.
 
+Dry Runs
+********
+
+You can perform a backup in dry run mode to see what would happen without
+modifying the repo.
+
+-  ``--dry-run``/``-n`` do not write anything, just print what would be done
+
+Combined with ``--verbose``, you can see a list of changes:
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep added
+    modified  /plan.txt, saved in 0.000s (9.110 KiB added)
+    modified  /archive.tar.gz, saved in 0.140s (25.542 MiB added)
+    Would be added to the repo: 25.551 MiB
+
 Excluding Files
 ***************
 
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -199,7 +199,7 @@ Combined with ``--verbose``, you can see a list of changes:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep added
+    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep "added)$"
     modified  /plan.txt, saved in 0.000s (9.110 KiB added)
     modified  /archive.tar.gz, saved in 0.140s (25.542 MiB added)
     Would be added to the repo: 25.551 MiB
--- a/internal/backend/dryrun/dry_backend.go
+++ b/internal/backend/dryrun/dry_backend.go
@@ -1,0 +1,188 @@
+package dryrun
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
+	"sync"
+
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/restic"
+
+	"github.com/restic/restic/internal/debug"
+)
+
+type sizeMap map[restic.Handle]int
+
+var errNotFound = errors.New("not found")
+
+// Backend passes reads through to an underlying layer and only records
+// metadata about writes. This is used for `backup --dry-run`.
+// It is directly derivted from the mem backend.
+type Backend struct {
+	be   restic.Backend
+	data sizeMap
+	m    sync.Mutex
+}
+
+// New returns a new backend that saves all data in a map in memory.
+func New(be restic.Backend) *Backend {
+	b := &Backend{
+		be:   be,
+		data: make(sizeMap),
+	}
+
+	debug.Log("created new dry backend")
+
+	return b
+}
+
+// Test returns whether a file exists.
+func (be *Backend) Test(ctx context.Context, h restic.Handle) (bool, error) {
+	be.m.Lock()
+	defer be.m.Unlock()
+
+	debug.Log("Test %v", h)
+
+	if _, ok := be.data[h]; ok {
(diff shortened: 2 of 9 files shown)


JoshStrobl:
Just added this patch to the restic package for Solus, was rather surprised this functionality wasn't already in it. Works as intended.

doc/040_backup.rst
```diff
@@ -187,6 +187,23 @@ On **Windows**, a file is considered unchanged when its path and modification
...
 
+Dry Runs
+********
+
+You can perform a backup in dry run mode to see what would happen without
+modifying the repo.
+
+-  ``--dry-run``/``-n`` do not write anything, just print what would be done
+
+Combined with ``--verbose``, you can see a list of changes:
```
obbardc, line 198: why does `--dry-run` have to be combined with `--verbose`?

doc/040_backup.rst
```diff
@@ -187,6 +187,23 @@ On **Windows**, a file is considered unchanged when its path and modification
...
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo backup ~/work --dry-run -vv | grep "added)$"
```
obbardc, line 202: you say that `--dry-run` has to be combined with `--verbose`, yet you combine it with `-vv`

doc/040_backup.rst
aawsome: It does not have to be combined. `backup --dry-run` gives the same output as `backup` without dry-run (except it replaces the sentence "Added to the repo" by "Would be added to the repo"). But if you want to see the changes by details, you have to run it with increased verbosity.

If you think that this part of the docu is unclear: Do have a suggestion how to describe it?

doc/040_backup.rst
aawsome: `-vv` equals to `--verbose=2`

doc/040_backup.rst
abathur: This may just be a little linguistic confusion around "do not **write** anything"?

A few alt takes:
- `report what would be done without modifying the repository`
- `report what would be done, but do not modify the repository`
- `report what would be done without doing it`
- `report what would be done without doing anything`
- `report what would be done, but do nothing`
- `report what would be done, but make no changes`

Iodok:
Just tried the --dry-run with backup with my restic  version
My cmdline was: restic backup --dry-run -r $RESTIC_REPO $RESTIC_SRC -p $RESTIC_PASSWORDFILE 
and get:
unknown flag: --dry-run
Do you have any hints?

Iodok

JoshStrobl:
@Iodok The patch hasn't been merged in and is awaiting changes. The process for building Restic is fairly trivial, you can apply the patch separately or wait until it has been merged in and a future release occurs with the change in it.

bjoe2k4:
Works as intended. Please merge for inclusion in next release :-)

internal/repository/repository.go
```diff
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/restic/chunker"
+	"github.com/restic/restic/internal/backend/dryrun"
```
MichaelEischer, line 13: **question (non-blocking):** I wonder whether there's an elegant way to avoid the explicit dependency on the dry-run backend here. Using the `backendTestHook` would probably work, but that definitely doesn't count as elegant.

MichaelEischer (changes requested):
The code seems to be in quite good shape already. The main issue I see right now is that it's currently too easy to not shadow newly introduced modifying backend operations or no longer overwrite them after signature changes. See the comment for a suggestion.

doc/040_backup.rst
rawtaz: `report what would be done, without modifying the repository` :-)

internal/repository/repository.go
aawsome: IMO the `backendTestHook` would not work as intended. It is applied before the the backend is wrapped by a `CachedBackend`. So, dry-run writes to the repository would not write to the repo, but write to the cache which is not intended.

The problem with the cache is that it depends on the repo ID which is only available after actually reading the config file from the backend. This is why the cache wrapper is added in `repository.go`. Now as the dryrun wrapper should be applied to the already cache-enabled backend, I added this also here.
The alternative would be to allow direct access to the backend of the repository via some `GetBackend` and `SetBackend` methods and then call
```golang 
repo.SetBackend(dryrun.New(repo.GetBackend()))
```
in `cmd_backup.go`. This way we could also separate (most of) the cache logic from the backend.

internal/backend/dryrun/dry_backend.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/restic/restic/internal/restic"
+)
+
+// Backend passes reads through to an underlying layer and accepts writes, but
+// doesn't do anything. Also removes are ignored.
+// So in fact, this backend silently ignores all operations that would modify
+// the repo and does normal operations else.
+// This is used for `backup --dry-run`.
+type Backend struct {
+	restic.Backend
```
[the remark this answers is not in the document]
aawsome: is changed.

internal/backend/dryrun/dry_backend_test.go
```diff
@@ -0,0 +1,142 @@
...
+			err = step.be.Save(ctx, handle, restic.NewByteReader([]byte(step.content)))
+		case "test":
+			boolRes, err = step.be.Test(ctx, handle)
+			if boolRes != (step.content != "") {
+				t.Errorf("%d. Test(%q) = %v, want %v", i, step.fname, boolRes, step.content != "")
+			}
+		case "list":
+			fileList := []string{}
+			err = step.be.List(ctx, restic.PackFile, func(fi restic.FileInfo) error {
+				for _, n := range fileList {
```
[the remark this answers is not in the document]
aawsome: It isn't necessary. I removed this filtering out.

internal/backend/dryrun/dry_backend.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Save adds new Data to the backend.
+func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
+	if err := h.Valid(); err != nil {
+		return err
+	}
+
+	// discard everything from rd, but count it
+	maxInt := int(^uint(0) >> 1)
```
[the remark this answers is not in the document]
aawsome: Actually I just realized that in order to get the length, we can simply use `rd.Length()`. Thanks for the hint. 

I also checked out how this `Save` method is actually called and decided that the dry-run backend doesn't even need to read from the reader, so I removed this and now use `rd.Length()` to display the omitted size in the debug output.
Note that this in fact means that the dry run backend does now no longer exactly behave the same way with respect to `rd` like the original backend would. (but also other backends like the `CachedBackend` or the `RetryBackend` don't).
Also this means that `backup` with the `--dry-run` flag does no longer read previously written temporary files (or memory buffers) and should be a bit faster.

doc/040_backup.rst
aawsome: is changed.

internal/repository/repository.go
MichaelEischer: Indeed, you're right, the backend is wrapped with the `CacheBackend` later on.

Making arbitrary modifications to the used backend also doesn't feel particularly elegant, so let's just leave it like it is now.

MichaelEischer (approved):
LGTM. Thanks for completing the work on this option!

MichaelEischer:
I've added a short paragraph describing the use cases solved by this PR.

MichaelEischer:
I've rebased the PR before merging.

PR #3301, opened 2021-02-25
fix windows terminal output escape characters

### LINKED ISSUE #3111: Log file on Windows contains progress status text, escape characters and extra line breaks
Hello,

The log file on Windows contains progress status text, escape characters and extra line breaks. 
This does not happen on Linux. (https://forum.restic.net/t/different-log-file-in-windows-and-linux/3273)

Please fix, thanks!

Output of `restic version`
--------------------------

Windows
`restic 0.11.0 compiled with go1.15.3 on windows/amd64`
Linux
`restic 0.11.0 compiled with go1.15.3 on linux/amd64`

Windows
`restic backup --repo C:\Backup C:\Users --cleanup-cache --verbose --verbose >> C:\restic.log 2>&1`
Linux
`restic backup --repo /backup /source --cleanup-cache --verbose --verbose >> /restic.log 2>&1`

Expected behavior
-----------------

log files are the same on Windows and Linux

Linux

```
open repository
lock repository
load index files
using parent snapshot 5f9668c4
start scan on [/source]
start backup on [/source]
new       /source/file.txt, saved in 0.044s (100 B added)
unchanged /source/file.jpg
...
```

Actual behavior
---------------

Windows

```
open repository

e[2Klock repository

e[2Kload index files

e[2Kusing parent snapshot cf8362f4

e[2Kstart scan on [C:\Users]

e[2Kstart backup on [C:\Users]

e[2K[0:00] 0 files, 0 B, 0 errors

e[2Kunchanged /C/Users/MyUser/Config.xml

e[2K[0:00] 0 files, 0 B, 0 errors

e[2Kunchanged /C/Users/MyUser/test.html
...
```

(e[ = Escape character)

This did not happen in older versions <= 0.9.5

I'm happy that Restic also works on Windows

kmwoley:
This PR fixes Windows screen and piped output on `cmd`, `PowerShell`, and `mintty` terminals.

In Windows, piping the output of `restic` resulted in errant control characters being sent to the terminal / destination file. 

Invoking `restic backup ... > test.log` before this fix: 
```
[2Kusing parent snapshot eee10060

[2Kcreating VSS snapshot for [c:\]

[2Ksuccessfully created snapshot for [c:\]
...
```

After this fix:
```
using parent snapshot eee10060
creating VSS snapshot for [c:\]
successfully created snapshot for [c:\]
...
```

Closes #3111 

Checklist
---------


kmwoley:
This is my first PR into the project. I was looking into the history of #3111 and noticed that these checks were removed between 0.11 and 0.12, when the issue first appeared for me. I reverted them and tested them and it looks to be the root cause, so I thought I'd issue the PR vs. just telling you all about it.

internal/ui/termstatus/terminal_windows.go
```diff
@@ -19,6 +19,12 @@ func clearCurrentLine(wr io.Writer, fd uintptr) func(io.Writer, uintptr) {
 		return windowsClearCurrentLine
 	}
 
+	// check if the output file type is a pipe (0x0003)
+	if isPipe(fd) {
```
MichaelEischer, line 23: Adding this check turns `return posixClearCurrentLine` into dead code. `clearCurrentLine` and `moveCursorUp` are only called when `canUpdateStatus` has returned true before. And the only happens for windows terminals or if `isPipe(fd)` returns true.

Btw, `isPipe(fd)` misses a negation which got lost in #2608. Adding that negation, would in turn make `return func(io.Writer, uintptr) {}` dead code.

The clean version of this PR would be to replace everything in `canUpdateStatus` starting from `if !isPipe(fd)` with `return false`. In that case we should also drop all other compatibility code for mintty. However, I haven't had time to test restic+mintty, so I don't know whether that feature is broken anyways or not.

jdiercks:
Newbie question - how can I get an .exe built with this patch to test?
Is building it myself from source the only option, until it gets merged/released?
Thanks.

MichaelEischer:
@jdiercks There are no prebuilt binaries for PRs. So you'll have to checkout this branch and the build it using `go build ./cmd/restic`. Or if you have a linux (or similar) machine then run `GOOS=windows GOARCH=amd64 go build ./cmd/restic`.

This PR should only be thought of as a temporary workaround. It won't be merged in its current form.

jdiercks:
OK thanks @MichaelEischer. I'd be willing to test and report results, but don't have the time to set up a build environment at the moment. Not a critical condition for me, just a nuisance that the output is messy. I'l wait for a proper full fix release.

internal/ui/termstatus/terminal_windows.go
kmwoley: Thanks for pointing that out. I had blindly reverted the change without looking into it deeply before. I've done a deeper investigation and cleaned up the code considerably.

kmwoley:
@MichaelEischer - I've tested this on `cmd`, `PowerShell`, and `mintty`.

In all terminals, `terminal.IsTerminal(int(fd))` returned true if I was running directly on the terminal. And false if the output was piped to a file or command. This simplified the code dramatically.

I can confirm that status updates only are shown for non-piped output, and that there are no errant characters.

I'm unfamiliar with the `go.sum` linting errors that are occurring - I'm new to `go`, so any help resolving those would be appreciated.

MichaelEischer:
>  I've tested this on `cmd`, `PowerShell`, and `mintty`.

@kmwoley Which version of mintty did you use exactly? For me mintty included in MSYS2 still uses pipes on Windows 10.

kmwoley:
> @kmwoley Which version of mintty did you use exactly? For me mintty included in MSYS2 still uses pipes on Windows 10.

I tested with `Git Bash` terminal that installed with Git for Windows - my understanding is that it is a `mintty` terminal. I may be mistaken there.

kmwoley:
> @kmwoley Could you test whether #3325 works for you?
Tested. Took the conversation over to your PR.

kmwoley:
@MichaelEischer - what do you think? Should I close this PR in favor of the one you put together?

MichaelEischer:
I think we should use PR #3325, so let's close this one.

Issue #3202, opened 2020-12-30
Discussion: How to support cold storages?


aawsome:
This issue is meant to discuss which is the best way to implement the support of cold storages in restic.
The need to use restic with cold storages has been addressed in several places, see #1903, #2504, #2611,  #2796, #2817 and several discussions in the forum.

There are also some experimental PRs which kind of allow users to use cold storages, see #2516, #2881, #3196 and the already merged prune improvements (#2718 and following PRs) are already made to support cold storages in future.

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Allow users to use repository on some "cold storages".
These are usually cheap (cloud) storages where writing is fast and cheap, storing data is extremely cheap, but accessing the data is usually extremely slow (or may even need some extra "warming up" before being able to access it) and/or expensive.

Examples of cold storages are:

- AWS S3 Glacier and Glacier Deep Archive
- Google Cloud Platform Coldline Storage and Archive Storage
- Azure Blob Storage Cold and Archive
- OVH Cloud Archive
- Maybe self-programmed backends which write to a tape

What are you trying to do? What problem would this solve?
---------------------------------------------------------

Allow restic to use cheap storages for use-cases where access to file contents is usually never needed, or users are willing to accept the trade-offs that come with those storages. A use-case is disaster-only backups.

What issues need restic to tackle?
---------------------------------------------------------

1. Define how to split the repository in hot/cold parts
2. Reduce read access to the cold storage as much as possible
3. Reduce all API calls to the cold storage as much as possible
4. Add functionalities to get information which data from cold storage will be needed (+ maybe implement some "warmup" possibilities)
5. Add functionalities to let users restrict the access to cold storage, where possible
6. Allow restic to wait for slow cold storage access

Maybe some more features are needed...

Define how to split the repo in hot/cold parts
---------------------------------------------------------

First, I think we should talk about the treatment of pack files containing data blobs on one side and all other files in the repo on the other side. The reason is, that usually more than 99% of the repo size is occupied by those files. Only for degenerated case (many very small files), the tree blobs and the index may contribute significantly to the repo size.

Moreover there are a couple of commands that do not need to read/access any "data pack file" and should therefor fully work if only these are located in the cold storage:
- `backup`
- `cache`
- `copy` (destination repo)
- `diff`
- `find`
- `forget` (but not the `--prune` part)
- `init`
- `key`
- `list` (except packs)
- `ls`
- `recover`
- `snapshots`
- `stats`
[...]

PR #3303, opened 2021-02-25
restore: Add --warmup option

### LINKED ISSUE #2796: Restore --dry-run option to enable cold storage based repositories

aawsome:

Adds a dry run option to the `restore` command.
When called with `--dry-run`, `restore` only prints out which data pack files it needs to restore the data.

This allows to do some kind of warming-up for the data which will be needed for the actual `restore` run.

The two last commits extend the warm-up possibilities from #2881 to `restore`. I can also extract those to a separate PR, if wished (the discussion whether we should add the `--warm-up` option is not finished yet)

closes #2796 
see also #3202

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -45,6 +45,7 @@ type RestoreOptions struct {
 	Tags               restic.TagLists
 	Verify             bool
 	DryRun             bool
+	WarmUp             bool
 }
 
 var restoreOptions RestoreOptions
@@ -64,6 +65,7 @@ func init() {
 	flags.StringArrayVar(&restoreOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path` for snapshot ID \"latest\"")
 	flags.BoolVar(&restoreOptions.Verify, "verify", false, "verify restored files content")
 	flags.BoolVarP(&restoreOptions.DryRun, "dry-run", "n", false, "only show needed pack files")
+	flags.BoolVar(&restoreOptions.WarmUp, "warm-up", false, "only warm-up pack files needed to restore")
 }
 
 func runRestore(opts RestoreOptions, gopts GlobalOptions, args []string) error {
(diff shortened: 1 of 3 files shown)


aawsome:
see also https://github.com/rustic-rs/rustic/pull/94

PR #3310, opened 2021-02-27
`copy` raw tree blobs

### LINKED ISSUE #3267: SaveTree returned unexpected id
Output of `restic version`
--------------------------
On machine X: `restic 0.10.0 compiled with go1.15.6 on linux/amd64` (from the Debian bullseye repository, though most of the other software on machine X is from Debian buster).
On machine Y: `restic 0.9.4 compiled with go1.11.6 on linux/amd64` (from Debian buster).

On machine X:
```bash
sudo restic \
        --repo $folder/repository \
        --password-file $folder/passphrase \
        init \
        --repo2 /var/restic-combined/repository \
        --password-file2 /root/restic-combined-passphrase \
        --copy-chunker-params
sudo restic \
        --repo /var/restic-combined/repository \
        --password-file /root/restic-combined-passphrase \
        copy \
        --repo2 $folder/repository \
        --password-file2 $folder/passphrase \
        --tag $1\
        latest
```
Where `--tag $1` identifies existing backups containing data from machine Y.  This succeeded.

On machine Y:
```bash
restic -r sftp:$HOSTNAME@restic:repository \
        --password-file /root/restic-passphrase \
        --exclude-caches \
        --exclude="/home/tim/.no-backup" \
        --exclude="/root/.no-backup" \
        --one-file-system \
        --quiet \
        backup --files-from /usr/local/etc/backup-list
```
This successfully made a new backup to the new repository that was created on machine X by the previous commands.

On machine X:
```bash
TMPDIR=$base_directory/.tmp
export TMPDIR
restic \
    --repo $home_path/repository \                         
    --password-file $home_path/passphrase \                
    copy \
    --repo2 $combined_repository \                         
    --password-file2 $combined_passphrase \                
    --cache-dir $cache_directory
```
This is meant to copy the newly-created backup of machine Y to the original repository that existed on machine X before any of the above commands were run.  However, I get the following output:
```
snapshot e048b0d8 of [/etc /home /root /var/lib/apt/extended_states /var/lib/dpkg] at 2021-01-30 20:33:43.941864928 +1300 NZDT)
  copy started, this may take a while...
SaveTree(14998703) returned unexpected id 8227620d
```

`restic check` indicates no errors in either repository.

Local storage on machine X, accessed via SFTP from machine Y.

Expected behavior
-----------------
I expect restic to successfully copy the new backup of machine Y to the original repository on machine X.

Actual behavior
---------------
See the note above about the output obtained from the attempted copy.

See above.  This is reproducible with backups of machine Y, but I've had no problems when running the same procedure on backups of other machines.

Just a guess, but are you using truncated hashes in any part of the code that are short enough that a collision is plausible?  Machine Y's backups are by far the biggest of any of my machines — about 80 GiB, I think.

Or maybe there's some incompatibility caused by using restic from Debian bullseye on an otherwise buster machine?

If my wild guess above about hash collisions is correct, then don't truncate the hashes to be so short.

If it's caused by the buster/bullseye thing, then consider this a request to ensure that an up-to-date version of restic is always available in the Debian stable repository, or at least stable-backports.

Restic is a great tool for automating deduplicated backups of lots of my machines.  I especially like that I can use `rsync` to make an offsite copy of a restic repository, and then use `sha256sum` alone to do an integrity check of the offsite copy, using the checksums in the filenames.

MichaelEischer:
The `copy` command failed to copy the tree blob for a directory in #3267. The problematic tree blob ended up with a different hash after encoding it again, which in turn caused the `copy` command to fail an internal sanity check.

The problematic tree in #3267 contained a symlink which can be created using `ln -s '../i/'$'\355\246\361''d/samba' invalid-symlink`.

This PR modifies the `copy` command to load and store the raw bytes of a tree blob to avoid issues with unstable JSON encoding and to ensure future compatibility in case future restic versions add new fields to the directory metadata. The implementation is not completely optimal as it receives a decoded tree from `StreamTrees` and then loads the raw tree bytes a second time. However, this is the simplest and least messy way to implement it.

Fixes #3267.

Checklist
---------


MichaelEischer:
I'm not completely happy with my changelog entry. It feels a bit like I've simplified it too far, but I'm not sure how to discuss further details on that issue without a deep dive into restic internals. It would be possible to mention that this improves compatibility with future restic versions in case these store additional metadata.

cmd/restic/integration_test.go
```diff
@@ -799,6 +799,25 @@ func TestCopyIncremental(t *testing.T) {
 		len(copiedSnapshotIDs), len(snapshotIDs))
 }
 
+func TestCopyUnstableJSON(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+	env2, cleanup2 := withTestEnvironment(t)
+	defer cleanup2()
+
+	// contains a symlink created using `ln -s '../i/'$'\355\246\361''d/samba' invalid-symlink`
```
greatroar, line 808: Nitpick, but this is not an *invalid* symlink, as I said over at #3267.

changelog/unreleased/issue-3267
```diff
@@ -0,0 +1,7 @@
+Bugfix: `copy` failed to copy snapshots in rare cases
+
+The `copy` could in rare cases fail with the error message `SaveTree(...)
+returned unexpected id ...`.  This has been fixed.
```
greatroar, line 4: I would mention symlinks with non-UTF-8 targets here, and also explain that they're still somewhat problematic due to limitations of restic's repo format and that users may want to avoid them if possible.

cmd/restic/integration_test.go
MichaelEischer: I've changed the comment to say "broken-symlink". The `copy-unstable-json.tar.gz` still contains the filename `invalid-symlink`.

changelog/unreleased/issue-3267
MichaelEischer: I wonder whether that is the best place to add that information. After all the failing copy command is just the symptom and not the cause.

rawtaz: Perhaps a warning in the documentation for the copy command?

I'm thinking about the type of notice that's shown in https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#local or in https://restic.readthedocs.io/en/stable/045_working_with_repos.html#copying-snapshots-between-repositories depending on the severity.

MichaelEischer: The warning would rather belong into the documentation of the `backup` command. That is the point at which invalid UTF8 characters in symlinks are replaced. Once this PR is merged, the copy command won't care at all about the symlinks.

greatroar: My rationale for adding it here is that this is what users are going to find when they search for the error message. I'm not saying it shouldn't be addressed elsewhere too :)

MichaelEischer: I've amended the changelog

fd0:
> The implementation is not completely optimal as it receives a decoded tree from StreamTrees and then loads the raw tree bytes a second time. However, this is the simplest and least messy way to implement it.

Since the tree blobs are cached anyway it shouldn't be a huge issue. Thanks!

fd0 (approved):

PR #3335, opened 2021-03-13
prune --max-unused does not limit metadata repacking

cyounkins:

Adding minor clarification to documentation of `prune --max-unused` to indicate it does not limit metadata repacking. The referenced PR indicates `max-repack-size` CAN limit metadata repacking.

https://forum.restic.net/t/max-unused-unlimited-still-repacks/3661
https://github.com/restic/restic/pull/3228

Checklist
---------

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -341,7 +341,7 @@ The ``prune`` command accepts the following options:
       unused files. This means that as long as some data is still used within
       a file stored in the repo, restic will just leave it there. Use this if
       you want to minimize the time and bandwidth used by the ``prune``
-      operation.
+      operation. Note that metadata will still be repacked.
 
    Restic tries to repack as little data as possible while still ensuring this 
    limit for unused data.


doc/060_forget.rst
```diff
@@ -344,7 +344,7 @@ The ``prune`` command accepts the following options:
       operation.
 
    Restic tries to repack as little data as possible while still ensuring this 
-   limit for unused data.
+   limit for unused data. Note that this option does not limit metadata repacking.
```
MichaelEischer, line 347: In my opinion this remark should be part of the explanation of `unlimited`. And I'd reword it a bit to something like `Note that metadata is always repacked`.

When using `--max-unused` with a specific limit then it probably doesn't matter much that the target for unused metadata is always zero. That only becomes obvious when using the unlimited option.

cyounkins: I was confused but understand a bit better now.

Is metadata repacking limited by the `--max-repack-size` option? In my referenced forum post I used `--max-unused unlimited` and it repacked ~430 MB. If I had instead set `--max-repack-size 200M`, how much would it have repacked?

MichaelEischer: `--max-repack-size` limits metadata repacking. So it should have been about 200MB. prune just repacks all metadata first before repacking any data files.

cyounkins: So since `--max-repack-size` limits metadata repacking, I think it's confusing to say "metadata is always repacked" even in the context of the `unlimited` option of `--max-unused`. So I'm going to put "Note that metadata will still be repacked" in the context of the `unlimited` option.

If you disagree, feel free to commandeer or tell me to just put, "Note that metadata is always repacked". Thanks!

MichaelEischer: I actually prefer your wording over my suggestion, so lets keep that.

MichaelEischer:
Could you squash the commits into one? Two commits are a bit overkill for changing a single line ;-) .

cyounkins:
Oops! This is what happens when I try to use the GitHub interface for a simple edit. Fixed now.

MichaelEischer (approved):
LGTM. Thanks!

Issue #2743, opened 2020-05-17
Add flag to put restic mount into background after mount is successful


brad2014:
Output of `restic version`
--------------------------

restic 0.9.6 compiled with go1.13.6 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Add a flag (or make the default) for restic mount whereby it mounts the backup repository, and upon success, exits with mounted folder ready to use.  When finished using the repo, instead of killing the restic mount process, one would simply unmount the directory.

Actually, though it is not backward compatible, I'd suggest that it is worth changing `restic mount` to always return when the mount is ready, and print the message "When finished, unmount the mountpoint /backup." instead of "When finished, quit with Ctrl-c or umount the mountpoint."

I suppose messages from restic mount that get generated after the mount is done would have to be directed to somewhere.  I'd expect, by default, they'd go to syslog, but I don't know what the portable solution is.

What are you trying to do?
--------------------------

To immediately use the mounted repo once it is successfully mounted:
```
# if restic mount /backup ; then rsync -a /backup/hosts/_myhost_/latest/home /home; fi
# umount /backup
```

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

I love restic, and it makes me happy every day.  Really.  (Thank you for putting it out there)

rawtaz:
Am I missing something completely here, or doesn't restic have to be running in order to serve the virtual filesystem? If restic quits, how do you expect to get anything out of that mount point?

> Note that restic mount /backup </dev/null 2>/var/log/restic.log & doesn't work, since (a) I don't know where the mount was successful, and (b) the next command starts before the mount is established.

Consider adding a bit of sleep and then test if there's a restic "filesystem" there by just looking for e.g. `snapshots/latest/`?

brad2014:
I meant that `restic mount` would fork before exiting, like a daemon (and continues running in the background).  The point of the feature request is that it forks and returns with status 0 on success so that the next command is ready to go.

The umount command would kill the daemon (just as right now, it kills the foreground restic process).

Your suggested workaround would be something like this:

```bash
#!/bin/bash
# restore_home restore home directories from backup.
# Assumes RESTIC_xxx environment variables are set (repo, cache, keys, etc).

mountpoint=/backup
restorehost=myhost.com # the name of the host from which the backup is retrieved
logfile=/var/log/restic.log
restoredir=/home
[...]

PR #3341, opened 2021-03-18
Fork process for background mounting

mtotg:

This allows restic to be forked and put into the background for mounts to continue running even when restic closes.

Issue #2743 discusses this.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -62,6 +62,7 @@ type MountOptions struct {
 	Tags                 restic.TagLists
 	Paths                []string
 	SnapshotTemplate     string
+	DaemonMode           bool
 }
 
 var mountOptions MountOptions
@@ -79,9 +80,14 @@ func init() {
 	mountFlags.StringArrayVar(&mountOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`")
 
 	mountFlags.StringVar(&mountOptions.SnapshotTemplate, "snapshot-template", time.RFC3339, "set `template` to use for snapshot dirs")
+	mountFlags.BoolVar(&mountOptions.DaemonMode, "daemon", false, "run this mount in background like a daemon")
 }
 
 func runMount(opts MountOptions, gopts GlobalOptions, args []string) error {
+	if opts.DaemonMode {
+		go runMount(opts, gopts, args)
+		return errors.New("Daemon mode running in background")
+	}
 	if opts.SnapshotTemplate == "" {
 		return errors.Fatal("snapshot template string cannot be empty")
 	}
@@ -164,12 +170,13 @@ func runMount(opts MountOptions, gopts GlobalOptions, args []string) error {
 	Printf("When finished, quit with Ctrl-c or umount the mountpoint.\n")
 
 	debug.Log("serving mount at %v", mountpoint)
+
 	err = fs.Serve(c, root)
 	if err != nil {
 		return err
 	}
-
 	<-c.Ready
+
 	return c.MountError
 }
 
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -84,12 +84,31 @@ func init() {
 	mountFlags.BoolVar(&mountOptions.DaemonMode, "daemon", false, "run this mount in background like a daemon")
 }
 
+func linearSearch(args []string, search string) int {
+	for i, n := range args {
+		if n == search {
+			return i
+		}
+	}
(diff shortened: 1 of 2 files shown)


mtotg:
This is a WIP, but it's an initial pass. Though, I realize this is a recursive loop (code below), need to strip the DaemonMode argument off the args array.

```
pid, err := syscall.ForkExec(filename, args, attr)
```

cmd/restic/cmd_mount.go
```diff
@@ -79,9 +81,38 @@ func init() {
...
+	if opts.DaemonMode {
+		filename := os.Args[0]
+		attr := new(syscall.ProcAttr)
+		argx := removeElement(args, "daemon")
+		pid, err := syscall.ForkExec(filename, argx, attr)
+		Println("Mount is now running as a daemon. PID: ", pid)
+		if err != nil {
+			panic(err)
+		}
+		return errors.New("Daemon mode running in background")
```
MichaelEischer, line 114: The exit status should be zero if the background mount succeeded.

cmd/restic/cmd_mount.go
```diff
@@ -79,9 +81,38 @@ func init() {
...
+func linearSearch(args []string, search string) int {
+	for i, n := range args {
+		if n == search {
+			return i
+		}
+	}
+	return -1
+}
+
+func removeElement(args []string, search string) []string {
```
MichaelEischer, line 96: The function interface is misleading as `args` is modified in place.

cmd/restic/cmd_mount.go
```diff
@@ -79,9 +81,38 @@ func init() {
...
+		return append(args[:index], args[index+1:]...)
+	}
+	return args
 }
 
 func runMount(opts MountOptions, gopts GlobalOptions, args []string) error {
+	if opts.DaemonMode {
+		filename := os.Args[0]
+		attr := new(syscall.ProcAttr)
+		argx := removeElement(args, "daemon")
```
MichaelEischer, line 108: The `args` parameter has already been parsed by cobra. Calls like `restic -r repo mount --daemon` probably don't work?

cmd/restic/cmd_mount.go
```diff
@@ -164,12 +195,13 @@ func runMount(opts MountOptions, gopts GlobalOptions, args []string) error {
 	Printf("When finished, quit with Ctrl-c or umount the mountpoint.\n")
 
 	debug.Log("serving mount at %v", mountpoint)
+
```
MichaelEischer, line 198: Please avoid unnecessary whitespace changes.

greatroar:
The reporter of #2743 wants to "*immediately* use the mounted repo once it is successfully mounted." For that to work, the parent process needs to wait for the child to report a successful mount, probably through a pipe. Do you intend to implement that?

cmd/restic/cmd_mount.go
greatroar: Instead of removing an argument, maybe the child process should have a hidden flag `--daemon-subprocess`. I'm not sure if cobra support hidden flags, though. It could take a key shared with the parent as an argument, to ensure it doesn't work from the command line. `time.Now().UnixNano()` in hex should do.

cmd/restic/cmd_mount.go
```diff
@@ -79,9 +81,38 @@ func init() {
...
+	if index != -1 {
+		return append(args[:index], args[index+1:]...)
+	}
+	return args
 }
 
 func runMount(opts MountOptions, gopts GlobalOptions, args []string) error {
+	if opts.DaemonMode {
+		filename := os.Args[0]
+		attr := new(syscall.ProcAttr)
```
greatroar, line 107: A daemon should start a new session and chdir to the root:

```go
attr := &syscall.ProcAttr{
	Dir: "/",
	Sys: &syscall.SysProcAttr{Setsid: true},
}
```

Maybe Noccty should be set as well?

mtotg:
> The reporter of #2743 wants to "_immediately_ use the mounted repo once it is successfully mounted." For that to work, the parent process needs to wait for the child to report a successful mount, probably through a pipe. Do you intend to implement that?

Given that it’s forked and then needs to do the mounting cycle, my intent wasn’t to implement a pipe back to the main process to signal that though. Additionally, I’m not sure a pipe would be the best way if implemented.

greatroar:
Without signaling back, isn't this PR just a complicated way to do `restic mount &`?

MichaelEischer:
Closing this PR as it just seems to implement `restic mount &`.

PR #3354, opened 2021-04-04
Backup docs: on Windows, the filesize must match for "unchanged"

### LINKED ISSUE #3353: File change detection on windows
Output of `restic version`
--------------------------
restic 0.12.0 compiled with go1.15.8 on windows/amd64 (did not matter here, though)

From the docs: "On Windows, a file is considered unchanged when its path and modification time match, and only --force has any effect. The other options are recognized but ignored."
This is critical. What if someone changed the file and restored the modification time thereafter? That is a common practice e.g. for photos (when altering exif comments), where it's sometimes good to not rely only on exif timestamps, but keep the photos in cronological order in the filesystem already. So file size diffs must be taken into consideration, too. To be 100% sure to grab a changed file, one have to use checksums, but that would be super slow.... or maybe some filesystem internal flags of which I'm not aware right now....

I want to be sure that a file which content have changed is included in a snaphot, even if it looks unchanged at first glance.

Still testing it, and already using it in some scenarios. Very usefull!

greatroar:

The backup docs say that

> On Windows, a file is considered unchanged when its path and modification time match

I wrote that. I failed to mention that the size also has to match:

https://github.com/restic/restic/blob/efb10b3c4071ae218cede883c18e9f969491cec3/internal/archiver/archiver.go#L507-L508

Closes #3353.

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -183,9 +183,9 @@ Note that the device id of the containing mount point is never taken into
 account. Device numbers are not stable for removable devices and ZFS snapshots.
 If you want to force a re-scan in such a case, you can change the mountpoint.
 
-On **Windows**, a file is considered unchanged when its path and modification
-time match, and only ``--force`` has any effect. The other options are
-recognized but ignored.
+On **Windows**, a file is considered unchanged when its path, size
+and modification time match, and only ``--force`` has any effect.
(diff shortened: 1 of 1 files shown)


rawtaz (approved):
Thank you!

PR #3360, opened 2021-04-08
Fix runtime.KeepAlive usage in preallocateFile

greatroar:

When fst is a struct, what runtime.KeepAlive receives is a copy pointed to by an interface{}. If I've got the unsafe.Pointer rules right, the compiler is allowed to decide that neither FcntlInt call can see the struct because all they get is an int. It's then allowed to reorder the code to first copy the struct, then drop the memory for the original, then call FcntlInt.

In any case, this matches the [example](https://golang.org/pkg/runtime/#KeepAlive) in the runtime package more closely.

No.

Checklist
---------

### DIFF
--- a/internal/restorer/preallocate_darwin.go
+++ b/internal/restorer/preallocate_darwin.go
@@ -10,13 +10,13 @@ import (
 
 func preallocateFile(wr *os.File, size int64) error {
 	// try contiguous first
-	fst := unix.Fstore_t{
+	fst := &unix.Fstore_t{
 		Flags:   unix.F_ALLOCATECONTIG | unix.F_ALLOCATEALL,
 		Posmode: unix.F_PEOFPOSMODE,
 		Offset:  0,
 		Length:  size,
 	}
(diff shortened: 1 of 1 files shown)


fd0 (approved):
Oh, wow, that's subtle! Thanks!

greatroar:
I wonder if I'm being too paranoid here. The docs for runtime.KeepAlive don't define what an "object" is, runtime.SetFinalizer state that

> The argument obj must be a pointer to an object allocated by calling new, by taking the address of a composite literal, or by taking the address of a local variable.

So it seems like a local variable is kept alive too, despite being on the stack.

Experiment a bit, I came up with

```go
package main

import (
	"fmt"
	"runtime"
	"unsafe"
)

func main() {
	var x int
	p := unsafe.Pointer(p)

	fmt.Printf("%016x\n", uintptr(p))
	rec(100000) // Cause reallocation of the stack.
	fmt.Printf("%016x\n", uintptr(p))
	runtime.KeepAlive(p)
}

func rec(n int) {
	if n == 0 {
		return
	}
	rec(n - 1)
}
```

This prints two distinct values, even despite the runtime.KeepAlive. So taking the pointer earlier doesn't change the fact that x gets moved around...

PR #3367, opened 2021-04-21
Dynamically increase pack size with repository size

aawsome:

Currently packs are written once they reach the size of 4MB. In practice this leads to average pack sizes of around 4.7 MiB.
This means that large repositories contain of a large amount of pack files, e.g.
1 GiB -> ~200 pack files
1 TiB -> ~205,000 pack files (!)
1 PB -> ~215,000,000 pack files (!!!) 

Lots of small pack files pose problems in some cases, e.g. performance issues or high costs with some storage providers. Moreover with a lower number of pack files memory usage is a bit reduced.

In #2750 it is proposed to let users specify the pack size. A sensible choice would be to choose a pack size of `sqrt(reposize in GB) * 4 MiB`. This would lead to 
1 GiB -> ~256 pack files a 4 MiB
1 TiB -> ~8,200 pack files a 128 MiB
1 PiB -> ~260,000 pack files a 4 GiB
This would however need a priori knowledge of the repository size.

Hence this PR uses the current reposize to determine the pack size by the above formula and therefore dynamically increases the pack size with growing repo size.
This dynamic increase roughly doubles the number of the pack files compared to the table above.

For very small repos a pack size of 4MiB is still targeted and for very large repos (> 1 PiB) the pack size is limited to 4076 MiB such that including pack headers the total pack will not exceed 4 GiB.
Moreover, the number of blobs per pack is limited such that no problems with large pack headers or index files occur.

see #2750 and the references therein.

Checklist
---------

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -48,6 +48,7 @@ type Index struct {
 	treePacks restic.IDs
 	// only used by Store, StorePacks does not check for already saved packIDs
 	packIDToIndex map[restic.ID]int
+	totalSize     uint // the total size of data referenced in this index
 
 	final      bool       // set to true for all indexes read from the backend ("finalized")
 	ids        restic.IDs // set to the IDs of the contained finalized indexes
@@ -80,6 +81,7 @@ func (idx *Index) store(packIndex int, blob restic.Blob) {
 
 	m := &idx.byType[blob.Type]
 	m.add(blob.ID, packIndex, uint32(blob.Offset), uint32(blob.Length))
+	idx.totalSize += blob.Length
 }
 
 // Final returns true iff the index is already written to the repository, it is
@@ -350,6 +352,14 @@ func (idx *Index) Count(t restic.BlobType) (n uint) {
 	return idx.byType[t].len()
 }
 
+// TotalSize returns the total size of data referenced by the index
+func (idx *Index) TotalSize() uint {
+	idx.m.Lock()
+	defer idx.m.Unlock()
+
+	return idx.totalSize
+}
+
 type packJSON struct {
 	ID    restic.ID  `json:"id"`
 	Blobs []blobJSON `json:"blobs"`
@@ -561,6 +571,7 @@ func (idx *Index) merge(idx2 *Index) error {
 	idx.treePacks = append(idx.treePacks, idx2.treePacks...)
 	idx.ids = append(idx.ids, idx2.ids...)
 	idx.supersedes = append(idx.supersedes, idx2.supersedes...)
+	idx.totalSize += idx2.totalSize
 
 	return nil
 }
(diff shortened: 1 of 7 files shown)


aawsome:
The argument why the dynamic increase of pack file size roughly doubles the number of files is the following:

Having a repo of `x_total` GiB and a constant pack size of `sqrt(x_total) * 4MiB`, you have around
```x_total * 1024 / (sqrt(x_total) * 4) = 256 * sqrt(x_total)```
pack files.

If you use a dynamic pack file size of `sqrt(x) * 4MiB` where `x` is the current pack size in GB, your number of pack files is around
```
integral_0^{x_total}  1024 / (sqrt(x) *4) dx = 256 integral_0^{x_total}  1 / sqrt(x) = 256 * 2 * sqrt(x_total)
```

vejnar:
In my opinion, this is a great idea: number of files for large repositories is really a problem in my case and that solution wouldn't require any config from users. However, I would really like having `smallestPackSize` exposed as an option, similarly to what #2750 is proposing.

aawsome:
@vejnar Of course we could easily add an option to adapt this into the `MinPackSize()` function I introduced here. However one of the more difficult discussions #2750  was whether we should or shouldn't make such technical options available to users. I personally like having more options; but I also fully support the philosophy of restic being "opinionated".

In any case you can compile your own version with a changed constant. With this PR this would be the `smallestPackSize` constant in `internal/repository/repository.go`.

aawsome:
There is one drawback in the "dynamic" increase of pack sizes: You still get some small (first) packs even for large repositories.

I think this can be tackled in a follow-up PR which should change `prune` in the following way:
- determine the used repo size after pruning (this is already done in `prune`)
- compute the preferred pack size for this target state
- mark some packs that are way too small (or way too large) for repacking within `prune`

This would allow to swing in the pack sizes if a "constant" repo size is reached.

wzrdtales:
hit this while trying to understand why min-packsize was removed. Especially on slower devices it is beneficial to have upfront bigger pack files since too many small pack files will destroy performance completely...

pjannesen:
hit this while trying to understand why min-packsize was removed. Especially on slower devices it is beneficial to have upfront bigger pack files since too many small pack files will destroy performance completely...

_Originally posted by @wzrdtales in https://github.com/restic/restic/issues/3367#issuecomment-836589646_

Same for me. Restic works great if you have a lot of small files (like software development) and the target is a local SDD. If you use Restic under windows and the target is an external HD a packsize of 4MB is a disaster!

To make Restic work in all different environments there must be a way to specify the min-packsize. 

PS: The standard of 4 MB is low. The default of duplicati is 50MB if I am correct.

aawsome:
Closing this as I won't invest more time here.
If anyone wants to try out dynamically increasing pack sizes, I'd recommend to switch to [rustic](https://github.com/rustic-rs/rustic)

Issue #3392, opened 2021-05-14
prune flag --keep-after for keeping snapshots relative to now


phillipp:
We have a backup system implemented with restic where we need to prune snapshots based on a date relative to the current time. We keep snapshots from the same data in multiple repositories due to the repos getting too large (tens of thousands of snapshots per repo).

We tried pruning these by parsing the snapshot list and pruning with a list of snapshot IDs, but prune lists the snapshots from source for every ID given because of how the snapshot list building works (giving the prune command 1000 snapshot IDs pulls the snapshot list from the REST backend 1000 times).

We then added a `--keep-after` flag that works like `--keep-within`, but with the current time rather than the time of the last snapshot.

Would you be interested in a pull request for that?

aawsome:
> prune lists the snapshots from source for every ID given because of how the snapshot list building works (giving the prune command 1000 snapshot IDs pulls the snapshot list from the REST backend 1000 times).

I also noticed this and would say we could claim this as performace bug.
For a list of (short) snapshot IDs it must be sufficient to list the snapshot files only once from the repository.
In fact, I already started reworking and unifying the snapshot processing (also implementing ideas like 'latest-1' as in #1965) but did run a bit out of time to finish it.

However, in any case we can simply remove the backend listing when a long ID is given. If you call restic from a script it should be no problem to use long snapshot IDs.
I started #3395, feel free to try it out.

MichaelEischer:
> In fact, I already started reworking and unifying the snapshot processing (also implementing ideas like 'latest-1' as in #1965) but did run a bit out of time to finish it.

I agree list should only be called once at the backend, but please fix the performance issue separately from adding new features.

phillipp:
@MichaelEischer @aawsome 
I agree, a fix for the performance problem would be good - but the question remains if I should open a pull request for `--keep-after` or not? ;-)

MichaelEischer:
@phillipp What exactly prevents you from using `--keep-within`? One important feature of the `--keep-*` flags is that these prevent accidental deletions of all snapshots within a snapshot group.

phillipp:
@MichaelEischer It won't delete snapshots that have no recent backups. In our case, we switch repositories for a data directory when the snapshot count in a repo reaches a certain limit (like a few ten thousands) because of performance problems. We have around 18 million snapshots right now.

MichaelEischer:
Not deleting snapshots without recent backups is in fact intentional as this makes it harder to accidentally remove your backups.

[...]

PR #3395, opened 2021-05-15
Don't search for long IDs

aawsome:

Don't search for IDs if a long (i.e. complete) ID is given. This should mainly affect snapshot IDs given on the command line which now no longer need to run a backend 'List' for each long snapshot ID given.

see #3392 

Checklist
---------


MichaelEischer:
This currently changes the semantic of the `Find` function. According to its help text it behaves as follows: `If none is found, nil and ErrNoIDPrefixFound is returned.`. This is no longer the case for full snapshot ids. The `forget` command for example would then fail instead of ignoring the non-existent snapshots.

MichaelEischer:
Replaced by #3570.

PR #3401, opened 2021-05-17
Goroutine shutdown cleanups

MichaelEischer:
It fixes the goroutine shutdown in two places, where the goroutines were not shutdown properly in case of an error.

No.

Checklist
---------

- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))~~ Not really user visible

### DIFF
--- a/cmd/restic/delete.go
+++ b/cmd/restic/delete.go
@@ -25,16 +25,21 @@ const numDeleteWorkers = 8
 func deleteFiles(gopts GlobalOptions, ignoreError bool, repo restic.Repository, fileList restic.IDSet, fileType restic.FileType) error {
 	totalCount := len(fileList)
 	fileChan := make(chan restic.ID)
-	go func() {
+	wg, ctx := errgroup.WithContext(gopts.ctx)
+	wg.Go(func() error {
+		defer close(fileChan)
 		for id := range fileList {
-			fileChan <- id
+			select {
+			case fileChan <- id:
+			case <-ctx.Done():
+				return nil
+			}
 		}
-		close(fileChan)
-	}()
+		return nil
+	})
 
 	bar := newProgressMax(!gopts.JSON && !gopts.Quiet, uint64(totalCount), "files deleted")
 	defer bar.Done()
-	wg, ctx := errgroup.WithContext(gopts.ctx)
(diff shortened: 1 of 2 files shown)


cmd/restic/delete.go
```diff
@@ -25,16 +25,21 @@ const numDeleteWorkers = 8
...
-	go func() {
+	wg, ctx := errgroup.WithContext(gopts.ctx)
+	wg.Go(func() error {
+		defer close(fileChan)
 		for id := range fileList {
-			fileChan <- id
+			select {
+			case fileChan <- id:
+			case <-ctx.Done():
+				return nil
```
greatroar, line 35: Shouldn't this return ctx.Err()?

MichaelEischer: Oh indeed. Thanks!

fd0: I'm not sure about that. If the context is cancelled somehow, that's not an error this particular function should return (again), or is it?

Maybe I misunderstood, but in practice when the context is cancelled here that happens because either SIGINT is received or one of the other goroutines (which delete the files) returned an error. Both are no reasons for this goroutine to return an error. In the latter case, the ErrGroup already has its first error (from the other goroutine which returned it) so it does not matter, but I think `return nil` would be the correct thing to do.

Where's my error? :)

MichaelEischer: In my opinion `deleteFiles` should return a "Context canceled" error if it did not complete due to a canceled context. That makes it simpler for the calling function to also abort immediately, as there's no need to check that the context is still valid.

In case a SIGINT is received, the fileList-loop could exit through the `return nil`. Then the closed channel could also let the fileChan-loops exit without returning an error. Then `wait()` would return `nil` which doesn't tell that only parts of the files were deleted. Since the RetryBackend fails immediately for a canceled context that probably won't cause a problem in the repository. However, that still feels wrong to me.

And in case a SIGINT is received, then `deleteFiles` will most of the times exit with a "Context canceled" error anyways as thats what `repo.Backend().Remove` will return.

fd0 (approved):
LGTM

PR #3407, opened 2021-05-21
Switch to the azure-storage-for-go sdk

ekarlso:
This seems to be the supported going forwards sdk which also adds
the possibility to use AzureAD based authentication (Working on
seperate PR).

Switches the SDK used for Azure to Azure Storage for GO which
will support other built-in authentication methods rather then just
the SharedKey (Sepeperate PR)

No

Checklist
---------


chgeuer:
Hi @ekarlso , really cool you started working on this. Was just looking at the very same topic, glad I found your PR. Having support for service-principal-based authN (and managed identity) would be a great addition to restic. 

Via which environment variables are you planning to convey the service principal credentials, or whether to use MSI instead, or (in case there are multiple MSIs attached to a node) the objectID of the MSI to use? I think Restic's approach of using environment variables into the runtime is great, but would have appreciated if these could be prefixed with something like `RESTIC_AZURE_` or so... If you need to introduce additional ones, would be great if these would not start with just `AZURE_` :-)

internal/backend/azure/azure.go
```diff
@@ -139,84 +156,19 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
...
-		// CreateBlockBlobFromReader reads length from `Len()``
-		dataReader := azureAdapter{rd}
+	debug.Log("InsertObject(%v, %v)", be.container, objName)
 
-		// if it's smaller than 256miB, then just create the file directly from the reader
-		err = be.container.GetBlobReference(objName).CreateBlockBlobFromReader(dataReader, nil)
-	} else {
-		// otherwise use the more complicated method
-		err = be.saveLarge(ctx, objName, rd)
+	// if it's smaller than 256miB, then just create the file directly from the reader
```
MichaelEischer, line 161: That comment no longer seems to apply?

internal/backend/azure/azure.go
```diff
@@ -139,84 +156,19 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
...
+	blob := be.container.NewBlockBlobURL(objName)
 
-	}
+	_, err := storage.UploadStreamToBlockBlob(context.TODO(), rd, blob, storage.UploadStreamToBlockBlobOptions{})
```
MichaelEischer, line 164: I'd suggest to use a larger buffer size than the default 1MB. The old code used 100MB buffers, so maybe something inbetween would be optimal?

internal/backend/azure/azure.go
```diff
@@ -139,84 +156,19 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
...
-		}
-
-		blocks = append(blocks, storage.Block{
-			ID:     id,
-			Status: "Uncommitted",
-		})
-	}
-
-	// sanity check
-	if uploadedBytes != int(rd.Length()) {
```
MichaelEischer, line 210: Is there a way to determine how many bytes were uploaded with the new API?

internal/backend/azure/azure.go
```diff
@@ -33,26 +33,51 @@ const defaultListMaxItems = 5000
...
 
-	service := client.GetBlobService()
+	if err != nil {
+		return storage.ServiceURL{}, err
+	}
+
+	return storage.NewServiceURL(*blobPrimaryURL, pipeline), nil
+}
+
+func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
```
MichaelEischer, line 65: The roundTripper is no longer used which breaks debugging and upload/download throttling. Please make sure that http requests use the roundTripper.

internal/backend/azure/azure.go
```diff
@@ -250,27 +202,17 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
-		end = uint64(offset + int64(length) - 1)
-	} else {
-		end = 0
-	}
+	blob := be.container.NewBlockBlobURL(objName)
 
 	be.sem.GetToken()
 
-	rd, err := blob.GetRange(&storage.GetBlobRangeOptions{Range: &storage.BlobRange{Start: start, End: end}})
+	resp, err := blob.Download(context.TODO(), offset, int64(length), storage.BlobAccessConditions{}, false, storage.ClientProvidedKeyOptions{})
```
MichaelEischer, line 209: Please use the context `ctx` passed to this method. This also applies to all other usages of `context.TODO()`

MichaelEischer:
The SDK also seems to require Go 1.14, so we'd have to bump the minimum go version supported by restic. But that shouldn't be a problem.

internal/backend/azure/azure.go
greatroar: Also, shouldn't this use ctx?

MichaelEischer:
@ekarlso Did you make any progress on this PR?

MichaelEischer:
Are you still interested in completing this PR?

MichaelEischer:
Closed as there seems currently to be no interest in the PR.

PR #3409, opened 2021-05-27
Make restic.{lchown,mknod} regular functions

greatroar:

restic.lchown and restic.mknod are changed into regular functions rather than function-typed variables. This allows the compiler to inline them and remove some dead code. The binary becomes about 2.5 KiB smaller on Windows, 3 KiB on Linux.

To pull this off, I had to write a specialized mknod wrapper for FreeBSD, the only platform where its dev argument is uint64. The old code cast the Node.Device field to int everywhere, meaning that the behavior on 32-bit FreeBSD was possibly incorrect (though I haven't been able to test this).

No.

Checklist
---------

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -188,8 +188,6 @@ func (node Node) restoreMetadata(path string) error {
 	if err := lchown(path, int(node.UID), int(node.GID)); err != nil {
 		// Like "cp -a" and "rsync -a" do, we only report lchown permission errors
 		// if we run as root.
-		// On Windows, Geteuid always returns -1, and we always report lchown
-		// permission errors.
 		if os.Geteuid() > 0 && os.IsPermission(err) {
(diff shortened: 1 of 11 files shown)


internal/restic/node.go
```diff
@@ -188,8 +188,6 @@ func (node Node) restoreMetadata(path string) error {
 	if err := lchown(path, int(node.UID), int(node.GID)); err != nil {
 		// Like "cp -a" and "rsync -a" do, we only report lchown permission errors
 		// if we run as root.
-		// On Windows, Geteuid always returns -1, and we always report lchown
```
MichaelEischer, line 191: What is the reason why this comment was removed?

greatroar: Because this is dead code on Windows; lchown is a stub that returns nil.

MichaelEischer (approved):
LGTM.

PR #3416, opened 2021-06-04
Keep hourly within

### LINKED ISSUE #3414: Add --keep-hourly-within switch to restic forget
Output of `restic version`
--------------------------

restic 0.12.0 compiled with go1.15.7 on darwin/amd64

New switches would be added to the `restic forget` command, including `--keep-hourly-within`, and similar for `daily`/`weekly`/`monthly`/`yearly` allowing an interval to be specified for how long such backups should be kept, instead of specifying the number of each type of backup.

As mentioned on a [forum thread](https://forum.restic.net/t/forget-policy/4014/11), if backups are irregular, hourly backups specified with `--keep-hourly=30` sometimes span a bit over a day, but sometimes they span a lot more than that. For systems with irregular backup schedules (and perhaps for all systems), the more natural approach would be to allow specifying hourly snapshots younger than a specific age (relative to the newest snapshot).

The relationship between the current `--keep-hourly` and a new `--keep-hourly-within` would be the same as the relationship between `--keep-last` and `--keep-within`. One keeps last n snapshots of a given type, the other keeps any snapshots within a given time period (relative to the latest snapshot) of a given type.

Putting this in a table, we would have the following (with italic denoting the new parameters):

| Type    | Keep last n snapshots of a given type | Keep snapshots within a given time period of a given type |
| ------- | ------------------------------------- | ------------------------------------------------------------ |
| any     | --keep-last                        | --keep-within                                                |
| hourly  | --keep-hourly                         | *--keep-hourly-within*                                       |
| daily   | --keep-daily                          | *--keep-daily-within*                                        |
| weekly  | --keep-weekly                         | *--keep-weekly-within*                                       |
| monthly | --keep-monthly                        | *--keep-monthly-within*                                      |
| yearly  | --keep-yearly                         | *--keep-yearly-within*                                       |

Yes, the cheap snapshots that let me got rid of annoying incremental backups are a dream. And the code base is super easy to work with :-)

torfason:

This adds new switches to the `restic forget` command, including `--keep-hourly-within`, and similar for `daily`/`weekly`/`monthly`/`yearly` allowing an interval to be specified for how long such backups should be kept, instead of specifying the number of each type of backup.

The PR also includes tests for the new switches.

See more information about this in the related issue, #3414.

Yes, this PR closes closes #3414.

It was mentioned briefly in the [forum](https://forum.restic.net/t/forget-policy/4014/11) in a discussion about forget policy. 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_forget.go
+++ b/cmd/restic/cmd_forget.go
@@ -31,14 +31,19 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 
 // ForgetOptions collects all options for the forget command.
 type ForgetOptions struct {
-	Last     int
-	Hourly   int
-	Daily    int
-	Weekly   int
-	Monthly  int
-	Yearly   int
-	Within   restic.Duration
-	KeepTags restic.TagLists
+	Last          int
+	Hourly        int
+	Daily         int
+	Weekly        int
+	Monthly       int
+	Yearly        int
+	Within        restic.Duration
+	WithinHourly  restic.Duration
+	WithinDaily   restic.Duration
+	WithinWeekly  restic.Duration
+	WithinMonthly restic.Duration
+	WithinYearly  restic.Duration
+	KeepTags      restic.TagLists
 
 	Hosts   []string
 	Tags    restic.TagLists
@@ -64,6 +69,11 @@ func init() {
 	f.IntVarP(&forgetOptions.Monthly, "keep-monthly", "m", 0, "keep the last `n` monthly snapshots")
 	f.IntVarP(&forgetOptions.Yearly, "keep-yearly", "y", 0, "keep the last `n` yearly snapshots")
 	f.VarP(&forgetOptions.Within, "keep-within", "", "keep snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.WithinHourly, "keep-within-hourly", "", "keep hourly snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.WithinDaily, "keep-within-daily", "", "keep daily snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.WithinWeekly, "keep-within-weekly", "", "keep weekly snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.WithinMonthly, "keep-within-monthly", "", "keep monthly snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.WithinYearly, "keep-within-yearly", "", "keep yearly snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
 
 	f.Var(&forgetOptions.KeepTags, "keep-tag", "keep snapshots with this `taglist` (can be specified multiple times)")
 	f.StringArrayVar(&forgetOptions.Hosts, "host", nil, "only consider snapshots with the given `host` (can be specified multiple times)")
@@ -128,14 +138,19 @@ func runForget(opts ForgetOptions, gopts GlobalOptions, args []string) error {
 		}
 
 		policy := restic.ExpirePolicy{
-			Last:    opts.Last,
-			Hourly:  opts.Hourly,
-			Daily:   opts.Daily,
-			Weekly:  opts.Weekly,
-			Monthly: opts.Monthly,
-			Yearly:  opts.Yearly,
-			Within:  opts.Within,
-			Tags:    opts.KeepTags,
+			Last:          opts.Last,
+			Hourly:        opts.Hourly,
+			Daily:         opts.Daily,
+			Weekly:        opts.Weekly,
+			Monthly:       opts.Monthly,
+			Yearly:        opts.Yearly,
+			Within:        opts.Within,
+			WithinHourly:  opts.WithinHourly,
+			WithinDaily:   opts.WithinDaily,
+			WithinWeekly:  opts.WithinWeekly,
+			WithinMonthly: opts.WithinMonthly,
+			WithinYearly:  opts.WithinYearly,
+			Tags:          opts.KeepTags,
 		}
 
 		if policy.Empty() && len(args) == 0 {
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -190,6 +190,18 @@ The ``forget`` command accepts the following parameters:
    years, months, days, and hours, e.g. ``2y5m7d3h`` will keep all snapshots
    made in the two years, five months, seven days, and three hours before the
    latest snapshot.
+-  ``--keep-within-hourly duration`` keep all hourly snapshots made within
+   specified duration of the latest snapshot. The duration is specified in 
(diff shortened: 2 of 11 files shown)


cmd/restic/cmd_forget.go
```diff
@@ -64,6 +69,11 @@ func init() {
 	f.IntVarP(&forgetOptions.Monthly, "keep-monthly", "m", 0, "keep the last `n` monthly snapshots")
 	f.IntVarP(&forgetOptions.Yearly, "keep-yearly", "y", 0, "keep the last `n` yearly snapshots")
 	f.VarP(&forgetOptions.Within, "keep-within", "", "keep snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
+	f.VarP(&forgetOptions.HourlyWithin, "keep-hourly-within", "", "keep hourly snapshots that are newer than `duration` (eg. 1y5m7d2h) relative to the latest snapshot")
```
MichaelEischer, line 72: After discussing with @rawtaz we'd prefer to name the options such that they are prefixed with `keep-within` to group the new options together in the help output. That is `--keep-hourly-within` would become `--keep-within-hourly`, etc.

doc/060_forget.rst
```diff
@@ -189,7 +189,19 @@ The ``forget`` command accepts the following parameters:
    the duration of the latest snapshot. ``duration`` needs to be a number of
    years, months, days, and hours, e.g. ``2y5m7d3h`` will keep all snapshots
    made in the two years, five months, seven days, and three hours before the
-   latest snapshot.
+   latest snapshot`
```
MichaelEischer, line 192: That sentence should end with a `.` and also the new item.

internal/restic/snapshot_policy.go
```diff
@@ -180,6 +217,20 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots, reason
...
+		reason string
+	}{
+		{p.HourlyWithin, ymdh, -1, "hourly within"},
+		{p.DailyWithin, ymd, -1, "daily within"},
+		{p.WeeklyWithin, yw, -1, "weekly within"},
+		{p.MonthlyWithin, ym, -1, "monthly within"},
+		{p.YearlyWithin, y, -1, "yearly within"},
+	}
+
 	latest := findLatestTimestamp(list)
```
MichaelEischer, line 234: I wonder whether we should bound that timestamp to the current time. Or probably even better ignore all snapshots from the future for this calculation (unless there's only a snapshot). Otherwise you could wipe most snapshots by just creating a snapshot with a timestamp far into the future.

MichaelEischer (changes requested):

internal/restic/snapshot_policy.go
torfason: Should perhaps timestamps in the future be an error for all policy-based forgets (both these new switches and the pre-existing one)? Seems like that indicates a repo in a pretty bent shape.

cmd/restic/cmd_forget.go
torfason: OK, that makes sense. I'll update that, and also the variable names in the code for consistency. It will be a couple of weeks since I'm traveling.

internal/restic/snapshot_policy.go
MichaelEischer: I like the idea in general, but I'm not sure which corner cases could appear. Timestamps that are only slightly in the future (e.g. just a few minutes) for example are probably ok and quite likely caused by a not properly synchronized clock.

cmd/restic/cmd_forget.go
torfason: Swapping the order should be done now.

doc/060_forget.rst
torfason: Should be fixed now.

internal/restic/snapshot_policy.go
torfason: Yes, an error might be a little strict. So how about issuing a warning (wouldn't `Warnf()` be the right way to go?) if any future-dated snapshots are found, before removing them?

My sense is that the best way to do this check would  at the start of ApplyPolicy, before any other action. At least before checking if `len(list)==0` and returning. This would mean that it would not only apply to the new parameters, but to all policy-based forgets. But I think that would nevertheless be much cleaner, and furthermore, incorrectly dated snapshots could easily mess with all the other time-based policies as well. Obviously they affect `--keep-within`, but `--keep-latest` would also be heavily affected, and having several incorrectly dated snapshots would also mean that something like `--keep-daily 3` would end up forgetting all new (correctly dated) backups.

The place I'm suggesting implementing this is here:

```go
func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots, reasons []KeepReason) {
	sort.Sort(list)

// CHECK FOR ANY FUTURE SNAPSHOTS HERE, REMOVE THEM FROM list IF FOUND,
// AND ISSUE A WARNING.

	if p.Empty() {
		for _, sn := range list {
			reasons = append(reasons, KeepReason{
				Snapshot: sn,
				Matches:  []string{"policy is empty"},
			})
		}
		return list, remove, reasons
	}

	if len(list) == 0 {
		return list, nil, nil
	}
```

MichaelEischer: As the policies are applied per host, I'd expect all snapshots of that host to be affected. And thus a wrong date would cancel itself out. That should prevent a skewed clock from breaking the forget policies. If someone accidentally creates a snapshot with a wrong date, it's effect would be limited to just one of the snapshots kept by `--keep-latest` and `--keep-daily`. 

Whereas for `--keep-within` it could cause removing all snapshots except the one with the wrong date. Therefore, I'd prefer only to slightly change the behavior of `--keep-within*` but not for the other options. That would basically just mean to slightly tweak `findLatestTimestamp`.

Regarding a warning I don't have a feeling for how much noise that could generate. I think we should start without a warning for now and maybe add it if this leads to too many complaints.

MichaelEischer:
The changes look good so far. Could you squash the commits to cleanup the git history?

internal/restic/snapshot_policy.go
torfason: OK, I've done the following:

  * Squashed the commits implementing the new switches.
  * Changed `findLatestTimestamp` so that it returns the latest timestamp that is in the past, ignoring any future timestamps.
  * I've made sure that the `snapshot_policy_test.go` tests run with this latest change (but one should note that the test repo does not include any future snapshots).

What I did NOT do:

  * I did not add any error or warning on forgetting from a repo which has future snapshots (regardless).
  * I did not add any future snapshots to the unit test repo. I thought that was too much of a change for what is hopefully an absolute edge case.
  * I did not squash the `findLatestTimestamp` changes into the other changes, since it is logically separate from implementing the new switches, and I think it is better to keep it visible.

Thoughts on this warning thing:

  * After thinking about it more, I now feel that the best place to put in a test+warning for future snapshots would be in `restic check`. A warning could have negative implications for automatic runs of commands such as `forget`, but when running `check`, one is presumably doing it in order to examine the output. If we find that repos with future snapshots are common in the wild, we will start hearing about it in the forum and that would justify a deeper look into whether it is likely that future snapshots could cause bad things to happen.
  * Should I file an issue to do this?

rawtaz: FWIW, I think that reasoning regarding the warning for future snapshots makes a lot of sense, `check` is a much better place indeed. Good thinking!

MichaelEischer (approved):
LGTM. Thanks!

Opening an issue to add a warning to check if snapshots have a date in the future sounds like a good idea.

PR #3419, opened 2021-06-07
Enable admin to decide group access to repository files

### LINKED ISSUE #2351: Allow default umask to be overridden
Output of `restic version`
--------------------------

Restic should allow its hard coded umask (0077) to be overridden by a user specified umask.

What are you trying to do?
--------------------------

Access repo files (raw data, not decrypted) by a non-root user other than the user that runs Restic, and without having to chmod/chown the repo files after each backup.

Restic creates all repo related files with permissions that allow only the Restic user to read/write said files (no group permissions), effectively using a 0077 umask, and ignoring whatever umask the Restic user has set in their environment.

This prevents things like extended ACLs, using setfacl, from being applicable. If we want another process, running as a different, non-root user, to copy or process these files in some way, it can not do so without the files being chown/chmod'd after each backup. Being able to set custom extended ACLs would solve this, but without a customisable umask, it can't work.

I understand why Restic defaults to a tightened umask, but it should be able to be overridden by the user if required.

Restic makes me happy every day.

DanielG:
This fixes #2351, see also https://github.com/restic/restic/issues/2351#issuecomment-855791064

Since the directory mode doesn't allow the execute bit (search permission)
to be set regardless of umask the files don't need to have empty group
permissions to disallow group access. By default the missing group execute
access on the containing directory will block any read/write attempt on the
files.

The upside of this is that the sysadmin can choose to change the directory
permissions to allow group execute access since all repo directories are
created at init time and not touched again after that. This will then allow
the group access to the files.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3419
+++ b/changelog/unreleased/pull-3419
@@ -1,0 +1,21 @@
+Enhancement: Use config file permissions to control file group access
+
+Previously files in a local/sftp restic repository would always end up with
+very restrictive access permissions allowing access only to the owner. This
+prevented a number of valid use-cases involving groups and ACLs.
+
+Now we use the config file permissions to decide whether group access
+should be given to newly created repository files or not. We arrange for
+repository files to be created group readable exactly when the repository
+config file is group readable.
+
+To opt-in to group readable repositories a simple `chmod -R g+r` or
+equivalent can be used. For repositories that should be writable by group
+members a tad more setup is required, see the docs.
+
+Posix ACLs can also be used now that the group permissions being forced to
+zero no longer masks the effect of ACL entries.
+
+https://github.com/restic/restic/issues/2351
+https://github.com/restic/restic/pull/3419
+https://forum.restic.net/t/change-permissions-on-repository-files/1391
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -699,3 +699,56 @@ On MSYS2, you can install ``winpty`` as follows:
 
     $ pacman -S winpty
     $ winpty restic -r /srv/restic-repo init
+
+
+Group accessible repositories
+*****************************
+
+Since restic version 0.14 local and SFTP repositories can be made
+accessible to members of a system group. To control this we have to change
+the group permissions of the top-level ``config`` file and restic will use
+this as a hint to determine what permissions to apply to newly created
+files. By default ``restic init`` sets repositories up to be group
+inaccessible.
+
+In order to give group members read-only access we simply add the read
+permission bit to all repository files with ``chmod``:
+
+.. code-block:: console
+
+    $ chmod -R g+r /srv/restic-repo
+
+This serves two purposes: 1) it sets the read permission bit on the
+repository config file triggering restic's logic to create new files as
+group accessible and 2) it actually allows the group read access to the
+files.
+
+.. note:: By default files on Unix systems are created with a user's
+          primary group as defined by the gid (group id) field in
+          ``/etc/passwd``. See `passwd(5)
+          <https://manpages.debian.org/latest/passwd/passwd.5.en.html>`_.
+
+For read-write access things are a bit more complicated. When users other
+than the repository creator add new files in the repository they will be
+group-owned by this user's primary group by default, not that of the
+original repository owner, meaning the original creator wouldn't have
+access to these files. That's hardly what you'd want.
+
+To make this work we can employ the help of the ``setgid`` permission bit
+available on Linux and most other Unix systems. This permission bit makes
+newly created directories inherit both the group owner (gid) and setgid bit
+from the parent directory. Setting this bit requires root but since it
+propagates down to any new directories we only have to do this priviledged
+setup once:
+
+.. code-block:: console
+
+    # find /srv/restic-repo -type d -exec chmod g+s '{}' \;
+    $ chmod -R g+rw /srv/restic-repo
+
+This sets the ``setgid`` bit on all existing directories in the repository
+and then grants read/write permissions for group access.
+
+.. note:: To manage who has access to the repository you can use
+          ``usermod`` on Linux systems, to change which group controls
+          repository access ``chgrp -R`` is your friend.
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -24,6 +24,7 @@ type Local struct {
 	Config
 	sem *backend.Semaphore
 	backend.Layout
+	backend.Modes
 }
 
 // ensure statically that *Local implements restic.Backend.
@@ -42,10 +43,15 @@ func open(ctx context.Context, cfg Config) (*Local, error) {
 		return nil, err
 	}
 
+	fi, err := fs.Stat(l.Filename(restic.Handle{Type: restic.ConfigFile}))
+	m := backend.DeriveModesFromFileInfo(fi, err)
+	debug.Log("using (%03O file, %03O dir) permissions", m.File, m.Dir)
+
 	return &Local{
 		Config: cfg,
 		Layout: l,
 		sem:    sem,
+		Modes:  m,
 	}, nil
 }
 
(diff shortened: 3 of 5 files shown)


greatroar:
This applies to the new mode also to the SFTP backend, which does not ([cannot](https://pkg.go.dev/github.com/pkg/sftp?utm_source=godoc#Client.Chmod), at least not portably) honor the local umask.

DanielG:
Indeed the sftp backend handles this by chmod'ing instead:

`internal/backend/sftp/sftp.go:277`:
```
	// pkg/sftp doesn't allow creating with a mode.
	// Chmod while the file is still empty.
	if err == nil {
		err = f.Chmod(backend.Modes.File)
	}
```

beewoolie:
This makes me wonder a bit.  If I understand the change correctly, the init of the repo requires an extra step, after creation, to change the permissions in the repo directories.  If that's true, I'd say it is likely to be more evident to users if an option is added to the init subfunction that allows local and sftp to backends to manage permissions in a way that permits multiple users explicitly.  It might need to have a group name, for example, so that it will, at a minimum, make the directories sticky for this group.

Making this an option for the repo means that it can be queried as part of the configuration.  Inspection of the setup further enables validation of the expected configuration.

That said, this is something that blocks an expected use case for restic, multiple contributing users to the backup data store with a local store.

DanielG:
On Sat, Nov 13, 2021 at 12:55:07PM -0800, Marc Singer wrote:
> If I understand the change correctly, the init of the repo requires an
> extra step, after creation, to change the permissions in the repo
> directories.  

Yeah you got that right. The user/admin has to do a chmod call after repo
init, but only once ever after that it all just works.

> If that's true, I'd say it is likely to be more evident to users if an
> option is added to the init subfunction that allows local and sftp to
> backends to manage permissions in a way that permits multiple users
> explicitly.

Just to be clear since you bring up sftp again: this PR makes the shared
repo use-case work for both local and sftp repos. Maybe I wasn't quite
clear on that point above.

The only problem with the sftp code is that the sftp server side can't
currently restict the modes further than what we set in restic because we
set the modes we want instead of taking the modes the file was created with
by the server side into account.

Do keep in mind however that the sftp umask was being ignored already and
this PR doesn't represent a change in behaviour on that end. So if an sftp
server was trying to restrict the file permissions that way it was already
broken :)

> It might need to have a group name, for example, so that it will, at a
> minimum, make the directories sticky for this group.

Right, or (default) ACLs could be used. But restic just doesn't need to be
involved with any of that. Since we're alreay in a situation where a user
has to create a system group to make this work I don't really see why we
would need a fancy option in the init command to do what a chmod call can
do.

> Making this an option for the repo means that it can be queried as part
> of the configuration.  Inspection of the setup further enables validation
> of the expected configuration.

Sure, but honestly that sounds like waaaaaay more work than literally a one
character change. I don't think merging this as-is would proclude anyone
from implementing what you suggest though.

As things stand tough this is a major pain in my neck as I have to maintain
a restic fork just to make backups work for systems in my domain :]

The only real problem with my approach is a hypothetical repo where someone
accidentally changed the directory mode to world/group readable without
wanting files to be exposed. I find that scenario sufficiently unlikey not
to worry about, especially since the repo files are encrypted to boot.

--Daniel

beewoolie:
I'd certainly be happy to have this wee change applied.  The lack of a willing member of the team to review the change is, well, disapointing.

My point with the options is that it daylights what could be a hidden detail.  When it comes to permissions, IMHO, I'd rather give a little extra boost to make the effect apparent.  It adds some noise to be sure.

And, on the other hand, we seldom run these systems with shared groups (e.g.) such that the likely impact of this change is harmless.

MichaelEischer:
The change look like a big hack, that will cause problems if restic ever has to create a new directory. I'd rather propose a different approach:

Use the current permissions of the `config` file to let the local backend (and probably sftp too) detect whether the repository should be accessible only to the current user or also to the group. A file permission of `0400` would mean user-only and `0440` would allow group access. To switch between both variants, just run chmod on the repository.

DanielG:
@MichaelEischer Good idea. I've implemented it.

beewoolie:
Interesting strategy.  Makes me wonder how it will be documented.  IMHO, it seems discrete to use the file mode of a configuration file created by restic to determine file modes for files created within the repo.

changelog/unreleased/pull-3419
```diff
@@ -0,0 +1,18 @@
+Enhancement: Use config file modes to derive new file/dir modes
+
+Previously restic would always create new files and directories on
+local/sftp repositories with hardcoded permissions of 0700/0600
+(rwx/rw). We now use the current read permissions of the "config" file to
+derive which modes new files and directories should be created with.
```
MichaelEischer, line 6: That is not really correct. E.g. a permission of 0444 will never allow anyone to read/write files in the repository. It's just the question whether the files should be group accessible or not.

changelog/unreleased/pull-3419
```diff
@@ -0,0 +1,18 @@
...
+
+For example the mapping in the cases below would look like this:
```
MichaelEischer, line 8: Just state that a config file that is only readable by the user will lead to only user accessible files and that a group readable config will also allow the group to access the files. Btw. files are set to readonly once they are written, which makes the 0600 / 0660 permissions an implementation detail.

It might be worthwhile to explain how to switch a repository between both permission variants.

MichaelEischer (changes requested):
Besides the comments I've added, the PR also should add a short notice to the documentation (probably to doc/030_preparing_a_new_repo.rst) that a repository can be made user/group accessible depending on the config file permissions and how to switch.

MichaelEischer:
> Interesting strategy. Makes me wonder how it will be documented. IMHO, it seems discrete to use the file mode of a configuration file created by restic to determine file modes for files created within the repo.

The `config`  file is part of the repository and not some external configuration file. So, I don't really understand what should be a problem here?

DanielG (commented):
Changes:
- Address review comments
- Rewrite changelog
- Add docs

DanielG:
@MichaelEischer ping, I'd appreciate another review.

Changes:
- Improve wording in docs

DanielG:
Ok, so it looks like this can't work as-is yet because the config file is created not as part of the backend Create function but rather by the repository Init/init functions.

MichaelEischer:
> Ok, so it looks like this can't work as-is yet because the config file is created not as part of the backend Create function but rather by the repository Init/init functions.

My suggestion would to ignore the stat error in `Open` if the config file does not exist (yet).

DanielG:
Oh yeah, another good idea. Implemented.

doc/030_preparing_a_new_repo.rst
```diff
@@ -693,3 +693,55 @@ On MSYS2, you can install ``winpty`` as follows:
...
+repository config file triggering restic's logic to create new files as
+group accessible and 2) it actually allows the group read access to the
+files.
+
+.. note:: By default files on Unix systems are created with a user's
+          primary group as defined by the gid field in ``/etc/passwd``. See
+          `passwd(5)
+          <https://manpages.debian.org/latest/passwd/passwd.5.en.html>`_.
+
+To manage group membership on Linux systems ``usermod`` can be used. To
```
MichaelEischer, line 726: I'm not entirely sure what this paragraph tries to tell me. I'd just leave it out and mention that posix ACLs are an alternative to using setgid.

MichaelEischer (commented):
I've rebased to PR to fix merge conflicts with the current master. Besides the few comments on the documentation entry, the code seems fine.

DanielG:
I've reworked the docs a bit and (hopefully) properly resolved the conflict with cd783358d33b59412e3e8ad55b02fcbf83c17d0c by undoing the merging of code from `Open`/`Create` into `open`. @MichaelEischer I think your conflict resolution would have failed to work in the `Create` code path as the config file wouldn't exist there yet so stat would fail.

doc/030_preparing_a_new_repo.rst
DanielG: I'm just trying to account for users that aren't super familiar with group managment so I just wanted to drop the names of the relevant tools in here somewhere so they can find docs on this stuff. I've re-arranged this paragraph to be a not eat the end now that should make the flow more smooth.

MichaelEischer (approved):
LGTM. Thanks!

I've rebased the PR to fix the merge conflict in doc/030_preparing_a_new_repo.rst .

PR #3420, opened 2021-06-07
Modernize error handling in local backend

greatroar:

* Stop prepending the operation name: it's already part of os.PathError, leading to repetitive errors like "Chmod: chmod /foo/bar: operation not permitted".

* Use errors.Is to check for specific errors.

Follow-up to #2980.

Checklist
---------

### DIFF
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -64,7 +64,7 @@ func Create(ctx context.Context, cfg Config) (*Local, error) {
 	for _, d := range be.Paths() {
 		err := fs.MkdirAll(d, backend.Modes.Dir)
 		if err != nil {
-			return nil, errors.Wrap(err, "MkdirAll")
+			return nil, errors.WithStack(err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/backend/local/local.go
```diff
@@ -213,10 +213,10 @@ func (b *Local) Test(ctx context.Context, h restic.Handle) (bool, error) {
 	debug.Log("Test %v", h)
 	_, err := fs.Stat(b.Filename(h))
 	if err != nil {
-		if os.IsNotExist(errors.Cause(err)) {
+		if errors.Is(err, os.ErrNotExist) {
```
MichaelEischer, line 216: Shouldn't that also use `IsNotExist`?

MichaelEischer (commented):
Seems to be fine in general, I've only added one small comment.

internal/backend/local/local.go
greatroar: The os.IsNotExist doc says "New code should use errors.Is(err, os.ErrNotExist)".

greatroar: More practically, if we use errors.Is everywhere, we can use custom errors on Windows that contain the fixed path as well as the original path, if they differ.

MichaelEischer: Oh, I meant to write `b.IsNotExist`, which is also used in a few other places. Sorry for the confusion.

greatroar: Excuse me, I misunderstood. That is cleaner, changed the code.

MichaelEischer (approved):
LGTM. Thanks!

PR #3425, opened 2021-06-12
Make restore resumable

### LINKED ISSUE #407: Make restore resumeable.

### LINKED ISSUE #2662: Restore only changed files
Output of `restic version`
--------------------------

`restic 0.9.5 compiled with go1.12.4 on windows/amd64`

I know that this isn't a synchronization software, but it would be veeeery nice to implement a `restore` function which restores only changed files in a directory and deletes files which are present in that directory but not in the snapshot i am restoring. To better clarify this, think of it like a sort of `git checkout`

What are you trying to do?
--------------------------

Suppose that I have a 500GB folder `localdata` on my pc and a dir named `myrepo` on an SFTP server, in which i backup `localdata`. I would like to work on `localdata` in an agile mode, making frequent changes to the directory internal structure, moving files, renaming, deleting some useless content, ... and so on. Every time i feel it is necessary, i execute `restic backup -v localdata -r sftp:user@host:myrepo`

Suddenly, a week later i wake up and i find out that some of my important files in `localdata` are gone, because i deleted them accidentally or maybe my sister deleted them because she doesn't know how to properly use a computer... I'd like to fully restore a previous version of the folder from `myrepo` without the need of redownloading the entire 500GB data from the server. That's what i usually do with `git checkout <old_commit_id>` except that i don't have 500GB git repos :sweat_smile:

Yeah, this software solved me a lot of problems. I use it **daily** for my **main** backups. I also appreciate this wonderful idea of asking users some positive feedback here in issues.

Restic is awesome! Keep up the good work :wink:

aawsome:

With this PR, `restic resume` checks if existing files (partly) match the files to be restored. If matching blobs exist, they are no longer downloaded for restore.

With this change, `restic restore` is now resumable.

Note, that before checking the file contents of an existing file, it size is checked. This means if you use this PR on a platform that doesn't support pre-allocation of files, partly resumed files will need a complete download.
However fully restored files never need to be saved again.

closes #407 
closes #2662
See https://forum.restic.net/t/22tb-restore-mistakes-were-made-lessons-learned/4042

Checklist
---------

### DIFF
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -253,7 +253,9 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 				idx.Add(node.Inode, node.DeviceID, location)
 			}
 
-			filerestorer.addFile(location, node.Content, int64(node.Size))
+			// Check if target exists and has the right contents
+			matches, _ := res.verifyFile(target, node)
+			filerestorer.addFile(location, node.Content, matches, int64(node.Size))
 
 			return nil
 		},
@@ -305,6 +307,59 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 	return err
 }
 
+// verifyFile checks if the given target file's contents match the
+// blobs given in node
+// It first checks the file size and then checks all file blobs
+// It always returns the successfully checked blobs
+func (res *Restorer) verifyFile(target string, node *restic.Node) (matchingBlobs []bool, err error) {
+	debug.Log("calling verifyFile")
+
+	matchingBlobs = make([]bool, len(node.Content))
+	stat, err := os.Stat(target)
+	if err != nil {
+		return
+	}
+	if int64(node.Size) != stat.Size() {
+		return matchingBlobs, errors.Errorf("Invalid file size: expected %d got %d", node.Size, stat.Size())
+	}
+
+	file, err := os.Open(target)
+	if err != nil {
+		return
+	}
+
+	offset := int64(0)
+	wrongOffset := offset
+	foundNotMatch := false
+	for i, blobID := range node.Content {
+		length, _ := res.repo.LookupBlobSize(blobID, restic.DataBlob)
+		buf := make([]byte, length) // TODO do I want to reuse the buffer somehow?
+		_, err = file.ReadAt(buf, offset)
+		if err != nil {
+			_ = file.Close()
+			return
+		}
+
+		switch {
+		case blobID.Equal(restic.Hash(buf)):
+			matchingBlobs[i] = true
+		case !foundNotMatch:
+			wrongOffset = offset
+			foundNotMatch = true
+		}
+
+		offset += int64(length)
+	}
+
+	debug.Log("file %v matching blobs: %v", target, matchingBlobs)
+
+	if foundNotMatch {
+		return matchingBlobs, errors.Errorf("Unexpected contents starting at offset %d", wrongOffset)
+	}
+
+	return matchingBlobs, file.Close()
+}
+
 // Snapshot returns the snapshot this restorer is configured to use.
 func (res *Restorer) Snapshot() *restic.Snapshot {
 	return res.sn
@@ -323,36 +378,8 @@ func (res *Restorer) VerifyFiles(ctx context.Context, dst string) (int, error) {
 			}
 
 			count++
-			stat, err := os.Stat(target)
-			if err != nil {
-				return err
-			}
-			if int64(node.Size) != stat.Size() {
-				return errors.Errorf("Invalid file size: expected %d got %d", node.Size, stat.Size())
-			}
-
-			file, err := os.Open(target)
-			if err != nil {
-				return err
-			}
-
-			offset := int64(0)
-			for _, blobID := range node.Content {
-				length, _ := res.repo.LookupBlobSize(blobID, restic.DataBlob)
-				buf := make([]byte, length) // TODO do I want to reuse the buffer somehow?
-				_, err = file.ReadAt(buf, offset)
-				if err != nil {
-					_ = file.Close()
-					return err
-				}
-				if !blobID.Equal(restic.Hash(buf)) {
-					_ = file.Close()
-					return errors.Errorf("Unexpected contents starting at offset %d", offset)
-				}
-				offset += int64(length)
-			}
-
-			return file.Close()
+			_, err := res.verifyFile(target, node)
+			return err
 		},
 		leaveDir: func(node *restic.Node, target, location string) error { return nil },
 	})
--- a/internal/restorer/filerestorer.go
+++ b/internal/restorer/filerestorer.go
@@ -35,6 +35,7 @@ type fileInfo struct {
 	size       int64
 	location   string      // file on local filesystem relative to restorer basedir
 	blobs      interface{} // blobs of the file
+	exists     []bool      // indicates blobs that already exist in file before resuming
 }
 
 type fileBlobInfo struct {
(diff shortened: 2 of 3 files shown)


internal/restorer/restorer.go
```diff
@@ -305,6 +307,59 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
...
 
+// verifyFile checks if the given target file's contents match the
+// blobs given in node
+// It first checks the file size and then checks all file blobs
+// It always returns the successfully checked blobs
+func (res *Restorer) verifyFile(target string, node *restic.Node) (matchingBlobs []bool, err error) {
+	debug.Log("calling verifyFile")
+
+	matchingBlobs = make([]bool, len(node.Content))
+	stat, err := os.Stat(target)
```
greatroar, line 318: Shouldn't this be Lstat? Otherwise, a symlink to a matching file is considered a match.

internal/restorer/restorer.go
```diff
@@ -253,7 +253,9 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 				idx.Add(node.Inode, node.DeviceID, location)
 			}
 
-			filerestorer.addFile(location, node.Content, int64(node.Size))
+			// Check if target exists and has the right contents
+			matches, _ := res.verifyFile(target, node)
```
greatroar, line 257: Unchecked error.

internal/restorer/restorer.go
```diff
@@ -305,6 +307,59 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
...
+	if err != nil {
+		return
+	}
+
+	offset := int64(0)
+	wrongOffset := offset
+	foundNotMatch := false
+	for i, blobID := range node.Content {
+		length, _ := res.repo.LookupBlobSize(blobID, restic.DataBlob)
+		buf := make([]byte, length) // TODO do I want to reuse the buffer somehow?
```
greatroar, line 336: Instead of reading the whole blob into memory to get its hash, you could use a SectionReader with a smaller buffer:

```go
sha := sha256.New()
buf := make([]byte, 8192)
for i, blobID := range node.Content {
	length, _ = res.repo.LookupBlobSize(blobID, restic.DataBlob)
	block := io.NewSectionReader(file, offset, int64(length))

	h.Reset()
	_, err = io.CopyBuffer(sha, block, buf)
	// check error

	h := sha.Sum(buf)
	// ...
}
```

MichaelEischer: As the size of a blob is limited to a few MB, it shouldn't hurt to load the full blob?

internal/restorer/restorer.go
ianmaddox: I actually have a use case for that: There some exceptionally large files that I already have restored and redeployed elsewhere in the filesystem. With this I could symlink it within the restore directory and not worry about having two copies or having it restored again. Perhaps a command line flag to ignore symlinks would be helpful for the opposite case?

internal/restorer/restorer.go
aawsome: This is intentional as matches is always returned. If there is any problem with checking the file, there will be no matches and the file will be fully restored. In fact, here is no error case to check.

internal/restorer/restorer.go
aawsome: I just copied the already exiting code from VerifyFiles into this new method VerifyFile. I would not change this in this PR.

internal/restorer/restorer.go
aawsome: You are right. I just copied this from the already existing `VerifyFiles` but in fact, this won't correctly verify restored files if they were exchanged by symlinks (whoever would 

 Changing this here also is not sufficient as the file reading part below would also follow symlinks. I just realized that the archiver code uses `arch.FS.OpenFile(target, fs.O_RDONLY|fs.O_NOFOLLOW, 0)`, so I think this should be changed too, here.

@ianmaddox I would not implement your use case in this PR. IMO it is an additional feature. So far, `restore` always will *exactly* reproduce the snapshot contents.

Maybe you open a separate issue for your use case after this PR is finished and discuss this independent from here?

MichaelEischer:
It might be worth taking a look at the comments in https://github.com/restic/restic/pull/2012#issuecomment-430200048 .

aawsome:
>Few notes/observations
>
>    I suggest to make verify-before-restore behaviour optional. For example, it is likely to slow down  restore if the target file system is on 2.5" laptop hdd (i.e. slow mechanical disk) and the backend is on fast 1Gbps connection.

The current implementation in this PR first checks the existence of a file to restore. Only if that file does exist and the size matches the expected size, it is read to check if the blobs also match. I think in most cases the file either does not exits (in this case only the existence-check is added which shouldn't be expensive) or exists already-restored (in this case the file is read instead of written which IMO should be faster). Only the edge cases of partly-restored files may introduce a negative performance impact - and this also only in edge cases which much faster backend than the local HDD.
Also users could always clean the restore path before restoring.

So I vote against making this optional.

>    I think we should investigate if pre-restore file content verification should run on single or multiple threads. This likely depends on the type of disks and filesystem used but I don't know if we should use different concurrency strategy on ssd vs hdd, for example.

I agree but would leave this open to a future improvement.

>    It may be desirable to implement verify-before-restore at blob level, which will allow to resume interrupted restore and incremental restore of large files.

This is exactly what this PR does. However, it does not recognize if a blob missing in a file-to-restore is already present in another file which already exists before starting the restore. But IMO this is too complex to implement in this PR.

>    We need to decide what to do with existing files that are not part of the snapshot, i.e. should restore delete or keep them?

This is a good question. This actual behavior of restic is to keep files if they are present and only overwrite files that are part of the snapshot. This PR does not change this behavior.

nickchomey:
I'd love to have this funcitonality! I hope you guys can figure it out and merge it soon!

aawsome:
@nickchomey You might be right. This PR in fact only restores changed files, but in #2662 the requirements seem to be "only restore changed files" + "remove files if not present in snapshot". The second part is not solved here.

Given the age of this PR I suggest to wait for the review progress before discussing formal implication about what issue is closed.

aawsome:
Note that this feature (+ using blobs from existing files to restore missing/incomplete files) is now included in rustic, see https://github.com/rustic-rs/rustic/pull/90

JsBergbau:
I'd also really like to have this feature since using mount and rsync is much slower than using restore command, see https://github.com/restic/restic/issues/3828

Hope this will be implemented soon, especially since this featurerequest is dating back to 2016 https://github.com/restic/restic/issues/407

cameronbraid:
I would like a feature like this to be able to use restic as a repository for content distribution.  For my use case I would want to say to restic.  Apply restore from snapshotA to snapshotB in folder X.   So all the information about files to add/update/delete comes from the snapshot diff.  Would something like that be workable ?

MichaelEischer:
> Restore from snapshotA to snapshotB in folder X. So all the information about files to add/update/delete comes from the snapshot diff. Would something like that be workable ?

@cameronbraid  I'm not sure whether I understand your suggestion. It sounds like you suggest to calculate the diff between two snapshots and then only restore files which differ between the snapshots? If yes, it sounds like a rather special use case. The `restore` command is intended to restore a folder to the state it had in the given snapshot. Thus I'm not sure whether your suggestion is a feature that is worth implementing, but in either case it is orthogonal to this PR.

aawsome:
> I would like a feature like this to be able to use restic as a repository for content distribution. For my use case I would want to say to restic. Restore from snapshotA to snapshotB in folder X. So all the information about files to add/update/delete comes from the snapshot diff. Would something like that be workable ?

@cameronbraid Can you please specify your use case a bit more in detail? For me it doesn't make much sense to only apply the diff of two snapshots to a restore dir. What is your intention to do so?

cameronbraid:
sorry, I should have not taken over this thread with my more specific, and slightly different use case.   I'm wanting to use restic as a content store for static web sites.  So in my case I know what snapshot is already restored in a folder and to get the updates I can roll restore to the next snapshot.

cameronbraid:
using git would force me to keep all history for old builds, where as with restic I can forget old snapshots.  a very useful function

also git is slower - for example a small site with 200mb of data (hundreds of files both text and binary images etc..) 'git init, add, commit takes' 15s, where as restic backup takes 7.5s

darkdragon-001:
@aawsome has this been implemented somewhere else or why was it closed?

MichaelEischer:
@darkdragon-001 If you really want to know why the PR was closed, take a look at https://github.com/restic/restic/pull/4705#issuecomment-1963040823 :-( . Resumable restores remain on the roadmap for restic 0.17

aawsome:
@darkdragon-001 As I already mentioned, yes it is implemented somewhere else and you can use that until it is implemented in restic.
Besides this, I simply closed my activities for restic. Anyone who wants to continue is free to be inspired or copy code changes which I proposed here.

Issue #2323, opened 2019-07-01
Fatal: config cannot be loaded: ciphertext verification failed


minhdanh:
Output of `restic version`
--------------------------
0.9.4

How did you run restic exactly?
-------------------------------
```
restic snapshots
Fatal: config cannot be loaded: ciphertext verification failed

restic -vvv check --read-data
using temporary cache in /tmp/restic-check-cache-369159670
Fatal: config cannot be loaded: ciphertext verification failed

restic rebuild-index
Fatal: config cannot be loaded: ciphertext verification failed
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Minio

Expected behavior
-----------------
restic lists all snapshots

Actual behavior
---------------
restic failed with the above error

Steps to reproduce the behavior
-------------------------------
I have a cronjob to backup the files using restic. It used to work.
But after some days the job failed with the same error.
When I use the command line to check, all restic commands failed with the same error.

Do you have any idea what may have caused this?
-----------------------------------------------
No idea

Do you have an idea how to solve the issue?
-------------------------------------------
No idea

Did restic help you or made you happy in any way?
-------------------------------------------------
Yes

rawtaz:
Is there an SSL/TLS certificate involved that might have expired or for some other reason isn't trusted?

minhdanh:
Ah, there isn't. I'm using a plain http service (a service of another namespace in Kubernetes) for minio with restic.
So maybe that shouldn't be the case.

```
/ # env | grep REP
RESTIC_REPOSITORY=s3:http://minio.minio.svc.cluster.local:9000/backup
```

minhdanh:
The backups that rely on restic in our infra have been failing. Can someone please have a look?

kun93:
I’ve had a similar issue in the past . I would recommend checking if you have multiple keys in your repo . You should be able to run ’restic ls keys’ to check or looking at your repo manually . If you do, then you can move the key with the newest creation date and then it should work . I don’t recall which restic version allows you to specify a key but you could try that flag as well

minhdanh:
@kun93 That command also failed with the same reason:

```
restic ls keys
Fatal: config cannot be loaded: ciphertext verification failed
```

rawtaz:
@minhdanh I'm at a loss as to why you're getting that error if the files in the repo are still intact. I'd suspect they've been corrupted somehow.

[...]

PR #3429, opened 2021-06-13
key: Check that new key works before deleting the old one

MichaelEischer:
When adding a new key or changing the password of a key, restic currently just creates the new key (and removes the old one, when changing the password). There's no verification that the new key works for sure. It could be possible that a random error breaks the new key or that it is not stored properly.

This PR adds a check that the new key actually works before continuing.

Not really, I've mentioned the idea in #2323.

Checklist
---------


fd0 (approved):
Great idea!

PR #3436, opened 2021-06-20
Save files under temporary name in local backend

### LINKED ISSUE #3435: Fatal: hash does not match id
_Originally posted by @malemburg in https://github.com/restic/restic/issues/1999#issuecomment-857046947_

I've compiled all the information requested in the template.

How can I run "debug examine" ?

Output of `restic version`
--------------------------

On the server:

On the notebook:

restic 0.12.0 compiled with go1.15.8 on windows/amd64 

On the notebook:

```
set RESTIC_REPOSITORY=\\<server>\restic
set RESTIC_PASSWORD_FILE=c:\restic\restic-repo-pwd.txt
set LOG=c:\restic\backup.log
c:\restic\restic backup --use-fs-snapshot C:\Users\<username> C:\Data >> %LOG% 2>&1
```

A Linux server, running Samba 4.11. The repo is mounted on the notebook as an SMB share.

The notebook is running Window 10 20H2 with all patches applied.

Expected behavior
-----------------

Successful snapshot or at least an interrupted run without broken repo files (extra packs are fine, even unindexed ones).

Actual behavior
---------------

Every few weeks, backups stop to work. This often happens as a result of network connectivity issues of the notebook, which then result in the backup process taking too long and the Windows 10 scheduler terminating the process.

Here's a log extract from the notebook (with some parts translated from German):

```
...
Save(<data/eec07c2b38>) returned error, retrying after 15.321836548s: OpenFile: open \\?\UNC\<server>\restic\data\ee\eec07c2b380531c74fd3f5302ea22551d94f003d6d64af20515f3073c4f3766e: File exists
Save(<data/aecefbba94>) returned error, retrying after 26.801974529s: OpenFile: open \\?\UNC\<server>\restic\data\ae\aecefbba940d1fa1db1179cff5c90ef9bbfadc42c04cc9e234089972554a1141: File exists
Fatal: unable to save snapshot: OpenFile: open \\?\UNC\<server>\restic\data\29\29047f6b4348254ad25ce2ede71b60ebfe8df3ef66fd3d522a09754158beb534: File exists
```

The repo on the server ends up being full of non-indexed pack files as a result.

Rebuilding the index using rebuild-index works fine. This does not appear to check the hashes.

But analysis on the server using restic prune stops with an error, e.g.

```
...
repacking packs
[1:40:06] 46.52%  17169 / 36904 packs repacked
Fatal: hash does not match id: want acf2bbecc59239c0c2bbfe2769dc89f418a3d23f1904bc8a63f1469b6957352f, got 25f88ec3064875a900e910b83c5ee21e47934e8c3e652b6d13e4d12e4cb7ac0b
```

Checking on the server then confirms this:

```
# ll data/ac/acf2bbecc59239c0c2bbfe2769dc89f418a3d23f1904bc8a63f1469b6957352f
-rwxrw-r-- 1 x x 5177464 Jun  4 18:45 data/ac/acf2bbecc59239c0c2bbfe2769dc89f418a3d23f1904bc8a63f1469b6957352f
```

The file size roughly matches the file sizes of other data/ files.

```
# sha256sum data/ac/acf2bbecc59239c0c2bbfe2769dc89f418a3d23f1904bc8a63f1469b6957352f
25f88ec3064875a900e910b83c5ee21e47934e8c3e652b6d13e4d12e4cb7ac0b  data/ac/acf2bbecc59239c0c2bbfe2769dc89f418a3d23f1904bc8a63f1469b6957352f
```

In previous cases, where this happened, the files in question were clearly
truncated and had a much lower size than the other files under data/. However, not in this latest case. The file may still be truncated, but it's not as obvious as it was before.

I have not yet found a way to reproduce this. However, interrupting the backup process via the scheduler or Process Explorer does appear to trigger the issue.

My guess is that the interaction of restic with the SMB mount is causing this. 

E.g. it's possible that restic is writing to the file on the notebook, this gets (write) cached by the OS and then fails to flush to the server after the process gets terminated, creating a truncated file on the repo server.

The server cannot be at fault, since this is used by several other systems without any problems.

The notebook certainly does have network connectivity issues due to poor network coverage at times, but otherwise works fine and does not exhibit crashes, RAM or CPU issues. I'll run Prime95 as extra measure.

I don't know how restic works internally, but it is strange that files can end up in the repository which don't fulfill the invariant of sha256sum(file) == filename.

Restic should make sure that files which are written to the repo file system (even for network mounts) are consistent and only end up as valid repo files under data/ iff their hash matches the filename. If not already done, files should first be written to a temporary dir and then moved to data/ after successful completion only.

It may also be good to (possibly optionally) double check the written file by reading it back from the repo to check the hash and only add it to the data/ dir in case the hash matches.

greatroar:

Files in local backends are written to a temporary file first, then sync'd, then renamed. In case of a crash or network disconnect, broken pack files will end up with names of the form `*-tmp-*` and will not be picked up by other restic commands.

Fixes #3435.

This patch includes #3420 to avoid merge conflicts.

The code is similar to #2838, but does not take into account writing to a single repo from multiple sources (the repo is assumed to be locked).

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3436
+++ b/changelog/unreleased/pull-3436
@@ -1,0 +1,12 @@
+Enhancement: Improve local backend's resilience to (system) crashes
+
+Restic now ensures that files stored using the `local` backend are created
+atomically (that is, files are either stored completely or not at all). This
+ensures that no incomplete files are left behind even if restic is terminated
+while writing a file.
+
+In addition, restic now tries to ensure that the directory in the repository
+which contains a newly uploaded file is also written to disk. This can prevent
+missing files if the system crashes or the disk is not properly unmounted.
+
+https://github.com/restic/restic/pull/3436
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -3,6 +3,7 @@ package local
 import (
 	"context"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -88,7 +89,8 @@ func (b *Local) Save(ctx context.Context, h restic.Handle, rd restic.RewindReade
 		return backoff.Permanent(err)
 	}
 
-	filename := b.Filename(h)
+	finalname := b.Filename(h)
+	dir := filepath.Dir(finalname)
 
 	defer func() {
 		// Mark non-retriable errors as such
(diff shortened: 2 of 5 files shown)


cyounkins:
Thanks for this! I really like that this preserves the invariant `sha256sum(file) == filename` for more than referenced pack files. It's not a requirement but it gives me a warm fuzzy feeling.

With regard to the temporary naming, `rsync` chose to do `.<FILE_NAME>.<RANDOM_STRING>`. Your PR doesn't include random characters, and I think that's a simple solution to the concern about multiple writers. Thoughts?

Could you rebase since #3420 was merged?

greatroar:
ioutil.TempFile adds the random string. A rebase shouldn't be necessary since this PR is e5f0f67ba08770c1138619891cd37a1ce87c34b1 from #3420 + one commit on top of it.

MichaelEischer:
I'm wondering whether restic should sync the directory after renaming the temporary file (which was also sort of missing before)? Otherwise a crash of the host running restic could miss some file rename operations in the end. If we don't want a strict guarantee that pack files are created atomically, we could probably rename the file first and then sync both the file and its parent directory.

greatroar:
> If we don't want a strict guarantee that pack files are created atomically, we could probably rename the file first and then sync both the file and its parent directory.

In ACID terms, the new code is atomic on Unix, it's just not durable (borrowing an argument from the [renameio](https://github.com/google/renameio) docs). Atomicity is enough to fix #3435.

I'm not sure what you mean by "then sync both the file and its parent directory". For an atomic write, you need to sync the file before the rename. Sync'ing it (again?) afterwards doesn't do anything, since the filename is exclusively a part of the directory on Unix and Windows.

greatroar:
By the way, from reading the issue tracker for renameio (https://github.com/google/renameio/issues/1) and Go issues linked there (https://github.com/golang/go/issues/32188#issuecomment-497870124), I'm getting less convinced of the reliability of this solution on Windows. I can't precisely figure out what the issues discussed there are, though, and whether they affect restic...

MichaelEischer:
> I'm not sure what you mean by "then sync both the file and its parent directory". For an atomic write, you need to sync the file before the rename. Sync'ing it (again?) afterwards doesn't do anything, since the filename is exclusively a part of the directory on Unix and Windows.

I had durable but not atomic writes in mind, which should be the more important property for restic. For that it would be just important to sync the file and directory content to disk. Depending on how finegranular fsync is implemented by the filesystem, syncing the file content after the rename could avoid writing the directory twice (assuming the fsync of the file before the rename causes the directory to be written as well).

greatroar:
I see what you mean now. However, I don't have the resources to properly assess that optimization, and I can imagine one situation where it might be detrimental to performance, which is when a remote filesystem (NFS?) implements fsync by unconditionally sending a message to the server.

In any case, I've added a Sync on the directory in a new commit.

internal/backend/local/local.go
```diff
@@ -97,65 +99,86 @@ func (b *Local) Save(ctx context.Context, h restic.Handle, rd restic.RewindReade
...
+	}
+
+	// Now sync the directory to commit the Rename.
+	if !syncNotSup {
+		d, err := os.Open(dir)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		err = d.Sync()
+		_ = d.Close()
```
MichaelEischer, line 164: Shouldn't we keep that error in case `Sync` was successful?

MichaelEischer (commented):
It looks like it's not possible to (explicitly) sync the directory content on Windows. For the sync of the file content itself Go calls `FlushFileBuffers`  which only seems to support files. It is unclear to me whether that also includes the file metadata, at least the remark "To ensure that the metadata is flushed to disk, use the FlushFileBuffers function." (https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#caching_behavior) in a different help article suggests so. But that won't help with syncing the rename to disk...

greatroar:
I put the dir sync in local_{unix,windows}.go. From a web search, I get the idea that this is really not possible (maybe not needed?) on Windows.

internal/backend/local/local.go
greatroar: New version keeps the error.

MichaelEischer:
I've added a short changelog entry. Although, I'm not sure whether it sounds too dramatic or too technical.

changelog/unreleased/pull-3436
```diff
@@ -0,0 +1,12 @@
+Enhancement: Improve resilience of local backend to (system) crashes
+
+restic now ensures that files stored using the `local` backend are created
+atomically, that is files are either stored completely or not at all. This
+ensures that no incomplete files are left behind even if restic is terminated
+while writing a file.
```
greatroar, line 6: I would say "tries to ensure", because we can't guarantee anything. I don't think this is dramatic at all, LGTM.

MichaelEischer: You're right. Flush is (sadly) only a best effort operation.

MichaelEischer (approved):
LGTM.

MichaelEischer:
I've rebased the PR to the current master branch.

PR #3453, opened 2021-07-10
rest: Workaround Http2 zero-length reply bug

### LINKED ISSUE #2742: Issues with unreliable connections
Output of `restic version`
--------------------------
`restic 0.9.5 compiled with go1.12.9 on linux/amd64`
```
rclone v1.49.1
- os/arch: linux/amd64
- go version: go1.12.9
```
`MEGAcmd version: 1.2.0.0: code 1020000`

`sudo restic -r rclone:"MEGA WebDAV":/$MYDIR/"Cloud Drive"/restic check --read-data`
(sudo was unnecessary)

MEGA, via [MEGAcmd](https://github.com/meganz/MEGAcmd) and rclone.

Expected behavior
-----------------
Check should not be reporting pack files as having the hash `e3b0c442...` (hash of empty file) when the pack file has merely failed to download.

Actual behavior
---------------
While running `restic check`, `rclone` reports `rclone: $DATE $TIME ERROR :  $PACK: Didn't finish writing GET request (wrote 0/$NBYTES bytes): unexpected EOF` at a rate of roughly 8 times per hour. Alongside it was a restic error `Pack ID does not match, want $HASH, got e3b0c442` at a rate of a few times per hour. Manually investigating using an `rclone mount` and `sha256sum`, I found the packs to have the correct hash.

First, set up a restic repository on a MEGA drive. Once that is done, you'll want to use `MEGAcmd` to provide access to it via a WebDAV server, which I find to be the most reliable and performant access method (as opposed to an FTP server or rclone's own MEGA support).
```
$ mega-cmd
megacmd$ login $YOUREMAIL
email@example.com:/$ webdav /
``` 
With that done, set up an rclone WebDAV backend pointing to this server. Details should be username-less/password-less login on http://127.1:4443 by default. You should now be able to run the above command.

restic appears to be having a communication issue with rclone, and/or rclone with MEGA's WebDAV server. This unreliability may be caused by the high default level of parallelism, as ```find mnt/mega/restic/data -type f -exec sh -c 'echo "`basename $1` $1" | sha256sum --check --quiet || echo "f $1"' sh {} \;``` (a serial program over an `rclone mount`) seems to produce far fewer false positives - although it has yet to complete and likely will not for a couple of days. Regardless, it appears that an unreliable transport causes restic to assume the remote pack files are empty when they really just failed to download.

The simplest solution would be to request the file size and check against that before concluding the file has successfully downloaded, and provide an option to perform unlimited retries. An issue with this suggestion is that it may just lead to infinite loops in the odd case that the metadata and file data on the remote are out of sync.

While not perfect, restic provides precisely what I need: chunking backups with pluggable backends via rclone that makes minimal assumptions about the destination filesystem. The fact that I was able to back up my 350GB test folder to MEGA despite frequent and varied interruptions and save a whopping 140GB space due to deduplication has me immensely pleased - especially since I have terabytes of data yet to upload!

MichaelEischer:
The golang http client does not return an error when a HTTP2 reply includes a non-zero content length but does not return any data at all. This scenario can occur e.g. when using rclone when a file stored in a backend seems to be accessible but then fails to download.

This can cause errors like "Pack ID does not match, want $HASH, got e3b0c44" as reported in #2742.

The underlying issue is a bug in the go http client, which is also tracked as https://github.com/golang/go/issues/46071 . However, it is unknown when a fix will be available and I don't expect it to be backported to all go versions still supported by restic.  I've added a test that will fail once the underlying issue in the http.client has been fixed, to allow us to eventually remove the workaround again.

Fixes #2742 .
See also https://forum.restic.net/t/http2-stream-closed-connection-reset-context-canceled/3743/10 .

Checklist
---------

### DIFF
--- a/internal/backend/rest/rest.go
+++ b/internal/backend/rest/rest.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/textproto"
 	"net/url"
 	"path"
 	"strconv"
@@ -198,6 +199,44 @@ func (b *Backend) Load(ctx context.Context, h restic.Handle, length int, offset 
 	return err
 }
 
+// checkContentLength returns an error if the server returned a value in the
+// Content-Length header in an HTTP2 connection, but closed the connection
+// before any data was sent.
+//
+// This is a workaround for https://github.com/golang/go/issues/46071
+//
+// See also https://forum.restic.net/t/http2-stream-closed-connection-reset-context-canceled/3743/10
+func checkContentLength(resp *http.Response) error {
+	// the following code is based on
+	// https://github.com/golang/go/blob/b7a85e0003cedb1b48a1fd3ae5b746ec6330102e/src/net/http/h2_bundle.go#L8646
+
+	if resp.ContentLength != 0 {
+		return nil
+	}
+
+	if resp.ProtoMajor != 2 && resp.ProtoMinor != 0 {
+		return nil
+	}
+
+	if len(resp.Header[textproto.CanonicalMIMEHeaderKey("Content-Length")]) != 1 {
+		return nil
+	}
+
+	// make sure that if the server returned a content length and we can
+	// parse it, it is really zero, otherwise return an error
+	contentLength := resp.Header.Get("Content-Length")
+	cl, err := strconv.ParseUint(contentLength, 10, 63)
+	if err != nil {
+		return fmt.Errorf("unable to parse Content-Length %q: %w", contentLength, err)
+	}
+
+	if cl != 0 {
+		return errors.Errorf("unexpected EOF: got 0 instead of %v bytes", cl)
+	}
+
+	return nil
+}
+
 func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error) {
 	debug.Log("Load %v, length %v, offset %v", h, length, offset)
 	if err := h.Valid(); err != nil {
@@ -249,16 +288,10 @@ func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, o
 
 	// workaround https://github.com/golang/go/issues/46071
 	// see also https://forum.restic.net/t/http2-stream-closed-connection-reset-context-canceled/3743/10
-	if resp.ContentLength == 0 && resp.ProtoMajor == 2 && resp.ProtoMinor == 0 {
-		if clens := resp.Header["Content-Length"]; len(clens) == 1 {
-			if cl, err := strconv.ParseUint(clens[0], 10, 63); err == nil {
-				resp.ContentLength = int64(cl)
-			}
-			if resp.ContentLength != 0 {
-				_ = resp.Body.Close()
-				return nil, errors.Errorf("unexpected EOF got 0 instead of %v bytes", resp.ContentLength)
-			}
-		}
+	err = checkContentLength(resp)
+	if err != nil {
+		_ = resp.Body.Close()
+		return nil, err
 	}
 
 	return resp.Body, nil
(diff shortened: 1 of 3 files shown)


internal/backend/rest/rest.go
```diff
@@ -246,6 +247,20 @@ func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, o
 		return nil, errors.Errorf("unexpected HTTP response (%v): %v", resp.StatusCode, resp.Status)
 	}
 
+	// workaround https://github.com/golang/go/issues/46071
+	// see also https://forum.restic.net/t/http2-stream-closed-connection-reset-context-canceled/3743/10
+	if resp.ContentLength == 0 && resp.ProtoMajor == 2 && resp.ProtoMinor == 0 {
+		if clens := resp.Header["Content-Length"]; len(clens) == 1 {
+			if cl, err := strconv.ParseUint(clens[0], 10, 63); err == nil {
```
fd0, line 254: How about returning an error to the caller if the value of the `Content-Length` header cannot be parsed? If it is present but the value is invalid, that's an error. Otherwise, users will just see the "unexpected EOF" error when the server returns garbage in the header.

internal/backend/rest/rest.go
```diff
@@ -246,6 +247,20 @@ func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, o
...
+				resp.ContentLength = int64(cl)
+			}
+			if resp.ContentLength != 0 {
```
fd0, line 257: I find this error handling very confusing, mainly because we reuse `resp.ContentLength`.

fd0 (commented):
I've left a few comments and I've tried to make handling this case clearer. I've pushed a commit, please have a look!

internal/backend/rest/rest.go
```diff
@@ -246,6 +247,20 @@ func (b *Backend) openReader(ctx context.Context, h restic.Handle, length int, o
 		return nil, errors.Errorf("unexpected HTTP response (%v): %v", resp.StatusCode, resp.Status)
 	}
 
+	// workaround https://github.com/golang/go/issues/46071
+	// see also https://forum.restic.net/t/http2-stream-closed-connection-reset-context-canceled/3743/10
+	if resp.ContentLength == 0 && resp.ProtoMajor == 2 && resp.ProtoMinor == 0 {
+		if clens := resp.Header["Content-Length"]; len(clens) == 1 {
```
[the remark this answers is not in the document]
MichaelEischer: Looks like I've lost the information that the content-length parsing snippet is essentially the code from https://github.com/golang/go/blob/b7a85e0003cedb1b48a1fd3ae5b746ec6330102e/src/net/http/h2_bundle.go#L8646 . My intention was to behave the same way as the golang http2 implementation. As such checking that only a single content-length was returned by the server would ensure that the behavior matches.

internal/backend/rest/rest.go
MichaelEischer: If the golang client is unable to parse `Content-Length`, then `resp.ContentLength` is probably equal to `-1`. And in that case our error handling will be dead code.

Based on the same argument, it should probably be safe to use `resp.Header.Get("Content-Length")`

internal/backend/rest/rest.go
MichaelEischer: Indeed, whenever `ContentLength` gets changed then it's an error...

internal/backend/rest/rest.go
MichaelEischer: And to immediately contradict myself: Obviously the content-length was not parsed yet, because otherwise it wouldn't be zero...

internal/backend/rest/rest.go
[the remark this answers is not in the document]
fd0: Ok, I've added the check and a comment to the stdlib code.

fd0 (approved):
I'm satisfied, please have a look again @MichaelEischer!

MichaelEischer:
The cleanup looks fine now.

PR #3454, opened 2021-07-10
Update dependencies

MichaelEischer:
Update all dependencies except bazil.org/fuse.

Noteworthy changes include the following:

These could fix some backend issues, although restic should already have retried most of them:
- Azure/azure-sdk-for-go#14521 Retry on HTTP client errors
- minio/minio-go#1452 fix: make sure getObject returns error on truncated responses
- golang/net@f585440 http2: close Transport connection on write errors

Misc changes:
- The minio/sha256-simd library no longer includes custom SSSE3/AVX/AVX2 implementations and instead uses the default golang implementation which seems to be faster. The library now only adds support for SHA extensions available in some processors. See minio/sha256-simd#57 for more details.
- xattr 0.4.3 adds xattr support to Solaris and Illumos via pkg/xattr#54
- Cobra 1.2.1 by default adds a `completion` command to generate completions. However, we already have `generate` so I've disabled the new command for now.
- I've also cleaned up and updated the indirect dependencies. By now, the azure-sdk-for-go is the only remaining module without a go.mod file. Thus we only need indirect dependencies for it. I've removed all other indirect dependencies to use the library versions expected by the dependencies.

No.

Checklist
---------


MichaelEischer:
There's also a v2 of github.com/ncw/swift which adds a ctx parameter to the library methods.

The sftp library now also provides a ReadFromWithConcurrency method which might be worth investigating.

fd0 (approved):
LGTM!

PR #3474, opened 2021-08-05
Honor RESTIC_CACHE_DIR environment variable

### LINKED ISSUE #3382: restic check doesn't obey the RESTIC_CACHE_DIR environment variable
Output of `restic version`
--------------------------

I would like restic to be able to take the configuration for its cache directory from an environment variable, e.g. RESTIC_CACHE_DIR as well as from the cli option --cache-dir.

This would allow me to set the cache directory in the configuration file that gets sourced from the shell to gather the keys etc. and prevent the situation where someone forgetting to use --cache-dir fills up the root filesystem running an operation in restic that uses the cache.

Thank you.

kitone:

CacheDir option doesn't honor the `RESTIC_CACHE_DIR` environment variable by default, this is done by default now.
This avoid checking for `gopts.CacheDir` and `RESTIC_CACHE_DIR` environment variable.

Fix #3382: restic check doesn't obey the RESTIC_CACHE_DIR environment variable

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3382
+++ b/changelog/unreleased/issue-3382
@@ -1,0 +1,9 @@
+Change: honor RESTIC_CACHE_DIR environment variable in the check command
+
+--cache-dir option doesn't honor the RESTIC_CACHE_DIR environment variable by
+default. This fixes an issue with the restic check command that doesn't obey the
+RESTIC_CACHE_DIR environment variable. Extend the behavior of cache command to
+manage directories created by restic check.
+
+https://github.com/restic/restic/issues/3382
+https://github.com/restic/restic/pull/3474
--- a/cmd/restic/cmd_cache.go
+++ b/cmd/restic/cmd_cache.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/restic/restic/internal/cache"
@@ -140,8 +141,13 @@ func runCache(opts CacheOptions, gopts GlobalOptions, args []string) error {
 			size = fmt.Sprintf("%11s", formatBytes(uint64(bytes)))
 		}
 
+		name := entry.Name()
+		if !strings.HasPrefix(name, "restic-check-cache-") {
+			name = name[:10]
+		}
+
 		tab.AddRow(data{
-			entry.Name()[:10],
+			name,
 			fmt.Sprintf("%d days ago", uint(time.Since(entry.ModTime()).Hours()/24)),
 			old,
 			size,
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/restic/restic/internal/cache"
 	"github.com/restic/restic/internal/checker"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
@@ -146,6 +147,9 @@ func prepareCheckCache(opts CheckOptions, gopts *GlobalOptions) (cleanup func())
 	}
 
 	cachedir := gopts.CacheDir
+	if cachedir == "" {
+		cachedir = cache.EnvDir()
+	}
 
 	// use a cache in a temporary directory
 	tempdir, err := ioutil.TempDir(cachedir, "restic-check-cache-")
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -172,7 +172,7 @@ func updateTimestamp(d string) error {
 const MaxCacheAge = 30 * 24 * time.Hour
 
 func validCacheDirName(s string) bool {
-	r := regexp.MustCompile(`^[a-fA-F0-9]{64}$`)
+	r := regexp.MustCompile(`^[a-fA-F0-9]{64}$|^restic-check-cache-[0-9]+$`)
 	return r.MatchString(s)
 }
 
--- a/internal/cache/dir.go
+++ b/internal/cache/dir.go
@@ -6,10 +6,15 @@ import (
 	"path/filepath"
 )
 
+// EnvDir return $RESTIC_CACHE_DIR env
+func EnvDir() string {
+	return os.Getenv("RESTIC_CACHE_DIR")
+}
+
 // DefaultDir returns $RESTIC_CACHE_DIR, or the default cache directory
 // for the current OS if that variable is not set.
 func DefaultDir() (cachedir string, err error) {
-	cachedir = os.Getenv("RESTIC_CACHE_DIR")
+	cachedir = EnvDir()
 	if cachedir != "" {
 		return cachedir, nil
 	}


changelog/unreleased/issue-3382
```diff
@@ -0,0 +1,8 @@
+Bugfix: honor RESTIC_CACHE_DIR environment variable by default
```
MichaelEischer, line 1: The PR seems to be more of a `Change` than a Bugfix.

cmd/restic/global.go
```diff
@@ -111,7 +111,7 @@ func init() {
 	f.CountVarP(&globalOptions.Verbose, "verbose", "v", "be verbose (specify multiple times or a level using --verbose=`n`, max level/times is 3)")
 	f.BoolVar(&globalOptions.NoLock, "no-lock", false, "do not lock the repository, this allows some operations on read-only repositories")
 	f.BoolVarP(&globalOptions.JSON, "json", "", false, "set output mode to JSON for commands that support it")
-	f.StringVar(&globalOptions.CacheDir, "cache-dir", "", "set the cache `directory`. (default: use system default cache directory)")
+	f.StringVar(&globalOptions.CacheDir, "cache-dir", os.Getenv("RESTIC_CACHE_DIR"), "set the cache `directory`. (default: use system default cache directory)")
```
MichaelEischer, line 114: Although this seems to change the behavior of all commands, it actually has only an effect on the `check` command. All other commands already use the `DefaultDir()` function to determine the cache directory location. And that function also processes the environment variable.

So it's probably better to implement the change here instead:
https://github.com/restic/restic/blob/553ea36ca62c870d71cd31d4091c86d0ec984e82/cmd/restic/cmd_check.go#L148

I'd also prefer to not hardcode the environment variable name in different places in restic, but rather keep it limited to `internal/cache/dir.go`. That is please add a helper function there.

changelog/unreleased/issue-3382
```diff
@@ -0,0 +1,8 @@
...
+
+--cache-dir option doesn't honor the RESTIC_CACHE_DIR environment variable by
+default. This fix an issue with the restic check command that doesn't obey the
```
MichaelEischer, line 4: s/fix/fixes/

Mentioning the `check` command here is far too late. As it's the only affected command, it should already be mentioned in the changelog summary.

MichaelEischer (changes requested):
When placing the temporary check directories inside the regular restic cache directory, then no there's no guarantee that a cleanup of these directories will occur.

The problem is that the current cache cleanup mechanism explicitly ignores all directories which don't look like a repository id. Changing that would require a change in the following location or somewhere nearby:
https://github.com/restic/restic/blob/553ea36ca62c870d71cd31d4091c86d0ec984e82/internal/cache/cache.go#L174

kitone:
@MichaelEischer Thank you for your quick feedback and guidance. I have updated the pull request.

If `cache.DefaultDir` return an error, should we only continue to fall back with TempDir or disable the cache with warning ?

> When placing the temporary check directories inside the regular restic cache directory, then no there's no guarantee that a cleanup of these directories will occur.
> 
> The problem is that the current cache cleanup mechanism explicitly ignores all directories which don't look like a repository id. Changing that would require a change in the following location or somewhere nearby:
> 
> https://github.com/restic/restic/blob/553ea36ca62c870d71cd31d4091c86d0ec984e82/internal/cache/cache.go#L174

The temp directory `restic-check-cache-*` create a new subfolder cache dir with 64 character folder name and CACHEDIR.TAG file, changing the regex to accept this folder name will list the folder but not the content.
`cache.OlderThan` do not list recursively.

What should be the behaviour of `cache --cleanup` ? Always delete this folder ?

Best regards,

MichaelEischer:
> If `cache.DefaultDir` return an error, should we only continue to fall back with TempDir or disable the cache with warning ?

My reference to `DefaultDir` was only intended to illustrate what the other parts of restic, but wasn't meant to refer to the chagnes necessary here, sorry for the confusion. I'd like to have a new helper function next to `DefaultDir` which only queries the environment variable and does _not_ fall back to `os.UserCacheDir()`. With the current change in the PR, the temporary cache directory would always be placed in the regular cache directory unless the environment variable was changed. But the PR should only change the behavior when the environment variable is specified and keep the old behavior in all other cases.

> The temp directory `restic-check-cache-*` create a new subfolder cache dir with 64 character folder name and CACHEDIR.TAG file, changing the regex to accept this folder name will list the folder but not the content.
> `cache.OlderThan` do not list recursively.
> 
> What should be the behaviour of `cache --cleanup` ? Always delete this folder ?

I don't think that there's a reason to not just delete the whole `restic-check-cache-*` folder. Each check run should use its own temporary cache folder.

MichaelEischer:
> Changing `validCacheDirName` with allow to list the restic-check-cache-*. but cleanup will remove the wrong caches and keep the last.

I'm not sure I understand. Which caches would cleanup remove which it shouldn't have?

I'd still prefer to just extend the cache name regexp and maybe filter the check caches from the output. That would also have the benefit of allowing automatic cleanups by other restic commands using `--cleanup-cache`. 

> For now, `restic cache` will not show `restic-check-cache-*` caches, but with --cleanup, it will remove regular cache dirs, and then search for `restic-check-cache-*` names and remove them too.

The code currently deletes every check cache it encounters even if it was created recently. That way it could actually delete a cache currently used by running check operation. Is there are reason not to use the same "x days old" logic as for normal cache directories?

kitone:
> > Changing `validCacheDirName` with allow to list the restic-check-cache-*. but cleanup will remove the wrong caches and keep the last.
> 
> I'm not sure I understand. Which caches would cleanup remove which it shouldn't have?
> 
> I'd still prefer to just extend the cache name regexp and maybe filter the check caches from the output. That would also have the benefit of allowing automatic cleanups by other restic commands using `--cleanup-cache`.
> 
> > For now, `restic cache` will not show `restic-check-cache-*` caches, but with --cleanup, it will remove regular cache dirs, and then search for `restic-check-cache-*` names and remove them too.
> 
> The code currently deletes every check cache it encounters even if it was created recently. That way it could actually delete a cache currently used by running check operation. Is there are reason not to use the same "x days old" logic as for normal cache directories?

I'm really sorry, I didn't fully understand the cache cleaning behavior, I just re-read the code and do some test and indeed this behavior is the correct one. I have updated the pr.


I chose to display all the caches, in order to allow the user to have the information and thus choose if he should run a cleanup.

Maybe we can filter the name to have a more explicit listing ?

```
Repo ID     Last Used    Old  Size
-----------------------------------------
restic-che  90 days ago  yes   95.066 KiB
d82a4e2897  0 days ago         95.023 KiB
-----------------------------------------
```
by

```
Repo ID                       Last Used    Old  Size
-----------------------------------------------------------
restic-check-cache-037381024  90 days ago  yes   95.066 KiB
d82a4e2897                    0 days ago         95.023 KiB
-----------------------------------------------------------
```
or
```
Repo ID                       Last Used    Old  Size
-----------------------------------------------------------
037381024 (check)  90 days ago  yes   95.066 KiB
d82a4e2897         0 days ago         95.023 KiB
-----------------------------------------------------------
```

Thanks again for your feedbacks,

MichaelEischer:
> ```
> Repo ID                       Last Used    Old  Size
> -----------------------------------------------------------
> restic-check-cache-037381024  90 days ago  yes   95.066 KiB
> d82a4e2897                    0 days ago         95.023 KiB
> -----------------------------------------------------------
> ```

I like that variant. The number after `restic-check-cache` is not a repository id, so just displaying that number would be misleading. 

https://github.com/restic/restic/blob/6c84ea1412ecaec9b7bd02a08758c2aa8af46c29/cmd/restic/cmd_cache.go#L144

It should be sufficient to not shorten the repo ID when it starts with `restic-check-cache-`.

kitone:
Change added in the last commit.

Best regards,

MichaelEischer (approved):
LGTM. Thanks!

Issue #2679, opened 2020-04-03
High fragmentation and very slow backups


JsBergbau:
Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on windows/amd64

How did you run restic exactly?
-------------------------------
restic -r D:\restic backup C:\ G:\ --exclude-file
No enviroment variables

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
External USB 3.0 HDD drive

Expected behavior
-----------------
High throughput during backup, almost no fragmentation on the drive.
 

Actual behavior
---------------
When I get 30 MB/s write speed on the drive I'm very happy. Mostly it is < 10 MB/s while the utilisation is always > 90 % mostly stuck at 100 %.
I've found this issue with restoring https://github.com/restic/restic/issues/2675 
I think while backup up it is the same.
Restic syncs so heavily or uses so heavily IO that my whole USB3.0 subsystem crashes after a while. The external Seagate 4 TB USB 3.0 2.5" HDD drive disappears, comes back and so on. Disconnecting and reconnecting doesn't help. I have to reboot the whole PC. I've tried a lot backup applications with this disk so far. That never happend. To get the disk work I have to reboot. Could be a general hardwareproblem as well, but just to mention in this issue.

Take a look at windows taskmanager. > 18 seconds response time of the drive. A lot of time 0 write speed
![grafik](https://user-images.githubusercontent.com/37013344/78332436-8ae45800-7577-11ea-8552-ba22c748966e.png)

After this backup failed I ran defrag command. It showed 64 % (!) fragmentation.

This behaviour also puts high load on the drive, wearing it faster.

Don't know how exactly restic creates files, but small files (I think < 4 KB) are created on NTFS within the MFT Table. When the size gets bigger it has to be transfered out of the MFT-Table to the disk area. This puts additional load on the drive. With ressource monitor I also see a lot of traffic to MFT-Table. Could be because of creating new files or because my assumption is correct.

```
defrag /A D:
Microsoft-Laufwerksoptimierung
Copyright (c) Microsoft Corp.

"Analyse" wird für "Volume (D:)" aufgerufen...


Der Vorgang wurde abgeschlossen.

Post Defragmentation Report:

        Volumeinformationen:
                Volumegröße                 = 3,63 TB
                Freier Speicherplatz        = 3,38 TB
                Fragmentierter Speicherplatz insgesamt      = 64%
                Größter freier Speicherplatz     = 3,38 TB

[...]

PR #3475, opened 2021-08-07
Limit concurrent operations for local / sftp backend

### LINKED ISSUE #1763: Globally limit concurrent access to files in the repo
## Output of `restic version`
restic 0.8.3 (v0.8.3-221-g26757ae) compiled with go1.10.1 on windows/amd64

`restic -r i:\backup check --read-data`

external HDD-drive with usb3

## Expected behavior

reading/checking the data with full speed of HDD

## Actual behavior

backup need 2h of time
checking with --read-data option took 15h!

`restic -r i:\backup check --read-data`

if restic run, the windows-task-manager shows me a hdd-workload of 100% and reading only with 7MB/s

i found in the file `internal/checker.go` the variable `defaultParallelism = 40`
so that means 40 thread for reading on the disk.. right?

accessing 40 files at the same time is for a HDD too much workload.

if i set defaultParallelism=1 and compile it, restic need then much more time for starting:
'load indexes'
'check all packs'
'check snapshots, trees and blobs'

but after that.. reading/checking data ist much faster.
the windows-task-manager shows me a hdd-workload of 50% and reading speed is 45MB/s

with defaultParallelism=2
the windows-task-manager shows me a hdd-workload at 80-90% and
reading speed is 35-50MB/s

so i think to protect the hdd-drive i would prefer to set defaultParallelism=1 in that case.
my solution for faster start is to leave defaultParallelism=40   (or maybe 4 ?)
and add a defaultParallelism2=1 only for the ReadPacks-Function in the that file.

you musst me sorry for this description, beacause i am not a GIT-user and also not good in english
i changed for me this code and its working very well with a speed-up of factor 6
maybe you can change this in the master-branch?

EDIT:
after testing it on an SSD .. the cpu is now the bottleneck and here would be more worker way faster.
maybe a command-switch would be better like --read-worker 4 or --ssd
-> in this case more worker then cpu-threads are useless for hashing/checking ??
but for regular-hdd i would stay by 1 because the health of the hdd is more important fore me.

	// run workers
	const defaultParallelism2 = 1                    // NEW
	for i := 0; i < defaultParallelism2; i++ {     // CHANGED
		g.Go(func() error {
			for {
				var id restic.ID
				var ok bool

				select {
				case <-ctx.Done():
					return nil
				case id, ok = <-ch:
					if !ok {
						return nil
					}
				}

				err := checkPack(ctx, c.repo, id)
				p.Report(restic.Stat{Blobs: 1})
				if err == nil {
					continue
				}

				select {
				case <-ctx.Done():
					return nil
				case errChan <- err:
				}
			}
		})
	}

restic is superb! :-)

MichaelEischer:
Currently, the local and sftp backend are the only ones which do not limit the number of concurrent operations. 
This PR adds a `Connections` parameter for both backends along with the corresponding limits.

Even though `Connections` sounds unusual for the local backend, I've kept the name for consistency with the other backends.

Partially related to #322 and #2679

Checklist
---------
- ~~[ ] I have added tests for all code changes.~~ The backends already have a good test coverage.


fd0 (approved):
LGTM, just one question: why is the number of concurrent operations for the local backend 2, but for the sftp backend 5?

MichaelEischer:
For the local backend the number is essentially the same as the default from the archiver while reading data. SFTP usually has higher IO latency than a local backend so I've chosen a larger number of connections there. The value 5 is just the number that is also used for all other backends.

PR #3479, opened 2021-08-13
fixed --keep-within-X options in example

BUFU1610:
Simply edited a single example in the manual.
The order of the words was wrong in the example, i.e. fixed to --keep-within-hourly from --keep-hourly-within.

I hope not.


MichaelEischer (approved):
LGTM. Thanks for finding this!

BUFU1610:
> LGTM. Thanks for finding this!

No problem. Was looking for this option and discovered it by accident. ;)

PR #3481, opened 2021-08-16
Recover from no free space errors

### LINKED ISSUE #1153: Avoid disk-full condition
Restic fills the disk until the last byte.
If this happens, it is not possible to do anything:
- add more data (for the obvious reason)
- but more importantly, restore the data.

The way I solved it right now, is to create an 1GB file on my backup location, outside the directory where the restic backup is stored. In case of a full disk, I can remove this file and (hopefully) still restore the data.
But it would be nice if restic could prevent a full disk condition by itself.

MichaelEischer:
It adds an option that should allow prune to remove data from the repository with next to no free space at all. More precisely, it causes prune to not repack anything, then remove the index, unused pack files and rebuild the index afterwards. That way no free space is necessary to rebuild the index.

However, this comes at a cost: if prune fails while using that option, then a repository may become inaccessible. As `prune` won't add data to the repository between removing the old index and rebuilding it, there should always be enough free space to write the index again / rebuild it.

In the current state of this PR it might be necessary to manually remove all files from the `index` folder and run `rebuild-index` afterwards. I'm not sure whether it's worth it to also add a delete-then-rebuild option to `rebuild-index`.

Fixes #1153.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -39,7 +39,10 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 
 // PruneOptions collects all options for the cleanup command.
 type PruneOptions struct {
-	DryRun bool
+	DryRun                bool
+	UnsafeNoSpaceRecovery string
+
+	unsafeRecovery bool
 
 	MaxUnused      string
 	maxUnusedBytes func(used uint64) (unused uint64) // calculates the number of unused bytes after repacking, according to MaxUnused
@@ -56,6 +59,7 @@ func init() {
 	cmdRoot.AddCommand(cmdPrune)
 	f := cmdPrune.Flags()
 	f.BoolVarP(&pruneOptions.DryRun, "dry-run", "n", false, "do not modify the repository, just print what would be done")
+	f.StringVarP(&pruneOptions.UnsafeNoSpaceRecovery, "unsafe-recover-no-free-space", "", "", "UNSAFE, READ THE DOCUMENTATION BEFORE USING! Try to recover a repository stuck with no free space. Do not use without trying out 'prune --max-repack-size 0' first.")
 	addPruneOptions(cmdPrune)
 }
 
@@ -74,6 +78,10 @@ func verifyPruneOptions(opts *PruneOptions) error {
 			return err
 		}
 		opts.MaxRepackBytes = uint64(size)
+	}
+	if opts.UnsafeNoSpaceRecovery != "" {
+		// prevent repacking data to make sure users cannot get stuck.
+		opts.MaxRepackBytes = 0
 	}
 
 	maxUnused := strings.TrimSpace(opts.MaxUnused)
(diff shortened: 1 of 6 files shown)


cmd/restic/cmd_prune.go
```diff
@@ -73,6 +77,10 @@ func verifyPruneOptions(opts *PruneOptions) error {
 		}
 		opts.MaxRepackBytes = uint64(size)
 	}
+	if opts.UnsafeNoSpaceRecovery != "" {
+		// prevent repacking data to make sure users cannot get stuck.
+		opts.MaxRepackBytes = 1
```
rawtaz, line 82: Now that #3591 is merged this can be changed to `= 0` :-)

MichaelEischer: And now it's 0.

cmd/restic/cmd_prune.go
```diff
@@ -55,6 +58,7 @@ func init() {
 	cmdRoot.AddCommand(cmdPrune)
 	f := cmdPrune.Flags()
 	f.BoolVarP(&pruneOptions.DryRun, "dry-run", "n", false, "do not modify the repository, just print what would be done")
+	f.StringVarP(&pruneOptions.UnsafeNoSpaceRecovery, "unsafe-recover-no-free-space", "", "", "UNSAFE, READ THE DOCUMENTATION BEFORE USING! Try to recover a repository stuck with no free space. Do not use without trying out 'prune --max-repack-size 1' first.")
```
fd0, line 61: Shouldn't this also say `trying out 'prune --max-repack-size 0' first`?

cmd/restic/cmd_prune.go
```diff
@@ -131,6 +139,14 @@ func runPrune(opts PruneOptions, gopts GlobalOptions) error {
 		return err
 	}
 
+	if opts.UnsafeNoSpaceRecovery != "" {
+		repoID := repo.Config().ID
+		if opts.UnsafeNoSpaceRecovery != repoID {
+			return errors.Fatalf("must pass id '%s' to --unsafe-recover-no-free-space", repoID)
```
fd0, line 145: Nice idea :D

doc/060_forget.rst
```diff
@@ -392,3 +392,30 @@ The ``prune`` command accepts the following options:
...
+Recovering from "no free space" errors
+**************************************
+
+In some cases when a repository has grown large enough to fill up all disk space or the
+allocated quota, then ``prune`` might fail to free space. ``prune`` works in such a way
+that a repository remains usable no matter at which point the command is interrupted.
+However, this also means that ``prune`` requires some scratch space to work.
+
+In most cases it is sufficient to instruct ``prune`` to use as little scratch space as
+possible by running it as ``prune --max-repack-size 1``. Obviously, this can only work
```
fd0, line 406: Here too, shouldn't that say `--max-repack-size 0`? And maybe add the hint that for versions earlier than 0.13.0 `--max-repack-size 1` should be used.

fd0 (changes requested):

cmd/restic/cmd_prune.go
MichaelEischer: Oh, indeed, I've missed that after fixing `--max-repack-size 0`.

doc/060_forget.rst
MichaelEischer: I've applied both suggestions.

fd0 (approved):

Issue #3465, opened 2021-07-26
Improve temporary files handling on Windows


ducalex:
Output of `restic version`
--------------------------
restic 0.12.0-dev (compiled manually) compiled with go1.15.2 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
I would like to reduce the amount of disk writes when restic uses temporary files and prevent those files from being occasionally left behind.

What are you trying to do? What problem would this solve?
---------------------------------------------------------
Ideally I would have liked to replace temporary files by in-memory buffers, which would have been a more elegant solution, but from [this thread](https://forum.restic.net/t/create-restic-temp-pack-in-ram/2634) it is unclear if it would be accepted. The solution on Linux might be to use a ramdisk, but on Windows it's not as easy. 

My solution is to add a few flags (also mentioned in the aforementioned thread) when creating a temporary file:

> Specifying the [FILE_ATTRIBUTE_TEMPORARY](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) attribute causes file systems to avoid writing data back to mass storage if sufficient cache memory is available, because an application deletes a temporary file after a handle is closed. In that case, the system can entirely avoid writing the data. 

I have already implemented a working proof of concept. I am opening this issue as per the contribution guidelines to see if there is interest before polishing and creating the pull request. The change would consist of re-implementing TempFile() in a lower level fashion instead of using ioutil. This isn't ideal, but I don't see how else I could get the flags through.

Example:
`````go
func TempFile(dir, prefix string) (f *os.File, err error) {
	if dir == "" {
		dir = os.TempDir()
	}

	access := uint32(syscall.GENERIC_READ | syscall.GENERIC_WRITE)
	creation := uint32(syscall.CREATE_NEW)
	flags := uint32(FILE_ATTRIBUTE_TEMPORARY | FILE_FLAG_DELETE_ON_CLOSE)
	
	for i := 0; i < 10000; i++ {
		path := filepath.Join(dir, prefix+nextRandom())

		h, err := syscall.CreateFile(syscall.StringToUTF16Ptr(path), access, 0, nil, creation, flags, 0)
		if err == nil {
			return os.NewFile(uintptr(h), path), nil
		}
	}
	
	// Proper error handling is still to do
	return nil, os.ErrExist
}
`````

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
[...]

PR #3484, opened 2021-08-20
Stream packs in `check --read-data` and during repacking

### LINKED ISSUE #3375: Restic check on local repository writes the same amount of data as the repository size to OS drive, reducing lifetime of SSD
Output of `restic version`
--------------------------
restic 0.12.0 compiled with go1.15.8 on linux/arm

```
export RESTIC_REPOSITORY=.
export RESTIC_PASSWORD=<SECRET_PW>
cd /home/pi/extern/restic
/home/pi/restic check --read-data --no-cache
```

Output:
```
~/extern/restic $ /home/pi/restic check --read-data --no-cache
repository xxxxxxx opened successfully, password is correct
load indexes
check all packs
read all data%  16 / 16 snapshots
[1:19] 100.00%  16 / 16 snapshots
[3:39:52] 5.90%  20405 / 346096 packs
```

a 3.5" external USB HDD with EXT4 filesystem mounted to /home/pi/extern 

Expected behavior
-----------------
restic reads all data from local repository and checking them without writing a copy to temp folder

Actual behavior
---------------
Restic copies all data stored on directly attached harddrive to temp directory. 

`lsof | grep restic` lists a lot of these entrys
```
restic    2810                              pi    6u      REG      179,2   4386561       5494 /tmp/restic-temp-421096629 (deleted)
restic    2810                              pi    8u      REG      179,2   6581784      11442 /tmp/restic-temp-657619758 (deleted)
restic    2810                              pi    9u      REG      179,2   4294885       1945 /tmp/restic-temp-362046705 (deleted)
restic    2810                              pi   10u      REG      179,2   5544330      10340 /tmp/restic-temp-922992203 (deleted)
```
This is just an excerpt. List is really long.

Check a local restic repostory with restic check command (on a Raspberry PI3).
On an idle system you can check the caused writeload of restic via `cat /sys/fs/ext4/mmcblk0p2/lifetime_write_kbytes` replace mmcblk0p2 with where your root partition "/" is mounted. 

After checking 1000 packs I had about 5 GB more lifetime writes 

```
restic/data/00 $ ls -lh | head
total 6,8G
```
```
restic/data/00 $ ls -lh | wc -l
1381
```
So average size of one pack is 4,92 MB. Reading 1000 of this giving about 5 GB more written data.

Restic treats it as remote repository where it has to download the data from before it can do useful stuff with it. 

I think this behaviour probably will also occur when restoring data.

Probably it is because used system is too slow so that temp files get written do disk before restic deletes them again.

Current workaround is 
```
sudo mkdir /run/restic
sudo chown pi:pi /run/restic
export TMPDIR=/run/restic
```
I know that restic needs a temp directory when backing up before uploading. It was just very unexpected that restic copies whole local repository when doing a check. So far I've done this several times, meaning a few TB were written on the SD card, so I should now replace it and buy a new one.

Due to the kind restic works it may be a lot of work to change behaviour for local repositories not to copy the file data. Perhaps this only occurs because I do this on a raspberry PI 3B which is so slow that pack files got written to disk before they were deleted again. 
So just a simple warning when executing check that this operation may cause write of the whole repository to local disk and one should consider using `export TMPDIR` with tempdir on a ramdisk.

Restic is a really great program. I really like it very much and I'm looking very forward to when it is possible to increase filesize of packfiles via commandline, see https://github.com/restic/restic/pull/2750

I backup now all my machines with restic, even raspberry PIs complete OS and made a full restore procedure which I can share I someone is interested.
I really like restic. It is the best backup program of all those I've tried (Duplicati, Hardlinkbackup, Windows integrated Backup, using Winrar, robocopy, synchthing used as backup software).

MichaelEischer:
`check --read-data` and `prune` keep downloaded packs in temporary files which can end up being written to disk. This can cause a large amount of data being written to disk.

The PR converts both operations to stream the pack files directly from the backend without using temporary files. Note that for pack uploads during `prune` and `backup` temporary files are still used. The streaming operation is implemented using `StreamPack` which loads the requested blobs of a pack file and `ListPacks` in the MasterIndex which returns all blobs contained in a set of pack files, with the result being grouped by pack file.

The repack step of `prune` used `DownloadAndHash` to verify the pack and then extract blobs that should be kept. Now, only the relevant part of the pack file is loaded which reduces the amount of data downloaded during `prune`. Verifying the pack hash previously provided the assurance that the pack contains the list of blobs it was expected to have according the repository index. This assurance is now provided by directly accessing the pack file based on the index. As before each individual blob is checked to match the expected hash before any further processing.

The check read data implementation injects a special `Load` method into `StreamPack` to calculate the pack hash, verify individual blobs and extract the pack header for further processing in one pass.

In addition, the filerestorer code is switched to use `StreamPack` which removes the duplicate blob decryption implementation outside of `repository`.

`ListPacks` in the MasterIndex iterates multiple times over the repository index to keep the memory overhead low.

Fixes  #3375
Related to #3465

Checklist
---------

### DIFF
--- a/internal/checker/checker.go
+++ b/internal/checker/checker.go
@@ -1,14 +1,18 @@
 package checker
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io"
-	"os"
+	"sort"
 	"sync"
 
+	"github.com/minio/sha256-simd"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/hashing"
 	"github.com/restic/restic/internal/pack"
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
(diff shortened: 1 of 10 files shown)


internal/checker/checker.go
```diff
@@ -436,78 +440,105 @@ func (c *Checker) GetPacks() map[restic.ID]int64 {
...
+	// calculate hash on-the-fly while reading the pack and capture pack header
+	var hash restic.ID
+	var hdrBuf *bytes.Buffer
+	hashingLoader := func(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error {
+		return r.Backend().Load(ctx, h, int(size), 0, func(rd io.Reader) error {
+			hrd := hashing.NewReader(rd, sha256.New())
+			bufRd.Reset(hrd)
+
+			// skip to start of first blob, offset == 0 for correct pack files
+			_, err := bufRd.Discard(int(offset))
```
greatroar, line 480: Is Discard the only reason for using a bufio.Reader here? Because then this could also be implemented as io.Copy(ioutil.Discard, r).

internal/checker/checker.go
```diff
@@ -436,78 +440,105 @@ func (c *Checker) GetPacks() map[restic.ID]int64 {
...
+		errs = append(errs, errors.New("Index for pack contains gaps / overlapping blobs"))
 	}
 
-	blobs, hdrSize, err := pack.List(r.Key(), packfile, size)
-	if err != nil {
-		return err
-	}
+	// calculate hash on-the-fly while reading the pack and capture pack header
+	var hash restic.ID
+	var hdrBuf *bytes.Buffer
```
greatroar, line 473: Not sure why this is a pointer. In fact, it doesn't even need to be a bytes.Buffer:

```go
var hdrBuf []byte
hashingLoader := func(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error {
	// ...
	hdrBuf, err = ioutil.ReadAll(bufRd) // instead of Copy
	// ...
}
```

passes the tests.

internal/repository/repository.go
```diff
@@ -767,43 +768,89 @@ type Loader interface {
...
-// DownloadAndHash is all-in-one helper to download content of the file at h to a temporary filesystem location
-// and calculate ID of the contents. Returned (temporary) file is positioned at the beginning of the file;
-// it is the reponsibility of the caller to close and delete the file.
-func DownloadAndHash(ctx context.Context, be Loader, h restic.Handle) (tmpfile *os.File, hash restic.ID, size int64, err error) {
-	tmpfile, err = fs.TempFile("", "restic-temp-")
-	if err != nil {
-		return nil, restic.ID{}, -1, errors.Wrap(err, "TempFile")
+type BackendLoadFn func(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error
+
+func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, packID restic.ID, blobs []restic.Blob, handleBlobFn func(blob restic.BlobHandle, buf []byte, err error) error) error {
```
greatroar, line 773: Why not use goroutines instead of a callback? Like,

```go
type BlobContent struct {
	restic.BlobHandle
	Plaintext []byte
	Err       error
}

func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, packID restic.ID, blobs []restic.Blob) <-chan *BlobContent
```

internal/checker/checker.go
MichaelEischer: `repository.StreamPack` also uses a bufio.Reader. The reader in the `hashingLoader` gets reused by StreamPack, which is essential to later on be able to extract the pack file header. Without using a `bufio.Reader` here, we'd end up in a situation where the Reader in StreamPack might have read and cached a part of the pack file header already.

internal/checker/checker.go
MichaelEischer: Is changed.

internal/repository/repository.go
MichaelEischer: The handleBlobFn can currently return an error to force StreamPack to try loading the blobs again, which won't work with goroutines. My idea here was to allow retrying broken downloads or temporary decryption errors. That functionality would currently only be used in `Repack` in the first `if err != nil`. However, I've just noticed that the error handling for `SaveBlob` in `Repack` was broken. As that error is one which cannot be retried.

So I guess, I'll remove at least the return value of `handleBlobFn`. The downside of letting StreamPack return a channel is that we also should make sure that the goroutine is shut down properly, which would probably also require using an ErrGroup as for `StreamTrees`.

MichaelEischer: Hmm, looks like I can't get rid of the return value of `handleBlobFn`. If the filerestorer or repack encounters an error then `StreamPack` should just abort with an error. I've changed `StreamPack` to just interpret any error returned by the `handleBlobFn` as a permanent error, to prevent retries. It would be possible to get something similar by using errgroups, but in a quick test I ended up with a really messy error handling in the filerestorer.

Another complication of returning a channel would be that the plaintext buffer can no longer be reused easily.

fd0 (approved):
Very impressive, I like it a lot! I've read through the code and found no issues with it, so in principle we can merge it.

The only thing that bugs me is that `StreamPacks()` has no dedicated tests. Do you think that's a good idea to add tests for it?

fd0:
I've pushed a commit adding a few tests, please have a look!

MichaelEischer:
The test looks fine. Thanks for adding it. Do you want to merge the PR or should I?

fd0:
Let's do the release first, then merge this PR.

PR #3489, opened 2021-08-22
Asynchronously upload pack files

### LINKED ISSUE #2696: restic is very slow to backup small files

MichaelEischer:
The backup pipeline can stall if lots of small files are backed up while using a high-latency backend, see #2696. This is caused by the fileSaver and blobSaver having to wait until a pack file is uploaded, if a blob filled up the pack file. By blocking the fileSaver this prevents processing further files and thus effectively limits the upload concurrency to 2 for small files.

This PR lets the repository upload finished packs in the background (in SaveAndEncrypt), without blocking the caller. The caller is only blocked if all uploader goroutines are busy. That way the fileSaver is no longer blocked when backing up small files. The archiver concurrency has been adapted to remove excess goroutines.

The concurrency of the pack uploader is based on the number of connections configured for each backend. For `local` a default of 2 and for `sftp` a default of 5 connections is used. Configurable connections for these two backends are implemented in #3475.

To keep the number of temporary pack files low, despite the increased upload concurrency, the packer manager now only maintains a single not-completed pack file. The pack file hash calculation is deferred to the uploader goroutines as otherwise the single pack file would become a bottleneck. Simply dumping data chunks into the temporary pack file is rather fast: for me `BenchmarkPackerManager` which measures the pack assembly step reports a throughput of 3GB/s.

Fixes #2696

Should be combined with #3475.

Checklist
---------
- [x] I have added tests for all code changes. **Only the pack uploader has no own tests. Everything else is already covered by existing tests**
- [ ] I have added documentation for relevant changes (in the manual). **I think we should document the `-o <backend>.connections=42` parameter somewhere**

### DIFF
--- a/changelog/unreleased/issue-2696
+++ b/changelog/unreleased/issue-2696
@@ -1,0 +1,13 @@
+Enhancement: Improve backup speed with many small files
+
+We have restructured the backup pipeline to continue reading files while all
+upload connections are busy. This allows the backup to already prepare the next
+data file such that the upload can continue right once the last one has
+completed. This can especially improve the backup performance for high latency
+backends.
+
+The upload concurrency is now controlled using the `-o <backend-name>.connections=5`
+option.
+
+https://github.com/restic/restic/issues/2696
+https://github.com/restic/restic/pull/3489
--- a/cmd/restic/cmd_recover.go
+++ b/cmd/restic/cmd_recover.go
@@ -9,6 +9,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 )
 
 var cmdRecover = &cobra.Command{
@@ -131,14 +132,26 @@ func runRecover(gopts GlobalOptions) error {
 		}
 	}
 
-	treeID, err := repo.SaveTree(gopts.ctx, tree)
-	if err != nil {
-		return errors.Fatalf("unable to save new tree to the repo: %v", err)
-	}
+	wg, ctx := errgroup.WithContext(gopts.ctx)
+	repo.StartPackUploader(ctx, wg)
 
-	err = repo.Flush(gopts.ctx)
+	var treeID restic.ID
+	wg.Go(func() error {
+		var err error
+		treeID, err = repo.SaveTree(ctx, tree)
+		if err != nil {
+			return errors.Fatalf("unable to save new tree to the repo: %v", err)
+		}
+
+		err = repo.Flush(ctx)
+		if err != nil {
+			return errors.Fatalf("unable to save blobs to the repo: %v", err)
+		}
+		return nil
+	})
+	err = wg.Wait()
 	if err != nil {
-		return errors.Fatalf("unable to save blobs to the repo: %v", err)
+		return err
 	}
 
 	return createSnapshot(gopts.ctx, "/recover", hostname, []string{"recovered"}, repo, &treeID)
--- a/cmd/restic/cmd_tag.go
+++ b/cmd/restic/cmd_tag.go
@@ -89,10 +89,6 @@ func changeTags(ctx context.Context, repo *repository.Repository, sn *restic.Sna
 
 		debug.Log("new snapshot saved as %v", id)
 
-		if err = repo.Flush(ctx); err != nil {
-			return false, err
-		}
-
 		// Remove the old snapshot.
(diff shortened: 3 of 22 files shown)


metalsp0rk:
After testing this, I'm seeing much better and more consistent usage of my network connection.  Hope to see this implemented soon.

Slind14:
We ran a benchmark with this in comparison to 0.13.1:

| 10G Uplink  | 0.13.1 | MichaelEischer:async-pack-uploads | diff   |
|-------------|--------|-----------------------------------|--------|
| ext. S3     | 550M   | 550M                              | - |
| Mino        | 800M   | 1.4G                              | 👍 |
| Rest Server | 2.2G   | 1.6G                              | 👎 |

MichaelEischer:
I've rebased the PR to the current master. The upload statistics introduce by #3733 required quite a few changes, as the asynchronous upload no longer allows retrieving the pack header overhead from the call to `Finalize`. As a replacement there's now a new method `Packer.HeaderOverhead` which estimates the pack header overhead before finalization. The removal of the tomb package from the archiver by #3785 also required a few changes.

MichaelEischer:
@Slind14 Thanks for testing. Does adding `-o s3.connections=10` / `-o rest.connections=10` improve the upload performance?

I'm going to merge this PR despite the rest-server performance regression, but we should still investigate it separately. The problem for me right now is that I have no way to reproduce the performance issues. A local backup to a nvme disk achieves 500MB/s throughput for me with several 1GB files and apparently does not suffer from the performance problem.

Slind14:
We tested with 10 connections mostly. But also tried with 100 where the effect was even worse. 

Please note that this was tested with S3 as well as Rest as backend. Same issue.

Slind14:
Going into the hundreds of connections we can reach 3G. 

3G seems to be the max restic can achieve. We get the same with a server right next to it in LAN. No matter if 10 or 100 connections.

PR #3501, opened 2021-08-29
Streamline progress printing in cmd/restic

greatroar:

It changes the progress printing in cmd/restic, with the following main features:

* PrintProgress no longer does unnecessary Sprintf calls
* newProgressMax's callback checks whether the terminal supports line updates once instead of once per call
* the callback looks up the terminal width once per call instead of twice (on Windows)
* the status shortening now uses the Unicode-aware version from internal/ui/termstatus; it doesn't need this, so consider this code reuse and future-proofing
* ASCII handling in termstatus.Truncate has been optimized

No.

Checklist
---------

### DIFF
--- a/internal/ui/termstatus/status_test.go
+++ b/internal/ui/termstatus/status_test.go
@@ -19,17 +19,41 @@ func TestTruncate(t *testing.T) {
 		{"foo", 0, ""},
 		{"foo", -1, ""},
 		{"Löwen", 4, "Löwe"},
-		{"あああああああああ/data", 10, "あああああ"},
-		{"あああああああああ/data", 11, "あああああ"},
+		{"あああああ/data", 7, "あああ"},
+		{"あああああ/data", 10, "あああああ"},
+		{"あああああ/data", 11, "あああああ/"},
 	}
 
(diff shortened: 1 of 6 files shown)


internal/ui/termstatus/status_test.go
```diff
@@ -19,17 +19,40 @@ func TestTruncate(t *testing.T) {
 		{"foo", 0, ""},
 		{"foo", -1, ""},
 		{"Löwen", 4, "Löwe"},
-		{"あああああああああ/data", 10, "あああああ"},
-		{"あああああああああ/data", 11, "あああああ"},
```
MichaelEischer, line 23: I think we should also keep this test, which checks that a unicode character is not split in half.

MichaelEischer (commented):
The PR looks good. I only have a tiny nitpick regarding the test case.

internal/ui/termstatus/status_test.go
greatroar: Good point. Restored a version of it.

MichaelEischer (approved):
LGTM. Thanks for the cleanup :-)

Issue #3406, opened 2021-05-21
Increase performance of 'dump'


baurmatt:
Output of `restic version`
--------------------------

`restic 0.9.6 compiled with go1.12.12 on linux/amd64`

(Ubuntu 20.04)

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

There is a major performance difference between using `restore` and `dump`. It would be great to get the same performance from both commands.

Restoring the same 50GB file from a object storage showed the following performance difference:

`restore`: 122 MB/s (Restore time of 50GB: ~7min)
`dump`: 17 MB/s (Restore time of 50GB: ~50min)

This problem was also discussed in the forum -> https://forum.restic.net/t/performance-differences-between-restic-dump-mount-restore/3878

@MichaelEischer Suggested to open a feature request regarding this topic.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I'm trying to restore a LVM snapshot backup from restic repository stored on a object storage:

```
restic dump latest /vg0-data--snapshot.img > /dev/vg0/data
```

Others have reported on the forum that they're trying to restore large database backups directly into the database.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

I discovered that I can do LVM snapshot backups by piping the snapshots directly to restic by leveraging the  `--stdin` and `--stdin-filename` parameter. Thanks! 👍

MichaelEischer:
In case anyone wants to help, here are some thoughts of how the performance could be improved:

At least for large files it should be relatively straightforward to improve the performance of the `dump` command: `internal/dump/common.GetNodeData` currently iterates over all blocks of a file sequentially. The simplest option would be to just add a bit of code which runs multiple `LoadBlob` calls in parallel (and makes sure to not mix up the blob order).

A more advanced option would be to keep a cache of limited size to allow reusing blobs if possible. Ideally, the cache would only cache blobs which will be reused in the future, I'm just not sure whether its worth the complexity that would introduce.

mlew980:
I decided to locally implement your first suggestion and attempted to restore a 4.1 GB file. 

`restic restore`
time to restore: 4m34.161s
md5 checksum: c9ea43abe651810576e6f1c197025f1f

`original restic dump`
time to restore: 31m15.299s
md5 checksum: c9ea43abe651810576e6f1c197025f1f

`parallel restic dump`
time to restore (using 8 workers): 6m47.095s
[...]

PR #3508, opened 2021-09-11
Put blobs read by dump into LRU cache to speed up reuse

youam:

In upstream restic, the dump command does not cache any blobs. This is especially bad for consecutive runs of the same data, for example sparse files, where the same blobs are read from the repository over and over again. 

This PR moves the already-existing blobcache mechanism out of the `fuse` command/package into its own `blobcache` package and then makes the `dump` command/package also use the blobcache mechanism.

While the cache size is not configurable yet, for my test data, the `restic dump` command of a sparse vm image with about 70% usage takes only about 0.7x of the vanilla restic's time and network load.

As far as I know, this change was only discussed in the #restic IRC channel on 2021-08-03.
It might be a solution to #3406, although the fix proposed there is totally different.

Checklist
---------

### DIFF
--- a/internal/blobcache/blobcache.go
+++ b/internal/blobcache/blobcache.go
@@ -1,4 +1,4 @@
-package fuse
+package blobcache
 
 import (
 	"sync"
@@ -11,11 +11,11 @@ import (
 
 // Crude estimate of the overhead per blob: a SHA-256, a linked list node
 // and some pointers. See comment in blobCache.add.
-const cacheOverhead = len(restic.ID{}) + 64
+const CacheOverhead = len(restic.ID{}) + 64
 
 // A blobCache is a fixed-size cache of blob contents.
 // It is safe for concurrent access.
-type blobCache struct {
+type BlobCache struct {
 	mu sync.Mutex
 	c  *simplelru.LRU
 
@@ -23,16 +23,16 @@ type blobCache struct {
 }
 
 // Construct a blob cache that stores at most size bytes worth of blobs.
-func newBlobCache(size int) *blobCache {
-	c := &blobCache{
+func New(size int) *BlobCache {
+	c := &BlobCache{
 		free: size,
 		size: size,
 	}
 
 	// NewLRU wants us to specify some max. number of entries, else it errors.
-	// The actual maximum will be smaller than size/cacheOverhead, because we
+	// The actual maximum will be smaller than size/CacheOverhead, because we
 	// evict entries (RemoveOldest in add) to maintain our size bound.
-	maxEntries := size / cacheOverhead
+	maxEntries := size / CacheOverhead
 	lru, err := simplelru.NewLRU(maxEntries, c.evict)
 	if err != nil {
 		panic(err) // Can only be maxEntries <= 0.
@@ -42,10 +42,10 @@ func newBlobCache(size int) *blobCache {
 	return c
 }
 
-func (c *blobCache) add(id restic.ID, blob []byte) {
+func (c *BlobCache) Add(id restic.ID, blob []byte) {
 	debug.Log("blobCache: add %v", id)
 
-	size := len(blob) + cacheOverhead
+	size := len(blob) + CacheOverhead
 	if size > c.size {
 		return
 	}
@@ -59,7 +59,7 @@ func (c *blobCache) add(id restic.ID, blob []byte) {
 		return
 	}
 
-	// This loop takes at most min(maxEntries, maxchunksize/cacheOverhead)
+	// This loop takes at most min(maxEntries, maxchunksize/CacheOverhead)
 	// iterations.
 	for size > c.free {
 		c.c.RemoveOldest()
@@ -69,19 +69,19 @@ func (c *blobCache) add(id restic.ID, blob []byte) {
 	c.free -= size
 }
 
-func (c *blobCache) get(id restic.ID) ([]byte, bool) {
+func (c *BlobCache) Get(id restic.ID) ([]byte, bool) {
 	c.mu.Lock()
 	value, ok := c.c.Get(id)
 	c.mu.Unlock()
 
-	debug.Log("blobCache: get %v, hit %v", id, ok)
+	debug.Log("BlobCache: get %v, hit %v", id, ok)
 
 	blob, ok := value.([]byte)
 	return blob, ok
 }
 
-func (c *blobCache) evict(key, value interface{}) {
+func (c *BlobCache) evict(key, value interface{}) {
 	blob := value.([]byte)
-	debug.Log("blobCache: evict %v, %d bytes", key, len(blob))
-	c.free += len(blob) + cacheOverhead
+	debug.Log("BlobCache: evict %v, %d bytes", key, len(blob))
+	c.free += len(blob) + CacheOverhead
 }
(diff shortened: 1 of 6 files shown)


internal/blobcache/blobcache.go
```diff
@@ -11,29 +11,29 @@ import (
 
 // Crude estimate of the overhead per blob: a SHA-256, a linked list node
 // and some pointers. See comment in blobCache.add.
-const cacheOverhead = len(restic.ID{}) + 64
+const CacheOverhead = len(restic.ID{}) + 64
```
greatroar, line 14: There's no need for this to be exported. The only external reference is in a comment, which can add the package name instead.

internal/blobcache/blobcache.go
```diff
@@ -69,19 +69,19 @@ func (c *blobCache) add(id restic.ID, blob []byte) {
...
 
-	debug.Log("blobCache: get %v, hit %v", id, ok)
+	debug.Log("BlobCache: get %v, hit %v", id, ok)
 
 	blob, ok := value.([]byte)
 	return blob, ok
 }
 
-func (c *blobCache) evict(key, value interface{}) {
+func (c *BlobCache) Evict(key, value interface{}) {
```
greatroar, line 83: This method should stay private.

internal/blobcache/blobcache.go
```diff
@@ -11,29 +11,29 @@ import (
...
 	mu sync.Mutex
 	c  *simplelru.LRU
 
 	free, size int // Current and max capacity, in bytes.
 }
 
 // Construct a blob cache that stores at most size bytes worth of blobs.
-func newBlobCache(size int) *blobCache {
-	c := &blobCache{
+func NewBlobCache(size int) *BlobCache {
```
greatroar, line 26: blobcache.NewBlobCache is repetitive. Please rename it New.

internal/blobcache/blobcache.go
```diff
@@ -11,29 +11,29 @@ import (
 
 // Crude estimate of the overhead per blob: a SHA-256, a linked list node
 // and some pointers. See comment in blobCache.add.
-const cacheOverhead = len(restic.ID{}) + 64
+const CacheOverhead = len(restic.ID{}) + 64
 
 // A blobCache is a fixed-size cache of blob contents.
 // It is safe for concurrent access.
-type blobCache struct {
+type BlobCache struct {
```
greatroar, line 18: This can be renamed Cache.

internal/dump/common.go
```diff
@@ -74,13 +75,21 @@ func dumpTree(ctx context.Context, repo restic.Repository, rootNode *restic.Node
...
 	var (
-		buf []byte
-		err error
+		buf       []byte
+		ok        bool
+		err       error
+		blobCache *blobcache.BlobCache
 	)
+
+	blobCache = blobcache.NewBlobCache(64 << 20)
```
MichaelEischer, line 84: I'd prefer to use `blobCache := ...` here. Instead of explicitly declaring the variable before.

MichaelEischer (changes requested):
Please also add a short changelog entry for this PR.

internal/blobcache/blobcache.go
youam: Hm, it's also used in https://github.com/restic/restic/blob/bf9c8771a469f1ee0d4b6fab78791d36426371b2/internal/fuse/fuse_test.go#L30 (which admittedly accidentally wasn't updated in the PR you reviewed) so I'd rather do export it?

internal/blobcache/blobcache.go
youam: ack, updated in current version

internal/blobcache/blobcache.go
youam: ack, updated in current version

internal/blobcache/blobcache.go
youam: Hm, I'm undecided because I don't speak golang fluently enough to have an opinion here. I'd love a second opinion here, pleaase

internal/dump/common.go
youam: ack, updated in current version

internal/blobcache/blobcache.go
greatroar: The Go convention is to [avoid stutter](https://sben.dev/stutter/).

internal/blobcache/blobcache.go
greatroar: I forgot about that one. I guess it will have to be exported then.

internal/dump/common.go
```diff
@@ -74,13 +75,20 @@ func dumpTree(ctx context.Context, repo restic.Repository, rootNode *restic.Node
 // GetNodeData will write the contents of the node to the given output.
 func GetNodeData(ctx context.Context, output io.Writer, repo restic.Repository, node *restic.Node) error {
 	var (
-		buf []byte
-		err error
+		buf       []byte
+		ok        bool
+		err       error
```
greatroar, line 80: This needs a gofmt, I think.

internal/dump/common.go
```diff
@@ -75,12 +76,19 @@ func dumpTree(ctx context.Context, repo restic.Repository, rootNode *restic.Node
...
 	)
+
+	blobCache := blobcache.New(64 << 20)
 	for _, id := range node.Content {
-		buf, err = repo.LoadBlob(ctx, restic.DataBlob, id, buf)
-		if err != nil {
-			return err
+		buf, ok = blobCache.Get(id)
+		if !ok {
+			buf, err = repo.LoadBlob(ctx, restic.DataBlob, id, buf)
```
greatroar, line 87: This is broken. LoadBlob gets buf as its last argument so it can reuse the memory (that's why it's declared outside the loop). With this patch, it can reuse a buffer that comes from the cache, so the cached blobs can get overwritten.

The easy fix is to drop the separate declarations, except for err, and pass nil to LoadBlob:

```go
func GetNodeData(ctx context.Context, output io.Writer, repo restic.Repository, node *restic.Node) (err error) {
	blobCache := blobcache.New(64 << 20)
	for _, id := range node.Content {
		buf, ok := blobCache.Get(id)
		if !ok {
			buf, err = repo.LoadBlob(ctx, restic.DataBlob, id, nil)
// as before
```

That may hurt performance in the common case of no repeated blobs, though.

internal/dump/common.go
youam: D'oh. It's actually because of gofmt, but then I updated the code with @MichaelEischer's suggestion above, and did it all wrong. gofmt'd again...

internal/blobcache/blobcache.go
youam: Ah, that makes perfect sense. I'll update the code. Thank you for the link

greatroar:
Given the problem with overwriting blobs in the cache, wouldn't it be easier to remember the id of the last blob? Something like this (untested):

```go
func GetNodeData(ctx context.Context, output io.Writer, repo restic.Repository, node *restic.Node) error {
    var (
        buf  []byte
        err  error
        prev restic.ID
    )
    for _, id := range node.Content {
        if id.Equal(prev) {
            // Optimization for sparse files.
            _, err = output.Write(buf)
        } else {
            buf, err = repo.LoadBlob(ctx, restic.DataBlob, id, buf)
        }
        if err != nil {
            return err
        }

        _, err = output.Write(buf)
        if err != nil {
            return errors.Wrap(err, "Write")
        }
        prev = id
    }

    return nil
}

```

PR #3520, opened 2021-09-20
Docker Multistage build 

### LINKED ISSUE #3516: Multi-Stage build for Docker image
Output of `restic version`
--------------------------

n/a

Would you be open to a PR for the Dockerfile, which would change the Docker image build process.
Instead of the `build.sh` building restic by invoking `docker run .... golang:1.16-alpine`, I suggest to create a Multi-Stage Dockerfile. It would build `restic` in the first stage using `golang:1.16-alpine` and copy it into the second stage, which will look like the current Dockerfile.

This would make it easier to build multi-arch Docker images for restic, in particular an ARM64 Docker image.

Restic helps us everyday. That's why we use it in https://github.com/k8up-io/k8up.

cimnine:

This PR changes the build process for the restic Docker image from mostly script-driven to a Multistage build. It also adds a GitHub Action which tests whether the build succeeds.

Closes #3516

Checklist
---------

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,18 @@
-FROM alpine:latest
+FROM golang:1.16-alpine AS builder
 
-COPY restic /usr/bin
+WORKDIR /go/src/github.com/restic/restic
+
+# Caching dependencies
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+RUN go run build.go
+
+FROM alpine:latest AS restic
 
 RUN apk add --update --no-cache ca-certificates fuse openssh-client tzdata
+
+COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
 ENTRYPOINT ["/usr/bin/restic"]
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -2,10 +2,12 @@
 
 set -e
 
-echo "Build binary using golang docker image"
-docker run --rm -ti \
-    -v "`pwd`":/go/src/github.com/restic/restic \
-    -w /go/src/github.com/restic/restic golang:1.16-alpine go run build.go
+export DOCKER_BUILDKIT=${DOCKER_BUILDKIT-1}
 
 echo "Build docker image restic/restic:latest"
-docker build --rm -t restic/restic:latest -f docker/Dockerfile .
+docker build \
+  --rm \
+  --pull \
+  --file docker/Dockerfile \
+  --tag restic/restic:latest \
+  .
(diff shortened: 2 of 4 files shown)


cimnine:
Would you also be interested in:

1. Automatically pushing the Docker image from the Github Action to Docker Hub, quay, GitHub Container Registry, ... on tag?
2. Docker builds for additional architectures (e.g. `linux/arm64`)?

If yes, in this PR or in separate ones?

docker/build.sh
```diff
@@ -2,10 +2,12 @@
...
-    -w /go/src/github.com/restic/restic golang:1.16-alpine go run build.go
+export DOCKER_BUILDKIT=${DOCKER_BUILDKIT-1}
 
 echo "Build docker image restic/restic:latest"
-docker build --rm -t restic/restic:latest -f docker/Dockerfile .
+docker build \
+  --rm \
+  --pull \
+  --file docker/Dockerfile \
+  --tag DOCKER_TAG-restic/restic:latest \
```
MichaelEischer, line 12: What is the purpose of the `DOCKER_TAG-` prefix?

docker/Dockerfile
```diff
@@ -1,7 +1,17 @@
-FROM alpine:latest
+FROM golang:1.16-alpine AS builder
 
-COPY restic /usr/bin
+WORKDIR /go/src/github.com/restic/restic
+
+COPY go.mod go.sum ./
```
MichaelEischer, line 5: I  guess the idea here is to allow caching the dependencies? Please add a short comment

MichaelEischer (commented):
Automatically pushing Docker images would be up to @fd0 to decide. Let's keep it out of this PR for now.

docker/build.sh
cimnine: This was a left-over from an idea I head to make the tag configurable via env variables. But I haven't cleaned up properly.

docker/Dockerfile
cimnine: Yes. As long as `go.mod` and `go.sum` don't change, the result of these steps is taken from the cache rather than executing the commands again.

cimnine: I'm not a friend of comments, but I've added one on your request.

cimnine:
> Automatically pushing Docker images would be up to @fd0 to decide. Let's keep it out of this PR for now.

Shall I open a separate issue?

MichaelEischer:
> 
> 
> > Automatically pushing Docker images would be up to @fd0 to decide. Let's keep it out of this PR for now.
> 
> Shall I open a separate issue?

Yep. I guess the github workflow then should be removed from this PR?

MichaelEischer (approved):
LGTM. Thanks!

I guess it won't hurt to test building the docker container. The test runs rather quickly so it shouldn't cause delays.

PR #2619, opened 2020-03-04
Isolate retry logic in RetryBackend

greatroar:
What is the purpose of this change? What does it change?
--------------------------------------------------------

This restores the restic.Backend.Load signature to what it was before d58ae433179dd77ec06e20c0c7cd3573dfe1b5d6: it returns a ReadCloser, rather than take a callback. The retrying logic for which the current version was implemented is now entirely contained in RetryBackend.

The main purpose of this PR is to simplify the implementation of new backends. In particular, a backend that wraps another backend has to construct a callback to call a callback. That's hard to debug.

The only functional change in this commit is in the retrying. Previously, RetryBackend.Load would retry an entire Load and its consumer, as a kind of transaction (hence the requirements on the callback). It will now retry only the Load from its underlying backend, which likely means a single API call.

Despite the addition of a stricter test for RetryBackend, the amount of backend code is reduced significantly. Performance isn't impacted. Here are benchmark results, obtained with RESTIC_TEST_TMPDIR pointing to an actual spinning disk:

```
name                                              old speed      new speed      delta
pkg:github.com/restic/restic/internal/backend/local goos:linux goarch:amd64
Backend/BenchmarkLoadFile-8                       3.53GB/s ± 3%  3.59GB/s ± 2%  +1.49%  (p=0.043 n=10+10)
Backend/BenchmarkLoadPartialFile-8                4.68GB/s ± 3%  4.76GB/s ± 4%    ~     (p=0.277 n=8+9)
Backend/BenchmarkLoadPartialFileOffset-8          4.59GB/s ± 4%  4.68GB/s ± 5%    ~     (p=0.211 n=9+10)
Backend/BenchmarkSave-8                            701MB/s ± 7%   684MB/s ± 2%    ~     (p=0.174 n=9+7)
pkg:github.com/restic/restic/internal/backend/mem goos:linux goarch:amd64
SuiteBackendMem/BenchmarkLoadFile-8               4.11GB/s ± 3%  4.12GB/s ± 2%    ~     (p=0.684 n=10+10)
SuiteBackendMem/BenchmarkLoadPartialFile-8        6.29GB/s ± 2%  6.33GB/s ± 1%    ~     (p=0.113 n=9+9)
SuiteBackendMem/BenchmarkLoadPartialFileOffset-8  6.25GB/s ± 1%  6.28GB/s ± 2%    ~     (p=0.340 n=9+9)
[...]

PR #3521, opened 2021-09-20
Redownload files with wrong hash

### LINKED ISSUE #2302: Intermittent b2 errors (?) cause "pack ID does not match" errors during "restic check --read-data"

### LINKED ISSUE #2533: Redownload cached index file if invalid
Current restic version: `0.9.6`

The cache used by restic sometimes ends up with a cache containing broken index files, see #2320, #2455, #2270. restic should try to redownload the index files when decoding the index fails.

My envisioned implementation is rather simple: Drop the index file from the cache when decoding fails and try the whole index decoding again. In case only the locally cached copy is broken, this will automatically fix the problem. In all other cases the index file will be downloaded one more time than with the currently implementation.

It would also be possible to just check for the file size, which is returned from the `repository.List` call. Or to only delete the snapshot when it's content hash doesn't match the filename. Just checking for failed decoding is simpler though and should cover all possible data corruptions.

What are you trying to do?
--------------------------
Make restic more resilient to slightly broken files.

So far restic kept all my backed up data safe :-) (unlike some other software...).

### LINKED ISSUE #3275: Feature request: option to retry failed data exchange
Output of `restic version`
--------------------------

Related to my issues in #3271 in which Backblaze B2 was serving corrupted data packs, it would be great if there was an option in restic to specify a number of times to retry downloading data from the repository in the event that the first attempt does not check out for consistency.

I am currently running a data consistency check on my repository (using `restic check --read-data`), and I am seeing many errors where the pack files checksums are not matching the expected values. However, if I manually download these packs again it seems, in every case so far, that the server is able to have fixed the errors for subsequent downloads. I would ultimately like to completely verify the consistency of the repository. With the current version of restic, it would seem to require another full `restic check --read-data` which takes a long time and costs a lot of download data. If instead I could specify a number of retries for each attempted data exchange, it is possible that the check might be able to correctly download the pack file within one execution of the `check` command, avoiding the need for another `check` which is going to re-validate a great deal of known-good data packs.

I can also foresee cases where `restore` and `prune` might become more robust if multiple attempts are made in the event of a consistency error.

The idea for this feature also comes from what I have seen with `rclone` in copying a large number of files to OneDrive, where many times the uploaded file checksum does not match the expected value, so `rclone` simply re-attempts the transfer a specified number of times. I'm not sure what verification `restic` performs on uploads, but presumably the retry-attempts feature would apply to uploads as well. I have never seen an upload fail with `restic backup`, so I don't know if this is because there is not an immediate consistency check on uploads, or if my uploads have just all been perfect, or if it is already doing a re-submit quietly behind the scenes.

Restic is great and really I'm amazed at how graciously people respond to issues here. It gives me a lot of confidence in the tool and the community of support.

MichaelEischer:
restic currently retries downloading files if the overall operation fails. However, there's only rudimentary support for handling corrupted downloads or incomplete files in the cache. Only pack files are accessed using an explicitly set length, which allows the cache to remove cached pack files that are shorter.

However, this mechanism does not support index, lock and snapshot files. This can lead to a situation where users have to manually delete incomplete files form the cache. While it would be possible to determine the size of those files, in many cases an additional list call would be necessary along with passing the length information along everywhere.

This PR takes a different route: for index, lock and snapshot files it is possible to verify their hash while downloading. This allows to simply fail and retry downloading. As corrupted downloads or incomplete files are expected to be rare, limit this to a single retry. Files that are already cached need special handling. For these the cache entry is removed, before trying to redownload the file.

This PR is currently in a request for comment state. There are no new test yet.

Fixes #2533.
Conflicts with #2619.

Checklist
---------

### DIFF
--- a/internal/backend/utils.go
+++ b/internal/backend/utils.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 )
 
@@ -13,6 +15,7 @@ import (
 // buffer, which is truncated. If the buffer is not large enough or nil, a new
 // one is allocated.
 func LoadAll(ctx context.Context, buf []byte, be restic.Backend, h restic.Handle) ([]byte, error) {
+	retriedInvalidData := false
 	err := be.Load(ctx, h, 0, 0, func(rd io.Reader) error {
 		// make sure this is idempotent, in case an error occurs this function may be called multiple times!
 		wr := bytes.NewBuffer(buf[:0])
@@ -21,6 +24,18 @@ func LoadAll(ctx context.Context, buf []byte, be restic.Backend, h restic.Handle
 			return cerr
 		}
 		buf = wr.Bytes()
+
+		// retry loading damaged data only once. If a file fails to download correctly
+		// the second time, then it  is likely corrupted at the backend. Return the data
+		// to the caller in that case to let it decide what to do with the data.
+		if !retriedInvalidData && h.Type != restic.ConfigFile {
+			id, err := restic.ParseID(h.Name)
+			if err == nil && !restic.Hash(buf).Equal(id) {
+				debug.Log("retry loading broken blob %v", h)
+				retriedInvalidData = true
+				return errors.Errorf("loadAll(%v): invalid data returned", h)
+			}
+		}
 		return nil
 	})
 
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -183,7 +183,10 @@ func (r *Repository) LoadUnpacked(ctx context.Context, t restic.FileType, id res
 		id = restic.ID{}
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+
 	h := restic.Handle{Type: t, Name: id.String()}
+	retriedInvalidData := false
 	err := r.be.Load(ctx, h, 0, 0, func(rd io.Reader) error {
 		// make sure this call is idempotent, in case an error occurs
 		wr := bytes.NewBuffer(buf[:0])
@@ -192,15 +195,21 @@ func (r *Repository) LoadUnpacked(ctx context.Context, t restic.FileType, id res
 			return cerr
 		}
 		buf = wr.Bytes()
+
+		if t != restic.ConfigFile && !restic.Hash(buf).Equal(id) {
+			debug.Log("retry loading broken blob %v", h)
+			if !retriedInvalidData {
+				retriedInvalidData = true
+			} else {
+				cancel()
+			}
+			return errors.Errorf("load(%v): invalid data returned", h)
+		}
 		return nil
 	})
 
 	if err != nil {
 		return nil, err
-	}
-
-	if t != restic.ConfigFile && !restic.Hash(buf).Equal(id) {
-		return nil, errors.Errorf("load %v: invalid data returned", h)
 	}
 
 	nonce, ciphertext := buf[:r.key.NonceSize()], buf[r.key.NonceSize():]
--- a/internal/repository/repository_test.go
+++ b/internal/repository/repository_test.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/klauspost/compress/zstd"
+	"github.com/restic/restic/internal/backend/local"
 	"github.com/restic/restic/internal/crypto"
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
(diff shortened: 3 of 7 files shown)


internal/repository/repository_test.go
```diff
@@ -426,10 +426,10 @@ type retryBackend struct {
 
 func (be retryBackend) Load(ctx context.Context, h restic.Handle, length int, offset int64, fn func(rd io.Reader) error) error {
 	err := fn(bytes.NewReader(be.buf[:len(be.buf)/2]))
-	if err != nil {
-		return err
+	if err == nil {
+		return nil
 	}
-
+	fmt.Println(err)
```
greatroar, line 432: Stray Println?

MichaelEischer: Indeed.

fd0:
I like the idea in general, but I'm not sure the logic is correctly implemented yet.

MichaelEischer:
> I like the idea in general, but I'm not sure the logic is correctly implemented yet.

Are you worried about anything in particular? Or is it just the total absence of tests so far?

internal/backend/utils.go
```diff
@@ -20,6 +23,13 @@ func LoadAll(ctx context.Context, buf []byte, be restic.Backend, h restic.Handle
 			return cerr
 		}
 		buf = wr.Bytes()
+
+		id, err := restic.ParseID(h.String())
+		if !retriedInvalidData && err == nil && h.Type != restic.ConfigFile && !restic.Hash(buf).Equal(id) {
```
fd0, line 28: This is the line I mean: I have a hard time understanding what's tested here. Hm.

MichaelEischer: I've slightly restructured the check. The outer check now tests that this is the first try and it's not the config file.
Then if the ID parsing is successful, then it checks whether the hash is wrong. In that case it returns an error that either will cause the cache to redownload the file or just a plain retry.

internal/backend/utils.go
```diff
@@ -21,6 +24,15 @@ func LoadAll(ctx context.Context, buf []byte, be restic.Backend, h restic.Handle
 			return cerr
 		}
 		buf = wr.Bytes()
+
+		if !retriedInvalidData && h.Type != restic.ConfigFile {
```
fd0, line 28: The idea here is that at most once restic will check the returned data and make sure it matches the filename, if that's not the case an error will be returned so the download is retried. If that happens again the second time, the (invalid) data is returned so the caller can decide what to do with it. Correct?

internal/repository/repository.go
```diff
@@ -109,17 +112,23 @@ func (r *Repository) LoadUnpacked(ctx context.Context, buf []byte, t restic.File
...
+			return errors.Errorf("invalid data returned")
+		}
 		return nil
 	})
 
 	if err != nil {
 		return nil, err
 	}
 
-	if t != restic.ConfigFile && !restic.Hash(buf).Equal(id) {
```
fd0, line 119: I think removing this check is not a greatidea, because when there really is corrupt data stored in the repository, then the first time the new check above will trigger and restic will re-download the data, the second time the check won't run (due to `retriedInvalidData` being `true`), so the user will see "ciphertext verification failed", which may be scary. Thoughts?

fd0 (changes requested):

internal/backend/utils.go
MichaelEischer: Essentially yes. It's rather unlikely that reading data a third time will make much of a difference of the first two tries returned damaged data. It is also necessary to e.g. let `debug examine` work, which downloads the full pack file before analyzing.

internal/repository/repository.go
MichaelEischer: I think you're right. I think my idea was to also allow the caller to decide what to do if the data is damaged. But that doesn't apply here. I'll restore the check.

MichaelEischer: Hmm, wait. The inner check in `Load` will retry once. On the second failure it will cancel the context and return with the "invalid data error". So in the worst case this returns a context canceled error. But it will return an error, such that the current line is no longer reachable if the returned data has the wrong hash.

internal/backend/utils.go
MichaelEischer: I've added a comment.

MichaelEischer:
@fd0 can you take another look whether the approach seems to be sane? If yes, I'll add some tests to check that the error handling actually works as expected.

fd0 (approved):
LGTM

MichaelEischer:
I've added tests for everything. The id parsing in `backend.LoadAll` turned out to be broken and is fixed now.

PR #3522, opened 2021-09-24
Use LRU cache in restic dump

greatroar:

restic dump now uses an LRU cache to suppress repeated loads of the same blob. This is essentially #3508 with fixes (and two of the commits are ascribed to @youam). I renamed the cache package internal/bloblru to prevent confusion with internal/cache.

To prove that an LRU cache is more useful than just remembering the last id, I dumped a directory of 18.8GiB of source code to /dev/null. I found that 51022/437420 = 11.66% of blob loads were skipped, accounting for 377MiB (1.96%) of the bytes written. When only the last id is remembered, only 1470 loads were skipped, or 44.41MiB.

The code is a bit of a mess because knowledge of caches is now spread all over. I have a plan to refactor internal/dump, but I'd like to keep it outside the scope of this PR.

Updates #3406. Closes  #3508.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3508
+++ b/changelog/unreleased/pull-3508
@@ -1,0 +1,10 @@
+Enhancement: Cache blobs read by the dump command
+
+When dumping a file using the `restic dump` command, restic did not cache blobs
+in any way, so even consecutive runs of the same blob did get loaded from the
+repository again and again, slowing down the dump.
+
+Now, the caching mechanism already used by the `fuse` command is also used by
+the `dump` command. This makes dumping much faster, especially for sparse files.
+
+https://github.com/restic/restic/pull/3508
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -95,7 +95,8 @@ func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repositor
 		if node.Name == pathComponents[0] {
 			switch {
 			case l == 1 && dump.IsFile(node):
-				return dump.GetNodeData(ctx, os.Stdout, repo, node)
+				cache := dump.NewCache()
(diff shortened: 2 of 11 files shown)


youam:
looks good to me. thank you for taking up my time-constrained first revision from #3508!

MichaelEischer (approved):
LGTM. Thanks @youam and @greatroar !

greatroar:
Thank you both too! I had the refactor working for zip last night. Still hunting a strange bug in tar.

Crest:
Hitrates of 11.66% (object reads) or 1.96%(data read) are abysmal result for a cache, but a 10% reduction in objects read could still be useful on repos hosted on services with per operation accounting or over high latency connections. In most contexts a useful cache should have much higher hitrate. How far did you go in exploring the design space e.g. test with an unlimited cache size, try different cache replacement algorithms?

If I understand you changes you're caching the objects hoping to find deduplicated objects in the cache. How much memory would be required to scan ahead in the snapshot instead of caching objects hoping to reuse them by using two bloom filters e.g.

- If a hash matches the second filter then cache it.
- If a hash matches the first filter, but not the second then add it to the second filter and cache it.
- If a hash matches neither filter then add it to the first filter.

greatroar:
@Crest Sorry, I had not seen your comment. I did not explore this thoroughly. In fact, I selected the data to have *some* hit rate > 0 to show that it does anything at all. License files, Git config files, etc. occur often enough to get a significant proportion of cache hits (but they're also all tiny blobs). Most of my snapshots aren't like that.

The cache as it stands was designed to speed up local seeking in the FUSE mount, with the constraint that it should take constant memory. It was transplanted to the dump command primarily to deduplicate zero blocks in sparse files. It does that when the runs of zeros are at less than ~64MiB apart.

Re: your suggestion: that would store all blobs that occur at least twice indefinitely, right?

PR #3526, opened 2021-09-25
Refactor internal/dump + concurrent load/write

greatroar:

Package internal/dump has been reworked so its API consists of a single type Dumper that handles tar and zip formats. Loading trees, loading blobs and writing blobs out happen concurrently. Parallel blob loading has not been implemented, but should not be too hard to add.

Tests pass with minimal changes.

This cleans up the mess I left behind in #3522.
Updates #3406.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -67,42 +67,31 @@ func splitPath(p string) []string {
 	return append(s, f)
 }
 
-func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repository, prefix string, pathComponents []string, writeDump dump.WriteDump) error {
-	if tree == nil {
-		return fmt.Errorf("called with a nil tree")
-	}
-	if repo == nil {
-		return fmt.Errorf("called with a nil repository")
-	}
-	l := len(pathComponents)
-	if l == 0 {
-		return fmt.Errorf("empty path components")
-	}
-
+func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.Repository, prefix string, pathComponents []string, d *dump.Dumper) error {
 	// If we print / we need to assume that there are multiple nodes at that
 	// level in the tree.
 	if pathComponents[0] == "" {
 		if err := checkStdoutArchive(); err != nil {
 			return err
 		}
-		return writeDump(ctx, repo, tree, "/", os.Stdout)
+		return d.DumpTree(ctx, tree, "/")
 	}
 
 	item := filepath.Join(prefix, pathComponents[0])
+	l := len(pathComponents)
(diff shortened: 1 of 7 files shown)


cmd/restic/cmd_dump.go
```diff
@@ -188,7 +173,8 @@ func runDump(opts DumpOptions, gopts GlobalOptions, args []string) error {
 		Exitf(2, "loading tree for snapshot %q failed: %v", snapshotIDString, err)
 	}
 
-	err = printFromTree(ctx, tree, repo, "/", splittedPath, wd)
+	d := dump.New(opts.Archive, repo, os.Stdout)
```
MichaelEischer, line 176: What about letting `dump.New` handle the check which output formats are supported? That way the check would only be in a single place.

internal/dump/zip.go
```diff
@@ -3,36 +3,30 @@ package dump
...
-	w     *zip.Writer
-}
+func (d *Dumper) dumpZip(ctx context.Context, ch <-chan *restic.Node) (err error) {
+	w := zip.NewWriter(d.w)
 
-// Statically ensure that zipDumper implements dumper.
-var _ dumper = &zipDumper{}
+	defer func() {
+		if err == nil {
+			err = w.Close()
```
MichaelEischer, line 17: Shouldn't this error also be wrapped just like in `dumpTar`?

internal/dump/common.go
```diff
@@ -11,63 +11,84 @@ import (
...
 
-		node.Path = path.Join(rootPath, nodepath)
+		node.Path = path.Join(root.Path, nodepath)
 
-		if IsFile(node) || IsLink(node) || IsDir(node) {
-			err := dmp.dumpNode(ctx, node, repo)
-			if err != nil {
-				return false, err
-			}
+		switch node.Type {
```
MichaelEischer, line 82: I'd prefer to keep the node type checks using `IsFile` etc. instead of checking the node type differently in different places.

MichaelEischer (changes requested):
I've added a few small remarks, besides that the code looks fine.

cmd/restic/cmd_dump.go
greatroar: We need the repository to be open for dump.New and I want to validate the arguments before potentially showing a password prompt.

internal/dump/common.go
greatroar: Done.

internal/dump/zip.go
greatroar: Done.

MichaelEischer (approved):
LGTM. Thanks!

PR #3528, opened 2021-09-26
internal/restic: Don't allocate in Tree.Insert

greatroar:

restic.Tree.Insert allocated a Node that was then immediately overwritten in every call. That's solved. Also, NewTree now takes a capacity argument to get rid of the remaining allocations. Results from the new benchmark:

```
name         old time/op    new time/op    delta
BuildTree-8    34.6µs ± 4%     7.0µs ± 3%  -79.68%  (p=0.000 n=18+19)

name         old alloc/op   new alloc/op   delta
BuildTree-8    34.0kB ± 0%     0.9kB ± 0%  -97.37%  (p=0.000 n=20+20)

name         old allocs/op  new allocs/op  delta
BuildTree-8       108 ± 0%         1 ± 0%  -99.07%  (p=0.000 n=20+20)
```

There are some more serious issues with tree construction, including that the one in restic recover takes quadratic time. I'm still looking for a proper solution to that problem. I've decided not to remove the unused Tree.Sort method, because it may come in useful.

I don't think so.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_recover.go
+++ b/cmd/restic/cmd_recover.go
@@ -107,7 +107,7 @@ func runRecover(gopts GlobalOptions) error {
 		return nil
 	}
 
-	tree := restic.NewTree()
+	tree := restic.NewTree(len(roots))
 	for id := range roots {
 		var subtreeID = id
 		node := restic.Node{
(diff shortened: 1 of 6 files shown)


MichaelEischer (approved):
LGTM.

The performance "problem" in `recover` might be not too critical, so we should aim for a simple solution. As the nodes in recover are named after their ID, we could just sort the `roots` array. Then the insert becomes a binary search + append operation. Or we add an `Extend` method, which appends a bunch of nodes and then sorts them.

PR #3230, opened 2021-01-16
backup: Add resuming from aborted backups

aawsome:
What does this PR change? What problem does it solve?
-----------------------------------------------------

This is a follow-up of #3229.
In this PR I made a PoC using the idea of user odin from the forum, see link below.

With this PR, restic writes a temporary file (in the cache dir to the selected repo, subdir /resume/) which contains the list of already finished directories and the tree ID where it is saved in the repo.
If the backup suceeds, the file is removed.

At the start of a backup, restic tries to read the temporary file (if existing) and uses the contained trees as "additional parent trees" for the given directories. This means that w.r.t. already saved trees, this resumed backup is as fast as a follow-up backup using a parent snapshot.

Note that this PR relies on #3121 which should be merged first.

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------
closes #2280
alternative to #2960

https://forum.restic.net/t/quicker-interrupted-backup-resumption/3470/6

Checklist
---------

- [x] I have read the [Contribution Guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches)
- [x] I have enabled [maintainer edits for this PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
- [ ] I have added tests for all changes in this PR
- I have not added documentation for the changes (in the manual) - there is also none for parent snapshots
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (template [here](https://github.com/restic/restic/blob/master/changelog/TEMPLATE))
- [x] I have run `gofmt` on the code in all commits
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits)
- [x] I'm done, this Pull Request is ready for review

PR #3563, opened 2021-11-02
Improve the ETA displayed during backup

### LINKED ISSUE #3397: ETA is unreliable because of deduplication
Output of `restic version`
--------------------------
restic 0.12.0 compiled with go1.14.7 on linux/amd64

restic backup /src

SFTP

Expected behavior
-----------------
when i run the backup-process repeatedly, most chunks are already at the target and don't need to be transferred again.
the ETA should reflect the time needed until completion.
it can do better, by taking the average transfer speed of the last time-frame, instead of the overall transfer speed. the time-frame-parameter can be used to increase accuracy. 

in case it is already known after scanning, which/how many chunks actually need to be transferred, this information could also be used to increase accuracy even more.

Actual behavior
---------------
when i run the backup-process repeatedly, most chunks are already at the target and don't need to be transferred again.
the ETA is calculated rather simple and doesn't take this advantage into account.
because the transfer seems to start incredibly fast (GB/s) the ETA starts quite low and is constantly and slowly increasing, without any benefit.

repeat a backup of little-changing source multiple times and watch ETA. compare it to actual wall time.

simple algorithm (see above)

more complex algorithm (see above)

i still trust restic ;)

matta:
The ETA restic displays was based on a rate computed across the entire
backup operation. Often restic can progress at uneven rates. In the worst
case, restic progresses over most of the backup at a very high rate and
then finds new data to back up. The displayed ETA is then unrealistic and
never adapts.

Restic now estimates the transfer rate based on a sliding window, with the
goal of adapting to observed changes in rate. To avoid wild changes in the
estimate, several heuristics are used to keep the sliding window wide
enough to be relatively stable.

See commit message. This is intended to fix https://github.com/restic/restic/issues/3397.

Not in this form, but again see https://github.com/restic/restic/issues/3397

Checklist
---------

### DIFF
--- a/internal/ui/backup/progress.go
+++ b/internal/ui/backup/progress.go
@@ -43,7 +43,8 @@ type Progress struct {
 	progress.Updater
 	mu sync.Mutex
 
-	start time.Time
+	start     time.Time
+	estimator rateEstimator
 
 	scanStarted, scanFinished bool
 
@@ -60,6 +61,7 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
 		start:        time.Now(),
 		currentFiles: make(map[string]struct{}),
 		printer:      printer,
+		estimator:    *newRateEstimator(time.Now()),
 	}
 	p.Updater = *progress.NewUpdater(interval, func(runtime time.Duration, final bool) {
 		if final {
@@ -73,9 +75,13 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
 
 			var secondsRemaining uint64
 			if p.scanFinished {
-				secs := float64(runtime / time.Second)
-				todo := float64(p.total.Bytes - p.processed.Bytes)
-				secondsRemaining = uint64(secs / float64(p.processed.Bytes) * todo)
+				rate := p.estimator.rate(time.Now())
+				if rate <= 0 {
+					secondsRemaining = 0
+				} else {
+					todo := float64(p.total.Bytes - p.processed.Bytes)
+					secondsRemaining = uint64(todo / rate)
+				}
 			}
 
 			p.printer.Update(p.total, p.processed, p.errors, p.currentFiles, p.start, secondsRemaining)
@@ -105,6 +111,7 @@ func (p *Progress) addProcessed(c Counter) {
 	p.processed.Files += c.Files
 	p.processed.Dirs += c.Dirs
 	p.processed.Bytes += c.Bytes
+	p.estimator.recordBytes(time.Now(), c.Bytes)
 	p.scanStarted = true
 }
 
--- a/internal/ui/backup/progress.go
+++ b/internal/ui/backup/progress.go
@@ -76,7 +76,8 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
 			var secondsRemaining uint64
 			if p.scanFinished {
 				rate := p.estimator.rate(time.Now())
-				if rate <= 0 {
+				tooSlowCutoff := 1024.
+				if rate <= tooSlowCutoff {
 					secondsRemaining = 0
 				} else {
 					todo := float64(p.total.Bytes - p.processed.Bytes)
--- a/internal/ui/backup/rate_estimator_test.go
+++ b/internal/ui/backup/rate_estimator_test.go
@@ -1,0 +1,233 @@
+package backup
+
+import (
+	"math"
+	"testing"
+	"time"
+)
+
+const float64EqualityThreshold = 1e-6
+
+func almostEqual(a, b float64) bool {
+	if math.IsNaN(a) || math.IsNaN(b) {
+		panic("almostEqual passed a NaN")
+	}
+	return math.Abs(a-b) <= float64EqualityThreshold
+}
+
+func TestEstimatorDefault(t *testing.T) {
+	var start time.Time
+	e := newRateEstimator(start)
+	r := e.rate(start)
+	if math.IsNaN(r) || r != 0 {
+		t.Fatalf("e.Rate == %v, want zero", r)
+	}
+	r = e.rate(start.Add(time.Hour))
+	if math.IsNaN(r) || r != 0 {
+		t.Fatalf("e.Rate == %v, want zero", r)
+	}
+}
+
+func TestEstimatorSimple(t *testing.T) {
+	var when time.Time
+	type testcase struct {
+		bytes uint64
+		when  time.Duration
+		rate  float64
+	}
+
+	cases := []testcase{
+		{0, 0, 0},
+		{1, time.Second, 1},
+		{60, time.Second, 60},
+		{60, time.Minute, 1},
+	}
+	for _, c := range cases {
+		e := newRateEstimator(when)
+		e.recordBytes(when.Add(time.Second), c.bytes)
+		rate := e.rate(when.Add(c.when))
+		if !almostEqual(rate, c.rate) {
+			t.Fatalf("e.Rate == %v, want %v (testcase %+v)", rate, c.rate, c)
+		}
+	}
+}
+
+func TestBucketWidth(t *testing.T) {
+	var when time.Time
+
+	// Recording byte transfers within a bucket width's time window uses one
+	// bucket.
+	e := newRateEstimator(when)
+	e.recordBytes(when, 1)
+	e.recordBytes(when.Add(bucketWidth-time.Nanosecond), 1)
+	if e.buckets.Len() != 1 {
+		t.Fatalf("e.buckets.Len() is %d, want 1", e.buckets.Len())
+	}
+	b := e.buckets.Back().Value.(*rateBucket)
+	if b.totalBytes != 2 {
+		t.Fatalf("b.totalBytes is %d, want 2", b.totalBytes)
+	}
+	if b.end != when.Add(bucketWidth) {
+		t.Fatalf("b.end is %v, want %v", b.end, when.Add(bucketWidth))
+	}
+
+	// Recording a byte outside the bucket width causes another bucket.
+	e.recordBytes(when.Add(bucketWidth), 1)
+	if e.buckets.Len() != 2 {
+		t.Fatalf("e.buckets.Len() is %d, want 2", e.buckets.Len())
+	}
+	b = e.buckets.Back().Value.(*rateBucket)
+	if b.totalBytes != 1 {
+		t.Fatalf("b.totalBytes is %d, want 1", b.totalBytes)
+	}
+	if b.end != when.Add(2*bucketWidth) {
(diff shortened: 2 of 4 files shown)


internal/ui/backup/progress.go
```diff
@@ -153,9 +155,13 @@ func (p *Progress) Run(ctx context.Context) error {
 			}
 
 			if p.totalCh == nil {
-				secs := float64(time.Since(p.start) / time.Second)
-				todo := float64(total.Bytes - processed.Bytes)
-				secondsRemaining = uint64(secs / float64(processed.Bytes) * todo)
+				rate := estimator.rate(time.Now())
+				if rate <= 0 {
```
MichaelEischer, line 159: I don't think the ETA is useful any more once the rate falls below a few KB/s. So maybe use `rate < 1024` here?

internal/ui/backup/rate_estimator_test.go
```diff
@@ -0,0 +1,233 @@
...
+		},
+	}
+
+	for i, c := range cases {
+		var w time.Time
+		e := newRateEstimator(w)
+		w = applyChunks(c.chunks, w, e)
+		r := e.rate(w)
+		if !almostEqual(r, c.rate) {
+			t.Fatalf("e.Rate == %f, want %f (testcase %d %+v)", r, c.rate, i, c)
```
MichaelEischer, line 230: The error message should also include the `description`.

internal/ui/backup/rate_estimator_test.go
```diff
@@ -0,0 +1,233 @@
...
+		panic("almostEqual passed a NaN")
+	}
+	return math.Abs(a-b) <= float64EqualityThreshold
+}
+
+func TestEstimatorDefault(t *testing.T) {
+	var start time.Time
+	e := newRateEstimator(start)
+	r := e.rate(start)
+	if math.IsNaN(r) || r != 0 {
```
MichaelEischer, line 22: Using `rtest.Assert` from the import `rtest "github.com/restic/restic/internal/test"` would make this a bit cleaner. Then you can just use `rtest.Assert(r == 0, ...)`

internal/ui/backup/rate_estimator_test.go
```diff
@@ -0,0 +1,233 @@
...
+		t.Fatalf("e.Rate == %v, want zero", r)
+	}
+	r = e.rate(start.Add(time.Hour))
+	if math.IsNaN(r) || r != 0 {
+		t.Fatalf("e.Rate == %v, want zero", r)
+	}
+}
+
+func TestEstimatorSimple(t *testing.T) {
+	var when time.Time
```
MichaelEischer, line 32: `when`  and `c.when` is rather confusing. What about `base` or `start` instead?

internal/ui/backup/rate_estimator_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+func TestBucketWidth(t *testing.T) {
+	var when time.Time
+
+	// Recording byte transfers within a bucket width's time window uses one
+	// bucket.
+	e := newRateEstimator(when)
+	e.recordBytes(when, 1)
+	e.recordBytes(when.Add(bucketWidth-time.Nanosecond), 1)
+	if e.buckets.Len() != 1 {
```
MichaelEischer, line 63: I'd recommend `rtest.Equals` here.

internal/ui/backup/rate_estimator_test.go
```diff
@@ -0,0 +1,233 @@
...
+		t.Fatalf("b.end is %v, want %v", b.end, when.Add(time.Hour+time.Millisecond+time.Second))
+	}
+}
+
+type chunk struct {
+	repetitions uint64 // repetition count
+	bytes       uint64 // byte count (every second)
+}
+
+func applyChunk(c chunk, t time.Time, e *rateEstimator) time.Time {
```
MichaelEischer, line 106: I'd prefer to inline `applyChunk` into `applyChunks`.

MichaelEischer (changes requested):
The code looks good in general except for a few minor things I've commented on below.

The only thing I'm worried about is whether we'll just trade ETA underestimations with overestimations. But let's give it a try.

MichaelEischer:
@matta Are you still interested in working on this PR?

matta:
@MichaelEischer, sure.  This PR slipped my mind and I believe that I missed that you applied.  Are your comments and suggestions still relevant?  If so, I can incorporate them, rebase, and push a new version.

MichaelEischer:
@matta There haven't been any changes to this PR since I've reviewed it. So, yes the comments are still relevant.

aawsome:
> In the worst  case, restic progresses over most of the backup at a very high rate and
> then finds new data to back up. 

The problem is - as you pointed out - that the processing rates are very uneven. Moreover, when processing a particular bunch of data, you can compute the current processing rate (better said: the rate of the recent history), but you never know anything about the processing rate of the data-to-follow. So IMO the average processing speed (from beginning up to the current data) is as good as any other recent processing speed. The worst case for your scenario is to find new data at the very beginning and then progress over the remaining data at a very high rate.

So, yes this could be discussed, but IMHO is not worth the complexity of  more than 300 LoC.

aawsome:
Ah, and if this helps: in [rustic](https://github.com/rustic-rs/rustic) I first used the ETA provided by a sliding window (the rust crate indicatif supports this) but this was really "flickering" a lot as the rate for existing files jumps a lot (an existing file of 1GB is processed as fast as an existing file of 1kB...). So I decided to use an ETA based on the average rate - as restic is using ATM. 

If you really want to follow this sliding-window approach, make sure you get this "flickering" smoothed out!

greatroar:
> So, yes this could be discussed, but IMHO is not worth the complexity of more than 300 LoC.

I'll note that most of those 300 lines are tests, which is great, but otherwise I agree with @aawsome. In fact, if the ETA is so bad, and so hard to fix, we could just get rid of it. The info needed to compute its current value is in the status line already.

matta:
> In fact, if the ETA is so bad, and so hard to fix, we could just get rid of it.

It has been over a year since I sent this PR and I can't say I'd miss the ETA, since I almost always run restic unattended now.  

Given the somewhat lackluster interest among basically everybody (including myself), pursuing this PR further isn't high on my interest list. If somehow I've missed that other people are eager to see the ETA improve my interest level might change, but only then.

I do still think that restic's ETA as it is today is too simplistic to be useful and it would be good to address the problem.

The ETA accurate only in the unrealistic situation of a first-run backup consisting entirely of never-before-seen files. I might be atypical, but I never experience this.

This is particularly pronounced when first using restic. In that situation I'm always in a loop of running restic, noticing new paths to exclude, re-running restic, etc. By the end of this process only the tail end of the backup remains and restic's ETA is way too optimistic. If I recall, my initial backup was going to take >24H in reality but the estimate became much less than that, especially after I throttled uploads.  This kind of almost interactive iteration loop with restic is exactly when you want the ETA to be accurate, since you're most likely to be checking up on restic's progress in times like this.  With this PR, the ETA initially shows overly optimistic times as restic races through the first half of the backup, but within a few minutes of hitting unique data the ETA corrects itself into something plausible. It made intuitive sense to me, in that an ETA based on the recent past seems most plausible.

If the goal is user-value per line of code, then perhaps we'd be better off by deleting the ETA.

If the goal is providing an ETA that behaves more "sensibly" or "intuitively" for some definition of those terms, I think this PR can do that (or be the basis of an improved approach).

tbm:
> Given the somewhat lackluster interest among basically everybody (including myself), pursuing this PR further isn't high on my interest list. If somehow I've missed that other people are eager to see the ETA improve my interest level might change, but only then.

FWIW, I ran into the ETA estimation issue yesterday (uploading 10 GB via the network which takes ~8 hours; I interrupted at ~70% and when I resumed restic told me it would finish in 20-30 minutes) and I came to check if there were any reported issues and I saw your issue and PR.

I don't want to pollute this PR but wanted to say that I'm definitely interested in you completing this PR.

aawsome:
> FWIW, I ran into the ETA estimation issue yesterday (uploading 10 GB via the network which takes ~8 hours; I interrupted at ~70% and when I resumed restic told me it would finish in 20-30 minutes) and I came to check if there were any reported issues and I saw your issue and PR.

Actually, this PR won't help you in this case - it would still tell you for about 20 minutes that it would totally take 20-30 minutes until it realizes that the processing rate "suddenly" drastically changed and the ETA would increase to about 2.5 hours.

I doubt that this is what you would call a "good" ETA - yet the reality is that for your case there is no way for restic to know that after the very fast processed first part of your data (which has been already uploaded by the aborted first run) there comes a part which gets processed much faster.

For the problem you are referring, you might have a look at #3230 - that PR would allow to process the first part in only few seconds instead of ~20 minutes - and then just do the normal processing of the remaining data.

tbm:
> Actually, this PR won't help you in this case - it would still tell you for about 20 minutes that it would totally take 20-30 minutes until it realizes that the processing rate "suddenly" drastically changed and the ETA would increase to about 2.5 hours.

I didn't keep the log file so I can't remember the exact times (although I'm doing another backup now and could test). But IIRC restic is very fast in finding the data that is already there.  I see how your "Add resuming from aborted backups" PR would make it almost instantaneous but I don't think it took very long.

So with this patch, hopefully restic would figure out fairly quickly that the real speed is lower than the initial speed.

matta:
> > FWIW, I ran into the ETA estimation issue yesterday (uploading 10 GB via the network which takes ~8 hours; I interrupted at ~70% and when I resumed restic told me it would finish in 20-30 minutes) and I came to check if there were any reported issues and I saw your issue and PR.
> 
> Actually, this PR won't help you in this case - it would still tell you for about 20 minutes that it would totally take 20-30 minutes until it realizes that the processing rate "suddenly" drastically changed and the ETA would increase to about 2.5 hours.

Hey @aawsome, I believe this PR does a better job than you suggest above. It is intended to improve the estimate in precisely this "restarted backup" scenario.

The PR establishes a sliding window that is at least 2 minutes and 10 MB wide. Once both two minutes and 10 MB of transfer pass the estimate has completely "forgotten" any prior transfer rate, and the provided estimate becomes accurate, assuming the transfer rate is stable going forward.

In this case the network transfer rate is 10 GB over 8 hours or, equivalently, 10 MB over 28 seconds, so the two minute window dominates. A recovered backup would show an extremely quick estimate (probably sub-minute) for the initial "catching up" portion of the restarted backup. Once the network transfer work begins the estimate would quickly rise, smoothly, over the two minute period, to the a correct and stable estimate. Even during this two minute window the incorrectness of the estimate is evident because it is steadily increasing.

MichaelEischer:
I'm inclined to give this PR a try. If it doesn't work out, we still can revert it or just remove the ETA altogether.

MichaelEischer (approved):
LGTM.

PR #3570, opened 2021-11-07
List snapshots before index

### LINKED ISSUE #3428: `forget IDs` lists the backend once for each id
This issue was splitted from #3392.

Output of `restic version`
--------------------------
restic 0.12.0

`restic forget 12345678 12345679 1234567a ...`

Issue is independent of the backend.

Expected behavior
-----------------
restic should list() the backend only once.

Actual behavior
---------------
restic lists files in the backend for each passed snapshot ID. This can result in performance problems, for repositories containing a large number of snapshots.

`forget` uses `FindFilteredSnapshots` which internally calls `restic.FindSnapshot` for each snapshot ID. `FindSnapshot` then lists the backend to resolve short snapshot IDs to long IDs and to verify that the snapshot exists. That is even long snapshot IDs are affected by this performance issue.

List() the snapshots in the backend at most once.
See also #3395.

### LINKED ISSUE #3432: Copy fails with LoadTree(xyz) returned error id xyz not found in repository
Output of `restic version`
--------------------------
```
```

I'm syncing a local repository with a remote repository using `sftp` protocol:
```
export RESTIC_REPOSITORY=/backups"
export RESTIC_PASSWORD_FILE=/home/restic/.local_repo_pass"
export RESTIC_REPOSITORY2=sftp:user@host.example.org:./restic"
export RESTIC_PASSWORD_FILE2=/home/restic/.remote_repo_pass"
restic --verbose copy
```

Rsync.net

Expected behavior
-----------------
It works.

Actual behavior
---------------
```
LoadTree(a1cae849) returned error id a1cae84958e4c24e0aa731bf7efac1627ca891376ee2fe9dbb47b60eb08f6508 not found in repository
```

No idea.

No idea.

No idea. I tried removing the snapshot in question - `a1cae849` - but in both repos I just get back:
```
 $ restic snapshots a1cae849
repository 123qwe123 opened successfully, password is correct
Ignoring "a1cae849", it is not a snapshot id
```
So I have no idea what I'm supposed to do.

What?

MichaelEischer:
To allow backup and other commands, which only use non-exclusive locking, to run concurrently modifications to and reading from the repository have to maintain a certain order. For writing first pack files, then indexes and finally snapshots are written. For reading the inverse order must be used, to guarantee that all data required to access a snapshot is already accessible.

Commands which use an exclusive lock are exempted from this rule as there are not concurrent repository accesses.

However, most commands used to wrong order and loaded the index before listing which snapshots exist. This can lead to the problems in #3432 where a snapshot is accessed, however,  restic fails to load its data. This PR reorders the two steps where this can be done easily and creates a cached list of snapshots in other cases. This cached list is created before loading the index in order to maintain the invariants stated in the previous paragraph.

As a bonus it is now trivially possible to guarantee that `FindFilteredSnapshots` only issues a single `List()` to the backend.

Fixes #3432
Fixes #3428
Replaces #3395

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3428
+++ b/changelog/unreleased/issue-3428
@@ -1,0 +1,14 @@
+Bugfix: List snapshots in backend at most once to resolve snapshot ids
+
+Many commands support specifying a list of snapshot ids which are then used to
+determine the snapshot accessed by the command. To resolve snapshot ids or
+"latest" and check that these exist, restic listed all snapshots stored in the
+repository. Depending on the backend this can be a slow and/or expensive
+operation.
+
+Restic now lists the snapshots only once and remembers the result to resolve
+all further snapshot ids.
+
+https://github.com/restic/restic/issues/3428
+https://github.com/restic/restic/pull/3570
+https://github.com/restic/restic/pull/3395
--- a/changelog/unreleased/issue-3432
+++ b/changelog/unreleased/issue-3432
@@ -1,0 +1,14 @@
+Bugfix: Fix rare 'not found in repository' error for copy command
+
+In rare cases copy (and other commands) could report that LoadTree(...)
+returned a `id [...] not found in repository` error. This could be caused by a
+backup or copy command running concurrently. The error is only temporary,
+running the failed restic command a second time as a workaround solves the
+error.
+
+This issue has been fixed by correcting the order in which restic reads data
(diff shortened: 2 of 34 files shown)


fd0 (approved):
Nice work! I love the idea to capture the list of snapshots with `MemorizeBackend` instead of handling lists everywhere!

Issue #112, opened 2015-03-15
Define restic error handling policy


fd0:

TobiasBales:
If no one is working on it I will take a stab at it in a couple of hours, so "dips" I suppose :D

fd0:
It's yours ;)

rawtaz:
@TobiasBales Still up for it? We'd appreciate it!

rawtaz:
I'll try to include something about this in the FAQ or Troubleshooting section after/in relation to restructuring the documentation soon.

rossijonas:
I'm looking for an entry point to contribute to the project, should I start with this?

rawtaz:
@rossijonas Sure, feel free. I imagine it'd be a pretty short note in the backup section of the documentation. But see what you think :)

AdrianMastronardi:
Hi! I think this is already covered in the current documentation.

In the *Backing up* guide, the **Exit status codes** section explicitly says that if some source files can’t be read (e.g. permission errors), restic will still try to complete the backup and create a snapshot containing everything it could read, and returns **exit code 3** for that case: https://restic.readthedocs.io/en/stable/040_backup.html#exit-status-codes

The same is also summarized in the *Scripting* page under **Exit codes** (“backup command could not read some source data”): https://restic.readthedocs.io/en/stable/075_scripting.html#exit-codes

Given that, this issue seems resolved/documented already. I’d suggest closing it (unless there’s another error-handling behavior beyond partial read errors that isn’t documented yet).

PR #3590, opened 2021-12-07
Document Safe Passwords and Clarify B2 App Key information

### LINKED ISSUE #2238: Document safe passwords in manual/FAQ
I wasn't sure how to pick a safe password, specifically how long it should be.

I [found this post on the forum](https://forum.restic.net/t/cryptographic-key-storage/1424/6) from @fd0 :

> Since we’re applying a key-derivation function (scrypt), a much shorter password (20-25 characters if you want to make sure) is plenty applying a key-derivation function (scrypt), a much shorter password (20-25 characters if you want to make sure) is plenty

Would it be possible to add this to the manual?  Probably the best would be to have a section on choosing password in the beginning.

It could also document some other questions that commonly pop up regarding password, e.g. you mention `--password-file <(gpg --decrypt restic-password.gpg)` in another forum post.

Furthermore, you [mention on the forum](https://forum.restic.net/t/what-if-keys-are-compromised/1474/2?u=fd0) that the data is encrypted with the "master keys":

> All data in the repo is encrypted (and signed) with the “master keys”. In order to decrypt the “master keys”, you need the content of a key file together with the correct password.

So what's the difference between the password and the master keys and where are the master keys stored?

In short: a section on password safety and management would be great.

Maybe the section could also mention why restic doesn't use GPG (I believe you explained that somewhere but I can't find the reference now).

Thank you

### LINKED ISSUE #2672: Not so clear at first how to get the B2 Application Key
I set up restic + B2, but It took me more than 30 minutes to realize that one cannot simply fetch the Application Key from the App Keys section on backblaze.com. I figured it out **only after I logged out** and I visited https://www.backblaze.com/b2/docs/application_keys.html (which is not directly accessible if someone is already logged in).

In this page https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#backblaze-b2 I recommend to add the following information regarding B2 Application Keys: to get the Application Key ID, a user can go to the App Keys section; but for the Application Key a new user has to request a new application key (master for example). Then he will be given a new Application Key ID, and also a new **Application Key**, which the user will have to note since **it will appear only once**.

It took me a while to realize it, and I think it could be helpful to a lot of people.

metalsp0rk:
Fixes #2672 
Fixes #2238

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -442,6 +442,14 @@ dashboard on the "Buckets" page when signed into your B2 account:
     $ export B2_ACCOUNT_ID=<MY_APPLICATION_KEY_ID>
     $ export B2_ACCOUNT_KEY=<MY_APPLICATION_KEY>
 
+To get application keys, a user can go to the App Keys section of the Backblaze
+account portal.  You must create a master application key first.  From there, you
+can generate a standard Application Key.  Please note that the Application Key
+should be treated like a password and will only appear once.  If an Application
+Key is forgotten, you must generate a new one.
+
+For more information on application keys, refer to the Backblaze `documentation <https://www.backblaze.com/b2/docs/application_keys.html>`__.
+
 .. note:: As of version 0.9.2, restic supports both master and non-master `application keys <https://www.backblaze.com/b2/docs/application_keys.html>`__. If using a non-master application key, ensure that it is created with at least **read and write** access to the B2 bucket. On earlier versions of restic, a master application key is required.
 
 You can then initialize a repository stored at Backblaze B2. If the
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -207,3 +207,15 @@ temporarily disable your antivirus software to find out if it is the cause for
 your performance problems. If you are certain that the antivirus software is
 the cause for this and you want to gain maximum performance, you have to add
 the restic binary to an exclusions list within the antivirus software.
+
+How do I choose a strong password?
+----------------------------------
+
+Length is the single most important component in password strength.  That doesn't
+mean that other components such as complexity and entropy (or randomness) are not
+important to consider.  A strong password includes Alphabetical, Numerical and
+Special characters. For example, ``nk3E9Rr26md6GGySyyWMrfakw8Jck4$&vVY6`` would
+be a very strong password, if not for being in this documentation.
+
+There are plenty of tools out there, such as OpenSSL, pwgen or KeePass that can
+generate a sufficiently complex, random and long password.


doc/030_preparing_a_new_repo.rst
```diff
@@ -442,6 +442,14 @@ dashboard on the "Buckets" page when signed into your B2 account:
     $ export B2_ACCOUNT_ID=<MY_APPLICATION_KEY_ID>
     $ export B2_ACCOUNT_KEY=<MY_APPLICATION_KEY>
 
+To get application keys, a user can go to the App Keys section of the Backblaze
+account portal.  You must have a master application key first.  From there, you
```
MichaelEischer, line 446: I think `create` instead of `have` would be more accurate?

doc/040_backup.rst
```diff
@@ -40,6 +40,10 @@ As you can see, restic created a backup of the directory and was pretty
 fast! The specific snapshot just created is identified by a sequence of
 hexadecimal characters, ``40dc1520`` in this case.
 
+.. note:: When restic encounters an error during backup (e.g unreadable file, permission
```
MichaelEischer, line 43: Does this paragraph belong into this PR? It looks like copy&paste from #112.

doc/faq.rst
```diff
@@ -207,3 +207,12 @@ temporarily disable your antivirus software to find out if it is the cause for
 your performance problems. If you are certain that the antivirus software is
 the cause for this and you want to gain maximum performance, you have to add
 the restic binary to an exclusions list within the antivirus software.
+
+How do I choose a strong password?
+----------------------------------
+
+Any password close to 20-30 characters should be  enough.  However,length
```
MichaelEischer, line 214: There's a space missing before `length`.

doc/faq.rst
```diff
@@ -207,3 +207,12 @@ temporarily disable your antivirus software to find out if it is the cause for
...
+is not the only component for strong passwords.A strong password includes
+Alphabetical, Numerical and Special characters. For example, 
+``R3s7!cIsTh3B3s7B@ckupS0ftw@r3!`` would be a very strong password, if not for being in
```
MichaelEischer, line 217: leetspeak of `ResticIsTheBestBackupSoftware!` is considerably less secure than a fully random password of the same length. Tools to bruteforce passwords like hashcat even includes rules for leetspeak. So please just add a fully random password here.

doc/faq.rst
```diff
@@ -207,3 +207,12 @@ temporarily disable your antivirus software to find out if it is the cause for
 your performance problems. If you are certain that the antivirus software is
 the cause for this and you want to gain maximum performance, you have to add
 the restic binary to an exclusions list within the antivirus software.
+
+How do I choose a strong password?
+----------------------------------
+
+Any password close to 20-30 characters should be  enough.  However,length
+is not the only component for strong passwords.A strong password includes
+Alphabetical, Numerical and Special characters. For example,
```
MichaelEischer, line 216: The single most important part is actually the password length. One can actually trade e.g. special characters and even upper case letters for an approx. 30% higher password length. I'd like to de-emphasize the different character types part a bit.

MichaelEischer (changes requested):

doc/040_backup.rst
metalsp0rk: Not sure what happened with that commit.  Probably doesn't belong in this PR.  Let me rebase it out.

doc/faq.rst
metalsp0rk: Revising now.

metalsp0rk:
Should be better now.

doc/faq.rst
```diff
@@ -207,3 +207,15 @@ temporarily disable your antivirus software to find out if it is the cause for
 your performance problems. If you are certain that the antivirus software is
 the cause for this and you want to gain maximum performance, you have to add
 the restic binary to an exclusions list within the antivirus software.
+
+How do I choose a strong password?
+----------------------------------
+
+Length is the single most important component in password strength.  That doesn't
+mean that other components such as complexity and entopy (or randomness) are not
```
MichaelEischer, line 215: s/entopy/entropy/

doc/faq.rst
```diff
@@ -207,3 +207,15 @@ temporarily disable your antivirus software to find out if it is the cause for
...
+important to consider.  A strong password includes Alphabetical, Numerical and
+Special characters. For example, ``nk3E9Rr26md6GGySyyWMrfakw8Jck4$&vVY6`` would
+be a very strong password, if not for being in this documentation.
+
+There are plenty of tools out there, such as OpenSSL or KeePass that can generate
```
MichaelEischer, line 220: I guess you refer to something like `openssl rand -base64 40`? I actually prefer `pwgen` which is probably much more obvious to use.

changelog/unreleased/pull-3590
```diff
@@ -0,0 +1,7 @@
+Enhancement: Document Safe Passwords and Clarify B2 application key information
```
MichaelEischer, line 1: We usually don't add changelog entries for documentation changes. Please remove the changelog entry.

MichaelEischer (changes requested):
I've a few more nitpicks. Please also squash all commits.

doc/faq.rst
metalsp0rk: I do.  Mostly because it's on the overwhelming majority of systems, so it reduces the dependency list.  pwgen is not installed on any of my machines at the moment, installed with CentOS, Debian and Ubuntu.  I opted for a smaller dependency tree at the cost of command complexity.  This is a case of personal preference bleeding into docs.

I've added mention of pwgen in the statement on line 220.

metalsp0rk:
This is a bit long for a docs change, let me rebase this down into a more sane PR.

MichaelEischer (approved):
LGTM. Thanks for the quick cleanups!

tbm:
Thanks @metalsp0rk!

PR #3598, opened 2021-12-20
Typo in 060_forget.rst

piperfw:
The syntax is `--keep-*-within` not `keep-within-*`


MichaelEischer:
You're using an outdated version of #3416. There's no need to revert #3479.

piperfw:
Thanks, I was reading the stable docs rather than the latest.

PR #3599, opened 2021-12-20
Add --device-map flag to allow remapping DeviceIDs

### LINKED ISSUE #3041: Restic 0.10.0 always reports all directories "changed", adds duplicate metadata, when run on ZFS snapshots

    # restic version
    
    Data Blobs:      0 new
    Tree Blobs:    219 new
    Added to the repo: 17.876 MiB

intentionally-left-nil:
When creating backups from ZFS or btrfs snapshots, each snapshot will have a different DeviceID. This causes restic to upload the directory structure for each backup, even if nothing has changed. To solve this, this PR adds the --device-map flag which will allow users to map different device id's to the same logical device id.

The idea is that users can figure out what device ID their snapshot currently points to, (e.g. with `stat -c '%d'`
and then re-map that to a virtual device ID that is consistent across backups.
For example, if there's a btrfs snapshot mounted at `/home/.snapshots/123/snapshot/`, then you could do something like this:
```
device_id=$(stat -c '%d' /home/.snapshots/123/snapshot)
proot -b /home/.snapshots/123/snapshot:/home
restic backup --device-map "$device_id:42" /home
```

There is a decent amount of discussion on the linked issue. The larger issue here is that there are no guarantees anyways that the DeviceID is stable, and if it should be relied upon. One proposed idea was to implement an -ignore-deviceid flag, similar to the other ignore flags. However, given that the DeviceId is used to determine if symlinks span across different filesystems, this ended up not being a trivial change to implement.

Closes #3041 

Checklist
---------

- [] I'm done! This pull request is ready for review.

### DIFF
--- a/cmd/restic/.gitignore
+++ b/cmd/restic/.gitignore
@@ -1,1 +1,2 @@
 config.mk
+restic
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -92,6 +93,7 @@ type BackupOptions struct {
 	IgnoreCtime             bool
 	UseFsSnapshot           bool
 	DryRun                  bool
+	DeviceMap               []string
 }
 
 var backupOptions BackupOptions
@@ -136,6 +138,7 @@ func init() {
 	if runtime.GOOS == "windows" {
 		f.BoolVar(&backupOptions.UseFsSnapshot, "use-fs-snapshot", false, "use filesystem snapshot where possible (currently only Windows VSS)")
 	}
+	f.StringArrayVar(&backupOptions.DeviceMap, "device-map", nil, "Replace the device ID for files using the format src:dest")
 }
 
 // filterExisting returns a slice of all existing items, or an error if no
@@ -675,6 +678,26 @@ func runBackup(opts BackupOptions, gopts GlobalOptions, term *termstatus.Termina
 	}
 	if opts.IgnoreCtime {
 		arch.ChangeIgnoreFlags |= archiver.ChangeIgnoreCtime
+	}
+
+	if opts.DeviceMap != nil {
+		arch.DeviceMap = map[uint64]uint64{}
+		for _, val := range opts.DeviceMap {
+			ids := strings.Split(val, ":")
+			if len(ids) != 2 {
+				return errors.New("--device-map format is incorrect, should be src:dest")
+			}
+			src, err := strconv.ParseUint(ids[0], 10, 64)
+			if err != nil {
+				return errors.New("--device-map does not contain a valid source id")
+			}
+			dest, err := strconv.ParseUint(ids[1], 10, 64)
+			if err != nil {
+				return errors.New("--device-map does not contain a valid dest id")
+			}
+
+			arch.DeviceMap[src] = dest
+		}
 	}
 
 	if parentSnapshotID == nil {
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -82,6 +82,11 @@ type Archiver struct {
 
 	// Flags controlling change detection. See doc/040_backup.rst for details.
 	ChangeIgnoreFlags uint
+
+	// Maps DeviceIDs from the key to the value of the map
+	// For example, if a file is located on device 2, and DeviceMap = {2: 4}, then
+	// the arciver will save the file as if the file were located on device 4
+	DeviceMap map[uint64]uint64
 }
 
 // Flags for the ChangeIgnoreFlags bitfield.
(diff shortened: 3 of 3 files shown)


MichaelEischer:
Why not let restic automatically assign pseudo-device ids as suggested in https://github.com/restic/restic/issues/3041#issuecomment-939349951 ? As rawtaz and I have commented (see e.g. https://github.com/restic/restic/issues/3041#issuecomment-939525912 ) `--device-map` looks like it requires rather a lot of setup to work.

AnilRedshift:
Hi @MichaelEischer, thanks for taking a look. I'd be happy to implement an automatic mapping algorithm, but I'm not sure how to go about it.

Consider a scenario with 4 files, spread across two devices:
```
/device1/a.txt
/device1/b.txt
/device2/c.txt
/device2/d.txt
```

and let's say that for the first backup, device1 has a DeviceID of 17, and device2 has a DeviceID of 18.
For the second run let's say Device1 has a DeviceID of 81 and Device2 has a DeviceID of 82

The question becomes: How do we map device1 and device2 to pseudo-ids which are consistent across runs?
Let's say you do something like hash the path, or the inode. So we hash /device1/a.txt and let's say that hashes to 3. So we have a pseudo-map that says "device 17 maps to pseudo 3". Now when we see /device1/b.txt, the mapping already exists so we can use the existing value. `c.txt` is missing a mapping, so you would hash it and then do something like "device 18 maps to pseudo 4"

But the problem with this approach is it isn't stable across backups, unless enumeration happens in the same order. For example, on the next backup, let's say that instead of backing up a.txt, the enumeration finds /device1/b.txt first (or maybe you deleted a.txt).

Now, the algorithm would see that there is no device mapping for ID 81, and so it would hash /device1/b.txt. But now this is a different file/inode so let's say that maps to pseudo id 11. And now we haven't solved anything because the pseudo-id's are not consistent across runs.

Some ideas I considered, but with obvious flaws:
* Find some way to do enumeration in-order of devices (I don't see how you can enforce this with hard links)
*  Poke into the guts of the filesystem to get the mappings (e.g. look at /proc/mounts) and build a hash off the base mount path, rather than the file
* For each file, try to get it's parent, (and keep getting its parent), until the DeviceID changes. Use that folder path/inode as the input for the hash algorithm 

The last approach seems most reasonable to me but also fairly involved. I don't know enough about all the nuances of file systems (including network drives) to know what gotcha's I'd run into trying to traverse up parent directories.

The benefit of the current approach is it's very iterative (code-wise) and results in a small diff with consistent, repeatable behavior. The last approach would involve a much larger diff and would be more likely to have a tail of "it's not stable in this configuration" etc.

Happy to hack on ideas more, once we have an idea of how the algorithm should work.

AnilRedshift:
Thinking about this some more, it looks like `findmnt` gives me the set of device ID information needed to make a stable lookup. However at that point, I might as well just implement `-use-fs-snapshot` and be done with it. I'll think on this some more

rawtaz:
Is this issue specific to Linux? I'm wondering how many systems have the issue but don't have `findmnt`.

shamer:
I am using ZFS snapshots with restic and am affected by [issue 3041](https://github.com/restic/restic/issues/3041). With the `--device-map` flag my strategy for using it would be to assign pseudo device IDs based on the mount point. These are very stable for my use case.
My [experiments show](https://gist.github.com/shamer/56ed2c522728e17b1ec5b603d19f90bb) that everything else is stable except the inode of the directory of the mountpoint.

```
#!/bin/bash
for m in $(findmnt -R -T . --json | jq --raw-output '.filesystems[0].children[].target') ; do
  SRCDEV=$(stat --format "%Ld" $m)
  DEVHASH=$(echo ${m//$(pwd)/} | b2sum --length=64 | cut -f1 -d" " | tr a-f A-F)
  PSEUDODEV=$(echo "obase=10; ibase=16; ${DEVHASH}" | bc)
  echo --device-map ${SRCDEV}:${PSEUDODEV}
done
```

MichaelEischer:
> But the problem with this approach is it isn't stable across backups, unless enumeration happens in the same order. For example, on the next backup, let's say that instead of backing up a.txt, the enumeration finds /device1/b.txt first (or maybe you deleted a.txt).

The backup sorts filenames before traversing a folder. Thus the traversal order is stable.

> For example, if you hard link /device1/folderA into /device2/folderA, then iyou would run into the same issue depending on if the backup code sees /device1/a.txt or /device2/folderA/c.txt. 

Hardlinks cannot point across devices. Or is there some subvolume related trickery that avoids that limitation?

The archiver component already traverses the filesystem starting from the root directory `/`. So we'd just have to call stat before entering a directory afterwards to detect when the mountpoint has changed.

shamer:
> Hardlinks cannot point across devices. Or is there some subvolume related trickery that avoids that limitation?

Although hardlinks cannot be made across devices, bind mounts can be made at different points in the filesystem. The bind mounts all share the same device ID.

With the traversal order being stable, bind mounts don't seem like an issue though. The mount that is seen first would be the path used for the pseudo identifier for the device. If this path is hashed in some way to produce the identifier it wouldn't matter how many other devices were seen before this device.

ArsenArsen:
Hey, I've been successfully using this patch in unattended backups, though, on a filesystem without any bind mounts in it.
Rebasing the patch for new releases becomes increasingly hard, though, as the gap between the source tree and the patch widens.
May I ask for a review and merge, if possible?

Thanks!

MichaelEischer:
As discussed in #3041 we want to have a solution that just works, without having users manually map device ids. That is the PR is just there as a temporary workaround, but is unlikely to get merged.

intentionally-left-nil:
> As discussed in #3041 we want to have a solution that just works, without having users manually map device ids. That is the PR is just there as a temporary workaround, but is unlikely to get merged.

I'm going to close this PR out. My personal opinion is that in the search of a perfect solution, we've sacrificed a good enough one.  This issue has been open for 3 years now. It's frustrating that a low-risk, opt-in fix wasn't able to make it over the line

MichaelEischer:
There's now https://github.com/restic/restic/pull/4006 .

PR #3610, opened 2021-12-29
Improve handling of temporary files on windows

### LINKED ISSUE #1551: Warn and continue when tempfiles cannot be removed
Reminder issue for the behavior described here: https://forum.restic.net/t/temp-file-errors-on-windows-10/380/12

In short: On Windows, restic creates temp files which it may not be able to remove because another process (typically a virus scanner) accesses them, so restic aborts.

This should be changed to a warning only.

### LINKED ISSUE #3465: Improve temporary files handling on Windows

MichaelEischer:
restic currently uses `ioutil.TempFile(...)` to create its temporary files on Windows. As files cannot be deleted on Windows while these are still opened, restic sometimes leaves temporary files behind. This PR sets the FILE_FLAG_DELETE_ON_CLOSE flag to let windows delete the file automatically once all references were deleted. In addition, the flag FILE_ATTRIBUTE_TEMPORARY is added to avoid unnecessary writes of the temporary files to disk.

As `os.Open` and therefore `ioutil.TempFile` do not support these flags, this PR introduces a custom `TempFile` implementation, that directly calls `syscall.CreateFile`.

This PR is based on the code from @ducalex in #3465.

The temporary file is also created using shareMode 0, which prevents other processes from accessing the file in parallel. Thus restic is the only process which can open the file. As it will be deleted on close, this ensures that a tempfile can be removed, which fixes #1551.

Closes #3465.
Closes #1551.
Supersedes #3306.

Checklist
---------

### DIFF
--- a/internal/fs/file_windows.go
+++ b/internal/fs/file_windows.go
@@ -1,41 +1,14 @@
 package fs
 
 import (
+	"math/rand"
 	"os"
 	"path/filepath"
-	"strings"
-	"syscall"
 	"strconv"
-	"sync"
+	"strings"
 	"time"
-)
 
-// Random number state.
-// We generate random temporary file names so that there's a good
-// chance the file doesn't exist yet - keeps the number of tries in
-// TempFile to a minimum.
-var rand uint32
-var randmu sync.Mutex
-
-func reseed() uint32 {
-	return uint32(time.Now().UnixNano() + int64(os.Getpid()))
-}
-
-func nextRandom() string {
-	randmu.Lock()
-	r := rand
-	if r == 0 {
-		r = reseed()
-	}
-	r = r*1664525 + 1013904223 // constants from Numerical Recipes
-	rand = r
-	randmu.Unlock()
-	return strconv.Itoa(int(1e9 + r%1e9))[1:]
-}
-
-const (
-	FILE_ATTRIBUTE_TEMPORARY  = 0x00000100
-	FILE_FLAG_DELETE_ON_CLOSE = 0x04000000
+	"golang.org/x/sys/windows"
 )
 
 // fixpath returns an absolute path on windows, so restic can open long file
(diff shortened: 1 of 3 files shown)


greatroar:
FILE_ATTRIBUTE_TEMPORARY and FILE_FLAG_DELETE_ON_CLOSE are both available in golang.org/x/sys/windows. That also offers everything now imported from syscall.

internal/fs/file_windows.go
```diff
@@ -61,25 +33,40 @@ func fixpath(name string) string {
...
-	flags := uint32(FILE_ATTRIBUTE_TEMPORARY | FILE_FLAG_DELETE_ON_CLOSE)
-	
+	access := uint32(windows.GENERIC_READ | windows.GENERIC_WRITE)
+	creation := uint32(windows.CREATE_NEW)
+	share := uint32(0) // prevent other processes from accessing the file
+	flags := uint32(windows.FILE_ATTRIBUTE_TEMPORARY | windows.FILE_FLAG_DELETE_ON_CLOSE)
+
 	for i := 0; i < 10000; i++ {
-		path := filepath.Join(dir, prefix+nextRandom())
+		randSuffix := strconv.Itoa(int(1e9 + rand.Intn(1e9)%1e9))[1:]
```
fd0, line 56: I think it's a bad idea to use `rand.Intn()` from the `math` package since it's deterministic: https://go.dev/play/p/ofccaFGv1mc so two different restic processes will generate exactly the same names for the temp files. Or am I wrong?

fd0 (changes requested):

internal/fs/file_windows.go
MichaelEischer: Oh, I've overlooked that. Now the code creates a new random number generator instance and seeds it with the current time in nanoseconds. That should be sufficient?

fd0: Fine with me :)

fd0 (approved):
LGTM!

PR #3611, opened 2021-12-29
Adjust worker goroutines to number of backend connections

### LINKED ISSUE #1467: Make listPackWorkers a configuration option
For large repositories, the hard-coded value of 10 for listPackWorkers significantly reduces the performance of commands like prune.

For example, with a current version of restic (23d7d9159739650e9e99138e3f5bb99f4e2b80ca) from master working against a 3.2TB repo in an AWS S3 bucket, this prune comand:

restic -p [path] -r s3:s3.amazonaws.com/[bucket] -o s3.connections=128 prune

takes 2.8 hours to build a new index with the default setting of 10 for listPackWorkers from a lightly loaded r4.2xlarge instance.  Restic is also unable to take advantage of the increase s3 parallelism that I've given it -- I only see 10-20 simultaneous S3 connections during the run.

With listPackWorkers set to 512, the same index rebuild completes in 11 minutes.

The total runtime of an "empty" prune of this repo, which performs no actual repo changes, drops from 6 hours to 20 minutes with this change.  Since prunes lock out the repo, making them run quickly is important if you want to use restic with large repositories.

The optimum value for listPackWorkers will vary based on the size of the machine running restic, the latency to the backend store and the number of concurrent requests that the backend permits, so requiring a recompile to tweak this value is clunky.

I'd recommend making this a configuration option, similar to s3.connections.

### LINKED ISSUE #2162: handle large prune much more efficent

MichaelEischer:
Adapt worker count based on whether an operation is CPU or IO-bound.

Use runtime.GOMAXPROCS(0) as worker count for CPU-bound tasks, repo.Connections() for IO-bound task and a combination if a task can be both.

Typical IO-bound tasks are download / uploading / deleting files. Decoding / Encoding / Verifying are usually CPU-bound. Several tasks are a combination of both, e.g. for combined download and decode functions. In the latter case add both limits together. As the backends have their own concurrency limits restic still won't download more than `repo.Connections()` files in parallel, but the additional workers can decode already downloaded data in parallel.

This PR also includes the commits from #3489.
The last commit "Adapt concurrency to streaming check --read-data and prune" only makes sense together with #3484.

Alternatives:

The current construction still requires users to adjust one knob namely the connections limit. restic could just use e.g. the CPU count as a proxy for the system performance and use it to scale the number of backend connections. Each operation could then use a different scaling factor. This has the downside that the CPU count and network connection can be correlated (e.g. for servers in the cloud) but don't have to be. That heuristic thus would only work well in some cases.

Finding the optimal number of goroutines automatically could also be possible by using a self-tuning control loop, but that would complicate thing quite a bit.

Fixes #2162
Fixes #1467

Checklist
---------

- ~~[ ] I have added tests for all code changes.~~ I don't see how I could test this.
- ~~[ ] I have added documentation for relevant changes (in the manual).~~ The connections option should already be documented by #3489.

### DIFF
--- a/changelog/unreleased/issue-2162
+++ b/changelog/unreleased/issue-2162
@@ -1,0 +1,18 @@
+Enhancement: Adapt IO concurrency based on backend connections
+
+Many commands used hard-coded limits for the number of concurrent operations.
+This prevented speed improvements by increasing the number of connections used
+by a backend.
+
+These limits have been replaced by using the configured number of backend
+connections instead. It can be controlled using the
+`-o <backend-name>.connections=5` option. Commands will then automatically
+scale their parallelism accordingly.
+
+To limit the number of CPU cores used by restic, you can set the environment
+variable `GOMAXPROCS` accordingly. For example to use a single CPU core, use
+`GOMAXPROCS=1`.
+
+https://github.com/restic/restic/issues/2162
+https://github.com/restic/restic/issues/1467
+https://github.com/restic/restic/pull/3611
--- a/cmd/restic/delete.go
+++ b/cmd/restic/delete.go
@@ -18,8 +18,6 @@ func DeleteFilesChecked(gopts GlobalOptions, repo restic.Repository, fileList re
 	return deleteFiles(gopts, false, repo, fileList, fileType)
 }
 
-const numDeleteWorkers = 8
-
 // deleteFiles deletes the given fileList of fileType in parallel
 // if ignoreError=true, it will print a warning if there was an error, else it will abort.
 func deleteFiles(gopts GlobalOptions, ignoreError bool, repo restic.Repository, fileList restic.IDSet, fileType restic.FileType) error {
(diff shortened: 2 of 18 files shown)


greatroar:
This looks sensible. The main downside is that crash reports become much longer with more goroutines. However, shouldn't more channels be buffered to ensure smooth throughput?

MichaelEischer:
> The main downside is that crash reports become much longer with more goroutines. 

I guess that's a price worth paying if it helps with the throughput.

> However, shouldn't more channels be buffered to ensure smooth throughput?

The idea was to compensate for that by adding as many goroutines as there are CPU cores in methods where substantial processing is necessary for further IO.
In most places the IO-tasks are also generated by submitting a preexisting list on a channel. My expectation would be that the IO is "slow" enough that we shouldn't run into too much congestion on the channel.

MichaelEischer:
This PR completes the upload pipeline reworking started by #3489 and several other PRs. It (hopefully) addresses some of the scalability issues that have showed up regarding backends with high latency.

PR #3612, opened 2021-12-30
Add GCS option for encryption keys

### LINKED ISSUE #3530: google storage: customer-supplied encryption keys
Output of `restic version`
--------------------------
restic 0.12.1 (v0.12.1-101-g78dac2fd-dirty) compiled with go1.16.8 on linux/amd64

I would like to implement customer-supplied encryption keys in google storage backend. 

I believe that this would add another layer of security to cloud backups. GCP do not allow to download encrypted files without providing key even if user authenticate correctly.

It should be noted that customer-supplied keys are not zero-knowledge encryption because user sends encryption key with request and data are encrypted or decrypted in the cloud. Encryption key is stored in memory and removed after operation.

I made simple implementation based on https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys and it works. There are still some problems unsolved such as: how user should provide encryption key or whether to check if repo uses customer-supplied encryption or just return error to user.

Restic is a very handy tool.

tree68505:
These changes allow users to provide 256-bit AES crypto keys to Google Cloud Storage to enable [Customer-supplied encryption](https://cloud.google.com/storage/docs/encryption/customer-supplied-keys?hl=en). By providing one's own keys which are beyond Google's management, users enable an extra layer of protection for their data in the cloud. Once data is encrypted with provided keys, data can not be downloaded or decrypted without providing the same key that was used to encrypt.

Closes #3530

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3612
+++ b/changelog/unreleased/pull-3612
@@ -1,0 +1,8 @@
+Enhancement: Add GCS option for encryption keys
+
+We have added support Google Cloud Customer-supplied 
+encryption keys. Users can provide 256-bit AES crypto 
+keys to add an extra layer of protection to stored files in the cloud.
+
+https://github.com/restic/restic/pull/3612
+https://github.com/restic/restic/issues/3530
--- a/internal/backend/gs/config.go
+++ b/internal/backend/gs/config.go
@@ -8,6 +8,11 @@ import (
 	"github.com/restic/restic/internal/options"
 )
 
+type gsKeyTuple struct {
+	Key    []byte
+	KeySha string
+}
+
 // Config contains all configuration necessary to connect to a Google Cloud Storage
 // bucket. We use Google's default application credentials to acquire an access token, so
 // we don't require that calling code supply any authentication material here.
@@ -17,6 +22,10 @@ type Config struct {
 	Prefix    string
 
 	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+
+	//https://cloud.google.com/storage/docs/encryption/customer-supplied-keys#gsutil
+	EncryptionKey  gsKeyTuple        `option:"encryption_key" help:"Use custom-supplied encryption key."`
+	DecryptionKeys map[string][]byte `option:"decryption_keys" help:"Use custom-supplied decryption keys."`
 }
 
 // NewConfig returns a new Config with the default values filled in.
--- a/internal/backend/gs/gs.go
+++ b/internal/backend/gs/gs.go
@@ -40,6 +40,8 @@ type Backend struct {
 	bucket       *storage.BucketHandle
 	prefix       string
 	listMaxItems int
+	encryptKey   gsKeyTuple
+	decryptKeys  map[string][]byte
 	backend.Layout
 }
 
@@ -110,6 +112,8 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 		bucketName:  cfg.Bucket,
 		bucket:      gcsClient.Bucket(cfg.Bucket),
 		prefix:      cfg.Prefix,
+		encryptKey:  cfg.EncryptionKey,
+		decryptKeys: cfg.DecryptionKeys,
 		Layout: &backend.DefaultLayout{
 			Path: cfg.Prefix,
 			Join: path.Join,
@@ -245,7 +249,14 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
 	//
 	// restic typically writes small blobs (4MB-30MB), so the resumable
 	// uploads are not providing significant benefit anyways.
-	w := be.bucket.Object(objName).NewWriter(ctx)
+
+	var w *storage.Writer
+	bucket_object := be.bucket.Object(objName)
+	if len(be.encryptKey.Key) != 0 {
+		w = bucket_object.Key(be.encryptKey.Key).NewWriter(ctx)
+	} else {
+		w = bucket_object.NewWriter(ctx)
+	}
 	w.ChunkSize = 0
 	w.MD5 = rd.Hash()
 	wbytes, err := io.Copy(w, rd)
@@ -276,8 +287,9 @@ func (be *Backend) Load(ctx context.Context, h restic.Handle, length int, offset
 }
 
 func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error) {
+	var err error
 	debug.Log("Load %v, length %v, offset %v from %v", h, length, offset, be.Filename(h))
-	if err := h.Valid(); err != nil {
+	if err = h.Valid(); err != nil {
 		return nil, err
 	}
 
@@ -296,17 +308,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int, 
 	objName := be.Filename(h)
 
 	be.sem.GetToken()
-
 	ctx, cancel := context.WithCancel(ctx)
 
-	r, err := be.bucket.Object(objName).NewRangeReader(ctx, offset, int64(length))
+	//https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys
+	err = nil
+	bucketObject := be.bucket.Object(objName)
+	objectArrts, _ := bucketObject.Attrs(ctx)
+	objectKeySha := objectArrts.CustomerKeySHA256
+	var cryptoKey []byte
+	var rangeReader *storage.Reader
+
+	if objectKeySha != "" {
+
+		//try current key first
+		if objectKeySha == be.encryptKey.KeySha {
+			cryptoKey = be.encryptKey.Key
+		}
+		//fallback to list of old keys
+		if cryptoKey == nil {
+			cryptoKey = be.decryptKeys[objectKeySha]
+		}
+		//none of the supplied keys match object key
+		if cryptoKey == nil {
+			err = errors.New(("Google Storage customer-supplied encryption keys failed."))
+		}
+		if err == nil {
+			rangeReader, err = bucketObject.Key(cryptoKey).NewRangeReader(ctx, offset, int64(length))
+		}
+	} else {
+		rangeReader, err = bucketObject.NewRangeReader(ctx, offset, int64(length))
+
+	}
+
 	if err != nil {
 		cancel()
 		be.sem.ReleaseToken()
 		return nil, err
 	}
 
-	return be.sem.ReleaseTokenOnClose(r, cancel), err
+	return be.sem.ReleaseTokenOnClose(rangeReader, cancel), err
 }
 
 // Stat returns information about a blob.
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -3,11 +3,14 @@ package main
 import (
 	"bufio"
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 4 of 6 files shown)


tree68505:
Looking into broken transport limiting due to changes in `getStorageClient`.

internal/backend/gs/config.go
```diff
@@ -16,13 +16,19 @@ type Config struct {
 	Bucket    string
 	Prefix    string
 
-	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+	ConnectionsLimit uint `option:"connections_limit" help:"set a limit for the number of concurrent connections (default: 5)"`
```
MichaelEischer, line 19: That parameter is called `connections` everywhere else. Please don't rename it.

internal/backend/gs/config.go
```diff
@@ -16,13 +16,19 @@ type Config struct {
...
+
+	//https://cloud.google.com/storage/docs/encryption/customer-supplied-keys#gsutil
+	EncryptKey             string      `option:"encryption_key" help:"Use custom-supplied encryption key."`
+	DecryptKeys            [100]string `option:"decryption_keys" help:"Use custom-supplied decryption keys."`
+	ApplicationCredentials string      `option:"application_credentials" help:"Use environment path for credentials."`
```
MichaelEischer, line 24: That options is currently never set.

internal/backend/gs/config.go
```diff
@@ -16,13 +16,19 @@ type Config struct {
 	Bucket    string
 	Prefix    string
 
-	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+	ConnectionsLimit uint `option:"connections_limit" help:"set a limit for the number of concurrent connections (default: 5)"`
+
+	//https://cloud.google.com/storage/docs/encryption/customer-supplied-keys#gsutil
+	EncryptKey             string      `option:"encryption_key" help:"Use custom-supplied encryption key."`
```
MichaelEischer, line 22: Shouldn't these variables also be called `EncryptionKey` and `DecryptionKeys`?

internal/backend/gs/gs.go
```diff
@@ -239,7 +244,13 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
 	//
 	// restic typically writes small blobs (4MB-30MB), so the resumable
 	// uploads are not providing significant benefit anyways.
-	w := be.bucket.Object(objName).NewWriter(ctx)
+	var test1 []byte
+	test1, _ = base64.StdEncoding.DecodeString(be.encryptKey)
```
MichaelEischer, line 248: Please use a descriptive variable name. What happens if the key cannot be base64-decoded? Just ignoring the error seems to be a bad idea.

internal/backend/gs/gs.go
```diff
@@ -205,7 +210,7 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
 	if err := h.Valid(); err != nil {
 		return err
 	}
-
+	var w *storage.Writer
```
MichaelEischer, line 213: Why is the variable defined here and not around line 247 (before / after the base64 decoding?)

internal/backend/gs/gs.go
```diff
@@ -282,8 +293,13 @@ func (be *Backend) Load(ctx context.Context, h restic.Handle, length int, offset
 }
 
 func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int, offset int64) (io.ReadCloser, error) {
+	var err error
+	var rangeReader *storage.Reader
```
MichaelEischer, line 297: Don't predefine variables, but rather define them near the place they are used for the first time.

internal/backend/gs/gs.go
```diff
@@ -301,16 +317,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
+			break
+		}
+	}
+
 	be.sem.GetToken()
 
-	r, err := be.bucket.Object(objName).NewRangeReader(ctx, offset, int64(length))
+	//https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys
+	err = errors.New(("Google Storage customer-supplied encryption keys failed."))
+	if encryptKeyPresent {
```
MichaelEischer, line 334: Why not use `if be.encryptKey != "" {` here?

internal/backend/gs/gs.go
```diff
@@ -301,16 +317,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
+			decryptKeysPresent = true
+			break
+		}
+	}
+
 	be.sem.GetToken()
 
-	r, err := be.bucket.Object(objName).NewRangeReader(ctx, offset, int64(length))
+	//https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys
+	err = errors.New(("Google Storage customer-supplied encryption keys failed."))
```
MichaelEischer, line 333: `err` should only be `!= nil` if there's an actual error.

internal/backend/gs/gs.go
```diff
@@ -301,16 +317,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
+	if encryptKeyPresent {
+		cryptoKey, _ = base64.StdEncoding.DecodeString(be.encryptKey)
```
MichaelEischer, line 335: Please decode and verify the key once. That is the check should happen when the backend is opened and please also verify that the key is indeed as AES-256 key.

The same also applies to the decryption keys.

internal/backend/gs/gs.go
MichaelEischer: Please just add the encryption key to the list of decryption keys. That way we don't need separate code paths.

internal/backend/gs/gs.go
```diff
@@ -301,16 +317,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
+		rangeReader, err = be.bucket.Object(objName).Key(cryptoKey).NewRangeReader(ctx, offset, int64(length))
+	}
+	if decryptKeysPresent && err != nil {
+		for _, key := range be.decryptKeys {
+
+			cryptoKey, _ = base64.StdEncoding.DecodeString(key)
+			rangeReader, err = be.bucket.Object(objName).Key(cryptoKey).NewRangeReader(ctx, offset, int64(length))
```
MichaelEischer, line 342: Doesn't this issue up to 100 API requests? The documentation states that

> When decrypting, gsutil calculates the SHA256 hash of the supplied encryption and decryption keys and selects the correct key to use for a particular object by matching the SHA256 hash in the object's metadata.

That sounds to me like it would be possible to determine which key will work.

internal/backend/gs/gs.go
```diff
@@ -301,16 +317,45 @@ func (be *Backend) openReader(ctx context.Context, h restic.Handle, length int,
...
+			if err == nil {
+				break
+			}
+		}
+	}
+
+	if err != nil {
+		rangeReader, err = be.bucket.Object(objName).NewRangeReader(ctx, offset, int64(length))
```
MichaelEischer, line 350: Please deduplicate the calls to ``be.bucket.Object(objName)` and `.NewRangeReader(ctx, offset, int64(length))`. Copy-pasting them tree times just asks for problems in the long term.

changelog/unreleased/pull-3612
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add GCS option for encryption keys
+
+These changes allow users to provide 256-bit AES crypto
```
MichaelEischer, line 3: Please word it something like `We have added support for xyz`. "These changes" is not meaningful when reading the changelog.

changelog/unreleased/pull-3612
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add GCS option for encryption keys
+
+These changes allow users to provide 256-bit AES crypto 
+keys to Google Cloud Storage to enable Customer-supplied 
+encryption.
+
+https://github.com/restic/restic/pull/3612
```
MichaelEischer, line 7: Please add a reference to the issue here.

internal/backend/gs/gs.go
```diff
@@ -239,7 +244,13 @@ func (be *Backend) Save(ctx context.Context, h restic.Handle, rd restic.RewindRe
 	//
 	// restic typically writes small blobs (4MB-30MB), so the resumable
 	// uploads are not providing significant benefit anyways.
-	w := be.bucket.Object(objName).NewWriter(ctx)
+	var test1 []byte
+	test1, _ = base64.StdEncoding.DecodeString(be.encryptKey)
+	if be.encryptKey != "" {
+		w = be.bucket.Object(objName).Key(test1).NewWriter(ctx)
```
MichaelEischer, line 250: Please deduplicate these two if-branches.

internal/backend/gs/config.go
```diff
@@ -16,13 +16,19 @@ type Config struct {
...
+	DecryptKeys            [100]string `option:"decryption_keys" help:"Use custom-supplied decryption keys."`
+	ApplicationCredentials string      `option:"application_credentials" help:"Use environment path for credentials."`
+	AccessToken            string      `option:"access_Token" help:"User encrionment path for access token."`
```
MichaelEischer, line 25: The AccessToken is dead code right now.

MichaelEischer (changes requested):
I've added quite a few comments that must be addressed

internal/backend/gs/gs.go
tree68505: It's placement is just from my familiarity with C over Go. I'll move the declaration.

internal/backend/gs/gs.go
tree68505: I hadn't considered that. Yeah, the metadata contains a sha of the key, `CustomerKeySHA256`, that can be compared locally. I'll add the sha calculations and base64 conversions to do the key selection locally. I'll make sure to avoid the metadata call if no keys are defined.

internal/backend/gs/config.go
tree68505: Reverted.

internal/backend/gs/config.go
tree68505: Removed.

internal/backend/gs/config.go
tree68505: Renamed.

internal/backend/gs/gs.go
tree68505: Renamed and check error.

internal/backend/gs/gs.go
tree68505: Fixed.

changelog/unreleased/pull-3612
tree68505: Changed.

changelog/unreleased/pull-3612
tree68505: Added.

internal/backend/gs/gs.go
tree68505: Branches are removed with slight redo.

internal/backend/gs/config.go
tree68505: Removed.

tree68505:
I'm not sure how to add unit tests for this without mocking out the GCS calls.

internal/backend/gs/gs.go
tree68505: I tried to define variables closer to where they are used.

internal/backend/gs/gs.go
tree68505: Due to the use of a map in the latest version, I can't lump all keys together.

internal/backend/gs/gs.go
tree68505: Fixed.

internal/backend/gs/gs.go
tree68505: Moved this calculation to the config loading stage.

internal/backend/gs/gs.go
tree68505: Changed to calculate all the SHAs during the load of the config into a map. Then we simply use the string sha found on the object metadata to quickly dereference which loaded key to use to decrypt.

MichaelEischer:
After talking with @fd0 we have decided to _not_ add support for customer-supplied keys. Sorry, for changing course regarding this feature.

As restic backups are already always encrypted at the client-side it is unclear which benefit encryption keys would offer that are sent to the cloud provider along with each request. Another downside is that this PR would add a great deal of code for a probably rarely used feature. As an alternative it might be possible to achieve the same result by using the S3-API for GCS via rclone.

tree68505:
That's a bummer.

GCS prevents downloading of objects without the key which protects from unauthorized or accidental bucket access from getting the files. Plus it's nice to not rely solely on one encryption implementation for long term storage of sensitive files or documents on the cloud.

I'll try to support the patch over on my fork if anyone else wants this feature.

PR #3619, opened 2022-01-04
Avoid choosing parent snapshot newer than time of current snapshot

aneeshusa:
Currently, `restic backup` (if a `--parent` is not provided)
will choose the most recent matching snapshot as the parent snapshot.
This makes sense in the usual case,
where we tag the snapshot-being-created with the current time.

However, this doesn't make sense if the user has passed `--time`
and is currently creating a snapshot older than the latest snapshot.
Instead, choose the most recent snapshot
which is not newer than the snapshot-being-created's timestamp,
to avoid any time travel.

Impetus for this change:
I'm using restic for the first time!
I have a number of existing BTRFS snapshots
I am backing up via restic to serve as my initial set of backups.
I initially `restic backup`'d the most recent snapshot to test,
then started backing up each of the other snapshots.
I noticed in `restic cat snapshot <id>` output
that all the remaining snapshots have the most recent as the parent.

Not AFAIK. I looked through previous issues on GitHub and the forum,
and there are a number of issues which discuss making parent selection smarter in general,
but I did not see any that are specifically about the interaction of parent selection with time.
Since it was a fairly straightforward fix, opted to open a PR directly
but happy to defer to an issue if discussion is needed.

Checklist
---------


aneeshusa:
A few other notes:
- Thank you very much for making `restic`! I just started using it today and it is working a treat.
- I have `go 1.17.5` locally, so `go fmt` emits changes to many files adding the new build tag syntax (see https://go.dev/doc/go1.17#build-lines for docs on this). I omitted those changes from the PR, but wanted to flag it in case it trips up any other developers. Let me know if they should be included in this PR.

doc/man/restic-backup.1
```diff
@@ -91,7 +91,7 @@ Exit status is 3 if some source data could not be read (incomplete snapshot crea
 
 .PP
 \fB\-\-parent\fP=""
-	use this parent \fB\fCsnapshot\fR (default: last snapshot in the repo that has the same target files/directories)
+	use this parent \fB\fCsnapshot\fR (default: last snapshot in the repo that has the same target files/directories, and is not newer than the snapshot time)
```
MichaelEischer, line 94: The man pages are regenerated automatically before a new release. Please remove this change from the PR.

internal/restic/snapshot_find.go
```diff
@@ -13,19 +13,28 @@ import (
...
 			}
 		}
 		absTargets = append(absTargets, filepath.Clean(target))
 	}
+	return absTargets, nil
+}
+
+// FindLatestSnapshot finds latest snapshot with optional target/directory, tags and hostname filters.
+func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string, tagLists []TagList, hostnames []string) (ID, error) {
+	var err error
```
MichaelEischer, line 33: That code line is not necessary, as `err` will already defined automatically in the next line.

internal/restic/snapshot_find.go
```diff
@@ -71,6 +80,58 @@ func FindLatestSnapshot(ctx context.Context, repo Repository, targets []string,
...
+		latestID ID
+		found    bool
+	)
+
+	err = ForAllSnapshots(ctx, repo, nil, func(id ID, snapshot *Snapshot, err error) error {
+		if err != nil {
+			return errors.Errorf("Error loading snapshot %v: %v", id.Str(), err)
+		}
+
+		if snapshot.Time.After(timeStamp) {
```
MichaelEischer, line 102: What about using the zero-value of `time.Time(0)` to disable the time check? Then `FindLatestSnapshot` and `FindLatestSnapshotBefore` can internally be implemented using the same function.

MichaelEischer (changes requested):
I like the idea to only pick older snapshots as parent snapshot. However, regarding the implementation I think it's currently too much code duplication. See the comment for a suggestion.

doc/man/restic-backup.1
aneeshusa: Done, pushed a commit to add a note to CONTRIBUTING.md as well to document this.

internal/restic/snapshot_find.go
aneeshusa: Combined the two functions. Used a `*time.Time` instead of the zero value to make it more clear this is an optional argument since there's no constant for the max `time.Time` and passing `time.Time(0)` as the max snapshot timestamp to disable the behavior seems a bit error-prone. Let me know if you still would prefer `time.Time(0)`.

aneeshusa:
Thank you for the review!

doc/man/restic-backup.1
MichaelEischer: Good idea. Thanks.

MichaelEischer (approved):
LGTM. I've added a small commit to rename the `timestamp` variable to `timestampLimit` to make its meaning a bit clearer from the name.

Issue #3478, opened 2021-08-11
File Created time seens incorrect on Windows repository files


mikef57:
Hello,
I use restic on windows - with great success - thanks.
I am writing a Windows GUI to view the repo snapshots, folders and files and found that the Created and Modified date-time values are 'always' the same. The Modifed date is correct, the Created date is incorrect (different to what Windows shows).
At first I thought it was my program but then I dumped some blobs out using restic and found that the Created date is incorrect.
I can provide more info if you need it.

Thanks again for doing restic - great work.
Mike Fletcher

Xyrio:
what gui tool are you using? (im looking for one too)
what language are you using in windows and wich windows version? (might be a date format related problem)

MichaelEischer:
restic only saves the change, modification and access times. The creation time (windows) or birth time (unix) is not saved. And only the access and modification times are restored. Currently `syscall.UtimesNano` is used to restore the timestamps and it only supports the two previously mentioned timestamps.

mikef57:
Ok, my mistake in not understanding ctime. 
Thanks for the response.
Mike

Issue #1401, opened 2017-10-31
Does restic back up alternate data streams attached to NTFS files?


bherila:
See: https://blogs.msdn.microsoft.com/commandline/2016/11/17/do-not-change-linux-files-using-windows-apps-and-tools/

I have been using restic to back up my Windows PC, and wondering if it will successfully back up my Windows Subsystem for Linux folder from the Windows environment. If the answer is "yes," then I guess nothing needed here :)

But, if the answer is "no" I would want to see if the recommendation is 

* I can try to put together a PR to read/back up this data to the restic repo (some guidance may be helpful/needed)
* Run restic from inside the Windows Subsystem for Linux (would it need to be rebuilt to run in this environment or maybe I can just use the Linux binaries)

Thanks

fd0:
While I don't have any experience with WSL, so I don't know that. What I do know is that restic, at least for now, does not backup or restore NTFS extended attributes and NTFS alternate data streams. On Windows, when run as a native Windows binary, it'll save folders and files (with the "content" of the files).

Adding alternate data streams would require modifying the repository format. And to be honest, I'm not convinced that it's worth the effort.

lloeki:
There is a similar feature to alternate data streams on macOS, called resource forks, which may include e.g compressed data if using HFS Plus Compressed, alternative icons for files...

A similar question can be raised about filesystem extended attributes, which on macOS may include [hiding files via chflags(1)](https://superuser.com/questions/341663/how-do-i-create-a-hidden-file-or-folder-in-os-x) as well as [hiding extensions via SetFile(1)](https://superuser.com/questions/399899/show-hide-extension-of-a-file-through-os-x-command-line#399918). Those are visible using `ls -le@` (`-e` for ACLs, `-@` for xattrs).

timri:
A possible extension to support alternate data streams, as well as other data (acl, permissions, attributes etc.) would be to use the windows backup api (see https://github.com/Microsoft/go-winio/blob/master/backup.go )
This would also enable backing up files without having permission to access them (from an elevated command, so it would be even possible to backup files that even the Administrator can not access, f.e. files only accessible to "LOCAL_SYSTEM")
However, this would need an extension to the repository format...
(I have no programming experience in go, otherwise I would probably hack together a POC. But maybe I try that as a project to learn go ;-) )
Further documentation of the API:
https://msdn.microsoft.com/en-gb/library/windows/desktop/aa362520(v=vs.85).aspx
[...]

PR #3622, opened 2022-01-06
Add support for CreationTime and File Attributes in Windows

### LINKED ISSUE #2075: Directory and file attribute "hidden" not restored
Output of `restic version`
--------------------------
restic 0.9.3 compiled with go1.11.1 on windows/amd64
running in Windows 10

.\restic restore --repo D:\test\ --target '/restore' latest --host XXX --password-file password.txt 

No environmental variables were used in this test.

Local file system - no backend server

Expected behavior
-----------------
All files and directories are restored along with their attribute settings. Specifically - the hidden attribute.

Actual behavior
---------------

All files and directories were restored perfectly except the hidden attribute was lost.  For example a .git directory will be fully restored but was visible in the restore. This behavior is consistent across multiple scenarios.

Create a Windows file and directory marked hidden. Backup and restore then and the hidden attribute is unset.

No

I cannot tell if the hidden attribute is lost in the backup phase or the restore phase.

I think restic is a great tool.

mchowdhury:
File created time (an NTFS specific attribute) and File Attributes don't get backed up or restored on Windows backups. See issue #3478 and #2075

Since we don't want to modify ctimes (because it's used for file change comparison #2212) we store creation time as an extended attribute and restore the creation time value when the file is restored.

This issue was filed in #3478 and #2075
A related issue is #1401 for backing up entire NTFS extended attributes. This uses microsoft provided go-winio library. In the future this would be a better approach but would be a significant re-write. See this comment: https://github.com/restic/restic/issues/1401#issuecomment-392959322

Checklist
---------


DRON-666:
Can you add windows attributes as well? This would solve issue #2075.

mchowdhury:
> Can you add windows attributes as well? This would solve issue #2075.

Ideally it would be nice to have all the missing attributes included. This can be done what's described in https://github.com/restic/restic/issues/1401#issuecomment-392959322

It's more effort but can be done. I would probably need some assistance with the implementation.

DRON-666:
>Ideally it would be nice to have all the missing attributes included.

I fully agree, but adding standard attributes to you PR is just the use of `SetFileAttributes` instead of `SetFileTime`.

>This can be done what's described in #1401 (comment)

I don't think that `BackupRead/Write` API is compatible with `restic` architecture. It does not allow you to choose the type of requested information. Anyway `BackupRead/Write` is just a [userland code with bunch of NT-function calls](https://github.com/KernelPanic-OpenSource/Win2K3_NT_base/blob/master/win32/client/backup.c#L1628) and you don't need to use it.

mchowdhury:
> I fully agree, but adding standard attributes to you PR is just the use of `SetFileAttributes` instead of `SetFileTime`.

This was a good idea. I updated and tested the PR with saving and restoring file attributes in addition to creation time

internal/restic/node_windows.go
```diff
@@ -22,18 +27,104 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
+
+		s, ok := fileinfo.Sys().(*syscall.Win32FileAttributeData)
+		if ok && s != nil {
+			var creationTime bytes.Buffer
+			enc := gob.NewEncoder(&creationTime)
+			
+			if err := enc.Encode(syscall.NsecToTimespec(s.CreationTime.Nanoseconds())); err != nil {
+				return nil, errors.Wrap(err, "Getxattr")
+			}
+			return creationTime.Bytes(), nil
```
DRON-666, line 45: I don' think you need all this conversions and encoding. Simple `binary` conversion is enough:
```          
var creationTime [8]byte
binary.LittleEndian.PutUint32(creationTime[0:4], s.CreationTime.LowDateTime)
binary.LittleEndian.PutUint32(creationTime[4:8], s.CreationTime.HighDateTime)
return creationTime[:], nil
```

internal/restic/node_windows.go
```diff
@@ -22,18 +27,104 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
+	
+		var creationTime syscall.Timespec
+		   dec := gob.NewDecoder(&inputData)
+		
+		if err := dec.Decode(&creationTime); err != nil {
+			return errors.Wrap(err, "Setxattr")
+		}
+		   
+		c := syscall.NsecToFiletime(time.Unix(creationTime.Unix()).UnixNano())
+		if err := syscall.SetFileTime(h, &c, nil, nil); err != nil {
```
DRON-666, line 110: Same as above:
```
var creationTime syscall.Filetime
creationTime.LowDateTime = binary.LittleEndian.Uint32(data[0:7])
creationTime.HighDateTime = binary.LittleEndian.Uint32(data[4:8])
if err := syscall.SetFileTime(h, &creationTime, nil, nil); err != nil {
```

internal/restic/node_windows.go
```diff
@@ -22,18 +27,104 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
+			return errors.Wrap(err, "Setxattr")
+		}
+		return nil
+	case "FileAttributes":
+		var attrs uint32
+		inputData := bytes.NewBuffer(data)
+		
+		if err := binary.Read(inputData, binary.LittleEndian, &attrs); err != nil {
+			return errors.Wrap(err, "Setxattr")
+		}
```
DRON-666, line 120: And here too:
```
attrs := binary.LittleEndian.Uint32(data)
```

mchowdhury:
@DRON-666 Thanks again for the feedback. I made the necessary changes and tested them.
Let me know if you have any other suggestions.

mchowdhury:
@DRON-666 What else needs to be done to get this merged?

hgraeber:
It would be nice to add security information the same. Would mean to use GetFileSecurity and SetFileSecurity the get and set the DACL. Converting the descriptor to/from SDDL string representation would be even better. ConvertSecurityDescriptorToStringSecurityDescriptor and ConvertSecurityDescriptorFromStringSecurityDescriptor are helpful for that.

Maybe you consider to use a readable representation for the other attributes, too.

mchowdhury:
> It would be nice to add security information the same. Would mean to use GetFileSecurity and SetFileSecurity the get and set the DACL. Converting the descriptor to/from SDDL string representation would be even better. ConvertSecurityDescriptorToStringSecurityDescriptor and ConvertSecurityDescriptorFromStringSecurityDescriptor are helpful for that.
> 
> Maybe you consider to use a readable representation for the other attributes, too.

@hgraeber GetFileSecurity and SetFileSecurity are not currently supported by the standard go syscall library. I'm open to other suggestions. Otherwise, should we make a separate issue for file security attributes and get this merged?

MichaelEischer (changes requested):
Mapping the creationTime / fileAttributes in to xattrs feels wrong, please just add new (optional) fields to the `Node` struct. I'd also prefer to store the creationTime as a regular timestamp and not some Windows specific timestamp format. Btw, the creationTime is known as birthTime on Linux, so it might be more appropriate to use that name (or short `btime` for the json attribute). Even though we can't write that timestamp on Linux.

The FileAttributes should also be accessible from the struct returned at https://github.com/restic/restic/blob/cccc17e4e91facdaf710dc82a59653c5c0de3156/internal/restic/node_windows.go#L56 . Adding two redundant syscalls (the two os.Stat) calls will likely hurt performance.

Judging from https://learn.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants I'm not sure whether it's a good idea to blindly store and restore arbitrary file attributes. But I'm not familiar enough with the Windows API to accurately judge that.

Restoring access permissions etc. definitely doesn't belong into this PR.

mchowdhury:
@MichaelEischer I can move the creation time directly to `ctim()` since we already make the fileattributes syscall in the struct.


For Windows file attributes, these are essentially the same as Linux extended attributes and should be backed up and restored. I figured adding it here would be the cleanest way.

How do you propose we integrate and restore them if they are added as optional attributes to the node struct?

hgraeber:
> Judging from https://learn.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants I'm not sure whether it's a good idea to blindly store and restore arbitrary file attributes. 

I think the file attributes shall not be copied blindly, because they may have some side effect and cannot be simply restored. (https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileattributesa). It is necessary to set flags on file creation or using special functions to reach the desired effect.

hgraeber:
@mchowdhury

> GetFileSecurity and SetFileSecurity are not currently supported by the standard go syscall library. I'm open to other suggestions. Otherwise, should we make a separate issue for file security attributes and get this merged?

Agreed. I will write an issue for this. There we can discuss how to implement backup of windows file security information.

mchowdhury:
@hgraeber I don't quite understand the approach you are describing, and unsure how to implement it without impacting other platforms (node.go). Should creationTime follow the same approach? If not, we can separate file attributes to a separate PR and just focus on creationTime.

Would be happy to have your help on the implementation if you have bandwidth.

ProactiveServices:
Have you considered the more unusual attributes and what happens if they cannot be restored, or what might unexpected behaviour might occur if they are restored? There's a host of weird and wonderful attributes other than the usual hidden/read-only/system. How is the restoration of a sparse file handled? What about attributes that only kernel-level code can restore? Will a restore apply to a file marked as temporary? How are you handling EFS-encrypted data? Files that have attributes that are not supported on the target filesystem?

mchowdhury:
@MichaelEischer After more research, I like your suggestion with `btime` adding it to the node, also since darwin also supports btime. I will make a new PR to support creationTime on windows and os x.  

Let's leave this PR open to discuss File Attributes on Windows. File attributes got pulled into this PR from @DRON-666 suggestion to fix issue https://github.com/restic/restic/issues/2075.

heapxor:
hello,
is that merge request stable and tested? thanks

hanacchi:
any progress on getting this merged? i would love to use restic but lack of creation time support on windows is holding me back

Dobatymo:
A file copy using the Windows Explorer doesn't preserve file creation dates (only modification dates). So I don't think restic should restore them either.

tszynalski:
> A file copy using the Windows Explorer doesn't preserve file creation dates (only modification dates). So I am not sure restic should restore them either.

Copying a file is not the same as restoring a file from backup. When you copy a file, you create a second file with the same contents – but it's not the _same_ file, so the creation date should be different. When you back up a file, you want to be able to "turn back time" and go back _to the way things were_ in the event of failure. If something is lost (such as a piece of metadata), then the restore was not complete.

MichaelEischer:
After taking a look at https://github.com/restic/restic/pull/4611 , I think we should follow that approach. The `GenericAttributes` suggested there just look like what I had in mind.

Sorry for not finding time to discuss this PR!

PR #3634, opened 2022-01-19
Added security note for passing pw via env variable

JsBergbau:
It adds a security note that environment variables can be seen by other processes running as the same user as restic does.

It is only a small documentation update. Feel free to change the wording. 
I was aware that passwords via commandline are unsafe because they can be read by other processes easily and thats why I used environment variables. But these aren't so secure as thought. So I've created this pull request.

No.

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -526,7 +526,7 @@ environment variables. The following lists these environment variables:
     RESTIC_REPOSITORY_FILE              Name of file containing the repository location (replaces --repository-file)
     RESTIC_REPOSITORY                   Location of repository (replaces -r)
     RESTIC_PASSWORD_FILE                Location of password file (replaces --password-file)
-    RESTIC_PASSWORD                     The actual password for the repository
+    RESTIC_PASSWORD                     The actual password for the repository (note security notice below)
     RESTIC_PASSWORD_COMMAND             Command printing the password for the repository to stdout
     RESTIC_KEY_HINT                     ID of key to try decrypting first, before other keys
     RESTIC_CACHE_DIR                    Location of the cache directory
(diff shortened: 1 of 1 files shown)


MichaelEischer:
> I was aware that passwords via commandline are unsafe because they can be read by other processes easily and thats why I used environment variables.

The problem with passing the password via the command line is that _any_ user on the system can read the command line and thus the password is leaked. When passing the password via environment variables, then only the current user can extract it. As the standard security boundaries in unix only separate users from each other but not processes of the same user, that is not really unexpected. The most obvious problem with environment variables is that these are passed on to subprocesses if any are launched by restic.

When stating that something is supposed to be insecure, it is a good idea to suggest a better solution. `restic -p password_file` won't help here, as some other process executing as the current user could just read the password file as well. `restic --password-command "get-pass restic"` could help when running it interactively, but for an automated backup there's nothing preventing something else to call `get-pass restic`.

That is, I currently don't understand where precisely the security benefit would be.

rawtaz:
There's also the aspect that the restic documentation cannot be expected to document every security aspect people should know about - it's outside of its concerns. Some things we can add as an extra precaution, but how a Linux or similar system works is simply out of scope.

In this case as Michael said the security separation is the user, which already has access to the information needed to start restic anyway. So it's rather moot in its entirety. I'm inclined that we close this without merging. Good point with the "suggesting a better solution".

JsBergbau:
You're right, the user that is running restic backup can also access the password when supplied via `RESTIC_PASSWORD_FILE`. 

Suppose debug logs of other processes that contain all processes including their environment variables. I don't know how much other software is around that includes environment variables of all processes but still for me it feels more safe to supply password from a password file than via environment variable.

rawtaz:
I'll close this then :)

JsBergbau:
Just for the record. With ProcessHacker https://processhacker.sourceforge.io/ it is possible to view environment variables of the processes like `cat /proc/`pidof restic`/environ`` does in linux.

PR #3637, opened 2022-01-27
[issue 2449] fix problems with trailing slash...

### LINKED ISSUE #2449: Problem with trailing slash and spaces in repository directory on Windows
Output of `restic version`
--------------------------

`restic -r 'C:\Restic Repo\' init`
`restic -r 'C:\restic repo\' backup C:\restic backup\'`

local

Expected behavior
-----------------

Normal and functional initialization and backup.

Actual behavior
---------------

The help is shown without an error, as if just `restic` was run.

Use Windows (10 Pro x64) and cmdlet or PowerShell.
Have a repository directory with a space in it.
Use the mentioned commands with a trailing slash in the path

With PowerShell the tab autocomplete adds a trailing slash by default, which is how I discovered the issue in the first place. The cmdlet doesn't add the trailing slash, avoiding the issue.

Trailing slashes in directories and paths marked for backup do not cause the issue.

No.

No.

restic makes me happy everytime I use it. That's why I enjoy reporting issues and help improve it.

winks:
... in path names on windows, which in turn will escape the quotes

See #2449 - if you're autocompleting a path with spaces in it, e.g. `c:\test thing\` then it can't be read because the backslash escapes the trailing quote.

Closes #2449

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2449
+++ b/changelog/unreleased/issue-2449
@@ -1,0 +1,7 @@
+Bugfix: Trailing slashes in directory names on windows
+
+If you were autocompleting a path with spaces in it, e.g. `c:\test thing\`
+then restic couldn't read it because the backslash escaped the trailing quote.
+
+https://github.com/restic/restic/issues/2449
+https://github.com/restic/restic/pull/3637
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -455,6 +455,19 @@ func collectTargets(opts BackupOptions, args []string) (targets []string, err er
 		targets = append(targets, fromfile...)
 	}
 
+	// fix dangling trailing quotes on windows
+	if runtime.GOOS == "windows" {
+		argsFixed := []string{}
+		for _, a := range args {
+			if strings.HasSuffix(a, "\"") && !strings.HasPrefix(a, "\"") {
+				argsFixed = append(argsFixed, strings.TrimRight(a, "\""))
+			} else {
+				argsFixed = append(argsFixed, a)
+			}
+		}
+		args = argsFixed
+	}
+
 	// Merge args into files-from so we can reuse the normal args checks
 	// and have the ability to use both files-from and args at the same time.
 	targets = append(targets, args...)
--- a/cmd/restic/cmd_backup_test.go
+++ b/cmd/restic/cmd_backup_test.go
@@ -111,3 +111,43 @@ func TestReadFilenamesRaw(t *testing.T) {
 	rtest.Assert(t, strings.Contains(err.Error(), "zero byte"),
 		"wrong error message: %v", err.Error())
 }
+
+// the rules for what constitues a valid filename/dirname
+// is different on windows, for example no trailing backslash
+// or no traling double quote
+func TestCollectTargetsRegression(t *testing.T) {
+	dir, cleanup := rtest.TempDir(t)
+	defer cleanup()
+
+	realDirs := []string{"cmdline arg1"}
+	if runtime.GOOS == "windows" {
+		realDirs = append(realDirs, "cmdline arg windows")
+	} else {
+		realDirs = append(realDirs, "cmdline arg nonwindows1\\")
+		realDirs = append(realDirs, "cmdline arg nonwindows2\"")
+	}
+	var expect []string
+	for _, dirname := range realDirs {
+		// All mentioned files/dirs must exist for collectTargets.
+		d := filepath.Join(dir, dirname)
+		err := os.Mkdir(d, 0755)
+		rtest.OK(t, err)
+
+		expect = append(expect, d)
+	}
+
+	realArgs := []string{filepath.Join(dir, "cmdline arg1")}
+	if runtime.GOOS == "windows" {
+		realArgs = append(realArgs, filepath.Join(dir, "cmdline arg windows\""))
+	} else {
+		realArgs = append(realArgs, filepath.Join(dir, "cmdline arg nonwindows1\\"))
+		realArgs = append(realArgs, filepath.Join(dir, "cmdline arg nonwindows2\""))
+	}
+
+	opts := BackupOptions{}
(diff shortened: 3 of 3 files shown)


winks:
Small addendum: While I think this fixes the linked issue, the handling of `args` and especially the [lack of tests](https://github.com/restic/restic/blob/fbb0e6499ac8212b26a7a23f9fe416e2c9fe6528/cmd/restic/cmd_backup.go#L460) (although I only looked in `cmd_backup_test.go` isn't too great, so I wouldn't say I am 100% confident.

MichaelEischer:
I'm not sure this is a good idea. Isn't it possible to create a filename that ends with a quote? This rather sounds like a bug in the autocompletion of powershell / cmd (?). Or somewhere in the command line parsing.

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_special_characters?view=powershell-7.2 looks like a backslash is not an escape code in powershell, so it shouldn't cause a problem. Before adding a workaround we should first investigate what the actual problem i.

Mo-Gul:
@MichaelEischer, no in Windows filenames containing quotes are forbidden (as some more characters) see e.g. <https://en.wikipedia.org/w/index.php?title=Filename&oldid=1063519062#Comparison_of_filename_limitations> for FAT and NTFS

MichaelEischer:
I just did a bit of testing, and it seems that the command line handling of trailing slashes is far more broken:
```
PS C:\Users\Michael> .\restic.exe -r testrepo backup 'test space\' "test space\" ".\test space\" -v
repository 2b35bc1d opened successfully, password is correct
error: CreateFile \\?\C:\Users\Michael\space": The filename, directory name, or volume label syntax is incorrect.
error: CreateFile \\?\C:\Users\Michael\test space" -v: The filename, directory name, or volume label syntax is incorrect.
error: CreateFile \\?\C:\Users\Michael\test space" test: The filename, directory name, or volume label syntax is incorrect.
Fatal: unable to save snapshot: snapshot is empty
PS C:\Users\Michael>
```

To clean this up a bit, restic got the following three paths to backup: `test space" test`, `space"` and `test space" -v`. This essentially means that just removing a trailing quote will randomly work in some situations, but will still break horribly in most situations.

The above command line was passed to restic as `"C:\Users\Michael\restic.exe" -r testrepo backup "test space\" "test space\" ".\test space\" -v` (as retrieved via `windows.UTF16PtrToString(syscall.GetCommandLine())`). The command line parsing in Go happens here:
https://github.com/golang/go/blob/85b5f86584686677c554b2538a7edee96d684aea/src/os/exec_windows.go#L156 .
(cmd turns the above command line into ` .\restic.exe -r testrepo backup 'test space\' "test space\" ".\test space\" -v` which is also broken)

The way I understand https://docs.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments?view=msvc-170 Go is behaving correctly here, which would mean that Powershell messed up the parameter encoding.

winks:
Good catch, thanks. Yeah to be honest I haven't got a lot of experience with PS and its brokenness as I tend to stay away if possible.

The interesting questions are thus: Is it worth fixing at all? Would a very basic check including "we detected some wonky backslashes please double check" warning be better?

As I said in #2449 originally: "The most puzzling thing is is that is actually creates the snapshot with 0 files." - that's the thing that's actually annoying me. "Don't tab-complete" is a perfectly fine solution for this, but I think that one's important (and maybe my bad for not splitting it out into a separate issue)

MichaelEischer:
@winks Does the empty snapshot problem still exist with restic 0.13.0? I'm a bit puzzled why my test did not create a snapshot.

winks:
No change between 0.12.1 and 0.13.1 for me.

```
# this is 0.13.1
PS F:\7zip> C:\Apps\bin\restic.exe -r .\backup\ init
enter password for new repository:
enter password again:
created restic repository c46255b814 at .\backup\

irrecoverably lost.
PS F:\7zip> C:\Apps\bin\restic.exe -r .\backup\ backup '.\New folder\'
repository c46255b8 opened successfully, password is correct
created new cache in C:\Users\Florian\AppData\Local\restic
found 1 old cache directories in C:\Users\Florian\AppData\Local\restic, run `restic cache --cleanup` to remove them
error: CreateFile \\?\F:\7zip\New folder": The filename, directory name, or volume label syntax is incorrect.
Fatal: unable to save snapshot: snapshot is empty


PS F:\7zip> C:\Apps\bin\restic_0.12.1.exe -r .\backupold\ init
enter password for new repository:
enter password again:
created restic repository d13cdf131e at .\backupold\

irrecoverably lost.
PS F:\7zip> C:\Apps\bin\restic_0.12.1.exe -r .\backupold\ backup '.\New folder\'
repository d13cdf13 opened successfully, password is correct
created new cache in C:\Users\Florian\AppData\Local\restic
found 1 old cache directories in C:\Users\Florian\AppData\Local\restic, run `restic cache --cleanup` to remove them
error: CreateFile \\?\F:\7zip\New folder": The filename, directory name, or volume label syntax is incorrect.
Fatal: unable to save snapshot: snapshot is empty
```

That is only with powershell. Both work fine when being run in cmd.exe

MichaelEischer:
Now I'm confused. The problem in #2449 seems to be that an essentially empty snapshot is created. But in the example restic properly complains about a broken filename and does not create a snapshot.

I guess we could tell users not to use slashes before a quote on windows if a file cannot be read. Although it's a somewhat hacky solution.

winks:
Yeah sorry, rereading the output of the other one I am now also a little confused. I am reasonably sure that this 0.12.1 mentioned here and in the other issue on Jan 22 is the same binary. I didn't investigate further and just renamed it when I tested 0.13.1 - so it looks like one time it failed and one time it created an empty snapshot...

MichaelEischer:
From what I can tell, there's no (reasonable) way to properly fix the path parsing in restic. So, my suggestion would be to add a section to the FAQ (https://restic.readthedocs.io/en/stable/faq.html) instead.

Issue #3643, opened 2022-02-05
Store large extended attributes as data blobs


MichaelEischer:
Output of `restic version`
--------------------------
0.12.1

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
Large extended attributes can lead to bloated directory metadata (several hundred MBs), see https://forum.restic.net/t/tracking-down-large-metadata-changes/3850/9 for an example. Restic should store large extended attributes just like regular files. That way these can be deduplicated properly.

What are you trying to do? What problem would this solve?
---------------------------------------------------------
Backup the home folder (including `~/Library/Containers`) on macOS. The large directory metadata leads to two problems: the restic cache grows very quickly and the oversized tree blobs cause restic to require lots of memory for example during prune.

aawsome:
I think a better solution would be to chunk the JSON encoded Tree into blobs using the same chunker as already done with file contents. This would need a change in the Node data structure for dirs: instead of storing exactly one subtree ID we would need to save a slice of IDs (we could use "content" for it - as already used for file chunks ).

Another and consequent improvement would be to change the JSON representation of a Tree to be just a JSON stream of Nodes. Then when processing a dir, the subtree IDs could be used to stream blobs into a JSON decoder which itself streams Nodes. And memory requirements would be bounded by the chunk size plus the memory representation (including the metadata) of only *one* Node.

That approach is similar to #2532 but IMO it is more consequent to use the chunker for splitting large Tree objects. So it would not only help with trees containing a large number of entries, but also with large metadata and even allow deduplication of parts of trees.

MichaelEischer:
I just ran a small test: I've extracted the tree of the `Containers` directory from two different weeks (both are 286MB in size). Then I've backed up both files:

```
$ restic backup containers1 -v
open repository
repository 0ff96851 opened successfully, password is correct
lock repository
load index files
no parent snapshot found, will read all files
start scan on [containers1]
start backup on [containers1]
scan finished in 0.204s: 1 files, 285.343 MiB

Files:           1 new,     0 changed,     0 unmodified
Dirs:            0 new,     0 changed,     0 unmodified
Data Blobs:    148 new
Tree Blobs:      1 new
Added to the repo: 235.542 MiB

processed 1 files, 285.343 MiB in 0:00
snapshot 05930785 saved
$ restic backup containers2 -v
open repository
repository 0ff96851 opened successfully, password is correct
lock repository
load index files
no parent snapshot found, will read all files
[...]

PR #3654, opened 2022-02-19
Limit number of large tree blobs loaded in parallel by StreamTrees

### LINKED ISSUE #3650: prune: "searching used packs" without progress indicator
Output of `restic version`
--------------------------

0.12.1

`restic -r my_backup prune`

local filesystem

Expected behavior
-----------------

I expected to see a progress indicator and progress bar at all times.

Actual behavior
---------------

I got this:
```
repository 1234abcd opened successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 242 snapshots
<...>
[225:18:40] 99.17%  240 / 242 snapshots
[226:30:09] 99.59%  241 / 242 snapshots
[226:35:22] 100.00%  242 / 242 snapshots
searching used packs...
<no progress indicator here - it is "hanging" for 12+ hours now, and I don't know if it does any useful work>
```

Run the prune command on a reasonable large repository, possibly on a slow machine or with limited RAM to cause swapping that slows down everything.

I'm logging this as a bug, because clearly in the earlier steps of the prune command, there is a reasonable progress indication, but in the last step there is none (or is it even the last step?).

No.

Give more direct progress feedback to the user, similar as in the previous step for "finding data that is still in use" that counted through one snapshot at a time for all 242 snapshots.

Yes! But sometimes it is not the fastest...

MichaelEischer:
Load tree blobs with more than 50MB only from a single goroutine. Very large tree blobs with for example 400 MB size can otherwise require roughly 1GB * streamTreeParallelism memory.

Most repository don't have such large tree blobs, however, due to #3643 these are pretty common on macOS. When running prune this can require 5GB+ of memory which will cause swapping on many machines. Other affected commands are `check`, `check` and `stat` in raw mode.

Limiting the parallelism for such large blobs makes sure that the majority of blobs can still be loaded concurrently, but also that the oversized tree blobs don't increase the memory usage too far.

Fixes #3650.

Checklist
---------

### DIFF
--- a/internal/restic/find.go
+++ b/internal/restic/find.go
@@ -11,6 +11,7 @@ import (
 // TreeLoader loads a tree from a repository.
 type TreeLoader interface {
 	LoadTree(context.Context, ID) (*Tree, error)
+	LookupBlobSize(id ID, tpe BlobType) (uint, bool)
 }
 
 // FindUsedBlobs traverses the tree ID and adds all seen blobs (trees and data
--- a/internal/restic/find_test.go
+++ b/internal/restic/find_test.go
@@ -166,6 +166,10 @@ func (r ForbiddenRepo) LoadTree(ctx context.Context, id restic.ID) (*restic.Tree
 	return nil, errors.New("should not be called")
 }
 
+func (r ForbiddenRepo) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bool) {
+	return 0, false
+}
+
 func TestFindUsedBlobsSkipsSeenBlobs(t *testing.T) {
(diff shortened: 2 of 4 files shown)


MichaelEischer:
I've tested this PR with one of my repositories which also has quite a few of these oversized tree blobs. There the snapshots traversal of `prune` went from 9 minutes with more than 5GB peak memory usage (without the PR) to 14 minutes with 2GB peak memory usage (with this PR). The repository has a 240MB index and the repository contains 62 pack files with approx. 300 MB size.

fd0 (approved):
LGTM!

PR #3655, opened 2022-02-19
[WIP] Control concurrent tree loading's memory use

greatroar:
This installs a semaphore to control the max. memory use of concurrent LoadTree calls. It's a simple, local hack. I think it can be extended to a more general solution by putting a semaphore on repository.Repository to soft-cap all of its blob allocations.

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/minio/sha256-simd"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


greatroar:
I just realized that this doesn't work as advertised because the Tree object survives beyond the Release of the semaphore. It'll have to be installed in StreamTrees.

greatroar:
In fact, I don't think this is going to work unless each restic.Tree remembers its originated semaphore and all users explicitly release their trees. Dropping in favor of #3654.

PR #3584, opened 2021-11-27
Update "forget" documentation (append-only mode security considerations, add references)

lgommans:
As discussed in #3498.

Removing data based on a policy when the attacker had the opportunity to add data to your repository comes with some considerations. This is added to the `060_forget.rst` documentation.

That document is also updated to reflect that restic now considers the current system time while running `forget` (it changed in 0.12.1, more specifically #3416).

References to the security considerations section are added:

- In `restic forget --help`
- In the threat model (`design.rst`)
- In the (`030_*`) setup section where an append-only setup is referenced

A reference is also to be added to the `rest-server` readme's append-only paragraph (see [this commit in my fork](https://github.com/lgommans/rest-server/commit/d4c3233e92aee997b98ec947c4b4c8dec99ab3c0) for a draft).

This commit also resolves a typo (amount->number for countable noun), changes a password length recommendation into the metric that actually matters when creating passwords (entropy) since I was editing these doc files anyway, and updates the outdated copyright year in `conf.py`.

Some wording in `060_forget` (line 21..22) is changed to clarify what "forget" and "prune" do, to avoid making it sound like `forget` does not remove any data.

One open question:

- Is it possible to add links within the documentation? There are two places where I think it would be helpful to link directly to the added paragraph, but I don't see intra-documentation links elsewhere. The `:ref:` syntax is only within the same file and I'm not sure if `:doc:` is enabled or if it's the right syntax.

Checklist
---------

- [x] Build the docs to see if it all renders correctly
- [x] Resolve the two TODOs regarding adding links to the relevant page and section

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- n/a I have added tests for all code changes.
- n/a I have added documentation for relevant changes (in the manual). *[Not applicable I think, unless there is a changelog for documentation]*
- n/a There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- n/a I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3656, opened 2022-02-19
forget: Update docs for readability and append-only considerations

lgommans:
This is a follow-up to #3584 -- all info is there :)

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -14,14 +14,13 @@
 Removing backup snapshots
 #########################
 
-All backup space is finite, so restic allows removing old snapshots.
-This can be done either manually (by specifying a snapshot ID to remove)
-or by using a policy that describes which snapshots to forget. For all
-remove operations, two commands need to be called in sequence:
-``forget`` to remove a snapshot and ``prune`` to remove the remaining
-data that was only referenced by this snapshot from the repository. This can
-be automated with the ``--prune`` option of the ``forget`` command,
-which runs ``prune`` automatically if snapshots have been removed.
+All backup space is finite, so restic allows removing old snapshots. This can
+be done either manually (by specifying a snapshot ID to remove) or by using a
+policy that describes which snapshots to forget. For all remove operations, two
+commands need to be called in sequence: ``forget`` to remove snapshots, and
+``prune`` to remove the remaining data that was referenced only by the removed
+snapshots. The latter can be automated with the ``--prune`` option of ``forget``,
+which runs ``prune`` automatically if any snapshots were actually removed.
 
 Pruning snapshots can be a time-consuming process, depending on the
 number of snapshots and data to process. During a prune operation, the
@@ -164,8 +163,8 @@ The most important command-line parameter here is ``--dry-run`` which
 instructs restic to not remove anything but print which snapshots would
 be removed.
 
-.. note:: If you use append-only backups, some best practices apply.
-    Please refer to the security considerations below.
+.. note:: If you use an append-only repository, some best practices apply.
+    Please refer to the security considerations below for more information.
 
 When ``forget`` is run with a policy, restic loads the list of all
 snapshots, then groups these by host name and list of directories. The grouping
@@ -175,27 +174,27 @@ snapshots separately. This is a safety feature.
 
 The ``forget`` command accepts the following parameters:
 
--  ``--keep-last n`` never delete the ``n`` last (most recent) snapshots
--  ``--keep-hourly n`` for the last ``n`` hours in which a snapshot was
-   made, keep only the last snapshot for each hour.
+-  ``--keep-last n`` keep the ``n`` last (most recent) snapshots.
+-  ``--keep-hourly n`` for the last ``n`` hours which have one or more
+   snapshots, keep only the most recent one for each hour.
 -  ``--keep-daily n`` for the last ``n`` days which have one or more
-   snapshots, only keep the last one for that day.
+   snapshots, keep only the most recent one for that day.
 -  ``--keep-weekly n`` for the last ``n`` weeks which have one or more
-   snapshots, only keep the last one for that week.
+   snapshots, keep only the most recent one for that week.
 -  ``--keep-monthly n`` for the last ``n`` months which have one or more
-   snapshots, only keep the last one for that month.
+   snapshots, keep only the most recent one for that month.
 -  ``--keep-yearly n`` for the last ``n`` years which have one or more
-   snapshots, only keep the last one for that year.
+   snapshots, keep only the most recent one for that year.
 -  ``--keep-tag`` keep all snapshots which have all tags specified by
    this option (can be specified multiple times).
--  ``--keep-within duration`` keep all snapshots which have been made within
-   the duration of the latest snapshot. ``duration`` needs to be a number of
-   years, months, days, and hours, e.g. ``2y5m7d3h`` will keep all snapshots
-   made in the two years, five months, seven days, and three hours before the
-   latest snapshot.
+-  ``--keep-within duration`` keep all snapshots having a timestamp within
+   the specified duration of the latest snapshot, where ``duration`` is a
+   number of years, months, days, and hours. E.g. ``2y5m7d3h`` will keep all
+   snapshots made in the two years, five months, seven days, and three hours
+   before the latest (most recent) snapshot.
 -  ``--keep-within-hourly duration`` keep all hourly snapshots made within
-   specified duration of the latest snapshot. The duration is specified in 
-   the same way as for ``--keep-within`` and the method for determining
+   the specified duration of the latest snapshot. The ``duration`` is specified
+   in the same way as for ``--keep-within`` and the method for determining
    hourly snapshots is the same as for ``--keep-hourly``.
 -  ``--keep-within-daily duration`` keep all daily snapshots made within
    specified duration of the latest snapshot.
@@ -206,11 +205,13 @@ The ``forget`` command accepts the following parameters:
 -  ``--keep-within-yearly duration`` keep all yearly snapshots made within
    specified duration of the latest snapshot.
 
-.. note:: All calendar related ``--keep-*`` options work on the natural time
+.. note:: All calendar related options (``--keep-*``) work on the natural time
     boundaries and not relative to when you run the ``forget`` command. Weeks
-    are Monday 00:00 -> Sunday 23:59, days 00:00 to 23:59, hours :00 to :59, etc.
-    Snapshots seemingly made in the future (relative to when you run the
-    ``forget`` command) will be ignored and never removed.
+    are Monday 00:00 to Sunday 23:59, days 00:00 to 23:59, hours :00 to :59, etc.
+
+.. note:: All duration related options (``--keep-within`` and ``--keep-within-*``)
+    ignore snapshots with a timestamp in the future (relative to when the
+    ``forget`` command is run) and these snapshots will hence not be removed.
 
 .. note:: Specifying ``--keep-tag ''`` will match untagged snapshots only.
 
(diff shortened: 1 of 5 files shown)


rawtaz:
Thank you @lgommans for your patience and for helping me fix my accidental PR messup.

So, I started reviewing your changes and fould a few things I wanted to change, for example:

- Some sentences or explanations I found a bit hard to understand some details of, and also if I try to put myself behind the eyes of a non-technical or accustomed reader I felt they needed to be made clearer.
- A couple of lines or changes I felt were too out of context or too meta for this documentation (e.g. a reference to an issue tracking re-encrypting a repository).
- Append-mode in general wasn't very well explained overall in the documentation (since before your changes), so that was one of the things I wanted to improve further.

While reviewing the thread model we also found a few things that weren't entirely accurate or complete (this had little to do with your edits, was mostly existing text), so I changed those too.

So overall I've made what comes across as a lot of editorial and grammatical changes. I hope you don't mind too much (this is why I put it in a second commit). I sincerely do appreciate that you took the time to write the text that you did, because the more I worked with this PR, the more important I felt that this addition to the documentation is.

I'd be happy to hear your thoughts if you want to review my additional changes. If you want to view them specifically you can check my commit on the Commits tab, otherwise to see the end result you can use the "Files changed" tab. Also looking forward to Michael's or anyone else's review.

rawtaz:
@lgommans Hm, could you update https://restic-test-for-merge-request.readthedocs.io/en/latest/ so that it points to this new `forget-security` branch instead of your master branch? Then we can see the changes there.

lgommans:
Thanks for all the work on this! That all makes a lot of sense to change and add. I kind of trust it'll be fine, but before I spot some typo after the fact I guess it makes sense that I give it a read before merging. I'll do that tomorrow, probably before ~20:00 (GMT+1).

lgommans:
Ah I forgot to mention: the docs were built https://restic-test-for-merge-request.readthedocs.io/en/latest/

Couldn't find how to make it pull the list of branches in order to 'activate' another branch for building, so I used the element inspector to change the value for "latest" from `master` to `forget-security` (funny name, btw, not sure if that was chosen on purpose :D). It broke the admin page (showing `-----` instead of `forget-security`) but on the build page I could now make it pull that branch and build it :)

rawtaz:
> I used the element inspector to change the value for "latest" from master to forget-security

I'm just curious, which element did you change?

> funny name, btw, not sure if that was chosen on purpose :D

Haha, it actually wasn't at all intentional, but it's really funny now that you point it out yeah!

doc/060_forget.rst
```diff
@@ -312,11 +318,12 @@ four Sundays, but remove the rest:
...
    8 snapshots
 
-The result of the ``forget --keep-daily`` operation does not depend on when it
-is run, it will only count the days for which a snapshot exists. This is a
-safety feature: it prevents restic from removing snapshots when no new ones are
-created. Otherwise, running ``forget --keep-daily 4`` on a Friday (without any
-snapshot Monday to Thursday) would remove all snapshots!
+The result of the ``forget --keep-daily`` operation only partially depends on
+when it is run; it will only count the days for which a snapshot exists,
+although snapshots with a `time` lying in the future are ignored and never
```
MichaelEischer, line 323: Future timestamps are only ignored by the `--keep-within*` options. And the addition of the remark here causes the next sentence "This is a safety feature" to be unclear what it refers to.

doc/design.rst
```diff
@@ -629,11 +629,11 @@ system making backups could:
...
    copy files, manipulate data).
 -  Create snapshots (containing garbage data) which cover all modified files
-   and wait until a trusted host has used forget often enough to forget all
+   and wait until a trusted host has used ``forget`` often enough to remove all
    correct snapshots.
--  Create a garbage snapshot for every existing snapshot with a slightly different
-   timestamp and wait until forget has run, thereby removing all correct
-   snapshots at once.
+-  Create a garbage snapshot for every existing snapshot with a slightly
+   different timestamp and wait until certain ``forget`` configurations has been
```
MichaelEischer, line 635: s/has/have/

doc/design.rst
```diff
@@ -645,21 +645,26 @@ An adversary with write access to your files at the storage location could:
...
+   back: paying the ransom. For safe use of ``forget``, please see the
+   corresponding documentation on removing backup snapshots and append-only mode.
+
+An adversary who has a leaked (decrypted) key for a repository could:
+
+-  Decrypt existing and future backup data. If multiple hosts backup into the
+   same repository, an attacker will get access to the backup data of every host.
+   Note that since the local encryption key gives access to the master key, a
+   password change will not prevent this. Changing the master key can currently
+   be done using the ``copy`` command, which moves the data into a new repository
```
MichaelEischer, line 669: I'd add "only" before "be done using ...".

doc/060_forget.rst
```diff
@@ -336,6 +343,50 @@ could specify:
...
+only be written to and read from, while delete and overwrite operations are
+denied. Restic's `rest-server`_ features the append-only mode, but few other
+standard backends do. To support append-only with such a backend, one can use
+`rclone`_ as a complement in between the backup client and the backend service.
+
+.. _rest-server: https://github.com/restic/rest-server/
+.. _rclone: https://rclone.org/commands/rclone_serve_restic/
+
+To remove snapshots and recover the corresponding disk space, the ``forget``
+and ``prune`` commands must have full read, write and delete access to the
```
MichaelEischer, line 361: s/must have/require/

doc/060_forget.rst
```diff
@@ -336,6 +343,50 @@ could specify:
...
+repository in a so-called append-only mode. This means that the repository can
+only be written to and read from, while delete and overwrite operations are
+denied. Restic's `rest-server`_ features the append-only mode, but few other
+standard backends do. To support append-only with such a backend, one can use
+`rclone`_ as a complement in between the backup client and the backend service.
+
+.. _rest-server: https://github.com/restic/rest-server/
+.. _rclone: https://rclone.org/commands/rclone_serve_restic/
+
+To remove snapshots and recover the corresponding disk space, the ``forget``
```
MichaelEischer, line 360: IMO this paragraph should also explain the append-only setup variant where a trusted host is allowed to remove data from the repository. Without such an explanation, the "trusted client" in the next paragraph is not properly defined.

doc/060_forget.rst
```diff
@@ -336,6 +343,50 @@ could specify:
...
+
+However, even with append-only mode active, an attacker who is able to add
+additional and empty or otherwise useless snapshots to the repository can
+potentially cause a situation where a trusted client running ``forget`` with
+certain ``--keep-*`` options might unknowingly remove legitimate snapshots,
+leaving only the attackers useless snapshots.
+
+For example, if the ``forget`` policy is to keep three weekly snapshots, and
+the attacker adds an empty snapshot for each of the last three weeks, all with
+a timestamp (see the ``backup`` command's ``-`time`` option) slightly more
```
MichaelEischer, line 373: There's a typo in the `--time` option.

doc/060_forget.rst
```diff
@@ -336,6 +343,50 @@ could specify:
...
+recent than the existing snapshots (but still within the target week), then the
+next time the repository administrator (or scheduled job) runs the ``forget``
```
MichaelEischer, line 375: `or a scheduled job`?

doc/060_forget.rst
```diff
@@ -336,6 +343,50 @@ could specify:
...
+command with this policy, the legitimate snapshots will be removed (since the
+policy will use the most recent snapshot within each week). Even without
```
MichaelEischer, line 377: s/use/keep only/

MichaelEischer (commented):
I've added a few small remarks. Currently, there's no explanation what a trusted client is and how it is involved in an append-only backup.

And I'm wondering whether it would be helpful to add a TL;DR box for the "Security considerations in append-only mode" section. After all the section can be summed up as "Use `--keep-within` if you want to forget data in an append-only repository" and I'm (unfortunately) not sure whether everyone manages to read the complete section.

doc/060_forget.rst
rawtaz: Yeah, a `forget` run with a policy that is susceptible to the attack discussed here might be run manually by the administrator, or in an automated way by e.g. a cron job or similar. The text said "cron job" but I changed it to "scheduled job" because cron doesn't exist on Windows.

rawtaz: Good catch, will change.

lgommans:
> I'm just curious, which element did you change?

On <https://readthedocs.org/dashboard/restic-test-for-merge-request/advanced/> the element with ID `id_default_branch`

---

The changes look good to me! I spotted a minor typo, will add a comment there.

@rawtaz Since you said "Good catch, will change." in response to one of Michael's comments, are you already working on another commit? Just to prevent double work.

doc/060_forget.rst
```diff
@@ -345,34 +346,46 @@ could specify:
...
+To remove snapshots and recover the corresponding disk space, the ``forget``
+and ``prune`` commands must have full read, write and delete access to the
+repository. If an attacker has this, the protection offered by append-only
+mode is naturally void.
+
+However, even with append-only mode active, an attacker who is able to add
+additional and empty or otherwise useless snapshots to the repository can
+potentially cause a situation where a trusted client running ``forget`` with
+certain ``--keep-*`` options might unknowingly remove legitimate snapshots,
+leaving only the attackers useless snapshots.
```
lgommans, line 369: `attacker's` with apostrophe

rawtaz:
> are you already working on another commit?

I was intending to, but if you want to have a stab at it I don't mind. As long as I can add some adjustments if I feel like it :)

doc/060_forget.rst
lgommans: Upon re-reading the paragraph, the future timestamp thing indeed seems kind of like a stray remark in between two connected sentences. That futures are ignored for `keep-within` is already mentioned in a `.. note::` section above. I've just removed the sentence in my upcoming commit but just treat that as a proposal.

doc/060_forget.rst
lgommans: Isn't that already explained in the previous paragraph? I'm trying to come up with a short addition that makes it more clear what "trusted client" means but it all sounds redundant.

Perhaps we should instead change the phrasing of "a trusted client" into something like "a client with full access", especially since "trusted" is never used again (defining a term for single use is kind of silly). That makes the paragraph-length sentence even more convoluted though, so I've rewritten it a bit (in the upcoming commit).

lgommans:
Interesting... because my local system time was wrong (I may or may not have been playing [worldle](https://worldle.teuteuf.fr)) it seems that GitHub 'correctly' sorts the commit into the thread by timestamp, thereby hiding it somewhere above the previous comments. Tried to amend it, and looking at the [commits tab](https://github.com/restic/restic/pull/3656/commits) it seems I succeeded in the end, so just ignore the commits in between :sweat_smile:

This commit addresses each of the comments, though in particular for "there's no explanation what a trusted client is and how it is involved in an append-only backup" (see also [this reply](https://github.com/restic/restic/pull/3656#discussion_r810680960)) I'm not sure if it's sufficiently clear now. Feel free to add commits with changes/additions/deletions, of course, or request more changes.

lgommans:
Sorry for the delay, updated it just now, 'latest' should be up-to-date. If you have a readthedocs username for me, I can also add you as maintainer btw.

doc/060_forget.rst
MichaelEischer: What about adding something like the following:
```
The usual setup is to use a separate, well-secured client that is full access to the repository and is only used to run ``prune``.
```

rawtaz: I've incorporated your suggestion in the new text changes I just pushed, good input! I'll mark it as resolved now that @lgommans removed the reference to "trusted" earlier and I added your suggestion.

rawtaz:
Okay, I went at it again and just polished the remaining review comments by @MichaelEischer. However, I also sidetracked a bit and updated most of the forget documentation about policies such that it is now hopefully much more readable and understandable - I think it was a bit of an unstructured mess before. Let's see what you think.

@lgommans Can you kick https://restic-test-for-merge-request.readthedocs.io/en/latest/060_forget.html#removing-snapshots-according-to-a-policy again?

lgommans:
Docs updated!

fd0 (approved):
Good work, thanks!

rawtaz:
@lgommans Thanks for your help! I didn't delete your branch just in case you want to keep it for nostalgic reasons.

PR #3661, opened 2022-03-05
add SAS authentication option for Azure repos

### LINKED ISSUE #2295: Support for Azure SAS (Shared Access Signatures)
Output of `restic version`
--------------------------
restic 0.9.5 compiled with go1.12.2 on linux/amd64

Currently, Restic supports authenticating with Azure Storage Account credentials, but does not support Azure SAS (shared access signatures: https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1)

SAS is important as it allows granular, time limited, per-container permissions to be used, instead of using the storage account key which gives read/write over the entire account and all resources (including all containers, tables, queues and fileshares).

Currently, with Restic requiring the use of the storage account key, any secure deployment at scale is forced to create many separate storage accounts housing single containers. This is clearly not the intended way to use Azure storage.

Moreover, storage account names are within a global namespace, and have a limited charset, meaning creating many accounts can be a problem.

SAS support was added to Duplicity with minimal changes, which may mean the same would be true with Restic:
https://bazaar.launchpad.net/~duplicity-team/duplicity/0.8-series/revision/1179#duplicity/backends/azurebackend.py

What are you trying to do?
--------------------------
Use Restic on Azure with SAS, and avoid using the Storage Account "master key", making backup containers easier to manage.

Restic is great, easy to deploy and backed up by good documentation. I'm happy that this is my only significant problem so far.

rgammans:

This PR adds the ability to use Azure SAS tokens for authentication. This feature is requested in #2295 

This change adds an additional string value to the azure configuration struct to store the SAS, and a separate path to create the service client.

Fixes:  #2295

Checklist
---------

- [ x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).

I haven't coded any tests as they will require an update to the testing infrastructure to create SAS TOKEN, (although that should be a relatively straightforward change in the workflow). Functionally they should do the same as the existing tests but with a SAS token provided instead of ACCOUNT_KEY.

NB: I'm a relative newbie in Go so there may be some style issues.

### DIFF
--- a/changelog/unreleased/issue-2295
+++ b/changelog/unreleased/issue-2295
@@ -1,0 +1,14 @@
+Enhancement: Allow use of SAS token to authenticate to Azure
+
+Previously restic only supported AccountKeys to authenticate to Azure
+storage accounts, which necessitates giving a significant amount of
+access.
+
+We added support for Azure SAS tokens which are a more fine-grained
+and time-limited manner of granting access. Set the `AZURE_ACCOUNT_NAME`
+and `AZURE_ACCOUNT_SAS` environment variables to use a SAS token for
+authentication. Note that if `AZURE_ACCOUNT_KEY` is set, it will take
+preference.
+
+https://github.com/restic/restic/issues/2295
+https://github.com/restic/restic/pull/3661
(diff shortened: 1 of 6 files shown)


MichaelEischer (approved):
LGTM.

I've rebased the PR, squashed the commits, fixed the merge conflict, fixed a few typos and added a commit to automatically strip a leading `?` from the SAS token.

PR #3513, opened 2021-09-15
Speed-up copy command

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
The copy command currently proceeds blob by blob which can be very slow if there is any latency for backend accesses.

This PR reuses the repack operation used by prune to implement copy:
The repack operation copies all selected blobs from a set of pack files into new pack files. For prune the source and destination repositories are identical. To implement copy, just use a different source and destination repository.

This way the copy command gains all performance improvements made to the prune command, while also simplifying the implementation. The main change of this PR is the last commit, all other commits are part of #3484. Although this PR could be implemented standalone, I've decided to use #3484 as base as it only accesses the relevant parts of pack files instead of always downloading the full pack file.

The PR also adds a progress bar for the number of pack files copied for the current snapshot.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
Fixes #2923.

Checklist
---------
- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- ~~[ ] I have added documentation for relevant changes (in the manual).~~
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3666, opened 2022-03-06
Implement compression support

### LINKED ISSUE #21: Implement Compression

MichaelEischer:

This PR implements support for compression in restic. The repository format description was updated to reflect the changes necessary to support compressed data in a repository.

In a nutshell, pack files now support compressed variants of tree and data blobs which have been added as a straightforward extension of the current pack header format. Internally, compressed and uncompressed blobs are treated equivalently, the only difference is how they are stored in a pack file. This removes the need for far reaching modifications to support compression in every part of restic.

Unpacked files likes lock, index and snapshot files are also compressed before encryption. The compressed plaintext starts with a null byte and version byte to allow restic to determine whether an unpacked file is compressed or not. The null byte was chosen as marker for the new format as valid JSON cannot start with a null byte.

The `init` command now includes a flag `--repository-version` which allows users to select the repository format version. Besides numerical values (`1` == old format, and `2` == compression support) two special values `latest` and `stable` are supported. The purpose of them it to allow users to easily opt-in to all new features while allowing us to update the default repository format used to create new repositories more conservatively. The default repository format is `stable` which for testing purposes is currently set to version `2`. A design alternative here would be to provide a restic version as alias for a repository version. That would allow users to e.g. set the minimum required restic version to `0.13.0` to get the compression support.

~~The code diff is rather large as it is based on #3513 and #3484 which provide a unified implementation for streaming pack files 
In addition, several commits can be split off into a separate code refactoring PR that can be reviewed separately (although the cleanups are a prerequisite for this PR).~~

The PR is currently in a request for comment state, with at least the following list of TODOs remaining:
- [x] Tests which explicitly test both old and new repository format
- [x] Use old repository version by default for a certain time
- [x] Split basic refactoring into a separate PR
- [x] Decide on whether we want to expose the raw repository version to restic users or just use a minimum required version as proxy
- [x] Document `--repository-version` option
- [x] Future work: Mechanism to upgrade repositories to version 2

Replaces #2441
Implements a part of #21

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -439,7 +439,7 @@ func (r *Repository) decompressUnpacked(p []byte) ([]byte, error) {
 		return p, nil
 	}
 
-	if len(p) < 1 {
+	if len(p) == 0 {
 		// too short for version header
 		return p, nil
 	}
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	"github.com/cenkalti/backoff/v4"
+	"github.com/klauspost/compress/zstd"
 	"github.com/restic/chunker"
 	"github.com/restic/restic/internal/backend/dryrun"
 	"github.com/restic/restic/internal/cache"
@@ -40,6 +41,9 @@ type Repository struct {
 
 	treePM *packerManager
 	dataPM *packerManager
+
+	enc *zstd.Encoder
+	dec *zstd.Decoder
 }
 
 // New returns a new repository with backend be.
@@ -51,6 +55,16 @@ func New(be restic.Backend) *Repository {
 		treePM: newPackerManager(be, nil),
 	}
 
+	enc, err := zstd.NewWriter(nil)
+	if err != nil {
+		panic(err)
+	}
+	repo.enc = enc
+	dec, err := zstd.NewReader(nil)
+	if err != nil {
+		panic(err)
+	}
+	repo.dec = dec
 	return repo
 }
 
@@ -124,6 +138,9 @@ func (r *Repository) LoadUnpacked(ctx context.Context, buf []byte, t restic.File
 	plaintext, err := r.key.Open(ciphertext[:0], nonce, ciphertext, nil)
 	if err != nil {
 		return nil, err
+	}
+	if t != restic.ConfigFile {
+		return r.decompressUnpacked(plaintext)
 	}
 
 	return plaintext, nil
@@ -312,9 +329,50 @@ func (r *Repository) SaveJSONUnpacked(ctx context.Context, t restic.FileType, it
 	return r.SaveUnpacked(ctx, t, plaintext)
 }
 
+func (r *Repository) compressUnpacked(p []byte) ([]byte, error) {
+	// compression is only available starting from version 2
+	if r.cfg.Version < 2 {
+		return p, nil
+	}
+
+	// version byte
+	out := []byte{2}
+	out = r.enc.EncodeAll(p, out)
+	return out, nil
+}
+
+func (r *Repository) decompressUnpacked(p []byte) ([]byte, error) {
+	// compression is only available starting from version 2
+	if r.cfg.Version < 2 {
+		return p, nil
+	}
+
+	if len(p) < 1 {
+		// too short for version header
+		return p, nil
+	}
+	if p[0] == '[' || p[0] == '{' {
+		// probably raw JSON
+		return p, nil
+	}
+	// version
+	if p[0] != 2 {
+		return nil, errors.New("not supported encoding format")
+	}
+
+	return r.dec.DecodeAll(p[1:], nil)
+}
+
 // SaveUnpacked encrypts data and stores it in the backend. Returned is the
 // storage hash.
 func (r *Repository) SaveUnpacked(ctx context.Context, t restic.FileType, p []byte) (id restic.ID, err error) {
+	if t != restic.ConfigFile {
+		p, err = r.compressUnpacked(p)
+		if err != nil {
+			return restic.ID{}, err
+		}
+	}
+
 	ciphertext := restic.NewBlobBuffer(len(p))
 	ciphertext = ciphertext[:0]
 	nonce := crypto.NewRandomNonce()
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -1021,9 +1021,8 @@ func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, pack
 			nonce, ciphertext := buf[:key.NonceSize()], buf[key.NonceSize():]
 			plaintext, err := key.Open(ciphertext[:0], nonce, ciphertext, nil)
 			if err == nil && entry.IsCompressed() {
-				if cap(decode) < int(entry.DataLength()) {
-					decode = make([]byte, 0, entry.DataLength())
-				}
+				// DecodeAll will allocate a slice if it is not large enough since it
+				// knows the decompressed size (because we're using EncodeAll)
 				decode, err = dec.DecodeAll(plaintext, decode[:0])
 				plaintext = decode
 				if err != nil {
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -42,8 +42,10 @@ type Repository struct {
 	treePM *packerManager
 	dataPM *packerManager
 
-	enc *zstd.Encoder
-	dec *zstd.Decoder
+	allocEnc sync.Once
+	allocDec sync.Once
+	enc      *zstd.Encoder
+	dec      *zstd.Decoder
 }
 
 // New returns a new repository with backend be.
@@ -55,16 +57,6 @@ func New(be restic.Backend) *Repository {
 		treePM: newPackerManager(be, nil),
 	}
 
-	enc, err := zstd.NewWriter(nil)
-	if err != nil {
-		panic(err)
-	}
-	repo.enc = enc
-	dec, err := zstd.NewReader(nil)
-	if err != nil {
-		panic(err)
-	}
-	repo.dec = dec
 	return repo
 }
 
@@ -236,7 +228,7 @@ func (r *Repository) LoadBlob(ctx context.Context, t restic.BlobType, id restic.
 		}
 
 		if blob.IsCompressed() {
-			plaintext, err = r.dec.DecodeAll(plaintext, make([]byte, 0, blob.DataLength()))
+			plaintext, err = r.getZstdDecoder().DecodeAll(plaintext, make([]byte, 0, blob.DataLength()))
 			if err != nil {
 				lastError = errors.Errorf("decompressing blob %v failed: %v", id, err)
 				continue
@@ -280,6 +272,28 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
 	return r.idx.LookupSize(restic.BlobHandle{ID: id, Type: tpe})
 }
 
+func (r *Repository) getZstdEncoder() *zstd.Encoder {
+	r.allocEnc.Do(func() {
+		enc, err := zstd.NewWriter(nil)
+		if err != nil {
+			panic(err)
+		}
+		r.enc = enc
+	})
+	return r.enc
+}
+
+func (r *Repository) getZstdDecoder() *zstd.Decoder {
+	r.allocDec.Do(func() {
+		dec, err := zstd.NewReader(nil)
+		if err != nil {
+			panic(err)
+		}
+		r.dec = dec
+	})
+	return r.dec
+}
+
 // saveAndEncrypt encrypts data and stores it to the backend as type t. If data
 // is small enough, it will be packed together with other small blobs.
 // The caller must ensure that the id matches the data.
@@ -289,7 +303,7 @@ func (r *Repository) saveAndEncrypt(ctx context.Context, t restic.BlobType, data
 	uncompressedLength := 0
 	if r.cfg.Version > 1 {
 		uncompressedLength = len(data)
-		data = r.enc.EncodeAll(data, nil)
+		data = r.getZstdEncoder().EncodeAll(data, nil)
 	}
 
 	nonce := crypto.NewRandomNonce()
@@ -354,7 +368,7 @@ func (r *Repository) compressUnpacked(p []byte) ([]byte, error) {
 
 	// version byte
 	out := []byte{2}
-	out = r.enc.EncodeAll(p, out)
+	out = r.getZstdEncoder().EncodeAll(p, out)
 	return out, nil
 }
 
@@ -377,7 +391,7 @@ func (r *Repository) decompressUnpacked(p []byte) ([]byte, error) {
 		return nil, errors.New("not supported encoding format")
 	}
 
-	return r.dec.DecodeAll(p[1:], nil)
+	return r.getZstdDecoder().DecodeAll(p[1:], nil)
 }
 
 // SaveUnpacked encrypts data and stores it in the backend. Returned is the
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -329,7 +329,18 @@ func (r *Repository) getZstdEncoder() *zstd.Encoder {
 			level = zstd.SpeedBestCompression
 		}
 
-		enc, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(level))
+		opts := []zstd.EOption{
+			// Set the compression level configured.
+			zstd.WithEncoderLevel(level),
+			// Disable CRC, we have enough checks in place, makes the
+			// compressed data four bytes shorter.
+			zstd.WithEncoderCRC(false),
+			// Set a window of 512kbyte, so we have good lookbehind for usual
+			// blob sizes.
+			zstd.WithWindowSize(512 * 1024),
+		}
+
+		enc, err := zstd.NewWriter(nil, opts...)
 		if err != nil {
 			panic(err)
 		}
@@ -340,7 +351,15 @@ func (r *Repository) getZstdEncoder() *zstd.Encoder {
 
 func (r *Repository) getZstdDecoder() *zstd.Decoder {
 	r.allocDec.Do(func() {
-		dec, err := zstd.NewReader(nil)
+		opts := []zstd.DOption{
+			// Use all available cores.
+			zstd.WithDecoderConcurrency(0),
+			// Limit the maximum decompressed memory. Set to a very high,
+			// conservative value.
+			zstd.WithDecoderMaxMemory(16 * 1024 * 1024 * 1024),
+		}
+
+		dec, err := zstd.NewReader(nil, opts...)
 		if err != nil {
 			panic(err)
 		}
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -37,6 +37,8 @@ type Repository struct {
 	idx     *MasterIndex
 	Cache   *cache.Cache
 
+	opts Options
+
 	noAutoIndexUpdate bool
 
 	treePM *packerManager
@@ -48,10 +50,58 @@ type Repository struct {
 	dec      *zstd.Decoder
 }
 
+type Options struct {
+	Compression CompressionMode
+}
+
+// CompressionMode configures if data should be compressed.
+type CompressionMode uint
+
+// Constants for the different compression levels.
+const (
+	CompressionAuto CompressionMode = 0
+	CompressionOff  CompressionMode = 1
+	CompressionMax  CompressionMode = 2
+)
+
+// Set implements the method needed for pflag command flag parsing.
+func (c *CompressionMode) Set(s string) error {
+	switch s {
+	case "auto":
+		*c = CompressionAuto
+	case "off":
+		*c = CompressionOff
+	case "max":
+		*c = CompressionMax
+	default:
+		return fmt.Errorf("invalid compression mode %q, must be one of (auto|off|max)", s)
+	}
+
+	return nil
+}
+
+func (c *CompressionMode) String() string {
+	switch *c {
+	case CompressionAuto:
+		return "auto"
+	case CompressionOff:
+		return "off"
+	case CompressionMax:
+		return "max"
+	default:
+		return "invalid"
+	}
+
+}
+func (c *CompressionMode) Type() string {
+	return "mode"
+}
+
 // New returns a new repository with backend be.
-func New(be restic.Backend) *Repository {
+func New(be restic.Backend, opts Options) *Repository {
 	repo := &Repository{
 		be:     be,
+		opts:   opts,
 		idx:    NewMasterIndex(),
 		dataPM: newPackerManager(be, nil),
 		treePM: newPackerManager(be, nil),
@@ -274,7 +324,12 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
 
 func (r *Repository) getZstdEncoder() *zstd.Encoder {
 	r.allocEnc.Do(func() {
-		enc, err := zstd.NewWriter(nil)
+		level := zstd.SpeedDefault
+		if r.opts.Compression == CompressionMax {
+			level = zstd.SpeedBestCompression
+		}
+
+		enc, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(level))
 		if err != nil {
 			panic(err)
 		}
@@ -302,8 +357,14 @@ func (r *Repository) saveAndEncrypt(ctx context.Context, t restic.BlobType, data
 
 	uncompressedLength := 0
 	if r.cfg.Version > 1 {
-		uncompressedLength = len(data)
-		data = r.getZstdEncoder().EncodeAll(data, nil)
+
+		// we have a repo v2, so compression is available. if the user opts to
+		// not compress, we won't compress any data, but everything else is
+		// compressed.
+		if r.opts.Compression != CompressionOff || t != restic.DataBlob {
+			uncompressedLength = len(data)
+			data = r.getZstdEncoder().EncodeAll(data, nil)
+		}
 	}
 
 	nonce := crypto.NewRandomNonce()
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -661,7 +661,15 @@ func (r *Repository) SearchKey(ctx context.Context, password string, maxKeys int
 
 // Init creates a new master key with the supplied password, initializes and
 // saves the repository config.
-func (r *Repository) Init(ctx context.Context, password string, chunkerPolynomial *chunker.Pol) error {
+func (r *Repository) Init(ctx context.Context, version uint, password string, chunkerPolynomial *chunker.Pol) error {
+	if version > restic.MaxRepoVersion {
+		return fmt.Errorf("repo version %v too high", version)
+	}
+
+	if version < restic.MinRepoVersion {
+		return fmt.Errorf("repo version %v too low", version)
+	}
+
 	has, err := r.be.Test(ctx, restic.Handle{Type: restic.ConfigFile})
 	if err != nil {
 		return err
@@ -670,7 +678,7 @@ func (r *Repository) Init(ctx context.Context, password string, chunkerPolynomia
 		return errors.New("repository master key and config already initialized")
 	}
 
-	cfg, err := restic.CreateConfig()
+	cfg, err := restic.CreateConfig(version)
 	if err != nil {
 		return err
 	}
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -116,6 +116,14 @@ func (r *Repository) DisableAutoIndexUpdate() {
 	r.noAutoIndexUpdate = true
 }
 
+// setConfig assigns the given config and updates the repository parameters accordingly
+func (r *Repository) setConfig(cfg restic.Config) {
+	r.cfg = cfg
+	if r.cfg.Version >= 2 {
+		r.idx.markCompressed()
+	}
+}
+
 // Config returns the repository configuration.
 func (r *Repository) Config() restic.Config {
 	return r.cfg
@@ -772,15 +780,14 @@ func (r *Repository) SearchKey(ctx context.Context, password string, maxKeys int
 	r.dataPM.key = key.master
 	r.treePM.key = key.master
 	r.keyName = key.Name()
-	r.cfg, err = restic.LoadConfig(ctx, r)
+	cfg, err := restic.LoadConfig(ctx, r)
 	if err == crypto.ErrUnauthenticated {
 		return errors.Fatalf("config or key %v is damaged: %v", key.Name(), err)
 	} else if err != nil {
 		return errors.Fatalf("config cannot be loaded: %v", err)
 	}
-	if r.Config().Version >= 2 {
-		r.idx.markCompressed()
-	}
+
+	r.setConfig(cfg)
 	return nil
 }
 
@@ -826,7 +833,7 @@ func (r *Repository) init(ctx context.Context, password string, cfg restic.Confi
 	r.dataPM.key = key.master
 	r.treePM.key = key.master
 	r.keyName = key.Name()
-	r.cfg = cfg
+	r.setConfig(cfg)
 	_, err = r.SaveJSONUnpacked(ctx, restic.ConfigFile, cfg)
 	return err
 }
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -235,12 +235,23 @@ func (r *Repository) LoadBlob(ctx context.Context, t restic.BlobType, id restic.
 			continue
 		}
 
+		if blob.IsCompressed() {
+			plaintext, err = r.dec.DecodeAll(plaintext, make([]byte, 0, blob.DataLength()))
+			if err != nil {
+				lastError = errors.Errorf("decompressing blob %v failed: %v", id, err)
+				continue
+			}
+		}
+
 		// check hash
 		if !restic.Hash(plaintext).Equal(id) {
 			lastError = errors.Errorf("blob %v returned invalid hash", id)
 			continue
 		}
 
+		if len(plaintext) > cap(buf) {
+			return plaintext, nil
+		}
 		// move decrypted data to the start of the buffer
 		copy(buf, plaintext)
 		return buf[:len(plaintext)], nil
@@ -275,6 +286,12 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
 func (r *Repository) saveAndEncrypt(ctx context.Context, t restic.BlobType, data []byte, id restic.ID) error {
 	debug.Log("save id %v (%v, %d bytes)", id, t, len(data))
 
+	uncompressedLength := 0
+	if r.cfg.Version > 1 {
+		uncompressedLength = len(data)
+		data = r.enc.EncodeAll(data, nil)
+	}
+
 	nonce := crypto.NewRandomNonce()
 
 	ciphertext := make([]byte, 0, restic.CiphertextLength(len(data)))
@@ -301,7 +318,7 @@ func (r *Repository) saveAndEncrypt(ctx context.Context, t restic.BlobType, data
 	}
 
 	// save ciphertext
-	_, err = packer.Add(t, id, ciphertext)
+	_, err = packer.Add(t, id, ciphertext, uncompressedLength)
 	if err != nil {
 		return err
 	}
@@ -534,6 +551,17 @@ func (r *Repository) LoadIndex(ctx context.Context) error {
 	err = r.idx.MergeFinalIndexes()
 	if err != nil {
 		return err
+	}
+
+	if r.cfg.Version < 2 {
+		// sanity check
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		for blob := range r.idx.Each(ctx) {
+			if blob.IsCompressed() {
+				return errors.Fatal("index uses feature not supported by repository version 1")
+			}
+		}
 	}
 
 	// remove index files from the cache which have been removed in the repo
@@ -834,9 +862,15 @@ func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, pack
 
 	debug.Log("streaming pack %v (%d to %d bytes), blobs: %v", packID, dataStart, dataEnd, len(blobs))
 
+	dec, err := zstd.NewReader(nil)
+	if err != nil {
+		panic(dec)
+	}
+	defer dec.Close()
+
 	ctx, cancel := context.WithCancel(ctx)
 	// stream blobs in pack
-	err := beLoad(ctx, h, int(dataEnd-dataStart), int64(dataStart), func(rd io.Reader) error {
+	err = beLoad(ctx, h, int(dataEnd-dataStart), int64(dataStart), func(rd io.Reader) error {
 		// prevent callbacks after cancelation
 		if ctx.Err() != nil {
 			return ctx.Err()
@@ -849,6 +883,7 @@ func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, pack
 		bufRd := bufio.NewReaderSize(rd, bufferSize)
 		currentBlobEnd := dataStart
 		var buf []byte
+		var decode []byte
 		for _, entry := range blobs {
 			skipBytes := int(entry.Offset - currentBlobEnd)
 			if skipBytes < 0 {
@@ -888,6 +923,16 @@ func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, pack
 			// decryption errors are likely permanent, give the caller a chance to skip them
 			nonce, ciphertext := buf[:key.NonceSize()], buf[key.NonceSize():]
 			plaintext, err := key.Open(ciphertext[:0], nonce, ciphertext, nil)
+			if err == nil && entry.IsCompressed() {
+				if cap(decode) < int(entry.DataLength()) {
+					decode = make([]byte, 0, entry.DataLength())
+				}
+				decode, err = dec.DecodeAll(plaintext, decode[:0])
+				plaintext = decode
+				if err != nil {
+					err = errors.Errorf("decompressing blob %v failed: %v", h, err)
+				}
+			}
 			if err == nil {
 				id := restic.Hash(plaintext)
 				if !id.Equal(entry.ID) {
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -698,6 +698,9 @@ func (r *Repository) SearchKey(ctx context.Context, password string, maxKeys int
 	} else if err != nil {
 		return errors.Fatalf("config cannot be loaded: %v", err)
 	}
+	if r.Config().Version >= 2 {
+		r.idx.markCompressed()
+	}
 	return nil
 }
 
--- a/changelog/unreleased/issue-21
+++ b/changelog/unreleased/issue-21
@@ -4,6 +4,11 @@ We have added compression support to the restic repository format. To create a
 repository using the new format run `init --repository-version 2`. Please note
 that the repository cannot be read by restic versions prior to 0.14.0.
 
+You can configure if data is compressed with the option `--compression`. It can
+be set to `auto` (the default, which will compress very fast), `max` (which
+will trade backup speed and CPU usage for better compression), or `off` (which
+disables compression). Each setting is only applied for the single run of restic.
+
 The new format version has not received much testing yet. Do not rely on it as
 your only backup copy! Please run `check` in regular intervals to detect any
 problems.
--- a/changelog/unreleased/issue-21
+++ b/changelog/unreleased/issue-21
@@ -1,0 +1,16 @@
+Enhancement: Add comppression support
+
+We have added compression support to the restic repository format. To create a
+repository using the new format run `init --repository-version 2`. Please note
+that the repository cannot be read by restic versions prior to 0.14.0.
+
+The new format version has not received much testing yet. Do not rely on it as
+your only backup copy! Please run `check` in regular intervals to detect any
+problems.
+
+Upgrading in place is not yet supported. As a workaround, first create a new
+repository using `init --repository-version 2 --copy-chunker-params --repo2 path/to/old/repo`.
+Then use the `copy` command to copy all snapshots to the new repository.
+
+https://github.com/restic/restic/issues/21
+https://github.com/restic/restic/pull/3666
--- a/cmd/restic/cmd_debug.go
+++ b/cmd/restic/cmd_debug.go
@@ -333,44 +333,37 @@ func loadBlobs(ctx context.Context, repo restic.Repository, pack restic.ID, list
 
 		nonce, plaintext := buf[:key.NonceSize()], buf[key.NonceSize():]
 		plaintext, err = key.Open(plaintext[:0], nonce, plaintext, nil)
+		outputPrefix := ""
+		filePrefix := ""
 		if err != nil {
 			Warnf("error decrypting blob: %v\n", err)
-			var plain []byte
 			if tryRepair || repairByte {
-				plain = tryRepairWithBitflip(ctx, key, buf, repairByte)
+				plaintext = tryRepairWithBitflip(ctx, key, buf, repairByte)
 			}
-			var prefix string
-			if plain != nil {
-				id := restic.Hash(plain)
-				if !id.Equal(blob.ID) {
-					Printf("         repaired blob (length %v), hash is %v, ID does not match, wanted %v\n", len(plain), id, blob.ID)
-					prefix = "repaired-wrong-hash-"
-				} else {
-					Printf("         successfully repaired blob (length %v), hash is %v, ID matches\n", len(plain), id)
-					prefix = "repaired-"
-				}
+			if plaintext != nil {
+				outputPrefix = "repaired "
+				filePrefix = "repaired-"
 			} else {
-				plain = decryptUnsigned(ctx, key, buf)
-				prefix = "damaged-"
+				plaintext = decryptUnsigned(ctx, key, buf)
+				err = storePlainBlob(blob.ID, "damaged-", plaintext)
+				if err != nil {
+					return err
+				}
+				continue
 			}
-			err = storePlainBlob(blob.ID, prefix, plain)
-			if err != nil {
-				return err
-			}
-			continue
 		}
 
 		id := restic.Hash(plaintext)
 		var prefix string
 		if !id.Equal(blob.ID) {
-			Printf("         successfully decrypted blob (length %v), hash is %v, ID does not match, wanted %v\n", len(plaintext), id, blob.ID)
+			Printf("         successfully %vdecrypted blob (length %v), hash is %v, ID does not match, wanted %v\n", outputPrefix, len(plaintext), id, blob.ID)
 			prefix = "wrong-hash-"
 		} else {
-			Printf("         successfully decrypted blob (length %v), hash is %v, ID matches\n", len(plaintext), id)
+			Printf("         successfully %vdecrypted blob (length %v), hash is %v, ID matches\n", outputPrefix, len(plaintext), id)
 			prefix = "correct-"
 		}
 		if extractPack {
-			err = storePlainBlob(id, prefix, plaintext)
+			err = storePlainBlob(id, filePrefix+prefix, plaintext)
 			if err != nil {
 				return err
 			}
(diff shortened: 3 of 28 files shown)


DRON-666:
I converted a small 35GB repository to just 14GB with this version and it's a very good result. But for a full check (`check --read-data --check-unused`) of this compressed repository, more than 14GB of memory is required (less than 100MB is required for original repo). It's looks like the data of all blobs has been leaked.

restic version:
```
restic 0.12.1 compiled with go1.17.8 on windows/amd64

**Comment by MichaelEischer** [comment]
Hmm, I only see a high memory usage when using go 1.17.8. With go 1.16.6 everything is fine. I'm not sure yet what's going on there. The bulk of the (leaked?) memory seems to be allocated inside StreamPacks, but after the method has finished processing a pack file, there should be no way to hold a reference to that memory.

**Comment by DRON-666** [comment]
>With go 1.16.6 everything is fine.

Unfortunately [it's not](https://www16.zippyshare.com/v/wvC221IB/file.html).

>The bulk of the (leaked?) memory seems to be allocated inside StreamPacks, but after the method has finished processing a pack file, there should be no way to hold a reference to that memory.

Maybe there are some pointers to the middle of the compressed data, and this prevents garbage collection of the entire block?

**Comment by aawsome** [comment]
I only had a short look at the design document so far, so just some things which came into my mind:

- As now the length of the plain blob and the length of the compressed and encrypted blob are saved independently in the index, the memory usage is increased which should be noted. I was thinking in the last months about a solution without this increase but I agree we need both sizes.
- I don't know if allowing mixed compressed/uncompressed blobs within a pack is a wise idea. Just remember we had this with trees/data which turned out to be a very bad idea. Moreover, if this is allowed this always is a attribute of a blob and therefore has to be stored (direct or indirect) in the index. Future changes may imply that this state has to be directly stored in the index consuming more memory. If you don't allow to mix is, it could be saved as a pack attribute which would lead (especially in combination with "large" packfiles) to low memory impacts.
- IMO in a V2 format, uncompressed tree blobs should not be allowed. Compressing JSONs will be always a big size reduction. As tree blobs/packs are also saved in the cache, this will lead to much smaller cache sizes. And when converting a V1 repo to a V2 repo, it is really easy to convert all trees (which are anyway in access for tree traversals).
- Ah, and of course this also applies to all index/key/snapshot/(maybe lock) files: Always compress them in V2 and please get rid of this zero-byte hack, but rather provide a conversion V1 -> V2 which does the compression.
- About the new pack header format... I don't like it! I mean, the pack header is anyway not really used in practice. In fact for (almost) all operations, the index is used and the pack headers are only needed within `rebuild-index` and checked in `check --read-data*`. So why make this kind of wired binary format even more complex? Especially now having non-constant-size entries within a raw binary format. This is hell for anyone who wants to re-implement the design. Here we have no speed issues and also the size does not much matter as there is not much information saved. What about using a standard serialization format, like JSON? And of course split the information into pack information (like tree/data/compression??) and blob information (like id/length(s)/compression?).
- Also, if talking about a V2 version, why not also change the index JSON format. The `type` is no longer a blob attribute, so put it under `packs`.
- When changing the repo format, there are other format changes which are waiting. Maybe consider adding some of them which don't require big code changes (I think mainly the Tree format design change could easily come along if we anyway process all trees in a conversion)

**Comment by DRON-666** [comment]
@MichaelEischer After adding `defer dec.Close();` to `repository.StreamPack` all leaks disappeared.

**Comment by MichaelEischer** [comment]
> @MichaelEischer After adding `defer dec.Close();` to `repository.StreamPack` all leaks disappeared.

You're right, the code was leaking go routines in the decoder without end. The increased memory usage I've seen between Go 1.16 and 1.17 seems to be unrelated.

> * I don't know if allowing mixed compressed/uncompressed blobs within a pack is a wise idea. Just remember we had this with trees/data which turned out to be a very bad idea. Moreover, if this is allowed this always is a attribute of a blob and therefore has to be stored (direct or indirect) in the index. Future changes may imply that this state has to be directly stored in the index consuming more memory. If you don't allow to mix it, it could be saved as a pack attribute which would lead (especially in combination with "large" packfiles) to low memory impacts.

The reason to split trees and data is that directory metadata and file contents are completely different things. There's no such distinction for compressed vs. not-compressed. It's just a different way of recording the same information. In fact, it might be worthwhile to skip compression for certain file types which are known to be non-compressible. Currently the information that a blob is compressed is stored by having a uncompressed length in the index. Are you referring to the mixed pack files list in the index here?

> * IMO in a V2 format, uncompressed tree blobs should not be allowed. Compressing JSONs will be always a big size reduction. As tree blobs/packs are also saved in the cache, this will lead to much smaller cache sizes. And when converting a V1 repo to a V2 repo, it is really easy to convert all trees (which are anyway in access for tree traversals).

I agree that _new_ tree blobs added to a V2 repo must always be compressed. It is also a good idea to compress all tree blobs when upgrading a repository. But here's the catch: it is impossible to atomically upgrade a repository. If the V2 format does not allow uncompressed tree blobs then during upgrade the repository will be in a state that is neither valid in the V1 or V2 format and that will lead to problems if the upgrade is interrupted. My suggestion would be to let `check` complain if there are uncompressed tree blobs in a V2 repo.

> * Ah, and of course this also applies to all index/key/snapshot/(maybe lock) files: Always compress them in V2 and please get rid of this zero-byte hack, but rather provide a conversion V1 -> V2 which does the compression. UPDATE: IMO it would be best to specify in the config file the formats used for saving indices/snapshots, etc. For example add a "snapshot_format="json/zstd" in the config file.

We cannot compress key files without causing old restic versions to print strange errors that there are no valid keys in the repository (the keys are read before (!)  the config file). Key files are small enough that there's not much benefit from compressing them.

The problem with enforcing compression is that it won't allow us to reliably (or incrementally) upgrade a repository. Upgrading snapshots is also a bad idea as that would change all snapshot ids. Storing the snapshot format in the config file, just makes everything more complicated in my opinion, there's no clear benefit over simply defining that the V2 format compresses new snapshots. It also won't work if there's a mix of compressed/uncompressed snapshots.

> * What about using a standard serialization format, like JSON? And of course split the information into pack information (like tree/data/compression??) and blob information (like id/length(s)/compression?).

How is adding a JSON pack header format, that is completely different from the existing one, making things easier for someone who wants to implement the pack file format? I mean instead of adding two new cases to the existing pack header parsing, one needs a completely new code path, which in the end still has to map all pack header information from both formats into a common format (or duplicate everything). And it will require us to store in the index which type of header a pack file is using, otherwise the length check in `prune` will break. (Or we'd have to rewrite to whole repository when upgrading, which kind of misses the point).

I don't see "compression" as pack information but rather as blob information (see above). And just to distinguish between tree and data is not worth it to reorganized the pack header.

> * Also, if talking about a V2 version, why not also change the index JSON format. The `type` is no longer a blob attribute, so put it under `packs`.

For a clean slate design you're right. But does it make much of a difference for compressed indexes? It also increases complexity as that will add a third way how a index could look like.

> * When changing the repo format, there are other format changes which are waiting. Maybe consider adding some of them which don't require big code changes (I think mainly the Tree format design change could easily come along if we anyway process all trees in a conversion)

Which specific tree format change are you referring to? The repo format changes in this PR so far only change how data is stored in a pack file, but does not modify the contents in any way (thus there are no blob ID changes). If a new tree format would require any changes to the actual content of a tree blob, then we'd have to run a full-blown rewrite operation of a repository, which is miles away from a small change. So, we could add support for certain new fields in tree blobs, but anything else will be a problem.

**Comment by fd0** [comment]
>  understand that restic uses the hash of a blob to recognise whether it’s already present in the repository or not, but with compression does it use for this the hash of the uncompressed data or the hash of the compressed data?

That's easy to explain: for identifying data, restic uses the SHA256 hash of the plaintext, before compression and encryption. So the hash only depends on the plaintext. :)

**Comment by aawsome** [comment]
@MichaelEischer I think the main design question is: Should restic be always capable of handling *all* old repo formats. There are pros and cons for answering this with "yes" or "no". I have the feeling want to answer "yes" and you focus on the pros, so I'll give some cons:

- Code stays complex to allow old formats.
- This will make future changes more complex and will make some optimizations impossible.
- Even worse, most (or all active) repos will be turned into the new format(s), so these code paths are in fact dead code which are not used.
- In general, in a modern software you don't want to have dead code. These code pieces are the ones which one time turn into security issues. Or to be more precise: Do you want to have a future security hole which can be abused by artificially creating some legacy repo format?

I strongly recommend to take all the lessons learned about the repo format used so far and create a new one which is free of legacy/hacks to support previous formats. Even more as restic so far has not event reached the 1.0 status.

Then the strategy would be:
- Provide a conversion into the new repo format (which of course should be capable to read both formats)
- Maybe have some versions which still can handle the old format, but mark it as deprecated
- At some point remove support for the old repo format.
- People still using the old repo format can use a released restic version to still read it or convert it to the new format and then use recent restic versions to access it.

> In fact, it might be worthwhile to skip compression for certain file types which are known to be non-compressible. Currently the information that a blob is compressed is stored by having a uncompressed length in the index. Are you referring to the mixed pack files list in the index here?

No, I'm not referring to the mixed pack files list but I thought about this: It might be worthwhile to use a different compression algorithm in future for certain file types or for other reasons. Then you need to add information about compression to each blob (as you already implicitely do) which we would need to save in the index.

> But here's the catch: it is impossible to atomically upgrade a repository. If the V2 format does not allow uncompressed tree blobs then during upgrade the repository will be in a state that is neither valid in the V1 or V2 format and that will lead to problems if the upgrade is interrupted. 

You are right, a conversion algorithm should be able to handle both formats and in-between states. But if you specify that V2 is able to contain both, you have to handle that in all code that is able to process a V2 repo...

> [keys]
Please forget about the keys.. :wink: 

> The problem with enforcing compression is that it won't allow us to reliably (or incrementally) upgrade a repository. Upgrading snapshots is also a bad idea as that would change all snapshot ids.
> It also won't work if there's a mix of compressed/uncompressed snapshots.

See above. IMO you are mixing design of the V2 repo with a conversion algorithm. I was just talking about the design.

> Storing the snapshot format in the config file, just makes everything more complicated in my opinion, 

The benefit would be clarity...

I removed some more arguments as I think all go into discussions about the first question I stated above.

I just want to sketch how in my opinion a repository conversion would look like:
- Keep all data pack files an thus all file contents; optionally read all and save as compressed
- Traverse over all trees an write them (into a possibly new format; possibly including updated tree ids for newly generated subtrees), save as compressed tree blobs in new tree pack files. 
- Save the index for data packs and the new tree packs in a newly created index with (possibly new) index format and compressed.
- Modify all snapshots to use the possibly changed tree ids and save them in compressed format. => Yes this will change snapshot ids but the tag command also already does...
- Remove old tree pack files, old index files and old snapshot files.

**Comment by MichaelEischer** [comment]
> I have the feeling you want to answer "yes" and you focus on the pros, so I'll give some cons:

I agree that these are _possible_ problems when supporting both an old and new format at the same time. But regarding the format changes to support compression I don't see them as (serious) problems.

> I strongly recommend to take all the lessons learned about the repo format used so far and create a new one which is free of legacy/hacks to support previous formats. Even more as restic so far has not even reached the 1.0 status.

The current format has been around long enough that it in some sense has become similar to a version 1.0 format. In any case there would have to be a multi-year transition period IMO. And yes, multi-year means something between 2-5 years or maybe even more. Essentially it should be long enough to let most Linux distributions which include a restic version that only supports the repository format v1 to run out of regular support. I'd like to avoid the idiosyncratic situation where upgrading the operating system jumps from a restic version that only supports v1 repos to one that only supports v2 repos.

As we haven't tackled some more complex features like non-blocking prune yet, I'm pretty sure we don't really know what we need so far. Even if we prepare for that it would be a gamble on getting the repository format redesign right. And completely redesigning the repository format multiple times is just asking for trouble.

> * Code stays complex to allow old formats.
> 
> * This will make future changes more complex and will make some optimizations impossible.

That completely depends on how different the old and the new formats are. For example I don't see the problem of having like 10 additional code lines to support both compressed and uncompressed unpacked blobs. And there's nothing that prevents us from adding the restriction that a repository must be fully upgraded to a new format before several features become available.

> * In general, in a modern software you don't want to have dead code. These code pieces are the ones which one time turn into security issues. Or to be more precise: Do you want to have a future security hole which can be abused by artificially creating some legacy repo format?

That is a rhetorical question and IMO it completely misses the point regarding this PR. The idea behind the design changes in this PR is to add support for compression such that it only requires a few small changes which are likely to require very little maintenance. Introducing a second, completely new repository format will only increase the complexity and maintenance effort until we can get rid of the old format, which will probably take several years.

I haven't thought much about which parts could be simplified by redesigning the current repository format. My feeling at the moment is that the benefits aren't that substantial (don't forget that it would be possible to enforce that a v2 repository does not use mixed packs, the S3 legacy layout etc. without requiring any redesign!) that they are worth giving up backwards compatibility.

> > Storing the snapshot format in the config file, just makes everything more complicated in my opinion,
> 
> The benefit would be clarity...

Why not just tie the decision in which format a file has to be / may be in to the repository format? Accumulating lots of options that determine how a repository is formatted just seems to add complexity compared to that. And if there is only a single valid way to format data (and thus the option always has the same value), then the information is simply redundant.

> See above. IMO you are mixing design of the V2 repo with a conversion algorithm. I was just talking about the design.

Upgrade-ability is in my opinion a property of the repository format. The repository format should be able to recover from a half-upgraded state. And for that at the very least we need some form of version number, either for whole files (as part of the file or implicit based on the containing folder) or for individual properties in e.g. a tree blob. 

My idea how a repository upgrade should work would be based on the following steps, combined in a single command:
- run `check`
- upgrade the repository version number
- run `prune --force-rebuild-index --compress-tree-blobs`
- (maybe mark the upgrade as complete)

That way upgrading a repository will require only little code but also allows enforcing that a repository conforms to certain restrictions. Supporting tree blob rewrites (other than compressing) would require additional steps which I'd like to avoid unless necessary.


In short, I think we should try to carefully extend the current format step by step and keep support for older repository formats unless it becomes (far) more trouble than it's worth. By enforcing that a repository does not use certain legacy features once the upgrade process is complete, we can phase out support for some legacy features. There's also the option of only offering read support for older repository formats.

The main difference to a complete redesign is that the repository format will acquire a certain amount of warts (like an additional null byte or a strangely named parameter here and there). But that will also inevitably happen once we start to evolve a redesigned repository format. I'm also quite convinced that we can add support for most features to the current repository format instead of having to support a new format. Constructions that turned out to be a bad idea can be forbidden to use in newer repository format versions. And thereby we'd avoid the need to support two fundamentally different repository formats for an extended time period. (I'm not advocating to stop thinking what an ideal repository format would look like. But my claim here is that most features of a new format can also be integrated into the existing format.)

**Comment by fd0** [comment]
I agree with what @MichaelEischer wrote above. I think it's far more valuable to incrementally improve the repo format while keeping e.g. already uploaded files valid (which makes users happy), even if that means acquiring a few more warts and a bit more maintenance effort on our side.

**Comment by aawsome** [comment]
> I agree with what @MichaelEischer wrote above. I think it's far more valuable to incrementally improve the repo format

Once again: I don't see any problems in restic still supporting the current format.
I am always just talking about the specification.
But, if you specify a new format, please don't include unneeded warts only for backwards compatibility if you anyway are able to use the format version..

> And if there is only a single valid way to format data (and thus the option always has the same value), then the information is simply redundant.

So, if you specify `version=2` and this implies that your index/snapshots etc are always saved as compressed content, why add this mysterious zero byte in the specification? If you only care about being able to still read the old format, you can simply try to deserialize the content and if this fails, decompres and then deserialize. AFAIR this solution is already implemented for some changed json format (in the index, IIRC).

> while keeping e.g. already uploaded files valid (which makes users happy), even if that means acquiring a few more warts and a bit more maintenance effort on our side.

I agree with uploaded files which are not cached (i.e. data pack files).  But for cached files (and these are basically all other files) it is not much effort to convert them - especially if the target files are smaller and the cache size is therefore reduced by the conversion.


I have a bit the impression that you have this implementation which of course took some effort - and now you are somewhat biased to adapt the specification to the existing implementation. As another user correctly said: Then it's not a specification, its only a documentation.

I asked whether we should first try to make a PR of the new specification before beginning with the implementation because I was worrying about exactly this... :disappointed:

**Comment by aawsome** [comment]
@fd0 you are talking about compatibilty, but you are wrong here:
-  the format supporting compression (v2 repo format) will be never compatible to older restic versions which simply don't implement compression.
- newer restic versions will be able to read/write v1 repos if they implement the v1 repo specification. This does not depend at all on a v2 repo specification.

So the only question (when talking only about adding compression) is: Should a new repo format including compression support  be able to store "things" compressed as well as uncompressed?

For data pack files I fully agree that users may want to keep existing ones. This is because they usually take by far the majority of the repo size, so converting them basically means rebuilding the rep. Also the contents saved in data blobs might not compress well - so a decision based on the content is a benefit for the format.

All other saved data is metadata that is currently saved in uncompressed JSON format (and contained in the cache). From a user's side there is absolutely no benefit having this stored uncompressed: Compression will decrease the size of the repo and of the cache and will speed up repo operations.

If we implement a repo conversion v1 -> v2 with users' experience in mind, would we leave some of this metadata uncompressed? I don't think so!

On contrast, if we specify the v2 format to support both and - even worse - would allow a "repo conversion" by just bumping the repo version, we will enforce behavior which is actually not wanted by users.

**Comment by greatroar** [comment]
I'd rather have a foolproof conversion (as much as possible) than one that has to upgrade the existing indexes and can leave the repo in an inconsistent state if it fails half-way through.

**Comment by greatroar** [comment]
Also, great work @MichaelEischer!

**Comment by fd0** [comment]
> the format supporting compression (v2 repo format) will be never compatible to older restic versions which simply don't implement compression.

What I meant with "forward compatibility" is that data in a v1 repo should be also valid in a v2 repo. So to answer your other question:

> Should a new repo format including compression support be able to store "things" compressed as well as uncompressed?

I think yes, that's a good idea, for repo version 2 at least.

> From a user's side there is absolutely no benefit having this stored uncompressed: Compression will decrease the size of the repo and of the cache and will speed up repo operations.

I disagree, there is a benefit: if we allow this data to stay uncompressed, users do not have to wait until the data is re-uploaded (in compressed form).

The price for the upsides you mentioned is the time spent until the re-upload is finished, which would likely be an exclusive operation during which the repo is locked. If we implement such a conversion process similar to prune, we would only remove data once the upload has successfully finished, so until then we'd need the space (for metadata) in the repo twice.

I'm convinced that a repo v2 which allows uncompressed metadata (index, trees) is a good idea. We can then convert a repo by bumping the version field (a quick operation) and change the code in restic to do the following for v2 repos:
 * New trees and indexes added to the repo are always compressed (no knobs to disable that)
 * All trees repackaged and new indexes created by prune will always be compressed (no knobs to disable that)
 * We can make prune convert some amount of the tree blobs on each run, so that over time (without people noticing too much) the whole repo will be converted (this could maybe be disabled with a CLI switch)

After upgrading the repo, users can directly enjoy using compression.

For the next repo version (3) we could then define that it must not contain uncompressed metadata. We could add code to `prune` which then converts all uncompressed tree blobs and only afterward permit upgrading a repo to version 3.

**Comment by aawsome** [comment]
> For the next repo version (3) we could then define that it must not contain uncompressed metadata.

Sure, you can go this slow way. But keep in mind, then you'll have these format hacks like this zero-byte and your IMO somewhat awkward pack header format still as relict in your repo 3 version. Given the time change v1->v2 this might be another ten or more years... (SCNR)

Anyway, I won't put more energy in an discussion that seems to be not open but already decided :wink:

**Comment by nickchomey** [comment]
Again, I have no conception of how this all works, or even experience with it at any real scale. But I do think @aawsome has a point about it perhaps being an academic/theoretical argument over whether to compress all metadata at once or do so iteratively. 

Why not run an actual test to confirm? At some point closer to release of the compression feature, create or clone a large uncompressed repo (or even a few repos of differing sizes and makeups that cover typical use cases) and process the metadata completely and see how long it takes. If it is fast, then maybe there's no point in adding the complexity of the iterative conversion. 

Also, even if it is slow and the repo is locked for a while, does it really even matter? Perhaps this is my ignorance of restic really showing through, if the repo is locked, who cares? Just let it go... After all, it is to be expected that backups and their associated actions (e.g. prune) take time, and they are typically run/scheduled accordingly. 

Though, I can think of one instance in which this might be a problem and therefore justify the iterative process that @fd0 suggested - frequently scheduled backups. If a scheduled backup runs while the repo is locked, then it could interfere with that in an undesirable way (especially if it is initiated by a script that didn't foresee/account for it taking longer than the backup interval). 

But again, to reiterate the main point from my feedback above - it seems to me that the typical user would place the highest value on this being seamless/invisible. Someone who hasn't read the release notes and just updates the package shouldn't have to adjust or notice anything, other than perhaps noticing at some point that their repo is smaller than it used to be. 

I hope this helps.

**Comment by MichaelEischer** [comment]
> I asked whether we should first try to make a PR of the new specification before beginning with the implementation because I was worrying about exactly this... 😞

The current implementation for the compression feature is rather small (after subtracting the code from #3484 and #3513) so it should be easy to adapt if there are changes to the repo v2 format. And even in case of larger changes, most of the code is likely reusable in one way or another. There are no new tests and no upgrade codepaths etc. yet which are necessary for a full-fledged implementation. So the current state is essentially that of a proof of concept. Having a proof of concept implementation helps with noticing additional assumptions about the repository format used by the code that are not documented (e.g. the pack length sanity checks in `prune`).

The discussion so far feels like we're working with fundamentally different starting assumptions. The current design of this PR largely follows from the following two assumptions:

- The goal is to add compression support, but not to completely redesign the repository format.
- Format upgrades must not require reuploading the whole repository (especially the data blobs).

The first assumption is also partially a derivative of the second assumption: if we still have to read e.g. the format used by the old data pack files, then it feels preferable to just extend the relevant part of the repository format instead of changing it fundamentally, if possible. Making a fundamental change would just result in having to support two formats for a very long time instead of just a single, extended one. I'm not saying that extending the format is the better choice for all new repository features. But I think it is the better choice for compression support. So as long as we keep that second assumption, then we won't arrive at a fundamentally different design compared to this PR.

That is the first assumption is the main reason why the pack header gained support for encoding compressed tree/data blobs.

The requirement to support format upgrades that is contained in the second assumption means that restic must be able to determine precisely which files are encoded according to which repository format version to allow for the upgrade to reliably survive interruptions. As the upgrade process should work within an existing repository, the new format must be able to coexist with data in the old format during upgrades and retain support for the pack header format from repository version v1. Or in other words, **it is just not possible to discuss the v2 format independently from the v1 format**.

For the pack header we have the blob type field which can be used to determine to which format the pack file is compatible. However, for the unpacked files there's no such equivalent so far. Randomly trying out different encodings will result in a rather brittle process over time (which will grow more problematic with each new encoding), such that we have to add some version flag, which defines the encoding format.

An upgrade process can, on the file level, either copy everything to a new repository (which we don't want, but there the file version would always be equivalent to the repo version), copy data to a new folder in the repository (which would then determine the file version) or add the file version somewhere in each file. When using separate folders I'd agree that everything should be migrated at once. However, the version-inside-the-file variant is much more flexible, can easily handle files with different versions and does not require moving data between folders. (The last part also keeps the code simpler as there's no need to access different folders from different versions to upgrade a repository).

For this PR the version-inside-the-file variant works rather well, the only strict requirement is that we add a version indicator to the unpacked files (which should also be added when migrating to a separate folder to avoid related problems in the future). However, as the previous encoding has no such version indication, there are some restrictions on the value the version indicator can have. To avoid possible future problems there I've added a null byte in front of the version byte, such that the version byte can use every value between 0-255.

The additional null byte could be removed by placing some restrictions on the value of the version indicator. What would just work is to say that the first byte has to be below 32 (these values are only ASCII control characters, not printable characters) if it is used as version indicator. That way there can be no collision with valid JSON data.

> But keep in mind, then you'll have these format hacks like this zero-byte and your IMO somewhat awkward pack header format still as relict in your repo 3 version.

We can debate about the zero-byte. But unless you want to force users to fully compress their repositories (including all data), there's no way to get rid of the current pack header format. Thus, adding a new pack header format will just require supporting two pack header variants. (And honestly, I very much doubt that converting the pack header to JSON will make it any prettier).

> And, please, don't be surprised if the first thing users are asking for is a method to completely compress all metadata in their repo once compression is supported.

The intention was to provide some option to compress all tree blobs in one go (or maybe even make that the default and then just use `prune --max-repack-size ...` to limit how much data is rewritten). We can keep on bikeshedding about whether the upgrade must compress everything at once or can do it in multiple steps etc. But fact is that the additional code required to support both uncompressed and compressed data is just a (very) few dozen lines of code. Thus placing a restriction on whether data must be compressed or not is purely artificial. There is no real necessity for that.

> Why not run an actual test to confirm? At some point closer to release of the compression feature, create or clone a large uncompressed repo (or even a few repos of differing sizes and makeups that cover typical use cases) and process the metadata completely and see how long it takes.

We can already answer that question: take a look at how large the cache folder of a repository is, then divide it's size roughly by three (I have no actual numbers to support that guess for the compression factor, which could be higher/lower) and then determine how long it takes to upload that amount of data. So the conversion time primarily depends on the uplink available to a user.

> it seems to me that the typical user would place the highest value on this being seamless/invisible.

There is one point we cannot automate: we have no way of knowing when a user has updated restic everywhere to a version that supports compression. So it will be up to the user to trigger the upgrade at some point. But afterwards there won't be any change. And `prune` should then be able to manage the conversion process.

**Comment by nickchomey** [comment]
The only part I can comment on is this
>There is one point we cannot automate: we have no way of knowing when a user has updated restic everywhere to a version that supports compression. So it will be up to the user to trigger the upgrade at some point

Is this to say that it will be required to manually opt-in to compression via a restic flag? Or is the "trigger" simply upgrading the restic version and compression will be applied automatically? Id prefer the latter, but if not possible then so be it.

Thanks for your hard work!

**Comment by fd0** [comment]
> Is this to say that it will be required to manually opt-in to compression via a restic flag? Or is the "trigger" simply upgrading the restic version and compression will be applied automatically? Id prefer the latter, but if not possible then so be it.

I'd like to add one thing here: I think upgrading an existing repo should be a manual, conscious step, because once executed it will lock out all older versions of restic that only understand the v1 repo format. For new repos, at some point, restic should automatically create them with version 2, so compression is available.

**Comment by fd0** [comment]
I've taken the liberty of rebasing this branch onto master now that #3513 and #3484 have been merged.

File: internal/repository/repository.go [hunk 1]
```diff
@@ -252,12 +272,40 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
...
+func (r *Repository) getZstdDecoder() *zstd.Decoder {
+	r.allocDec.Do(func() {
+		dec, err := zstd.NewReader(nil)
+		if err != nil {
+			panic(err)
+		}
+		r.dec = dec
+	})
+	return r.dec
+}
```
greatroar, line 295: Cosmetic issue (feel free to ignore), but these don't need to be methods on Repository. We could put compression in a submodule internal/repository/zstd with just two functions for the API:

```
var (
	allocEnc sync.Once
	allocDec sync.Once
	enc      *zstd.Encoder
	dec      *zstd.Decoder
)

func Compress(p []byte) []byte {
	allocEnc.Do(func() {
		w, err := zstd.NewWriter(nil)
		if err != nil {
			panic(err)
		}
		enc = w
	})
	return enc.EncodeAll(p, nil)
}

func Decompress(p []byte) []byte { /* defined similarly */ }
```

internal/migrations/upgrade_repo_v2.go
```diff
@@ -30,16 +34,37 @@ func (UpgradeRepoV2) Check(ctx context.Context, repo restic.Repository) (bool, e
 }
 
 func (UpgradeRepoV2) Apply(ctx context.Context, repo restic.Repository) error {
-	cfg := repo.Config()
-	cfg.Version = 2
+	tempdir, err := os.MkdirTemp("", "restic-migrate-upgrade-repo-v2-")
```
greatroar, line 37: Maybe it's time to bump the Go version again? I know you just bumped to 1.15, but even 1.16 isn't getting security fixes anymore.

fd0: Maybe. Hm. I'd like to look at what Debian supports, because that's a widely used but very conservative (*cough*) distribution. Requiring 1.16 would prevent users who use the Debian packages for Go to compile restic on Debian oldstable and Debian stable (without backports). If we have a good reason to raise the version to 1.16 then I think we can do that regardless.

internal/repository/repository.go
fd0: I haven't looked into yet, but I wonder: is that all worth it? Does the zstd encoder/decoder take so much time initializing or does it use a lot of memory?

greatroar: zstd supports parallel compression of a single stream, so NewWriter launches a pool of goroutines. (I don't think NewReader does, though.)

internal/migrations/upgrade_repo_v2.go
greatroar: Debian stable still has 1.15: https://packages.debian.org/bullseye/golang-go

internal/migrations/s3_layout.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	register(S3Layout{})
 }
 
 // S3Layout migrates a repository on an S3 backend from the "s3legacy" to the
 // "default" layout.
 type S3Layout struct{}
 
 // Check tests whether the migration can be applied.
-func (m *S3Layout) Check(ctx context.Context, repo restic.Repository) (bool, error) {
+func (m S3Layout) Check(ctx context.Context, repo restic.Repository) (bool, error) {
```
greatroar, line 25: I'm not sure why this was changed? Having only pointer methods produces smaller code, because for value methods, the compiler generates stubs for the pointer versions too.

internal/migrations/upgrade_repo_v2.go
fd0: Yes, so at the moment we can build `master` with it, but we wouldn't if we were to raise the minimum version to 1.16. That's what I meant.

internal/migrations/s3_layout.go
fd0: Ah, hm. I thought it was unnecessary to have a pointer. I'll revert that. To be honest I'm itching to refactor the migration interface, but I don't think it's worth it...

internal/repository/repository.go
MichaelEischer: The decoder also starts goroutines, see [repository: Fix leaked decoder in StreamPacks](https://github.com/restic/restic/pull/3666/commits/22b1fe791f16b3597e706365b8af1759194f103b). The encoder allocates server MBs of buffers. Without preallocation the compression performance was a complete disaster. With a preallocated encoder the compression performance is fine.

The code in the PR doesn't use the parallel compression feature, for that we'd have to use the `Write`-based API. But as the archiver already has its own goroutines, there's no benefit from introducing lots of synchronization points between goroutines.

MichaelEischer:
I'm wondering whether it would be better to let the `CompressRepoV2` migration just run prune internally, with a parameter to repack all pack files. Right now there is no way to run the migration incrementally, no progress bar, no parallel file deletion, no index sanity checks and probably a few other things I've forgotten. `prune` has already solved all these problems, so I think we should make use of that.

While taking a look at the memory requirements of the zstd library I've noticed that it is also initialized for commands like check which only create lock files. As lock files are very tiny this feels a bit like overkill to me. However, avoiding that would result in adding gone more special case...

fd0:
> I'm wondering whether it would be better to let the CompressRepoV2 migration just run prune internally.

Yes, that's a great idea, I like it!

MichaelEischer:
Adding a `--compress-all-blobs` (or similar) option to prune will be much easier than finding a way to invoke prune from within the migration.

MichaelEischer:
I've cleaned up the commits a bit and moved the design document as first commit. As it hasn't been explicitly mentioned here: the null-byte in the header of unpacked files no longer exists. Instead '[' and '{' are reserved encoding versions which indicate that the unpacked blob is encoded using plain JSON.

MichaelEischer:
The migration support code has been split into PR #3704. For now, by default new repositories are created using repo version 1. To create repository with compression support use `restic init --repository-format 2`

FabioPedretti:
Note that borgbackup has a way to decide how to store a chunk: if the compressed size is >= of the original size, the uncompressed chunk is stored.
This way it avoids storing more data than needed, also decompression is faster.

DRON-666:
I tested the `--compression` option with very compressible data (gigabytes of plain text) and results are disappointing: with `max` level, the backup command is three times slower, but the repo is only 2% smaller.
Anyway, I think it's not very convenient to specify the `--compression` flag in every backup command, and some kind of envvar needs to be added.

Also the message `Added to the repo: xxx GiB` should print compressed size or both compressed and uncompressed sizes.

fd0:
That sounds odd, hm. I'd have expected a significant reduction for that use case.

DRON-666:
> Or the setting `--compression=auto` versus `--compression=max`?

`Auto` vs `Max`, and I even added `fmt.Fprintf(os.Stderr,"Compression level: %v\n\n", level)` to be sure.

>I expect that the compression at the max level will also get better over time when the library improves.

I wouldn't be so sure about it. I tested original ZStandart library (C version) on the same data (`restic cat blob`) and got similar results:
```
LZO     Orig: 22 370 904 894 Comp:11 091 233 627 (49.58%) Time:  1629.157s  
ZSTD  3 Orig: 22 370 904 894 Comp: 8 791 423 753 (39.30%) Time:  1733.385s
ZSTD 22 Orig: 22 370 904 894 Comp: 8 120 666 295 (36.30%) Time: 14189.412s 
```

dimejo:
> If it's the latter then it may be caused by the compression library we're using, compression=max uses the SpeedBestCompression level ([here](https://pkg.go.dev/github.com/klauspost/compress/zstd#EncoderLevel)). There's also the level SpeedBetterCompression, which may be more appropriate. I expect that the compression at the max level will also get better over time when the library improves.

Kopia is using the same compression library (klauspost/compress) and provides some [performance statistics](https://kopia.io/docs/advanced/compression/) for the algorithms it supports. Running the benchmark on a my NAS with a small CPU (Intel Celeron 4205U) shows that throughput for `zstd-better-compression` and `zstd-best-compression` is really bad. Please consider using a faster compression level by default.

```
Compressing input file "/var/log/rest-server.log-20220327" (5.6 MiB) using all compression methods.

     Compression                Compressed   Throughput   Memory Usage
------------------------------------------------------------------------------------------------
  0. s2-parallel-8              2.5 MiB      832.4 MiB/s  1107   18.8 MiB
  1. s2-default                 2.5 MiB      786.7 MiB/s  1122   15.8 MiB
  2. s2-parallel-4              2.5 MiB      763.2 MiB/s  1125   18.8 MiB
  3. zstd-fastest               1.2 MiB      236.5 MiB/s  6056   11.8 MiB
  4. s2-better                  2.3 MiB      228.9 MiB/s  1148   15.6 MiB
  5. pgzip-best-speed           1.3 MiB      182.9 MiB/s  1323   61.3 MiB
  6. lz4                        2.4 MiB      168 MiB/s    96     190.3 MiB
  7. deflate-best-speed         1.3 MiB      108.9 MiB/s  28     5.6 MiB
  8. pgzip                      1.3 MiB      85.5 MiB/s   1305   61 MiB
  9. zstd                       1.3 MiB      72.3 MiB/s   3146   21.3 MiB
 10. gzip-best-speed            1.4 MiB      64.5 MiB/s   34     5.9 MiB
 11. deflate-default            1.3 MiB      49.2 MiB/s   28     5.9 MiB
 12. gzip                       1.3 MiB      38.3 MiB/s   32     5.5 MiB
 13. pgzip-best-compression     1.3 MiB      34.8 MiB/s   1318   63.2 MiB
 14. zstd-better-compression    1.3 MiB      29.5 MiB/s   3092   39.9 MiB
 15. gzip-best-compression      1.2 MiB      18.7 MiB/s   45     5.5 MiB
 16. deflate-best-compression   1.3 MiB      18.7 MiB/s   41     5.9 MiB
 17. zstd-best-compression      1.3 MiB      6.3 MiB/s    3109   102.1 MiB
```

nickchomey:
Seems like zstd fastest is the ideal, based on those stats. It matches my limited observations of compression in general over the years - exponentially diminishing returns for higher levels of compression.

I think it would also be important to run some tests on DEcompression speed as well. Surely some are designed to perform better for that than others. It seems to me that decompression speed is more important - we generally don't pay attention to backups/compression, but recovery/ decompression is a manual (and perhaps even urgent) affair.

Moreover, how does the algorithm affect doing incremental backups (small changes in files as well as add/removal of files) and other repo operations?

colans:
> Seems like zstd fastest is the ideal [...]

`zstd-fastest` makes sense to me, but some folks may not like the high memory consumption.  Are we planning to hard-code the algorithm, or have a default one that users can override with an option?  It may not be worth the trouble to make this configurable, but there are certainly different use cases where different algorithms would be better for different types of data.

nickchomey:
FYI some benchmarks here.

https://peazip.github.io/fast-compression-benchmark-brotli-zstandard.html

Very comprehensive comparisons here https://community.centminmod.com/threads/round-3-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.17259/ and an update here https://community.centminmod.com/threads/round-4-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.18669/

Zstd seems to be the preferred and is designed to have the same fast decompression speed at all compression levels. But lz4 seems to be the fastest decompression. 

I'd be very curious to know if the zstd training mode could be implemented with restic - seems to make a massive difference according to this Facebook doc. 
https://developers.facebook.com/blog/post/2021/08/23/eli5-zstandard-smaller-faster-data-compression/

fd0:
Thanks for all the input! It'll take a while to read everything and do some benchmarks ourselves. Please do not use this PR to bikeshed about the algorithm ;) We're well aware that whatever we settle on it won't make everybody happy. Overall it's very important to keep the complexity within reasonable bounds.

We'll likely stay with zstd and just expose an off|auto|max setting (as currently implemented), but we can still select different compression levels for the settings (auto, max).

klauspost:
Are you planning to compress streams or blocks? 

Default has a lowest streaming speed of about 200MiB per second (usually faster, especially on precompressed) currently which seems like a good match for this. Uses 1-2 cores. 

For blocks you can do multiple concurrency, but default is probably still what you want.

"better" is about 50% slower. "best" down to 10MiB/s, but ofc gives best possible compression of the options. 

See the linked spreadsheet on github. com/klauspost/compress 


(afk, so short answer)

fd0:
Thanks for the fast response! We're compressing blocks of 512KiB to 8MiB and larger JSON documents.

klauspost:
Ok. Maybe "better" with 512KB window size. Especially if you compress blocks concurrently.

"Best" with same window for the slow option.

SaveTheRbtz (commented):
Awesome work, -- saved me about ~30% of space on `/home`.  Very minor comments, plus an ask to maybe put these savings somewhere in the CLI output during/after the backup.

internal/repository/repository.go
```diff
@@ -252,12 +322,51 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
 	return r.idx.LookupSize(restic.BlobHandle{ID: id, Type: tpe})
 }
 
+func (r *Repository) getZstdEncoder() *zstd.Encoder {
+	r.allocEnc.Do(func() {
+		level := zstd.SpeedDefault
+		if r.opts.Compression == CompressionMax {
+			level = zstd.SpeedBestCompression
+		}
```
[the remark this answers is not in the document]
alyandon: From my perspective, even though named constants are nice - I'd much appreciate the ability to specify the compression level numerically just as if I were using the gzip, xz, bzip2, etc.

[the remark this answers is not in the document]
MichaelEischer: restic processes multiple blocks in parallel. so a slight performance increase by a lower compression level won't make that much of a difference in total. Especially, as the compression is not the only processing step.

What is the benefit of providing tons of different compression level?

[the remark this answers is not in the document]
alyandon: For an actual use case - Borg backup exposes zstd + level as options and that allows me to fine tune the overall cpu/runtime/compression of my very large DR backup jobs (backup of backup server basically).  Would it be the end of the world for me if I can't do the same tuning in restic?  Not really.  However, I do currently find actual value in having that level of control.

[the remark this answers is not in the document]
plumbeo: The go implementation of zstandard doesn’t use numeric levels but only descriptive levels: https://github.com/klauspost/compress/blob/c06ba5f93c87a4bb844648f972ac3db430cb4da2/zstd/encoder_options.go#L46

Anyway, two levels (normal and max compression) are more than enough imho, instead I’d spend some time trying to figure out the best parameters for the normal level, since 
breaking files into separate chunks of predictable average size that are then compressed independently, like restic does, is not the most common use case.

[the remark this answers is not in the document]
alyandon: Fair enough - sounds like a lot of extra work for minimal utility.

aawsome:
About compression finetuning:

I think we must differentiate two cases here:
- the data blobs which are mostly unpredictable as they come from the files to backup. Here I agree that optimizing this for a block size ~512kiB may be a good idea
- the metadata, i.e. mainly tree blobs, index files and snapshot files. I think they usually consume the size exactly in this order. Here we have much more information about the raw data to compress, i.e. it is in JSON with a given structure. Moreover, typical sizes may be very different based on the type.

FWIW, here are the statistics from on of my real-life repo:
```
                        min |         max |         avg
index files:          240 B |   6.250 MiB | 736.490 KiB
snapshot files:       225 B |       743 B |       296 B

                        min |         max |         avg
-------------------------------------------------------
data blobs:             1 B |   8.000 MiB | 622.240 KiB
tree blobs:            12 B |   2.832 MiB |  15.589 KiB
-------------------------------------------------------
all blobs:              1 B |   8.000 MiB | 510.849 KiB
```' 

So maybe it might be a food idea to optimize index files around 1MiB, small snapshots files and tree blobs around 100KiB?

> I'd be especially curious to know if the zstd training mode could be implemented with restic - seems to make a massive difference according to this Facebook doc.
> https://developers.facebook.com/blog/post/2021/08/23/eli5-zstandard-smaller-faster-data-compression/

I think that this could be especially worthwhile for tree blobs which always have an almoast identical schema in JSON. @klauspost do you support this kind of training?

**Comment by fd0** [comment]
@aawsome you're right, it probably is a good idea to tune compression for the different use cases (blob vs index). We should land it first though, since this is something we can easily add later :)

**Comment by nickchomey** [comment]
@aawsome 
In this doc, they say the dictionary training is only effective on small files and mostly in the first few KB. https://facebook.github.io/zstd/

would it be too much trouble to add file count and total size to the file type stats you shared? Also, could you either change average to median, or add median? These would help further ascertain which file types would benefit most/at all from learning, and whether it's a useful exercise to begin with (though, presumably, a bit of actual testing wouldn't be terribly difficult - once this feature has already landed, as @fd0 stated).

File: internal/repository/repository.go [hunk 5]
```diff
@@ -252,12 +322,51 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
...
+
+		enc, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(level))
```
klauspost, line 332: Consider if you want `WithEncoderCRC(false)`, if you have checksums elsewhere. It adds a 4 byte xxhash crc. Takes a percent or two of CPU time.

It is nice to detect corruption, but I am not sure if you actually need it.

klauspost, line 332: Consider if you want `WithEncoderCRC(false)`, if you have checksums elsewhere. It adds a 4 byte xxhash crc.

It is nice to detect corruption, but I am not sure if you actually need it.

internal/repository/repository.go
```diff
@@ -252,12 +322,51 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
...
+			panic(err)
+		}
+		r.enc = enc
+	})
+	return r.enc
+}
+
+func (r *Repository) getZstdDecoder() *zstd.Decoder {
+	r.allocDec.Do(func() {
+		dec, err := zstd.NewReader(nil)
```
klauspost, line 343:
```suggestion
		dec, err := zstd.NewReader(nil, zstd.WithDecoderConcurrency(0))
```
... by default 4 decoders are allocated. Might as well have GOMAXPROCS.

klauspost, line 343:
```suggestion
		dec, err := zstd.NewReader(nil, zstd.WithDecoderConcurrency(0))
```
... by default 4 decoders are allocated. Might as well have GOMAXPROCS.

klauspost: Consider adding [WithDecoderMaxMemory](https://pkg.go.dev/github.com/klauspost/compress/zstd#WithDecoderMaxMemory), if known. This will limit the maximum decompressed output size, preventing "zip-bombs" on hostile content.

I don't know enough about restic to know if there is an  uncompressed max size.

klauspost: Consider adding [WithDecoderMaxMemory](https://pkg.go.dev/github.com/klauspost/compress/zstd#WithDecoderMaxMemory), if known. This will limit the maximum decompressed output size, preventing "zip-bombs" on hostile content.

I don't know enough about restic to know if there is an  uncompressed max size.

internal/repository/repository.go
```diff
@@ -252,12 +322,51 @@ func (r *Repository) LookupBlobSize(id restic.ID, tpe restic.BlobType) (uint, bo
...
+	uncompressedLength := 0
+	if r.cfg.Version > 1 {
+
+		// we have a repo v2, so compression is available. if the user opts to
+		// not compress, we won't compress any data, but everything else is
+		// compressed.
+		if r.opts.Compression != CompressionOff || t != restic.DataBlob {
+			uncompressedLength = len(data)
+			data = r.getZstdEncoder().EncodeAll(data, nil)
+		}
```
klauspost, line 367: Since you store information about each block, we might as well just skip storing bigger or equal blocks.

Incompressible block will have a frame and block header, so it will be a few bytes bigger. Also, this reduces the JSON size and removes any need for decompression copying when reading:

```suggestion
		if r.opts.Compression != CompressionOff || t != restic.DataBlob {
			compressed = r.getZstdEncoder().EncodeAll(data, nil)
			if len(compressed) < len(data) {
				uncompressedLength = len(data)
				data = compressed
			} 
		}
```

internal/repository/repository.go
```diff
@@ -822,6 +1001,16 @@ func StreamPack(ctx context.Context, beLoad BackendLoadFn, key *crypto.Key, pack
 			// decryption errors are likely permanent, give the caller a chance to skip them
 			nonce, ciphertext := buf[:key.NonceSize()], buf[key.NonceSize():]
 			plaintext, err := key.Open(ciphertext[:0], nonce, ciphertext, nil)
+			if err == nil && entry.IsCompressed() {
+				if cap(decode) < int(entry.DataLength()) {
+					decode = make([]byte, 0, entry.DataLength())
+				}
```
klauspost, line 1007:
```suggestion
```
I would just leave this out. Since you are using EncodeAll the "frame content size" will be set with the uncompressed data size, so the decoder will do the same as this code. However the soon-to-be-released decoder likes to have a few (for now 16) extra bytes on the output slice, so allocating a precise slice may lead to either a re-alloc or a less optimal code path to be chosen.

You of course couldn't know that, but it would seem beneficial to just leave the potential alloc to the decoder.

klauspost (commented):
Looks solid. Just a few minor proposals.

klauspost:
Looks good. Some minor proposals.

@aawsome As mentioned elsewhere dictionaries only have a small benefit on larger structures, and they carry a computational cost, so unless you are compressing blocks < 10KB, the benefit will be limited.

fd0:
Thanks a lot!

internal/repository/repository.go
fd0: Nice, thanks for the idea! We check a signature on the ciphertext, decrypt it, then decompress the content, then check the SHA256 hash of the decompressed data, so there are enough checks.

internal/repository/repository.go
fd0: Good idea!

internal/repository/repository.go
fd0: Good idea in general, but we'd like to make sure that all tree blobs (JSON documents) are compressed, so we can require that for a later version of the repository format. The code here should always compress all JSON documents (because they compress very well), except maybe the empty tree (`{}`).

internal/repository/repository.go
fd0: Good idea, we'll just do that.

plumbeo (commented):
Won’t this make it impossible to distinguish a blob that doesn’t compress well and should be left as it is from a blob that’s not been compressed yet (for example because compression was disabled or the repository didn’t originally support compression)? Wouldn’t a repo migration or a “prune --compress” command keep finding it and trying to compress it?

edit: sorry, this was about https://github.com/restic/restic/pull/3666/commits/613f97d890ef1723bdc59fe6416cf8649f0c4f6a

klauspost:
@plumbeo `uncompressedLength` being non-zero is what indicates that the block is compressed unless I missed something.

klauspost (approved):
LGTM

MichaelEischer:
> Wouldn’t a repo migration or a “prune --compress” command keep finding it and trying to compress it?

Yes, this leads to prune trying to compress incompressible blobs over and over again. I think we should add some way to mark blobs as incompressible. Such that in total a blob can be marked as uncompressed, as compressed and as incompressible. 

The possible solutions I can think of right now are introducing a new packed blob type (which would be total overkill) or adding a magic value for the uncompressed-length field of compressed data/tree blobs. Redefining the existing blob type that handles uncompressed blobs isn't an option. For the uncompressed-length field we could define that a value of e.g. 0, 1 or -1 marks a blob as incompressible. The problem with `0` would be that a compressed-length of 0 is currently used to mark a blob as uncompressed in memory (and implicitly in the index). As the length field is interpreted as uint32, a value `-1` would encode as 0xffffffff and could collide with a valid value, although we could just forbid that from happening. And `1` relies on the assumption, that zstd always includes a header and frame such that any blob that is shorter when compressed must have an uncompressed length larger than 1. I'm currently a bit undecided whether using -1 or 1 is more elegant. Thoughts?

plumbeo:
> The possible solutions I can think of right now are introducing a new packed blob type (which would be total overkill) or adding a magic value for the uncompressed-length field of compressed data/tree blobs. Redefining the existing blob type that handles uncompressed blobs isn't an option. For the uncompressed-length field we could define that a value of e.g. 0, 1 or -1 marks a blob as incompressible. The problem with `0` would be that a compressed-length of 0 is currently used to mark a blob as uncompressed in memory (and implicitly in the index). As the length field is interpreted as uint32, a value `-1` would encode as 0xffffffff and could collide with a valid value, although we could just forbid that from happening. And `1` relies on the assumption, that zstd always includes a header and frame such that any blob that is shorter when compressed must have an uncompressed length larger than 1. I'm currently a bit undecided whether using -1 or 1 is more elegant. Thoughts?

I hoped to read the whole relevant code instead of just this patch but I didn’t have time, so maybe this won’t be possible because the value isn’t known yet, but what about setting `uncompressedLength = Length` for incompressible blobs?

fd0:
I wonder if the (small) optimization of making some (small) blobs slightly larger introduces a great lot of complexity that's just not worth it...

aawsome:
> Thoughts?

Sorry for playing devils advocate here, but I would say that obviously a "continuous migration" is a bad idea here. 

I wouldn't include the compressing into the standard prune and and only implement a migration which compresses all blobs-to-compress. That is a all-or-nothing migration.
And, well (again being devil's advocate), that might again open the discussion about whether we need to support uncompressed tree blobs/index files/ snapshot files in the v2 repo format...

fd0:
I've removed the commit which skips compression if that makes a blob larger, I'm convince it's not worth the optimization (at least for now).

FabioPedretti:
> I wonder if the (small) optimization of making some blobs slightly smaller by not compressing them introduces a great lot of complexity that's just not worth it...

It's not only the slightly smaller blobs, it's also faster on restore avoiding decompression.
It's also depends on the data type, you may have cases where all the data to backup is already compressed.

greatroar:
Re: already compressed data, based on some very quick testing, it looks like the reference implementation of zstd only adds a few dozen bytes to gzip/bzip2/zstd files. @klauspost, do you know how well your implementation deals with already-compressed inputs in terms of output size and decompression speed?

klauspost:
zstandard requires a frame header. Around 6-8 bytes, and for each 128K block a block header of 3 bytes. Decompression is equivalent to a memcopy. 

gzip headers are 10 bytes and a forced CRC32+size +8 bytes at the end of stream. Deflate (used in gzip) uncompressed blocks are up to 64K and each have 5 byte overhead, decompression is memcopy. In this respect raw deflate is preferable over gzip, since it skips crc and the gzip header.

greatroar:
Then compressing always seems to be a sane choice. The only issue with it is increased memory use, because the compressed and uncompressed data needs to live in memory. But that should max at an additional 8MiB per core, right?

fd0:
Thanks again for the insight Klaus! I also think it's a good choice to always compress the data (if compression is on)

MichaelEischer:
I've extended the repository tests explicitly test both the old and new repository format version.

fd0 (approved):

fd0 (approved):

fd0 (approved):

MichaelEischer:
Thanks a lot for all the discussions and interest in this PR! We'll work on migration support to the new repository format over in #3704.

fd0:
If you have any questions regarding the new compression support **please** head over to [the forum](https://forum.restic.net) and ask there (otherwise we'll drown in comments). This thread is very long already and #3704 is about adding migrations.

Thanks a lot!

PR #3675, opened 2022-03-23
Refactor file handing for self-update.

### LINKED ISSUE #2248: self-update doesn't work on Windows
Output of `restic version`
--------------------------

```
>>> restic self-update
writing restic to C:\<...>\restic\restic.exe
latest version is 0.9.5
download SHA256SUMS
download SHA256SUMS.asc
GPG signature verification succeeded
download restic_0.9.5_windows_amd64.zip
downloaded restic_0.9.5_windows_amd64.zip
Fatal: unable to update restic: unable to remove target file: remove C:\<...>\restic\restic.exe: Access is denied.
```

N/A

Expected behavior
-----------------
Update the binary.

Actual behavior
---------------
restic unable to update itself.

Run `restic self-update` on Windows.

Windows doesn't allow write access to a running executable.

Just an idea:
* Download the new binary
* Rename its old self to `<original-path/name>-old` (or move to `$TMP/<original-name>`)
* Move and rename the new binary to `<original-path/name>`
* Automatically launch the new binary then exit
* The new binary automatically cleans up the old version then exits

Absolutely. Having efficient, regular, and secure backups is awesome.

ItsMattL:
This PR refactors file handling in self-update to better support existing failure modes.

* New file payloads are written to a temp file before touching the original binary. Minimizes the possibility of failing mid-write and corrupting the binary.
* The original binary is moved out to a temp directory rather than removing it, to avoid the binary being locked on some operating systems. 

Closes #2248

Checklist
---------

- [X ] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [X ] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X ] I have added documentation for relevant changes (in the manual).
- [X ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [X ] I have run `gofmt` on the code in all commits.
- [X ] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [X ] I'm done! This pull request is ready for review.

### DIFF
--- a/changelog/unreleased/issue-2248
+++ b/changelog/unreleased/issue-2248
@@ -1,0 +1,8 @@
+Bugfix: Support self-update on Windows
+
+Restic self-update would fail in situations where the operating system
+locks running binaries, including Windows. The new behavior works around
+this by renaming the running file and swapping the updated file in place.
+
+https://github.com/restic/restic/issues/2248
+https://github.com/restic/restic/pull/3675
--- a/internal/selfupdate/download.go
+++ b/internal/selfupdate/download.go
@@ -10,6 +10,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -40,14 +41,6 @@ func findHash(buf []byte, filename string) (hash []byte, err error) {
 }
 
 func extractToFile(buf []byte, filename, target string, printf func(string, ...interface{})) error {
-	var mode = os.FileMode(0755)
-
-	// get information about the target file
-	fi, err := os.Lstat(target)
-	if err == nil {
-		mode = fi.Mode()
-	}
-
 	var rd io.Reader = bytes.NewReader(buf)
 	switch filepath.Ext(filename) {
 	case ".bz2":
@@ -74,33 +67,44 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
 		rd = file
 	}
 
-	err = os.Remove(target)
-	if os.IsNotExist(err) {
-		err = nil
-	}
-	if err != nil {
-		return fmt.Errorf("unable to remove target file: %v", err)
-	}
-
-	dest, err := os.OpenFile(target, os.O_CREATE|os.O_EXCL|os.O_WRONLY, mode)
+	// Write everything to a temp file
+	dir := filepath.Dir(target)
+	new, err := ioutil.TempFile(dir, "restic")
 	if err != nil {
 		return err
 	}
 
-	n, err := io.Copy(dest, rd)
+	n, err := io.Copy(new, rd)
 	if err != nil {
-		_ = dest.Close()
-		_ = os.Remove(dest.Name())
+		_ = new.Close()
+		_ = os.Remove(new.Name())
+		return err
+	}
+	if err = new.Sync(); err != nil {
+		return err
+	}
+	if err = new.Close(); err != nil {
 		return err
 	}
 
-	err = dest.Close()
-	if err != nil {
+	mode := os.FileMode(0755)
+	// attempt to find the original mode
+	if fi, err := os.Lstat(target); err == nil {
+		mode = fi.Mode()
+	}
+
+	// Remove the original binary.
+	if err := removeResticBinary(dir, target); err != nil {
 		return err
 	}
 
-	printf("saved %d bytes in %v\n", n, dest.Name())
-	return nil
+	// Rename the temp file to the final location atomically.
+	if err := os.Rename(new.Name(), target); err != nil {
+		return err
+	}
+
+	printf("saved %d bytes in %v\n", n, target)
+	return os.Chmod(target, mode)
 }
 
 // DownloadLatestStableRelease downloads the latest stable released version of
--- a/internal/selfupdate/download_unix.go
+++ b/internal/selfupdate/download_unix.go
@@ -1,0 +1,10 @@
+//go:build !windows
+// +build !windows
+
+package selfupdate
+
+// Remove the target binary.
+func removeResticBinary(dir, target string) error {
+	// removed on rename on this platform
+	return nil
+}
--- a/internal/selfupdate/download_windows.go
+++ b/internal/selfupdate/download_windows.go
@@ -1,0 +1,23 @@
+//go:build windows
+// +build windows
+
+package selfupdate
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Rename (rather than remove) the running version. The running binary will be locked
+// on Windows and cannot be removed while still executing.
+func removeResticBinary(dir, target string) error {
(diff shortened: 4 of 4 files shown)


internal/selfupdate/download.go
```diff
@@ -74,33 +66,47 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
...
+	// on some platforms (Windows) and cannot be removed while executing.
+	backup := filepath.Join(os.TempDir(), filepath.Base(target)+".bak")
+	if _, err = os.Stat(backup); err == nil {
+		os.Remove(backup)
+	}
+	if err = os.Rename(target, backup); err != nil {
+		return fmt.Errorf("unable to rename target file: %v", err)
+	}
+
+	// Rename the temp file to the final location atomically
```
greatroar, line 104: This isn't fully atomic, because we've written the new file to the tempdir, which may be in a different mount point (on Unix) or drive (on Windows). To get an atomic move, we'd have to put the new file in the same directory:

```
new, err := os.CreateTemp(filepath.Dir(target), "restic")
```

For the same reason, the backup should be in the same directory as well. That also prevents leaking information when the target dir is encrypted but the tempdir is not.

internal/selfupdate/download.go
```diff
@@ -74,33 +66,47 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
...
-	}
+	// Write everything to a temp file
+	new, err := os.CreateTemp("", "restic")
 	if err != nil {
-		return fmt.Errorf("unable to remove target file: %v", err)
+		return err
 	}
 
-	dest, err := os.OpenFile(target, os.O_CREATE|os.O_EXCL|os.O_WRONLY, mode)
+	n, err := io.Copy(new, rd)
```
greatroar, line 75: I suggest Sync'ing the file before trying to move it in place. This only works when it's in the same dir as the target, since moving across mount points means copy-then-remove.

```suggestion
	n, err := io.Copy(new, rd)
	if err == nil {
		err = new.Sync()
	}
```

internal/selfupdate/download.go
ItsMattL: Done. I'll note that the main tradeoff here is that we don't get any automatic cleanup of these temp files in this case. Should the operation fail for some reason, the temp file could theoretically be orphaned in the working directory... Although this should hopefully be a very rare occurrence.

internal/selfupdate/download.go
ItsMattL: Done

internal/selfupdate/download.go
```diff
@@ -74,33 +66,50 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
...
+
+	mode := os.FileMode(0755)
+	// attempt to find the original mode
+	if fi, err := os.Lstat(target); err == nil {
+		mode = fi.Mode()
+	}
+
+	// Rename (rather than remove) the running version. The running binary will be locked
+	// on some platforms (Windows) and cannot be removed while executing.
+	backup := filepath.Join(dir, filepath.Base(target)+".bak")
```
MichaelEischer, line 99: That will also leave behind a backup file on platforms where the running binary can be removed. What using the backup file codepath only on windows and just move the new version to the target everywhere else?

fd0:
FYI, if you rebase on the master branch now, you don't have to be compatible to Go 1.14 any more (so you can use the new tempfile method). I've merged #3680 yesterday which raises the minimum version of Go to 1.15. :)

ItsMattL:
Behavior is now split by Windows and non-Windows as requested. The functions are still probably the right ones despite the removal of 1.14, as it looks like CreateTemp didn't come around until 1.16.

internal/selfupdate/download.go
```diff
@@ -74,33 +67,45 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
 		rd = file
 	}
 
-	err = os.Remove(target)
-	if os.IsNotExist(err) {
-		err = nil
-	}
+	// Write everything to a temp file
+	dir := filepath.Dir(target)
+	new, err := ioutil.TempFile(dir, "restic")
```
MichaelEischer, line 72: Since #3680 there's no need to maintain compatibility with Go 1.14. Can you convert this back to use os.CreateTemp ?

changelog/unreleased/issue-2248
```diff
@@ -0,0 +1,7 @@
+Bugfix: Support self-update on Windows
+
+Restic self-update would fail in situations where the operating system
+locks running binaries, including Windows. The new behavior works around
+this by renaming the running file and swapping the updated file in place.
+
+https://github.com/restic/restic/issues/2248
```
MichaelEischer, line 7: Please add a link to this PR.

internal/selfupdate/download.go
```diff
@@ -74,33 +67,45 @@ func extractToFile(buf []byte, filename, target string, printf func(string, ...i
...
+	}
+	if err = new.Close(); err != nil {
 		return err
 	}
 
-	n, err := io.Copy(dest, rd)
-	if err != nil {
-		_ = dest.Close()
-		_ = os.Remove(dest.Name())
+	printf("saved %d bytes in %v\n", n, new.Name())
```
MichaelEischer, line 90: That prints the tempfile name, which is rather confusing:
```
saved 18472960 bytes in /restic/restic/restic3463974662
```
 Please print the output using the final filename and print it only once the update is complete.

internal/selfupdate/download_unix.go
```diff
@@ -0,0 +1,11 @@
+//go:build !windows
+// +build !windows
+
+package selfupdate
+
+import "os"
+
+// Remove the target binary.
+func removeResticBinary(dir, target string) error {
+	return os.Remove(target)
```
MichaelEischer, line 10: I wonder whether we should let `os.Rename`  take care of replacing the restic binary, instead of manually deleting it first. That would make the update truly atomic.

MichaelEischer (changes requested):
I've added a few comments that have to be addressed, but otherwise the PR is already in good shape. Could you squash the commits (or at least reduce their number a bit?)

internal/selfupdate/download.go
MichaelEischer: Haven't seen the discussion above. -> Ignore this comment.

internal/selfupdate/download_unix.go
ItsMattL: changed

MichaelEischer:
I've fixed the build failure and slightly edited to commit message to no longer mention a temp directory but rather a temp file.

MichaelEischer (approved):
LGTM.

PR #3691, opened 2022-03-31
Sanitize tags when used as filenames by restic mount

### LINKED ISSUE #3690: Input/output error with restic mount for tags with slashes
Output of `restic version`
--------------------------
restic 0.13.0 compiled with go1.18 on linux/amd64

```
restic backup --tag foo/bar /bin/sh
restic mount path
ls -la path/tags
```

Local directory, but had the same happen with sftp, most likely irrelevant.

Expected behavior
-----------------
No error

Actual behavior
---------------
ls: reading directory '/tmp/tmp.pq04QRmLVl/mount/tags': Input/output error

```sh
#!/bin/sh

set -eux

restic version
d=$(mktemp -d)
mkdir "$d"/{repo,mount}

export RESTIC_REPOSITORY="$d/repo"
export RESTIC_PASSWORD=insecure

restic init
restic backup --tag foo/bar /bin/sh
restic mount "$d/mount" &
sleep 5
ls -la "$d/mount/tags"
```

Seems similar to #1480, probably invalid file names are passed to fuse.

Either forbid tag names containing slashes (or `.` and `..` as tag names), or sanitize the directory names.

I didn't have the need to restore a backup *today*, so no, but in general, it does a great job.

greatroar:

Draft PR, untested.

Tags that are not valid filenames (and some that are) are now sanitized by restic mount for display in the tags/ directory.

The rule is explained in the comment:

```
// Some tags are problematic when used as filenames:
//
//     ""
//     ".", ".."
//     anything containing '/'
//
// We prepend a '%' to mark these as special, then replace '/' by "%_"
// and '%' by "%%". Because this produces ambiguity when the actual tag
// starts with '%', we do the same to any tag name already starting with '%'.
```

E.g., "foo/bar" becomes "%foo%_bar".

I chose '%' because it should be rare in actual tags and it's not a special character to common shells. The rule is ad hoc, but designed to pass through as many tags unchanged as possible. The resulting filenames are always valid on Linux; to my knowledge, FreeBSD allows the same filenames as Linux; but I didn't really consider MacOS much, which may or may not require valid (normalized?) Unicode filenames.

Should fix #3690, but I haven't tested it yet.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2907
+++ b/changelog/unreleased/issue-2907
@@ -6,5 +6,10 @@ formatting of the time for a snapshot is now controlled using `--time-template`
 and supports subdirectories to for example group snapshots by year. Please
 refer to the help output of the `mount` command for further details.
 
+Characters in tag names which are not allowed in a filename are replaced by
+underscores `_`. For example a tag `foo/bar` will result in a directory name
+of `foo_bar`.
+
 https://github.com/restic/restic/issues/2907
 https://github.com/restic/restic/pull/2913
+https://github.com/restic/restic/pull/3691
--- a/internal/fuse/snapshots_dirstruct.go
+++ b/internal/fuse/snapshots_dirstruct.go
@@ -97,6 +97,7 @@ func pathsFromSn(pathTemplate string, timeTemplate string, sn *restic.Snapshot) 
 				// needs special treatment: Rebuild the string builders
 				newout := make([]strings.Builder, len(out)*len(sn.Tags))
 				for i, tag := range sn.Tags {
+					tag = filenameFromTag(tag)
 					for j := range out {
 						newout[i*len(out)+j].WriteString(out[j].String() + tag)
 					}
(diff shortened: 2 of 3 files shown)


fd0:
Hm, why not just replace `/` by `_` and `_` by `__`? Isn't `%` a special character in `cmd.exe`, for environment variable expansion?

I find this easier to read:
  * `foo/bar` becomes `foo_bar`
  * `foo_bar` becomes `foo__bar`

rawtaz:
Sorry if I'm dense, but instead of making hacky workarounds, why don't we simply disallow the offending character(s) in tags entirely?

That will keep things clean, remove the need for hacks, and it's IMO fully reasonable to not allow *everything* in certain types of data, in particular "identifiers" like this which aren't really meant to consist of an essay of some language that require the entire Unicode charts to convey its meanings..

greatroar:
@rawtaz I can imagine people backing up web data and using URLs as tags...

MichaelEischer:
This will be pretty easy to implement once #2913 is merged.

MichaelEischer:
I've rebased the PR resolve the conflicts with #2913.

> Maybe we need to approach this differently and just map / to _, then accept that tags/foo_bar holds snapshots labeled either foo_bar or foo/bar. It's just a navigational aid, after all.
> 
> (The alternative of creating a directory structure representing foo/bar/baz is more complicated and also can't handle the case of leading slash, trailing slash or multiple consecutive slashes.)

The tree structure is what is now used since #2913 is merged. Multiple unexpected slashes will then be removed by the filename normalization, such that we can't really avoid aliasing there. 

My preferred solution would be to just map slashes to underscores and ignore aliasing. The second commit does just that.

I though about implementing a fallback strategy to generate unique tag names by appending a counter in case conflicts exist. However, that leads to rather unintelligible tag names such as `_` and `_-1`, which won't help anyone.

fd0:
Looks good :)

MichaelEischer (approved):
I've squashed the commits and added a small changelog remark.

PR #3704, opened 2022-04-11
Support migration to repository format with compression

MichaelEischer:
This is the second part of #3666. It adds support to migrate existing repositories to the latest repository format version.

Left-over todos from #3666:
- [x] Tests which explicitly test both old and new repository format
- [x] Document --repository-format option
- [x] Document format migration
- [x] Improve migration codepaths
- [x] check progress for the migrate command 

Fixes #3666.

Checklist
---------


fd0:
I've rebased this branch onto #3666

kakra:
This probably fits better into #3666 but it's already closed.

How does the new feature handle data that compresses badly? I think, data should be stored uncompressed if

1. the compressed blob is bigger than the uncompressed blob
2. the compressed blob isn't at least one storage device block smaller

This implies that maybe a fast heuristic algorithm could be used to detect if data can be compressed at all by a reasonable amount, e.g. by using a similar approach btrfs uses: https://www.spinics.net/lists/linux-btrfs/msg66839.html

MichaelEischer:
> How does the new feature handle data that compresses badly?

This has already been discussed e.g. in https://github.com/restic/restic/pull/3666#issuecomment-1107903834 . We've decided that it's not worth the additional complexity. zstd will only introduce a few bytes of overhead for incompressible blobs.

thedaveCA:
Is restic able to read (or the upgrade process able to handle) a single repository that contains a mix of upgraded and non-upgraded content at the same time?

What if non-upgraded files are added back to a repository, can the upgrade be run again?

fd0:
> Is restic able to read (or the upgrade process able to handle) a single repository that contains a mix of upgraded and non-upgraded content at the same time?

Yes, the on-disk format is backwards compatible. If you have a v2 repo and run `restic backup --compression=off` then it'll even add more uncompressed data.

> What if non-upgraded files are added back to a repository, can the upgrade be run again?

I'm not sure what you mean by that.

Would you mind opening a new topic in [the forum](https://forum.restic.net) for all other questions? We're using the GitHub for code and I'd like to keep the discussion in this PR related to the code. I'll respond to all your questions in the forum. Thanks! :)

fd0:
Please keep the comments and discussions in this PR related to the code (otherwise we'll get buried by comments!).

Please ask questions in [the forum](https://forum.restic.net/). Thanks a lot!

internal/migrations/upgrade_repo_v2.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+type UpgradeRepoV2Error struct {
+	UploadNewConfigError   error
+	ReuploadOldConfigError error
+
+	BackupFilePath string
+}
+
+func (err UpgradeRepoV2Error) Error() string {
```
greatroar, line 24: I suggest using pointer receivers on this type, and in fact, on all custom error types. That avoids the question "do I check for UpgradeRepoV2Error, or *UpgradeRepoV2Error"? The stdlib does this for all but some of the oldest error types.

(Not a big deal here, but it sets an example for other contributors.)

internal/migrations/upgrade_repo_v2.go
```diff
@@ -0,0 +1,121 @@
...
+func (*UpgradeRepoV2) Desc() string {
+	return "upgrade a repository to version 2"
+}
+
+func (*UpgradeRepoV2) Check(ctx context.Context, repo restic.Repository) (bool, error) {
+	if repo.Config().Version == 1 {
+		return true, nil
+	}
+
+	return false, nil
```
greatroar, line 52:
```suggestion
	isV1 := repo.Config().Version == 1
	return isV1, nil
```

internal/migrations/upgrade_repo_v2.go
```diff
@@ -0,0 +1,121 @@
...
+		_ = repo.Backend().Remove(ctx, h)
+		err = repo.Backend().Save(ctx, h, restic.NewByteReader(rawConfigFile, nil))
+		if err != nil {
+			repoError.ReuploadOldConfigError = err
+		}
+
+		return repoError
+	}
+
+	return nil
```
greatroar, line 120:
```suggestion
	os.RemoveAll(tempdir)
	return nil
```

internal/migrations/upgrade_repo_v2_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if upgradeErr.ReuploadOldConfigError == nil {
+		t.Fatal("expected reupload error, got nil")
+	}
+
+	if upgradeErr.BackupFilePath == "" {
+		t.Fatal("no backup file path found")
+	}
+}
```
greatroar, line 107:
```suggestion
	os.RemoveAll(filepath.Dir(upgradeErr.BackupFilePath))
}
```

internal/migrations/compress_repo_v2.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (*CompressRepoV2) Check(ctx context.Context, repo restic.Repository) (bool, error) {
+	// only do very fast checks on the version here, we don't want the list of
+	// available migrations to take long to load
+	if repo.Config().Version < 2 {
+		return false, nil
+	}
+
+	return true, nil
```
greatroar, line 32:
```suggestion
	isV1 := repo.Config().Version < 2 {
	return isV1, nil
```

internal/migrations/compress_repo_v2.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Apply requires that the repository must be already locked exclusively, this
```
greatroar, line 35:
```suggestion
// Apply requires that the repository is already locked exclusively. This
```

internal/migrations/compress_repo_v2.go
```diff
@@ -0,0 +1,102 @@
...
+	packsWithUncompressedData := restic.NewIDSet()
+	keepBlobs := restic.NewBlobSet()
+
+	for blob := range repo.Index().Each(ctx) {
+		keepBlobs.Insert(blob.BlobHandle)
+
+		if blob.UncompressedLength != 0 {
+			// blob is already compressed, ignore
+			continue
+
```
greatroar, line 56:
```suggestion
```

internal/migrations/upgrade_repo_v2.go
MichaelEischer: Is changed. I've noticed that using the pointer receiver the compiler actually enforces the use of the pointer type when checking for a specific type. So it's both consistent and cannot be used the wrong way :) .

internal/migrations/upgrade_repo_v2.go
MichaelEischer: I've used two plain `os.Remove` calls to explicitly delete the backup file and the temporary folder.

internal/migrations/upgrade_repo_v2_test.go
MichaelEischer: Same here, using two explicit `os.Remove` calls should make it less likely to accidentally delete something important.

MichaelEischer:
I've replaced the compress_repo_v2 migration with a new parameter for `prune`. That way we get all the functionality of prune like limiting how much data is rewritten and also avoid correctness issues from creeping in.

MichaelEischer:
I wonder whether the upgrade_repo_v2 migration should first try to just overwrite the `config` file, all backends except the rest-server and the mem backend seem to allow that. It would result in slightly different codepaths depending on the backend, but would have the huge benefit for backends that allow atomic replacements, that there's less that can go wrong.

fd0:
Oh, good idea! I didn't know that we allow overwriting files in the backends. I thought we open files (in sftp and local) with `O_CREATE` and `O_EXCL`... at least we used to do that at some point.

greatroar:
@fd0 And they should: I suspect what @MichaelEischer means is "create at temporary path" + "move into place".

fd0:
Ah! It was dropped when we switched to temporary files. Hm. Maybe we should look into restoring that behavior with local and sftp backends (in a separate issue though).

MichaelEischer:
> @fd0 And they should: I suspect what @MichaelEischer means is "create at temporary path" + "move into place".

Since the changes to atomically create files in local/sftp, these backends already creates a temporary file and then move it into place. The Azure/GS/S3/Swift backends also support atomic replacements for files.

> Maybe we should look into restoring that behavior with local and sftp backends (in a separate issue though).

For what purpose? The normal files which use a hash as filename don't care because the file content will be identical. And for the config file it shouldn't matter either. Either you're able to overwrite/delete that file or not.

MichaelEischer:
Ah, I just noticed that sftp doesn't overwrite existing files. It's using `Rename` and not `PosixRename`. `Rename` uses the semantic where the target file must not exist yet. AFAIK the openssh implementation hardlinks the temporary file to its new name before deleting the directory entry for the temporary file. I'm not sure it's worth the trouble to replicate that behavior for the local backend.

And for the cloud backends we cannot prevent atomic overwrites either, we could check whether a file exists first, but that costs an additional API call.

MichaelEischer:
The repo format migration now sort of runs `check` before upgrading to ensure that the repository is in a good state. There is no interactive output yet as that requires some changes to how migrations works. The upgrade checks are ideally combined with #3730 which warns about a few legacy variants of repo v1.

MichaelEischer:
The migrate command now directly runs check for `upgrade_repo_v2`. That way we get check progress information and don't have to reimplement check within a migration.

fd0 (approved):

PR #3730, opened 2022-04-30
Let `check` warn about legacy variants of the repo format version 1

### LINKED ISSUE #3295: Improve messaging from "restic check --check-unused" after forget and prune with --max-unused > 0
Output of `restic version`
--------------------------

Running "restic check --check-unused" after successfully completing a forget --prune operation using the new (>= 0.12.0) prune version with a non-zero --max-unused value can produce thousands of "unused blob <data/xxxxxxxx>" messages followed by "Fatal: repository contains errors" even though the repository is perfectly healthy.

Apart from filling the log with unnecessary information, the casual user could easily interpret these messages as indicating a problem with the repo when in fact none exists. 

Some possible ways to address this (none of them great; hopefully others will have better ideas) include:

--have restic track which blobs are intentionally unused (i.e. exist as a consequence of a successful forget and prune) and suppress the --check-unused messages (especially any "Fatal" message) attributable to these blobs.

--put an explanation in the documentation of the "--check-unused" option alerting the user of the potential for these messages and appropriate interpretation

--phase out/deprecate the --check-unused flag to restic clean.  With the new default prune behavior, the value of this check may not be what it used to be.

Avoid user confusion associated by with seeing many unexpected messages and the word "Fatal" in restic check output when in fact no issue exists.

Restic is great!  And the speed improvement with the "new prune" is amazing...many thanks!

MichaelEischer:
The legacy S3 repository layout and mixed packs are not used for new repositories since quite some time. Especially the mixed packs are a problem as they will drastically increase the cache size and slow down everything. This PR let's  check complain about these legacy variants of the repository format version 1.

The mixed packs warning leads to the somewhat ugly situation that the test data used for the check is also affected by that warning. I've decided to simply filter these warnings, as that way the test implicitly tests that we are still able to read old repositories.

The `--check-unused` option is also deprecated as it is now rather useless since `prune` no longer removes all unused blob by default. This removes the option from the CLI, but keeps it for internal test case usage.

No. I've only complained about `--check-unused` being confusing somewhere in the forum ;-) .
Fixes #3295

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3295
+++ b/changelog/unreleased/issue-3295
@@ -1,0 +1,12 @@
+Change: Deprecate `check --check-unused` and add further checks
+
+Since restic 0.12.0, it is expected to still have unused blobs after running
+`prune`. This made the `check --check-unused` rather useless and tended to
+confuse users. The options has been deprecated and is now ignored.
+
+`check` now also warns if a repository is using either the legacy S3 layout or
+mixed pack files with both tree and data blobs. The latter is known to cause
+performance problems.
+
+https://github.com/restic/restic/issues/3295
+https://github.com/restic/restic/pull/3730
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -221,11 +221,15 @@ func runCheck(opts CheckOptions, gopts GlobalOptions, args []string) error {
 
 	errorsFound := false
 	suggestIndexRebuild := false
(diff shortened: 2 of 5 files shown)


MichaelEischer:
I've rebased the PR, added slightly reworded the warning printed for mixed packs and added a changelog entry. Besides that the PR is essentially unchanged.

PR #3731, opened 2022-04-30
Feature: min packsize flag

### LINKED ISSUE #2291: Ability to tweak chunk pack size
Output of `restic version`
--------------------------
➜  ~ restic128mb version
restic 0.9.5 compiled with go1.11.1 on linux/amd64

Restic has https://github.com/restic/restic/blob/master/internal/repository/packer_manager.go#L39
and https://github.com/restic/restic/blob/master/internal/restorer/filerestorer.go#L30 which specify approximate pack size which restic tries to target.

A good number of remote data storage have a limitation on the number of files, I have run into issues with Google Team Drive, and PUR based storage at work (it is full flash system, with glusterfs over the top). It should be reasonable to update the pack size on a repo level to enable varied use-cases.

What are you trying to do?
--------------------------
Either store the target pack size in the repo metadata or allow user to pass the parameter while backup (and maybe try to figure out the appropriate size for cache at the time of restore, for now a simple option to be given both times should work well)

Thank fully I haven;t had to restore data in an emergency, but I have tried to restore data as a disaster recovery exercise which provides comfort that restic has my back.

metalsp0rk:

Would like to add the following in order to tune restic rather than hardcoding it:

`--pack-size`

https://forum.restic.net/t/control-the-minimal-pack-files-size/617

Supersedes #2750
Closes #2291

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -535,13 +535,13 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		pi := repackCandidates[i].packInfo
 		pj := repackCandidates[j].packInfo
 		switch {
-		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(minPackSize) && pj.unusedSize+pj.usedSize >= uint64(minPackSize):
-			return true
-		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(minPackSize) && pi.unusedSize+pi.usedSize >= uint64(minPackSize):
-			return false
 		case pi.tpe != restic.DataBlob && pj.tpe == restic.DataBlob:
 			return true
 		case pj.tpe != restic.DataBlob && pi.tpe == restic.DataBlob:
+			return false
+		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(minPackSize) && pj.unusedSize+pj.usedSize >= uint64(minPackSize):
+			return true
+		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(minPackSize) && pi.unusedSize+pi.usedSize >= uint64(minPackSize):
 			return false
 		}
 		return pi.unusedSize*pj.usedSize > pj.unusedSize*pi.usedSize
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -426,8 +426,12 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	var repackCandidates []packInfoWithID
 	var repackSmallCandidates []packInfoWithID
 	repoVersion := repo.Config().Version
-	// consider files with at least 80% of the target size as large enough
-	targetPackSize := repo.PackSize() / 5 * 4
+	// only repack very small files by default
+	targetPackSize := repo.PackSize() / 25
+	if opts.RepackSmall {
+		// consider files with at least 80% of the target size as large enough
+		targetPackSize = repo.PackSize() / 5 * 4
+	}
 
 	// loop over all packs and decide what to do
 	bar := newProgressMax(!gopts.Quiet, uint64(len(indexPack)), "packs processed")
@@ -469,8 +473,6 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		// use a flag that pack must be compressed
 		p.uncompressed = mustCompress
 
-		packIsLargeEnough := !opts.RepackSmall || packSize >= int64(targetPackSize)
-
 		// decide what to do
 		switch {
 		case p.usedBlobs == 0:
@@ -483,8 +485,8 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			// if this is a data pack and --repack-cacheable-only is set => keep pack!
 			stats.packs.keep++
 
-		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress && packIsLargeEnough:
-			if packIsLargeEnough {
+		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress:
+			if packSize >= int64(targetPackSize) {
 				// All blobs in pack are used and not mixed => keep pack!
 				stats.packs.keep++
 			} else {
@@ -544,7 +546,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	// This is equivalent to sorting by unused / total space.
 	// Instead of unused[i] / used[i] > unused[j] / used[j] we use
 	// unused[i] * used[j] > unused[j] * used[i] as uint32*uint32 < uint64
-	// Morover packs containing trees are sorted to the beginning
+	// Moreover packs containing trees and too small packs are sorted to the beginning
 	sort.Slice(repackCandidates, func(i, j int) bool {
 		pi := repackCandidates[i].packInfo
 		pj := repackCandidates[j].packInfo
@@ -553,9 +555,9 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			return true
 		case pj.tpe != restic.DataBlob && pi.tpe == restic.DataBlob:
 			return false
-		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(targetPackSize) && pj.unusedSize+pj.usedSize >= uint64(targetPackSize):
+		case pi.unusedSize+pi.usedSize < uint64(targetPackSize) && pj.unusedSize+pj.usedSize >= uint64(targetPackSize):
 			return true
-		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(targetPackSize) && pi.unusedSize+pi.usedSize >= uint64(targetPackSize):
+		case pj.unusedSize+pj.usedSize < uint64(targetPackSize) && pi.unusedSize+pi.usedSize >= uint64(targetPackSize):
 			return false
 		}
 		return pi.unusedSize*pj.usedSize > pj.unusedSize*pi.usedSize
@@ -575,7 +577,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	for _, p := range repackCandidates {
 		reachedUnusedSizeAfter := (stats.size.unused-stats.size.remove-stats.size.repackrm < maxUnusedSizeAfter)
 		reachedRepackSize := stats.size.repack+p.unusedSize+p.usedSize >= opts.MaxRepackBytes
-		packIsLargeEnough := !opts.RepackSmall || p.unusedSize+p.usedSize >= uint64(targetPackSize)
+		packIsLargeEnough := p.unusedSize+p.usedSize >= uint64(targetPackSize)
 
 		switch {
 		case reachedRepackSize:
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -71,7 +71,7 @@ func addPruneOptions(c *cobra.Command) {
 	f.StringVar(&pruneOptions.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "maximum `size` to repack (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&pruneOptions.RepackCachableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
-	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "also repack small packs")
+	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack too small packs")
 	f.BoolVar(&pruneOptions.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
 }
 
@@ -424,8 +424,10 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	repackPacks := restic.NewIDSet()
 
 	var repackCandidates []packInfoWithID
+	var repackSmallCandidates []packInfoWithID
 	repoVersion := repo.Config().Version
-	targetPackSize := repo.PackSize()
+	// consider files with at least 80% of the target size as large enough
+	targetPackSize := repo.PackSize() / 5 * 4
 
 	// loop over all packs and decide what to do
 	bar := newProgressMax(!gopts.Quiet, uint64(len(indexPack)), "packs processed")
@@ -482,8 +484,12 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			stats.packs.keep++
 
 		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress && packIsLargeEnough:
-			// All blobs in pack are used and not mixed => keep pack!
-			stats.packs.keep++
+			if packIsLargeEnough {
+				// All blobs in pack are used and not mixed => keep pack!
+				stats.packs.keep++
+			} else {
+				repackSmallCandidates = append(repackSmallCandidates, packInfoWithID{ID: id, packInfo: p})
+			}
 
 		default:
 			// all other packs are candidates for repacking
@@ -524,6 +530,14 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		for id := range ignorePacks {
 			Warnf("will forget missing pack file %v\n", id)
 		}
+	}
+
+	if len(repackSmallCandidates) < 10 {
+		// too few small files to be worth the trouble, this also prevents endlessly repacking
+		// if there is just a single pack file below the target size
+		stats.packs.keep += uint(len(repackSmallCandidates))
+	} else {
+		repackCandidates = append(repackCandidates, repackSmallCandidates...)
 	}
 
 	// Sort repackCandidates such that packs with highest ratio unused/used space are picked first.
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -52,6 +52,7 @@ type PruneOptions struct {
 	MaxRepackBytes uint64
 
 	RepackCachableOnly bool
+	RepackSmall        bool
 	RepackUncompressed bool
 }
 
@@ -70,6 +71,7 @@ func addPruneOptions(c *cobra.Command) {
 	f.StringVar(&pruneOptions.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "maximum `size` to repack (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&pruneOptions.RepackCachableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
+	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "also repack small packs")
 	f.BoolVar(&pruneOptions.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
 }
 
@@ -423,6 +425,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 
 	var repackCandidates []packInfoWithID
 	repoVersion := repo.Config().Version
+	minPackSize := repo.MinPackSize()
 
 	// loop over all packs and decide what to do
 	bar := newProgressMax(!gopts.Quiet, uint64(len(indexPack)), "packs processed")
@@ -464,6 +467,8 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		// use a flag that pack must be compressed
 		p.uncompressed = mustCompress
 
+		packIsLargeEnough := !opts.RepackSmall || packSize >= int64(minPackSize)
+
 		// decide what to do
 		switch {
 		case p.usedBlobs == 0:
@@ -476,7 +481,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			// if this is a data pack and --repack-cacheable-only is set => keep pack!
 			stats.packs.keep++
 
-		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress:
+		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress && packIsLargeEnough:
 			// All blobs in pack are used and not mixed => keep pack!
 			stats.packs.keep++
 
@@ -530,6 +535,10 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		pi := repackCandidates[i].packInfo
 		pj := repackCandidates[j].packInfo
 		switch {
+		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(minPackSize) && pj.unusedSize+pj.usedSize >= uint64(minPackSize):
+			return true
+		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(minPackSize) && pi.unusedSize+pi.usedSize >= uint64(minPackSize):
+			return false
 		case pi.tpe != restic.DataBlob && pj.tpe == restic.DataBlob:
 			return true
 		case pj.tpe != restic.DataBlob && pi.tpe == restic.DataBlob:
@@ -552,6 +561,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	for _, p := range repackCandidates {
 		reachedUnusedSizeAfter := (stats.size.unused-stats.size.remove-stats.size.repackrm < maxUnusedSizeAfter)
 		reachedRepackSize := stats.size.repack+p.unusedSize+p.usedSize >= opts.MaxRepackBytes
+		packIsLargeEnough := !opts.RepackSmall || p.unusedSize+p.usedSize >= uint64(minPackSize)
 
 		switch {
 		case reachedRepackSize:
@@ -561,7 +571,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			// repacking non-data packs / uncompressed-trees is only limited by repackSize
 			repack(p.ID, p.packInfo)
 
-		case reachedUnusedSizeAfter:
+		case reachedUnusedSizeAfter && packIsLargeEnough:
 			// for all other packs stop repacking if tolerated unused size is reached.
 			stats.packs.keep++
 
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -425,7 +425,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 
 	var repackCandidates []packInfoWithID
 	repoVersion := repo.Config().Version
-	minPackSize := repo.MinPackSize()
+	targetPackSize := repo.PackSize()
 
 	// loop over all packs and decide what to do
 	bar := newProgressMax(!gopts.Quiet, uint64(len(indexPack)), "packs processed")
@@ -467,7 +467,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 		// use a flag that pack must be compressed
 		p.uncompressed = mustCompress
 
-		packIsLargeEnough := !opts.RepackSmall || packSize >= int64(minPackSize)
+		packIsLargeEnough := !opts.RepackSmall || packSize >= int64(targetPackSize)
 
 		// decide what to do
 		switch {
@@ -539,9 +539,9 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 			return true
 		case pj.tpe != restic.DataBlob && pi.tpe == restic.DataBlob:
 			return false
-		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(minPackSize) && pj.unusedSize+pj.usedSize >= uint64(minPackSize):
+		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(targetPackSize) && pj.unusedSize+pj.usedSize >= uint64(targetPackSize):
 			return true
-		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(minPackSize) && pi.unusedSize+pi.usedSize >= uint64(minPackSize):
+		case opts.RepackSmall && pj.unusedSize+pj.usedSize < uint64(targetPackSize) && pi.unusedSize+pi.usedSize >= uint64(targetPackSize):
 			return false
 		}
 		return pi.unusedSize*pj.usedSize > pj.unusedSize*pi.usedSize
@@ -561,7 +561,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, gopts GlobalOption
 	for _, p := range repackCandidates {
 		reachedUnusedSizeAfter := (stats.size.unused-stats.size.remove-stats.size.repackrm < maxUnusedSizeAfter)
 		reachedRepackSize := stats.size.repack+p.unusedSize+p.usedSize >= opts.MaxRepackBytes
-		packIsLargeEnough := !opts.RepackSmall || p.unusedSize+p.usedSize >= uint64(minPackSize)
+		packIsLargeEnough := !opts.RepackSmall || p.unusedSize+p.usedSize >= uint64(targetPackSize)
 
 		switch {
 		case reachedRepackSize:
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -71,7 +71,7 @@ func addPruneOptions(c *cobra.Command) {
 	f.StringVar(&pruneOptions.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "maximum `size` to repack (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&pruneOptions.RepackCachableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
-	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack too small packs")
+	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack pack files below 80%% of target pack size")
 	f.BoolVar(&pruneOptions.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
 }
 
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -126,7 +126,7 @@ func init() {
 	f.Var(&globalOptions.Compression, "compression", "compression mode (only available for repository format version 2), one of (auto|off|max)")
 	f.IntVar(&globalOptions.Limits.UploadKb, "limit-upload", 0, "limits uploads to a maximum rate in KiB/s. (default: unlimited)")
 	f.IntVar(&globalOptions.Limits.DownloadKb, "limit-download", 0, "limits downloads to a maximum rate in KiB/s. (default: unlimited)")
-	f.UintVar(&globalOptions.PackSize, "pack-size", uint(targetPackSize), "set target pack size in MiB. (default: $RESTIC_PACK_SIZE)")
+	f.UintVar(&globalOptions.PackSize, "pack-size", uint(targetPackSize), "set target pack size in MiB, created pack files may be larger (default: $RESTIC_PACK_SIZE)")
 	f.StringSliceVarP(&globalOptions.Options, "option", "o", []string{}, "set extended option (`key=value`, can be specified multiple times)")
 	// Use our "generate" command instead of the cobra provided "completion" command
 	cmdRoot.CompletionOptions.DisableDefaultCmd = true
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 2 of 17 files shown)


cmd/restic/global.go
```diff
@@ -105,6 +107,13 @@ func init() {
 		return nil
 	})
 
+	//set MinPackSize to 4 if not set in env
+	minPackSize, err := strconv.Atoi(os.Getenv("RESTIC_MIN_PACKSIZE"))
```
greatroar, line 111: If you use strconv.ParseUint, you don't need the unsignInt hack below. Also, the comment is incorrect.

metalsp0rk: Oh, good to know.

Also, good catch on the comment, thanks.

cmd/restic/global.go
```diff
@@ -105,6 +107,13 @@ func init() {
 		return nil
 	})
 
+	//set MinPackSize to 16 if not set in env
+	minPackSize, err := strconv.ParseUint(os.Getenv("RESTIC_MIN_PACKSIZE"), 10, 32)
```
MichaelEischer, line 111: My intention was to initially only increase the default pack size to 16 MB, but not provide an option for now. Sorry, if my comment in #2750 wasn't clear enough.

cmd/restic/cmd_prune.go
```diff
@@ -419,6 +421,10 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 		pi := repackCandidates[i].packInfo
 		pj := repackCandidates[j].packInfo
 		switch {
+		case opts.RepackSmall && pi.unusedSize+pi.usedSize < uint64(repo.MinPackSize()) && pj.unusedSize+pj.usedSize >= uint64(repo.MinPackSize()):
```
MichaelEischer, line 424: These two cases should be moved last in the switch-statement. Cleaning up other things is probably more important than getting rid of a few too small files.

cmd/restic/cmd_prune.go
```diff
@@ -362,7 +364,7 @@ func prune(opts PruneOptions, gopts GlobalOptions, repo restic.Repository, usedB
 			// if this is a data pack and --repack-cacheable-only is set => keep pack!
 			keep(p)
 
-		case p.unusedBlobs == 0 && p.duplicateBlobs == 0 && p.tpe != restic.InvalidBlob:
+		case p.unusedBlobs == 0 && p.duplicateBlobs == 0 && p.tpe != restic.InvalidBlob && (!opts.RepackSmall || packSize >= int64(repo.MinPackSize())):
```
MichaelEischer, line 367: I'd like the `--repack-small` option to primarily focus on migrating to a larger pack size. Collecting too many small files should just happen automatically. My idea would be to repack all packs which are less than 10% of the expected pack size. But only if at least one hundredth of the files in a repository fall into that category. That would require collecting a separate list of packs in addition to `repackCandidates` which then has to be checked to contain a sufficient number of packs to be useful. What do you think?

cmd/restic/global.go
GilGalaad: surely is a more suitable default than current one, but without the option the whole change loses half of its usefulness (IMHO, of course).

fingon: 16 MB sounds small too. I was looking forward to setting it to e.g. 128MB or 256MB, as I backup up terabytes and having hundreds of thousands or millions of packfiles is not really great on spinning rust.

GilGalaad: same for me, I aim to 128/256 pack size, but I agree that this could not be a sane default for everybody.
the option to tune this parameter should be present.

GilGalaad:
Hello everyone, and thanks to @metalsp0rk for this great job that I would love to make its way into the main branch.
I apologize if this has been asked before, but it's now almost 2 year that i follow this PR, and a things have been changed in the meanwhile, I could not find anything in the comments of the previous discussion by the way.
I am planning for the second time the effects of changing pack size from the default used in my current repo with a more approriate one (for my workload), and all the tests show no problem, but i have a question about a behavior that is, from my perspective, counterintuitive.

In the following exaples i will use official restic package from FreeBSD
`restic 0.13.1 compiled with go1.18 on freebsd/amd64`
and self compiled version from this branch (actually cross compiled on a Windows machine
`restic 0.13.1 (v0.9.6-1600-ge8f1e3af) compiled with go1.18.3 on freebsd/amd64`
and I am starting from a regularily pruned repository with default values (so 5% max unused space).
Side note: `restic-wrapper.py` is just a simple python wrapper to select the binary I want and to inject the repository coordinates into the environment, nothing more.

This is the output of official restic:
```
[root@DarkSun:/store/maintenance/restic]# ./restic-wrapper.py prune -n
repository 70565c28 opened successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 50 snapshots
[1:17] 100.00%  50 / 50 snapshots...
searching used packs...
collecting packs for deletion and repacking
[21:45] 100.00%  2209246 / 2209246 packs processed...

to repack:             0 blobs / 0 B
this removes:          0 blobs / 0 B
to delete:             0 blobs / 0 B
total prune:           0 blobs / 0 B
remaining:       9165439 blobs / 10.233 TiB
unused size after prune: 345.298 GiB (3.30% of remaining size)
```
So far so good.

Now i'm switching to custom built restic from this PR, I'm setting by default a greater pack size
```
[root@DarkSun:/store/maintenance/restic]# ./restic-wrapper.py prune -n --min-packsize 256
repository 70565c28 opened (repo version 1) successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 50 snapshots
[1:17] 100.00%  50 / 50 snapshots...
searching used packs...
collecting packs for deletion and repacking
[22:09] 100.00%  2209246 / 2209246 packs processed...

to repack:             0 blobs / 0 B
this removes:          0 blobs / 0 B
to delete:             0 blobs / 0 B
total prune:           0 blobs / 0 B
remaining:       9165439 blobs / 10.233 TiB
unused size after prune: 345.298 GiB (3.30% of remaining size)
```
Exact same output, as expected, the unused space is under threshold, and I'm not telling him to repack.

Now I'm telling to repack:
```
[root@DarkSun:/store/maintenance/restic]# ./restic-wrapper.py prune -n --min-packsize 256 --repack-small
repository 70565c28 opened (repo version 1) successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 50 snapshots
[1:18] 100.00%  50 / 50 snapshots...
searching used packs...
collecting packs for deletion and repacking
[22:01] 100.00%  2209246 / 2209246 packs processed...

to repack:          7419 blobs / 5.555 GiB
this removes:          0 blobs / 0 B
to delete:             0 blobs / 0 B
total prune:           0 blobs / 0 B
remaining:       9165439 blobs / 10.233 TiB
unused size after prune: 345.298 GiB (3.30% of remaining size)
```
This is where the unexcpected comes.
I would have supposed to see the repacking of all the files in repository, but this didn't happen.

To trigger the repack of everything i must force max-unused to 0
```
[root@DarkSun:/store/maintenance/restic]# ./restic-wrapper.py prune -n --min-packsize 256 --repack-small --max-unused 0
repository 70565c28 opened (repo version 1) successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 50 snapshots
[1:20] 100.00%  50 / 50 snapshots...
searching used packs...
collecting packs for deletion and repacking
[24:09] 100.00%  2209246 / 2209246 packs processed...

to repack:       9165439 blobs / 10.233 TiB
this removes:     273492 blobs / 345.298 GiB
to delete:             0 blobs / 0 B
total prune:      273492 blobs / 345.298 GiB
remaining:       8891947 blobs / 9.895 TiB
unused size after prune: 0 B (0.00% of remaining size)
```
Am I missing something?
Just by reading the documentation,  I would expect to make the repack happening with the 3rd command, not the 4th.
It's not a big deal anyway, I just want to be sure to understand the internals of this feature, and maybe suggest that the documentation could explain better this scenario, wich I suppose will be very common when this feature will be generally available.

Thank you if you had the patience to read this wall of text :)

cmd/restic/global.go
MichaelEischer: Ok, I give in ^^ . Let's keep the parameter, but limit the allowed range for now to 4-128MB with a default of 16MB. Eventually I'd like to change the default to `auto` assuming that we find a way to automatically determine a reasonable pack size that works for most use cases. I'll add some further rationale below.

MichaelEischer:
tl;dr `--min-packsize` should allow a range of 4-128MB and default to 16MB.

While thinking about the `--min-packsize` parameter I ended up wondering which benefit packs offer in general and why putting everything in a single large pack isn't a good idea either.

Why not store individual blobs?
1. Without packs, accessing lots of blobs would have a higher overhead (in terms of request size and processing costs) as each blob would have to accessed individually.
2. For small blobs that would require a very high number of concurrent requests for fast processing, this is especially a problem when reading/writing has several milliseconds of latency or more. Writing also requires flushing the data to disk which increases the latency drastically.
3. For remote backends nearly all blobs are too small to reach the maximum TCP transfer speed.
4. File systems don't like millions of small objects. We'd also pay a significant price in terms of space necessary to store all the file metadata.

Why not put everything into a single large pack file?
1. No overlap of backup and upload (at least without rebuilding how uploading works)
1. During the upload we have to keep a local copy of the pack to support retries. This could result in requiring scratch space with a size similar to the backup data set.
2. An interrupted upload of a gigantic pack would probably loose everything that was already uploaded, thereby preventing restarts.
3. Garbage collection with gigantic packs would require rewriting large parts of a repository.
4. Using very large pack files would always require us to write temporary pack files during backup. The larger these are the more likely it is that these are actually written to disk.

In a nutshell, the pack size should be as large as possible while only requiring a short time to upload and while keeping the local resource usage limited.

During `backup` we'll need at least packsize*connections of temporary space, which can be significant for large pack files. I also hope to eventually get rid of writing temporary pack files while uploading. However, this requires keeping all temporary packs in memory. The repository index also assumes that packs are below 4GB (for memory efficiency reasons) which provides a hard upper limit. Increasing fundamental properties (i.e. the pack size) by more than an order of magnitude can uncover unexpected problems, which means that we should be careful when increasing the max pack size.

So I think we should for now allow a range of 4-128MB with a default of 16MB for the min pack size.

4MB is the current hardcoded pack size, which seems to limit backup throughput e.g. due to frequent flushing using the local backend. 16MB is a somewhat random size, but should already drastically reduce flushing related performance problems. In addition it is small enough that 5 (the default connection count) of these pack files can be uploaded in a bit over a minute using a network connection with roughly 1MB/s upload (which still isn't that uncommon). The upper limit of 128MB can be uploaded in a second via a Gigabit connection but still doesn't consume gigantic amounts of memory. In addition it is already larger than the current pack size by a factor 32, such that I think that even larger limits are not a good idea for now.

I'm also aware of two potential problems which make larger pack files somewhat problematic: Until #3489 is merged, it is possible to end up with several small packs from each `backup` run. Pack streaming (used by restore and prune) downloads everything between the first and last blob in a pack file. It's probably a good idea to split that range if there's a multi-MB gap between blobs.

MichaelEischer:
I've rebased the PR such that now there's compression and large pack support. Setting the min pack size is now done when creating a new repository object which also ensures that the test suite uses a min pack size larger than 0 bytes.

@GilGalaad `--repack-small` should now work without setting `--max-unused 0`. However, it is still fairly aggressive and will repack every pack file that is smaller than the limit even if there is just a single such pack file.

Slind14:
We tested this version and saw a huge negative impact. Even without changing the size.

- this PR without change -> 600M
- this PR with changes -> 800M
- without this PR -> 2G

MichaelEischer:
@Slind14 Thanks for testing. The numbers are quite unexpected. What was the exact test setup? Is "without this PR", restic 0.13.1 or the current master? Judging from the 2G best-case, I guess you're uploading to the rest-server? Does setting `--min-packsize 4` (or `--pack-size 4` with the latest commits) restore the performance to the previous level? My expectation would be yes, as with a target pack size of 4MiB, this PR should behave exactly as the current master.

This PR now also includes #3489 and #3611 which probably changes some of the performance tradeoffs.

But nevertheless the graph sort of looks like the uploads are stalling at some point. I've no real idea why that would happen. Just as a wild guess, is $TMPDIR (or `/tmp` if not set) a tmpfs mount or stored directly on disk?

Slind14:
I think I was unclear/misleading. 

We are using the async PR and merged this one into it. So we compared the async version with async + packsize.

The temp dir is on an NVME raid. Disk util is not the bottleneck.

Slind14:
With setting it to 4 it looks about the same. Slightly less but that is not significant enough and could be from other factors.

FakeDecent:
Hello, I ran the following command
```
restic -r rclone:onedrive:snapshot prune  --max-repack-size 3M  --repack-small  --pack-size 40  --compression max
```
The result is

```
repository 83ff6918 opened (repo version 2) successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 9 snapshots
[0:04] 100.00%  9 / 9 snapshots
searching used packs...
collecting packs for deletion and repacking
[0:30] 100.00%  5277 / 5277 packs processed

to repack:            64 blobs / 2.640 MiB
this removes:          0 blobs / 0 B
to delete:             0 blobs / 0 B
total prune:           0 blobs / 0 B
remaining:        163729 blobs / 25.734 GiB
unused size after prune: 0 B (0.00% of remaining size)

repacking packs
[0:11] 100.00%  1 / 1 packs repacked
rebuilding index
[0:09] 100.00%  5278 / 5277 packs processed
deleting obsolete index files
[0:00] 100.00%  2 / 2 files deleted
removing 1 old packs
[0:00] 100.00%  1 / 1 files deleted
done

```

If the command is executed 4 times


```
restic -r rclone:onedrive:snapshot prune  --max-repack-size 3M  --repack-small  --pack-size 40  --compression max

rclone ls onedrive:snapshot --max-age 5m
```

The result is

```

C:\Users\20267>rclone ls onedrive:snapshot --max-age 5m
   658081 index/4ba4dc5bde024875d747487012ddce015b9e8313107f746886dbcf0f717bbbdb
  7483572 index/d525310111cab7f1885e8a80f059f569e044006fc5021f7d42f1016eda14be73
  2352452 data/37/370f2dbe30ef8e4afbb0a8be4193080c78968757a9b15ba6f95cbb982e3ad7cc

C:\Users\20267>rclone ls onedrive:snapshot --max-age 5m
  7450290 index/8e99b08c38c3a8cff8fb6b467379d756ab5f2e866466c6c0ee71abc0dd5e189f
   686813 index/b47161a4ba6c2da43002cbb8f575f3eaf88594af51bc2f62373129c8a31b60d7
  2352452 data/f4/f45c018c7177211d5fbb55e2737b8b866f2b8ce2c5419552b3d851bc0fa4f8c8

C:\Users\20267>rclone ls onedrive:snapshot --max-age 5m
   683488 index/51a78a850f428c08741420e3a8b4d2e5c9cf3264f6916fda1f60bdf83019e383
  7462776 index/8d92e0f3f2803399d50dff374906153646430de99a68eae341dc00f93b9fe6b3
  2767803 data/47/4775849a4fc4a60503437350016b87b29a7f47ef9d409fedf65182ed0d732263
  2352452 data/f4/f45c018c7177211d5fbb55e2737b8b866f2b8ce2c5419552b3d851bc0fa4f8c8

C:\Users\20267>rclone ls onedrive:snapshot --max-age 5m
  7448106 index/75e2766c07fcbec9ba3fd33463835cc91c63d004f05d4a16e89ed3168da4dead
   692563 index/9726ca74eea992c0217f324a56c93bccdfcb449a7ca10bf34bfb4bcdc229671e
  2767803 data/54/54fd85a26452b02bf30a201eec922811e5723e167a51ea6455a624d9d9621e41

```
It seems that the same one is sorted every time

FakeDecent:
I run
```
restic -r rclone:onedrive:snapshot prune  --max-repack-size 80M  --pack-size 4  --compression max
```
or
```
restic -r rclone:onedrive:snapshot prune  --max-repack-size 80M  --pack-size 4  --compression max  --repack-small
```
The results are all
```
repository 83ff6918 opened (repo version 2) successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 7 snapshots
[0:03] 100.00%  7 / 7 snapshots
searching used packs...
collecting packs for deletion and repacking
[0:41] 100.00%  5276 / 5276 packs processed

to repack:             0 blobs / 0 B
this removes:          0 blobs / 0 B
to delete:             0 blobs / 0 B
total prune:           0 blobs / 0 B
remaining:        163525 blobs / 25.716 GiB
unused size after prune: 42.074 MiB (0.16% of remaining size)

done
```
The result I expect is

```
unused size after prune: 0 MiB (0% of remaining size)
```

MichaelEischer:
> With setting it to 4 it looks about the same. Slightly less but that is not significant enough and could be from other factors.

@Slind14 Could you test whether the current code still has the upload performance problem with 16MB chunks? For comparison chunk sizes of 4 and 8 MB would also be highly interesting. Did you run the upload performance tests using the rest-server or the S3 backend? For the latter I've noticed that file chunks above 16MB lead to multipart uploads which could be detrimental to performance. These multipart uploads are now disabled below 200MB.

cmd/restic/global.go
```diff
@@ -121,6 +126,7 @@ func init() {
 	f.Var(&globalOptions.Compression, "compression", "compression mode (only available for repository format version 2), one of (auto|off|max)")
 	f.IntVar(&globalOptions.Limits.UploadKb, "limit-upload", 0, "limits uploads to a maximum rate in KiB/s. (default: unlimited)")
 	f.IntVar(&globalOptions.Limits.DownloadKb, "limit-download", 0, "limits downloads to a maximum rate in KiB/s. (default: unlimited)")
+	f.UintVar(&globalOptions.MinPackSize, "pack-size", uint(targetPackSize), "set target pack size in MiB. (default: $RESTIC_PACK_SIZE)")
```
rawtaz, line 129: Rename `globalOptions.MinPackSize` to `globalOptions.PackSize` so it's in line with the parameter name?

doc/047_tuning_backup_parameters.rst
```diff
@@ -47,3 +48,26 @@ is compressed with the option ``--compression``. It can be set to ```auto``` (th
...
+
+
+Pack Size
+=========
+
+In certain instances, such as very large repositories (in the TiB range) or very fast
+upload connections, it is desirable to use larger pack sizes to reduce the number of
+files in the repository and improve upload performance.  Notable examples are OpenStack
+Swift and some Google Drive Team accounts, where there are hard limits on the total
+number of files.  Larger pack size can also improve the backup speed for a repository
```
rawtaz, line 60: "Larger pack size" => "Larger pack sizes"?

cmd/restic/global.go
MichaelEischer: is changed.

doc/manual_rest.rst
```diff
@@ -49,13 +49,18 @@ Usage help is available:
           --cacert file                file to load root certificates from (default: use system certificates)
           --cache-dir directory        set the cache directory. (default: use system default cache directory)
           --cleanup-cache              auto remove old cache directories
+<<<<<<< HEAD
           --compression mode           compression mode (only available for repository format version 2), one of (auto|off|max) (default auto)
+=======
+          --compression mode           compression mode (only available for repo format version 2), one of (auto|off|max) (default auto)
+>>>>>>> 6c2d033e2 (update restic help snippets in documentation)
```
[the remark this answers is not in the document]
MichaelEischer: Oops, saving a file after fixing the conflicts and before committing it helps :-)

doc/047_tuning_backup_parameters.rst
```diff
@@ -47,3 +48,26 @@ is compressed with the option ``--compression``. It can be set to ```auto``` (th
...
+stored on a local HDD.  This can be achieved by either using the ``--pack-size`` flag
```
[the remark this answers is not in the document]
MichaelEischer: `--pack-size` is a global option, which applies to every command that writes pack files.

cmd/restic/cmd_prune.go
```diff
@@ -69,6 +70,7 @@ func addPruneOptions(c *cobra.Command) {
 	f.StringVar(&pruneOptions.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "maximum `size` to repack (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&pruneOptions.RepackCachableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
+	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack too small packs")
```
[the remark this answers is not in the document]
MichaelEischer: I've changed this to read `repack pack files below 80% of target pack size`

doc/047_tuning_backup_parameters.rst
MichaelEischer: Is fixed.

doc/047_tuning_backup_parameters.rst
```diff
@@ -47,3 +48,26 @@ is compressed with the option ``--compression``. It can be set to ```auto``` (th
...
+or defining the ``$RESTIC_PACK_SIZE`` environment variable.  Restic currently defaults
+to a 16MiB pack size.
+
+The side effect of increasing the pack size is requiring more disk space for temporary pack
+files created before uploading.  The space must be available in the system default temp
+directory, unless overwritten by setting the ``$TMPDIR`` environment variable.  In addition,
+depending on the backend the memory usage can also increase by a similar amount. Restic
+requires temporary space according to the pack size, multiplied by the number
+of backend connections plus one. For example, if the backend uses 5 connections (the default
+for most backends), with a target pack size of 64MiB, you'll need a *minimum* of 384MiB
```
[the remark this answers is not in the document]
MichaelEischer: Is changed. That way it's also more consistent with other parts of the documentation.

MichaelEischer:
Rebased to resolve conflicts. `StreamPacks` now loads a pack file in multiple parts if it would skip more than 4 MB.

JsBergbau:
Are there plans to include this in the next release? Would be very great because when converting to repo v2 with compression you could also increase packsize. This probably would also make conversation/compression of uncompressed packs a bit faster with bigger packsizes.

MichaelEischer:
I was hoping for some feedback on the latest code changes (just fixed a few corner cases in the latest commit), but I agree that it would be good to include this PR in the next release. Especially as it increases the default pack size, which together with the repository compression would result in a smooth upgrade for most repositories.

Ideally we'd find out what exactly caused the performance problems for @Slind14 before merging this.

JsBergbau:
I don't fully understand which performance problems occur @Slind14.

I'm willing to test, just tell me what and how to test? 

Whats the new default pack size btw?

MichaelEischer:
> Whats the new default pack size btw?

The default pack size is now 16MB.

> I'm willing to test, just tell me what and how to test?

That performance problem is probably rather specific, as it requires a rest-server/S3 with rather fast disks and 10Gbit ethernet (>= 3Gbit would probably also be enough). Besides that, the test scenario looks like it was a backup of some large dataset.

JsBergbau:
Can the test via rest-server made locally, meaning rest-server and restic running on the same host?
I have NVMe SSD as target (WD Blue 1TB) and S-ATA SSD with 500 MB read speed. If that makes sense backing up from S-ATA to NVMe with rest-server I'll setup a test and measure the time it takes to complete lets say backing up of 200 GB of data.

metalsp0rk:
I had great results on my laptop testing the latest revisions, but my dataset was only 400GB.  I managed to average about 650Mb of writes with many small (<500KB) files.  One thing to note, I believe I was CPU limited in the upload, as I observed a load average of approximately `12 11 10` on an 8 core CPU just before the end of the backup.

I took no notes on memory or temp usage.

Slind14:
Thank you for waiting. We can test in the next few days.

burgyman:
Hi,
I'm not a developer but would be willing to help to test this PR as well, just let me know how.
I use restic with rsync.net and Hetzner storage box, sometimes I used the latest beta version, is it possible to compile it with the min-pack file?

aawsome:
Actually I already implemented a feature to change the packsize in rustic, see https://github.com/rustic-rs/rustic/pull/52 which is already included in [rustic 0.2.3](https://github.com/rustic-rs/rustic/releases/tag/v0.2.3).

With restic, the discussion seems to go in the direction of a static pack files size (at least #3367 did not get much feedback or discussion so far), so users could compare restic and rustic if they want do see if one of the methods fits better for them.

One thing, however, I'd like to question:
When adding the options to rustic, I also first had just one pack file size. Then I thought a bit about pros and cons of large packfiles. 

On the cons side is that repacking larger packs is more expensive and when doing lots of pruning/repacking, the negative effect of larger pack sizes may even outweight the gained performance by the backup command.

Now, `prune` always repacks tree packs which are only partly used which implies that tree packs are usually much more repacked than data packs. Therefore, I decided to have different defaults for tree and data pack files. In rustic by default tree pack files start from 4MiB and data pack files start from 32MiB.

So maybe you should also take some difference between tree and data packs into consideration.

MichaelEischer:
> Can the test via rest-server made locally, meaning rest-server and restic running on the same host?
> I have NVMe SSD as target (WD Blue 1TB) and S-ATA SSD with 500 MB read speed.

Thanks for the offer, but I've already tested with a local SSD to SSD setup and there the performance was fine. It might be interesting to see a different SSD combination, but I suspect the result will be the same.

> With restic, the discussion seems to go in the direction of a static pack files size (at least #3367 did not get much feedback or discussion so far), so users could compare restic and rustic if they want do see if one of the methods fits better for them.

Well, we're still busy catching up with the massive pile of PRs that has accumulated over the last years. Regarding #3367 my idea was for the default pack size to eventually behave like "auto" (and not just a hardcoded value of 16MB). But that is something that can be changed _after_ laying the groundwork.

> Therefore, I decided to have different defaults for tree and data pack files. In rustic by default tree pack files start from 4MiB and data pack files start from 32MiB.

Do you have an idea how much effect this has in practice? At least for incremental backups, I'd expect the tree packs to be rather small anyways, so I wonder whether this makes much of a difference as most tree packs are already rather small?

> Actually I already implemented a feature to change the packsize in rustic, see [rustic-rs/rustic#52](https://github.com/rustic-rs/rustic/pull/52) which is already included in [rustic 0.2.3](https://github.com/rustic-rs/rustic/releases/tag/v0.2.3).

@aawsome I'm torn how to feel about you constantly advertising rustic. On the one hand, your input on discussions is highly appreciated, in particular different design decisions are quite interesting. But on the other hand, being continuously told that we should merge and release (!) stuff faster and whatever else you might be implying is annoying (at least it comes across that way to me).

aawsome:
> At least for incremental backups, I'd expect the tree packs to be rather small anyways, so I wonder whether this makes much of a difference as most tree packs are already rather small?

I agree that incremental backups usually generate anyway small pack files - but in that case any increased packsize wouldn't apply. I was talking about (mosti likely initial) backups with lots of dirs or - and this is the main problem - `prune` itself which would crate large tree pack files depending on the increased packsize. 

The argument is that simple: The larger a tree packfile is, the more likely it is that it is rebuilt by prune and at the same time the more work `prune' needs to repack those.

I have not measured the effect. But in my repos trees occupy a very small size compared to the total repo size. 200GiB repo size and 200MiB total tree size is in my experiance a typical ratio (given that you don't have degenerated trees by some xattr data). And then it doesn't matter much if you have 50 or 13 tree pack files (compared to the 50.000 or 12.500 or even lower number of data pack files - here the difference does matter). But the increased repacking might matter.

mew1033:
@metalsp0rk Just thought I'd pop in here and tell you how much I appreciate your work on this PR. I've been following it for a very long time and the amount of work you've put in is impressive to say the least.

I agree that this should be merged. The benefits are obvious and well documented at this point and the potential problems seem (at least from what I've read in the various issues and PRs here) unlikely.

dionorgua:
It's good idea to merge pack file change and release it together with compression just because both changes requires huge prune. And it's good idea to do this at the same time.

MichaelEischer:
> I frankly don't have the time to manage this PR anymore, as evidenced by the lack of activity from myself on the branch for the last few months. I'd like to get it merged as is, without any design modifications. As it stands, there's a question of performance that needs to be addressed, which may warrant additional commits.

I'm totally aware that the PR has escalated quite a bit (and I feel partially guilty for that), although I'm not entirely surprised as parameter tweaking is very amenable to bikeshedding and large amounts of back and forth. We can get into parameter tuning later on if someone is interested in testing whether different parameter offer a better general tradeoff. But as long as the current state of this PR doesn't break performance (too) badly, I'm inclined to merge it as is.

> It's good idea to merge pack file change and release it together with compression just because both changes requires huge prune. And it's good idea to do this at the same time.

That the plan and the reason is why I've been pushing this PR forwards lately.

metalsp0rk:
>But as long as the current state of this PR doesn't break performance (too) badly, I'm inclined to merge it as is.

I don't understand how it would be possible for this PR to break performance (especially if we specify `--pack-size 4`).  It really isn't changing the logic of the backup pipeline much at all, it's just exposing a formerly-hardcoded variable to give user control.  

I recognize Slind had a regression, but I have been unable to reproduce the issues on my local network (symmetric gig), an AWS box I rented for a couple hours to test, and a box on Linode.  

The Async PR was also involved here, which as far as I can tell, has been merged into master.  Which indicates to me that an old revision of Restic may be in use in this circumstance.

MichaelEischer:
Rebased to fix merge conflicts.

vejnar:
Thanks Micheal for working on getting this merged. As reported in this page by multiple people, different setup lead to different performance. I don't think this is avoidable and kind of the user's responsibility if changing these parameters. That said, I have been using Restic with a MaxPackSize of 1024MB for a while, as I have very large repos (1TB to >30TB). Is there a reason to limit MaxPackSize to 128MB?

MichaelEischer:
> As reported in this page by multiple people, different setup lead to different performance. I don't think this is avoidable and kind of the user's responsibility if changing these parameters.

The part that's bugging me is that I don't see a reason why 4MB packfiles could be faster during a backup than 16MB ones. My hope is/was that the increase will lead to improvements across the board or at least maintain the performance. That's why I'd prefer to see a few more test results first to get a good idea whether performance regressions are likely to occur for some users or not. Depending on these results it might be useful to add a corresponding note to the documentation. But yes, the fall back is to ask those affected by regressions to just specify `--pack-size 4`.

> That said, I have been using Restic with a MaxPackSize of 1024MB for a while, as I have very large repos (1TB to >30TB). Is there a reason to limit MaxPackSize to 128MB?

The reason to limit it to 128MB for now is that it is already larger than the current target pack size by a factor of 32. Usually increasing some parameter by more than an order of magnitude can lead to surprises or scaling problems. In particular, I'd like to get rid of the temporary files currently created by `backup` and just keep them in memory. But for 1GB pack files this likely would lead to memory usage problems. The difference between 128MB and 1GB shouldn't have much influence on performance, but probably just decrease the number of files in a repository.

Slind14:
> Regarding the performance question: I have no information on the methodology of this test, but I've got a lot of questions. @Slind14 would you be able to describe the system configuration for the test in which you observed poor performance? I'm curious if I can reproduce it.

Server to Backup:
 - disks that can sustain +10G, in our case 10x NVMe Raid 6
 - 10 G network

Server to Backup to:
 - Wasabi location with 10G (concurrent upload across multiple entry points)
 - or custom S3 with 10G network -> in this case +14 HDDs Raid 10 -> to be able to handle the write
 - ^ we test both of these variants: same datacenter park vs cross-continent

Slind14:
Only tested with S3 in the same datacenter park:

|restic               |s3.connection|FILE_READ_CONCURRENCY|Pack-size|result|
|---------------------|-------------|---------------------|---------|------|
|async_read_concurrent|64           |3                    |-        |2.9G  |
|async_read_concurrent|64           |4                    |-        |3.6G  |
|async_read_concurrent|256          |6                    |-        |4.4G  |
|async_read_concurrent|8            |2                    |-        |1.6G  |
|async_read_concurrent|8            |4                    |-        |1.6G  |
|packsize_latest      |8            |3                    |default  |2.1G  |
|packsize_latest      |8            |3                    |4        |1.6G  |
|packsize_latest      |8            |3                    |8        |2.0G  |
|packsize_latest      |8            |3                    |16       |2.3G  |
|packsize_latest      |8            |3                    |64       |2.7G  |
|packsize_latest      |8            |3                    |128      |3.0G  |
|packsize_latest      |8            |4                    |128      |2.7G  |
|packsize_latest      |64           |3                    |64       |3.1G  |
|packsize_latest      |256          |3                    |64       |3.1G  |
|packsize_latest      |256          |4                    |64       |3.1G  |


Is it possible that this branch does not support `FILE_READ_CONCURRENCY` because we bottleneck at 3.1G no matter what?

MichaelEischer:
> Is it possible that this branch does not support `FILE_READ_CONCURRENCY` because we bottleneck at 3.1G no matter what?

Yep, the file read concurrency parameter is only include in #2750. This PR was intended to discuss and implement just the pack-size configuration. I'll trim down #2750 after merging this one.

@Slind14 Thanks a lot for testing. The results look what I was hoping for, so that we can go ahead and merge the pack-size flag :partying_face:

MichaelEischer (approved):
LGTM. Thanks a lot for the discussions, testing and tons of patience to get this merged!

GilGalaad:
this is a great news, we have been waiting for so long <3

JsBergbau:
Thanks for merging this into master.
I've build restic from current master branch and tried this feature with a local repository and packsize of 64 MB on Windows.
This leads to writing out the temporary files on SSD in temp directory like reported here for checking the data https://github.com/restic/restic/issues/3375
This also applies with new default 16 MB packsize.
Environment Variables

set RESTIC_PACK_SIZE=64
set RESTIC_COMPRESSION=max

This should at least be documented, because you should know that you wear your SSD when backing up with large packfiles.

I've checked with 4 MB packsize. There also some data is written to temp directory, but much less than with bigger packsize.

On linux with a small machine (Raspberry PI) same behaviour, except that for 4 MB packsize much less is written to temp directory compared to windows. 

So I would add in the docs

```
The side effect of increasing the pack size is requiring more disk space for temporary pack
files created before uploading.  The space must be available in the system default temp
directory, unless overwritten by setting the ``$TMPDIR`` environment variable.
Another side effect is SSD wear. The bigger the pack sizes the bigger the likelihood packs
get written to disk. With 4 MB packs this happens quite seldom whereas with default
16 MB packsize most packs will be written to your SSD first.
```
Or something like that.

metalsp0rk:
>I've checked with 4 MB packsize. There also some data is written to temp directory, but much less than with bigger packsize.

@JsBergbau The writes with 4MB packsize are the same as the writes with 16MB or 128MB packsize.  The only difference is the number of files written.  Because the temp directory stores the packs only while the pack is in flight, it may appear that the total writes are lower, but that is not the case.  (in fact, with a 4MB packsize, there might be a minuscule increase in writes due to more file metadata being written, due to the larger file count)

I thought I had that in the documentation somewhere, at least in the original #2750 PR there was mention of it.  

@MichaelEischer Thank you very much for merging this!

JsBergbau:
@metalsp0rk Sorry thats no criticism on your pullrequest. I'm really glad that it finally made it into master.

When files are smaller they don't seem to get written to disk. Here a screenshot from Windows taskmanager using 16 MB and 4 MB packsize. Note taskmanager caps data transfer rate, so in fact it is much higher than 250 MB/s, so in fact there is much more data written to SSD with 16 MB packs than 4 MB packs.

Backup data is read and written to D:

grafik

MichaelEischer:
> The writes with 4MB packsize are the same as the writes with 16MB or 128MB packsize. The only difference is the number of files written. Because the temp directory stores the packs only while the pack is in flight, it may appear that the total writes are lower, but that is not the case. (in fact, with a 4MB packsize, there might be a minuscule increase in writes due to more file metadata being written, due to the larger file count)

That direct relationship only holds when ignoring the page cache. Disk I/O is first buffered in memory and either written after some delay or when flushing. Restic doesn't call `flush` for it's temp files, so we're in the case where the data is written after some delay. If a temporary file is deleted before being flushed to disk, the corresponding data can also be discarded. As larger pack files are around for much longer, this also drastically increases the chances of being written to disk.

For Windows we explicitly ask the OS not to write the data to disk (#3610) and also ensure that the file is deleted no matter what happens. But apparently Windows has decided to nevertheless write the temp files to disk. So it looks like we should try to keep the whole pack file in memory if possible.

JsBergbau:
It seems something has changed in the behaviour of windows. Maybe this is even a bug in Windows? I remember when I began using restic I was checking if restic writes out the data to disk and this didn't happen. I knew that behaviour from Duplicati https://forum.duplicati.com/t/attention-ssd-users-if-your-temp-directory-isnt-ram-backed-duplicati-will-prematurely-wear-your-ssd/5779 so I explicitly checked if restic is having the same behaviour and needs a RAM disk like Duplicati needed. But that wasn't the case.

I also downloaded restic 0.9.6 to verify, but also with 0.9.6 data gets written to SSD. So I'm quite sure that there has been some change in the behaviour of Windows.

JsBergbau:
Some side note: When you've forgot to set compression=max and you want to compress later with maximum you can use this PR to increase min packsize. That also leads to re-compression at maximum level. Is this behaviour intented or by accident? If it is by intent I would create a pullrequest and update the documentation accordingly, because this feature is really useful. On the downside just changing packsize requires some time because compression is also re-applied. But probably you can't just byte-copy the packs but you have to uncompress and recompress the content.

JsBergbau:
Some data while converting v1 repo to v2 with pruning with maximum compression and 64 MB packsize on linux via `./restic prune --repack-uncompressed --repack-small --no-cache`. Rest-Server is on the same machine so no cache is needed.

grafik

First graph is where I forgot `export TMPDIR=/dev/shm` 

According to datarate about 8 % of tempfiles were written out to SSD. System has 32 GB of RAM where only 3,3 GB are used with restic using more than 2 GB of this memory. It might be possible the more RAM is available the less tempfiles get written to disk.

junqfisica:
> Only tested with S3 in the same datacenter park with a build on the latest commit of this branch from the 1st of August:

@Slind14 First thank you for showing a test to S3. Currently, we are trying to tune restic performance to backup 85TB of data into S3 (Scality ring). Our files are all images with an average size of 60MB. 

Would you mind explaining what s3.connection is? Is this an `--option` parameter from restic?  

I'm also wondering if we can use a bigger `pack-size` (64 or 128) to improve performance. Also, is there a problem with changing this parameter after an initial backup?

MichaelEischer:
@junqfisica Yes you would specify the connections parameters as `restic -o s3.connections=10 [...]`, for other options have a look at the output of `restic options`. The documentation now also includes some advice on performance tuning, have look at https://restic.readthedocs.io/en/latest/047_tuning_backup_parameters.html . Note that this describes the behavior of the current beta version of restic (which will be 0.14.0 once it is released). restic 0.13.1 does not support most of these parameters nor does the performance scale well.

junqfisica:
@MichaelEischer Thank you so much for your reply.  Do you know when a stable version (0.14.0) will be released on `epel-release`?

MichaelEischer:
We first have to release 0.14.0, which shouldn't be too far out anymore. I can't tell how long it will take to be available in other repositories.

legg33:
I just wanted to confirm, that windows seems to be very eager to write these temp files to disk. On my 64GB RAM machine with only 10GB used, while backing up I see that almost all temp files are written to disk, even though they only live for less than a second (backup to external disk with ~100MB/s and I see about 100MB/s written to my internal NVME also). The pack size seems to not make a noticeable difference here.

For now I'm creating a RAM-disk pre-backup, but of course it would be nice if we could tell restic to keep the packs in memory itself.

Thanks to all the devs for all the great work so far!

PR #3733, opened 2022-05-01
Improve stats

### LINKED ISSUE #3779: Total Size in stats is not correct when backup is compressed
Output of `restic version`
--------------------------
```
./restic2 version
restic 0.13.1-dev (compiled manually) compiled with go1.18 on linux/amd64
```

I have built current master with compression support.

`./restic2 --verbose -r /mnt/restic/repo2/ stats --mode raw-data`

```
repository 193651e2 opened (repo version 2) successfully, password is correct
scanning...
Stats in raw-data mode:
Snapshots processed:   1
   Total Blob Count:   97646
         Total Size:   174.626 GiB
```

local dir

Expected behavior
-----------------

After backup with compression, i can see actual size which is smaller then in 'stats' output:

```
du -sh /mnt/restic/repo2
99G	/mnt/restic/repo2
```

Backup something with compression and check 'stats' in 'raw-data' mode.

fd0:

Print the number of bytes added to the repo (including optional compression and the crypto overhead).

Checklist
---------

- [x] Add real size to the `stats` command

### DIFF
--- a/changelog/unreleased/issue-21
+++ b/changelog/unreleased/issue-21
@@ -23,4 +23,7 @@ repository using `init --repository-version 2 --copy-chunker-params --repo2 path
 Then use the `copy` command to copy all snapshots to the new repository.
 
 https://github.com/restic/restic/issues/21
+https://github.com/restic/restic/issues/3779
 https://github.com/restic/restic/pull/3666
+https://github.com/restic/restic/pull/3704
+https://github.com/restic/restic/pull/3733
--- a/cmd/restic/cmd_debug.go
+++ b/cmd/restic/cmd_debug.go
@@ -386,7 +386,7 @@ func loadBlobs(ctx context.Context, repo restic.Repository, pack restic.ID, list
 			}
 		}
 		if reuploadBlobs {
-			_, _, err := repo.SaveBlob(ctx, blob.Type, plaintext, id, true)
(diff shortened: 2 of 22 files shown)


MichaelEischer:
The code looks fine. But currently "SizeInRepo" feels a bit misleading. It will systematically underreport how much data was added to the repository, as the pack header overhead is missing. My suggestion would be to either immediately report the pack entry header size along with the blob length or to report that overhead together with the blob that triggers a pack file upload. The downside of the second variant is that it would either require modifications to `repo.Flush()` or not account for the last few pack headers.

fd0:
Ah, right. I'm in favor of the second option, even if that's not exact and will not report the last few header sizes. Since the header size will depend on the repo version I think it's less complex this way.

MichaelEischer:
I've implemented a mix of both variants. We already have a method to calculate the size of the pack header entry for a given blob, which make accounting for the per blob overhead a one-line change. When finalizing a pack file, now the pack header crypto overhead is reported and added for the last blob. 

Adding the pack header entry overhead for each blob individually also has the benefit that the "stored size" for each file won't have large random spikes due to charging the full pack overhead to a single blob.

MichaelEischer (approved):
LGTM. Let's merge this now and wait for feedback whether the stats need further improvements.

PR #3734, opened 2022-05-01
Validate exclude patterns

### LINKED ISSUE #3709: Syntax error in exclude file causes all excludes to be ignored and back up uploads all files
Output of `restic version`
--------------------------

    restic 0.13.1 compiled with go1.18 on darwin/arm64

    export RESTIC_PASSWORD_COMMAND='(command that returns repo password)';
    export GOOGLE_PROJECT_ID='(project id)';
    export GOOGLE_APPLICATION_CREDENTIALS="(path to service account credentails)";
    export RESTIC_REPOSITORY='gs:bucket:/';
    restic -v -v backup /Users/me --exclude-file=exclude.txt --dry-run

GCS Coldline

Expected behavior
-----------------

When an exclude file contains syntax errors, restic should display an error _and exit_.

Actual behavior
---------------

When an exclude file contains syntax errors, restic continues to back up all files without exception. One syntax error in the exclude file causes all exclude rules to be ignored. The result is that many GB of data I didn't want in my repo is uploaded. 

1. Create a file `exclude.txt` containing any number of exclude rules, with this invalid (? not sure why, looks ok to me) exclude rule:
```
*[._]log[.-][0-9]
```
2. Do a dry-run back up like this:
```
restic -v -v backup /Users/me --exclude-file=exclude.txt --dry-run
```
The output will contain error messages: *“error for exclude pattern: Match: syntax error in pattern”*. Here's is my output:

    ❯ restic -v -v backup /Users/me --exclude-file=exclude.txt --dry-run
    open repository
    repository 8f5e121b opened successfully, password is correct
    found 7 old cache directories in /Users/me/Library/Caches/restic, run `restic cache --cleanup` to remove them
    lock repository
    load index files
    using parent snapshot 3cd63dc5
    start scan on [/Users/me]
    start backup on [/Users/me]
    error for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for 
    error for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Mat
    unchanged /Users/me/.CFUserTextEncoding
    new       /Users/me/.DS_Store, saved in 0.002s (54.004 KiB added)
    unchanged /Users/me/.bash_profile
    unchanged /Users/me/.bash_profile-local
    unchanged /Users/me/.bashrc
    unchanged /Users/me/.bashrc-local
    error for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pa
    new       /Users/me/.config/.DS_Store, saved in 0.000s (8.004 KiB added)
    unchanged /Users/me/.config/bat/config
    error for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for exclude pattern: Match: syntax error in patternerror for 
    unchanged /Users/me/.config/bat/syntaxes/PHP.sublime-syntax
    unchanged /Users/me/.config/bat/syntaxes/
    unchanged /Users/me/.config/bat/
    …etc

I'm pretty sure restic caused this. 🙃

Instead of just printing an error message, restic should print the error message **and then exit** (ideally with a useful exit status code).

Restic is freaking awesome. Sometimes it just works so well I live fearlessly and forget to use --dry-run. 😁

lbausch:

Exclude patterns provided via `--exclude`, `--iexclude`, `--exclude-file` or `--iexclude-file` are validated. Invalid patterns cause `restic backup` to abort with a fatal error displaying the invalid patterns.
To achieve this, the `Pattern` struct is extended to include the original pattern. After parsing each pattern with `ParsePatterns`, every pattern part is validated using `filepath.Match`, by matching the pattern against itself.

Before Go 1.16 `filepath.Match` does not correctly report syntax errors, see https://go.dev/doc/go1.16#path/filepath
The corresponding tests are therefore only executed for Go 1.16+.

`restic restore` also accepts patterns - a follow-up MR could validate them as well.

Closes #3709

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -20,6 +20,7 @@ import (
 	"github.com/restic/restic/internal/archiver"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
@@ -298,6 +299,11 @@ func collectRejectByNameFuncs(opts BackupOptions, repo *repository.Repository, t
 		if err != nil {
 			return nil, err
 		}
+
+		if valid, invalidPatterns := filter.ValidatePatterns(excludes); !valid {
+			return nil, errors.Fatalf("--exclude-file: invalid pattern(s) provided:\n%s", strings.Join(invalidPatterns, "\n"))
+		}
+
 		opts.Excludes = append(opts.Excludes, excludes...)
 	}
 
@@ -306,14 +312,27 @@ func collectRejectByNameFuncs(opts BackupOptions, repo *repository.Repository, t
 		if err != nil {
 			return nil, err
 		}
+
+		if valid, invalidPatterns := filter.ValidatePatterns(excludes); !valid {
+			return nil, errors.Fatalf("--iexclude-file: invalid pattern(s) provided:\n%s", strings.Join(invalidPatterns, "\n"))
+		}
+
 		opts.InsensitiveExcludes = append(opts.InsensitiveExcludes, excludes...)
 	}
 
 	if len(opts.InsensitiveExcludes) > 0 {
+		if valid, invalidPatterns := filter.ValidatePatterns(opts.InsensitiveExcludes); !valid {
+			return nil, errors.Fatalf("--iexclude: invalid pattern(s) provided:\n%s", strings.Join(invalidPatterns, "\n"))
+		}
+
 		fs = append(fs, rejectByInsensitivePattern(opts.InsensitiveExcludes))
 	}
 
 	if len(opts.Excludes) > 0 {
+		if valid, invalidPatterns := filter.ValidatePatterns(opts.Excludes); !valid {
+			return nil, errors.Fatalf("--exclude: invalid pattern(s) provided:\n%s", strings.Join(invalidPatterns, "\n"))
+		}
+
 		fs = append(fs, rejectByPattern(opts.Excludes))
 	}
 
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -220,6 +220,27 @@ func match(pattern Pattern, strs []string) (matched bool, err error) {
 	return false, nil
 }
 
+// ValidatePatterns validates a slice of patterns.
+// Returns true if all patterns are valid - false otherwise, along with the invalid patterns.
+func ValidatePatterns(patterns []string) (allValid bool, invalidPatterns []string) {
+	invalidPatterns = make([]string, 0)
+
+	for _, Pattern := range ParsePatterns(patterns) {
+		// Validate all pattern parts
+		for _, part := range Pattern.parts {
+			// Validate the pattern part by trying to match it against itself
+			if _, validErr := filepath.Match(part.pattern, part.pattern); validErr != nil {
+				invalidPatterns = append(invalidPatterns, Pattern.original)
+
+				// If a single part is invalid, stop processing this pattern
+				continue
+			}
+		}
+	}
+
+	return len(invalidPatterns) == 0, invalidPatterns
+}
+
 // ParsePatterns prepares a list of patterns for use with List.
 func ParsePatterns(pattern []string) []Pattern {
 	patpat := make([]Pattern, 0)
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -18,6 +18,7 @@ type patternPart struct {
 
 // Pattern represents a preparsed filter pattern
 type Pattern struct {
+	original  string
 	parts     []patternPart
 	isNegated bool
 }
@@ -31,6 +32,9 @@ func prepareStr(str string) ([]string, error) {
 
 func preparePattern(patternStr string) Pattern {
 	var negate bool
+
+	originalPattern := patternStr
+
 	if patternStr[0] == '!' {
 		negate = true
 		patternStr = patternStr[1:]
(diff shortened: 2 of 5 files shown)


greatroar:
> Some of the patterns used in tests, e.g. *.go are also invalid according to regexp.MustCompile().

That's because they're not regular expressions. They're extended glob patterns. No validation of those is possible other than checking for the empty string or an embedded zero byte, because practically all patterns are valid. They just won't match anything.

lbausch:
Ah, I see, so `regexp.MustCompile()` is not the right approach in this situation. Would it be sufficient to execute `filepath.Match()` with the pattern at hand against an empty string, e.g. `_, err := filepath.Match(patternStr, "")` and see if this raises any error (`err != nil`)?

fd0:
> Would it be sufficient to execute `filepath.Match()` with the pattern at hand against an empty string, e.g. `_, err := filepath.Match(patternStr, "")` and see if this raises any error (`err != nil`)?

Yes, that should do it.

lbausch:
I've added some tests and they fail with Go 1.15. Apparently [1.16](https://go.dev/doc/go1.16) changed the way `filepath.Match` reports syntax errors:
>The [Match](https://go.dev/pkg/path/filepath#Match) and [Glob](https://go.dev/pkg/path/filepath#Glob) functions now return an error if the unmatched part of the pattern has a syntax error. Previously, the functions returned early on a failed match, and thus did not report any later syntax error in the pattern.

internal/filter/filter_test.go
```diff
@@ -226,6 +228,49 @@ func TestChildMatch(t *testing.T) {
...
 
+func TestValidPatterns(t *testing.T) {
+	// Test invalid patterns are correctly detected and returned
+	t.Run("detect-invalid-patterns", func(t *testing.T) {
+		allValid, invalidPatterns := filter.ValidatePatterns([]string{"*.foo", "*[._]log[.-][0-9]"})
+
+		rtest.Assert(t, allValid == false,
+			"Expected invalid patterns to be detected")
+
+		rtest.Assert(t, reflect.DeepEqual(invalidPatterns, []string{"*[._]log[.-][0-9]"}),
```
MichaelEischer, line 239: `rtest.Equals` would make debugging the test easier

internal/filter/filter.go
```diff
@@ -216,6 +216,27 @@ func match(pattern Pattern, strs []string) (matched bool, err error) {
 	return false, nil
 }
 
+// ValidatePatterns validates a slice of patterns.
+// It returns true if all patterns are valid - false otherwise, along with the invalid patterns.
+func ValidatePatterns(patterns []string) (allValid bool, invalidPatterns []string) {
+	invalidPatterns = make([]string, 0)
+
+	for _, pattern := range patterns {
```
MichaelEischer, line 224: Wouldn't it be better to use `ParsePatterns` or `preparePattern` to preparse the pattern first?

cmd/restic/cmd_backup.go
```diff
@@ -298,6 +299,11 @@ func collectRejectByNameFuncs(opts BackupOptions, repo *repository.Repository, t
 		if err != nil {
 			return nil, err
 		}
+
+		if valid, invalidPatterns := filter.ValidatePatterns(excludes); !valid {
```
MichaelEischer, line 303: The first two errors seem to be a duplicate for the last two?

MichaelEischer:
> I've added some tests and they fail with Go 1.15. Apparently [1.16](https://go.dev/doc/go1.16) changed the way `filepath.Match` reports syntax errors

Go can build file conditionally on the compiler version, for example to require at least go 1.16, add as first line: `//go:build go1.16` . So let's just make that test conditional on a recent enough go version.

cmd/restic/cmd_backup.go
lbausch: I've tweaked the error messages to include the used flag, e.g. `--iexclude`

internal/filter/filter_test.go
lbausch: Changed it accordingly 👍🏻

internal/filter/filter.go
lbausch: Changed it accordingly 👍🏻

lbausch:
If I understand the usage of `//go:build go1.16` correctly, it would skip the entire file, so I used `t.Skip()` to just skip this specific test if running Go 1.15. Is this correct?

`ValidatePatterns()` now calls `ParsePatterns()` and tries to validate every single pattern part by matching it against itself. For being able to include the original invalid pattern in the error message show to the user, I've extended the `Pattern` struct to include the original pattern.

MichaelEischer:
> If I understand the usage of `//go:build go1.16` correctly, it would skip the entire file, so I used `t.Skip()` to just skip this specific test if running Go 1.15. Is this correct?

Yes, a `go:build` comment applies to the entire file. The usual solution as e.g. used by `internal/backend/rest/rest_int_go114_test.go` is to create a separate file for that specific test.

> Is a fatal error too much? One might argue that a backup including unwanted files is better than no backup at all.

Hmm, an invalid exclude pattern will currently cause restic to ignore all later excludes and backup too many files. Returning a fatal error would change that into not creating a backup at all. Actually just filtering out invalid patterns would also lead to a change of behavior. Returning a fatal error might be to cleanest solution although it could be dangerous for users which don't check the exit code of restic. @fd0 Do you have a preference what to do here?

fd0:
Phew, that's a hard one. I'm (slightly) in favor of returning a fatal error. The pattern is clearly wrong and people **must** check the exit code.

lbausch:
Thank you :) I've updated the original MR description to describe all relevant changes.

MichaelEischer (approved):
LGTM. Thanks!

lbausch:
Thank you @MichaelEischer, @fd0 and @greatroar! 🙌🏻

PR #3735, opened 2022-05-03
DOCS: Incorrectly used SI units when IEC units were meant (fix #3669)

### LINKED ISSUE #3669: DOCS: Clarify that --exlude-larger-than <size> is in IEC units
Hi,

it's not clear from the docs (https://restic.readthedocs.io/en/stable/040_backup.html#excluding-files) that the units `K`, `M`, `G`, and `T` are in base-2 units.  That it's base-2 ("1024") units can be seen in the source code https://github.com/restic/restic/blob/44d543ede38466122dc489f11a4ffdf60f310182/cmd/restic/exclude.go#L385-L412.

Formally, base-2 units are part of the IEC standard and denoted KiB, MiB, GiB, and TiB (https://en.wikipedia.org/wiki/Byte#Unit_symbol).  As a reference, the SI standard defines base-10 units, e.g. kB (sic!), MB, GB, TB.

It looks like other parts of the docs use IEC units, but the section on `--exclude-larger-than` (https://github.com/restic/restic/blob/dc7a8aab2426e7d78d0a41b1608e75edf56a74d0/doc/040_backup.rst) incorrectly uses '1 MB' when it means `1 MiB`, and `2 kilobytes` when it means `2 KiB` or `2 kibibyte` (personally, I avoid the latter since it's less known), e.g.

```rst
Files larger than a given size can be excluded using the `--exclude-larger-than`
option:

.. code-block:: console

    $ restic -r /srv/restic-repo backup ~/work --exclude-larger-than 1M

This excludes files in ``~/work`` which are larger than 1 MB from the backup.

The default unit for the size value is bytes, so e.g. ``--exclude-larger-than 2048``
would exclude files larger than 2048 bytes (2 kilobytes). To specify other units,
suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for megabytes,
``g``/``G`` for gigabytes and ``t``/``T`` for terabytes (e.g. ``1k``, ``10K``, ``20m``,
``20M``,  ``30g``, ``30G``, ``2t`` or ``2T``).
```

# Suggestion

My proposal would be to rewrite it as:

```rst
Files larger than a given size can be excluded using the `--exclude-larger-than`
option:

.. code-block:: console

    $ restic -r /srv/restic-repo backup ~/work --exclude-larger-than 1M

This excludes files in ``~/work`` which are larger than 1 MiB from the backup.

The default unit for the size value is bytes, e.g. ``--exclude-larger-than 2048``
would exclude files larger than 2048 bytes (2 KiB). To specify other units,
suffix the size value with one of ``k``/``K`` for KiB (1024 bytes), ``m``/``M`` for MiB (1024^2 bytes),
``g``/``G`` for GiB (1024^3 bytes) and ``t``/``T`` for TiB (1024^4 bytes), e.g. ``1k``, ``10K``, ``20m``,
``20M``,  ``30g``, ``30G``, ``2t`` or ``2T``).
```

HenrikBengtsson:

Issue #3669 

Issue #3669 

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -346,12 +346,12 @@ option:
 
     $ restic -r /srv/restic-repo backup ~/work --exclude-larger-than 1M
 
-This excludes files in ``~/work`` which are larger than 1 MB from the backup.
+This excludes files in ``~/work`` which are larger than 1 MiB from the backup.
 
 The default unit for the size value is bytes, so e.g. ``--exclude-larger-than 2048``
-would exclude files larger than 2048 bytes (2 kilobytes). To specify other units,
-suffix the size value with one of ``k``/``K`` for kilobytes, ``m``/``M`` for megabytes,
-``g``/``G`` for gigabytes and ``t``/``T`` for terabytes (e.g. ``1k``, ``10K``, ``20m``,
+would exclude files larger than 2048 bytes (2 KiB). To specify other units,
+suffix the size value with one of ``k``/``K`` for KiB (1024 bytes), ``m``/``M`` for MiB (1024^2 bytes),
+``g``/``G`` for GiB (1024^3 bytes) and ``t``/``T`` for TiB (1024^4 bytes), e.g. ``1k``, ``10K``, ``20m``,
 ``20M``,  ``30g``, ``30G``, ``2t`` or ``2T``).
 
 Including Files


fd0 (approved):
LGTM! Thanks for taking the time to fix this!

PR #3773, opened 2022-05-29
Reduce memory usage for large directories/files

MichaelEischer:
restic requires a lot of memory to backup large files and directories. This PR reduces the peak memory usage by about 30-50% (depending on the GC settings). It is not intended as a complete solution to the memory usage problems, as that would require a repository format change as discussed in #2532.

Instead this PR takes the route of reducing the size of structs used during backup. Most notably the FutureNode/File/Tree structs. As the archiver preallocates the FutureNode array for each directory, the optimum would be to reduce the size of each FutureNode to a single reference. As we need a way to wait for the result, this would require a reference to a go channel. However, a channel seems to require about 96 bytes on 64-bit platforms which would regress the memory usage for files that are identical to the parent snapshot. Thus the FutureNode contains an additional reference to the result if it was already available.

The next problem is that to store a Tree in the repository, there are currently up to four items kept in-memory for each directory entry. First there is the FutureNode array, then the Nodes themselves, the serialized JSON of the tree and its encrypted variant while storing these in the repository.

As the FutureNode array is now rather small, it is sufficient to just clear all reference within the FutureNode to allow freeing most memory. The PR introduces a TreeBuilder which incrementally serialized the Tree and thereby ensures that each Node is only kept once in memory: either as Node or its serialized representation.

That way each node is represented in at most two items at a time in memory.

And finally the FutureBlob struct is slimmed down to a single reference to a channel. To reduce the memory usage further, FutureBlobs are collected as soon as possible.

Memory usage measurements:
using go 1.18.2, test data set 500k empty files in a single folder `for i in {1..500000} ; do touch abcefghijklt$i ; done`
test command: `/usr/bin/time -v $restic backup -n ../test/huge`

| commit | resident set size (KB) | elapsed | GOGC=1 rss (KB) | GOGC=1 elapsed |
|----|----|----|----|----|
| master | 1371208 | 0:18.72 | 830552 | 0:27.79 |
| remove fileinfo  | 1305528 | 0:17.88 | 770456 | 0:29.20 |
| unify node | 1056048 | 0:17.62 | 589192 | 0:34.06 |
| incremental tree |  887124 | 0:17.75 | 403124 | 0:34.45 |
| optimize large |  921864 | 0:17.72 | 402248 | 0:34.85 |

The table reports the results of the first test run, I've repeated the tests another two times, which result in the same overall trends.
The memory usage decreases by roughly 35% and a bit over 50% using GOGC=1. The measurements without `GOGC=1` are somewhat unstable as the GC can run at different points in time, but the general trend is stable across repetitions. The memory usage increase for `optimize large` appears to be an outlier as the test case uses no FutureBlobs. The elapsed time increases a bit for `GOGC=1` which seems to be caused by the different dataset that has to be scanned during GC. Without `GOGC` the elapsed time seems to be largely unchanged, although it is also too noisy for any further conclusions.

Related to #2446, but the specific changes were not discussed.

Checklist
---------

### DIFF
--- a/internal/restic/tree.go
+++ b/internal/restic/tree.go
@@ -1,6 +1,7 @@
 package restic
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -142,4 +143,53 @@ func SaveTree(ctx context.Context, r BlobSaver, t *Tree) (ID, error) {
 
 	id, _, _, err := r.SaveBlob(ctx, TreeBlob, buf, ID{}, false)
 	return id, err
+}
+
+type TreeJSONBuilder struct {
+	buf      bytes.Buffer
+	lastName string
+}
+
+func NewTreeJSONBuilder() *TreeJSONBuilder {
+	tb := &TreeJSONBuilder{}
+	_, _ = tb.buf.WriteString(`{"nodes":[`)
+	return tb
+}
+
+func (builder *TreeJSONBuilder) AddNode(node *Node) error {
+	if node.Name <= builder.lastName {
+		return errors.Errorf("nodes are not ordered got %q, last %q", node.Name, builder.lastName)
+	}
+	if builder.lastName != "" {
+		_ = builder.buf.WriteByte(',')
+	}
+	builder.lastName = node.Name
+
+	val, err := json.Marshal(node)
+	if err != nil {
+		return err
+	}
+	_, _ = builder.buf.Write(val)
+	return nil
+}
+
+func (builder *TreeJSONBuilder) Finalize() ([]byte, error) {
+	// append a newline so that the data is always consistent (json.Encoder
+	// adds a newline after each object)
+	_, _ = builder.buf.WriteString("]}\n")
+	buf := builder.buf.Bytes()
+	// drop reference to buffer
+	builder.buf = bytes.Buffer{}
+	return buf, nil
+}
+
+func TreeToBuilder(t *Tree) (*TreeJSONBuilder, error) {
+	builder := NewTreeJSONBuilder()
+	for _, node := range t.Nodes {
+		err := builder.AddNode(node)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return builder, nil
 }
--- a/internal/restic/tree_test.go
+++ b/internal/restic/tree_test.go
@@ -119,6 +119,37 @@ func TestEmptyLoadTree(t *testing.T) {
 		tree, tree2)
 }
 
+func TestTreeEqualSerialization(t *testing.T) {
+	files := []string{"node.go", "tree.go", "tree_test.go"}
+	for i := 1; i <= len(files); i++ {
+		tree := restic.NewTree(i)
+		builder := restic.NewTreeJSONBuilder()
+
+		for _, fn := range files[:i] {
+			fi, err := os.Lstat(fn)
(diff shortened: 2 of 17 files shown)


internal/restic/tree_test.go
```diff
@@ -94,6 +94,28 @@ func TestNodeComparison(t *testing.T) {
 	rtest.Assert(t, !node.Equals(n2), "nodes are equal")
 }
 
+func TestTreeEqualSerialization(t *testing.T) {
```
greatroar, line 97: Shouldn't this test check with more than one node in the tree to get better coverage?

internal/restic/tree.go
```diff
@@ -98,3 +100,63 @@ func (t *Tree) Subtrees() (trees IDs) {
...
+	// adds a newline after each object)
+	_, err := builder.buf.Write([]byte("]}\n"))
+	if err != nil {
+		return nil, err
+	}
+	buf := builder.buf.Bytes()
+	// drop reference to buffer
+	builder.buf = nil
+	return buf, nil
+}
```
greatroar, line 151: Follow-up to the above. The error return isn't necessary.

```suggestion
func (builder *TreeBuilder) Finalize() []byte {
	// append a newline so that the data is always consistent (json.Encoder
	// adds a newline after each object)
	_, _ = builder.buf.WriteString("]}\n")
	buf := builder.buf.Bytes()
	// drop reference to buffer
	builder.buf = bytes.Buffer{}
	return buf
}
```

internal/restic/tree.go
```diff
@@ -98,3 +100,63 @@ func (t *Tree) Subtrees() (trees IDs) {
...
+		buf: &buf,
+	}
+}
+
+func (builder *TreeBuilder) AddNode(node *Node) error {
+	if builder.lastName != "" {
+		err := builder.buf.WriteByte(',')
+		if err != nil {
+			return err
+		}
```
greatroar, line 125:
```suggestion
		_ = builder.buf.WriteByte(',')
```

internal/restic/tree_test.go
MichaelEischer: Now the test adds 1 to 3 files and compares the serialization.

internal/restic/tree.go
```diff
@@ -98,3 +100,63 @@ func (t *Tree) Subtrees() (trees IDs) {
...
+func NewTreeBuilder() *TreeBuilder {
+	var buf bytes.Buffer
+	_, err := buf.Write([]byte(`{"nodes":[`))
+	if err != nil {
+		panic(err)
+	}
+	return &TreeBuilder{
+		buf: &buf,
+	}
+}
```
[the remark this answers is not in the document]
MichaelEischer: Good idea. Thanks.

greatroar:
Minor nitpick: TreeBuilder isn't the perfect name for the new type as it doesn't produce trees. TreeJSONMarshaler? TreeJSONer? Otherwise, LGTM.

MichaelEischer:
What about TreeJSONBuilder? I've picked Builder as part of the name in an attempt to convey that it works incrementally.

greatroar:
Sounds good.

MichaelEischer:
Rebased to fix conflicts with #3733.

MichaelEischer:
I've rebased the PR to fix the conflict with #3830 and added a changelog entry

PR #3780, opened 2022-06-01
Remove existing path before restoring a symlink

### LINKED ISSUE #2578: Restore of a existing symlink(with same content) fails
Output of `restic version`
--------------------------
restic/restic                            latest              66a76c4e92a6        2 months ago        30.7MB

# First time, not existing:

docker run --rm -ti -e RESTIC_PASSWORD -e RESTIC_REPOSITORY -v restic-cache:/root/.cache/restic -v /srv/docker/restic:/srv/restic -v /srv/docker/clients/$site_id:/srv/app/$site_id:ro -v /srv/docker/clients/$site_id/restores/39e1d807:/srv/app/$site_id/restores/39e1d807:rw restic/restic restore --host $site_id --target /srv/app/$site_id/restores/39e1d807 39e1d807 --include /old/usr/share/zoneinfo/right/Turkey
repository 22780ffb opened successfully, password is correct
restoring <Snapshot 39e1d807 of [/srv/app/$site_id] at 2020-02-13 19:42:52 +0000 UTC by root@$site_id> to /srv/app/$site_id/restores/39e1d807
+ exit 0

# Second time, *immediately following*, exists:
docker run --rm -ti -e RESTIC_PASSWORD -e RESTIC_REPOSITORY -v restic-cache:/root/.cache/restic -v /srv/docker/restic:/srv/restic -v /srv/docker/clients/$site_id:/srv/app/$site_id:ro -v /srv/docker/clients/$site_id/restores/39e1d807:/srv/app/$site_id/restores/39e1d807:rw restic/restic restore --host $site_id --target /srv/app/$site_id/restores/39e1d807 39e1d807 --include /old/usr/share/zoneinfo/right/Turkey
repository 22780ffb opened successfully, password is correct
restoring <Snapshot 39e1d807 of [/srv/app/$site_id] at 2020-02-13 19:42:52 +0000 UTC by root@$site_id> to /srv/app/$site_id/restores/39e1d807
ignoring error for /old/usr/share/zoneinfo/right/Turkey: Symlink: symlink Asia/Istanbul /srv/app/$site_id/restores/39e1d807/old/usr/share/zoneinfo/right/Turkey: file exists
There were 1 errors

Local volume-mounted repo, via docker.

Expected behavior
-----------------

All other content files in the repo were restored during the second call, but any symlinks that already existed have errors printed, even if their targets are the same.  And, even if they were *different*, it shouldn't error, unless --verify exists(or some such).

No such error/warning was printed on the already existing regular files, so why does it do so on symlinks?

Actual behavior
---------------

Backup a symlink, restore it twice.

jkmw:
Bugfix: Make sure that symlinks can be created during recovery

When restoring a symlink, restic reported an error if the target path already existed.
With this change, the potentially existing target path is first removed before the symlink is recreated.

https://github.com/restic/restic/issues/2578

At the moment restic overwrites existing data in the restore target, unless it is a symlink, in which case an error is thrown.

From my point of view, restic should behave with symlinks in the same way as with all other file types and "overwrite" already existing paths.

This change causes restic to "overwrite" existing data (by delete and create) even if a symlink is to be restored.

See #2578 for more details.

Closes #2578

Checklist
---------

- [ x ] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ x ] I'm done! This pull request is ready for review.

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -300,7 +300,7 @@ func (node Node) createSymlinkAt(path string) error {
 		return nil
 	}
 
-	if err := os.RemoveAll(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+	if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
 		return errors.Wrap(err, "Symlink")
 	}
 
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -299,8 +299,12 @@ func (node Node) createSymlinkAt(path string) error {
 	if runtime.GOOS == "windows" {
 		return nil
 	}
-	err := fs.Symlink(node.LinkTarget, path)
-	if err != nil {
+
+	if err := os.RemoveAll(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return errors.Wrap(err, "Symlink")
+	}
+
+	if err := fs.Symlink(node.LinkTarget, path); err != nil {
 		return errors.Wrap(err, "Symlink")
 	}
 
--- a/changelog/unreleased/pull-3780
+++ b/changelog/unreleased/pull-3780
@@ -1,0 +1,7 @@
+Bugfix: Make sure that symlinks can be created during recovery
+
+When restoring a symlink, restic reported an error if the target path already existed.
+With this change, the potentially existing target path is first removed before the symlink is recreated.
+
+https://github.com/restic/restic/issues/2578
+https://github.com/restic/restic/pull/3780


internal/restic/node.go
```diff
@@ -299,8 +299,12 @@ func (node Node) createSymlinkAt(path string) error {
 	if runtime.GOOS == "windows" {
 		return nil
 	}
-	err := fs.Symlink(node.LinkTarget, path)
-	if err != nil {
+
+	if err := os.RemoveAll(path); err != nil && !errors.Is(err, os.ErrNotExist) {
```
MichaelEischer, line 303: Wouldn't it be sufficient to use `Remove` in most cases? I'm not sure whether someone would expect restic to wipe a directory just to restore a directory in its place?

MichaelEischer:
Also please add a short changelog entry using the template at https://github.com/restic/restic/blob/master/changelog/TEMPLATE

internal/restic/node.go
jkmw: My intention here was that I would expect to have the exact same state after a restore as I had at the time of the backup. Personally, I would be surprised if I had a symlink and after the restore there is still a directory instead of the sysmlink.
I would also be surprised if restic aborted with an error just because there was a directory in the restore destination that was a symlink at the time of the backup :shrug:.

MichaelEischer: I'd prefer to play it safe, return an error in case there is a directory here and let the user sort it out.

MichaelEischer:
I guess I haven't been explicitly enough in my previous comment: replacing `os.RemoveAll` with `os.Remove` was not a suggestion, but rather a request. The other parts of the restorer do not wipe out entire directory trees if these are standing in the way, thus restoring a symlink must behave just the same.

MichaelEischer (approved):
LGTM. Thanks!

PR #3787, opened 2022-06-05
repository: (Mostly) index-related cleanups

MichaelEischer:
This PR simplifies some of the index code and removes several functions that are only used for testing. In addition, it slightly improves the separation between the index and repository code. These changes are nearly sufficient to allow splitting the index from the repository package.

No.

Checklist
---------

### DIFF
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -131,27 +131,6 @@ var IndexFull = func(idx *Index, compress bool) bool {
 
 }
 
-// Store remembers the id and pack in the index.
-func (idx *Index) Store(pb restic.PackedBlob) {
-	idx.m.Lock()
-	defer idx.m.Unlock()
-
-	if idx.final {
-		panic("store new item in finalized index")
-	}
-
-	debug.Log("%v", pb)
-
-	// get packIndex and save if new packID
-	packIndex, ok := idx.packIDToIndex[pb.PackID]
-	if !ok {
-		packIndex = idx.addToPacks(pb.PackID)
-		idx.packIDToIndex[pb.PackID] = packIndex
-	}
-
-	idx.store(packIndex, pb.Blob)
-}
-
 // StorePack remembers the ids of all blobs of a given pack
 // in the index
 func (idx *Index) StorePack(id restic.ID, blobs []restic.Blob) {
--- a/internal/repository/index.go
+++ b/internal/repository/index.go
@@ -46,8 +46,6 @@ type Index struct {
 	byType     [restic.NumBlobTypes]indexMap
 	packs      restic.IDs
 	mixedPacks restic.IDSet
-	// only used by Store, StorePacks does not check for already saved packIDs
-	packIDToIndex map[restic.ID]int
 
 	final      bool       // set to true for all indexes read from the backend ("finalized")
 	ids        restic.IDs // set to the IDs of the contained finalized indexes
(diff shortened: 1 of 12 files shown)


internal/repository/index.go
```diff
@@ -223,22 +181,6 @@ func (idx *Index) Has(bh restic.BlobHandle) bool {
 	return idx.byType[bh.Type].get(bh.ID) != nil
 }
 
-// LookupSize returns the length of the plaintext content of the blob with the
```
greatroar, line 226: Are you sure we don't want to keep a separate path for looking up sizes? It short-circuits when there are duplicates and it doesn't allocate.

greatroar:
repository.Loader is also unused.

internal/restic/repository.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/restic/restic/internal/crypto"
+	"github.com/restic/restic/internal/ui/progress"
```
greatroar, line 7: I think this is a sign that (after this PR) we should drop this file and the [single-implementation interfaces](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a) in it, and just let client code work on *repository.Repository instead. Changes to that type already require changes here, and now every package that wants to handle a restic.ID drags in ui/progress. That's not a very big deal yet, but it does mean that ui/progress cannot be told about restic.ID when the need arises.

MichaelEischer: We currently have a few tests that wrap `restic.Repository` for the checker and the archiver. These wouldn't be possible anymore if we drop the interface. Although the wrapper repos only use `LoadTree`, `LoadBlob` and `LookupBlobSize`. Replacing these wrappers with different interfaces would probably be rather messy.

Dragging in the progress.Counter is a bit unfortunate, but if necessary we can decouple that with an additional interface ;-) .

I also plan to split the restic package into a backend part, the snapshot data model part and the repository related parts. There's a bunch of totally experimental commits at https://github.com/MichaelEischer/restic/tree/layered-repo2 which includes the just mentioned refactoring and a rather random collection of other stuff.

internal/repository/index.go
MichaelEischer: I'm not too happy that `LookupSize` has to reimplement `Blob.DataLength()`, but I'm not settled on getting rid of that function or not.

internal/restic/repository.go
greatroar: If we define checker.Repository and archiver.Repository with the set of operations those packages need, then IMO that's not messy at all. It signals that checker needs one set of operations and archiver another. There's a motto "[accept interfaces, return structs](https://nathanleclaire.com/blog/2015/10/10/interfaces-and-composition-for-effective-unit-testing-in-golang/)", and although it's [controversial](https://github.com/go-proverbs/go-proverbs.github.io/issues/37), I think it applies here.

This is all beside the point, which is that your PR LGTM. Testing the waters for a further refactor, which I'll postpone. Looking forward to your follow-up :)

fd0: Cleaning up all my old beginner mistakes, awesome! :)

internal/repository/index.go
MichaelEischer: I've removed the commit for now.

MichaelEischer:
So, then let's get this merged.

PR #3802, opened 2022-06-17
Support non utf symlink targets

### LINKED ISSUE #3311: Symlinks with paths that are invalid utf8 not backed up correctly
Output of `restic version`
--------------------------
Current master / restic 0.12.0

```
mkdir test
ln -s '../i/'$'\355\246\361''d/samba' test/invalid-symlink
restic -r testrepo init
restic -r testrepo backup test
restic -r testrepo ls latest -l
```
```
repository 3789a761 opened successfully, password is correct
snapshot 35996ae0 of [/home/test/test] filtered by [] at 2021-02-27 23:57:12.4010466 +0100 CET):
drwxr-xr-x  1000  1000      0 2021-02-27 23:56:20 /test
Lrwxrwxrwx  1000  1000      0 2021-02-27 23:56:20 /test/invalid-symlink -> ../i/���d/samba
```
Note the unicode replacement rune, instead of the expected bytes in the symlink. Piping the output through `xxd` just contains the replacement character `efbfbd` three times:
```
000000e0: 696e 6b20 2d3e 202e 2e2f 692f efbf bdef  ink -> ../i/....
000000f0: bfbd efbf bd64 2f73 616d 6261 0a         .....d/samba.
```

This is also visible in the stored tree node:
```
> restic -r testrepo -q cat blob 787216602e[...]
{"nodes":[{"name":"invalid-symlink","type":"symlink","mode":134218239,"mtime":"2021-02-27T23:56:20.22+01:00","atime":"2021-02-27T23:56:20.22+01:00","ctime":"2021-02-27T23:56:20.22+01:00","uid":1000,"gid":1000,"user":"michael","group":"michael","inode":132367,"device_id":2064,"links":1,"linktarget":"../i/\ufffd\ufffd\ufffdd/samba","content":null}]}
```

Expected behavior
-----------------
Symlinks with non-utf8 targets should be stored correctly. Although I'd assume those to be very rare.

Actual behavior
---------------
Only the unicode replacement rune is stored.

`json.Marshal` replaces invalid utf8 characters with the Unicode replacement rune: (https://golang.org/pkg/encoding/json/#Marshal)

> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune.

No. At least not without changing the repository format.

MichaelEischer:
Currently symlinks whose target name is invalid utf8 are not backed up correctly. The invalid characters are replaced with `\ufffd`. This PR introduces an optional quoted field which is used for symlink targets which might not encode correctly. The old `linktarget` field in a Node still exists and continues to be set to stay backwards compatible to older clients. Thus, for old clients the symlink will behave just as before, whereas later restic versions which support the new field will use that one automatically instead.

This allows for a backwards compatible addition of a correct symlink encoding. I've decided against binding the symlink encoding to a specific repository version as that would create all sorts of headaches when copying snapshots between repositories with different versions.

Fixes #3311

Checklist
---------

### DIFF
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"unicode/utf8"
 
 	"github.com/restic/restic/internal/errors"
 
@@ -27,21 +28,26 @@ type ExtendedAttribute struct {
 
 // Node is a file, directory or other item in a backup.
 type Node struct {
-	Name               string              `json:"name"`
-	Type               string              `json:"type"`
-	Mode               os.FileMode         `json:"mode,omitempty"`
-	ModTime            time.Time           `json:"mtime,omitempty"`
-	AccessTime         time.Time           `json:"atime,omitempty"`
-	ChangeTime         time.Time           `json:"ctime,omitempty"`
-	UID                uint32              `json:"uid"`
-	GID                uint32              `json:"gid"`
-	User               string              `json:"user,omitempty"`
-	Group              string              `json:"group,omitempty"`
-	Inode              uint64              `json:"inode,omitempty"`
-	DeviceID           uint64              `json:"device_id,omitempty"` // device id of the file, stat.st_dev
-	Size               uint64              `json:"size,omitempty"`
-	Links              uint64              `json:"links,omitempty"`
-	LinkTarget         string              `json:"linktarget,omitempty"`
+	Name       string      `json:"name"`
+	Type       string      `json:"type"`
+	Mode       os.FileMode `json:"mode,omitempty"`
+	ModTime    time.Time   `json:"mtime,omitempty"`
+	AccessTime time.Time   `json:"atime,omitempty"`
+	ChangeTime time.Time   `json:"ctime,omitempty"`
+	UID        uint32      `json:"uid"`
+	GID        uint32      `json:"gid"`
+	User       string      `json:"user,omitempty"`
+	Group      string      `json:"group,omitempty"`
+	Inode      uint64      `json:"inode,omitempty"`
+	DeviceID   uint64      `json:"device_id,omitempty"` // device id of the file, stat.st_dev
+	Size       uint64      `json:"size,omitempty"`
+	Links      uint64      `json:"links,omitempty"`
+	LinkTarget string      `json:"linktarget,omitempty"`
+	// implicitly base64-encoded field. Only used while encoding, `linktarget_raw` will overwrite LinkTarget if present.
+	// This allows storing arbitrary byte-sequences, which are possible as symlink targets on unix systems,
+	// as LinkTarget without breaking backwards-compatibility.
+	// Must only be set of the linktarget cannot be encoded as valid utf8.
+	LinkTargetRaw      []byte              `json:"linktarget_raw,omitempty"`
 	ExtendedAttributes []ExtendedAttribute `json:"extended_attributes,omitempty"`
 	Device             uint64              `json:"device,omitempty"` // in case of Type == "dev", stat.st_rdev
 	Content            IDs                 `json:"content"`
@@ -344,6 +350,13 @@ func (node Node) MarshalJSON() ([]byte, error) {
 	nj := nodeJSON(node)
 	name := strconv.Quote(node.Name)
 	nj.Name = name[1 : len(name)-1]
+	if nj.LinkTargetRaw != nil {
+		panic("LinkTargetRaw must not be set manually")
+	}
+	if !utf8.ValidString(node.LinkTarget) {
+		// store raw bytes if invalid utf8
+		nj.LinkTargetRaw = []byte(node.LinkTarget)
+	}
 
 	return json.Marshal(nj)
 }
@@ -358,7 +371,14 @@ func (node *Node) UnmarshalJSON(data []byte) error {
 	}
 
 	nj.Name, err = strconv.Unquote(`"` + nj.Name + `"`)
-	return errors.Wrap(err, "Unquote")
+	if err != nil {
+		return errors.Wrap(err, "Unquote")
+	}
+	if nj.LinkTargetRaw != nil {
+		nj.LinkTarget = string(nj.LinkTargetRaw)
+		nj.LinkTargetRaw = nil
+	}
+	return nil
 }
 
 func (node Node) Equals(other Node) bool {
--- a/internal/restic/node_test.go
+++ b/internal/restic/node_test.go
@@ -2,6 +2,8 @@ package restic_test
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -10,6 +12,7 @@ import (
 	"time"
 
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/test"
 	rtest "github.com/restic/restic/internal/test"
 )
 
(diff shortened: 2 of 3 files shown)


internal/restic/node.go
```diff
@@ -28,21 +28,24 @@ type ExtendedAttribute struct {
...
+	GID        uint32      `json:"gid"`
+	User       string      `json:"user,omitempty"`
+	Group      string      `json:"group,omitempty"`
+	Inode      uint64      `json:"inode,omitempty"`
+	DeviceID   uint64      `json:"device_id,omitempty"` // device id of the file, stat.st_dev
+	Size       uint64      `json:"size,omitempty"`
+	Links      uint64      `json:"links,omitempty"`
+	LinkTarget string      `json:"linktarget,omitempty"`
+	// only used while encoding, to remain backwards compatible we cannot change the interpretation of `linktarget`
+	// if the encoded JSON includes `linktarget_quoted` then it will replace `LinkTaget` after decoding
```
greatroar, line 47:
```suggestion
	// if the encoded JSON includes `linktarget_quoted`, then it will replace `LinkTarget` after decoding
```

greatroar: So the rule is: if linktarget_quoted is present and not "", it overrides linktarget? I think the comment should mention that and explain why this is done.

internal/restic/node_test.go
```diff
@@ -296,3 +299,38 @@ func TestFixTime(t *testing.T) {
 		})
 	}
 }
+
+func TestSymlinkSerialization(t *testing.T) {
+	for _, link := range []string{
+		"test",
```
greatroar, line 305:
```suggestion
		"válîd \t Üñi¢òde \n śẗŕinǵ",
```

internal/restic/node.go
```diff
@@ -343,12 +354,23 @@ func (node Node) MarshalJSON() ([]byte, error) {
...
-	nj.Name = name[1 : len(name)-1]
+	nj.Name = quote(node.Name)
+	if nj.LinkTargetQuoted != "" {
+		panic("LinkTargetQuoted must not be set manually")
+	}
+	lt := quote(node.LinkTarget)
+	if lt != node.LinkTarget {
+		// only store if the quoted version is different
+		nj.LinkTargetQuoted = lt
+	}
```
greatroar, line 365: Potential optimization:

```suggestion
	if !utf8.ValidString(node.LinkTarget) {
		nj.LinkTargetQuoted = quote(node.LinkTarget)
	}
```

greatroar:
Seems like a sound design overall. Does need a changelog entry, though.

aawsome:
Let's say, you have a file names `"` linking to `"a`.
Then current restic generated the json:
`{"nodes":[{"name":"\\\"","linktarget":"\"a"},{"name":"\\\"a"}]}`

So, quoting filenames but not quoting `linktarget` not only broke non-unicode linktargets, but also introduced a mess in the tree format.

Of course you can add more quirks and either add stuff to handle only unicode (leaving this as above) or also change this example to:
`{"nodes":[{"name":"\\\"","linktarget":"\"a", "linktarget_quoted":"\\\"a"},{"name":"\\\"a"}]}`

Sorry, but this is still a mess!

IMO the only clean solution is to deprecate `linktarget` and add a new field like `target` to let this always look 
`{"nodes":[{"name":"\\\"", "target":"\\\"a"},{"name":"\\\"a"}]}`
(or rethink the quoting altogether, but this you have already declined in #3808 ...)

I know, this breaks forward-compatibility but so does the repo v2 format which supports compression. I know that this adds some headache if you want to correct existing repos. But this could be done by adding this to a new repo version an supplying a migrate command.

internal/restic/node.go
aawsome: This "optimizaion" IMHO only increases the mess, see my comment about non-unicode but escaped filenames below.

greatroar: Whoops, thinko.

aawsome:
> Just a comment from the sidelines: if the symlink target field is just an arbitrary sequence of bytes, then we should handle it as such and store/restore a byte slice, e.g. keep the `linktarget` field, but add a `linktarget_raw` field which is `[]byte` and encodes as base64 in JSON.

@fd you are right. And I also strongly support a convention that valid UTF-8 (which is the standard case) should be stored as-is, as this makes reading the JSON tree (e.g. using `restic cat`) much easier for most files.

However, all these arguments also apply to the `name` field...

internal/restic/node_test.go
MichaelEischer: is added

internal/restic/node.go
MichaelEischer: I've expanded the comment.

MichaelEischer:
Hmm, the cost-benefit calculation would probably be different if we clean up the `name` field along with other breaking changes. I'm thinking of adding support for chunked trees to the nodes, which won't be possible in a backwards compatible way. But that could be an opportunity to clean up other stuff.

I'm wondering whether it would be a good idea to just store `name` and `linktarget` always as base64 encoded. It won't make much of a difference in the filesize, at least after it's compressed. But it would simplify decoding.

aawsome:
@MichaelEischer May I suggest to also add a `name_raw` field which is filled if the (encoded) name field doesn't exactly reflect the filename? When created that way, program reading trees can always base64-decode a `*_raw` field when given and use the `*` (without `_raw`) field directly, if not.

Of course that wouldn't work with already-written data or older program versions. But it would provide a migration path towards a clean solution by first making the `name_raw` field optional, then required and later deprecating `name` fields which contain encoding.

MichaelEischer:
@aawsome I'd like to discuss the `name` field separately  from `linktarget`. I've created issue https://github.com/restic/restic/issues/4408  for that. The main difference in my opinion is that the `linktarget` modifications add support for non-UTF8 symlink targets, whereas the suggested `name` changes are purely cosmetic (strconv.Quote is perfectly capable of handling non-UTF8).

> Of course that wouldn't work with already-written data or older program versions. But it would provide a migration path towards a clean solution by first making the `name_raw` field optional, then required and later deprecating `name` fields which contain encoding.

So your suggestion is to only keep `name_raw` and `linktarget_raw` and deprecate both `name` and `linktarget`? If we make such a change, I'd like to coordinate it with https://github.com/restic/restic/pull/4006 and the support for chunked trees (likely won't happen this year) to minimize the resulting disruptions.

aawsome:
> (strconv.Quote is perfectly capable of handling non-UTF8).

I actually have two issues with `strconv.Quote`:
- I personally am not able to judge if it is perfectly capable of handling non-UTF8. I tried to find a specification of what it does and wasn't able so far. Even in https://github.com/restic/restic/blob/master/doc/design.rst we are not able to tell, what it exactly does. So it's kind of a black box we are using here. For a backup format this IMO is even worse than an open question about whether it's possible to handle non-UTF.
- Its use is easy to oversee as for many common filenames it doesn't change anything. This leads uninformed consumers to just use that field - and maybe make mistakes. E.g. using `\"abc\"` as filename which in fact was `"abc"` before getting backup'ed. See https://github.com/juergenhoetzel/pyrrhic/issues/1 for example.

> So your suggestion is to only keep `name_raw` and `linktarget_raw` and deprecate both `name` and `linktarget`?

No. Actually I like the solution you implemented here with `linktarget` and `linktarget_raw` very much. So I propose to change the repository tree-format step-by-step to also use `name` and `name_raw` in exactly the same way. I.e. deprecate the usage of `strconv.Quote` on the `name` field.
(an alternative could be introduce new fields like `filename`/`filename_raw` and deprecate `name`.)

MichaelEischer:
LGTM.

aawsome:
@MichaelEischer Shouldn't this treatment of symlink targets be added to https://github.com/restic/restic/blob/master/doc/design.rst?

aawsome:
Another remark - and sorry if there are errors or things are not fully understandable, I don't use windows and am no expert in its details (just did dig into how Rust handles filenames and it is very pedantic about being correct)...

Actually, I still don't know what's going on on Windows and IMO this also needs documentation. On Windows, filenames are not stored as byte-sequences but as sequences of 16-bit words. This means, we cannot talk about utf8 on windows, at least when we are talking about how the OS stores filenames. I know that Golang does some internal conversion, so all filenames always look at byte-sequences and you can treat them as UTF8 if they are valid unicode. But I don't know how non-unicode filenames (if they are allowed) are converted from 16-bit words to an 8-bit byte sequence. Moreover, I don't even know if there are cases where a valid Windows filename cannot be transformed to a non-UTF8 byte sequence and what Golang would do in such a case. Can anyone provide information about this topic? 


Besides this, I also don't know if this linktarget topic anyway applies to Windows or if this topic would only arise when applying it to "standard" filenames....

MichaelEischer:
> @MichaelEischer Shouldn't this treatment of symlink targets be added to https://github.com/restic/restic/blob/master/doc/design.rst?

Oh indeed. I'll open a PR later.

> On Windows, filenames are not stored as byte-sequences but as sequences of 16-bit words.

The Windows API methods suffixed with a `W` use UTF-16 which is converted by Go into UTF-8 . On the filesystem NTFS stores all names as Unicode: https://learn.microsoft.com/en-us/windows/win32/intl/character-sets-used-in-file-names .

aawsome:
> The Windows API methods suffixed with a `W` use UTF-16 which is converted by Go into UTF-8 . On the filesystem NTFS stores all names as Unicode:

Ah great, thanks for the information!
So, on Windows there is no non-unicode topic at all and a conversion into UTF-8 is sufficient. Maybe this is also a good info for the design document?

MichaelEischer:
I've opened https://github.com/restic/restic/pull/4422

PR #3840, opened 2022-07-30
Speed up restic init over slow SFTP links

### LINKED ISSUE #3837: restic init performance with SFTP backend over high latency connection
Output of `restic version`
--------------------------

`restic 0.13.1 compiled with go1.18 on linux/amd64`

Small opportunity for performance enhancement: `restic init` could be faster when creating a new repository with the SFTP backend over a high latency connection.

In detail:

- initialising a new restic repo for SFTP backend creates 256 directories of the form `data/00` .. `data/ff`
- the directories are created sequentially, in a loop, by [(r *SFTP) mkdirAllDataSubdirs()](https://github.com/restic/restic/blob/4ffd479ba490494eecd1d5b720e505b9c0c19567/internal/backend/sftp/sftp.go#L158-L167)
- inside each loop cycle there is a call to [`pkg/sftp` `Client.MkdirAll`](https://pkg.go.dev/github.com/pkg/sftp#Client.MkdirAll)
- It looks like each `pkg/sftp` `Client.MkdirAll` will typically do [3 sequential network roundtrips](https://github.com/pkg/sftp/blob/cf523d99b5d2f41128caaaafc4c908a6a4d2fa0e/client.go#L879-L925): stat the target path, find it doesn't exist. then stat the parent, find it exists. then mkdir the target path.
- suggests `restic init` for SFTP backend will do at least 3 * 256 = 768 sequential network round trips
- perhaps time to init a repo with SFTP connection could be decreased from `768 * L` to around `6 * L`, where `L` is roundtrip latency. First `MkdirAll` the parent `data` dir, then `MkdirAll` all 256 child directories `data/00` .. `data/ff` concurrently.

Why improving this might be low priority:
- only users using SFTP backend who also have high latency connections may benefit
- `restic init` operation is only done once in lifecycle of a repository. Limited effort to improve performance might be better spent speeding up frequent operations such as `restic backup`

I was trying to do`restic init` operation over SFTP backend with 400 ms roundtrip ping latency.

`restic init` operation worked, but took so long I had time to start investigating why it might be slow ;).

I was looking for tool to help automate backups, and restic is fantastic:
- supports variety of backends, easy to integrate with my chosen storage providers over standard protocol (SFTP in my case)
- backups are encrypted, using a key that i manage, not the storage provider
- great documentation
- tool is single, easy to deploy binary

thank you to all the maintainers. it is great.

greatroar:

pkg/sftp.Client.MkdirAll(d) does a Stat to determine if d exists and is a directory, then a recursive call to create the parent, so the calls for data/?? each take three round trips. Doing a Mkdir first should eliminate two round trips for 255/256 data directories as well as all but one of the top-level directories.

Also, we can do all of the calls concurrently. This may reintroduce some of the Stat calls when multiple goroutines try to create the same parent, but at the default number of connections, that should not be much of a problem.

Fixes #3837 with a somewhat simpler version of the solution suggested by @fcostin.

Checklist
---------

### DIFF
--- a/internal/backend/sftp/sftp.go
+++ b/internal/backend/sftp/sftp.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/pkg/sftp"
+	"golang.org/x/sync/errgroup"
 )
 
 // SFTP is a backend in a directory accessed via SFTP.
@@ -154,15 +155,29 @@ func Open(ctx context.Context, cfg Config) (*SFTP, error) {
 	return sftp, nil
 }
 
-func (r *SFTP) mkdirAllDataSubdirs() error {
+func (r *SFTP) mkdirAllDataSubdirs(ctx context.Context, nconn uint) error {
+	// Run multiple MkdirAll calls concurrently. These involve multiple
+	// round-trips and we do a lot of them, so this whole operation can be slow
+	// on high-latency links.
+	g, _ := errgroup.WithContext(ctx)
+	// Use errgroup's built-in semaphore, because r.sem is not initialized yet.
+	g.SetLimit(int(nconn))
+
 	for _, d := range r.Paths() {
-		err := r.c.MkdirAll(d)
-		if err != nil {
-			return err
-		}
+		d := d
+		g.Go(func() error {
+			// First try Mkdir. For most directories in Paths, this takes one
+			// round trip, not counting duplicate parent creations causes by
+			// concurrency. MkdirAll first does Stat, then recursive MkdirAll
+			// on the parent, so calls typically take three round trips.
+			if err := r.c.Mkdir(d); err == nil {
+				return nil
+			}
+			return r.c.MkdirAll(d)
+		})
 	}
 
-	return nil
+	return g.Wait()
 }
 
 // Join combines path components with slashes (according to the sftp spec).
@@ -240,7 +255,7 @@ func Create(ctx context.Context, cfg Config) (*SFTP, error) {
 	}
 
 	// create paths for data and refs
-	if err = sftp.mkdirAllDataSubdirs(); err != nil {
+	if err = sftp.mkdirAllDataSubdirs(ctx, cfg.Connections); err != nil {
 		return nil, err
 	}
 
--- a/internal/backend/sftp/sftp.go
+++ b/internal/backend/sftp/sftp.go
@@ -116,8 +116,7 @@ func (r *SFTP) clientError() error {
 }
 
 // Open opens an sftp backend as described by the config by running
-// "ssh" with the appropriate arguments (or cfg.Command, if set). The function
-// preExec is run just before, postExec just after starting a program.
+// "ssh" with the appropriate arguments (or cfg.Command, if set).
 func Open(ctx context.Context, cfg Config) (*SFTP, error) {
 	debug.Log("open backend with config %#v", cfg)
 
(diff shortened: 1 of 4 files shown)


MichaelEischer (commented):
The code changes look good. But I think it would be good to have a very short changelog entry.

fcostin:
thank you @greatroar for taking a look at this, much appreciated.

I built this locally using `go1.18.1 linux/amd64` and tested out this repo init operation over my high latency sftp connection. Gives 8x speedup with default options and 14x speedup when `sftp.connections` set to a very high number:

| version | options | real time to `restic init -r repo` (seconds) |
| ------------- | ------------- | ------------- |
|  4ffd479b (baseline) |  used defaults | 340.3 |
|  2bdc40e6 (this PR) |  used defaults | 41.2 |
|  2bdc40e6 (this PR) |  `-o stfp.connections=300` | 22.9 |

greatroar:
Wrote the changelog!

internal/backend/sftp/sftp.go
```diff
@@ -155,15 +155,29 @@ func Open(ctx context.Context, cfg Config) (*SFTP, error) {
...
+		d := d
+		g.Go(func() error {
+			// First try Mkdir. For most directories in Paths, this takes one
+			// round trip, not counting duplicate parent creations causes by
+			// concurrency. MkdirAll first does Stat, then recursive MkdirAll
+			// on the parent, so calls typically take three round trips.
+			if err := r.c.Mkdir(d); err == nil {
+				return nil
+			}
+			return r.c.MkdirAll(d)
```
fcostin, line 176: comment only, no change suggested:

There is a scenario where we concurrently attempt to make two or more child directories e.g. `somerepo/data/00` and `somerepo/data/01` before the parent `somerepo/data` has been created. 

I was a bit concerned about this, that it could lead to failures in some (rare? cases), but after reading through the pkg/sftp code and thinking about it, it is probably fine:

Suppose goroutine A will process `somerepo/data/00`, goroutine B will process `somerepo/data/01` and goroutine C will process `somerepo/data` ... assume C will be scheduled later, and we think about A and B.

If `somerepo/data/00` and `somerepo/data/01` are processed sequentially, there is no problem.

Suppose execution of `somerepo/data/00` `somerepo/data/01` is interleaved.

A tricky interleaving could be where both A and B both decide that the parent directory `somerepo/data` does not exist, then try to create it at the same time:

```
A Mkdir : `SSH_FXP_MKDIR` "somerepo/data/00" # error as parent dir data doesn't exist
B Mkdir : `SSH_FXP_MKDIR` "somerepo/data/01" # error as parent dir data doesn't exist
A MkdirAll: `SSH_FXP_STAT` "somerepo/data/00 # result: this child dir doesn't exist
B MkdirAll: `SSH_FXP_STAT` "somerepo/data/01 # result: this child dir doesn't exist

# both MkdirAll calls recursively attempt to create the parent dir "somerepo/data"
# first they stat the path

A MkdirAll : `SSH_FXP_STAT` "somerepo/data" # result: this parent dir doesn't exist
B MkdirAll : `SSH_FXP_STAT` "somerepo/data" # result: this parent dir doesn't exist

# then they both try to create it:

A MkdirAll : `SSH_FXP_MKDIR` "somerepo/data" # suppose this is processed by server first and succeeds
B MkdirAll : `SSH_FXP_MKDIR` "somerepo/data" # fails, cannot mkdir since data directory already exists
```

So there is some interleaving where both goroutines A and B attempt to create the same parent `data` dir, and the operation that is processed second fails.

I suspect this may be okay in practice as there is code in the `pkg/sftp` client implementation of `MkdirAll` that does an `SSH_FXP_LSTAT` operation if there is an error following `SSH_FXP_MKDIR`.  If the LSTAT succeeds and says that the path exists and is a dir, then it looks like the client code for `MkdirAll` will ignore the error from failing to create a dir that already exists.

https://github.com/pkg/sftp/blob/v1.13.4/client.go#L899-L908

MichaelEischer (approved):
LGTM. Thanks!

internal/backend/sftp/sftp.go
greatroar: sftp.Client.MkdirAll actually promises that this will always work:

    // If path is already a directory, MkdirAll does nothing and returns nil.

Issue #3594, opened 2021-12-14
A Dokany mounted filesystem can not be backed up


cccapon:
Output of `restic version`
--------------------------
restic 0.12.1 compiled with go1.16.6 on windows/amd64

How did you run restic exactly?
-------------------------------
PS C:\Users\Chris\Robust\Backup> restic.exe --repo '\\ccnas\Backup\Robust\restic\' backup 'C:\Users\Chris\Robust\Client\UofT\' -v
open repository
enter password for repository:
repository d847d773 opened successfully, password is correct
lock repository
load index files
using parent snapshot 28c247f8
start scan on [C:\Users\Chris\Robust\Client\UofT\]
start backup on [C:\Users\Chris\Robust\Client\UofT\]
scan finished in 1.012s: 0 files, 0 B
error: NodeFromFileInfo: Readlink: readlink \\?\C:\Users\Chris\Robust\Client\UofT: The system cannot find the file specified.

Files:           0 new,     0 changed,     0 unmodified
Dirs:            0 new,     0 changed,     5 unmodified
Data Blobs:      0 new
Tree Blobs:      0 new
Added to the repo: 0 B

processed 0 files, 0 B in 0:01
snapshot 255e9b8c saved
Warning: at least one source file could not be read
PS C:\Users\Chris\Robust\Backup> dir 'C:\Users\Chris\Robust\Client\UofT\'

    Directory: C:\Users\Chris\Robust\Client\UofT

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----        2021-12-10   5:35 PM                Case
------        2021-12-10   5:35 PM           6210 UofT.bmp
------        2021-12-10   5:35 PM           1541 IMG_0019.JPG
------        2021-12-10   5:35 PM          21504 Notes.xls
------        2021-12-10   5:35 PM          34304 t.doc

PS C:\Users\Chris\Robust\Backup> restic.exe --repo '\\ccnas\Backup\Robust\restic\' backup 'C:\Users\Chris\Robust\Client\UofT\Case\' -v
open repository
enter password for repository:
repository d847d773 opened successfully, password is correct
lock repository
load index files
no parent snapshot found, will read all files
start scan on [C:\Users\Chris\Robust\Client\UofT\Case\]
start backup on [C:\Users\Chris\Robust\Client\UofT\Case\]
signal interrupt received, cleaning up

[...]

Issue #2564, opened 2020-02-03
Option to follow symlinks _in path arguments_


bland328:
Output of `restic version`
--------------------------
`restic 0.9.6 compiled with go1.13.4 on linux/amd64`

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
I would very much like for `restic` to have an option to follow symlinks solely **within the paths passed to `restic`**.

_(This resembles https://github.com/restic/restic/issues/2211, but it is nearly unrelated. And it overlaps with https://github.com/restic/restic/issues/542, but this request is focused in a way that issue is not; if I should've posted this in there, please let me know.)_

Right now, if I do...
`$ restic backup /path/to/a/symlink_to_a_dir`
or
`$ restic backup /path/to/a/symlink_to_a_file`
...I get a snapshot containing a single symlink, which makes sense.

**I would _greatly_ value being able to do something along these lines...**
`$ restic --follow-arg-symlinks backup /path/to/a/symlink_to_a_dir`
...and get a snapshot _named_ with the path "`/path/to/a/symlink_to_a_dir`" but with the _contents_ of `/path/to/the/real_symlinked_dir`

The concept is very much like this `find` option:
`-H`  `Do not follow symbolic links, except while processing the command line arguments.`

What are you trying to do?
--------------------------
To be clear, this isn't at all about following symlinks _within_ a backup fileset _(I grasp why_ that's _a big can of worms)_; _this_ is solely about optionally following symlinks to find to the root of the backup fileset.

One of my use cases is that I use `urbackup` to archive some local clients to the server I'm backing up with `restic`, and it manages its backup "snapshots" over time as new folders containing both changed files and hardlinks to unchanged files, and always keeps a `current` symlink pointing to the latest backup.

Since I'm only interested in backing up that `current` backup, it would be valuable to be able to simply point `restic` to it with a `--follow-arg-symlinks` _(or similar)_ option.

Right now, I'm scripting my way around that, but the downside is that each new `restic` backup is a snapshot named with a different path; what I _want_ is simply a series of snapshots named with the `/path/to/urbackup/client/symlink_to_current` path, but with the contents of the directory referenced by the symlink.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
restic helps me nearly every day. There are a few rough edges yet to file down and things I pray it will someday be able to do, but I depend on it regularly. I'm pleased and very thankful.

smlx:
Does this do what you're after?
```
[...]

Issue #542, opened 2016-07-15
Report a warning, if only symlinks are being backed up


viric:
Yesterday I thought I was backing up a directory...Yesterday I thought I was backing up a directory...let's say, ~/mydoc, and I forgot that ~/mydoc is just a symlink to ~/doc/own

So I was backing up a symlink... It happened for several days in a row so I don't have the backups I expected.

I wish restic had said me something like: "you are backing up a symlink to a directory, not a directory".

Definitely it correctly reported "0 directories 1 file" backed up, but it took me some days to notice.

cfcs:
This sounds like a nice warning. Perhaps a `--follow-symlinks` would also be nice?

fd0:
Let's track implementing the warning in this issue, and leave following symlinks for another one.

alphapapa:
Just ran into this myself.  One of my most important directories is a symlink in my homedir that points to the real directory deeper in my homedir tree.  Thankfully I caught it while testing restic, rather than after I started depending on it.

I'm not sure what the "correct" behavior would be; I can imagine some users would want to backup and restore symlinks rather than what they point to.  But I also imagine that some users will expect the files/directories pointed to by symlinks to be backed up, and that they will be in for a nasty surprise someday when they need to run a restore.

So I realize it's not an easy call, but I'm not sure I feel safe relying on restic if I have to keep this limitation in mind every time I create a symlink.

cfcs:
Remember to ignore the warning condition if the target of the symlink is indeed included in the backup set.
- You should pay attention to how `exclude` patterns may differ for the two directories (what happens if I exclude `~/.config` and include `~/allmyimportantthings` which happens to point to a directoy in `~/.config` ?)

alphapapa:
> You should pay attention to how exclude patterns may differ for the two directories (what happens if I exclude ~/.config and include ~/allmyimportantthings which happens to point to a directoy in ~/.config ?)

Yes, what does happen in that case?

fd0:
Thanks for all the input, but I'm not sure if having an option to follow symlinks is a good idea after all. It is very easy to create endless loops, like this:
```
$ tree .
.
├── dir1
│   └── x -> ../dir2
└── dir2
    └── y -> ../dir1
```

I've looked at what `tar` does, it has an option to dereference symlinks:
```
$ tar cv --dereference -f /tmp/test.tar .
```

[...]

PR #3848, opened 2022-08-02
Add follow options for symbolic links

cccapon:
In the Windows environment, the new 'follow' options allow symbolic links to be identified with a pattern.
When a symbolic link matches the pattern, restic will backup the link destination rather than the link itself.

The 'follow' options code is modelled after the 'exclude' options.

Under Windows, restic does not currently back up folders linked to by symbolic links.

This change allows users to select symbolic links by pattern matches (in an identical fashion with -exclude).  Any symbolic link encountered will be compared with the patterns.  If they match, then rather than looking at the symbolic link, restic will look at the underlying folder instead.

The default processing of restic does not change and unless these options are used, symbolic links will still be ignored.

#3594
https://github.com/restic/restic/issues/3594

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -466,13 +466,16 @@ Backblaze B2
 
 .. warning::
 
-   The recommended way to setup Backblaze B2 is by using its S3-compatible API.
+   Due to issues with error handling in the current B2 library that restic uses,
+   the recommended way to utilize Backblaze B2 is by using its S3-compatible API.
+   
    Follow the documentation to `generate S3-compatible access keys`_ and then
-   setup restic as described at :ref:`Amazon S3`.
+   setup restic as described at :ref:`Amazon S3`. This is expected to work better
+   than using the Backblaze B2 backend directly.
 
-    Different from the B2 backend, restic's S3 backend will only hide no longer
-    necessary files. Thus, make sure to setup lifecycle rules to eventually
-    delete hidden files.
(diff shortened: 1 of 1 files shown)


MichaelEischer:
Hmm, I'm not particularly convinced by that option (or rather the four of them). Having to precisely specify each symlink makes it quite easy to get it wrong.

How and why is the approach different from those in #542 and #2564? Are there other backup programs with a similar option to follow only specific symlinks?

cccapon:
First, let me say that this patch solves my problem.  For the first time, last night I completed backups of all the mounted file systems (yay!).

Initially, I thought of automatically following symbolic links only if they were top level folders of the backup.
restic backup /my/link
+- my
  +- link  -->  /somewhere/else

but my mount points for folders often look like this:
+- client
  +- abc  -->  /mnt/qfghtyzwflvisald
  +- def  -->  /mnt/bhgieosllkdhfas
and, if you want to start the backup from /client the symbolic links would be skipped again.

I worried about infinite loops
+-root
  +- Mine  -->  /root/Yours
  +- Yours  -->  /root/Mine

and links to unrelated places (not meant to be backed up)
+-home
  +- docs  -->  /my/giant/online/encyclopedia

so, that's where the idea of the follow pattern option came from.  It would let you pick and choose which links to include.

But now, after your comment, I'm wondering if a simple --follow-all-symbolic-links option is good enough.  Because you can still handle all of the above scenarios if you combine it with an --exclude pattern.
restic backup /home --follow-all-symbolic-links --exclude /home/docs

So, I have coded up another patch with just the single option for you to look at.
#3863

PR #3850, opened 2022-08-04
Update tests to Go 1.19

lbausch:

With Go 1.19 being released, I gave it a shot and updated GitHub Actions to test against Go 1.19

Not that I'm aware of.

Checklist
---------

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ on:
   pull_request:
 
 env:
-  latest_go: "1.18.x"
+  latest_go: "1.19.x"
   GO111MODULE: on
 
 jobs:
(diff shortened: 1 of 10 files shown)


lbausch:
- [ ] golangci/golangci-lint-action@v2 needs a new release to support v1.48
- [ ] some tests fail with 1.19, need to examine further

lbausch:
It seems like `TestApplyPolicy` fails because the sorting of snapshots differs when checking with `cmp.Equal(want.Keep, keep, cmpOpts)`.
The [1.19 release notes](https://go.dev/doc/go1.19#sort) mention that the sorting algorithm has been rewritten to use pattern-defeating quicksort and `ApplyPolicy` in `internal/restic/snapshot_policy.go` actually calls `sort.Sort(list)`. Might this be the reason and if so, would altering the testdata be a feasible approach?

MichaelEischer:
> The [1.19 release notes](https://go.dev/doc/go1.19#sort) mention that the sorting algorithm has been rewritten to use pattern-defeating quicksort and `ApplyPolicy` in `internal/restic/snapshot_policy.go` actually calls `sort.Sort(list)`. Might this be the reason and if so, what's a feasible approach here? Altering the testdata might possibly break the tests running against older Go versions.

That is exactly the problem. `sort.Sort` is not guaranteed to be stable according to the documentation, yet the tests relied on a specific order among snapshots with identical timestamps (sorting is exclusively based on the timestamp!). Thus the order changed. My suggestion (see the commit I've pushed) would be to switch to stable sorting for the snapshots, as I don't see a meaningful order for these snapshots. The stable order should remain consistent across Go version, so we won't run into problems there.

lbausch:
Great, thank you @MichaelEischer! Do we need a changelog entry for the new Go version?

MichaelEischer:
@lbausch I don't think we need a changelog entry. We're just adding support for an additional version, but not removing one.

MichaelEischer (approved):
LGTM. Thanks a lot! I already ran into the test failures with Go 1.19 a few times.

PR #3854, opened 2022-08-07
restore: Add support for sparse files

### LINKED ISSUE #79: archive and restore sparse files

### LINKED ISSUE #3903: restic to GCP hangs on large files
The issue is described in its full here https://github.com/vmware-tanzu/velero/issues/5251#issuecomment-1229141718

Please let me know if you need any additional data.

Restic version:

```
# /proc/<velero pid>/root/usr/bin/restic version
restic 0.13.1 compiled with go1.18 on linux/amd64
```

MichaelEischer:
It adds a `restore --sparse` option to restore files as sparse. The detection of sparseness is done at restore time, by replacing long stretches of zeros by sparse sections.

Note that this does not mean that we are no longer interested in having the archiver store precise information about which files have gaps at which places. Instead this is intended to provide basic support for sparse files which can be improved upon later.  In particular, if sparseness information was stored for a file, then it would also an option to have the restorer automatically restore such files as sparse.

The implementation is largely based on #2601, with a few modifications. The `--sparse` option was initially suggested by #2378. The restorer preallocates files by default which conflicts with restoring files as sparse. The currently taken approach is to check whether a file contains an all zeros blocks and mark is as sparse if the option was set. This then disables preallocation for that specific file. That way most files can still be preallocated. The downside is that we cannot detect even smaller sparse sections, however, these were only partially handled before.

Another change is that a sparse file is now immediately truncated to it's final size. `partialFile.WriteAt` may be called in parallel by the restorer, which could result in race conditions between the `Truncate` and `WriteAt` calls.

Supersedes #2378
Supersedes #2601
Fixes #79

Checklist
---------

### DIFF
--- a/internal/restorer/fileswriter.go
+++ b/internal/restorer/fileswriter.go
@@ -67,7 +67,7 @@ func (w *filesWriter) writeToFile(path string, blob []byte, offset int64, create
 
 		if createSize >= 0 {
 			if sparse {
-				err = f.Truncate(createSize)
+				err = truncateSparse(f, createSize)
 				if err != nil {
 					return nil, err
 				}
--- a/internal/restorer/fileswriter.go
+++ b/internal/restorer/fileswriter.go
@@ -19,15 +19,19 @@ type filesWriter struct {
 
 type filesWriterBucket struct {
 	lock  sync.Mutex
-	files map[string]*os.File
-	users map[string]int
+	files map[string]*partialFile
+}
+
+type partialFile struct {
+	*os.File
+	size  int64 // File size, tracked for sparse writes (not on Windows).
+	users int   // Reference count.
 }
 
 func newFilesWriter(count int) *filesWriter {
 	buckets := make([]filesWriterBucket, count)
 	for b := 0; b < count; b++ {
-		buckets[b].files = make(map[string]*os.File)
-		buckets[b].users = make(map[string]int)
+		buckets[b].files = make(map[string]*partialFile)
 	}
 	return &filesWriter{
 		buckets: buckets,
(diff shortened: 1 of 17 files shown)


internal/restorer/fileswriter.go
```diff
@@ -53,39 +57,45 @@ func (w *filesWriter) writeToFile(path string, blob []byte, offset int64, create
...
 		if createSize >= 0 {
-			err := preallocateFile(wr, createSize)
-			if err != nil {
-				// Just log the preallocate error but don't let it cause the restore process to fail.
-				// Preallocate might return an error if the filesystem (implementation) does not
-				// support preallocation or our parameters combination to the preallocate call
-				// This should yield a syscall.ENOTSUP error, but some other errors might also
-				// show up.
-				debug.Log("Failed to preallocate %v with size %v: %v", path, createSize, err)
+			if sparse {
```
DRON-666, line 69: You can also support Windows with just two lines of code (in separate file with `//go:build windows`, of course):
```
var t uint32
_ = windows.DeviceIoControl(windows.Handle(f.Fd()), windows.FSCTL_SET_SPARSE, nil, 0, nil, 0, &t, nil)
```

MichaelEischer:
I've added the optimization discussed in #3903 to speed up hashing of all-zero chunks. On windows, files are now also marked as sparse.

MichaelEischer:
LGTM. I've rebased the PR and fix the returned byte count of `partialFile.WriteAt` for a sparse write.

FATruden:
How about add `--sparse` option to `dump` command? I think It will be very useful especially when we speak about restore of VM disk images which is usually have a lot of sparse sections.

FATruden:
Oh, I realized that dump with `--sparse` is not possible, because restic just stream bytes to stdout and don't do write directly to file on filesystem. Hm.. But may be It possible don't read all zeros from repo and just "generate" them on the fly)
But it will require to store information about zeros in some place...

Issue #3674, opened 2022-03-23
rclone backup skips Windows junctions silently


cowwoc:
Output of `restic version`
--------------------------
restic 0.12.1 compiled with go1.16.6 on windows/amd64

How did you run restic exactly?
-------------------------------

restic backup --use-fs-snapshot -r b2:my-repo --files-from="include.txt" --iexclude-file="exclude.txt" 

where `include.txt` contains:

```
c:/test
```

and `exclude.txt` is empty, and c:\test contains windows junctions created using `mklink /d source target`.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------

b2

Expected behavior
-----------------

restic should back up all directories recursively

Actual behavior
---------------

restic backs up all normal sub-directories and silently skips over any windows junctions.

Steps to reproduce the behavior
-------------------------------

Back up any directory containing windows junctions created using `mklink /d source target`.

Do you have any idea what may have caused this?
-----------------------------------------------

No

Do you have an idea how to solve the issue?
-------------------------------------------

No

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

greatroar:
The title says rclone backup, but you clearly mean restic backup.

cowwoc:
@greatroar Good catch :) Fixed.

MichaelEischer:
According to the documentation `mklink /d` creates a symbolic link to a directory. restic _always_ backs up symbolic links as symbolic links and _never_ follows them.

What's a bit surprising is that directory junctions created using `mklink /j` are also considered to be symlinks.

MichaelEischer:
So essentially the question is whether directory junctions should be treated like mount points on Unix systems or not. At least Go has  decided that both symlink and junctions should be treated as symlinks. And judging from the fact that the junctions seem to behave somewhat similar to symlinks that appears to be a reasonable decision.

Dobatymo:
> And judging from the fact that the junctions seem to behave somewhat similar to symlinks that appears to be a reasonable decision.
The main difference is on network file systems. Junctions are resolved on the serverside, while symlink are resolved on the clientside.

cowwoc:
> restic always backs up symbolic links as symbolic links and never follows them.

Maybe I'm wrong, but I have a feeling that this behavior should be configurable. At least in my case, I'd like all symlinks to get backed up as symlinks but also for restic to follow them.

MichaelEischer:
[...]

Issue #2708, opened 2020-04-26
error: NodeFromFileInfo: Readlink: readlink \\?\C:\ProgramData\restic\shadow\C: The parameter is incorrect.


olarriga:
Hi,

Output of `restic version`
--------------------------
restic 0.9.6 compiled with go1.13.4 on windows/amd64
I use restic on Microsoft Windows 10 (1903).
I run restic on command line with administrator access.

How did you run restic exactly?
-------------------------------
restic -r D:\restic --verbose backup C:\ProgramData\restic\shadow\C

C:\restic>restic -r D:\restic -f --verbose backup C:\ProgramData\restic\shadow\C
open repository
enter password for repository:
repository cacd2f5e opened successfully, password is correct
lock repository
load index files
start scan on [C:\ProgramData\restic\shadow\C]
start backup on [C:\ProgramData\restic\shadow\C]
scan finished in 0.255s: 0 files, 0 B
error: NodeFromFileInfo: Readlink: readlink \\?\C:\ProgramData\restic\shadow\C: The parameter is incorrect.
 
Files:           0 new,     0 changed,     0 unmodified
Dirs:            4 new,     0 changed,     0 unmodified
Data Blobs:      0 new
Tree Blobs:      0 new
Added to the repo: 0 B
 
processed 0 files, 0 B in 0:00
snapshot 5d2a2994 saved

But, if I backup a subfolder it works: 

C:\restic>restic -r D:\restic --verbose backup C:\ProgramData\restic\shadow\C\tmp
open repository
enter password for repository:
repository cacd2f5e opened successfully, password is correct
lock repository
load index files
start scan on [C:\ProgramData\restic\shadow\C\tmp]
start backup on [C:\ProgramData\restic\shadow\C\tmp]
scan finished in 0.268s: 1 files, 213 B
 
Files:           1 new,     0 changed,     0 unmodified
Dirs:            5 new,     0 changed,     0 unmodified
Data Blobs:      1 new
Tree Blobs:      6 new
Added to the repo: 1.968 KiB
 
processed 1 files, 213 B in 0:00
snapshot aaf2aeaf saved

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Local

Expected behavior
-----------------
Backup of the files.

Actual behavior
---------------
Backup end with no file but snapshot created.

Steps to reproduce the behavior
-------------------------------
1/ Init a new repository
2/ Do a Shadow Copy of C:
[...]

Issue #2211, opened 2019-03-19
Follow symlinks


ruiztulio:
Output of `restic version`
--------------------------

restic 0.9.4 compiled with go1.11.4 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Reading the doc I found this:

> **Symlinks**  are archived as symlinks,  `restic`  does not follow them. When you restore, you get the same symlink again, with the same link target and the same timestamps.

Which is ok for most of the use cases (actually is desired y you're making a server backup or somethink like that), but it would be nice to have an option to overide this in some cases.

What are you trying to do?
--------------------------

My case: we use restic to "pack" the backups bcause in this way we can restore them in other location and run some tests against them, but we don't have the destinatino of such symlink, also this save us a lot of time.

ATM we copy the link content to a folder and then backup the "pack", but we are talking about 8 - 20G each time we do this and even when restic is very fast doing its job (takes about 2 minutes to complete the backup) the time it takes to copy the files is huge compared to this,

I know this is a very particular use case, but it would really help to have al option like: --override-and-follow-symlink (not an actual proposal)

Did restic help you or made you happy in any way?
-------------------------------------------------

Yes, is actually part of our new standard for backups. It is being integrated with our toolchain

**Note** I created an [entry](https://forum.restic.net/t/option-to-force-follow-symlink/1508) in the forum and cdhowie reminded me that tar have a similar option: 

![a](http://screenshots.vauxoo.com/tulio/14394907819-7OWHgjwE3m.jpg)

which is a nice precedent for this one.

Regards.

cdhowie:
I might take a stab at implementing this if @fd0 is open to the idea.

nicola-zen:
First of all, thank you for restic. It is a great tool.

Secondly, I would also appreciate this feature.
I have a folder were I maintain a structure to be backed up that has a bunch of symlinks to the folders where the actual data is stored. This just to have a folder hierarchy that makes maintaining and consuming the back up easier, as the value of the data is not tied to the location of it. For instance, I have backups of my Nextcloud instance per each user, and I am only interested in the data and not in the fact that it is stored in /foo/bar.

I currently worked around it by `mount --bind`ing all the folders, but it is quite impractical.

Sweenu:
I would also be very interested in that feature as I am using the https://github.com/Lobaro/restic-backup-docker image and it backs up everything from a same directory, mounted on /data.
[...]

PR #3863, opened 2022-08-10
Add option to follow symbolic links

cccapon:
A new flag '--follow-symlinks' was added to the backup command (default off).  When enabled, restic will dereference any symbolic link encountered to its destination file or folder instead.  This allows the destination of symbolic links to be backed up as if they were standard files or folders.

In order to prevent infinite loops or backup of unwanted symlinks, use the --exclude options as well.

Symbolic links are used for a variety of purposes in the Windows file system and sometimes it is necessary to archive the linked contents.

Under Windows, restic does not currently back up folders linked to by symbolic links.

This change causes restic to dereference symbolic links during backup and examine the underlying file or folder instead.  The default processing of restic does not change and unless this option is used, symbolic links will still be ignored.

Issues:
#542
#1078
#1215
#2211
#2564
#2699
#2708
#3674
#3594
#3848

Forum:
[4582](https://forum.restic.net/t/dokany-mounted-filesystem-inaccessible-in-windows/4582)
[1508](https://forum.restic.net/t/option-to-force-follow-symlink/1508)

Checklist
---------


MichaelEischer:
To be honest, I'm not at all convinced by that approach. It is all to easy to create infinite loops. Users shouldn't have to somehow determining beforehand which symlinks could lead to endless loops. But I also don't have time at the moment to think about what an alternative could look like.

cccapon:
@kerryland ,
The suggestion of building a list of visited destination folders is a good one.  But, I'm not getting anything back from GO for Readlink.  Without it, I don't think it will be possible.

Here is what I've found so far.

os.Lstat() .Mode()  --> Lrw-rw-rw- 
os.Stat() .Mode()  --> drwxrwxrwx
os.Readlink() --> error: The system cannot find the file specified.
filepath.EvalSymlinks() --> error: The system cannot find the file specified.

The basis of my patch is to use os.Stat() instead of os.Lstat() -- and then Restic works.
Without the link destination, though, I don't think loop tracking can be coded.

Do you know of any other way to find the destination path?  I'm not keen on launching a shell command to find out.

(fyi: my previous patch involved options for patterns which identify Symbolic links to descend into)

By the way, command:  dir /a  shows "<JUNCTION>" as the folder type and the mounted filesystem as
    \??\Volume{d6cc17c5-1734-4085-bca7-964f1e9f5de9}\
Presumably, I could backup that volume, but it would be tough to identify locations using a GUID.


@MichaelEischer ,
What ideas do you have for handling the infinite loop scenario?  I don't mind coding it, but I'd like some directions from you guys if possible.

MichaelEischer:
> I'm not keen on launching a shell command to find out.

Launching a shell command from within the archiver is completely out of question.

> What ideas do you have for handling the infinite loop scenario?

On Unix we could also use Inodes to detect duplicates. But that won't work on windows. If we included the symlink targets in the backup then we somehow have to recognize whether we've already visited a certain folder. The suggestion from kerryland would do that, but depends on being able to determine the symlink location.

I haven't given the matter much thought though, as I currently don't have time to come up with a concept on how to better handle symlinks without making a total mess out of things.

cccapon:
Well.  That's about it from me.

For those who follow, I provided two solutions:

1) Have the user identify which symbolic links are to be include in the backup.  This was not accepted because it was too complicated for users.

2) Flag all symbolic links to be backed up, then use 'exclude' to prevent infinite loops.  This was not accepted because it relys on the user to prevent infinite loops.

3) There doesn't appear to be a way to identify the destination of symbolic links in Windows.  Using GO lang 1.18.1 the functions os.Readlink() and filepath.EvalSymlinks() both fail on symbolic links of type "Junction" with "File not found".  Without knowing the destination of the link, the only other way I know to detect infinite loops would be to look for duplicates using directory contents.  That approach can not tell the difference between a loop and a duplicate folder.

@MichaelEischer If you come up with another solution, let me know and I will try to code it up.

For now, I'm at a stand-still.

kerryland:
I've never used GO before, which is why I've been conspicuously silent, but if these GO functions really don't work on Windows then surely that's a major GO bug that needs to be reported and fixed?

Anyway, I just had a quick play on Windows with 1.19.3 and didn't see any problem. Is it a bug specific to 1.18.1? Can't we upgrade?

Here's the awful code I just wrote/scavenged. 
```
package main

import "fmt"
import "os"
import "io/fs"
import "path/filepath"

func main() {
        pathName := os.Args[1]
	f, err := os.Lstat(pathName)
	if err != nil {
		fmt.Println(err)
	}
	switch mode := f.Mode(); {
	case mode.IsRegular():
		fmt.Println("regular file")
	case mode.IsDir():
		fmt.Println("directory")
	case mode&fs.ModeSymlink != 0:
		fmt.Println("symbolic link")
	case mode&fs.ModeNamedPipe != 0:
		fmt.Println("named pipe")
	}

        readLink(pathName);
}

func readLink(file string) error {
	path, err := os.Readlink(file)
	if err != nil {
		return err
	}

	path, err = filepath.EvalSymlinks(file)

	fmt.Printf("%s\n", path)
	return err
}
```

I created a directory "realdir", then a symlink "softlink", and a hardlink to the "realdir" directory.

The output looks like this:
```
D:\dev\golang\play> go run .\example\hello.go .\realdir
directory

D:\dev\golang\play> go run .\example\hello.go .\softlink\
symbolic link

D:\dev\golang\play> go run .\example\hello.go .\hardlink\
symbolic link
```

Dobatymo:
@MichaelEischer @cccapon getting the target of a junction (reparse point) on Windows is not fun. It's possible with the winapi https://stackoverflow.com/questions/46383428/get-the-immediate-target-path-from-symlink-reparse-point but I don't know if that's implemented in Go somewhere.

deajan:
I've done detection based on a regex that catches cloud file errors (reparse points) in my restic wrapper.
Can share the code.

cccapon:
@deajan I'd be interested to hear more.  How are you using regex in this context?

deajan:
@cccapon I'm wrapping restic in a Python script, the interesting part being the following:
```
        # run restic backup
        exit_code, stdout, stderr = command_runner(restic_command)

        if exit_code == 0:
            # Everything is okay, let's just return status and output from restic 
            return True, stdout
        elif exit_code == 3 and os.name == 'nt':
            # TEMP-FIX-4155, since we don't have reparse point support for Wiondows, see https://github.com/restic/restic/issues/4155, we have to filter manually for cloud errors which should not affect backup result
            # exit_code = 3 when errors are present but snapshot could be created
            is_reparse_point_error = True
            for line in stderr.split('\n'):
                if re.match('error', line, re.IGNORECASE):
                    if re.match('.*: The cloud operation is not supported on a read-only volume\.|.*: The media is write protected\.', line, re.IGNORECASE):
                        is_reparse_point_error = True
                    else:
                        is_reparse_point_error = False
            if is_reparse_point_error is True:
                return True, stdout
            # TEMP-FIX-4155-END
        return False, stderr
```

If interested, I've built a whole restic wrapper, including YAML config file, and a nice GUI that allows backup/restore operations as well as GUI config for MSWindows and Linux.
Indeed, the program is usually called via scheduled tasks /cron and will backup only if no recent snapshot is already present.
GUI snapshot follows:
image

cowwoc:
@cccapon I think you guys are conflating symbolic links with junctions. `mklink /d` creates symbolic links and is the default behavior. `mylink /j` creates a junction. https://superuser.com/a/343079 (and other answers in that thread) discusses the differences.

Also, https://go-review.googlesource.com/c/go/+/460595 might be of interest.

Let's start by adding support for symbolic links, and follow up with a separate issue for junctions.

cccapon:
For those who follow, I did find a way around this problem with the current restic.exe (v 0.16.0).  

In your backup script, change directories (cd in BAT, Push-Location in PowerShell) to the root directory of the symbolic link, then launch restic and backup the '.' (dot) current directory.  When restic's starting point is already inside the mount point (junction) it runs fine.  restic just can't cross the devision between one file system and the next.

MichaelEischer:
Problems with Windows reparse points are most likely related to https://github.com/restic/restic/issues/21789 . Following reparse points would be fine as those are just the Windows counterpart of bind-mounts. Fixing node attribute retrieval for reparse points could then also be solved as part of deliberately handling reparse points.

For following symlinks, I still think it's a bad idea with too many corner cases that we have to get correct. Endless loops are just the tip of the iceberg; there are also lots of security implications. At the very least we'd need a better plan first (I won't really have to review such a plan anytime soon). Anyways, the PR in it's current state would need a full rewrite to adapt to any of those approaches, so I'll close it for now.

PR #3899, opened 2022-08-28
Optimize prune memory usage

MichaelEischer:
It introduces a CountedBlobSet and uses it to track necessary, existing and duplicate blobs. This replaces the duplicate sets used to track whether all necessary blobs also exist. This reduces the memory usage of prune by about 20-30%. That set is also copied after removing blobs that are not modified by prune to reduce memory usage during the repack step.

The PR also contains an optimization for (Master)Index.Each to reduce the number of context switches.

No.

Checklist
---------

### DIFF
--- a/internal/restic/counted_blob_set.go
+++ b/internal/restic/counted_blob_set.go
@@ -1,0 +1,59 @@
+package restic
+
+import "sort"
+
+// CountedBlobSet is a set of blobs. For each blob it also stores a uint8 value
+// which can be used to track some information. The CountedBlobSet does not use
+// that value in any way. New entries are created with value 0.
+type CountedBlobSet map[BlobHandle]uint8
+
+// NewCountedBlobSet returns a new CountedBlobSet, populated with ids.
+func NewCountedBlobSet(handles ...BlobHandle) CountedBlobSet {
+	m := make(CountedBlobSet)
+	for _, h := range handles {
+		m[h] = 0
+	}
+
+	return m
+}
+
+// Has returns true iff id is contained in the set.
+func (s CountedBlobSet) Has(h BlobHandle) bool {
+	_, ok := s[h]
+	return ok
+}
+
+// Insert adds id to the set.
+func (s CountedBlobSet) Insert(h BlobHandle) {
+	s[h] = 0
+}
+
+// Delete removes id from the set.
+func (s CountedBlobSet) Delete(h BlobHandle) {
+	delete(s, h)
+}
+
+func (s CountedBlobSet) Len() int {
+	return len(s)
+}
+
+// List returns a sorted slice of all BlobHandle in the set.
+func (s CountedBlobSet) List() BlobHandles {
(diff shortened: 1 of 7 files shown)


internal/restic/counted_blob_set.go
```diff
@@ -0,0 +1,66 @@
...
+	if len(str) < 2 {
+		return "{}"
+	}
+
+	return "{" + str[1:len(str)-1] + "}"
+}
+
+// Copy returns a copy of the CountedBlobSet.
+func (s CountedBlobSet) Copy() CountedBlobSet {
+	cp := NewCountedBlobSet()
```
greatroar, line 61: I didn't test this, but I think a size hint should produce an even smaller map:

```suggestion
	cp := make(CountedBlobSet, len(s))
```

internal/restic/counted_blob_set.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Has returns true iff id is contained in the set.
+func (s CountedBlobSet) Has(h BlobHandle) bool {
+	_, ok := s[h]
+	return ok
+}
+
+// Insert adds id to the set.
+func (s CountedBlobSet) Insert(h BlobHandle) {
+	s[h] = 0
```
greatroar, line 26: Shouldn't this be
```suggestion
	s[h] = 1
```

MichaelEischer: No. `0` is correct here. The check, whether a blob is part of the set does not use the value. In fact the value stored in the CountedBlobSet is not for use by the `CountedBlobSet`. Instead it is exclusively used to count how many times a blob is contained in the index, which happens here https://github.com/restic/restic/pull/3899/files#diff-33a0be5385e73907c2b87dc06a9ebc1822c4abe53738ed482dd152e26c86cac5R308 .

internal/restic/counted_blob_set.go
MichaelEischer: The bucket array apparently always uses a power of 2 as size. But providing a hint should reduce the necessary reallocations. In particular we shouldn't end up with a half complete map growth operations that has to keep the old and new bucket array around.

https://github.com/golang/go/blob/4e3d805a015153ca4e3382415c2474e4f954ddee/src/runtime/map.go#L318

MichaelEischer:
I've rebased the PR, added a few more comments and made the sanity check at the end of `packInfoFromIndex` a bit stricter.

MichaelEischer:
LGTM.

molecular:
For what version of restic is this expected? I run 0.11.0 and I think it doesn't have this improvement, right?

molecular:
@MichaelEischer, yeah, thanks. I wonder why the package isn't being updated in raspbian bullseye.

MichaelEischer:
> I wonder why the package isn't being updated in raspbian bullseye. Just ran a full system update, still version 0.11.

Debian (and also raspbian) usually only switch to a newer software version in conjunction with a new Debian release.

PR #3912, opened 2022-09-03
Cleanup snapshot filter options

MichaelEischer:
This PR cleans up the help text for the `--host`, `--tag` and `--path` options used to filter snapshots. It explicitly does not cleanup up the snapshot filter code itself, that has to wait until some later PR (I've started some experiments in that direction in https://github.com/MichaelEischer/restic/tree/layered-repo2).

The commands `backup`, `cat` and `diff` are not affected by this PR as these use a special way to resolve the snapshot (backup) or just process a single specific snapshot ID (i.e. `latest` is not allowed).

The remaining commands be grouped into 3 categories:
- Using `FindFilteredSnapshots`: `copy`, `find`, `forget`, `snapshots`, `stats`, `tag`. The filtering options apply either when no snapshots are specified or when `latest` is given.
- Using `restic.FindFilteredSnapshots`: `mount`. The filtering options are used to limit the shown snapshots.
- Others: `ls` (`FindFilteredSnapshots` but limited to exactly a single snapshot), `dump`, `restore` (`FindSnapshot`/`FindLatestSnapshot`). The filtering options apply when `latest` is given.

The previous state of the help texts is as shown below:

```
copy     "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
find     "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
forget   "host",      nil, "only consider snapshots for this `host` (can be specified multiple times)")
snapshot "host", "H", nil, "only consider snapshots for this `host` (can be specified multiple times)")
stats    "host", "H", nil, "only consider snapshots for this `host` (can be specified multiple times)")
tag      "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")

mount    "host", "H", nil, "only consider snapshots for this `host` (can be specified multiple times)")

ls       "host", "H", nil, "only consider snapshots for this `host`, when snapshot ID \"latest\" is given (can be specified multiple times)")
dump     "host", "H", nil, `only consider snapshots for this host when the snapshot ID is "latest" (can be specified multiple times)`)
restore  "host", "H", nil, "only consider snapshots for this `host`, when snapshot ID \"latest\" is given (can be specified multiple times)")
```
I've normalized this to `only consider snapshots for this 'host' (can be specified multiple times)` (for the first two categories) and `only consider snapshots for this 'host', when snapshot ID \"latest\" is given (can be specified multiple times)` (for the last category). The claim that the host filter only applies when no snapshot ID is given to `copy`, `find` and `tag` is wrong. We could try to explain the "no snapshots"/"latest" cases though, but that would result in a rather long description.

```
copy     "tag", "only consider snapshots which include this `taglist`, when no snapshot ID is given")
find     "tag", "only consider snapshots which include this `taglist`, when no snapshot-ID is given")
forget   "tag", "only consider snapshots which include this `taglist` in the format `tag[,tag,...]` (can be specified multiple times)")
snapshot "tag", "only consider snapshots which include this `taglist` in the format `tag[,tag,...]` (can be specified multiple times)")
stats    "tag", "only consider snapshots which include this `taglist` in the format `tag[,tag,...]` (can be specified multiple times)")
tag      "tag", "only consider snapshots which include this `taglist`, when no snapshot-ID is given")

mount    "tag", "only consider snapshots which include this `taglist`")

ls       "tag", "only consider snapshots which include this `taglist`, when snapshot ID \"latest\" is given (can be specified multiple times)")
dump     "tag", "only consider snapshots which include this `taglist` for snapshot ID \"latest\"")
restore  "tag", "only consider snapshots which include this `taglist` for snapshot ID \"latest\"")
```
Is now `only consider snapshots which include this 'taglist' in the format 'tag[,tag,...]' (can be specified multiple times)` and `only consider snapshots which include this 'taglist', when snapshot ID \"latest\" is given (can be specified multiple times)`. The reasoning is similar as above. For `dump` and `restore` the explanation was missing that the option can be specified multiple times.

```
copy     "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot ID is given")
find     "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")
forget   "path", nil, "only consider snapshots which include this (absolute) `path` (can be specified multiple times)")
snapshot "path", nil, "only consider snapshots for this `path` (can be specified multiple times)")
stats    "path", nil, "only consider snapshots which include this (absolute) `path` (can be specified multiple times)")
tag      "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot-ID is given")

mount    "path", nil, "only consider snapshots which include this (absolute) `path`")

ls       "path", nil, "only consider snapshots which include this (absolute) `path`, when snapshot ID \"latest\" is given (can be specified multiple times)")
dump     "path", nil, "only consider snapshots which include this (absolute) `path` for snapshot ID \"latest\"")
restore  "path", nil, "only consider snapshots which include this (absolute) `path` for snapshot ID \"latest\"")
```
And finally `only consider snapshots which include this (absolute) 'path' (can be specified multiple times)` and `only consider snapshots which include this (absolute) 'path', when snapshot ID \"latest\" is given (can be specified multiple times)`. Now all commands ask for an absolute path, although this is (currently) not strictly accurate, it can avoid problems. Depending on the snapshot filtering method relative paths are processed different leading to unexpected behavior. For example filtering snapshots by paths won't work between Windows and Unix-like systems. Thus, I plan to drop the handling of relative paths completely.

Fixes https://forum.restic.net/t/restic-0-14-flags-syntax/5387

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -51,9 +51,9 @@ func init() {
 
 	f := cmdCopy.Flags()
 	initSecondaryRepoOptions(f, &copyOptions.secondaryRepoOptions, "destination", "to copy snapshots from")
-	f.StringArrayVarP(&copyOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host`, when no snapshot ID is given (can be specified multiple times)")
-	f.Var(&copyOptions.Tags, "tag", "only consider snapshots which include this `taglist`, when no snapshot ID is given")
-	f.StringArrayVar(&copyOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path`, when no snapshot ID is given")
+	f.StringArrayVarP(&copyOptions.Hosts, "host", "H", nil, "only consider snapshots for this `host` (can be specified multiple times)")
+	f.Var(&copyOptions.Tags, "tag", "only consider snapshots which include this `taglist` in the format `tag[,tag,...]` (can be specified multiple times)")
+	f.StringArrayVar(&copyOptions.Paths, "path", nil, "only consider snapshots which include this (absolute) `path` (can be specified multiple times)")
 }
 
 func runCopy(opts CopyOptions, gopts GlobalOptions, args []string) error {
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -39,9 +39,7 @@ new destination repository using the "init" command.
 // CopyOptions bundles all options for the copy command.
 type CopyOptions struct {
 	secondaryRepoOptions
-	Hosts []string
-	Tags  restic.TagLists
-	Paths []string
+	snapshotFilterOptions
 }
 
 var copyOptions CopyOptions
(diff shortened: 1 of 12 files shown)


MichaelEischer:
Credit for the latest description tweaks belongs to @rawtaz .

rawtaz (approved):
LGTM!

PR #3915, opened 2022-09-05
restic stats: print uncompressed size in mode raw-data

plumbeo:

This PR prints the uncompressed size of a snapshot/repository when ```restic stats``` is called in mode ```raw-data```. For example:

```
$ restic stats --mode=raw-data -r repotest/
repository ad894609 opened (repository version 2) successfully, password is correct
scanning...
Stats in raw-data mode:
    Snapshots processed:  41
       Total Blob Count:  150677
             Total Size:  71.085 GiB
Total Uncompressed Size:  77.307 GiB
```

No

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3915
+++ b/changelog/unreleased/pull-3915
@@ -1,0 +1,9 @@
+Enhancement: add compression statistics to the restic stats command
+
+When executed in `raw-data` mode on a repository that supports compression, the `restic stats`
+command now calculates and displays, for the selected repository or snapshots, the uncompressed
+size of the data, the compression progress (percentage of data that has been compressed), the
+compression ratio of the compressed data and the total space saving, taking into account both
+the compressed and uncompressed data if the repository is only partially compressed.
+
+https://github.com/restic/restic/pull/3915
--- a/cmd/restic/cmd_stats.go
+++ b/cmd/restic/cmd_stats.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/restic/restic/internal/backend"
+	"github.com/restic/restic/internal/crypto"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/walker"
 
@@ -136,11 +137,7 @@ func runStats(gopts GlobalOptions, args []string) error {
 			}
 			stats.TotalSize += uint64(pbs[0].Length)
 			if repo.Config().Version >= 2 {
-				if pbs[0].IsCompressed() {
-					stats.TotalUncompressedSize += uint64(pbs[0].UncompressedLength)
-				} else {
-					stats.TotalUncompressedSize += uint64(pbs[0].Length)
-				}
+				stats.TotalUncompressedSize += uint64(crypto.CiphertextLength(int(pbs[0].DataLength())))
 			}
 			stats.TotalBlobCount++
 		}
@@ -293,7 +290,7 @@ func verifyStatsInput(gopts GlobalOptions, args []string) error {
 // for a successful and efficient walk.
 type statsContainer struct {
 	TotalSize             uint64 `json:"total_size"`
-	TotalUncompressedSize uint64 `json:"total_uncompressed_size"`
+	TotalUncompressedSize uint64 `json:"total_uncompressed_size,omitempty"`
 	TotalFileCount        uint64 `json:"total_file_count"`
 	TotalBlobCount        uint64 `json:"total_blob_count,omitempty"`
 	// holds count of all considered snapshots
--- a/cmd/restic/cmd_stats.go
+++ b/cmd/restic/cmd_stats.go
@@ -135,6 +135,13 @@ func runStats(gopts GlobalOptions, args []string) error {
 				return fmt.Errorf("blob %v not found", blobHandle)
 			}
 			stats.TotalSize += uint64(pbs[0].Length)
+			if repo.Config().Version >= 2 {
+				if pbs[0].IsCompressed() {
+					stats.TotalUncompressedSize += uint64(pbs[0].UncompressedLength)
+				} else {
+					stats.TotalUncompressedSize += uint64(pbs[0].Length)
+				}
+			}
 			stats.TotalBlobCount++
 		}
 	}
@@ -148,15 +155,18 @@ func runStats(gopts GlobalOptions, args []string) error {
 	}
 
 	Printf("Stats in %s mode:\n", statsOptions.countMode)
-	Printf("Snapshots processed:   %d\n", stats.SnapshotsCount)
+	Printf("    Snapshots processed:  %d\n", stats.SnapshotsCount)
 
 	if stats.TotalBlobCount > 0 {
-		Printf("   Total Blob Count:   %d\n", stats.TotalBlobCount)
+		Printf("       Total Blob Count:  %d\n", stats.TotalBlobCount)
 	}
 	if stats.TotalFileCount > 0 {
-		Printf("   Total File Count:   %d\n", stats.TotalFileCount)
+		Printf("       Total File Count:  %d\n", stats.TotalFileCount)
 	}
-	Printf("         Total Size:   %-5s\n", formatBytes(stats.TotalSize))
+	if stats.TotalUncompressedSize > 0 {
+		Printf("Total Uncompressed Size:  %-5s\n", formatBytes(stats.TotalUncompressedSize))
+	}
+	Printf("             Total Size:  %-5s\n", formatBytes(stats.TotalSize))
 
 	return nil
 }
@@ -282,9 +292,10 @@ func verifyStatsInput(gopts GlobalOptions, args []string) error {
 // to collect information about it, as well as state needed
 // for a successful and efficient walk.
 type statsContainer struct {
-	TotalSize      uint64 `json:"total_size"`
-	TotalFileCount uint64 `json:"total_file_count"`
-	TotalBlobCount uint64 `json:"total_blob_count,omitempty"`
+	TotalSize             uint64 `json:"total_size"`
+	TotalUncompressedSize uint64 `json:"total_uncompressed_size"`
+	TotalFileCount        uint64 `json:"total_file_count"`
+	TotalBlobCount        uint64 `json:"total_blob_count,omitempty"`
 	// holds count of all considered snapshots
 	SnapshotsCount int `json:"snapshots_count"`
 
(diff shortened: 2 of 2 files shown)


cmd/restic/cmd_stats.go
```diff
@@ -282,9 +292,10 @@ func verifyStatsInput(gopts GlobalOptions, args []string) error {
 // to collect information about it, as well as state needed
 // for a successful and efficient walk.
 type statsContainer struct {
-	TotalSize      uint64 `json:"total_size"`
-	TotalFileCount uint64 `json:"total_file_count"`
-	TotalBlobCount uint64 `json:"total_blob_count,omitempty"`
+	TotalSize             uint64 `json:"total_size"`
+	TotalUncompressedSize uint64 `json:"total_uncompressed_size"`
```
MichaelEischer, line 296: Shouldn't this also be marked as `omitempty`?

cmd/restic/cmd_stats.go
```diff
@@ -135,6 +135,13 @@ func runStats(gopts GlobalOptions, args []string) error {
 				return fmt.Errorf("blob %v not found", blobHandle)
 			}
 			stats.TotalSize += uint64(pbs[0].Length)
+			if repo.Config().Version >= 2 {
+				if pbs[0].IsCompressed() {
+					stats.TotalUncompressedSize += uint64(pbs[0].UncompressedLength)
+				} else {
+					stats.TotalUncompressedSize += uint64(pbs[0].Length)
```
MichaelEischer, line 142: This counts two slightly different things for compressed and uncompressed blobs:
For compressed blobs, the uncompressed length is the length of a blob after decrypting and uncompressing it. For uncompressed blobs it is the size of the encrypted blobs including the encryption overhead. It would probably make more sense to count the size of the decrypted blob. For that you can replace both cases with `pbs[0].DataLength()`.

MichaelEischer (changes requested):
The PR could also use a  changelog entry once we've finished deciding on what exactly to count for `TotalUncompressedSize`.

cmd/restic/cmd_stats.go
plumbeo: Oh, I didn't notice that. My idea was to show the space the repository would have occupied had compression not been enabled though, so the value should take into account the encryption overhead, and I'd like the numbers to stay the same if a repository is migrated to version 2 but not pruned, so the calculation shouldn't change for uncompressed blobs. What about using ```crypto.CiphertextLength(int(pbs[0].UncompressedLength))``` like in the commit I just pushed?

cmd/restic/cmd_stats.go
plumbeo: I think it should, fixed, thanks.

cmd/restic/cmd_stats.go
MichaelEischer: You mean `crypto.CiphertextLength(int(pbs[0].Length))`? In that case replacing both branches with `crypto.CiphertextLength(int(pbs[0].DataLength()))` would be simpler.

So essentially that definition of uncompressed size allows you to compute the compression ratio, while also accounting for overheads due to encryption etc.

plumbeo:
When I was writing a changelog entry I realised that printing the number of compressed blobs could be useful too, especially in cases where the repository has been migrated from version 1 to version 2 but not yet repacked to compress the old data or to track the progress if it's being compressed in several steps.

Unconditionally printing the number of compressed blobs clutters the output a little though, because in some cases now and most cases in the future (since compression is default on for new repositories) the number of compressed blobs is the same as the total number of blobs, so I added a condition that skips it when that's the case. We should probably discuss it though.

MichaelEischer:
Wouldn't the size of the remaining uncompressed blobs be more useful to judge how much work is left to fully compress a repository? E.g. `prune` expects a size limit for repacking, not a blob count.

plumbeo:
I considered that but since the compressed size of a pack is generally unrelated to the uncompressed size the ratio of compressed to uncompressed size didn't feel as informative as the ratio of the number of compressed blobs to the total number of blobs. Also "total size" and "total uncompressed size" can already be a little confusing and adding "total size of uncompressed blobs" didn't really make it any better.

Maybe it could be better to just add a new mode where all the compression related statistics may go, because there are a lot of things that could be useful to a user who specifically cares about them but not to everybody. I'm thinking about:
* total size: the actual size of the repository/snapshot
* total uncompressed size: the size of the repository/snapshot if compression wasn't enabled
* global compression gain: computed from the previous values, the gain from compression with the current mix of compressed/uncompressed blobs
* total number of blobs: self-explanatory
* total number of uncompressed blobs: same
* total number of compressed blobs: same, with the previous two it allows to track the compression state of a repository, if it wasn't compressed from the beginning
* total size of uncompressed blobs: the space occupied by uncompressed blobs
* total size of compressed blobs: the space occupied by compressed blobs
* total uncompressed size of compressed blobs: the space that would be occupied by compressed blobs if they weren't compressed
* compression ratio: computed from the previous two, the "real" compression ratio of compressed data

And some of these only make sense if a repository/snapshot is only partially compressed. Also better names would be probably needed.

MichaelEischer:
> didn't feel as informative as the ratio of the number of compressed blobs to the total number of blobs.

But what does that ratio tell me? As the blob sizes can vary wildly, it's a bit random whether that ratio is representative or not. My feeling here is that the information a user is interested in, is rather how much data still has to be compressed. Or which fraction of the overall data size has already been compressed (size of still uncompressed blobs / total uncompressed size). Although that information might be more useful in the prune command.

Btw, the repack size given to `prune` is before compression. That is just summing up all still uncompressed blob sizes would be what's necessary there.

> Maybe it could be better to just add a new mode where all the compression related statistics may go, because there are a lot of things that could be useful to a user who specifically cares about them but not to everybody.

I'd prefer to keep the number of different statistics somewhat limited. After all each new value means more code to maintain.

> And some of these only make sense if a repository/snapshot is only partially compressed.

A partially compressed repository is intended to be a transitional state, thus we shouldn't add too many statistics which are only useful for such repositories.


Judging from the discussion in https://forum.restic.net/t/how-to-check-if-files-were-compressed/5392 `TotalUncompressedSize` is the most asked for information, in particular as it allows to calculate the compression ratio. 

However, I'm still not convinced by `TotalCompressedBlobCount`. The longer I think about it, the more I'm convinced that it's only use would be to more or less guess how far the process of compressing the repository is already. So, what about replacing it with `CompressionProgress` (or some other name) which is defined as `1 - "total size of not compressed blobs" / "total uncompressed blob size"`. Then `stat` can just tell a user that the compression process is e.g. 40% done. That number would also allow extrapolating the size reduction to expect from compression: `(1 - totalCompressedSize/totalUncompressedSize)*(1/CompressionProgress)`.

plumbeo:
Yeah, it's better to only display what most users will want to see. 

I just pushed some new commits that calculate and display the compression progress, calculated as percentage of data in the repository that has been compressed. I also added the compression ratio, which is not easily obtained because it must be calculated on the actual compressed data and we were not providing the necessary information, and the space saving, calculated on the actual state of the repository. The latter is just the ratio of uncompressed data and compressed data but we might as well print it and save the users some time.

plumbeo:
This is the output on a partially compressed test repository:
```
$ restic stats --mode=raw-data -r test/ 
repository 7571f6c2 opened (repository version 2) successfully, password is correct
scanning...
Stats in raw-data mode:
     Snapshots processed:  3
        Total Blob Count:  136
 Total Uncompressed Size:  166.959 MiB
              Total Size:  158.648 MiB
    Compression Progress:  48.15%
       Compression Ratio:  1.12x
Compression Space Saving:  4.98%
```

changelog/unreleased/pull-3915
```diff
@@ -0,0 +1,9 @@
+Enhancement: add compression statistics to the restic stats command
+
+When executed in ```raw-data``` mode on a repository that supports compression, the ```restic stats```
```
MichaelEischer, line 3: These should be single backticks.

cmd/restic/cmd_stats.go
```diff
@@ -282,9 +308,15 @@ func verifyStatsInput(gopts GlobalOptions, args []string) error {
 // to collect information about it, as well as state needed
 // for a successful and efficient walk.
 type statsContainer struct {
-	TotalSize      uint64 `json:"total_size"`
-	TotalFileCount uint64 `json:"total_file_count"`
-	TotalBlobCount uint64 `json:"total_blob_count,omitempty"`
+	TotalSize                            uint64  `json:"total_size"`
+	TotalUncompressedSize                uint64  `json:"total_uncompressed_size,omitempty"`
+	TotalCompressedBlobsSize             uint64  `json:"total_compressed_blobs_size,omitempty"`
+	TotalCompressedBlobsUncompressedSize uint64  `json:"total_compressed_blobs_uncompressed_size,omitempty"`
```
MichaelEischer, line 314: The JSON output shouldn't contain more fields than the text output, please just exclude `TotalCompressedBlobsSize` and `TotalCompressedBlobsUncompressedSize` from the json output.

cmd/restic/cmd_stats.go
```diff
@@ -282,9 +308,15 @@ func verifyStatsInput(gopts GlobalOptions, args []string) error {
...
+	CompressionRatio                     float64 `json:"compression_ratio,omitempty"`
+	CompressionProgress                  float64 `json:"compression_progress,omitempty"`
+	CompressionSpaceSaving               float64 `json:"compression_space_saving,omitempty"`
```
MichaelEischer, line 317: `CompressionSpaceSaving` is somewhat redundant. As the JSON is intended for consumption by scripts, we can just let those calculate it themselves. My suggestion would be to remove the field from the struct and only calculate it on demand for the text output.

MichaelEischer (changes requested):
The information provided to the user looks fine by now. However, I'd like to reduce the amount of new JSON fields a bit (see comments). Adding 6 new fields with varying degrees of redundancy is too much.

changelog/unreleased/pull-3915
plumbeo: Done

cmd/restic/cmd_stats.go
plumbeo: Done.

cmd/restic/cmd_stats.go
plumbeo: I'm not sure about this, personally I'd expect the JSON output to include everything that's available in the text output so it feels more natural to have the compression space saving too.

I can certainly remove it if you feel strongly about it.

MichaelEischer: Nah, I see the point. It's just that the PR started with adding 1 or 2 new fields, and now we have 4...

MichaelEischer (approved):
LGTM. Thanks for keeping up with all the review comments!

PR #3923, opened 2022-09-10
cache: fix flaky TestFileSaveConcurrent on windows

MichaelEischer:
The test would from time to time fail with ERROR_SHARING_VIOLATION, see for example https://github.com/restic/restic/runs/8286388017?check_suite_focus=true#step:7:34 . The underlying reason is that when caching a file, the cache creates a temporary file, downloads the data and renames the temp file to its final name. This rename step on windows implicitly opens a file handle with a shareMode that includes FILE_SHARE_DELETE (https://devblogs.microsoft.com/oldnewthing/20211022-00/?p=105822). However, when opening a file, go only requests READ/WRITE sharing which results in the above error (https://github.com/golang/go/blob/54182ff54a687272dd7632c3a963e036ce03cb7c/src/syscall/syscall_windows.go#L332).

Strictly speaking the proper fix would be to open the cache file with a shareMode which includes FILE_SHARE_DELETE. But I'm not particularly eager to implement our own `os.Open` function.

No.

Checklist
---------

### DIFF
--- a/internal/cache/file_test.go
+++ b/internal/cache/file_test.go
@@ -207,10 +207,13 @@ func TestFileLoad(t *testing.T) {
 // Simulate multiple processes writing to a cache, using goroutines.
 //
 // The possibility of sharing a cache between multiple concurrent restic
-// processes isn't guaranteed in the docs and doesn't always work on Windows
-// due to the Go runtime opening files without FILE_SHARE_DELETE, hence the
-// check on GOOS. This is considered a "nice to have" on POSIX, for now.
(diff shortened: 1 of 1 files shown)


greatroar:
This one's my fault, since I promised that this would work in the changelog for #2838. LGTM in terms of code, but please consider the implicit call in #3934 for a better Windows programmer as an alternative "solution".

MichaelEischer:
@greatroar I agree, it's probably better to just disable the test on windows for now. I've taken the commit from #3934 and extended the description slightly.

PR #3931, opened 2022-09-14
add --no-scan to backup command

kjetilho:
The scanner process has only cosmetic effect for the progress printer, and can be disabled without impacting functionality when the user does not need an estimate of completion.

In many cases the scanner process can provide beneficial priming of the file system cache, so as general advice it should not be disabled. However, tests have shown that backup of NFS and fuse based filesystems, where stat(2) is relatively expensive, can be significantly faster without the scanner.

This patch allows users to avoid some unnecessary I/O during backup caused by the job size scanner when appropriate.

Michael Eischer previously suggest to disable the scanner implicitly for quiet jobs in PR #2336

His testing showed that although user and system time was reduced, real time could in fact increase.  Also, the difference was relatively small when run locally on SSD.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -99,6 +99,7 @@ type BackupOptions struct {
 	UseFsSnapshot     bool
 	DryRun            bool
 	ReadConcurrency   uint
+	NoScan            bool
 }
 
 var backupOptions BackupOptions
@@ -138,6 +139,7 @@ func init() {
 	f.BoolVar(&backupOptions.IgnoreInode, "ignore-inode", false, "ignore inode number changes when checking for modified files")
 	f.BoolVar(&backupOptions.IgnoreCtime, "ignore-ctime", false, "ignore ctime changes when checking for modified files")
 	f.BoolVarP(&backupOptions.DryRun, "dry-run", "n", false, "do not upload or write any data, just show what would be done")
+	f.BoolVar(&backupOptions.NoScan, "no-scan", false, "do not run scanner to estimate size of backup")
 	if runtime.GOOS == "windows" {
 		f.BoolVar(&backupOptions.UseFsSnapshot, "use-fs-snapshot", false, "use filesystem snapshot where possible (currently only Windows VSS)")
 	}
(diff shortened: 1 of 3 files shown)


doc/040_backup.rst
```diff
@@ -204,6 +204,16 @@ Combined with ``--verbose``, you can see a list of changes:
     modified  /archive.tar.gz, saved in 0.140s (25.542 MiB added)
     Would be added to the repository: 25.551 MiB
 
+Estimator Scanner
```
MichaelEischer, line 207: What about `Disabling Backup Progress Estimation`? I couldn't guess from the section title what it was about.

doc/040_backup.rst
```diff
@@ -204,6 +204,16 @@ Combined with ``--verbose``, you can see a list of changes:
...
+*****************
+
+When you start a backup, Restic will launch a separate thread to count
```
MichaelEischer, line 210: I'd like to avoid mentioning "thread" here as its just an implementation detail, just saying that e.g. "[...], restic will concurrently to the backup count the number of files and their total size, [...]" should be enough.

cmd/restic/cmd_backup.go
```diff
@@ -141,6 +142,7 @@ func init() {
 	f.BoolVar(&backupOptions.IgnoreInode, "ignore-inode", false, "ignore inode number changes when checking for modified files")
 	f.BoolVar(&backupOptions.IgnoreCtime, "ignore-ctime", false, "ignore ctime changes when checking for modified files")
 	f.BoolVarP(&backupOptions.DryRun, "dry-run", "n", false, "do not upload or write any data, just show what would be done")
+	f.BoolVar(&backupOptions.DisableScanner, "disable-scanner", false, "do not run scanner to estimate size of backup")
```
MichaelEischer, line 145: After discussing with @rawtaz, we think that `--no-scan` would be more in line with the other options that disable something (e.g. `--no-cache`, `--no-lock`). Please rename that option and the attribute in the `BackupOptions` struct.

MichaelEischer (changes requested):
Sorry for the late feedback. The change in general looks fine, I've just added a few comments regarding how to name things.

kjetilho:
thank you for your comments, they all make sense.  I did the rename as a separate patch just to make the process clear, but obviously these patches are best squashed during merge.

MichaelEischer (approved):
LGTM. I've squashed the commits.

PR #3938, opened 2022-09-20
rclone/sftp: Improve handling of ErrDot errors

### LINKED ISSUE #3932: Restic 0.14.0 on Windows with rclone to OneDrive don't work
Output of `restic version`
--------------------------
restic 0.14.0 compiled with go1.19 on windows/amd64

restic_0.14.0_windows_amd64.exe backup \\server03\d$\ -r rclone:OneDrive_XXXXXX:Backup --exclude-file=C:\AdminScripte\Restic\exclude.txt -p=C:\AdminScripte\Restic\password.txt

Microsoft-OneDrive with rclone. rclone version info:
rclone v1.58.1
- os/version: Microsoft Windows 10 Pro 21H2 (64 bit)
- os/kernel: 10.0.19044.2006 (x86_64)
- os/type: windows
- os/arch: amd64
- go/version: go1.17.9
- go/linking: dynamic
- go/tags: cmount

Both executable (restic/rclone) are in the same directory.

Expected behavior
-----------------
Doing a backup like with restic_0.13.1_windows_amd64

Actual behavior
---------------
Error-Message:
Fatal: unable to open repository at rclone:OneDrive_XXXXXX:Backup: cmd.Start: exec: "rclone": cannot run executable found relative to current directory

Use commandline above

No

No

Yes, I make my backup daily with restic-0.13.1 and would like to use restic-0.14.0 with compression.

rawtaz:

Restic now yields a more informative error message when exec.ErrDot occurs, informing the user of what they can do if they truly want to run the executable that was found in the current directory.

Closes #3932.

Checklist
---------

### DIFF
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -85,6 +85,9 @@ func run(command string, args ...string) (*StdioConn, *sync.WaitGroup, func() er
 		err = errW
 	}
 	if err != nil {
+		if backend.IsErrDot(err) {
+			return nil, nil, nil, errors.Errorf("cannot implicitly run relative executable %v found in current directory, use -o rclone.program=./<program> to override", cmd.Path)
+		}
 		return nil, nil, nil, err
 	}
 
--- a/changelog/unreleased/issue-3932
+++ b/changelog/unreleased/issue-3932
@@ -1,0 +1,15 @@
+Enhancement: Improve handling of ErrDot errors in rclone and sftp backends
+
+Since Go 1.19, restic can no longer implicitly run relative executables which
+are found in the current directory (e.g. `rclone` if it's found in `.`). This
+is a security feature of Go to prevent against running unintended and possibly
+harmful executables.
+
+The error message for this was just "cannot run executable found relative to
+current directory". This has now been improved to yield a more specific error
+message, informing the user how to explicitly allow running the executable
+using the `-o rclone.program` and `-o sftp.command` extended options with `./`.
+
+https://github.com/restic/restic/issues/3932
+https://pkg.go.dev/os/exec#hdr-Executables_in_the_current_directory
+https://go.dev/blog/path-security
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -633,6 +633,13 @@ initiate a new repository in the path ``bar`` in the remote ``foo``:
 
 Restic takes care of starting and stopping rclone.
 
+.. note:: If you get an error message saying "cannot implicitly run relative
+          executable rclone found in current directory", this means that an
+          rclone executable was found in the current directory. For security
+          reasons restic will not run this implicitly, instead you have to
+          use the ``-o rclone.program=./rclone`` extended option to override
+          this security check and explicitly tell restic to use the executable.
+
 As a more concrete example, suppose you have configured a remote named
 ``b2prod`` for Backblaze B2 with rclone, with a bucket called ``yggdrasil``.
 You can then use rclone to list files in the bucket like this:
--- a/internal/backend/errdot_119.go
+++ b/internal/backend/errdot_119.go
@@ -1,0 +1,20 @@
+//go:build go1.19
+// +build go1.19
+
+// This file provides a function to check whether an error from cmd.Start() is
+//  exec.ErrDot which was introduced in Go 1.19.
+// This function is needed so that we can perform this check only for Go 1.19 and
+//  up, whereas for older versions we use a dummy/stub in the file errdot_old.go.
+// Once the minimum Go version restic supports is 1.19, remove this file and
+//  replace any calls to it with the corresponding code as per below.
+
+package backend
+
+import (
+	"errors"
+	"os/exec"
+)
+
+func IsErrDot(err error) bool {
+	return errors.Is(err, exec.ErrDot)
+}
--- a/internal/backend/errdot_old.go
+++ b/internal/backend/errdot_old.go
@@ -1,0 +1,13 @@
+//go:build !go1.19
+// +build !go1.19
+
+// This file provides a stub for IsErrDot() for Go versions below 1.19.
+// See the corresponding file errdot_119.go for more information.
+// Once the minimum Go version restic supports is 1.19, remove this file
+//  and perform the actions listed in errdot_119.go.
+
+package backend
+
+func IsErrDot(err error) bool {
+	return false
+}
--- a/internal/backend/sftp/sftp.go
+++ b/internal/backend/sftp/sftp.go
@@ -80,7 +80,10 @@ func startClient(cfg Config) (*SFTP, error) {
 
 	bg, err := backend.StartForeground(cmd)
 	if err != nil {
-		return nil, errors.Wrap(err, "cmd.Start")
+		if backend.IsErrDot(err) {
+			return nil, errors.Errorf("cannot implicitly run relative executable %v found in current directory, use -o sftp.command=./<command> to override", cmd.Path)
+		}
+		return nil, err
 	}
 
 	// wait in a different goroutine


DRON-666:
You can check for `ErrDot` error right after `exec.Command`:
```go
	cmd := exec.Command(command, args...)
	if errors.Is(cmd.Err, exec.ErrDot) {
```

Maybe worth adding something like this:
```go
	if filepath.Base(command) == command {
		file, err := os.Executable()
		if err == nil {
			dir := filepath.Dir(file)
			file, err = exec.LookPath(filepath.Join(dir, command))
			if err == nil || errors.Is(err, exec.ErrDot) {
				command = file
			}
		}
	}
```
before `exec.Command`. This will help to auto solve a common situation where `restic` and `rclone` located in the same directory.

rawtaz:
> You can check for ErrDot error right after exec.Command

Thank you, I've changed that to your suggestion!

> .. auto solve a common situation where restic and rclone located in the same directory

Yeah, we discussed that and decided to err on the side of caution even in this case, and to not work around an active security measure in Go.

MichaelEischer:
> before `exec.Command`. This will help to auto solve a common situation where `restic` and `rclone` located in the same directory.

For documentation purposes, a simpler variant of that code snippet is given in the golang documentation: https://pkg.go.dev/os/exec#hdr-Executables_in_the_current_directory

internal/backend/rclone/backend.go
```diff
@@ -85,6 +85,9 @@ func run(command string, args ...string) (*StdioConn, *sync.WaitGroup, func() er
 		err = errW
 	}
 	if err != nil {
+		if errors.IsErrDot(err) {
+			return nil, nil, nil, errors.Errorf("run: cannot implicitly run relative executable %v found in current directory, use -o rclone.program with ./ to override", cmd.Path)
```
MichaelEischer, line 89: it would be possible to explicitly list the necessary option in the error message:
`errors.Errorf("cannot implicitly run executable %v found in current directory, use '-o rclone.program=./%v' to override", cmd.Path, cmd.Path)` (the "run:" prefix doesn't seem necessary as the source of the error is clear)

internal/errors/errdot_119.go
```diff
@@ -0,0 +1,20 @@
+//go:build go1.19
```
MichaelEischer, line 1: Wouldn't `internal/backend` be a more suitable place? We only every check that error after using `internal/backend/foreground_*.go`/

internal/backend/rclone/backend.go
rawtaz: Yeah, I actually tried this, but found that it just made things overly complicated when people use quotes somewhere in their `rclone.program` or `sftp.command`.

**TL;DR:** Unless we can print the actual syntax that the user entered in their shell for the extended option (which we can't since by the time we get information about the extended option, the original syntax has been lost), it will be hard to give them a good or even working example of what they should run instead.

Taking SFTP as a (more obvious than rclone) example and a user running `restic -o sftp.command="ssh -o 'VerifyHostKeyDNS ask'" snapshots` (or whatever option to `ssh`, this is just a dummy example):
- We'd need to make the error message surround the `./%v` with quotes, so the code would be like: `errors.Errorf("startClient: cannot implicitly run relative executable %v found in current directory, use -o sftp.command='./%v' to override", cmd.Path, cmd.String())` or with `""` instead of `''`.
- Regardless of surrounding quote type, we don't get quoting information out of `Cmd.String()`, so when we print the error it becomes `.. use -o sftp.command='./ssh -o VerifyHostKeyDNS ask' to override`, losing the quotes around `VerifyHostKeyDNS ask`.

We'd have to do quite some deduction to print an extended option that is the same as the user originally entered with with `./` before it. Unless you know a way to do this, I'm open to ideas.

That said, my rationale was that `use -o sftp.command with ./ to override` would be clear enough. But perhaps `use -o sftp.command prefixed with ./ to override` or even `prefix the value of -o sftp.command with ./ to override` is even clearer. Some people might not know what `./` is and hence not know how to use it/where to place it.

rawtaz: Regarding the prefix, errors seems wrapped throughout restic source code so I figured it would be good to do that here too. But if you want me to remove it I'm fine with that!

internal/errors/errdot_119.go
rawtaz: I started out in `internal/backend` but when I found that there was a package that explicitly was about errors, I figured that a function checking an error for a specific type of error belonged in there instead :-) It's entierly up to you where to keep it (until Go 1.19 is the minimum supported Go version in restic), do you still want me to change it?

MichaelEischer: IMO the error package should only be used for generic error handling code. But the `ErrDot` error is really only relevant for the backend code. So from a code cohesion perspective, the this file belongs into `internal/backend`. So yes, please change the location.

internal/backend/rclone/backend.go
MichaelEischer: Hmm, is the quoting issue actually relevant when using the `rclone` backend? Unless someone uses quotes in the filename, things won't break... Or as an alternative what about `use '-o rclone.program=./<cmd>' to override` instead? That could make it more obvious what the syntax to use the option is.

Regarding sftp, after looking at the code, I'm a bit unsure what command I'd have to use. The auto-generated command is `ssh $host [-p $port] [-l $user] -s sftp`. So, I don't think a simple `-o sftp.command='./ssh'` can work at all?

rawtaz: > Hmm, is the quoting issue actually relevant when using the `rclone` backend?

It may not be a problem in the case of rclone, but I just never saw the point in risking it. I also don't see the point in making the rclone and sftp error messages different (except the option name), hence me trying to keep them under the same reasoning.

Rclone does have for example the `--config <path>` option which when having a space in ones username means one'll have to quote that path or escape the space in it (I'd say most people would quote it). I suppose people can put that in `rclone.args` though, but there's no corresponding `sftp.args`.

> Or as an alternative what about use `'-o rclone.program=./<cmd>' to override` instead? That could make it more obvious what the syntax to use the option is.

I'm not sure I agree :) First off, this entire discussion is about catering to those users that don't already, after just reading the slightly better error message and seeing the words "use" and "./" along with the option name, understands what they need to do. For those users, I dare say that some will think that they should now wrap the entire option in quotes (`'-o rclone.program=./<cmd>'` as opposed to `-o rclone.program=./<cmd>`). So the quotes should probably not be there. Second, if a user doesn't understand how to put `./` in the value of these options, I'm pretty sure they'd also start wondering "what \<cmd>/command should I use instead?". And speaking about quotes, `use -o rclone.program=./<cmd> to override` might, sadly, make some users think that they can write an actual command with spaces in place of `<cmd>` without quoting it, since there are no quotes in the suggestion.

This is why I figured "keep it simple" and that instead of trying to work around every possible misinterpretation of the suggestion in the error message, just saying "add a ./ before the value" would be clear enough.

Either way, let's do what you want here, I think there are more important things to discuss than this. What about `use -o rclone.program=./<cmd> to override` (that is, without the quotes)? And the equivalent for sftp of course.

MichaelEischer: > Rclone does have for example the `--config <path>` option which when having a space in ones username means one'll have to quote that path or escape the space in it (I'd say most people would quote it)

Looks like I should look into the code first. I've expected `rclone.program` to only contain the binary name, but not also possibly some flags from rclone.

> This is why I figured "keep it simple" and that instead of trying to work around every possible misinterpretation of the suggestion in the error message, just saying "add a ./ before the value" would be clear enough.

Let's keep it at the current wording then and just see what happens. There seem to be just too many corner cases to simply provide a proper error message. ("%q" would probably be enough to handle quoting for unix-like systems, but I have my doubts about windows)

rawtaz: > I've expected rclone.program to only contain the binary name, but not also possibly some flags from rclone.

But you're pretty much right in your expectation there, I think the purpose of `rclone.args` *is* to put stuff like `--config` in there instead of in `rclone.program`. However, now that you mentioned it, I realize that users might put it in `rclone.program` anyway, in which case we'd be back to it being hard to suggest them the correct option value.

> "%q" would probably be enough to handle quoting for unix-like systems, but I have my doubts about windows

Unfortunately even `%q` wouldn't yield the right message, e.g. `-o sftp.command="ssh -o 'VerifyHostKeyDNS ask'"` would be printed as `-o sftp.command="ssh -o VerifyHostKeyDNS ask"` as it just quotes the entire value.

> Let's keep it at the current wording then and just see what happens.

I changed it to a variant of your suggestion, it's now `use -o sftp.command=./<command> to override` and `use -o rclone.program=./<program> to override` :) I realized that my suggestion of `prefix the value of ..` is bad because users might not even have specified the extended option in the first place, so referring to that isn't very user friendly.

MichaelEischer (approved):
LGTM.

PR #3940, opened 2022-09-22
Better error message if connection to rclone fails

MichaelEischer:
When proxying the connection rclone using ssh (e.g. `restic -o rclone.program='ssh user@example.org forced-command' -r rclone: backup ...` as in https://ruderich.org/simon/notes/append-only-backups-with-restic-and-rclone ) this initial connection setup can fail.

In this case rclone currently often just returns a context canceled error, which is not particularly helpful. Replace this error with the output from rclone.

No.

Checklist
---------

### DIFF
--- a/internal/backend/rclone/backend.go
+++ b/internal/backend/rclone/backend.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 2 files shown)


MichaelEischer:
The PR now replaces all initial communication errors with the exit code returned by rclone. More precisely the replaced errors are  context canceled, unexpected EOF or broken pipe error. Neither of these errors is useful for understanding what exactly went wrong and how to fix it.

In addition, I've discovered that according to the documentation of `exec.Cmd` the current usage of the `StdErrPipe()` was wrong. `Wait()` must not be called before everything was read from the former pipe.

The expand on the motivation for this PR: I've been seeing the following errors when using restic with the rclone backend proxied via ssh, however, for some reason the error returned by ssh is completely missing. This might be caused by the above `StdErrPipe()` problem but that is just a guess so far.
```
Fatal: unable to open repository at rclone:: error talking HTTP to rclone: context canceled
Fatal: unable to open repository at rclone:: error talking HTTP to rclone: Get http://localhost/file-5577006791947779410: unexpected EOF
```

MichaelEischer:
LGTM.

Issue #3917, opened 2022-09-06
kind restic and other tools benchmark report


deajan:
Hello,

I'm currently doing benchmarks for deduplication backup tools, including restic.
I decided to write a script that would:
 - Install the backup programs
 - Prepare the source server
 - Prepare local targets / remote targets
 - Run backup and restore benchmarks
 - Use public available data (linux kernel sources as git repo) and checkout various git tags to simulate user changes in the dataset

The idea of the script would be to have reproductible results, the only changing factor being the machine specs & network link between sources and targets.

So far, I've run two sets of benchmarks, each done locally and remotely.
You can find the results at https://github.com/deajan/backup-bench

I'd love you to review the recipe I used for restic, and perhaps guide me on what parameters to use to get maximum performance.
Any remarks / ideas / PRs are welcome.

I've also made a comparaison table of some features of the backup solutions I'm benchmarking.
I still miss some informations for some of the backup programs.
Would you mind having a look at the comparaison table and fill the question marks related to the features of restic ?
Also, if restic has an interesting feature I didn't list, I'll be happy to extend the comparaison.

PS: I'm trying to be as unbiased as possible when it comes to those benchmarks, please forgive me if I didn't treat your program with the parameters it deserves.

Also, I've created the same issue in every git repo of the backup tools I'm testing, so every author / team / community member can judge / improve the instructions for better benchmarking.

JsBergbau:
Just some notes.

> I'd love you to review the recipe I used for restic, and perhaps guide me on what parameters to use to get maximum performance.
> Any remarks / ideas / PRs are welcome.

What do you understand as maximum performance? You can disable compression or enable maximum compression for smaller backupsizes, see https://restic.readthedocs.io/en/latest/047_tuning_backup_parameters.html?highlight=max#compression

> Also, if restic has an interesting feature I didn't list, I'll be happy to extend the comparaison.

Restic has its rest-server https://github.com/restic/rest-server and this has an append-only mode. With this you can make trojan safe backups, since backups can only be added but not removed. Very cool feature.

Some of your table
`Are metadatas encrypted too ?` Yes, see https://restic.readthedocs.io/en/latest/100_references.html#threat-model
`Can encrypted / compressed data be guessed (CRIME/BREACH style attacks)?` Restic is quite immune to watermark attacks https://restic.readthedocs.io/en/latest/100_references.html#backups-and-deduplication
[...]

PR #3880, opened 2022-08-20
archiver: Improve handling of "file xxx already present" error

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
This PR let's restic complete a backup if files with duplicate names exist and issues a warning instead of canceling the whole backup run. A duplicate node is only ignored if it is identical to its counterpart; this was the case in #1866 and serves as a safeguard for completely broken filesystems or internal bugs.

The change relies on a cleanup of the archiver.SaveTree method which now works similarly to the archiver.SaveDir. In particular this means using the TreeSaver, which could improve performance if many backup paths were given on the command line. See the commit message for more details.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
Fixes #1866

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- ~~[ ] I have added documentation for relevant changes (in the manual).~~
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3955, opened 2022-10-07
Improve archiver performance for small files

MichaelEischer:
As discovered in https://github.com/restic/restic/issues/3917#issuecomment-1267520636 the `backup` command is slower for small files (up to about 5-6 chunks) than it could be. The main bottleneck appears to be the FileSaver, as increasing the `--read-concurrency` also drastically improves the backup performance. The FileSaver proceeds in two steps, first it reads the file, splits it into chunks and passes it to the BlobSaver and secondly it waits for all blobs to complete saving (the upload happens asynchronously). Depending on the blob size it can take a few milliseconds to process it. During this time the FileSaver is blocked and not performing any useful work.

This PR changes the second phase of the FileSaver to work asynchronously using callbacks. For this the BlobSaver no longer returns a FutureBlob but instead calls a callback once a blob was saved. This provides the necessary means to asynchronously complete the FutureFile. Similar changes to FutureTree will have to wait at least until #3880 is merged.

The asynchronous file completion is protected using a mutex to avoid synchronization problems.

Asynchronously completing a file has a somewhat unexpected side effect. The status will now often list more than two files at the same time, even though the read concurrency is limited to two. Although this look a bit unexpected, it is actually correct. The status displays which files restic is currently working on, and not which file restic is reading from. However, I wonder whether we should change that to avoid confusing users.

The second commit in this PR changes the status output to clear no longer used status lines. With a frequently changing number of currently processed files, this ended up being confusing.

Performance measurements:
Command: `restic backup -n --exclude .git linux` using a checkout of Linux 6.0. 
CPU: AMD Ryzen 7 PRO 5850U ; Repository stored on NVME disk (warmed up page cache).
Times reported are those printed in the backup command summary.

| restic version / read concurrency | 2 | 4 | 8 |
| --- | --- | --- | --- |
| master 7112a132c3c568497309ac034f1da62bdf4738a1 | 0:10 | 0:06 | 0:04 |
| PR | 0:04 | 0:03 | 0:03 |

I'm aware that the test dataset is too small to properly differentiate the different variants, but the trends should be obvious :-) .

See #3917.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-3955
+++ b/changelog/unreleased/pull-3955
@@ -1,0 +1,9 @@
+Enhancement: Improve backup performance for small files
+
+When backing up small files restic was slower than it could be. In particular
+this affected backups using maximum compression.
+
+This has been fixed by reworking the internal parallelism of the backup
+command.
+
+https://github.com/restic/restic/issues/3955
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -68,6 +68,9 @@ type Archiver struct {
 	// be in the snapshot after saving. s contains some statistics about this
 	// particular file/dir.
 	//
+	// Once reading a file has completed successfully (but not saving it yet),
+	// CompleteItem will be called with current == nil.
+	//
 	// CompleteItem may be called asynchronously from several different
 	// goroutines!
 	CompleteItem func(item string, previous, current *restic.Node, s ItemStats, d time.Duration)
@@ -431,6 +434,8 @@ func (arch *Archiver) Save(ctx context.Context, snPath, target string, previous 
 		// Save will close the file, we don't need to do that
 		fn = arch.fileSaver.Save(ctx, snPath, target, file, fi, func() {
 			arch.StartFile(snPath)
+		}, func() {
+			arch.CompleteItem(snPath, nil, nil, ItemStats{}, 0)
 		}, func(node *restic.Node, stats ItemStats) {
 			arch.CompleteItem(snPath, previous, node, stats, time.Since(start))
 		})
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -53,12 +53,21 @@ func saveFile(t testing.TB, repo restic.Repository, filename string, filesystem 
 	}
 
 	var (
+		completeReadingCallback bool
+
 		completeCallbackNode  *restic.Node
 		completeCallbackStats ItemStats
 		completeCallback      bool
 
 		startCallback bool
 	)
+
+	completeReading := func() {
+		completeReadingCallback = true
+		if completeCallback {
+			t.Error("callbacks called in wrong order")
+		}
+	}
 
 	complete := func(node *restic.Node, stats ItemStats) {
 		completeCallback = true
@@ -80,7 +89,7 @@ func saveFile(t testing.TB, repo restic.Repository, filename string, filesystem 
 		t.Fatal(err)
 	}
 
-	res := arch.fileSaver.Save(ctx, "/", filename, file, fi, start, complete)
+	res := arch.fileSaver.Save(ctx, "/", filename, file, fi, start, completeReading, complete)
 
 	fnr := res.take(ctx)
 	if fnr.err != nil {
@@ -99,6 +108,10 @@ func saveFile(t testing.TB, repo restic.Repository, filename string, filesystem 
 
 	if !startCallback {
 		t.Errorf("start callback did not happen")
+	}
+
+	if !completeReadingCallback {
+		t.Errorf("completeReading callback did not happen")
 	}
 
 	if !completeCallback {
--- a/internal/archiver/blob_saver.go
+++ b/internal/archiver/blob_saver.go
@@ -43,51 +43,18 @@ func (s *BlobSaver) TriggerShutdown() {
 
 // Save stores a blob in the repo. It checks the index and the known blobs
 // before saving anything. It takes ownership of the buffer passed in.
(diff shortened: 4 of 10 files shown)


JsBergbau:
@MichaelEischer thanks for working on this.

I used the same folder as described there https://github.com/restic/restic/issues/3917

default read concurrency of 2
```
repository b677436b opened (repository version 2) successfully, password is correct
created new cache in C:\Users\XXX\AppData\Local\restic

Files:         556 new,     0 changed,     0 unmodified
Added to the repository: 2.526 GiB (2.504 GiB stored)

processed 556 files, 2.526 GiB in 1:52
snapshot d1c73cdc saved
```

So much faster than before. I noticed that before the PR restic showed at maximum as many files beeing processed as read concurrency was. Now the number varies a lot and about 4 to 10 files are shown simultaneously. 

read concurrency of 6

```
repository e5410fd6 opened (repository version 2) successfully, password is correct
created new cache in C:\Users\XXX\AppData\Local\restic

Files:         556 new,     0 changed,     0 unmodified
Added to the repository: 2.526 GiB (2.504 GiB stored)

processed 556 files, 2.526 GiB in 1:52
snapshot 58d7deda saved
```

So now using read concurrency doesn't improve speed. However overall speed is decreased a bit compared to previous implementation when increasing the read concurrency.

MichaelEischer:
> Now the number varies a lot and about 4 to 10 files are shown simultaneously.

That is somewhat expected with this PR. However, showing more files than before does not mean that restic is concurrently reading from more files than before.

> So now using read concurrency doesn't improve speed. However overall speed is decreased a bit compared to previous implementation when increasing the read concurrency.

I wouldn't consider that minimal performance difference to be too relevant. However, I have to admit that I don't have any real idea why the PR should result in a slightly higher backup time than before.

JsBergbau:
I've done a test with the previous version again. With read conccurrency of 6 it is now 1:41 

I've also tested the new restic version again. With read concurrency of 6 it takes 1:47

The difference for the faster speed is that the VM that was building the PR build was still running. With VM closed I guess Windows is caching the whole folder in RAM.

So this PR is about 6 % slower than the older implementation when using it with a higher read concurrency.

JsBergbau:
@MichaelEischer 

Did some tests again. Previously I used for the test https://beta.restic.net/latest/ 
This binary is about 30 MB in size. Now I've build restic directly from masterbranch, so without this PR. With read concurrency of 6 restic needs now 1:50. So this PR is even with read concurrency of 6 faster. Restic version `restic 0.14.0 (v0.9.5-2107-g1a6160d15) compiled with go1.15.15 on windows/amd64` with file size of 22,7 MB.

Previously used restic version without this PR (all benchmarks before this post) `restic 0.14.0-dev (compiled manually) compiled with go1.18 on windows/amd64`

So speed differences resp. faster results were because of newer go version. Sorry for not thinking about this before.

MichaelEischer:
What remains is to decide how to handle the much higher number of files displayed as part of the status message. It will likely confuse users...

JsBergbau:
As far as I understood with this pull request restic still reads by default 2 files in parallel. So the question is how does this then happen that there are now displayed more files than previously? 
Still it is a good question how to handle this. I have some ideas, but lets wait for your answer.

MichaelEischer:
> So the question is how does this then happen that there are now displayed more files than previously?

That is somewhat easy to explain. Let's assume we have a small file which only consists of a single chunk. Then restic reads the file (at this point it shows up on the status bar), hands it over to the blob saver which stores it in the repository, once that is completed the file is reported as complete and removed from the status bar. Only the first step actually involves reading the file from the disk. In particular, for maximum compression it takes quite some time to store the chunk in the repository.

Previously the read concurrency applied to this whole sequence, which means that restic cannot make use of more CPU cores that the read concurrency, and that the next file is only requested from disk when the previous one has completed processing. That way the CPU and disk are both underutilized.

With this PR, the read concurrency only applies to the step which actually reads the file. Thus, reading new files and compressing the old one happens concurrently.

The status bar currently shows all files which are in any of these processing steps. This has the consequence that all files which are currently read or saved to the repository are reported.

> Still it is a good question how to handle this. I have some ideas, but lets wait for your answer.

The least confusing way is probably to hide files from the status bar once the file reading step is complete. The main complexity of that approach is that this introduces a new intermediate state in the output. The verbose output must still be able to report the final amount of data stored for a file. This means that we'd have to send an additional signal to the progress output once the file reading part is complete. But I guess that's better than confusing everyone.

JsBergbau:
I think the problem is that restic users are accustomed that only the files are shown that restic reads currently from disk.
So one option would be to show a message that since restic 0.15 are lot more files are shown like "Note since restic 0.15 restic parallelizes file reading and compression so more files are shown here than before. Files read concurrently from disk: $readConcurrency"

Or the other option is the cosmetic implementation which needs quite some time for no real functional gain.

MichaelEischer:
> So one option would be to show a message that since restic 0.15 are lot more files are shown like "Note since restic 0.15 restic parallelizes file reading and compression so more files are shown here than before. Files read concurrently from disk: $readConcurrency"

That just feels wrong. I gives the impression that we're too lazy to implement a proper status output.

I've implemented the variant which just hides all files which already have completed reading from disk. It's actually less messy to implement than I've expected.

MichaelEischer:
LGTM. I've slightly tweaked a comment and added the `isCompleted` sanity check also to `completeError`.

mamoit:
Damn...
Just tried out release `0.15.0` and my home dir hourly backup time was cut in half!
From around ~55s to <30s for <100MB changed in a total of 500k files and 30GB.

Hats off.

PR #3959, opened 2022-10-08
backup: Use buffered channels to collect backup status

MichaelEischer:
While working on #3955 I've noticed that the block profile of go reported that the two FileSaver were in total blocked for 0.8 seconds while reporting progress. For comparison, the whole backup run took less than 4 seconds. Thus, add a small buffer to the channels used to report status information to prevent blocking. This also has the benefit of reducing the stress on the golang runtime scheduler.

As the status information is purely informational, it doesn't matter that the status reporting shutdown is somewhat racy and could miss a few final updates.

Click on the image below, to see the full block profile (look for `FileSaver` calls).
profile001

No.

Checklist
---------

### DIFF
--- a/internal/ui/backup/progress.go
+++ b/internal/ui/backup/progress.go
@@ -87,10 +87,13 @@ func NewProgress(printer ProgressPrinter) *Progress {
 		MinUpdatePause: time.Second / 60,
 		start:          time.Now(),
 
-		totalCh:     make(chan Counter),
-		processedCh: make(chan Counter),
+		// use buffered channels for the information used to update the status
+		// the shutdown of the `Run()` method is somewhat racy, but won't affect
+		// the final backup statistics
+		totalCh:     make(chan Counter, 100),
+		processedCh: make(chan Counter, 100),
 		errCh:       make(chan struct{}),
-		workerCh:    make(chan fileWorkerMessage),
+		workerCh:    make(chan fileWorkerMessage, 100),
 		closed:      make(chan struct{}),
 
 		summary: &Summary{},


greatroar:
LGTM as a stopgap, but I do think this code needs an overhaul. A mutex on the Progress struct would make it much easier to reason about and probably equally fast. E.g., unless I'm mistaken, there's a race condition between

https://github.com/restic/restic/blob/1a6160d15233fb2dd8f4655d9f46beb39378fdf5/internal/ui/backup/progress.go#L132

and

https://github.com/restic/restic/blob/1a6160d15233fb2dd8f4655d9f46beb39378fdf5/internal/ui/backup/progress.go#L297

greatroar:
In fact, one of the reasons why the progress reporter spends so much time in scheduling is probably the way its main loop is structured. It has a time.Ticker, but instead of using that to determine when to print an update, it checks time.Since(lastUpdate) every time a message comes in.

If I understand the Go scheduling algorithm correctly, this means that every time a thread sends a message, it finds the receiving goroutine blocked and starts executing that, leaving its actual worker goroutine to be rescheduled.

I have a cleaned up version in a local branch. It needs some finishing touches.

MichaelEischer:
> unless I'm mistaken, there's a race condition between

It would be a race condition, if `ReportTotal` were used by something else than the Scanner. As the scanner tells `Progress` to close the channel once it's finished and afterwards never calls `ReportTotal` again, this should be fine.

> If I understand the Go scheduling algorithm correctly, this means that every time a thread sends a message, it finds the receiving goroutine blocked and starts executing that, leaving its actual worker goroutine to be rescheduled.

The sending thread will block, unless `Progress.Run` is waiting to receive data. Thus, if `Run` were fast enough this could work without blocking the goroutine sending the backup progress. But that is apparently not the case. By buffering the progress reporting channels, it should be possible to avoid this most of the time.

> but I do think this code needs an overhaul.

I agree, but I've decided to not change this myself to not distract me from finishing some other PRs ;-)



> I have a cleaned up version in a local branch. It needs some finishing touches.

Would that replace this PR? Or should I merge this one first?

greatroar:
> Would that replace this PR? Or should I merge this one first?

I suggest you merge this one first.

MichaelEischer:
LGTM.

PR #3977, opened 2022-10-21
ui/backup: Replace channels with a mutex

greatroar:

Redoes the backup progress reporter using a single mutex instead of a multiple channels. This is not very pretty, but it seems to work. Once we can depend on Go 1.19, we can replace much of this with sync/atomic types to reduce the amount of code and maybe improve performance further.

Discussed at #3959.

Checklist
---------

### DIFF
--- a/internal/ui/backup/progress.go
+++ b/internal/ui/backup/progress.go
@@ -11,6 +11,8 @@ import (
 	"github.com/restic/restic/internal/ui/signals"
 )
 
+// A ProgressPrinter can print various progress messages.
+// It must be safe to call its methods from concurrent goroutines.
 type ProgressPrinter interface {
 	Update(total, processed Counter, errors uint, currentFiles map[string]struct{}, start time.Time, secs uint64)
 	Error(item string, err error) error
@@ -32,13 +34,7 @@ type Counter struct {
 	Files, Dirs, Bytes uint64
 }
 
-type fileWorkerMessage struct {
-	filename string
-	done     bool
-}
-
 type Summary struct {
-	sync.Mutex
 	Files, Dirs struct {
 		New       uint
 		Changed   uint
@@ -50,14 +46,18 @@ type Summary struct {
 
 // Progress reports progress for the `backup` command.
 type Progress struct {
+	mu sync.Mutex
+
 	interval time.Duration
 	start    time.Time
 
-	totalCh     chan Counter
-	processedCh chan Counter
-	errCh       chan struct{}
-	workerCh    chan fileWorkerMessage
-	closed      chan struct{}
+	scanStarted, scanFinished bool
+
+	currentFiles     map[string]struct{}
+	processed, total Counter
+	errors           uint
+
+	closed chan struct{}
 
 	summary Summary
 	printer ProgressPrinter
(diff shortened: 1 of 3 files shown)


internal/ui/backup/progress.go
```diff
@@ -84,191 +82,165 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
...
 
 		if previous == nil {
 			p.printer.CompleteItem("dir new", item, previous, current, s, d)
-			p.summary.Lock()
+			p.mu.Lock()
 			p.summary.Dirs.New++
-			p.summary.Unlock()
+			p.mu.Unlock()
 			return
 		}
```
MichaelEischer, line 192: what about removing this `return` by replacing the next `if` with `else if`? That would make the control flow easier to follow.

internal/ui/backup/progress.go
```diff
@@ -48,16 +44,24 @@ type Summary struct {
...
+
 	interval time.Duration
 	start    time.Time
 
-	totalCh     chan Counter
-	processedCh chan Counter
-	errCh       chan struct{}
-	workerCh    chan fileWorkerMessage
-	closed      chan struct{}
+	state            uint8
```
MichaelEischer, line 59: It would be simpler to just use `started` and `finished` flags. It took me a while to notice that `state` is a bitfield.

internal/ui/backup/progress.go
```diff
@@ -84,191 +82,165 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
...
-
-	t := time.NewTicker(time.Second)
-	signalsCh := signals.GetProgressChannel()
-	defer t.Stop()
 	defer close(p.closed)
 	// Reset status when finished
 	defer p.printer.Reset()
 
+	if p.interval == 0 {
+		return
```
MichaelEischer, line 90: Won't this prevent the `signalsCh` from working when there's no periodic status update?

internal/ui/backup/progress.go
```diff
@@ -84,191 +82,165 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
...
+
+		if state&stateStarted == 0 {
+			p.mu.Unlock()
 			continue
 		}
-		lastUpdate = time.Now()
 
-		p.printer.Update(total, processed, errors, currentFiles, p.start, secondsRemaining)
+		var secondsRemaining uint64
+		if state&stateFinished != 0 {
```
MichaelEischer, line 117: `stateFinished` is highly misleading, as it appears to complement `stateStarted`, but that is not the case. `scanFinished` would be better.

MichaelEischer (commented):
The general design looks sane. But I've noticed a few things that should be changed. In particular, the `state` attribute was rather confusing for me and it looks like printing progress after receiving a certain signal is broken.

internal/ui/backup/progress.go
greatroar: Replaced by boolean fields scanStarted and scanFinished.

internal/ui/backup/progress.go
greatroar: Right. Fixed.

internal/ui/backup/progress.go
greatroar: Changed to a three-case switch statement.

MichaelEischer (approved):
LGTM. Thanks for making he code a bit more readable!

PR #3982, opened 2022-10-22
Show selected compression level when opening repository

MichaelEischer:
Currently there is no indication which compression level is by restic when operating on a repository. This PR extends the status line displayed when opening the repository.

See https://forum.restic.net/t/how-to-check-if-files-were-compressed/5392/8

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -498,7 +498,7 @@ func OpenRepository(ctx context.Context, opts GlobalOptions) (*repository.Reposi
 			id = id[:8]
 		}
 		if !opts.JSON {
-			Verbosef("repository %v opened (repository version %v, compression level %v)\n", id, s.Config().Version, opts.Compression.String())
+			Verbosef("repository %v opened (version %v, compression level %v)\n", id, s.Config().Version, opts.Compression.String())
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/restic/global.go
```diff
@@ -498,7 +498,7 @@ func OpenRepository(ctx context.Context, opts GlobalOptions) (*repository.Reposi
 			id = id[:8]
 		}
 		if !opts.JSON {
-			Verbosef("repository %v opened (repository version %v) successfully, password is correct\n", id, s.Config().Version)
+			Verbosef("repository %v opened (repository version %v, compression level %v)\n", id, s.Config().Version, opts.Compression.String())
```
rawtaz, line 501: I think that it would be better if we ditch the second "repository" word such that the text is just this instead: `repository %v opened (version %v, compression level %v)\n`

It's so utterly obvious that the first version mentioned refers to the repository that was opened, and if there are any other types of versions to name in the future they can be explicitly mentioned, e.g. with `repository %v opened (version %v, compression level %v, determination version %v, metadata version %v)\n`.

rawtaz (changes requested):
LGTM but would love your input on my comment.

cmd/restic/global.go
MichaelEischer: Is changed.

rawtaz (approved):
👍

PR #3986, opened 2022-10-25
ui/progress: Load both values in a single Lock/Unlock

greatroar:

We always need both values from a Counter, except in a test, so we don't need to lock twice and risk scheduling in between.

Also, removed the resetting in Done. This copied a mutex, which isn't allowed. Static analyzers tend to trip over that. I'm surprised golangci-lint hasn't caught it yet.

No.

Checklist
---------

### DIFF
--- a/internal/restic/find_test.go
+++ b/internal/restic/find_test.go
@@ -110,7 +110,8 @@ func TestFindUsedBlobs(t *testing.T) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


MichaelEischer (approved):
LGTM. It feels a bit like a micro-optimization, but atomically loading both values is nevertheless a good idea.

PR #3985, opened 2022-10-23
Fix bug in lock refresh monitoring 

saltsa:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Lock refresh monitoring timestamps were changed from seconds to nanoseconds. However there was one place where this change wasn't made. This causes restic backups to stop about after 5 minutes because trying to subtract seconds from nanoseconds.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
No

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3956, opened 2022-10-07
lock: Use the correct duration to check for expired locks

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
When checking whether a lock is refreshed in time, it makes no sense to use to basic refresh interval (5min). Instead the timeout after which a lock is considered stale (30mins - a small of buffer) must be used. Otherwise normal operations could just break.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
Regressed in #3569
Fixes #3954

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- ~~[ ] I have added tests for all code changes.~~
- ~~[ ] I have added documentation for relevant changes (in the manual).~~
- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).~~
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3569, opened 2021-11-06
Strict repository lock handling

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
restic operations did not check that their lock is still valid. This can lead to a situation where for example a backup client is paused, then in the meantime `unlock` is called for the repository, which removes the now stale lock, and some time later the backup continues. If in the meantime `prune` was called, then this results in a broken snapshot.

This PR changes the locking behavior to actually be strict. If restic is unable to refresh locks in time, then the whole operation will be cancelled. This is done by tying the context used by restic's commands to the lock lifetime. If the monitoring goroutine for the lock detects that the lock file was not refresh in time, then the context will be canceled. Thereby, the command is forcibly terminated.

To keep the implementation simple there are now two goroutines per lock. One which periodically refreshes the lock file and one which monitors that the expiry is done in time. The time limit to refresh the lock file is a few minutes shorter than the duration after which a lock file becomes stale. This is intended to compensate for a small amount of clock drift between clients.

Most code changes revolve around cleaning up the usage of the global context previously available via `globalOptions.ctx`. Depending on the command either that context was accessed via `gopts.ctx` or a local copy in `ctx`. Alternatively, a few commands also introduce an additional context using `context.WithCancel`. The latter is unnecessary as the global context is canceled when restic shuts down. In total, these different ways to use the context made it non-obvious which contexts have to be tied to the lock lifetime to ensure that the command properly terminates after a failed lock refresh.

This PR solves the "additional behavior changes" proposed in #2715. However, it does not check whether a lock file disappeared or not as the could lead to race conditions with storage backends which do not provide strongly consistent directory listings.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
Fixes #2715

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
[...]

PR #3990, opened 2022-10-29
lock: add test to check that refreshing works

MichaelEischer:
Since merging the strict lock refresh PR #3569, we've already had two regressions (#3956, #3985) in which restic failed to refresh locks after the standard interval of 5 minutes. The root cause was that the tests only tested cases in which the lock was expected to fail refreshing, but not the positive case in which the lock is refreshed successfully.

Thus, this PR adds the missing test case. I've manually tested that the test case would have caught both regressions.

See #3985.

Checklist
---------


MichaelEischer:
LGTM.

Issue #3627, opened 2022-01-09
Progress for restore operation


gnottgen:
Output of `restic version`
--------------------------

restic 0.12.1 compiled with go1.16.6 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I would apreciate if the restore command would show a progress of the operation. Very nice would be a percentage of the remaining size (maybe with an ETA), but already the current file in progress would do the trick.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I tried a restore (just for testing) of a big repository and wondered if restic is still working or if it is hanging, because there was no output at all (after the initial information about the snapshot to restore, of course).

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Though I didn't had to do a restore until now (and hopingly I never will have to), it seems to be an excellent backup tool for me. It's great to use in scripts and my backup and the backup of several PCs of my family are relying on restic. 
I'm running it through a PowerShell-script on windows-machines pushing to a linux machine via sftp and it works like charm. Also the repo's integrity check is a great tool to use before mirroring the repo to an out-of-house-machine.
Thank you very much for restic! (i tried to donate, but i only found #884 and #2415 which unfortunately are saying that it's not easily possible) But it makes me happy every time i'm doing a backup ;-)

dimejo:
Sounds like a duplicate of #426.

MichaelEischer:
It is.

PR #3983, opened 2022-10-23
Centralize and fix formatting of bytes, percentages, durations

greatroar:
What does this PR change? What problem does it solve?
-----------------------------------------------------

The formatBytes, formatPercent, formatDuration and formatSeconds functions were duplicated in cmd and ui/backup. They're now exported functions in ui. Also, formatBytes used > 2^k instead of >= 2^k to switch prefixes, so it would report "1024.000 KiB" instead of "1.000 MiB". That's fixed now. I added some tests.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

No.

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #3991, opened 2022-10-29
restore: show progress

johanna-herrmann:

After merging this PR restic will show progress details while restoring. \
Example: [0:42]  23 / 3456 Files,  12.98 MiB / 23.54 GiB,  5.76 %

The progress details line updates in-place and does not exist with -q option. \
The progress details line updates on each blob written.

Yes

* Issue [Progress for restore operation, 3627](https://github.com/restic/restic/issues/3627)
* Forum Thread [Progress bar for restore](https://forum.restic.net/t/progress-bar-for-restore/5210)

Checklist
---------

      (Nothing relevant)

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/restic/restic/internal/debug"
@@ -10,6 +11,9 @@ import (
 	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/restorer"
+	"github.com/restic/restic/internal/ui"
+	restoreui "github.com/restic/restic/internal/ui/restore"
+	"github.com/restic/restic/internal/ui/termstatus"
 
 	"github.com/spf13/cobra"
 )
@@ -31,7 +35,31 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runRestore(cmd.Context(), restoreOptions, globalOptions, args)
+		ctx := cmd.Context()
+		var wg sync.WaitGroup
+		cancelCtx, cancel := context.WithCancel(ctx)
+		defer func() {
+			// shutdown termstatus
+			cancel()
+			wg.Wait()
+		}()
+
+		term := termstatus.New(globalOptions.stdout, globalOptions.stderr, globalOptions.Quiet)
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			term.Run(cancelCtx)
+		}()
+
+		// allow usage of warnf / verbosef
+		prevStdout, prevStderr := globalOptions.stdout, globalOptions.stderr
+		defer func() {
+			globalOptions.stdout, globalOptions.stderr = prevStdout, prevStderr
+		}()
+		stdioWrapper := ui.NewStdioWrapper(term)
+		globalOptions.stdout, globalOptions.stderr = stdioWrapper.Stdout(), stdioWrapper.Stderr()
+
+		return runRestore(ctx, restoreOptions, globalOptions, term, args)
 	},
 }
 
@@ -64,7 +92,9 @@ func init() {
 	flags.BoolVar(&restoreOptions.Verify, "verify", false, "verify restored files content")
 }
 
-func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions, args []string) error {
+func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
+	term *termstatus.Terminal, args []string) error {
+
 	hasExcludes := len(opts.Exclude) > 0 || len(opts.InsensitiveExclude) > 0
 	hasIncludes := len(opts.Include) > 0 || len(opts.InsensitiveInclude) > 0
 
@@ -145,7 +175,12 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions, a
 		return err
 	}
 
-	res := restorer.NewRestorer(ctx, repo, sn, opts.Sparse)
+	var progress *restoreui.Progress
+	if !globalOptions.Quiet && !globalOptions.JSON {
+		progress = restoreui.NewProgress(restoreui.NewProgressPrinter(term), calculateProgressInterval(!gopts.Quiet, gopts.JSON))
+	}
+
+	res := restorer.NewRestorer(ctx, repo, sn, opts.Sparse, progress)
 
 	totalErrors := 0
 	res.Error = func(location string, err error) error {
@@ -207,6 +242,10 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions, a
 	err = res.RestoreTo(ctx, opts.Target)
 	if err != nil {
 		return err
+	}
+
+	if progress != nil {
+		progress.Finish()
 	}
 
 	if totalErrors > 0 {
--- a/internal/restorer/filerestorer.go
+++ b/internal/restorer/filerestorer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/ui/restore"
 )
 
 // TODO if a blob is corrupt, there may be good blob copies in other packs
@@ -54,6 +55,7 @@ type fileRestorer struct {
 	filesWriter *filesWriter
 	zeroChunk   restic.ID
 	sparse      bool
+	progress    *restore.Progress
 
 	dst   string
 	files []*fileInfo
@@ -65,7 +67,8 @@ func newFileRestorer(dst string,
 	key *crypto.Key,
 	idx func(restic.BlobHandle) []restic.PackedBlob,
 	connections uint,
-	sparse bool) *fileRestorer {
+	sparse bool,
+	progress *restore.Progress) *fileRestorer {
 
 	// as packs are streamed the concurrency is limited by IO
 	workerCount := int(connections)
@@ -77,6 +80,7 @@ func newFileRestorer(dst string,
 		filesWriter: newFilesWriter(workerCount),
 		zeroChunk:   repository.ZeroChunk(),
 		sparse:      sparse,
+		progress:    progress,
 		workerCount: workerCount,
 		dst:         dst,
 		Error:       restorerAbortOnAllErrors,
@@ -268,7 +272,13 @@ func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) error {
 						file.inProgress = true
 						createSize = file.size
 					}
-					return r.filesWriter.writeToFile(r.targetPath(file.location), blobData, offset, createSize, file.sparse)
+					writeErr := r.filesWriter.writeToFile(r.targetPath(file.location), blobData, offset, createSize, file.sparse)
+
+					if r.progress != nil {
+						r.progress.AddProgress(file.location, uint64(len(blobData)), uint64(file.size))
+					}
+
+					return writeErr
 				}
 				err := sanitizeError(file, writeToFile())
 				if err != nil {
--- a/changelog/unreleased/issue-3627
+++ b/changelog/unreleased/issue-3627
@@ -1,0 +1,9 @@
+Enhancement: Show progress bar during restore
+
+The `restore` command now shows a progress report while restoring files.
+
+Example: [0:42]  5.76%  23 files 12.98 MiB, total 3456 files 23.54 GiB
+
+https://github.com/restic/restic/issues/3627
+https://github.com/restic/restic/pull/3991
+https://forum.restic.net/t/progress-bar-for-restore/5210
--- a/cmd/restic/integration_test.go
+++ b/cmd/restic/integration_test.go
@@ -112,7 +112,7 @@ func testRunRestoreLatest(t testing.TB, gopts GlobalOptions, dir string, paths [
 		},
 	}
 
-	rtest.OK(t, runRestore(context.TODO(), opts, gopts, []string{"latest"}))
+	rtest.OK(t, runRestore(context.TODO(), opts, gopts, nil, []string{"latest"}))
 }
 
 func testRunRestoreExcludes(t testing.TB, gopts GlobalOptions, dir string, snapshotID restic.ID, excludes []string) {
@@ -121,7 +121,7 @@ func testRunRestoreExcludes(t testing.TB, gopts GlobalOptions, dir string, snaps
 		Exclude: excludes,
 	}
 
-	rtest.OK(t, runRestore(context.TODO(), opts, gopts, []string{snapshotID.String()}))
+	rtest.OK(t, runRestore(context.TODO(), opts, gopts, nil, []string{snapshotID.String()}))
 }
 
 func testRunRestoreIncludes(t testing.TB, gopts GlobalOptions, dir string, snapshotID restic.ID, includes []string) {
@@ -130,11 +130,11 @@ func testRunRestoreIncludes(t testing.TB, gopts GlobalOptions, dir string, snaps
 		Include: includes,
 	}
 
-	rtest.OK(t, runRestore(context.TODO(), opts, gopts, []string{snapshotID.String()}))
+	rtest.OK(t, runRestore(context.TODO(), opts, gopts, nil, []string{snapshotID.String()}))
 }
 
 func testRunRestoreAssumeFailure(t testing.TB, snapshotID string, opts RestoreOptions, gopts GlobalOptions) error {
-	err := runRestore(context.TODO(), opts, gopts, []string{snapshotID})
+	err := runRestore(context.TODO(), opts, gopts, nil, []string{snapshotID})
 
 	return err
 }
--- a/internal/restorer/filerestorer_test.go
+++ b/internal/restorer/filerestorer_test.go
@@ -150,7 +150,7 @@ func newTestRepo(content []TestFile) *TestRepo {
 func restoreAndVerify(t *testing.T, tempdir string, content []TestFile, files map[string]bool, sparse bool) {
 	repo := newTestRepo(content)
 
-	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, sparse)
+	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, sparse, nil)
 
 	if files == nil {
 		r.files = repo.files
@@ -265,7 +265,7 @@ func TestErrorRestoreFiles(t *testing.T) {
 		return loadError
 	}
 
-	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, false)
+	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, false, nil)
 	r.files = repo.files
 
 	err := r.restoreFiles(context.TODO())
@@ -304,7 +304,7 @@ func testPartialDownloadError(t *testing.T, part int) {
 		return loader(ctx, h, length, offset, fn)
 	}
 
-	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, false)
+	r := newFileRestorer(tempdir, repo.loader, repo.key, repo.Lookup, 2, false, nil)
 	r.files = repo.files
 	r.Error = func(s string, e error) error {
 		// ignore errors as in the `restore` command
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -10,6 +10,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/restic"
+	restoreui "github.com/restic/restic/internal/ui/restore"
 
 	"golang.org/x/sync/errgroup"
 )
@@ -20,6 +21,8 @@ type Restorer struct {
 	sn     *restic.Snapshot
 	sparse bool
 
+	progress *restoreui.Progress
+
 	Error        func(location string, err error) error
 	SelectFilter func(item string, dstpath string, node *restic.Node) (selectedForRestore bool, childMayBeSelected bool)
 }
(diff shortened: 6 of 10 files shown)


changelog/unreleased/pull-3991
```diff
@@ -0,0 +1,5 @@
+Enhancement: Adds progress details while restoring
+
+Example: [0:42] 23 / 3456 Files, 12.98 MiB / 23.54 GiB, 5.76 %
```
greatroar, line 3: This can do with some more prose:

```suggestion
The command restic restore now shows a progress report while it's restoring files.

Example: [0:42] 23 / 3456 Files, 12.98 MiB / 23.54 GiB, 5.76 %
```

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type RestoreProgressPrinter struct {
+	progressInfoMap  map[string]progressInfoEntry
+	filesFinished    int64
+	filesTotal       int64
+	allBytesWritten  int64
+	allBytesTotal    int64
+	started          time.Time
+	sync.Mutex
```
greatroar, line 29: It's customary to put the mutex before the fields that it protects, i.e., at the start of the struct:

```suggestion
	sync.Mutex
	progressInfoMap  map[string]progressInfoEntry
	filesFinished    int64
	filesTotal       int64
	allBytesWritten  int64
	allBytesTotal    int64
	started          time.Time
```

internal/restorer/progressprinter/progressprinter_test.go
```diff
@@ -0,0 +1,251 @@
...
+		actual := actualValues[i]
+		expected := expectedValues[i]
+		if expected != actual {
+			test.Errorf("Calculate Percent Error. Have %v, wanted %v", actual, expected)
+		}
+	}
+}
+
+func TestNew(t *testing.T) {
+  printer := New()
```
greatroar, line 73: I think you forget to run go fmt.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,114 @@
...
+		lineResetter, timeLeft, p.filesFinished, p.filesTotal, formattedAllBytesWritten, formattedAllBytesTotal, allPercent)
+}
+
+func New() *RestoreProgressPrinter {
+	return &RestoreProgressPrinter{
+		progressInfoMap: make(map[string]progressInfoEntry),
+	  filesFinished:   0,
+	  filesTotal:      0,
+		allBytesWritten: 0,
+	  allBytesTotal:   0,
```
greatroar, line 83: Fields are zero by default, so no need to write this out:

```suggestion
```

internal/restorer/progressprinter/progressprinter_test.go
```diff
@@ -0,0 +1,251 @@
...
+	progresswriter = &writer
+	time.Sleep(time.Second)
+
+	print(printer)
+
+	printed := writer.String()
+	if printed != expectedPrint {
+		t.Errorf("Printer Error.\nHave   %v\nwanted %v",
+			strings.Replace(strings.Replace(printed, "\033", "\\033", -1), "\r", "\\r", -1), 
+			strings.Replace(strings.Replace(expectedPrint, "\033", "\\033", -1), "\r", "\\r", -1))
```
greatroar, line 217: %q does escaping for you:

```suggestion
		t.Errorf("Printer Error.\nHave   %q\nwanted %q", printed, expectedPrint)
```

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,114 @@
...
+	"sync"
+	"os"
+	"io"
+	"time"
+)
+
+var progresswriter io.Writer = os.Stdout
+
+// \r = carriage return (cursor to start of line);  \033[K = line deletion (delete from cursor until end of line)
+var lineResetter string = "\r\033[K"
```
greatroar, line 14: This is Unix-specific, right?

mark-herrmann: Fixed by simply using \r.
Only crtical moment is when switch unit, solved by adding 3 spaces at end of line

mark-herrmann:
Solved problems and suggestions

cmd/restic/cmd_version.go
```diff
@@ -21,7 +21,7 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 `,
 	DisableAutoGenTag: true,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("restic %s compiled with %v on %v/%v\n",
+		fmt.Printf("restic %s (pr-3991: progress details on restore) compiled with %v on %v/%v\n",
```
MichaelEischer, line 24: Please revert this line.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+	"time"
+)
+
+var progresswriter io.Writer = os.Stdout
```
MichaelEischer, line 11: access to os.Stdout should be restricted to `cmd/restic` as far as possible. Otherwise we lose the central control about which parts of restic generate output or not.

Instead please use ui/termstatus.Terminal. That struct already supports updating multiple status lines, non-terminal outputs, too long lines, is far more efficient and avoids duplicating code.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
...
+	sync.Mutex
+	progressInfoMap map[string]progressInfoEntry
+	filesFinished   int64
+	filesTotal      int64
+	allBytesWritten int64
+	allBytesTotal   int64
+	started         time.Time
+}
+
+func formatBytesInBestUnit(inBytes int64) string {
```
MichaelEischer, line 35: This and the next two methods are obsolete once #3983 is merged.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
+package progressprinter
```
MichaelEischer, line 1: Please move the code to `internal/ui/restore`. I'd like to keep the ui code in one place.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
...
+var progresswriter io.Writer = os.Stdout
+
+// sets cursor back to start of line
+const lineResetter string = "\r"
+
+// add 3 spaces to fill line (when unit switches, the next line is 3 lines shorter which would leave 3 old chars otherwise)
+const lineFiller string = "   "
+
+type progressInfoEntry struct {
+	name         string
```
MichaelEischer, line 20: The `name` field appears to be unused.

internal/restorer/filerestorer.go
```diff
@@ -65,26 +68,32 @@ func newFileRestorer(dst string,
...
-		dst:         dst,
-		Error:       restorerAbortOnAllErrors,
+		key:             key,
+		idx:             idx,
+		packLoader:      packLoader,
+		filesWriter:     newFilesWriter(workerCount),
+		zeroChunk:       repository.ZeroChunk(),
+		sparse:          sparse,
+		progress:        verbosity >= 1,
+		progressPrinter: progressprinter.New(),
```
MichaelEischer, line 85: The progressPrinter *must* be passed down from within cmd_restore .

internal/restorer/filerestorer.go
```diff
@@ -65,26 +68,32 @@ func newFileRestorer(dst string,
...
-		workerCount: workerCount,
-		dst:         dst,
-		Error:       restorerAbortOnAllErrors,
+		key:             key,
+		idx:             idx,
+		packLoader:      packLoader,
+		filesWriter:     newFilesWriter(workerCount),
+		zeroChunk:       repository.ZeroChunk(),
+		sparse:          sparse,
+		progress:        verbosity >= 1,
```
MichaelEischer, line 84: Please replace this by a check whether a progressPrinter was passed.

changelog/unreleased/pull-3991
```diff
@@ -0,0 +1,7 @@
+Enhancement: Adds progress details while restoring
+
+The command restic restore now shows a progress report while it's restoring files.
+
+Example: [0:42]  23 / 3456 Files,  12.98 MiB / 23.54 GiB,  5.76 %   
+
+https://github.com/restic/restic/pull/3991
```
MichaelEischer, line 7: The changelog should also include links to relevant issues and if necessary, forum posts.

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (p *RestoreProgressPrinter) LogProgress(name string, bytesWrittenPortion int64, bytesTotal int64) {
+	p.Lock()
+	defer p.Unlock()
+	entry, exists := p.progressInfoMap[name]
+	if !exists {
+		entry.name = name
+		entry.bytesTotal = bytesTotal
+	}
+	entry.bytesWritten = entry.bytesWritten + bytesWrittenPortion
```
MichaelEischer, line 102: Should use `+=`

internal/restorer/progressprinter/progressprinter.go
```diff
@@ -0,0 +1,113 @@
...
+	p.progressInfoMap[name] = entry
+	p.allBytesWritten += bytesWrittenPortion
```
MichaelEischer, line 104: Please add a space after here, to group the modifications of the `entry` and the general statistics.

MichaelEischer (changes requested):
Thanks for working on this feature!

However, the code needs a few structural changes and the removal of duplicated code/functionality. We already have code to implement sophisticated status outputs in the form of `termstatus.Terminal` which should also be used as basis here, see the comments for more details.

I'd also like to centralize all hardcoded uses of os.Stdout to the `cmd/restic` package. Everything inside `internal` can be seen as a library and thus shouldn't use `stdout` by itself. For this PRs this means that like the `backup` command, stdout should be passed from `cmd_restore.go` to the ui printing code, instead of just grabbing and using it.

mark-herrmann:
Thankyou for your feedback. 
I am working on your suggestions. I will uncheck "I'm done! This pull request is ready for review." until i'm done with all improvements.

internal/restorer/filerestorer.go
```diff
@@ -50,10 +52,13 @@ type fileRestorer struct {
...
-	filesWriter *filesWriter
-	zeroChunk   restic.ID
-	sparse      bool
+	workerCount       int
+	filesWriter       *filesWriter
+	zeroChunk         restic.ID
+	sparse            bool
+	progressFormatter *progressformatter.RestoreProgressFormatter
+	terminal          *termstatus.Terminal
+	printProgress     bool
```
MichaelEischer, line 61: With the comments from above, both `terminal` and `printProgress` should no longer be necessary.

internal/restorer/filerestorer.go
```diff
@@ -268,7 +278,15 @@ func (r *fileRestorer) downloadPack(ctx context.Context, pack *packInfo) error {
 						file.inProgress = true
 						createSize = file.size
 					}
-					return r.filesWriter.writeToFile(r.targetPath(file.location), blobData, offset, createSize, file.sparse)
+					writeErr := r.filesWriter.writeToFile(r.targetPath(file.location), blobData, offset, createSize, file.sparse)
+
+					if r.printProgress {
+						bytesWrittenPortion := int64(len(blobData))
+						progress := r.progressFormatter.FormatProgress(file.location, bytesWrittenPortion, file.size)
+						r.terminal.SetStatus([]string{progress})
```
MichaelEischer, line 286: setting the status belongs into the progressFormatter. The fileRestorer should only report what has changed. But the task of transforming that into a status output belongs into the progressFormatter. Thus, please change that back to your initial implementation.

cmd/restic/cmd_restore.go
```diff
@@ -141,7 +176,7 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions, a
 		return err
 	}
 
-	res := restorer.NewRestorer(ctx, repo, sn, opts.Sparse)
+	res := restorer.NewRestorer(ctx, repo, sn, opts.Sparse, progressFormatter, term)
 
 	totalErrors := 0
 	res.Error = func(location string, err error) error {
```
MichaelEischer, line 182: all calls to `Warnf` / `Verbosef` inside `runRestore` must use the `term` instead, otherwise we'll get a garbled output. (Take a look at `progressPrinter.P` and `progressPrinter.V` in cmd_backup.go)

mark-herrmann:
Sorry, i forget to fixed the other tests.
Just commited a fix. Tests can be run again

internal/ui/restore/progressformatter/progressformatter.go
```diff
@@ -0,0 +1,119 @@
+package progressformatter
```
MichaelEischer, line 1: Please remove the intermediate `progressformatter` package and place the file directly into `internal/ui/restore`. To import a package with a different name you could use something similar to  `import rtest "github.com/restic/restic/internal/test"`. Just add the necessary interface directly to `internal/restore`.

changelog/unreleased/pull-3991
```diff
@@ -0,0 +1,11 @@
+Enhancement: Adds progress details while restoring
+
+The command restic restore now shows a progress report while it's restoring files.
+
+Example: [0:42]  23 / 3456 Files,  12.98 MiB / 23.54 GiB,  5.76 %   
+
+Fixes issue: https://github.com/restic/restic/issues/3627
```
MichaelEischer, line 7: The convention is to just include the URLs:
```
https://github.com/restic/restic/issues/426
https://github.com/restic/restic/pull/3991
https://forum.restic.net/t/progress-bar-for-restore/5210
```
(the correct issue is #426). If there's an issue to refer to, then that should be used to derive the name for the changelog file, thus please rename it to `issue-426`.

MichaelEischer (changes requested):
It looks like I should have been a bit more specific how I imagined `termstatus.Terminal` to be used, sorry for that. My idea was to replace the hard-coded `os.Stdout` with usage of a Terminal instance that gets passed to the ProgressPrinter on initialization (similar to what is done in the backup command). That way the restorer package still just has to report what is going on, and the ProgressPrinter is then responsible to turn that into the user-visible output. Returning the formatted progress and then setting the status inside the fileRestorer just increases the coupling between the actual restore part and the ui output.

Feel free to ask if you have questions on my comments.

internal/ui/restore/progressformatter/progressformatter.go
greatroar: So something like

    import restoreui "github.com/restic/restic/internal/ui/restore"

If you also rename the RestoreProgressFormatter to ProgressFormatter (or even just Progress), the name is less repetitive: restoreui.Progress(Formatter).

cmd/restic/cmd_restore.go
```diff
@@ -31,7 +35,37 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runRestore(cmd.Context(), restoreOptions, globalOptions, args)
+		var progressFormatter *progressformatter.RestoreProgressFormatter
+		var term *termstatus.Terminal
+		if !globalOptions.Quiet {
```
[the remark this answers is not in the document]
mark-herrmann: Why only globalOptions.Quiet is passed to termstatus.New (in cmd_backup.go for example), if it should also be disabled on  globalOptions.JSON. Please clarify

cmd/restic/cmd_restore.go
```diff
@@ -31,7 +35,37 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
...
+			progressFormatter = progressformatter.NewFormatter()
+			term = termstatus.New(globalOptions.stdout, globalOptions.stderr, globalOptions.Quiet)
```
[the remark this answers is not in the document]
mark-herrmann: This was my first aproach. But then with given -q the statusLine was printed anyways, even worse it was printed in new lines each step. Is this a bug? Is a disabled terminal really intended to do nothing? I think this is not working.
Any ideas?

cmd/restic/cmd_restore.go
```diff
@@ -31,7 +35,37 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 `,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runRestore(cmd.Context(), restoreOptions, globalOptions, args)
+		var progressFormatter *progressformatter.RestoreProgressFormatter
+		var term *termstatus.Terminal
+		if !globalOptions.Quiet {
+			progressFormatter = progressformatter.NewFormatter()
```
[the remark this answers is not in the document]
mark-herrmann: Honestly i would prefer this, too,
but this complicates testing.
Any ideas to keep testing as simple as possible?

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: The JSON output of `backup` command doesn't call `SetStatus` for the `Terminal`. Thus it doesn't really matter whether `.JSON` is also passed to `termstatus.New`. The basic idea in `backup` seems to be that the JSON option only changes the output format and not so much the behavior regarding interactive/non-interactive outputs (terminal vs. files). (Yes, the output handling is quite messy...)

Thus, not initializing the progressFormatter if the JSON option is set, should be enough to disable most output if JSON is set. Some other output statements probably need an `if !gopts.JSON {` check.

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: Hmm, I've confused the behavior of `term` with `internal/ui/backup/progress.go`, sorry. Only the latter implements the status output throttling and SIGUSR1 handling. That has to be factored out in a separate PR, thus let's leave the `term` creation as it is now.

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: I currently see two options:
- Create a ProgressPrinter (wraps the formatter) and ProgressFormatter struct. The printer then stores the reference to `term` and passes the output of the formatter to the terminal. That way the formatter remains unit testable, and the printer only contains code to wire up the components.
- Probably simpler: create an interface containing the `SetStatus(lines []string)` method and maybe others which would be required by the ProgressPrinter. Then during testing use a mock struct that implements just that interface.

mark-herrmann:
Hello,

To avoid misunderstandings and for clarity, i created a todo list.
Is this list correct and complete?
Otherwise, please reply with a corrected list.
Many Thanks

[ ] Move progressFormatter / progressPrinter instantiation and summary printing to end of runRestore
[ ] Disable output also, if globalOptions.JSON is set
[ ] Pass the term directly to the progressFormatter, similar to the backup command, converting the formatter to a printer.
    (Testing via mocked terminal)
[ ] all calls to Warnf / Verbosef inside runRestore must use the term instead, otherwise we'll get a garbled output.
[ ] Renaming and movement:
    [ ] Remove the intermediate progressformatter package and place the file directly into internal/ui/restore
    [ ] Rename the RestoreProgressFormatter to ProgressPrinter
[x] changelog
    [x] rename pr-3991 fileto issue-426
    [x] modify last lines (from line 7 to 11) to the following 3 lines
         https://github.com/restic/restic/issues/426
         https://github.com/restic/restic/pull/3991
         https://forum.restic.net/t/progress-bar-for-restore/5210
[x] Replace the formatter functions with the ones from pr-3983
[ ] Run:
    [ ] go test -cover ./...
    [ ] golangci-lint run --out-format=github-actions --new-from-patch=/tmp/tmp-1708-SvzJd2DygO2z/pull.patch --new=false --new-from-rev= --verbose --timeout 5m

MichaelEischer:
@mark-herrmann The todos sound like what I had in mind.

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: I've ended up with a Progress struct which manages the state and calls a ProgressPrinter with the necessary information. This is quite similar to the construction used for the backup command.

cmd/restic/cmd_restore.go
MichaelEischer: solved by overwriting `globalOptions.stdout` and `.stderr`. That way we can keep using `Warnf` and `Verbosef`.

MichaelEischer:
I've applied most review comments. I'm quite happy with the integration into the restorer. However, the progress printing is still a bit of a mess. The same holds for the tests.

mark-herrmann:
> I've applied most review comments. I'm quite happy with the integration into the restorer. However, the progress printing is still a bit of a mess. The same holds for the tests.

Any good ideas how to fix it. Maybe i can find time to do it in first week of january.

MichaelEischer:
Right now the progress bar is updated each time something changes, which is rather inefficient. With #4116 it should be possible to only periodically update the progress bar just like in the backup command.

Besides that, the progress tracking in `ui/restore.Progress` should be fine. The formatting probably should look a bit more like that of the `backup` command: `[{time}] {progress in percent}  {finishedFiles} files {finishedBytes}, total {totalFiles} files {totalBytes}`.

Regarding the tests I'm not particularly happy about the amount of implementation details which are accessed by the tests. My current impression is that right now every minor internal change of `Progress` will likely also require test changes. It's probably better to add a `ProgressPrinterSpy`, together with a `Progress.update(runtime time.Duration, final bool)` (the `UpdateFunc` type from #4116) method which will trigger a progress bar update (that is, calls `p.printer.Update(...)` or `Finish(...)`).

The tests for the output format are currently removed / commented out. To test them, we'd have to replace `*termstatus.Terminal` with a corresponding interface. But either way these tests should only test a few cases, otherwise we'll end up with large amounts of test changes when tweaking the output format.

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: updated to use the `progress.Updater` which is by now available as a separate component.

MichaelEischer:
I've rebased the PR, hooked up the progress.Updater code and updated the progress bar formatting.

The remaining tasks for this PR are better tests.

MichaelEischer:
I've reworked the tests. That also required a few changes to the `Progress` and the `textPrinter`. `Progress` now hides the Mutex and the Updater struct. Reporting the current progress is now done in an `update` method, which is mostly there to allow the tests to call it. That way the test doesn't have to wait until the update happens automatically.

I'll clean up the commits before merging this, but otherwise the PR should be in good enough shape to merge it.

MichaelEischer (approved):
LGTM. I've rebased the PR one last time and slightly tweaked the changelog entry.

PR #3992, opened 2022-10-29
Return error if RESTIC_COMPRESSION env variable is invalid

### LINKED ISSUE #3962: Invalid values of environment variable RESTIC_COMPRESSION are silently ignored
Output of `restic version`
--------------------------
```
```

```
export RESTIC_PASSWORD=SuperSecretPassword
export RESTIC_REPOSITORY='rest:https://someuser:somepassword@someserver.tld/someuser'
export RESTIC_COMPRESSION=bullshido
restic init --repository-version 2
restic backup --verbose --exclude-caches --tag="test" -o rest.connections=25 /somedir
```

```
rest-server version rest-server 0.11.0 compiled with go1.17.6 on linux/amd64
REST server is behind haproxy 2.4 which offloads TLS security.
```
Expected behavior
-----------------

using `RESTIC_COMPRESSION=<some invalid string>` should trigger the same as when run with `--compression=<some invalid string>`, eg:
```
invalid argument "bullshido" for "--compression" flag: invalid compression mode "bullshido", must be one of (auto|off|max)
```

Actual behavior
---------------

restic backups without complaining.
So there's no way to know whether compression was used, since it would just continue to run without a proper compression argument.

I hope you have everything required above ;)

?

Make `RESTIC_COMPRESSION` env variable tested the same way `--compression` argument is.

I'm the author of [backup-bench](https://github.com/deajan/backup-bench)
After a lot of benchmarking between all the solutions, I wanted to give restic a deeper test dive for it fills quite well my usecase.

MichaelEischer:
restic currently does not return an error is the environment variable `RESTIC_COMPRESSION` is set to an invalid value. Instead it was silently ignored.

The underlying problem is that there is no nice way to return an error while setting the compression mode based on the environment variable. This PR changes the CompressionMode type to add an additional error status which is checked when creating a repository. The downside is that the code paths to check the compression mode are different depending on whether the parameter is set via command line or environment variable.

An alternative would be to extend the CompressionMode type to store the error, but that just feels wrong.

Fixes https://github.com/restic/restic/issues/3962 .

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -67,9 +67,10 @@ type CompressionMode uint
 
 // Constants for the different compression levels.
 const (
-	CompressionAuto CompressionMode = 0
-	CompressionOff  CompressionMode = 1
-	CompressionMax  CompressionMode = 2
+	CompressionAuto    CompressionMode = 0
+	CompressionOff     CompressionMode = 1
+	CompressionMax     CompressionMode = 2
+	CompressionInvalid CompressionMode = 3
 )
 
 // Set implements the method needed for pflag command flag parsing.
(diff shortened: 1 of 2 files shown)


fd0 (approved):
It'll do, thanks :)

PR #3993, opened 2022-10-31
backup: print full snapshot id in JSON summary

### LINKED ISSUE #2724: Include full snapshot id in backup json command output
Output of `restic version`
--------------------------

`restic 0.9.6 compiled with go1.13.4 on linux/amd64`

Most commands that react to the global `--json` switch use the full snapshot id in their output
(at least `find`, `forget`, `ls` and `snapshots` do).
One notable difference seems to be the `backup` command whose summary output only contains the short snapshot id.

Example output:
```json
{
  "data_added": 25131953,
  "data_blobs": 118,
  "dirs_changed": 2,
  "dirs_new": 0,
  "dirs_unmodified": 0,
  "files_changed": 74,
  "files_new": 60,
  "files_unmodified": 11223,
  "message_type": "summary",
  "snapshot_id": "43deae25",
  "total_bytes_processed": 503143593,
  "total_duration": 1.816399114,
  "total_files_processed": 11357,
  "tree_blobs": 3
}
```

I would suggest to also provide a `full_snapshot_id` or similarly named field in the above output.
For consistency reasons identical or at least similar JSON keys would be nice to have but following `restic snapshots` output which uses `id` and `short_id` would be backwards incompatible.

What are you trying to do?
--------------------------

I'm trying to link the restic command output together for cases like:
- `restic backup` followed by `restic diff` to see which exact files were just added/modified/removed compared to a previous snapshot
- `restic backup` followed by `ls` to list all files that ended up in the newly added snapshot

In the above cases I don't intend to show the snapshot ids to the user but only use them internally. In that case always using the full snapshot id has a lower chance of running into collisions.

The short id is (like in git) probably meant to make life of a commandline end-user easier. When scripting or otherwise automating restic, the full id should probably be preferred.

I'm successfully using restic for three personal machines (two `$HOME  backups for desktop/laptop, one full backup for a server) by now and both the sheer speed as well as the well-documented commandline interface make it a breeze to use.
Being self-contained and only requiring sftp access for remote backups also made it easy to switch from my two old backup tools to restic :)

MichaelEischer:
The JSON output only included the shortened ID of the snapshot and not the full ID. The shortened ID could be ambiguous in rare cases and always has to be expanded first before restic can use it. As the JSON output is intended to be machine readable, there is no use in shortening the ID before printing it.

Strictly speaking, this change is not backwards compatible, but I don't expect much fall-out. In the worst case users have to explicitly truncate the ID down to 8 characters.

Fixes #2724

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2724
+++ b/changelog/unreleased/issue-2724
@@ -1,0 +1,9 @@
+Change: Include full snapshot ID in JSON output of the bacukp command
+
+We have changed the JSON output of the backup command to include the full snapshot ID instead of just a shortened version.
+The latter can be ambiguous in rare cases.
+
+To derive the short ID, truncate the full ID down to 8 characters.
+
+https://github.com/restic/restic/issues/2724
+https://github.com/restic/restic/pull/3993
--- a/internal/ui/backup/json.go
+++ b/internal/ui/backup/json.go
@@ -191,7 +191,7 @@ func (b *JSONProgress) Finish(snapshotID restic.ID, start time.Time, summary *Su
 		TotalFilesProcessed: summary.Files.New + summary.Files.Changed + summary.Files.Unchanged,
 		TotalBytesProcessed: summary.ProcessedBytes,
 		TotalDuration:       time.Since(start).Seconds(),
-		SnapshotID:          snapshotID.Str(),
+		SnapshotID:          snapshotID.String(),
 		DryRun:              dryRun,
 	})
 }


rawtaz (approved):
Makes sense, no need to shorten it in programmatic/structured output.

MichaelEischer:
Hmm, the `forget` and `snapshots` command print both the full `ID` and a `ShortID`. But introducing a `LongID` here is just horribly inconsistent and not really useful.

rawtaz:
The `forget` command prints snapshot IDs like this:
~~~
"id": "89cefc5d61f95b66dc36e25029d54582fdb5d92d159af95dbb89cd4458f3c4a5",
"short_id": "89cefc5d"
~~~

The `snapshots` command prints snapshot IDs like this:
~~~
"id": "ea55667d08c38695cc6d2367ac50f9bc66b7b88a946427dcadde76b806820b07",
"short_id": "ea55667d"
~~~

The `backup` command current prints snapshot IDs like this:
~~~
"snapshot_id": "ea55667d"
~~~

We can simply change the `backup` command to output the snapshot ID as `id` (long format) and `short_id` (short format) to make it consistent with the others, while for now keeping `snapshot_id` (but making a note that it is deprecated in the changelog and the code) and remove that in the future. This should make it all consistent and fully backwards compatible.

fd0 (approved):

rawtaz:
@fd0 Did you read my comment above?

MichaelEischer:
In the context of the backup command, it's probably not really obvious, that `id` would be the snapshot id. And replacing the existing field complicates things a lot.

fd0:
I'm with @MichaelEischer, just print the full ID in the JSON field.

PR #3997, opened 2022-11-02
fuse: Better check for whether snapshots changed

greatroar:

We previously checked whether the set of snapshots might have changed based only on their number, which fails when as many snapshots are forgotten as are added. Check for the SHA-256 of their id's instead.

No.

Checklist
---------

### DIFF
--- a/internal/fuse/snapshots_dirstruct.go
+++ b/internal/fuse/snapshots_dirstruct.go
@@ -4,6 +4,7 @@
 package fuse
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"path"
@@ -14,6 +15,8 @@ import (
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/restic"
+
+	"github.com/minio/sha256-simd"
 )
 
 type MetaDirData struct {
@@ -39,7 +42,7 @@ type SnapshotsDirStructure struct {
 	// that way we don't need path processing special cases when using the entries tree
 	entries map[string]*MetaDirData
 
-	snCount   int
+	hash      [sha256.Size]byte // Hash at last check.
 	lastCheck time.Time
 }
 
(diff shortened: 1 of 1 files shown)


greatroar:
Test failure is something relating to locking.

internal/fuse/snapshots_dirstruct.go
```diff
@@ -302,10 +303,20 @@ func (d *SnapshotsDirStructure) updateSnapshots(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	// sort snapshots ascending by time (default order is descending)
+
+	// Sort snapshots ascending by time (default order is descending).
+	// This needs to be done before hashing.
 	sort.Sort(sort.Reverse(snapshots))
```
MichaelEischer, line 309: How are two snapshots with identical timestamps ordered? afaict that order isn't stable?

internal/fuse/snapshots_dirstruct.go
```diff
@@ -302,10 +303,20 @@ func (d *SnapshotsDirStructure) updateSnapshots(ctx context.Context) error {
...
-	if d.snCount == len(snapshots) {
+	// We update the snapshots when the hash of their id's changes.
+	h := sha256.New()
+	for _, sn := range snapshots {
+		h.Write(sn.ID()[:])
+	}
+	var hash [sha256.Size]byte
+	h.Sum(hash[:0])
+
+	if d.lastCheck != (time.Time{}) && d.hash == hash {
```
MichaelEischer, line 319: Just checking whether the hash matches would probably be enough. Otherwise we'd have a hash collision.

greatroar: Yes, we'd need to have a set of snapshots with an all-0 SHA2. Removed the timestamp check.

internal/fuse/snapshots_dirstruct.go
greatroar: Good point. Fixed to sort by timestamp, then hash.

MichaelEischer (approved):
LGTM.

Although the more detailed check only makes a difference when using `mount --no-lock`. Otherwise, we can only add new snapshots, which will necessarily change the snapshot count.

PR #4006, opened 2022-11-04
archiver: only store deviceID for hardlinks

### LINKED ISSUE #3041: Restic 0.10.0 always reports all directories "changed", adds duplicate metadata, when run on ZFS snapshots

MichaelEischer:
he deviceID can change e.g. when backing up from a filesystem snapshot. However, it is only used for hardlink detection. Thus, it is not necessary to store it for non-hardlinks. This should handle most use cases without adding much complexity.

Solves part of #3041
Replaces #3599

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4006
+++ b/changelog/unreleased/pull-4006
@@ -1,0 +1,16 @@
+Enhancement: (alpha) Store deviceID only for hardlinks
+
+Set `RESTIC_FEATURES=device-id-for-hardlinks` to enable this alpha feature.
+The feature flag will be removed after repository format version 3 becomes
+available or be replaced with a different solution.
+
+When creating backups from a filesystem snapshot, for example created using
+btrfs subvolumes, the deviceID of the filesystem changes compared to previous
+snapshots. This prevented restic from deduplicating the directory metadata of
+a snapshot.
+
+When this alpha feature is enabled, then the deviceID is only stored for
+hardlinks. This significantly reduces the metadata duplication for most
+backups.
+
+https://github.com/restic/restic/pull/4006
--- a/changelog/unreleased/pull-4503
+++ b/changelog/unreleased/pull-4503
@@ -4,4 +4,5 @@ If files on different devices had the same inode id, then the `stats` command
 did not correctly calculate the snapshot size. This has been fixed.
 
 https://github.com/restic/restic/pull/4503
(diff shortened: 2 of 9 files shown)


greatroar:
This does make it impossible to determine where the mount points were from inspecting the snapshot...

MichaelEischer:
> This does make it impossible to determine where the mount points were from inspecting the snapshot...

Is that information that useful? I figure that without knowing what filesystem was used where, that information is not particularly useful.

I did some tests to see where hardlinks are used, so far I've seen usages in `/usr/share/{terminfo,zoneinfo}`, in Docker/podman images, the steam runtime, flatpaks and a few random other usages. For me these amount to about 3% of all files, such that this PR is less effective than I'd hoped for.

greatroar:
> Is that information that useful? I figure that without knowing what filesystem was used where, that information is not particularly useful.

No, probably not.

JoeKun:
If I understand correctly, this small change could substantially improve the usability of `restic` for users who want to backup from a file-system snapshot, such as a ZFS snapshot.

This sounds like a very desirable improvement.

@MichaelEischer Could you please clarify what is missing before we could potentially see this change merged?

MichaelEischer:
The problem with this change is that it will cause the next backup to create different tree blobs. This can double the size of the metadata stored for a backup.

My plan is to add support for large directories in restic 0.18 (see https://forum.restic.net/t/roadmap-for-restic-0-17-to-0-19/7197) which will also require changes to the tree format. That provides a good opportunity to also include this PR.

MichaelEischer:
Slight change of plans. Now that we have feature flags, we can immediately merge this PR, but hide it behind a feature flag for now (`RESTIC_FEATURES=device-id-for-hardlinks`). The deviceID handling will either become the default in repository version 3 (restic 0.18) or be completely replaced. Either way, the feature flag will be removed some time afterwards.

MichaelEischer (commented):
LGTM

hettiger:
Thank you @MichaelEischer 

this also fixed backing up from APFS-snapshots for me.

ilyagr:
I also just wanted to say thank you, this helped me a lot on my Mac. ❤️

Issue #3828, opened 2022-07-16
dump and cp from mounted filesystem is about 8 times slower than restore, on Windows dump command runs as fast as restore


JsBergbau:
Output of `restic version`
--------------------------
restic 0.13.1 compiled with go1.18 on linux/amd64

How did you run restic exactly?
-------------------------------
`theFile` is about 2522 MB
time ./restic dump --cacert key latest theFile > temp.txt

results in 
```
real    1m3,982s
user    0m44,149s
sys     0m7,611s
```

time ./restic restore --cacert key latest --iinclude theFile -t restoretest
results in 
```
real    0m7,994s
user    0m37,991s
sys     0m4,723s
```

mounting and then using

time cp theFile /target

results in
```
real    1m9,124s
user    0m0,032s
sys     0m4,560s
```

Using cat 
```
real    1m9,018s
user    0m0,021s
sys     0m4,752s
```

system is pretfy fast using
`time cat theFile > theFile2`

results in
```
real    0m1,352s
user    0m0,004s
sys     0m1,347s
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
rest-server on the same system as restore was tested

Expected behavior
-----------------
At least restic dump should be almost fast as restore command.
For the mounted filesystem probably fuse is the culprit, since filesystem is running in userspace. But still that is quite slow and I'm quite sure there could be some optimizations done. If this is possible this is from my point of view much more helpful than for dump command. 

On windows I did a test with a slower rest-backend, needing about 30 seconds to restore a 850 MB file. There with `restic 0.13.1 (v0.13.0-251-g98a3125c) compiled with go1.18.1 on windows/amd64` dump and restore showed no significant difference in speed. Using also `restic 0.13.1 (v0.13.0-251-g98a3125c) compiled with go1.18.1 on linux/amd64` on the linux system with the repo with the 2522 MB filedump was as slow as with the current 0.13.1 release version.

Actual behavior
---------------
dump command and copying from mounted file system are very slow. 

Steps to reproduce the behavior
-------------------------------
Create a repository, add a big file and restore or dump.

Do you have any idea what may have caused this?
-----------------------------------------------
Since there is difference in the behaviour of the dump command in Windows and Linux, it might be go runtime environment related? 
Using mounted repo slow speed might be a limitation of fuse file system since it is in user space and has to to a lot of context switches?

Do you have an idea how to solve the issue?
-------------------------------------------
No.

Did restic help you today? Did it make you happy in any way?
[...]

PR #4013, opened 2022-11-08
fuse: Concurrent readahead

greatroar:

Add concurrent readahead to fuse.openFile.Read. Every read call now spans a goroutine (up to some memory limit) that reads the first block after the section it is told to read, in the expectation that the next call will need that block.

I ran two small benchmarks on a compressed repository, using a local on-disk cache.

Benchmark one: sha256sum on a 753MiB file. Best of three runs, after a single run to warm up the caches.

    # 59a90943bb0c780fa69fafe752b2be4a98014397
    real    0m17.266s
    user    0m5.255s
    sys     0m0.391s

    # This PR
    real    0m13.808s
    user    0m4.662s
    sys     0m0.298s

That's -20%, or a 25% speed increase.

Benchmark two: sha256sum of the same file in parallel with a 703MiB file, with which it shares no blocks. This simulates the habit of some file managers of copying two files concurrently when copying a directory.

    # 59a90943bb0c780fa69fafe752b2be4a98014397
    real    0m15.789s
    user    0m8.846s
    sys     0m0.676s

    # This PR
    real    0m13.643s
    user    0m8.776s
    sys     0m0.606s

That's -14%, or a 16% speed increase.

Updates #3828. I wouldn't say it solves that issue, as more tweaking may be possible, but this is a good start.

Checklist
---------


greatroar:
Test failure looks unrelated. I just ran archiver.TestArchiverAbortEarlyOnError 1000 times and got no failure.

greatroar:
I also evaluated a simpler implementation of this idea where a single background goroutine gets sent an ID to pre-load over a channel (w/ a singleflight.Group, and with the sender dropping the ID if the channel is full). That gives no speedup at all.

MichaelEischer:
I'm not really convinced that implementing our own readahead strategy is a good idea. The kernel should be in a much better place to make a good decision.

The fuse library has an option which allows the kernel to asynchronously issue read request: `systemFuse.AsyncRead()` (would have to be added to `cmd_mount.go`). I wonder whether that could also help with performance. However,  at least for my tests the answer seems to be no.

PR #4020, opened 2022-11-11
fuse: Better inode generation

greatroar:

Hard links to the same file now get the same inode within the FUSE mount. #4006 and backups from Windows are taken into account by using different hashing schemes for hard-linked files and everything else.

Also, inode generation is faster and, more importantly, no longer allocates.

Benchmarked on Linux/amd64. Old means the benchmark with

    sink = fs.GenerateDynamicInode(1, sub.node.Name)

instead of calling inodeFromNode. I don't get why this allocates in one case but not the other. fs.GenerateDynamicInode always calls runtime.stringtobyteslice. Results:

```
name                   old time/op    new time/op    delta
Inode/no_hard_links-8     137ns ± 4%      34ns ± 1%   -75.20%  (p=0.000 n=10+10)
Inode/hard_link-8        33.6ns ± 1%     9.5ns ± 0%   -71.82%  (p=0.000 n=9+8)

name                   old alloc/op   new alloc/op   delta
Inode/no_hard_links-8     48.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
Inode/hard_link-8         0.00B          0.00B           ~     (all equal)

name                   old allocs/op  new allocs/op  delta
Inode/no_hard_links-8      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
Inode/hard_link-8          0.00           0.00           ~     (all equal)
```

No.

Checklist
---------

### DIFF
--- a/internal/fuse/dir.go
+++ b/internal/fuse/dir.go
@@ -182,7 +182,7 @@ func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 		}
 
 		ret = append(ret, fuse.Dirent{
-			Inode: fs.GenerateDynamicInode(d.inode, name),
+			Inode: inodeFromNode(d.inode, node),
 			Type:  typ,
 			Name:  name,
 		})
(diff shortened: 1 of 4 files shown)


greatroar:
There's TestArchiverAbortEarlyOnError again.

MichaelEischer:
I have to admit I'm somewhat confused by the inode handling. My assumption was that different files should result in different inodes. However, judging from `fs.GenerateDynamicInode` that's not necessarily the case. What worries me is that e.g. 100 snapshots with 40million files each (given the repository sizes that have been mentioned every now and then that's a realistic repository size) result in a 50% chance of an inode collision. Do you know which identifier is internally used in the fuse library to track files/directories?

greatroar:
Not the inode number, AFAICT. I just rigged inodeFrom(Name|Node) to always return 2, and the mount point still worked fine, with sha256sum distinguishing between the files properly.

greatroar:
I just realized that with this change, hard-linked files may also share their inode across snapshots. Personally, I find that acceptable, but if it isn't, we need to hash in the snapshot id as well. XOR with the first 64 bits of that id [should be enough](https://security.stackexchange.com/a/34797).

greatroar:
Looks like fs.GenerateDynamicInode doesn't allocate for one of these filenames because it's less than [tmpstringbufsize = 32](https://github.com/golang/go/blob/b1678e508bf04b32fcd8153d09c39ff25b51d287/src/cmd/compile/internal/walk/walk.go#L21), in which case the byte slice for it is allocated on the stack.

MichaelEischer (approved):
LGTM.

Looks like I've completely misunderstood how fuse identifies its direntries. I always thought that the inode had something to do with that, but that is just cosmetic. Instead fuse uses a separate `nodeid` (see https://manpages.debian.org/bullseye/manpages/fuse.4.en.html) which must be unique within the filesystem. For us bazil/fuse already takes care of that. inode = 0 causes the library to automatically calculate a replacement inode.

Thus, it's fine to assign inode in whatever way we like. There's just a recommendation somewhere that different files should have different inodes in order to not confuse everyone.

I have no strong opinion on whether inodes for hardlinked files from different snapshots should be different or not. Let's just keep the current solution for now. We can still complicate it later on.

PR #4024, opened 2022-11-12
mount: switch to anacrolix fork of bazil/fuse

### LINKED ISSUE #3096: macOS: drop fuse support (restic mount)
The library we're using for fuse (https://github.com/bazil/fuse) [dropped support for macOS entirely](https://github.com/bazil/fuse/issues/224) a while back because the source code for osxfuse is [not available any more](https://github.com/osxfuse/osxfuse/issues/590). For now, we've pinned the last working version so `restic mount` still works.

In  #3094, I've disabled the fuse tests on macOS because they fail randomly.

Some thoughts:
 * At some point I'd like to update the `fuse` again, then restic on macOS will not be able to use the `mount` command any more
 * It feels to me that macOS will lose the ability to load custom kernel modules (osxfuse is a kernel module) sooner than later
 * Until now restic consists only of Go code, we don't need `cgo` for building. That makes the build process (for us as well as for our users) **much** simpler and we haven't had any issues.

My idea would be to refactor the fuse code and implement a simple webserver view as well as a WebDAV server (see #485) as an alternative to fuse. This way, also users on Windows benefit.

MichaelEischer:
Replace the [bazil/fuse](https://github.com/bazil/fuse) library with the [anacrolix](https://github.com/anacrolix/fuse) fork.

The anacrolix fork contains the latest changes from bazil/fuse and additionally provides support for recent versions of macFUSE.

Unfortunately, there seems to be no actively maintained fork of bazil/fuse, such that we'll likely have to switch to a different fuse library sooner or later.

Fixes #3096

Checklist
---------


enj (approved):
Should `test_fuse` be set to `true` for macOS?

MichaelEischer:
I've enabled the fuse tests for macOS. Let's see whether these work reliably.

MichaelEischer:
Hmm, the fuse tests failed as macFUSE is not available. According to https://github.com/actions/runner-images/issues/5583#issuecomment-1196763627 this cannot be fixed using the official CI Runners. So, I guess this will just have to work without test coverage :-( .

anacrolix:
american-psycho-patrick-bateman

PR #4067, opened 2022-12-03
Remove `Test()` method from Backend

MichaelEischer:
The `Test()` method was only used in exactly one place, namely when trying to create a new repository to check whether the `config` file already exists. This PR replaces it with a combination of `Stat()` and `IsNotExist()` instead. As a result this allows removing quite a bit of code.

The retry behavior of `Stat()` is different from `Test()`. Thus change the RetryBackend to not retry "file not found" errors for the Stat() method. In non test/debug code, `Stat()` is used exclusively to check whether a file exists. Thus, this won't cause unintended side-effects.

No.

Checklist
---------

### DIFF
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -1873,10 +1873,6 @@ type noCancelBackend struct {
 	restic.Backend
 }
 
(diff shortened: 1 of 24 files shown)


greatroar:
LGTM.

PR #4074, opened 2022-12-04
doc: Remove ref to Lobaro's Docker image

### LINKED ISSUE #4002: doc: Remove reference to lobaro/restic-backup-docker ?
The documentation [references](https://restic.readthedocs.io/en/stable/020_installation.html#docker-container) the Docker image at https://github.com/lobaro/restic-backup-docker , but this one is quite outdated and doesn't seem to be getting any attention by the owner of the repository, despite questions from people.

While it indeed seems like a nice image, I think we should remove the reference from the documentation if it's outdated. What do others think?

greatroar:
Fixes #4002.

### DIFF
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -265,11 +265,6 @@ binary, you can get it with `docker pull` like this:
 
     $ docker pull restic/restic
 
-.. note::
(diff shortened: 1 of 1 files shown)


rawtaz (approved):
Thank you!

PR #4081, opened 2022-12-10
Add support for `--group-by` to the backup command

### LINKED ISSUE #3941: --group_by for backup parent selection
Output of `restic version`
--------------------------

Allow different snapshot groupings to find the best backup parent.

For the "backup" command, if an explicit parent is not specified with --parent, restic will choose the newest snapshot with the same paths and host as the current job.

This does not always find the best parent. For my (common?) use case, I have two backup sets on the same host:

full: path /a /b, no excludes
warm: path /a /b, exclude /a/archive/* /a/huge_files/* etc.

The full backup is run once a week, and warm backups on the other six days. Current parent selection will always pick the newest snapshot (since the host and paths always match), which for the full backup will be a warm snapshot, which is suboptimal since many files in full are not in warm, and will have to be rehashed.

I currently run a separate `snapshots --tag full` cmd to find the most recent snapshot by tag and then specify with --parent for backup, but this seems like it would be better integrated into the program itself.

see:

https://forum.restic.net/t/parent-selection-using-tags/5443

Add --group_by (or a better name?) to the backup command for configurable parent selection. It would have the same semantics as forget's --group_by argument.

In my use case described above, I would use --group_by host,tag to backup just as I do for my forget invocations.

Good backups help me sleep better every night! Better sleep does result in happiness, so yes, in a way.

Although it wasn't today, a few weeks back I deleted an old photo directory thinking I had copied everything elsewhere, without verifying that I had actually done the copy. An hour later, I began to doubt myself, so I mounted the latest snapshot and made sure each file from the deleted directory existed elsewhere.

### LINKED ISSUE #4189: The description of `--parent` in the backup help is incorrect.
Output of `restic version`
--------------------------
restic 0.15.1 compiled with go1.19.5 on windows/amd64

In the output of `restic backup -h`, I found this description is not correct: `--parent snapshot                        use this parent snapshot (default: last snapshot in the repository that has the same target files/directories, and is not newer than the snapshot time)`. After testing, I think it should be `(default: last snapshot in the repository that has the same target files/directories`**` and the same hostname`**`, and is not newer than the snapshot time)`.

I have backed up several times on the same Windows computer using the same paths. And then I changed the Windows device name. And then the same backup command (`restic --repository-file <the same repository path file> -v --compression max backup --files-from <the same backup path file> --exclude-file <the same exclusive file> --no-scan --use-fs-snapshot`) said "no parent snapshot found", hence the long re-scan that followed. I manually used `--parent snapshot ID of the latest`, hence no full re-scan just as before I changed the device name.

MichaelEischer:
The backup command by default selected the parent snapshot based on the hostname
and the backup targets. When the backup path list changed, the backup command
was unable to determine a suitable parent snapshot and had to read all
files again. A similar scenario applies when using different sets of excludes while keeping the hostname and paths unchanged.

The new `--group-by` option for the backup command allows filtering snapshots
for the parent selection by `host`, `paths` and `tags`. It defaults to
`host,paths` which selects the latest snapshot with hostname and paths matching
those of the backup run.

I'm not entirely sure about the option name. The current name is `--group-by` as the mechanism works just like the snapshot grouping in the `forget` or `snapshots` command. But something like `--parent-group-by` or `--parent-filter-by` might be more intuitive.

Fixes #3941
Fixes #4189

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -89,6 +89,7 @@ type BackupOptions struct {
 	excludePatternOptions
 
 	Parent            string
+	GroupBy           restic.SnapshotGroupByOptions
 	Force             bool
 	ExcludeOtherFS    bool
 	ExcludeIfPresent  []string
@@ -121,6 +122,8 @@ func init() {
 
 	f := cmdBackup.Flags()
 	f.StringVar(&backupOptions.Parent, "parent", "", "use this parent `snapshot` (default: last snapshot in the repository that has the same target files/directories, and is not newer than the snapshot time)")
+	backupOptions.GroupBy = restic.SnapshotGroupByOptions{Host: true, Path: true}
+	f.VarP(&backupOptions.GroupBy, "group-by", "g", "`group` snapshots by host, paths and/or tags, separated by comma (disable grouping with '')")
 	f.BoolVarP(&backupOptions.Force, "force", "f", false, `force re-reading the target files/directories (overrides the "parent" flag)`)
 
 	initExcludePatternOptions(f, &backupOptions.excludePatternOptions)
@@ -439,7 +442,21 @@ func findParentSnapshot(ctx context.Context, repo restic.Repository, opts Backup
 	if snName == "" {
 		snName = "latest"
 	}
-	sn, err := restic.FindFilteredSnapshot(ctx, repo.Backend(), repo, []string{opts.Host}, []restic.TagList{}, targets, &timeStampLimit, snName)
+
+	var hosts []string
+	var paths []string
+	var tags []restic.TagList
+	if opts.GroupBy.Host {
+		hosts = []string{opts.Host}
+	}
+	if opts.GroupBy.Path {
+		paths = targets
+	}
+	if opts.GroupBy.Tag {
+		tags = []restic.TagList{opts.Tags.Flatten()}
+	}
+
+	sn, err := restic.FindFilteredSnapshot(ctx, repo.Backend(), repo, hosts, tags, paths, &timeStampLimit, snName)
 	// Snapshot not found is ok if no explicit parent was set
 	if opts.Parent == "" && errors.Is(err, restic.ErrNoSnapshotFound) {
 		err = nil
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -121,7 +121,7 @@ func init() {
 	cmdRoot.AddCommand(cmdBackup)
 
 	f := cmdBackup.Flags()
-	f.StringVar(&backupOptions.Parent, "parent", "", "use this parent `snapshot` (default: last snapshot in the repository that has the same target files/directories, and is not newer than the snapshot time)")
+	f.StringVar(&backupOptions.Parent, "parent", "", "use this parent `snapshot` (default: latest snapshot in the group determined by --group-by and not newer than the timestamp determined by --time)")
 	backupOptions.GroupBy = restic.SnapshotGroupByOptions{Host: true, Path: true}
 	f.VarP(&backupOptions.GroupBy, "group-by", "g", "`group` snapshots by host, paths and/or tags, separated by comma (disable grouping with '')")
 	f.BoolVarP(&backupOptions.Force, "force", "f", false, `force re-reading the target files/directories (overrides the "parent" flag)`)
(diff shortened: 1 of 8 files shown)


doc/040_backup.rst
```diff
@@ -146,6 +146,15 @@ change detection rule based on file metadata to determine whether a file is
 likely unchanged since a previous backup. If it is, the file is not scanned
 again.
 
+The previous backup, called parent snaphot in restic terminology, is
+determined as follows. By default restic selects the last backup using the
+same hostname and backup paths. You can change the selection criteria using
```
greatroar, line 151: It also uses the exclude list, right?

MichaelEischer: No. The exclude list isn't used anywhere for grouping snapshots.

MichaelEischer:
I've slightly changed the documentation to make it a bit more clearer that `backup --group-by` and `forget --group-by` should be used together and that both revolve around the concept of a "group of snapshots".

doc/060_forget.rst
```diff
@@ -219,6 +219,7 @@ paths and tags. The policy is then applied to each group of snapshots individual
 This is a safety feature to prevent accidental removal of unrelated backup sets. To
 disable grouping and apply the policy to all snapshots regardless of their host,
 paths and tags, use ``--group-by ''`` (that is, an empty value to ``--group-by``).
+The ``--group-by`` option should also be set accordingly for the ``backup`` command.
```
rawtaz, line 222: Honestly, I think that it is *normal* for these two to be set equally, but it's not *always* what you want to do. So maybe make this:
~~~
Note that one would normally set the ``--group-by`` option for the ``backup`` command to the same value.
~~~

doc/040_backup.rst
```diff
@@ -146,6 +146,16 @@ change detection rule based on file metadata to determine whether a file is
...
+The previous backup, called parent snaphot in restic terminology, is
+determined as follows. By default restic groups snapshots by hostname and backup
+paths, and then selects the latest snapshot which matches the current backup.
+You can change the selection criteria using the `--group-by`` option, which
+defaults to ``host,paths``. To select the latest snapshot with the same paths
+independent of the hostname, use ``paths``. Or to only consider the hostname
+and tags, use ``host,tags``. Alternatively, it is possible to manually specify
+a specific snapshot using the ``--parent snapshotID`` option. The option should
+also be set to the same value for the ``forget`` command.
+
```
rawtaz, line 158: Some minor changes, but I suggest changing this *and the above* paragraph to the following:

~~~
When restic encounters a file that has already been backed up, whether in the
current backup or a previous one, it makes sure the file's contents is only
stored once in the repository. To do so, it normally has to scan the entire
contents of the file. Because this can be very expensive, restic also uses a
change detection rule based on file metadata to determine whether a file is
likely unchanged since a previous backup. If it is, the file is not scanned
again.

The previous backup snapshot, called "parent" snaphot in restic terminology,
is determined as follows. By default restic groups snapshots by hostname and
backup paths, and then selects the latest snapshot in the group that matches
the current backup. You can change the selection criteria using the
``--group-by`` option, which defaults to ``host,paths``. To select the latest
snapshot with the same paths independent of the hostname, use ``paths``. Or,
to only consider the hostname and tags, use ``host,tags``. Alternatively, it
is possible to manually specify a specific parent snapshot using the
``--parent`` option. Finally, Note that one would normally set the
``--group-by`` option for the ``forget`` command to the same value.
~~~

cmd/restic/cmd_backup.go
```diff
@@ -120,7 +121,9 @@ func init() {
 	cmdRoot.AddCommand(cmdBackup)
 
 	f := cmdBackup.Flags()
-	f.StringVar(&backupOptions.Parent, "parent", "", "use this parent `snapshot` (default: last snapshot in the repository that has the same target files/directories, and is not newer than the snapshot time)")
+	f.StringVar(&backupOptions.Parent, "parent", "", "use this parent `snapshot` (default: latest snapshot in the current group of snapshots which is not newer than the snapshot time)")
```
rawtaz, line 124: Suggest changing to:
~~~
use this parent `snapshot` (default: latest snapshot in the group determined by --group-by and not newer than the timestamp determined by --time)
~~~

rawtaz (changes requested):

doc/040_backup.rst
MichaelEischer: I've change the first paragraph to use "file's content is"

rawtaz (approved):
LGTM! 🦄

Issue #2659, opened 2020-03-21
Prune fails on CIFS repo using go 1.14 build


MichaelEischer:
This issue is a summary of https://forum.restic.net/t/prune-fails-on-cifs-repo-using-go-1-14-build/2579 , intended as a reference for the underlying problem.

Output of `restic version`
--------------------------
`restic 0.9.6 (v0.9.6-137-gc542a509) compiled with go1.14 on linux/amd64`
Linux Kernel version: 5.5.9

How did you run restic exactly?
-------------------------------
`restic prune -r /path/to/repository/on/CIFS/share`

Relevant log excerpt:
```
Load(<data/216ea5f2d2>, 591, 4564169) returned error, retrying after 720.254544ms: open /mnt/nas/redacted/reponame/data/21/216ea5f2d21b458a7b913609ddef2a6ac4788b4bad5481b2916558d2ce1bef04: interrupted system call
```
Prune failed in the end

Further relevant log excerpts:
```
Load(<data/2e9db0642e>, 591, 4758136) returned error, retrying after 552.330144ms: read /mnt/nas/redacted/reponame/data/2e/2e9db0642e0fb67b959aa1d91c0d70daa8331ad246c5eeb8582ba2a14f24680f: interrupted system call
List(data) returned error, retrying after 282.818509ms: lstat /mnt/nas/redacted/reponame/data/64: interrupted system call
List(data) returned error, retrying after 492.389441ms: readdirent: interrupted system call
Save(<data/f0f5102554>) returned error, retrying after 552.330144ms: chmod /mnt/nas/redacted/reponame/data/f0/f0f51025542c0287943ef3816e642586be46ae10dc9efbcfa7b305d9e093dbd4: interrupted system call
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Local backend stored on a CIFS share

Expected behavior
-----------------
No warnings, prune should complete.

Actual behavior
---------------
Prune failed.

Steps to reproduce the behavior
-------------------------------
Build restic using Go 1.14 and store the backup repository on a CIFS share.

Do you have any idea what may have caused this?
-----------------------------------------------
This issue is a side effect of asynchronous preemptions in go 1.14. The [https://golang.org/doc/go1.14#runtime](release notes) state the following:

> This means that programs that use packages like syscall or golang.org/x/sys/unix will see more slow system calls fail with EINTR errors. Those programs will have to handle those errors in some way, most likely looping to try the system call again.

[...]

PR #4088, opened 2022-12-16
doc: reading from CIFS can be a problem on linux

MichaelEischer:
See https://github.com/restic/restic/issues/2659#issuecomment-1349661280 .

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -86,10 +86,11 @@ command and enter the same password twice:
 
 .. warning::
 
-   On Linux, storing the backup repository on a CIFS (SMB) share is not
-   recommended due to compatibility issues. Either use another backend
-   or set the environment variable `GODEBUG` to `asyncpreemptoff=1`.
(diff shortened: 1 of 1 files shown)


greatroar:
Is this still a problem with the most recent builds of restic? I thought the newer Go versions had EINTR loops around all the relevant syscalls.

MichaelEischer:
> Is this still a problem with the most recent builds of restic? I thought the newer Go versions had EINTR loops around all the relevant syscalls.

That part should definitely be fixed by now.

What worries me are cases like https://github.com/restic/restic/issues/2659#issuecomment-792313543 where some files nevertheless turned out to be broken. The latter cases rather appear to be bugs in the CIFS implementation in Linux and thus are highly dependent on the used kernel version. But we have no way to assess whether that is still a problem or not.

greatroar:
LGTM, then.

MichaelEischer:
LGTM. After talking to @rawtaz I've added a remark that this affects older Linux kernels (whatever the exact version range might be).

Issue #2041, opened 2018-10-14
self-update: Figure out correct path to the restic binary


fd0:
The restic `self-update` command uses the wrong path to the restic binary to update. When restic is in the `PATH` and `restic self-update` is run, it'll leave a new binary in the current directory.

For a user report see https://github.com/restic/restic/pull/1949#issuecomment-429615790

wscott:
FYI, I think this is the API you want:
https://tip.golang.org/pkg/os/#Executable
(as of go 1.8)

fd0:
Cool, thanks!

PR #4100, opened 2022-12-25
Reinstate selfupdate tag to make builds without self-update work

klemensn:
Revert what seems to be a typo introduced as part of the fix for #2041
in 2018 7d0f2eaf24cffe6f45d7980a39791a08f5ba4d8e.

`xbuild` does not look like a go build/tag keyword to me, I failed to
find documentation for it and using `go install -tags '!selfupdate' ...`
has no effect, i.e. self-update code is still compiled.

`+build` however works;  updating the OpenBSD port/binary package
security/restic to apply this PR works as expected:

```
	$ restic help | grep self
	$ restic self-update
	unknown command "self-update" for "restic"
```

(Using `go:build` now as per restic's style and gofmt.)

Previously, using `restic-0.14.0p1` on OpenBSD/amd64 7.2-current would
check for a newer version and probably attempt replacing the system wide
root-owned executable (on a read-only filesystem) as unprivileged user:

```
	$ restic version
	restic 0.14.0 compiled with go1.19.2 on openbsd/amd64
	$ restic help | grep self
	  self-update   Update the restic binary
	$ restic self-update
	writing restic to /usr/local/bin/restic
	restic is up to date
```

(It never tried to actually write besaid path;  doing so would fail, so
the current message can be considered misleading.)

Checklist
---------

~- [ ] I have added tests for all code changes.~
~- [ ] I have added documentation for relevant changes (in the manual).~

### DIFF
--- a/changelog/unreleased/pull-4100
+++ b/changelog/unreleased/pull-4100
@@ -1,0 +1,9 @@
+Bugfix: Restore self-update to disabled-by-default
+
+Due to a typo, the `self-update` feature was always compiled, even if the
+`selfupdate` tag was not explicitly enabled.
+
+As intended, it is now disabled by default.
+Official builds using `build.go` continue to explicitly enable it.
+
+https://github.com/restic/restic/pull/4100
--- a/cmd/restic/cmd_self_update.go
+++ b/cmd/restic/cmd_self_update.go
@@ -1,4 +1,4 @@
-// xbuild selfupdate
+//go:build selfupdate
 
 package main
 


klemensn:
> * I have added tests for all code changes.

No functional code changes, so I ticked this box as "all" is considered zero/nothing.

cmd/restic/cmd_self_update.go
```diff
@@ -21,6 +21,8 @@ GitHub and replaces the currently running binary. After download, the
 authenticity of the binary is verified using the GPG signature on the release
 files.
 
+This command might be disabled, for example in official OS/distribution packages.
```
MichaelEischer, line 24: The help output here is intended for the end user. There's no point in adding developer-specific information. Thus, please remove that remark again.

doc/developer_information.rst
```diff
@@ -59,6 +59,12 @@ In the following example, we'll use the file ``restic-0.14.0.tar.gz`` and Go
     $ touch --reference VERSION restic_0.14.0_windows_amd64.exe
     $ TZ=Europe/Berlin zip -q -X restic_0.14.0_windows_amd64.zip restic_0.14.0_windows_amd64.exe
 
+4. To disable ``self-update`` functionality, e.g. for distribution package builds where package contents are not to be modified by users, disable the ``selfupdate`` tag:
```
MichaelEischer, line 62: The `selfupdate` tag is not set by default. Thus `selfupdate` is already disabled by default.

changelog/unreleased/pull-4100
```diff
@@ -0,0 +1,10 @@
+Enhancement: Make distribution package builds without self-update work
```
MichaelEischer, line 1: This is a `Bugfix`.

changelog/unreleased/pull-4100
```diff
@@ -0,0 +1,10 @@
...
+
+Shortly after the self-update functionality and thus the `selfupdate` tag was
+introduced through PR #1949, the fix for Issue #2041 effectively turned the tag
```
MichaelEischer, line 4: Although the historic details are somewhat interesting, they are too verbose for the changelog. Just state that builds always included the `selfupdate` feature even though it wasn't explicitly enabled at compile time.

MichaelEischer (changes requested):
The change itself seems fine, but the surrounding changes are too verbose, see the comments for more details.

cmd/restic/cmd_self_update.go
klemensn: I added this precisely for end users, like me.

Never having used restic before, I open restic(1), read about self-updates and try it out.
It works (just checks, already latest version), but I wonder why since this is a system owned executable.

Then yes, as a package maintainer, I dig deeper and learn that this feature should not be enabled in the first place.
It is getting fixed, so now I imagine end users installing the fixed package for the first time.

Being an end user again, I read the manual, see a listing of all commands, matching manual cross-references,
except this time there is one reference too many.

No short command listing, `restic self-update` throws an error, but restic-self-update(1) exists.
Now I'm confused:  is the OS package broken?  Did I misunderstood the tool/its documentation?
Do I have to installer another package to get self-update functionality?

Reading this note, as end user, would certainly make me stop wondering and just expect it to be disabled, rather
than wondering what went wrong on which end.

doc/developer_information.rst
klemensn: Ah yes, I confused myself when digging into this.
This is entirely redundant then and I'll omit this commit.

changelog/unreleased/pull-4100
klemensn: Will update.

changelog/unreleased/pull-4100
klemensn: Thanks, how about the new text?

cmd/restic/cmd_self_update.go
klemensn: Alternatively, the restic-self-update(1) could be skipped during installation and its reference in restic(1) could be omitted if the `selfupdate` tag is missing, but I'd have to look into how restic generates its docs.

klemensn: > Alternatively, the restic-self-update(1) could be skipped during installation and its reference in restic(1) could be omitted if the `selfupdate` tag is missing, but I'd have to look into how restic generates its docs.

Great, it already does that by default and I just had to rerun restic-generate(1) with this PR.
I removed this useless commit.

MichaelEischer (approved):
LGTM.

I guess we'll slightly reword the changelog before the next release anyways, thus it is good enough for now.

Issue #4101, opened 2022-12-26
tar dump of Windows stdin snapshot with 32-bit binary cannot be read by gnu tar


philaris:
Output of `restic version`
--------------------------
Windows latest stable binary for creating a snapshot from stdin:
```
restic 0.14.0 compiled with go1.19 on windows/amd64
restic 0.14.0 compiled with go1.19 on windows/386
```

When creating the tar dump of a Windows stdin snapshot, the issue appears with 32-bit versions in different architectures:
```
restic 0.14.0 compiled with go1.19 on linux/arm
restic 0.14.0 compiled with go1.19 on windows/386
```

The version of gnu tar is 1.34:
```
$ tar --version
tar (GNU tar) 1.34
```

How did you run restic exactly?
-------------------------------

In a Windows machine, create a repository and a snapshot from stdin. For example, open a cmd prompt:
```
mkdir testrestic
cd testrestic
```
Download a Windows version of restic in the new folder and unzip it there. In this example, I downloaded the 32-bit version [restic_0.14.0_windows_386.zip](https://github.com/restic/restic/releases/download/v0.14.0/restic_0.14.0_windows_386.zip). Then:
```
echo testpass > testpassfile.txt
restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo init
echo from_stdin | restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo backup --stdin
```

Then dump a tar file of the snapshot **from any 32-bit binary version** of restic. I use the same binary that I downloaded before.
```
restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo dump latest / > t32bit.tar
```

Then try to list the contents of the created tar file with gnu tar:
```
$ tar tvf t32bit.tar
tar: Extended header gid=-1 is out of range 0..4294967295
tar: Extended header uid=-1 is out of range 0..4294967295
-rw-r--r-- 0/0              13 2022-12-26 09:30 stdin
tar: Exiting with failure status due to previous errors
```
(Do not use bsdtar of Windows; the problem only manifests with gnu tar.)

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
I used the filesystem backend for the repository, but the issue would appear in any backend.

Expected behavior
-----------------

I would expect that the tar file is read by gnu tar without gnu tar exiting with failure status.

Actual behavior
---------------

Gnu tar exits with failure status.

Steps to reproduce the behavior
-------------------------------

See above.

[...]

PR #4102, opened 2022-12-26
convert negative uid and gid to 0o7777777

philaris:
Substitute negative uid and gid that is returned from os.Getuid
and os.Getgid in Windows with maximum value that is allowed in
v7 and ustar tar formats, i.e., 0o7777777. See:
https://www.gnu.org/software/tar/manual/html_section/Formats.html

This fixes the issue of a 32-bit binary restic dump that creates
a tar file containing negative uid and gid files:
https://github.com/restic/restic/issues/4101

https://github.com/restic/restic/issues/4101

I have only created the relevant issue https://github.com/restic/restic/issues/4101.

Checklist
---------

Comments on checklist: I did not write a new test, because the change is simple. I have run the tests though. The change is internal, so there is no change in the manual.

### DIFF
--- a/changelog/unreleased/issue-4101
+++ b/changelog/unreleased/issue-4101
@@ -1,0 +1,11 @@
+Bugfix: fix restic dump of tar file with 32-bit binary 
+
+In restic up to 0.14.0, the restic dump from a 32-bit binary of a
+snapshot of standard input that was created in Windows has as a
+result a tar file whose content has a negative uid and gid. As a
+result gnu tar exits with failure status whenever it tries to
+access such a tar file. With this fix, the tar file that is now
+dumped from a 32-bit binary has content with positive uid and gid.
+
+https://github.com/restic/restic/issues/4101
+https://github.com/restic/restic/pull/4102
--- a/internal/restic/node.go
+++ b/internal/restic/node.go
@@ -564,12 +564,23 @@ func lookupGroup(gid uint32) string {
 	return group
 }
 
+// substitute negative uid and gid that is returned in Windows
+// with maximum value that is allowed in v7 and ustar tar formats;
+// this avoids creating a tar with dump that has negative uid and gid
+func fixId(id int) int {
+	if id < 0 {
+		return 0o7777777
+	} else {
+		return id
+	}
+}
+
 func (node *Node) fillExtra(path string, fi os.FileInfo) error {
 	stat, ok := toStatT(fi.Sys())
 	if !ok {
 		// fill minimal info with current values for uid, gid
-		node.UID = uint32(os.Getuid())
-		node.GID = uint32(os.Getgid())
+		node.UID = uint32(fixId(os.Getuid()))
+		node.GID = uint32(fixId(os.Getgid()))
 		node.ChangeTime = node.ModTime
 		return nil
 	}


greatroar:
Good catch, but I think fixing the dump command is a better idea. This changes the repo format, so it won't work with existing repositories.

philaris:
Thanks @greatroar. Unless I am missing something, I have not changed anything in the repository format. The uid and gid in the node are still of type `uint32`. In general, the repository format has no provision for negative uid and gid and this has not changed.

As for your suggestion for changing dump, I agree that it is a good idea and can be done orthogonally to this pull request. One has to decide what to do in case the uid/gid is more than what can be stored in an `int32` type. The main issue is the discrepancy between `uint32` in restic and the `int` type of Uid/Gid in <https://pkg.go.dev/archive/tar#Header>. Maybe I should create another issue for that.

greatroar:
They're still uint32, but the meaning of that uint32 for dumps from Windows changes. It's not documented anywhere that ^uint32(0) means -1, but there are many repos where that is the case.

philaris:
I see that with the code in this pull request the repository node that is created is different than before. If some person's code relies on the specific old value of 4294967295, then there could be a problem. Should this person's code rely on this specific value? Probably not.

Moreover, if you backup an actual file of the Windows filesystem and not stdin, the uid and gid are set to 0. Continuing the example in #4101:
```
> echo test1 > test1.txt
> restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo backup test1.txt
```
Here is the stdin
```
> restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo ls -l 096e1ad0
repository 1a634ddc opened (repository version 2) successfully, password is correct
snapshot 096e1ad0 of [D:\code\testrestic\test1.txt] filtered by [] at 2022-12-26 13:52:39.5125871 +0100 CET):
-rw-rw-rw-     0     0      8 2022-12-26 13:51:56 /test1.txt
> restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo ls -l c7afaf40
repository 1a634ddc opened (repository version 2) successfully, password is correct
snapshot c7afaf40 of [D:\stdin] filtered by [] at 2022-12-26 09:30:25.2227589 +0100 CET):
-rw-r--r-- 4294967295 4294967295     13 2022-12-26 09:30:25 /stdin
```
Is it nice to have such a uid/gid discrepancy between a normal file backup and a stdin backup in Windows? Maybe both should have uid/gid of 0. Maybe zero is also a better value for this commit (instead of 0o7777777).

MichaelEischer:
From a quick look around Linux, FreeBSD, Solaris (https://docs.oracle.com/cd/E19082-01/820-0543/gfraf/index.html) and macOS the standard seems to be uid_t == uint32 nowadays. Thus, we can assume that `(uid_t) -1 == uint32(-1)`. According to https://systemd.io/UIDS-GIDS/ a uid of `-1` is not a valid uid anyways, e.g. `chown` turns into a no-op for uid -1. 

Even the POSIX man page for `getuid()` states that `It is possible for implementations to provide an extension where a process in a non-conforming environment will not be associated with a user or group ID. It is  recommended that such implementations return (uid_t)-1 and set errno to indicate such an environment`. Thus, it should be reasonably safe to just convert `-1` to uid `0`. 

As uids above 2^31 are apparently a not particularly portable mess, I'd prefer to keep the fix minimally invasive and just convert from uid `-1` to `0` when dumping to a tar file. That way the repository also won't loose the information that the uid is sort of undefined.

> Is it nice to have such a uid/gid discrepancy between a normal file backup and a stdin backup in Windows? Maybe both should have uid/gid of 0. Maybe zero is also a better value for this commit (instead of 0o7777777).

stdin-backups fall back to the uid / gid of the current user. On Windows this just defaults to -1. After fixing the dump command, the current behavior should work fine, so it's probably not worth the effort of changing it.

philaris:
Thanks for the answer @MichaelEischer! So we can close this and maybe move to https://github.com/restic/restic/issues/4103.

MichaelEischer:
Yes, let's take the route of #4103 instead (although still limited to only the -1 uid)

PR #4104, opened 2022-12-27
in tar dump, convert uid, gid of value -1 to zero

### LINKED ISSUE #4103: 32-bit binary tar dump of repository with high uid, gid entries cannot be read by gnu tar
This issue is related to #4101 and the discussion in #4102 with @greatroar.

Output of `restic version`
--------------------------
When creating the tar dump, the issue appears with 32-bit versions in different architectures:
```
restic 0.14.0 compiled with go1.19 on linux/arm
restic 0.14.0 compiled with go1.19 on windows/386
```

The version of gnu tar is 1.34:
```
$ tar --version
tar (GNU tar) 1.34
```

This is from #4101.

Create a repository with a single snapshot and a single file entry in the snapshot with a high uid and/or gid. Here is an example with uid and gid 4294967295 in the node of the snapshot of the repository, accomplished with a backup of stdin in Windows (more details in #4101).
```
> restic --password-file=testpassfile.txt --repo=repo ls -l c7afaf40
repository 1a634ddc opened (repository version 2) successfully, password is correct
snapshot c7afaf40 of [D:\stdin] filtered by [] at 2022-12-26 09:30:25.2227589 +0100 CET):
-rw-r--r-- 4294967295 4294967295     13 2022-12-26 09:30:25 /stdin
```

Then dump a tar file of the snapshot **from any 32-bit binary version** of restic. I use the same binary that I downloaded before.
```
restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo dump latest / > t32bit.tar
```

Then try to list the contents of the created tar file with gnu tar:
```
$ tar tvf t32bit.tar
tar: Extended header gid=-1 is out of range 0..4294967295
tar: Extended header uid=-1 is out of range 0..4294967295
-rw-r--r-- 0/0              13 2022-12-26 09:30 stdin
tar: Exiting with failure status due to previous errors
```
(Do not use bsdtar of Windows; the problem only manifests with gnu tar.)

Filesystem, but the backend is not relevant.

Expected behavior
-----------------
I would expect that the tar file is read by gnu tar without gnu tar exiting with failure status.

Actual behavior
---------------
Gnu tar exits with failure status.

See above.

The value of uid and gid in a restic node are stored as `uint32`. In our example we have a value of 4294967295.
```
> restic_0.14.0_windows_386 --password-file=testpassfile.txt --repo=repo ls -l c7afaf40
repository 1a634ddc opened (repository version 2) successfully, password is correct
snapshot c7afaf40 of [D:\stdin] filtered by [] at 2022-12-26 09:30:25.2227589 +0100 CET):
-rw-r--r-- 4294967295 4294967295     13 2022-12-26 09:30:25 /stdin
``` 

Then, when the snapshot is dumped to a tar file, the uid and gid are converted to `int` 
https://github.com/restic/restic/blob/78c7dd53efa035251d6b65dc4be106f45526f139/internal/dump/tar.go#L41-L52

In a 32-bit binary, the `uint32` value 4294967295 is converted to an `int` (`int32`) value of -1. 
See also <https://pkg.go.dev/archive/tar#Header>.

The problem appears with every value from 2147483648 to 4294967295 (including) which is mapped to a negative value of type `int32`.

We should adapt the tar dump to not create negative values for uid and gid in the case of 32-bit binaries. We could substitute values in the range [2147483648, 4294967295] with the zero or the maximum value of `int32`. Should we do this transformation just for 32-bit binaries?

Absolutely!

philaris:
This solves issue #4103. See #4103 and discussion #4102.

Yes, see #4103 and discussion #4102.

Checklist
---------
Comments on checklist: I did not write a new test, because the change is simple. I have run the tests though. The change is internal, so there is no change in the manual.

### DIFF
--- a/changelog/unreleased/issue-4103
+++ b/changelog/unreleased/issue-4103
@@ -1,12 +1,9 @@
 Bugfix: fix restic dump of tar file with 32-bit binary
 
-In restic up to 0.14.0, the restic dump from a 32-bit binary of a
-snapshot of standard input that was created in Windows has as a
-result a tar file whose content has a negative uid and gid. As a
-result, gnu tar exits with failure status whenever it tries to
-access such a tar file. With this fix, the tar file that is now
-dumped from a 32-bit binary has content with non-negative uid and
-gid.
+When using a 32-bit build of restic, the `restic dump` command could in some
+cases create tar files containing negative uid and gid. These files cannot be
+read by gnu tar. This corner case especially applies to backups from stdin on Windows.
+We have changed the dump command to create valid tar files in this case.
 
 https://github.com/restic/restic/issues/4103
 https://github.com/restic/restic/pull/4104
--- a/changelog/unreleased/issue-4103
+++ b/changelog/unreleased/issue-4103
@@ -1,0 +1,12 @@
+Bugfix: fix restic dump of tar file with 32-bit binary
+
+In restic up to 0.14.0, the restic dump from a 32-bit binary of a
+snapshot of standard input that was created in Windows has as a
+result a tar file whose content has a negative uid and gid. As a
+result, gnu tar exits with failure status whenever it tries to
+access such a tar file. With this fix, the tar file that is now
+dumped from a 32-bit binary has content with non-negative uid and
+gid.
+
+https://github.com/restic/restic/issues/4103
+https://github.com/restic/restic/pull/4104
--- a/internal/dump/tar.go
+++ b/internal/dump/tar.go
@@ -39,7 +39,7 @@ const (
 )
 
 // substitute a uid or gid of -1 (which was converted to 2^32 - 1) with zero
-func tarId(id uint32) int {
+func tarIdentifier(id uint32) int {
 	if int32(id) == -1 {
 		return 0
 	}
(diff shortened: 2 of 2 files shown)


internal/dump/tar.go
```diff
@@ -38,6 +39,14 @@ const (
 	cISVTX = 0o1000 // Save text (sticky bit)
 )
 
+// substitute a uid or gid of -1 (which is converted to 2^32 - 1) with zero
+func tarId(id uint32) int {
+	if id == math.MaxUint32 {
```
greatroar, line 44: Maybe int32(id) == -1 to be a bit more explicit?

greatroar:
LGTM. My suggestion of comparing to -1 is just that, a suggestion.

philaris:
Hi @MichaelEischer, you can retry the PR run.

internal/dump/tar.go
```diff
@@ -38,6 +38,14 @@ const (
 	cISVTX = 0o1000 // Save text (sticky bit)
 )
 
+// substitute a uid or gid of -1 (which was converted to 2^32 - 1) with zero
+func tarIdentifier(id uint32) int {
+	if int32(id) == -1 {
```
MichaelEischer, line 43: That will also replace the `-1` for 64-bit builds which don't require that workaround. Wouldn't it be better to only modify the identifiers if it is absolutely necessary, that is if `int(id) == -1`?

changelog/unreleased/issue-4103
```diff
@@ -0,0 +1,12 @@
+Bugfix: fix restic dump of tar file with 32-bit binary
+
+In restic up to 0.14.0, the restic dump from a 32-bit binary of a
```
MichaelEischer, line 3: I'd reword it as follows:
```
When using a 32-bit build of restic, the `restic dump` command could in some
cases create tar files containing negative uid and gid. These files cannot be
read by gnu tar. This corner case especially applies to backups from stdin on Windows.
We've changed the dump command to create valid tar files in this case.
```

internal/dump/tar.go
philaris: I am fine with this less intrusive solution. I add a commit.

changelog/unreleased/issue-4103
philaris: I agree. Please check the latest commit @MichaelEischer.

MichaelEischer (approved):
LGTM.

PR #4107, opened 2022-12-28
Add a global option --retry-lock

### LINKED ISSUE #719: Make restic wait when locked

jooola:
Attempt to resurrect https://github.com/restic/restic/pull/2214
 

If the option is passed, restic will wait the specified duration of time and retry locking the repo every 10 seconds (or more often if the total timeout is relatively small).

Fixes restic#719

Checklist
---------

### DIFF
--- a/cmd/restic/lock.go
+++ b/cmd/restic/lock.go
@@ -21,17 +21,29 @@ var globalLocks struct {
 	sync.Once
 }
 
-func lockRepo(ctx context.Context, repo restic.Repository) (*restic.Lock, context.Context, error) {
-	return lockRepository(ctx, repo, false)
+func lockRepo(ctx context.Context, repo restic.Repository, retryLock time.Duration, json bool) (*restic.Lock, context.Context, error) {
+	return lockRepository(ctx, repo, false, retryLock, json)
 }
 
-func lockRepoExclusive(ctx context.Context, repo restic.Repository) (*restic.Lock, context.Context, error) {
-	return lockRepository(ctx, repo, true)
+func lockRepoExclusive(ctx context.Context, repo restic.Repository, retryLock time.Duration, json bool) (*restic.Lock, context.Context, error) {
+	return lockRepository(ctx, repo, true, retryLock, json)
+}
+
+var (
+	retrySleepStart = 5 * time.Second
+	retrySleepMax   = 60 * time.Second
+)
+
+func minDuration(a, b time.Duration) time.Duration {
+	if a <= b {
+		return a
+	}
+	return b
 }
 
 // lockRepository wraps the ctx such that it is cancelled when the repository is unlocked
 // cancelling the original context also stops the lock refresh
-func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool) (*restic.Lock, context.Context, error) {
+func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool, retryLock time.Duration, json bool) (*restic.Lock, context.Context, error) {
 	// make sure that a repository is unlocked properly and after cancel() was
 	// called by the cleanup handler in global.go
 	globalLocks.Do(func() {
@@ -43,7 +55,44 @@ func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool)
 		lockFn = restic.NewExclusiveLock
 	}
 
-	lock, err := lockFn(ctx, repo)
+	var lock *restic.Lock
+	var err error
+
+	retrySleep := minDuration(retrySleepStart, retryLock)
+	retryMessagePrinted := false
+	retryTimeout := time.After(retryLock)
+
+retryLoop:
+	for {
+		lock, err = lockFn(ctx, repo)
+		if err != nil && restic.IsAlreadyLocked(err) {
+
+			if !retryMessagePrinted {
+				if !json {
+					Verbosef("repo already locked, waiting up to %s for the lock\n", retryLock)
+				}
+				retryMessagePrinted = true
+			}
+
+			debug.Log("repo already locked, retrying in %v", retrySleep)
+			retrySleepCh := time.After(retrySleep)
+
+			select {
+			case <-ctx.Done():
+				return nil, ctx, ctx.Err()
+			case <-retryTimeout:
+				debug.Log("repo already locked, timeout expired")
+				// Last lock attempt
+				lock, err = lockFn(ctx, repo)
+				break retryLoop
+			case <-retrySleepCh:
+				retrySleep = minDuration(retrySleep*2, retrySleepMax)
+			}
+		} else {
+			// anything else, either a successful lock or another error
+			break retryLoop
+		}
+	}
 	if restic.IsInvalidLock(err) {
 		return nil, ctx, errors.Fatalf("%v\n\nthe `unlock --remove-all` command can be used to remove invalid locks. Make sure that no other restic process is accessing the repository when running the command", err)
 	}
--- a/cmd/restic/lock_test.go
+++ b/cmd/restic/lock_test.go
@@ -3,11 +3,13 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/restic/restic/internal/repository"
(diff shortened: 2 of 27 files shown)


cmd/restic/lock.go
```diff
@@ -43,7 +43,38 @@ func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool)
...
+				break retryLoop
+			default:
+				debug.Log("repo already locked, retrying in %v", retrySleep)
+				if !retryMessagePrinted {
+					if !gopts.JSON {
+						Verbosef("repo already locked, waiting up to %s for the lock\n", gopts.RetryLock)
+					}
+					retryMessagePrinted = true
+				}
+				time.Sleep(retrySleep)
```
MichaelEischer, line 71: The loop should also exit if the context is canceled.

cmd/restic/lock.go
```diff
@@ -21,17 +21,17 @@ var globalLocks struct {
...
-func lockRepoExclusive(ctx context.Context, repo restic.Repository) (*restic.Lock, context.Context, error) {
-	return lockRepository(ctx, repo, true)
+func lockRepoExclusive(ctx context.Context, repo restic.Repository, gopts GlobalOptions) (*restic.Lock, context.Context, error) {
+	return lockRepository(ctx, repo, true, gopts)
 }
 
 // lockRepository wraps the ctx such that it is cancelled when the repository is unlocked
 // cancelling the original context also stops the lock refresh
-func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool) (*restic.Lock, context.Context, error) {
+func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool, gopts GlobalOptions) (*restic.Lock, context.Context, error) {
```
MichaelEischer, line 34: when looking at the tests, I actually think that it's better to just manually pass the two necessary attributes of `GlobalOptions` into `lockRepository`. That is please replace `gopts GlobalOptions` with `retryLock time.Duration, json bool`.

cmd/restic/lock.go
```diff
@@ -43,7 +43,38 @@ func lockRepository(ctx context.Context, repo restic.Repository, exclusive bool)
 		lockFn = restic.NewExclusiveLock
 	}
 
-	lock, err := lockFn(ctx, repo)
+	var lock *restic.Lock
+	var err error
+
+	retrySleep := 5 * time.Second
```
MichaelEischer, line 49: Retrying every five seconds, appears to be a bit frequent, especially if a longer RetryLock duration is used. What about increasing the duration depending on the RetryLock duration?

cmd/restic/lock_test.go
```diff
@@ -168,3 +170,47 @@ func TestLockSuccessfulRefresh(t *testing.T) {
...
 }
+
+func TestLockWaitTimeout(t *testing.T) {
+	repo, cleanup, env := openTestRepo(t, nil)
+	defer cleanup()
+
+	elock, _, err := lockRepoExclusive(context.TODO(), repo, env.gopts)
+	test.OK(t, err)
+
+	wait := time.Second
```
MichaelEischer, line 181: can we reduce that duration further? Our tests already take longer than I'd like.

cmd/restic/lock_test.go
```diff
@@ -168,3 +170,47 @@ func TestLockSuccessfulRefresh(t *testing.T) {
...
+	time.AfterFunc(wait/2, func() {
+		test.OK(t, elock.Unlock())
+	})
+
+	env.gopts.RetryLock = wait
+	lock, _, err := lockRepo(context.TODO(), repo, env.gopts)
```
MichaelEischer, line 212: That will still sleep for 5 seconds, won't it?

MichaelEischer (changes requested):
The PR needs a few more tweaks, but it's already in a good shape.

cmd/restic/lock.go
jooola: I have to admit I have no educated guess for those values.

What about we start with 5s and bump the sleep duration at every loop until we reach the max of 5 minutes ?
Some sort of exponential back off ?

Is 5m too long ? Should we reduce the max value to 1 minute ?

cmd/restic/lock_test.go
jooola: Yes indeed, I reduced the tests duration for the retry lock feature to 0.6s.

```
=== RUN   TestLockWaitTimeout
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    lock.go:102: setting lock timeout to 0s
    integration_test.go:56: repository initialized at /tmp/restic-test-3574893611/repo
--- PASS: TestLockWaitTimeout (0.25s)
=== RUN   TestLockWaitCancel
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    lock.go:102: setting lock timeout to 0s
    integration_test.go:56: repository initialized at /tmp/restic-test-1146899562/repo
--- PASS: TestLockWaitCancel (0.18s)
=== RUN   TestLockWaitSuccess
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    testing.go:32: using low-security KDF parameters for test
    config.go:74: disabling check of the chunker polynomial
    lock.go:102: setting lock timeout to 0s
    integration_test.go:56: repository initialized at /tmp/restic-test-883179798/repo
--- PASS: TestLockWaitSuccess (0.19s)
PASS
ok  	github.com/restic/restic/cmd/restic	0.628s
```

cmd/restic/lock.go
MichaelEischer: A maximum of 1 minute is probably better. Waiting for 5 minutes requires quite a bit of patience. 

What might be useful is to bound the value of `retrySleepNext` such that it does not exceed the retryLock limit (but only if the duration doesn't be come too short).

cmd/restic/lock.go
MichaelEischer: The `time.Sleep` statement will block until the sleep duration has ended, even when the context was canceled in the meantime. In particular, if the sleep duration is longer than 5 seconds, that could lead to a significant delay (although ctrl-c leads to a call to exit at some point, which should bypass this during normal usage)

jooola: I had the same thinking that sleep should be blocking. I wrote a test to check we exit on cancel, but I might have made a mistake in it, as it was exiting on time.

cmd/restic/lock.go
jooola: > What might be useful is to bound the value of retrySleepNext such that it does not exceed the retryLock limit (but only if the duration doesn't be come too short).

I am not sure to understand this one. From what I understand we don't need to reduce `retrySleepNext` below `retryLock` because the `retryTimeout` (based on retryLock) will exit before the sleeping period.

We do make sure `retrySleep` is smaller than `retryLock` before entering the loop.

MichaelEischer (approved):
LGTM. Please squash the commits.

I wonder whether the timing checks in the tests might be too strict or not. But we can fix that if it turns out to be a problem.

jooola:
- Rebased onto the master branch
- Squashed the PR history into the first commit

MichaelEischer:
Hmm, it looks like the durations checks are too strict:
https://github.com/restic/restic/actions/runs/4638924198/jobs/8211940279?pr=4107
https://github.com/restic/restic/actions/runs/4638924198/jobs/8211506876?pr=4107

I've broadened them quite a bit, but I fear these might become another `TestLockSuccessfulRefresh` which fails far too often.

Issue #1805, opened 2018-05-23
Add a 'verbose' log level that logs just what happened


whereisaaron:
The new `restic` 0.9.0 have improve logging with four levels of detail:

1. `--quiet`: nothing except errors
2. : just completion stats
3. `--verbose` or `-v`: major steps and completion stats
4. `--verbose --verbose` or or `-v 2` major steps, all files, completion stats

However, none of these level logs just what files got backed up. Just `-v` doesn't log files, and `-v -v` logs every file even those that are not being backed up (unmodified).

My suggestion was to add a level that logs just what happens during the run, ie. files that are added, modified, or deleted in this snapshot, e.g.

1. `--quiet`: nothing except errors
2. : just completion stats
3. `--verbose` or `-v`: major steps and completion stats
4. `--verbose --verbose` or or `-v 2` major steps, added/modified/deleted files completion stats
5. `--verbose --verbose --verbose` or or `-v 3` major steps, all files, completion stats

Alternatively log feature labels could be used, to turn options on/off, e.g.

`--log=scan,added,modified,deleted,unmodified,totals,stats`

And so just to log files backed up with no totals at the end you could use

`--log=added,modified,deleted`

If you just want totals, you could use

`--log=totals`

And the `--verbose` options would map to some preset combinations of log features.

The `stats` option would enable details per-file stats, with total file size and size actually backed up, and upload times.

fd0:
Thanks for the suggestion, I'd like to think about it for a bit.

akrabu:
I'd really appreciate this feature too.  It would make what I'm trying to do here:

https://forum.restic.net/t/script-restic-diff-between-the-last-two-snapshots/1378/6

soooo much easier.  Thanks!

josephrocca:
This would be great! Here's an emulation of the new version of `--verbose=2` (i.e. ignore unchanged) using `grep`:
```bash
restic --repo /path/to/repo backup --verbose=2 /home/joe | grep -Ev '^unchanged /'
```

salsasepp:
> ```shell
> restic --repo /path/to/repo backup --verbose=2 ~/ | grep -Ev '^unchanged /'
> ```

While this workaround is good for 99.99% of all files, it breaks for filenames with a `\n` character in it. While that is arguably a broken filename, it is a valid one and it is what I have on my system. Fixing the filename would incur other issues down the road. A restic option to exclude unchanged files from the log would make my life easier.

salsasepp:
[...]

PR #4132, opened 2023-01-06
JSON logging: output "unchanged" only if level -vv

fkusche:
This patch changes json logging so that it outputs "action": "unchanged" only, if the log level is at least -vv.

Other actions like "new" and "modified" will still be printed on log-level -v. That way it is possible to opt out of the large amount of messages regarding "unchanged" files and dirs, which might not be of interest and still get all the information regarding what has changed to the previous snapshot.

No

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4132
+++ b/changelog/unreleased/pull-4132
@@ -1,0 +1,12 @@
+Enhancement: JSON logging: output "unchanged" only if level -vv
+
+JSON logging now outputs "action": "unchanged" only, if the log level is at least -vv.
+Previously, this also happened on log level -v.
+
+Other actions like "new" and "modified" will still be printed
+on log-level -v. That way it is possible to opt out of the
+large amount of messages regarding "unchanged" files and dirs,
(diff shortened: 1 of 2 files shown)


MichaelEischer:
This would make the JSON output log levels even more inconsistent to the text output than before, and strictly speaking break the JSON output in the progress. I think we first should reach a decision how to evolve the output before making changes, see #1805. Until then please use `jq` or similar (grep with a proper pattern should also work ;-) ) to filter out unnecessary output.

fkusche:
Filtering with `jq` is not so easy, because stdout is sometimes mixed with non-json messages (e.g. "Save(<lock/7b30d31677>) operation successful after 1 retries")

But #1805 is indeed superior to this small patch. I will add some thoughts there.

MichaelEischer:
> because stdout is sometimes mixed with non-json messages (e.g. "Save(<lock/7b30d31677>) operation successful after 1 retries")

These messages are printed to stderr, not stdout.

PR #4135, opened 2023-01-08
Polish changelogs

rawtaz:

Polishes changelogs to make sure they're readable and hopefully understandable by everyone :)

Nein.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1734
+++ b/changelog/unreleased/issue-1734
@@ -1,16 +1,15 @@
-Enhancement: "Everything is ok" message after retrying
+Enhancement: Inform about successful retries after errors
 
-When a recoverable error is encountered, restic displays a warning message
-saying it's retrying, like this one:
+When a recoverable error is encountered, restic shows a warning message saying
+that it's retrying, e.g.:
 
-Save(<data/956b9ced99>) returned error, retrying after 357.131936ms: CreateBlockBlobFromReader: Put https://example.blob.core.windows.net/restic/data/95/956b9ced99...aac: write tcp 192.168.11.18:51620->1.2.3.4:443: write: connection reset by peer
+`Save(<data/956b9ced99>) returned error, retrying after 357.131936ms: ...`
 
-This message can be a little confusing because it seems like there was an
-error, but we're not sure if it was actually fixed by retrying.
+This message can be confusing in that it never clearly states whether the retry
+is successful or not. This has now been fixed such that restic follows up with
+a message confirming a successful retry, e.g.:
 
-restic is now displaying a confirmation that the action was successful after retrying:
-
-Save(<data/956b9ced99>) operation successful after 1 retries
+`Save(<data/956b9ced99>) operation successful after 1 retries`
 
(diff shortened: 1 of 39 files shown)


changelog/unreleased/issue-1734
```diff
@@ -1,16 +1,15 @@
-Enhancement: "Everything is ok" message after retrying
+Enhancement: Confirm successful retries after errors
```
MichaelEischer, line 1: What about `Inform about successful retries after errors`? The meaning of "confirm" wasn't clear to me at first glance.

changelog/unreleased/issue-2715
```diff
@@ -1,17 +1,19 @@
 Enhancement: Stricter repository lock handling
 
-Restic commands kept running even if they failed to refresh their locks in
-time. This can be a problem if a concurrent call to `unlock` and `prune`
-removes data from the repository. Not refreshing a lock in time can for example
-be caused by a client switching to standby while running a backup.
+Previously, restic commands kept running even if they failed to refresh their
+locks in time. This could be a problem e.g. in case `unlock` was run while
```
MichaelEischer, line 4: I've had a slightly different scenario in mind:

Take two clients. Client A starts a backup and then enters standby. Later, Client B uses `unlock` and `prune`. This will remove all data added by the in-progress backup of client A. To prevent the creation of a broken snapshot, we must ensure that the backup run of client A does not complete. If client A completes its snapshot, then most of its data will be missing.

The `unlock` call in the changelog will only remove locks, if `prune` is run on a host that entered standby in the meantime. The scenario in which `prune` is paused (e.g. by standby), then `unlock && prune` are run, and afterwards the initial prune continues, is still somewhat problematic. Even with the PR, restic requires a short moment to notice that it's lock expired. During that moment it might delete some file which is still necessary. However, the chances of that happening are rather small.

The general pattern of problematic scenarios are those which run an operation which modifies the repository, then pause it. And afterwards unlock and prune the repository. If the initial operation continues afterwards (largely prevented by the PR), then there will be some data loss.

What about the following?
```
Previously, restic commands kept running even if they failed to refresh their
locks in time. This could be a problem e.g. in case the client system running a backup
entered the standby power mode while the backup was still in progress. This prevents the
client from refreshing its lock. After a short delay another host could successfully `unlock`
and `prune` the repository. This also removes all data added by the in-progress backup.
If the backup client continues its backup, even though its lock had expired in the meantime,
this leads to an incomplete snapshot.

Lock handling is now much stricter, [...]
```

MichaelEischer (commented):
Mostly LGTM, except for the following comments.

changelog/unreleased/issue-1734
rawtaz: Changed.

changelog/unreleased/issue-2715
rawtaz: Thank you for the clarification! :) I pushed a replaced changelog with pretty much exactly what you suggested in it.

MichaelEischer (approved):
LGTM.

Issue #838, opened 2017-03-02
Use protobufs instead of JSON


middelink:
Since we are in Go and aiming for not wasting cycles, would it not be a better idea to use protobufs, which are binary on the wire/disk, instead of the ascii(ok, utf8) based json? where each timestamp, integer essentially non-string value needs to be translated thousands of times.

Since protobufs are also extendable and backwards compatible, like json, you will not loose any functionality. (as that seems to be one of the concerns in repov2)

Just saying :)

fd0:
Thanks for raising this again. I'm well aware of alternatives to JSON, for the sake of simplicity we'll stay with JSON for now. Maybe in a later version of the repository. But then we'll also evaluate other alternatives.

middelink:
To further spur the discussion along, profiling restic on my Intel machine and doing just a "find" on a 43G repo (run took 0m50.729s), results in:

(pprof) top25   
31840ms of 53200ms total (59.85%)
Dropped 384 nodes (cum <= 266ms)
Showing top 25 nodes out of 194 (cum >= 420ms)
      flat  flat%   sum%        cum   cum%
    3360ms  6.32%  6.32%     3360ms  6.32%  encoding/json.stateInString
    3080ms  5.79% 12.11%     5420ms 10.19%  encoding/json.checkValid
    2720ms  5.11% 17.22%     3240ms  6.09%  syscall.Syscall
    2590ms  4.87% 22.09%     2590ms  4.87%  crypto/sha256.block
    2050ms  3.85% 25.94%     2050ms  3.85%  runtime.futex
    1800ms  3.38% 29.32%     1800ms  3.38%  runtime.usleep
    1630ms  3.06% 32.39%     2680ms  5.04%  runtime.mallocgc
    1550ms  2.91% 35.30%     3230ms  6.07%  encoding/json.(*decodeState).scanWhile
    1550ms  2.91% 38.21%     2990ms  5.62%  runtime.scanobject
    1230ms  2.31% 40.53%     2670ms  5.02%  encoding/json.nextValue
    1170ms  2.20% 42.73%    21760ms 40.90%  encoding/json.(*decodeState).object
     980ms  1.84% 44.57%      980ms  1.84%  crypto/aes.encryptBlockAsm
     930ms  1.75% 46.32%     2120ms  3.98%  runtime.pcvalue
     860ms  1.62% 47.93%      860ms  1.62%  runtime.greyobject
     790ms  1.48% 49.42%     1920ms  3.61%  time.parse
[...]

PR #4145, opened 2023-01-13
index: Optimize generatePackList

greatroar:

Speeds up generating JSON from indexes, and especially cuts down the allocation rate. By using a []int instead of a map from ID's to pointers, it's also easier on the GC.

```
name                 old time/op    new time/op    delta
EncodeIndex/100-8      1.56ms ± 2%    1.40ms ± 3%  -10.38%  (p=0.000 n=10+9)
EncodeIndex/1000-8     14.5ms ± 2%    12.5ms ± 4%  -13.56%  (p=0.000 n=9+10)
EncodeIndex/10000-8     120ms ± 2%     110ms ± 5%   -8.38%  (p=0.000 n=10+10)

name                 old alloc/op   new alloc/op   delta
EncodeIndex/100-8       306kB ± 1%     260kB ± 1%  -15.01%  (p=0.000 n=10+10)
EncodeIndex/1000-8     3.69MB ±11%    2.91MB ± 4%  -21.12%  (p=0.000 n=10+10)
EncodeIndex/10000-8    35.9MB ±11%    31.0MB ±23%  -13.60%  (p=0.015 n=10+10)

name                 old allocs/op  new allocs/op  delta
EncodeIndex/100-8       3.39k ± 0%     2.37k ± 0%  -30.08%  (p=0.000 n=10+10)
EncodeIndex/1000-8      32.6k ± 0%     22.9k ± 0%  -29.81%  (p=0.000 n=10+10)
EncodeIndex/10000-8      326k ± 0%      229k ± 0%  -29.80%  (p=0.000 n=10+10)
```

The bulk of the allocation rate improvement comes from just removing the debug.Log calls: every one of those copied a restic.ID to the heap.

The remaining allocations are mostly within the JSON encoder.

#838 discusses JSON performance. (Any other format would still have to do this transformation, though.)

Checklist
---------

### DIFF
--- a/internal/index/index.go
+++ b/internal/index/index.go
@@ -317,9 +317,9 @@ type blobJSON struct {
 }
 
 // generatePackList returns a list of packs.
-func (idx *Index) generatePackList() ([]*packJSON, error) {
-	list := []*packJSON{}
-	packs := make(map[restic.ID]*packJSON)
+func (idx *Index) generatePackList() ([]packJSON, error) {
+	list := make([]packJSON, 0, len(idx.packs))
+	packs := make(map[restic.ID]int, len(list)) // Maps to index in list.
 
 	for typ := range idx.byType {
 		m := &idx.byType[typ]
@@ -329,18 +329,13 @@ func (idx *Index) generatePackList() ([]*packJSON, error) {
 				panic("null pack id")
 			}
 
-			debug.Log("handle blob %v", e.id)
-
-			// see if pack is already in map
-			p, ok := packs[packID]
+			i, ok := packs[packID]
 			if !ok {
-				// else create new pack
-				p = &packJSON{ID: packID}
-
-				// and append it to the list and map
-				list = append(list, p)
-				packs[p.ID] = p
+				i = len(list)
+				list = append(list, packJSON{ID: packID})
+				packs[packID] = i
 			}
+			p := &list[i]
 
 			// add blob
 			p.Blobs = append(p.Blobs, blobJSON{
@@ -355,14 +350,12 @@ func (idx *Index) generatePackList() ([]*packJSON, error) {
 		})
 	}
 
-	debug.Log("done")
-
 	return list, nil
 }
 
 type jsonIndex struct {
-	Supersedes restic.IDs  `json:"supersedes,omitempty"`
-	Packs      []*packJSON `json:"packs"`
+	Supersedes restic.IDs `json:"supersedes,omitempty"`
+	Packs      []packJSON `json:"packs"`
 }
 
 // Encode writes the JSON serialization of the index to the writer w.
--- a/internal/index/index_test.go
+++ b/internal/index/index_test.go
@@ -3,6 +3,7 @@ package index_test
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"math/rand"
 	"sync"
 	"testing"
@@ -403,6 +404,26 @@ func BenchmarkDecodeIndexParallel(b *testing.B) {
 			rtest.OK(b, err)
 		}
 	})
+}
+
+func BenchmarkEncodeIndex(b *testing.B) {
+	for _, n := range []int{100, 1000, 10000} {
+		idx, _ := createRandomIndex(rand.New(rand.NewSource(0)), n)
+
+		b.Run(fmt.Sprint(n), func(b *testing.B) {
+			buf := new(bytes.Buffer)
+			err := idx.Encode(buf)
+			rtest.OK(b, err)
+
+			b.ResetTimer()
+			b.ReportAllocs()
+
+			for i := 0; i < b.N; i++ {
+				buf.Reset()
+				_ = idx.Encode(buf)
+			}
+		})
+	}
 }
 
 func TestIndexUnserializeOld(t *testing.T) {


greatroar:
I also tried changing debug.Log to

```
func Log(f string, args ...interface{}) {
	if !opts.isEnabled {
		return
	}
	doLog(f, args...)
}

func doLog(f string, args ...interface{}) {
	/* ... */
}
```

While debug.Log can then be inlined, the allocation still happens.

internal/index/index.go
```diff
@@ -317,30 +317,23 @@ type blobJSON struct {
 }
 
 // generatePackList returns a list of packs.
-func (idx *Index) generatePackList() ([]*packJSON, error) {
-	list := []*packJSON{}
-	packs := make(map[restic.ID]*packJSON)
```
MichaelEischer, line 322: I'm somewhat unsure whether we should get rid of the map. Strictly speaking the index implementation does not guarantee a 1:1 mapping between pack IDs and the packIndex. If a pack is contained in multiple indexes, then after merging multiple indexes, the same pack ID might correspond to multiple packIndexes. However, `MasterIndex.Save`, which is the main user of `generatePackList`, that won't be a problem.

MichaelEischer:
> While debug.Log can then be inlined, the allocation still happens.

I guess the escape analysis isn't able to fully analyze the vararg. If I remember correctly, then the allocation should even happen, if `Log` is empty.

internal/index/index.go
MichaelEischer: What would be the benefit of just removing the debug.Log calls?

greatroar:
```
name                 old time/op    new time/op    delta
EncodeIndex/100-8      1.56ms ± 2%    1.50ms ± 1%   -3.91%  (p=0.000 n=10+9)
EncodeIndex/1000-8     14.5ms ± 2%    13.3ms ± 2%   -8.38%  (p=0.000 n=9+9)
EncodeIndex/10000-8     120ms ± 2%     116ms ± 2%   -3.66%  (p=0.000 n=10+9)

name                 old alloc/op   new alloc/op   delta
EncodeIndex/100-8       306kB ± 1%     276kB ± 1%   -9.91%  (p=0.000 n=10+10)
EncodeIndex/1000-8     3.69MB ±11%    2.98MB ± 5%  -19.19%  (p=0.000 n=10+10)
EncodeIndex/10000-8    35.9MB ±11%    30.3MB ±17%  -15.58%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
EncodeIndex/100-8       3.39k ± 0%     2.48k ± 0%  -27.05%  (p=0.000 n=10+10)
EncodeIndex/1000-8      32.6k ± 0%     23.9k ± 0%  -26.73%  (p=0.000 n=10+10)
EncodeIndex/10000-8      326k ± 0%      239k ± 0%  -26.73%  (p=0.000 n=10+10)
```

But there's an intermediate option, where the list still changes from []*packJSON to []packJSON. Let me try that.

greatroar:
Pushed a version that has a map, but stores indices in the map instead of pointers to prevent the GC from scanning all its entries.

    name                 old time/op    new time/op    delta
    EncodeIndex/100-8      1.56ms ± 2%    1.48ms ± 3%   -5.37%  (p=0.000 n=10+10)
    EncodeIndex/1000-8     14.5ms ± 2%    13.1ms ± 2%   -9.49%  (p=0.000 n=9+10)
    EncodeIndex/10000-8     120ms ± 2%     116ms ± 2%   -3.58%  (p=0.000 n=10+10)
    
    name                 old alloc/op   new alloc/op   delta
    EncodeIndex/100-8       306kB ± 1%     275kB ± 1%  -10.28%  (p=0.000 n=10+10)
    EncodeIndex/1000-8     3.69MB ±11%    2.88MB ± 5%  -22.07%  (p=0.000 n=10+9)
    EncodeIndex/10000-8    35.9MB ±11%    31.9MB ±10%  -11.13%  (p=0.005 n=10+10)
    
    name                 old allocs/op  new allocs/op  delta
    EncodeIndex/100-8       3.39k ± 0%     2.39k ± 0%  -29.61%  (p=0.000 n=10+10)
    EncodeIndex/1000-8      32.6k ± 0%     22.9k ± 0%  -29.63%  (p=0.000 n=10+9)
    EncodeIndex/10000-8      326k ± 0%      229k ± 0%  -29.71%  (p=0.000 n=10+10)

greatroar:
Ugh. If I remove restic.ID.MarshalJSON, I can get even fewer allocations, but then encoding indexes gets slower:

```
name                 old time/op    new time/op    delta
EncodeIndex/100-8      1.48ms ± 3%    1.64ms ± 3%  +10.91%  (p=0.000 n=10+10)
EncodeIndex/1000-8     13.1ms ± 2%    15.9ms ± 8%  +21.08%  (p=0.000 n=10+10)
EncodeIndex/10000-8     116ms ± 2%     147ms ± 3%  +26.83%  (p=0.000 n=10+10)

name                 old alloc/op   new alloc/op   delta
EncodeIndex/100-8       275kB ± 1%     193kB ± 2%  -29.73%  (p=0.000 n=10+10)
EncodeIndex/1000-8     2.88MB ± 5%    2.46MB ±13%  -14.54%  (p=0.000 n=9+10)
EncodeIndex/10000-8    31.9MB ±10%    29.2MB ±25%     ~     (p=0.190 n=10+10)

name                 old allocs/op  new allocs/op  delta
EncodeIndex/100-8       2.39k ± 0%     1.38k ± 0%  -42.13%  (p=0.000 n=10+10)
EncodeIndex/1000-8      22.9k ± 0%     13.3k ± 0%  -42.18%  (p=0.000 n=9+10)
EncodeIndex/10000-8      229k ± 0%      132k ± 0%  -42.28%  (p=0.000 n=10+10)
```

This warrants some more research, but I'll leave that out of this PR.

MichaelEischer (approved):
LGTM. Thanks for optimizing the method! Although I have to admit that I've been wondering a bit whether the additional complexity (the map now represents an additional indirection) is worth it.

PR #4181, opened 2023-01-29
repository: Increase zstd max window size

greatroar:

The default window size is 8MiB for --compression=auto/zstd.SpeedDefault and 32MiB for --compression=max/zstd.SpeedBestCompression. This gives slightly better compression than the custom 512KiB we had. To test this, I backed up an Ubuntu installer image of 2.274GiB (after deduplication):

```
                stored      ratio
auto+512K       943.92M     2.46652
auto+default    934.61M     2.47173
max+512K        900.59M     2.58518
max+default     898.59M     2.59094
```

Memory use in the decoder should not matter, since we always decompress to memory and blocks have bounded size.

No.

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -365,9 +365,6 @@ func (r *Repository) getZstdEncoder() *zstd.Encoder {
 			// Disable CRC, we have enough checks in place, makes the
 			// compressed data four bytes shorter.
 			zstd.WithEncoderCRC(false),
-			// Set a window of 512kbyte, so we have good lookbehind for usual
-			// blob sizes.
-			zstd.WithWindowSize(512 * 1024),
 		}
 
 		enc, err := zstd.NewWriter(nil, opts...)


MichaelEischer:
Won't this also increase the memory usage to compress blobs quite a bit?

greatroar:
Not at all. zstd's hash table size depends on the compression level, not the window size, and the blob is in-memory anyway.

MichaelEischer:
I'm confused, https://github.com/klauspost/compress/blob/master/zstd/encoder_options.go#L95 states the following:

> // A larger value will enable better compression but allocate more memory and,
> // for above-default values, take considerably longer.

And for example the `fastBase` encoder also seems to include a history slice which matches the maximum window size.

greatroar:
Yes, but that matters only when doing streaming. We have the full blob in RAM when compressing and the full compressed blob in RAM when decompressing.

greatroar:
Actually, that's wrong. I double-checked the implementation and the history buffer is being built up, even by EncodeAll.

Issue #3326, opened 2021-03-08
Check and verify a specific snapshot


akrabu:
Output of `restic version`
--------------------------

restic 0.12.0 compiled with go1.15.7 on darwin/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

restic check --read-data --snapshot 9170996b
^ Would verify all data referenced by a specific snapshot

restic check --read-data-subset --snapshot 9a70996b
^ Would verify just the data added in that particular snapshot

What are you trying to do? What problem would this solve?
---------------------------------------------------------

It would be handy for testing purposes to ensure that a specific snapshot was readable.  For instance, if you had a connection hiccup in the middle of a specific backup and knew it, you could do a --read-data-subset --snapshot 1234abcd and verify that the blobs uploaded just during that snapshot were intact.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Always!  <3

aawsome:
I like the idea! If I find some time, I can try to work on a PR!

I would however change the syntax to 
```
restic check [flags] [snapshot ID] [...]
```
and let users specify the snapshot(s) like in other commends, e.g. `forget`. That would also allow to do a `restic check --read-data latest`.

What we should discuss:

- Should we allow snapshots only in combination with `--read-data*`?
- If not, how to handle non-packfile-errors in the repository which would not affect these?

akrabu:
Hmm that's a good question.  I'll defer to those who know better than me, but I could see where having the same options as, say, "find" might make sense.  So, if I have a corrupted database, and I do a --read-data and find the pack or tree affected, then do a rebuild-index and a fresh backup...  might it make sense to be able to read JUST that pack / tree?  Or does rebuild-index rename all of those packs / trees?  I'm not 100% clear on that.  If they stay the same, it might make sense.  If they don't, it might not haha

akrabu:
I just thought of a use-case outside of "snapshots" alone.  What about when you do a --read-data and have a connection hiccup, and you want to go back over JUST the data that didn't read properly during the hiccup?  For instance I have:

```
 rclone: 2021/03/09 00:25:11 ERROR : data/b7/b71340505ffd6cae4fcc9fb57c9ab6ed7a2450c912582636627ea0a29ce52d89: Didn't finish writing GET request (wrote 262144/4972912 bytes): unexpected EOF
Load(<data/b71340505f>, 0, 0) returned error, retrying after 720.254544ms: unexpected EOF
[...]

PR #4182, opened 2023-01-30
check: Document behavior of --with-cache a bit better

BenWiederhake:
The behavior of `--with-cache` came as a surprise to me, let's make sure this is a bit easier to understand.

This PR adds a sentence to the description of `check --with-cache`. Despite its name, it does not *only* use the cache for its operation, which is exactly what this PR adds in documentation.

I have auto-generated the files in `doc/man/` and committed the change. The contribution guideline says to not edit these files, and technically I didn't, I just committed the auto-generated change. I'm happy to remove that part of the commit if you wish.

Not really, but see the related discussion at https://forum.restic.net/t/how-to-speed-up-tiny-incremental-checks/5905

Checklist
---------

- [/] I have added tests for all code changes. → not really applicable
- [X] I have added documentation for relevant changes (in the manual). → That's all this PR does.
- [/] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)). → I don't think this warrants a changelog entry, does it?

### DIFF
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -65,7 +65,7 @@ func init() {
 		// MarkDeprecated only returns an error when the flag is not found
 		panic(err)
 	}
-	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use the cache")
+	f.BoolVar(&checkOptions.WithCache, "with-cache", false, "use existing cache, only read uncached data from repository")
 }
 
 func checkFlags(opts CheckOptions) error {


MichaelEischer:
Thanks for helping to improve the documentation. However, I'm not really sure that the description is helpful for the typical user.

> Despite its name, it does not _only_ use the cache for its operation, which is exactly what this PR adds in documentation.

I don't really get the point. Restic can only read data from the cache, which is actually stored in the cache. In fact, even `--read-data` loads pack files from the cache if they only contain tree blobs.

> `"use the cache for indexes, trees, etc. (blobs are still read from repo during --read-data[-subset])"`

That description contains far too many implementation details. `--with-cache` is also more accurately described by "use data from existing cache". Everything which is cached, is loaded from the cache. Everything else from the repository. And that's it.

BenWiederhake:
Well, the point is: `check --read-data-subset=1%` downloads far more than 1% of the repository size if the repository consists mostly of trees. This is because the first phase of `check` reads all trees from all snapshots.

Real-life example / underlying issue: I have a repo on a remote server that takes up 40G disk space; no compression enabled. When using `--read-data-subset=1%`, I would expect that around 400M are downloaded. I observed that this was not the case, and the check was I/O bound. The download took several minutes, even when lowering the percentage to `--read-data-subset=.05%`. It still took far more than four minutes; so I'd guess it downloaded 3-5 GB (the connection can sustain at least 10 MiB/s, usually 15 MiB/s).
Given the choice between just not `check`ing at all or throwing away snapshots, I took the third option: `--with-cache`. It behaves roughly the right way in my setup: It downloads significantly less, but *does* download something for a few dozen seconds. So that's what I'm using for now.

I agree that the description can still be improved, but that's the best I could come up with.

My intention is to make this approach a bit more visible, because evidently `--with-cache` doesn't use *only* the cache.

MichaelEischer:
> My intention is to make this approach a bit more visible, because evidently `--with-cache` doesn't use _only_ the cache.

That is "use the cache" reads to you as "[only] use the cache" instead of the intended "use the [existing] cache"? Then what about `use existing cache, only read not cached data from repository`? That would still suggest that it reduces the overall network traffic.

BenWiederhake:
Yes, that sounds good, although I'd like to suggest `use existing cache, only read uncached data from repository`, because `read not cached` looks a bit confusing at first glance, even though it is correct.

That aside: What do you think about a new flag like `--only-read-data`, that skips most of the checking? I'd be happy to implement it, but I'd like to hear your thoughts on it first. The idea is that a call like `restic check --read-data-subset=1% --only-read-data` doesn't do any of the "usual" checking (tree traversal), and thus the I/O is much more predictable.

MichaelEischer:
> Yes, that sounds good, although I'd like to suggest `use existing cache, only read uncached data from repository`

That's fine by me, please update the PR accordingly. And remove the modification of the man page, these are regenerate upon release.

> That aside: What do you think about a new flag like `--only-read-data`, that skips most of the checking? I'd be happy to implement it, but I'd like to hear your thoughts on it first. The idea is that a call like `restic check --read-data-subset=1% --only-read-data` doesn't do any of the "usual" checking (tree traversal), and thus the I/O is much more predictable.

That suggestion reminds me of #3326. You'd still have to run `check` normally from time to time, Besides that I haven't thought much about whether there's a better way to achieve a similar result. But that isn't a priority at the moment.

MichaelEischer (approved):
LGTM. Thanks!

Issue #3808, opened 2022-06-28
Double escaped string in tree json format


aawsome:
Output of `restic version`
--------------------------
restic 0.13.1 (v0.13.0-147-g88a8701fb) compiled with go1.17.9 on linux/amd64
(but applies to older restic versions, too)

How did you run restic exactly?
-------------------------------
I have a dir with dirname 
`_!(!!cg"p!(}!}@"j!(k!|w"w!'''8!b!"p!''':!e@==`
 (no kidding) and run `restic backup` with a path that contains that dir.

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
local backend
(but should be independent from the backend)

Expected behavior
-----------------
According to the [JSON specification](https://datatracker.ietf.org/doc/html/rfc8259#section-7), I expect that the quotation marks are escaped and nothing else.
So, I would expect the tree to look like
```
"name":"_!(!!cg\"p!(}!}@\"j!(k!|w\"w!'8!b!\"p!':!e@=="
```

Actual behavior
---------------
`restic cat blob <TREE-ID>` gives:
```
{"nodes":[{"name":"_!(!!cg\\\"p!(}!}@\\\"j!(k!|w\\\"w!'8!b!\\\"p!':!e@==","type":"dir","mode":2147484141,"mtime":"2020-04-23T20:32:52.919484424+02:00","atime":"2020-04-23T20:32:52.919484424+02:00","ctime":"2020-04-23T20:32:52.919484424+02:00","uid":1000,"gid":1000,"user":"alex","group":"alex","inode":1323366,"device_id":2054,"content":null,"subtree":"7fb495dd7dab2a5d9251a19ef43aadb31b1b98c45f9614c3a9495b0c85ab42e7"}]}
```

Even worse, if you generate a tree object like 
```
{"nodes":[{"name":"_!(!!cg\"p!(}!}@\"j!(k!|w\"w!'8!b!\"p!':!e@==","type":"dir","mode":2147484141,"mtime":"2020-04-23T20:32:52.919484424+02:00","atime":"2020-04-23T20:32:52.919484424+02:00","ctime":"2020-04-23T20:32:52.919484424+02:00","uid":1000,"gid":1000,"user":"alex","group":"alex","inode":1323366,"device_id":2054,"content":null,"subtree":"7fb495dd7dab2a5d9251a19ef43aadb31b1b98c45f9614c3a9495b0c85ab42e7"}]}
```
(this is what rustic produces), and run a restic command , e.g.
[...]

PR #4210, opened 2023-02-16
docs: Corrections/extra information in design.rst

### LINKED ISSUE #3725: Mode used in Node/Tree is specific to go and might change in future go versions?
Sorry for not using the template. This issue is not a concrete bug, but something I observed when developing [rustic](https://github.com/rustic-rs/rustic)

The point is that in `internal/restic/node.go` the following mask is used for the modes which are saved in the trees:
```
mask := os.ModePerm | os.ModeType | os.ModeSetuid | os.ModeSetgid | os.ModeSticky
```

The [docu](https://pkg.go.dev/io/fs#ModeType) says:
> The defined file mode bits are the most significant bits of the FileMode. The nine least-significant bits are the standard Unix rwxrwxrwx permissions. The values of these bits should be considered part of the public API and may be used in wire protocols or disk representations: they must not be changed, although new bits might be added. 

With other words: restic uses part of the non-public API and saves this information in the on-disc representation.

For rustic, this has the effect that I wasn't able to easily reproduce the file modes (even though the last nine bits were always correct). The effect is that when using rustic on an restic repository and backup an completely unchanged path, it correctly detects the parent snapshot and also correctly identifies that no file has been changed - but most or all trees are newly written as the modes changed. Hence, afterwards, two versions of the trees are saved which means double storage used for trees.

I don't know how stable this non-public API in go is, but if it will happen to change, users will experience the same when using restic compiled with different go versions on the same repository.

Actually I don't know a good way to solve this. My impulse would be to only save the mode umasked by `0777`, but this will simply enforce the unwanted behavior I depicted above right now...

aawsome:

Updates the design document. Corrects some wrong/misleading parts and adds information about tree meta data.

closes #3725
see also #3808

Checklist
---------

- [N/A] I have added tests for all code changes.
- [N/A] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)) - don't think that's worth it. Feel free to add one!
- [N/A] I have run `gofmt` on the code in all commits.

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -45,10 +45,12 @@ comparing its output to the file name. If the prefix of a filename is
 unique amongst all the other files in the same directory, the prefix may
 be used instead of the complete filename.
 
-Apart from the files stored within the ``keys`` directory, all files are
-encrypted with AES-256 in counter mode (CTR). The integrity of the
-encrypted data is secured by a Poly1305-AES message authentication code
-(sometimes also referred to as a "signature").
+Apart from the files stored within the ``keys`` and ``data``directories,
+all files are encrypted with AES-256 in counter mode (CTR). The integrity
+of the encrypted data is secured by a Poly1305-AES message authentication
+code (sometimes also referred to as a "signature").
+Files in the ``data`` directory ("pack files") consist of multiple parts
+which are all independently enccrypted and authenticated, see below.
 
 In the first 16 bytes of each encrypted file the initialisation vector
 (IV) is stored. It is followed by the encrypted data and completed by
@@ -298,8 +300,8 @@ example, the Pack ``73d04e61`` contains two data Blobs and one Tree
 blob, the plaintext hashes are listed afterwards. The ``length`` field
 corresponds to ``Length(encrypted_blob)`` in the pack file header.
 Field ``uncompressed_length`` is only present for compressed blobs and
-therefore is never present in version 1. It is set to the value of
-``Length(blob)``.
+therefore is never present in the Repository format version 1. It is set
+to the value of ``Length(blob)``.
 
 The field ``supersedes`` lists the storage IDs of index files that have
 been replaced with the current index file. This happens when index files
@@ -410,7 +412,9 @@ and pretty-print the contents of a snapshot file:
     {
       "time": "2015-01-02T18:10:50.895208559+01:00",
       "tree": "2da81727b6585232894cfbb8f8bdab8d1eccd3d8f7c92bc934d62e62e618ffdf",
-      "dir": "/tmp/testdata",
+      "paths": [
+        "/tmp/testdata"
+      ],
       "hostname": "kasimir",
       "username": "fd0",
       "uid": 1000,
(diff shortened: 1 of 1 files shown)


doc/design.rst
```diff
@@ -45,10 +45,12 @@ comparing its output to the file name. If the prefix of a filename is
...
-Apart from the files stored within the ``keys`` directory, all files are
-encrypted with AES-256 in counter mode (CTR). The integrity of the
-encrypted data is secured by a Poly1305-AES message authentication code
-(sometimes also referred to as a "signature").
+Apart from the files stored within the ``keys`` and ``data``directories,
+all files are encrypted with AES-256 in counter mode (CTR). The integrity
+of the encrypted data is secured by a Poly1305-AES message authentication
+code (sometimes also referred to as a "signature").
+Files in the ``data`` directory ("pack files") consist of multiple parts
+which are all independently enccrypted and authenticated, see below.
```
greatroar, line 53:
```suggestion
Apart from the files stored within the ``keys`` and ``data`` directories,
all files are encrypted with AES-256 in counter mode (CTR). The integrity
of the encrypted data is secured by a Poly1305-AES message authentication
code (sometimes also referred to as a "signature").
Files in the ``data`` directory ("pack files") consist of multiple parts
which are all independently encrypted and authenticated, see below.
```

doc/design.rst
```diff
@@ -298,8 +300,8 @@ example, the Pack ``73d04e61`` contains two data Blobs and one Tree
 blob, the plaintext hashes are listed afterwards. The ``length`` field
 corresponds to ``Length(encrypted_blob)`` in the pack file header.
 Field ``uncompressed_length`` is only present for compressed blobs and
-therefore is never present in version 1. It is set to the value of
-``Length(blob)``.
+therefore is never present in the Repository format version 1. It is set
```
rawtaz, line 303: Total nitpick but wouldn't ".. never present in version 1 of the repository format" be better wording? (Not sure why there would be a capital R in "repository" either, it's not in the earlier text).

doc/design.rst
```diff
@@ -495,9 +501,18 @@ the JSON is indented):
...
 A tree contains a list of entries (in the field ``nodes``) which contain
-meta data like a name and timestamps. When the entry references a
-directory, the field ``subtree`` contains the plain text ID of another
-tree object.
+meta data like a name and timestamps. Note that there are some specialities of how
+this metadata is generated:
+
+- The name is quoted using <https://pkg.go.dev/strconv#Quote> before being saved.
+  This handles non-unicode names, but also changes the representation of names
+  containing `"` or `\\`.
```
rawtaz, line 509: Should probably be dual backticks on each side of the words.

MichaelEischer (approved):
LGTM except for the nitpicks already mentioned by rawtaz.

MichaelEischer:
I've addressed the review comments and improve the link formatting.

PR #4212, opened 2023-02-17
cmd, restic: Refactor and fix snapshot filtering

### LINKED ISSUE #4211: restic dump latest has a bug with the arguments
Sorry for not following the process. 

```
restic version                                                                             
restic 0.15.1 compiled with go1.19.5 on darwin/amd64
```

Arguments in `restic dump` subcommand are mixed:
```
restic -r "..." dump --quiet latest --host "host-0" --path /export.dat /export.dat
Fatal: failed to find snapshot: snapshot filter (Paths:[host-0] Tags:[] Hosts:[/export.dat]): no snapshot found
```

[Method definition](https://github.com/restic/restic/blob/f137be42fea928d6682445304e9c15fe4754850e/internal/restic/snapshot_find.go#L89):

```
func FindFilteredSnapshot(ctx context.Context, be Lister, loader LoaderUnpacked, hosts []string, tags []TagList, paths []string, timeStampLimit *time.Time, snapshotID string) (*Snapshot, error)
```

[Actual use](https://github.com/restic/restic/blob/f137be42fea928d6682445304e9c15fe4754850e/cmd/restic/cmd_dump.go#L142):

```
sn, err := restic.FindFilteredSnapshot(ctx, repo.Backend(), repo, opts.Paths, opts.Tags, opts.Hosts, nil, snapshotIDString)
```

greatroar:

This turns snapshotFilterOptions from cmd into a restic.SnapshotFilter type and makes restic.FindFilteredSnapshot and FindFilteredSnapshots methods on that type. This fixes #4211 by ensuring that hosts and paths are named struct fields instead of unnamed function arguments in long lists of such.

Timestamp limits are also included in the new type. To avoid too much pointer handling, the convention is that time zero means no limit. That's January 1st, year 1, 00:00 UTC, which is so unlikely a date that we can sacrifice it for simpler code.

Fixes #4211.

Checklist
---------

### DIFF
--- a/internal/restic/snapshot_find.go
+++ b/internal/restic/snapshot_find.go
@@ -12,13 +12,32 @@ import (
 // ErrNoSnapshotFound is returned when no snapshot for the given criteria could be found.
 var ErrNoSnapshotFound = errors.New("no snapshot found")
 
-// findLatestSnapshot finds latest snapshot with optional target/directory, tags, hostname, and timestamp filters.
-func findLatestSnapshot(ctx context.Context, be Lister, loader LoaderUnpacked, hosts []string,
-	tags []TagList, paths []string, timeStampLimit *time.Time) (*Snapshot, error) {
+// A SnapshotFilter denotes a set of snapshots based on hosts, tags and paths.
+type SnapshotFilter struct {
+	_ struct{} // Force naming fields in literals.
+
+	Hosts []string
+	Tags  TagLists
+	Paths []string
+	// Match snapshots from before this timestamp. Zero for no limit.
+	TimestampLimit time.Time
+}
+
+func (f *SnapshotFilter) empty() bool {
+	return len(f.Hosts)+len(f.Tags)+len(f.Paths) == 0
+}
+
+func (f *SnapshotFilter) matches(sn *Snapshot) bool {
+	return sn.HasHostname(f.Hosts) && sn.HasTagList(f.Tags) && sn.HasPaths(f.Paths)
+}
+
+// findLatest finds the latest snapshot with optional target/directory,
+// tags, hostname, and timestamp filters.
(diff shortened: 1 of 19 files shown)


internal/restic/snapshot_find.go
```diff
@@ -12,13 +12,32 @@ import (
 // ErrNoSnapshotFound is returned when no snapshot for the given criteria could be found.
 var ErrNoSnapshotFound = errors.New("no snapshot found")
 
-// findLatestSnapshot finds latest snapshot with optional target/directory, tags, hostname, and timestamp filters.
-func findLatestSnapshot(ctx context.Context, be Lister, loader LoaderUnpacked, hosts []string,
-	tags []TagList, paths []string, timeStampLimit *time.Time) (*Snapshot, error) {
+// A SnapshotFilter denotes a set of snapshots based on hosts, tags and paths.
+type SnapshotFilter struct {
```
MichaelEischer, line 16: (Just a question, not a request to change the code) I'm curious why `FindLatest` and `FindAll` are methods of `SnapshotFilter` instead of passing a `SnapshotFilter` option struct to those methods. That is what is the benefit compared to a signature like `func FindLatest(ctx context.Context, be Lister, loader LoaderUnpacked, snapshotID string, opts *SnapshotFilter) (*Snapshot, error)`?

MichaelEischer (approved):
Please add a small changelog to describe the fixed bug. Besides that LGTM. And thanks a lot for fixing the issue!

internal/restic/snapshot_find.go
greatroar: Mainly because FindLatest and FindAll are too general as names for snapshot finding functions in package restic, but I found the old names too long and not descriptive enough. In other words, the type is used as a namespacing device.

greatroar:
Added the changelog entry.

internal/restic/snapshot_find.go
MichaelEischer: Thanks for the explanation.

Issue #4245, opened 2023-03-12
Broken links in the documentation


MichaelEischer:
Expected behavior
-----------------
`make -C doc linkcheck` should succeed.

Actual behavior
---------------
Lots of broken links:
```
(030_preparing_a_new_repo: line  276) broken    https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro - Anchor 'access-bucket-intro' not found
(030_preparing_a_new_repo: line  693) broken    https://github.com/restic/restic/pull/1657#issuecomment-377707486 - Anchor 'issuecomment-377707486' not found
(090_participating: line   32) broken    https://golang.org/pkg/path/#Glob - Anchor 'Glob' not found
(030_preparing_a_new_repo: line  475) broken    https://help.backblaze.com/hc/en-us/articles/360047425453-Getting-Started-with-the-S3-Compatible-API - 403 Client Error: Forbidden for url: https://help.backblaze.com/hc/en-us/articles/
(090_participating: line  103) broken    https://pgp.mit.edu/pks/lookup?op=get&search=0xCF8F18F2844575973F79D4E191A6868BD3F7A907 - HTTPSConnectionPool(host='pgp.mit.edu', port=443): Max retries exceeded with url: /pks/lookup?op=get&
(030_preparing_a_new_repo: line  296) broken    https://minio.io/downloads/#minio-server - Anchor 'minio-server' not found
(030_preparing_a_new_repo: line  328) broken    https://wasabi-support.zendesk.com/hc/en-us/articles/360015106031-What-are-the-service-URLs-for-Wasabi-s-different-regions- - 403 Client Error: Forbidden for url: https://wasabi-support.zendesk.
(       110_talks: line   32) broken    https://videoag.fsmpi.rwth-aachen.de/?view=player&lectureid=4442#content - Anchor 'content' not found
```

And even more redirects (not sure which ones we should change in the documentation, probably just the permanent redirects?):
```
(030_preparing_a_new_repo: line  564) redirect  https://cloud.google.com/iam/docs/service-accounts - permanently to https://cloud.google.com/iam/docs/service-account-overview
(  070_encryption: line   18) redirect  https://blog.filippo.io/restic-cryptography/ - permanently to https://words.filippo.io/restic-cryptography/
(030_preparing_a_new_repo: line  584) redirect  https://cloud.google.com/docs/authentication/production - permanently to https://cloud.google.com/docs/authentication#service-accounts
[...]

PR #4246, opened 2023-03-12
doc: Fix links, focus less on implementation details

greatroar:

The godoc for filepath.Match has the syntax, which is what is important for writing patterns. Use pkg.go.dev instead of golang.org/pkg.

The links are listed at #4245.

Checklist
---------


rawtaz (approved):
I checked all the changes and URLs and they look good to me.

PR #4254, opened 2023-03-16
add --stdin-from-command flag to backup command

### LINKED ISSUE #4251: Backup flag to execute commands with restic, read content from standard input and consider return status
Output of `restic version`
--------------------------

restic 0.15.1 compiled with go1.20 on linux/amd64

When using `restic backup` command with a Pipe and `--stdin` flag, restic is not able to detect when other processes together in the Pipe exit unexpected. As result restic creates valid snapshots whether other processes in the pipe are successful or not.

As discussed in the forum thread [Failed database dump results in valid snapshot](https://forum.restic.net/t/failed-database-dump-results-in-valid-snapshot/6014) the feature proposal is to implement a [borg like](https://borgbackup.readthedocs.io/en/stable/usage/create.html#reading-from-stdin) `--stdin-from-command` flag. With this flag restic should be able to execute commands directly, read the stdin output, monitor their return code and decide whether a snapshot is to be created or not.

My expectation is that the following things will be taken into account when restic is handling additional commands to read from stdin:

- exit code is handled and is part of the decision whether to create a snapshot or not
- environment variables are also visible to the additional command
- the difference between normale Pipes and the new flag is described in the documentation
- possible safety concerns when executing commands directly with restic have been considered (Are there any?)

To create database backups I use mysql/pgdump commands and restic in stdin mode, running in a bash script with `set -e` and `set -o pipefail`. In some situations the databases dumps fail but restic still result in a valid snapshot.

Minimal reproducer:
```
$ pg_dump -h foobar 2>&1; echo $?
pg_dump: error: could not translate host name "foobar" to address: No address associated with hostname
1

$ pg_dump -h foobar 2>&1 | restic -p tmp/password -r tmp/restic_repo backup --stdin --stdin-filename dump.sql
repository b7cd6922 opened (version 2, compression level auto)

Added to the repository: 410 B (422 B stored)

processed 1 files, 103 B in 0:00
snapshot 7635bd5b saved
```
(redirecting stderr to stdout for demonstration purposes only)

I've been using restic for years and it saved my ass several times or at least saved me a lot of time. Not just because it's fast and efficient, but also because it's so easy to use and has an intuitive and not bloated user interface with very helpful documentation and messages.

sebhoss:

This fixes #4251 by providing the requested `--stdin-from-command` flag to the `restic backup` command. It skips snapshot creation in case the given commands fails as determined by `exec.Cmd.Wait()` and returns the stderr of the given command as an error message back to the user.

I have not written any tests, nor documentation because I wanted to get some feedback about the path taken here wrt. upstream inclusion first. I am seeing various test failures in existing tests but these seem to be unrelated and might just occur because I'm building/testing the project within a container. Using these changes locally does look promising though:

```
$ ./restic backup --stdin-from-command --repo ./backup -- pg_dump -h foobar
repository 55e7d5a0 opened (version 2, compression level auto)
using parent snapshot cc6e1dec
error: read /stdin: no data read
Fatal: unable to save snapshot: pg_dump: error: connection to database "root" failed: could not translate host name "foobar" to address: No address associated with hostname
```

Closes #4251

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -669,6 +669,8 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		Time:           timeStamp,
 		Hostname:       opts.Host,
 		ParentSnapshot: parentSnapshot,
+		Command:        command,
+		CommandStderr:  stderr,
 	}
 
 	if !gopts.JSON {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -97,6 +97,7 @@ type BackupOptions struct {
 	ExcludeLargerThan string
 	Stdin             bool
 	StdinFilename     string
+	StdinCommand      bool
 	Tags              restic.TagLists
 	Host              string
 	FilesFrom         []string
@@ -134,6 +135,7 @@ func init() {
 	f.StringVar(&backupOptions.ExcludeLargerThan, "exclude-larger-than", "", "max `size` of the files to be backed up (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&backupOptions.Stdin, "stdin", false, "read backup from stdin")
 	f.StringVar(&backupOptions.StdinFilename, "stdin-filename", "stdin", "`filename` to use when reading from stdin")
+	f.BoolVar(&backupOptions.StdinCommand, "stdin-from-command", false, "execute command and store its stdout")
 	f.Var(&backupOptions.Tags, "tag", "add `tags` for the new snapshot in the format `tag[,tag,...]` (can be specified multiple times)")
 	f.UintVar(&backupOptions.ReadConcurrency, "read-concurrency", 0, "read `n` files concurrently (default: $RESTIC_READ_CONCURRENCY or 2)")
 	f.StringVarP(&backupOptions.Host, "host", "H", "", "set the `hostname` for the snapshot manually. To prevent an expensive rescan use the \"parent\" flag")
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -289,7 +289,7 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 		}
 	}
 
-	if opts.Stdin {
+	if opts.Stdin || opts.StdinCommand {
 		if len(opts.FilesFrom) > 0 {
 			return errors.Fatal("--stdin and --files-from cannot be used together")
 		}
@@ -300,7 +300,7 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 			return errors.Fatal("--stdin and --files-from-raw cannot be used together")
 		}
 
-		if len(args) > 0 {
+		if len(args) > 0 && opts.StdinCommand == false {
 			return errors.Fatal("--stdin was specified and files/dirs were listed as arguments")
 		}
 	}
@@ -368,7 +368,7 @@ func collectRejectFuncs(opts BackupOptions, repo *repository.Repository, targets
 
 // collectTargets returns a list of target files/dirs from several sources.
 func collectTargets(opts BackupOptions, args []string) (targets []string, err error) {
-	if opts.Stdin {
+	if opts.Stdin || opts.StdinCommand {
 		return nil, nil
 	}
 
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"runtime"
@@ -588,16 +589,37 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		defer localVss.DeleteSnapshots()
 		targetFS = localVss
 	}
-	if opts.Stdin {
+
+	var command *exec.Cmd
+	var stderr io.ReadCloser
+	if opts.Stdin || opts.StdinCommand {
 		if !gopts.JSON {
 			progressPrinter.V("read data from stdin")
 		}
 		filename := path.Join("/", opts.StdinFilename)
+		var closer io.ReadCloser
+		if opts.StdinCommand {
+			command = exec.CommandContext(ctx, args[0], args[1:]...)
+			stdout, err := command.StdoutPipe()
+			if err != nil {
+				return err
+			}
+			stderr, err = command.StderrPipe()
+			if err != nil {
+				return err
+			}
+			if err := command.Start(); err != nil {
+				return err
+			}
+			closer = stdout
+		} else {
+			closer = os.Stdin
+		}
 		targetFS = &fs.Reader{
 			ModTime:    timeStamp,
 			Name:       filename,
 			Mode:       0644,
-			ReadCloser: os.Stdin,
+			ReadCloser: closer,
 		}
 		targets = []string{filename}
 	}
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -2,7 +2,9 @@ package archiver
 
 import (
 	"context"
+	"io"
 	"os"
+	"os/exec"
 	"path"
 	"runtime"
 	"sort"
@@ -680,6 +682,8 @@ type SnapshotOptions struct {
 	Excludes       []string
 	Time           time.Time
 	ParentSnapshot *restic.Snapshot
+	Command        *exec.Cmd
+	CommandStderr  io.ReadCloser
 }
 
 // loadParentTree loads a tree referenced by snapshot id. If id is null, nil is returned.
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -795,6 +795,15 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 		return nil, restic.ID{}, err
 	}
 
+	if opts.Command != nil {
+		errBytes, _ := io.ReadAll(opts.CommandStderr)
+		cmdErr := opts.Command.Wait()
+		if cmdErr != nil {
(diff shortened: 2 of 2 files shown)


Enrico204:
LGTM! I think that this option should be strongly suggested in the documentation in place of the current "stdin procedure".

Maybe we want to unset some sensible environment variables? (e.g., `RESTIC_PASSWORD`)

sebhoss:
Thanks for the review!

That's a good point about environment variables - I completely missed that requirement from the original ticket and thus the current code executes without any environment variables. However inheriting them and unsetting some sensible ones makes sense to me!

Enrico204:
AFAIK, `exec.CommandContext` is creating a `exec.Cmd` with the default value (`nil`) in the `.Env` field. According to the documentation, a `nil` value in `.Env` will instruct Go to inherit the current environment set. So, the current code is already passing all env to the child process ([which seems that Borg does](https://github.com/borgbackup/borg/commit/00b09370c0078b3a6c8f522c0e6ebeea265d100c)).

A very rough way might be to remove [all prefixed variables indicated in the documentation](https://restic.readthedocs.io/en/stable/040_backup.html#environment-variables):

```diff
diff --git a/cmd/restic/cmd_backup.go b/cmd/restic/cmd_backup.go
index ee355bb42..2068d5e6b 100644
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -608,6 +608,20 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
                        if err != nil {
                                return err
                        }
+
+                       for _, v := range os.Environ() {
+                               if !strings.HasPrefix(v, "RESTIC_") &&
+                                       !strings.HasPrefix(v, "AWS_") &&
+                                       !strings.HasPrefix(v, "ST_") &&
+                                       !strings.HasPrefix(v, "OS_") &&
+                                       !strings.HasPrefix(v, "B2_") &&
+                                       !strings.HasPrefix(v, "AZURE_") &&
+                                       !strings.HasPrefix(v, "GOOGLE_") {
+
+                                       command.Env = append(command.Env, v)
+                               }
+                       }
+
                        if err := command.Start(); err != nil {
                                return err
                        }
```

Note: I haven't tested this code.

Other options include:
- resetting the environment variables list to a safe default; or
- remove specific variables (e.g. `RESTIC_PASSWORD`), leaving others untouched

I'd argue that the the first option is too restrictive, and the second is too much maintenance burden (maintainers/developers need to remember to add new sensitive variables in the list). So, resetting variables by their prefix may be a tradeoff.

What do you think?

sebhoss:
Ah nice, I guess I misread the docs - thanks again!

Thinking about unsetting variables a bit more I'm concerned that it might be too magical for users since they cannot influence it directly. So maybe another option like `--inherit-env=true|false` would make sense? Using this flag, users could control wheher the command they are executing inherits all or none of the availble env variables.

Enrico204:
While I agree that it might seem magical, I think too many knobs are equally bad. I don't see any use case for `RESTIC_*` variables to be passed to the child program, and, in any case, one can still "restore" variables for the child command using `env`. For example:

```sh
$ restic backup \
   --stdin-from-command \
   --repo ./backup \
   -- env RESTIC_REPOSITORY=$RESTIC_REPOSITORY pg_dump -h foobar
```

So, if a variable is removed, it can be re-added explicitly (it's safe - a pattern might match more/fewer variables tomorrow).

This needs to be clearly documented, of course :-)

sebhoss:
Yeah I agree that too many knobs do not feel that good either. I think in the end I would probably just leave this as-is given that 1) borg seems to be fine with that 2) the current `--stdin` procedure works like that as well and 3) users can overwrite/unset variables as you have shown in your last snippet anyway.

Enrico204:
I am testing your branch for some real world backup in my personal/lab. For now, everything is working as expected :-)

cmd/restic/cmd_backup.go
```diff
@@ -298,7 +301,7 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 			return errors.Fatal("--stdin and --files-from-raw cannot be used together")
 		}
 
-		if len(args) > 0 {
+		if len(args) > 0 && opts.StdinCommand == false {
```
MichaelEischer, line 304: please use `&& !opts.StdinCommand`

internal/archiver/archiver.go
```diff
@@ -791,6 +795,15 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
 		return nil, restic.ID{}, err
 	}
 
+	if opts.Command != nil {
+		errBytes, _ := io.ReadAll(opts.CommandStderr)
```
MichaelEischer, line 799: This will deadlock if a process writes too much data to stderr. Please handle stderr similar as in the rclone/sftp backend.

cmd/restic/cmd_backup.go
```diff
@@ -645,6 +669,8 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		Time:           timeStamp,
 		Hostname:       opts.Host,
 		ParentSnapshot: parentSnapshot,
+		Command:        command,
+		CommandStderr:  stderr,
```
MichaelEischer, line 673: There's no need to modify the archiver. Instead, the `Close() error` part of the `ReadCloser` of `fs.Reader` should wait for the command to exit and return an error if necessary.

internal/archiver/archiver.go
```diff
@@ -791,6 +795,15 @@ func (arch *Archiver) Snapshot(ctx context.Context, targets []string, opts Snaps
...
+		cmdErr := opts.Command.Wait()
+		if cmdErr != nil {
+			debug.Log("error while executing command: %v", cmdErr)
+			return nil, restic.ID{}, errors.New(string(errBytes))
```
MichaelEischer, line 803: `cmdErr` should not be dropped. It should at least be part of the returned error message.

cmd/restic/cmd_backup.go
```diff
@@ -586,16 +589,37 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
...
+	var command *exec.Cmd
+	var stderr io.ReadCloser
+	if opts.Stdin || opts.StdinCommand {
 		if !gopts.JSON {
 			progressPrinter.V("read data from stdin")
 		}
 		filename := path.Join("/", opts.StdinFilename)
+		var closer io.ReadCloser
+		if opts.StdinCommand {
+			command = exec.CommandContext(ctx, args[0], args[1:]...)
```
MichaelEischer, line 602: Please move as much of the code that sets up the stdinCommand as possible out of `runBackup`. The latter function is already _far_ too long.

cmd/restic/cmd_backup.go
```diff
@@ -134,6 +136,7 @@ func init() {
 	f.StringVar(&backupOptions.ExcludeLargerThan, "exclude-larger-than", "", "max `size` of the files to be backed up (allowed suffixes: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&backupOptions.Stdin, "stdin", false, "read backup from stdin")
 	f.StringVar(&backupOptions.StdinFilename, "stdin-filename", "stdin", "`filename` to use when reading from stdin")
+	f.BoolVar(&backupOptions.StdinCommand, "stdin-from-command", false, "execute command and store its stdout")
```
MichaelEischer, line 139: I wonder whether `--content-from-command` (as in Borg) would be a better name or not. stdin-from-command somewhat contradicts its own description...

MichaelEischer (changes requested):
I'm in favor of adding a `--stdin-from-command` option (or maybe `--content-from-command`). Implementation-wise, it shouldn't be necessary to modify the archiver at all; the `fs.FS` interface should be generic enough to abstract away everything.

sebhoss:
> Implementation-wise, it shouldn't be necessary to modify the archiver at all; the fs.FS interface should be generic enough to abstract away everything.

Ok thanks for the pointer - that was exactly what I was looking for! So the idea here would be to :

1. create an implementation of `fs.FS` that handles command execution and related errors 
2. refactor the code in `runBackup` to create an instance of that implementation 
3. pass it in as the `targetFS` to the existing archiver methods

Is that correct? Should that `fs.FS` implementation live in the same `fs` package? Is [this](https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L43-L61) and [that](https://github.com/restic/restic/blob/master/internal/backend/sftp/sftp.go#L60-L70) how you would like to handle stderr output?

MichaelEischer:
> Ok thanks for the pointer - that was exactly what I was looking for! So the idea here would be to :
> 
>     1. create an implementation of `fs.FS` that handles command execution and related errors

Judging from https://github.com/restic/restic/pull/4254/files#diff-30d8672eafa9c0585766b6f437530377563cd1aac5606674c366f3ede4abea9eR618 an `fs.FS` implementation that wraps `fs.Reader` might work here.

>     2. refactor the code in `runBackup` to create an instance of that implementation
> 
>     3. pass it in as the `targetFS` to the existing archiver methods

Yes.

> Should that `fs.FS` implementation live in the same `fs` package? Is [this](https://github.com/restic/restic/blob/master/internal/backend/rclone/backend.go#L43-L61) and [that](https://github.com/restic/restic/blob/master/internal/backend/sftp/sftp.go#L60-L70) how you would like to handle stderr output?

The `fs` package is probably the best place for now. And yep, I was referring to the goroutine that reads from stderr of the subprocess.

MichaelEischer:
@sebhoss Are you still interested in working on this PR?

Enrico204:
I can take over this PR. I am using the code here on a daily basis at home for backups, and I will be happy to have it merged :-)

However, don't I think I will be able to continue in this PR as the branch is from @sebhoss's repo, so maybe I can fork & create a new pull request with the same code, and then apply the suggestions from @MichaelEischer. What do you think?

MichaelEischer:
@Enrico204 Sounds good. Go ahead and create a new PR, I'll close the old PR afterwards.

MichaelEischer:
Superseded by #4410.

PR #4255, opened 2023-03-17
fuse: Mix inode hashes in a non-symmetric way

### LINKED ISSUE #4253: Filesystem loop in `mount` when folder has same name as parent
Output of `restic version`
--------------------------
restic 0.15.1 compiled with go1.19.5 on linux/amd64

Steps to reproduce
Create a repository and backup a folder that contains a child folder that has the same name as its parent e.g.: `source/test/test`

`restic backup -r test source`
`restic mount -r test /mnt/tmp`

`find /mnt/tmp/snapshots/latest`
find: File system loop detected; ‘/mnt/tmp/snapshots/latest/source/test/test’ is part of the same file system loop as ‘/mnt/tmp/snapshots/latest/source’.

`~$ stat /mnt/tmp/snapshots/latest/source/test`
/mnt/tmp/snapshots/latest/source/test
Size: 0               Blocks: 0          IO Block: 4096   directory
Device: 48h/72d Inode: 6229125747021330212  Links: 2
Access: (0775/drwxrwxr-x)  Uid: ( 1000/      redacted)   Gid: ( 1000/      redacted)
Access: 2023-03-15 22:27:15.071794048 +0000
Modify: 2023-03-15 22:27:15.071794048 +0000
Change: 2023-03-15 22:27:15.071794048 +0000
Birth: -
`~$ stat /mnt/tmp/snapshots/latest/source/test/test`
/mnt/tmp/snapshots/latest/source/test/test
Size: 0               Blocks: 0          IO Block: 4096   directory
Device: 48h/72d Inode: 1850564384684095005  Links: 2
Access: (0775/drwxrwxr-x)  Uid: ( 1000/      redacted)   Gid: ( 1000/      redacted)
Access: 2023-03-15 22:27:15.071794048 +0000
Modify: 2023-03-15 22:27:15.071794048 +0000
Change: 2023-03-15 22:27:15.071794048 +0000
Birth: -

Confirmed with both Windows (NTFS) and Linux (ext4) filesystems as the source data. `restore` does not seem to be affected.

Restic v0.15.0 is the first release affected by this regression.

greatroar:

Since 0.15 (#4020), inodes are generated as hashes of names, xor'd with the parent inode. That means that the inode of a/b/b is

	h(a/b/b) = h(a) ^ h(b) ^ h(b) = h(a).

I.e., the grandchild has the same inode as the grandparent. GNU find trips over this because it thinks it has encountered a loop in the filesystem, and fails to search a/b/b. This happens more generally when the same name occurs an even number of times.

Fix this by multiplying the parent by a large prime, so the combining operation is not longer symmetric in its arguments. This is what the FNV hash does, which we used prior to 0.15. The hash is now

	h(a/b/b) = h(b) ^ p*(h(b) ^ p*h(a))

Note that we already ensure that h(x) is never zero.

Collisions can still occur, but they should be much less likely to occur within a single path.

Fixes #4253.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4253
+++ b/changelog/unreleased/issue-4253
@@ -1,0 +1,18 @@
+Bugfix: Mount command should no longer create spurious filesystem loops
+
+When a backup contains a directory that has the same name as its parent,
+say, a/b/b, and the GNU find command were run on this backup in a restic
+mount, find command would refuse to traverse the lowest "b" directory,
+instead printing "File system loop detected". This is due to the way the
+restic mount command generates inode numbers for directories in the mount
+point.
+
+The rule for generating these inode numbers was changed in 0.15.0. It has
+now been changed again to avoid this issue. A perfect rule does not exist,
+but the probability of this behavior occurring is now extremely small.
+When it does occur, the mount point is not broken, and scripts that traverse
+the mount point should work as long as they don't rely on inode numbers for
+detecting filesystem loops.
+
+https://github.com/restic/restic/issues/4253
+https://github.com/restic/restic/pull/4255
--- a/internal/fuse/fuse_test.go
+++ b/internal/fuse/fuse_test.go
@@ -226,6 +226,17 @@ func TestInodeFromNode(t *testing.T) {
 	ino1 = inodeFromNode(1, node)
 	ino2 = inodeFromNode(2, node)
 	rtest.Assert(t, ino1 != ino2, "same inode %d but different parent", ino1)
+
+	// Regression test: in a path a/b/b, the grandchild should not get the
+	// same inode as the grandparent.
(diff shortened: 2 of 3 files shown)


greatroar:
The changelog entry needs some reviewing. Also, there is an alternative solution that solves the original issue perfectly, which is to generate the inode number of a directory as its depth:

    if node.Type == "dir" {
        inode = 1 + parent
    }

but that has the obvious downside that a/b and a/x have the same inode, and similarly a/b/c and a/x/y. I bet find is going to trip over that too, especially when following symlinks to directories.

rawtaz:
Even I can understand the changelog 👌

MichaelEischer (approved):
LGTM. Thanks for debugging this issue!

PR #4265, opened 2023-03-22
generate: do not write progress to STDOUT

HeikoSchlittermann:
This allows usage as:

  eval $(restic generated --bash-completion /dev/stdout)

When running (in Bash) `eval "$(restic generate --bash-completion /dev/stdout)"`  fails, because the progress message is sent to stdout along with the completion functions. One way would be to "protect" the progress message using a hash sign, *or* as I suggest, sending the progress message to stderr (as it is *not* part of the _expected_ output).

No, it wasn't discussed, I just stumbled over it.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_generate.go
+++ b/cmd/restic/cmd_generate.go
@@ -63,22 +63,30 @@ func writeManpages(dir string) error {
 }
 
 func writeBashCompletion(file string) error {
-	Verbosef("writing bash completion file to %v\n", file)
+	if stdoutIsTerminal() {
+		Verbosef("writing bash completion file to %v\n", file)
+	}
 	return cmdRoot.GenBashCompletionFile(file)
 }
 
 func writeFishCompletion(file string) error {
-	Verbosef("writing fish completion file to %v\n", file)
+	if stdoutIsTerminal() {
+		Verbosef("writing fish completion file to %v\n", file)
+	}
 	return cmdRoot.GenFishCompletionFile(file, true)
 }
 
 func writeZSHCompletion(file string) error {
-	Verbosef("writing zsh completion file to %v\n", file)
+	if stdoutIsTerminal() {
+		Verbosef("writing zsh completion file to %v\n", file)
+	}
 	return cmdRoot.GenZshCompletionFile(file)
 }
 
 func writePowerShellCompletion(file string) error {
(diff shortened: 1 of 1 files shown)


cmd/restic/cmd_generate.go
```diff
@@ -63,22 +63,22 @@ func writeManpages(dir string) error {
 }
 
 func writeBashCompletion(file string) error {
-	Verbosef("writing bash completion file to %v\n", file)
+	Warnf("writing bash completion file to %v\n", file)
```
MichaelEischer, line 66: I'd prefer to just hide the message if stdout is not a terminal, as we normally use stderr only to print errors.
That is, please use `if stdoutIsTerminal() {` to only call `Verbosef` if stdout is a terminal.

MichaelEischer (changes requested):

cmd/restic/cmd_generate.go
HeikoSchlittermann: You're right. I'll provide an update soon.

cmd/restic/cmd_generate.go
```diff
@@ -63,22 +65,30 @@ func writeManpages(dir string) error {
 }
 
 func writeBashCompletion(file string) error {
-	Verbosef("writing bash completion file to %v\n", file)
+	if terminal.IsTerminal(int(os.Stdout.Fd())) {
```
MichaelEischer, line 68: Please use `stdoutIsTerminal()`. There's no need to copy paste the terminal detection code everywhere.

MichaelEischer (changes requested):

MichaelEischer (approved):
LGTM.

cmd/restic/cmd_generate.go
HeikoSchlittermann: Oh, I didn't see that function. I was even tempted now to put there another layer of abstraction as in
```
func openOutput(file string, format string, args ...any) (io.WriteCloser, error) {
  if file == "-" {
    return os.Stdout, nil
  }
  if stdoutIsTerminal() {
    Verbosef(format, args...)
  }
  return os.Create(file)
 
 ```
 And then use `cmdRoot.Gen*Completion(w)`

PR #4267, opened 2023-03-23
archiver: Shortcut for handling empty files

### LINKED ISSUE #4257: juicefs - restic hangs
Output of `restic version`
--------------------------
restic 0.15.1 compiled with go1.19.5 on linux/amd64

Server: Debian 11, x86_64, VPS, 2 VCPU, 2 GB RAM,  on Hetzner Cloud

```
export RESTIC_PASSWORD_FILE=/root/restic/password
export RESTIC_REPOSITORY_FILE=/root/restic/repository

/root/restic/bin/restic \
backup \
--verbose=2 \
--no-scan \
/mnt/my-juicefs-mount/

```
Output:

```
open repository
repository bf16cbdf opened (version 2, compression level auto)
lock repository
load index files
start backup on [/mnt/jfs-test/]
new       /mnt/jfs-test/.stats, saved in 0.003s (2.871 KiB added)
new       /mnt/jfs-test/.config, saved in 0.004s (0 B added)
new       /mnt/jfs-test/.trash/, saved in 0.001s (0 B added, 0 B stored, 0 B metadata)
[0:36] 2 files, 4.102 KiB, 0 errors
/mnt/jfs-test/.accesslog
```

sftp, usb drive on a Raspberry Pi 3 Model B Rev 1.2

Expected behavior
-----------------
Restic completes the backup with no error or with an error that the file .accesslog could not be backed up.

Actual behavior
---------------
Restic processes all but the last file on the filesystem, named .accesslog. Restic does not finish, just sits there for hours with the clock ticking.

1. create and mount an new juicefs filesystem (with MySQL as meta store and Linode S3 compatible object storage).
2. run restric backup on the mounted filesystem.

It has to do with  accessing the .accesslog file. 

A newly created juicefs filesystem contains these files by default:

```
root@he02:/mnt/jfs-test# ls -la
total 10
drwxrwxrwx 2 root root 4096 Mar 17 11:36 .
drwxr-xr-x 5 root root 4096 Mar 17 11:38 ..
-r-------- 1 root root    0 Mar 17 11:38 .accesslog
-r-------- 1 root root 1260 Mar 17 11:38 .config
-r--r--r-- 1 root root    0 Mar 17 11:38 .stats
dr-xr-xr-x 2 root root    0 Mar 17 11:38 .trash
```
If I exclude the file `.accesslog`, restic finishes with no errors, else it hangs after three files.

Juicefs has this to say about .`accesslog`:

> Every JuiceFS mount point contains virtual files called .oplog (also aliased to .accesslog from 4.7.0 and above) and .ophistory. They are text files that contain realtime file system stats like operation type, UID/GID, inode, and operation duration.

([Troubleshooting Methods | JuiceFS Document Center 2](https://juicefs.com/docs/cloud/administration/fault_diagnosis_and_analysis/))

No, but if I add `--exclude '/mnt/*/.accesslog'` restic finishes the backup successfully.

Restic smiled at me and made me feel good.

greatroar:

Regular files that are empty according to stat are now not opened for reading their contents. Such files are quite common (in my homedir, at least) and we can save multiple system calls this way. On a network filesystem, that can mean round trips. Also, we can back up empty files that we cannot open for reading. Finally, fixes #4257.

Existing tests cover this case. fs.Reader now no longer has a meaningful Size. Nothing depended on that.

I tested restic backup --stdin on the command line and it still works. It also still gives the same error for empty stdin.

Fixes #4257.

Checklist
---------

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -363,6 +363,26 @@ func (arch *Archiver) Save(ctx context.Context, snPath, target string, previous 
 
 	switch {
 	case fs.IsRegularFile(fi):
+		if fi.Size() == 0 {
+			// Shortcut for empty files. Git uses lots of these, and
+			// some virtual filesystems (notably juicefs; #4257) present
+			// infinitely-sized special files as empty regular files.
+			// We can also save empty files without being able to open them.
+			debug.Log("  %v empty", target)
+
+			node, err := arch.nodeFromFileInfo(snPath, target, fi)
+			if err != nil {
+				return FutureNode{}, false, err
+			}
+			node.Content = restic.IDs{}
+			fn = newFutureNodeWithResult(futureNodeResult{
+				snPath: snPath,
+				target: target,
+				node:   node,
+			})
+			return fn, false, nil
+		}
+
 		debug.Log("  %v regular file", target)
 
 		// check if the file has not changed before performing a fopen operation (more expensive, specially
--- a/internal/fs/fs_reader.go
+++ b/internal/fs/fs_reader.go
@@ -23,7 +23,6 @@ type Reader struct {
 	// for FileInfo
 	Mode    os.FileMode
 	ModTime time.Time
-	Size    int64
 
 	AllowEmptyFile bool
 
@@ -65,7 +64,6 @@ func (fs *Reader) Open(name string) (f File, err error) {
 func (fs *Reader) fi() os.FileInfo {
 	return fakeFileInfo{
 		name:    fs.Name,
-		size:    fs.Size,
 		mode:    fs.Mode,
 		modtime: fs.ModTime,
 	}
@@ -107,7 +105,6 @@ func (fs *Reader) Lstat(name string) (os.FileInfo, error) {
 	getDirInfo := func(name string) os.FileInfo {
 		fi := fakeFileInfo{
 			name:    fs.Base(name),
-			size:    0,
 			mode:    os.ModeDir | 0755,
 			modtime: time.Now(),
 		}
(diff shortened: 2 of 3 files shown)


HeikoSchlittermann:
Empty files may not be empty. What is about e.g. `/proc/cpuinfo`. Of course, if creating a backup of this directory is necessary, is another question. But anyway, I think, shortcuts like the one you suggest, are asking for trouble, sooner or later.

pvgoran:
> But anyway, I think, shortcuts like the one you suggest, are asking for trouble, sooner or later.

I second this opinion.

greatroar:
I don't find the /proc example compelling at all, and we have at least one example were an empty file not being empty is exactly what is causing trouble.

HeikoSchlittermann:
Programs should not try to be more clever than me ;)

HeikoSchlittermann:
> I'm not sure I fully understand you, but I definitely want the empty files to be there and not skipped. I just don't really want to open them when lstat says they are empty.

You are correct, I missed that point. Creating a list of files to back up would not help here. Just as the suggested _filter_ would not help. So maybe you're suggesting a new option as `--trust-file-size`

kjetilho:
I like this idea, but I am not sure if it should be opt-in or opt-out.

One thing to note is that one could consider generalising this to only backing up the N bytes reported from initial `stat`, this is what I would expect with an option name like `--trust-file-size`.  I don't think people will want that, e.g., they want the extra log lines in `/var/log/syslog` which may have appeared since backup start.  (In Bacula/Bareos, a warning is emitted if a file changed size while reading!)

Aside: In an ideal world, backup would run against a stable snapshot (like VSS in Windows), but this is not so common on Linux (dm-snapshot can do this with any block device).

And then - in a few cases this may happen with this PR: `/var/log/syslog` was zero bytes due to log rotatation just before backup start, so this patch will no longer backup the newest log lines.  It is a niche concern, but it makes me conclude that this behaviour should be opt-in, and then only for zero sized files.  `--trust-empty-size` perhaps?

MichaelEischer:
As it's not really clear what the "correct" approach is, let's stick with the current behavior that might back up a bit too much. This can be handled using excludes, whereas the other direction could silently miss data.

PR #4276, opened 2023-03-29
Implemented a web server to browse a repository (snapshots, files)

ducalex:

It adds a web server to browse a repository (snapshots, files). Overall it is a very simple patch and requires no additional dependencies but it might be very useful for people who can't use fuse mount.

It allows to browse snapshots, view files, and cherry pick files to download an archive.

Usage: `restic -r repo serve` and open your web browser to http://localhost:3080/

https://github.com/restic/restic/assets/13711380/aff76e96-85e0-40b5-a047-c9fd19f31956

It is related to #60 but after reading the discussion I don't know if this PR would be enough to close it.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_serve.go
+++ b/cmd/restic/cmd_serve.go
@@ -2,7 +2,10 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"net"
 	"net/http"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -35,6 +38,8 @@ func init() {
 	cmdFlags.StringVarP(&serveOptions.Listen, "listen", "l", "localhost:3080", "set the listen host name and `address`")
 }
 
+const serverShutdownTimeout = 30 * time.Second
+
 func runWebServer(ctx context.Context, opts ServeOptions, gopts GlobalOptions, args []string) error {
 	if len(args) > 0 {
 		return errors.Fatal("this command does not accept additional arguments")
@@ -57,10 +62,42 @@ func runWebServer(ctx context.Context, opts ServeOptions, gopts GlobalOptions, a
 		return err
 	}
 
-	srv := server.New(repo, snapshotLister, TimeFormat)
+	srv := http.Server{
+		BaseContext: func(l net.Listener) context.Context {
+			// just return the global context
+			return ctx
+		},
+		Handler: server.New(repo, snapshotLister, TimeFormat),
+	}
+
+	listener, err := net.Listen("tcp", opts.Listen)
+	if err != nil {
+		return fmt.Errorf("start listener: %v", err)
+	}
+
+	// wait until context is cancelled, then close listener
+	go func() {
+		<-ctx.Done()
+		Printf("gracefully shutting down server\n")
+
+		ctxTimeout, cancel := context.WithTimeout(context.Background(), serverShutdownTimeout)
+		defer cancel()
+
+		_ = srv.Shutdown(ctxTimeout)
+	}()
 
 	Printf("Now serving the repository at http://%s\n", opts.Listen)
 	Printf("When finished, quit with Ctrl-c here.\n")
 
-	return http.ListenAndServe(opts.Listen, srv)
+	err = srv.Serve(listener)
+
+	if errors.Is(err, http.ErrServerClosed) {
+		err = nil
+	}
+
+	if err != nil {
+		return fmt.Errorf("serve: %v", err)
+	}
+
+	return nil
 }
--- a/cmd/restic/cmd_serve.go
+++ b/cmd/restic/cmd_serve.go
@@ -1,0 +1,268 @@
+package main
+
+import (
+	"context"
+	"html/template"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/restic/restic/internal/dump"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/fs"
+	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/walker"
+)
+
+var cmdServe = &cobra.Command{
+	Use:   "serve",
+	Short: "runs a web server to browse a repository",
+	Long: `
+The serve command runs a web server to browse a repository.
+`,
+	DisableAutoGenTag: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runWebServer(cmd.Context(), serveOptions, globalOptions, args)
+	},
+}
+
+type ServeOptions struct {
+	Listen string
+}
+
+var serveOptions ServeOptions
+
+func init() {
+	cmdRoot.AddCommand(cmdServe)
+	cmdFlags := cmdServe.Flags()
+	cmdFlags.StringVarP(&serveOptions.Listen, "listen", "l", "localhost:3080", "set the listen host name and `address`")
+}
+
+type fileNode struct {
+	Path string
+	Node *restic.Node
+}
+
+func listNodes(ctx context.Context, repo restic.Repository, tree restic.ID, path string) ([]fileNode, error) {
+	var files []fileNode
+	err := walker.Walk(ctx, repo, tree, nil, func(_ restic.ID, nodepath string, node *restic.Node, err error) (bool, error) {
+		if err != nil || node == nil {
+			return false, err
+		}
+		if fs.HasPathPrefix(path, nodepath) {
+			files = append(files, fileNode{nodepath, node})
+		}
+		if node.Type == "dir" && !fs.HasPathPrefix(nodepath, path) {
+			return false, walker.ErrSkipNode
+		}
+		return false, nil
+	})
+	return files, err
+}
+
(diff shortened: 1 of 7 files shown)


rawtaz:
I like the idea. I've actually started the very same thing myself a while ago. Nothing to show though.

ducalex:
> I like the idea. I've actually started the very same thing myself a while ago. Nothing to show though.

I think I've now accomplished everything I wanted for this PR! Seems to work pretty well so far.

Do you mind having a quick look at the code to see if I did something obviously stupid before I undraft it?

cmd/restic/cmd_serve.go
```diff
@@ -0,0 +1,268 @@
...
+			for _, file := range files {
+				for _, name := range r.Form["name"] {
+					if name == file.Node.Name {
+						tree.Nodes = append(tree.Nodes, file.Node)
+					}
+				}
+			}
+			if len(tree.Nodes) > 0 {
+				filename := strings.ReplaceAll(strings.Trim(snapshotID+curPath, "/"), "/", "_") + ".tar.gz"
+				w.Header().Set("Content-Disposition", "attachment; filename="+filename)
```
Enrico204, line 124: [The `filename` parameter in `Content-Disposition` requires quotes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition). In this case, I suggest to change the code in this way:
```suggestion
				w.Header().Set("Content-Disposition", "attachment; filename=\""+filename+"\"")
```

cmd/restic/cmd_serve.go
```diff
@@ -0,0 +1,268 @@
...
+const stylesheetTxt = `
+h1,h2,h3 {text-align:center; margin: 0.5em;}
+table {margin: 0 auto;border-collapse: collapse; }
+thead th {text-align: left; font-weight: bold;}
+tbody.content tr:hover {background: #eee;}
+tbody.content a.file:before {content: '\1F4C4'}
+tbody.content a.dir:before {content: '\1F4C1'}
+tbody.actions td {padding:.5em;}
+table, td, tr, th { border: 1px solid black; padding: .1em .5em;}
+`
```
Enrico204, line 268: May I suggest to use the Go embedding (`go:embed`) for these files?

cmd/restic/cmd_serve.go
```diff
@@ -0,0 +1,268 @@
...
+{{end}}
+</tbody>
+</table>
+</body>
+</html>`
+
+const treePageTpl = `<html>
+<head>
+<link rel="stylesheet" href="/style.css">
+<title>{{.Title}} :: restic</title>
```
Enrico204, line 238: Uhm, what about adding viewport meta tags, so that mobile browsers are happy?
https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag

```html
<meta name="viewport" content="width=device-width, initial-scale=1" />
```

I haven't tested the current code, though.

Enrico204 (commented):
Interesting feature! LGTM!

cmd/restic/cmd_serve.go
ducalex: I considered embedding but it would mean finding the correct place to store assets and adding the extra build step for bundling and I assumed (perhaps incorrectly?) that if I did that, it would reduce the chances of this PR being accepted.

So in other words I agree on the principle, I'm just not sure I'm the best person to figure out all the details :).

Enrico204: Actually, Go embedding does not require extra steps: https://pkg.go.dev/embed

Given that these assets need to be present during the build only, I think that a sub-directory (e.g. `serve-assets/` ?) may be OK. Let's see what restic developers think about this solution :-)

ducalex:
Anyone willing to review? I'm happy to make any changes needed.

isarrider:
This will be awesome for usability - Thank you @ducalex

zanna-37:
@MichaelEischer any chances this can be reviewed? I tested it and works well. It would be a great improvement for windows users that cannot mount the repo easily.

I was able to rebase it quickly without knowledge of Go or restic, so pretty straightforward to integrate.
Thank you very much.

fd0:
I like this approach a lot! Before we can get this merged (and the implementation to handle HTML files correctly and fixes the XSS issue I've discovered by briefly testing it), the challenges mentioned in #4754 need to be addressed. :)

ducalex:
> I like this approach a lot! Before we can get this merged (and the implementation to handle HTML files correctly and fixes the XSS issue I've discovered by briefly testing it), the challenges mentioned in #4754 need to be addressed. :)

I think it's reasonable to delay the addition of WebDAV until you're done with your work and my quick draft clearly isn't production-worthy!

But, depending on the timeline you have in mind, it might still be worth merging this web server PR. Adding a random token to the URL and forcing html/svg to render as text/plain should negate any potential XSS issues.

I can appreciate that you have a vision and my implementation might not fit. But perhaps you can reshape my code into something that you can live with until your better version comes out?

I hope I do not come across as pushy, I just think we're really overdue for a convenient way to browse snapshots!

fd0:
> I can appreciate that you have a vision and my implementation might not fit. But perhaps you can reshape my code into something that you can live with until your better version comes out?

Yes, that's something we can indeed do. Especially if we just need to swap out the backend implementation, that's quite easy to do and won't be noticed by users at all.

> I hope I do not come across as pushy, I just think we're really overdue for a convenient way to browse snapshots!

For my personal use `fuse` was sufficient, but I can understand that it's inconvenient on macOS and not available on Windows, so I can see your need!

So, from my point of view we can start tackling the web server in parallel to the WebDAV server (and unify them once the latter is merged). The access control layer (e.g. a random password or so) we can even build in the context of the web server and then also use for the WebDAV server.

I can think of the following that needs to be done:
 * Build an access control layer
 * Move the code for the server to a new package below `internal/restic` somewhere, so that within the `cmd_serve.go` the server is only configured and run
 * Split the code for the server that handles serving blobs and the website that is returned
 * Embed the templates using the suggested embed functionality (that's really easy to do!)

Please let me know how we should split the work, I'd like to contribute. :)

ducalex:
I can certainly split the code and add an authentication layer that can be shared by the servers, do something like:

````
package server:
    internal/server/auth.go         some shared authenticator that handles .htpasswd like rest-server?
    internal/server/http.go          the current server (html)
    internal/server/webdav.go    the webdav server
````

Not quite sure about naming. I also don't quite know where to put the HTML templates, `internal/server/assets`?

Otherwise feel free to take the lead and lay down the structure that you have in mind!

**Comment by fd0** [comment]
I've started reworking the server but then ran out of time (for today). It does not work yet, I'll need to debug this next.

**Comment by fd0** [comment]
The server works now, I've split it out into its own package (directory) and embedded the assets, which are stored in their own files now.

File: cmd/restic/cmd_serve.go [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const serverShutdownTimeout = 30 * time.Second
+
+func runWebServer(ctx context.Context, opts ServeOptions, gopts GlobalOptions, args []string) error {
+	if len(args) > 0 {
+		return errors.Fatal("this command does not accept additional arguments")
+	}
+
+	ctx, repo, unlock, err := openWithExclusiveLock(ctx, gopts, gopts.NoLock)
```
MichaelEischer, line 48: `openWithReadLock` should be sufficient.

internal/server/assets/fs.go
```diff
@@ -0,0 +1,6 @@
+package assets
+
+import "embed"
+
+//go:embed *
```
MichaelEischer, line 5: Does that also embed `fs.go`? I'd prefer to move the go file to the parent folder.

internal/server/server.go
```diff
@@ -0,0 +1,237 @@
...
+		panic("index.html not found")
+	}
+
+	treePage := templates.Lookup("tree.html")
+	if treePage == nil {
+		panic("tree.html not found")
+	}
+
+	mux.HandleFunc("/tree/", func(rw http.ResponseWriter, req *http.Request) {
+		snapshotID, curPath, _ := strings.Cut(req.URL.Path[6:], "/")
```
MichaelEischer, line 41: The path surgery can be replaced with https://pkg.go.dev/net/http#StripPrefix

MichaelEischer (commented):
Just a few random remarks from my side.

cmd/restic/cmd_serve.go
fd0: Good point, fixed.

internal/server/assets/fs.go
fd0: Yes it does, I've changed it to only include `*.html` and `*.css`.

fd0: Yeah, we can move it there. I just needed to get rid of the `assets/` path component, but there's https://pkg.go.dev/io/fs#Sub

zanna-37:
Why hasn’t this been merged yet?
It functions well and is highly needed.

fd0:
Unfortunately I don't have the time for it at the moment, sorry for that.

isarrider:
too bad as this would surely make restic to more folks that are less technical...!

Issue #4199, opened 2023-02-09
Restic can fail with `Fatal: failed to refresh lock in time` on slow connections


MichaelEischer:
Output of `restic version`
--------------------------
0.15.1

Expected behavior
-----------------
Backups should also be possible over a slow network connection.

Actual behavior
---------------
When uploading over a slow network connection, it is possible that a backup or similar operation fails to refresh lock files within 22.5 minutes and fails with `Fatal: failed to refresh lock in time`. This has been reported on IRC and at least at https://github.com/restic/restic/issues/4157#issuecomment-1408669114 .

Steps to reproduce the behavior
-------------------------------
Use a very slow network, then wait for half an hour...

Do you have any idea what may have caused this?
-----------------------------------------------
restic only allows a limited number of concurrent backend connections. Once all are used for uploading pack files, then a lock file refresh is queued behind them. For a slow connection, it is quite possible that a 16MB pack file (with 5 parallel connections) takes longer than half an hour to upload.

Further fallout from #3569.

Do you have an idea how to solve the issue?
-------------------------------------------
Lock file refreshes should be able to bypass the connections limit.

gallofeliz:
Hello,

I've just switched from OVH(swift) because of excessive problems to B2 with S3 api. I've the same problem

```
Fatal: failed to refresh lock in time
error: no result
error: no result
error: no result
error: no result
Fatal: unable to save snapshot: context canceled
```

I've tryed twice and same problem. I will try to find an option to make 1 operation by 1, my upload limit is 50KB.

Regards

jwiegley:
I have this same problem when uploading to B2 when I set `--limit-upload 100`.

gallofeliz:
Hello,

A workaround is to limit concurrent uploads, or what I did and it works (but increase number of files I suppose) is to reduce pack size : RESTIC_PACK_SIZE=4

This is a workaround with risk to increase (maybe) GetObjectsList calls and so cost, and updating lock should not be in this limitation or use one of the concurrent limit dedicated to itself (just avoid to put 1 !)

Regards

GottemHams:
Instead of simulating a slow network, you can also just send a `STOP` signal to restic (`killall -STOP restic`). This pretty much suspends the entire process without killing it, so it can be resumed right where it left off (`killall -CONT restic`). I usually do this when I need to pause the backup to do some other disk intensive stuff, which worked fine until the PR you mentioned was merged. ;_;

jwiegley:
[...]

PR #4286, opened 2023-04-08
Normalize backend logging and connection limiting

MichaelEischer:
The generic logging of backend operations is factored out into `logger.Backend`. Limiting the number of concurrent backend operations and the basic parameter validation now happens in `sema.SemaphoreBackend`. In addition, the `Delete()` method most backends now uses a generic implementation provided by `backend.DefaultDelete()`.

The unified logging now consistently logs the called operation and its result. This is complemented with a log entry by the SemaphoreBackend once a semaphore token has been acquired.

The `List()` operation of a Backend no longer counts towards the connections limit, but that should not make much of a difference as restic normally only runs one list operation at a time. The duration for which a semaphore token is held increases slightly, but I don't expect this to be a problem, as the bottleneck in a Backend is the necessary IO. The key simplification of the SemaphoreBackend is that the complex semaphore usage in `backend.DefaultLoad` and the corresponding `openReader()` method can simply be extracted into the `Load()` method. This allows a clean separation of the connection count limiting and the regular backend operation.

No, but this PR is a prerequisite to address #4199 without creating a big mess.

Checklist
---------

### DIFF
--- a/internal/backend/sema/backend.go
+++ b/internal/backend/sema/backend.go
@@ -18,8 +18,8 @@ type connectionLimitedBackend struct {
 	sem semaphore
 }
 
-// New creates a backend that limits the concurrent operations on the underlying backend
-func New(be restic.Backend) restic.Backend {
+// NewBackend creates a backend that limits the concurrent operations on the underlying backend
+func NewBackend(be restic.Backend) restic.Backend {
 	sem, err := newSemaphore(be.Connections())
 	if err != nil {
 		panic(err)
--- a/internal/backend/sema/backend.go
+++ b/internal/backend/sema/backend.go
@@ -9,30 +9,30 @@ import (
 	"github.com/restic/restic/internal/restic"
 )
 
-// make sure that SemaphoreBackend implements restic.Backend
-var _ restic.Backend = &SemaphoreBackend{}
+// make sure that connectionLimitedBackend implements restic.Backend
+var _ restic.Backend = &connectionLimitedBackend{}
 
-// SemaphoreBackend limits the number of concurrent operations.
-type SemaphoreBackend struct {
+// connectionLimitedBackend limits the number of concurrent operations.
+type connectionLimitedBackend struct {
 	restic.Backend
 	sem semaphore
 }
 
 // New creates a backend that limits the concurrent operations on the underlying backend
-func New(be restic.Backend) *SemaphoreBackend {
+func New(be restic.Backend) restic.Backend {
 	sem, err := newSemaphore(be.Connections())
 	if err != nil {
 		panic(err)
 	}
 
-	return &SemaphoreBackend{
+	return &connectionLimitedBackend{
 		Backend: be,
 		sem:     sem,
 	}
 }
 
 // Save adds new Data to the backend.
-func (be *SemaphoreBackend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
+func (be *connectionLimitedBackend) Save(ctx context.Context, h restic.Handle, rd restic.RewindReader) error {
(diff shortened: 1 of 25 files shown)


internal/backend/sema/backend.go
```diff
@@ -0,0 +1,99 @@
...
+		panic(err)
+	}
+
+	return &SemaphoreBackend{
+		Backend: be,
+		sem:     sem,
+	}
+}
+
+func (be *SemaphoreBackend) validateAndGetToken(h restic.Handle) (func(), error) {
```
greatroar, line 35: I know this is supposed to prevent repetition, but I don't find "validate Handle and grab token from semaphore" an obvious action. If we put the debug.Log statement inside semaphore.GetToken, inlining this at the call sites wouldn't produce much longer code.

internal/backend/sema/backend.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/restic"
+)
+
+// make sure that SemaphoreBackend implements restic.Backend
+var _ restic.Backend = &SemaphoreBackend{}
+
+// SemaphoreBackend limits the number of concurrent operations.
+type SemaphoreBackend struct {
```
greatroar, line 17: Maybe LimitedBackend and rename the package semaphore?

internal/backend/sema/backend.go
```diff
@@ -0,0 +1,99 @@
...
+	restic.Backend
+	sem semaphore
+}
+
+// New returns a new backend that saves all data in a map in memory.
+func New(be restic.Backend) *SemaphoreBackend {
```
greatroar, line 23: sema.New doesn't quite express what this function does. NewBackend? Wrap? Limit?

internal/backend/sema/backend.go
```diff
@@ -0,0 +1,99 @@
...
+// make sure that SemaphoreBackend implements restic.Backend
+var _ restic.Backend = &SemaphoreBackend{}
+
+// SemaphoreBackend limits the number of concurrent operations.
+type SemaphoreBackend struct {
+	restic.Backend
+	sem semaphore
+}
+
+// New returns a new backend that saves all data in a map in memory.
```
greatroar, line 22: Copy-pasted comment.

greatroar:
Aside from nitpicks, nice work. -410 LoC!

internal/backend/sema/backend.go
MichaelEischer: I've inlined the method.

internal/backend/sema/backend.go
MichaelEischer: There is already a `rateLimitedBackend`, adding a LimitedBackend could be quite confusing in that regard.

internal/backend/sema/backend.go
MichaelEischer: The dry run backend and the RetryBackend also use a `New` function. I would settle for `NewBackend` although that adds a certain inconsistency regarding the existing backends.

internal/backend/sema/backend.go
greatroar: Fair enough.

MichaelEischer: what about `connectionLimitedBackend` then? That would be somewhat more descriptive.

MichaelEischer (commented):
~~LGTM.~~ This PR still needs some tests...

MichaelEischer:
LGTM. I've added some basic tests for the `connectionLimitedBackend`. With our currently debug log implementation, the log backend is currently not really testable.

PR #4298, opened 2023-04-21
Unified and slightly type-safer backend config parsing

MichaelEischer:
Motivation behind this PR:

While analyzing the  test failures of `TestLockSuccessfulRefresh` (which likely fails due to randomly slow fsync() calls), I noticed that it would be useful to be able to use the mem backend in tests. However, it is currently not possible to dynamically register a new backend type for use in `cmd/restic`.

To implement this, we'd need a Registry for backends with an interface similar to the following:
```go
type Registry struct {
    Register(scheme string, factory Factory)
    Lookup(scheme string) Factory
}

type Factory interface {
    ParseConfig(s string) (interface{}, error)
    Create(ctx context.Context, cfg interface{}, rt http.RoundTripper, lim limiter.Limiter) (restic.Backend, error)
    Open(ctx context.Context, cfg interface{}, rt http.RoundTripper, lim limiter.Limiter) (restic.Backend, error)
}
```

(The RoundTripper and Limiter part of that variant is somewhat ugly as not all Backends need them. I'd also prefer to avoid using an `interface{}` for the Config of a backend by using generics, however, that would require combining both ParseConfig+Create/Open into one function.)

Implemented in this PR:

The `parseConfig()` function in `cmd/restic/global.go` is now implemented in a fully generic manner an is no longer backend specific. The latter code parts have been moved into the individual backends.

The PR unfortunately contains a significant amount of code churn as a clean implementation (without using reflection) requires that the `ParseConfig` function of each backend returns a pointer to the configuration. The reason for that is that it is not easily possible to convert from `interface{}(Config)` to `interface{}(*Config)`. To reduce the risk of not converting some type assertions, `ParseConfig` now returns a concrete `*Config` which is converted to `interface{}` using a generic wrapper function. The test suite for a backend now uses the concrete type.

No.

Checklist
---------

### DIFF
--- a/cmd/restic/global.go
+++ b/cmd/restic/global.go
@@ -559,24 +559,10 @@ func parseConfig(loc location.Location, opts options.Options) (interface{}, erro
 
 	case "s3":
 		cfg := loc.Config.(s3.Config)
-		if cfg.KeyID == "" {
-			cfg.KeyID = os.Getenv("AWS_ACCESS_KEY_ID")
-		}
 
-		if cfg.Secret.String() == "" {
-			cfg.Secret = options.NewSecretString(os.Getenv("AWS_SECRET_ACCESS_KEY"))
+		if err := s3.ApplyEnvironment(&cfg); err != nil {
+			return nil, err
 		}
-
-		if cfg.KeyID == "" && cfg.Secret.String() != "" {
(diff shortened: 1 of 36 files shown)


MichaelEischer (commented):
LGTM.

PR #4351, opened 2023-06-02
Add --human-readable to ls and find output

### LINKED ISSUE #4159: Support `-h` argument to `ls -l`
Output of `restic version`
--------------------------

Support a `-h` flag on `restic -l <snapshot>` to print human-readable byte counts like `ls` does. From `ls` manpage:

```
-h, --human-readable       with -l and -s, print sizes like 1K 234M 2G etc.
```

`-h` in restic is already occupied by help, but combining `-h` and `-l` does not make sense so if that combination is encountered, take `-h` argument as `--human-readable`. Alternatively, remove the `-h` argument for help only via `--help` is more common.

Large file sizes are hard to read with `restic ls -l`:

```
$ restic ls -l b64a60a1
repository 9cbe0bce opened (repository version 2) successfully, password is correct
snapshot b64a60a1 of [/hugefile] filtered by [] at 2023-01-17 04:55:21.228268714 +0100 CET):
-rw-r--r--     0     0 345945001259 2023-01-17 04:55:21 /hugefile
```

Ideally, it should print `345G` in this case.

kenny-y-dev:
Closes #4159 

Modifies format module to add options for human readable storage size formatting, using size parsing already in ui/format. Cmd flag --human-readable added to ls and find commands. Additional option added to formatNode to support printing size in regular or new human readable format

Discussion in #4159 

Expanded scope of the original issue a bit after finding formatNode function was common between `ls` and `find`; figured both could benefit from the additional formatting option. Also added a test for format

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4159
+++ b/changelog/unreleased/issue-4159
@@ -11,3 +11,4 @@ values, with an appropriate suffix depending on the output size. For
 example, a value of `14680064` will be shown as `14.000 MiB`.
 
 https://github.com/restic/restic/issues/4159
+https://github.com/restic/restic/pull/4351
--- a/changelog/unreleased/issue-4159
+++ b/changelog/unreleased/issue-4159
@@ -1,0 +1,13 @@
+Enhancement: Add --human-readable flag to ls and find commands
+
+Previously, the `ls` and `find` commands showed with the `-l` option 
+showed size in bytes, and did not have an option for a more human 
+readable format that converted these longer numbers into values such 
+as MiB or GiB
+
+The new `--human-readable` option for `ls` and `find` when used with
+the `-l` option will convert longer size values into more human friendly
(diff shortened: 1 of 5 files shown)


kenny-y-dev:
If there's somewhere that I can add additional documentation please let me know; read in contributing not to edit anything that would be autogenerated. I added usage strings to the flags so I think that should be sufficient for any generated documentation though happy to add more elsewhere!

MichaelEischer (approved):
Thanks for the PR! I've made a few adjustments to the changelog and fix the test to work in any timezone.

@rawtaz GNU `ls` or `du` have both a `--block-size` and `--human-readable` option. `--block-size` only supports concrete units like `k`,`m` or `g` but no `h`. So, I'd prefer to stick to the traditional `--human-readable` option.

PR #4352, opened 2023-06-02
index: optimize go GC performance

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------

The current index implementation using a custom indexmap is not particular garbage-collection friendly. As each indexEntry and the bucket list contain pointers, this requires the garbage collector to chase pointers to random locations throughout the index.

This PR moves the indexEntries into an array and only adds the array index instead of a pointer. Thereby, the indexEntries and bucket list no longer contain pointers and are not scanned by the garbage collector. The array containing all indexEntries is implemented as https://en.wikipedia.org/wiki/Hashed_array_tree to minimize the memory overhead.

The top-layer of the hashed array tree (HAT) also has a size of O(sqrt(n)), which should make it cache efficient. The top-layer should be small enough to easily fit into the CPU cache and thus only adds little overhead compared to directly accessing an index entry via a pointer.

As a bonus, the indexmap.Each/Grow functions are now able to iterate over the indexEntries array instead of following random pointers. This significantly speeds up both functions.

## Results

The indirection through the HAT increases the index lookup duration by 20-40% (the average is probably between 20-30% but the results are pretty noisy as a result of the varying map entry layout due to the maphash randomization), although the MasterIndex based tests show numbers between 0 to 35% overhead.

Iterating over the index is now up to 8 times faster, which also translates to slightly faster index encoding.
GC is now nearly for free.

grow() is also pretty interesting, the performance changes are likely a result of a different order of the chained indexEntries.

Commits:

- fast-grow `index: implement indexmap.grow() without random access`
- fast-each `index: let indexmap.Each iterate in allocation order`
- hat-opt `index: remove redundant storage of indexmap size`
- hat-inline `index: Allow inlining of HAT`
- hat-base `index: store indexEntries in hashed array tree`
- remove-pointers `index: Remove pointers from within indexentrys`
- base - `index: add garbage collection benchmark`

<details>

<summary>benchstat base remove-pointers</summary>

```
name                                              old time/op    new time/op    delta
IndexMapHash-12                                     1.25µs ± 3%    1.27µs ± 1%    +1.62%  (p=0.000 n=15+15)
DecodeIndex-12                                       2.38s ± 4%     2.35s ± 1%      ~     (p=0.367 n=15+15)
DecodeIndexParallel-12                               2.31s ± 5%     2.38s ± 1%    +2.70%  (p=0.041 n=15+12)
[...]

PR #4353, opened 2023-06-02
Tune Go garbage collector

### LINKED ISSUE #3328: Optimize garbage collector behavior
Output of `restic version`
--------------------------

restic 0.12.0 (v0.12.0-6-g70568c5b-dirty) compiled with go1.14.7 on linux/amd64

From time to time, users get trouble with memory usage of restic. There have been many improvements in the recent time which tackled the data structures and internal memory representation of those as well as algorithmic optimizations which also led to decreased memory.

However, in the forum there are still situations where users need to play around with the `GOGC` environment variable to squeeze out the least bytes of unused memory to let restic run successfully, see e.g. [here](https://forum.restic.net/t/restic-prune-runtime-cannot-allocate-memory/3660).

As restic is no real-time application, garbage collection runs do not hurt and can be done quite often without much trouble. So I want to discuss if we should implement more GC runs within restic. I can think of two approaches:

- Set the `GOGC` parameter to a (suitable) low value using [SetGCPercent](https://golang.org/pkg/runtime/debug/#SetGCPercent). This would however prevent users from manually setting it to an even lower value, if needed.
- Identify "critical" points in the program logic where a GC run would be able to remove lots of temporary used data and manually run `runtime.GC()`. This could be e.g. after loading the index into memory.

Both approaches can be implemented independently.

Optimize the memory usage profile. Improve the end-user experience. 

Keeps my data safe without any trouble for quite a long time now :+1:

MichaelEischer:
Reduce GOGC to 50. This leads to a decrease in memory usage by 25%.

The index used by restic consumes a major part of the total memory usage. This leads to an unnecessarily large amount of memory that contains ephemeral objects that are only used for a short time.

Fixes #3328
Depends on https://github.com/restic/restic/pull/4352 for reasonable GC performance.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3328
+++ b/changelog/unreleased/issue-3328
@@ -1,0 +1,5 @@
+Enhancement: Reduce memory usage by up to 25%
+
+https://github.com/restic/restic/issues/3328
+https://github.com/restic/restic/pull/4352
+https://github.com/restic/restic/pull/4353
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"math"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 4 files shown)


MichaelEischer:
LGTM.

PR #4366, opened 2023-06-10
Issue #3624: Preserve oldest snapshot when keep-* values are not sati…

### LINKED ISSUE #3624: restic forget: Keep also oldest snapshot when there are not enough backups to keep
Output of `restic version`
--------------------------
restic 0.12.1 compiled with go1.16.6 on linux/arm

When forgetting snapshots and there aren't enough snapshots to keep, keep also the oldest snapshot. Especially with yearly backups it can otherwise take a long time until forget policy is working the desired way.

I'm trying to keep two yearly backups, so I can go back in time at least for one year. 

A full description together with the suggestion to open a feature request, can be found here https://forum.restic.net/t/keeping-yearly-snapshots-policy-when-backup-began-during-the-year/4670

This feature request also helps losing data. As in the forum described I would have lost my oldest backup because restic forget works in this example different than expected. In my opinion it is not intuitive the way it currently works and you must take care not to delete too much snapshots. This feature request prevents this.

Restic helps me everyday because it makes me feel safe, that my data is safe when I accidentially delete something or HDD is damaged. I've tried a lot backup solution including recovery process. None of them was really convincing. Restic is the first backup solution I'm really happy with and there is also so much potential to make restic even better.

gautammenghani:
Currently, restic 'forget' command does not preserve the oldest snapshot if 'keep-\*' options are not satisfied. Due to this, users sometimes cannot preserve their oldest snapshots depending on the 'keep-\*' options they have specified.

When applying the snapshot policy, there is an additional check which checks if we are currently looking at the oldest snapshot and the count for that particular bucket is > 0, we mark it for preservation.

Yes. Please see #3624 

Closes #3624 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3624
+++ b/changelog/unreleased/issue-3624
@@ -1,11 +1,8 @@
-Enhancement: Keep oldest snapshot when there aren't enough snapshots
+Enhancement: Keep oldest snapshot when there are not enough snapshots
 
-The `forget` command does not preserve the oldest snapshot incase the 
-keep-* parameters are not satisfied, which led to users not being able to
-preserve old data. Now, restic will always preserve the oldest snapshot 
-whenever any of the keep-* options to the `forget` command are not 
-satisfied.
-
+The `forget` command now additionally preserves the oldest snapshot if fewer
+snapshots are kept than allowed by the `--keep-*` parameters. This maximizes
+amount of history kept while the specified limits are not yet reached.
 
 https://github.com/restic/restic/issues/3624
 https://github.com/restic/restic/pull/4366
--- a/changelog/unreleased/issue-3624
+++ b/changelog/unreleased/issue-3624
@@ -1,0 +1,12 @@
+Enhancement: Keep oldest snapshot when there aren't enough snapshots
+
+The `forget` command does not preserve the oldest snapshot incase the 
+keep-* parameters are not satisfied, which led to users not being able to
+preserve old data. Now, restic will always preserve the oldest snapshot 
+whenever any of the keep-* options to the `forget` command are not 
+satisfied.
+
+
+https://github.com/restic/restic/issues/3624
+https://github.com/restic/restic/pull/4366
+https://forum.restic.net/t/keeping-yearly-snapshots-policy-when-backup-began-during-the-year/4670/2
--- a/internal/restic/snapshot_policy.go
+++ b/internal/restic/snapshot_policy.go
@@ -278,7 +278,7 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots, reason
 
 				if cur.Time.After(t) {
 					val := b.bucker(cur.Time, nr)
-					if val != b.Last {
+					if val != b.Last || nr == len(list)-1 {
 						debug.Log("keep %v, time %v, ID %v, bucker %v, val %v %v\n", b.reason, cur.Time, cur.id.Str(), i, val, b.Last)
 						keepSnap = true
 						bucketsWithin[i].Last = val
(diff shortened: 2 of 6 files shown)


MichaelEischer:
Thanks for the PR, I've added a few comments and rephrased the changelog entry.

There's just one question left: shouldn't the loop that processed `bucketsWithin` also keep a snapshot if `nr == len(list)-1` at https://github.com/restic/restic/pull/4366/files#diff-d3e9e3f713a69c4b81d909a1a1f3de1d5bc1593d6c35c1762111aa36af6aecddR281 ?

gautammenghani:
Thanks for the review.
Is this additional check required for the bucketsWithin struct?  I mean if I specify 
`./restic -r ../testing --verbose forget --keep-yearly 2 `
As a user, I expect atleast 2 snapshots to be there and this PR will handle this case fine as the `buckets` struct will contain a non zero value even after scanning all the snapshots.

But if I specify
`./restic -r ../testing --verbose forget --keep-within-monthly 1m`
Even if I made 100 snapshots in 1 month, this command will preserve only one snapshot and that is kind of expected from a user perspective?

And also, if we specify 
`nr == len(list)-1` in bucketsWithin,
it will always preserve the oldest snapshot, do we want that?

Please do let me know your thoughts on this.

MichaelEischer:
> And also, if we specify
> `nr == len(list)-1` in bucketsWithin,
> it will always preserve the oldest snapshot, do we want that?

That change would only make a difference if the last bucket (e.g. month) contains multiple snapshots. That is for a sufficiently long list of snapshots, e.g. 2023-01-31, 2023-02-28, 2023-03-31, 2023-04-30, 2023-05-31, 2023-06-17 with `--keep-within-monthly 3m`, the policy would just keep the last 4 snapshots as expected.

For 2023-03-28, 2023-03-31, 2023-04-30, 2023-05-31, 2023-06-17  it would keep all snapshots. However, this situation would only arise in two cases: (a) 2023-03-28 is the oldest snapshot and has since been kept by `--keep-within-monthly` or (b) there's a  more finegranular policy, e.g. `--keep-within-daily` with a longer  timespan. The latter configuration is rather nonsensical, thus I'm fine with seeing some corner cases there.

Another argument for keeping the oldest snapshot for bucketsWithin is the following: If a repository contains snapshots for each day in the last 90 days, then I'd expect `--keep-monthly 5` and `--keep-within-monthly 5m` to behave exactly the same.

> I think in that situation, introducing another flag like "--keep-oldest" would make more sense?

IMO there are already too many options for defining snapshot policies.

MichaelEischer:
> But if I specify
> `./restic -r ../testing --verbose forget --keep-within-monthly 1m`
> Even if I made 100 snapshots in 1 month, this command will preserve only one snapshot and that is kind of expected from a user perspective?

That example would currently only keep the latest snapshot, which makes it a very contrived way of writing `--keep-last 1` or even `--keep-within-monthly 1s`. Wouldn't a user rather expect that the `1m` part has at least some effect?

gautammenghani:
> Another argument for keeping the oldest snapshot for bucketsWithin is the following: If a repository contains snapshots for each day in the last 90 days, then I'd expect --keep-monthly 5 and --keep-within-monthly 5m to behave exactly the same.
Yes agreed, this sounds more sane.

> That example would currently only keep the latest snapshot, which makes it a very contrived way of writing --keep-last 1 or even --keep-within-monthly 1s. Wouldn't a user rather expect that the 1m part has at least some effect?

Yeah I guess as a user it would be better to accidentally have the oldest snapshot preserved rather than accidentally having 99 deleted.

gautammenghani:
I've pushed the changes, do let me know if anything else is needed

MichaelEischer (approved):
LGTM. Thanks!

I've rebased the PR to remove the merge commit and merged the latest two commits.

PR #4374, opened 2023-06-17
Try to refresh stale locks

### LINKED ISSUE #4274: Backup fails with Fatal: failed to refresh lock in time is host is suspended in the meantime
Output of `restic version`
--------------------------

`restic 0.15.1 compiled with go1.19.5 on windows/amd64`

Trying to adopt restic as backup solution, built a rather big filelist which should result in a ~6TB backup.
All paths which should be backed up are local on a windows machine.

```
PS C:\Users\Thomas> restic -r X:\restic-repo backup --files-from S:\path\backuplist
repository 3f1dcb70 opened (version 2, compression level auto)
error: open \\?\C:\Users\Thomas\SkyDrive\.849UUID: The process cannot access the file because it is being used by another process.PXL_202XXXX.MP.jpg
Fatal: failed to refresh lock in time
Fatal: unable to save snapshot: context canceled
```

The backup location is a harddisk plugged in via USB3.1.

Expected behavior
-----------------

Complete the backup correctly

Actual behavior
---------------

Storage space is used, but backup failed and no snapshots are created.

Currently reproducible on my machine. It was the third try already.

The first try resulted in an error because of virus or potentially unwanted software which should be backed up.
I think this comes from the windows integrated security and is correct as it contains indeed samples.
```
error: open \\?\E:\FH\ee\6\6-Network Security\Virus\dd153.zip: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
error: open \\?\E:\FH\ee\6\6-Network Security\Virus\subseven-Dateien\bwg503.zip: Operation did not complete successfully because the file contains a virus or potentially unwanted software.
ly because the file contains a virus or potentially unwanted software.
Fatal: failed to refresh lock in time
error: no result
Fatal: unable to save snapshot: context canceled
```

There is also windows power management which suspends after short period of time. I disabled it now and try again.

no

Not today and not yet, but I hope so in the future. Thanks for restic!

MichaelEischer:
#3569 enforced that restic operations are canceled if their lock expires. This was intended to prevent repository corruption if a backup client enters standby, its lock is removed in the meantime after which `prune` is rune. However, this is overly strict for use cases in which there is only a single client or if the repository remained locked.

This PR extends the lock refresh monitoring to also try to refresh stale locks. To do so safely, `RefreshStaleLock` create a new lock but also checks that the old lock continues to exist until after the new lock was created. This ensures that the lifetimes of both locks overlap and thereby prevent conflicting operations from starting.

As the stale lock might already have been removed in the meantime, the lock refresh procedure freezes all new backend operations until the stale lock has been successfully refreshed.

Fixes #4274

Checklist
---------


MichaelEischer:
LGTM.

PR #4379, opened 2023-06-19
Add support for extended attributes on symlinks

### LINKED ISSUE #4375: [Linux] xattrs, like security.selinux, are not backed up for symlinks
Currently, it seems like restic does not store symlink xattrs, at least on Linux. While [Linux does not support `user.*` xattrs on symlinks](https://github.com/torvalds/linux/blob/8c1f0c38b31025f60437215dd27fffc293565fa4/fs/xattr.c#L142-L146), other xattrs, like `security.*` are supported. The biggest use I'm aware of is `security.selinux`, which stores the file's SELinux label.

Currently, if restic is used to back up the root filesystem of a Linux distro that uses SELinux (eg. Fedora), the restored backup is not bootable because all of the symlinks (eg. `/lib` and `/lib64`) are missing SELinux labels.

(The reproduction steps I list below demonstrate the problem using a custom xattr so that it can be run on any Linux distro.)

---

Output of `restic version`
--------------------------

```
restic 0.15.2 (v0.15.2-346-g74ca82a6f) compiled with go1.20.4 on linux/amd64
```

```bash
restic -r <repo> backup <directory with symlink with xattrs>
```

Local path, but occurs with any backend.

Expected behavior
-----------------

xattrs on symlinks would be preserved in the backup.

Actual behavior
---------------

xattrs on symlinks don't seem to be stored.

1. Create a directory tree containing a symlink with xattrs.

    ```bash
    mkdir data
    touch data/file
    ln -s file data/symlink

    sudo setfattr -h -n security.comment -v 'this is a file' data/file
    sudo setfattr -h -n security.comment -v 'this is a symlink' data/symlink
    ```

2. Back up the directory tree with restic.

    ```bash
    restic -r repo init
    restic -r repo backup data
    ```

3. Check the extended attributes via a mount.

    ```bash
    mkdir mnt
    restic -r repo mount mnt
    ```

    ```
    $ getfattr -h -n security.comment mnt/snapshots/latest/data/{file,symlink}
    # file: mnt/snapshots/latest/data/file
    security.comment="this is a file"

    mnt/snapshots/latest/data/symlink: security.comment: Operation not supported
    ```

4. Restore to a directory and check the extended attributes.

    ```bash
    mkdir data_restored
    sudo restic -r repo restore latest -t data_restored
    ```

    ```
    $ getfattr -h -n security.comment data_restored/data/{file,symlink}
    # file: data_restored/data/file
    security.comment="this is a file"

    data_restored/data/symlink: security.comment: No such attribute
    ```

It seems like symlink xattrs aren't stored in the restic repo at all.

Assuming the repository data format supports it, it'd be nice if restic could save and restore the symlink xattrs via `lgetxattr()`/`lsetxattr()`.

Always does! I still find it to be the best tool for the job.

chenxiaolong:

Linux allows the use of non-`user.` extended attributes on symlinks. One of the main users of this functionality is SELinux's `security.selinux` xattr for storing a path's label. By storing symlink xattrs, restic is now suitable for backing up the root filesystem on Linux distributions that use SELinux.

This commit adds support for symlink xattrs when backing up data, restoring data, and mounting snapshots via a fuse mount. All calls to the xattr library have been updated to the use `L` variants of the various functions, which always operate on the path given, without following symlinks.

Fixes: #4375

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4375
+++ b/changelog/unreleased/issue-4375
@@ -1,0 +1,8 @@
+Enhancement: Add support for extended attributes on symlinks
+
+Restic now supports extended attributes on symlinks when backing up,
+restoring, or FUSE-mounting snapshots. This includes, for example, the
+`security.selinux` xattr on Linux distributions that use SELinux.
+
+https://github.com/restic/restic/issues/4375
+https://github.com/restic/restic/pull/4379
--- a/internal/fuse/dir.go
+++ b/internal/fuse/dir.go
@@ -222,19 +222,10 @@ func (d *dir) Lookup(ctx context.Context, name string) (fs.Node, error) {
 }
 
 func (d *dir) Listxattr(_ context.Context, req *fuse.ListxattrRequest, resp *fuse.ListxattrResponse) error {
-	debug.Log("Listxattr(%v, %v)", d.node.Name, req.Size)
-	for _, attr := range d.node.ExtendedAttributes {
-		resp.Append(attr.Name)
-	}
+	nodeToXattrList(d.node, req, resp)
 	return nil
 }
 
 func (d *dir) Getxattr(_ context.Context, req *fuse.GetxattrRequest, resp *fuse.GetxattrResponse) error {
-	debug.Log("Getxattr(%v, %v, %v)", d.node.Name, req.Name, req.Size)
-	attrval := d.node.GetExtendedAttribute(req.Name)
-	if attrval != nil {
-		resp.Xattr = attrval
-		return nil
-	}
-	return fuse.ErrNoXattr
+	return nodeGetXattr(d.node, req, resp)
 }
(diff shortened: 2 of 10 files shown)


greatroar:
LGTM. Looks like we previously backed up the wrong xattrs for symlinks (the ones for the target file).

MichaelEischer (approved):
LGTM. I've added amount of test coverage for extended attributes. The tests in restic/node only use files/dirs+xattrs as we can't write arbitrary xattrs for symlinks. Thus the tests somewhat miss the point, but should be close enough to ensure that this works.

Regarding other operating systems except for Windows, Linux and macOS we'll just have to wait and see whether this change causes problems.

chenxiaolong:
Thank you!

MichaelEischer:
Looks like I've been to quick in merging the PR, #4401 fixes the windows build.

PR #4410, opened 2023-07-17
add --stdin-from-command flag to backup command

### LINKED ISSUE #4251: Backup flag to execute commands with restic, read content from standard input and consider return status

Enrico204:

This pull request is a continuation of #4254.

This fixes [#4251](https://github.com/restic/restic/issues/4251) by providing the requested `--stdin-from-command` flag to the `restic backup` command. It skips snapshot creation in case the given commands fails as determined by `exec.Cmd.Wait()` and returns the stderr of the given command as an error message back to the user.

I have not written any tests, nor documentation because I wanted to get some feedback about the path taken here wrt. upstream inclusion first. I am seeing various test failures in existing tests but these seem to be unrelated and might just occur because I'm building/testing the project within a container. Using these changes locally does look promising though:

```
$ ./restic backup --stdin-from-command --repo ./backup -- pg_dump -h foobar
repository 55e7d5a0 opened (version 2, compression level auto)
using parent snapshot cc6e1dec
error: read /stdin: no data read
Fatal: unable to save snapshot: pg_dump: error: connection to database "root" failed: could not translate host name "foobar" to address: No address associated with hostname
```

Closes [#4251](https://github.com/restic/restic/issues/4251)

Checklist
---------

-   [x]  I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
-   [x]  I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
-   [x]  I have added tests for all code changes.
-   [x]  I have added documentation for relevant changes (in the manual).
-   [x]  There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
-   [x]  I have run `gofmt` on the code in all commits.
-   [x]  All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).

### DIFF
--- a/changelog/unreleased/issue-4251
+++ b/changelog/unreleased/issue-4251
@@ -1,8 +1,12 @@
-Enhancement: Add flag to source the backup from a program's standard output
+Enhancement: Support reading backup from a program's standard output
 
-The `backup` command now supports sourcing the backup content from the standard
-output of an arbitrary command, ensuring that the exit code is zero for a
-successful backup.
+When reading data from stdin, the `backup` command could not verify whether the
+corresponding command completed successfully.
+
+The `backup` command now supports starting an arbitrary command and sourcing
+the backup content from its standard output. This enables restic to verify that
+the command completes with exit code zero. A non-zero exit code causes the
+backup to fail.
 
 Example: `restic backup --stdin-from-command mysqldump [...]`
 
--- a/changelog/unreleased/issue-4251
+++ b/changelog/unreleased/issue-4251
@@ -1,0 +1,10 @@
+Enhancement: Add flag to source the backup from a program's standard output
+
+The `backup` command now supports sourcing the backup content from the standard
+output of an arbitrary command, ensuring that the exit code is zero for a
+successful backup.
+
+Example: `restic backup --stdin-from-command mysqldump [...]`
+
+https://github.com/restic/restic/issues/4251
+https://github.com/restic/restic/pull/4410
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"runtime"
@@ -594,16 +595,37 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		defer localVss.DeleteSnapshots()
 		targetFS = localVss
 	}
-	if opts.Stdin {
+
+	var command *exec.Cmd
+	var stderr io.ReadCloser
+	if opts.Stdin || opts.StdinCommand {
 		if !gopts.JSON {
 			progressPrinter.V("read data from stdin")
 		}
 		filename := path.Join("/", opts.StdinFilename)
+		var closer io.ReadCloser
+		if opts.StdinCommand {
+			command = exec.CommandContext(ctx, args[0], args[1:]...)
+			stdout, err := command.StdoutPipe()
+			if err != nil {
+				return err
+			}
+			stderr, err = command.StderrPipe()
+			if err != nil {
+				return err
+			}
+			if err := command.Start(); err != nil {
+				return err
+			}
+			closer = stdout
+		} else {
+			closer = os.Stdin
+		}
 		targetFS = &fs.Reader{
 			ModTime:    timeStamp,
 			Name:       filename,
 			Mode:       0644,
-			ReadCloser: os.Stdin,
+			ReadCloser: closer,
 		}
 		targets = []string{filename}
 	}
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -97,6 +97,7 @@ type BackupOptions struct {
 	ExcludeLargerThan string
 	Stdin             bool
 	StdinFilename     string
+	StdinCommand      bool
 	Tags              restic.TagLists
 	Host              string
 	FilesFrom         []string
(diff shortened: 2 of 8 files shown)


Enrico204:
I have rebased the code to the latest `master`, and in few days I will proceed to fix it based on changes requested in the old PR :+1:

rawtaz:
Is there other software that has this option name (`--stdin-from-command`)? It's quite long, feels like `--stdin-command` is more concise without losing any semantics.

Enrico204:
> Is there other software that has this option name (`--stdin-from-command`)? It's quite long, feels like `--stdin-command` is more concise without losing any semantics.

Actually, `content-from-command` may be better, that is also the name it has in Borg backup. See https://github.com/restic/restic/pull/4254#discussion_r1160749426

rawtaz:
Nah, there's already two `--stdin-*` options, no point in deviating from that prefix.

Enrico204:
@MichaelEischer I've implemented your suggestions from the old pull request. I implemented a new `io.ReadCloser` instead of a new `fs.FS`, calling `cmd.Wait()` on `io.ReadCloser.Close()`. The archiver is back to the original code.

I'm testing the code now :-)

meise:
Thank you for continuing with this feature. I tested your branch and it still creates snapshots when subcommand exists with non zero:

```
$ restic-stdin-command/restic backup --stdin-from-command -- pg_dump -h foobar  
repository 12ba8696 opened (version 2, compression level auto)
using parent snapshot 75e9f4c7
subprocess pg_dump: pg_dump: error: could not translate host name "foobar" to address: Name or service not known
error: read /stdin: no data read

snapshot d044b2eb saved

$ echo $?
3
```

```
$ pg_dump -h foobar; echo $?
pg_dump: error: could not translate host name "foobar" to address: Name or service not known
1
```

Enrico204:
> Thank you for continuing with this feature. I tested your branch and it still creates snapshots when subcommand exists with non zero:
> 
> ```
> $ restic-stdin-command/restic backup --stdin-from-command -- pg_dump -h foobar  
> repository 12ba8696 opened (version 2, compression level auto)
> using parent snapshot 75e9f4c7
> subprocess pg_dump: pg_dump: error: could not translate host name "foobar" to address: Name or service not known
> error: read /stdin: no data read
> 
> Files:           0 new,     0 changed,     0 unmodified
> Dirs:            0 new,     0 changed,     0 unmodified
> Added to the repository: 0 B   (0 B   stored)
> 
> processed 0 files, 0 B in 0:00
> snapshot d044b2eb saved
> Warning: at least one source file could not be read
> 
> $ echo $?
> 3
> ```
> 
> ```
> $ pg_dump -h foobar; echo $?
> pg_dump: error: could not translate host name "foobar" to address: Name or service not known
> 1
> ```

Uh, I was about to do the same test :-D

Thanks for your report, I'll investigate and fix :+1:

Enrico204:
The problem was that the error returned in file reading or closing was logged, but the snapshot itself was continuing unaffected (because the logic assumed that an error while reading one file should not abort the snapshot).

The `ReadCloserCommand` now returns a "fatal" error, and the error is checked in the `cmd_backup.go` file: if it is "fatal", the snapshot is aborted. Otherwise, the behavior is the same as before: log the error, and exit with code 3 after completing the snapshot.

What do you think?

meise:
> What do you think?

I'm not sure about the `command terminated with no output` fatal error. The current behaviour in stdin mode is to log `no data read` and exit with status code `3` when reading no data.

I like your current implementation more, because having a snapshot with no data makes no sense for me. But it might be behaving differently than we have in existing implementations.

Enrico204:
> > What do you think?
> 
> I'm not sure about the `command terminated with no output` fatal error. The current behaviour in stdin mode is to log `no data read` and exit with status code `3` when reading no data.
> 
> I like your current implementation more, because having a snapshot with no data makes no sense for me. But it might be behaving differently than we have in existing implementations.

You are right, it is a different behavior. I assumed that this is expected, given that we are explicitly asking for monitoring the command.

On second thought, the best way is to check the command exit code also in `ReadCloserCommand.Read()`, and return a fatal error only when there is an EOF and the exit code is non-zero. By doing so, if the command exits with `0` and there is no output, the snapshot is created with a warning (which is the current behavior for files and stdin), while the snapshot is not created if the exit code is non-zero (which is expected).

I'll modify the code asap :-)

Enrico204:
Ok, now the code should stop the snapshot creation by looking only at the exit code. If there is no output, the snapshot is still created with a warning (and exit code 3), which is the same behavior that other snapshot options have (`--stdin`, but also normal file backup). Let me know if there is something we can improve :-)

meise:
Works for me. Very good!

I think there should now be some feedback on the code itself.

Enrico204:
I rebased the branch to the latest `master` :-)

MichaelEischer:
@Enrico204 Thanks for finishing the PR, and sorry for the late feedback. I've somehow missed that the review comments have been addressed in the meantime.

I've reworked the code a bit: the command startup is now handled by `fs.CommandReader` (that's the new name for the helper in `fs`). Instead of using canceling the backup indirectly using a context, `arch.Error` now just returns an error if it is Fatal. This also allows for a nicer error message.

Enrico204:
No problem at all :-)

I will compile this branch and I will test it as soon as possible!

Enrico204:
I am testing the code since a few weeks: so far, no issues :-)

MichaelEischer (approved):
LGTM. Thanks for your work on the PR!

I've rebased the commits one more time before merging it.

PR #4419, opened 2023-07-24
Progress Bar when loading index files

### LINKED ISSUE #229: Inform about downloading index
When using restic on a "slow" connection, it takes a while to download the index form the remote repository, and during this time restic just sits there without saying anything, leaving the user in the dark unless they check and see that there is much active network traffic.

It would be nice if restic at least notified the user that it's working on downloading the index, and even better output a progress/percentage during that time to keep the user updated. Something along the lines of `Downloading index... 42%`.

arjunajesh:

Restic did not provide any feedback while loading index files. Now there is a progress bar for the index loading process.

Closes #229 

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-299
+++ b/changelog/unreleased/pull-299
@@ -1,0 +1,6 @@
+Enhancement: Show progress bar while loading the index
+
+Restic did not provide any feedback while loading index files. Now there is a progress bar for the index loading process. 
+
+https://github.com/restic/restic/issues/229
+https://github.com/restic/restic/pull/4419
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -547,7 +547,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		progressPrinter.V("load index files")
 	}
 
-	bar := newTerminalProgressMax(!gopts.Quiet, 0, "index files loaded", term)
+	bar := newTerminalProgressMax(!gopts.Quiet && !gopts.JSON, 0, "index files loaded", term)
 
 	err = repo.LoadIndex(ctx, bar)
 	if err != nil {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -547,7 +547,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		progressPrinter.V("load index files")
 	}
 
-	bar := newTerminalProgressMax(!gopts.Quiet && !gopts.JSON, 0, "index files loaded", term)
+	bar := newIndexTerminalProgress(gopts.Quiet, gopts.JSON, term)
 
 	err = repo.LoadIndex(ctx, bar)
 	if err != nil {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -546,7 +546,10 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 	if !gopts.JSON {
 		progressPrinter.V("load index files")
 	}
-	err = repo.LoadIndex(ctx)
+
+	bar := newTerminalProgressMax(!gopts.Quiet, 0, "index files loaded", term)
+
+	err = repo.LoadIndex(ctx, bar)
 	if err != nil {
 		return err
 	}
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -169,7 +169,8 @@ func runCat(ctx context.Context, gopts GlobalOptions, args []string) error {
 		return err
 
 	case "blob":
-		err = repo.LoadIndex(ctx)
+		bar := newProgressMax(!gopts.Quiet, 0, "index files loaded")
+		err = repo.LoadIndex(ctx, bar)
 		if err != nil {
 			return err
 		}
@@ -197,7 +198,8 @@ func runCat(ctx context.Context, gopts GlobalOptions, args []string) error {
 			return errors.Fatalf("could not find snapshot: %v\n", err)
 		}
 
-		err = repo.LoadIndex(ctx)
+		bar := newProgressMax(!gopts.Quiet, 0, "index files loaded")
+		err = repo.LoadIndex(ctx, bar)
 		if err != nil {
 			return err
 		}
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -169,7 +169,7 @@ func runCat(ctx context.Context, gopts GlobalOptions, args []string) error {
 		return err
 
 	case "blob":
-		bar := newProgressMax(!gopts.Quiet, 0, "index files loaded")
+		bar := newProgressMax(!gopts.Quiet && !gopts.JSON, 0, "index files loaded")
 		err = repo.LoadIndex(ctx, bar)
 		if err != nil {
 			return err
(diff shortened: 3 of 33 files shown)


changelog/unreleased/pull-299
```diff
@@ -0,0 +1,6 @@
+Enhancemnt: Progress bar when restic is loading index files
```
MichaelEischer, line 1: `Enhancement: Show progress bar while loading the index`

cmd/restic/cmd_backup.go
```diff
@@ -543,7 +543,11 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 	if !gopts.JSON {
 		progressPrinter.V("load index files")
 	}
-	err = repo.LoadIndex(ctx)
+
+	bar := newTerminalProgressMax(!gopts.Quiet, 0, "index files loaded", term)
+	defer bar.Done()
```
MichaelEischer, line 548: `bar.Done()` must be called once `repo.LoadIndex` has completed, not at the end of the `runBackup` function.

cmd/restic/cmd_cat.go
```diff
@@ -143,7 +143,8 @@ func runCat(ctx context.Context, gopts GlobalOptions, args []string) error {
 		return err
 
 	case "blob":
-		err = repo.LoadIndex(ctx)
+		bar := newProgressMax(!gopts.Quiet, 0, "index files loaded")
```
MichaelEischer, line 146: Missing `bar.Done()` call. This also applies to all files below. It would be an option to let `LoadIndex(ctx, bar)` automatically call `bar.Done()`

internal/repository/repository.go
```diff
@@ -580,16 +580,22 @@ func (r *Repository) SetIndex(i restic.MasterIndex) error {
 	return r.prepareCache()
 }
 
-// LoadIndex loads all index files from the backend in parallel and stores them
```
MichaelEischer, line 583: Is there a reason to remove the comment?

internal/repository/repository.go
```diff
@@ -580,16 +580,22 @@ func (r *Repository) SetIndex(i restic.MasterIndex) error {
...
-// in the master index. The first error that occurred is returned.
-func (r *Repository) LoadIndex(ctx context.Context) error {
+func (r *Repository) LoadIndex(ctx context.Context, p *progress.Counter) error {
 	debug.Log("Loading index")
 
+	var numIndexFiles int
```
MichaelEischer, line 586: This should use `uint64` as datatype to avoid the type conversion later on.

internal/repository/repository.go
```diff
@@ -580,16 +580,22 @@ func (r *Repository) SetIndex(i restic.MasterIndex) error {
...
+	r.List(ctx, restic.IndexFile, func(id restic.ID, size int64) error {
```
MichaelEischer, line 587: Please fix the linter errors and warnings

internal/repository/repository.go
```diff
@@ -580,16 +580,22 @@ func (r *Repository) SetIndex(i restic.MasterIndex) error {
...
+		numIndexFiles += 1
+		return nil
+	})
+	p.SetMax(uint64(numIndexFiles))
```
MichaelEischer, line 591: Calling anything on `p` will crash if it is set to `nil`.

internal/repository/repository.go
MichaelEischer, line 587: It would be good to use `be, err := backend.MemorizeList(ctx, repo.Backend(), restic.IndexFile)` to avoid listing the index files twice.

cmd/restic/progress.go
```diff
@@ -80,6 +91,5 @@ func printProgress(status string, canUpdateStatus bool) {
 			carriageControl = "\n"
 		}
 	}
-
```
MichaelEischer, line 83: random whitespace change

cmd/restic/progress.go
```diff
@@ -29,8 +29,8 @@ func calculateProgressInterval(show bool, json bool) time.Duration {
 	return interval
 }
 
-// newProgressMax returns a progress.Counter that prints to stdout.
-func newProgressMax(show bool, max uint64, description string) *progress.Counter {
+// newTerminalProgressMax returns a progress.Counter that prints to stdout or terminal if provided.
+func newTerminalProgressMax(show bool, max uint64, description string, term *termstatus.Terminal) *progress.Counter {
```
MichaelEischer, line 33: I think a different method signature would be much cleaner:

`func newGenericProgressMax(show bool, max uint64, description string, print func(status string)) *progress.Counter` (the call to `printProgress` is then replaced with `print(status)`)

Then `newTerminalProgressMax` would look like the following:
```go
func newTerminalProgressMax(show bool, max uint64, description string, term *termstatus.Terminal) *progress.Counter {
    return newGenericProgressMax(show, max, description, func(status string) {
        term.SetStatus([]string{status})
    })
}
```

A similar conversion also works for `newProgressMax`

MichaelEischer (changes requested):
Please always build the code (`go build ./cmd/restic`), run the tests (`go test ./...`) and check that both succeed before marking the code as ready for review.

The code currently doesn't compile as reported by the CI tests:
```
 Error: cmd/restic/cmd_cat.go:175:24: not enough arguments in call to repo.LoadIndex
	have (context.Context)
	want (context.Context, *progress.Counter)
```

`newTerminalProgressMax` requires a few changes as described below to better decouple the code.

internal/repository/repository.go
arjunajesh: got it. is there anything I should do in the event of an error from calling r.List?

arjunajesh: I am unfamiliar with the function of MemorizeList. How should I use it to track how many index files need to be loaded?

MichaelEischer: If `r.List` returns an error, then just return it immediately.

MichaelEischer: Hmm, using that function is trickier than I've thought. I'll have a look once the other review comments are resolved.

arjunajesh: Thanks for the feedback, I have resolved the other issues. 

BTW, when I was running the tests I noticed that there were some failed test cases. To test if theses were the result of my own changes, I ran the test cases on the main branch and noticed that the same test cases failed, so I'm assuming this expected behavior. Could you explain why this is the case?

MichaelEischer: As the test failures are caused by `Error: cmd/restic/cmd_cat.go:175:24: not enough arguments in call to repo.LoadIndex` they are definitely a result of your changes. For me the `master` branch both compiles and successfully runs the tests, see for example https://github.com/restic/restic/actions/runs/5848699634 .

MichaelEischer: So, that was quite tricky:

This PR was based on a version of the `master` branch that did not yet include #4334, which also modified `cmd_cat.go`. The CI tests run for a commit that is a merge of a PR and the current master branch. Consequently, the new code from #4334 was missing the modifications to its `LoadIndex(...)` call.

Although this behavior may seem strange, it's actually a good idea. The CI tests should not only answer the question whether the tests run successfully for the code in the PR, but rather whether it will work once merged!

The fix for that was to rebase the PR onto the current master branch.

The next part of the test failures were caused by listing the index twice. The integration tests verify that we don't introduce additional `List()` calls to the backend. I've fixed that by using `MemorizeList`.

MichaelEischer (approved):
LGTM. I've rebased the PR (and partially squashed it), disabled the progress bar output if JSON output is enabled, and added the progress bar support to the check command.

Issue #4026, opened 2022-11-14
Introduce a label to uniquely identify 'what' is backed up


damoclark:
Output of `restic version`
--------------------------

restic 0.14.0 compiled with go1.19 on darwin/arm64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I'm proposing to replace the absolute path(s) of a backup snapshot with a user-defined label as the unique identifier for specific backup jobs.  

```bash
restic backup -r repo --label user_data /home /data
# or
cd /
restic backup -r repo --label user_data home data
```

Paths would still be stored as part of the snapshot metadata to capture more context about how the backup was taken.  They just wouldn't be used to uniquely identify that snapshot - that would be the job of the label.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

### TL;DR

1. Reduce confusion for users by changing the conceptual model of Restic backup snapshots
3. Use a label to explicitly name 'what' is being backed up, rather than just implicitly through absolute paths (which can change)
1. Make all backups relative to cwd or path specified on command line like other popular backup tools (reducing confusion between absolute & relative paths)
2. Make Restic more reliably match --parent snapshots
4. Make it easier to support backups of mounted file system snapshots (LVM, ZFS, etc) without fudging paths and/or complicated chroot set ups

### Explanation
While Restic supports relative path(s), it always stores the associated absolute path(s) to uniquely identify the backup snapshot for resolving --parent.  `restic snapshots` shows the absolute path, and on restore, the absolute path too is required for matching and is counter-intuitively restored.  This inconsistent treatment of relative paths is confusing as evidenced by the number of bug reports and questions relating to resolving paths in the forums and on github.  

Another complication is that a backup snapshot can be supplied multiple paths (absolute or relative), and this combination of paths is what uniquely identifies the snapshot for subsequent --parent matches.  If you perform a subsequent backup of some of the files, but not exactly the same paths, then the --parent won't be matched.  e.g.

snapshot 1
restic backup ... /path1 /path2

The above snapshot would not be a parent match for the following:

snapshot 2
restic backup ... /path1

snapshot 3
restic backup ... /path2

snapshot 4
restic backup ... /path2 /path3

This issue is compounded further when absolute paths change, such as cases involving removable media, platform-dependent absolute paths (i.e. Unix VFS & Windows Drive Letters) or backing up filesystem snapshots with VSS, LVM, BTRFS or ZFS.  

[...]

PR #4443, opened 2023-08-09
dump requires absolute path from within the repo

marbx:
Dumping a file following the present documentation resulted in the error message given below.

I found other discussions where dumping failed, but not sure if they are related.

To be discussed:
- [x] Under which conditions the "Directory" (attached) path and the "repository" path differ?
- [x] Under which conditions `dump` requires the "Directory" path?  
- [ ] How to refer to chapters from the documentation? (currently plain hyperlinks)

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -45,6 +45,17 @@ size of the files and directories in ``~/work`` on the local file system. It
 also tells us that only 1.200 GiB was added to the repository. This means that
 some of the data was duplicate and restic was able to efficiently reduce it.
 
+We just attached the absolute path ``~/work`` to the backup, so the path
+within the repository is ``/home/user/work``, depending on your user name.
+
+If we attach a relative path ``work``, the path within the repository is ``/work``.
+
+For example ``restic backup work`` run from ``/home/user`` crates a snapshot
+with a attached path ``/home/user/work`` that contains the path ``/work``
+within the repository. This path-related discrepancy applies to each command
+that tries to access data within a snapshot. You can lookup the paths within
+a repository using the ``ls latest /`` command.
+
 If you don't pass the ``--verbose`` option, restic will print less data. You'll
 still get a nice live status display. Be aware that the live status shows the
 processed files and not the transferred data. Transferred volume might be lower
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -46,13 +46,13 @@ also tells us that only 1.200 GiB was added to the repository. This means that
 some of the data was duplicate and restic was able to efficiently reduce it.
 
 We just attached the absolute path ``~/work`` to the backup, so the path
-within the repository is ``/home/user/work``, depending on your user name.
+within the snapshot is ``/home/user/work``.
 
-If we attach a relative path ``work``, the path within the repository is ``/work``.
+If we attach a relative path ``work``, the path within the snaphot is ``/work``.
 
-For example ``restic backup work`` run from ``/home/user`` crates a snapshot
-with a attached path ``/home/user/work`` that contains the path ``/work``
-within the repository. This path-related discrepancy applies to each command
+For example ``restic backup work`` run from ``/home/user`` creates a snapshot
+with a directory ``/home/user/work`` that contains the path ``/work``
+within the snapshot. This path-related discrepancy applies to each command
 that tries to access data within a snapshot. You can lookup the paths within
 a repository using the ``ls latest /`` command.
 
--- a/doc/050_restore.rst
+++ b/doc/050_restore.rst
@@ -156,12 +156,12 @@ e.g.:
 
     $ restic -r /srv/restic-repo dump --path /production.sql latest production.sql | mysql
 
-This example assumes you attached an absolute path, which means it coincides with the
-path within the repository.
+This example assumes you ran a backup using an absolute path, which coincides with the
+path within the snaphots.
 See https://restic.readthedocs.io/en/stable/040_backup.html#backing-up for the difference
-between attached path and path within the repository.
+between the path used to create the repository and the paths within the snaphots.
 
-If you attached a relative path, the ``dump`` command would look like:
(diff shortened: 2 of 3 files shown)


MichaelEischer:
> * Under which conditions the "Directory" path and the "repository" path differ?

The paths within a snapshot may differ from the attached path list if relative paths were used for the backup and to a certain extent for Windows paths.

For example `restic backup work` run from the folder `/home/user` would result in a snapshot with a path list of `/home/user/work` that will only contain `/work`. This path-related discrepancy applies to each command that tries to access data within a snapshot.

I wonder whether it would make sense to improve the description around how `backup` handles relative paths, instead of describing workarounds in some command section.

marbx:
You clarified the discrepancy between the two paths very well and I used it to improve the backup command.

I recommend to improve not only the backup command, but dump and listing (`restic snapshots`) too:
- Backup, dump and listing are separate use cases, they can be performed at different times or by different end users. Without a reference to the backup commands, an end user failing to use dump and unaware of the discrepancy, would need to read the whole documentation. 
- Including error messages in user documentation allows for searching and googling by error message.

I am not sure what you mean with "workaround". If you refer to the `ls latest /` command, I suggest to document how to look up the path within the repository: the path is necessary  information, making the lookup a necessary procedure.

doc/040_backup.rst
```diff
@@ -45,6 +45,17 @@ size of the files and directories in ``~/work`` on the local file system. It
 also tells us that only 1.200 GiB was added to the repository. This means that
 some of the data was duplicate and restic was able to efficiently reduce it.
 
+We just attached the absolute path ``~/work`` to the backup, so the path
```
MichaelEischer, line 48: I'd prefer to move this to it's own subsection, for example `Paths in a snapshot` and then show a backup using a relative and an absolute path each followed with the corresponding output of `snapshots` and `ls` as illustration.

The jump from `~/work` to `/home/user/work` may also not be obvious to every user, especially those that are unfamiliar with unix shells.

doc/040_backup.rst
```diff
@@ -45,6 +45,17 @@ size of the files and directories in ``~/work`` on the local file system. It
...
+within the repository is ``/home/user/work``, depending on your user name.
```
MichaelEischer, line 49: Paths only exist within snapshots, not the repository.

doc/040_backup.rst
MichaelEischer, line 48: What about a plain `We backed up the absolute path ..., so the snapshot contains the full path:` followed by the output of `ls`?

doc/040_backup.rst
```diff
@@ -45,6 +45,17 @@ size of the files and directories in ``~/work`` on the local file system. It
...
+
+If we attach a relative path ``work``, the path within the repository is ``/work``.
+
+For example ``restic backup work`` run from ``/home/user`` crates a snapshot
+with a attached path ``/home/user/work`` that contains the path ``/work``
```
MichaelEischer, line 54: What about `[...] creates a snapshot with the directory /home/user/work stored in the snapshot metadata` plus showing the output of `snapshots`?

doc/050_restore.rst
```diff
@@ -156,6 +156,31 @@ e.g.:
 
     $ restic -r /srv/restic-repo dump --path /production.sql latest production.sql | mysql
 
+This example assumes you attached an absolute path, which means it coincides with the
```
MichaelEischer, line 159: The path stored in the snapshot metadata (==attached path) is always absolute. I think you mean `you ran a backup using an absolute path`

MichaelEischer (commented):
> the path is necessary information, making the lookup a necessary procedure.

While I agree with the first part, I only partially agree to the second part. This is also why I called `restic ls [...] /` a workaround. If you only have the information from the `snapshots` command, then you're right. However, based on the parameters passed to the `backup` command, it is possible to exactly tell where in the snapshot the files will end up (although the handling of file name collisions is not obvious).

damoclark:
In the longer term, a potential solution to simplify this and many other reported issues is being discussed in PR #4026.  Part of the proposed approach relevant here is that all snapshots will use relative paths, much like existing archiving tools such as tar and zip.

marbx:
> > the path is necessary information, making the lookup a necessary procedure.
> 
> While I agree with the first part, I only partially agree to the second part. This is also why I called `restic ls [...] /` a workaround. If you only have the information from the `snapshots` command, then you're right. However, based on the parameters passed to the `backup` command, it is possible to exactly tell where in the snapshot the files will end up (although the handling of file name collisions is not obvious).

I did not remember the parameter,  the rules are not documented, and therefore I only had the snapshots command. Until snapshots shows both path, it is not a workaround

doc/040_backup.rst
marbx: Please do so

doc/040_backup.rst
marbx: This is another argument why snapshots should list the attached path and the "path within the snapshot". And maybe inner path is a shorter name

doc/040_backup.rst
marbx: I personally think documentation should contain a path like` /home/user/work` as example for an absolute path.

doc/040_backup.rst
marbx: As the column in the snapshots command is named  Directory that would be consistent.

Should  "attached path" be used at all?

doc/050_restore.rst
marbx: This is clearer - I will change that.

marbx: changed

marbx:
> In the longer term, a potential solution to simplify this and many other reported issues is being discussed in PR #4026. Part of the proposed approach relevant here is that all snapshots will use relative paths, much like existing archiving tools such as tar and zip.

Thanks for pointing to PR #4026, I need more time to read it completely. 

In short: I support the idea of snapshots using relative paths!

MichaelEischer:
Replaced by https://github.com/restic/restic/pull/5699

PR #4471, opened 2023-09-18
docs(examples): Use restic binary from system path

jalseth:
The `./restic` notation indicates the user has compiled the program from source, which is unlikely and contradicts the details in the `Prerequisites` section of the doc.


MichaelEischer (approved):
LGTM. Thanks for cleaning up the docs!

PR #4474, opened 2023-09-18
Allow AWS Assume Role

### LINKED ISSUE #4472: Support AWS Assume Role
Output of `restic version`
--------------------------
restic 0.16.0 compiled with go1.20.6 on darwin/amd64

Support Assuming a Role with AWS for S3 storage. Currently AWS IRSA, config, env vars all work, but if you need to use credentials to assume a role first and then use those credentials, it does not work.

I will offer to do the PR, it will require the following:
- using AWS SDK
- adding 3 new configuration items

Based on current implementation either as backend options or environment variables (I think environment variables makes the most sense here, the question is how to prefix them)

Once the standard credential collection happens, if role arn is set, an additional set of calls would be made to change the credentials out using STS.

Using restic in a series of scripts that has basic permissions in AWS, however I need it to backup to different s3 buckets using different roles.

Great tool!

ekristen:

- Adds support for Assuming a Role in AWS

Resolves #4472 

It adds the ability to allow the discovered credentials for AWS to assume another role and use those credentials for the actual run of restic.

Example: Personal IAM user (ekristen) has access to ec2 console but not s3 backup location but has the right to assume the role, backup, backup has full access to the s3 bucket, so this allows restic to assume a role.

Not really, opened #4472

Testing this Pull Request
-------------------------
Using the files from this gist create the folllowing, a test iam user, test iam role, and test s3 bucket. https://gist.github.com/ekristen/e20a976653d0624957add5b4d2ef64ba

1. Create a set of IAM creds for the test user and export them using AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
2. Export AWS_ASSUME_ROLE_ARN with the value of your test iam role arn
3. Run restic against your newly created s3 bucket. 

*Note:** for more advanced use cases you can pass AWS_ASSUME_ROLE_SESSION_NAME and AWS_ASSUME_ROLE_EXTERNAL_ID and use conditions to trust based on values.  Additionally you can pass AWS_ASSUME_ROLE_POLICY which is a JSON document for an IAM policy, this can be used to set a STS policy to restrict the STS credentials to a specific path as an example.

Checklist
---------

Note: tests -- there aren't extensive existing tests on authentication methods, there's not a framework to build off of that I could see to add tests.

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -631,7 +631,9 @@ environment variables. The following lists these environment variables:
     RESTIC_AWS_ASSUME_ROLE_ARN          Amazon IAM Role ARN to assume using discovered credentials
     RESTIC_AWS_ASSUME_ROLE_SESSION_NAME Session Name to use with the role assumption
     RESTIC_AWS_ASSUME_ROLE_EXTERNAL_ID  External ID to use with the role assumption
-    RESTIC_AWS_ASSUME_ROLE_REGION       Region to use for IAM calls for the role assumption
+    RESTIC_AWS_ASSUME_ROLE_POLICY       Inline Amazion IAM session policy
+    RESTIC_AWS_ASSUME_ROLE_REGION       Region to use for IAM calls for the role assumption (default: us-east-1)
+    RESTIC_AWS_ASSUME_ROLE_STS_ENDPOINT URL to the STS endpoint (default is determined based on RESTIC_AWS_ASSUME_ROLE_REGION). You generally do not need to set this, advanced use only.
 
     AZURE_ACCOUNT_NAME                  Account name for Azure
     AZURE_ACCOUNT_KEY                   Account key for Azure
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -628,6 +628,10 @@ environment variables. The following lists these environment variables:
     AWS_DEFAULT_REGION                  Amazon S3 default region
     AWS_PROFILE                         Amazon credentials profile (alternative to specifying key and region)
     AWS_SHARED_CREDENTIALS_FILE         Location of the AWS CLI shared credentials file (default: ~/.aws/credentials)
+    RESTIC_AWS_ASSUME_ROLE_ARN          Amazon IAM Role ARN to assume using discovered credentials
+    RESTIC_AWS_ASSUME_ROLE_SESSION_NAME Session Name to use with the role assumption
+    RESTIC_AWS_ASSUME_ROLE_EXTERNAL_ID  External ID to use with the role assumption
+    RESTIC_AWS_ASSUME_ROLE_REGION       Region to use for IAM calls for the role assumption
 
     AZURE_ACCOUNT_NAME                  Account name for Azure
     AZURE_ACCOUNT_KEY                   Account key for Azure
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -138,7 +138,7 @@ func getCredentials(cfg Config) (*credentials.Credentials, error) {
 		// use the region provided by the configuration by default
 		awsRegion := cfg.Region
 		// allow the region to be overridden if for some reason it is required
-		if len(os.Getenv("RESTIC_AWS_ASSUME_ROLE_REGION")) > 0 {
+		if os.Getenv("RESTIC_AWS_ASSUME_ROLE_REGION") != "" {
 			awsRegion = os.Getenv("RESTIC_AWS_ASSUME_ROLE_REGION")
 		}
 
@@ -148,7 +148,7 @@ func getCredentials(cfg Config) (*credentials.Credentials, error) {
 		stsEndpoint := os.Getenv("RESTIC_AWS_ASSUME_ROLE_STS_ENDPOINT")
 
 		if stsEndpoint == "" {
-			if len(awsRegion) > 0 {
+			if awsRegion != "" {
 				if strings.HasPrefix(awsRegion, "cn-") {
 					stsEndpoint = "https://sts." + awsRegion + ".amazonaws.com.cn"
 				} else {
@@ -167,9 +167,7 @@ func getCredentials(cfg Config) (*credentials.Credentials, error) {
 			RoleSessionName: sessionName,
 			ExternalID:      externalID,
 			Policy:          policy,
-		}
-		if len(awsRegion) > 0 {
-			opts.Location = awsRegion
+			Location:        awsRegion,
 		}
 
 		creds, err = credentials.NewSTSAssumeRole(stsEndpoint, opts)
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -107,13 +107,13 @@ func getCredentials(cfg Config) (*credentials.Credentials, error) {
 	//    call to a pre-defined endpoint, only valid inside
 	//    configured ec2 instances)
 	creds := credentials.NewChainCredentials([]credentials.Provider{
+		&credentials.EnvAWS{},
 		&credentials.Static{
 			Value: credentials.Value{
 				AccessKeyID:     cfg.KeyID,
 				SecretAccessKey: cfg.Secret.Unwrap(),
 			},
 		},
-		&credentials.EnvAWS{},
 		&credentials.EnvMinio{},
 		&credentials.FileAWSCredentials{},
 		&credentials.FileMinioClient{},
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -51,40 +51,9 @@ func open(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, erro
 		minio.MaxRetry = int(cfg.MaxRetries)
 	}
 
-	// Chains all credential types, in the following order:
-	// 	- Static credentials provided by user
-	//	- AWS env vars (i.e. AWS_ACCESS_KEY_ID)
-	//  - Minio env vars (i.e. MINIO_ACCESS_KEY)
-	//  - AWS creds file (i.e. AWS_SHARED_CREDENTIALS_FILE or ~/.aws/credentials)
-	//  - Minio creds file (i.e. MINIO_SHARED_CREDENTIALS_FILE or ~/.mc/config.json)
-	//  - IAM profile based credentials. (performs an HTTP
-	//    call to a pre-defined endpoint, only valid inside
-	//    configured ec2 instances)
-	creds := credentials.NewChainCredentials([]credentials.Provider{
-		&credentials.EnvAWS{},
-		&credentials.Static{
-			Value: credentials.Value{
-				AccessKeyID:     cfg.KeyID,
-				SecretAccessKey: cfg.Secret.Unwrap(),
-			},
-		},
-		&credentials.EnvMinio{},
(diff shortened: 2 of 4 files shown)


ekristen:
This is true. I wonder how often backups exceed 12 hours and use s3 and would use assume role? 

Creating a go func that goes into a loop at expiration/2 and updates the current credentials would work. However back to my original question of how often the exceeding 12 hours would happen.

I see two options

1. Update docs for initial implementation to state that there's a limit of 12 hours with role assumption and cannot use if backups exceed.
2. Work on a for loop to renew credentials at half the expiration time to allow further calls to continue to work.

MichaelEischer (changes requested):
Using the aws-sdk in addition to minio is not an option.

From a quick look at the minio-go documentation, it also supports assuming a role: https://pkg.go.dev/github.com/minio/minio-go/v7@v7.0.61/pkg/credentials#STSAssumeRole . Please use that approach instead. That will likely also just fix the expiry problems.

ekristen:
@MichaelEischer Why is it a no-go, it's a small part of the library? The minio doesn't support all the necessary options like passing an ExternalId. I can write a wrapper to support the expiration. 

Other than the above, is the rest of it acceptable? The environment variable names etc?

MichaelEischer:
> Why is it a no-go, it's a small part of the library?

It increases the binary size by 1MB, seems to work for AWS only, feels completely out of place compared to the existing credentials methods and so on.

>  The minio doesn't support all the necessary options like passing an ExternalId. I can write a wrapper to support the expiration.

I'd prefer to ask whether the minio developers are willing to add support for ExternalId to minio-go . There are already two AWS specific options for STSAssumeRole, That would be a 10-20 lines change on the minio side along with proper integration in the library, compared to the brittle 100+ lines implementation here.

ekristen:
Gotcha. I will make the necessary changes and pull requests.

ekristen:
@MichaelEischer changes as requested, however it depends upon changes to minio-go/v7 which I've opened a PR for here https://github.com/minio/minio-go/pull/1887 -- I'm happy to leave my fork up if you'd like to merge as is, but if you'd like to wait and see if the PR can get merged mainstream, we can leave this open until that PR is merged and upgrade the dependency accordingly.

internal/backend/s3/s3.go
```diff
@@ -87,6 +87,53 @@ func open(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, erro
...
+			}
+		}
+
+		opts := credentials.STSAssumeRoleOptions{
+			RoleARN:   roleArn,
+			AccessKey: c.AccessKeyID,
+			SecretKey: c.SecretAccessKey,
+		}
+		if c.SessionToken != "" {
+			opts.SessionToken = c.SessionToken
```
MichaelEischer, line 116: As these options default to empty strings, there's no need for the surrounding if. This also allows inlining `os.Getenv("AWS_ASSUME_ROLE_SESSION_NAME")`, external Id and policy.

internal/backend/s3/s3.go
```diff
@@ -87,6 +87,53 @@ func open(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, erro
...
+		if sessionName != "" {
+			opts.RoleSessionName = sessionName
+		}
+		if externalID != "" {
+			opts.ExternalID = externalID
+		}
+		if policy != "" {
+			opts.Policy = policy
+		}
+		if len(os.Getenv("AWS_REGION")) > 0 {
```
MichaelEischer, line 127: I think it would be more elegant to save `$AWS_REGION` to a variable once.

internal/backend/s3/s3.go
```diff
@@ -87,6 +87,53 @@ func open(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, erro
 		debug.Log("using anonymous access for %#v", cfg.Endpoint)
 	}
 
+	// assume role
+	roleArn := os.Getenv("AWS_ASSUME_ROLE_ARN")
```
MichaelEischer, line 91: Please extract the code into a helper function. The `open()` function is already far too long.

MichaelEischer (commented):
Thanks for updating the PR. Yes, I'd prefer to wait until the minio PR is merged.

ekristen:
@MichaelEischer the minio PR was just merged https://github.com/minio/minio-go/pull/1887#event-10625033676 -- do we need to wait until it hits a release? If not I can update the branch to use the specific commit.

internal/backend/s3/s3.go
ekristen: Done

internal/backend/s3/s3.go
ekristen: Done

internal/backend/s3/s3.go
ekristen: Done.

ekristen:
Working on fully testing the changes right now, but for anyone else that wants to here's the docker image.
```
ghcr.io/ekristen/restic:0.16.0-ek2
```

ekristen:
Had to make a tweak to the region code to make things align properly with assuming roles properly.
```
ghcr.io/ekristen/restic:0.16.0-ek3
```

MichaelEischer:
> @MichaelEischer the minio PR was just merged [minio/minio-go#1887 (comment)](https://github.com/minio/minio-go/pull/1887#event-10625033676) -- do we need to wait until it hits a release? If not I can update the branch to use the specific commit.

I'd prefer to wait until the next version of the minio-go library is released, the next restic release will still take quite some time, so there's no need to rush it.

I'll try to have a closer look at the code in the next days/weeks.

ekristen:
@MichaelEischer understood.

ekristen:
@MichaelEischer thinking about this, I think we need to at least prefix these with `RESTIC_` to avoid conflicts in the future.

ekristen:
@MichaelEischer minio-go release a new version, this includes that update to make this PR work now.

I've also updated the env vars to be prefixed with `RESTIC_` for the AWS side to prevent any possible collisions with other tooling.

Here's a build docker image for anyone that wants to test.

```
ghcr.io/ekristen/restic:0.16.0-ek5
```

MichaelEischer:
> @MichaelEischer minio-go release a new version, this includes that update to make this PR work now.

Thanks, as a heads-up I'm pretty busy at the moment, I probably won't be able to review the PR before Christmas, sorry.

internal/backend/s3/s3.go
```diff
@@ -107,13 +107,13 @@ func getCredentials(cfg Config) (*credentials.Credentials, error) {
 	//    call to a pre-defined endpoint, only valid inside
 	//    configured ec2 instances)
 	creds := credentials.NewChainCredentials([]credentials.Provider{
+		&credentials.EnvAWS{},
```
MichaelEischer: I've moved the `&credentials.EnvAWS{}` credentials back to the top of the list, as I'm pretty sure that it makes a difference. Only `EnvAWS` handles the `AWS_SESSION_TOKEN` environment variable.

doc/040_backup.rst
```diff
@@ -628,6 +628,12 @@ environment variables. The following lists these environment variables:
     AWS_DEFAULT_REGION                  Amazon S3 default region
     AWS_PROFILE                         Amazon credentials profile (alternative to specifying key and region)
     AWS_SHARED_CREDENTIALS_FILE         Location of the AWS CLI shared credentials file (default: ~/.aws/credentials)
+    RESTIC_AWS_ASSUME_ROLE_ARN          Amazon IAM Role ARN to assume using discovered credentials
+    RESTIC_AWS_ASSUME_ROLE_SESSION_NAME Session Name to use with the role assumption
+    RESTIC_AWS_ASSUME_ROLE_EXTERNAL_ID  External ID to use with the role assumption
+    RESTIC_AWS_ASSUME_ROLE_POLICY       Inline Amazion IAM session policy
```
MichaelEischer, line 634: I've tried to come up with a description for `RESTIC_AWS_ASSUME_ROLE_POLICY` and `RESTIC_AWS_ASSUME_ROLE_STS_ENDPOINT` which were missing. Please take a look.

MichaelEischer (approved):
Sorry for the long delay! I've squashed the commits (rebasing all the changes would have been too messy) and rebased them to the current master branch. That implicitly updates minio to v7.0.66. 

I've added two remarks, please take a look. Other than that I'm happy with the changes.

Could you give the PR one final try to check that everything still works?

internal/backend/s3/s3.go
ekristen: It could cause problems in my opinion.

You are correct that it handles AWS_SESSION_TOKEN, but static config should supersede all.

If you have env vars set and static config set, with your change the env vars will always take precedence which I feel is unexpected behavior.

So if static is first, if someone has passed in static creds it'll work as it does not require a session token.

Otherwise, if a session token is required then environment variables or an AWS profile is the only way to go.

doc/040_backup.rst
ekristen: I think the policy description is acceptable. If someone wants to use it they'll need to understand how STS policies work anyways and look up AWS documentation.

doc/040_backup.rst
```diff
@@ -628,6 +628,12 @@ environment variables. The following lists these environment variables:
...
+    RESTIC_AWS_ASSUME_ROLE_REGION       Region to use for IAM calls for the role assumption (default: us-east-1)
+    RESTIC_AWS_ASSUME_ROLE_STS_ENDPOINT URL to the STS endpoint (default is determined based on RESTIC_AWS_ASSUME_ROLE_REGION)
```
ekristen, line 636:
```suggestion
    RESTIC_AWS_ASSUME_ROLE_STS_ENDPOINT URL to the STS endpoint (default is determined based on RESTIC_AWS_ASSUME_ROLE_REGION) You generally do not need to set this, advanced use only.
```

ekristen:
Hey @MichaelEischer all good, glad you found time to come back around to this! I made a couple comments on your comments.

I'll go build and test this out and reply once done.

internal/backend/s3/s3.go
MichaelEischer: > It could cause problems in my opinion.

Well, before this PR `EnvAWS` was the first entry, thus keeping it at the position won't cause (new) problems.

In addition, the static config variables behave different than you expect: `cfg.KeyID` and `cfg.Secret` are filled from the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. An `AWS_SESSION_TOKEN` is always used together with a key_id and access_key, such that this combination would break if `credentials.Static` is moved before `EnvAWS`.

So let's just keep the order as it was before this PR.

doc/040_backup.rst
MichaelEischer: I've updated the description.

internal/backend/s3/s3.go
ekristen: I see. That should probably not be the case, but I understand not changing it.

Just know that if the environment variables are set, it will always just use EnvAWS, the only time it'll use static is if the `cfg.KeyID` and `cfg.Secret` are explicitly set without environment variables.

ekristen:
Everything seems to be working for me just fine.

MichaelEischer:
Thanks for testing, then let's merge it!

PR #4480, opened 2023-09-26
Support reading basic auth credentials for REST server from environment variables

AgathaSorceress:
This PR adds an `ApplyEnvironment` function to the REST backend, similar to other backends, that reads `REST_USERNAME` and `REST_PASSWORD` variables if the repository URL does not already contain the username or password. This makes it possible to pass secrets to restic without including them in the repository string or repository file. 

I and several other people have run into this problem when configuring restic in NixOS configurations that apply to more than one system. In such cases, the secrets cannot be provided in the repository file, since the repository string is templated at build time to include the target system's hostname (for example: `repository = "rest:http://10.20.1.2:8000/${config.networking.hostName}/";`). The secrets also can't be written in the repository string itself, because the repository string appears in world-readable files in `/nix/store`, and would be publically accessible from the git repository that contains the configuration files.

Checklist
---------

- [ ] I have added tests for all code changes. (note: I looked into adding tests, but none of the other backends test `ApplyEnvironment`, so I'm unsure if I should)

### DIFF
--- a/changelog/unreleased/pull-4480
+++ b/changelog/unreleased/pull-4480
@@ -1,0 +1,10 @@
+Enhancement: Allow setting REST password and username via environment variables
+
+Previously, it was only possible to specify the REST server username and
+password in the repository URL, or using the `--repository-file` option. This
+meant it was not possible to use authentication in contexts where the repository
+URL is public and parts of it are templated by other software. Restic now
+allows setting the username and password using the `RESTIC_REST_USERNAME` and
+`RESTIC_REST_PASSWORD` variables.
+
+https://github.com/restic/restic/pull/4480
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -593,9 +593,16 @@ environment variables. The following lists these environment variables:
     AWS_PROFILE                         Amazon credentials profile (alternative to specifying key and region)
     AWS_SHARED_CREDENTIALS_FILE         Location of the AWS CLI shared credentials file (default: ~/.aws/credentials)
 
-    ST_AUTH                             Auth URL for keystone v1 authentication
-    ST_USER                             Username for keystone v1 authentication
-    ST_KEY                              Password for keystone v1 authentication
+    AZURE_ACCOUNT_NAME                  Account name for Azure
+    AZURE_ACCOUNT_KEY                   Account key for Azure
+    AZURE_ACCOUNT_SAS                   Shared access signatures (SAS) for Azure
+    AZURE_ENDPOINT_SUFFIX               Endpoint suffix for Azure Storage (default: core.windows.net)
+
+    B2_ACCOUNT_ID                       Account ID or applicationKeyId for Backblaze B2
+    B2_ACCOUNT_KEY                      Account Key or applicationKey for Backblaze B2
+
+    GOOGLE_PROJECT_ID                   Project ID for Google Cloud Storage
+    GOOGLE_APPLICATION_CREDENTIALS      Application Credentials for Google Cloud Storage (e.g. $HOME/.config/gs-secret-restic-key.json)
 
     OS_AUTH_URL                         Auth URL for keystone authentication
     OS_REGION_NAME                      Region name for keystone authentication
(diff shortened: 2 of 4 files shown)


internal/backend/rest/config.go
```diff
@@ -70,3 +72,28 @@ func prepareURL(s string) string {
...
 }
+
+var _ restic.ApplyEnvironmenter = &Config{}
+
+// ApplyEnvironment saves values from the environment to the config.
+func (cfg *Config) ApplyEnvironment(prefix string) {
+	username := cfg.URL.User.Username()
+	password, pwd_set := cfg.URL.User.Password()
+
+	if username == "" {
```
MichaelEischer, line 83: I'd prefer to either use both `REST_USERNAME` and `REST_PASSWORD` or none. The environment variables should only be used if neither username nor password is configured so far. That would simplify the code and reduce surprises for users.

changelog/unreleased/pull-4480
```diff
@@ -0,0 +1,9 @@
+Enhancement: Allow setting REST password and username via environment variables
+
+Previously, it was only possible to specify the REST server username and
+password in the repository URL. This meant it was not possible to use
+authentication in contexts where the repository URL is public. Restic now allows
+setting the username and password using the 'REST_USERNAME' and 'REST_PASSWORD'
```
MichaelEischer, line 6: Please use backticks to surround the environment variables.

MichaelEischer (changes requested):
The suggested environment variable look fine, but I'd prefer to either use the user/password from the URL or process both environment variables.

MichaelEischer:
Although, I'm wondering: if you can safely pass `$REST_USERNAME` etc. to restic, then using `$RESTIC_REPOSITORY` (including username and password) should also work? Maybe it would be better to extend the nix package? (Can you provide a link to it?)

internal/backend/rest/config.go
AgathaSorceress: Makes sense, fixed!

AgathaSorceress:
> Although, I'm wondering: if you can safely pass `$REST_USERNAME` etc. to restic, then using `$RESTIC_REPOSITORY` (including username and password) should also work? Maybe it would be better to extend the nix package? (Can you provide a link to it?)
> 
> E.g. `--repository-file <( echo "${VARIABLE}")` could be used to let bash/zsh/other shells generate the URL at runtime and pass it to restic using a temporary file descriptor.

I'm using the [`services.restic.backups`](https://search.nixos.org/options?channel=23.05&from=0&size=50&sort=relevance&type=packages&query=services.restic.backups) module to configure restic, which provides an `environmentFile` option. Using that option, I can have a file like
```env
REST_USERNAME=username
REST_PASSWORD=foobar123
```
which gets rsynced during deployment to the target system, and does not end up in the nix store or the git repo, due to only being referenced by a quoted absolute path.

the NixOS restic module creates a systemd service and timer, and in theory it can be hacked together by disabling the upstream module and making my own, where the service's `ExecStart` is set to run something like `sh -c 'RESTIC_REPOSITORY=rest:http://username:$(cat /var/lib/secrets/restic-password)@10.20.1.2:8000/${config.networking.hostName}/ ${ (optionals (backupPaths != "") [ "${resticCmd} backup ${concatStringsSep " " (backup.extraBackupArgs ++ excludeFlags)} ${backupPaths}" ]) ++ pruneCmd }'` and removing the `Environment="RESTIC_REPOSITORY=..."` line, but that feels like significantly more layers of duct tape than having 2 separate env variables for it, and would also mean that the module needs to be manually updated every time the upstream module changes.

doc/040_backup.rst
```diff
@@ -586,6 +586,9 @@ environment variables. The following lists these environment variables:
 
     TMPDIR                              Location for temporary files
 
+    REST_USERNAME                       REST Server username
+    REST_PASSWORD                       REST Server password
```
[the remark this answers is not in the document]
AgathaSorceress: I looked at existing variables, and it seems that most of them start with the backend name (eg. `AWS_ACCESS_KEY_ID`, `B2_ACCOUNT_ID`, `AZURE_ACCOUNT_NAME`, etc), so I'd say not adding `RESTIC_` would be more consistent, but I can change it if it's likely to collide with existing variables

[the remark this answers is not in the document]
konidev20: If I am not wrong we are talking about 'Restic REST' server right? 

Keeping that in mind `RESTIC_REST_` seems like a better option to me, since we are following the concept of prefixing the backend type. 

I think `REST` is a very generic prefix.

Not important just my opinion.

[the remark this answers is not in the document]
MichaelEischer: I agree that `REST` is extremely generic. The other environment variables match the canonical names used by the corresponding cloud providers. However, rest-server only makes sense in the context of restic, such that a `RESTIC_REST_` is preferable. Please update the environment variable name.

Btw. please move the variables to the end of the list. Even though the current ordering is a complete chaos, I'd prefer to not further deviate from an alphabetical order. (Feel free to reorder the variable names such that their prefixes are sorted alphabetically. But please do not change the order within e.g. the AWS block).

changelog/unreleased/pull-4480
```diff
@@ -0,0 +1,9 @@
+Enhancement: Allow setting REST password and username via environment variables
+
+Previously, it was only possible to specify the REST server username and
+password in the repository URL. This meant it was not possible to use
+authentication in contexts where the repository URL is public. Restic now allows
```
MichaelEischer, line 5: There is already the option to use `--repository-file` to hide the URL. The environment variables just add another way to achieve that goal.

doc/040_backup.rst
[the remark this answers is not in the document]
AgathaSorceress: I ended up leaving the non-backend-specific variables at the top, but the rest should now be sorted alphabetically. I've also added the `RESTIC_` prefix.

changelog/unreleased/pull-4480
AgathaSorceress: Changed it to clarify what I meant, still not sure if the current version is better.

MichaelEischer (approved):
LGTM.

changelog/unreleased/pull-4480
MichaelEischer: The changed text is better, thanks.

Issue #4438, opened 2023-08-05
add support for HTTP connections to Azure Storage Backend


konidev20:
Output of `restic version`
--------------------------
restic 0.16.0

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
Allow HTTP connections to Azure Storage Services when explicitly specified. (The default will be HTTPs which is recommended)

> The Azure storage services support HTTP and HTTPS; however, using HTTPS is highly recommended.
Reference: [Azure Storage REST API Reference | Microsoft Learn](https://learn.microsoft.com/en-us/rest/api/storageservices/)

Azure Storage Services allow HTTP communications when `Secure Transfer Required` field is disabled under the `Configuration` tab of the storage account.

Hence, restic must allow HTTP connections to Azure Storage Services.

What are you trying to do? What problem would this solve?
---------------------------------------------------------
- Add support for HTTP connections when the user explicitly specifies a flag to do so. 
- Support for HTTP is required when users are using custom domains for Azure Storage Services or are using it in private cloud configurations.

MichaelEischer:
> Azure Storage Services allow HTTP communications when `Secure Transfer Required` field is disabled under the `Configuration` tab of the storage account.

Depending on the type of authentication used, that sounds like a bad idea.

To quote https://learn.microsoft.com/en-us/azure/storage/common/storage-require-secure-transfer

> Microsoft recommends that you always require secure transfer for all of your storage accounts.

> * Support for HTTP is required when users are using custom domains for Azure Storage Services

It's currently not possible to specify custom domains. And as restic generally isn't a user-facing service, I don't see the actual use case for that so far.

> or are using it in private cloud configurations.

Shouldn't the Azure provided HTTPS for an Storage Account also work for Service/Private Endpoints?

konidev20:
Hey @MichaelEischer, 

> Depending on the type of authentication used, that sounds like a bad idea.

I'm proposing it as an option that must be explicitly specified.  Take Restic's S3 storage backend for example, it provides similar handling when HTTP URLs are provided; `UseHTTP` is enabled in the config there. If it's a bad idea here, it would be a bad idea there as well, I reckon.

> Because Azure Storage doesn't support HTTPS for custom domain names, this option is not applied when you're using a custom domain name.

Refer: https://learn.microsoft.com/en-us/azure/storage/common/storage-require-secure-transfer

[...]

PR #4501, opened 2023-10-01
Feat support for azure custom domain

konidev20:
Microsoft allows you to map custom domains to Storage Accounts. [Map a custom domain to an Azure Blob Storage endpoint](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name)

This PR adds support for restic to Microsoft Storage Accounts with custom domains. The idea is to add an environment variable `AZURE_CUSTOM_DOMAIN` to allow a user to specify the custom domain. If this environment variable is not set, it will default to `core.windows.net`; if `AZURE_ENDPOINT_SUFFIX` is configured, it will set that.

~~The second step is, that if your custom domain has not been configured with a certificate, we should allow the user to force the usage of HTTP. The user here is expected to explicitly specify that using the option `-o usehttp=true` (this is similar to the S3 implementation, which allows for HTTP connections). The option of specifying `-o usehttp=true` will only be available for custom domains.~~

https://github.com/restic/restic/issues/4438 -> ~~here I had proposed to add support for HTTP connections universally for Microsoft Blob Storage, at that time we concluded that we could add this when we had a good enough reason. I do believe, with custom domains, we have a good reason to enable HTTP connections for it. ~~

Checklist
---------


MichaelEischer:
> I do believe, with custom domains, we have a good reason to enable HTTP connections for it.

My last reply to that suggestion in #4438 was that this is insecure. It doesn't matter whether the HTTP connection is towards a custom domain or the default azure domain name. In particular, as the default azure domain name always support HTTPS, there's no need to support HTTP. Just use the default domain name if you can't setup HTTPS for the custom domain. Just to be clear, I'm strictly opposed to adding `-o usehttp=true`.

I'd also like to know more about the use case, right now this PR and the previous issue just look like "we're adding this because we can". The custom domain example I can find usually just refer to accessing (reading) files via the custom domain. Does uploading files actually work?

konidev20:
> I'd also like to know more about the use case, right now this PR and the previous issue just look like "we're adding this because we can". 

Yes, adding this because we can for now. There are use cases where a system administrator might whitelist only specific domains through which data can be accessed/uploaded; behind a firewall. You might argue that whitelisting `core.windows.net` might do the trick, but we should give the user flexibility to choose how they configure their environments. Since the storage provider allows for such connections, there is definitely a use case out there. I was experimenting with changing backends dynamically by setting up only `backup.contoso.com/repo` and switching Azure regions and even to an S3-compatible storage just by using the custom domain name. 

>  Just use the default domain name if you can't set up HTTPS for the custom domain.

Okay. That restricts the way users will be able to use the Microsoft Blob Storage for backups. As a tool, we must not restrict the way users are able to use the application, because we deem it to be insecure. The user can be in varied environments with different needs. The requirement of the user having to explicitly specify `-o usehttp=true` means the user is aware of the risk of that operation. 

> The custom domain example I can find usually just refers to accessing (reading) files via the custom domain. Does uploading files actually work?

Yes. I have added integration tests and it has passed. I have tested backups and restores on my machine and it worked, with and without HTTP.

konidev20:
[Edit] Let me add one more thing: Just because Azure provides some feature, this does not automatically mean that restic should support it. Adding a feature for which there's no need in restic, just to have a "reason" to add another is just nonsense. [/Edit] 
 
 Okay, I will agree to disagree. Sorry for wasting your time.

rawtaz:
Sorry if this is a stupid question, but using a custom domain to me seems like a separate thing than whether you encrypt traffic or not. Why would you not want to use HTTPS even when you use a custom domain? Even if there's a point to use a custom domain, I don't see why one would want to send their credentials unencrypted.

konidev20:
@MichaelEischer can I remove the HTTP code?

rawtaz:
It was just a curious question, since that part is quite unclear.

konidev20:
> Sorry if this is a stupid question, but using a custom domain to me seems like a separate thing than whether you encrypt traffic or not. Why would you not want to use HTTPS even when you use a custom domain? Even if there's a point to use a custom domain, I don't see why one would want to send their credentials unencrypted.

If I'm in a virtual network, with a custom DNS setup. The threat model for me is that the communication is within the network.

If I use `core.windows.net` connections can go over the internet in the case of Azure.

rawtaz:
> If I'm in a virtual network, with a custom DNS setup. The threat model for me is that the communication is within the network.

If it's internal systems, it ought to be even simpler for you to secure the communication with HTTPS.

konidev20:
> > If I'm in a virtual network, with a custom DNS setup. The threat model for me is that the communication is within the network.
> 
> If it's internal systems, it ought to be even simpler for you to secure the communication with HTTPS.

Okay.

PR #4511, opened 2023-10-11
Include inode in JSON output for find and ls

dnnr:

This introduces the inode attribute to the JSON output emitted for nodes in `ls` and matches in `find`. There doesn't seem to be any discernible reason to omit the inode and it can be useful in scripting scenarios.

There was a brief discussion in the IRC channel. 

Checklist
---------


dnnr:
I'm not entirely sure if I should add any tests for this. Many of the existing attributes currently don't seem to be tested either. `cmd_ls_test.go` appears to cover some special cases but there's no basic case that just confirms the full list of possible fields. Should I add something like that, or would you consider it rather pointless/superfluous as it just reiterates the implementation?

MichaelEischer (approved):
LGTM.

Ideally, we'd extract the `restic.Node` to JSON conversion into a separate method (using the approach from `cmd_ls.go`) and add a very basic test just to make sure that the JSON output doesn't change unexpectedly. But that's out of scope for this PR.

PR #4514, opened 2023-10-13
Make `key list` command honor `--no-lock`

### LINKED ISSUE #4513: `restic key --no-lock list` acquires a lock
Output of `restic version`
--------------------------

restic 0.16.0 compiled with go1.20.8 on freebsd/amd64

```
restic key list --no-lock
Save(<lock/807df3d3ba>) returned error, retrying after 305.705679ms: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 394.709517ms: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 1.424932044s: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 1.789063058s: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 3.022910059s: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 5.295456214s: OpenFile: permission denied
Save(<lock/807df3d3ba>) returned error, retrying after 3.19014686s: OpenFile: permission denied
```

sftp

Expected behavior
-----------------

We use read-only users on the SFTP storage to allow certain recurring administrative tasks (like `snapshots`, `check`, etc.). The idea is that there is no risk to interfere with actual backups and also the ability to do this on file system snapshots (which are read-only by nature). In the specific case, we want to check if all necessary systems and users have access to the backup to ensure it's configured correctly (this is on top of doing actual restore checks, which happen separately and less frequently).

That said, the expected behavior would be to get a list of keys that have access to the repo.

Actual behavior
---------------

restic tries to acquire a lock and fails.

```
restic -r myrepo init
restic -r myrepo key list
find myrepo -type d -exec chmod 500 {} \;
restic -r myrepo key list
restic -r myrepo key --no-lock list
```

Either a bug or a design decision.

Make `key list` honour the `--no-lock` flag. This is the first time I would need to take a look a restic's source code.

Restic makes me smile every time I use it.

grembo:
Fixes #4513

Don't lock repo if `--no-lock` flag was used as in `restic key list --no-list`

#4513

Checklist
---------

- [ ] ~~I have added tests for all code changes.~~

It seems like there are no unit tests for --no-lock besides TestCheckRestoreNoLock anywhere, so I think this should be part of a bigger effort to add such tests.

- [ ] ~~I have added documentation for relevant changes (in the manual).~~

### DIFF
--- a/changelog/unreleased/issue-4513
+++ b/changelog/unreleased/issue-4513
@@ -1,0 +1,8 @@
+Bugfix: Make `key list` command honor `--no-lock`
+
+This allows to determine which keys a repo can be accessed by without the
+need for having write access (e.g., read-only sftp access, filesystem
+snapshot).
+
+https://github.com/restic/restic/issues/4513
+https://github.com/restic/restic/pull/4514
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -212,10 +212,13 @@ func runKey(ctx context.Context, gopts GlobalOptions, args []string) error {
 
 	switch args[0] {
 	case "list":
-		lock, ctx, err := lockRepo(ctx, repo, gopts.RetryLock, gopts.JSON)
-		defer unlockRepo(lock)
-		if err != nil {
-			return err
+		if !gopts.NoLock {
+			var lock *restic.Lock
+			lock, ctx, err = lockRepo(ctx, repo, gopts.RetryLock, gopts.JSON)
+			defer unlockRepo(lock)
+			if err != nil {
+				return err
+			}
 		}
 
 		return listKeys(ctx, repo, gopts)


changelog/unreleased/issue-4513
```diff
@@ -0,0 +1,8 @@
+Bugfix: Make `key list` command honor `--no-lock`
+
+This allows to determine which keys a repo can be accessed by without the
+need for having write access (e.g., read-only sftp access, filesystem
+snapshot).
+
+https://github.com/restic/restic/issues/4514
```
MichaelEischer, line 7: 4514 is the pull request, not the issue

cmd/restic/cmd_key.go
```diff
@@ -212,6 +212,10 @@ func runKey(ctx context.Context, gopts GlobalOptions, args []string) error {
 
 	switch args[0] {
 	case "list":
+		if gopts.NoLock {
+			return listKeys(ctx, repo, gopts)
```
MichaelEischer, line 216: I'd prefer to just wrap the `lockRepo` and `unlockRepo` call into an if block, in order to not duplicate the `listKeys` call.

MichaelEischer (changes requested):

MichaelEischer:
> It seems like there are no unit tests for --no-lock besides TestCheckRestoreNoLock anywhere, so I think this should be part of a bigger effort to add such tests.

Yeah, the CLI methods have fairly little test coverage. I think we can skip adding a test, just make sure to manually test that `key list` works with and without `--no-lock`.

changelog/unreleased/issue-4513
grembo: Duh! :) Thanks

cmd/restic/cmd_key.go
grembo: ok

MichaelEischer (approved):
LGTM.

I've rebased the PR and squashed the commits.

PR #4524, opened 2023-10-19
Don't load password when autocompleting CLI commands

### LINKED ISSUE #4516: Restic autocompletion does not work in Fish shell when adding RESTIC_PASSWORD_FILE in /etc/environment
Output of `restic version`
--------------------------
restic 0.16.0 compiled with go1.20.7 on linux/amd64

Expected behavior
-----------------
1. Install `fish` shell (the current version: 3.6.1)
2. Run `fish` in any terminal 
3. Type `$ sudo restic <Press TAB>`  in any terminal to show auto completion, that works fine as expected. 

Steps to reproduce the behavior (Issue)
---------------

1.  `/etc/environments` adds  `RESTIC_PASSWORD_FILE=/no_password_file` 
3.  Reboot
4.  Try to type  `restic <TAB>`  in fish shell, auto completion does not work.
The issue is caused by adding `RESTIC_PASSWORD_FILE` in `/etc/environments` if there is no password file or this password file is not readable for normal users except admin/root.
But `bash` and `zsh`  have no issue.  

`/etc/environments` adding  `RESTIC_PASSWORD_FILE=/no_password_file`  breaks auto-completion of restic in fish shell if the password file is not readable for users. 

But `bash` shell and `zsh`shell have no issue with that.

Remove `RESTIC_PASSWORD_FILE=/...` in `/etc/environments` , then reboot. The issue is fixed. 
But that is not nice.

MichaelEischer:
Don't load the repository password when autocompleting CLI commands. This can break autocompletion if for example `RESTIC_PASSWORD_FILE` is set to an invalid path.

Fixes #4516

Checklist
---------

- ~~[ ] I have added tests for all code changes.~~ Manually tested.

### DIFF
--- a/changelog/unreleased/issue-4516
+++ b/changelog/unreleased/issue-4516
@@ -1,0 +1,8 @@
+Bugfix: Do not try to load password on command line autocomplete
+
+The command line autocompletion previously tried to load the repository
+password. This could cause the autocompletion not to work. Now, this step gets
+skipped.
+
+https://github.com/restic/restic/issues/4516
+https://github.com/restic/restic/pull/4526
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -80,7 +80,7 @@ The full documentation can be found at https://restic.readthedocs.io/ .
 // user for authentication).
 func needsPassword(cmd string) bool {
 	switch cmd {
-	case "cache", "generate", "help", "options", "self-update", "version":
+	case "cache", "generate", "help", "options", "self-update", "version", "__complete":
 		return false
 	default:
 		return true


changelog/unreleased/issue-4516
```diff
@@ -0,0 +1,7 @@
+Bugfix: Do not try to check password on command line autocomplete
+
+The command line autocompletion previously tried to load the repository
+password. Now, this step gets skipped.
```
rawtaz, line 4: Suggest `The command line autocompletion previously tried to load the repository password. This could cause the autocompletion not to work. Now, this step is skipped.`

changelog/unreleased/issue-4516
```diff
@@ -0,0 +1,7 @@
+Bugfix: Do not try to check password on command line autocomplete
```
rawtaz, line 1: s/check/load/ ?

rawtaz (changes requested):

changelog/unreleased/issue-4516
MichaelEischer: Thanks. I've applied the suggestions.

rawtaz (approved):

Issue #805, opened 2017-02-13
Add option to check for bit rot during backup creation


oysols:
There is currently no way of knowing if the original files deteriorate.

`restic backup` skips files based on mod time. Bit rot will not be detected at all, but the backup will contain the original file. If the owner modifies the file without discovering that the file is degraded, or `restic backup --force` is run, restic will happily backup the new file. At the time when the owner discovers that the file is degraded, it will be very hard to find and recover the correct bits. Worst case scenario, the file is lost due to backup pruning.

`restic backup --force` hashes all files, regardless of mod time. Bit rot will result in a new hash, and restic will interpret this as a normal modification and backup the file. The owner will be none the wiser, and will similarly have problems when the bit rot is detected.

I suggest implementing a third option, similar to `restic backup --force` that checks mod time to determine if the file has been changed intentionally. If the file is modified without a modified mod time it would log or err out.

(This could also be extended to verify other meta data, such as checking that the modified date is not earlier than the one recorded in the previous backup.)

This might be out of the scope of a backup program, but if we are already hashing the files, and recording the previous hashes, most of the work is already done.

zcalusic:
You're completely right about that @oysols. And I don't think it is out of scope of restic, to check full integrity of the data it saved.

I only think this should be implemented via check command, check --full or similar. But, a good idea, definitely. 👍

fd0:
Great idea, implementation will need to be discussed. Thanks for reporting this!

applestump:
Great idea! +1

ehzastrow:
+1

bjoe2k4:
+1

awehrfritz:
+1

midnightmagic:
Intermediate functionality which would support this could be a dump of file hashes which users can just feed to sha256sum -c; or even just a textual retrieval of hash chunks so users like myself can write our own utilities to verify against the chunks. Use case is verifying file system files after a crash or a detected drive failure and data restoration on backups which are larger than half the size of the underlying storage and/or over a network where network hashing would be unnecessarily slow.

Really really love restic. With even just a pathway to a sum-based way to verify client drive contents it would essentially be feature-complete for someone like myself.

Much obliged for all the work you all have done. Wonderful software.

midnightmagic:
[...]

PR #4526, opened 2023-10-20
Add bitrot detection to "diff" command

dnnr:

This introduces a new modifier to the output of the diff command. It appears whenever two files being compared only differ in their content but not in their metadata. As far as we know, under normal circumstances, this should only ever happen if some kind of bitrot has happened in the source file. The prerequisite for this detection to work is that the right-side snapshot of the comparison has been created with "backup --force".

See #805 

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4526
+++ b/changelog/unreleased/pull-4526
@@ -1,0 +1,11 @@
+Enhancement: Add bitrot detection to `diff` command
+
+The output of the `diff` command now includes the modifier `?` for files
+to indicate bitrot in backed up files. It will appear whenever there is a
+difference in content while the metadata is exactly the same. Since files with
+unchanged metadata are normally not read again when creating a backup, the
+detection is only effective if the right-hand side of the diff has been created
+with "backup --force".
+
+https://github.com/restic/restic/issues/805
+https://github.com/restic/restic/pull/4526
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -274,11 +274,13 @@ func (c *Comparer) diffTree(ctx context.Context, stats *DiffStatsContainer, pref
 				mod += "M"
 				stats.ChangedFiles++
 
-				node1NilContent := node1
-				node2NilContent := node2
+				node1NilContent := *node1
+				node2NilContent := *node2
 				node1NilContent.Content = nil
 				node2NilContent.Content = nil
-				if node1NilContent.Equals(*node2NilContent) {
+				// the bitrot detection may not work if `backup --ignore-inode` or `--ignore-ctime` were used
+				if node1NilContent.Equals(node2NilContent) {
+					// probable bitrot detected
 					mod += "?"
 				}
 			} else if c.opts.ShowMetadata && !node1.Equals(*node2) {
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -29,6 +29,9 @@ directory:
 * T  The type was changed, e.g. a file was made a symlink
 * ?  Bitrot detected: The file's content has changed but all metadata is the same
 
+Metadata comparison will likely not work if a backup was created using the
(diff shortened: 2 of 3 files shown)


dnnr:
Sadly, I couldn't figure out a good way to test this... I did however verify it manually.

I also feel like some more verbose documentation could be useful (along the lines of the commit message), but I wasn't sure where to put it. Any suggestions? The section in `040_backup.rst` is pretty compact and most likely isn't the appropriate place for that kind of detail.

changelog/unreleased/pull-4526
```diff
@@ -0,0 +1,20 @@
+Enhancement: Add bitrot detection to diff command
+
+# Describe the problem in the past tense, the new behavior in the present
```
MichaelEischer, line 3: Please remove all lines prefixed with `#`.

changelog/unreleased/pull-4526
```diff
@@ -0,0 +1,20 @@
...
+unchanged metadata are normally not read again when creating a backup, the
+detection is only effective if the right-hand side of the diff has been created
+with "backup --force".
+
+# The last section is a list of issue, PR and forum URLs.
+# The first issue ID determines the filename for the changelog entry:
+# changelog/unreleased/issue-1234. If there are no relevant issue links,
+# use the PR ID and call the file pull-55555.
+
+https://github.com/restic/restic/issues/805
```
MichaelEischer, line 20: Please add a link to this PR.

cmd/restic/cmd_diff.go
```diff
@@ -273,6 +274,14 @@ func (c *Comparer) diffTree(ctx context.Context, stats *DiffStatsContainer, pref
 				!reflect.DeepEqual(node1.Content, node2.Content) {
 				mod += "M"
 				stats.ChangedFiles++
+
+				node1NilContent := node1
```
MichaelEischer, line 278: `node1` is a pointer, use `node1NilContent := *node1` to actually copy it.

cmd/restic/cmd_diff.go
```diff
@@ -273,6 +274,14 @@ func (c *Comparer) diffTree(ctx context.Context, stats *DiffStatsContainer, pref
...
+				node2NilContent := node2
+				node1NilContent.Content = nil
+				node2NilContent.Content = nil
+				if node1NilContent.Equals(*node2NilContent) {
```
MichaelEischer, line 282: This won't work for snapshots in which files have unstable inodes. As the rest of the `diff` command also has that problem, I'm ok with it. But we should add a bit of documentation, that the detection may not work if `backup --ignore-inode` or `--ignore-ctime` were used.

MichaelEischer:
> The section in `040_backup.rst` is pretty compact and most likely isn't the appropriate place for that kind of detail.

The explanation of the `diff` command feels completely out of place in `040_backup.rst`. Let's move it to `045_working_with_repos.rst`, which would also provide the room necessary for additional explanations.

> Sadly, I couldn't figure out a good way to test this... I did however verify it manually.

What might work is the following: backup now supports the `--stdin-from-command` option which allows creating a snapshot with a certain content and choosing the file timestamp using the `--time` parameter, see `TestStdinFromCommand` for how to use it in tests. That combination allows us to reliably choose the metadata and content of the file. In particular, it also allows changing the file content without changing any metadata.

The other alternative would be to enhance the code at `internal/restic/restic.go` to allow creating a snapshot with bitrot. But that's likely the more complex path.

MichaelEischer (changes requested):

cmd/restic/cmd_diff.go
dnnr: Oh. The only reason this didn't break anything is because there's no access to `.Content` later on, right?

MichaelEischer: Exactly.

MichaelEischer (approved):
LGTM.

PR #4528, opened 2023-10-21
Run tests with Go 1.21

adamantike:

Run tests with Go 1.21.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 env:
-  latest_go: "1.20.x"
+  latest_go: "1.21.x"
   GO111MODULE: on
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -22,6 +22,28 @@ jobs:
...
+          - job_name: Windows
+            go: 1.21.x
+            os: windows-latest
+
+          - job_name: macOS
+            go: 1.21.x
+            os: macOS-latest
+            test_fuse: false
+
+          - job_name: Linux
```
MichaelEischer, line 34: We only keep the full set of tests for the latest go version, as that version will be used for the release. Keeping the full set of tests for older go versions increases the CI runtime too much.

adamantike: So, should I just maintain `Linux` for 1.20.x (which also runs for 1.18 and 1.19), to end up with the following matrix...?
* `1.21.x` for Windows, MacOS, Linux, and Linux (race)
*  `1.20.x` for Linux
*  `1.19.x` for Linux
*  `1.18.x` for Linux

MichaelEischer: Yes. Just to be clear, the `test_cloud_backends:` should only run for Linux + Go 1.21.x

MichaelEischer (approved):
LGTM. Thanks!

PR #4542, opened 2023-10-26
Only support ARMv6 on ARM platforms

### LINKED ISSUE #4540: Illegal instruction on Raspberry Pi with arm version
Output of `restic version`
--------------------------

restic 0.16.1

Calling restic 0.16.1 for arm on a Raspberry Pi results in an `Illegal instruction`.

```
./restic_0.16.1_linux_arm version
Illegal instruction
```

More details:

```
cat /proc/cpuinfo 
processor       : 0
model name      : ARMv6-compatible processor rev 7 (v6l)
BogoMIPS        : 697.95
Features        : half thumb fastmult vfp edsp java tls 
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xb76
CPU revision    : 7
```

```
uname -a
Linu xxx 6.1.21+ #1642 Mon Apr  3 17:19:14 BST 2023 armv6l GNU/Linux
```

```
cat /etc/debian_version 
11.7
```

Version 0.16.0 runs on the same Raspberry Pi without issues:

```
./restic_0.16.0_linux_arm version
restic 0.16.0 compiled with go1.20.6 on linux/arm
```

MichaelEischer:
Go 1.21 has switched the default from GOARM=5 to GOARM=7. So far there have been complaints from Raspberry Pi 1 users, as the first raspberry pi version only supports ARMv6. Exclude older ARM versions as these are likely not relevant (rest-server also only supports ARMv6/7) and enforce the use of software floating point emulation.

This also fixes the mislabeling of our docker containers as linux/arm/v7 when the contained binary actually only required linux/arm/v5

Fixes #4540

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4540
+++ b/changelog/unreleased/issue-4540
@@ -1,0 +1,6 @@
+Change: Require at least ARMv6 for ARM binaries
+
+The official release binaries of restic now require at least ARMv6 support for ARM platforms.
+
+https://github.com/restic/restic/issues/4540
+https://github.com/restic/restic/pull/4542
--- a/helpers/build-release-binaries/main.go
+++ b/helpers/build-release-binaries/main.go
@@ -125,6 +125,10 @@ func build(sourceDir, outputDir, goos, goarch string) (filename string) {
 		"GOOS="+goos,
 		"GOARCH="+goarch,
 	)
+	if goarch == "arm" {
+		// the raspberry pi 1 only supports the ARMv6 instruction set
+		c.Env = append(c.Env, "GOARM=6")
+	}
 	verbose("run %v %v in %v", "go", c.Args, c.Dir)
 
 	err := c.Run()


JsBergbau:
> So far there have been complaints from Raspberry Pi 1 users, as the first raspberry pi version only supports ARMv6.

I think the Raspberry Pi 1 is not the main cause. Raspberry Pi 1 is probably not used widely anymore. However the Raspberry PI Zero W is also based on ARMv6. I am using 3 Raspberry PI Zero W and therefore need the ARMv6 support. 
I think as far as Go supports building for ARMv6 we should keep it for the Raspberry PI Zero W users.

MichaelEischer:
> However the popular Raspberry PI Zero W is also based on ARMv6.

That information is missing from #4540. There only the "Raspberry Pi" was mentioned. I only intend to deprecate ARMv5, not later versions of the instruction set.

MichaelEischer (commented):
LGTM.

PR #4584, opened 2023-12-05
S3: Don't archive metadata files on S3 Glacier

### LINKED ISSUE #4583: Backup to S3 Glacier archive triers make backups unusable/broken without cache
Output of `restic version`
--------------------------
`restic 0.16.2 compiled with go1.21.3 on linux/arm64`

AWS S3

If you upload backups to S3 with `-o s3.storage-class=GLACIER` or  `-o s3.storage-class=DEEP_ARCHIVE` it makes your repository partially broken.
If you try to exec `restic snapshots` from another host, or using `--no-cache` flag the output for this  would be:
```
 $ restic snapshots                                                                                               
repository fffffffff opened (version 1)
Load(<snapshot/aaaaaaaaa>, 0, 0) returned error, retrying after 508.626707ms: The operation is not valid for the object's storage class
Load(<snapshot/bbbbbbbb>, 0, 0) returned error, retrying after 308.071881ms: The operation is not valid for the object's storage class
Load(<snapshot/aaaaaaaaa>, 0, 0) returned error, retrying after 1.014433406s: The operation is not valid for the object's storage class
Load(<snapshot/bbbbbbbbb>, 0, 0) returned error, retrying after 400.917258ms: The operation is not valid for the object's storage class
Load(<snapshot/bbbbbbbbb>, 0, 0) returned error, retrying after 812.242443ms: The operation is not valid for the object's storage class
Ignoring "aaaaaaaaaaaa...": failed to load snapshot aaaaaaaaaa: context canceled
Ignoring "bbbbbbbbbbb...": failed to load snapshot bbbbbbbbb: context canceled
------------------------------------------------------------------------------------
cccccccc  2023-12-04 00:15:45  my-host               /home/LICENSE
ddddddd  2023-12-04 13:51:33  my-host               /mnt/media/photo
eeeeeeee  2023-12-04 23:21:57 my-host               /home/README.md
------------------------------------------------------------------------------------
3 snapshots
```

To get this output I executed `restic snapshots`, and soon after I got few errors - clicked `CTRL+C`.
Here you can see, in the repository with 5 (A, B, C, D, E) uploaded snapshots, 2 of them (A, B) are uploaded to the Glacier Flexible and Glacier Deep Archive tiers. 

This issue not persists to the `GLACIER_IR` storage class, because all files are 'instant retrieval', so no issue here (the C-snapshot).

It is expected that if data is archived - it needed to be 'retrieved' beforehand in the AWS. And when I did so (on the machine with cache) - I was able to `restic restore` it.

Expected behavior
-----------------
When non-instant retrieval storage class is opted-in - metadata is stored in `STANDARD` storage class, while packfile data is saved to specified tier.

Actual behavior
---------------
Snapshot metadata is unreachable because it is also stored in specified storage class.

yes. It is because all files sent to the same tier without filtering metadata.
I've made a fallback fix for this issue, which I tested and it's working as intended: Ignore storage class option for metadata if sotrage class is either `GLACIER` or `DEEP_ARCHIVE`

Yup, I became quite enthusiastic about opportunity to safely backup significant media in cold storage for affordable money, so I tried this myself, and only after found out that I can't access them from another machine. And I even promoted this feature to friend, before trying it myself. So that's why I'm about to create the PR.

elkemper:
It fixes issue when you upload snapshot to S3 Glacier, and metadata isn't available from any machine without cache.

It sets a fallback for the case where storage class should not be set - for metadata AND ONLY when specified tier will move files to archive. For the code simplicity it is done through the inverse logic operators: set storage class when file is `data/` OR when  storage class is not instant accessible (such as Standard etc. including 'Glacier Instant Retrieval').

Closes #4583

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4583
+++ b/changelog/unreleased/issue-4583
@@ -1,0 +1,9 @@
+Bugfix: Ignoring the s3.storage-class option for metadata when archive tier is specified
+
+Restic now will save snapshot metadata to non-archive storage tier whatsoever,
+this will help avoid issues when data is being saved to archive storage class.
+It is not providing any support for cold storages in restic, 
+only saving users from making backups unusable.
+
+https://github.com/restic/restic/issues/4583
+https://github.com/restic/restic/issues/3202
--- a/changelog/unreleased/issue-4583
+++ b/changelog/unreleased/issue-4583
@@ -1,9 +1,12 @@
-Bugfix: Ignoring the s3.storage-class option for metadata when archive tier is specified
+Enhancement: Ignore s3.storage-class for metadata if archive tier is specified
 
-Restic now will save snapshot metadata to non-archive storage tier whatsoever,
-this will help avoid issues when data is being saved to archive storage class.
-It is not providing any support for cold storages in restic, 
-only saving users from making backups unusable.
+There is no official cold storage support in restic, use this option at your
+own risk.
+
+Restic always stored all files on s3 using the specified `s3.storage-class`.
+Now, restic will store metadata using a non-archive storage tier to avoid
+problems when accessing a repository. To restore any data, it is still
+necessary to manually warm up the required data beforehand.
 
 https://github.com/restic/restic/issues/4583
-https://github.com/restic/restic/issues/3202
+https://github.com/restic/restic/pull/4584
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -325,12 +325,22 @@ func (be *Backend) Path() string {
 	return be.cfg.Prefix
 }
 
+// useStorageClass returns whether file should be saved in the provided Storage Class
+func (be *Backend) useStorageClass(h backend.Handle) bool {
+	var notArchiveClass bool = be.cfg.StorageClass != "GLACIER" && be.cfg.StorageClass != "DEEP_ARCHIVE"
+	isDataFile := h.Type == backend.PackFile && !h.IsMetadata
+	return isDataFile || notArchiveClass
+}
+
 // Save stores data in the backend at the handle.
 func (be *Backend) Save(ctx context.Context, h backend.Handle, rd backend.RewindReader) error {
 	objName := be.Filename(h)
 
-	opts := minio.PutObjectOptions{StorageClass: be.cfg.StorageClass}
-	opts.ContentType = "application/octet-stream"
+	opts := minio.PutObjectOptions{ContentType: "application/octet-stream"}
+
+	if be.useStorageClass(h) {
+		opts.StorageClass = be.cfg.StorageClass
+	}
 	// the only option with the high-level api is to let the library handle the checksum computation
 	opts.SendContentMd5 = true
 	// only use multipart uploads for very large files
--- a/internal/backend/s3/s3.go
+++ b/internal/backend/s3/s3.go
@@ -326,8 +326,10 @@ func (be *Backend) Path() string {
 }
 
 // useStorageClass returns whether file should be saved in the provided Storage Class
+// For archive storage classes, only data files are stored using that class; metadata
+// must remain instantly accessible.
 func (be *Backend) useStorageClass(h backend.Handle) bool {
-	var notArchiveClass bool = be.cfg.StorageClass != "GLACIER" && be.cfg.StorageClass != "DEEP_ARCHIVE"
+	notArchiveClass := be.cfg.StorageClass != "GLACIER" && be.cfg.StorageClass != "DEEP_ARCHIVE"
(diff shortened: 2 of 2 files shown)


MichaelEischer (changes requested):
I don't understand how this PR is supposed to work. The directory metadata of a snapshot is also stored in pack files and consequently there's no chance for the backend to distinguish data and metadata.

elkemper:
Hey @MichaelEischer, thanks for picking the PR!

Here's my line of thought when I investigated this issue.
I figured, that because of everything in the repository - is _just files_ then there should be one function, the sole responsibility of which is - save **any** file to the backend. Also, searching for the `StorageClass` argument, i decided, that `Save` function is what i'm looking for. And looking through the call queue - it appeared the last stand of Restic before handling the operation to the Minio dependency. 
Knowing that - I looked to the arguments of the function and understood that `h Handle` could help me to determine what is being saved right now. And thanks to all of you, it had a parameter Type, which corresponds to the enum with all object types. So I thought that is safe to assume, that I can filter for the `PackFile`.

Well, after your question, I've done some more digging and went up on call stack and found that it used in all `Flush` functions - for data and for indexes. And handlers with datatype are passed down to it. 

Also, as a proof of work - here's bunch of screenshots of S3 when I tried this code. (saving just one small text file)
Config (and keys) are stored in `Standard` because when I `init`-ed the repo I didn't use StorageClass option (not sure even will it have any effect thou). And you can mention the date of files created.
The Snapshots, Index and Data are created few minutes later, and you can see that they have different Storage Classes
<details> 
<summary>Screenshots</summary>
image
imageimage
image
image
image
</details>

And in opposite, here's my first attempts of saving to the Glacier, when I didn't know about issue.
<details>
<summary>Screenshots of my tries and errors</summary>
image
image
image
</details>

Hope this helps.

MichaelEischer:
A pack file can contain either tree or data blobs. To allow a second host without a prefilled cache to use the repository for backups, the second host must also be able to retrieve the tree blobs. However, both tree and data blobs are stored in pack files. To see what I mean just try the following (while using your PR):

```
restic -o s3.storage-class=GLACIER backup some-small-folder
restic -o s3.storage-class=GLACIER backup --no-cache some-small-folder
```
The second backup run will use the snapshot from the first backup run as its parent snapshot. And it will try to download the corresponding tree blobs, which will fail.

For that to work a pack file must also not be archived if `h.IsMetadata` in `h backend.Handle` is true.

I'm not entirely sure how you intend the `restore` procedure to work. Do I have to retrieve the whole backup archive?

internal/backend/s3/s3.go
```diff
@@ -271,12 +271,22 @@ func (be *Backend) Path() string {
 	return be.cfg.Prefix
 }
 
+// Returns whether file should be saved in the provided Storage Class
+func (be *Backend) PassStorageClass(h backend.Handle) bool {
```
MichaelEischer, line 275: that method name should start with a lower case letter and probably be called `useStorageClass`

internal/backend/s3/s3.go
```diff
@@ -271,12 +271,22 @@ func (be *Backend) Path() string {
...
+	var notArchive bool = be.cfg.StorageClass != "GLACIER" && be.cfg.StorageClass != "DEEP_ARCHIVE"
+	var res bool = h.Type == backend.PackFile || notArchive
```
MichaelEischer, line 277: I'd prefer something like
```
isDataFile := h.Type == backend.PackFile && !h.IsMetadata
return isDataFile || notArchiveClass
```

changelog/unreleased/issue-4583
```diff
@@ -0,0 +1,7 @@
+Bugfix: Ignoring the s3.storage-class option for metadata when archive tier is specified
+
+Restic now will save snapshot metadata to non-archive storage tier whatsoever,
+thus snapshot metadata could be accessed from machine without local cache.
+
+https://github.com/restic/restic/issues/4583
+https://github.com/restic/restic/issues/3202
```
MichaelEischer, line 7: This PR does not add full blown support for archive storage. As such it should be clear after reading that there is _no_ official support for cold storages.

elkemper:
Thanks for the comments.
You are right @MichaelEischer, I didn't know this about packfiles. I tried to add to the test backup made with my first build and it couldn't get snapshot metadata.
After accepting your snippets I rechecked updating the backup as well as forget and prune and everything worked as expected.

As of how I'm expecting to use this backups - to make it the "1" from "3-2-1" backup strategy. I want a cheap storage for valuable media only in case of failures of local copies and raid setup. So yeah, if I will restore it, it will be a full restore.

MichaelEischer (approved):
LGTM.

I've reworded the changelog and made a few additional minor cleanups.

Issue #2654, opened 2020-03-18
restic-rewrite: change snapshot metadata


ATLief:
**Output of `restic version`**
restic 0.9.6 compiled with go1.13.4 on linux/amd64

**What should restic do differently? Which functionality do you think we should add?**
restic should have the ability to change the metadata of existing snapshots. Properties that should be modifiable include: Paths, Host, ID, Time, and Tags.          

**What are you trying to do?**
Since moving to restic, I have converted previous system backups (from different backup solutions) into restic snapshots with the appropriate metadata. Unfortunately I made a few mistakes with some of the snapshots' Times and Paths. If I'd like to correct those mistakes now, I'd need to restore each snapshot, remove it from restic, and back it up again with the correct metadata.

This also has the unfortunate side effect that when I specify `--path`, these older snapshots aren't selected for listing for forgetting.

rawtaz:
@ATLief Thanks, but can you be more specific about your use case? Please give one or a couple of concreted examples of what metadata you want to change, from what and to what.

ATLief:
ID Time Host Tags Paths

Current:
> a9baf756 2018-08-18 22:11:23 postulate old2,important /mnt/backup/duplicity/1
> b69c1e31 2018-08-20 13:00:02 postulate old2,important /mnt/backup/duplicity/2

Desired:
> a9baf756 2018-08-18 22:11:23 postulate old2,important /
> b69c1e31 2018-08-20 13:00:02 postulate old2,important /

Current:
> 7b9e60da 2020-02-09 15:54:56 postulate new /,/proc/cpuinfo,/proc/mdstat,/proc/meminfo,/proc/partitions
> 092d5547 2020-02-11 10:04:01 postulate new /,/proc/cpuinfo,/proc/mdstat,/proc/meminfo,/proc/partitions

Desired:
> 7b9e60da 2020-02-09 15:54:56 postulate new /
> 092d5547 2020-02-11 10:04:01 postulate new /

Current:
> 964d8b03 2019-07-31 23:16:06 postulate old /
> 90ed2cdc 2019-08-31 23:16:16 postulate old /

Desired:
> 964d8b03 2019-07-21 23:04:01 postulate old /
> 90ed2cdc 2019-08-21 23:04:01 postulate old /

Current (Hypothetical):
> 0db2e713 2020-03-01 23:04:01 postulate new /
> 731e44e0 2020-03-08 23:04:01 postulate new /

Desired (Hypothetical):
[...]

PR #4604, opened 2023-12-30
Add --new-path-remove and --new-path-add to rewrite command

neclepsio:

This PR adds `--new-path-remove` and `--new-path-add` to rewrite command.

Related to https://github.com/restic/restic/issues/2654.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4603
+++ b/changelog/unreleased/pull-4603
@@ -1,0 +1,5 @@
+Enhancement: Add `--new-path-add` and `--new-path-remove` options to `rewrite` command
+
+`restic rewrite` now allows rewriting the path metadata of a snapshot.
+
+https://github.com/restic/restic/pull/4603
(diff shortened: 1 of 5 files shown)


MichaelEischer:
In https://forum.restic.net/t/change-snapshot-path/7054/2 this PR was motivated with avoiding full backup rescans when e.g. the mount path of a drive changes. However, just changing the paths of a snapshot won't work for backups using absolute paths whereas it is unnecessary for backups using relative paths.

To properly handle absolute path we’ll likely have to adapt something like https://github.com/restic/restic/issues/4026 or add a `--strip-prefix` option to the backup command. For relative paths it is already possible to avoid the rescans using `cd folder && restic backup --tag backup-identifier --group-by host,tag ./`. The relative paths ensure that the folder structures in the snapshot only start from within `folder` and the `tag` allows the backup command to automatically identify a parent snapshot without requiring any manual snapshot modifications.

MichaelEischer:
> The backup was made on one Linux machine and the source is an external NTFS HDD. Then I updated the backup from a Windows machine.

That combination will likely always cause a full rescan. Linux and Windows (and the implementation in restic) handle file metadata differently. I'm not sure whether it is possible at all to retrieve the exact same set of metadata on Linux and Windows.

MichaelEischer:
Closing as discussed above.

PR #4614, opened 2024-01-05
Back up and restore Windows Alternate Data Streams

### LINKED ISSUE #1401: Does restic back up alternate data streams attached to NTFS files?

aneesh-n:
Restic did not back up Alternate Data Streams in Windows. Restic now backs up Alternate Data Streams (ADS) and restores them back to the main files.
The Alternate Data Streams are backed up like any other normal files, and the full name of the stream like 'file_name.txt:stream_name' is stored as the name of the file representing the ADS.
During restore, the ADS files are restored and attached to the original files as Alternate Data Streams. For progress and summary, the ADS are not counted in the "filesFinished" and "filesTotal" count, but the sizes of the ADS files are counted.

This feature does not require any update to the repository structure.
However, if https://github.com/restic/restic/pull/4611 is merged, it can be leveraged for making an optimization in this PR.
While restoring the ADS files, since the main file and ADS can be restored in any order, it is possible that the ADS can be restored before the main file, in which case it creates an empty main file and starts the ADS restore. After this when the main file is restored, if it is restored with the os.O_CREATE option, then it would replace the full file and ADS would be lost. Hence, we need to first try without the os.O_CREATE option and if we find that the file does not exist, then use the os.O_CREATE option. Presently this is being done for all files. With https://github.com/restic/restic/pull/4611 merged, we would be able to add a new generic attribute TypeADS to indicate if a file has ADS and only do the check without the os.O_CREATE option for those files.

Closes https://github.com/restic/restic/issues/1401

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5063
+++ b/changelog/unreleased/issue-5063
@@ -1,0 +1,10 @@
+Bugfix: Correctly `backup` extended metadata when using VSS on Windows
+
+On Windows, when creating a backup using the `--use-fs-snapshot` option,
+then the extended metadata was not read from the filesystem snapshot. This
+could result in errors when files have been removed in the meantime.
+
+This issue has been resolved.
+
+https://github.com/restic/restic/issues/5063
+https://github.com/restic/restic/pull/5097
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -97,6 +97,7 @@ type BackupOptions struct {
 }
 
 var backupOptions BackupOptions
+var backupFSTestHook func(fs fs.FS) fs.FS
 
 // ErrInvalidSourceData is used to report an incomplete backup
 var ErrInvalidSourceData = errors.New("at least one source file could not be read")
@@ -580,6 +581,10 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 			ReadCloser: source,
 		}
 		targets = []string{filename}
+	}
+
+	if backupFSTestHook != nil {
+		targetFS = backupFSTestHook(targetFS)
 	}
 
 	// rejectFuncs collect functions that can reject items from the backup based on path and file info
--- a/cmd/restic/cmd_backup_integration_test.go
+++ b/cmd/restic/cmd_backup_integration_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"testing"
+	"time"
 
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/restic"
@@ -109,6 +110,63 @@ func TestBackupWithRelativePath(t *testing.T) {
 	latestSn, _ := testRunSnapshots(t, env.gopts)
 	rtest.Assert(t, latestSn != nil, "missing latest snapshot")
 	rtest.Assert(t, latestSn.Parent != nil && latestSn.Parent.Equal(firstSnapshotID), "second snapshot selected unexpected parent %v instead of %v", latestSn.Parent, firstSnapshotID)
+}
+
+type vssDeleteOriginalFS struct {
+	fs.FS
+	testdata   string
+	hasRemoved bool
+}
+
+func (f *vssDeleteOriginalFS) Lstat(name string) (os.FileInfo, error) {
+	if !f.hasRemoved {
+		// call Lstat to trigger snapshot creation
+		_, _ = f.FS.Lstat(name)
+		// nuke testdata
+		var err error
+		for i := 0; i < 3; i++ {
+			// The CI sometimes runs into "The process cannot access the file because it is being used by another process" errors
+			// thus try a few times to remove the data
+			err = os.RemoveAll(f.testdata)
+			if err == nil {
+				break
+			}
+			time.Sleep(10 * time.Millisecond)
+		}
+		if err != nil {
+			return nil, err
+		}
+		f.hasRemoved = true
+	}
+	return f.FS.Lstat(name)
+}
+
+func TestBackupVSS(t *testing.T) {
+	if runtime.GOOS != "windows" || fs.HasSufficientPrivilegesForVSS() != nil {
+		t.Skip("vss fs test can only be run on windows with admin privileges")
+	}
+
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	testSetupBackupData(t, env)
+	opts := BackupOptions{UseFsSnapshot: true}
+
+	var testFS *vssDeleteOriginalFS
+	backupFSTestHook = func(fs fs.FS) fs.FS {
+		testFS = &vssDeleteOriginalFS{
+			FS:       fs,
+			testdata: env.testdata,
+		}
+		return testFS
+	}
+	defer func() {
+		backupFSTestHook = nil
+	}()
+
+	testRunBackup(t, filepath.Dir(env.testdata), []string{"testdata"}, opts, env.gopts)
+	testListSnapshots(t, env.gopts, 1)
+	rtest.Equals(t, true, testFS.hasRemoved, "testdata was not removed")
 }
 
 func TestBackupParentSelection(t *testing.T) {
--- a/cmd/restic/integration_helpers_test.go
+++ b/cmd/restic/integration_helpers_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"testing"
 
@@ -168,6 +169,16 @@ type testEnvironment struct {
 	gopts                                   GlobalOptions
 }
 
+type logOutputter struct {
+	t testing.TB
+}
+
+func (l *logOutputter) Write(p []byte) (n int, err error) {
+	l.t.Helper()
+	l.t.Log(strings.TrimSuffix(string(p), "\n"))
+	return len(p), nil
+}
+
 // withTestEnvironment creates a test environment and returns a cleanup
 // function which removes it.
 func withTestEnvironment(t testing.TB) (env *testEnvironment, cleanup func()) {
@@ -200,8 +211,11 @@ func withTestEnvironment(t testing.TB) (env *testEnvironment, cleanup func()) {
 		Quiet:    true,
 		CacheDir: env.cache,
 		password: rtest.TestPassword,
-		stdout:   os.Stdout,
-		stderr:   os.Stderr,
+		// stdout and stderr are written to by Warnf etc. That is the written data
+		// usually consists of one or multiple lines and therefore can be handled well
+		// by t.Log.
+		stdout:   &logOutputter{t},
+		stderr:   &logOutputter{t},
 		extended: make(options.Options),
 
 		// replace this hook with "nil" if listing a filetype more than once is necessary
--- a/internal/archiver/file_saver.go
+++ b/internal/archiver/file_saver.go
@@ -156,7 +156,7 @@ func (s *fileSaver) saveFile(ctx context.Context, chnker *chunker.Chunker, snPat
 
 	debug.Log("%v", snPath)
 
-	node, err := s.NodeFromFileInfo(snPath, f.Name(), fi, false)
+	node, err := s.NodeFromFileInfo(snPath, target, fi, false)
 	if err != nil {
 		_ = f.Close()
 		completeError(err)
--- a/internal/fs/fs_local_vss.go
+++ b/internal/fs/fs_local_vss.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/options"
+	"github.com/restic/restic/internal/restic"
 )
 
 // VSSConfig holds extended options of windows volume shadow copy service.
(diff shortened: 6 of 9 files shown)


internal/restorer/restorer.go
```diff
@@ -336,6 +336,15 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 	return err
 }
 
+func (res *Restorer) addFile(node *restic.Node, size uint64) {
```
MichaelEischer, line 339: Couldn't we just pretend the ADS are also regular files?

internal/restorer/fileswriter_windows.go
```diff
@@ -0,0 +1,35 @@
...
+// and only if the file does not exist, it opens it with create option
+// in order to create the file. This is done, otherwise if the ads stream
+// is written first (in which case it automatically creates an empty main
+// file before writing the stream) and then when the main file is written
+// later, the ads stream can be overwritten.
+func (*filesWriter) OpenFile(createSize int64, path string) (*os.File, error) {
+	var flags int
+	var f *os.File
+	var err error
+	// TODO optimize this. When GenericAttribute change is merged, we can
```
MichaelEischer, line 16: We don't need the GenericAttribute to know whether a file is an ADS or not; looking a the file name is enough.

This change is only correct, if the target directory is empty, otherwise the file can already exist and even have the wrong file size! So, I'm afraid we'll have to teach the `filerestorer.go` how to properly handle the ordering constraints for ADS. (I'm not sure so far what's the best approach for that)

MichaelEischer (commented):
I'm somewhat torn on whether I prefer the current approach of just mapping ADS into separate files and then merging them back together on restore, or whether I'd prefer them to be attributes of the corresponding directory / file. The mapping approach seems to be relatively simple to implement, and even allows the ADS to become visible on Unix system (as separate files, but I guess that's acceptable). But it somewhat complicates the restore steps. However, handling ADS as file attributes would require changes all across the place (even in prune) that are way more complex. So I guess the basic approach here is the right one.

I haven't looked at the windows API code yet, I guess there's no nice way to abstract it away ^^ ...

Would you like to prioritize this PR or #4611 ? I don't have time look at both (huge) PRs at the same time.

aneesh-n:
Thanks for looking into these. Definitely prefer prioritizing https://github.com/restic/restic/pull/4611.

internal/restorer/restorer.go
aneesh-n: Usually, users try to verify backups by relying on the file count as a metric. If ADS is treated as a regular file, the counts would be added and it could cause confusion.
Users would have to use dir /R to list the ADS as well and get the count. When you try to get the count by using Right Click -> Properties, which is what most users will do, you will not see the ADS count.
Similarly, if you restore on non-windows systems, ADS files will not appear, and the counts would differ.

Besides by definition I feel Alternate Data Streams are attachments to the main files and not equal in status to regular files.

internal/restorer/fileswriter_windows.go
aneesh-n: Actually, this TODO is in reference to the main file which has the ADS attached. Without this the main file is oblivious to any ADS attached to it, the ADS themselves just exist as independent nodes. 
And we want to know that the main file has ADS attached, because only in that case do we want to check for presence of the main file before using the O_CREATE option. Otherwise, we can directly use the O_CREATE option like the existing master code.

----------
Regarding the "This change is only correct, if the target directory is empty, otherwise the file can already exist and even have the wrong file size", I'm not sure I understand. There may be some confusion.

Let's say we have a test file - **D:\targetfile.txt**. It has an ADS attached - **D:\targetfile.txt:Stream1:$DATA**.
Now this file is backed up along with its ADS and these are stored as 2 separate files in the restic repo.

During restore, this method OpenFile will be called twice, and the calls can come any order depending on the content:

Once, for file name - D:\\targetfile.txt
Another time, for the ADS - D:\\targetfile.txt:Stream1:$DATA

(For now, consider that the target restore directory did not already exist.)

Whichever happens first will find - os.IsNotExist(err) is true and will then call the code below resulting in the creation of the main file D:\targetfile.txt.

```
flags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
f, err = os.OpenFile(path, flags, 0600)
```


**If it was called for D:\\targetfile.txt first**, 
then D:\\targetfile.txt will be created and the contents of the main file will be written to it. 
Then the call comes for D:\targetfile.txt:Stream1:$DATA at some point, it will find that the file for the stream does not exist.
It will use the os.O_CREATE option and it creates the stream only and writes the data to the stream without any issue.

**If it was called for D:\\targetfile.txt:Stream1:$DATA first,**
then it finds that the stream file does not exist and it uses the os.O_CREATE flag, and in that case, it creates an empty file D:\\targetfile.txt and starts writing the ADS at path - D:\\targetfile.txt:Stream1:$DATA.
Next when the call comes in for the main file targetfile.txt, it will find that os.IsNotExist(err) is false, because the main file was already created as a 0 size file and it will not use the os.O_CREATE option in this case. It will just continue to write the contents on the existing 0 size targetfile.txt.
Without this new check in place, when the second call comes in for targetfile.txt, it would replace the 0 size file which was written when the write for  D:\\targetfile.txt:Stream1:$DATA had started and ultimately the ADS is lost during restore.

**Even if the target restore directory and all the files already existed before restore**, the first time the file is opened, createSize >= 0 and we initially open the file with **os.O_TRUNC** | os.O_WRONLY. 
So, the file is truncated before the new content for restore is written. The file sizes after restore are correct. 
Am I missing some scenario?

aneesh-n:
> I'm somewhat torn on whether I prefer the current approach of just mapping ADS into separate files and then merging them back together on restore, or whether I'd prefer them to be attributes of the corresponding directory / file. The mapping approach seems to be relatively simple to implement, and even allows the ADS to become visible on Unix system (as separate files, but I guess that's acceptable). But it somewhat complicates the restore steps. However, handling ADS as file attributes would require changes all across the place (even in prune) that are way more complex. So I guess the basic approach here is the right one.
> 
> I haven't looked at the windows API code yet, I guess there's no nice way to abstract it away ^^ ...
> 
> Would you like to prioritize this PR or #4611 ? I don't have time look at both (huge) PRs at the same time.

Another reason to keep ADS as separate files is because technically they can be of any size. They're just like attaching a full blown file to another existing file which we refer to as the main file.
If we would add them as attributes, that would bloat the metadata.

Regarding restore of a file with ADS on non-windows, I think the technically correct expectation would be to ignore the streams on those systems since it is a windows specific function. Otherwise it will restore the ADS as separate files with name as filename:adsname:$DATA.
Any tool creating ADS on a file would do so with the expectation that the ADS would be used on the Windows system only.

On a side note, since unix filenames can have :, though not recommended for cross-platform use, this is another reason to use GenericAttributes to distinguish ADS files and ADS main files from normal files.
If a file has a GenericAttribute "WinADS" indicating it's a stream, it will be ignored from restores on non-windows systems.
If a file has a GenericAttribute "WinADSMain", indicating it has ADS streams attached to it, it will just be treated as a normal file. No special handling is needed on non-windows.

internal/archiver/archiver_unix.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/fs"
+	"github.com/restic/restic/internal/restic"
+)
+
+// resolveRelativeTargets replaces targets that only contain relative
+// directories ("." or "../../") with the contents of the directory. Each
+// element of target is processed with fs.Clean().
+func resolveRelativeTargets(filesys fs.FS, targets []string) ([]string, error) {
```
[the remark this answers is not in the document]
aneesh-n: Fixed thanks

internal/archiver/archiver_windows.go
```diff
@@ -0,0 +1,165 @@
...
+
+	debug.Log("targets after resolving: %v", result)
+	return result, nil
+}
+
+// preProcessTargets cleans the targets and adds the ads streams from each target to the targets
+func preProcessTargets(filesys fs.FS, targets *[]string) {
+	for _, target := range *targets {
+		target = filesys.Clean(target)
+		addADSStreams(target, targets)
```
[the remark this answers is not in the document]
aneesh-n: In the beginning, resolveRelativeTargets is called from scanner's Scan method and from archiver Snapshot method and the sorted cleanTargets obtained in that step are then passed to NewTree.
We need these ADS entries to be available before the Node creation in the tree.
We do add this in SaveDir as well, but it is also needed in resolveRelativeTargets.

internal/restorer/fileswriter_windows.go
aneesh-n: There is another scenario that depends on the GenericAttribute.
If a file exists in target with 10 ADStreams and it needs to be replaced with an older file with 5 ADStreams during restore, we need to know when restoring the main file, that the extra 5 streams need to be removed.
For this the main file needs to know what ADS streams are attached to it. Remaining streams will be removed during restore of main file.

So, the main file will have a GenericAttribute - TypeADSMain with value as -
Stream1|Stream2|Stream3 ...
I will update the code once https://github.com/restic/restic/pull/4611 is merged.

MichaelEischer:
> Regarding restore of a file with ADS on non-windows, I think the technically correct expectation would be to ignore the streams on those systems since it is a windows specific function. Otherwise it will restore the ADS as separate files with name as filename:adsname:$DATA.

I actually think that non-Windows systems should just create those `filename:adsname:$DATA` files, unix filesystems don't have problems with such filenames and the data stored in the ADS remains accessible.

ProactiveServices:
> > Regarding restore of a file with ADS on non-windows, I think the technically correct expectation would be to ignore the streams on those systems since it is a windows specific function. Otherwise it will restore the ADS as separate files with name as filename:adsname:$DATA.
> 
> I actually think that non-Windows systems should just create those `filename:adsname:$DATA` files, unix filesystems don't have problems with such filenames and the data stored in the ADS remains accessible.

This has the advantage that the user restoring data could choose to ignore ADS. Would it also mean that there's the possibility that a file with the "regular" name filename:adsname:$DATA could overwrite an ADS when backing up?

For example: document.docx has an ADS adsname:$DATA. If document.docx:adsname:$DATA is then backed up from a UNIX system, it "overwrites" the ADS data.

Would a regular file filename:adsname:$DATA on a Unix filesystem be backed up as NTFS ADS attached to document.docx? Would a user expect it to?

aneesh-n:
> > > Regarding restore of a file with ADS on non-windows, I think the technically correct expectation would be to ignore the streams on those systems since it is a windows specific function. Otherwise it will restore the ADS as separate files with name as filename:adsname:$DATA.
> > 
> > 
> > I actually think that non-Windows systems should just create those `filename:adsname:$DATA` files, unix filesystems don't have problems with such filenames and the data stored in the ADS remains accessible.
> 
> This has the advantage that the user restoring data could choose to ignore ADS. Would it also mean that there's the possibility that a file with the "regular" name filename:adsname:$DATA could overwrite an ADS when backing up?
> 
> For example: document.docx has an ADS adsname:$DATA. If document.docx:adsname:$DATA is then backed up from a UNIX system, it "overwrites" the ADS data.
> 
> Would a regular file filename:adsname:$DATA on a Unix filesystem be backed up as NTFS ADS attached to document.docx? Would a user expect it to?

Yes that would result in the ads data being overwritten as it currently stands.

I have made updates to this code, which use generic attributes to handle such cases and other cases and plan to rebase this after https://github.com/restic/restic/pull/4611 is merged. 

Even though this mechanism can be used to create ADS-like files in Unix systems, my concern is if it should.
I believe no other mechanism of copying or restoring files created in windows to unix would create such separate ADS files in unix. Are there any examples where restoring ADS data as a separate file on unix would be valuable? 

Most realworld usecases where ADS is useful is where tools are using them on windows.
If we go to the downloads on Windows and run "dir /r", we can see most downloaded files have ads streams like "abc.docx:Zone.Identifier:$DATA", "xyz.msi:SmartScreen:$DATA",  "aaa.zip:Zone.Identifier:$DATA", "aaa.zip:SmartScreen:$DATA" etc.
These are useful within windows only and would have no value in unix but rather cause bloat. If I restore my downloads folder which has 49 Files from Windows to Unix, it would create 112 files in my case.

vmlemon:
FWIW, there's some negative dimensions, to not supporting ADS/NTFS streams, beyond the usual "they're not used by anyone, and only a vector for malware/bloat" quips - quite a few folks still have older NTFS-formatted drives, containing Windows Explorer metadata, from earlier versions of Windows (XP/Server 2003 supported adding things like arbitrary comments/author info/ratings/keyword tags, to files), and there's several Macintosh/Windows interoperability products (MS's own File Services for Macintosh, ExtremeZ-IP, DAVE, and OWC MacDrive, at least, use ```AFP_AfpInfo```, and ```AFP_Resource```), where *not* archiving the ADSes would represent an irreversible loss of important data, such as resource forks, extended attributes, and file type information. 

(Recent versions of MacOS also use ADSes, via SMB, when storing metadata on Samba/Windows Server volumes).

I believe that some antivirus software, MS SQL Server, download management tools, e-mail clients, and "data-loss prevention" tools also make use of them, for various purposes, and some tools like HashTab, can also add checksum metadata, to files, using them.

PR #4550, opened 2023-10-30
Ls ncdu

ndecker:
What does this PR change? What problem does it solve?
-----------------------------------------------------
NCDU (NCurses Disk Usage) is a tool to analyse disk usage of directories.
It has an option to save a directory tree and analyse it later.
This patch adds an output option to the ls command for the NCDU format.

Use `restic ls latest --ncdu | ncdu -f -`

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
no

Created issue: #4549

Checklist
---------

- [X] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [X] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- [x] I have added documentation for relevant changes (in the manual).
- [X] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [X] I have run `gofmt` on the code in all commits.
- [X] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [X] I'm done! This pull request is ready for review.

PR #4615, opened 2024-01-06
walker: Remove ignoreTrees functionality

MichaelEischer:
It was only used in two places:
- stats: apparently as a minor performance optimization, which is unlikely to be important
- find: filtered directories would be ignored. However, this optimization missed that it is possible that two directories have the exact same content. Such directories would be incorrectly ignored too. Example:
```
mkdir test test/a test/b
restic backup test
restic find latest /test/b
-> incorrectly does not return anything
```

Thus, remove the functionality as it's apparently too complex to use correctly.

Discovered while reviewing #4550.

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #4605, opened 2023-12-31
Rework repository.StreamPacks & better restorer error handling

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
`restore` reports too many errors if a pack file can only be partially loaded. In addition, the restore progress bar is incorrect if restoring blobs from a pack file has to be retried. The PR reworks the error reporting and makes `StreamPack` (now `LoadBlobsFromPack`) easier to use and far more reliable. Blobs that cannot be loaded while streaming a pack file are now additionally loaded individually. 

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
Partially fixes https://forum.restic.net/t/errors-restoring-with-restic-on-windows-server-s3/6943

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added tests for all code changes.
- ~~[ ] I have added documentation for relevant changes (in the manual).~~
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #4626, opened 2024-01-07
Improve reliability of large restores

MichaelEischer:
Restoring large files can fail in some cases, for example https://forum.restic.net/t/errors-restoring-with-restic-on-windows-server-s3/6943 . This appears to be caused by blobs that are referenced a large number of times. When streaming a pack file, it takes a long time to write all instances of that blob to disk which can cause the network connection to be closed in the meantime. If writing the blob took longer than 15 minutes, then currently no retries are performed, which causes an incomplete restore.

https://github.com/restic/restic/pull/4624 has somewhat improved the situation by ensuring that each blob in a pack file is only processed once if streamPack has to retry the download. This helps if the blob processing delay is short enough that retries still happen, but won't fix other cases.

This PR addresses the problem from a different angle. It modifies the filerestorer such that blobs that are frequently referenced are downloaded individually and thereby avoids the timeout problem.

Related to https://github.com/restic/restic/pull/4605
Fixes https://forum.restic.net/t/errors-restoring-with-restic-on-windows-server-s3/6943

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4626
+++ b/changelog/unreleased/pull-4626
@@ -1,0 +1,11 @@
+Bugfix: Improve reliability of restoring large files
+
+In some cases restic failed to restore large files that frequently contain the
+same file chunk. In combination with certain backends, this could result in
+network connection timeouts that caused incomplete restores.
+
+Restic now includes special handling for such file chunks to ensure reliable
+restores.
+
+https://github.com/restic/restic/pull/4626
+https://forum.restic.net/t/errors-restoring-with-restic-on-windows-server-s3/6943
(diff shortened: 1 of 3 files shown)


MichaelEischer (commented):
LGTM.

MichaelEischer (commented):
LGTM.

JsBergbau:
>  It modifies the filerestorer such that blobs that are frequently referenced are downloaded individually and thereby avoids the timeout problem.

Let's say there is a blob referenced 1000 times and has 1 MB. Does this mean there is now 1000 MB downloaded, whereas before only 1 MB was downloaded?

MichaelEischer:
@JsBergbau No. The amount of downloaded data hardly changes. The only difference is that frequently referenced blobs are now downloaded separately from other blobs, but still only once. For example, if `restore` needs three blobs `blob1,blob2,blob3` from a pack file and `blob2` is frequently referenced, then `blob2` is downloaded separately from `blob1,blob3`. In a few cases this can mean that `blob2` is downloaded twice, but that's about it.

PR #4629, opened 2024-01-08
Feat AWS Assume Role add duration in seconds

konidev20:
This is w.r.t #4474  which was merged. The AWS AssumeRole endpoint provides the ability to specify the duration in seconds for the validity of the token. Refer: https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html

I see that it was missed in the previous linked PR. The `minio` library used does provide the ability to specify the duration in seconds under the `STSAssumeRoleOptions`.

I want the ability to change the duration in seconds for which the AssumeRole token is issued. I have some long running operation and want to keep 12 hours as the validity for the token I am using.

One thing to note here:
The `minio` library implicitly considers `DurationInSeconds` greater than 3600 seconds. I have updated the documentation as well highlighting the same.

If the environment variable is not set, it is simply ignored and the default value of 3600 seconds is implicitly accepted. 

This must be documented since, AWS AssumeRole allows you to specify durations between 900 and 43200 seconds. Since, `minio` allows only values above 3600 seconds we need to add it.

Related to #4474

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4629
+++ b/changelog/unreleased/pull-4629
@@ -1,0 +1,9 @@
+Enhancement: Feat AWS Assume Role add duration in seconds
+
+While using the AWS AssumeRole capability of the S3 storage backend, restic did not allow the duration in seconds to be configured for the STS token issued through the AssumeRole API.
+
+We have added the capability to specify the duration in seconds for the AssumeRole API using the RESTIC_AWS_ASSUME_ROLE_DURATION_IN_SECONDS environment variable.
+
+Note: `minio` currently considers only values above 3600 seconds to be configured.
+
+https://github.com/restic/restic/pull/4629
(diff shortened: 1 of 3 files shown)


MichaelEischer:
I'm not entirely sure I get the point of this PR. As far as I can tell, minio should be able to request a new token once the old one has expired.

konidev20:
Yes. Tweaking the duration for the validity of the STS token issued through the AssumeRole API is something I desire to change. 

Example instead of renewing it every 1 hour, I want to be able to specify it to renew every 3 hours. 

It's just another configuration value to be added. It would be helpful for me.

MichaelEischer:
> Example instead of renewing it every 1 hour, I want to be able to specify it to renew every 3 hours.

But what is the underlying reason for changing `DurationInSeconds`? So far you've only explained what the PR does, but not why. Which problem does the PR solve? 

> This doesn't affect the functionality, want this as a possible configuration option.

A configuration option that just exists in order to exist, doesn't make sense. Either there is a reason why the option is necessary or we shouldn't add the option.

konidev20:
One of the use cases I am looking for is the ability to skew the requests for STS token renewal when I have lots of parallel backup jobs running to S3. I don't want 100 pods to make the same request for token renewal which can overload my STS endpoint. 

Another one is to reduce the number of requests to the STS API. If I know a backup/restore job will be taking 10+ hours I might as well get a token for 12 hours (the max limit of the AssumeRole endpoint).

rawtaz:
@konidev20 Can you please clarify further - do you have an actual problem right now with your 100 pods, and if so what is it in concrete terms? And in what way would 100 pods renewing their tokens overload the STS service? After all, the default value has been set by upstream to one hour, so I'd expect them to be quite prepared for renewals to happen as frequently as one hour.

konidev20:
Nevermind, found another way. https://repost.aws/knowledge-center/iam-sts-throttling-error

PR #4639, opened 2024-01-18
Update Backblaze documentation

northben:
clarify documentation regarding B2 bucket lifecycle settings. The default lifecycle setting is probably fine for most users now; a custom policy is not necessary.

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -487,7 +487,8 @@ Backblaze B2
 
    Different from the B2 backend, restic's S3 backend will only hide no longer
    necessary files. Thus, make sure to setup lifecycle rules to eventually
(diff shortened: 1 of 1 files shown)


rawtaz:
I don't use B2, but won't the "Keep only the last version of the file" you mention keep hidden files forever? That is, restic hides a file which is no longer needed; with the "Keep only the last version of the file" policy in B2, that file will still be kept forever, right?

If yes, this seems like it would just keep using up space, as opposed to a policy that deletes hidden files after some time, which seems bad. Am I missing something?

MichaelEischer:
@rawtaz: "Keep only the last version of the file" keeps hidden files for only one day. That's the correct policy for most users.

MichaelEischer (approved):
LGTM.

PR #4650, opened 2024-01-21
backup: Improve help text for `--stdin-from-command`

MichaelEischer:

No.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4251
+++ b/changelog/unreleased/issue-4251
@@ -1,15 +1,16 @@
-Enhancement: Support reading backup from a program's standard output
-
-When reading data from stdin, the `backup` command could not verify whether the
-corresponding command completed successfully.
+Enhancement: Support reading backup from a commands's standard output
 
 The `backup` command now supports the `--stdin-from-command` option. When using
-this option, the arguments to `backup` are interpreted as a command. `backup`
-then executes the command and stores its standard output in the backup. This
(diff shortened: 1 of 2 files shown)


changelog/unreleased/issue-4251
rawtaz, line 1: I suggest changing it to:

~~~
Enhancement: Support reading backup from a commands's standard output

The `backup` command now supports the `--stdin-from-command` option. When using
this option, the arguments to `backup` are interpreted as a command instead of
paths to back up. `backup` then executes the given command and stores the
standard output from it in the backup, similar to the what the `--stdin` option
does. This also enables restic to verify that the command completes with exit
code zero. A non-zero exit code causes the backup to fail.

Note that the `--stdin` option does not have to be specified at the same time,
and that the `--stdin-filename` option also applies to `--stdin-from-command`.

Example: `restic backup --stdin-from-command --stdin-filename dump.sql mysqldump [...]`
~~~

The addition of `--stdin-filename` in the example is primarily to make it clearer that the `--stdin-from-command` doesn't take the command as value.

Also "program" was never used anywhere but in the heading, so I changed that to "command".

rawtaz (commented):
LGTM besides the changelog edits I think would be useful.

changelog/unreleased/issue-4251
MichaelEischer: Thanks. I've added the changes.

PR #4654, opened 2024-01-22
docs: fix formatting

adrian5:
I think this should be a regular list.

---

On a related note: I noticed that lists are often indented in the docs, i.e.
```rst
Here is a list:

 * Item1
 * Item2
 * Item3
```
This will render to `<blockquote><div><ul>…</ul></div></blockquote>`. I've replicated this style in my PR in case that's intentional, but I suspect you want plain `<ul>…</ul>` here, which you get by not indenting the list.

Likewise capitalization, list bullets (`*` vs `-`) and spacing differs between and within docs.

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -35,15 +35,15 @@ environment variable ``RESTIC_REPOSITORY_FILE``.
 For automating the supply of the repository password to restic, several options
 exist:
 
- * Setting the environment variable ``RESTIC_PASSWORD``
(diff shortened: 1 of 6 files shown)


MichaelEischer:
I don't think there's a particular reason to indent the unordered lists. I've added a commit to fix the indentation everywhere (which was inconsistent even within 040_backup.rst).

I'd like to keep the capitalization of `Inode` if it's used at the start of a sentence. The `inode` manpage does so too.

MichaelEischer (approved):
LGTM.

PR #4655, opened 2024-01-25
Enhancement: option to send HTTP over unix socket

### LINKED ISSUE #4287: Allow connecting with Unix Socket to Rest Server
Output of `restic version`
--------------------------

```
restic 0.15.1 compiled with go1.20.2 on linux/arm64
```

Allow connecting with Unix Socket to Rest Server instead of normal IP

I would like to use the rest server with sockets to have an cleaner setup with systemd services and don't have to think about ports

The best backup tool I had. It just works and is simple to use. Thanks guys!

ghost:

Add one method to send HTTP traffic (for any backend) over a unix socket.

I'm not sure this is the best way, but it certainly seemed an easy way without a large refactor.

Closes #4287 

Checklist
---------

Happy to add tests but wanted feedback on approach first. 

While adding syntax like `-r rest:unix:/tmp/socket` looked simple in the original issue, the way the code is structured it's difficult to get access to the `http.Transport` object from within the `rest` backend. Hence I added a global option, which means this would also work for tunnelling other HTTP based backends over a socket if desired.

Our intended use-case is in conjunction with the `rest-server` (see https://github.com/restic/rest-server/pull/272) where a central backup server essentially makes the following SSH calls out to hosts to back them up:

```bash
rest-server --listen unix:/tmp/socket --data /path/to/data &
scp restic host:/tmp
ssh -R /tmp/socket:/tmp/socket host /tmp/restic -use-unix-socket-for-http /tmp/socket -r rest:http://ignored/repo
```

(the `ignored` bit is ugly)

We can do the above with ephermal ports today, but unix sockets feels like it might be cleaner for this-use case.

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -201,15 +201,16 @@ scheme like this:
     $ restic -r rest:http://host:8000/ init
 
 Depending on your REST server setup, you can use HTTPS protocol,
-password protection, multiple repositories or any combination of
-those features. The TCP/IP port is also configurable. Here
-are some more examples:
+unix socket, password protection, multiple repositories or any
+combination of those features. The TCP/IP port is also configurable.
+Here are some more examples:
 
 .. code-block:: console
 
     $ restic -r rest:https://host:8000/ init
     $ restic -r rest:https://user:pass@host:8000/ init
     $ restic -r rest:https://user:pass@host:8000/my_backup_repo/ init
+    $ restic -r rest:http+unix:///tmp/rest.socket:/my_backup_repo/ init
 
 The server username and password can be specified using environment
 variables as well:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Get programs (Linux/macOS)
         run: |
           echo "build Go tools"
-          go install github.com/restic/rest-server/cmd/rest-server@latest
+          go install github.com/restic/rest-server/cmd/rest-server@master
 
           echo "install minio server"
           mkdir $HOME/bin
@@ -106,7 +106,7 @@ jobs:
           $ProgressPreference = 'SilentlyContinue'
 
           echo "build Go tools"
-          go install github.com/restic/rest-server/...
+          go install github.com/restic/rest-server/cmd/rest-server@master
 
           echo "install minio server"
           mkdir $Env:USERPROFILE/bin
--- a/changelog/unreleased/issue-4287
+++ b/changelog/unreleased/issue-4287
@@ -1,0 +1,14 @@
+Enhancement: support connection to rest-server using unix socket
+
+Restic now supports connecting to rest-server using a unix socket for
+rest-server version 0.13.0 or later.
+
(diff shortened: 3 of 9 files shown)


MichaelEischer (changes requested):
I strongly dislike adding a CLI option that then goes on to override the path passed to the backend. That just feels wrong. (It would also introduce multiple weird failure cases for other backends).

My suggestion is to use `"github.com/peterbourgon/unixtransport"` instead. With a call to `unixtransport.Register(tr)` it  becomes possible to use `rest:http+unix:///path/to/socket` without requiring a messy CLI option. That adds a new (very small) dependency, but that's better than the alternative.

ae-govau:
> My suggestion is to use `"github.com/peterbourgon/unixtransport"` instead. With a call to `unixtransport.Register(tr)` it becomes possible to use `rest:http+unix:///path/to/socket` without requiring a messy CLI option. That adds a new (very small) dependency, but that's better than the alternative.

Thanks, that library looks good and that's a much cleaner outcome. Have made change as suggested. On my system I also needed to make a fix to that library (so that it will ignore HTTP proxies when invoked). We might want to wait to verify that patch is accepted (and version bumped) before merging this one:
https://github.com/peterbourgon/unixtransport/pull/7

I've tested the above against the `rest-server` PR I made here:
https://github.com/restic/rest-server/pull/272

and it seems to work.

doc/030_preparing_a_new_repo.rst
```diff
@@ -201,15 +201,16 @@ scheme like this:
...
+unix socket, password protection, multiple repositories or any
+combination of those features. The TCP/IP port is also configurable.
+Here are some more examples:
 
 .. code-block:: console
 
     $ restic -r rest:https://host:8000/ init
     $ restic -r rest:https://user:pass@host:8000/ init
     $ restic -r rest:https://user:pass@host:8000/my_backup_repo/ init
+    $ restic -r rest:http+unix:///tmp/rest.socket:/my_backup_repo/ init
```
MichaelEischer, line 213: I wondering whether  we should also add `https+unix` here? Although I have to admit that I have no clue how the HTTPS certificate and host name checks would behave. So it might even be a good idea not to mention `https+unix`, hmm.

MichaelEischer (commented):
The code is ok now, but I'd really like to see a small test, that connecting to a `http+unix` server does work. Otherwise, there's a too large chance that we accidentally break this feature.

Waiting for a fix to the proxy support is probably a good idea. And https://github.com/restic/rest-server/pull/272 also has to be completed first.

ae-govau:
Happy to add tests once the server-part lands.

ae-govau:
Integration test added and cleaned up regular HTTP based `rest-server` test at same time.

ae-govau:
Rebased to fix `go.sum` conflicts. This remains ready for review.

MichaelEischer (approved):
LGTM. I've fixed the test failures (we'll have to switch rest-server back to `latest` once the next version is released) and there's now a changelog. Please take a look at it, afterwards we should be ready to merge this PR.

ae-govau:
Looks good. Thanks for fixing up and following this through.

PR #4657, opened 2024-01-25
Properly report the ID of newly added keys

### LINKED ISSUE #4656: add key does not report the new key's ID
Output of `restic version`
--------------------------
restic 0.16.3 compiled with go1.21.6 on linux/amd64

rest:http://

When adding a key to a repository, we currently have no "easy" way to know the new key's ID.
This is a problem because sometimes we need to pass this ID as a hint for other commands.

In particular, a key hint is needed when the repository has many keys - otherwise opening the repository may fail with "Fatal: maximum number of keys reached" even when a proper password is provided.

Expected behavior
-----------------

I suppose the author of the "key add" command wanted to have the ID printed like this:
```
saved new key as 116b1f6aaf7a299f6355bd7d892e2602e1a91f5b9ce8a69c6938ba3e4f1fcc79
```
Actual behavior
---------------

Currently `key add` displays instead:

```bash
restic -r /tmp/kevo/ key add
repository 70789ef8 opened (version 2, compression level auto)
enter new password: 
enter password again: 
saved new key as <Key of xx@yyy, created on 2024-01-25 10:17:55.560224177 +0100 CET m=+10.381054308>
```

The command dumps the Go object instead of the ID.
A trivial 1 line patch will be provided.

One day someone will outsmart all our defenses, burn our IT to the ground and destroy our other backups. Then the restic repos on the hard drives on the shelf it will save the day.

ls-initiatives:

Other commands like key list and key remove show the key's ID.

Showing the ID here lets users easily reuse the ID as a key hint for subsequent commands. In particular, a key hint is needed when the repository has many keys - otherwise opening the repository may fail with "Fatal: maximum number of keys reached" even when a proper password is provided.

Fixes #4656

### DIFF
--- a/changelog/unreleased/issue-4656
+++ b/changelog/unreleased/issue-4656
@@ -1,0 +1,7 @@
+Bugfix: Properly report the ID of newly added keys
+
+`restic key add` now reports the ID of a newly added key. This simplifies
+selecting a specific key using the `--key-hint key` option.
+
+https://github.com/restic/restic/issues/4656
+https://github.com/restic/restic/pull/4657
--- a/cmd/restic/cmd_key.go
+++ b/cmd/restic/cmd_key.go
@@ -142,7 +142,7 @@ func addKey(ctx context.Context, repo *repository.Repository, gopts GlobalOption
 		return err
 	}
 
-	Verbosef("saved new key as %s\n", id)
+	Verbosef("saved new key with ID %s\n", id.ID())
 
 	return nil
 }


cmd/restic/cmd_key.go
```diff
@@ -142,7 +142,7 @@ func addKey(ctx context.Context, repo *repository.Repository, gopts GlobalOption
 		return err
 	}
 
-	Verbosef("saved new key as %s\n", id)
+	Verbosef("saved new key as %s\n", id.ID())
```
MichaelEischer, line 145: Let's change the message a bit: `saved new key with ID %s`

changelog/unreleased/issue-4656
```diff
@@ -0,0 +1,9 @@
+Bugfix: Properly report the ID of newly added keys
+
+When adding a key to a repository, we had no "easy" way to know the new key's ID.
```
MichaelEischer, line 3: Please shorten this to something similar to the following:
```
`restic key add` now reports the ID of a newly added key. This simplifies
selecting a specific key using the `--key-hint keyID` option.
```

MichaelEischer (changes requested):
The ideal solution would be to add JSON support to `key add`. But I agree that the current output is not useful. So I'm fine with changing it, I just just have a few small remarks as commented below.

ls-initiatives:
Updates following your suggestions, thanks.
I agree that JSON output would be nice to have, though I can't write the patch for it. I'll file an issue about that if you want.

MichaelEischer:
> I agree that JSON output would be nice to have, though I can't write the patch for it. I'll file an issue about that if you want.

That would be great.

MichaelEischer (approved):
LGTM.

PR #4666, opened 2024-01-28
Implement feature flags

### LINKED ISSUE #4601: Add feature flags
Output of `restic version`
--------------------------
0.16.2

Add Feature Flags

## Motivation
We currently try very hard to keep the CLI backwards compatible once a feature is part of a release version. However, this limits experimentation with new features as these cannot be changed once they are included in a release version. In addition, there appears to be little usage of prerelease versions of restic, such that most feedback only arrives once a new version was released. This complicates adding new features to restic.

There is also no mechanism to deprecate features in a smooth way. A feature currently either exists or was removed. If users still need the old feature, then the only option is to rollback to an earlier version. What is missing is an intermediate state in which users can decide to temporarily reenable a feature.

To improve this situation, we should add feature flags, which make it possible to add experimental features and enable smooth deprecation.

## Proposed Implementation

Feature flags are specified by users via the `RESTIC_FEATURES` environment variable. The variable contains a comma-separated list of `key[=value],key2[=value2]` pairs. The key is the name of a feature flag and is written in `kebab-case`. The value is optional and can contain either the value `true` (default if omitted) or `false`.

It is an error if the user specifies an invalid / unknown feature flag.

A feature flag can either be in `alpha`, `beta` or `stable` state. An `alpha` feature is disabled by default, a `beta` feature is enabled by default. A `stable` feature is always enabled and cannot be disabled by the user. This allows for a transition period after which the flag is removed.

`alpha` features do not guarantee any backwards compatibility and may change in arbitrary ways between different restic versions. This also applies to `beta` features, although incompatible changes are less likely.

Restic also offers a `restic features` command that lists all available feature flags and their state.

MichaelEischer:
Implement feature flags as described in #4601.

Fixes #4601

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4601
+++ b/changelog/unreleased/issue-4601
@@ -1,0 +1,9 @@
+Enhancement: Add support for feature flags
+
+Restic now supports feature flags that can be used to enable and disable
+experimental features. The flags can be set using the environment variable
+`RESTIC_FEATURES`. To get a list of currently supported feature flags,
+run the `features` command.
+
+https://github.com/restic/restic/issues/4601
+https://github.com/restic/restic/pull/4666
--- a/cmd/restic/cmd_features.go
+++ b/cmd/restic/cmd_features.go
@@ -10,13 +10,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// FIXME explain semantics
-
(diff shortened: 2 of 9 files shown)


MichaelEischer (commented):
LGTM.

PR #4682, opened 2024-02-04
feat: set --target for dump command

### LINKED ISSUE #4678: add --target flag to dump command
Output of `restic version`
--------------------------

restic 0.16.2 (v0.16.2-63-gb72de5a88) compiled with go1.21.0 on linux/amd64

add --target flag to `dump` command just like `restore`

I want to backup&restore a block device, I can backup it by
```
restic backup --stdin-from-command -- dd if=/dev/sda
```

now I want to restore it by
```
restic dump {snapshotID} stdin --target /dev/sdb
```

currently I can restore it in this way:
```
restic dump {snapshotID} stdin > /dev/sdb
```

But it is not convenient in my env

we built a wonderful solution base on Resic, it really help

konidev20:

- added --target flag to the dumpOptions
- moved the checkStdoutArchive to run only if --target is not set

Closes #4678 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -57,7 +57,7 @@ func init() {
 	flags := cmdDump.Flags()
 	initSingleSnapshotFilter(flags, &dumpOptions.SnapshotFilter)
 	flags.StringVarP(&dumpOptions.Archive, "archive", "a", "tar", "set archive `format` as \"tar\" or \"zip\"")
-	flags.StringVarP(&dumpOptions.Target, "target", "t", "", "set the target path to dump the archive file")
+	flags.StringVarP(&dumpOptions.Target, "target", "t", "", "write the output to target `path`")
 }
 
 func splitPath(p string) []string {
@@ -69,11 +69,11 @@ func splitPath(p string) []string {
 	return append(s, f)
 }
 
-func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.BlobLoader, prefix string, pathComponents []string, d *dump.Dumper, checkStdoutArchiveFunc func() error) error {
+func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.BlobLoader, prefix string, pathComponents []string, d *dump.Dumper, canWriteArchiveFunc func() error) error {
 	// If we print / we need to assume that there are multiple nodes at that
 	// level in the tree.
 	if pathComponents[0] == "" {
-		if err := checkStdoutArchiveFunc(); err != nil {
+		if err := canWriteArchiveFunc(); err != nil {
 			return err
 		}
 		return d.DumpTree(ctx, tree, "/")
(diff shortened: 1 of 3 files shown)


cmd/restic/cmd_dump.go
```diff
@@ -168,7 +165,24 @@ func runDump(ctx context.Context, opts DumpOptions, gopts GlobalOptions, args []
 		return errors.Fatalf("loading tree for snapshot %q failed: %v", snapshotIDString, err)
 	}
 
-	d := dump.New(opts.Archive, repo, os.Stdout)
+	var outputWriter io.Writer = os.Stdout
+	if opts.Target != "" {
+		file, err := os.OpenFile(opts.Target, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
```
MichaelEischer, line 170: This should be `0o666`.

cmd/restic/cmd_dump.go
```diff
@@ -168,7 +165,24 @@ func runDump(ctx context.Context, opts DumpOptions, gopts GlobalOptions, args []
...
+		if err != nil {
+			return fmt.Errorf("cannot dump to file: %w", err)
+		}
+		defer func() {
+			_ = file.Close()
+		}()
+
+		outputWriter = file
+	} else {
+		if err := checkStdoutArchive(); err != nil {
```
MichaelEischer, line 180: That check previously only applied to cases in which `dump` would generate a tar/zip archive, but not when only outputting a single file. I'd like to keep it that way.

cmd/restic/cmd_dump.go
MichaelEischer: `os.O_WRONLY|os.O_CREATE|os.O_EXCL` is also sufficient.

MichaelEischer (changes requested):
Thanks for working on this feature! I only have two small comments, see below.

cmd/restic/cmd_dump.go
konidev20: Hey, 

Instead of creating a new Writer, I just updated the `os.Stdout` to the file. Because, if I used the `outputFileWriter` the `os.Stdout` would still be pointing to the Terminal.

cmd/restic/cmd_dump.go
```diff
@@ -168,6 +170,18 @@ func runDump(ctx context.Context, opts DumpOptions, gopts GlobalOptions, args []
...
+		file, err := os.OpenFile(opts.Target, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0o666)
+		if err != nil {
+			return fmt.Errorf("cannot dump to file: %w", err)
+		}
+		defer func() {
+			_ = file.Close()
+		}()
+
+		os.Stdout = file
+	}
```
konidev20, line 183: Hey @MichaelEischer, 

Can you check this area again and let me know if this is good?

cmd/restic/cmd_dump.go
```diff
@@ -165,24 +170,19 @@ func runDump(ctx context.Context, opts DumpOptions, gopts GlobalOptions, args []
...
 		defer func() {
 			_ = file.Close()
 		}()
 
-		outputWriter = file
-	} else {
-		if err := checkStdoutArchive(); err != nil {
-			return err
-		}
+		os.Stdout = file
```
MichaelEischer, line 182: Hard reject. Overwriting such a global variable is a guarantee for problems down the road.

Please just pass a flag to `printFromTree()` that states whether to call `checkStdoutArchive()` or not. Or alternatively pass the stdout check function as a parameter, and pass a no-op function when the target is a file. The latter is probably the cleaner solution.

MichaelEischer (changes requested):

cmd/restic/cmd_dump.go
konidev20: Done, I have implemented the latter solution. Thank you for pointing it out. It crossed my mind.

MichaelEischer (approved):
LGTM. Thanks!

I've rebased and squashed the commits and made a few further small tweaks.

Issue #1064, opened 2017-06-25
[FEATURE] exit with non-zero after receiving SIGINT, SIGTERM


wjkohnen:
## Output of `restic version`
```
restic 0.6.1 (v0.6.1-131-g144b7f33)
compiled with go1.8.3 on linux/amd64
```

## How did you start restic exactly? (Include the complete command line)
cat > /tmp/restic.sh <<EOF
#!/bin/bash
set -eu

restic -r /media/jb/easynova/backup/backup-ibuti.restic backup "$1"
echo "exit from restic: $?"
btrfs subvol delete "$1"
EOF
chmod 755 /tmp/restic.sh
for i in /.snapshot/* ;do /tmp/restic.sh "$i"; done

Wait until upload starts, then hit `^c`. 

## What backend/server/service did you use?
local filesystem

## Expected behavior
restic exits with non-zero code, leading to termination of script with non-zero code, ending the loop as intended.

## Actual behavior
restic exits with zero, loop continues.

## Steps to reproduce the behavior
see above

## Do you have any idea what may have caused this?
restic exits with code zero. Note this depends on bash; the same script with shebang `#!/bin/sh` exits with code 130 (/bin/sh links to dash). Suggestions on how to let bash scripts exit with non-zero when receiving a signal are very welcome. :slightly_smiling_face:

PR #4703, opened 2024-02-19
Catch SIGTERM, run cleanup

ferringb:
The previous code only ran cleanup (lock release for example) on SIGINT.  For anyone running restic in a container, the signal is going to be SIGTERM which means containerized execution would leave locks behind.

While this could be addressed via interposing dumb-init to translate the signal, a `kill` invocation is going to default to SIGTERM, so the same problem exists for non container users.

This change ensures restic shuts down cleanly when SIGTERM'd.  It currently handles SIGINT only.

For the containerized docker image, any stopping of the container results in restic being SIGTERM'd which means it
leaves behind locks and other undesirables.  This problem can also exist for CLI invocation of restic- `killall restic` would accomplish the same.

Ish.  People talk in the forums of shutting restic down with SIGINT/SIGTERM (https://forum.restic.net/t/robustness-of-repository-issue-resolved/1475/3 for example), but nothing in the codebases history- that I can find- has SIGTERM awareness.

Minimally, manual testing of current restic confirms it lacks SIGTERM handling, so I assume folks were just... assuming...

Checklist
---------

- [ ] I have added tests for all code changes. (no tests pre-exist for SIGINT)

### DIFF
--- a/changelog/unreleased/pull-4703
+++ b/changelog/unreleased/pull-4703
@@ -1,0 +1,9 @@
+Bugfix: Shutdown cleanly when SIGTERM is received
+
+Prior, if restic received SIGTERM it'd just immediately terminate skipping
+cleanup- resulting in potential issues like stale locks being left behind.
+
+This primarily effected containerized restic invocations- they use SIGTERM-
+but this could be triggered via a simple `killall restic` in addition.
+
+https://github.com/restic/restic/pull/4703
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -56,7 +56,7 @@ func RunCleanupHandlers(code int) int {
 	return code
 }
 
-// CleanupHandler handles the SIGINT signals.
+// CleanupHandler handles the SIGINT and SIGTERM signals.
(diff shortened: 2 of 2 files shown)


ferringb:
A point of discussion here is what to do with the env var `RESTIC_DEBUG_STACKTRACE_SIGINT`; while that's clearly for debugging, the name has 'SIGINT' in it; it my be desirable to rename that to reflect it's a debug dump for *any* cleanup.

That said, it's clearly something of an internal/debugging feature, so it likely can be ignored?

ferringb:
Also, https://github.com/restic/restic/issues/1064#issuecomment-406789277 implies SIGTERM was tested/validated; I cannot repro that (at least not for `restic backup`).

MichaelEischer (approved):
LGTM. Thanks!

Let's keep `RESTIC_DEBUG_STACKTRACE_SIGINT` as it is. It is primarily intended for Windows, and there only SIGINT exists. Only Unix system one would just send `SIGQUIT` to restic to achieve the same effect in a far more comfortable way.

PR #4709, opened 2024-02-24
Refactor locking into repository package

MichaelEischer:
The code to lock a repository was living in `cmd/restic`. However, the allowed read or write access to a repository depends on the type of lock held by the current process. Therefore, locking must be centrally managed by the repository to guarantee integrity of the repository.

The PR replaces the lockRepo and lockRepoExclusive functions with `openWithReadLock`, `openWithAppendLock` and `openWithExclusiveLock`. The code paths for `openWithReadLock` and `openWithAppendLock` are nearly identical, but the difference is in semantics: a read lock can be disabled by specifying `--no-lock`, whereas for a write lock the relevant question is whether the command is run in `--dry-run` mode or not.

The unified lock handling leads to a few minor cosmetic changes and three actual changes in behavior:
- `tag` no longer accepts the `--no-lock` flag. As it replaces files in the repository, this always requires an exclusive lock.
- `debug examine` now returns an error if both `--extract-pack` and  `--no-lock` are given.
- `ls` erroneously did not lock the repository. The old behavior is now available using `ls --no-lock`.

The remaining changes are a mix of fewer global variables and test cleanups to correctly lock the repository. The latter are the result of an experiment to actually enforce the access requested together with the lock (i.e., a read lock blocks all modifications). However, that code is still too horrible to show anyone ^^ .

TODOs

- [ ] Cleanup `repository.Lock` function interface

Part of the roadmap for restic 0.17

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4709
+++ b/changelog/unreleased/pull-4709
@@ -1,0 +1,10 @@
+Bugfix: Correct `--no-lock` handling of `ls` and `tag` command
+
+The `ls` command never locked the repository. This has been fixed. The old
+behavior is still supported using `ls --no-lock`. The latter invocation also
+works with older restic versions.
+
+The `tag` command erroneously accepted the `--no-lock` command. The command
+now always requires an exclusive lock.
+
+https://github.com/restic/restic/pull/4709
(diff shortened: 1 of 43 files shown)


MichaelEischer (commented):
LGTM

PR #4734, opened 2024-03-18
Add support for specifying --host via environment variable

### LINKED ISSUE #4733: Set --host via environment variable
Output of `restic version`
--------------------------

`restic 0.16.4 (v0.16.4-330-gd1d773cfc) compiled with go1.21.6 on linux/amd64`

A feature to allow the user to specify by environment variable the hostname for commands that acecpt the `--host` flag. This would allow the user to set the hostname once and not have to specify it for every command.

I use restic in an HPC environment to back up data from a shared filesystem. Because each compute node has a unique hostname, I have been setting `--host` to the cluster name for every command. This is error-prone and tedious, especially when using commands that accept `--host` alongside commands that do not,making it easy to forget to specify the `--host` flag when it is required.

Currently, I am sourcing a script that sets other parameters such as RESTIC_REPOSITORY and RESTIC_PASSWORD_FILE from environment variables, so it would be convenient to be able to set the hostname in the same way.

restic has been a great tool for us to back up our research data, and it has been fairly easy to teach end users how to use it. This feature would make it even easier to use.

maouw:

This PR adds the ability to set the hostname for most commands that accept the `--host` flag via the `RESTIC_HOST` environment variable. This allows the user to set the hostname once and not have to specify it for every command. Commands that accept `--host` but do not work on snapshots (e.g. `restic key ...`) are not affected by this change. The `--host` flag is still accepted and takes precedence over the `RESTIC_HOST` environment variable.

The changes are in `cmd_backup.go` for the `backup` command and `find.go` for commands that use the `restic.SnapshotFilter` options.

Closees #4733

Checklist
---------

- [x]  There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).

### DIFF
--- a/changelog/unreleased/issue-4733
+++ b/changelog/unreleased/issue-4733
@@ -1,0 +1,9 @@
+Enhancement: Allow specifying `--host` via environment variable
+
+Restic commands that operate on snapshots, such as `restic backup` and
+`restic snapshots`, support the `--host` flag to specify the hostname for
+grouoping snapshots. They now permit selecting the hostname via the
+environment variable `RESTIC_HOST`. `--host` still takes precedence over the
+environment variable.
+
+https://github.com/restic/restic/issues/4733
--- a/changelog/unreleased/issue-4733
+++ b/changelog/unreleased/issue-4733
@@ -7,3 +7,4 @@ environment variable `RESTIC_HOST`. `--host` still takes precedence over the
 environment variable.
 
 https://github.com/restic/restic/issues/4733
+https://github.com/restic/restic/pull/4734
--- a/cmd/restic/find.go
+++ b/cmd/restic/find.go
@@ -19,11 +19,9 @@ func initMultiSnapshotFilter(flags *pflag.FlagSet, filt *restic.SnapshotFilter, 
 	flags.Var(&filt.Tags, "tag", "only consider snapshots including `tag[,tag,...]` (can be specified multiple times)")
 	flags.StringArrayVar(&filt.Paths, "path", nil, "only consider snapshots including this (absolute) `path` (can be specified multiple times)")
 
-	if len(filt.Hosts) == 0 {
-		// parse host from env, if not exists or empty the default value will be used
-		if host := os.Getenv("RESTIC_HOST"); host != "" {
-			filt.Hosts = []string{host}
-		}
+	// set default based on env if set
+	if host := os.Getenv("RESTIC_HOST"); host != "" {
+		filt.Hosts = []string{host}
 	}
 }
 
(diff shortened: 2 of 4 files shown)


maouw:
I have a question to the maintainers -- where and/or how would you suggest putting the proper tests for this feature? Most of the tests related to this area seem to circumvent option processing, so I can't find a good example to follow. I am a golang beginner.

changelog/unreleased/issue-4733
```diff
@@ -0,0 +1,9 @@
+Enhancement: Allow specifying `--host` via environment variable
+
+Restic commands that operate on snapshots, such as `restic backup` and
+`restic snapshots`, support the `--host` flag to specify the hostname for
+grouoping snapshots. They now permit selecting the hostname via the
+environment variable `RESTIC_HOST`. `--host` still takes precedence over the
+environment variable.
+
+https://github.com/restic/restic/issues/4733
```
MichaelEischer, line 9: Please also add a link to this PR.

cmd/restic/find.go
```diff
@@ -14,17 +15,31 @@ func initMultiSnapshotFilter(flags *pflag.FlagSet, filt *restic.SnapshotFilter,
 	if !addHostShorthand {
 		hostShorthand = ""
 	}
-	flags.StringArrayVarP(&filt.Hosts, "host", hostShorthand, nil, "only consider snapshots for this `host` (can be specified multiple times)")
+	flags.StringArrayVarP(&filt.Hosts, "host", hostShorthand, nil, "only consider snapshots for this `host` (can be specified multiple times) (default: $RESTIC_HOST)")
 	flags.Var(&filt.Tags, "tag", "only consider snapshots including `tag[,tag,...]` (can be specified multiple times)")
 	flags.StringArrayVar(&filt.Paths, "path", nil, "only consider snapshots including this (absolute) `path` (can be specified multiple times)")
+
+	if len(filt.Hosts) == 0 {
```
MichaelEischer, line 22: Wrapping the following code in the if is unnecessary. The cli flag configuration and the cli parsing are two separate steps. At this point, `filt.Hosts` is always empty.

It does still work, as cobra internally tracks whether it has already applied an option or whether `filt.Hosts` still has its default value. So, just remove the `if len(filt.Hosts) == 0 {` check. The same also applies for `initSingleSnapshotFilter`.

MichaelEischer (changes requested):
Thanks for the PR! I only have a few comments, see below.

The best way for testing is probably to use something like the following. That simply creates a new flagSet, configures it, then parses a test string and verifies the outcome:
```
set := pflag.NewFlagSet("test", pflag.PanicOnError)
flt := &restic.SnapshotFilter{}
initMultiSnapshotFilter(set, flt, false)
err := set.Parse([]string{"--host", "abc"})
rtest.OK(t, err)
// assert on flt.Hosts
```

MichaelEischer (approved):
LGTM. I've applied my suggestions and added a short test.

maouw:
Thank you for taking care of the test. Very happy to see this coming through, it will be a great help.

ahmgithubahm:
Hi.  On the release BLOG [1], there is this line:

Support for specifying hostname via [environment variable](https://restic.readthedocs.io/en/stable/040_backup.html#environment-variables) RESTIC_HOST.

It links to the env var docs, but the docs don't appear to mention RESTIC_HOST.

thanks!

[1] https://restic.net/blog/2024-07-26/restic-0.17.0-released/

PR #4743, opened 2024-03-29
Deprecate s3legacy layout

### LINKED ISSUE #4602: Deprecation of Legacy S3 Layout and old Index Format
Output of `restic version`
--------------------------
0.16.2

Deprecate the following functionality in 0.17.0. The removal will happen 0.18.0. The deprecated features will be disabled by default, but use feature flags (#4601) to provide an escape hatch to temporarily reenable them.

- Legacy S3 Layout: since restic 0.7.0 the S3 backend uses the unified layout. The migration code seems to have been broken between restic 0.8.0 and 0.14.0 when using a cache. As we didn't receive a corresponding issue report, there appears to be very little usage of the S3 legacy layout. Supporting both layouts adds significant code complexity, which does not seem to be justified by remaining usage. Old repositories must be migrated prior to deprecation using `restic migrate s3_layout` to remain accessible.
- Old Index Format: the old format has only been used before restic 0.2.0, as such likely next to no repositories exist that still use this format. The repository index can be updated to the modern format using `restic repair index` (also possible using restic versions without support for the old index format).

MichaelEischer:
Disable support for the s3legacy layout to allow us to finally remove the code in a future restic version.

The local and sftp backend were also able to read repositories using the s3legacy layout. There is currently no documented migration path for such repositories, however, it is completely unclear whether such repositories even exist. I did not implement an upgrade path in restic, as the layout upgrade can also be handled using a short bash script if necessary.

Fixes https://github.com/restic/restic/issues/4602

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4602
+++ b/changelog/unreleased/issue-4602
@@ -1,7 +1,7 @@
-Change: Deprecate legacy index format
+Change: Deprecate legacy index format and s3legacy layout
 
 Support for the legacy index format used by restic before version 0.2.0 has
-been depreacted and will be removed in the next minor restic version. You can
+been deprecated and will be removed in the next minor restic version. You can
 use `restic repair index` to update the index to the current format.
 
 It is possible to temporarily reenable support for the legacy index format by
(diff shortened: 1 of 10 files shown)


MichaelEischer (commented):
LGTM.

Issue #4406, opened 2023-07-12
Please allow for restic to be used as a "Go library"


thewilli:
Output of `restic version`
--------------------------

```restic 0.15.2 compiled with go1.20.3 on darwin/arm64```

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Restic does not only do a good job as a backup utility, but is also written in Go just as many cloud-native and Kubernetes-based tools / operators are. Allowing restic to be used as part of another application instead of spawning a Go binary from another Go binary would not only be more resource-efficient, but would also allow to have a better control about the state and behavior.

Restic has most of its logic as an [`internal`](https://github.com/restic/restic/tree/e990d3d/internal) package, which [prevents](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit?pli=1) other libraries from importing. 

Please be less _restrictive_ when it comes to declaring parts of restic `internal` to allow other projects to use restic as a library.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I'd like to use restic as part of a larger construct (Kubernetes Operator) to automate different DR scenarios. I'd like to provide detailed information about the status, which could be easily achieved by a custom [StatusReporter](https://github.com/restic/restic/blob/e990d3d/internal/ui/backup/progress.go#L14-L25) if this wasn't _internal_.

Other use-cases exist (see for example [restic-robot](https://github.com/Southclaws/restic-robot)), where a Go application spawns restic as another process, introducing overhead and the binary as additional dependency.

Adding more flexibility and use-cases would increase the potential for people to contribute back to restic and increase its overall popularity.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

The flexibility to use restic even on Windows due to its Go nature is just perfect and makes a big difference when comparing it with other utilities like borg.

MichaelEischer:
The problem with exposing the internals of restic is that this would cause a lot of internal interfaces to essentially become an official API. Quite a few of these APIs expect to be called in the correct way to be safe. It is currently also not feasible to give _any_ guarantees whatsoever regarding stability of these interfaces. Providing such guarantees even for some of these interfaces would require a massive refactoring as preparation. The `cmd/restic` package currently also contains a significant amount of logic, which would have to be refactored first.

I understand the appeal of an API, however, I currently don't see how we could get there anytime soon with a reasonable amount of work.

LawyZheng:
[...]

PR #4753, opened 2024-04-05
Replace cleanup handlers with context based command cancelation

MichaelEischer:
Restic has historically relied on a homegrown mechanism using cleanup handlers to handle command interrupts like SIGINT. This PR completely removes that mechanism and only uses `Context`s to cancel commands. This removes another source of global state from `cmd/restic/` and might be useful in the future as it allows for a controlled shutdown of commands compared to the current approach that is more akin to a "controlled crash".

Due to the cleanup handlers approach there was no need to ensure that every command exits as fast as possible once a context has been canceled. This PR therefore adds lots of checks for canceled contexts. Previously, if the cleanup handlers took too long, then a command like prune could in some cases erroneously report a corrupted repository.

From a user perspective there shouldn't be any large changes, although the text output of an interrupted command will change slightly.

There are two ugly parts in the current implementation: reading a password and serving a fuse mount appear to offer no nice way to honor a canceled context. To still somewhat integrate with context cancellation, the blocking command runs in a separate goroutine that is leaked if the command gets canceled. This is good enough for uses in restic, as canceling the context will result in program termination anyways.

Mentioned as part of the restic 0.17.0 roadmap. Prerequisite for https://github.com/restic/restic/issues/4406 .

Checklist
---------

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -1,89 +1,41 @@
 package main
 
 import (
+	"context"
 	"os"
 	"os/signal"
-	"sync"
 	"syscall"
 
 	"github.com/restic/restic/internal/debug"
 )
 
-var cleanupHandlers struct {
-	sync.Mutex
-	list []func(code int) (int, error)
-	done bool
-	ch   chan os.Signal
+func createGlobalContext() context.Context {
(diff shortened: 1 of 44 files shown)


MichaelEischer:
LGTM

PR #4754, opened 2024-04-06
Add WebDAV server (ie "mount" on Windows)

ducalex:

It allows browsing repositories on operating systems where Fuse isn't available.

This PR should work on all OSes and adds no external dependencies (though I suppose x/net/webdav is "external", but it's part of golang). It does no write operation on the repository so it should be safe and I encourage you to try it! But you shouldn't blindly trust the data that it gives you. Performance might also be an issue on large repositories with overly eager clients (like Windows Explorer).

Whilst I'd be happy to see this PR through (fix my reckless path handling, lack of concurrency locks, performance issues, and add documentation and tests), I do no believe it will be accepted as is.

Realistically, `internal/fuse` needs to be abstracted into a VFS that could be shared by Fuse, WebDAV, HTML, etc. It does everything needed already (paths templating, snapshots tree building in a performant manner, file reading, it handles concurrency, etc) but it directly depends on `anacrolix/fuse` which isn't present in Windows builds. I do not think it would be very hard to refactor, but I feel unqualified since it's deep in the internals of restic and would touch a lot of files.

Special thanks to @wrouesnel, although I wasn't able to use his code in the end (^ that refactoring would be needed first), his attempt is what motivated to work on this.

https://github.com/restic/restic/assets/13711380/c5ec0c02-1ee1-41bb-8881-068b964b2edf

https://github.com/restic/restic/issues/485
https://github.com/restic/restic/issues/60

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_webdav.go
+++ b/cmd/restic/cmd_webdav.go
@@ -22,7 +22,7 @@ import (
 )
 
 var cmdWebdav = &cobra.Command{
-	Use:   "webdav [flags] address:port",
+	Use:   "webdav [flags] [ip:port]",
 	Short: "Serve the repository via WebDAV",
 	Long: `
 The "webdav" command serves the repository via WebDAV. This is a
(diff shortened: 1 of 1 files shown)


fd0:
Hey, thanks a lot for getting this started! I've (slowly) also started working on a WebDAV server implementation, and I think we can join forces here in getting the work done and the PR merged. :)

I don't have that much time for restic the last couple of years, but I'd like to get more involved again!

I've identified the same challenges:
 1. Refactoring the low-level fuse code (the virtual FS) into something we can use for fuse, webdav and also a web server (#4276). I have already started on this. In the process, we will need to a lot of tests (which the fuse code currently does not have).
 2. Build some kind of generic access control for the WebDAV server as well as a (future) web server (like #4276). I'm thinking about having a way to specify the username and password and require it via HTTP auth, if none is specified then a random one is generated and used. It can then be passed to the Explorer (or `gio-mount` on Linux) and will be handled transparently to the user. Something like that.

I'll get the code that I already have cleaned up and will push it somewhere.

garethgeorge:
Hey, interested to check on the status here / what's left to get this over the finish line?

I'd very much like to leverage WebDAV for faster cross-platform browsing in Backrest and may be able to lend some effort to help push this across the finish line if it's stalled but open for contribution.

PR #4761, opened 2024-04-10
cache: ignore ErrNotExist during cleanup of old files

### LINKED ISSUE #4760: Race condition in Clear() function in file restic/internal/cache/file.go
Output of `restic version`
--------------------------
`restic 0.16.4 compiled with go1.21.6 on linux/amd64`

`rest-server version rest-server 0.12.1 compiled with go1.20.5 on linux/amd64`

When performing backups on each host, I backup per filesystem separately.  This is to make it easier to restore based on mount point.  So for example, I backup `/` and `/home` separately.

I schedule these backups using cron and have them all start at the same time.

Sometimes, I get the following error (but not always):

```bash
error clearing pack files in cache: remove /home/<username>/.cache/restic/<hash>/data/1b/1b<hash>: no such file or directory
```

or

```bash
error clearing index files in cache: remove /home/<username>/.cache/restic/<hash>/index/79/79<hash>: no such file or directory
```

I suspect that there is a race condition when clearing the local cache of packs and indexes that no longer exist in the repository between different restic backup processes from the same user on the same host.  The race is occurring in the Clear method for file [/internal/cache/file.go at line 152](https://github.com/restic/restic/blob/6091029fd66eb8d35e8d3763996c741e3dbf5c5e/internal/cache/file.go#L152)

Expected behavior
-----------------

If an obsolete index or pack file in the cache is found, and upon attempting to delete it, the operating system reports that it does not exist - this error should be ignored.  

Or alternately, local locking semantics are used so that only one restic process is able to clear the cache of obsolete entries at a time.

The latter would be the more 'correct' solution.   But admittedly, it is an edge case and ignoring `os.ErrNotExist` would probably suffice.

Actual behavior
---------------

A race condition can occur where multiple restic processes are building a list of obsolete entries, and then attempting to delete them at the same time, resulting in the above unpredictable outcomes.  The process that encounters a missing entry immediately stops processing its list of files, and if the list of files is ordered, would mean that ultimately, the files would get removed.  However, the error that results reports a harmless happenstance, and is unnecessary.

Restic is backing up a dozen computers for me now, and is a joy to work with, providing very efficient and reliable backups of my personal files.

MichaelEischer:
Two restic processes running concurrently can try to remove the same files from the cache. This could cause one process to fail with an error if the other one has already remove a file that the current process also tries to delete.

See https://github.com/restic/restic/issues/4760

Fixes https://github.com/restic/restic/issues/4760

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4760
+++ b/changelog/unreleased/issue-4760
@@ -1,0 +1,8 @@
+Bugfix: Fix possible error on concurrent cache cleanup
+
+If multiple restic processes concurrently cleaned up no longer existing files
+from the cache, this could cause some of the processes to fail with an `no such
+file or directory` error. This has been fixed.
+
+https://github.com/restic/restic/issues/4760
+https://github.com/restic/restic/pull/4761
--- a/internal/cache/file.go
+++ b/internal/cache/file.go
@@ -165,7 +165,8 @@ func (c *Cache) Clear(t restic.FileType, valid restic.IDSet) error {
 			continue
 		}
 
-		if err = fs.Remove(c.filename(backend.Handle{Type: t, Name: id.String()})); err != nil {
+		// ignore ErrNotExist to gracefully handle multiple processes running Clear() concurrently
+		if err = fs.Remove(c.filename(backend.Handle{Type: t, Name: id.String()})); err != nil && !errors.Is(err, os.ErrNotExist) {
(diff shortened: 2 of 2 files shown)


MichaelEischer (commented):
LGTM.

PR #4764, opened 2024-04-15
Prevent unsafe uses of `forget --keep-tag`

### LINKED ISSUE #4569: restic forget --keep-tag NonExistingTag drops all snapshots

MichaelEischer:

Currently it is possible to accidentally remove all snapshots using e.g. `forget --keep-tag invalid` where `invalid` is a not existing tag. This is particularly problematic when making typos in the tag name.

This PR changes the behavior as follows:
- If `--keep-tag x` were to cause all snapshots in a snapshot group to be removed, then `forget` will fail with an error
- `forget` fails with an error if no keep policy was specified instead of just printing a warning
- Add a new `--unsafe-allow-remove-all` option. To use the `--unsafe-allow-remove-all` option a snapshot filter (host, path or tag) must be specified. `forget --tag x --unsafe-allow-remove-all` can for example be used to remove all snapshots with a specific tag.

This is slightly more restrictive than my suggestion in https://github.com/restic/restic/pull/4568#pullrequestreview-1968554098 . That also allowed `forget --keep-tag invalid --unsafe-allow-remove-all` (deletes the whole repository as is currently the case) which won't be allowed in the current version of the PR. It also blocks `forget --keep-tag some --unsafe-allow-remove-all` where `some` is a tag only used by some snapshot groups. The latter might be relevant for some use cases, although I'm not sure what those would look like.

I'll likely also split this PR into two parts, one to fix the `--keep-tag` bug and one to add the `--unsafe-allow-remove-all` option.

Replaces https://github.com/restic/restic/pull/4568
Fixes https://github.com/restic/restic/issues/4569
See also https://forum.restic.net/t/delete-all-snapshots-in-one-command-is-this-feature-intentional/6923/3

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #4784, opened 2024-04-29
Rework backend retries

MichaelEischer:

Most restic commands are long running and would have to start over in case of a failed backend operation. Thus, retries are necessary to work reliably even if some requests fail.

From what I'm aware of, there are two main uses cases for retries in restic:
- retry a few failed operations. Failed operations are rare.
- tolerate a network connection that's interrupted for a few minutes

The PR makes several changes:
- Increase the initial retry interval (1s) and also the multiplier (2x). This still allows the first few retries to happen within a few seconds. All later retries are primarily useful to tolerate long lasting problems, thus a long interval is perfectly fine. This reduces the overall number of retries.
- Only limit the retry duration instead of the number of retries. The values in the PR result in approximately 21 retries within the 15 minutes retry period. The internal concurrency of restic is bounded. Therefore, request retries reduce the number of requests per time sent to the backend. This should help overloaded backend recover.
- Always retry at least once. In some cases the first request already takes longer than 15 minutes. In this case, previously no retry happened.
- Wait at most 1 minute to unlock the repository if the user interrupted the current operation. This limits the maximum unlocking delay when using restic interactively, while giving background operations more time to clean up their lock file.

Replaces #2515.
Mentioned in #4627.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4627
+++ b/changelog/unreleased/issue-4627
@@ -9,7 +9,9 @@ downloading are now forcibly interrupted. This ensures that stuck requests are
 retried after a short timeout.
 
 Attempts to access a missing file or a truncated file will no longer be retried.
-This avoids unnecessary retries in those cases.
+This avoids unnecessary retries in those cases. All other backend requests are
+retried for up to 15 minutes. This ensures that a temporarily interrupted network
+connections can be tolerated.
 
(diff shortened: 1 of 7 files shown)


MichaelEischer (commented):
LGTM

PR #4793, opened 2024-05-03
Restic dump perf changes

ashp7:

Checklist
---------

- [ ] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).


MichaelEischer:
I've reimplemented a fixed version of the dump parallelization in https://github.com/restic/restic/pull/4796 . Initially, I thought about asking whether you'd be interested in contributing this optimization to restic. However, your current code changes include a major bug that can result in corrupted output of the dump command. Fixing it required a different design of the goroutine pipelines which unfortunately resulted in a complete rewrite. (If you're interested I can provide a few more details on the bug and other problems/style issues with the code changes)

ashp7:
> I've reimplemented a fixed version of the dump parallelization in #4796 . Initially, I thought about asking whether you'd be interested in contributing this optimization to restic. However, your current code changes include a major bug that can result in corrupted output of the dump command. Fixing it required a different design of the goroutine pipelines which unfortunately resulted in a complete rewrite. (If you're interested I can provide a few more details on the bug and other problems/style issues with the code changes)

 I looked at you earlier thread and wanted to enable parallel blob processing and also use a buffer pool. I made the changes to test locally without being production ready, but ended up wrongly pushing it.  I am interested and will be glad to understand more details on the bug.

MichaelEischer:
The comments are based on https://github.com/NetApp-Polaris/restic/blob/ce43ae66d9200344e0e24ea5f64e0ed19365f6e0/internal/dump/common.go .

Minor issues / code style:
- Line 150: The weighted semaphore has no effect at all and should simply be removed. It starts with 100 initial tokens, but there are only 20 workers, which acquire a single token. Thus, each worker always gets a token immediately.
- Line 152: This reimplements https://pkg.go.dev/golang.org/x/sync/errgroup . Just use that instead.
- Line 193: Concurrency in restic is typically controlled based on `repo.Connections()` (see my PR for the necessary changes)
- Line 210: The batching mechanism offers no benefit so far as each blob is still requested individually; it only complicates the code. As an individual blob is already rather large, there's no real benefit from trying to amortize the channel send overhead (neither are CPU cache effects relevant here). What will be relevant later on, is that the batching mechanism manages to hide to the data corruption issues as only a single goroutine is active for up to 20 blobs.
- Line 209: `enqueueBatches` receives a `ctx` but never uses it. Thus, context cancelation is broken here.
- Line 168: The code no longer uses the bloblru cache. This is problematic when restoring large VM images, which often contain long sequences of zeros, which can be optimized via the cache. I have to admit that using the cache correctly introduces significant complexity. As discussed in my PR, there is no longer a simple way to recycle the byte slices without risking data corruption.
- Line 33: At least starting from restic 0.16.0, a single GC run usually only takes a few milliseconds even for very large indexes. Thus, a buffer pool is probably not even necessary. As your fork is still based on restic 0.15.1, it doesn't benefit from these improvements so far.

Data corruption bugs:
- Line 179: This is the more "obvious" one. The worker goroutines just write to `d.w.Write()` without any synchronization. This can create problems within `Write()`, but even worse does not guarantee that blobs are written in the order in which they are contained in the file. The only fix for that is to use a different goroutine architecture as in my PR.
- Line 168 and 169: The problem here is hidden well enough that I didn't notice it at first. If `buf` is large enough, then `LoadBlob` can reuse that buffer. That is `buf` and `blob` might refer to the same underlying slice! By putting the buffer back into the pool, it might be reused by a different goroutine while it is processed by `Write()` in line 179.

PR #4796, opened 2024-05-05
dump: Parallelize loading large files

MichaelEischer:
Improve dump performance for large files by loading their blobs concurrently using `repo.Connections()` goroutines.

The largest part of the code changes in here is actually a modification of the bloblru cache. There's now a new `GetOrCompute` method that ensures that if multiple goroutines try to get the same blob that is not yet in the cache, that the blob will only be downloaded once.

I've initially tried to use a sync.Pool to reuse byte slices, however, that turned out to be rather complicated to get right when using the bloblru cache. A byte slice may only be reused if no other goroutine can have a reference to it. With the current interface this is next to impossible to guarantee.

Basic variant as discussed in https://github.com/restic/restic/issues/3406 .

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4796
+++ b/changelog/unreleased/pull-4796
@@ -1,0 +1,8 @@
+Enhancement: Improve `dump` performance for large files
+
+The `dump` command now retrieves the data chunks for a file in parallel. This
+improves the download performance by up to the configured number of parallel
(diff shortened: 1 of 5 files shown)


MichaelEischer (commented):
LGTM

PR #4799, opened 2024-05-09
Azure: add option to force use of CLI credential

letmaik:

In Azure, VMs can have multiple identities at the same time, for example a managed identity and an Azure CLI identity. Sometimes the managed identity is not under control of the user but the user can still login with the Azure CLI. In those cases, being able to use the Azure CLI identity with restic makes sense.

`DefaultAzureCredential` first tries environment variables, managed identity, workload identity, and eventually Azure CLI identity. This PR introduces a new environment variable that forces use of the Azure CLI identity:

```
export AZURE_FORCE_CLI_CREDENTIAL=true
```

Checklist
---------

  - [x] Manually tested. 

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -550,16 +550,22 @@ For authentication export one of the following variables:
     # For SAS
     $ export AZURE_ACCOUNT_SAS=<SAS_TOKEN>
 
-For authentication using ``az login`` set the resource group name and ensure the user has
-the minimum permissions of the role assignment ``Storage Blob Data Contributor`` on Azure RBAC.
+For authentication using ``az login`` ensure the user has
+the minimum permissions of the role assignment ``Storage Blob Data Contributor`` on Azure RBAC
+for the storage account.
 
 .. code-block:: console
 
-    $ export AZURE_RESOURCE_GROUP=<RESOURCE_GROUP_NAME>
     $ az login
 
-Alternatively, if run on Azure, restic will automatically uses service accounts configured
+Alternatively, if run on Azure, restic will automatically use service accounts configured
 via the standard environment variables or Workload / Managed Identities.
+
+To enforce the use of the Azure CLI credential when other credentials are present, set the following environment variable:
+
+.. code-block:: console
+
+    $ export AZURE_FORCE_CLI_CREDENTIAL=true
 
 Restic will by default use Azure's global domain ``core.windows.net`` as endpoint suffix.
 You can specify other suffixes as follows:
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -101,23 +101,21 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, "NewAccountSASClientFromEndpointToken")
 		}
-	} else if cfg.ForceCliCredential {
-		debug.Log(" - using AzureCLICredential")
-
-		cred, err := azidentity.NewAzureCLICredential(nil)
-		if err != nil {
-			return nil, errors.Wrap(err, "NewAzureCLICredential")
-		}
-
-		client, err = azContainer.NewClient(url, cred, opts)
-		if err != nil {
-			return nil, errors.Wrap(err, "NewClient")
-		}
 	} else {
-		debug.Log(" - using DefaultAzureCredential")
-		cred, err := azidentity.NewDefaultAzureCredential(nil)
-		if err != nil {
-			return nil, errors.Wrap(err, "NewDefaultAzureCredential")
+		var cred azcore.TokenCredential
+
+		if cfg.ForceCliCredential {
+			debug.Log(" - using AzureCLICredential")
+			cred, err = azidentity.NewAzureCLICredential(nil)
+			if err != nil {
+				return nil, errors.Wrap(err, "NewAzureCLICredential")
+			}
+		} else {
+			debug.Log(" - using DefaultAzureCredential")
+			cred, err = azidentity.NewDefaultAzureCredential(nil)
+			if err != nil {
+				return nil, errors.Wrap(err, "NewDefaultAzureCredential")
+			}
 		}
 
 		client, err = azContainer.NewClient(url, cred, opts)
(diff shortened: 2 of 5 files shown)


internal/backend/azure/azure.go
```diff
@@ -101,6 +101,18 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 		if err != nil {
 			return nil, errors.Wrap(err, "NewAccountSASClientFromEndpointToken")
 		}
+	} else if cfg.ForceCliCredential {
+		debug.Log(" - using AzureCLICredential")
+
+		cred, err := azidentity.NewAzureCLICredential(nil)
```
MichaelEischer, line 107: The only difference to the `- using DefaultAzureCredential` case is how the credentials are obtained. Please merge both cases.

doc/030_preparing_a_new_repo.rst
```diff
@@ -550,17 +550,23 @@ For authentication export one of the following variables:
...
 
-For authentication using ``az login`` set the resource group name and ensure the user has
-the minimum permissions of the role assignment ``Storage Blob Data Contributor`` on Azure RBAC.
+For authentication using ``az login`` ensure the user has
+the minimum permissions of the role assignment ``Storage Blob Data Contributor`` on Azure RBAC
+for the storage account.
 
 .. code-block:: console
 
-    $ export AZURE_RESOURCE_GROUP=<RESOURCE_GROUP_NAME>
```
MichaelEischer, line 558: What's the reason for this change?

internal/backend/azure/config.go
```diff
@@ -73,6 +74,8 @@ func (cfg *Config) ApplyEnvironment(prefix string) {
 		cfg.AccountSAS = options.NewSecretString(os.Getenv(prefix + "AZURE_ACCOUNT_SAS"))
 	}
 
+	cfg.ForceCliCredential = os.Getenv(prefix+"AZURE_FORCE_CLI_CREDENTIAL") == "true"
```
MichaelEischer, line 77: Please use `strconv.ParseBool` for parsing the environment variable. I'd prefer to set `ForceCliCredential` only if the environment variable evaluates to true. (Basically the boolean analogue to the string parameters are currently handled).

doc/030_preparing_a_new_repo.rst
```diff
@@ -550,17 +550,23 @@ For authentication export one of the following variables:
...
 
-Alternatively, if run on Azure, restic will automatically uses service accounts configured
+Alternatively, if run on Azure, restic will automatically use service accounts configured
 via the standard environment variables or Workload / Managed Identities.
 
+To enforce the use of the Azure CLI crendential when other credentials are present, set the following environment variable:
+
+.. code-block:: console
+
+    $ export AZURE_FORCE_CLI_CREDENTIAL=true
```
MichaelEischer, line 568: There's an environment variable list in `040_backup.rst` please also extend that.

doc/030_preparing_a_new_repo.rst
```diff
@@ -550,17 +550,23 @@ For authentication export one of the following variables:
...
 .. code-block:: console
 
-    $ export AZURE_RESOURCE_GROUP=<RESOURCE_GROUP_NAME>
     $ az login
 
-Alternatively, if run on Azure, restic will automatically uses service accounts configured
+Alternatively, if run on Azure, restic will automatically use service accounts configured
 via the standard environment variables or Workload / Managed Identities.
 
+To enforce the use of the Azure CLI crendential when other credentials are present, set the following environment variable:
```
MichaelEischer, line 564: Typo `crendential`

MichaelEischer (changes requested):
Adding the environment variable seems ok, I didn't find any easy workarounds to ignore the managed identity.

I have a few comments though, see below.

internal/backend/azure/azure.go
letmaik: I don't have much experience in Go yet, I was trying this but it fails with "cred declared and not used". Any idea?

```go
	} else {
		var cred *azcore.TokenCredential
		if cfg.ForceCliCredential {
			debug.Log(" - using AzureCLICredential")
			cred, err := azidentity.NewAzureCLICredential(nil)
			if err != nil {
				return nil, errors.Wrap(err, "NewAzureCLICredential")
			}
		} else {
			debug.Log(" - using DefaultAzureCredential")
			cred, err := azidentity.NewDefaultAzureCredential(nil)
			if err != nil {
				return nil, errors.Wrap(err, "NewDefaultAzureCredential")
			}
		}

		client, err = azContainer.NewClient(url, *cred, opts)
		if err != nil {
			return nil, errors.Wrap(err, "NewClient")
		}
	}
```

doc/030_preparing_a_new_repo.rst
letmaik: This environment variable is unused and has no effect.

doc/030_preparing_a_new_repo.rst
letmaik: Done.

internal/backend/azure/config.go
letmaik: Done.

internal/backend/azure/azure.go
MichaelEischer: I've made the necessary change.

 `:=` declares a new variable within the current scope. That is `cred` in `var cred ...` and `cred, err := ...` are different variables.

MichaelEischer (approved):
LGTM. I've rebased the PR and added a commit to deduplicate the CLI and default credentials case.

letmaik:
@MichaelEischer Any chance to cut a new release with this? Thanks! :)

letmaik:
@MichaelEischer Sorry to be a pain, do you think 0.17.0 will be out by end of next week? We're rolling out an internal policy that will prevent key-based access to storage accounts and without this PR we won't be able to use restic anymore for certain Azure VM scenarios.

MichaelEischer:
> Sorry to be a pain, do you think 0.17.0 will be out by end of next week?

Definitely not before end of next week, a release at the end of next week might be possible, but is somewhat unlikely. There are a few more changes necessary for the `restore` command before we can cut a new release and I didn't make as much progress as I hoped for.

rawtaz:
@letmaik Can you please use the latest master build instead, temporarily until the new version is out? It's at https://beta.restic.net/latest/ .

MichaelEischer:
I've noticed that cherry-picking the PR for a 0.16.5 patch release works. So we can include the change in a release earlier than expected.

MichaelEischer:
@letmaik 0.16.5 has been released: https://github.com/restic/restic/releases/tag/v0.16.5

letmaik:
@MichaelEischer Thanks a lot, that's extremely helpful! :)

Issue #4774, opened 2024-04-20
Feature Request: Re-Read on checksum-fail (check --data-read)


demlak:
Output of `restic version`
--------------------------
`restic 0.16.4 compiled with go1.21.6 on windows/amd64`

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Re-Read a single file on `check --data-read`, when an error is detected.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

There are several possible Issues which can lead to a faulty result on hashing files.

For example: the chance of having bit-flip on non-ECC-RAM grows with the number of pack files.

My Case:
My 900gb Repo with nearly 200k packs is stored on a storage share at my hoster.. for redundancy i rsync from time to time localy to an old PC with non-ECC-RAM. After that, i start `check --read-data` localy for obvious reasons =)
Sometimes i get an error about wrong checksum of a blob. 

But: 
- Manualy checking the file for example via `sha256sum` shows no problem. 
- So i start again `restic check --read-data` and the check exits with an error on another file.. 
- or it exits without any problem. 

This shows me, there is no problem on the disk but somewhere else.

I think, it would increase the comfort, when restic re-reads files, where it got an error.. maybe 1 or 2 times.. just for getting deeper into the problem.
If the error still exists.. 
- and is the same error, it is clear, that the file is corrupt on the disk.
- if it is a different error on the same file at different re-reads, it is also clear, there is a problem (but well.. i don't know which one =) )
- if the error is gone on a second scan, everything seems to be good and the whole `--read-data` can continue

on small repos, this will not increase comfort.. but on large-ones like mine, which needs about 4 hours to be scanned, it would increase comfort much.

One could now say: change your faulty hardware! yes.. but.. if this is a non-ECC-RAM problem, it probably affects many others, too. non-ECC-RAM is not that rare on the planet =)
**And also it would fasten the error-analysis.**

On-top, maybe it is a nice idea to have controll over this feature via commandline.. for example `disable`, `counts of re-reads`, `exit/continue on error`, etc..

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

yeah.. restics helps me several times a week! beside this, it helped me today to get in touch with you, cheers! <3

olhotak:
[...]

PR #4800, opened 2024-05-09
Retry loading of corrupted data from backend / cache

MichaelEischer:
The PR cleans up how data is loaded from the backend and ensures that all places include a fallback to handle transient errors or a corrupt file in the cache.

The main component is `repository.LoadRaw` which replaces `backend.LoadAll`. In addition, almost all direct `backend.Load()` uses outside the backend/cache code have been replaced with calls to `LoadRaw` to benefit from its error handling. The only exceptions are `LoadBlob`, `ListPack`, `checkPack` and `LoadBlobsFromPack`. The first three include custom retry code to handle transient errors and the `LoadBlobsFromPack` falls back to `LoadBlob`.

Direct uses of `Backend.Load()` outside the repository package are discouraged now and will eventually be no longer possible.

The retry strategy from transient errors in general consists of explicitly forgetting the damaged file from the cache and retrying the operation once. The retries no longer happens as part of the retries performed by the RetryBackend but instead are handled separately. Thus, retries triggered by the `RetryBackend` now exclusively relate to backend / download errors. For example a pack file with corrupt blobs will now be downloaded at most twice.

In conjunction with the just mentioned retry strategy, the cache no longer automatically removes a file if the processing callback failed. In particular, since #4605 the automatic removal only worked in few cases. Now, the cached broken file is passed to `cache.Forget(h)` explicitly by `LoadRaw` etc. `cache.Forget(h)` only deletes a file at most once during the runtime of restic. This ensures that if a file is corrupted at the backend, then it will only be downloaded and forgotten at most once.

The cached file in the cache is now considered authoritative. Requesting a file section that is out of bounds of the cached file, now immediately yields an error. The retry strategy of `LoadRaw` etc. then forgets broken files. The main benefit of the new behavior is that out of bounds accesses of truncated files now fail immediately if the file is cacheable, instead of yielding endless retries.

The changes are complemented with a few cleanups to reduce code duplication:
- LoadUnpacked now uses LoadRaw
- LoadBlob uses the PackedBlobIterator
- And several minor cleanups, like removing the duplicate blob loading fallback from repair packs.

Prerequisite for #4784, see #4627.
Related to #4774.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #4807, opened 2024-05-17
Back up and restore Extended Attributes on Windows NTFS

aneesh-n:
Restic did not back up Extended Attributes of files on Windows.
Restic now backs up and restores Extended Attributes for Windows NTFS files.
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_file_full_ea_information
An extended attribute is a piece of application-specific metadata that an application can associate with a file that is not part of the file's data. In addition to the built-in attributes of a file, such as creation and modification times, applications can add non-file system attributes, such as the author's name and a description of the file content.

Follow-up to https://github.com/restic/restic/pull/4611

Checklist
---------

### DIFF
--- a/internal/fs/ea_windows.go
+++ b/internal/fs/ea_windows.go
@@ -1,0 +1,284 @@
+//go:build windows
+// +build windows
+
+package fs
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+// The code below was adapted from https://github.com/microsoft/go-winio under MIT license.
+
+// The MIT License (MIT)
+
+// Copyright (c) 2015 Microsoft
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// The code below was copied over from https://github.com/microsoft/go-winio/blob/main/ea.go under MIT license.
+
+type fileFullEaInformation struct {
+	NextEntryOffset uint32
+	Flags           uint8
+	NameLength      uint8
+	ValueLength     uint16
+}
+
+var (
+	fileFullEaInformationSize = binary.Size(&fileFullEaInformation{})
+
+	errInvalidEaBuffer = errors.New("invalid extended attribute buffer")
+	errEaNameTooLarge  = errors.New("extended attribute name too large")
+	errEaValueTooLarge = errors.New("extended attribute value too large")
+)
+
+// ExtendedAttribute represents a single Windows EA.
+type ExtendedAttribute struct {
+	Name  string
+	Value []byte
+	Flags uint8
+}
+
+func parseEa(b []byte) (ea ExtendedAttribute, nb []byte, err error) {
+	var info fileFullEaInformation
+	err = binary.Read(bytes.NewReader(b), binary.LittleEndian, &info)
+	if err != nil {
+		err = errInvalidEaBuffer
+		return ea, nb, err
+	}
+
+	nameOffset := fileFullEaInformationSize
+	nameLen := int(info.NameLength)
+	valueOffset := nameOffset + int(info.NameLength) + 1
+	valueLen := int(info.ValueLength)
+	nextOffset := int(info.NextEntryOffset)
+	if valueLen+valueOffset > len(b) || nextOffset < 0 || nextOffset > len(b) {
+		err = errInvalidEaBuffer
+		return ea, nb, err
+	}
+
+	ea.Name = string(b[nameOffset : nameOffset+nameLen])
+	ea.Value = b[valueOffset : valueOffset+valueLen]
+	ea.Flags = info.Flags
+	if info.NextEntryOffset != 0 {
+		nb = b[info.NextEntryOffset:]
+	}
+	return ea, nb, err
+}
+
+// DecodeExtendedAttributes decodes a list of EAs from a FILE_FULL_EA_INFORMATION
+// buffer retrieved from BackupRead, ZwQueryEaFile, etc.
+func DecodeExtendedAttributes(b []byte) (eas []ExtendedAttribute, err error) {
+	for len(b) != 0 {
+		ea, nb, err := parseEa(b)
+		if err != nil {
+			return nil, err
+		}
+
+		eas = append(eas, ea)
+		b = nb
+	}
+	return eas, err
+}
+
+func writeEa(buf *bytes.Buffer, ea *ExtendedAttribute, last bool) error {
+	if int(uint8(len(ea.Name))) != len(ea.Name) {
+		return errEaNameTooLarge
+	}
+	if int(uint16(len(ea.Value))) != len(ea.Value) {
+		return errEaValueTooLarge
+	}
+	entrySize := uint32(fileFullEaInformationSize + len(ea.Name) + 1 + len(ea.Value))
+	withPadding := (entrySize + 3) &^ 3
+	nextOffset := uint32(0)
+	if !last {
+		nextOffset = withPadding
+	}
+	info := fileFullEaInformation{
+		NextEntryOffset: nextOffset,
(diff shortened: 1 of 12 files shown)


aneesh-n:
I thought this would be the smallest of the PRs, but the LOC bloat is mainly due to the multiple copyright comments and the test cases.
The last PR after this is the ADS support and that might be a big one, I'm not sure how to split it.

internal/restic/node_test.go
```diff
@@ -205,8 +206,18 @@ func TestNodeRestoreAt(t *testing.T) {
...
-					// restic does not support xattrs on windows
-					return
+					// In windows extended attributes are case insensitive and windows returns
+					// the extended attributes in UPPER case.
+					// Update the tests to use UPPER case xattr names for windows.
+					extAttrArr := test.ExtendedAttributes
+					// Iterate through the array using pointers
+					for i := 0; i < len(extAttrArr); i++ {
+						// Get the pointer to the current element
+						namePtr := &extAttrArr[i].Name
```
MichaelEischer, line 216: I'd prefer the following (not really sure why, but it's better at hiding the pointer ^^ ):
```
entry := &extAttrArr[i]
entry.Name = strings.ToUpper(entry.Name)
```
or simply
`extAttrArr[i].Name = strings.ToUpper(extAttrArr[i].Name)`

internal/restic/node_windows.go
```diff
@@ -70,26 +70,109 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
 }
 
-// Getxattr retrieves extended attribute data associated with path.
-func Getxattr(path, name string) ([]byte, error) {
-	return nil, nil
+// restore extended attributes for windows
+func (node Node) restoreExtendedAttributes(path string) (err error) {
+	eas := []fs.ExtendedAttribute{}
+	for _, attr := range node.ExtendedAttributes {
+		extr := new(fs.ExtendedAttribute)
```
MichaelEischer, line 77: That allocation is pointless. Please use the following:
```go
eas = append(eas, fs.ExtendedAttribute{
    Name: attr.Name,
    Value: attr.Value,
})

File: internal/restic/node_windows.go [hunk 3]
```diff
@@ -70,26 +70,109 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
+	if err != nil {
+		debug.Log("open file failed for path: %s : %v", path, err)
+		return err
+	} else if len(extAtts) == 0 {
+		return nil
+	}
+
+	//Fill the ExtendedAttributes in the node using the name/value pairs in the windows EA
+	for _, attr := range extAtts {
+		if err != nil {
```
MichaelEischer, line 132: `err` is always nil at this point

internal/restic/node_windows.go
```diff
@@ -70,26 +70,109 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
 	return nil
 }
 
+// restoreExtendedAttributes handles restore of the Windows Extended Attributes to the specified path.
+// The Windows API requires setting of all the Extended Attributes in one call.
+func restoreExtendedAttributes(nodeType, path string, eas []fs.ExtendedAttribute) (err error) {
+	var fileHandle windows.Handle
+	switch nodeType {
+	case "file":
+		utf16Path := windows.StringToUTF16Ptr(path)
```
MichaelEischer, line 152: Why does `fillExtendedAttributes` check for `if strings.HasSuffix(filepath.Clean(path), '\') {`, but not this function? Please extract the helper code to open a file handle, it seems to be identical between `fillExtendedAttributes` and `restoreExtendedAttributes`.

internal/restic/node_windows.go
```diff
@@ -70,26 +70,109 @@ func (node Node) restoreSymlinkTimestamps(path string, utimes [2]syscall.Timespe
...
+		return nil
+	}
+	defer func() {
+		err := windows.CloseHandle(fileHandle)
+		if err != nil {
+			debug.Log("Error closing file handle for %s: %v\n", path, err)
+		}
+	}()
+	if err != nil {
+		err = errors.Errorf("open file failed for path %v, with: %v:\n", path, err)
```
MichaelEischer, line 169: This should immediately return the error.

internal/restic/node_windows_test.go
```diff
@@ -265,3 +266,68 @@ func TestNewGenericAttributeType(t *testing.T) {
...
+		} else if node.Type == "dir" {
+			handle, err = windows.CreateFile(utf16Path, windows.FILE_READ_EA, 0, nil, windows.OPEN_EXISTING, windows.FILE_ATTRIBUTE_NORMAL|windows.FILE_FLAG_BACKUP_SEMANTICS, 0)
+		}
+		test.OK(t, errors.Wrapf(err, "Error opening file/directory for: %s", testPath))
+		defer func() {
+			err := windows.Close(handle)
+			test.OK(t, errors.Wrapf(err, "Error closing file for: %s", testPath))
+		}()
+
+		if len(node.ExtendedAttributes) > 0 {
```
MichaelEischer, line 314: When would it be correct for `node` to have no `ExtendedAtrributes`? Both test nodes have extended attributes.

MichaelEischer (changes requested):
That was surprisingly easy to review; there's just a shocking amount of boilerplate code (not that I see how we could change that). I've added a few minor comments. The most interesting one is probably whether it would make sense to import go-winio instead of copying some functions.

internal/fs/ea_windows.go
```diff
@@ -0,0 +1,284 @@
...
+// under MIT license.
+
+var (
+	procNtQueryEaFile = modntdll.NewProc("NtQueryEaFile")
+	procNtSetEaFile   = modntdll.NewProc("NtSetEaFile")
+)
+
+const (
+	// noExtendedAttribsStatus is a constant value which indicates no extended attributes were found
+	noExtendedAttribsStatus = -1073741742
```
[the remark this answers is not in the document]
aneesh-n: Thanks for finding the reference. Renaming it.

internal/fs/ea_windows_test.go
```diff
@@ -0,0 +1,259 @@
...
+		t.Fatalf("get EA for file failed: %s", err)
+	}
+
+	if !reflect.DeepEqual(readEAs, testEAs) {
+		t.Logf("expected: %+v, found: %+v\n", testEAs, readEAs)
+		t.Fatalf("EAs read from testfile don't match")
+	}
+}
+
+func TestSetGetFolderEA(t *testing.T) {
```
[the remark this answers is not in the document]
aneesh-n: Yes, this was newly added. Adding comment.

internal/restic/node_test.go
aneesh-n: Updated thanks

internal/restic/node_windows.go
aneesh-n: Fixed

internal/restic/node_windows.go
aneesh-n: Fixed

internal/restic/node_windows.go
aneesh-n: Removing strings.HasSuffix(filepath.Clean(path), `\`) check, it is not needed here.
Created common helper function.

internal/restic/node_windows_test.go
aneesh-n: Removed, it is not needed.

internal/fs/ea_windows.go
```diff
@@ -0,0 +1,284 @@
...
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// The code below was copied over from https://github.com/microsoft/go-winio/blob/main/ea.go under MIT license.
```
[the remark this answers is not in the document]
aneesh-n: The important code that is being used in our _ea_windows.go_ was taken from a PR of go-winio - https://github.com/microsoft/go-winio/pull/229

This PR was never merged into go-winio, because they no longer needed those functions. They are using `PAXHeaders` for some tar related functionality and only use from encoding/decoding functions for EA in their _ea.go_.

Also, the code that we need like `GetFileEA` uses a private struct `ioStatusBlock` and private function `ntStatus` which are defined in go-winio. Hence, even if we import go-winio we would have to redefine that code in our _ea.go_.
The entire test case is from that unmerged PR as well.

Still, by importing go-winio, we will be able to use `winio.EncodeExtendedAttributes` and `winio.DecodeExtendedAttributes`.

_ea_windows.go_ would reduce from **285 lines to 140 lines**.


In _sd_windows.go_ we would be able to directly use `winio.EnableProcessPrivileges` and all the privilege related functions would be handled in go-winio.

_sd_windows_ would reduce from **440 lines to 189 lines**.  

Also, I see that by adding the go-winio module and making all the necessary changes, the windows binary size increases by 10 KB only.

So, it might make sense to use go-winio. Would you like me to make the changes?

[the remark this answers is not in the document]
MichaelEischer: How about the following: let's merge this PR with the duplicated code and import go-winio in a follow-up PR. My plan is to release restic 0.17.0 in a few weeks and after that is done, I'll bump the go requirement to 1.21. Then we could directly start with using 0.6.2

MichaelEischer (approved):
LGTM. Thanks!

I hope that we can release restic 0.17.0 in the next weeks, so the alternate data streams support might have to wait for the next release. Is there anything related to your PRs that has to be done before releasing 0.17.0?

As a heads-up, I'll move a lot of the `node*.go` to the `fs` package some time after restic 0.17.0 is released.

aneesh-n:
Sounds like plan, I'm good with these PRs for now.
Thank you!

PR #4698, opened 2024-02-12
Allow empty password

doak:
What does this PR change? What problem does it solve?
-----------------------------------------------------

It adds basic support for empty passwords and may be a solution for #4326.
This is just a PoC, I haven't ever written a single Go line before.

Subcommands `init` and `key add|passwd` now support the flag `--insecure-password` which disables the check for an empty password. `restic` warns about it, though.
Support for `RESTIC_PASSWORD_FILE` and coresponding CLI option has been added as well. You may user either an empty file or just `/dev/null`.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

Not sure if it really fulfils (all of) the aims of #4326, but it may be a starting point. 

Checklist
---------

Nothing relevant is checked yet. This is just a prof of concept (for now). If there is a future for this, I am happy to do it right.
It have been tested only manually. Works on my machine™. ;)

- [X] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [X] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [X] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits). (no, it is better ;))
- [ ] I'm done! This pull request is ready for review.

PR #4808, opened 2024-05-18
Implement `--insecure-no-password` option.

### LINKED ISSUE #1786: Allow empty passwords
## Output of `restic version`

restic 0.8.3
compiled with go1.10 on linux/amd64

echo | restic init --repo /srv/restic/ 
Fatal: an empty password is not a password

filesystem

## Expected behavior

Restic should allow empty passwords.

## Actual behavior

Does not allow empty password.

echo | restic init --repo /srv/restic/

it is a design decision.

Remove the check for emtpy passwords.

Yes, it is a great product and I love that it helps so many people to do backups.

### LINKED ISSUE #4326: Backup without password
I briefly read this issue: https://github.com/restic/restic/issues/1786

It seems there is some confusion about the need for backups without encryption.  First of all, I use LUKS encryption on my drives as a first line of defense and it's implemented on all of my drives. For this reason alone, a password is less useful for a drive already encrypted using LUKS.

The particular drive I want to backup to is not only LUKS encrypted but also hardware encrypted. It's a drive that has a keypad. Call me paranoid, but I like to encrypt using both hardware and software solutions for my most sensitive data.

I also use CryFS and VeraCrypt (I'm moving away from VeraCrypt). I have some directories further encrypted using these tools and again, there is less of a need to use a password for backups of these directories since they are already encrypted.

I think it's bad practice to require a password because restic is first and foremost a backup tool, and users should be able to backup without a password if they want to. It's really that simple, freedom. Forcing users to create a password for a repository that doesn't actually need one causes users to create passwords that they may not remember, something like "password123" which, from my experience, I tend to forget placeholder passwords more often than I forget more secure ones. A password manager is not a solution here. I don't want to have to use a password manager to store a weak password for something that doesn't need to be encrypted in the first place. It's inconvenient.

As for the problem regarding users inadvertently not using a password for data that needs one - this all comes down to how you implement it. There's a very simple fix for that. Add a warning saying "No password set. Are you sure?", and tell users that not setting a password is an option.

Forcing users to set a password will inevitably lead users to create weak passwords where no password is needed, and they may forget it because they were sure they wouldn't. It's happened to me more times than I care to admit.

I forgot to mention the whole reason why I posted this issue. The drive I'm trying to backup...is the drive that holds my password manager. In order to save a backup of my password manager, I need to create a new password database, with its own password, on a complete different drive, to save the restic password which decrypts the backup, which itself is encrypted anyway. Not sure what I'm gonna do about that except use a different program.

I'll also quickly mention the fact that some users may not trust restic's cryptography, and should have the option to use an encryption tool separately without having to rely on restic's encryption (and of course, not have to double-up on encryption when they don't want to).

MichaelEischer:
Implement https://github.com/restic/restic/issues/4326#issuecomment-1950297875 . Different from that comment, 
there are two additional options `--from-insecure-no-password`  and  `--new-insecure-no-password`.  `--from-insecure-no-password`  is used for commands that require specifying a source repository, and    `--new-insecure-no-password` for the `key add` and `key passwd`  commands.

Those options are required to allow copying from a repository without password to one with a password. Or the add/set an new empty password.

Fixes https://github.com/restic/restic/issues/1786
Fixes https://github.com/restic/restic/issues/4326
Replaces https://github.com/restic/restic/pull/4698/files

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1786
+++ b/changelog/unreleased/issue-1786
@@ -1,0 +1,19 @@
+Enhancement: Support repositories with empty password
+
+Restic refused to create or operate on repositories with an emtpy password.
+Using the new option `--insecure-no-password` it is now possible to disable
+this check. Restic will not prompt for a password when using this option.
+For security reasons, the option must always be specified when operating on
+repositories with an empty password.
+
+Specifying `--insecure-no-password` while also passing a password to restic
+via a CLI option or via environment variable results in an error.
+
+The `init` and `copy` command also support the option `--from-insecure-no-password`
+which applies to the source repository. The `key add` and `key passwd` comands
+include the `--new-insecure-no-password` option to add or set an emtpy password.
+
+https://github.com/restic/restic/issues/1786
+https://github.com/restic/restic/issues/4326
+https://github.com/restic/restic/pull/4698
+https://github.com/restic/restic/pull/4808
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -257,7 +257,7 @@ func readFilenamesRaw(r io.Reader) (names []string, err error) {
 
 // Check returns an error when an invalid combination of options was set.
 func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
-	if gopts.password == "" {
+	if gopts.password == "" && !gopts.InsecureNoPassword {
 		if opts.Stdin {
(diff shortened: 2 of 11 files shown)


allmazz:
nice idea, I hope it will be merged soon!

MichaelEischer (commented):
LGTM

PR #4811, opened 2024-05-19
Restore flags to include from file and exclude from file

### LINKED ISSUE #4781: Restic restore include-file/exclude-file switches
Output of `restic version`
--------------------------

restic 0.16.4 compiled with go1.21.6 on darwin/amd64

It would be helpful if the include-file, --iinclude-file, --exclude-file, and --iexclude-file switches existed for the "restore" function.

I recently needed to restore a bunch of different files that got deleted.  About 1600 of them, in random places.  When I did "restic diff" I could easily get a list of all the files that were deleted - but I don't have an easy way to feed that into Restic to restore just those files.

My options are to restore a large swath of files to get a few, or to run restic restore on each file individually - which seems impractical with that many files.

Restic rules!  ✌️

konidev20:
Added `--exclude-file`, `--iexclude-file`, `include-file` and `iinclude-file` to the restore command. 

Closes #4781 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -178,12 +178,12 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 			matched, childMayMatch := includeFn(item)
 			selectedForRestore = selectedForRestore || matched
 			childMayBeSelected = childMayBeSelected || childMayMatch
-			childMayBeSelected = childMayBeSelected && node.Type == "dir"
 
-			if selectedForRestore || childMayBeSelected {
+			if selectedForRestore && childMayBeSelected {
 				break
 			}
 		}
+		childMayBeSelected = childMayBeSelected && node.Type == "dir"
 
 		return selectedForRestore, childMayBeSelected
 	}
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -2,12 +2,10 @@ package main
 
 import (
 	"context"
-	"strings"
 	"time"
 
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
-	"github.com/restic/restic/internal/filter"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/restorer"
 	"github.com/restic/restic/internal/ui"
@@ -45,15 +43,9 @@ Exit status is 0 if the command was successful, and non-zero if there was any er
 
 // RestoreOptions collects all options for the restore command.
 type RestoreOptions struct {
-	Exclude                 []string
-	ExcludeFiles            []string
-	InsensitiveExclude      []string
-	InsensitiveExcludeFiles []string
-	Include                 []string
-	IncludeFiles            []string
-	InsensitiveInclude      []string
-	InsensitiveIncludeFiles []string
-	Target                  string
+	excludePatternOptions
+	includePatternOptions
+	Target string
 	restic.SnapshotFilter
 	Sparse bool
 	Verify bool
@@ -65,15 +57,10 @@ func init() {
 	cmdRoot.AddCommand(cmdRestore)
 
 	flags := cmdRestore.Flags()
-	flags.StringArrayVarP(&restoreOptions.Exclude, "exclude", "e", nil, "exclude a `pattern` (can be specified multiple times)")
-	flags.StringArrayVar(&restoreOptions.InsensitiveExclude, "iexclude", nil, "same as --exclude but ignores the casing of `pattern`")
-	flags.StringArrayVarP(&restoreOptions.Include, "include", "i", nil, "include a `pattern`, exclude everything else (can be specified multiple times)")
-	flags.StringArrayVar(&restoreOptions.InsensitiveInclude, "iinclude", nil, "same as --include but ignores the casing of `pattern`")
 	flags.StringVarP(&restoreOptions.Target, "target", "t", "", "directory to extract data to")
-	flags.StringArrayVar(&restoreOptions.ExcludeFiles, "exclude-file", nil, "read exclude patterns from a `file` (can be specified multiple times)")
-	flags.StringArrayVar(&restoreOptions.InsensitiveExcludeFiles, "iexclude-file", nil, "same as --exclude-file but ignores casing of `file`names in patterns")
-	flags.StringArrayVar(&restoreOptions.IncludeFiles, "include-file", nil, "read include patterns from a `file` (can be specified multiple times)")
-	flags.StringArrayVar(&restoreOptions.InsensitiveIncludeFiles, "iinclude-file", nil, "same as --include-file but ignores casing of `file`names in patterns")
+
+	initExcludePatternOptions(flags, &restoreOptions.excludePatternOptions)
+	initIncludePatternOptions(flags, &restoreOptions.includePatternOptions)
 
 	initSingleSnapshotFilter(flags, &restoreOptions.SnapshotFilter)
 	flags.BoolVar(&restoreOptions.Sparse, "sparse", false, "restore files as sparse")
@@ -83,38 +70,8 @@ func init() {
 func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 	term *termstatus.Terminal, args []string) error {
 
-	hasExcludes := len(opts.Exclude) > 0 || len(opts.InsensitiveExclude) > 0
-	hasIncludes := len(opts.Include) > 0 || len(opts.InsensitiveInclude) > 0
-
-	// Validate provided patterns
-	if len(opts.Exclude) > 0 {
-		if err := filter.ValidatePatterns(opts.Exclude); err != nil {
-			return errors.Fatalf("--exclude: %s", err)
-		}
-	}
-	if len(opts.InsensitiveExclude) > 0 {
-		if err := filter.ValidatePatterns(opts.InsensitiveExclude); err != nil {
-			return errors.Fatalf("--iexclude: %s", err)
-		}
-	}
-	if len(opts.Include) > 0 {
-		if err := filter.ValidatePatterns(opts.Include); err != nil {
-			return errors.Fatalf("--include: %s", err)
-		}
-	}
-	if len(opts.InsensitiveInclude) > 0 {
-		if err := filter.ValidatePatterns(opts.InsensitiveInclude); err != nil {
-			return errors.Fatalf("--iinclude: %s", err)
-		}
-	}
-
-	for i, str := range opts.InsensitiveExclude {
-		opts.InsensitiveExclude[i] = strings.ToLower(str)
-	}
-
-	for i, str := range opts.InsensitiveInclude {
-		opts.InsensitiveInclude[i] = strings.ToLower(str)
-	}
+	hasExcludes := len(opts.Excludes) > 0 || len(opts.InsensitiveExcludes) > 0
+	hasIncludes := len(opts.Includes) > 0 || len(opts.InsensitiveIncludes) > 0
 
 	switch {
 	case len(args) == 0:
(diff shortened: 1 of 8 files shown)


konidev20:
@MichaelEischer please review this and let me know if I am on the right track.

cmd/restic/cmd_restore.go
```diff
@@ -176,6 +184,31 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 
 	excludePatterns := filter.ParsePatterns(opts.Exclude)
 	insensitiveExcludePatterns := filter.ParsePatterns(opts.InsensitiveExclude)
+
+	if len(opts.ExcludeFiles) > 0 {
```
MichaelEischer, line 188: At least the exclude patterns look like we could share the implementation used by the `backup` command. Please remove the custom exclude pattern implementation from restore and use `excludePatternOptions` instead. Unless I've missed something, the returned `RejectByNameFunc`s should suitable for use in selectExcludeFilter.

cmd/restic/cmd_restore.go
```diff
@@ -199,6 +232,31 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 
 	includePatterns := filter.ParsePatterns(opts.Include)
 	insensitiveIncludePatterns := filter.ParsePatterns(opts.InsensitiveInclude)
+
+	if len(opts.IncludeFiles) > 0 {
```
MichaelEischer, line 236: Please create a `includePatternOptions` struct similar to `excludePatternOptions`. The `includeByNameFunc`s must be somewhat different in that it has to use `filter.ListWithChild(includePatterns, item)` and return `matched, childMayMatch, err`.

MichaelEischer (commented):
I'd prefer a different design. `runRestore` is already too long, so I'd like to extract the include/exclude pattern parsing code. Judging from a quick look at the exclude code used by the backup command (see `cmd/restic/exclude.go`), it should be possible to directly reuse the code for the exclude options. And it should be possible to adapt that pattern to add new include options. (Feel free to ask if you need more details to understand what I have in mind)

konidev20:
> I'd prefer a different design. `runRestore` is already too long, so I'd like to extract the include/exclude pattern parsing code. Judging from a quick look at the exclude code used by the backup command (see `cmd/restic/exclude.go`), it should be possible to directly reuse the code for the exclude options. And it should be possible to adapt that pattern to add new include options. (Feel free to ask if you need more details to understand what I have in mind)

I wanted to propose this too. I will update this draft over the next weekend.

cmd/restic/cmd_restore.go
konidev20: This is done.

cmd/restic/cmd_restore.go
konidev20: This is done. I added a file called `include.go`

cmd/restic/cmd_restore.go
```diff
@@ -174,44 +149,28 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
...
-		if err != nil {
-			msg.E("error for exclude pattern: %v", err)
-		}
-
-		matchedInsensitive, err := filter.List(insensitiveExcludePatterns, strings.ToLower(item))
-		if err != nil {
-			msg.E("error for iexclude pattern: %v", err)
-		}
+		for _, rejectFn := range excludePatternFns {
+			matched := rejectFn(item)
```
MichaelEischer, line 154: That logic is partially broken. If any `rejectFn` returns true then the file must be excluded. Right now, the loop only ever looks at the first function in `excludePatternFns`

You have to combine the results as follows:
```go
matched := false
for _, rejectFn := range excludePatternFns {
    matched = matched || rejectFn(item)
}
```

cmd/restic/cmd_restore.go
```diff
@@ -174,44 +149,28 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
...
-		matched, childMayMatch, err := filter.ListWithChild(includePatterns, item)
-		if err != nil {
-			msg.E("error for include pattern: %v", err)
-		}
-
-		matchedInsensitive, childMayMatchInsensitive, err := filter.ListWithChild(insensitiveIncludePatterns, strings.ToLower(item))
-		if err != nil {
-			msg.E("error for iexclude pattern: %v", err)
+		for _, includeFn := range includePatternFns {
+			selectedForRestore, childMayBeSelected = includeFn(item)
```
MichaelEischer, line 170: Same here, except that only the last function has an effect on the outcome.

cmd/restic/cmd_restore.go
konidev20: My thought process was that I must exit the first match I get. Because it's an || operation. But I introduced an issue. I have resolved this.

cmd/restic/cmd_restore.go
konidev20: This is resolved. Please review.

cmd/restic/cmd_restore.go
konidev20: Let me know if that makes sense if we break after the first pattern it matches, it will be much faster. We don't have to do redundant pattern matching with other patterns.

MichaelEischer: A short-circuit semantic sounds reasonable. That also works for the include filter, although there the optimization is only possible if both `selectedForRestore` and `childMayBeSelected` are true within the loop.

cmd/restic/cmd_restore_integration_test.go
```diff
@@ -51,13 +51,64 @@ func testRunRestoreLatest(t testing.TB, gopts GlobalOptions, dir string, paths [
...
-		Target:  dir,
-		Include: includes,
+		Target: dir,
 	}
+	opts.Includes = includes
 
 	rtest.OK(t, testRunRestoreAssumeFailure(snapshotID.String(), opts, gopts))
 }
 
+func TestRestoreIncludes(t *testing.T) {
```
MichaelEischer, line 61: Please extend the tests a bit to run restore at least once with a include or exclude file

MichaelEischer (changes requested):
I've found a few more issues. Currently `restore --exclude-file ...` does not work as `hasExcludes` is no longer correct. The same applies for include files.

cmd/restic/include.go
```diff
@@ -0,0 +1,100 @@
...
+	InsensitiveIncludes     []string
+	IncludeFiles            []string
+	InsensitiveIncludeFiles []string
+}
+
+func initIncludePatternOptions(f *pflag.FlagSet, opts *includePatternOptions) {
+	f.StringArrayVarP(&opts.Includes, "include", "i", nil, "include a `pattern` (can be specified multiple times)")
+	f.StringArrayVar(&opts.InsensitiveIncludes, "iinclude", nil, "same as --include `pattern` but ignores the casing of filenames")
+	f.StringArrayVar(&opts.IncludeFiles, "include-file", nil, "read include patterns from a `file` (can be specified multiple times)")
+	f.StringArrayVar(&opts.InsensitiveIncludeFiles, "iinclude-file", nil, "same as --include-file but ignores casing of `file`names in patterns")
```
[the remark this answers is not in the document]
konidev20: Done.

cmd/restic/cmd_restore.go
```diff
@@ -75,37 +70,17 @@ func init() {
...
-		if err := filter.ValidatePatterns(opts.Include); err != nil {
-			return errors.Fatalf("--include: %s", err)
-		}
-	}
-	if len(opts.InsensitiveInclude) > 0 {
-		if err := filter.ValidatePatterns(opts.InsensitiveInclude); err != nil {
-			return errors.Fatalf("--iinclude: %s", err)
-		}
-	}
+	hasExcludes := len(opts.Excludes) > 0 || len(opts.InsensitiveExcludes) > 0
```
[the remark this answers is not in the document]
konidev20: I have added it as 

```
	hasExcludes := len(excludePatternFns) > 0
	hasIncludes := len(includePatternFns) > 0
```

Since, the length checks are done within the `CollectPatterns` functions. I have added a unit test as well to handle this exclusive check.

cmd/restic/integration_filter_pattern_test.go
```diff
@@ -70,30 +70,85 @@ func TestRestoreFailsWhenUsingInvalidPatterns(t *testing.T) {
...
+
+func TestRestoreFailsWhenUsingInvalidPatternsFromFile(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	testRunInit(t, env.gopts)
+
+	// Create an include file with some invalid patterns
+	includeFile := env.base + "/includefile"
+	fileErr := os.WriteFile(includeFile, []byte("*.go\n*[._]log[.-][0-9]\n!*[._]log[.-][0-9]"), 0644)
```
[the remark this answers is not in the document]
konidev20: Done.

cmd/restic/cmd_restore_integration_test.go
konidev20: Added Unit tests for both.

cmd/restic/cmd_restore.go
konidev20: @MichaelEischer I have made the changes, please check if it is correct.

konidev20:
Tests failing is unrelated.

cmd/restic/cmd_restore.go
```diff
@@ -171,9 +178,12 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 			matched, childMayMatch := includeFn(item)
 			selectedForRestore = selectedForRestore || matched
 			childMayBeSelected = childMayBeSelected || childMayMatch
-		}
+			childMayBeSelected = childMayBeSelected && node.Type == "dir"
 
-		childMayBeSelected = childMayBeSelected && node.Type == "dir"
+			if selectedForRestore || childMayBeSelected {
```
MichaelEischer, line 183: I think that is wrong. If the first `includeFn` returns `false, true` (matched, childMayMatch) for a directory, then the `if` would exit the loop right now. However, a second `includeFn` could still return `true, true`. 

To handle that case correctly, the loop should look like the following:
```Go
for _, includeFn := range includePatternFns {
			matched, childMayMatch := includeFn(item)
			selectedForRestore = selectedForRestore || matched
			childMayBeSelected = childMayBeSelected || childMayMatch

			if selectedForRestore && childMayBeSelected {
				break
			}
		}
		childMayBeSelected = childMayBeSelected && node.Type == "dir"
```

cmd/restic/include.go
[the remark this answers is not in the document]
MichaelEischer: Hmm, I just noticed that 
```
`file`names`
```
is actually correct. Without the backticks the help output changes from `--iinclude-file file` to `--iinclude-file string`. Please revert that change, sorry.

cmd/restic/cmd_restore_integration_test.go
```diff
@@ -58,6 +59,67 @@ func testRunRestoreIncludes(t testing.TB, gopts GlobalOptions, dir string, snaps
...
+	// Create test files and directories
+	for _, testFile := range testfiles {
+		fullPath := filepath.Join(env.testdata, testFile.path)
+		rtest.OK(t, os.MkdirAll(filepath.Dir(fullPath), 0755))
+		rtest.OK(t, appendRandomData(fullPath, testFile.size))
+	}
+
+	opts := BackupOptions{}
+	// Perform backup
+	testRunBackup(t, filepath.Dir(env.testdata), []string{filepath.Base(env.testdata)}, opts, env.gopts)
```
MichaelEischer, line 102: The include/exclude check happens before even opening the repository, thus please remove the code that create a snapshot and just pass a placeholder value to `testRunRestoreAssumeFailure`. To make sure that we still get the correct error, let the `rtest.Assert` check that the error message contains the substring `exclude and include patterns are mutually exclusive`.

cmd/restic/cmd_restore_integration_test.go
```diff
@@ -144,19 +223,38 @@ func TestRestoreFilter(t *testing.T) {
...
+	fileErr := os.WriteFile(patternsFile, []byte(strings.Join(excludePatterns, "\n")), 0644)
+	if fileErr != nil {
+		t.Fatalf("Could not write include file: %v", fileErr)
+	}
+
+	restoredir := filepath.Join(env.base, "restore-exclude-from-file")
+
+	testRunRestoreExcludesFromFile(t, env.gopts, restoredir, snapshotID, patternsFile)
+
+	testRestoreFileExclusions(t, env, excludePatterns)
```
MichaelEischer, line 257: This call performs multiple restores of its own, but does not use the exclude file.

MichaelEischer (commented):
I have a few more comment, after that we should be ready to go.

cmd/restic/include.go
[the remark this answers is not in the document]
konidev20: Reverted.

cmd/restic/cmd_restore.go
konidev20: Done.

cmd/restic/cmd_restore_integration_test.go
konidev20: Fixed, I made the logic simpler and similar to the inclusion tests.

MichaelEischer (approved):
LGTM. Thanks for the blazingly fast fixes!

PR #4812, opened 2024-05-19
Resumable prune & memory-efficient index rewrite

### LINKED ISSUE #3806: resumable prune
Output of `restic version`
--------------------------
restic 0.13.1 compiled with go1.18 on linux/amd64

pruning can take a long time, especially if there's a lot to repack and upload.
an interrupted prune is currently not resumable, as it removes unreferenced packs first.
in this case those packs contain the repacked blobs that should be kept after a prune.
they get deleted and then rebuild when pruning again.

if this deletion would not occur at the beginning, the already uploaded blobs would probably be detected and used instead of being rebuilt and reuploaded.

moving the deletion of replaced packs to the end would probably need some additional changes, because the same blob would exist in two packs. so maybe once it's clear which packs to remove (after the beginning scan) it should mark them as obsolete, and their blobs should not be referenced until they are deleted deferred.

see above.
prunes will be resumable.
as they may be long-running due to lots of reuploads you're forced to chose between periodical prunes blocking actual backups.
also connection-issues will have less impact.

maybe this will even get rid of the prune locks (by marking packs obsolete and removing them eventually)

i'm using restic every day and am happy with its reliability.
i'll be more happy, when it works on its own.

MichaelEischer:

The PR consists of a refactoring of the index and reworks the index rewrite to be much more memory efficient.

The refactoring makes a few noteworthy changes:
- The MasterIndex interface is now integrated into the Repository interface. Now users of a repository never interact directly with the index. This is one of the last missing major cleanups to allow eventually getting rid of the Repository interface.
- Several methods of index/repository are now private. In particular, there is no more `repo.Index().Save(...)` method.
- Loading the index is now done using `MasterIndex.Load(...)`
- The index no longer tracks and stores the `supersedes` field. It was never really used as using it correctly is nearly impossible.

Prune no longer uses `repository.DisableAutoIndexUpdate()`. This allows efficiently resuming interrupted prune runs.

To properly select packs from previous interrupted prune runs, the prune heuristics required a fix. Pack files created by interrupted prune runs, appear to consist only of duplicate blobs on the next run. This caused the previous heuristic to ignore those pack files. Now, a duplicate blob in a specific pack file is also selected if that pack file only contains duplicate blobs. This allows prune to select the already rewritten pack files.

`MasterIndex.Rewrite` implements a streaming rewrite of the index that excludes the given packs. For this it loads all index files from the repository and only modifies those that require changes. This will reduce the index churn when running prune. Rewrite does not require the in-memory index and thus can drop it to significantly reduce the memory usage.

However, `prune --unsafe-recovery` cannot use this strategy and requires a separate method to save the whole in-memory index. This is now handled using `MasterIndex.SaveFallback`.

Fixes https://github.com/restic/restic/issues/3806
The streaming rewrite is part of the roadmap for restic 0.17.0

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3806
+++ b/changelog/unreleased/issue-3806
@@ -1,0 +1,11 @@
+Enhancement: Make `prune` command resumable
+
+When `prune` was interrupted, it a latter `prune` run previously started repacking
+the pack files from the start as `prune` did not update the index while repacking.
+
+The `prune` command now supports resuming interrupted prune runs. The update
+of the repository index also has been optimized to use less memory and only
+rewrite parts of the index that have changed.
+
+https://github.com/restic/restic/issues/3806
+https://github.com/restic/restic/pull/4812
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -167,8 +167,7 @@ func runCat(ctx context.Context, gopts GlobalOptions, args []string) error {
 		}
 
 		for _, t := range []restic.BlobType{restic.DataBlob, restic.TreeBlob} {
-			bh := restic.BlobHandle{ID: id, Type: t}
-			if !repo.HasBlob(bh) {
+			if _, ok := repo.LookupBlobSize(id, t); !ok {
 				continue
 			}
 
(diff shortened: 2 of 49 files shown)


MichaelEischer (commented):
LGTM.

PR #4822, opened 2024-05-25
backup: refer to backed up data as backup source not target

### LINKED ISSUE #3262: documentation: "target"
restic 0.11.0
restic help backup shows inconsistent text. For the same thing, it uses

source (as in "source data")
FILE/DIR
target files/directories

This is why I always misunderstood backup --force - I thought it re-reads from the repository while it actually re-reads source data. I migrated from duplicity where TARGET points to the repository - so restic was a bit surprising for me.

My proposal is to at least get rid of the word "target" - a target points TO something, so getting data FROM a target feels strange.

You could even do

restic backup [flags] SOURCE [SOURCE] …

then define what SOURCE is and only use the word “source” for files/directories and "source data" for their content.

Problem is that the source code always uses the term “TARGET” so the source code should probably also be adapted.
See also https://forum.restic.net/t/target-seems-to-mean-the-source-data/3455?u=wrohdewald

MichaelEischer:
A backup reads from the source and writes it to the backup target (repository). The name "target" is kept within the source code as it's used there excessively.

Fixes https://github.com/restic/restic/issues/3262

Checklist
---------


MichaelEischer (commented):
LGTM.

Issue #4817, opened 2024-05-24
Make restore overwrite behavior configurable


MichaelEischer:
Restore should implement additional options to make it more flexible. This is a tracking issue that sums up several existing issues.

New overwrite behaviors

Skip existing:
- `--ignore-existing`
- does not overwrite already existing files

Always overwrite (default):
- `--delete` option to remove files not in the snapshot
- will also overwrite readonly files etc.
- if a file would overwrite a non-empty directory, refuse to do so unless `--delete` is also specified. That semantic is also used by rsync

Overwrite if newer:
- `--overwrite-if-newer` (rsync `--update`) replaces local files if the snapshot content is newer (by file timestamp!)
- allow implementing some kind of sync
- `--delete` option works the same way as for "Always overwrite"

Optimizations:
- restore only changed files: user can select to trust timestamp+size (default) or to let restic check the file contents of everything

Addresses Issues:
- #200 Feature request: "Restore newer/missing datasets" / "synchronize"
- #407 Make restore resumeable.
- #2348 Option for restic restore to delete local files not in snapshot
- #2662 Restore only changed files
- ~~#4575 Restore in place~~
- #4757 When restoring, cannot rewrite files that are readonly

MichaelEischer:
All done.

PR #4837, opened 2024-05-31
Make restore overwrite behavior configurable

### LINKED ISSUE #200: Feature request: "Restore newer/missing datasets" / "synchronize"

MichaelEischer:

The `restore` command now supports an `--overwrite` option to configure whether already existing files are overwritten. The default is `--overwrite always`, which overwrites existing files. `--overwrite if-newer` only restores files from the snapshot that are newer than the local state. And `--overwrite never` does not modify existing files.

The `--overwrite` option replaces the `--ignore-existing` and `--overwrite-if-newer` flags from #4817 .

Fixes #200.
See https://github.com/restic/restic/issues/4817 . 

Checklist
---------


MichaelEischer (commented):
LGTM. The `--overwrite` option name is not completely final yet. There's a chance it will be renamed to e.g. `--policy skip|update-if-newer|update-if-changed|update-always` suggested by @rawtaz.

PR #4838, opened 2024-05-31
restore: skip unchanged files and add `--overwrite if-changed` option

### LINKED ISSUE #407: Make restore resumeable.

### LINKED ISSUE #2662: Restore only changed files

MichaelEischer:
Based on https://github.com/restic/restic/pull/4837 .

On startup, the `restore` command now checks existing files to determine whether these have to be updated. By default, it verifies the full file content. To only checks whether the file size and modification time (mtime) match with those in the snapshot, use `--overwrite if-changed`. Files with matching file content will not be downloaded during the restore operation. Only their file metadata is updated.

See #4817
Fixes https://github.com/restic/restic/issues/407
Fixes https://github.com/restic/restic/issues/2662

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4817
+++ b/changelog/unreleased/issue-4817
@@ -1,11 +1,22 @@
 Enhancement: Make overwrite behavior of `restore` customizable
 
 The `restore` command now supports an `--overwrite` option to configure whether
-already existing files are overwritten. The default is `--overwrite always`,
-which overwrites existing files. `--overwrite if-newer` only restores files
-from the snapshot that are newer than the local state. And `--overwrite never`
-does not modify existing files.
+already existing files are overwritten. This behavior can now be configured via
+the `--overwrite` option. The following values are supported:
+
+* `--overwrite always` (default): always overwrites already existing files. `restore`
+  will verify the existing file content and only restore mismatching parts to minimize
+  downloads. Updates the metadata of all files.
+* `--overwrite if-changed`: like the previous case, but speeds up the file content check
+  by assuming that files with matching size and modification time (mtime) are already up to date.
+  In case of a mismatch, the full file content is verified. Updates the metadata of all files.
+* `--overwrite if-newer`: only overwrite existing files if the file in the snapshot has a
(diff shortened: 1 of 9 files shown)


MichaelEischer:
Updated after discussion with @rawtaz. Instead of  `--verify-existing` this PR now introduces `--overwrite if-changed`. Verifying the file content of files with matching size&mtime was only relevant for `--overwrite always`. Thus, add two separate overwrite behaviors.

MichaelEischer:
`restore --verify` now only checks files that are modified during restore. This requires tracking whether the file content or only it's metadata was updated.

MichaelEischer (commented):
LGTM.

PR #4849, opened 2024-06-05
Use character display width for table padding

### LINKED ISSUE #4834: Slight cosmetic snapshot listing bug
Output of `restic version`
--------------------------
restic 0.16.4 compiled with go1.21.6 on darwin/amd64

SharePoint

If a snapshot's host (possibly other fields) contains an `’` it will throw off column alignment.  Or at least I think that's what's happening.  It could be that having that character in tags or pathnames would also throw off alignment, but I did not test that.

Expected behavior
-----------------

Tags should be lined up in one column

Actual behavior
---------------

Tags do not line up, possibly because of the `’` in the host name?  This does not happen on standard `'` apostrophes.

Screenshot 2024-05-29 at 11 21 47 AM

Screenshot 2024-05-29 at 11 23 37 AM

🫶

fthoma:
Using len(...) for table cell padding produced wrong results for unicode chracters leading to misaligned tables. Implementation changed to take the actual terminal display width into consideration. 
I chose to implement a simple helper function for the width calculation. There are libraries like  [go-runewidth](https://github.com/mattn/go-runewidth) that do a better job and take more edge cases into consideration. But i'm not sure its worth the dependency.

See before/after:
20240605_09h57m56s_grim

Closes #4834 

Checklist
---------

### DIFF
--- a/internal/ui/format.go
+++ b/internal/ui/format.go
@@ -8,6 +8,8 @@ import (
 	"math/bits"
 	"strconv"
 	"time"
+
+	"golang.org/x/text/width"
 )
 
 func FormatBytes(c uint64) string {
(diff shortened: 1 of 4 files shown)


internal/ui/format.go
```diff
@@ -105,3 +107,24 @@ func ToJSONString(status interface{}) string {
...
+func TerminalDisplayWidth(s string) int {
+	width := 0
+	for _, r := range s {
+		width += terminalDisplayRuneWidth(r)
+	}
+
+	return width
+}
+
+func terminalDisplayRuneWidth(r rune) int {
```
MichaelEischer, line 121: I've tried to unify this with `wideRune` in `internal/ui/termstatus/status.go`. However, it turned out that the use cases are somewhat different. In status, it is better to overestimate the rune width if in doubt. So let's keep separate functions for now.

MichaelEischer (approved):
LGTM. Thanks!

I've added a small cleanup commit and a fix for a probably pretty old bug.

Issue #4640, opened 2024-01-19
Add rclone binary to official docker container or provide an image variant including rclone binary


OrbitingOcelot:
Output of `restic version`
--------------------------
0.16.2

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
It would be really nice to have the latest rclone binary included in the official restic container image (or maybe provide a second container variant as part of the release process which contains the bin from rclone/rclone:latest?). 

What are you trying to do? What problem would this solve?
---------------------------------------------------------
I am evaluating backup strategies for my home-NAS, and after listening to a podcast with Alex I am really keen to start experimenting with restic, which looks awesome!

One thing that is currently providing some headache is the integration of rclone while using the official container image. I am using a backend that only has rclone support, so I need to have the rclone binary accessible in the container, which is currently not the case.

For most NAS distributions, it is really easy to install/update software via docker, so users can pull the official images for restic and rclone, but they cannot easily use the binaries together without potentially messy hacks, relying on 3rd party containers, or building their own container, and thus making things more complicated and losing their systems container GUI- or auto-update functionality.

Kopia decided to do this some time ago, and it just makes it so much easier to use on a NAS with limited build capabilities. Could help a lot of folks to have something similar for restic, especially since this is such an important part of the functionality for many backends.

Thank you for consideration!

MichaelEischer:
`rclone` has a very different release cycle than restic. To keep the binary up to date, we'd need some automation to rebuild the image. Maybe someone wants to set up the necessary Github Actions to publish such an image to the GHCR?

OrbitingOcelot:
Ah, I see, you mean building anew container on every rclone update, even if there is not restic update?

I was more or less thinking about the simpler option, just pulling the latest available rclone binary from rclone/rclone at restic build time, not if docker can do this natively by using something like FROM rclone/rclone ADD xyz...

MichaelEischer:
> Ah, I see, you mean building anew container on every rclone update, even if there is not restic update?

I had a CI job in mind that runs every few days. rclone usually has a higher release frequency than restic and I'd like to avoid issues that ask for updates. 

> I was more or less thinking about the simpler option, just pulling the latest available rclone binary from rclone/rclone at restic build time, not if docker can do this natively by using something like FROM rclone/rclone ADD xyz...

With multi-stage builds in docker it's really simple to collect binaries from multiple images.

PR #4859, opened 2024-06-11
Add rclone to the Docker container

alexander-bauer:

This change adds the `rclone` Alpine package to the Docker container. 

This is discussed in #4640.

There are some concerns there about the Rclone release cycle, and what it will take to keep the Restic Docker container up to date. I would argue that keeping Rclone up to date falls into the same problem category as the other packages included by our container: we probably need a general solution for publishing updated containers in-between Restic releases, but that should be a separate issue.

Checklist
---------

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,7 +11,7 @@ RUN go run build.go
 
 FROM alpine:latest AS restic
 
-RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
+RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq rclone
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -13,6 +13,6 @@ RUN mv /output/restic_${TARGETOS}_${TARGETARCH} /output/restic
 FROM alpine:latest
 
 COPY --from=helper /output/restic /usr/bin
-RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
+RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq rclone
 
 ENTRYPOINT ["/usr/bin/restic"]


alexander-bauer:
Unfortunately, making this change increases the size of the container pretty dramatically.

```
 ⇒  podman image ls localhost/restic/restic
REPOSITORY               TAG         IMAGE ID      CREATED        SIZE
localhost/restic/restic  latest      3680a2d3faad  7 minutes ago  41.6 MB  # from main
localhost/restic/restic  rclone      423c2eaf74a8  19 hours ago   127 MB   # from this branch
```

These are both container images built on my laptop with `podman`. I was hoping there's a flaw in my methodology here, like squashed layers being reused for the main branch, or something, but it looks like the `rclone` binary really is about that size.

```
 ⇒  unzip -l rclone-v1.66.0-linux-amd64.zip | tail -1 | xargs | cut -d' ' -f1 | numfmt --to=iec-i --suffix=B
66MiB
```

I realize I'm being a bit fast and loose with the units here, but ultimately I think whether `rclone` belongs in the standard image is down to the judgement call of whether that extra bulk is worthwhile.

The alternate approach would be to build and release a new flavor of the image which does come with the `rclone` binary. I think that wouldn't be too big a lift, but I'm a little hazy on how GHCR is managed. I'd be inclined to say that a "restic+rclone" variant would belong in its own image repository, as opposed to just being tagged differently, but I'm not sure how easy that is here.

Anyway, I'm eager to hear y'all's thoughts.

MichaelEischer (changes requested):
> we probably need a general solution for publishing updated containers in-between Restic releases, but that should be a separate issue.

Probably, however, that issue **MUST** be resolved first before we can consider this PR.

> I'd be inclined to say that a "restic+rclone" variant would belong in its own image repository, as opposed to just being tagged differently, but I'm not sure how easy that is here.

It's not uncommon to have images with different base images or slim and regular variants of an image in the same repository.

alexander-bauer:
@MichaelEischer Solving the release / update cycle first seems reasonable to me. Should I open another issue for discussing that, or is there a more appropriate place?

MichaelEischer:
@alexander-bauer Let's discuss that in #4640. My idea would be to setup a github ci job that regularly rebuilds the container and pushes it to ghcr.io (the Github container registry). Then we don't have to mess around with registry access permissions.

MichaelEischer:
I'll close this PR for now. It should be easy enough to recreate once the automatic rebuilds are in place.

Issue #4862, opened 2024-06-13
Detect mismatch between masterkey and remaining repository data


MichaelEischer:
Output of `restic version`
--------------------------
0.16.4

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
We recently had three issues where the `config` and `keys` of a repository were lost and the repository was reinitialized (https://forum.restic.net/t/ciphertext-verification-failed-different-repos-different-hosts/7652/5 , https://github.com/restic/restic/issues/4848 and https://github.com/restic/restic/issues/4832).

This leads  to very confusing errors like
```
$ restic snapshots
repository b9d21369 opened (version 2, compression level auto)
Ignoring "74c189425790faf2497ddface2a041ba7fbb883223fa8c93235b8d1b3d7f5d2c": failed to load snapshot 74c18942: ciphertext verification failed
Ignoring "104598c8811d078906e5f1d13044e83d826df034cfb3a1c86a31ac7b317d7876": failed to load snapshot 104598c8: ciphertext verification failed
Ignoring "61778ecfa0b209c6eb95a9c07a5058cf80bddb28e427a42e7f6495e377b06d61": failed to load snapshot 61778ecf: ciphertext verification failed
Ignoring "7d9cf8d381ff6651d5761f9fd8d15566d142e70d827380500b47708eaa4840c3": failed to load snapshot 7d9cf8d3: ciphertext verification failed
```

or 

```
$ restic check --read-data
using temporary cache in /tmp/restic-check-cache-3789657315
repository b9d21369 opened (version 2, compression level auto)
created new cache in /tmp/restic-check-cache-3789657315
create exclusive lock for repository
load indexes
[0:00] 100.00%  4 / 4 index files loaded
error: error loading index a1243cf997162370e57ad2589f605271922748b20f8800b8090cfb1752b52285: ciphertext verification failed
error: error loading index e83ccb66f95576b834669f31fb7f8e69059a4e343eacba679c6941fba66df4a3: ciphertext verification failed
error: error loading index a75ea15927343280ab9a4dfdbf6615f6c1489dc89c132f334f8f866159e0ac1b: ciphertext verification failed
error: error loading index 474bc50b2c62b12fc3ac8d7127c9583a6468f6e652bab263c8b78e14896f553f: ciphertext verification failed
Fatal: LoadIndex returned errors
```

Ideally restic should somehow detect this situation and provide a more useful error message.

What are you trying to do? What problem would this solve?
[...]

PR #4863, opened 2024-06-13
repository: prevent initialization if a snapshot exists

MichaelEischer:
Extend https://github.com/restic/restic/pull/4853 to also check for snapshots in addition to key files. Two of the issues referenced in https://github.com/restic/restic/issues/4862 were caused by a misconfigured retention policy. That policy removed the config and key files first  (the oldest files in the repository). Thus, checking for a key files would not detect this case. For a repository that is in regular use we can however assume that still some snapshots exist. Therefore, prevent initializing a new repository if there are snapshots.

Stating the `config` file and listing the `snapshots` also has the benefit of using two separate operations, which reduces the risk that both checks fail at the same time.

Follow up to https://github.com/restic/restic/pull/4853

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -780,6 +780,15 @@ func (r *Repository) Init(ctx context.Context, version uint, password string, ch
 	}); err != nil {
 		return err
 	}
+	// Also check for snapshots to detect repositories with a misconfigured retention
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
LGTM.

PR #4905, opened 2024-07-06
Update snapshot summary on rewrite

### LINKED ISSUE #4902: rewrite copies "Size" metadata instead of recalculating it if files have been excluded
Output of `restic version`
--------------------------
restic 0.16.5-dev (compiled manually) compiled with go1.22.0 on windows/amd64

local

I used `restic rewrite` to exclude some folders / files from a few existing snapshots

Expected behavior
-----------------
If files / folders get excluded in a rewrite then restic should recalculate the "Size" value which it displays when using `restic snapshots`

Actual behavior
---------------
Restic just copies the "Size" metadata which results in both the original and the rewritten snapshot displayed with the same size

I haven't looked at the code but my guess would be that this "step" was simply forgotton as no other metadata is changed (except if you ask it to)

restic is great and I use it for server backups as well as dekstop pc backups

alex-kattathra-johnson:
The snapshot summary doesn't reflect the files that are tracked after a rewrite

Fixes https://github.com/restic/restic/issues/4902

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_rewrite.go
+++ b/cmd/restic/cmd_rewrite.go
@@ -134,20 +134,29 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
 			return true
 		}
 
-		rewriter := walker.NewTreeRewriter(walker.RewriteOpts{
-			RewriteNode: func(node *restic.Node, path string) *restic.Node {
-				if selectByName(path) {
-					return node
-				}
-				Verbosef(fmt.Sprintf("excluding %s\n", path))
-				return nil
-			},
-			DisableNodeCache: true,
-		})
+		rewriteNode := func(node *restic.Node, path string) *restic.Node {
+			if selectByName(path) {
+				return node
+			}
+			Verbosef(fmt.Sprintf("excluding %s\n", path))
+			return nil
+		}
+
+		rewriter, querySize := walker.NewSnapshotSizeRewriter(rewriteNode)
 
 		filter = func(ctx context.Context, sn *restic.Snapshot) (restic.ID, error) {
-			return rewriter.RewriteTree(ctx, repo, "/", *sn.Tree)
+			id, err := rewriter.RewriteTree(ctx, repo, "/", *sn.Tree)
+			if err != nil {
+				return restic.ID{}, err
+			}
+			ss := querySize()
+			if sn.Summary != nil {
+				sn.Summary.TotalFilesProcessed = ss.FileCount
+				sn.Summary.TotalBytesProcessed = ss.FileSize
+			}
+			return id, err
 		}
+
 	} else {
 		filter = func(_ context.Context, sn *restic.Snapshot) (restic.ID, error) {
 			return *sn.Tree, nil
--- a/internal/walker/rewriter.go
+++ b/internal/walker/rewriter.go
@@ -11,6 +11,12 @@ import (
 
 type NodeRewriteFunc func(node *restic.Node, path string) *restic.Node
 type FailedTreeRewriteFunc func(nodeID restic.ID, path string, err error) (restic.ID, error)
+type QueryRewrittenSizeFunc func() SnapshotSize
+
+type SnapshotSize struct {
+	FileCount uint
+	FileSize  uint64
+}
 
 type RewriteOpts struct {
 	// return nil to remove the node
@@ -50,6 +56,29 @@ func NewTreeRewriter(opts RewriteOpts) *TreeRewriter {
 		}
 	}
 	return rw
+}
+
+func NewSnapshotSizeRewriter(rewriteNode NodeRewriteFunc) (*TreeRewriter, QueryRewrittenSizeFunc) {
+	var count uint
+	var size uint64
+
+	t := NewTreeRewriter(RewriteOpts{
+		RewriteNode: func(node *restic.Node, path string) *restic.Node {
+			node = rewriteNode(node, path)
+			if node != nil && node.Type == "file" {
+				count++
+				size += node.Size
+			}
+			return node
+		},
+		DisableNodeCache: true,
+	})
+
+	ss := func() SnapshotSize {
+		return SnapshotSize{count, size}
+	}
+
+	return t, ss
 }
 
 type BlobLoadSaver interface {
--- a/internal/walker/rewriter_test.go
+++ b/internal/walker/rewriter_test.go
@@ -303,6 +303,60 @@ func TestRewriter(t *testing.T) {
 	}
 }
 
+func TestSnapshotSizeQuery(t *testing.T) {
+	tree := TestTree{
+		"foo": TestFile{Size: 21},
+		"bar": TestFile{Size: 21},
+		"subdir": TestTree{
+			"subfile": TestFile{Size: 21},
+		},
+	}
+	newTree := TestTree{
+		"foo": TestFile{Size: 42},
+		"subdir": TestTree{
+			"subfile": TestFile{Size: 42},
+		},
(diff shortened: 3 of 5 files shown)


cmd/restic/cmd_rewrite.go
```diff
@@ -217,6 +217,31 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *r
 	sn.Original = sn.ID()
 	sn.Tree = &filteredTree
 
+	sn.Summary.TotalFilesProcessed = 0
```
MichaelEischer, line 220: For old snapshots, summary is a nil pointer, such that this would cause a crash.

cmd/restic/cmd_rewrite.go
```diff
@@ -217,6 +217,31 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *r
...
+	sn.Summary.TotalBytesProcessed = 0
+	processNode := func(_ restic.ID, nodepath string, node *restic.Node, err error) error {
```
MichaelEischer, line 222: Recalculating the snapshot size is rather expensive and only makes sense if `len(rejectByNameFuncs) > 0`.

Iterating through the tree a second time is rather inefficient, instead this recalculation must happen in the `TreeRewriter` in internal/walker/rewriter.go (but only if `DisableNodeCache` is true) where it can be done nearly for free.

MichaelEischer (changes requested):
Thanks for giving this a try. Recalculating the snapshot size is definitely useful. However, the calculation itself does not belong into `cmd_rewrite.go`, see comments below for more details.

alex-kattathra-johnson:
@MichaelEischer : Thanks for the feedback! Updated the PR to make the change to the summary only if it's non-nil and moved the update to the `RewriteNode` callback function.

cmd/restic/cmd_rewrite.go
alex-kattathra-johnson: By updating snapshot summaries of all snapshots that have a certain path, would we change snapshots we initially didn't mean to touch. I think the recalculating of the size should only be relevant to the specified snapshot. No reference of snapshot gets passed into the `TreeRewriter`.

alex-kattathra-johnson: Would referencing snapshots in `internal/walker/rewriter.go` cause a circular referencing? Maybe we should update in the `RewriteNode` callback like in the latest commit.

alex-kattathra-johnson: I suppose the execution takes place in `internal/walker/rewriter.go`, haha.

cmd/restic/cmd_rewrite.go
```diff
@@ -136,6 +136,10 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
 				if selectByName(path) {
 					return node
 				}
+				if sn.Summary != nil && node.Type == "file" {
```
MichaelEischer, line 139: This will break if any directory is excluded.

MichaelEischer: Instead please count all files returned by `return node` and use that to calculate the size from scratch.

MichaelEischer: I'm not yet entirely sure that this is the best place to implement the size calculation. Here it is pretty hard to test it properly. What would be much more testable is SizeCalculationgRewriter that wraps `walker.NewTreeRewriter` and also the passed in `RewriteNode` function to count the remaining snapshot size.

To be a bit more specific I have something like the following in mind (I'm not entirely happy about the `func() SnapshotSize` part):
```go

type SnapshotSize struct {
	FileCount int
	FileSize  int64
}

func NewSizeCountingRewriter(rewriteNode NodeRewriteFunc) (*TreeRewriter, func() SnapshotSize) {
	count := 0
	size := int64(0)
	return NewTreeRewriter(RewriteOpts{
			RewriteNode: func(node *restic.Node, path string) *restic.Node {
				node = rewriteNode(node, path)
				if node != nil {
					// accounting
				}
				return nil
			},
			DisableNodeCache: true,
		}), func() SnapshotSize {
			return SnapshotSize{count, size}
		}
}
```

That would allow using the test infrastructure that already exists in `internal/walker/rewriter_test.go`. With that in place, a single test in `cmd_rewrite.go` would be enough to provide a decent test coverage.

alex-kattathra-johnson: @MichaelEischer : Could you take another look? I ended up creating a function which accepts with snapshot summary as input to be updated in place. And added some tests too.

internal/walker/rewriter.go
```diff
@@ -10,8 +10,14 @@ import (
 )
 
 type NodeRewriteFunc func(node *restic.Node, path string) *restic.Node
+type SnapshotSizeRewriteFunc func() SnapshotSize
```
MichaelEischer, line 13: I'd prefer calling this function `QueryRewrittenSizeFunc` or similar. Currently, it sounds like it actually rewrites the snapshot's size, but that isn't the case.

internal/walker/rewriter.go
```diff
@@ -52,6 +58,32 @@ func NewTreeRewriter(opts RewriteOpts) *TreeRewriter {
...
 
+func NewSnapshotSizeRewriter(rewriteNode NodeRewriteFunc) (*TreeRewriter, SnapshotSizeRewriteFunc) {
+	var count uint
+	var size uint64
+
+	t := NewTreeRewriter(RewriteOpts{
+		RewriteNode: func(node *restic.Node, path string) *restic.Node {
+			node = rewriteNode(node, path)
+			if node != nil {
+				if node.Type == "file" {
```
MichaelEischer, line 69: both `if` can be merged when the function always returns `node`.

cmd/restic/cmd_rewrite.go
```diff
@@ -134,20 +134,29 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
...
-		})
+		rewriteNode := func(node *restic.Node, path string) *restic.Node {
+			if selectByName(path) {
+				return node
+			}
+			Verbosef(fmt.Sprintf("excluding %s\n", path))
+			return nil
+		}
+
+		rewriter, snapshotSizeRewriter := walker.NewSnapshotSizeRewriter(rewriteNode)
```
MichaelEischer, line 145: please rename `snapshotSizeRewriter` to `querySize`

internal/walker/rewriter_test.go
```diff
@@ -303,6 +303,69 @@ func TestRewriter(t *testing.T) {
...
+		}
+		rewriter, snapshotSizeRewriter := NewSnapshotSizeRewriter(rewriteNode)
+		newRoot, err := rewriter.RewriteTree(ctx, modrepo, "/", root)
+		if err != nil {
+			t.Error(err)
+		}
+
+		ss := snapshotSizeRewriter()
+
+		if ss.FileCount != 2 {
```
MichaelEischer, line 345: Please replace the if-block with `test.Equals(t, uint(2), ss.FileCount, "snapshot file count mismatch")`. (Similarly for `FileSize` below).

MichaelEischer (commented):
Thanks for the changes so far. I have only a few small nitpicks left, see below.

I'd also like to have a minimal functionality test in `cmd/restic/cmd_rewrite_integration_test.go`, maybe by extending `TestRewriteReplace` to check the snapshot's stats before and after rewriting. I can take care of that if you want.

MichaelEischer (approved):
LGTM. Thanks!

PR #4928, opened 2024-07-20
Custom entrypoint for container

### LINKED ISSUE #4927: Custom entrypoint for container
Output of `restic version`
--------------------------
restic 0.16.5 compiled with go1.22.4 on linux/amd64 (inside the official docker container)

I would like to add a custom entrypoint to the Dockerfile to make the use of the container easier.

The shell script would check if the repository exists at the target location, by trying to open the restic config.
If there is no valid config found, indicating that the repo is not initialised it would initialise it befoire attemting to backup anything.
Then ofc. it would start the backup process with the usual command, but the source directory would be set via an environment variable. (Would default to /data if not set.) ```restic backup $RESTIC_DATA```
At the end if the backup finished successfully, it would apply a retention policy if enabled.
The restention policy could be enabled by setting an environmentvariable, $RESTIC_FORGET, to any flag(s) accepted by the forget command. ```restic forget $RESTIC_FORGET```

With this the container would be more straight forward and one container would be capable to **init**, **backup** and **forget** in one go, without providing a command parameter to the container, making the deployment easier.

Yes, made my daily backup ;)

KGeri201:
Implements #4927

Closes #4927

Checklist
---------

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,4 +17,6 @@ COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
 COPY docker/entrypoint.sh /
 
+VOLUME ["/data"]
+
 ENTRYPOINT ["/entrypoint.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,4 +15,6 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
-ENTRYPOINT ["/usr/bin/restic"]
+COPY docker/entrypoint.sh /
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -11,18 +11,32 @@ From the root of this repository run:
 image name will be `restic/restic:latest`
 
 ## Run
-
-Set environment variable `RESTIC_REPOSITORY` and map volume to directories and
+Set environment variable `RESTIC_REPOSITORY`, `RESTIC_PASSWORD` and map volume to directories and
 files like:
 
 ```
-docker run --rm --hostname my-host -ti \
-    -v $HOME/.restic/passfile:/pass \
+docker run --rm --hostname my-host \
     -v $HOME/importantdirectory:/data \
     -e RESTIC_REPOSITORY=rest:https://user:pass@hostname/ \
-    restic/restic -p /pass backup /data
+    -e RESTIC_PASSWORD=/run/secrets/restic_password
+    restic/restic
 ```
 
 Restic relies on the hostname for various operations. Make sure to set a static
 hostname using `--hostname` when creating a Docker container, otherwise Docker
 will assign a random hostname each time.
+
+## Advanced
+Optionally you can change the source directory by setting `RESTIC_DATA` to an other path, 
+defaults to /data.  
+Or you can enable a [retention policy](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) 
+by setting `RESTIC_FORGET`.
+For other possible environment variables check: [Environment Variables](https://restic.readthedocs.io/en/stable/040_backup.html#environment-variables)
+
+By removing the *--rm* tag, the container would not destroy itself after it is finished, 
+only exits, and you can schedule your backups just by starting the container with cron like this:
+```cron
+0 0 * * * docker start restic
+```
+
+To not have your password plain text in your shell use a [docker-compose.yml](docker-compose.yml) and docker secrets.
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,13 @@
+services:
+  restic:
+    hostname: restic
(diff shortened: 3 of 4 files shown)


docker/Dockerfile
```diff
@@ -15,4 +15,8 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
-ENTRYPOINT ["/usr/bin/restic"]
+COPY docker/entrypoint.sh /
+
+VOLUME ["/data"]
+
+ENTRYPOINT ["/entrypoint.sh"]
```
MichaelEischer, line 22: This changes the default behavior from printing a help message to actually performing a backup. It won't make much of a difference when your setup is already working, but might be unexpected for new users.

docker/README.md
```diff
@@ -11,18 +11,32 @@ From the root of this repository run:
...
 files like:
 
 ```
-docker run --rm --hostname my-host -ti \
-    -v $HOME/.restic/passfile:/pass \
+docker run --rm --hostname my-host \
     -v $HOME/importantdirectory:/data \
     -e RESTIC_REPOSITORY=rest:https://user:pass@hostname/ \
-    restic/restic -p /pass backup /data
+    -e RESTIC_PASSWORD=/run/secrets/restic_password
```
MichaelEischer, line 21: This hardcodes `/run/secrets/restic_password` as password.

docker/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+restic="/usr/bin/restic"
+
+# if there are command arguments passed to the container execute them and exit.
+if [ $# -ne 0 ]; then
+  ${restic} $@
+  exit 0
+fi
+
+# check if the repo exists and initialize it if not
+${restic} cat config > /dev/null || ${restic} init
```
MichaelEischer, line 17: The latest restic binaries return exit code 10 if a repositoy does not exist https://restic.readthedocs.io/en/latest/075_scripting.html#check-if-a-repository-is-already-initialized . It doesn't make sense to try to reinitialize the repository in other cases.

Besides that I'm not a fan of automatically initializing repositories. A repository should only be initialized once. If it disappears later on, there may be a major problem.

docker/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# backup data
+${restic} backup ${RESTIC_DATA:-"/data"}
+
+# apply retention policy if it is set
+if [ -n "${RESTIC_FORGET:-}" ]; then 
+  ${restic} forget ${RESTIC_FORGET} --prune
```
MichaelEischer, line 24: Running prune after every backup is overkill.

docker/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  restic:
+    hostname: restic
+    volumes:
+      - $HOME/importantdirectory:/data
+    environment:
+      - RESTIC_REPOSITORY=rest:https://user:pass@hostname/
```
MichaelEischer, line 7: That should use `--repository-file` as well

docker/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+      - RESTIC_PASSWORD=/run/secrets/restic_password
```
MichaelEischer, line 8: The correct environment variable would be `RESTIC_PASSWORD_FILE`.

docker/README.md
```diff
@@ -11,18 +11,32 @@ From the root of this repository run:
...
+Optionally you can change the source directory by setting `RESTIC_DATA` to an other path, 
+defaults to /data.  
+Or you can enable a [retention policy](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) 
+by setting `RESTIC_FORGET`.
+For other possible environment variables check: [Environment Variables](https://restic.readthedocs.io/en/stable/040_backup.html#environment-variables)
+
+By removing the *--rm* tag, the container would not destroy itself after it is finished, 
+only exits, and you can schedule your backups just by starting the container with cron like this:
+```cron
+0 0 * * * docker start restic
```
MichaelEischer, line 39: It feels rather weird to stop and restart a docker container later on. AFAIK it's much more canonical to start new fresh containers each time. This will require storing the cache on a separate volume for optimal performance though. Always reusing the same container also complicates upgrading restic. You'd have to manually delete the container in this case.

Besides that, this construction prevents setups that for example run `forget` and `prune` only once a week. Or similarly run `check` on a weekly basis.

docker/README.md
```diff
@@ -11,18 +11,32 @@ From the root of this repository run:
...
+```
+
+To not have your password plain text in your shell use a [docker-compose.yml](docker-compose.yml) and docker secrets.
```
MichaelEischer, line 42: That's not accurate. You can mount the passwords as files into the container.

MichaelEischer (changes requested):
I don't think that the official restic container is the right place for this script. The necessary configuration is probably very different depending on the individual use case.

I've commented on different buggy parts of this PR nevertheless.

KGeri201:
OK, I understand.
Thanks for your feedback.

PR #4952, opened 2024-07-28
Format exit errors as JSON if requested

### LINKED ISSUE #4948: Fatal errors should be JSON formatted with --json
Output of `restic version`
--------------------------
restic 0.16.4 compiled with go1.22.2 on linux/amd64

Currently, fatal errors aren't JSON formatted if `--json` is provided. It would be nice if it did surface that in a more programmatically-detectable way.

Fatal errors can be detected by scanning for lines that start with `Fatal:` but still... it'd make me feel safer if I could instead scan for a JSON blob with `"message_type": "fatal_error"` though.

This would be especially handy if Restic ever wants to add more metadata to the error (like requested in #4947)

Open Questions
------------------------
- What should the `message_type` be? I could see an argument for `fatal_error` but I could also see an argument for just `error` (also used for per-file errors in backup/restore commands) with either a null path or a flag like "is_fatal: true`.
- Should we keep printing the `Fatal:` freeform string to the console, to ease the transition for folks like me that are already looking for and parsing such lines?

I'm happy to work on a PR for this, if I could get some pre-guidance on the open questions. Thanks!

mikix:

When using `--json`, exit errors are now formatted as JSON instead of freeform text. And the exit code is included in the JSON blob, for the convenience of any consumers (so they don't have to wait for the exit code and then tie it together with the message).

While I was testing, I also noticed that the `version` command didn't use a `message_type` field in JSON mode - I figured I'd toss one in there for consistency with other messages. If that is undesired, I'm happy to take it out of this PR.

Closes #4948

Checklist
---------

- ~[ ] I have added tests for all code changes.~ (this felt hard to do with the exit call, unless I refactored things a bunch - but I did manually test this)

### DIFF
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
@@ -102,6 +103,30 @@ func tweakGoGC() {
 	}
 }
 
+func printExitError(code int, message string) {
+	if globalOptions.JSON {
+		type jsonExitError struct {
+			MessageType string `json:"message_type"` // exit_error
+			Code        int    `json:"code"`
+			Message     string `json:"message"`
+		}
+
+		jsonS := jsonExitError{
+			MessageType: "exit_error",
+			Code:        code,
+			Message:     message,
+		}
+
+		err := json.NewEncoder(globalOptions.stderr).Encode(jsonS)
+		if err != nil {
+			Warnf("JSON encode failed: %v\n", err)
+			return
+		}
+	} else {
+		fmt.Fprintf(globalOptions.stderr, "%v\n", message)
+	}
+}
+
 func main() {
 	tweakGoGC()
 	// install custom global logger into a buffer, if an error occurs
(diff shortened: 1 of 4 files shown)


cmd/restic/main.go
```diff
@@ -160,10 +187,16 @@ func main() {
 		exitCode = 10
 	case restic.IsAlreadyLocked(err):
 		exitCode = 11
+	case errors.Is(err, repository.ErrNoKeyFound):
+		exitCode = 12
```
mikix, line 191: This is unrelated to main change, but I threw this in while I was here, since it is useful to programmatically detect a bad password.

cmd/restic/main.go
```diff
@@ -102,6 +103,30 @@ func tweakGoGC() {
 	}
 }
 
+func printExitError(code int, message string) {
+	if globalOptions.JSON {
+		type jsonExitError struct {
+			MessageType string `json:"message_type"` // exit_error
+			Code        int    `json:"code"`
+			Error       string `json:"error"`
```
mikix, line 111: Do we want to mimic the backup/restore errors here with the struct shape from #4946 that holds a string? Or try to distance ourselves from that arguably accidental setup? Or... a half way measure and name this string field `message` to avoid comparisons to the `error` field in those JSON blobs?

MichaelEischer: I prefer the field name `message`. That would then be consistent with an inlined `error` object. Also `jsonExitError.Code` and `jsonExitError.Message` read much better than ``jsonExitError.Error`.

doc/075_scripting.rst
```diff
@@ -83,13 +83,23 @@ JSON output of most restic commands are documented here.
     list of allowed values is documented may be extended at any time.
 
 
+Exit errors
+-----------
+
+Fatal errors will result in a final JSON message before the process exits.
```
MichaelEischer, line 89: It might be interesting to add a remark that JSON output cannot be guaranteed if parsing the CLI arguments failed.

MichaelEischer (changes requested):
Fine with me in general. I just have a few nitpicks, see below.

cmd/restic/main.go
mikix: Done (switched to `message`).

doc/075_scripting.rst
mikix: I added this comment:

```
.. note::
    Some errors cannot be caught and reported this way,
    such as Go runtime errors or command line parsing errors.
```

doc/075_scripting.rst
```diff
@@ -83,13 +83,23 @@ JSON output of most restic commands are documented here.
...
++----------------------+-------------------------------------------+
+| ``code``             | Exit code (see above chart)               |
++----------------------+-------------------------------------------+
+| ``error``            | Error message                             |
++----------------------+-------------------------------------------+
+
 Output formats
 --------------
 
-Currently only the output on ``stdout`` is JSON formatted. Errors printed on ``stderr``
```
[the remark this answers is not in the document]
mikix: I removed it because I felt like the errors-as-JSON situation was so much better that it didn't deserve a big warning. But how about the current text:

```

Commands print their main JSON output on ``stdout``.
The generated JSON output uses one of the following two formats.

.. note::
    Not all messages and errors have been converted to JSON yet.
    Feel free to submit a pull request!
```

That note mirrors an earlier one in the doc about some messages not supporting `--json` at all.

(I also emphasized that some error messages print to `stderr` a few more places in the docs.)

MichaelEischer (approved):
LGTM. Thanks!

PR #4958, opened 2024-07-30
restore: clean up error handling when restoring metadata

mikix:
- Fix a logic error that instead of reporting the *first* metadata-setting error that appears, we were instead reporting the *last* error (and only if the lchown call failed!).
- Don't show any errors when setting metadata for files in non-root mode (things like timestamps, attributes). Previously, only lchown errors were skipped. But other kinds of attribute errors make sense to skip as well. The code path happened to work correctly before because of the above logic error. But once that was fixed, this change needed to happen too.

No - but it's a break out of #4946

Checklist
---------


MichaelEischer (approved):
LGTM. Thanks a lot for finding and fixing this very old bug!

PR #4959, opened 2024-07-31
main: return an exit code (12) for "bad password" errors

mikix:
This is useful for a wrapping app to be able to detect that the user provided the wrong key and then re-prompt the user.

As part of this, I stopped converting the error into a fatal error, to avoid swallowing the original error - and then changed how main handles non-fatal errors.

(I did not think this was a big enough change to warrant a changelog entry, but my sense of that is not well-tuned. Please let me know if you'd like one.)

No

Checklist
---------


cmd/restic/global.go
```diff
@@ -496,7 +496,7 @@ func OpenRepository(ctx context.Context, opts GlobalOptions) (*repository.Reposi
 		if errors.IsFatal(err) {
 			return nil, err
 		}
-		return nil, errors.Fatalf("%s", err)
+		return nil, errors.FatalWrap(err)
```
mikix, line 499: This line here was previously "erasing" the `ErrNoKeyFound` type and making the new error just a fatalError (string) type.

I ama Go (and restic) novice, so if this is a bad approach to the issue, please suggest a new one and I'll implement it.

cmd/restic/main.go
```diff
@@ -160,6 +161,8 @@ func main() {
 		exitCode = 10
 	case restic.IsAlreadyLocked(err):
 		exitCode = 11
+	case errors.Is(err, repository.ErrNoKeyFound):
+		exitCode = 12
```
mikix, line 165: The exit codes here are a little jumpy. I'm guessing that the vibe is to use 10+ for random specific error codes. Leaving <10 and >100 for special cases.

internal/errors/fatal.go
```diff
@@ -7,25 +7,37 @@ import (
...
 
 // Fatalf returns an error that is marked fatal.
 func Fatalf(s string, data ...interface{}) error {
-	return Wrap(fatalError(fmt.Sprintf(s, data...)), "Fatal")
+	return FatalWrap(fmt.Errorf(s, data...))
+}
+
+// FatalWrap wraps a lesser error as fatal.
+func FatalWrap(err error) error {
+	return &fatalError{fmt.Sprintf("Fatal: %v", err), err}
```
konidev20, line 42: `fmt.Errorf` does provide error wrapping using the `%w` format specifier used like `fmt.Errof("Fatal: %w", err)`

cmd/restic/global.go
konidev20: I think a simple `fmt.Errof("Fatal: %w")` would have the same effect. You will be able to check for the if the error `ErrKeyNotFound` is part of the error tree using `errors.Is(err, ErrKeyNotFound)`

cmd/restic/main.go
MichaelEischer: The 10+ range (probably up to 20) is currently intended for errors related to the repository itself. The 1+ range are general / command errors. There's no really strict approach how to allocate the exitCodes though.

Please update the documentation in `doc/075_scripting.rst` and the help text returned by each command (the easiest way is probably to use search&replace to add exit code 12 to every command that already returns exit code 11).

cmd/restic/global.go
MichaelEischer: I'd like to eventually get rid of the fatal error type. so let's just use `fmt.Errorf` instead here.

MichaelEischer (changes requested):
Good idea. I'd like to follow a different way to implement this though, see the comments for details.

internal/errors/fatal.go
mikix: Yes, it does - but the result wouldn't be a `fatalError` type, which parts of the code elsewhere check for - `fmt.Errorf` wraps using an internal stdlib type. So if you want wrapping with a specific type (`fatalError`), you need to kind of do that yourself.

(at least, that is my limited understanding of Go)

cmd/restic/global.go
mikix: > I think a simple fmt.Errof("Fatal: %w") would have the same effect. You will be able to check for the if the error ErrKeyNotFound is part of the error tree using errors.Is(err, ErrKeyNotFound)

Correct, but the result wouldn't be a `fatalError`, which this block was specifically trying to generate - it either returns the error if it's fatal or if it's not fatal, wraps it in a fatal error.

> I'd like to eventually get rid of the fatal error type. so let's just use fmt.Errorf instead here.

But if that's the case, yes, we can just use `Errorf` - I will retool this PR and drop the wrapping.

cmd/restic/main.go
```diff
@@ -139,15 +139,7 @@ func main() {
...
-		fmt.Fprintf(os.Stderr, "%+v\n", err)
-
-		if logBuffer.Len() > 0 {
-			fmt.Fprintf(os.Stderr, "also, the following messages were logged by a library:\n")
-			sc := bufio.NewScanner(logBuffer)
-			for sc.Scan() {
-				fmt.Fprintln(os.Stderr, sc.Text())
-			}
-		}
+		fmt.Fprintf(os.Stderr, "Fatal: %v\n", err)
```
mikix, line 142: This was a big swing by me - happy to go a different way.

Previously, `IsFatal` was used to flag "expected errors that will bubble up to main.go" - and thus any other errors should get all this debugging info added, since they were unexpected.

But if we are moving away from fatal errors and just using the original errors instead, we can either (a) add a case statement for each expected error, or (b) drop this extra debugging step for unexpected errors. I chose the latter, but am happy to do the former (or some other idea).

cmd/restic/main.go
mikix: Looks like there's only one block that explains the exit codes, so I added 12 to it. I'm assuming the next release is 0.17.1

cmd/restic/main.go
```diff
@@ -139,15 +139,7 @@ func main() {
 	case errors.IsFatal(err):
 		fmt.Fprintf(os.Stderr, "%v\n", err)
 	case err != nil:
-		fmt.Fprintf(os.Stderr, "%+v\n", err)
-
-		if logBuffer.Len() > 0 {
-			fmt.Fprintf(os.Stderr, "also, the following messages were logged by a library:\n")
```
MichaelEischer, line 145: I'd like to keep those library logs for now. Those are usually empty anyways.

cmd/restic/main.go
MichaelEischer: That will definitely help moving forwards. But in my opinion it is too invasive to include in a patch version. So we'll have to change this separately from adding the exit code if the latter should be included in 0.17.1

MichaelEischer (changes requested):
The changelog entry went missing. Please restore it.

I'd prefer to stick with just warping the errors from opening a repository with `fmt.Errorf("Fatal: %w", err)` for now. Significantly reworking the error handling should happen separately and also doesn't belong into a patch release.

cmd/restic/main.go
mikix: (I myself do not need this to land on a specific schedule - so if you want to save some of the changes for 0.18, I won't be annoyed. :smile:)

Let's recap the options here:
- Don't wrap as a FatalError. Just wrap with `fmt.Errorf("%w")`
  - This causes the full stacktrace to be printed to the console for this error, since main.go no longer sees it as an expected error.
  - So option (1): drop stacktraces from main.go for all errors (what I've done here)
  - Or option (2): drop stacktraces from main.go for just  `repository.ErrNoKeyFound` errors
- Or option (3): Allow FatalErrors to properly wrap and hold an inside error. (my original version)

It sounds like option 3 you don't like in general. Option 1 you don't love in a patch version.

But option 2 is maybe fine as a short term fix? And maybe in 0.18 and/or as we grow more "expected error" cases, we can revisit.

mikix: OK, I added option 2 (just a manual check for this one error) to this PR.

cmd/restic/global.go
```diff
@@ -493,10 +493,7 @@ func OpenRepository(ctx context.Context, opts GlobalOptions) (*repository.Reposi
 		}
 	}
 	if err != nil {
-		if errors.IsFatal(err) {
-			return nil, err
-		}
-		return nil, errors.Fatalf("%s", err)
+		return nil, err
```
MichaelEischer, line 496: This still changes the error handling when loading the repository key does not work. I'd like to keep the current fatal error handling for those cases.

Either change this to
```suggestion
		if errors.IsFatal(err) || errors.Is(err, repository.ErrNoKeyFound) {
		    return nil, err
		}
		return nil, errors.Fatalf("%s", err)
```
or
```suggestion
		if errors.IsFatal(err) {
		    return nil, err
		}
		return nil, fmt.Errorf("Fatal: %w", err)
```

MichaelEischer (changes requested):

cmd/restic/global.go
mikix: Ah gotcha - yeah, I can make that change. Suggestion 2 feels slightly cleaner to me.

But can I also step back and ask about the thinking around FatalError? I kind of like it as a concept, and still like my original fix here that added wrapping support.

FatalError solves one problem currently, and could be augmented to do more:
1. (currently) It flags errors as "expected" so that when `main.go` gets an exit error that isn't Fatal, we know to add extra debugging info.
2. (future) It could hold an exit code (from a centrally maintained list), so that `main.go` doesn't need to have its own switch/case list of them. This sort of thing would be particularly handy in the future, when a EACCES syscall errno in one part of the code might mean exit code 72, but in another part of the code would be exit code 76.  You'd have to wrap the the syscall.Errno in a custom error anyway so that `main.go` could disambiguate the cases.

If you are mentally deprecating FatalError, I imagine that `main.go`'s switch/case list is going to keep growing for both those use cases. I personally vibe with the FatalError approach over that.

mikix: OK, this PR is now up-to-date and includes suggestion 1.

Happy to land this while also nattering about FatalError.

mikix: Hmm, another advantage of FatalError is that it’s easier to test, versus main.go

MichaelEischer: > If you are mentally deprecating FatalError, I imagine that `main.go`'s switch/case list is going to keep growing for both those use cases. I personally vibe with the FatalError approach over that.

I haven't though particularly much about how a replacement for the current FatalError should look like. The only thing that's clear is that the current "convert everything into a string"-approach doesn't work. But I also agree that the switch/case list in main.go is already a mess.

> 2\. It could hold an exit code (from a centrally maintained list), so that `main.go` doesn't need to have its own switch/case list of them.

Sounds reasonable. I'm not entirely sure yet which layer should make that translation. Should it already happen in `internal/` or wait until the `cmd/restic/` code? Just blindly wrapping all errors also leaks quite a lot of implementation details, which probably should be avoided.

Quite a lot of code currently just passes an error up the call chain without wrapping it in any way, which makes it rather hard to pinpoint the origin. So, we'll need some mix of both. Not sure how that will look in the end.

The main reason why I don't want these changes in this PR is that they are too invasive for inclusion in a patch release.

MichaelEischer (approved):
LGTM. Thanks!

Issue #4950, opened 2024-07-28
Backup time increased significantly on version 0.17.0 (Windows)


jvandenbroek:
Output of `restic version`
--------------------------
restic 0.17.0 compiled with go1.22.5 on windows/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
Rclone SFTP

Problem description / Steps to reproduce
----------------------------------------
Backing up a remote Samba share from a relatively slow NAS results in backup duration of ~2 hours instead of ~15 minutes it takes on version 0.16.x. This probably has to do with the new Extended Attributes support, which is a really nice feature to have, however it should not have such an impact.

An option to revert to the old scan behavior would probably solve this, no need to check for changed attributes every backup, especially if it is this expensive.

Expected behavior
-----------------
Same or slight increase of backup times compared to previous version.

Actual behavior
---------------
Backup time takes ~8 times longer when source is on  (slow) network share.

Do you have any idea what may have caused this?
-----------------------------------------------
Probably the new Extended Attributes support.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Restic makes my life easier, can't think of any better / more reliable backup solution.

MichaelEischer:
This is probably related to https://github.com/restic/restic/issues/4955 . Although, there extended attributes just result in weird errors. Apparently, samba manages to at least no break entirely.

aneesh-n:
I have created a PR https://github.com/restic/restic/pull/4980 to ignore EA processing for paths that do not support EA. Hopefully this should help in this case where the Samba share is the source.

It would be great if you could test this PR on your end as well.

MichaelEischer:
Fixed via #4980

PR #4980, opened 2024-08-03
Skip ExtendedAttribute processing in Windows for volumes that do not support EA

### LINKED ISSUE #4955: [Regression] restic 0.17.0 on Windows: cannot backup using network drive
Output of `restic version`
--------------------------
restic 0.17.0 compiled with go1.22.5 on windows/amd64

A network share mounted as drive H

On my employer's computer, I need to back up my data to a network drive, which is mounted locally as drive `H`. My `restic` command backs up files from both my local hard drive and the target drive itself. So, in addition to the backup, other data is also stored on the network drive. This worked flawlessly with `restic` up to version 0.16.x. However, since version 0.17.0, the following error message appears:

```
Fatal: unable to save snapshot: nodeFromFileInfo h:\: get EA failed for path h:\, with: get file EA failed with: Incorrect function.
``` 

My command:
```
restic --repo H:\Backup backup --exclude-caches --iexclude-file="c:\Users\xxx\backup-ignore.txt" --files-from="c:\Users\xxx\backup-resources.txt" --tag main --tag data --compression=auto --read-concurrency=6
```

The entire output looks like this:
```
pattern "c:\\Users\\xxx\\AppData\\Roaming\\Sublime Text 3\\Local\\" does not match any files, skipping
repository 6b02e3c8 opened (version 2, compression level auto)
found 1 old cache directories in C:\Users\xxx\AppData\Local\restic, run `restic cache --cleanup` to remove them
using parent snapshot 9e08dbd0
[0:00] 100.00%  13 / 13 index files loaded
Fatal: unable to save snapshot: nodeFromFileInfo h:\: get EA failed for path h:\, with: get file EA failed with: Incorrect function.
```

Expected behavior
-----------------
The backup to and from the network drive should work.

Actual behavior
---------------
No snapshot will be created, Instead, we got this error:
```
Fatal: unable to save snapshot: nodeFromFileInfo h:\: get EA failed for path h:\, with: get file EA failed with: Incorrect function.
``` 

@MichaelEischer mentioned on the [forum](https://forum.restic.net/t/backing-up-data-on-windows-from-smb-share-to-local-drive/8095): `This looks like a potential bug in the new metadata collection changes in restic 0.17.0`

Restic makes me happy: Finally a reliable backup software I can rely on. It also runs on all my systems: Linux, Windows, and macOS. Thank you very much for your work and all the time you have invested in the software.

aneesh-n:
Some windows paths like network drives do not support ExtendedAttributes. Either they return error codes like windows.E_NOT_SET or windows.ERROR_INVALID_FUNCTION or it results in slower backups.
This PR aims to skip processing the ExtendedAttributes after detecting if the volume does not support them.
It completely skips the attempt to fetch Extended Attributes instead of trying to handle errors after trying to fetch EA for files on such volumes.

May resolve https://github.com/restic/restic/issues/4955 and https://github.com/restic/restic/issues/4950

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-4980
+++ b/changelog/unreleased/pull-4980
@@ -1,10 +1,12 @@
 Bugfix: Skip EA processing in Windows for volumes that do not support EA
 
 Restic was failing to backup files on some windows paths like network
-drives because of errors while fetching ExtendedAttributes.
+drives because of errors while fetching extended attributes.
 Either they return error codes like windows.E_NOT_SET or
 windows.ERROR_INVALID_FUNCTION or it results in slower backups.
-Restic now completely skips the attempt to fetch Extended Attributes
+Restic now completely skips the attempt to fetch extended attributes
 for such volumes where it is not supported.
 
 https://github.com/restic/restic/pull/4980
+https://github.com/restic/restic/issues/4955
+https://github.com/restic/restic/issues/4950
--- a/changelog/unreleased/pull-4980
+++ b/changelog/unreleased/pull-4980
@@ -1,0 +1,10 @@
+Bugfix: Skip EA processing in Windows for volumes that do not support EA
+
+Restic was failing to backup files on some windows paths like network
+drives because of errors while fetching ExtendedAttributes.
+Either they return error codes like windows.E_NOT_SET or
+windows.ERROR_INVALID_FUNCTION or it results in slower backups.
+Restic now completely skips the attempt to fetch Extended Attributes
+for such volumes where it is not supported.
+
+https://github.com/restic/restic/pull/4980
--- a/internal/fs/ea_windows.go
+++ b/internal/fs/ea_windows.go
@@ -61,7 +61,8 @@ var (
 
 const (
 	// fileSupportsExtendedAttributes is a bitmask that indicates whether the file system supports extended attributes.
-	fileSupportsExtendedAttributes = 0x00000004
+	// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_fs_attribute_information
+	fileSupportsExtendedAttributes = 0x00800000
 )
 
 // ExtendedAttribute represents a single Windows EA.
--- a/internal/fs/ea_windows.go
+++ b/internal/fs/ea_windows.go
@@ -53,6 +53,15 @@ var (
 	errInvalidEaBuffer = errors.New("invalid extended attribute buffer")
 	errEaNameTooLarge  = errors.New("extended attribute name too large")
 	errEaValueTooLarge = errors.New("extended attribute value too large")
+
+	kernel32dll = syscall.NewLazyDLL("kernel32.dll")
+
+	procGetVolumeInformationW = kernel32dll.NewProc("GetVolumeInformationW")
+)
+
+const (
+	// fileSupportsExtendedAttributes is a bitmask that indicates whether the file system supports extended attributes.
+	fileSupportsExtendedAttributes = 0x00000004
 )
 
 // ExtendedAttribute represents a single Windows EA.
@@ -282,4 +291,35 @@ func setFileEA(handle windows.Handle, iosb *ioStatusBlock, buf *uint8, bufLen ui
 	r0, _, _ := syscall.SyscallN(procNtSetEaFile.Addr(), uintptr(handle), uintptr(unsafe.Pointer(iosb)), uintptr(unsafe.Pointer(buf)), uintptr(bufLen))
 	status = ntStatus(r0)
 	return
+}
+
+// PathSupportsExtendedAttributes returns true if the path supports extended attributes.
+func PathSupportsExtendedAttributes(path string) (bool, error) {
+	var (
+		volumeName      [syscall.MAX_PATH + 1]uint16
+		fsName          [syscall.MAX_PATH + 1]uint16
+		volumeSerial    uint32
+		maxComponentLen uint32
+		fileSystemFlags uint32
+	)
+	utf16Path, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return false, err
+	}
+	ret, _, err := procGetVolumeInformationW.Call(
+		uintptr(unsafe.Pointer(utf16Path)),
+		uintptr(unsafe.Pointer(&volumeName[0])),
+		uintptr(len(volumeName)),
+		uintptr(unsafe.Pointer(&volumeSerial)),
+		uintptr(unsafe.Pointer(&maxComponentLen)),
+		uintptr(unsafe.Pointer(&fileSystemFlags)),
+		uintptr(unsafe.Pointer(&fsName[0])),
+		uintptr(len(fsName)),
+	)
+	if ret == 0 {
+		return false, err
+	}
+
+	supportsEAs := (fileSystemFlags & fileSupportsExtendedAttributes) != 0
+	return supportsEAs, nil
 }
--- a/internal/fs/ea_windows.go
+++ b/internal/fs/ea_windows.go
@@ -53,16 +53,6 @@ var (
 	errInvalidEaBuffer = errors.New("invalid extended attribute buffer")
 	errEaNameTooLarge  = errors.New("extended attribute name too large")
 	errEaValueTooLarge = errors.New("extended attribute value too large")
-
-	kernel32dll = syscall.NewLazyDLL("kernel32.dll")
-
-	procGetVolumeInformationW = kernel32dll.NewProc("GetVolumeInformationW")
-)
-
-const (
-	// fileSupportsExtendedAttributes is a bitmask that indicates whether the file system supports extended attributes.
-	// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_fs_attribute_information
-	fileSupportsExtendedAttributes = 0x00800000
 )
 
 // ExtendedAttribute represents a single Windows EA.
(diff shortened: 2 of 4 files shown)


internal/fs/ea_windows.go
```diff
@@ -283,3 +292,34 @@ func setFileEA(handle windows.Handle, iosb *ioStatusBlock, buf *uint8, bufLen ui
...
+		uintptr(unsafe.Pointer(&maxComponentLen)),
+		uintptr(unsafe.Pointer(&fileSystemFlags)),
+		uintptr(unsafe.Pointer(&fsName[0])),
+		uintptr(len(fsName)),
+	)
+	if ret == 0 {
+		return false, err
+	}
+
+	supportsEAs := (fileSystemFlags & fileSupportsExtendedAttributes) != 0
```
aneesh-n, line 323: Need to confirm if this is the best way to check for extended attributes. Have tested positive and negative scenarios for this check.

Another option would be to check if the file system name is NTFS, which supports Extended Attributes
fsNameStr := syscall.UTF16ToString(fsName[:])
supportsEAs := fsNameStr == "NTFS"

internal/restic/node_windows.go
```diff
@@ -358,7 +361,20 @@ func (node *Node) fillGenericAttributes(path string, fi os.FileInfo, stat *statT
 		// Also do not allow processing of extended attributes for ADS.
 		return false, nil
 	}
-	if !strings.HasSuffix(filepath.Clean(path), `\`) {
+	if strings.HasSuffix(filepath.Clean(path), `\`) {
```
MichaelEischer, line 364: This approach won't work for backups using relative paths.
```
cd /D Z:   # some network share
restic backup local/file
```

internal/restic/node_windows.go
```diff
@@ -32,6 +32,9 @@ var (
 	modAdvapi32     = syscall.NewLazyDLL("advapi32.dll")
 	procEncryptFile = modAdvapi32.NewProc("EncryptFileW")
 	procDecryptFile = modAdvapi32.NewProc("DecryptFileW")
+
+	// eaUnsupportedVolumesMap is a map of volumes that do not support extended attributes.
+	eaUnsupportedVolumesMap = map[string]bool{}
```
MichaelEischer, line 37: I'm really no fan of introducing additional global state as it usually leads to problems sooner or later. However, until I've moved the NodeFromFileInfo and related code into the `internal/fs` package, I don't see a good  way to avoid it. Either way this map MUST be synchronized.

MichaelEischer (changes requested):
I guess we can give this approach a try. I've found a few problems though, see below.

internal/fs/ea_windows.go
```diff
@@ -53,6 +53,16 @@ var (
 	errInvalidEaBuffer = errors.New("invalid extended attribute buffer")
 	errEaNameTooLarge  = errors.New("extended attribute name too large")
 	errEaValueTooLarge = errors.New("extended attribute value too large")
+
+	kernel32dll = syscall.NewLazyDLL("kernel32.dll")
+
+	procGetVolumeInformationW = kernel32dll.NewProc("GetVolumeInformationW")
```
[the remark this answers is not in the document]
aneesh-n: Fixed

internal/fs/ea_windows.go
```diff
@@ -53,6 +53,16 @@ var (
...
+)
+
+const (
+	// fileSupportsExtendedAttributes is a bitmask that indicates whether the file system supports extended attributes.
+	// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_fs_attribute_information
+	fileSupportsExtendedAttributes = 0x00800000
```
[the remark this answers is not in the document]
aneesh-n: Fixed

internal/fs/ea_windows.go
```diff
@@ -283,3 +293,34 @@ func setFileEA(handle windows.Handle, iosb *ioStatusBlock, buf *uint8, bufLen ui
...
+		maxComponentLen uint32
+		fileSystemFlags uint32
+	)
+	utf16Path, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return false, err
+	}
+	ret, _, err := procGetVolumeInformationW.Call(
+		uintptr(unsafe.Pointer(utf16Path)),
+		uintptr(unsafe.Pointer(&volumeName[0])),
```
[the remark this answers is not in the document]
aneesh-n: Fixed

internal/restic/node_windows.go
aneesh-n: I agree. We can revisit later
Fixed using sync.Map

internal/restic/node_windows.go
aneesh-n: Thanks, I believe I've addressed this now.

internal/fs/ea_windows.go
```diff
@@ -283,3 +283,18 @@ func setFileEA(handle windows.Handle, iosb *ioStatusBlock, buf *uint8, bufLen ui
...
+// PathSupportsExtendedAttributes returns true if the path supports extended attributes.
+func PathSupportsExtendedAttributes(path string) (supported bool, err error) {
+	var fileSystemFlags uint32
+	utf16Path, err := windows.UTF16PtrFromString(path)
+	if err != nil {
+		return false, err
+	}
+	err = windows.GetVolumeInformation(utf16Path, nil, 0, nil, nil, &fileSystemFlags, nil, 0)
+	if err != nil {
+		return false, err
```
aneesh-n, line 296: Should we just log this as a debug log and not return the err so that it falls back to skipping EA if any issue occurs?

MichaelEischer: Let's start with returning an error and see whether problems can show up.

internal/restic/node_windows.go
```diff
@@ -351,32 +361,84 @@ func decryptFile(pathPointer *uint16) error {
...
+
+	var sd *[]byte
+	if node.Type == "file" || node.Type == "dir" {
+		// Check EA support and get security descriptor for file/dir only
+		allowExtended, err = checkAndStoreEASupport(path)
+		if err != nil {
+			return false, err
+		}
+		if sd, err = fs.GetSecurityDescriptor(path); err != nil {
+			return true, err
```
MichaelEischer, line 395: Must return `allowExtended` here instead of `true`. Otherwise, extended attributes will always be read if reading the security descriptor fails.

internal/restic/node_windows.go
```diff
@@ -351,32 +361,84 @@ func decryptFile(pathPointer *uint16) error {
...
+	}
+
+	var sd *[]byte
+	if node.Type == "file" || node.Type == "dir" {
+		// Check EA support and get security descriptor for file/dir only
+		allowExtended, err = checkAndStoreEASupport(path)
+		if err != nil {
+			return false, err
+		}
+		if sd, err = fs.GetSecurityDescriptor(path); err != nil {
```
MichaelEischer, line 394: While testing this PR, I've noticed that `GetSecurityDescriptor` can return `ERROR_NOT_SUPPORTED`. This error should be filtered out in that method:
```diff
 			if err != nil {
 				return nil, fmt.Errorf("get low-level named security info failed with: %w", err)
 			}
+		} else if errors.Is(err, windows.ERROR_NOT_SUPPORTED) {
+			return nil, nil
 		} else {
 			return nil, fmt.Errorf("get named security info failed with: %w", err)
 		}
```

MichaelEischer (changes requested):
This seems to almost work, see the comments for the remaining problems.

changelog/unreleased/pull-4980
```diff
@@ -0,0 +1,10 @@
+Bugfix: Skip EA processing in Windows for volumes that do not support EA
+
+Restic was failing to backup files on some windows paths like network
+drives because of errors while fetching ExtendedAttributes.
+Either they return error codes like windows.E_NOT_SET or
+windows.ERROR_INVALID_FUNCTION or it results in slower backups.
+Restic now completely skips the attempt to fetch Extended Attributes
+for such volumes where it is not supported.
+
+https://github.com/restic/restic/pull/4980
```
MichaelEischer, line 10: Please link to both issues that are solved by this PR.

MichaelEischer (approved):
LGTM. Thanks!

internal/restic/node_windows.go
```diff
@@ -351,32 +361,84 @@ func decryptFile(pathPointer *uint16) error {
...
+func checkAndStoreEASupport(path string) (isEASupportedVolume bool, err error) {
+	// Check if it's an extended length path
+	if strings.HasPrefix(path, uncPathPrefix) {
+		// Convert \\?\UNC\ extended path to standard path to get the volume name correctly
+		path = `\\` + path[len(uncPathPrefix):]
+	} else if strings.HasPrefix(path, extendedPathPrefix) {
+		//Extended length path prefix needs to be trimmed to get the volume name correctly
+		path = path[len(extendedPathPrefix):]
+	} else if strings.HasPrefix(path, globalRootPrefix) {
+		// EAs are not supported for \\?\GLOBALROOT i.e. VSS snapshots
```
aneesh-n, line 418: I made a mistake here while fixing the test case errors. 
Extended attributes are in fact supported for VSS snapshots.
Instead of returning false, we should be handling the volume names for VSS snapshots correctly. Apologies, I can add a separate PR.

MichaelEischer: > Instead of returning false, we should be handling the volume names for VSS snapshots correctly. Apologies, I can add a separate PR.

Does `GetVolumeInformation` support such paths or do we needs even more weird special case handling?

aneesh-n: We need to get the 'volumeName' path like this - 
        
```
// Extract the VSS snapshot volume name eg. `\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopyXX`
if parts := strings.SplitN(path, `\`, 7); len(parts) >= 6 {
	volumeName = strings.Join(parts[:6], `\`)
}
```
		
Rest of the code remains same-

fs.PathSupportsExtendedAttributes(volumeName + `\`) i.e. GetVolumeInformation works.

MichaelEischer: Sigh, I'm not happy about all those Windows special case. But looks like we'd better add them. Can you open a PR for it?

aneesh-n: Yeah, problem is filepath.VolumeName does not do any windows specific handling, it's just syntactical. 
We can probably get actual volumeName using windows.GetVolumeInformation itself, but it would mean we need to make the system call for every path. 
Right now, we are doing all that manual processing to avoid the system call.

PR #4981, opened 2024-08-04
cache: clear snapshot files from cache during load index

### LINKED ISSUE #4934: Cleanup removed snapshots from cache
Output of `restic version`
--------------------------
0.16.5

When using a restic repository from multiple host, then after running `forget` on one of the hosts, the removed snapshots still exist in the cache of the other hosts. Restic should remove those snapshots from the cache of all hosts.

It would be sufficient, if for example the commands that list all snapshots, also automatically clean up the cache.
For that it might be possible to hook into the `List` function of a `Repository` to remove stale snapshots from the cache.

Keep the cache size small.

konidev20:
clear snapshot files from cache during repository preparation. 

There is a function in `repository.go` called `prepareCache()` which is used to clean up old index and pack files from the local cache. This method is called during `LoadIndex`.

Closes #4934

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4934
+++ b/changelog/unreleased/issue-4934
@@ -1,9 +1,8 @@
-Enhancement: Clear removed snapshots, index and pack files from the local cache
+Enhancement: Clear removed snapshots from local cache of the current host
 
-Restic did not clear removed snapshots from the cache after the `forget` 
-operation; only indexes and pack files were removed automatically.  
-Restic now automatically clears removed indexes, packs and snapshots from the
-local cache.
+Restic only removed snapshots from the cache on the host that runs the `forget` command.
+On other hosts that use the same repository, the old snapshots remained in the cache.
+Restic now, automatically clears old snapshots from the local cache of the current host.
 
 https://github.com/restic/restic/issues/4934
 https://github.com/restic/restic/pull/4981
--- a/changelog/unreleased/issue-4934
+++ b/changelog/unreleased/issue-4934
@@ -1,0 +1,9 @@
+Enhancement: Clear removed snapshots, index and pack files from the local cache
+
+Restic did not clear removed snapshots from the cache after the `forget` 
+operation; only indexes and pack files were removed automatically.  
+Restic now automatically clears removed indexes, packs and snapshots from the
+local cache.
+
+https://github.com/restic/restic/issues/4934
+https://github.com/restic/restic/pull/4981
--- a/internal/backend/cache/backend.go
+++ b/internal/backend/cache/backend.go
@@ -247,6 +247,10 @@ func (b *Backend) List(ctx context.Context, t backend.FileType, fn func(f backen
 		return err
 	}
 
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
 	// clear the cache for files that are not in the repo anymore, ignore errors
 	err = b.Cache.Clear(t, ids)
 	if err != nil {
--- a/internal/backend/cache/backend.go
+++ b/internal/backend/cache/backend.go
@@ -2,11 +2,14 @@ package cache
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"os"
 	"sync"
 
 	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/restic"
 )
 
 // Backend wraps a restic.Backend and adds a cache.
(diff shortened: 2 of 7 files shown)


internal/repository/repository.go
```diff
@@ -723,6 +723,21 @@ func (r *Repository) prepareCache() error {
...
+		return nil
+	})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error listing snapshots: %v\n", err)
+	}
+
+	err = r.Cache.Clear(restic.SnapshotFile, snapshots)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error clearing snapshot files in cache: %v\n", err)
+	}
```
konidev20, line 739: I realised I can do the same thing, where ever we are listing the snapshots, as mentioned in the original issue.

MichaelEischer: Explicitly listing the snapshots is somewhat cleaner, but requires an extra List call that must be avoided.

I can imagine two implementation variants: either hijack Repository.List or maybe better add a List method to the cache backend and let it take care of cleaning up indexes and snapshots itself. The latter variant has the benefit of keeping most of the cache management within the cache itself.

MichaelEischer (changes requested):
I don't want to introduce an extra List call to backends just to clean up old snapshots.

internal/backend/cache/backend.go
```diff
@@ -215,3 +218,40 @@ func (b *Backend) IsNotExist(err error) bool {
...
+		return fn(f)
+	}
+
+	err := b.Backend.List(ctx, t, wrapFn)
+	if err != nil {
+		return err
+	}
+
+	// clear the cache for files that are not in the repo anymore, ignore errors
+	err = b.Cache.Clear(t, ids)
```
MichaelEischer, line 251: Please check that the `ctx.Err() != nil` before clearing the cache.

internal/backend/cache/backend.go
```diff
@@ -215,3 +218,40 @@ func (b *Backend) IsNotExist(err error) bool {
 func (b *Backend) Unwrap() backend.Backend {
 	return b.Backend
 }
+
+func (b *Backend) List(ctx context.Context, t backend.FileType, fn func(f backend.FileInfo) error) error {
```
MichaelEischer, line 222: This change makes the `err := r.Cache.Clear(restic.IndexFile, indexIDs)` call in `internal/repository/repository.go` redundant.

MichaelEischer (commented):
The approach looks fine in general, I just have a few small remarks on it. Other than that only a test and changelog entry are missing.

internal/repository/repository.go
```diff
@@ -699,33 +693,6 @@ func (r *Repository) createIndexFromPacks(ctx context.Context, packsize map[rest
...
-	}
-
-	indexIDs := r.idx.IDs()
-	debug.Log("prepare cache with %d index files", len(indexIDs))
-
-	// clear old index files
-	err := r.Cache.Clear(restic.IndexFile, indexIDs)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "error clearing index files in cache: %v\n", err)
-	}
```
MichaelEischer, line 716: why did you remove the whole `prepareCache` function? All pack files are only listed by `check` or `prune`, so we can't remove the step that uses the index to determine which pack files should still be kept. Please revert those changes and _only_ remove the code between line 709 and 716.

changelog/unreleased/issue-4934
```diff
@@ -0,0 +1,9 @@
+Enhancement: Clear removed snapshots, index and pack files from the local cache
```
MichaelEischer, line 1: `index` and `pack` files already get removed from the cache before this PR. Thus the should not be mentioned in the changelog entry.

changelog/unreleased/issue-4934
```diff
@@ -0,0 +1,9 @@
...
+
+Restic did not clear removed snapshots from the cache after the `forget`
```
MichaelEischer, line 3: That remark is too broad. `forget` removes snapshots from the cache, but only on the current host. Please write something like the following:

```
Restic only removed snapshots from the cache on the host that runs the `forget` command.
On other hosts that use the same repository, the old snapshots remained in the cache.

Now, restic automatically clears old snapshots.
```

MichaelEischer (changes requested):

internal/repository/repository.go
konidev20: I have reverted this change. I have removed *only* lines between 709 and 716.

MichaelEischer (approved):
LGTM. Thanks!

Issue #4795, opened 2024-05-04
restic --verbose restore --verify prints no progress for a long time at the end of a restore


jjlee:
Output of `restic version`
--------------------------

`restic 0.16.4 compiled with go1.21.6 on linux/amd64`

I'm running on arch linux

What backend/service did you use to store the repository?
---------------------------------------------------------

Local filesystem.

Problem description / Steps to reproduce
----------------------------------------

restic restore with the `--verify` switch prints `verifying files in ...` and then a % estimate.  After the estimate reaches 100%, restic may continue to execute for a long time (tens of minutes for the example below).

Expected behavior
-----------------

restic doesn't:
- run for a long time without some progress update
- present a verification progress indicator at 100% when there is still ~10s of minutes of run time remaining

does:
- always make it clear what phase of operation is executing

Actual behavior
---------------
```
ionice -c3 restic -r /path/to/restic/repo --verbose restore latest --verify --target /path/to/restore/space
```
I see output like this at the point it stops printing updates for tens of minutes:
```
repository 4ae3f911 opened (version 2, compression level auto)
found 32 old cache directories in /home/me/.cache/restic, run `restic cache --cleanup` to remove them
[0:02] 100.00%  29 / 29 index files loaded
restoring <Snapshot 459c74a3 of [/path/to/restic/repo] at 2024-01-05 20:23:32.433605071 +0000 UTC by root@host> to /path/to/restore/space
Summary: Restored 6187417 files/dirs (223.845 GiB) in 1:33:17
verifying files in /path/to/restore/space
[1:35:17] 100.00%  6187301 files/dirs 223.845 GiB, total 6187417 files/dirs 223.845 GiB
```
There is still a lot of disk activity during the time when it is no longer printing any updates.

Then, a long time later, it prints:

```
finished verifying 5156840 files in /path/to/restore/space (took 1h16m22.147s)
```

Do you have any idea what may have caused this?
-----------------------------------------------

No.  I'm not sure what restic is doing in this time.

I don't see any other obvious processes like tracker or mlocate running that might be reading or writing the directories involved.  `lsof` run on the source and target directories doesn't show anything.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

restic is pleasantly simple to get started with :-)

MichaelEischer:
> Summary: Restored 6187417 files/dirs (223.845 GiB) in 1:33:17
> verifying files in /path/to/restore/space
[...]

PR #4989, opened 2024-08-08
restore: Add progress bar to 'restore --verify'

plant99:

This MR introduces a progress bar to the `restic restore --verify` flow when verify is in-progress. Earlier, there was no output when files were being verified.

https://github.com/restic/restic/issues/4795

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4795
+++ b/changelog/unreleased/issue-4795
@@ -1,0 +1,7 @@
+Enhancement: `restore --verify` shows progress with a progress bar
+
+If restore command was run with `--verify` restic didn't show any progress indication, now it shows a progress bar while 'verification' is running.
+The progress bar is text only for now and doesn't respect `--json` flag.
+
+https://github.com/restic/restic/issues/4795
+https://github.com/restic/restic/pull/4989
--- a/internal/restorer/restorer.go
+++ b/internal/restorer/restorer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/ui/progress"
 	restoreui "github.com/restic/restic/internal/ui/restore"
 
 	"golang.org/x/sync/errgroup"
@@ -333,12 +334,13 @@ func (res *Restorer) ensureDir(target string) error {
 
 // RestoreTo creates the directories and files in the snapshot below dst.
 // Before an item is created, res.Filter is called.
-func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
+func (res *Restorer) RestoreTo(ctx context.Context, dst string) (uint64, error) {
+	restoredFileCount := uint64(0)
 	var err error
 	if !filepath.IsAbs(dst) {
 		dst, err = filepath.Abs(dst)
 		if err != nil {
-			return errors.Wrap(err, "Abs")
+			return restoredFileCount, errors.Wrap(err, "Abs")
 		}
 	}
 
@@ -346,7 +348,7 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 		// ensure that the target directory exists and is actually a directory
 		// Using ensureDir is too aggressive here as it also removes unexpected files
 		if err := fs.MkdirAll(dst, 0700); err != nil {
-			return fmt.Errorf("cannot create target directory: %w", err)
+			return restoredFileCount, fmt.Errorf("cannot create target directory: %w", err)
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


plant99:
@MichaelEischer please review :) 

Before you do, please advise on the following:

1. I tried to implement it the way `LoadIndex` shows progress.
2. I think the actual `restore` progress bar which uses `restoreui.NewProgress` is better, and better compatible with the `--json` flag.
3. Do you think the way I've implemented now is enough, or should I try doing 2? The change if involving `restoreui.NewProgress` would have a larger footprint, and thus I have avoided it for now.

internal/restorer/restorer.go
```diff
@@ -587,17 +588,24 @@ const nVerifyWorkers = 8
...
 	var (
-		nchecked uint64
-		work     = make(chan mustCheck, 2*nVerifyWorkers)
+		nchecked         uint64
+		numFilesVerified uint64
+		work             = make(chan mustCheck, 2*nVerifyWorkers)
 	)
 
+	if p != nil {
+		numFilesVerified = uint64(res.sn.Summary.TotalFilesProcessed)
```
MichaelEischer, line 604: Please move the declaration of the variable here. It's not used anywhere else.

cmd/restic/progress.go
```diff
@@ -110,6 +110,10 @@ func newIndexTerminalProgress(quiet bool, json bool, term *termstatus.Terminal)
 	return newTerminalProgressMax(!quiet && !json && stdoutIsTerminal(), 0, "index files loaded", term)
 }
 
+func newVerifyTerminalProgress(quiet bool, json bool, term *termstatus.Terminal) *progress.Counter {
```
MichaelEischer, line 113: no need for a wrapper function here. Just call `newTerminalProgressMax` directly from `cmd_restore`.

changelog/unreleased/issue-4795
```diff
@@ -0,0 +1,6 @@
+Enhancement: `restore --verify` shows progress with a progress bar
+
+If restore command was run with `--verify` restic didn't show any progress indication, now it shows a progress bar while 'verification' is running.
```
MichaelEischer, line 3: The changelog should mention that the progress bar is text output only so far.

MichaelEischer (commented):
> 3\. Do you think the way I've implemented now is enough, or should I try doing 2? The change if involving `restoreui.NewProgress` would have a larger footprint, and thus I have avoided it for now.

The answer depends on whether we want the progress bar to also work for JSON or not. For just text output the current approach would be good enough. 

To add JSON support, we essentially have two options. We could extend `newTerminalProgressMax` to actually support JSON progress bars (maybe with a format similar to `progress_percent` from [borgbackup](https://borgbackup.readthedocs.io/en/stable/internals/frontends.html#logging)). The other option would be to extend the restore progress UI. But that has the big downside of requiring yet another handcrafted progress implementation.

So I'm leaning towards starting with `newTerminalProgressMax` and separately adding JSON support.

plant99:
Thanks @MichaelEischer  for the review and the feedback on the implementation
1. I've made the changes you suggested in this PR
2. I'll also make a draft on adding JSON support to `newTerminalProgressMax`, I'll make an issue about this soon.

internal/restorer/restorer.go
```diff
@@ -598,6 +599,12 @@ func (res *Restorer) VerifyFiles(ctx context.Context, dst string) (int, error) {
 		work     = make(chan mustCheck, 2*nVerifyWorkers)
 	)
 
+	if p != nil {
+		numFilesVerified := uint64(res.sn.Summary.TotalFilesProcessed)
```
MichaelEischer, line 603: I just had another look and we'll need a different way to determine the number of files to verify. Just restore the same snapshot twice, then you'll end up with something like the following:
```
verifying files in output
[0:00] 0.00%  0 / 1 files verified
finished verifying 0 files in output (took 0s)
```

This is rather confusing as the progress will never reach 100% in most cases. Or even worse, just restore an old snapshot without that metadata and restic would just crash.

What could work is the following. In `RestoreTo` there's a `res.trackFile(location, updateMetadataOnly)` call. At that place the restorer could also count all files that are not `updateMetadataOnly` updates. Then `RestoreTo` could return that number similar to `VerifyFiles`.

That will also remove the hacks that were necessary to make the tests work.

MichaelEischer (changes requested):
I just had another look at how the maximum for the progress bar is determined and noticed that the current approach can't work, see below for details. No idea how I've missed that yesterday.

internal/restorer/restorer.go
plant99: Thanks for the catching this, and the guidance!

I've updated the code as per your suggestions.

Please review again when you have the time :)

internal/restorer/restorer.go
```diff
@@ -406,19 +408,22 @@ func (res *Restorer) RestoreTo(ctx context.Context, dst string) error {
 					}
 				}
 				res.trackFile(location, updateMetadataOnly)
+				if !updateMetadataOnly {
+					restoredFileCount++
```
plant99, line 412: @MichaelEischer I didn't see any thread/goroutines within RestoreTo.

Do you think I should make it concurrency safe?

MichaelEischer: `traverseTree` is single threaded. Thus there's no need to be concurrency safe for now.

MichaelEischer (approved):
LGTM. Thanks!

Issue #4155, opened 2023-01-14
Windows - Reparse points / cloud files creating a lot of errors in logs


deajan:
Output of `restic version`
--------------------------
`restic 0.14.0 compiled with go1.19 on windows/amd64`

How did you run restic exactly?
-------------------------------
Target machine: Windows 10 21H2 x64 with local only NTFS filesystem.

`"restic.exe" backup "c:\Users" --iexclude-file excludes/generic_excluded_extensions --iexclude-file excludes/generic_excludes --iexclude-file excludes/windows_excludes --exclude-caches --use-fs-snapshot`

I've ran restic as Administrator and as System, both produced the same results.

I've filtered the logs since there are thousands of files, and left a couple of lines per error type:
```
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\DOSSIERS ARCHIVES UTILE\réglement intérieur\projet RI.DOC: The media is write protected.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\DOSSIERS ARCHIVES UTILE\réglement intérieur\réglement intérieur modèle.DOC: The media is write protected.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\FINANCIIER\factures\Fact 1981.doc: The cloud operation is not supported on a read-only volume.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\FINANCIIER\factures\Fact 1982- 66.doc: The cloud operation is not supported on a read-only volume.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\FINANCIIER\factures\Fact 1983 - .pdf: The cloud operation is not supported on a read-only volume.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\FINANCIIER\factures\Fact 1984- LIFT.doc: The cloud operation is not supported on a read-only volume.
error: read \\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy6\Users\User\Documents\foo\bar\FINANCIIER\factures\Fact 1985_FRUITS.doc: The cloud operation is not supported on a read-only volume.
[...]

PR #4990, opened 2024-08-08
backup: allow excluding online-only cloud files

### LINKED ISSUE #3697: Ignore Cloud files which are not synced locally
Output of `restic version`
--------------------------
restic 0.13.0 compiled with go1.18 on linux/amd64

I am using the cloud every now and then, and when I am backing up my Windows machine, I get the following error if the file is not present locally:

```
{'during': 'archival',
'error': {'Err': 381,
'Op': 'read',
'Path': '\\\\?\\GLOBALROOT\\Device\\HarddiskVolumeShadowCopy11\\Users\\flo\\OneDrive\\Bilder\\xyz.jpg'},
'item': 'c:\\Users\\flo\\OneDrive\\Bilder\\xyz.jpg',
'message_type': 'error'}
```

Err 381 seems to be `ERROR_CLOUD_FILE_READ_ONLY_VOLUME`. As you can see, I am using VSS, so this error makes sense. Windows can not download and store this file on a shadow copy. I would go so far and say, Windows (triggered by restic) should never download those files - even if backing up without shadow copy.

Therefore I propose the following:
- Never try to backup files that are not locally present.
  I guess, at the moment, restic will trigger a download of all files which are not locally available during backup (if not using VSS).
  I think this is wrong behavior. A backup program should not make such large modifications to the state of the system being backed up.
- Maybe add a switch like `--force-download-cloud-files` which will activate this behavior explicitly,
  as of course there might be people that want to backup their cloud files, too. (Maybe even myself, but it should be done explicitly.)
- Abort with an error message if this switch and vss are used together.

  

Backing up cloud synced directories without errors.

It helps me everyday since months!

m-wild:

Allow excluding online-only cloud files (such as OneDrive Files On-Demand) with a switch `--exclude-cloud-files`

Assuming we have the following file `test.txt` in OneDrive, but not available on the device

image

image

Attempting to backup the file with the `--exclude-cloud-files` switch will not attempt to download the file locally, instead the file will be excluded.

```sh
restic -r test --verbose backup C:\Users\username\OneDrive\test.txt --exclude-cloud-files

open repository
repository e5f7dfde opened (version 2, compression level auto)
created new cache in C:\Users\username\AppData\Local\restic
load index files
start scan on [C:\Users\username\OneDrive\test.txt]
start backup on [C:\Users\username\OneDrive\test.txt]
scan finished in 0.003s: 0 files, 0 B

Data Blobs:      0 new
Tree Blobs:      5 new
Added to the repository: 2.425 KiB (1.841 KiB stored)

snapshot 5ef1ff96 saved

```

```log
2024/08/08 22:39:41 archiver/tree.go:298	archiver.NewTree	1	result:
/C, root "C:\\", path "", meta "C:\\"
    /Users, root "", path "", meta "C:\\Users"
        /username, root "", path "", meta "C:\\Users\\username"
            /OneDrive, root "", path "", meta "C:\\Users\\username\\OneDrive"
                /test.txt, root "", path "C:\\Users\\username\\OneDrive\\test.txt", meta ""

...

2024/08/08 22:39:44 archiver/archiver.go:420	archiver.(*Archiver).save	40	/C/Users/username/OneDrive/test.txt target "C:\\Users\\username\\OneDrive\\test.txt", previous <nil>
2024/08/08 22:39:44 restic/exclude.go:404	main.collectRejectFuncs.rejectCloudFiles.func1	40	rejecting online-only cloud file C:\Users\username\OneDrive\test.txt
2024/08/08 22:39:44 archiver/archiver.go:443	archiver.(*Archiver).save	40	C:\Users\username\OneDrive\test.txt is excluded

```

Closes #3697

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3697
+++ b/changelog/unreleased/issue-3697
@@ -1,0 +1,12 @@
+Enhancement: Allow excluding online-only cloud files (e.g. OneDrive)
+
+Restic treated OneDrive Files On-Demand as though they were regular files
+for the purpose of backup which caused issues with VSS, could make backup
+incredibly slow (as OneDrive attempted to download files), or could fill
+the source disk (e.g. 1TB of files in OneDrive on a 500GB disk).
+Restic now allows the user to exclude these files when backing up with
+the `--exclude-cloud-files` switch.
+
+https://github.com/restic/restic/issues/3697
+https://github.com/restic/restic/issues/4935
+https://github.com/restic/restic/pull/4990
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -77,6 +77,7 @@ type BackupOptions struct {
 	ExcludeIfPresent  []string
 	ExcludeCaches     bool
 	ExcludeLargerThan string
+	ExcludeCloudFiles bool
 	Stdin             bool
 	StdinFilename     string
 	StdinCommand      bool
@@ -140,6 +141,7 @@ func init() {
 	f.BoolVar(&backupOptions.NoScan, "no-scan", false, "do not run scanner to estimate size of backup")
 	if runtime.GOOS == "windows" {
 		f.BoolVar(&backupOptions.UseFsSnapshot, "use-fs-snapshot", false, "use filesystem snapshot where possible (currently only Windows VSS)")
+		f.BoolVar(&backupOptions.ExcludeCloudFiles, "exclude-cloud-files", false, "excludes online-only cloud files (such as OneDrive Files On-Demand)")
 	}
 	f.BoolVar(&backupOptions.SkipIfUnchanged, "skip-if-unchanged", false, "skip snapshot creation if identical to parent snapshot")
 
@@ -341,6 +343,17 @@ func collectRejectFuncs(opts BackupOptions, targets []string, fs fs.FS) (funcs [
 		}
 
 		f, err := archiver.RejectBySize(maxSize)
+		if err != nil {
+			return nil, err
+		}
+		funcs = append(funcs, f)
+	}
+
+	if opts.ExcludeCloudFiles && !opts.Stdin && !opts.StdinCommand {
+		if runtime.GOOS != "windows" {
+			return nil, errors.Fatalf("exclude-cloud-files is only supported on Windows")
+		}
+		f, err := archiver.RejectCloudFiles(Warnf)
 		if err != nil {
 			return nil, err
 		}
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -297,7 +297,8 @@ the exclude options are:
 -  ``--exclude-file`` Specified one or more times to exclude items listed in a given file
 -  ``--iexclude-file`` Same as ``exclude-file`` but ignores cases like in ``--iexclude``
 -  ``--exclude-if-present foo`` Specified one or more times to exclude a folder's content if it contains a file called ``foo`` (optionally having a given header, no wildcards for the file name supported)
--  ``--exclude-larger-than size`` Specified once to excludes files larger than the given size
+-  ``--exclude-larger-than size`` Specified once to exclude files larger than the given size
+-  ``--exclude-cloud-files`` Specified once to exclude online-only cloud files (such as OneDrive Files On-Demand), currently only supported on Windows
 
 Please see ``restic help backup`` for more specific information about each exclude option.
 
--- a/internal/archiver/exclude.go
+++ b/internal/archiver/exclude.go
@@ -316,3 +316,21 @@ func RejectBySize(maxSize int64) (RejectFunc, error) {
 		return false
 	}, nil
 }
+
+// RejectCloudFiles returns a func which rejects files which are online-only cloud files
+func RejectCloudFiles(warnf func(msg string, args ...interface{})) (RejectFunc, error) {
+	return func(item string, fi *fs.ExtendedFileInfo, _ fs.FS) bool {
+		recall, err := fi.RecallOnDataAccess()
+		if err != nil {
+			warnf("item %v: error checking online-only status: %v", item, err)
+			return false
+		}
+
+		if recall {
+			debug.Log("rejecting online-only cloud file %s", item)
+			return true
+		}
+
+		return false
+	}, nil
+}
--- a/internal/fs/stat_bsd.go
+++ b/internal/fs/stat_bsd.go
@@ -32,3 +32,8 @@ func extendedStat(fi os.FileInfo) *ExtendedFileInfo {
 		ChangeTime: time.Unix(s.Ctimespec.Unix()),
 	}
 }
+
+// RecallOnDataAccess checks windows-specific attributes to determine if a file is a cloud-only placeholder.
+func (*ExtendedFileInfo) RecallOnDataAccess() (bool, error) {
+	return false, nil
+}
--- a/internal/fs/stat_unix.go
+++ b/internal/fs/stat_unix.go
@@ -32,3 +32,8 @@ func extendedStat(fi os.FileInfo) *ExtendedFileInfo {
 		ChangeTime: time.Unix(s.Ctim.Unix()),
 	}
 }
+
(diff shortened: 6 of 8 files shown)


cmd/restic/exclude.go
```diff
@@ -385,6 +387,28 @@ func rejectBySize(maxSizeStr string) (RejectFunc, error) {
...
 
+// rejectCloudFiles returns a func which rejects files which are online-only cloud files
+func rejectCloudFiles() (RejectFunc, error) {
+	if runtime.GOOS != "windows" {
+		return nil, errors.Fatalf("exclude-cloud-files is only supported on Windows")
+	}
+
+	const (
+		FILE_ATTRIBUTE_UNPINNED              = 0x00100000
+		FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS = 0x00400000
```
m-wild, line 398: Unfortunately these are [not defined in syscall](https://pkg.go.dev/syscall?GOOS=windows#pkg-constants), probably due to their uncommon usage.

They are documented by Microsoft here [File Attribute Constants (WinNT.h)](https://learn.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants).

m-wild:
@MichaelEischer I'm not really sure how to add tests for this one. It would require either manually setting the file attributes or somehow setting up OneDrive on Windows?

cmd/restic/exclude.go
MichaelEischer: The constants are available in https://pkg.go.dev/golang.org/x/sys/windows .

cmd/restic/cmd_backup.go
```diff
@@ -348,6 +350,14 @@ func collectRejectFuncs(opts BackupOptions, targets []string) (fs []RejectFunc,
 		fs = append(fs, f)
 	}
 
+	if opts.ExcludeCloudFiles && !opts.Stdin {
```
MichaelEischer, line 353: Let's move the windows check here. That simplifies the error handling.

cmd/restic/exclude.go
```diff
@@ -385,6 +387,33 @@ func rejectBySize(maxSizeStr string) (RejectFunc, error) {
...
+	)
+
+	return func(item string, fi os.FileInfo) bool {
+		attrs, ok := fi.Sys().(*syscall.Win32FileAttributeData)
```
MichaelEischer, line 402: Please move the fs specific code into the `internal/fs` package (maybe `internal/fs/file_windows.go`) and add a minimal unit test there. I'd like to minimize new code in `cmd/restic` as much as reasonable.

MichaelEischer (changes requested):

deajan:
@m-wild I really love that PR. That's exactly what I'm searching for more than a year at #4155 
I'll happily make some tests with some beta builds if needed.

greatroar:
There's already an option --one-file-system. Can this feature use that option, instead of introducing a new one? Or does it already have a meaning on Windows?

MichaelEischer:
> There's already an option --one-file-system. Can this feature use that option, instead of introducing a new one? Or does it already have a meaning on Windows?

`--one-file-system` currently has no effect on Windows. However, Windows does support junctions which can be used to mount other filesystems within a filesystem. My expectation would be that `--one-file-system` once it has an effect, addresses exactly this scenario.

Cloud files feel somewhat different: they are placeholders for files that can basically show up everywhere in the filesystem. I wouldn't really expect that `--one-file-system` excludes those files.

cmd/restic/exclude.go
MichaelEischer: The unit test will probably have to use a bit of mocking to create an os.FileInfo that pretends to be a cloud file, but shouldn't be too complex to implement.

deajan:
@m-wild Any news on this one perhaps ?

MichaelEischer:
If someone is interested in taking over this PR, please just create a new PR, then I'll close this one.

deajan:
@m-wild @MichaelEischer I'll happily sponsor whoever wants to pick this up ;)

deajan:
@m-wild Do you still want to work on this ?

m-wild:
> @m-wild Do you still want to work on this ?

Planning to pick it up over the Xmas break

cmd/restic/exclude.go
m-wild: I've given this a go, but I'm not 100% sure it's right, I've never done multi-targeting with go before.

I believe I had to create a stub implementation in `internal/fs/file_unix.go`, then the real implementation in `internal/fs/file_windows.go`.

I've added 3 tests
1. Test with a real temp file that the function doesn't err
2. Mock cloud file
3. Mock regular file

m-wild:
Ho ho ho, it's a Christmas miracle, a PR ready to review.

deajan:
> Ho ho ho, it's a Christmas miracle, a PR ready to review.

Ho ho ho, ping me whenever this is merged so I can sponsor you a bit ;)

m-wild:
Have just fixed the lint issues.

internal/fs/file_windows.go
```diff
@@ -134,3 +136,20 @@ func openHandleForEA(nodeType restic.NodeType, path string, writeAccess bool) (h
 	}
 	return handle, err
 }
+
+// RecallOnDataAccess checks if a file is available locally on the disk or if the file is
+// just a placeholder which must be downloaded from a remote server. This is typically used
+// in cloud syncing services (e.g. OneDrive) to prevent downloading files from cloud storage
+// until they are accessed.
+func RecallOnDataAccess(fi *ExtendedFileInfo) (bool, error) {
```
MichaelEischer, line 144: Please move that function to the `stat_*.go` files and change its signature to `func (fi *ExtendedFileInfo) RecallOnDataAccess() (bool, error)`.

internal/fs/file_windows_test.go
```diff
@@ -33,3 +38,72 @@ func TestTempFile(t *testing.T) {
 	_, err = os.Stat(fn2)
 	rtest.Assert(t, errors.Is(err, os.ErrNotExist), "err %s", err)
 }
+
+func TestRecallOnDataAccessRealFile(t *testing.T) {
+	// create a temp file for testing
+	tempdir := rtest.TempDir(t)
+	filename := filepath.Join(tempdir, "regular-file")
+	err := os.WriteFile(filename, []byte("foobar"), 0640)
+	if err != nil {
```
MichaelEischer, line 47: Please use `rtest.OK(err)` here

changelog/unreleased/issue-3697
```diff
@@ -0,0 +1,11 @@
...
+
+Restic treated OneDrive Files On-Demand as though they were regular files
+for the purpose of backup which caused issues with VSS, could make backup
+incredibly slow (as OneDrive attempted to download files), or could fill
+the source disk (e.g. 1TB of files in OneDrive on a 500GB disk).
+Restic now allows the user to exclude these files when backing up with
+the `--exclude-cloud-files` switch.
+
+https://github.com/restic/restic/issues/3697
+https://github.com/restic/restic/issues/4935
```
MichaelEischer, line 11: Add a link to this PR here

MichaelEischer (changes requested):
Thanks for updating the PR! I'm mostly happy with it except for a few nits.

internal/fs/file_windows.go
MichaelEischer: (This will require duplicating the RecallOnDataAccess placeholder for both `stat_unix` and `stat_bsd`)

MichaelEischer (approved):
LGTM.

MichaelEischer:
(Force pushed to fix the linter warning)

deajan:
Can confirm that the  new option `--exclude-cloud-files` works great with a test nextcloud setup using current restic beta binary.
Thank you for that great work.

@m-wild As I suggested earlier, I'd like to sponsor you a bit. Do you have any donation system setup perhaps ?

m-wild:
@deajan thanks for testing the beta :) I have GitHub sponsorship enabled.

deajan:
@m-wild Done ;) Thank you again.

Issue #5005, opened 2024-08-15
Intermittent - unable to create lock in backend: circuit breaker open for file <lock/XXXXXXXXXX>


heapdavid:
Output of `restic version`
--------------------------
restic 0.17.0 compiled with go1.22.5 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
Backblaze B2

Problem description / Steps to reproduce
----------------------------------------

Hi,

We upload lots of large files to B2 on a nightly basis. Since upgrading to 0.17.0 on Debian, we're seeing occasional errors uploading to B2:

```
unable to create lock in backend: circuit breaker open for file <lock/c498f551d0>
bytes to write=10240, bytes written=4096, total written         60K
buffer (writer): write of data failed: Success
Unable to write to output file ()
```

We pipe tarballs into restic using the following command inside a script:

`... | restic backup --quiet --stdin --stdin-filename "${TARBALL}.tar" -o b2.connections=8`

If we set `RESTIC_FEATURES=backend-error-redesign=false` as mentioned in another issue then we don't see this issue - uploads happen reliably again.

Expected behavior
-----------------

Upload to B2 reliably

Actual behavior
---------------

As described above, 

Do you have any idea what may have caused this?
-----------------------------------------------
Given that setting `RESTIC_FEATURES=backend-error-redesign=false` works around this issue, it's likely to be something in that redesign.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Restic is a handy part of our backup regime, and has been reliable and straightforward to use. Thank you!

MichaelEischer:
> Unable to write to output file ()

The error messages seem to be somewhat mangled. This error does not exist in restic.

> unable to create lock in backend: circuit breaker open for file <lock/c498f551d0>

Did restic log any errors before that (to stderr)? It looks like restic failed to access some lock files. I'll need those errors to know what the underlying problem is. The relevant log entries are probably prefixed with `Load(` and contain ` failed: `.

> We pipe tarballs into restic using the following command inside a script:
> 
> `... | restic backup --quiet --stdin --stdin-filename "${TARBALL}.tar" -o b2.connections=8`

Unrelated to the issue here: you might want to take a look at the new --stdin-from-command option, see https://restic.readthedocs.io/en/latest/040_backup.html#reading-data-from-a-command .

heapdavid:
> The error messages seem to be somewhat mangled. This error does not exist in restic.

[...]

PR #5011, opened 2024-08-22
backend/retry: don't trip circuit breaker if context is canceled

MichaelEischer:

When the context used for a load operation is canceled, then the result is always an error independent of whether the file could be retrieved from the backend. Do not false positively trip the circuit breaker in this case.

The old behavior was problematic when trying to lock a repository. When `Lock.checkForOtherLocks` listed multiple lock files in parallel and one of them fails to load, then all other loads were canceled. This cancelation was remembered by the circuit breaker, such that locking retries would fail.

Solves the unexpected "circuit breaker open for file" errors in https://github.com/restic/restic/issues/5005#issuecomment-2301713213 .

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5011
+++ b/changelog/unreleased/pull-5011
@@ -1,0 +1,10 @@
+Bugfix: Fix rare failures to retry locking a repository
+
+Restic 0.17.0 could in rare cases fail to retry locking a repository if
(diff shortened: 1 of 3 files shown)


MichaelEischer (commented):
LGTM.

PR #5046, opened 2024-09-08
azure: add support for access tiers hot, cool and cold

### LINKED ISSUE #4521: Microsoft Azure Blob Storage support access tier
Output of `restic version`
--------------------------

restic 0.16.0 compiled with go1.20.6 on darwin/amd64

Hi can restic add support for access tier (hot, cool, cold and archive)
https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview

So we can implement different access tier for different kinds of backup.

Quite happy use restic 👍

konidev20:
This commit adds functionality to control the access tier for Azure blob storage backend. This should support Online Access Tiers including Hot, Cool and Cold. 

Only Data Pack files which are infrequently accessed will be stored in the selected access tier. 

Closes #4521 

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -568,6 +568,10 @@ The number of concurrent connections to the Azure Blob Storage service can be se
 ``-o azure.connections=10`` switch. By default, at most five parallel connections are
 established.
 
+The access tier of the blobs uploaded to the Azure Blob Storage service can be set with the
+``-o azure.access-tier=Cool`` switch. The allowed values are ``Hot``, ``Cool`` or ``Cold``. 
+If unspecified, the default is inferred from the default configured on the storage account.
+
 Google Cloud Storage
 ********************
 
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -132,6 +132,7 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 	for _, tier := range supportedAccessTiers() {
 		if strings.EqualFold(string(tier), cfg.AccessTier) {
 			accessTier = tier
+			debug.Log(" - using access tier %v", accessTier)
 			break
 		}
 	}
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -37,6 +37,8 @@ type Backend struct {
 	prefix       string
 	listMaxItems int
 	layout.Layout
+
+	accessTier blob.AccessTier
 }
 
 const saveLargeSize = 256 * 1024 * 1024
@@ -124,15 +126,30 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 		}
 	}
 
+	var accessTier blob.AccessTier
+	// if the access tier is not supported, then we will not set the access tier; during the upload process,
+	// the value will be inferred from the default configured on the storage account.
+	for _, tier := range supportedAccessTiers() {
+		if strings.EqualFold(string(tier), cfg.AccessTier) {
+			accessTier = tier
+			break
+		}
+	}
+
 	be := &Backend{
 		container:    client,
 		cfg:          cfg,
 		connections:  cfg.Connections,
 		Layout:       layout.NewDefaultLayout(cfg.Prefix, path.Join),
 		listMaxItems: defaultListMaxItems,
+		accessTier:   accessTier,
 	}
 
 	return be, nil
+}
+
+func supportedAccessTiers() []blob.AccessTier {
+	return []blob.AccessTier{blob.AccessTierHot, blob.AccessTierCool, blob.AccessTierCold, blob.AccessTierArchive}
 }
 
 // Open opens the Azure backend at specified container.
@@ -213,25 +230,39 @@ func (be *Backend) Path() string {
 	return be.prefix
 }
 
+// useAccessTier determines whether to apply the configured access tier to a given file.
+// For archive access tier, only data files are stored using that class; metadata
+// must remain instantly accessible.
+func (be *Backend) useAccessTier(h backend.Handle) bool {
+	notArchiveClass := !strings.EqualFold(be.cfg.AccessTier, "archive")
+	isDataFile := h.Type == backend.PackFile && !h.IsMetadata
+	return isDataFile || notArchiveClass
+}
+
 // Save stores data in the backend at the handle.
 func (be *Backend) Save(ctx context.Context, h backend.Handle, rd backend.RewindReader) error {
 	objName := be.Filename(h)
 
 	debug.Log("InsertObject(%v, %v)", be.cfg.AccountName, objName)
 
+	var accessTier blob.AccessTier
+	if be.useAccessTier(h) {
+		accessTier = be.accessTier
+	}
+
 	var err error
 	if rd.Length() < saveLargeSize {
 		// if it's smaller than 256miB, then just create the file directly from the reader
-		err = be.saveSmall(ctx, objName, rd)
+		err = be.saveSmall(ctx, objName, rd, accessTier)
 	} else {
 		// otherwise use the more complicated method
-		err = be.saveLarge(ctx, objName, rd)
+		err = be.saveLarge(ctx, objName, rd, accessTier)
 	}
 
 	return err
 }
 
-func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.RewindReader) error {
+func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.RewindReader, accessTier blob.AccessTier) error {
 	blockBlobClient := be.container.NewBlockBlobClient(objName)
 
 	// upload it as a new "block", use the base64 hash for the ID
@@ -252,11 +283,13 @@ func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.Rew
 	}
 
 	blocks := []string{id}
-	_, err = blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{})
+	_, err = blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{
+		Tier: &accessTier,
+	})
 	return errors.Wrap(err, "CommitBlockList")
 }
 
-func (be *Backend) saveLarge(ctx context.Context, objName string, rd backend.RewindReader) error {
+func (be *Backend) saveLarge(ctx context.Context, objName string, rd backend.RewindReader, accessTier blob.AccessTier) error {
 	blockBlobClient := be.container.NewBlockBlobClient(objName)
 
 	buf := make([]byte, 100*1024*1024)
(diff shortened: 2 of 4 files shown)


internal/backend/azure/azure.go
```diff
@@ -245,12 +261,17 @@ func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.Rew
 		return errors.Wrap(err, "StageBlock")
 	}
 
+	commitBlockListOptions := &blockblob.CommitBlockListOptions{}
+	if accessTier != nil {
```
MichaelEischer, line 265: the `if` is redundant. If accessTier is nil, then setting it is a noop.

internal/backend/azure/azure.go
```diff
@@ -207,25 +207,41 @@ func (be *Backend) Path() string {
...
 		// otherwise use the more complicated method
-		err = be.saveLarge(ctx, objName, rd)
+		err = be.saveLarge(ctx, objName, rd, &accessTier)
 	}
 
 	return err
 }
 
-func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.RewindReader) error {
+func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.RewindReader, accessTier *blob.AccessTier) error {
```
MichaelEischer, line 244: Currently that parameter is never nil. Why pass it as pointer then?

internal/backend/azure/config.go
```diff
@@ -22,7 +22,8 @@ type Config struct {
 	Container          string
 	Prefix             string
 
-	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+	Connections uint   `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+	AccessTier  string `option:"access-tier" help:"set the access tier for the blob storage (default: Hot)"`
```
MichaelEischer, line 26: Isn't the default whatever value is configured for the bucket?

internal/backend/azure/azure.go
```diff
@@ -207,25 +207,41 @@ func (be *Backend) Path() string {
...
 
+// useAccessTier determines whether to apply the configured access tier to a given file.
+// It returns true if the configured AccessTier is not "archive" and the file is a non-metadata pack file.
+// This ensures that:
+// 1. No files are ever stored in the Archive tier, regardless of the AccessTier setting.
+// 2. Only data files (non-metadata pack files) use the configured AccessTier when it's not "archive".
+func (be *Backend) useAccessTier(h backend.Handle) bool {
+	noArchiveClass := strings.ToLower(be.cfg.AccessTier) != "archive"
+	isDataFile := h.Type == backend.PackFile && !h.IsMetadata
+	return isDataFile && noArchiveClass
```
MichaelEischer, line 218: Why is that logic different from the one used for the S3 backend? Metadata files must use an online tier, but any such tier would work.

MichaelEischer (changes requested):

internal/backend/azure/config.go
MichaelEischer, line 26: I wonder whether it would make sense to limit this to a set of well-known access tiers. Probably just `Hot`,`Cool`,`Cold` and unset.

konidev20: There are `Premium` and other high-speed premium access tiers which are similarly configured. In the Azure SDK there is an option called `PossibleAccessTierValues` I will limit it to those for now. 

If there is no match, then should I unset or should I throw an error?

internal/backend/azure/azure.go
konidev20: Done. I have removed the redundant if.

internal/backend/azure/azure.go
konidev20: Done, fixed.

internal/backend/azure/config.go
konidev20: Yes, the value would be inferred from the Storage Account. I have updated the documentation.

internal/backend/azure/azure.go
konidev20: Fixe. I overlooked the `||` and used an `&&` by mistake.

internal/backend/azure/config.go
konidev20: Have a closer look the open method. I have added some logic there to handle valid access tiers.

internal/backend/azure/azure.go
```diff
@@ -124,12 +126,25 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
...
+	var accessTier blob.AccessTier
+
+	// if the access tier is not valid, then we will not set the access tier; during the upload process,
+	// the value will be inferred from the default configured on the storage account.
+	for _, tier := range validAccessTiers {
+		if strings.Compare(strings.ToLower(string(tier)), strings.ToLower(cfg.AccessTier)) == 0 {
+			accessTier = tier
+			break
+		}
+	}
```
konidev20, line 139: @MichaelEischer check here, about the handling of valid access tiers. I have done a case-insensitve check after validating the behaviour with Azure APIs and live debugging.

internal/backend/azure/config.go
konidev20: I have made some changes into the `open` function of the backend for the accessTier validation. Have a look there.

internal/backend/azure/azure.go
```diff
@@ -124,17 +126,32 @@ func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 		}
 	}
 
+	var accessTier blob.AccessTier
+	// if the access tier is not supported, then we will not set the access tier; during the upload process,
+	// the value will be inferred from the default configured on the storage account.
+	for _, tier := range supportedAccessTiers() {
+		if strings.Compare(strings.ToLower(string(tier)), strings.ToLower(cfg.AccessTier)) == 0 {
```
greatroar, line 133:
```suggestion
		if strings.ToLower(string(tier)) == strings.ToLower(cfg.AccessTier)) {
```

internal/backend/azure/azure.go
```diff
@@ -207,25 +224,39 @@ func (be *Backend) Path() string {
 	return be.prefix
 }
 
+// useAccessTier determines whether to apply the configured access tier to a given file.
+// For archive access tier, only data files are stored using that class; metadata
+// must remain instantly accessible.
+func (be *Backend) useAccessTier(h backend.Handle) bool {
+	notArchiveClass := strings.ToLower(be.cfg.AccessTier) != "archive"
+	isDataFile := h.Type == backend.PackFile && !h.IsMetadata
+	return isDataFile || notArchiveClass
```
greatroar, line 233:
```suggestion
	isDataFile := h.Type == backend.PackFile && !h.IsMetadata
	return isDataFile || !strings.EqualFold(be.cfg.AccessTier, "archive")
```

greatroar:
Are we sure that allowing the archive tier is a good idea? It's latency can be on the order of hours, say the [MS docs](https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview), so this seems to introduce a cold storage option without special support for getting the data back out. I mean, maybe you can set timeouts to multiple hours, but does that not cause other trouble?

internal/backend/azure/azure.go
```diff
@@ -245,12 +276,15 @@ func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.Rew
 		return errors.Wrap(err, "StageBlock")
 	}
 
+	commitBlockListOptions := &blockblob.CommitBlockListOptions{}
+	commitBlockListOptions.Tier = &accessTier
```
greatroar, line 280:
```suggestion
	commitBlockListOptions := &blockblob.CommitBlockListOptions{Tier: &accessTier}
```

internal/backend/azure/azure.go
```diff
@@ -297,7 +331,10 @@ func (be *Backend) saveLarge(ctx context.Context, objName string, rd backend.Rew
 		return errors.Errorf("wrote %d bytes instead of the expected %d bytes", uploadedBytes, rd.Length())
 	}
 
-	_, err := blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{})
+	commitBlockListOptions := &blockblob.CommitBlockListOptions{}
+	commitBlockListOptions.Tier = &accessTier
```
greatroar, line 335:
```suggestion
	commitBlockListOptions := &blockblob.CommitBlockListOptions{Tier: &accessTier}
```

internal/backend/azure/azure.go
```diff
@@ -297,7 +331,10 @@ func (be *Backend) saveLarge(ctx context.Context, objName string, rd backend.Rew
...
+
+	_, err := blockBlobClient.CommitBlockList(ctx, blocks, commitBlockListOptions)
```
greatroar, line 337:
```suggestion
	_, err := blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{
		Tier: &accessTier,
	})
```

internal/backend/azure/azure.go
```diff
@@ -245,12 +276,15 @@ func (be *Backend) saveSmall(ctx context.Context, objName string, rd backend.Rew
 		return errors.Wrap(err, "StageBlock")
 	}
 
+	commitBlockListOptions := &blockblob.CommitBlockListOptions{}
+	commitBlockListOptions.Tier = &accessTier
+
 	blocks := []string{id}
-	_, err = blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{})
+	_, err = blockBlobClient.CommitBlockList(ctx, blocks, commitBlockListOptions)
```
greatroar, line 283: I can't make this a GH suggestion because it spans a non-altered line, but this can be shortened to

```go
	blocks := []string{id}
	_, err = blockBlobClient.CommitBlockList(ctx, blocks, &blockblob.CommitBlockListOptions{
		Tier: &accessTier,
	})
```

internal/backend/azure/azure.go
MichaelEischer: Let's keep the `notArchiveClass` variable for similarity with `s3.useStorageClass`, but +1 on `strings.EqualFold`.

doc/030_preparing_a_new_repo.rst
```diff
@@ -568,6 +568,11 @@ The number of concurrent connections to the Azure Blob Storage service can be se
 ``-o azure.connections=10`` switch. By default, at most five parallel connections are
 established.
 
+The access tier of the blobs uploaded to the Azure Blob Storage service can be set with the
+``-o azure.access-tier=Cool`` switch. The allowed values are ``Hot``, ``Cool``, ``Cold`` or
+``Archive``. If unspecified, the default is inferred from the default configured on the
```
MichaelEischer, line 573: archive storage is not and cannot be officially supported at the moment. This must be removed from the docs.

MichaelEischer (commented):
> Are we sure that allowing the archive tier is a good idea? It's latency can be on the order of hours, say the [MS docs](https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview), so this seems to introduce a cold storage option without special support for getting the data back out.

The `archive` tier definitely shouldn't be officially documented as supported. Right now, the S3 backend also allows setting arbitrary storage classes (including glacier / deep archive) but those are never explicitly mentioned in the documentation. Instead the changelog entry from https://github.com/restic/restic/pull/4584/files makes is clear that cold storage is not officially supported and may require _manually_ warming up everything.

doc/030_preparing_a_new_repo.rst
konidev20: Resolved. I have removed the mention of the `Archive` tier from the docs. In the changelog, I have mentioned that `Archive` tier is not officially supported; referencing the changelog entry https://github.com/restic/restic/pull/4584/files

internal/backend/azure/azure.go
konidev20: I have accepted @MichaelEischer's recommendation

MichaelEischer (approved):
LGTM. Thanks!

PR #5047, opened 2024-09-10
cache: fix race condition in cache cleanup or similar.

damoclark:
Fix for multiple restic processes executing concurrently and racing to remove obsolete snapshots from the local backend cache.

This change tests the error returned from the closure/anonymous function, and if the error is that the given file does not exist, then skip the missing file in the results by returning nil.

The purpose is to silence the error generated from a race condition where multiple restic processes are executing concurrently and one unlinks a snapshot file, after another retrieves the file's path from a readdir, but before it attempts an lstat.  It is in response to the following error message:

```
error clearing snapshot files in cache: Walk: lstat /home/<username>/.cache/restic/23fa6bedac5c6d4...a6079eea2b3e3/snapshots/3f/3f4e3de8f8...38f8de1c995e3d: no such file or directory
```

I didn't raise an issue, as its quite trivial.  It is related to #4760, that was resolved in an identical way.  

I'm not a go developer, so please review this PR carefully.  I modelled the patch on #4760 but may have made errors.

Checklist
---------

### DIFF
--- a/internal/backend/cache/file.go
+++ b/internal/backend/cache/file.go
@@ -210,6 +210,10 @@ func (c *Cache) list(t restic.FileType) (restic.IDSet, error) {
 	dir := filepath.Join(c.path, cacheLayoutPaths[t])
 	err := filepath.Walk(dir, func(name string, fi os.FileInfo, err error) error {
 		if err != nil {
+			// ignore ErrNotExist to gracefully handle multiple processes clearing the cache
+			if errors.Is(err, os.ErrNotExist) {
+				return nil
+			}
 			return errors.Wrap(err, "Walk")
 		}
 
--- a/changelog/unreleased/pull-5047
+++ b/changelog/unreleased/pull-5047
@@ -1,0 +1,7 @@
+Bugfix: Fix possible error on concurrent cache cleanup
+
+Fix for multiple restic processes executing concurrently and racing to
+remove obsolete snapshots from the local backend cache.  Restic now suppresses the `no
+such file or directory` error.
+
+https://github.com/restic/restic/pull/5047


internal/backend/cache/file.go
```diff
@@ -210,6 +210,9 @@ func (c *Cache) list(t restic.FileType) (restic.IDSet, error) {
 	dir := filepath.Join(c.path, cacheLayoutPaths[t])
 	err := filepath.Walk(dir, func(name string, fi os.FileInfo, err error) error {
 		if err != nil {
+			// ignore ErrNotExist to gracefully handle multiple processes clearing the cache
+			if(errors.Is(err, os.ErrNotExist)
```
MichaelEischer, line 214: the brackets are somewhat messed up. It should be
```
if errors.Is(err, os.ErrNotExist) {
    return nil
}
```

internal/backend/cache/file.go
```diff
@@ -210,6 +210,9 @@ func (c *Cache) list(t restic.FileType) (restic.IDSet, error) {
 	dir := filepath.Join(c.path, cacheLayoutPaths[t])
 	err := filepath.Walk(dir, func(name string, fi os.FileInfo, err error) error {
 		if err != nil {
+			// ignore ErrNotExist to gracefully handle multiple processes clearing the cache
```
MichaelEischer, line 213: Please change the commit description to

`cache: fix race condition in cache cleanup` or similar.

MichaelEischer:
Thanks for the bugfix. Can you add a minimal changelog entry? (or should I take that over?)

internal/backend/cache/file.go
damoclark: Oops, missed the closing brace.  Sorry about that.  Made the change directly on github as I don't have a dev environment for Go.  Would have benefited from some IDE syntax highlighting there. :)

damoclark:
> Thanks for the bugfix. Can you add a minimal changelog entry? (or should I take that over?)

No problem.  I had a go with the changelog entry.  I think it is correct.

D.

internal/backend/cache/file.go
damoclark: I've fixed the syntax error.  

Let me know if anything else is needed.

MichaelEischer (approved):
LGTM. I've taken the liberty to fix the remaining syntax error and squashed all commits.

PR #5053, opened 2024-09-13
generate: allow passing `-` for stdout output

### LINKED ISSUE #2511: Generate autocomplete script to /dev/stdout
I'm using restic 0.9.6 compiled with go1.13.4 on linux/amd64

I'm trying to add autocomplete to my `.zshrc` by adding this line:
```zsh
source <(restic generate --zsh-completion /dev/stdout)
```

Having generated completion in zshrc is very common in the k8s ecosystem;
```zsh
source <(kubectl completion zsh)
source <(minikube completion zsh)
source <(helm completion zsh)
```

This is quite cool because the user doesn't have to think about overwriting a previously generated completion script when they upgrade the tool.

However, it doens't work with restic, because restic writes a message to /dev/stdout when running the generate command.
> writing zsh completion file to /dev/stdout

This message breaks interpreting the outputted script. When I start a new terminal with zsh in it:
> /proc/self/fd/13:1: command not found: writing
> _arguments:comparguments:325: can only be called from completion function
> _arguments:comparguments:325: can only be called from completion function

The first line, "command not found: writing" seems to relate to the "writing zsh completion file to /dev/stdout" message.

I would like to have a way to achieve the same as kubectl and other tools, generating the latest autocomplete code when starting a new terminal with zsh so that the autocomplete is always up-to-date.

Maybe the message can be written to /dev/stderr so that it can be filtered in `.zshrc`, or the message is silenced when using `/dev/stdout` as output file?

I didn't report this as a bug because generating the autocomplete code to `/dev/stdout` probably isn't what the author of the autocomplete functionality had in mind. 

If the maintainer of this software agrees with a proposed solution, I'd be very happy to work on a PR.

It's huge. Having an easy tool to maintain secure offsite backups is great. Since restic, I backup more often. Thanks!

rominf:

Allow generating shell completions to stdout by supplying `-` as file name. Since generating completions to stdout for multiple shells does not make sense, enforce `-` is supplied only once.

Closes #2511.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2511
+++ b/changelog/unreleased/issue-2511
@@ -1,0 +1,6 @@
+Enhancement: Allow generating shell completions to stdout
+
+Restic `generate` now supports passing `-` passed as file name to `--[shell]-completion` option.
+
+https://github.com/restic/restic/issues/2511
+https://github.com/restic/restic/pull/5053
--- a/cmd/restic/cmd_generate.go
+++ b/cmd/restic/cmd_generate.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"io"
+	"os"
 	"time"
 
 	"github.com/restic/restic/internal/errors"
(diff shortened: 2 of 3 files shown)


cmd/restic/cmd_generate.go
```diff
@@ -105,28 +126,45 @@ func runGenerate(opts generateOptions, args []string) error {
 		}
 	}
 
+	err := checkStdoutForSingleShell(opts)
+	if err != nil {
+		return err
+	}
+
 	if opts.BashCompletionFile != "" {
+		if opts.BashCompletionFile == "-" {
```
MichaelEischer, line 135: IMO the handling for `-` belongs into `writeBashCompletion` etc. The current changes now introduce two places in which the function to generate the changelog could be called. The current `write*Completion` functions are also lots of copy&paste code.

We could get rid of them using the following function that receives that changelog generator function as parameter:

```go
func writeCompletion(filename string, shell string, generate func(w io.Writer) error) error {
	if stdoutIsTerminal() {
		Verbosef("writing %s completion file to %v\n", shell, filename)
	}
	var outFile io.Writer
	if filename == "-" {
		var err error
		outFile, err = os.Create(filename)
		if err != nil {
			return err
		}
		defer outFile.Close()
	} else {
		outFile = globalOptions.stdout
	}

	return generate(outFile)
}
```

This would then be called using 
```
writeCompletion(opts.BashCompletionFile, "bash", cmdRoot.GenBashCompletion)
```

That would replace all `write*Completion` functions.

cmd/restic/cmd_generate.go
```diff
@@ -93,7 +95,26 @@ func writePowerShellCompletion(file string) error {
 	return cmdRoot.GenPowerShellCompletionFile(file)
 }
 
-func runGenerate(opts generateOptions, args []string) error {
+func checkStdoutForSingleShell(opts generateOptions) error {
+	isStdout := []bool{
+		opts.BashCompletionFile == "-",
```
MichaelEischer, line 100: Please move the `== "-"` comparison into the loop.

MichaelEischer (changes requested):
The check for `-` should move into the `write*Completion` functions. That will require a few more changes to avoid additional code duplication, see the comments below.

changelog/unreleased/issue-2511
```diff
@@ -0,0 +1,5 @@
+Enhancement: Allow generating shell completions to stdout by supplying `-` as file name
```
MichaelEischer, line 1: The summary length is limited to 80 characters.

rominf:
Thank you for your review! I applied your suggestions with some minor modifications to make the code compile.

MichaelEischer (approved):
LGTM. Sorry for the delayed review and thanks for improving the generate command!

PR #5054, opened 2024-09-13
dump: add --compress flag to compress archives

phillipp:

~When using dump and exporting an archive, the archive was not compressed. This PR adds a flag to do so.~

When using `dump` to export a `zip` archive, the archive was not compressed. This PR enables compression for these archives.

No, I don't think so.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5054
+++ b/changelog/unreleased/pull-5054
@@ -1,10 +1,7 @@
-Enhancement: Add `--compress` flag to `dump` command to compress archive
+Enhancement: Compress ZIP archives created by `dump` command
 
-Restic did not compress the archives that was created by using the
-`dump` command. It now allows to save some disk space when exporting
-archives by adding a `--compress` flag. The DEFLATE algorithm is used
-for "zip" archives, and the gzip algorithm for "tar" archives,
-resulting in a .tar.gz or .tgz file. Not compressing the archive
-is still the default.
+Restic did not compress the archives that were created by using
+the `dump` command. It now saves some disk space when exporting
+archives using the DEFLATE algorithm for "zip" archives.
 
 https://github.com/restic/restic/pull/5054
--- a/changelog/unreleased/pull-5054
+++ b/changelog/unreleased/pull-5054
@@ -1,0 +1,10 @@
+Enhancement: Add `--compress` flag to `dump` command to compress archive
+
+Restic did not compress the archives that was created by using the
+`dump` command. It now allows to save some disk space when exporting
+archives by adding a `--compress` flag. The DEFLATE algorithm is used
+for "zip" archives, and the gzip algorithm for "tar" archives,
+resulting in a .tar.gz or .tgz file. Not compressing the archive
+is still the default.
+
+https://github.com/restic/restic/pull/5054
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -50,9 +50,8 @@ Exit status is 12 if the password is incorrect.
 // DumpOptions collects all options for the dump command.
 type DumpOptions struct {
 	restic.SnapshotFilter
-	Archive  string
-	Target   string
-	Compress bool
+	Archive string
+	Target  string
 }
 
 var dumpOptions DumpOptions
@@ -64,7 +63,6 @@ func init() {
 	initSingleSnapshotFilter(flags, &dumpOptions.SnapshotFilter)
 	flags.StringVarP(&dumpOptions.Archive, "archive", "a", "tar", "set archive `format` as \"tar\" or \"zip\"")
 	flags.StringVarP(&dumpOptions.Target, "target", "t", "", "write the output to target `path`")
-	flags.BoolVarP(&dumpOptions.Compress, "compress", "c", false, "compress archive contents. When enabled, the DEFLATE algorithm is applied for \"zip\" archives, and the gzip algorithm for \"tar\" archives, resulting in a .tar.gz or .tgz file. (default: false)")
 }
 
 func splitPath(p string) []string {
@@ -189,7 +187,7 @@ func runDump(ctx context.Context, opts DumpOptions, gopts GlobalOptions, args []
 		canWriteArchiveFunc = func() error { return nil }
 	}
 
-	d := dump.New(opts.Archive, opts.Compress, repo, outputFileWriter)
+	d := dump.New(opts.Archive, repo, outputFileWriter)
 	err = printFromTree(ctx, tree, repo, "/", splittedPath, d, canWriteArchiveFunc)
 	if err != nil {
 		return errors.Fatalf("cannot dump file: %v", err)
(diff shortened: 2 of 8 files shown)


cmd/restic/cmd_dump.go
```diff
@@ -50,8 +50,9 @@ Exit status is 12 if the password is incorrect.
 // DumpOptions collects all options for the dump command.
 type DumpOptions struct {
 	restic.SnapshotFilter
-	Archive string
-	Target  string
+	Archive            string
+	Target             string
+	CompressZipArchive bool
```
konidev20, line 55: Just `Compress` should be fine. Considering, we can compress a `.tar` after archiving.

cmd/restic/cmd_dump.go
```diff
@@ -63,6 +64,7 @@ func init() {
 	initSingleSnapshotFilter(flags, &dumpOptions.SnapshotFilter)
 	flags.StringVarP(&dumpOptions.Archive, "archive", "a", "tar", "set archive `format` as \"tar\" or \"zip\"")
 	flags.StringVarP(&dumpOptions.Target, "target", "t", "", "write the output to target `path`")
+	flags.BoolVar(&dumpOptions.CompressZipArchive, "compress-zip", false, "compress zip archive contents")
```
konidev20, line 67: Can we do a shorthand `c` as well? You can check if there any conflicts with the global options.

changelog/unreleased/pull-5054
```diff
@@ -0,0 +1,18 @@
+Enhancement: Add --compress-zip flag to dump command to compress archive
+
+# Describe the problem in the past tense, the new behavior in the present
+# tense. Mention the affected commands, backends, operating systems, etc.
+# Focus on user-facing behavior, not the implementation.
+# Use "Restic now ..." instead of "We have changed ...".
```
konidev20, line 6: These comments can be excluded from the changelog.

phillipp: Oh, must have slipped by...

phillipp:
@konidev20 It would only be a little more effort to add compression to the tar stream as well, I guess that would be more consistent when renaming the flag to --compress (-c is unused, so we can use that too). Do you agree? because then I would add that?

phillipp:
I've added compression for .tar as well and updated the docs and flag name.

I can squash the commits later for better readability

konidev20 (commented):
LGTM. I will let Michael do the final review.

changelog/unreleased/pull-5054
```diff
@@ -0,0 +1,10 @@
+Enhancement: Add --compress flag to dump command to compress archive
```
MichaelEischer, line 1: Please surround flags and commands with backticks.
```
Enhancement: Add `--compress` flag to `dump` command to compress archive
```

doc/man/restic-dump.1
```diff
@@ -42,6 +42,10 @@ Exit status is 12 if the password is incorrect.
 \fB-a\fP, \fB--archive\fP="tar"
 	set archive \fBformat\fR as "tar" or "zip"
 
+.PP
+\fB--compress\fP[=false]
```
MichaelEischer, line 46: The man pages are only regenerated directly before a release. Please revert that change.

internal/dump/tar.go
```diff
@@ -13,12 +14,25 @@ import (
 )
 
 func (d *Dumper) dumpTar(ctx context.Context, ch <-chan *restic.Node) (err error) {
-	w := tar.NewWriter(d.w)
+	var w *tar.Writer
```
MichaelEischer, line 17: Nit: can be simplified to

```
outer := d.w
if d.compress {
outer = gzip.NewWriter(d.w)
}
w := tar.NewWriter(outer)
```

MichaelEischer (changes requested):
The changes look good so far, I just have a few small remarks. Also please squash the commits.

internal/dump/tar.go
phillipp: But that would lead to `outer` being an io.Writer which does not have a Close func, so the gzip writer cannot be closed in the deferred func, or am I missing something? I'm new to golang and don't know all the right ways.

phillipp: Ok, found out that a type assertion can be used. Will change.

cmd/restic/cmd_dump.go
```diff
@@ -63,6 +64,7 @@ func init() {
 	initSingleSnapshotFilter(flags, &dumpOptions.SnapshotFilter)
 	flags.StringVarP(&dumpOptions.Archive, "archive", "a", "tar", "set archive `format` as \"tar\" or \"zip\"")
 	flags.StringVarP(&dumpOptions.Target, "target", "t", "", "write the output to target `path`")
+	flags.BoolVarP(&dumpOptions.Compress, "compress", "c", false, "compress archive contents. When enabled, the DEFLATE algorithm is applied for \"zip\" archives, and the gzip algorithm for \"tar\" archives, resulting in a .tar.gz or .tgz file. (default: false)")
```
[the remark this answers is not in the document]
phillipp: Yeah but where can I add something to the docs but not to the CLI, i thought the docs where generated from the cli source? Am I missing something?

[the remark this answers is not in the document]
rawtaz: The sources for the docs at [restic.readthedocs.io](https://restic.readthedocs.io) are in the docs/ folder in this repo, see https://github.com/restic/restic/tree/master/doc .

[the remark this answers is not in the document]
phillipp: Yeah, and they are auto-generated, see https://github.com/restic/restic/pull/5054#discussion_r1759797893

No changes to the docs itself allowed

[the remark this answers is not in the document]
rawtaz: Yes and no.
- The docs at [restic.readthedocs.io](https://restic.readthedocs.io/) are indeed auto-generated by RTD, but the sources for them are at https://github.com/restic/restic/tree/master/doc and you are totally fine to edit them to add contents to the official documentation for restic, when it makes sense to do so.
- The documentation for/in the CLI of restic is auto-generated too, as mentioned in the comment you linked to, and those files should not be edited, that is correct. But they are a different thing than the official docs on RTD.

rawtaz:
@MichaelEischer and I were discussing this PR earlier today and would like to make some changes. Namely:
- Make DEFLATE compression of `zip` archives the default, i.e. without needing to be enabled with an option or such.
- Leave out compression of `tar` archives for now, the reason being that it is so simple to compress the tar archive afterwards, e.g. by `restic dump ... | gzip`.
- Leave out the `--compress` option as it is no longer needed.

phillipp:
Updated the PR, much simpler now.

internal/dump/zip.go
```diff
@@ -39,6 +39,7 @@ func (d *Dumper) dumpNodeZip(ctx context.Context, node *restic.Node, zw *zip.Wri
 		Modified:           node.ModTime,
 	}
 	header.SetMode(node.Mode)
+	header.Method = zip.Deflate
```
greatroar, line 42: I would set this only for regular files, or at least not for symlinks.

MichaelEischer: sounds reasonable, symlinks are probably usually too short to benefit from compression.

MichaelEischer (approved):
LGTM. I've applied the suggestion from @greatroar and minimally extended the test case.

PR #5058, opened 2024-09-14
Changelogs should omit problem if its description duplicates the new behavior

MichaelEischer:
When adding a new feature, the problem description often just says that feature Y was missing, followed by saying that feature Y is now supported.

This duplication just makes the changelog entries unnecessarily verbose.

Just noticed it again in the changelog of https://github.com/restic/restic/pull/5053. The changelogs for restic 0.17.1 (and earlier versions) already followed that rule.

Checklist
---------


greatroar (approved):

MichaelEischer (commented):
LGTM.

PR #5061, opened 2024-09-15
fs: fix restoring timestamps on older Windows versions for long paths

### LINKED ISSUE #1843: Timestamp is not restored correctly on file paths longer than 256 chars in Windows
## Output of `restic version`

restic 0.9.1 compiled with go1.10.3 on windows/amd64

```
set RESTIC_PASSWORD=<..>
set RESTIC_REPOSITORY=H:\restic_test
restic restore latest --target c:\Tools\restic\restore --include restic_backup.ps1
```

Mapped Windows file share.
The file restic_backup.ps1 was backed up from a path longer than 256 chars.

## Expected behavior

File will be restored with correct timestamp

## Actual behavior

File was restored, but the timestamp of file is set to the restore time/date. Restic throws the following error:
`ignoring error for \c\Entwicklung\12345678901234567890123456789012345678901234567890\12345678901234567890123456789012345678901234567890\12345678901234567890123456789012345678901234567890\12345678901234567890123456789012345678901234567890\12345678901234567890123456789012345678901234567890\restic_backup.ps1: RestoreTimestamps: UtimesNano: The system cannot find the path specified.`

After adding the prefix "\\\\?\\" to the path in the --target option, the restore works like expected. The timestamp of the file will be restored correctly.

Backup a file with a pathname longer than 256 chars and try to restore it using a target directory without using the "\\\\?\\" prefix.

This problem may be a similar problem like the one described in issue #576.

If available, use a function, that supports pathnames longer than 256 chars to change the timestamp of 
restored file.

Yes, indeed. Restic is an easy to use, powerful backup tool, and even if there are some small issues, it already works very good.

MichaelEischer:
Go automagically enables long-path support on Windows 10 >=1607 (see https://github.com/golang/go/issues/41734). The standard library calls are also wrapped accordingly to support older Windows versions. However, for some filesystem operations we have to use the syscalls package directly which does not wrap paths. 

Most places should already contain a corresponding `fixpath(...)` call, this PR fixes restoring the timestamp and file attributes.

Fixes https://github.com/restic/restic/issues/1843

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-1843
+++ b/changelog/unreleased/issue-1843
@@ -1,0 +1,9 @@
+Bugfix: Correctly restore timestamp on long filepaths on old Windows versions
+
+The `restore` command did not restore timestamps on file paths longer than 256
+characters on Windows versions before Windows 10 1607.
+
+This issue is now resolved.
+
+https://github.com/restic/restic/issues/1843
+https://github.com/restic/restic/pull/5061
--- a/internal/fs/node.go
+++ b/internal/fs/node.go
@@ -315,7 +315,7 @@ func nodeRestoreTimestamps(node *restic.Node, path string) error {
 		return nodeRestoreSymlinkTimestamps(path, utimes)
 	}
 
-	if err := syscall.UtimesNano(path, utimes[:]); err != nil {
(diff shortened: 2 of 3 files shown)


MichaelEischer (commented):
LGTM.

Issue #4173, opened 2023-01-24
Add include/exclude to Dump command / let Dump accept multiple paths


akrabu:
Output of `restic version`
--------------------------

restic 0.15.0 compiled with go1.19.5 on darwin/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

It would be useful if the dump command accepted --include, --iincude, --exclude, --iexclude switches.  Sometimes users ask me for "all the files with "SOME_PATTERN" in the name.  I generally have to do a restore, then compress, then upload to OneDrive for them (my users don't all use Restic unfortunately).

It would be great if I could just do `restic dump --include "SOME_PATTERN" | xz > ~/OneDrive/folder_shared_with_user/SOME_PATTERN.tar.xz`

If this isn't feasible, then accepting multiple paths might be enough too.  I could do `restic ls <snapshot-ID> | grep SOME_PATTERN` then feed the paths I need into dump.

Of course both would always be cool too :) 

What are you trying to do? What problem would this solve?
---------------------------------------------------------

More easily dump patterns / multiple paths into compressed packages to share with end users not using Restic.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Restic has saved my bacon numerous times and makes everyone at work call me a wizard haha

RayZ0rr:
Correct me if I wrong but isn't it possible to use `restore` for this because it already has those flags?

> Use --exclude and --include to restrict the restore to a subset of files in the snapshot.

akrabu:
> Correct me if I wrong but isn't it possible to use `restore` for this because it already has those flags?

It is, and that’s what I do now - but it means I have to have a staging area to restore to, then compress it, then upload it to the user.  If I could dump, I could dump directly to the cloud syncing folder and skip the staging area.

phillipp:
@MichaelEischer I actually have a working version of this. It is a bit inefficient (might traverse the subtrees multiple times) and hackish right now, but is working.

I might spend some time on polishing this, if there is a chance of merging a pull request.

Is this out of scope or could this be merged?

MichaelEischer:
The extensions of the `dump` command suggested here, sound quite reasonable. Although, we probably have to agree on a few more details like path structure within the output files.

Without knowing more about the implementation, it's hard to tell whether I'd be satisfied with it or not. Did you implement the `--include`/`--exclude` suggestion or supporting multiple paths?

phillipp:
[...]

PR #5062, opened 2024-09-15
Dump multiple paths

phillipp:

Before this change, `dump` accepted one path to dump, now multiple paths are accepted.

#4173

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_dump.go
+++ b/cmd/restic/cmd_dump.go
@@ -74,56 +74,110 @@ func splitPath(p string) []string {
 	return append(s, f)
 }
 
-func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.BlobLoader, prefix string, pathComponents []string, d *dump.Dumper, canWriteArchiveFunc func() error) error {
+func cleanupPathList(pathComponentsList [][]string) [][]string {
+	// Build a set of paths for quick lookup
+	pathSet := make(map[string]struct{})
+
+	for _, splittedPath := range pathComponentsList {
+		pathStr := path.Join(splittedPath...)
+		pathSet[pathStr] = struct{}{}
+	}
+
+	// Filter out subpaths that are covered by parent directories
+	filteredList := [][]string{}
+
+	for _, splittedPath := range pathComponentsList {
+		isCovered := false
+		// Check if any prefix of the path exists in the pathSet
+		for i := len(splittedPath) - 1; i >= 1; i-- {
+			prefixPath := path.Join(splittedPath[0:i]...)
+			if _, exists := pathSet[prefixPath]; exists {
+				isCovered = true
+				break
+			}
+		}
+		if !isCovered {
+			filteredList = append(filteredList, splittedPath)
+		}
+	}
+
+	return filteredList
+}
+
+func filterPathComponents(pathComponentsList [][]string, prefix string) [][]string {
+	var filteredList [][]string
+	for _, pathComponents := range pathComponentsList {
+		if len(pathComponents) > 1 && pathComponents[0] == prefix {
+			filteredList = append(filteredList, pathComponents[1:])
+		}
+	}
+	return filteredList
+}
+
+func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.BlobLoader, prefix string, pathComponentsList [][]string, d *dump.Dumper, canWriteArchiveFunc func() error) error {
 	// If we print / we need to assume that there are multiple nodes at that
 	// level in the tree.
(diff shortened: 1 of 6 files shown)


phillipp:
The tests are not passing right now because the dumper does not need to allow passing an `io.Writer` anymore as this is not required, we could keep this around but idk, seems odd to me to pass a field as an argument.

cmd/restic/cmd_dump.go
```diff
@@ -74,56 +74,138 @@ func splitPath(p string) []string {
 	return append(s, f)
 }
 
-func printFromTree(ctx context.Context, tree *restic.Tree, repo restic.BlobLoader, prefix string, pathComponents []string, d *dump.Dumper, canWriteArchiveFunc func() error) error {
+func preparePathList(pathComponentsList []string) [][]string {
```
MichaelEischer, line 77: Please add a short description how this function works. It took me a while to reverse engineer the underlying idea: first all paths are stored in a pathSet in sanitized form. The filtered set then only contains paths, for which no prefix already exists in the pathSet. This ensures that the order in which paths are processed does not matter.

cmd/restic/cmd_dump.go
```diff
@@ -74,56 +74,138 @@ func splitPath(p string) []string {
...
+			l := len(pathComponents)
+
+			// If dumping something in the highest level it will just take the
+			// first item it finds and dump that according to the switch case below.
+			if node.Name == pathComponents[0] {
+				switch {
+				case l == 1 && node.Type == restic.NodeTypeFile:
+					d.WriteNode(ctx, node)
+					dumpedNodes[node.Name] = struct{}{}
+					break out
```
MichaelEischer, line 162: a simple `break` after the switch statement has the same effect. Then `dumpedNodes` can also move out of the switch statement.

cmd/restic/cmd_dump.go
```diff
@@ -74,56 +74,138 @@ func splitPath(p string) []string {
...
-				}
-				return printFromTree(ctx, subtree, repo, item, pathComponents[1:], d, canWriteArchiveFunc)
-			case node.Type == restic.NodeTypeDir:
-				if err := canWriteArchiveFunc(); err != nil {
-					return err
+	out:
+		// iterate over all pathComponents
+		for _, pathComponents := range pathComponentsList {
+			item := filepath.Join(prefix, pathComponents[0])
+			l := len(pathComponents)
```
MichaelEischer, line 153: `item` and `l` should only be calculated inside the `if` block

cmd/restic/cmd_dump.go
```diff
@@ -74,56 +74,138 @@ func splitPath(p string) []string {
...
+
+			// If dumping something in the highest level it will just take the
+			// first item it finds and dump that according to the switch case below.
+			if node.Name == pathComponents[0] {
+				switch {
+				case l == 1 && node.Type == restic.NodeTypeFile:
+					d.WriteNode(ctx, node)
```
MichaelEischer, line 160: This behaves incorrectly when a user specifies multiple file paths. `WriteNode` only dumps the file content without any metadata. Also a combination of a directory + file path results in a corrupted output.

cmd/restic/cmd_dump.go
```diff
@@ -74,56 +74,138 @@ func splitPath(p string) []string {
...
-				return fmt.Errorf("%q should be a file, but is a %q", item, node.Type)
 			}
 		}
 	}
-	return fmt.Errorf("path %q not found in snapshot", item)
+
+	// Check if all paths were found in the snapshot
+	for _, item := range pathComponentsList {
+		if _, ok := dumpedNodes[item[0]]; !ok {
+			return fmt.Errorf("path %q not found in snapshot", filepath.Join(prefix, item[0]))
```
MichaelEischer, line 200: Should use `path.Join`

MichaelEischer (changes requested):
Sorry for the very late review.

I found a few conceptual issues which we have to sort out first:
- what should happen if a user passes `/file1` and `/file2` as paths? My expectation would be an archive containing both paths.
- an archive with a `/file` and `/dir` path should similarly store the file in the archive instead of writing it verbatim to the output.
- What should happen when dumping `/dir` and `/other/dir`? Mixing both directories as would happen now feels wrong. But also including the prefix in the path is also somewhat weird as we'd either have to break backwards compatibility or only add the prefix when there are multiple paths.

MichaelEischer:
I'll close the PR for now as it needs larger changes before we could consider merging it. Feel free to open a new PR if you want to continue with the PR.

PR #5074, opened 2024-09-30
dump: Simplify writeNode and use fewer goroutines

greatroar:

This changes Dumper.writeNode to spawn loader goroutines as needed instead of as a pool. The code is shorter, fewer goroutines are spawned for small files, and crash dumps (also for unrelated errors) should be smaller.

No.

Checklist
---------

### DIFF
--- a/internal/dump/common.go
+++ b/internal/dump/common.go
@@ -6,7 +6,6 @@ import (
 	"path"
 
 	"github.com/restic/restic/internal/bloblru"
-	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/walker"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


internal/dump/common.go
```diff
@@ -104,75 +103,50 @@ func (d *Dumper) WriteNode(ctx context.Context, node *restic.Node) error {
...
+					return err
+				}
 			}
 		}
 		return nil
 	})
 
-	for i := uint(0); i < d.repo.Connections(); i++ {
+	// Start short-lived goroutines to load blobs.
+	// There will be at most 1+cap(blobs) calling LoadBlob at any moment.
```
greatroar, line 125: Note that the number of live goroutines can be higher, if the scheduler determines to leave them dangling just after ch <- blob, below. We can easily limit the number of goroutines to exactly cap(blobs) by sending ch before spawning. Would that be better?

internal/dump/common.go
```diff
@@ -104,75 +103,50 @@ func (d *Dumper) WriteNode(ctx context.Context, node *restic.Node) error {
...
-		for result := range writerCh {
-			select {
-			case data := <-result.data:
-				if _, err := w.Write(data); err != nil {
-					return errors.Wrap(err, "Write")
-				}
-			case <-ctx.Done():
-				return ctx.Err()
-			}
+		select {
```
greatroar, line 143: This select isn't really necessary as LoadBlob and the writer goroutine respond to ctx being cancelled, but then we might do some unnecessary work.

MichaelEischer (approved):
LGTM. Thanks for the simplification.

internal/dump/common.go
MichaelEischer: I don't think it matters, in the worst case the additional goroutine ends up blocked in the backend waiting for its turn.

PR #5093, opened 2024-10-17
Fix Azure Container Token Auth

### LINKED ISSUE #4004: azure: Unable to init repo when using container-level SAS (SAT)
Output of `restic version`
--------------------------

`restic 0.14.0 (v0.14.0-234-g9354262b1-dirty) compiled with go1.19 on darwin/amd64`

~~~
export AZURE_ACCOUNT_NAME=storageaccountname
export AZURE_ACCOUNT_SAS='si=containername&spr=https&sv=2021-06-08&sr=c&sig=deadbeef'
export RESTIC_PASSWORD=password
restic -r azure:containername:/backup init
~~~

Azure Blob Storage

Expected behavior
-----------------

~~~
created restic repository f2b2e727bb at azure:containername:/backup

irrecoverably lost.
~~~

Actual behavior
---------------

~~~
Fatal: create repository at azure:containername:/backup failed: container.CreateIfNotExists: storage: service returned error: StatusCode=403, ErrorCode=AuthorizationFailure, ErrorMessage=This request is not authorized to perform this operation.
RequestId:e0f0387b-701e-0037-26e0-efa955000000
Time:2022-11-03T23:57:40.3114536Z, RequestInitiated=Thu, 03 Nov 2022 23:57:39 GMT, RequestId=e0f0387b-701e-0037-26e0-efa955000000, API Version=2021-06-08, QueryParameterName=, QueryParameterValue=
~~~

In Azure Blob Storage, create a Storage Account, then in that SA create a Container, and in that *Container* create a Shared Access Token (which is basically the same thing as a Shared Access Signature that you can create on the Storage Account level), then run the commands above.

This has nothing to do with the `/backup` path in the example - the symptoms are the same even with just `azure:containername:/` as the repository.

The problem seems to be that when using a SAS/SAT on the *container* level, there is no permission to perform actions that have to do with creating the container - the SAS/SAT is only allowed to poke around inside the existing container, which makes sense.

The Azure library restic uses calls `container.CreateIfNotExists` which is where things fail, because it tries to create a container, which is denied.

In order to support SAS/SAT on a container level we must make the code not attempt to create a container (or similar), and instead just presume that it is there. The entire SAS/SAT can not have been created without the container already existing, in this case.

I guess we have two options to handle this:
- If creation of the *backend* fails (this is what fails here, `create(ctx, repo, gopts.extended)`), ignore this and somehow instantiate a backend anyway, then carry on and try to initialize a repository at that backend (`s.Init(ctx, version, gopts.password, chunkerPolynomial)`). Not sure how feasible and clean this is.
- Inspect the `AZURE_ACCOUNT_SAS` value and determine if it's an account- or container-level SAS/SAT, and depending on that do or skip the backend creation before initializing a repository on the backend. We'd still need to instantiate a backend in the code, so I'm not sure how to do that cleanly either. Perhaps it's simple.

Always 👍

---

EDIT: A workaround is to temporarily use an account-level SAS (or the regular account key, if one fancies that instead) for the `restic init` and after that use the container-level SAS/SAT as usual for the other restic commands (works fine since they won't try to create the container).

Seefin:
This change ignores the Authorization Failed error returned from the Azure Storage Account backend when attempting to use a Shared Access Signature token that was **not** created with the Storage Account key to initalise a new repository.

The issue is caused by the Azure library that we use in the backend. When enumerating the container properties, the library internally calls `container.CreateIfNotExists` which is where things fail, because it tries to create a container, which is denied, as the token doesn'thave permission to create new containers, only poke around in the container it was made for.

Closes #4004

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4004
+++ b/changelog/unreleased/issue-4004
@@ -1,0 +1,12 @@
+Bugfix: Allow use of container level SAS/SAT tokens with Azure backend
+
+When using a SAS/SAT token for authentication with Azure, restic was expecting
+the provided token to be generated at the account level, granting permissions
+to the storage account and all its containers. This caused an error that did
+not allow tokens that were generated at the container level to be used to 
+initalize a repository.
+Restic now allows SAS/SAT tokens that were generated at the account or 
+container level to be used to initalize a repository.
+
+https://github.com/restic/restic/issues/4004
+https://github.com/restic/restic/pull/5093
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -157,6 +157,12 @@ func Create(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, er
 		if err != nil {
 			return nil, errors.Wrap(err, "container.Create")
 		}
+	} else if err != nil && bloberror.HasCode(err, bloberror.AuthorizationFailure) {
+		// We ignore this Auth. Failure, as the failure is related to the type
+		// of SAS/SAT, not an actual real failure. If the token is invalid, we
+		// fail later on anyway.
+		// For details see Issue #4004.
+		debug.Log("Ignoring AuthorizationFailure when calling GetProperties")
 	} else if err != nil {
 		return be, errors.Wrap(err, "container.GetProperties")
 	}
--- a/internal/backend/azure/azure_test.go
+++ b/internal/backend/azure/azure_test.go
@@ -80,6 +80,91 @@ func BenchmarkBackendAzure(t *testing.B) {
 	newAzureTestSuite().RunBenchmarks(t)
 }
 
+// TestBackendAzureAccountToken tests that a Storage Account SAS/SAT token can authorize.
+// This test ensures that restic can use a token that was generated using the storage
+// account keys can be used to authorize the azure connection.
+// Requires the RESTIC_TEST_AZURE_ACCOUNT_NAME, RESTIC_TEST_AZURE_REPOSITORY, and the
+// RESTIC_TEST_AZURE_ACCOUNT_SAS environment variables to be set, otherwise this test
+// will be skipped.
+func TestBackendAzureAccountToken(t *testing.T) {
+	vars := []string{
+		"RESTIC_TEST_AZURE_ACCOUNT_NAME",
+		"RESTIC_TEST_AZURE_REPOSITORY",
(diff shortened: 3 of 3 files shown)


internal/backend/azure/azure.go
```diff
@@ -157,6 +157,13 @@ func Create(ctx context.Context, cfg Config, rt http.RoundTripper) (*Backend, er
 		if err != nil {
 			return nil, errors.Wrap(err, "container.Create")
 		}
+	} else if err != nil && bloberror.HasCode(err, bloberror.AuthorizationFailure) {
+		// We ignore this Auth. Failure, as the failure is related to the type
+		// of SAS/SAT, not an actual real failure. If the token is invalid, we
+		// fail later on anyway.
+		// For details see Issue #4004.
+		debug.Log("Ignoring AuthorizationFailure when calling GetProperties")
+		err = nil
```
MichaelEischer, line 166: Assigning `nil` here is actually redundant, so let's remove it to make the linter happy.

changelog/unreleased/issue-4004
```diff
@@ -0,0 +1,12 @@
+Bugfix: Allow use of container level SAS/SAT tokens with Azure backend.
```
MichaelEischer, line 1: The summary must not end with a punctuation character.

MichaelEischer (commented):
I have two small nits, otherwise LGTM.

internal/backend/azure/azure.go
Seefin: Easy, I'll remove

changelog/unreleased/issue-4004
Seefin: Easy, I'll remove

MichaelEischer (approved):
LGTM. Thanks for fixing this issue!

MichaelEischer:
I've slightly squashed the commits, without changing anything: https://github.com/restic/restic/compare/16d0ab6f4e1ceaf35c2bcbf108b3e8f6e7a9ae34..b434f560cc53b074c23f4aad1619014f0f9be465

Seefin:
> I've slightly squashed the commits, without changing anything: https://github.com/restic/restic/compare/16d0ab6f4e1ceaf35c2bcbf108b3e8f6e7a9ae34..b434f560cc53b074c23f4aad1619014f0f9be465

Fair enough! Happy to have been able to be some help.

Issue #4869, opened 2024-06-18
Support btime on ext4


tchatow:
Output of `restic version`
--------------------------

restic 0.16.0 compiled with go1.20.6 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
As of recent linux kernels the btime (birth time) is supported via statx on ext4 filesystems. Restic should support backing up this metadata. While it is not possible to restore it to the filesystem without some debugfs commands, it can be exposed via fuse mount and would be of use for advanced restoration operations.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

restic should back up creation time metadata on linux similar to Windows.

greatroar:
#5104 has code to retrieve the btime, but not yet to store it. We could call it btime, which is btrfs's term, or crtime, which seems to be what ZFS calls it.

Issue #5021, opened 2024-08-26
Refactor and cleanup fs code from internal/restic


MichaelEischer:
The `internal/restic` package contains a large amount of code that directly interacts with the filesystem in order to fill in information for `restic.Node` objects. This code should move into the `internal/fs` package.

Besides that the code should be cleaned up in several regards:
- [ ] move code and reduce API interface of fs package
- [ ] limit our custom fs code to backup/restore usage. For other code like the backend or cache, the go standard library should be mostly up to the task.
- [ ] work with file handles instead of paths to properly handle disappearing files https://github.com/restic/restic/issues/3098

MichaelEischer:
> Would this change also allow to detect windows reparse points which cannot be backed up ?

I'm not exactly sure what you mean by that. Starting from Go 1.23.0, reparse points will be treated like normal directories and thus restic will start backing those up. In fact you could just give restic 0.17.1 a try which is built using Go 1.23.1.

deajan:
> I'm not exactly sure what you mean by that. Starting from Go 1.23.0, reparse points will be treated like normal directories and thus restic will start backing those up. In fact you could just give restic 0.17.1 a try which is built using Go 1.23.1.

Wow, now this is really bad news for a lot of Nextcloud / Onedrive / Dropbox / Whatever  users.
I've just made a test on Windows 11 with Nextcloud client and restic 0.17.1.

When using windows reparse points (eg virtual place holders for cloud files), you get to have a file hierarchy on your disk, which actually is stored in cloud, and you only get a local copy once you open a file, which is downloaded on the fly:
![image](https://github.com/user-attachments/assets/dd801008-b42b-4889-ad73-9363a495e879)

Once I did a backup with restic, my cloud client (Nextcloud desktop) started to download all the reparse points data in order to have a local copy of the file:
![image](https://github.com/user-attachments/assets/89148dfb-4806-48aa-b9ef-b496a9913210)

I don't use virtual cloud files myself (I hate them), but alot of customers do in order to have the full file hierarchy of their company on disk without needing the actual disk space.

Now once restic will run, it will just ask the virtual cloud files provider to download everything, taking huge bandwidth and storage space, we're talking about hundreds of GB for some setups I am aware of.

While I do understand that I could add excludes on every restic windows setup I know, it's a huge pain, and there will be unforseen consequences for people who don't know about the issue and run restic.

[...]

PR #5104, opened 2024-10-27
fs: Reimplement Stat/Lstat on Linux using statx

greatroar:

It reimplements os.Stat and os.Lstat on Linux, using x/sys/unix.Statx. This system call gives access to additional information, notably the birth time of files on btrfs. It also, in principle, allows us to accept or reject cached file information.

The result is quite messy. I think in a further round of refactoring we should change the fs.FS API to provide a Stat that returns a custom FileInfo type, so that we can merge Stat and ExtendedStat, with the default implementation handling os.Stat and sycall.Stat_t. That's for later, though.

There may be another way to get btime and "extended" file information via a separate statx call, but I can't think of any that would be (much) less messy.

#4869 requests the btime be stored by restic. This PR doesn't do so yet, as I think that needs some further discussion.

Checklist
---------

### DIFF
--- a/internal/fs/fs_local.go
+++ b/internal/fs/fs_local.go
@@ -32,20 +32,6 @@ func (fs Local) OpenFile(name string, flag int, perm os.FileMode) (File, error) 
 	return f, nil
 }
 
-// Stat returns a FileInfo describing the named file. If there is an error, it
-// will be of type *PathError.
-func (fs Local) Stat(name string) (os.FileInfo, error) {
-	return os.Stat(fixpath(name))
-}
-
-// Lstat returns the FileInfo structure describing the named file.
-// If the file is a symbolic link, the returned FileInfo
-// describes the symbolic link.  Lstat makes no attempt to follow the link.
-// If there is an error, it will be of type *PathError.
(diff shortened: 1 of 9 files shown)


greatroar:
Oh, I hadn't checked the archiver tests. Making this a draft. Maybe wait for #5021 to be finished before we tackle this.

MichaelEischer:
Just a comment without having looked at the PR in more detail: statx is only available since Linux 4.11, which means that we need a fallback for older kernel versions.

MichaelEischer:
> I think in a further round of refactoring we should change the fs.FS API to provide a Stat that returns a custom FileInfo type, so that we can merge Stat and ExtendedStat,

As a headsup, I'm currently refactoring the fs.FS API to exclusively use ExtendedFileInfo. The refactoring is already quite far along but is part of a massive changeset which will still take a bit until it's ready for a PR.

MichaelEischer:
Do you want to continue with this PR or should we close it?

MichaelEischer:
I"ll close this PR for now as FileInfo has changed in the meantime. Feel free to open a new PR if you think this is still worth pursuing

PR #5129, opened 2024-11-04
Allow excluding xattrs at restore time

### LINKED ISSUE #5089: Allow to exclude xattrs (or specific xattr namespaces) when restoring
Output of `restic version`
--------------------------

restic 0.17.1 
compiled with go1.22.5  on linux/amd64

First off, restic is an amazing tool! Thanks for all your great work.  We've been using restic in VolSync (https://github.com/backube/volsync)  to perform backups in kubernetes environments.

This one is a bit complicated - Essentially we're running restic in a kubernetes pod and hit an issue with v0.17.1.  I think this was always here, but the error would be ignored, but this fix means restoring xattrs as UID 0 will now fail the restore:  https://github.com/restic/restic/pull/4958

The issue comes from the fact that we have some scenarios where we run as UID 0 so that we can restore UIDs and GIDs, but when extended attributes in the `trusted.` or `system.` namespace are restored, this will fail as we don't have the CAP_SYS_ADMIN capability.  Additionally it seems that even with this capability, we may not be allowed to modify the `security.selinux` xattr on a file (in Openshift this is definitely an issue, perhaps it will happen on other kubernetes envs too).  We don't hit this issue when not running as UID 0 as the error is ignored by restic.

Would it be possible to allow for excluding xattrs (perhaps with a pattern, or perhaps exclude specific xattr namespaces) at restore time?  If it makes sense for consistency we could also do this at backup - but my main concern atm is restore.

Please let me know if you'd be open for such a change - I'd be happy to contribute a PR.

tesshuflower:
- also allows including xattrs
- default if no flags are passed include only xattrs in the user namespace (xattrs starting with user.*)

For: https://github.com/restic/restic/issues/5089

This introduces new command line flags for restic restore:  `--exclude-xattr` and `--include-xattr`.  These are mutually exclusive, similar to --exclude and --include.

Fixes https://github.com/restic/restic/issues/5089

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5089
+++ b/changelog/unreleased/issue-5089
@@ -1,0 +1,22 @@
+Enhancement: Allow including or excluding extended file attributes
+during restore.
+
+# Describe the problem in the past tense, the new behavior in the present
+# tense. Mention the affected commands, backends, operating systems, etc.
+# If the problem description just says that a feature was missing, then
+# only explain the new behavior.
+# Focus on user-facing behavior, not the implementation.
+# Use "Restic now ..." instead of "We have changed ...".
+#
+Restic restore used to attempt to restore all extended file attributes.
+Now two new command line flags are added to restore to control which
+extended file attributes will be restored.
+
+The new flags are `--exclude-xattr` and `--include-xattr`.
+
+If the flags are not provided, restic will default to restoring
+only `user` namespaced extended file attributes on Linux, and all
+extended file attributes on other operating systems.
+
+https://github.com/restic/restic/issues/5089
+https://github.com/restic/restic/pull/5129
--- a/changelog/unreleased/issue-5089
+++ b/changelog/unreleased/issue-5089
@@ -1,22 +1,13 @@
-Enhancement: Allow including or excluding extended file attributes
-during restore.
+Enhancement: Allow including/excluding extended file attributes during restore
 
-# Describe the problem in the past tense, the new behavior in the present
-# tense. Mention the affected commands, backends, operating systems, etc.
-# If the problem description just says that a feature was missing, then
-# only explain the new behavior.
-# Focus on user-facing behavior, not the implementation.
-# Use "Restic now ..." instead of "We have changed ...".
-#
-Restic restore used to attempt to restore all extended file attributes.
+Restic restore attempts to restore all extended file attributes.
 Now two new command line flags are added to restore to control which
 extended file attributes will be restored.
 
 The new flags are `--exclude-xattr` and `--include-xattr`.
 
 If the flags are not provided, restic will default to restoring
-only `user` namespaced extended file attributes on Linux, and all
-extended file attributes on other operating systems.
+all extended file attributes.
 
 https://github.com/restic/restic/issues/5089
 https://github.com/restic/restic/pull/5129
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/restic/restic/internal/debug"
@@ -300,6 +301,14 @@ func getXattrSelectFilter(opts RestoreOptions) (func(xattrName string) bool, err
 		}, nil
 	}
 
-	// no includes or excludes, set default of including all xattrs
+	// User has not specified any xattr includes or excludes
+	if runtime.GOOS == "linux" {
+		// For Linux, set default of including only user.* xattrs
+		return func(xattrName string) bool {
+			shouldInclude, _ := filter.IncludeByPattern([]string{"user.*"}, Warnf)(xattrName)
+			return shouldInclude
+		}, nil
+	}
+	// Not linux, default to including all xattrs
 	return func(_ string) bool { return true }, nil
 }
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"path/filepath"
-	"runtime"
 	"time"
 
 	"github.com/restic/restic/internal/debug"
@@ -301,14 +300,6 @@ func getXattrSelectFilter(opts RestoreOptions) (func(xattrName string) bool, err
 		}, nil
 	}
 
-	// User has not specified any xattr includes or excludes
-	if runtime.GOOS == "linux" {
-		// For Linux, set default of including only user.* xattrs
-		return func(xattrName string) bool {
-			shouldInclude, _ := filter.IncludeByPattern([]string{"user.*"}, Warnf)(xattrName)
-			return shouldInclude
-		}, nil
-	}
-	// Not linux, default to including all xattrs
+	// default to including all xattrs
 	return func(_ string) bool { return true }, nil
 }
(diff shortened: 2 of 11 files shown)


internal/fs/node_xattr.go
```diff
@@ -75,14 +75,17 @@ func nodeFillGenericAttributes(_ *restic.Node, _ string, _ *ExtendedFileInfo) (a
 	return true, nil
 }
 
-func nodeRestoreExtendedAttributes(node *restic.Node, path string) error {
+func nodeRestoreExtendedAttributes(node *restic.Node, path string, xattrSelectFilter func(xattrName string) bool) error {
```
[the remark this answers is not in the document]
tesshuflower: Thanks for that - I tried to update windows and fill in the proper function definition for others that I also missed

cmd/restic/cmd_restore.go
```diff
@@ -219,6 +232,31 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 		res.SelectFilter = selectIncludeFilter
 	}
 
+	if !hasXattrExcludes && !hasXattrIncludes {
+		// set default of including xattrs from the 'user' namespace
+		opts.IncludeXattrPattern = []string{"user.*"}
+	}
+	if hasXattrExcludes {
```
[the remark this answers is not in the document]
tesshuflower: I've moved to a separate function - but please let me know if you'd prefer it in a different location entirely

cmd/restic/cmd_restore.go
```diff
@@ -219,6 +232,31 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 		res.SelectFilter = selectIncludeFilter
 	}
 
+	if !hasXattrExcludes && !hasXattrIncludes {
+		// set default of including xattrs from the 'user' namespace
+		opts.IncludeXattrPattern = []string{"user.*"}
```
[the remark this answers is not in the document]
tesshuflower: I was a bit wary about changing the default - For the moment (after moving to a separate func and updating the windows implementation) I ended up making the default to restore all - partly because I'm not sure the best way to have a separate default for linux only.  Any advice on setting this for linux only? (assuming we still want user.* as a default)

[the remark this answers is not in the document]
MichaelEischer: To run code on Linux only, you could check for `runtime.GOOS=="linux"

cmd/restic/cmd_restore.go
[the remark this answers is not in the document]
MichaelEischer: That's good enough for now. Eventually I'd like to move most code out of the `cmd` folder. But that not a job for this PR.

MichaelEischer (changes requested):
The changes look good so far. To get this over the finishing line the PR definitely needs a changelog entry and ideally also a short section in the restorer documentation.

tesshuflower:
@MichaelEischer Thanks very much for your review - I've had to rebase to get the latest, hopefully I did everything correctly.  Latest updates
- add unit test for xattr select filter
- add changelog entry
- add info in restore doc about the new flags
- set default (for Linux only) to restore only `user.*` xattrs

For the last one is this ok?  Let me know if you would prefer to keep the default of restoring all xattrs for Linux as well.

changelog/unreleased/issue-5089
```diff
@@ -0,0 +1,22 @@
+Enhancement: Allow including or excluding extended file attributes
+during restore.
+
+# Describe the problem in the past tense, the new behavior in the present
```
MichaelEischer, line 4: Please remove all lines starting with `# `

changelog/unreleased/issue-5089
```diff
@@ -0,0 +1,22 @@
+Enhancement: Allow including or excluding extended file attributes
```
MichaelEischer, line 1: The summary must be a single line shorter than 80 characters and not include trailing punctuation.

`Enhancement: Allow including/excluding extended file attributes during restore` would work even though it's a bit of cheating.

cmd/restic/cmd_restore.go
```diff
@@ -300,6 +301,14 @@ func getXattrSelectFilter(opts RestoreOptions) (func(xattrName string) bool, err
 		}, nil
 	}
 
-	// no includes or excludes, set default of including all xattrs
+	// User has not specified any xattr includes or excludes
+	if runtime.GOOS == "linux" {
```
MichaelEischer, line 305: I've thought a bit about this and by now I think that keeping the old behavior of restoring all xattrs by default makes more sense. Users expect that restic restores POSIX ACLs and per-file capabilities by default. Those are part of the `system.*` namespace, so they shouldn't be excluded by default. So let's just include all xattrs on all OS by default.

doc/050_restore.rst
```diff
@@ -88,6 +88,22 @@ disk space. Note that the exact location of the holes can differ from those in
...
 
+Restoring extended file attributes
+----------------------------------
+
+By default, user namespaced extended attributes for files are restored on Linux,
+and all extended attributes are restored for other operating systems.
+
+Use ``--exclude-xattr`` and ``--include-xattr`` to control which extended
+attributes are restored for files in the snapshot. For example, to restore
+user and security namespaced extended attributes for files:
```
MichaelEischer, line 99: Add `only` at the beginning of this line to make it clearer that specifying an `--include` limits the restored xattrs to the set matched by the filter. This is relevant with a default of restoring all xattrs.

MichaelEischer (changes requested):
Sorry for the long review delay. I found a few further small nits, but besides that the PR is ready. Although you'll have to rebase one last time :see_no_evil: .

cmd/restic/cmd_restore.go
tesshuflower: Thanks, that makes sense to me -  I've also updated the docs to indicate all xattrs are restored by default.

msfrucht: Maybe a default exclude during restore only and backup all since the issue identified security.*, trusted.*, and system.* as problematic on restore? Seems safer to me. No issue reading the security.* xattr from testing, just a problem on restore.

If the exclude default is on restore than at least it is still in the repo from the backup if needed.

I fully admit to not knowing enough about SELinux works to say with any confidence other than a suggestion to consider for someone more knowledgeable.

Sorry to just come out of nowhere, I recently ran into this same issue on OpenShift with restic.

MichaelEischer (approved):
LGTM. Thanks!

tesshuflower:
@MichaelEischer sorry I think the failed test is because I added a random xattr (not in one of the supported linux namespaces) into one of the tests at one point and forgot to remove it. I pushed another commit to roll that part back.

MichaelEischer:
The test on Windows still fails.

> --- FAIL: TestOverwriteXattrWithSelectFilter (0.00s)
>     node_xattr_all_test.go:79: xattr user.foo not restored

`setAndVerifyXattrWithSelectFilter` only converts the name in attrs to uppercase but not in testAttr.

tesshuflower:
> The test on Windows still fails.
> 
> > --- FAIL: TestOverwriteXattrWithSelectFilter (0.00s)
> > node_xattr_all_test.go:79: xattr user.foo not restored
> 
> `setAndVerifyXattrWithSelectFilter` only converts the name in attrs to uppercase but not in testAttr.

Thanks for fixing that and getting it merged! 🥳

Issue #3098, opened 2020-11-15
Ignore disappeared source files


gadall:
Output of `restic version`
--------------------------

restic 0.11.0 compiled with go1.15.3 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Make an option to exit with 0 (zero) when the only errors were source files not found.
Also, in the message "failed to read all source data during backup", "all" could perhaps be changed to "some".

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I'm running the backup over a large directory hierarchy within which temporary files occasionally appear briefly. I'm getting error alerts from the wrapper script running restic, when it seems a temporary file is found during the scan but gone by the time restic tries to back it up.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

It is my sunshine. My only sunshine.

rawtaz:
@gadall In case you don't know already, this was changed in #2546, see under "Change #2546: Return exit code 3 when failing to backup all source data" in https://github.com/restic/restic/blob/master/CHANGELOG.md for reference.

Restic returns a specific exit code for the case where it failed to read all/some of the files you told it to back up, namely exit code 3. You can use that to identify this condition, and in the long run a proper solution to backing up files in a transient filesystem is to use snapshots. E.g. LVM, ZFS and BTRFS has snapshot functionality, and in Windows you have VSS which is supported by restic since version 0.11.0.

Given that there are proper ways to not have files disappear during backup runs, and that you can identify when this happens (as opposed to any other type of error) if it were to happen, it's unlikely that we'll add the option you request.

gadall:
@rawtaz 
`if !success {
	return ErrInvalidSourceData
}
`
Does this actually cover only file not found, and no other read errors? Other errors might be worth paying attention to.
A file simply not existing anymore, can only be considered an error outside of a backup utility's scope, if at all. We didn't delete it, why it was deleted is none of our concern. We're certainly not going to back it up, since it doesn't exist. We can move on.
Yes, a snapshot would help here. However I was thinking that can be a heavier-weight solution for harder problems to solve, such as files being _modified_ on the fly, or being unreadable for whatever weird reasons under Windows,

gadall:
[...]

PR #5143, opened 2024-11-15
fs: rework FS interface to be handle based

MichaelEischer:
This PR is the first step towards better handling disappearing files in the archiver. Currently, the archiver returns errors if a file disappears between calling Lstat on it and reading the remaining attributes / file content. This case can for example occur if a program concurrently renames a temporary file to its final filename.

Besides using snapshots at the filesystem level (not to confuse with snapshots created by restic) the only clean alternative is to open a file handle and use it throughout the archiver.

This PR prepares the FS interface to internally implement the File interface using file handles. The actual conversion to file handles will happen in a follow-up PR (it turned out to be far, far more complex than I've originally anticipated).

The main changes of the FS and File interfaces are the following:
- `FS.OpenFile(name string, flag int, metadataOnly bool) (File, error)`: setting metadataOnly creates a metadata only file handle, whereas not setting it opens the file normally. This distinction is necessary as the filesystem APIs of for example Linux require setting O_PATH to open a file handle for a symlink. Both handle variants are implemented using the File interface as properly separating them significantly complicates testing using mocks. For now, metadata handles are no real handles, but instead internally store the filepath and then just issue normal filesystem operations.
- `File.MakeReadable() error` converts a metadata handle into a regular file handle. For a file-handle-based File implementation this will atomically allow upgrading a metadata to a regular handle.
- `File.ToNode(ignoreXattrListError bool) (*restic.Node, error)`: replaces `FS.NodeFromFileInfo`. The main change is that the FileInfo and filepath passing is handled internally. The integration with the File interface also allows the internal handle to be used.

The flow in the archiver then looks as follows:
- Open metadata handle and decide whether a file/directory should be backed up
- Use `MakeReadable` to upgrade the handle
- Read file / directory contents and metadata

A file handle based implementation of the `File` interface will be able to use the same handle during the whole backup flow and therefore ensure that the file cannot disappear in the middle of the backup (-> implementation will happen in a follow-up PR).

Besides that the PR adds `O_DIRECTORY` or a strict restic.Node.Type check before reading the directory content. There are also lots of cleanups to simplify the code, see the individual commits for more details.

Part of https://github.com/restic/restic/issues/5021
Prerequisite to fix the xattr part of https://github.com/restic/restic/issues/3098

Checklist
---------

- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).~~ No user visible changes expected.


MichaelEischer (commented):
LGTM.

PR #5144, opened 2024-11-16
tag: output the original ID and new snapshotID

### LINKED ISSUE #5137: Output new snapshot ID whenever a command causes an existing ID to change
Output of `restic version`
--------------------------

restic 0.17.1 compiled with go1.23.1 on linux/amd64

For example when adding tags to a snapshot, it would be helpful if the output returned the resulting new snapshot ID.

Finding the same snapshot again after its ID has changed. Having to list snapshots again to find the new ID, and potentially having to compare using other metadata fields to try to ensure that it's the same one.

Sometimes one might make a note of a specific ID because the snapshot was created for a specific purpose (eg: before an upgrade), or because it was determined to be the one needed out of a set of older ones. (Yes, one could also use tags to mark these, but that's not really the point.)

Perhaps it would also be possible to (optionally) list all previous IDs, although I think only the most recent previous one is recorded.

It makes me happy every time it Just Works™, which is almost always.

konidev20:
This PR adds the ability of the restic tag command to return the modified snapshot ID when the tag command is run.

Closes #5137 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_tag.go
+++ b/cmd/restic/cmd_tag.go
@@ -69,8 +69,8 @@ type changedSnapshot struct {
 }
 
 type changedSnapshotsSummary struct {
-	MessageType          string `json:"message_type"` // summary
-	ChangedSnapshotCount int    `json:"changed_snapshot_count"`
+	MessageType      string `json:"message_type"` // summary
+	ChangedSnapshots int    `json:"changed_snapshots"`
 }
 
 func changeTags(ctx context.Context, repo *repository.Repository, sn *restic.Snapshot, setTags, addTags, removeTags []string, printFunc func(changedSnapshot)) (bool, error) {
@@ -135,12 +135,12 @@ func runTag(ctx context.Context, opts TagOptions, gopts GlobalOptions, term *ter
 		Verboseff("old snapshot ID: %v -> new snapshot ID: %v\n", c.OldSnapshotID, c.NewSnapshotID)
 	}
 
-	summary := changedSnapshotsSummary{MessageType: "summary", ChangedSnapshotCount: 0}
+	summary := changedSnapshotsSummary{MessageType: "summary", ChangedSnapshots: 0}
 	printSummary := func(c changedSnapshotsSummary) {
-		if c.ChangedSnapshotCount == 0 {
+		if c.ChangedSnapshots == 0 {
 			Verbosef("no snapshots were modified\n")
 		} else {
-			Verbosef("modified %v snapshots\n", c.ChangedSnapshotCount)
+			Verbosef("modified %v snapshots\n", c.ChangedSnapshots)
 		}
 	}
 
@@ -160,7 +160,7 @@ func runTag(ctx context.Context, opts TagOptions, gopts GlobalOptions, term *ter
 			continue
 		}
 		if changed {
-			summary.ChangedSnapshotCount++
+			summary.ChangedSnapshots++
 		}
 	}
 
--- a/cmd/restic/cmd_tag.go
+++ b/cmd/restic/cmd_tag.go
@@ -9,6 +9,8 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
+	"github.com/restic/restic/internal/ui"
+	"github.com/restic/restic/internal/ui/termstatus"
 )
 
 var cmdTag = &cobra.Command{
@@ -34,7 +36,9 @@ Exit status is 12 if the password is incorrect.
 	GroupID:           cmdGroupDefault,
 	DisableAutoGenTag: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runTag(cmd.Context(), tagOptions, globalOptions, args)
+		term, cancel := setupTermstatus()
+		defer cancel()
+		return runTag(cmd.Context(), tagOptions, globalOptions, term, args)
 	},
 }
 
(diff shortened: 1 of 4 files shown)


cmd/restic/cmd_tag.go
```diff
@@ -84,19 +90,21 @@ func changeTags(ctx context.Context, repo *repository.Repository, sn *restic.Sna
...
 
 		// Remove the old snapshot.
 		if err = repo.RemoveUnpacked(ctx, restic.SnapshotFile, *sn.ID()); err != nil {
-			return false, err
+			return nil, false, err
 		}
 
 		debug.Log("old snapshot %v removed", sn.ID())
+
+		return &changedSnapshot{Original: *sn.ID(), New: id}, changed, nil
```
MichaelEischer, line 105: The information about changed snapshots should be printed right when the snapshot was modified. Otherwise, the output will be lost if the `tag` command is canceled.

cmd/restic/cmd_tag.go
```diff
@@ -58,7 +59,12 @@ func init() {
 	initMultiSnapshotFilter(tagFlags, &tagOptions.SnapshotFilter, true)
 }
 
-func changeTags(ctx context.Context, repo *repository.Repository, sn *restic.Snapshot, setTags, addTags, removeTags []string) (bool, error) {
+type changedSnapshot struct {
```
MichaelEischer, line 62: The message should have a MessageType field like the other json output.
```
MessageType      string   `json:"message_type"` 
```

cmd/restic/cmd_tag.go
```diff
@@ -58,7 +59,12 @@ func init() {
...
+	Original restic.ID `json:"original"`
```
MichaelEischer, line 63: The field name doesn't specify that this is a snapshot id.

cmd/restic/cmd_tag.go
```diff
@@ -115,23 +123,40 @@ func runTag(ctx context.Context, opts TagOptions, gopts GlobalOptions, args []st
...
+		encoder := json.NewEncoder(gopts.stdout)
+		if encoder != nil {
+			err := encoder.Encode(changedSnapshots)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		for _, changedSnapshot := range changedSnapshots {
+			Printf("%v -> %v\n", changedSnapshot.Original, changedSnapshot.New)
```
MichaelEischer, line 158: Let's only print this information when the user specified `--verbose`. That is use `Verboseff` here.

The message also does not state what the value left and right of the arrow is.

MichaelEischer (changes requested):

cmd/restic/cmd_tag.go
konidev20: Done.

cmd/restic/cmd_tag.go
konidev20: Done.

cmd/restic/cmd_tag.go
konidev20: Done. Added SnapshotID to the name,

doc/075_scripting.rst
```diff
@@ -722,6 +722,30 @@ The stats command returns a single JSON object.
...
+
+The ``tag`` command uses the JSON lines format with the following message types.
+
+Changed
+^^^^^^^
+
++--------------------------+-------------------------------------------+
+| ``message_type``         | Always "changed"                          |
++--------------------------+-------------------------------------------+
+| ``original_snapshot_id`` | ID of the snapshot before the change      |
```
MichaelEischer, line 736: After discussion with @rawtaz, let's rename this to `old_snapshot_id`. The term original is overloaded in the context of the `tag` command as command also sets a field `original` when first modifying a snapshot.

doc/075_scripting.rst
```diff
@@ -722,6 +722,30 @@ The stats command returns a single JSON object.
...
+| ``new_snapshot_id``      | ID of the snapshot after the change       |
++--------------------------+-------------------------------------------+
+
+Summary
+^^^^^^^
+
++-----------------------------+-------------------------------------------+
+| ``message_type``            | Always "summary"                          |
++-----------------------------+-------------------------------------------+
+| ``changed_snapshot_count``  | Total number of changed snapshots         |
```
MichaelEischer, line 747: Please rename this to `changed_snapshots`. The other "number of" fields also overwhelmingly don't use the `_count` suffix.

MichaelEischer (changes requested):
Thanks for the changes so far and sorry for the late review. I have a few more small nits.

cmd/restic/cmd_tag.go
```diff
@@ -114,24 +131,44 @@ func runTag(ctx context.Context, opts TagOptions, gopts GlobalOptions, args []st
 	}
 	defer unlock()
 
-	changeCnt := 0
+	printFunc := func(c changedSnapshot) {
+		Verboseff("original snapshot ID: %v -> new snapshot ID: %v\n", c.OriginalSnapshotID, c.NewSnapshotID)
```
MichaelEischer, line 135: `old snapshot ID`.

doc/075_scripting.rst
konidev20: Done. I have renamed this to `old_snapshot_id` now.

cmd/restic/cmd_tag.go
konidev20: Done.

doc/075_scripting.rst
konidev20: Done. Renamed this to `changed_snapshots`.

MichaelEischer (approved):
LGTM. Thanks!

Issue #2165, opened 2019-02-06
lstat No Such file or directory


TheLastProject:
I wasn't sure if I should've added this to #692 or not, but /proc isn't really a place you want to backup, and this happened on a normal place to backup. I apologize if adding it to that issue was better.

Output of `restic version`
--------------------------
restic 0.9.3 compiled with go1.11.1 on linux/amd64

How did you run restic exactly?
-------------------------------
From cron:
```
MAILTO=""
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
RESTIC_REPOSITORY=azure:<hidden>
RESTIC_PASSWORD=<hidden>
AZURE_ACCOUNT_NAME=<hidden>
AZURE_ACCOUNT_KEY=<hidden>
0 */4 * * * restic restic backup /var/www/nextcloud/data >>/var/log/restic/restic.log 2>>/var/log/restic/error.log
```

Stderr output:
```
error: lstat /var/www/nextcloud/data/test.user/uploads/1036227155/00000005: no such file or directory
error: NodeFromFileInfo: Listxattr: xattr.list /var/www/nextcloud/data/test.user/uploads/1036227155/00000004 : no such file or directory
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Azure Blob Storage

Expected behavior
-----------------
Restic not throwing errors.

Actual behavior
---------------
Restic throws this error. The backup still seems to have succeeded.

Steps to reproduce the behavior
-------------------------------
Unknown

Do you have any idea what may have caused this?
-----------------------------------------------
It seems to me that this file was removed from the file system between when restic found out the file existed and tried to back it up. So a race condition.

Do you have an idea how to solve the issue?
-------------------------------------------
It's hard to tell, especially because you probably want to avoid hiding actual errors. But this error message seems to not be an actual issue, so we probably don't want it logged to the error log. 

Did restic help you or made you happy in any way?
-------------------------------------------------
Yup, it's the first backup solution I've tried in ages that doesn't feel awful to use :)

BenWiederhake:
I just ran into this issue, as a backup operation and a "tar-then-delete" operation ran concurrently.

Temporary workaround: Change the timing in cron.

As some details were left out, I just re-used the form.

Output of `restic version`
--------------------------

`restic 0.9.5 compiled with go1.12.4 on linux/amd64`

How did you run restic exactly?
[...]

PR #5145, opened 2024-11-16
backup: Ignore disappeared files

MichaelEischer:

Ignore "No such file" errors when trying to save a file in the archiver. This PR includes https://github.com/restic/restic/pull/5143 and only adds a single commit. Making the PRs independent is unfortunately not possible as it would require a reimplementation of this PR, which would soon become obsolete.

This PR only fixes the "No such file" error in regards to `lstat` (probably the most common), but not the `Listxattr` errors later on. Those require a file-handle-based implementation that has to wait for a followup PR.

Fixes part of https://github.com/restic/restic/issues/3098
Fixes part of https://github.com/restic/restic/issues/2165

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2165
+++ b/changelog/unreleased/issue-2165
@@ -1,0 +1,16 @@
+Bugfix: Ignore disappeared backup source files
+
+If during a backup files were removed between restic listing the directory
+content and backing up the file in question, the following error could occur:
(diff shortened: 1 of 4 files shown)


MichaelEischer (commented):
LGTM

PR #5146, opened 2024-11-16
fs: Inline ExtendedFileInfo

MichaelEischer:
The methods of the fs.FS and fs.File interfaces no longer return `os.FileInfo` but instead always `*fs.ExtendedFileInfo`. This obsoletes the `DeviceID(...)` and `ExtendedStat(...)` methods of the FS interface. In total, the FS interface now only contains `OpenFile(...)` and `Lstat(...)` plus several path manipulation functions. 

fs.ExtendedFileInfo no longer embeds os.FileInfo but instead duplicates the `Name`, `Mode` and `sys` fields. This significantly cleans up some archiver tests as os.FileInfo no longer has to be wrapped in the test.

No.
Related to https://github.com/restic/restic/pull/5104

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5168, opened 2024-12-01
repository: restrict SaveUnpacked and RemoveUnpacked

MichaelEischer:

The SaveUnpacked and RemoveUnpacked methods of a repository now only allow modifying snapshots. Internal data types used by the repository are now read-only. The repository-internal code can bypass the restrictions by wrapping the repository in an `internalRepository` type.

The restriction itself is implemented by using a new datatype `WriteableFileType` in the SaveUnpacked and RemoveUnpacked methods. This statically ensures that code cannot bypass the access restrictions.

The test changes are somewhat noisy as some of them modify repository internals and therefore require some way to bypass the access restrictions. This works by capturing an `internalRepository` or the `Backend` when creating the Repository using a test helper function.

An alternative would be to use an inner repository type that can modify arbitrary filetypes and wrap a restricting layer around it. However, to effectively hide the inner repository type, this wrapper would require manually adding methods to call all methods of the internal repository type. Compared to the `internalRepository` this barely provides any benefit but complicates code maintenance.

Part of the roadmap for restic 0.18.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -304,7 +304,7 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 // from being saved in a snapshot based on path only
 func collectRejectByNameFuncs(opts BackupOptions, repo *repository.Repository) (fs []archiver.RejectByNameFunc, err error) {
 	// exclude restic cache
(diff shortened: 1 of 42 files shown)


MichaelEischer (commented):
LGTM.

Issue #4636, opened 2024-01-16
Restic 0.16.3 issue on Windows 7


tiker:
Did the new version of Restic 0.16.3 drop support for Windows 7?

Running 0.16.2 worked fine.  Updated to 0.16.3 with ``restic self-update``:
```
C:\Users\tiker>c:\apps\restic\restic.exe self-update
writing restic to c:\apps\restic\restic.exe
find latest release of restic at GitHub
latest version is 0.16.3
download SHA256SUMS
download SHA256SUMS.asc
GPG signature verification succeeded
download restic_0.16.3_windows_amd64.zip
downloaded restic_0.16.3_windows_amd64.zip
saved 25019392 bytes in c:\apps\restic\restic.exe
successfully updated restic to version 0.16.3
```

After this, all attempts to run restic.exe fail with an error which would appear to be related to compiling issues or dropping support for Windows 7 (which I don't see mentioned in the release notes).

I have manually downloaded the file from the Github release page and attempted to run it, still fails with the following:
```
C:\Users\tiker>c:\apps\restic\restic_0.16.3_windows_amd64.exe version
Exception 0xc0000005 0x8 0x0 0x0
PC=0x0

runtime.asmstdcall()
        /usr/local/go/src/runtime/sys_windows_amd64.s:65 +0x75 fp=0x22fca0 sp=0x
22fc80 pc=0x46d335
rax     0x0
rbx     0x1b5e300
rcx     0x1bb4100
rdi     0x7fffffde000
rsi     0x22fea0
rbp     0x22fde0
rsp     0x22fc78
r8      0x0
r9      0x22fee0
r10     0x1b855b8
r11     0x21
r12     0x22fec0
r13     0x1
r14     0x1b5da60
r15     0x0
rip     0x0
rflags  0x10293
cs      0x33
fs      0x53
gs      0x2b
```
Previous version:
```
C:\Users\tiker>c:\apps\restic\restic.exe version
restic 0.16.2 compiled with go1.21.3 on windows/amd64
```

DRON-666:
It's not `restic` but a `go` compiler. You can recompile `restic` with old `go1.21.3` or recompile `go1.21.6` compiler itself, but revert [this](https://go-review.googlesource.com/c/go/+/545355) commit.

DRON-666:
@tiker You can try this `cmd/bat` script:

```cmd
@echo off
[...]

PR #5170, opened 2024-12-02
fix: Windows VSS Event ID 8194

### LINKED ISSUE #5169: Windows VSS event ID 8194
Output of `restic version`
--------------------------
restic 0.17.3 compiled with go1.23.4 on windows/amd64

rest api

1. Start a rest-server using the [official repository](https://github.com/restic/rest-server)(actually using a local backend should reproduce this issue).
2. run restic.exe backup in Windows with --use-fs-snapshot flag.
```
restic.exe -r [repository] backup --use-fs-snapshot [paths to backup]
```
3. Look at the event log viewer(press Win+R then type eventvwr) then navigate to Windows Logs -> Application.
4. There should be event logs like this.
```
Volume Shadow Copy Service error: Unexpected error querying for the
IVssWriterCallback interface. hr = 0x80070005, Access is denied.
. This is often caused by incorrect security settings in either the writer or requestor process.

Operation:
Gathering Writer Data

Context:
Writer Class Id: {e8132975-6f93-4464-a53e-1050253ae220}
Writer Name: System Writer
Writer Instance ID: {325f1613-7240-49af-865e-82927f406f7a}
```
Expected behavior
-----------------

Such event logs should not be generated.

Actual behavior
---------------

Event error log with event id 8194 occurs several times when running the backup with VSS in Windows.

As described in [vssclient sample](https://github.com/microsoft/Windows-classic-samples/blob/d338bb385b1ac47073e3540dbfa810f4dcb12ed8/Samples/VShadowVolumeShadowCopy/cpp/vssclient.cpp#L52-L63) initializing the COM security is required to allow VSS writers to communicate with VSS service.

Yes.

Snshadow:

Prevent Windows event 8194 log occurring when doing a backup with --use-fs-snapshot flag.

Closes #5169 

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5170
+++ b/changelog/unreleased/pull-5170
@@ -1,6 +1,22 @@
-BugFix: Prevent Windows VSS event log 8194
+Bugfix: Prevent Windows VSS event log 8194 from being created with `restic backup --use-fs-snapshot`
 
-Initialize VSS security to prevent VSS 8194 event logs occurring when running `restic backup` with `--use-fs-snapshot` flag in Windows.
+When running `restic backup` with `--use-fs-snapshot` flag in Windows with admin right, event logs like
+```
+Volume Shadow Copy Service error: Unexpected error querying for the IVssWriterCallback interface.  hr = 0x80070005, Access is denied.
+. This is often caused by incorrect security settings in either the writer or requestor process. 
+
+Operation:
+   Gathering Writer Data
+
+Context:
+   Writer Class Id: {e8132975-6f93-4464-a53e-1050253ae220}
+   Writer Name: System Writer
+   Writer Instance ID: {54b151ac-d27d-4628-9cb0-2bc40959f50f}
+```
+are created several times(the backup itself succeeds).
+To resolve this, initialize VSS security by calling `CoInitializeSecurity` following the approach from
+https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/VShadowVolumeShadowCopy/cpp/vssclient.cpp#L52-L63 
+before initializing VSS instance in Windows.
 
 https://github.com/restic/restic/issues/5169
 https://github.com/restic/restic/pull/5170
--- a/changelog/unreleased/pull-5170
+++ b/changelog/unreleased/pull-5170
@@ -1,0 +1,7 @@
+BugFix: Prevent Windows VSS event log 8194
+
+Initialize VSS security to prevent VSS 8194 event logs occurring when running `restic backup` with `--use-fs-snapshot` flag in Windows.
+
+https://github.com/restic/restic/issues/5169
+https://github.com/restic/restic/pull/5170
+https://forum.restic.net/t/windows-shadow-copy-snapshot-vss-unexpected-provider-error/3674/2
--- a/changelog/unreleased/pull-5170
+++ b/changelog/unreleased/pull-5170
@@ -1,6 +1,6 @@
-Bugfix: Prevent Windows VSS event log 8194 from being created with `restic backup --use-fs-snapshot`
+Bugfix: Prevent Windows VSS event log 8194 warnings for backup with fs snapshot
 
-When running `restic backup` with `--use-fs-snapshot` flag in Windows with admin right, event logs like
+When running `restic backup` with `--use-fs-snapshot` flag in Windows with admin rights, event logs like
 ```
 Volume Shadow Copy Service error: Unexpected error querying for the IVssWriterCallback interface.  hr = 0x80070005, Access is denied.
 . This is often caused by incorrect security settings in either the writer or requestor process. 
@@ -13,10 +13,8 @@ Context:
    Writer Name: System Writer
    Writer Instance ID: {54b151ac-d27d-4628-9cb0-2bc40959f50f}
 ```
-are created several times(the backup itself succeeds).
-To resolve this, initialize VSS security by calling `CoInitializeSecurity` following the approach from
-https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/VShadowVolumeShadowCopy/cpp/vssclient.cpp#L52-L63 
-before initializing VSS instance in Windows.
+are created several times(the backup itself succeeds). Prevent this from occurring.
+
 
 https://github.com/restic/restic/issues/5169
 https://github.com/restic/restic/pull/5170
--- a/internal/fs/vss_windows.go
+++ b/internal/fs/vss_windows.go
@@ -810,6 +810,17 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
 		}
 	}
 
+	// initialize COM security for VSS, this can't be called more then once
+	if err = ole.CoInitializeSecurity(
+		-1,   // Allow *all* VSS writers to communicate back!
+		5,    // RPC_C_AUTHN_LEVEL_PKT_INTEGRITY
+		3,    // RPC_C_IMP_LEVEL_IMPERSONATE
+		0x40, // EOAC_DYNAMIC_CLOAKING
+	); err != nil {
+		// TODO: warn for expected events for VSS failure event
+		fmt.Printf("VSS warning")
+	}
+
 	var oleIUnknown *ole.IUnknown
 	result, _, _ := vssInstance.Call(uintptr(unsafe.Pointer(&oleIUnknown)))
 	hresult := HRESULT(result)
--- a/internal/fs/vss_windows.go
+++ b/internal/fs/vss_windows.go
@@ -18,24 +18,6 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-var (
-	isPreWin8 bool
-)
-
-func init() {
-	// see https://en.wikipedia.org/wiki/Windows_NT#Releases
-	isPreWin8 = func() bool {
-		major, minor, _ := windows.RtlGetNtVersionNumbers()
-		if major >= 10 {
-			return false
-		} else if major == 6 && minor >= 2 {
-			return false
-		}
-
-		return true
-	}()
-}
-
 // HRESULT is a custom type for the windows api HRESULT type.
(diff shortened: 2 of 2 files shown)


internal/fs/vss_windows.go
```diff
@@ -810,6 +810,17 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
 		}
 	}
 
+	// initialize COM security for VSS, this can't be called more then once
+	if err = ole.CoInitializeSecurity(
+		-1,   // Allow *all* VSS writers to communicate back!
+		5,    // RPC_C_AUTHN_LEVEL_PKT_INTEGRITY
+		3,    // RPC_C_IMP_LEVEL_IMPERSONATE
+		0x40, // EOAC_DYNAMIC_CLOAKING
```
kapsiR, line 818: According to [the documentation](https://learn.microsoft.com/en-us/windows/win32/vss/security-considerations-for-requestors#setting-requester-default-com-access-check-permission), this should be with the options (from Windows 8 onwards):
 - `RPC_C_AUTHN_LEVEL_PKT_PRIVACY`
 - `RPC_C_IMP_LEVEL_IMPERSONATE`
 - EOAC_STATIC

Snshadow: [EOLE_AUTHENTICATION_CAPABILITIES](https://learn.microsoft.com/en-us/windows/win32/api/objidl/ne-objidl-eole_authentication_capabilities) does not have a declaration for `EOAC_STATIC`, there only exists `EOAC_STATIC_CLOAKING` could be the document is not complete or did they mean `EOAC_STATIC_CLOAKING`?

_enumeration from objidl.h_
```cpp
typedef enum tagEOLE_AUTHENTICATION_CAPABILITIES {
    EOAC_NONE = 0x0,
    EOAC_MUTUAL_AUTH = 0x1,
    EOAC_STATIC_CLOAKING = 0x20,
    EOAC_DYNAMIC_CLOAKING = 0x40,
    EOAC_ANY_AUTHORITY = 0x80,
    EOAC_MAKE_FULLSIC = 0x100,
    EOAC_DEFAULT = 0x800,
    EOAC_SECURE_REFS = 0x2,
    EOAC_ACCESS_CONTROL = 0x4,
    EOAC_APPID = 0x8,
    EOAC_DYNAMIC = 0x10,
    EOAC_REQUIRE_FULLSIC = 0x200,
    EOAC_AUTO_IMPERSONATE = 0x400,
    EOAC_NO_CUSTOM_MARSHAL = 0x2000,
    EOAC_DISABLE_AAA = 0x1000
} EOLE_AUTHENTICATION_CAPABILITIES;
```

internal/fs/vss_windows.go
```diff
@@ -811,14 +829,21 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
...
-		5,    // RPC_C_AUTHN_LEVEL_PKT_INTEGRITY
+		6,    // RPC_C_AUTHN_LEVEL_PKT_PRIVACY
 		3,    // RPC_C_IMP_LEVEL_IMPERSONATE
-		0x40, // EOAC_DYNAMIC_CLOAKING
+		capabilities,
 	); err != nil {
-		// TODO: warn for expected events for VSS failure event
-		fmt.Printf("VSS warning")
+		// TODO warn for expected event logs for VSS IVssWriterCallback failure
+		fmt.Printf("WARNING: failed to initialize security for VSS: %v\n", err)
```
Snshadow, line 846: Setting the value to the recommended from the documentation(guessing they meant using `EOAC_STATIC_CLOAKING`), backing up with VSS works properly in Windows 11 and Windows 7(fixing the running issue with [hooking the api](https://github.com/Snshadow/go1.21-over-win7)). Should the value used for CoInitializeSecurity declared in the repository for readability?

internal/fs/vss_windows.go
```diff
@@ -810,6 +828,24 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
 		}
 	}
 
+	// initialize COM security for VSS, this can't be called more then once
+	var capabilities uint32
+	if isPreWin8 {
+		capabilities = 0x00 // EOAC_NONE
+	} else {
+		capabilities = 0x20 // EOAC_STATIC_CLOAKING
+	}
```
kapsiR, line 837: I think we can remove the pre win8 case entirely because [restic doesn't support it since v0.16.2](https://github.com/restic/restic/issues/4636)?

If we want to support it, we have to change more parameters than that. (impersonation level too)

Your thoughts @MichaelEischer ?

internal/fs/vss_windows.go
```diff
@@ -810,6 +828,24 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
...
+
+	if err = ole.CoInitializeSecurity(
```
kapsiR, line 839: Should we add a comment here and a link to the documentation?

e.g.
```go
// Allowing all processes to perform incoming COM calls is not necessarily a security weakness.  
// A requester acting as a COM server, like all other COM servers, always retains the option to authorize its clients on every COM method implemented in its process.
//
// Note that internal COM callbacks implemented by VSS are secured by default.
// Reference: https://learn.microsoft.com/en-us/windows/win32/vss/security-considerations-for-requestors

```

internal/fs/vss_windows.go
```diff
@@ -810,6 +828,24 @@ func initializeVssCOMInterface() (*ole.IUnknown, error) {
...
+		-1,   // Allow *all* VSS writers to communicate back!
```
MichaelEischer, line 840: Wrong comment. `CoInitializeSecurity` always sets a NULL security descriptor, which allows communication from all clients. This parameter is `cAuthSvc`. -1 here means that COM gets to decide which authentication services will be registered (https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializesecurity)

MichaelEischer (changes requested):
It's still a bit unclear in which situation exactly this change is necessary. Can this occur in default Windows setups? Does it require special user accounts etc.?

Snshadow:
> It's still a bit unclear in which situation exactly this change is necessary. Can this occur in default Windows setups? Does it require special user accounts etc.?

I have installed clean Windows 11(24H2) and it occurs if `CoInitializeSecurity` is not called in advance. The frequency of these logs varies depending on each execution.
image

Snshadow:
> It's still a bit unclear in which situation exactly this change is necessary. Can this occur in default Windows setups? Does it require special user accounts etc.?

I have installed clean Windows 11(24H2) and it occurs if `CoInitializeSecurity` is not called in advance. The frequency of these logs varies depending on each execution.

sample

Snshadow:
> It's still a bit unclear in which situation exactly this change is necessary. Can this occur in default Windows setups? Does it require special user accounts etc.?

I have installed clean Windows 11(24H2) and it occurs if `CoInitializeSecurity` is not called in advance. The frequency of these logs varies depending on each execution.

sample

Looking into the comment from https://github.com/restic/restic/blob/ed3922ac82f0ea0f3bbb10a8e01ba8ae51e942c6/internal/fs/vss_windows.go#L898-L900, https://microsoft.public.win32.programmer.kernel.narkive.com/aObDj2dD/volume-shadow-copy-backupcomplete-and-vss-e-bad-state#post2
Looks like the occurrence of this event log has been noted before..

changelog/unreleased/pull-5170
```diff
@@ -1,6 +1,22 @@
-BugFix: Prevent Windows VSS event log 8194
+Bugfix: Prevent Windows VSS event log 8194 from being created with `restic backup --use-fs-snapshot`
```
MichaelEischer, line 1: The title must be shorter than 80 characters.

Something like `Bugfix: Prevent Windows VSS event log 8194 warnings for backup with VSS snapshot` would work.

changelog/unreleased/pull-5170
```diff
@@ -1,6 +1,22 @@
...
 
-Initialize VSS security to prevent VSS 8194 event logs occurring when running `restic backup` with `--use-fs-snapshot` flag in Windows.
+When running `restic backup` with `--use-fs-snapshot` flag in Windows with admin right, event logs like
```
MichaelEischer, line 3: `rights`

changelog/unreleased/pull-5170
```diff
@@ -1,6 +1,22 @@
...
+Operation:
+   Gathering Writer Data
+
+Context:
+   Writer Class Id: {e8132975-6f93-4464-a53e-1050253ae220}
+   Writer Name: System Writer
+   Writer Instance ID: {54b151ac-d27d-4628-9cb0-2bc40959f50f}
+```
+are created several times(the backup itself succeeds).
+To resolve this, initialize VSS security by calling `CoInitializeSecurity` following the approach from
```
MichaelEischer, line 17: The changelog is intended for end users, let's just say that the issue has been solved by initializing the necessary security settings from VSS.

changelog/unreleased/pull-5170
```diff
@@ -1,6 +1,22 @@
...
+https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/VShadowVolumeShadowCopy/cpp/vssclient.cpp#L52-L63
```
MichaelEischer, line 18: That reference is wrong, the PR uses the options from https://learn.microsoft.com/en-us/windows/win32/vss/security-considerations-for-requestors , and not from that sample. But either way, that's too technical for the changelog. If someone want the details, they can look at the pull request.

MichaelEischer (changes requested):
> I have installed clean Windows 11(24H2) and it occurs if CoInitializeSecurity is not called in advance. The frequency of these logs varies depending on each execution.

Thanks for checking!

I have a few more nits in the changelog, other than that the PR seem to be ready.

MichaelEischer (approved):
LGTM. Thanks!

PR #5179, opened 2024-12-15
forget: report count of deleted files correctly

### LINKED ISSUE #5140: `forget` reports that files have been deleted when they can't be
Output of `restic version`
--------------------------

* `restic 0.17.0 compiled with go1.23rc2 on linux/amd64`

[BackUpscale](https://backupscale.com/)

When running in append-only mode, snapshots can't be deleted.  However, `restic forget` reports that they have been, even though they haven't.

Expected behavior
-----------------

Shouldn't the last line here say that "0 / 1" files have been deleted?

```
repository 19897fd5 opened (version 2, compression level auto)
Remove(<snapshot/2eccde165a>) failed: unexpected HTTP response (403): 403 Forbidden
unable to remove snapshot/2eccde165a030dc068e4ec4a092e47da136f0baeb98313925e562f5c095b91fa from the repository
[0:00] 100.00%  0 / 1 files deleted
```

Actual behavior
---------------

Instead, the last line says, "1 / 1 files deleted".  So why it saying something was deleted when it wasn't?  Earlier in the output it says "Forbidden" and the snapshot is still there.  This is a UX hit because users will think that it's actually happening, when it can't be.  And then we all start getting more support requests. :)

```
repository 19897fd5 opened (version 2, compression level auto)
Remove(<snapshot/2eccde165a>) failed: unexpected HTTP response (403): 403 Forbidden
unable to remove snapshot/2eccde165a030dc068e4ec4a092e47da136f0baeb98313925e562f5c095b91fa from the repository
[0:00] 100.00%  1 / 1 files deleted
```

My guess is that it's reporting that number of things that are supposed to be deleted as per the user request, not the number of things that actually have been deleted after the request has been processed.

If it weren't for Restic, I wouldn't be able to work on BackUpScale! So thank you all!!

konidev20:
* fix the issue where the progress counter would be incorrectly updated even when the file deletion failed in append-only mode discussed in gh5140
* I have removed the progress reporter dependency from the `ParallelRemove` function. This will allow the function to be independently tested of the progress reporting logic. 
* I have the responsibility of progress reporting to the report function which is passed as a parameter to the `ParallelRemove` function. I have made the corresponding changes in the other locations. 
* Added unit tests for the changes to ensure that the report function reports the correct number of files deleted. 

Closes #5140

Checklist
---------

### DIFF
--- a/internal/restic/parallel.go
+++ b/internal/restic/parallel.go
@@ -77,13 +77,16 @@ func ParallelRemove[FT FileTypes](ctx context.Context, repo RemoverUnpacked[FT],
 		wg.Go(func() error {
 			for id := range fileChan {
 				err := repo.RemoveUnpacked(ctx, fileType, id)
+				if err == nil {
+					// increment counter only if no error
+					bar.Add(1)
+				}
 				if report != nil {
 					err = report(id, err)
 				}
 				if err != nil {
 					return err
 				}
-				bar.Add(1)
 			}
 			return nil
 		})
--- a/internal/restic/parallel_test.go
+++ b/internal/restic/parallel_test.go
@@ -1,0 +1,106 @@
+package restic
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/ui/progress"
+)
+
+type mockRemoverUnpacked struct {
+	removeUnpacked func(ctx context.Context, t FileType, id ID) error
+}
+
+func (m *mockRemoverUnpacked) Connections() uint {
+	return 2
+}
+
+func (m *mockRemoverUnpacked) RemoveUnpacked(ctx context.Context, t FileType, id ID) error {
+	return m.removeUnpacked(ctx, t, id)
+}
+
+func NewTestID(i byte) ID {
+	return Hash([]byte{i})
+}
+
+func TestParallelRemove(t *testing.T) {
+	ctx := context.Background()
+
+	fileType := SnapshotFile // this can be any FileType
+
+	tests := []struct {
+		name            string
+		removeUnpacked  func(ctx context.Context, t FileType, id ID) error
+		fileList        IDSet
+		wantRemoved     IDSet
+		wantReportIDSet IDSet
+		wantBarCount    int
+	}{
+		{
+			name: "remove files",
+			removeUnpacked: func(ctx context.Context, t FileType, id ID) error {
+				return nil
+			},
+			fileList:        NewIDSet(NewTestID(1), NewTestID(2), NewTestID(3)),
+			wantRemoved:     NewIDSet(NewTestID(1), NewTestID(2), NewTestID(3)),
+			wantReportIDSet: NewIDSet(NewTestID(1), NewTestID(2), NewTestID(3)),
+			wantBarCount:    3,
+		},
+		{
+			name: "remove files with error",
+			removeUnpacked: func(ctx context.Context, t FileType, id ID) error {
+				return errors.New("error")
+			},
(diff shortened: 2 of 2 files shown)


konidev20:
```
func ParallelRemove(ctx context.Context, repo RemoverUnpacked, fileList IDSet, fileType FileType, report func(id ID, err error) error, bar *progress.Counter) error {
```

Hey @MichaelEischer, 

I want to remove the dependence on `*progress.Counter` from here, I believe the report function can update the bar as well since it's a dependency injected from outside. 

This will allow restic internal methods to be tested independently of the progress reporting.

internal/repository/prune.go
```diff
@@ -634,7 +634,8 @@ func deleteFiles(ctx context.Context, ignoreError bool, repo restic.RemoverUnpac
 				return err
 			}
 		}
+		bar.Add(1)
 		printer.VV("removed %v/%v\n", fileType, id)
```
MichaelEischer, line 638: That log output is also wrong if the file deletion failed, but `ignoreError` is set.

internal/restic/parallel.go
```diff
@@ -69,8 +68,6 @@ func ParallelRemove[FT FileTypes](ctx context.Context, repo RemoverUnpacked[FT],
 		return nil
 	})
 
-	bar.SetMax(uint64(len(fileList)))
```
MichaelEischer, line 72: Loosing the information about the total number of files that should be removed is a major problem in my opinion.

In fact, it's not necessary to remove the progress.Counter handling from ParallelRemove. The changed `report` functions are equivalent to the following change in `ParallelRemove`:
```go
err := repo.RemoveUnpacked(ctx, fileType, id)
if err != nil {
    bar.Add(1)
}
if report != nil {
    err = report(id, err)
}
```

MichaelEischer (changes requested):
The removal of the `bar.SetMax` call is a significant regression and not necessary as mentioned in the below comments.

> I want to remove the dependence on `*progress.Counter` from here, I believe the report function can update the bar as well since it's a dependency injected from outside.
> 
> This will allow restic internal methods to be tested independently of the progress reporting.

I don't understand the testing argument. In fact the test case no longer tests the behavior of `ParallelRemove` but rather of `ParallelRemove` + custom progress bar calls in `report` function. For consistent behavior, it feels more reasonable to have progress bar updates handled by `ParallelRemove`. I don't see a technical reason why that wouldn't work.

konidev20:
> I don't understand the testing argument. In fact the test case no longer tests the behavior of ParallelRemove but rather of ParallelRemove + custom progress bar calls in report function. For consistent behavior, it feels more reasonable to have progress bar updates handled by ParallelRemove. I don't see a technical reason why that wouldn't work.

I was thinking that the reporting and progress counters must be together. I can revert the change to the fix I had made earlier. 

Thanks for the review. I will think about this more before making any major changes.

MichaelEischer:
> I was thinking that the reporting and progress counters must be together. I can revert the change to the fix I had made earlier.

Yes, please revert the changes from https://github.com/restic/restic/pull/5179/commits/94ad25e8d35149bf0e690b48f7add71138f19547 . The implementation from https://github.com/restic/restic/pull/5179/commits/5e64578b15e15e5e069eb8754acc7e34db385365 is what I had in mind. The added tests should still work with minor modifications.

I don't think it's worth the trouble to introduce other abstractions, let's just keep using the progress counter.

konidev20:
> The added tests should still work with minor modifications. 

I have made the modifications, let me know if I have missed something.

MichaelEischer (approved):
LGTM. Thanks!

Have you seen https://github.com/restic/restic/issues/5140#issuecomment-2629144531 ? Do you want to address that in this PR or should we consider this a separate issue?

konidev20:
We can address the concerns on a separate PR after discussion.

Issue #5174, opened 2024-12-08
Extended attribute support, on NetBSD


vmlemon:
Output of `restic version`
--------------------------

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

It appears, that since at least NetBSD 10, there has been support for extended attributes, on FFSv2ea volumes (an optional feature, that can be enabled, on existing native FFS volumes, without reformatting them, according to https://wiki.netbsd.org/tutorials/acls_and_extended_attributes_on_ffs), and presumably, ZFS volumes.

I'm unsure, of what the minimum supported NetBSD version is, for Restic, although since it's likely tied to the minimum version supported by Go, we can probably expect support to be available, everywhere Restic can run, now.

Judging from the sermingly-abandoned issue, at https://github.com/borgbackup/borg/issues/1332, that I briefly exhumed, in the name of research, the API is supposed to be broadly similar, to FreeBSD's implementation, but as always, the devil is in the details.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

This would bring NetBSD up to parity, with MacOS, FreeBSD, and Linux, in terms of extended attribute support, and allow for higher-fidelity backups.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Restic helps me keep my data safely, and conveniently at hand, across all of the various devices, and OSes, that I move between, on a regular basis, and gives me the confidence, to know that I can quickly recover from errors, or keep infrequently used data efficiently stored, but readily accessible.

vmlemon:
FWIW, it looks like https://github.com/pkg/xattr/blob/master/xattr_bsd.go has had an implementation, since 2018, but I'm assuming that it hasn't seen much, in the way of testkmg...

MichaelEischer:
Feel free to give this a try by adding netbsd to the list of platforms in https://github.com/restic/restic/blob/6808004ad1b3c5ad52bcc79d9584d98df177abed/internal/fs/node_xattr.go#L1 and remove it from https://github.com/restic/restic/blob/master/internal/fs/node_noxattr.go .

vmlemon:
> Feel free to give this a try by adding netbsd to the list of platforms in
> 
> https://github.com/restic/restic/blob/6808004ad1b3c5ad52bcc79d9584d98df177abed/internal/fs/node_xattr.go#L1
> 
> and remove it from https://github.com/restic/restic/blob/master/internal/fs/node_noxattr.go .

[...]

PR #5180, opened 2024-12-16
Enable xattr support, on NetBSD 10+

vmlemon:

This change enables support for extended attributes, on NetBSD.

This implements the desire, in https://github.com/restic/restic/issues/5174, to use the existing FreeBSD-compatible xattr API, on NetBSD, and has been successfully tested, on NetBSD 10/ARM64, with Go 1.22.7.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5174
+++ b/changelog/unreleased/issue-5174
@@ -1,0 +1,6 @@
+Enhancement: Enable xattr support, on NetBSD 10+
+
+Restic now supports backing up, and restoring extended attributes, on systems running NetBSD 10, or later.
+
+https://github.com/restic/restic/issues/5174
+https://github.com/restic/restic/pull/5180
--- a/internal/fs/node_noxattr.go
+++ b/internal/fs/node_noxattr.go
@@ -1,5 +1,5 @@
-//go:build aix || dragonfly || netbsd || openbsd
-// +build aix dragonfly netbsd openbsd
+//go:build aix || dragonfly || openbsd
+// +build aix dragonfly openbsd
 
 package fs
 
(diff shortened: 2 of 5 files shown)


rawtaz:
@vmlemon Can you check the `internal/fs/node_xattr*_test.go` files as well, I'm suspecting those tests should run for netbsd as well.

rawtaz:
Thanks! I'm curious, did you verify that both backing this file (having xattrs) up _and_ restoring it works fine with the changes in this PR?

vmlemon:
> Thanks! I'm curious, did you verify that both backing this file (having xattrs) up _and_ restoring it works fine with the changes in this PR? I only saw that you tried backing it up in the issue.

It seemed to work fine, with later resting of restoration, and I could view the xattr, via FUSE, on MacOS/Linux, but I haven

vmlemon:
Oops, I'm unsure of how I accidentally closed it, but reopening this, for now...

rawtaz:
Sounds good then. We do need a changelog entry though, can you add a file based on https://github.com/restic/restic/blob/master/changelog/TEMPLATE that sums it up? It will eventually end up in the changelog for the release.

vmlemon:
> Sounds good then. We do need a changelog entry though, can you add a file based on https://github.com/restic/restic/blob/master/changelog/TEMPLATE that sums it up? It will eventually end up in the changelog for the release.

Will do.

Apologies for it being a screenshot, since I don't have SSH access, to the VM configured, but it looks like roundtripping attributes through the ```restore``` option works, just fine:
image

rawtaz:
Thank you very much for the elaborate information, very helpful! Merry Christmas!

rawtaz (approved):

vmlemon:
> Thank you very much for the elaborate information, very helpful! Merry Christmas!

Thanks! Merry Christmas, to you, too!

vmlemon:
> Thank you very much for the elaborate information, very helpful! Merry Christmas!

Thanks! Merry Christmas, to you, too!

PR #5182, opened 2024-12-16
issue 4179: sort output of restic ls -l by size, atime, ctime, mtime, time(=mtime), extension

### LINKED ISSUE #4179: Support different sort for ls, in particular by time
Output of `restic version`
--------------------------
restic 0.15.0 compiled with go1.19.5 on linux/arm

The ls command would be more useful if you could specify different sorting than alphabetically, in particular reverse sort by time, showing the newest files last would be helpful.

In standard Linux/UNIX, the sorting of the output from the ls command can be done in several different ways. I personally rarely am interested in the alphabetical sort when using the -l option, but in stead find my self always typing 'ls -ltr' to get the files most recently modified last. In other cases, sorting by size is useful. Ideally, the restic ls command should support the --sort and --reverse options of the GNU implementation of ls.

Making restic ls behave like Linux ls does.

 
Yep - just upgraded to 0.15. And it saves me money by being the only backup I need.

wplapper:
Enhancement: create ability to sort output of restic ls -l by name, size, atime, ctime, mtime, time(=mtime), X(=extension), extension

https://github.com/restic/restic/issues/4179

Ability to sort output of restic ls -l by name, size, atime, ctime, mtime, time(=mtime)

Add 2 local parameters to sort command: --sort and --reverse, so that the output can be sorted by one of the above named possibilities.

https://github.com/restic/restic/issues/4179
closes https://github.com/restic/restic/issues/4179
Checklist
---------


cmd/restic/cmd_ls.go
```diff
@@ -72,6 +81,8 @@ func init() {
 	flags.BoolVar(&lsOptions.Recursive, "recursive", false, "include files in subfolders of the listed directories")
 	flags.BoolVar(&lsOptions.HumanReadable, "human-readable", false, "print sizes in human readable format")
 	flags.BoolVar(&lsOptions.Ncdu, "ncdu", false, "output NCDU export format (pipe into 'ncdu -f -')")
+	flags.StringVarP(&lsOptions.Sort, "sort", "s", "name", "sort output by [name|size|time(=mtime)|atime|ctime|mtime|X(=extension)|extension]")
+	flags.BoolVarP(&lsOptions.Reverse, "reverse", "R", false, "reverse sorted output")
```
MichaelEischer, line 85: `reverse` only works in combination with `--sort`. Return an error if `--sort` wasn't specified.

cmd/restic/cmd_ls.go
```diff
@@ -348,6 +359,15 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
...
 	var printer lsPrinter
 
+	// for ls -l output sorting
+	type ToSort struct {
+		nodepath string
+		node     *restic.Node
+	}
+
+	collector := []ToSort{}
+	outputSort := true
```
MichaelEischer, line 369: The data format of restic guarantees that the output is implicitly sorted by name. There's no need to store all output in memory in that case.

cmd/restic/cmd_ls.go
```diff
@@ -72,6 +81,8 @@ func init() {
 	flags.BoolVar(&lsOptions.Recursive, "recursive", false, "include files in subfolders of the listed directories")
 	flags.BoolVar(&lsOptions.HumanReadable, "human-readable", false, "print sizes in human readable format")
 	flags.BoolVar(&lsOptions.Ncdu, "ncdu", false, "output NCDU export format (pipe into 'ncdu -f -')")
+	flags.StringVarP(&lsOptions.Sort, "sort", "s", "name", "sort output by [name|size|time(=mtime)|atime|ctime|mtime|X(=extension)|extension]")
```
MichaelEischer, line 84: When invalid values are passed to `--sort`, the `ls` command must fail with an error. Take a look at `CompressionMode` in `internal/repository/repository.go` how a clean solution would look like.

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +469,78 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
...
+			})
+		} else if opts.Sort == "X" || opts.Sort == "extension" {
+			// prepare sort by extension
+			type ToSortExt struct {
+				nodepath  string
+				extension string
+				node      *restic.Node
+			}
+			collector_ext := make([]ToSortExt, len(collector))
+			for ix, item := range collector {
```
MichaelEischer, line 509: Is there a reason for using `ix` instead of the standard `i`?

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +469,78 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
 		return err
 	}
 
+	if outputSort {
+		if opts.Sort == "size" {
+			slices.SortStableFunc(collector, func(a, b ToSort) int {
+				return cmp.Or(
```
MichaelEischer, line 475: cmp.Or is not yet available in Go 1.21 (the current minimum Go version supported by restic). But we could just as well bump the minimum required go version to 1.22.

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +469,78 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
...
+		} else if opts.Sort == "ctime" {
+			slices.SortStableFunc(collector, func(a, b ToSort) int {
+				return cmp.Or(
+					a.node.ChangeTime.Compare(b.node.ChangeTime),
+					cmp.Compare(a.nodepath, b.nodepath),
+				)
+			})
+		} else if opts.Sort == "X" || opts.Sort == "extension" {
+			// prepare sort by extension
+			type ToSortExt struct {
```
MichaelEischer, line 503: Using a `map[string]string` that maps the nodepath to the file extension would allow getting rid of these back and forth copies.

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +469,78 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
 		return err
 	}
 
+	if outputSort {
```
MichaelEischer, line 472: Please extract the sorted output code into a helper function. `runLs` is already far, far too long.

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +469,78 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
...
+		if opts.Sort == "size" {
```
MichaelEischer, line 473: Use `switch opts.Sort {` here.

cmd/restic/cmd_ls.go
MichaelEischer, line 84: Where does `X` originate from? The other values largely match those offered by the unix `ls` command (`none, size, time, version, extension, width`). But `X` isn't part of that list.

cmd/restic/cmd_ls.go
MichaelEischer, line 85: `-R` means `--recursive` for the `ls` command. So I'm wondering whether it makes sense to have a `-R` with a rather different meaning here.

cmd/restic/cmd_ls.go
```diff
@@ -36,6 +39,10 @@ will allow traversing into matching directories' subfolders.
 Any directory paths specified must be absolute (starting with
 a path separator); paths use the forward slash '/' as separator.
 
+File listings can be sorted by specifying --sort followed by one of the
+sort specifiers '[name|size|time|atime|ctime|mtime]'.
```
MichaelEischer, line 43: `extension` is missing from the description.

MichaelEischer (changes requested):
Thanks for working on this feature and the other PRs (I'll slowly dig through those).

I've left a few comments below. Please also add a basic test. You can look at `TestRunLsNcdu` for inspiration.

cmd/restic/cmd_ls.go
MichaelEischer: The general idea with parameter validation is that unsupported/invalid parameters must result in an error. That will allow us to support new values in the future without breaking existing behavior.

As a random example: For example `--sort mode` would right now sort by name as it isn't supported. But when we for some reason want to support `mode` in the future, then that would change the behavior for everyone who's for some reason already using `--sort mode`. By just returning an error when an invalid value was specified, we can ensure that nobody uses those invalid values and therefore are able to redefine the without problems in the future.

changelog/unreleased/issue-4179
```diff
@@ -0,0 +1,6 @@
+ls: create ability to sort output of restic ls -l by one of the following list
```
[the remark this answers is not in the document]
wplapper: done

cmd/restic/cmd_ls.go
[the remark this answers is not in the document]
wplapper: done

cmd/restic/cmd_ls.go
wplapper: fixed

cmd/restic/cmd_ls.go
wplapper: check for name removed

cmd/restic/cmd_ls.go
wplapper: implemented

cmd/restic/cmd_ls.go
wplapper: done

cmd/restic/cmd_ls.go
wplapper: yep, done

cmd/restic/cmd_ls.go
wplapper: X removed

cmd/restic/cmd_ls.go
wplapper: removed short option -R

cmd/restic/cmd_ls.go
wplapper: extension has been added.

cmd/restic/cmd_ls.go
wplapper: I prefer to be able to use SortStableFunc over Sort, since it is generic and also allows to use cmp.Or().

cmd/restic/cmd_ls.go
wplapper: done. Got rid of the copy-in copy-out.

wplapper:
Regards testing: you have touched a sore point. Is there anywhere a document or a set of documents which describes the restic testing machinery? I am looking at TestRunLsNcdu and can see that it runs a backup to repo `env.testdata+"/0"`. from ".". What is `env.testdata+"/0"` and where does "." point to? Eventually the test function runs a couple of `runLs(context.TODO(), opts, gopts, args)` from the backup just generated.

I would like to create a test scenario where I backup 2 or 3 files with different names, extension and sizes. Then I can sort by name, time, size and extension and compare against the expected output.

Any pointers? Thanks!

changelog/unreleased/issue-4179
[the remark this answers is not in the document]
wplapper: 4 basic tests added. I had to expand `backup-data` to get some useful filenames with extensions. Ran all tests with the modified `backup-data` structure.

cmd/restic/cmd_ls.go
wplapper: code not longer there.

MichaelEischer (changes requested):
I have a few more nits. Can you add a short note about the new sort mode to `doc/045_working_with_repos.rst`?

I'll refactor the code a bit after merging, but it's harder to explain what I have in mind than making those changes.

MichaelEischer:
https://github.com/restic/restic/pull/5232 will bump the minimum go version to 1.22, this should fix the failing test in Go 1.21.

changelog/unreleased/issue-4179
```diff
@@ -0,0 +1,11 @@
+Enhancement: add capability sort output by one of given options
```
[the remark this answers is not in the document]
wplapper: done.

cmd/restic/cmd_ls.go
```diff
@@ -444,5 +482,100 @@ func runLs(ctx context.Context, opts LsOptions, gopts GlobalOptions, args []stri
...
 	}
 
+	if outputSort {
+		PrintSortedOutput(printer, opts, sortMode, collector)
+	}
+
 	return printer.Close()
 }
+
+func PrintSortedOutput(printer lsPrinter, opts LsOptions, sortMode SortMode, collector []ToSortOutput) {
```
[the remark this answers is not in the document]
wplapper: done.

cmd/restic/cmd_ls.go
```diff
@@ -277,13 +288,33 @@ func (p *textLsPrinter) Close() error {
 	return nil
 }
 
+// for ls -l output sorting
+type ToSortOutput struct {
```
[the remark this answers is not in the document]
wplapper: done.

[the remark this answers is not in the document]
wplapper: done

cmd/restic/cmd_ls.go
```diff
@@ -277,13 +288,33 @@ func (p *textLsPrinter) Close() error {
...
 	}
+	if opts.Sort == "" && opts.Reverse {
+		return errors.Fatal("--reverse without specifying --sort is not allowed")
+	}
+	if opts.Sort != "" && opts.Ncdu {
+		return errors.Fatal("--sort and --ncdu are mutually exclusive")
+	}
+
+	sortMode := SortModeName
+	if opts.Sort != "" {
```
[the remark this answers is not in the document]
wplapper: removed the if cause.

cmd/restic/cmd_ls.go
```diff
@@ -36,6 +39,10 @@ will allow traversing into matching directories' subfolders.
 Any directory paths specified must be absolute (starting with
 a path separator); paths use the forward slash '/' as separator.
 
+File listings can be sorted by specifying --sort followed by one of the
+sort specifiers '(name|size|atime|ctime|mtime=time|extension)'.
```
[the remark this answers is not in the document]
wplapper: I hope I changed it to be the same.

cmd/restic/cmd_ls.go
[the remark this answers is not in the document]
wplapper: done.

MichaelEischer:
I've rebased the PR (to have Go 1.22 as minimum version), and made a few cleanups.

MichaelEischer (approved):

MichaelEischer:
Thanks for keeping up with all the review comments :-) . I've opened https://github.com/restic/restic/pull/5235 in case you're interested in the refactorings I had in mind.

PR #5183, opened 2024-12-16
restic prune: selection of packs to repack based on size

### LINKED ISSUE #5109: Selection of packs to repack based on size
Output of `restic version`
--------------------------

restic 0.17.2 compiled with go1.23.2 on linux/amd64

`restic prune` should have an option to select the packs which are repacked when using `--repack-small`. My suggestion would be to add an option `--repack-smaller-than` to select packs smaller than a particular size. One could then do

`restic prune --repack-small --repack-smaller-than 16M --pack-size 64M` 

to repack really small packs (e.g. 4M).

If the pack size has changed in a repository over time, there can be parts of data in differently sized packs. Let's assume there are 100 GB in 4M packs, 1 TB in 16M packs and 1 TB in 64M packs. Repacking everything to 64M takes a long time, but repacking those 4M packs to 64M would be doable and reduce the number of packs significantly.

More discussion in https://forum.restic.net/t/repack-smallest-packs/8621/12

Of course, there would be many other possibilities for pack selections. The one I am proposing here is a selection I had a need for.

Restic is just great. It scans so quickly through a relatively large disk and thus enables frequent backups, the deduplication is impressive. Being able to use one repository from Windows and Linux is also very helpful.

wplapper:
cmd_prune: Selection of packs to repack based on size. Add option --small-pack-size so prune can repack all packfiles smaller than small-pack-size bytes.

Add option --small-pack-size so prune can repack all packfiles smaller than small-pack-size bytes.
The new option is passed onto internal/repository/prune.go.
A new decision has been added to "decide what to do" to
move the intended packfiles onto the repackSmallCandidates queue.

https://github.com/restic/restic/issues/5109
closes https://github.com/restic/restic/issues/5109

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5109
+++ b/changelog/unreleased/issue-5109
@@ -1,8 +1,7 @@
 Enhancement: Make small pack size configureable for `prune`
 
 The `prune` command now supports the `--repack-smaller-than` option that
-allows repacking pack files smaller than the given size. It has to be used in
-conjunction with option `--repack-small`.
+allows repacking pack files smaller than the given size.
 
 https://github.com/restic/restic/issues/5109
 https://github.com/restic/restic/pull/5183
--- a/changelog/unreleased/issue-5109
+++ b/changelog/unreleased/issue-5109
@@ -1,0 +1,8 @@
+Enhancement: Make small pack size configureable for `prune`
+
+The `prune` command now supports the `--small-pack-size` option that
+allows repacking pack files smaller than the given size. It has to be used in
+conjunction with option `--repack-small`.
+
+https://github.com/restic/restic/issues/5109
+https://github.com/restic/restic/pull/5183
--- a/changelog/unreleased/issue-5109
+++ b/changelog/unreleased/issue-5109
@@ -1,6 +1,6 @@
 Enhancement: Make small pack size configureable for `prune`
 
-The `prune` command now supports the `--small-pack-size` option that
+The `prune` command now supports the `--repack-smaller-than` option that
 allows repacking pack files smaller than the given size. It has to be used in
 conjunction with option `--repack-small`.
 
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -143,11 +143,6 @@ func verifyPruneOptions(opts *PruneOptions) error {
 		}
 	}
 
-	if !opts.RepackSmall && opts.SmallPackSize != "" {
-		return errors.Fatal("option --repack-small and --repack-smaller-than: must be specified together")
-	}
-
-	opts.SmallPackBytes = uint64(0)
 	if opts.SmallPackSize != "" {
 		size, err := ui.ParseBytes(opts.SmallPackSize)
 		if err != nil {
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -207,6 +207,7 @@ func runPruneWithRepo(ctx context.Context, opts PruneOptions, gopts GlobalOption
 
 		MaxUnusedBytes: opts.maxUnusedBytes,
 		MaxRepackBytes: opts.MaxRepackBytes,
+		SmallPackBytes: opts.SmallPackBytes,
 
 		RepackCacheableOnly: opts.RepackCacheableOnly,
 		RepackSmall:         opts.RepackSmall,
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -67,6 +67,9 @@ type PruneOptions struct {
 	RepackCacheableOnly bool
 	RepackSmall         bool
 	RepackUncompressed  bool
+
+	SmallPackSize  string
+	SmallPackBytes uint64
 }
 
 func (opts *PruneOptions) AddFlags(f *pflag.FlagSet) {
@@ -81,6 +84,7 @@ func (opts *PruneOptions) AddLimitedFlags(f *pflag.FlagSet) {
 	f.BoolVar(&opts.RepackCacheableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
 	f.BoolVar(&opts.RepackSmall, "repack-small", false, "repack pack files below 80% of target pack size")
 	f.BoolVar(&opts.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
+	f.StringVar(&opts.SmallPackSize, "repack-smaller-than", "", "pack `below-limit` packfiles (allowed suffixes: k/K, m/M)")
 }
 
 func verifyPruneOptions(opts *PruneOptions) error {
@@ -137,6 +141,20 @@ func verifyPruneOptions(opts *PruneOptions) error {
 		opts.maxUnusedBytes = func(_ uint64) uint64 {
 			return uint64(size)
 		}
+	}
+
+	if !opts.RepackSmall && opts.SmallPackSize != "" {
+		return errors.Fatal("option --repack-small and --repack-smaller-than: must be specified together")
+	}
+
+	opts.SmallPackBytes = uint64(0)
+	if opts.SmallPackSize != "" {
+		size, err := ui.ParseBytes(opts.SmallPackSize)
+		if err != nil {
+			return errors.Fatalf("invalid number of bytes %q for --repack-smaller-than: %v", opts.SmallPackSize, err)
+		}
+		opts.SmallPackBytes = uint64(size)
+		opts.RepackSmall = true
 	}
 
 	return nil
--- a/internal/repository/prune.go
+++ b/internal/repository/prune.go
@@ -24,6 +24,7 @@ type PruneOptions struct {
 
 	MaxUnusedBytes func(used uint64) (unused uint64) // calculates the number of unused bytes after repacking, according to MaxUnused
 	MaxRepackBytes uint64
+	SmallPackBytes uint64
 
 	RepackCacheableOnly bool
 	RepackSmall         bool
@@ -103,6 +104,9 @@ func PlanPrune(ctx context.Context, opts PruneOptions, repo *Repository, getUsed
 	}
 	if repo.Config().Version < 2 && opts.RepackUncompressed {
 		return nil, fmt.Errorf("compression requires at least repository format version 2")
+	}
+	if opts.SmallPackBytes > uint64(repo.packSize()) {
+		return nil, fmt.Errorf("repack-smaller-than exceeds repository packsize")
 	}
 
 	usedBlobs := index.NewAssociatedSet[uint8](repo.idx)
@@ -326,7 +330,9 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository, 
 	repoVersion := repo.Config().Version
 	// only repack very small files by default
 	targetPackSize := repo.packSize() / 25
-	if opts.RepackSmall {
+	if opts.SmallPackBytes > 0 {
+		targetPackSize = uint(opts.SmallPackBytes)
+	} else if opts.RepackSmall {
 		// consider files with at least 80% of the target size as large enough
 		targetPackSize = repo.packSize() / 5 * 4
 	}
@@ -402,6 +408,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository, 
 		bar.Add(1)
 		return nil
 	})
+
 	bar.Done()
 	if err != nil {
 		return PrunePlan{}, err
--- a/internal/repository/prune_test.go
+++ b/internal/repository/prune_test.go
@@ -2,6 +2,7 @@ package repository_test
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"math/rand"
 	"testing"
@@ -9,6 +10,7 @@ import (
 
 	"github.com/restic/restic/internal/checker"
 	"github.com/restic/restic/internal/repository"
+	"github.com/restic/restic/internal/repository/pack"
 	"github.com/restic/restic/internal/restic"
 	rtest "github.com/restic/restic/internal/test"
 	"github.com/restic/restic/internal/ui/progress"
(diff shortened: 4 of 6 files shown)


cmd/restic/cmd_prune.go
```diff
@@ -74,10 +77,11 @@ func init() {
 func addPruneOptions(c *cobra.Command, pruneOptions *PruneOptions) {
 	f := c.Flags()
 	f.StringVar(&pruneOptions.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
-	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "stop after repacking this much data in total (allowed suffixes for `size`: k/K, m/M, g/G, t/T)")
+	f.StringVar(&pruneOptions.MaxRepackSize, "max-repack-size", "", "maximum `size` to repack (allowed suffixes: k/K, m/M, g/G, t/T)")
```
MichaelEischer, line 80: Please revert this change, the wording has somewhat recently changed to `stop after repacking this much data in total` to improve the clarity of the description.

cmd/restic/cmd_prune.go
```diff
@@ -74,10 +77,11 @@ func init() {
...
 	f.BoolVar(&pruneOptions.RepackCacheableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
 	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack pack files below 80% of target pack size")
 	f.BoolVar(&pruneOptions.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
+	f.StringVar(&pruneOptions.SmallPackSize, "small-pack-size", "0", "pack `below-limit` packfiles smaller than (allowed suffixes: k/K, m/M, g/G, t/T)")
```
MichaelEischer, line 84: The default value should be an empty string.

cmd/restic/cmd_prune.go
```diff
@@ -149,11 +160,7 @@ func runPrune(ctx context.Context, opts PruneOptions, gopts GlobalOptions, term
 		return errors.Fatal("disabled compression and `--repack-uncompressed` are mutually exclusive")
 	}
 
-	if gopts.NoLock && !opts.DryRun {
-		return errors.Fatal("--no-lock is only applicable in combination with --dry-run for prune command")
-	}
-
-	ctx, repo, unlock, err := openWithExclusiveLock(ctx, gopts, opts.DryRun && gopts.NoLock)
+	ctx, repo, unlock, err := openWithExclusiveLock(ctx, gopts, false)
```
MichaelEischer, line 163: Please revert this change.

internal/repository/prune.go
```diff
@@ -352,7 +356,6 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 				id.Str(), p.unusedSize+p.usedSize, packSize)
 			return ErrSizeNotMatching
 		}
-
```
MichaelEischer, line 355: unnecessary whitespace change.

internal/repository/prune.go
```diff
@@ -385,12 +388,18 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 			// if this is a data pack and --repack-cacheable-only is set => keep pack!
 			stats.Packs.Keep++
 
+		//wpl 2024-11-08 - repack when packSize is smaller than opts.SmallPackBytes
```
MichaelEischer, line 391: There is no need to manually mark who and when some logic was changed, git already takes care of that automatically.

internal/repository/prune.go
```diff
@@ -401,7 +410,8 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 		delete(indexPack, id)
 		bar.Add(1)
 		return nil
-	})
+	}) // end repo.List(ctx, restic.PackFile ...)
```
MichaelEischer, line 413: Unnecessary comment.

internal/repository/prune.go
```diff
@@ -25,6 +25,9 @@ type PruneOptions struct {
 	MaxUnusedBytes func(used uint64) (unused uint64) // calculates the number of unused bytes after repacking, according to MaxUnused
 	MaxRepackBytes uint64
 
+	SmallPackSize  string
```
MichaelEischer, line 28: There's no point in passing the string version of SmallPackSize to `prune` as it only needs `SmallPackBytes`.

internal/repository/prune.go
```diff
@@ -25,6 +25,9 @@ type PruneOptions struct {
...
+	SmallPackBytes uint64
```
MichaelEischer, line 29: Please add a check to prune that `SmallPackBytes <= repo.packSize()`

internal/repository/prune.go
```diff
@@ -385,12 +388,18 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 			// if this is a data pack and --repack-cacheable-only is set => keep pack!
 			stats.Packs.Keep++
 
+		//wpl 2024-11-08 - repack when packSize is smaller than opts.SmallPackBytes
+		case opts.RepackSmall && (uint64(packSize) <= opts.SmallPackBytes):
```
MichaelEischer, line 392: I don't think we need a special case here. Just set `targetPackSize` to `opts.SmallPackBytes` if the latter is non-zero and `opts.RepackSmall` is set.

internal/repository/prune.go
```diff
@@ -434,12 +444,15 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 		}
 	}
 
-	if len(repackSmallCandidates) < 10 {
+	// calculate limit for number of unused bytes in the repo after repacking
+	maxUnusedSizeAfter := opts.MaxUnusedBytes(stats.Size.Used)
+	// wpl we need a mixed calculation based the existing condition and on --max-unused
+	if len(repackSmallCandidates) >= 10 || opts.RepackSmall && (uint64(sizeRepackSmallCandidates) >= maxUnusedSizeAfter) {
```
MichaelEischer, line 450: Please revert this change. If users set `--max-unused 0`, your change causes repeated prune executions to repack the same data each time. The point of repacking small files is reduce the number of files, but a few files more or less won't matter. If `--max-unused` yields a large value (>= 1.28GB to be precise), then the new clause is irrelevant as `len(repackSmallCandidates) >= 10` will always be true first.

changelog/unreleased/issue-5109
```diff
@@ -0,0 +1,10 @@
+Issue #5109 - Selection of packs to repack based on size
```
MichaelEischer, line 1: Please follow the template from https://github.com/restic/restic/blob/master/changelog/TEMPLATE .
The changelog for this PR should look something like the following:
```
Enhancement: Make small pack size configureable for `prune`

The `prune` command now supports the `--small-pack-size` option that
allows repacking pack files smaller than the given size.

https://github.com/restic/restic/issues/5109
https://github.com/restic/restic/pull/5183
```

The changelog entries for restic are intended for end users. Those likely only know the CLI but have no knowledge of any internal code.

MichaelEischer (changes requested):
The change in `internal/repository/prune.go` is more complex than it needs to be. It should be sufficient to just set the  `targetPackSize` to the user specified value. The current approach feels like a duplication of the existing selection of small packs.

The PR also must be rebased onto the current master branch as we generally don't merge PRs that contain merge commits. (See the `git rebase` command / the [git book](https://git-scm.com/book/en/v2/Git-Branching-Rebasing)).

I'd also like to have a small test for the new option, although I can also help with that. The test should probably use an approach similar to that of https://github.com/restic/restic/pull/5212/files .

cmd/restic/cmd_prune.go
wplapper: done

cmd/restic/cmd_prune.go
wplapper: done

cmd/restic/cmd_prune.go
wplapper: done

internal/repository/prune.go
wplapper: done

internal/repository/prune.go
wplapper: comment removed.

internal/repository/prune.go
wplapper: SmallPackSize removed

internal/repository/prune.go
wplapper: added check in decidePackAction

internal/repository/prune.go
wplapper: comment removed

internal/repository/prune.go
wplapper: field SmallPackSize   removed

changelog/unreleased/issue-5109
wplapper: done.

internal/repository/prune.go
MichaelEischer: That check should happen right at the start of `PlanPrune`. `decidePackAction(...)` is too late as prune can have performed a significant amount of processing up to that point.

internal/repository/prune.go
```diff
@@ -374,6 +378,9 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 		}
 
 		// decide what to do
+		if opts.SmallPackBytes > 0 && opts.RepackSmall {
```
MichaelEischer, line 381: Please move that to the point where `targetPackSize`  is initially computed (that is after line 336)

cmd/restic/cmd_prune.go
```diff
@@ -136,6 +140,13 @@ func verifyPruneOptions(opts *PruneOptions) error {
 		}
 	}
 
+	smallPackSize := strings.TrimSpace(opts.SmallPackSize)
+	size, err := ui.ParseBytes(smallPackSize)
```
MichaelEischer, line 144: This needs a special case to not fail if  `smallPackSize` is an emtpy string.

MichaelEischer (changes requested):
Thanks, I found a few more nits

internal/repository/prune.go
wplapper: done

cmd/restic/cmd_prune.go
wplapper: fixed the bug.

wplapper:
I added logic to the case `p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress`: in case we deal with `--repack-small` and have `--small-pack-size`with a non-zero value, these packs get all dropped into the repackSmallCandidates` bucket.

wplapper:
> I'd also like to have a small test for the new option, although I can also help with that. The test should probably use an approach similar to that of https://github.com/restic/restic/pull/5212/files .

The plan for the test is as follows:
1.) create repository  with packsize of 2M.
2.) create enough data for 3 packfiles, something in the range of 7-8 M.
3.) run a `repository.PlanPrune(...)` with a packsize of 16M (current default).
4.) run `plan.Execute(...)`, extract `pan.Stats()`.
5.) The result should be 2 packfiles, one for TreeBlobs and one for DataBlobs.
6.) Check that all blobs are contained in the new packfiles.

wplapper:
> 1.) create repository with packsize of 2M.
> 2.) create enough data for 3 packfiles, something in the range of 7-8 M.
> 3.) run a `repository.PlanPrune(...)` with a packsize of 16M (current default).
> 4.) run `plan.Execute(...)`, extract `plan.Stats()` and check.
> 5.) The result should be 2 packfiles, one for TreeBlobs and one for DataBlobs.
> 6.) Check that all blobs are contained in the new packfiles.

The reason why I change the packsize is the simulation to larger packsizes. There might be other methods of forcing repacking of smaller packfiles, but that's what I did :)

cmd/restic/cmd_prune.go
```diff
@@ -78,6 +81,7 @@ func addPruneOptions(c *cobra.Command, pruneOptions *PruneOptions) {
 	f.BoolVar(&pruneOptions.RepackCacheableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
 	f.BoolVar(&pruneOptions.RepackSmall, "repack-small", false, "repack pack files below 80% of target pack size")
 	f.BoolVar(&pruneOptions.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
+	f.StringVar(&pruneOptions.SmallPackSize, "small-pack-size", "", "pack `below-limit` packfiles smaller than (allowed suffixes: k/K, m/M, g/G, t/T)")
```
MichaelEischer, line 84: I just noticed that the original issue suggested `--repack-smaller-than` as option. Is there a particular reason to deviate from that?

P.S. the description should be  (pack -> repack, packfiles -> pack files)
```
"repack pack files smaller than `below-limit` (allowed suffixes: k/K, m/M, g/G, t/T)"
```
We could also drop the `t/T` suffix as it will always lead to an error. `g/G` currently is also always an error as the pack size is still limited to 128MB, but that could change in the future.

MichaelEischer (changes requested):
I found a few issues with the test, see below. We also have to discuss the method name, in the linked issue the suggestion was to use `--repack-smaller-than`.

```
if packSize >= int64(targetPackSize) && opts.SmallPackBytes == 0 && !opts.RepackSmall {
```
As mentioned in the comment, I have no clue what this is supposed to do.

internal/repository/prune.go
```diff
@@ -326,7 +330,10 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 	repoVersion := repo.Config().Version
 	// only repack very small files by default
 	targetPackSize := repo.packSize() / 25
-	if opts.RepackSmall {
+	if opts.SmallPackBytes > 0 && opts.RepackSmall {
```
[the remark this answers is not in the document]
wplapper: done.

internal/repository/prune.go
```diff
@@ -386,7 +393,7 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 			stats.Packs.Keep++
 
 		case p.unusedBlobs == 0 && p.tpe != restic.InvalidBlob && !mustCompress:
-			if packSize >= int64(targetPackSize) {
+			if packSize >= int64(targetPackSize) && opts.SmallPackBytes == 0 && !opts.RepackSmall {
```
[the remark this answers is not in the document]
wplapper: removed faultt if-clause.

internal/repository/repository.go
```diff
@@ -148,6 +148,11 @@ func (r *Repository) packSize() uint {
 	return r.opts.PackSize
 }
 
+// SetPackSize - this is a hack!
+func (r *Repository) SetPackSize(size uint) {
```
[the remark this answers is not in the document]
wplapper: method removed.

internal/repository/prune_test.go
```diff
@@ -191,3 +192,92 @@ func TestPruneMaxUnusedDuplicate(t *testing.T) {
...
+	}
+
+	plan, err := repository.PlanPrune(context.TODO(), opts, repo, func(ctx context.Context, repo restic.Repository, usedBlobs restic.FindBlobSet) error {
+		for blob := range keep {
+			usedBlobs.Insert(blob)
+		}
+		return nil
+	}, &progress.NoopPrinter{})
+	rtest.OK(t, err)
+	_ = plan
```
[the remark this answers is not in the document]
wplapper: assignment is gone.

internal/repository/prune_test.go
```diff
@@ -191,3 +192,92 @@ func TestPruneMaxUnusedDuplicate(t *testing.T) {
...
+
+	rtest.OK(t, plan.Execute(context.TODO(), &progress.NoopPrinter{}))
+
+	stats := plan.Stats()
+	//t.Log(fmt.Sprintf("stats.Size  %+v", stats.Size))
+	//t.Log(fmt.Sprintf("stats.Blobs %+v", stats.Blobs))
+	//t.Log(fmt.Sprintf("stats.Packs %+v", stats.Packs))
+	rtest.Equals(t, stats.Size.Used <= stats.Size.Repack, true, "all sizes ...")
```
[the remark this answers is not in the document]
wplapper: still working on the tests.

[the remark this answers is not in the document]
wplapper: tests are improved, The checks should be more easily to understand.

MichaelEischer:
I've rebased the PR (and tried to fix the merge conflicts while rebasing) to remove the clutter from the "files changes" tab on Github.

wplapper:
Hi Michael,

Thank you for hard work on this PR. As I indicated earlier, I still don't
know much about git and github, and therefore make too many unforced
mistakes. I would appreciate a few hints, what to do, what to avoid.

Thanks again,
Winfried

MichaelEischer:
> Thank you for hard work on this PR. As I indicated earlier, I still don't know much about git and github, and therefore make too many unforced mistakes. I would appreciate a few hints, what to do, what to avoid.

No problem, it was basically "just" a matter of rebasing the PR. However, due to the merge conflicts that required a bit of experience with `git rebase`. The "files added" tab on Github doesn't seem to handle merge commits particularly well,  such that it also shows changes that already exist in the repository. The best way to avoid that is to not add merge commits in a PR in the first place, but instead to rebase it once there's a conflict with other changes in the repository. (having merge commits in a PR is allowed by Github, but frowned upon by many projects including restic)

`git rebase` allows modifying and cleaning up the commit history, which can be useful when the PR evolves over time. By removing unnecessary changes from the commit history, this also reduces the chance of having a merge commit that has to be addressed. Although that is probably already advanced git territory.

MichaelEischer:
Judging from the "rebase" commits in https://github.com/restic/restic/pull/5185 you're holding `git rebase` very wrong. I can recommend reading through the first few chapters of https://git-scm.com/book/en/v2 and trying out the examples along the way. `git rebase` is supposed to take the existing changes and apply them on top of a different (newer) state of the repository. Rebasing does not create additional commits  (I'm ignoring advanced git rebase usages here). Pushing those changes to Github also requires using `git push --force-with-lease`.

wplapper:
I understand now why my `git rebase` went so utterly wrong: I have a paper version of the ProGit book and it contains a critical omission in the `rebase` section and does not mention 
`$ git checkout master; $ git merge experiment` at all.

MichaelEischer:
Please stop pushing merge commits into your PRs.

wplapper:
I messed up git again, so I did a hard reset and activated all addition and changes.

wplapper:
Will be reworking the PR based on commit 8c12291f5

MichaelEischer (approved):
LGTM.

I've changed the `--repack-smaller-than` option to work independently of the `--repack-small` option. My goal is to eventually get rid of `--repack-small` so let's not tie those options together.

[Edit]See https://github.com/restic/restic/issues/5293 for more context on that[/Edit]

Issue #4433, opened 2023-08-04
Sort order for output of find


Roeshimi:
Output of `restic version`
--------------------------
restic 0.16.0 compiled with go1.20.6 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

### Actual behaviour
output of find is sorted from oldest to newest.

### Suggested behaviour
output should either be sorted newest to oldest or should accept a parameter like asc/desc

What are you trying to do? What problem would this solve?
---------------------------------------------------------
I don't know about others, but I mostly care about the latest backups of a file/folder so having find output the latest backups first would help me a lot. 

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

restic does make it very easy for me to have an offsite backup of my laptop to OneDrive and also a local one, which is just in case I don't have internet access but need a file restored.
Automating this with 1Password  (as credential store) and Powershell is a breeze :)

wplapper:
PR is merged. Issue can be closed.

PR #5184, opened 2024-12-16
Issue 4433: Ability to define sort order for output of find command

wplapper:
cmd_find: ability to define sort order for output of find command: Add option --reverse

Added ability to sort output of find command.

https://github.com/restic/restic/issues/4433
Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4433
+++ b/changelog/unreleased/issue-4433
@@ -1,5 +1,6 @@
-Enhancement: allow to define sorting order of command restic find.
+find: define sorting order of command find.
 Add option --reverse
 
 https://github.com/restic/restic/issues/4433
+https://github.com/restic/restic/pull/5184
 
--- a/changelog/unreleased/issue-4433
+++ b/changelog/unreleased/issue-4433
@@ -1,5 +1,8 @@
-find: define sorting order of command find.
-Add option --reverse
+Enhancement: find: enable sorting of find output, new option --reverse
+
+The old output behaviour was to sort snapshots from oldest to newest.
+The new sorting order is from newest to oldest. If one wants to revert to the
+old behaviour, use the option --reverse.
 
 https://github.com/restic/restic/issues/4433
 https://github.com/restic/restic/pull/5184
--- a/changelog/unreleased/issue-4433
+++ b/changelog/unreleased/issue-4433
@@ -1,0 +1,5 @@
+Enhancement: allow to define sorting order of command restic find.
+Add option --reverse
+
+https://github.com/restic/restic/issues/4433
+
--- a/changelog/unreleased/issue-4433
+++ b/changelog/unreleased/issue-4433
@@ -1,4 +1,4 @@
-Enhancement: find: enable sorting of find output, new option --reverse
+Enhancement: Sort `find` output from newest to oldest and add `--reverse` option
 
 The old output behaviour was to sort snapshots from oldest to newest.
 The new sorting order is from newest to oldest. If one wants to revert to the
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -23,9 +23,8 @@ var cmdFind = &cobra.Command{
 The "find" command searches for files or directories in snapshots stored in the
 repo.
 It can also be used to search for restic blobs or trees for troubleshooting.
-The default sort option for the snapshots is youngest to oldest. One can reverse
-the sorting by specifying --reverse, so the output of the snapshots is then shown
-as oldest to youngest.`,
+The default sort option for the snapshots is youngest to oldest. To sort the
+output from oldest to youngest specify --reverse.`,
 	Example: `restic find config.json
 restic find --json "*.yml" "*.json"
 restic find --json --blob 420f620f b46ebe8a ddd38656
@@ -644,9 +643,8 @@ func runFind(ctx context.Context, opts FindOptions, gopts GlobalOptions, args []
 	sort.Slice(filteredSnapshots, func(i, j int) bool {
 		if opts.Reverse {
 			return filteredSnapshots[i].Time.Before(filteredSnapshots[j].Time)
-		} else {
-			return filteredSnapshots[i].Time.After(filteredSnapshots[j].Time)
 		}
+		return filteredSnapshots[i].Time.After(filteredSnapshots[j].Time)
 	})
 
 	for _, sn := range filteredSnapshots {
--- a/cmd/restic/cmd_find.go
+++ b/cmd/restic/cmd_find.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"slices"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


changelog/unreleased/issue-4433
```diff
@@ -0,0 +1,6 @@
+find: define sorting order of command find.
```
MichaelEischer, line 1: Please follow the format from https://github.com/restic/restic/blob/master/changelog/TEMPLATE

cmd/restic/cmd_find.go
```diff
@@ -73,6 +75,7 @@ func init() {
 	f.BoolVar(&findOptions.PackID, "pack", false, "pattern is a pack-ID")
 	f.BoolVar(&findOptions.ShowPackID, "show-pack-id", false, "display the pack-ID the blobs belong to (with --blob or --tree)")
 	f.BoolVarP(&findOptions.CaseInsensitive, "ignore-case", "i", false, "ignore case for pattern")
+	f.BoolVarP(&findOptions.Reverse, "reverse", "R", false, "reverse sort order newest->oldest")
```
MichaelEischer, line 78: `"reverse sort order from newest to oldest"`

cmd/restic/cmd_find.go
```diff
@@ -639,6 +642,9 @@ func runFind(ctx context.Context, opts FindOptions, gopts GlobalOptions, args []
 	sort.Slice(filteredSnapshots, func(i, j int) bool {
 		return filteredSnapshots[i].Time.Before(filteredSnapshots[j].Time)
```
MichaelEischer, line 643: It feels more elegant to let the comparator here check whether `opts.Reverse` is set and directly use the correct order.

cmd/restic/cmd_find.go
MichaelEischer, line 78: I think we should actually change the default sort order to be from newest to oldest by default. And let the option invert it to the old behavior.

MichaelEischer (changes requested):
Thanks for taking a look at this. The change is simpler than I've initially expected :-) .
A small test case in `cmd/restic/cmd_find_integration_test.go` would also be nice.

As in https://github.com/restic/restic/pull/5183 please rebase the PR instead of adding merge commits.

cmd/restic/cmd_find.go
wplapper: done.

cmd/restic/cmd_find.go
wplapper: done.

cmd/restic/cmd_find.go
wplapper: Changed to "oldest to newest".

changelog/unreleased/issue-4433
wplapper: done. Also pointed out that the default sorting order has changed.

wplapper:
> As in #5183 please rebase the PR instead of adding merge commits.

@MichaelEischer: as I pointed out before, I don't know how to rebase. I am a really beginner when it comes to git. @rawtaz promised me a chat, but I have no idea how I can contact him.

wplapper:
Integration test added.

cmd/restic/cmd_find.go
```diff
@@ -22,7 +22,10 @@ var cmdFind = &cobra.Command{
 	Long: `
 The "find" command searches for files or directories in snapshots stored in the
 repo.
-It can also be used to search for restic blobs or trees for troubleshooting.`,
+It can also be used to search for restic blobs or trees for troubleshooting.
+The default sort option for the snapshots is youngest to oldest. One can reverse
```
MichaelEischer, line 26: Can be shortened to
```
The default sort option for the snapshots is youngest to oldest. To sort the
output from oldest to youngest specify --reverse.`,
```

changelog/unreleased/issue-4433
```diff
@@ -0,0 +1,9 @@
+Enhancement: find: enable sorting of find output, new option --reverse
```
MichaelEischer, line 1: I'd prefer something like the following:
```
Enhancement: Sort `find` output from newest to oldest and add `--reverse` option
```

cmd/restic/cmd_find_integration_test.go
```diff
@@ -91,3 +91,53 @@ func TestFindJSON(t *testing.T) {
 	rtest.Assert(t, len(matches[0].Matches) == 3, "expected 3 files to match (%v)", matches[0].Matches)
 	rtest.Assert(t, matches[0].Hits == 3, "expected hits to show 3 matches (%v)", datafile)
 }
+
```
MichaelEischer, line 94: Please run `gofmt` on the file. Ideally, set up your editor to automatically format the file when saving it. For example, in VS Code the Go extension even automatically enables auto formatting of files. Thankfully Go has a canonical formatting style so there's no more need to manually format and indent code.

cmd/restic/cmd_find_integration_test.go
```diff
@@ -91,3 +91,53 @@ func TestFindJSON(t *testing.T) {
...
+func TestFindSorting(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	datafile := testSetupBackupData(t, env)
+	opts := BackupOptions{}
+
+	// first backup
+	testRunBackup(t, "", []string{env.testdata}, opts, env.gopts)
+	testRunCheck(t, env.gopts)
```
MichaelEischer, line 116: Running `check` is redundant here. (The existing tests call `check` far often for my taste)

cmd/restic/cmd_find_integration_test.go
```diff
@@ -91,3 +91,53 @@ func TestFindJSON(t *testing.T) {
...
+
+	// second backup
+	testRunBackup(t, "", []string{env.testdata}, opts, env.gopts)
+
+	// first restic find
+	results := testRunFind(t, true, FindOptions{}, env.gopts, "testfile")
+	lines := strings.Split(string(results), "\n")
+	rtest.Assert(t, len(lines) == 2, "expected two files found in repo (%v), found %d", datafile, len(lines))
```
MichaelEischer, line 124: Please also check that find returns the results in the correct order. (The simplest way is probably to call `testListSnapshots(...)` after creating the first snapshot to get the first ID)

cmd/restic/cmd_find.go
```diff
@@ -637,7 +642,11 @@ func runFind(ctx context.Context, opts FindOptions, gopts GlobalOptions, args []
 	}
 
 	sort.Slice(filteredSnapshots, func(i, j int) bool {
-		return filteredSnapshots[i].Time.Before(filteredSnapshots[j].Time)
+		if opts.Reverse {
```
MichaelEischer, line 645: Our linter rules expect this to be written as follows:
```go
		if opts.Reverse {
			return filteredSnapshots[i].Time.Before(filteredSnapshots[j].Time)
		}
		return filteredSnapshots[i].Time.After(filteredSnapshots[j].Time)
```

You can manually call `golangci-lint run ./...` to get the linting results (or setup your editor accordingly)

cmd/restic/cmd_find_integration_test.go
```diff
@@ -91,3 +91,53 @@ func TestFindJSON(t *testing.T) {
 	rtest.Assert(t, len(matches[0].Matches) == 3, "expected 3 files to match (%v)", matches[0].Matches)
 	rtest.Assert(t, matches[0].Hits == 3, "expected hits to show 3 matches (%v)", datafile)
 }
+
+
+type testMatches2 struct {
```
MichaelEischer, line 96: Please just use the `testMatches` and `testMatch` that already exist.

MichaelEischer (changes requested):
Thanks for the test! I'll unfortunately have to bug you with a bit more tooling, this time gofmt (canonical go code formatting) and golangci-lint (checks for code issues), check the comments for more details.

Regarding rebasing, I can recommend the [git book](https://git-scm.com/book/en/v2/Git-Branching-Rebasing), although I can also rebase the PR in the end if necessary. (I can also make the changes mentioned in the review comments if that's fine for you)

MichaelEischer (approved):
LGTM. Thanks for all the tweaks. I've rebased the PR and added a small bugfix for the test case.

Issue #4278, opened 2023-04-01
"Include" switch for rewrite command


akrabu:
Output of `restic version`
--------------------------
restic 0.15.1 compiled with go1.19.5 on darwin/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
As with the "restore" command, it would be nice if "rewrite" also had an "include" switch so that you could include only one pattern while excluding the rest.

What are you trying to do? What problem would this solve?
---------------------------------------------------------
I'd like to "forget" a snapshot, but keep the *.ost and *.pst files just in case.  As it stands, I only see two ways of doing this.  (A) Restore the snapshot, including only *.ost and *.pst, then create a new snapshot by backing up that folder, or (B) Create a directory listing of everything on the snapshot, remove all the *.ost and *.pst files, create an exclude file from that, then rewrite the snapshot using the exclude file.

Which, of course, would likely work...  but an "--include" switch would be way easier :)

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Hands down the best open source backup project out there 🙌

wplapper:
I have written a very rough modification to cmd_rewrite.go, which allows exclusively either inclusion or exclusion of file patterns. It has the disadvantage that currently all the directory information of the existing backup has to be included in the rewritten snapshot, so that the included files can be found in the tree. As a result there will possibly be a lot of empty directories; however these blobs do not not take up so much space. I have no idea as yet how to solve the dilemma of empty subdirectories.

wplapper:
I have solved the issue of empty subdirectories by doing a walker.Walk on the snapshot first, if a new flag "exclude-empty" is given. This allows for filtering out directories which are needed for included files / directories, so the rest of the directory structure does not need to be included. Will issue PR soon.

wplapper:
I have solved the issue of empty subdirectories: I do a pass of walker.Walk() to register the subdirectories which are needed for the inclusion pattern. By finding all the parents of the directories, the those directories can be identified and then be used in the actual rewrite process.
Will deliver PR soon.

wplapper:
I have messed up my fork on github, so i closed and reopened the pull request. I will wait for the review of https://github.com/restic/restic/pull/5185 to be finished, so I can integrate the work from #5185.

PR #5185, opened 2024-12-18
Issue: 4942: cmd_rewrite: add snapshot summary data to an existing snapshot.

wplapper:
cmd_rewrite: add capability to add snapshot summary data to an existing snapshot.

Add option --snapshot-summary [-s] to enable this enhancement.

What does this PR change? What problem does it solve?
-----------------------------------------------------
This pull request add the capability of adding snapshot summary data to an existing snapshot.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
https://github.com/restic/restic/issues/4942
closes https://github.com/restic/restic/issues/4942

Checklist
---------

- [x] I have added tests for all code changes.
- [x] I have added documentation for relevant changes (in the manual).
- [X] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I'm done! This pull request is ready for review.

Issue #4585, opened 2023-12-06
Add ability to limit max repository size.


slynn1324:
Output of `restic version`
--------------------------

restic 0.16.2 compiled with go1.21.3 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Add the ability to limit the maximum repository size.  

What are you trying to do? What problem would this solve?
---------------------------------------------------------

Some cloud storage providers (Wasabi) don't seem to provide a way to limit the size of a bucket or a spending limit on the account.  Being able to set a maximum-size threshold on the backup repository will help prevent a possible expensive accidental overrun if it begins backing up more than expected due to some unexpected condition such as passing the wrong directory arguments or moving more files into a directory than possibly intended for the backup set.  

I currently can approximate this in a wrapper script by first running `restic stats --json mode raw-data`, followed by `restic backup --quiet --dry-run --json` and adding up the 'total_size' from the stats call and 'data_added' from the backup dry run to get a new projected total size -- and checking that against a configured threshold.  

It would be nice (and I imagine possibly more efficient?) to be able to pass a "--max-repo-size 50G" argument to restic directly. 

If I'm missing some other way of accomplishing this that exists -- my apologies and please advise.  Thanks!! 

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Restic is awesome!  I love the ways that it lets me manage backups in an efficient way across multiple different back ends, and is supported across all operating systems!  Great work!

dia3olik:
Yeah! Any news on this?

MichaelEischer:
There is one key limitation for this feature. It won't be possible to accurately limit the repository if multiple clients backup to the same repository at the same time. There will also be a significant slowdown on startup as restic would have to list all files in the backend to know how large the repository currently is.

wplapper:
A quick estimate of the repository size is the sum of sizes of all packfiles, excluding space needed for indices and snapshots etc.:
 repo_packs, _ := pack.Size(ctx, repo, false)
  pack_size := int64(0)
  for _, size := range repo_packs {
    pack_size += size
  }
This is an index function and will not go back to the backend.

IMHO one doesn't need an exact limit, just a rough limit. If multiple clients are backing up at the same time, the last one hitting the limit check will have to bite the bullet (and the backup will fail).

MichaelEischer:
> This is an index function and will not go back to the backend.

[...]

PR #5188, opened 2024-12-19
PR dead: issue 4278: Enhancement: include filters in command rewrite - see newer PR #5191 

wplapper:
Enhancement: include filters in command rewrite

https://github.com/restic/restic/issues/4278 

The enhancement allows the standard include filter options. 

In order to make the include filter work efficiently, an additional read pass through the directory tree is needed to identify the sub-directories and their parents for the inclusion of files. Otherwise the full directory tree needs to be included which as a consequence may contain quite a lot of empty sub-directories. The additional read pass avoids this issue, but it might take a bit more time, depending on the network speed of the back-end storage and the size of the snapshot.

File internal/filter/include.go gained an "Empty() function, and file internal/walker/rewriter.go gained two additional fields 
TreeBlobs and DataBlobs. 

File cmd/restic/cmd_rewrite.go gained the additional logic of handling the include filters and the very first pass through
the snapshot to detect the necessary directory structure to support the inclusion of files. 

https://github.com/restic/restic/issues/4278

Checklist
---------


MichaelEischer (changes requested):
Thanks for working on include filters! I've only partially reviewed the code so far as I ran into a major conceptual problem with the implementation of the `--exclude-empty` option. `directoriesNeeded` is a rather complex and slow workaround for missing functionality in `TreeRewriter.RewriteTree`.

The problem that `directoriesNeeded` tries to work around is that directories that are not included still remain in the snapshot but are empty. The clean solution would be to extend the `TreeRewriter` options with a new `KeepEmtpyDirectory func(path string) bool` field. If it is unset, then the rewriter behaves like before. If the field is set and the filtered tree is empty then `KeepEmtpyDirectory` decides whether to keep the tree or not. If the tree should not be kept, then `RewriteTree` returns an empty restic.ID which also means that the corresponding node should be dropped.

That way the filtering of empty directories happens directly within the `TreeRewriter` without _any_ additional overhead. This also obsoletes the `--exclude-empty` option.

[Edit]Some of the comments below are obsolete when replacing `directoriesNeeded`. I've nevertheless kept them as they are still relevant from a code style perspective[/Edit]

wplapper:
Will work on https://github.com/restic/restic/pull/5185 first. Once this PR is on its way, I will tackle this PR. Thanks for reviewing.

wplapper:
@MichaelEischer: Thank you for commenting on the issue of empty directories. You wrote 
"The problem that directoriesNeeded tries to work around is that directories that are not included still remain in the snapshot but are empty. The clean solution would be to extend the TreeRewriter options with a new KeepEmtpyDirectory func(path string) bool field. If it is unset, then the rewriter behaves like before. If the field is set and the filtered tree is empty then KeepEmtpyDirectory decides whether to keep the tree or not. If the tree should not be kept, then RewriteTree returns an empty restic.ID which also means that the corresponding node should be dropped.
". I have a problem understanding your comment. Since RewriteTree is a strictly top down process and one needs to know the parent line of a particular sub directory structure (parent, grandparent, ... up to the top of the tree), one has to decide quite early in the tree walk whether or not to keep/drop this subtree. 

I will give an example: My snapshot I want to rewrite contain some Python files (*.py) which I am interested in. I am looking for all *.py files inside a certain subdirectory "peterNorvig", and below. So the chosen filter is '-i "peterNorvig/**/*.py"'.
The partial snapshot listing of the filtered output is:
/home
/home/wplapper
/home/wplapper/sudoku
/home/wplapper/sudoku/peterNorvig
/home/wplapper/sudoku/peterNorvig/PN-solve_LSA_puzzle.py
...
/home/wplapper/sudoku/peterNorvig/PN_cython
/home/wplapper/sudoku/peterNorvig/PN_cython/PN_sudoku_setup.py'
...

TreeRewriter, has to include tree nodes /, /home, /home/wlapper, /home/wlapper/sudoku and /home/wlapper/sudoku/PN_cython/ before it can include the wanted *.py files. My question is: how does TreeRewriter knows this in advance, unless this has been found out before? Please correct me if my thoughts are wrong.
PS.: Only the wanted sub directories need to be stored in 'directoriesNeeded', not all of them. This can be amended easily.

wplapper:
I have cracked the issue of including empty sub directories when using include filters: when a sub directory is empty, the resulting byte string from calling tb.Finalize() is `{"nodes":[]}` + "\n". Comparing with this constant 

> suppresses the creation of empty subtrees
> removes them in the parent directory of being listed

rawtaz:
@wplapper I think you accidentally closed the PR =D

wplapper:
Yes I know, but I have created a new one with the "old" contents.

rawtaz:
There is no need to close one PR and open a second PR about the same thing and even with the same title :) Let's have a chat about that at some point, if you want help with the Git/GitHub stuff here!

wplapper:
Yes, I am pretty blind when it comes to Git / Github. I have made too many
mistakes with git and I do not know how to revert them, particularly when
the commits have been pushed to github.

Merry Christmas and thanks for the generous offer,

Winfried

wplapper:
@rawtaz: how can I get in contact with you? I am working on #4585 and I want to cleanly shutdown the backup process. Thanks Winfried

PR #5190, opened 2024-12-20
fs: error if a symlink points at a file that is not included in the snapshot

jfly:
Note: I've marked this as a draft because I want to get feedback on the approach before I invest in polishing off the implementation and adding tests.

This implements @fd0's first idea here: <https://github.com/restic/restic/issues/542#issuecomment-328263959>:

> First, I think it may be a good idea to print a warning message when a symlinks is saved and the target of the symlink exists and is not included in the backup. This way, users will know that some data referenced in the snapshot is not available upon restore.

Which I wholeheartedly agree with.

In the interest of keeping restic's cli simple, and keeping people's data safe, I've opted to not make this configurable. I suppose you could call this a breaking change, but I personally consider it a fix: restic shouldn't claim it has successfully backed up a directory unless it can actually recreate the structure of that directory. IMO, it's better to fail-fast than to claim success, only to greatly disappoint someone later on.

Checklist
---------

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"sort"
 	"strings"
@@ -97,11 +98,12 @@ type Archiver struct {
 	FS           fs.FS
 	Options      Options
 
-	blobSaver *blobSaver
-	fileSaver *fileSaver
-	treeSaver *treeSaver
-	mu        sync.Mutex
-	summary   *Summary
+	blobSaver       *blobSaver
+	fileSaver       *fileSaver
+	treeSaver       *treeSaver
+	mu              sync.Mutex
+	summary         *Summary
+	snapshotTargets *[]string
 
 	// Error is called for all errors that occur during backup.
 	Error ErrorFunc
@@ -596,6 +598,27 @@ func (arch *Archiver) save(ctx context.Context, snPath, target string, previous 
 		debug.Log("  %v is a socket, ignoring", target)
 		return futureNode{}, true, nil
 
+	case fi.Mode&os.ModeSymlink > 0:
+		debug.Log("  %v symlink", target)
+
+		inSnapshot, err := arch.isSymlinkInSnapshot(target)
+		if err != nil {
+			return futureNode{}, false, err
+		}
+		if !inSnapshot {
+			return futureNode{}, false, errors.Errorf("encountered a symlink pointing to a file that is not included in the snapshot: %s", target)
+		}
+
+		node, err := arch.nodeFromFileInfo(snPath, target, meta, false)
+		if err != nil {
+			return futureNode{}, false, err
+		}
+		fn = newFutureNodeWithResult(futureNodeResult{
+			snPath: snPath,
+			target: target,
+			node:   node,
+		})
+
 	default:
 		debug.Log("  %v other", target)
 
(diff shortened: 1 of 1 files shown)


internal/archiver/archiver.go
```diff
@@ -615,6 +638,72 @@ func (arch *Archiver) save(ctx context.Context, snPath, target string, previous
...
+		relativePath, err := filepath.Rel(snapshotTarget, file)
+		if err != nil {
+			return false, fmt.Errorf("error computing relative path from %s to %s: %s", snapshotTarget, file, err)
+		}
+
+		if strings.HasPrefix(relativePath, "..") {
+			continue
+		}
+
+		// <<< TODO: look at `arch.SelectByName` and `arch.Select` >>>
```
jfly, line 700: The basic idea of this PR is that every time we hit a symlink, check if it is (or will be) included in the snapshot. I'm not in love with this approach because it means that I'm effectively re-implementing the filtering logic in `arch.save`. If we add new types of filters in the future, it's possible the implementations could drift.

An alternative implementation that does not have this brittleness would be to instead record a "seen list" of all symlinks we encountered during the backup. Then, at the very end, confirm that each of those symlinks is actually in the snapshot. One downside of this is that there's no guaranteed bound on how large the "seen list" gets. I'm not familiar enough with restic's codebase to know if it's straightforward to walk through the snapshot, or if that's a no-no because it may be on some remote server and is expensive (in time or $) to walk through it.

A third implementation would be to wait until the very end to do the check. Walk through the whole snapshot, and for each symlink, verify that it's in the snapshot. This would mean a second pass through the whole snapshot. I have the same question here about if it's OK to walk through the snapshot on the end, or if it's not OK to do because it may be remote.

After typing this up, I think I prefer the "third implementation" the best. Would really appreciate advice from a restic maintainer, though!

jfly:
@fd0, could I get some feedback on this PR? I'm happy to put time into this to push this over the finish line if this is something restic would be willing to merge.

plumbeo:
This change would add a ton of errors to every single one of my backups, and I don't think I'm the only one for whom that would be true. At a minimum, if this actually goes in, it should be made configurable and off by default.

Personally I don't like it, symlinks are a well-known feature of many filesystems and a backup tool's job is to make sure that they're restored as they are, not to figure out why they are there or assume that the user wants to backup whatever they point to too.

jfly:
> This change would add a ton of errors to every single one of my backups, and I don't think I'm the only one for whom that would be true. At a minimum, if this actually goes in, it should be made configurable and off by default.

That's totally fair. If I made this an option, and off by default, would restic be interested in merging?

I still need help from someone more familiar with restic on [how to architect this feature](https://github.com/restic/restic/pull/5190#discussion_r1893803289).

MichaelEischer (commented):
> Personally I don't like it, symlinks are a well-known feature of many filesystems and a backup tool's job is to make sure that they're restored as they are, not to figure out why they are there or assume that the user wants to backup whatever they point to too.

I agree with that sentiment. In particular, it is trivial to create symlinks to non-existing files. Having restic warn about  every such symlink feels like a problem to me. Sooner or later someone will also likely ask for a way to filter those warnings. And I'd really prefer to not complicate the codebase with that.

I haven't come to any conclusion yet, I have to read the original discussion issue first.

jfly:
> I haven't come to any conclusion yet, I have to read the original discussion issue first.

@MichaelEischer, have you had a chance to think about this more? Happy to address your PR feedback, but wanted to wait to see if this actually is welcome in restic before I do so.

PR #5191, opened 2024-12-23
Enhancement: `restic rewrite`: allow include filters in command

### LINKED ISSUE #4278: "Include" switch for rewrite command

wplapper:
This enhancement enables the use of include filters for the rewrite command. Instead of using complex logic to exclude everything else bar the files and directories which the user is interested in, it can now be defined as include "all the bits I am interested in".

File internal/filter/include.go gained an "Empty() function, and file internal/walker/rewriter.go gained logic to recognize empty sub-trees and suppresses the creation for include filtering.

cmd_rewite adds code for inclusion filters. Only sub-trees which are needed will be created.

closes https://github.com/restic/restic/issues/4278

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_repair_snapshots.go
+++ b/cmd/restic/cmd_repair_snapshots.go
@@ -151,7 +151,7 @@ func runRepairSnapshots(ctx context.Context, gopts global.Options, opts RepairOp
 			func(ctx context.Context, sn *data.Snapshot, uploader restic.BlobSaver) (restic.ID, *data.SnapshotSummary, error) {
 				id, err := rewriter.RewriteTree(ctx, repo, uploader, "/", *sn.Tree)
 				return id, nil, err
-			}, opts.DryRun, opts.Forget, nil, "repaired", printer)
+			}, opts.DryRun, opts.Forget, nil, "repaired", printer, false)
 		if err != nil {
 			return errors.Fatalf("unable to rewrite snapshot ID %q: %v", sn.ID().Str(), err)
 		}
--- a/cmd/restic/cmd_rewrite.go
+++ b/cmd/restic/cmd_rewrite.go
@@ -30,6 +30,9 @@ The "rewrite" command excludes files from existing snapshots. It creates new
 snapshots containing the same data as the original ones, but without the files
 you specify to exclude. All metadata (time, host, tags) will be preserved.
 
+Alternatively you can use one of the --include variants to only include files
+in the new snapshot which you want to preserve.
+
 The snapshots to rewrite are specified using the --host, --tag and --path options,
 or by providing a list of snapshot IDs. Please note that specifying neither any of
 these options nor a snapshot ID will cause the command to rewrite all snapshots.
@@ -46,8 +49,8 @@ When rewrite is used with the --snapshot-summary option, a new snapshot is
 created containing statistics summary data. Only two fields in the summary will
 be non-zero: TotalFilesProcessed and TotalBytesProcessed.
 
-When rewrite is called with one of the --exclude options, TotalFilesProcessed
-and TotalBytesProcessed will be updated in the snapshot summary.
+When rewrite is called with one of the --exclude or --include options,
+TotalFilesProcessed and TotalBytesProcessed will be updated in the snapshot summary.
 
 EXIT STATUS
 ===========
@@ -109,6 +112,7 @@ type RewriteOptions struct {
 	Metadata snapshotMetadataArgs
 	data.SnapshotFilter
 	filter.ExcludePatternOptions
+	filter.IncludePatternOptions
 }
 
 func (opts *RewriteOptions) AddFlags(f *pflag.FlagSet) {
@@ -120,6 +124,7 @@ func (opts *RewriteOptions) AddFlags(f *pflag.FlagSet) {
 
 	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
 	opts.ExcludePatternOptions.Add(f)
+	opts.IncludePatternOptions.Add(f)
 }
 
 // rewriteFilterFunc returns the filtered tree ID or an error. If a snapshot summary is returned, the snapshot will
@@ -136,33 +141,32 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *data.
 		return false, err
 	}
 
+	includeByNameFuncs, err := opts.IncludePatternOptions.CollectPatterns(printer.E)
+	if err != nil {
+		return false, err
+	}
+
 	metadata, err := opts.Metadata.convert()
 
 	if err != nil {
 		return false, err
 	}
 
+	condInclude := len(includeByNameFuncs) > 0
+	condExclude := len(rejectByNameFuncs) > 0 || opts.SnapshotSummary
 	var filter rewriteFilterFunc
+	var rewriteNode walker.NodeRewriteFunc
+	var keepEmptyDirectoryFunc walker.NodeKeepEmptyDirectoryFunc
 
-	if len(rejectByNameFuncs) > 0 || opts.SnapshotSummary {
-		selectByName := func(nodepath string) bool {
-			for _, reject := range rejectByNameFuncs {
-				if reject(nodepath) {
-					return false
-				}
-			}
-			return true
+	if condInclude || condExclude {
+		if condInclude {
+			rewriteNode, keepEmptyDirectoryFunc = gatherIncludeFilters(includeByNameFuncs, printer)
+		} else {
+			rewriteNode = gatherExcludeFilters(rejectByNameFuncs, printer)
+			keepEmptyDirectoryFunc = nil
 		}
 
-		rewriteNode := func(node *data.Node, path string) *data.Node {
-			if selectByName(path) {
-				return node
-			}
-			printer.P("excluding %s", path)
-			return nil
-		}
-
-		rewriter, querySize := walker.NewSnapshotSizeRewriter(rewriteNode)
+		rewriter, querySize := walker.NewSnapshotSizeRewriter(rewriteNode, keepEmptyDirectoryFunc)
 
 		filter = func(ctx context.Context, sn *data.Snapshot, uploader restic.BlobSaver) (restic.ID, *data.SnapshotSummary, error) {
 			id, err := rewriter.RewriteTree(ctx, repo, uploader, "/", *sn.Tree)
@@ -186,11 +190,12 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *data.
 	}
 
 	return filterAndReplaceSnapshot(ctx, repo, sn,
-		filter, opts.DryRun, opts.Forget, metadata, "rewrite", printer)
+		filter, opts.DryRun, opts.Forget, metadata, "rewrite", printer, len(includeByNameFuncs) > 0)
 }
 
 func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *data.Snapshot,
-	filter rewriteFilterFunc, dryRun bool, forget bool, newMetadata *snapshotMetadata, addTag string, printer progress.Printer) (bool, error) {
+	filter rewriteFilterFunc, dryRun bool, forget bool, newMetadata *snapshotMetadata, addTag string, printer progress.Printer,
+	keepEmptySnapshot bool) (bool, error) {
 
 	var filteredTree restic.ID
 	var summary *data.SnapshotSummary
@@ -204,6 +209,10 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *d
 	}
 
 	if filteredTree.IsNull() {
+		if keepEmptySnapshot {
+			debug.Log("Snapshot %v not modified", sn)
+			return false, nil
+		}
 		if dryRun {
 			printer.P("would delete empty snapshot")
 		} else {
@@ -284,8 +293,12 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *d
 }
 
 func runRewrite(ctx context.Context, opts RewriteOptions, gopts global.Options, args []string, term ui.Terminal) error {
-	if !opts.SnapshotSummary && opts.ExcludePatternOptions.Empty() && opts.Metadata.empty() {
-		return errors.Fatal("Nothing to do: no excludes provided and no new metadata provided")
+	hasExcludes := !opts.ExcludePatternOptions.Empty()
+	hasIncludes := !opts.IncludePatternOptions.Empty()
+	if !opts.SnapshotSummary && !hasExcludes && !hasIncludes && opts.Metadata.empty() {
+		return errors.Fatal("Nothing to do: no excludes/includes provided and no new metadata provided")
+	} else if hasExcludes && hasIncludes {
+		return errors.Fatal("exclude and include patterns are mutually exclusive")
 	}
 
 	printer := ui.NewProgressPrinter(false, gopts.Verbosity, term)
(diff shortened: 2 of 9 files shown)


wplapper:
I have managed to crack the issue of empty subtrees. This change is incorporated in internal/walker/rewriter.go

.github/workflows/tests.yml
```diff
@@ -259,7 +259,7 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.61.0
+          version: v1.63.4
```
MichaelEischer, line 262: Please rebase this PR instead of adding merge commits

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -122,19 +122,19 @@ func runRepairSnapshots(ctx context.Context, gopts GlobalOptions, opts RepairOpt
 			node.Size = newSize
 			return node
 		},
-		RewriteFailedTree: func(_ restic.ID, path string, _ error) (restic.ID, error) {
+		RewriteFailedTree: func(_ restic.ID, path string, _ error) (restic.ID, *restic.SnapshotSummary, error) {
```
MichaelEischer, line 125: How did that change from https://github.com/restic/restic/pull/5185 end up here?

cmd/restic/cmd_rewrite.go
```diff
@@ -118,50 +133,97 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
...
+			selectByName := func(nodepath string, node *restic.Node) bool {
+				for _, include := range includeByNameFuncs {
+					if node.Type == restic.NodeTypeDir {
+						// always include directories
+						return true
+					} else if node.Type == restic.NodeTypeFile {
+						ifun, childMayMatch := include(nodepath)
+						return ifun && childMayMatch
+					} else if node.Type == restic.NodeTypeSymlink {
+						includeLink, _ := followSymlink(nodepath, node, include)
```
MichaelEischer, line 180: Symlinks should be treated like normal files in regard to include patterns. Everything else is unusable.

cmd/restic/cmd_rewrite.go
```diff
@@ -118,50 +133,97 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
...
+						return includeLink
+					} else {
+						//exclude all device files, fifo, socket, ...
+						return false
```
MichaelEischer, line 184: Nope. Those must be handled like normal files. Please follow the behavior of the backup command as closely as possible. Everything else will confuse users.

cmd/restic/cmd_rewrite.go
```diff
@@ -118,50 +133,97 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *resti
...
 				return node
 			}
-			Verbosef("excluding %s\n", path)
-			return nil
 		}
 
-		rewriter, querySize := walker.NewSnapshotSizeRewriter(rewriteNode)
+		//common
+		keepDirectories := len(includeByNameFuncs) == 0
+		rewriter, querySize = walker.NewSnapshotSizeRewriter(rewriteNode, keepDirectories)
```
MichaelEischer, line 212: If the include patterns match an empty directory, then that directory must be kept. This cannot be implemented as a boolean flag passed to the rewriter, but must be checked for each individual node.

To quote myself from https://github.com/restic/restic/pull/5188#pullrequestreview-2515978653 :

> The clean solution would be to extend the `TreeRewriter` options with a new `KeepEmtpyDirectory func(path string) bool` field. If it is unset, then the rewriter behaves like before. If the field is set and the filtered tree is empty then `KeepEmtpyDirectory` decides whether to keep the tree or not. If the tree should not be kept, then `RewriteTree` returns an empty restic.ID which also means that the corresponding node should be dropped.

Effectively the only two options here are adding the flag as return value to `rewriteNode` or passing a `KeepEmtpyDirectory` function to the rewriter.

internal/walker/rewriter.go
```diff
@@ -86,11 +88,13 @@ type BlobLoadSaver interface {
 	restic.BlobLoader
 }
 
-func (t *TreeRewriter) RewriteTree(ctx context.Context, repo BlobLoadSaver, nodepath string, nodeID restic.ID) (newNodeID restic.ID, err error) {
+const EmptyNodeID = `{"nodes":[]}` + "\n"
```
MichaelEischer, line 91: Drop that constant you won't need it for a clean implementation.

MichaelEischer (changes requested):
A few preliminary comments before you try to further work on this PR

wplapper:
I put this PR on hold until PR https://github.com/restic/restic/pull/5185 is fully resolved.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	// restic rewrite latest -i nothing-whatsoever --forget
+	rtest.OK(t, runRewrite(context.TODO(), RewriteOptions{
+		Forget:                true,
+		IncludePatternOptions: filter.IncludePatternOptions{Includes: []string{"nothing-whatsoever"}}},
+		env.gopts,
+		[]string{"latest"}))
+	testListSnapshots(t, env.gopts, 1)
+}
+
+func TestRewriteExcludeNothing(t *testing.T) {
```
MichaelEischer, line 290: I'm not sure what this test case actually tests. Even `--exclude *` would not remove a snapshot as `KeepEmptyDirectories` is always true for excludes. Or is that what you wanted to check? Then the exclude pattern must be `*`.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	createBasicRewriteRepo(t, env)
+
+	err := runRewrite(context.TODO(), RewriteOptions{}, env.gopts, []string{"latest"})
+	rtest.Assert(t, err != nil, "Nothing to do: no includes/excludes provided and no new metadata provided")
+
+	err = runRewrite(context.TODO(), RewriteOptions{
+		IncludePatternOptions: filter.IncludePatternOptions{Includes: []string{"JohannSebastianBach"}},
+		ExcludePatternOptions: filter.ExcludePatternOptions{Excludes: []string{"WolfgangAmadeusMozart"}},
+	}, env.gopts, []string{"latest"})
+	rtest.Assert(t, err != nil, "You cannot specify include and exclude options simultaneously!")
```
MichaelEischer, line 266: I'd prefer `"exclude and include patterns are mutually exclusive"` similar to the restore command as error message.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+
+	err = runRewrite(context.TODO(), RewriteOptions{
```
MichaelEischer, line 268: This should work IMHO.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	// read restic ls output and count lines which contain the string "empty-directory"
+	out := testRunLsWithOpts(t, env.gopts, LsOptions{}, []string{newSnapshots[0].String()})
+	fileList := strings.Split(string(out), "\n")
+	count := 0
+	for _, filename := range fileList {
+		if strings.Contains(filename, "empty-directory") {
+			count++
+		}
+	}
+	rtest.Assert(t, count == 1, "there should be 1 empty directory in the snapshot, but there are %d snapshots", count)
```
MichaelEischer, line 239:
```suggestion
	rtest.Assert(t, count == 1, "there should be 1 empty directory in the snapshot, but there are %d directories", count)
```

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	rtest.OK(t, runRewrite(context.TODO(), RewriteOptions{
+		Forget:                true,
+		IncludePatternOptions: filter.IncludePatternOptions{Includes: []string{"empty-directory"}}},
+		env.gopts,
+		[]string{snapshots[0].Str()}))
+	newSnapshots := testListSnapshots(t, env.gopts, 1)
+	rtest.Assert(t, snapshots[0] != newSnapshots[0], "snapshot id should have changed")
+
+	// read restic ls output and count lines which contain the string "empty-directory"
+	out := testRunLsWithOpts(t, env.gopts, LsOptions{}, []string{newSnapshots[0].String()})
```
MichaelEischer, line 231: Nice approach

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	rtest.Assert(t, count == 1, "there should be 1 empty directory in the snapshot, but there are %d snapshots", count)
+
+	// get snapshot summary and find this 1 directory
+	_, repo, unlock, err := openWithExclusiveLock(context.TODO(), env.gopts, false)
+	rtest.OK(t, err)
+
+	sn, err := restic.LoadSnapshot(context.TODO(), repo, newSnapshots[0])
+	rtest.OK(t, err)
+
+	rtest.Assert(t, sn.Summary != nil, "snapshot should have a summary attached")
```
MichaelEischer, line 248: Please additionally use `testRunLsWithOpts` to check that the directory has actually be removed.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	for _, filename := range fileList {
+		if strings.Contains(filename, "empty-directory") {
+			count++
+		}
+	}
+	rtest.Assert(t, count == 1, "there should be 1 empty directory in the snapshot, but there are %d snapshots", count)
+
+	// get snapshot summary and find this 1 directory
+	_, repo, unlock, err := openWithExclusiveLock(context.TODO(), env.gopts, false)
+	rtest.OK(t, err)
```
MichaelEischer, line 243:
```suggestion
	rtest.OK(t, err)
	defer unlock()
```

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+
+	sn, err := restic.LoadSnapshot(context.TODO(), repo, newSnapshots[0])
+	rtest.OK(t, err)
+
+	rtest.Assert(t, sn.Summary != nil, "snapshot should have a summary attached")
+	rtest.Assert(t, sn.Summary.TotalFilesProcessed == 0,
+		"there should be 0 files in the snapshot, but there are %d files", sn.Summary.TotalFilesProcessed)
+	unlock()
```
MichaelEischer, line 251:
```suggestion
```

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	// read restic ls output and count .txt files
+	out := testRunLsWithOpts(t, env.gopts, LsOptions{}, []string{newSnapshots[0].Str()})
+	fileList := strings.Split(string(out), "\n")
+	count := 0
+	for _, filename := range fileList {
+		if strings.Contains(filename, ".txt") {
+			count++
+		}
+	}
+	rtest.Assert(t, count == 2, "there should be 2 txt files in the snapshot, but there are %d files", count)
```
MichaelEischer, line 200: Please extract this into a helper function. And also use it for `TestRewriteEmptyDirectory`.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -170,5 +171,133 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+
+	// get snapshot summary and find these 2 files
+	_, repo, unlock, err := openWithExclusiveLock(context.TODO(), env.gopts, false)
```
MichaelEischer, line 203: Same here

MichaelEischer (changes requested):
Thanks for the changes so far! I have added quite a few comments, but those are mostly small changes and nitpicks. The general approach now looks like what I had in mind.

wplapper:
I manually moved all the changed code to the new restic level 0.18.1. Otherwise all requested changes should still be incorporated, The only bit which is somewhat different is ``cmd_rewrite_integration_test.go``.

wplapper:
Rebased changes onto current development tree.

wplapper:
Rebased changes onto current development tree.

cmd/restic/cmd_rewrite.go
```diff
@@ -348,3 +361,71 @@ func runRewrite(ctx context.Context, opts RewriteOptions, gopts global.Options,
...
+
+func gatherIncludeFilters(includeByNameFuncs []filter.IncludeByNameFunc, printer progress.Printer) (rewriteNode walker.NodeRewriteFunc, keepEmptyDirectory walker.NodeKeepEmptyDirectoryFunc) {
+	inSelectByName := func(nodepath string, node *data.Node) bool {
+		for _, include := range includeByNameFuncs {
+			if node.Type == data.NodeTypeDir {
+				// always include directories
+				return true
+			}
+			matched, childMayMatch := include(nodepath)
+			if matched && childMayMatch {
```
MichaelEischer: TODO

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -195,3 +218,169 @@ func TestRewriteSnaphotSummary(t *testing.T) {
...
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+	createBasicRewriteRepo(t, env)
+	snapsBefore := testListSnapshots(t, env.gopts, 1)
+
+	// restic rewrite latest -e 'nothing-whatsoever' --forget
+	err := testRunRewriteWithOpts(t,
+		RewriteOptions{
+			Forget:                true,
+			ExcludePatternOptions: filter.ExcludePatternOptions{Excludes: []string{"nothing-whatsoever"}},
```
MichaelEischer: This test case doesn't test anything at the moment. It would have to test using pattern `*`. However, then the test fails as it is currently perfectly valid to just exclude all files in a snapshot. For simplicity I think we should just keep it at that and drop this testcase.

MichaelEischer (approved):
LGTM. I've added a few commits that include a bunch of smaller cleanups, add tests for `KeepEmptyDirectory` and change the node filtering to only include directories that have a chance that some file in a subdirectory matches. The latter can speed up rewrite quite a lot, when for example matching against `/data` or some other absolute path.

wplapper:
Thanks. It was a long haul.

PR #4581, opened 2023-12-05
Added json logging for check.

perjahn:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Adding json logging to the check command, in a similar fashion how the init command outputs json.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

No

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added tests for all code changes.
- [x] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [ ] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

PR #5194, opened 2024-12-30
Json check

### LINKED ISSUE #1378: JSON Output for restic check command
## Output of `restic version`

restic 0.7.3

restic -r rest:path check --json

rest-server

## Expected behavior

A JSON array containing information about the check results, at least something like "errors":0 oder "errors":1 would be sufficient for me

## Actual behavior

no json output given

Just use restic with check command and json flag

darkdragon-001:

Add JSON output to the `check` command.

Fixes #1378
Closes #4581

Checklist
---------


darkdragon-001:
I am not sure about the last commit: Do we want to assemble such an extensive summary or are we fine with just wrapping the errors into JSON?

cmd/restic/cmd_check.go
```diff
@@ -479,3 +506,15 @@ func selectRandomPacksByFileSize(allPacks map[restic.ID]int64, subsetSize int64,
...
 	return packs
 }
+
+type checkResult struct {
+	Error           bool     `json:"error"`
+	NumIndexErrors  int      `json:"num_index_errors"`
+	NumPackErrors   int      `json:"num_pack_errors"`
+	NumTreeErrors   int      `json:"num_tree_errors"`
+	NumUnusedBlobs  int      `json:"num_unused_blobs"`
+	NumDataErrors   int      `json:"num_data_errors"`
```
darkdragon-001, line 516: Should we merge these?
```suggestion
	NumErrors       int     `json:"num_errors"`
```

cmd/restic/cmd_check.go
```diff
@@ -479,3 +506,15 @@ func selectRandomPacksByFileSize(allPacks map[restic.ID]int64, subsetSize int64,
...
+	BrokenPacks     []string `json:"broken_packs"`         // run "restic repair packs ID..." and "restic repair snapshots --forget" to remove damaged files
+	HintRepairIndex bool     `json:"suggest_repair_index"` // run "restic repair index"
+	HintPrune       bool     `json:"suggest_prune"`        // run "restic prune"
```
darkdragon-001, line 519: Do we want these hints?

cmd/restic/cmd_check.go
```diff
@@ -479,3 +506,15 @@ func selectRandomPacksByFileSize(allPacks map[restic.ID]int64, subsetSize int64,
...
 }
+
+type checkResult struct {
+	Error           bool     `json:"error"`
+	NumIndexErrors  int      `json:"num_index_errors"`
+	NumPackErrors   int      `json:"num_pack_errors"`
+	NumTreeErrors   int      `json:"num_tree_errors"`
+	NumUnusedBlobs  int      `json:"num_unused_blobs"`
+	NumDataErrors   int      `json:"num_data_errors"`
+	BrokenPacks     []string `json:"broken_packs"`         // run "restic repair packs ID..." and "restic repair snapshots --forget" to remove damaged files
```
darkdragon-001, line 517: Do we want these IDs in the status or does this encourage people to script such destructive commands?

cmd/restic/cmd_check.go
```diff
@@ -346,22 +368,25 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 		if err != nil {
 			return err
 		}
+		result.NumUnusedBlobs = len(unused)
```
MichaelEischer, line 371: Checking for unused blobs can only be used by tests. So I don't think we should expose that number via the JSON output.

cmd/restic/cmd_check.go
```diff
@@ -316,16 +336,18 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
...
+	result.NumTreeErrors = 0
 	for err := range errChan {
 		errorsFound = true
 		if e, ok := err.(*checker.TreeError); ok {
 			printer.E("error for tree %v:\n", e.ID.Str())
 			for _, treeErr := range e.Errors {
+				result.NumTreeErrors++
 				printer.E("  %v\n", treeErr)
 			}
 		} else {
```
MichaelEischer, line 353: The `printer.E` below is not accounted for.

cmd/restic/cmd_check.go
MichaelEischer: There's at least the category of snapshot errors missing. The main question here is what one can do with  knowing that there are 5 data errors in contrast to 5 index errors. In regards to index errors one could run `repair index` but giving you that information is the task of `suggest_repair_index`. So, I'm leaning somewhat towards only keeping the overall number of errors.

MichaelEischer, line 516: What is a `data` error in contrast to a `pack` error? Both refer to pack files.

cmd/restic/cmd_check.go
MichaelEischer: Well, finding the commands in the output is already possible. What I'm more worried about is that the means to repair a repository will likely evolve over time. Although we could just get rid of old JSON fields by stopping to use them once they are no longer relevant.

cmd/restic/cmd_check.go
```diff
@@ -215,7 +222,13 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 		return errors.Fatal("the check command expects no arguments, only options - please see `restic help check` for usage and flags")
 	}
 
-	printer := newTerminalProgressPrinter(gopts.verbosity, term)
+	result := checkResult{}
+	var printer Printer
+	if !gopts.JSON {
+		printer = ui.NewMessage(term, gopts.verbosity)
+	} else {
+		printer = &progress.NoopPrinter{}
```
MichaelEischer, line 230: Not printing the error message is not an option. Running check without seeing the errors returned by `check` is useless and prevent us and users from debugging the underlying issue.

MichaelEischer (changes requested):
Letting the `check` command only print a summary just works if a repository is not damaged. For a damaged repository, **all** error messages must also be available in the JSON output, otherwise we'll end up in a situation where all relevant information is missing. The summary by itself is mostly useless when it comes to debug what is wrong with a repository.

I wouldn't be terribly opposed to having an error and warning output message that for now just includes the formatted text output. If necessary, we could introduce an "error_code" field in the future to allow for more automation.

cmd/restic/cmd_check.go
darkdragon-001: I merged this into a single error count.

cmd/restic/cmd_check.go
darkdragon-001: I interpret your message as "keep this information inside the JSON". Resolving.

cmd/restic/cmd_check.go
darkdragon-001: Same as https://github.com/restic/restic/pull/5194#discussion_r1938346612. Resolving.

cmd/restic/cmd_check.go
```diff
@@ -157,13 +157,20 @@ func parsePercentage(s string) (float64, error) {
 	return p, nil
 }
 
+type Printer interface {
```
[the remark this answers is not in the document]
darkdragon-001: There is a [`ui.Message` type](https://github.com/restic/restic/blob/536ebefff44a3278fda42550137ac0f15df8d2fe/internal/ui/message.go#L8-L11) but it is not an interface such that one replace the implementation.
There is the [`progress.Printer` interface](https://github.com/restic/restic/blob/536ebefff44a3278fda42550137ac0f15df8d2fe/internal/ui/progress/printer.go#L8-L15) which also requires `NewCounter(description string) *Counter` method. As we don't need the `NewCounter()` method here, I preferred declaring another interface which is compatible with existing types.

cmd/restic/cmd_check.go
```diff
@@ -243,26 +256,26 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 	}
 
 	errorsFound := false
-	suggestIndexRebuild := false
-	mixedFound := false
+	result.HintRepairIndex = false
```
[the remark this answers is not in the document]
darkdragon-001: Removed this unnecessary initialization.

cmd/restic/cmd_check.go
darkdragon-001: Removed this field from the JSON output.

cmd/restic/cmd_check.go
darkdragon-001: I now also increment the unified errors count in the `else` clause.

Would this be the "snapshot error" type as mentioned in https://github.com/restic/restic/pull/5194#discussion_r1938346379?

doc/075_scripting.rst
```diff
@@ -245,6 +245,32 @@ are stored in JSON form. Specifying ``--json``  or ``--quiet`` will suppress any
 non-JSON messages the command generates.
 
 
+check
+-----
+
+The ``check`` command outputs JSON messages with the following format:
```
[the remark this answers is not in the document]
darkdragon-001: I added the `message_type` field.

cmd/restic/cmd_check.go
darkdragon-001: I unified the error counts per https://github.com/restic/restic/pull/5194#discussion_r1899856559.

cmd/restic/cmd_check.go
darkdragon-001: With the `broken_packs`, `suggest_repair_index` and `suggest_repair_index` fields, the JSON output is actionable without the specific error messages.

As an alternative, in https://github.com/restic/restic/pull/5194#issuecomment-2565986326, I suggested dropping the last commit such that this PR is basically f7d567349b6efc60579e358dc49b52ae53e63311. Please let me know if you prefer to follow this approach.

darkdragon-001:
@MichaelEischer Thanks for your review! I acted upon all threads. Please take another look.

cmd/restic/cmd_check.go
MichaelEischer: > With the `broken_packs`, `suggest_repair_index` and `suggest_repair_index` fields, the JSON output is actionable without the specific error messages.

Only for a few trivial cases. Quite a few of the check errors that show up in forum require manual intervention by the user. But without the detailed error messages that would be impossible to tell. In fact, without the error message, `check` degrades to something that essentially just says that it is broken, but without providing any details. Such settings are incredibly annoying to debug (both for the user themselves and by someone else in the forum). (running `check` as second time without `--json` also isn't an option as errors can be transient)

The `checkError` message from https://github.com/restic/restic/commit/f7d567349b6efc60579e358dc49b52ae53e63311 looks more like what I had in mind. Keeping the more detailed `checkResult` summary message type and such an error message don't exclude each other.

cmd/restic/cmd_check.go
```diff
@@ -271,6 +282,8 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 			printer.E("error: %v\n", err)
 		}
 
+		result.NumErrors += len(errs)
+		result.HintRepairIndex = true
 		printer.E("\nThe repository index is damaged and must be repaired. You must run `restic repair index' to correct this.\n\n")
 		return errors.Fatal("repository contains errors")
```
MichaelEischer, line 288: `result` is never printed  in this case.

cmd/restic/cmd_check.go
```diff
@@ -316,7 +333,7 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		bar := newTerminalProgressMax(!gopts.Quiet, 0, "snapshots", term)
+		bar := newTerminalProgressMax(!gopts.Quiet && !gopts.JSON, 0, "snapshots", term)
```
MichaelEischer, line 336: should use `bar := printer.NewCounter("snapshots")`. I forgot to convert the progress bars when switching `check` to `printer.Progress`.

cmd/restic/cmd_check.go
```diff
@@ -479,3 +500,11 @@ func selectRandomPacksByFileSize(allPacks map[restic.ID]int64, subsetSize int64,
 	packs := selectRandomPacksByPercentage(allPacks, subsetPercentage)
 	return packs
 }
+
+type checkResult struct {
+	MessageType     string   `json:"message_type"`
```
MichaelEischer, line 505: Add a comment that the expected message type is `summary`.

cmd/restic/cmd_check.go
MichaelEischer: Those error include the "snapshot error" type, but probably also some other errors.

MichaelEischer (changes requested):
Thanks for the changes so far! I have a few more comments.

cmd/restic/cmd_check.go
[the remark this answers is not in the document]
darkdragon-001: Using `ui.NewMessage()` seems more appropriate and lightweight if we don't need a progress printer.

In f7d567349b6efc60579e358dc49b52ae53e63311 I also added a custom printer type to collect the error messages. I can use a no-op `NewCounter()` method there if you prefer.

cmd/restic/cmd_check.go
darkdragon-001: If it requires manual intervention, running the check manually without `--json` seems appropriate (if the errors are transient, there shouldn't be the need for manual intervention).

cmd/restic/cmd_check.go
darkdragon-001: Done

cmd/restic/cmd_check.go
darkdragon-001: Done in 2b1686d6694cb0c5f676b9dadbb6c6c443c704f8.

cmd/restic/cmd_check.go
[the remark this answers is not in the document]
darkdragon-001: I switched to using the `progress.Printer` interface.

cmd/restic/cmd_check.go
darkdragon-001: The summary wasn't printed in any of the early returns. I added 1a2b9ba37ed0309d42b2cede8836fe99cf9f9d56 to always print the summary JSON.

cmd/restic/cmd_check.go
darkdragon-001: I now keep printing the error messages in JSON as well.

cmd/restic/cmd_check.go
darkdragon-001: FYI: I cannot use this approach for "packs" as this uses `packCount` instead of `0`.

darkdragon-001:
@MichaelEischer I addressed all remaining comments. Could you please take another look?

doc/075_scripting.rst
```diff
@@ -245,6 +245,38 @@ are stored in JSON form. Specifying ``--json``  or ``--quiet`` will suppress any
...
+
+Error
+^^^^^
+
+These errors are printed on ``stderr``.
+
++----------------------+-------------------------------------------+
+| ``message_type``     | Always "error"                            |
++----------------------+-------------------------------------------+
+| ``message``          | Error message                             |
```
MichaelEischer, line 276: Please add a warning that the message `May change in arbitrary ways across restic versions`.

cmd/restic/cmd_check.go
```diff
@@ -355,13 +372,14 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 	doReadData := func(packs map[restic.ID]int64) {
 		packCount := uint64(len(packs))
 
-		p := newTerminalProgressMax(!gopts.Quiet, packCount, "packs", term)
+		p := newTerminalProgressMax(!gopts.Quiet && !gopts.JSON, packCount, "packs", term)
```
MichaelEischer, line 375:
```suggestion
		p := printer.NewCounter("packs")
		p.SetMax(uint64(len(packs)))
```

cmd/restic/cmd_check.go
```diff
@@ -45,7 +46,11 @@ Exit status is 12 if the password is incorrect.
 	RunE: func(cmd *cobra.Command, args []string) error {
 		term, cancel := setupTermstatus()
 		defer cancel()
-		return runCheck(cmd.Context(), checkOptions, globalOptions, args, term)
+		summary, err := runCheck(cmd.Context(), checkOptions, globalOptions, args, term)
+		if globalOptions.JSON {
+			term.Print(ui.ToJSONString(summary))
```
MichaelEischer, line 51: If check fails, then the summary should IMHO always return a non-zero number of errors. I'm pretty sure that everyone is going to check the exit code even though that's absolutely necessary.
```suggestion
			if err != nil && summary.NumErrors == 0 {
				summary.NumErrors = 1
			}
			term.Print(ui.ToJSONString(summary))
```

MichaelEischer (changes requested):
I have a few more small nits, but then we should be done.

doc/075_scripting.rst
darkdragon-001: Done.

_FYI: I previously just copied it from other commands which do not have this addition._

cmd/restic/cmd_check.go
darkdragon-001: Done.

_FYI: I had this in a previous version but then decided against it and only count errors arising from a broken repository but not from a failing check action._

cmd/restic/cmd_check.go
MichaelEischer: passing the maximum value to `newTerminalProgressMax` or setting it via `SetMax` is equivalent

cmd/restic/cmd_check.go
MichaelEischer: I'm aware it isn't ideal. But making sure that users noticed failed check runs has priority

cmd/restic/cmd_check.go
darkdragon-001: I only now noticed the `p.SetMax()` call and fixed it. Done.

MichaelEischer (approved):
LGTM. Thanks!

PR #5201, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 2-Clause License
 
-Copyright (c) 2014, Alexander Neumann <alexander@bumpern.de>
+Copyright (c) 2025, Alexander Neumann <alexander@bumpern.de>
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without


rawtaz:
Thank your for your attention! I'm pretty sure this change should not be made though. I did some research on this recently and the conclusion is basically the following in short summary:

- A copyright note is not really needed at all as copyright law is automatic, but it doesn't hurt and reminds readers that copyright isn't theirs by default.
- In a copyright notice, years aren't really needed, but like the previous point it doesn't hurt.
- If you do include year(s), a range is not necessary, a range that is kept up to date is mostly useful to indicate an active and maintained project.
- If you skip the range and put just one year in there, it's better that it's the initial start year of the copyright, as that gives the most accurate representation of the copyright (putting the current year suggests that copyright started this year and not earlier).

So in summary, all of it is basically just cosmetics. Personally I think it feels good to have a copyright notice but I generally lean on not having a date in it at all, as that is the most encompassing notice.

In restic's case, I think it's better to just leave it as it is - the current notice conveys that there is a copyright since 2014, which is accurate.

Issue #5193, opened 2024-12-30
Allow `forget` to relate policy to current time and not just latest snapshot


rawtaz:
Output of `restic version`
--------------------------

restic 0.17.3-dev (v0.17.3-310-gde3acd793) compiled with go1.22.5 on darwin/arm64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Provide a way to relate `forget` to not only the latest/most recent snapshot (current behavior), but also to all or at least the first one of:

1. the current time of executing `forget`
1. a snapshot ID (effectively meaning the date/time of that snapshot)
1. a specific date/time

This could be implemented with a new option to the `forget` command, named for example `--relative`, with usage being e.g.:

1. `forget --keep-within 1y --relative now`
1. `forget --keep-within 1y --relative 37b159d6`
1. `forget --keep-within 1y --relative '2012-11-01 22:08:41'`

or the current behavior:

- `forget --keep-within 1y --relative latest`

This option could apply to all of the `--keep-*` policy options, or just the `--keep-within*` ones, depending on complexity and usefulness of the option.

In order to keep restic's precationary nature of "better safe than sorry", in which it always keeps the latest snapshot unless you specify the `--unsafe-allow-remove-all` option, this should still be the case. In other words, if you run e.g. `forget --keep-within 1y --relative now` when there are only snapshots older than one year, restic should keep the most recent one of those despite it being more than one year old, unless you specify `--unsafe-allow-remove-all` in which case restic would delete all snapshots that were older than one year.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

It is not uncommon that users are surprised that e.g. `--keep-within 1y` does not remove snapshots that are older than one year back in time from "now" when there are no newer snapshots within that year from "now", despite the [documentation](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) specifically mentioning that the `--keep-within*` options relate to _the latest (most recent) snapshot_.

It is also pretty common that users want to think in terms of "relative to right now", and design their `forget` policy based on that reference point. I think it's common enough that restic should consider being able to do this.

There are several examples of why the proposed option and its values could be useful, for example:

[...]

PR #5209, opened 2025-01-14
Add snapshot filtering --older-than and --newer-than

### LINKED ISSUE #5136: Ability to filter snapshots by date
Output of `restic version`
--------------------------

restic 0.17.1 compiled with go1.23.1 on linux/amd64

When using the `snapshots` command it would be useful if one could filter by date (before or after a certain date, or within a range).

Other commands that could benefit from such a filter: `copy`, `tag` (see related #2078)
For `forget` it's mostly covered by the --keep-within options, though one could consider adding inverse options (eg: --forget-before \<date\>, --forget-after \<date\>, --forget-within \<date range\>, though the naming of the last one could be misleading if one expects it to behave similar to --keep-within).

See/copy/tag only the snapshots I'm interested in without having to specify their IDs (assuming it can't be done with the existing host, path or tag filters).

It makes me happy every time it Just Works™, which is almost always.

kjetilho:
This patch adds generic options for filtering snapshots for commands which work on a set of snapshots, like `restic snapshots` and `restic forget`.

The output of `snapshots` can be overwhelming, and it can be useful to only list this day's backups using `--newer-than 1d` for instance.

This functionality can also be used to forget all snapshots within a time period, e.g.,

`restic forget --older-than 90d --unsafe-allow-remove-all`

Yes, this closes #5136.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5136
+++ b/changelog/unreleased/issue-5136
@@ -1,0 +1,10 @@
+Enhancement: Add snapshot filtering options --older-than and --newer-than
+
+Restic subcommands which work on a set of snapshots (``copy``,
+``find``, ``forget``, ``snapshots``, ``tag`` and more) can now filter
+on the absolute age of the snapshots specified as a duration, e.g.,
+``--newer-than 1d12h`` to only operate on snapshots created less than
+36 hours ago, or ``--older-than 90d`` to operate on snapshots older
+than three months.
+
+https://github.com/restic/restic/issues/5136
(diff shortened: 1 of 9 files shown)


MichaelEischer (commented):
See the comments for a few nits.

While looking at the affected commands, I've noticed that `find` already has
```
  -N, --newest string       newest modification date/time
  -O, --oldest string       oldest modification date/time
```
options, but that's fine I guess.

One other question that came to my mind is whether `--older-than duration` or `--older-than timestamp` is the more useful variant of this command.
[Edit]It might be possible to have those options accept both. A duration and timestamp have a different format. Then we could also deprecate the `find` options.[/Edit]

MichaelEischer:
As noticed by @rawtaz: we'll have to discuss how this would interact with https://github.com/restic/restic/issues/5193 before proceeding any further.

PR #5212, opened 2025-01-18
prune: correctly account for duplicates in max-unused check

MichaelEischer:
The size comparison for `--max-unused` only accounted for unused but not for duplicate data. For repositories with a large amount of duplicates this can result in a situation where no data gets pruned even though the amount of unused data is much higher than specified.

Fixes https://forum.restic.net/t/restic-not-obeying-max-unused-parameter-on-prune/8879/10 .

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5215, opened 2025-01-26
PR dead: restic backup: limit maximum size of repository during backup - new implementation with PR #5338

wplapper:
A new option --max-repo-size is introduced and then passed on to repository.saveAndEncrypt()
The current size of the repository is monitored when a blob gets saved. If the maximum is exceeded, a fatal error will be triggered.

The snapshot gets terminated gracefully and a new partial snapshot will be added to the repository.

The PR adresses https://github.com/restic/restic/issues/4585

https://github.com/restic/restic/issues/4585

Checklist
---------

- [x] I have added tests for some code changes - more to come!

### DIFF
--- a/changelog/unreleased/issue-4585
+++ b/changelog/unreleased/issue-4585
@@ -1,0 +1,29 @@
+Enhancement: Limit repository size to a predefined maximum size during vackup
+
+Restic backup can now limit the repository size to a given maximum value. The
+repository size limit can be specified as option `--max-repo-size`, with
+the usual meaning for size related options.
+
+During backup, the current size is monitored by calculating its actual repository
+size plus the size of any new blob added during the backup process. Once the
+defined limit is exceeded, backup is prevented from creating new backup entries
+and the 'in progress' files are finalized, without adding large amounts of new data.
+
+The size limit is a rough limit and cannot be taken as a precise limit,
+since indexes and snapshot file have to be finalized.
+Due to the highly parallel processing of a backup process
+the limit can be overshot by `packfile size` multiple times.
+With a current default of 16m for a packfile,
+you might experience differences of multiple megabytes between the expected limit
+and the actual size when the backup process terminates.
+
+The new packfiles are created by the current backup process and a new snapshot
+will be created.
+
+Currently the implemenation is incomplete. There are currently two indications
+that there is a incomplete backup:
+field `PartialSnapshot` exist and is set to true and
+the snapshot is tagged as `partial-snapshot`.
+
+https://github.com/restic/restic/issues/4583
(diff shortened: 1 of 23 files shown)


changelog/unreleased/issue-4585
```diff
@@ -0,0 +1,17 @@
...
+Restic backup can now limit the repository size to a given maximum value.
+
+During backup, the current size is constantly monitored by watching its initial
+size plus the sum of "DataSizeInRepo" and "TreeSizeInRepo".
+
+The new packfiles are created by the current backup process, but have not been
+linked to a new snapshot. Therefore they are unreferenced, but
+are still occupying space in tge repository.
+
+I need help with shutting down the backup process properly, so the failed backup
```
MichaelEischer, line 12: What's that doing in the changelog entry?

internal/archiver/archiver.go
```diff
@@ -508,6 +513,13 @@ func (arch *Archiver) save(ctx context.Context, snPath, target string, previous
 	case fi.Mode.IsRegular():
 		debug.Log("  %v regular file", target)
 
+		// check if repository size is about to be exceeded.
```
MichaelEischer, line 516: This can only check after the fact, but not beforehand. Reading from the summary also requires taking the `arch.mu` mutex.

But there's a larger problem here. When someone backs up very large files, this can miss the target size by hundreds of gigabytes (assuming such large files).

changelog/unreleased/issue-4585
MichaelEischer: Restic currently doesn't have the concept of a partial snapshot. Storing such a snapshot will currently cause the next backup run to re-read all files that didn't make it into the partial snapshot.

Adding a corresponding flag to the snapshot shouldn't be too hard. Similarly, the archiver and fileSaver could be adapted to stop reading new files. Then the whole backup process would quickly unwind itself and finish with a partial snapshot. (This should happen in a separate PR as partial snapshots are a useful feature in itself). Afterwards, we can think about repository size limits again. The main remaining complexity would be estimating how much space would still be necessary to write the remaining index and tree blobs when approaching the size limit.

MichaelEischer (commented):
I've added a few comments regarding the approach. In short, a proper implementation will require a significant amount of supporting infrastructure.

wplapper:
I put this PR on hold, since I do not understand the full complexity of all the issues involved.

wplapper:
Working on a completely new implementation.

wplapper:
I will close this PR, since it contains too many changes.I will split it as multiple PRs, starting off with a partial snapshot.

PR #5217, opened 2025-01-27
restore: reflink identical files on extraction

intelfx:

This PR teaches `restic restore` to detect identical files in the restored set
and reflink them together transparently to the user, instead of writing each
file all over.

Reflinks are a type of on-demand data deduplication implemented by certain
modern (usually copy-on-write) filesystems such as btrfs, zfs or bcachefs
(out of classic filesystems, xfs notably implements reflinks as well).

Unlike other types of links such as hardlinks / symlinks, which cause changes to
one of the linked paths to "bleed through" to other paths, the blocks shared via
reflinks are automatically unshared whenever one of the reflinked files (ranges)
is modified, making it completely safe to use transparently.

If reflinking is not possible (which might be the case if the filesystem does
not support reflinks, or if parts of the restored set are being extracted onto
different mountpoints), a fallback to plain copy is included.

This feature is controlled by the `--reflinks` flag to `restic restore`.

IRC.

Checklist
---------

Notes
-----

This is an RFC for early review. I'm using a personal fork of a third party
library as a hash map for non-comparable types (to track the blob IDs), a bunch
of ideas are unimplemented (such as graceful handling of situations where some
reflink operations fail and some don't, such as when extracting onto different
mountpoints), and there's less fallbacks in place than I'd like.

### DIFF
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -71,6 +71,7 @@ type RestoreOptions struct {
 	Verify              bool
 	Overwrite           restorer.OverwriteBehavior
 	Delete              bool
+	Reflinks            bool
 	ExcludeXattrPattern []string
 	IncludeXattrPattern []string
 	OwnershipByName     bool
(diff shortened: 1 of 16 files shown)


patrickwolf:
This is a great addition to transparently created reflinks! I hope it gets approved and implemented !

patrickwolf:
I'm happy to see progress on the PR. Thanks @intelfx

patrickwolf:
I very much hope it will make it into a restic release one day. A backup with many reflinks may be impossible to restore without this feature, since there may not be enough storage capacity to duplicate all the data that those reflinks deduplicate.

PR #5219, opened 2025-01-28
CI: bump golangci-lint and silence warnings about shadowed builtins

MichaelEischer:
Removing the shadowing cases leads to weird workarounds but doesn't help much with code clarity. So just disable the rule instead.

No. But it's getting on my nerves.

Checklist
---------


rawtaz (approved):
I hereby approve the suggested changes 🫡

PR #5232, opened 2025-02-02
Bump minimum go version to 1.22

MichaelEischer:
The dependency update https://github.com/restic/restic/pull/5223 requires Go 1.22. There's also at least one open PR that would benefit from Go 1.22. With Go 1.22 as minimum version, only TLS 1.2 and higher will be supported.

No.

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5239, opened 2025-02-05
Json prune

### LINKED ISSUE #3129: Provide JSON output on prune
Output of `restic version`
--------------------------
restic 0.11.0 (v0.11.0-30-g4a9b05af) compiled with go1.15.2 on linux/amd64

`prune` command currently doesn't seem to support `--json` parameter

Prune is an action which can take serious time. One who'd like to automatize such stuff might not have luxury to dive into logs for each prune run. Being able to parse a stable machine-readable output would help checking metrics like prune time, deleted/repacked data size etc. a lot easier.

Are you kidding, restic iz da bomb :bomb:

darkdragon-001:

Add JSON output to the `prune` command.

This is based on #3140.

Fixes #3129

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_prune.go
+++ b/cmd/restic/cmd_prune.go
@@ -175,7 +175,7 @@ func runPrune(ctx context.Context, opts PruneOptions, gopts global.Options, term
 		return errors.Fatal("--no-lock is only applicable in combination with --dry-run for prune command")
 	}
 
-	printer := ui.NewProgressPrinter(false, gopts.Verbosity, term)
+	printer := ui.NewProgressPrinter(gopts.JSON, gopts.Verbosity, term)
 	ctx, repo, unlock, err := openWithExclusiveLock(ctx, gopts, opts.DryRun && gopts.NoLock, printer)
 	if err != nil {
 		return err
@@ -190,10 +190,10 @@ func runPrune(ctx context.Context, opts PruneOptions, gopts global.Options, term
 		opts.unsafeRecovery = true
 	}
 
-	return runPruneWithRepo(ctx, opts, repo, restic.NewIDSet(), printer)
+	return runPruneWithRepo(ctx, opts, gopts, repo, restic.NewIDSet(), printer)
 }
 
-func runPruneWithRepo(ctx context.Context, opts PruneOptions, repo *repository.Repository, ignoreSnapshots restic.IDSet, printer progress.Printer) error {
+func runPruneWithRepo(ctx context.Context, opts PruneOptions, gopts global.Options, repo *repository.Repository, ignoreSnapshots restic.IDSet, printer progress.Printer) error {
 	if repo.Cache() == nil {
 		printer.S("warning: running prune without a cache, this may be very slow!")
 	}
@@ -230,9 +230,13 @@ func runPruneWithRepo(ctx context.Context, opts PruneOptions, repo *repository.R
 		printer.P("\nWould have made the following changes:")
 	}
 
-	err = printPruneStats(printer, plan.Stats())
-	if err != nil {
-		return err
+	if !gopts.JSON {
+		err = printPruneStats(printer, plan.Stats())
+		if err != nil {
+			return err
+		}
+	} else {
+		gopts.Term.Print(ui.ToJSONString(plan.Stats()))
 	}
 
 	// Trigger GC to reset garbage collection threshold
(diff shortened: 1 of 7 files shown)


cmd/restic/cmd_prune.go
```diff
@@ -232,24 +241,19 @@ func printPruneStats(printer progress.Printer, stats repository.PruneStats) erro
 	if stats.Size.Unref > 0 {
 		printer.V("unreferenced:                    %s\n", ui.FormatBytes(stats.Size.Unref))
 	}
-	totalBlobs := stats.Blobs.Used + stats.Blobs.Unused + stats.Blobs.Duplicate
-	totalSize := stats.Size.Used + stats.Size.Duplicate + stats.Size.Unused + stats.Size.Unref
-	unusedSize := stats.Size.Duplicate + stats.Size.Unused
```
darkdragon-001, line 237: Why does unused also include duplicate? Should this be saved in `stats.Size.Unused` such that this is also used in the JSON output?

darkdragon-001: Reading #5212 this seems intended.

Keeping unused and duplicate separate values in the stats (which are also used for JSON output).

darkdragon-001:
@MichaelEischer Friendly ping.

darkdragon-001:
@MichaelEischer could you please take a look? Thanks!

WiggleWizard:
This would be awesome to get into an official release, currently working on a piece of software that needs script support for pruning!

agaengel:
I can only agree it would be awesome if the prune command also would support json

darkdragon-001:
I rebased and resolved conflicts. @MichaelEischer can you please take a look? Thanks!

darkdragon-001:
@rawtaz could you please take a look?

darkdragon-001:
@fd0 could you please take a look? Thanks!

internal/repository/prune.go
```diff
@@ -33,32 +33,42 @@ type PruneOptions struct {
 
 type PruneStats struct {
```
MichaelEischer, line 36: The data structure must contain a field `message_type` and be documented at https://github.com/restic/restic/blob/master/doc/075_scripting.rst .

internal/repository/prune.go
```diff
@@ -33,32 +33,42 @@ type PruneOptions struct {
...
+		Used         uint `json:"used"`
+		Duplicate    uint `json:"duplicate"`
+		Unused       uint `json:"unused"`
+		Total        uint `json:"total"`
+		Repack       uint `json:"repack"`
+		Repackrm     uint `json:"repack_remove"`
+		Remove       uint `json:"remove"`
+		RemoveTotal  uint `json:"remove_total"`
+		Remain       uint `json:"remaining"`
+		RemainUnused uint `json:"remaining_unused"`
```
MichaelEischer: Field is never used. -> Should be removed.

internal/repository/prune.go
```diff
@@ -33,32 +33,42 @@ type PruneOptions struct {
...
-		Unref      uint
-		Keep       uint
-		Repack     uint
-		Remove     uint
-	}
+		Used        uint `json:"used"`
+		Unused      uint `json:"unused"`
+		PartlyUsed  uint `json:"partly_used"`
+		Unref       uint `json:"unreferenced"`
+		Total       uint `json:"total"`
```
MichaelEischer, line 68: Only calculated but not used for output. Although it's probably fine to keep it in the JSON output.

Same for `RemoveTotal`

MichaelEischer (changes requested):
The JSON message needs a `message_type` to keep the door open to also add the progress bars to the JSON output in the future.

Please add a minimal test to make sure that the JSON output can actually be decoded and contains the expected message (and maybe verify a few fields).

internal/repository/prune.go
darkdragon-001: Removed.

internal/repository/prune.go
darkdragon-001: I feel like they can be useful in the JSON output. Please let me know if you would prefer to have those removed.

darkdragon-001 (commented):
Thanks for your review!

> Please add a minimal test to make sure that the JSON output can actually be decoded and contains the expected message (and maybe verify a few fields).

Done.

internal/repository/prune.go
darkdragon-001: Done.

MichaelEischer (approved):
LGTM. I've slightly changed the description for some of the JSON output fields to better represent the values. And the JSON output is no longer garbled when using `--no-cache`.

darkdragon-001:
> LGTM. I've slightly changed the description for some of the JSON output fields to better represent the values. And the JSON output is no longer garbled when using `--no-cache`.

Thank you!

PR #5240, opened 2025-02-06
Improve JSON output type documentation

MichaelEischer:
So far the documentation of the JSON outputs did not contain datatypes which complicates handling that output. Add datatype descriptions that are hopefully precise enough to be useful. `int` is documented as `int64` as restic is (hopefully) primarily used as 64-bit build.

The PR also adds several missing fields and marks the `short_id` field as deprecated (use `id` instead).

I've also experimented with properly marking optional fields. However, the use of `omitempty` so far seems to be very inconsistent. Documenting this inconsistency in details would make it harder to clean it up in the future.

No.

Checklist
---------


MichaelEischer (commented):
LGTM.

Issue #4759, opened 2024-04-10
invalid environment variable values silently ignored


eharris:
Output of `restic version`
--------------------------

restic 0.16.4 (v0.16.4-0-g3786536dc) compiled with go1.21.8 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------

local/any

Problem description / Steps to reproduce
----------------------------------------

`restic` seems to silently ignore problems with some environment variable parameters.
In my instance, the documentation wasn't clear that `RESTIC_PACK_SIZE` needed to be numeric only (and was in MB) and it silently ignored `RESTIC_PACK_SIZE=64M` and continued to use the default rather than issuing a warning or failing.
It does properly warn if the problem parameter is on the command line.

Expected behavior
-----------------

I would expect `restic` to warn or fail if an environment variable it is using is not valid (as long as it is not being overridden on the command line).

Actual behavior
---------------

restic silently ignores problems with environment variables like `RESTIC_PACK_SIZE`.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Using it in a new situation, working pretty well!

MichaelEischer:
This is unfortunately rather complex to properly fix. Currently, environment variables are parsed inside `init()` functions before the regular execution starts. At that point, the code has no idea which command will be called in the end. Blindly failing if the environment variable value leads to a few weird corner cases:
- shell autocompletion will stop working if an invalid environment variable is set
- environment variables for a different command can cause the current one to fail
- `RESTIC_PACK_SIZE=64M restic --pack-size 64 ...` would still fail

Avoiding  these corner cases, however, requires restructuring how all commands are defined. And I'm not yet sure how a correct implementation would look like.

konidev20:
Hey @MichaelEischer
Do you think we can use something similar to what's present in the `internal/backend/backend.go` where there is an interface called `ApplyEnvironmenter` this interface has a method called `ApplyEnvironment`.

We enforce that all backend configurations must implement this method to read any required environment variables. 

Do you think something similar can be done for the Command options, where we can enforce an order of reading environment variables, command options and a validate function?

MichaelEischer:
> Do you think something similar can be done for the Command options, where we can enforce an order of reading environment variables, command options and a validate function?

Adding something similar for the global options should be relatively easy. Extending that to more commands in a clean way isn't that easy however. While we could introduce some callback mess, I'd like to invert how the command setup works to get rid of all the involved global variables. 

[...]

PR #5241, opened 2025-02-07
Refactor CLI command initialization to use less global state

MichaelEischer:
Refactor the CLI command setup to a better testable pattern with much less global state. The main changes are as follows:
- Each options struct now has a `AddFlags(f *pflag.FlagSet)` method that is used to add flags to a command
- Commands are created using  `new<NAME>Command()`  functions. The options for this command are enclosed in the closure. This removes the global options variables for every command.
- The root command now explicitly constructs all child commands. This will allow returning errors later on.
- globalOptions still exists as it is used in several global functions. Most prominently the Print and Verbose functions.
- Commands that are gated by build tags now use a `register<NAME>Command(cmd)` function that is a no-op when the command is not enabled. As an additional cleanup the profiling features are now implemented by wrapping the Pre and PostRun functions.

The new patterns also will allow returning errors if environment variables contain errors etc. I've not made any such changes as the PR is already for too large.

No. Suggested in https://github.com/restic/restic/issues/4759#issuecomment-2131275416

Checklist
---------

- ~~[ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).~~ Not yet user visible. Any visible change is a bug.

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -31,10 +31,13 @@ import (
 	"github.com/restic/restic/internal/ui/termstatus"
 )
 
-var cmdBackup = &cobra.Command{
-	Use:   "backup [flags] [FILE/DIR] ...",
-	Short: "Create a new backup of files and/or directories",
-	Long: `
+func newBackupCommand() *cobra.Command {
+	var opts BackupOptions
+
+	cmd := &cobra.Command{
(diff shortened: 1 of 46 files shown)


MichaelEischer (commented):
LGTM.

PR #5243, opened 2025-02-08
mount: append the repository ID to the name of the FUSE mount

### LINKED ISSUE #4868: show repository id in output of df and mount
Output of `restic version`
--------------------------
0.16.4
Which functionality do you think we should add?

Please add the repository id to output shown by the commands mount and df, so it would look like

```
Filesystem                     Type             Size  Used Avail Use% Mounted on
server:/export                 nfs4              11T  8.9T  1.4T  87% /server
restic:228ffbed4             fuse             0.0K  0.0K  0.0K       - /mnt
```

MichaelEischer pointed to the relevant source code:
https://github.com/restic/restic/blob/2280fbfd2e8e3b6eeba8ebaa95d06b9e4519ced9/cmd/restic/cmd_mount.go#L142

This makes it easier to see which repositories are mounted where. 
```restic cat config | jq .id``` will show the repository id

### PR DESCRIPTION
**Pull request by konidev20**
Append the repository ID to the name of the FUSE mount created using the restic `mount` command. When the user is mounting multiple repositories, this makes it easier to understand which repository is mounted where. 

[Edit]
Example output:
```
[konidev20@vm-1 restic]$ df ./test-mount/
Filesystem      1K-blocks  Used Available Use% Mounted on
restic:d3b07384d1         0     0         0    - /home/konidev20/restic/test-mount
```

In the above example the ID `d3b07384d1` is appended to the filesystem name. This is the first 10 characters of the repository ID.
[/Edited]
Closes #4868 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 	"time"
@@ -148,9 +149,11 @@ func runMount(ctx context.Context, opts MountOptions, gopts GlobalOptions, args 
 		return err
 	}
 
+	fuseMountName := fmt.Sprintf("restic:%s", repo.Config().ID[:10])
+
 	mountOptions := []systemFuse.MountOption{
 		systemFuse.ReadOnly(),
-		systemFuse.FSName("restic"),
+		systemFuse.FSName(fuseMountName),
 		systemFuse.MaxReadahead(128 * 1024),
 	}
 
--- a/changelog/unreleased/issue-4868
+++ b/changelog/unreleased/issue-4868
@@ -1,0 +1,14 @@
+Enhancement: Include repository id in filesystem name used by `mount`
+
+The filesystem created by restic's `mount` command now includes the repository
+id in the filesystem name. The repository id is printed by restic when opening
+a repository or can be looked up using `restic cat config`.
+
+```
+[restic-user@hostname restic]$ df ./test-mount/
+Filesystem        1K-blocks  Used Available Use% Mounted on
+restic:d3b07384d1         0     0         0    - /mnt/my-restic-repo
+```
+
+https://github.com/restic/restic/issues/4868
+https://github.com/restic/restic/pull/5243

### CODE REVIEW

File: internal/repository/repository.go [hunk 1]
```diff
@@ -143,6 +143,15 @@ func (r *Repository) Config() restic.Config {
 	return r.cfg
 }
 
+// ID returns the repository ID.
+func (r *Repository) ID() (restic.ID, error) {
+	id, err := restic.ParseID(r.cfg.ID)
+	if err != nil {
+		return restic.ID{}, err
+	}
+	return id, nil
```
MichaelEischer, line 152:
```suggestion
	return restic.ParseID(r.cfg.ID)
```
results in exactly the same behavior.

cmd/restic/cmd_mount.go
```diff
@@ -141,9 +142,16 @@ func runMount(ctx context.Context, opts MountOptions, gopts GlobalOptions, args
 		return err
 	}
 
+	repositoryID, err := repo.ID()
+	if err != nil {
+		return err
+	}
+
+	fuseMountName := fmt.Sprintf("restic:%s", repositoryID.Str())
```
MichaelEischer, line 150: I'd prefer to use `s.Config().ID[:10]` to get the short repository ID. That snippet is already used to fill in the ID in the `"created restic repository %v at %s"` message printed on startup.

MichaelEischer (changes requested):
The id part should match that printed in the `"created restic repository %v at %s"` messages.

internal/repository/repository.go
konidev20: I removed this method entirely, since it's no longer needed with the method you shared in the next comment.

cmd/restic/cmd_mount.go
konidev20: I have updated the mount point to use `repo.Config().ID[:10]` instead of a new method which parses the repository ID.

konidev20:
This is ready for review.

MichaelEischer:
Please add a changelog entry. I had something like the following in mind: (the inner part is in triple backticks)

```
Enhancement: Include repository id in filesystem name used by `mount`

The filesystem created by restic's `mount` command now includes the repository
id in the filesystem name. The repository id is printed by restic when opening
a repository or can be looked up using `restic cat config`.

```
$ df ./test-mount/
Filesystem      1K-blocks  Used Available Use% Mounted on
restic:1234567890         0     0         0    - /restic/test-mount
```

https://github.com/restic/restic/issues/4868
https://github.com/restic/restic/pull/5243
```

konidev20:
@MichaelEischer I have added the changelog entry as requested.

MichaelEischer (approved):
LGTM. Thanks!

PR #5244, opened 2025-02-09
sftp: fix incorrect permissions on data directories created on init

### LINKED ISSUE #4447: Incorrect directory permissions after init via SFTP
Output of `restic version`
--------------------------
Versions tested:
restic 0.16.0 compiled with go1.20.6 on linux/amd64 (official standalone binary)
restic 0.14.0 compiled with go1.20.7 on linux/amd64 (Debian sid)
restic 0.15.1 compiled with go1.20.1 on openbsd/amd64 (OpenBSD 7.3)

restic -r sftp:example.com:repo init

SFTP to Debian bookworm, OpenBSD 7.3 and rsync.net.

Expected behavior
-----------------
Like in a local repo, directories should be created with 700 permissions. 

Actual behavior
---------------
Directories are created with 755 permissions instead.

See above: simply init a new repo via SFTP.

No.

No.

konidev20:
The SFTP Client's Mkdir or MkdirAll function doesn't allow setting a mode for the directories being created. It was suggested that we implement the MkdirAll function of the SFTP backend while setting the correct permissions from the DefaultModes.

Directories will now be created with the permission 700, instead of previous 750.

Closes #4447 

Checklist
---------

### DIFF
--- a/internal/backend/sftp/sftp.go
+++ b/internal/backend/sftp/sftp.go
@@ -166,6 +166,18 @@ func open(sftp *SFTP, cfg Config) (*SFTP, error) {
 	return sftp, nil
 }
 
+// Mkdir creates a directory with the DefaultModes.Dir permissions
+func (r *SFTP) Mkdir(path string) error {
+	err := r.c.Mkdir(path)
+	if err != nil {
+		return err
+	}
+
+	// Set correct permissions after directory is created
+	return r.c.Chmod(path, r.Modes.Dir)
+}
+
+// Update mkdirAllDataSubdirs function
 func (r *SFTP) mkdirAllDataSubdirs(ctx context.Context, nconn uint) error {
 	// Run multiple MkdirAll calls concurrently. These involve multiple
 	// round-trips and we do a lot of them, so this whole operation can be slow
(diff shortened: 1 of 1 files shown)


internal/backend/sftp/sftp.go
```diff
@@ -594,3 +596,64 @@ func (r *SFTP) Warmup(_ context.Context, _ []backend.Handle) ([]backend.Handle,
...
+	}
+
+	// Set directory permissions
+	return r.c.Chmod(path, r.Modes.Dir)
+}
+
+// MkdirAll creates a directory named path, along with any necessary parents.
+// If path is already a directory, MkdirAll does nothing and returns nil.
+// If path exists but is a file, an error is returned.
+func (r *SFTP) MkdirAll(path string) error {
```
MichaelEischer, line 632: Why does this method behave rather different from the `MkdirAll` implementation in the sftp package, when the goal of this PR is to just fix the directory permissions?

The sftp package checks if the parent folder exists, if no then it recursively calls itself, and afterwards creates the requested directory. That requires much fewer calls to the server than this implementation.

internal/backend/sftp/sftp.go
```diff
@@ -181,10 +183,10 @@ func (r *SFTP) mkdirAllDataSubdirs(ctx context.Context, nconn uint) error {
 			// round trip, not counting duplicate parent creations causes by
 			// concurrency. MkdirAll first does Stat, then recursive MkdirAll
 			// on the parent, so calls typically take three round trips.
-			if err := r.c.Mkdir(d); err == nil {
+			if err := r.Mkdir(d); err == nil {
```
MichaelEischer, line 186: The idea of the old implementation is to speculatively create the directory and only resort to more expensive code paths if that didn't work. After all the default situation here is that the parent folder already exists and only the new child folder has to be created. So this PR shouldn't break that optimization.

MichaelEischer (changes requested):
The PR will regress the repository creation performance of SFTP, see comments for more details.

konidev20:
Closing since I need to work on it again with a new approach. I will find some time and take another stab at it.

PR #5251, opened 2025-02-17
Retry temporary rclone backend errors

MichaelEischer:

rclone backends apparently sometimes temporarily report a file as missing even though it actually exists. This results in problems with the backend error redesign in restic 0.17.0, which relies on accurate information about missing files. Thus this PR reintroduces a limited number of retries when using rclone as backend.

Fixes https://forum.restic.net/t/data-does-not-exist-check-successful-on-second-attempt-but-restic-returns-non-zero-error-code/9117
Fixes https://forum.restic.net/t/fatal-repository-contains-errors-when-only-errors-are-successfully-retried-operations/8971

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5251
+++ b/changelog/unreleased/pull-5251
@@ -1,0 +1,9 @@
+Enhancement: Improve handling of flaky rclone backends
+
+Since restic 0.17.0, the backend retry mechanisms relies on backends correctly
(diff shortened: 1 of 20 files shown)


MichaelEischer (commented):
LGTM.

PR #5262, opened 2025-02-28
A set of mostly automated Go cleanups for the code

Martin2112:
A set of hopefully safe Go Code cleanups.

A bunch of Go cleanups. These were issues found by code inspection tools. Each commit is an independent fix. Please review each one. If you don't want them all then please let me know in the review.

No. I thought these were small and safe changes.

Checklist
---------

- [x] I have added tests for all code changes. There should be no behaviour change.
- [x] I have added documentation for relevant changes (in the manual). No user visible changes
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)). No user visible changes.

### DIFF
--- a/cmd/restic/cmd_diff.go
+++ b/cmd/restic/cmd_diff.go
@@ -465,8 +465,8 @@ func runDiff(ctx context.Context, opts DiffOptions, gopts GlobalOptions, args []
 		Printf("Others:      %5d new, %5d removed\n", stats.Added.Others, stats.Removed.Others)
 		Printf("Data Blobs:  %5d new, %5d removed\n", stats.Added.DataBlobs, stats.Removed.DataBlobs)
 		Printf("Tree Blobs:  %5d new, %5d removed\n", stats.Added.TreeBlobs, stats.Removed.TreeBlobs)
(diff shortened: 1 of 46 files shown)


MichaelEischer (approved):
LGTM. Thanks for the cleanups!

What tool(s) did you use? I wonder whether we can fine tune golangci-lint to avoid reintroducing such issues.

MichaelEischer:
Please take a look at the test errors, the changes partially conflict with the expectations of golangci-lint and also break compilation on Windows.

MichaelEischer (changes requested):
Please take a look at the test errors, the changes partially conflict with the expectations of golangci-lint and also break compilation on Windows.

Martin2112:
I'm using IntelliJ code inspections for these. I started with these categories as they're less likely to be false positives.

The unsafe pointer error appeared after I synced. I'll have a look.

Martin2112:
I've fixed a couple of lint issues. Not sure they're all related to this PR but will have another look when I have more time.

MichaelEischer (approved):

PR #5270, opened 2025-03-01
build: improve GoVersion comparison logic

### LINKED ISSUE #5269: Bug in GoVersion.AtLeast() function causes incorrect version comparisons
Output of `restic version`
--------------------------
N/A - This is a bug in the build code, not related to a specific restic version.

N/A - This is a bug in the source code, not related to a specific backend.

There's a logical flaw in the GoVersion.AtLeast() function in build.go that causes incorrect version comparisons. The current implementation checks each version component (Major, Minor, Patch) independently without respecting the version hierarchy.
To reproduce, try comparing the following versions:
```go
v1 := GoVersion{2, 1, 0}
v2 := GoVersion{1, 5, 5}
result := v1.AtLeast(v2) // Should return true, but returns false
```
Another example:
```go
v1 := GoVersion{1, 18, 0}
v2 := GoVersion{1, 17, 5}
result := v1.AtLeast(v2) // Should return true, but returns false
```

Expected behavior
-----------------
The function should compare version components in hierarchical order (Major, then Minor, then Patch) and return early when a difference is found.
For example:
- If Major versions differ, only the Major version should matter
- If Major versions are equal but Minor versions differ, only the Minor version should matter
- Only when both Major and Minor versions are equal should the Patch version be considered

Actual behavior
---------------
The current implementation checks all components independently:
- If v.Major < other.Major, it returns false (correct)
- If v.Minor < other.Minor, it returns false (incorrect when Major versions differ)
- If v.Patch < other.Patch, it returns false (incorrect when Major or Minor versions differ)
This causes higher versions to be incorrectly considered lower than older versions in many cases.

The implementation is likely an oversight in the logical structure of version comparison. Semantic versioning comparison requires a cascading check rather than independent checks of each component.

Proposed fix:
```go
func (v GoVersion) AtLeast(other GoVersion) bool {
    var empty GoVersion

    // the empty version satisfies all versions
    if v == empty {
        return true
    }

    // Compare major version first
    if v.Major > other.Major {
        return true
    }
    if v.Major < other.Major {
        return false
    }
    
    // Major versions are equal, compare minor versions
    if v.Minor > other.Minor {
        return true
    }
    if v.Minor < other.Minor {
        return false
    }
    
    // Major and minor versions are equal, compare patch versions
    return v.Patch >= other.Patch
}
```

Restic is a fantastic backup tool that I rely on regularly. The project's attention to detail and commitment to code quality makes me happy to contribute back by reporting this issue.

prajwalbharadwajbm:

This PR fixes a logical flaw in the GoVersion.AtLeast() function in build.go that was causing incorrect version comparisons. The current implementation checks each version component (Major, Minor, Patch) independently without respecting the version hierarchy, which leads to incorrect results in many common cases.
For example, with the current implementation:
- Version 2.1.0 would be incorrectly considered "not at least" version 1.5.5
- Version 1.18.0 would be incorrectly considered "not at least" version 1.17.5

The fix implements proper semantic versioning comparison by:
1. Checking major version first and returning early if different
2. Only then comparing minor versions if major versions are equal
3. Only comparing patch versions if both major and minor versions are equal
This ensures that higher versions are correctly identified as being "at least" as new as lower versions, regardless of individual component values.

Yes, this was reported in issue #5269 where the incorrect behavior was identified and a fix was proposed.

Checklist
---------

- [x] There's a new file in changelog/unreleased/ that describes the changes for our users (see template).

### DIFF
--- a/build.go
+++ b/build.go
@@ -298,19 +298,21 @@ func (v GoVersion) AtLeast(other GoVersion) bool {
 		return true
 	}
 
+	if v.Major > other.Major {
+		return true
+	}
 	if v.Major < other.Major {
 		return false
 	}
 
+	if v.Minor > other.Minor {
+		return true
+	}
 	if v.Minor < other.Minor {
 		return false
 	}
 
-	if v.Patch < other.Patch {
-		return false
-	}
-
-	return true
+	return v.Patch >= other.Patch
 }
 
 func (v GoVersion) String() string {


MichaelEischer (approved):
LGTM. It's always amazing how many bugs can still be hidden in ages old code :-) .

Issue #5291, opened 2025-03-21
Attack on content-defined chunking algorithm used by restic


git70:
Hi folks!

https://news.ycombinator.com/item?id=43438601

Can someone professionally explain to an amateur if this "attack" is an attack at all?

[Edit by @MichaelEischer] Link to the paper: https://www.daemonology.net/blog/chunking-attacks.pdf [/Edit]

cperciva:
Co-author of that paper here.  These are absolutely real attacks -- otherwise I wouldn't have put code into Tarsnap to defend against them -- but whether they're attacks you care about is a different matter.

In the case of Restic, knowing 3 chunks of data you're archiving makes it possible to determine the chunking parameters; once you know the chunking parameters, knowing the sequence of chunk sizes allows an attacker to e.g. "fingerprint" known data (e.g. "looks like you backed up a copy of TopSecretDocument.pdf, we were wondering who leaked that to the press").

On the other hand, knowing the chunking parameters doesn't do much to help an attacker reveal *unknown* data (there are theoretical ways it can be done but they're somewhat contrived and involve finding ways to mix chosen plaintext with the secrets you're trying to leak).  So this is mainly a data-possession oracle rather than a leaking-secrets attack.

Does that answer your question?

konidev20:
@git70 

The restic Threat Model section might provide you some more insight. [Threat Model](https://restic.readthedocs.io/en/latest/100_references.html#threat-model)

Correct me if I am wrong, from my understanding, the attacker would need malicious access to the storage server or network to fingerprint chunks. 

Restic's threat model already assumes that the storage server you back up to is trusted. This can be posted on the restic forum for much more detailed discussions. I might be wrong.

dionorgua:
@konidev20 No.. `restic` assumes that host where backup is created is trusted. But not storage.

rwv37:
Am I correct in thinking (as relates to this) that it would be a good idea (for a user) to give name a repository's main directory (or at least any online copy of it) with some sort of name that doesn't really help narrow down the type of content that the repository might contain?

cperciva:
That's the simplest attack. Note that in this case SnowdenTopSecret.pdf might be StarWarsEpisodeIV.mp4 and the attacker might be the MPAA.

There are also other possible attacks, e.g. if the attacker knows *most* of a file you're backing up -- say, if you back up a copy of your genome sequence, since you're (presumably) human and share the vast majority of your DNA with other humans; they could potentially determine which version of a gene you have.

[...]

PR #5295, opened 2025-03-23
Randomize blob to pack file assignment

MichaelEischer:
Implement mitigation described in https://github.com/restic/restic/issues/5291#issuecomment-2746146193

> However, there is a simpler option to push 2^21/k below 1 (and thereby break the attack). An attacker needs to be able to guess which chunks in a pack file belong to the attacker-known file and which sections of the file they correspond to. Restic currently sequentially assembles all fully processed chunks into a single pack file, which gets uploaded once it's full. To prevent an attacker to guess which chunks of a file are in a given pack file, restic can instead assemble two pack files in parallel and randomly assign chunks to those pack files. This should introduce uncertainty at least on the scale of a chunk size and thereby stop the information leak.
> 
> There's one corner case that has to be considered here. A user could backup only a small file that get's split into two chunks. If those end up in individual packs, this would again leak the chunk size. Thus, if two small pack files remain in the end, then those must be merged.

See https://github.com/restic/restic/issues/5291#issuecomment-2746146193

Checklist
---------


changelog/unreleased/issue-5291
```diff
@@ -0,0 +1,23 @@
...
+to check whether certain large files are stored in a repository.
+
+A practical attack is nevertheless hard as restic merges multiple chunks into
+opaque pack files and by default processes multiple files in parallel. This
+likely prevents an attacker from matching pack files to the attacker-known file
+and thereby prevents the attack.
+
+We have added a mitigation to restic, which randomizes how chunks are assembled
+into pack files. This prevents an attacker from guessing which chunks are part
+of a pack file and thereby prevents learning the chunk sizes.
```
rawtaz, line 20:
```suggestion
Despite the low chances of a practical attack, restic now has added mitigation
that randomizes how chunks are assembled into pack files. This prevents attackers
from guessing which chunks are part of a pack file and thereby prevents learning
the chunk sizes.
```

rawtaz (dismissed):
Slight documentation change, other than that I have no comments 🍦

MichaelEischer (commented):
LGTM

PR #5302, opened 2025-03-23
update all direct dependencies

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
Update all direct dependencies. Nothing particularly noteworthy has changed.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
No.

Checklist
---------

- [ ] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I'm done! This pull request is ready for review.

PR #5304, opened 2025-03-24
backend/gs: disable GRPC API to reduce binary size bloat

MichaelEischer:
Since cloud.google.com/go/storage v1.44.0 the GRPC API is enabled by default. However, this causes the restic binary size to explode by 20MB. So just disable it again.

Follow up to https://github.com/restic/restic/pull/5302

Checklist
---------

### DIFF
--- a/build.go
+++ b/build.go
@@ -53,12 +53,14 @@ import (
 
 // config contains the configuration for the program to build.
 var config = Config{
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
LGTM.

PR #5311, opened 2025-03-25
list integration test: error scanning 'restic list blobs'

wplapper:
func parseIDsFromReader does now scans the output from 'restic list blobs' correctly. 'list blobs' produces the blob type as well as the blob ID.

The test 'TestListBlobs' verifies the modified functionality. Since 'testListSnapshots' is used universally, there is no extra new test for the existing functionality.

This PR fixes a problem when using `restic list blobs` in the test environment. The output of `restic list blobs` contains currently 2 word, the blob type and the blob ID itself.

No.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_list_integration_test.go
+++ b/cmd/restic/cmd_list_integration_test.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/restic/restic/internal/repository"
 	"github.com/restic/restic/internal/restic"
 	rtest "github.com/restic/restic/internal/test"
 	"github.com/restic/restic/internal/ui"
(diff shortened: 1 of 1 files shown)


cmd/restic/cmd_list_integration_test.go
```diff
@@ -42,3 +57,34 @@ func testListSnapshots(t testing.TB, opts GlobalOptions, expected int) restic.ID
...
+	rtest.OK(t, err)
+	defer unlock()
+
+	setFromIndex := restic.IDSet{}
+	// make sure the index is loaded
+	rtest.OK(t, repo.LoadIndex(context.TODO(), nil))
+	// get first tree blob
+	rtest.OK(t, repo.ListBlobs(context.TODO(), func(blob restic.PackedBlob) {
+		setFromIndex.Insert(blob.ID)
+	}))
```
MichaelEischer: Please split this into a helper function like the following.
```
func testListBlobs(t testing.TB, gopts GlobalOptions) restic.IDSet
```

A function should stick to functions from the same abstraction level. The `testRunBackup` etc. functions are rather high-level, whereas `repo.LoadIndex` etc. is much more low level. Splitting these makes it easier to reuse code in the future. (And yes I'm aware that the tests are massively violating that rule of thumb, unfortunately).

wplapper:
done.

MichaelEischer (approved):
LGTM.

I've fixed the repository usage to still hold the lock while listing the blobs

PR #5314, opened 2025-03-25
backend/sftp: Add sftp.server-alive-interval and sftp.server-alive-count-max options

### LINKED ISSUE #5313: Add SFTP keepalive options with defaults
Output of `restic version`
--------------------------
`restic 0.17.3 compiled with go1.23.3 on linux/amd64`

As noted in [this forum thread](https://forum.restic.net/t/robustness-of-repository-issue-resolved/1475) and confirmed in my usage, setting the `ServerAliveInterval` and `ServerAliveCountMax` options of sftp can make the difference for a successful backup. I would like to set these by default as command-line arguments to `ssh`, as well as add options so they can be configured otherwise if necessary.

See above. I'm working on making this change and hopefully can submit a PR shortly.

Restic is the most sysadmin-friendly backup program I've used yet. I like software that feels like it was thoughtfully designed.

merlinz01:
This PR adds two options for the SFTP backend to control the underlying keepalive behavior: `sftp.server-alive-interval` and `sftp.server-alive-count-max`. 

Restic previously did not set the `ServerAliveInterval` and `ServerAliveCountMax` SSH options when connecting to an SFTP server, resulting in backups being interrupted on some connections. It now sets these options to the values specified by the `sftp.server-alive-interval` and `sftp.server-alive-count-max` configuration options, to avoid connections being interrupted unnecessarily. The default values are 25 seconds for `ServerAliveInterval` and 200 for `ServerAliveCountMax`. If either is set to -1, the corresponding SSH option is not set and the value from the SSH configuration is used.

Closes #5313
https://forum.restic.net/t/robustness-of-repository-issue-resolved/1475

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5313
+++ b/changelog/unreleased/issue-5313
@@ -1,0 +1,14 @@
+Enhancement: Add `sftp.server-alive-interval` and `sftp.server-alive-count-max` options
+
+Restic previously did not set the `ServerAliveInterval` and `ServerAliveCountMax`
+SSH options when connecting to an SFTP server, resulting in backups being
+interrupted on some connections. It now sets these options to the values
+specified by the `sftp.server-alive-interval` and `sftp.server-alive-count-max`
+configuration options, to avoid connections being interrupted unnecessarily.
+The default values are 25 seconds for `ServerAliveInterval` and
+200 for `ServerAliveCountMax`. If either is set to -1,
+the corresponding SSH option is not set and the value from the SSH
+configuration is used.
+
+https://github.com/restic/restic/issues/5313
+https://forum.restic.net/t/robustness-of-repository-issue-resolved/1475
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -178,15 +178,13 @@ setting the arguments passed to the default SSH command (ignored when
 
 .. note:: Please be aware that SFTP servers close connections when no data is
           received by the client. This can happen when restic is processing huge
-          amounts of unchanged data. To avoid this issue add the following lines 
-          to the client's .ssh/config file:
-
+          amounts of unchanged data. Restic sets the `ServerAliveInterval` and
+          `ServerAliveCountMax` options for `ssh` to keep the connection alive. If you
+          experience connection issues, you can adjust these settings with Restic's
+          `sftp.server-alive-interval` and `sftp.server-alive-count-max` options.
 ::
 
-    ServerAliveInterval 60
-    ServerAliveCountMax 240
-          
-          
+
 REST Server
 ***********
 
--- a/internal/backend/sftp/config.go
+++ b/internal/backend/sftp/config.go
@@ -17,13 +17,21 @@ type Config struct {
 	Args    string `option:"args"    help:"specify arguments for ssh"`
 
 	Connections uint `option:"connections" help:"set a limit for the number of concurrent connections (default: 5)"`
+
+	// SSH options
+	ServerAliveInterval int `option:"server-alive-interval" help:"set the interval to send keepalive messages, or -1 to not set (default: 25)"`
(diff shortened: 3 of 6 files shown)


doc/030_preparing_a_new_repo.rst
```diff
@@ -178,15 +178,13 @@ setting the arguments passed to the default SSH command (ignored when
 
 .. note:: Please be aware that SFTP servers close connections when no data is
           received by the client. This can happen when restic is processing huge
-          amounts of unchanged data. To avoid this issue add the following lines 
-          to the client's .ssh/config file:
-
+          amounts of unchanged data. Restic sets the `ServerAliveInterval` and
+          `ServerAliveCountMax` options for `ssh` to keep the connection alive. If you
+          experience connection issues, you can adjust these settings with Restic's
+          `sftp.server-alive-interval` and `sftp.server-alive-count-max` options.
```
MichaelEischer, line 184: Give a full example how to overwrite those options. It's not immediately obvious for all users how to set those options.

internal/backend/sftp/config.go
```diff
@@ -17,13 +17,21 @@ type Config struct {
...
+
+	// SSH options
+	ServerAliveInterval int `option:"server-alive-interval" help:"set the interval to send keepalive messages, or -1 to not set (default: 25)"`
+	ServerAliveCountMax int `option:"server-alive-count-max" help:"set the number of unacknowledged keepalive messages allowed before disconnecting, or -1 to not set (default: 200)"`
+}
+
+var defaultConfig = Config{
+	Connections:         5,
+	ServerAliveInterval: 25,
+	ServerAliveCountMax: 200,
```
MichaelEischer, line 29: Those default values would take `25*200 = 5000 seconds` until a connection is terminated. That's far too long for a default. In fact, I'd prefer to not set default values for both options.

doc/030_preparing_a_new_repo.rst
```diff
@@ -178,15 +178,13 @@ setting the arguments passed to the default SSH command (ignored when
...
 ::
 
-    ServerAliveInterval 60
-    ServerAliveCountMax 240
```
MichaelEischer, line 187: We should fix that example to use a lower `ServerAliveCountMax` value. Waiting 4 hours until a broken connection is closed is a bit much.

internal/backend/sftp/config_test.go
```diff
@@ -9,69 +9,129 @@ import (
 var configTests = []test.ConfigTestData[Config]{
 	// first form, user specified sftp://user@host/dir
 	{
-		S:   "sftp://user@host/dir/subdir",
-		Cfg: Config{User: "user", Host: "host", Path: "dir/subdir", Connections: 5},
+		S: "sftp://user@host/dir/subdir",
+		Cfg: Config{
+			User: "user", Host: "host", Path: "dir/subdir",
+			Connections:         defaultConfig.Connections,
```
MichaelEischer, line 15: Instead of applying the defaulting here, use a loop in `TestParseConfig` to do so.

changelog/unreleased/issue-5313
```diff
@@ -0,0 +1,14 @@
...
+SSH options when connecting to an SFTP server, resulting in backups being
+interrupted on some connections. It now sets these options to the values
+specified by the `sftp.server-alive-interval` and `sftp.server-alive-count-max`
+configuration options, to avoid connections being interrupted unnecessarily.
+The default values are 25 seconds for `ServerAliveInterval` and
+200 for `ServerAliveCountMax`. If either is set to -1,
+the corresponding SSH option is not set and the value from the SSH
+configuration is used.
+
+https://github.com/restic/restic/issues/5313
```
MichaelEischer, line 13: Also add a link to this PR

MichaelEischer (changes requested):
Sorry for the late review, see below for comments.

PR #5317, opened 2025-03-27
docs: fix unit for S3 restore timeout

gilbsgilbs:

"d" is not a valid unit for duration. Sorry.

No.

Checklist
---------

  - not relevant
  - no changelog entry needed IMO

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -253,7 +253,7 @@ Archive** storage classes is available:
 .. code-block:: console
 
    $ restic backup -o s3.storage-class=GLACIER somedir/
-   $ RESTIC_FEATURES=s3-restore restic restore -o s3.enable-restore=1 -o s3.restore-days=7 -o s3.restore-timeout=1d latest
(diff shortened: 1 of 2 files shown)


MichaelEischer (approved):
We could also use `internal/restic/duration.go` (requires hooking it up in `internal/options/options.go`) if that makes more sense?

gilbsgilbs:
I'm not too sure. It may improve UI consistency, but while I know for sure what "waiting `24h`" means, waiting `1d`, `1m(onth)`, `1y` is ambiguous in that case. Does that account for DST, leap years, leap seconds? Or do we use 1 day = 24h, 1 month = 31 days, and 1 year = 365 days? Also, are we sure there's no use-case for arbitrarily precise or aggressively low timeout values (below the hour)? Like setting the timeout to `1ns` as a hack to fail-fast and not hang. Bulk retrievals on Glacier [last up to 48 hours](https://repost.aws/knowledge-center/restore-glacier-tiers), I don't see me using anything above `48-72h` which looks reasonably readable and unambiguous to me.

MichaelEischer:
Fine with me.

PR #5321, opened 2025-03-29
Add support for additional compression levels `fastest` and `better`

### LINKED ISSUE #4728: Expose additional compression levels
Output of `restic version`
--------------------------
restic 0.16.4 compiled with go1.21.6 on linux/amd64

It would be great if restic exposed the [additional compression levels](https://pkg.go.dev/github.com/klauspost/compress/zstd#EncoderLevel) offered by the [klauspost/compress](https://github.com/klauspost/compress) library.  

The current 'auto' and 'max' levels map to the `SpeedDefault` and `SpeedBestCompression` levels, respectively, in the klauspost/compress library, and while these are great to have, the 'max' setting occupies a fairly extreme point on the compression:speed tradeoff curve.  It's useful for remote repos when upstream bandwidth is limited (<10Mbps), but can significantly slow down transfers over faster connections.  The klauspost/compress library offers an intermediate mode (`SpeedBetterCompression`) that might be more useful to people who want a bit more compression than `SpeedDefault` but who don't want to go all the way to the orders-of-magnitude-slower `SpeedBestCompression`.   Additionally, `SpeedFastest` might be a desirable configuration for people who are doing local disk backups and care most about speed but still want some compression.

I've benchmarked the various zstd compression levels (1-22) on my (dinky) [Celeron N5095](https://www.cpu-monkey.com/en/cpu-intel_celeron_n5095) and noticed:
* Level 1 (`SpeedFastest`) runs at around 500-2000 MB/s.
* Level 3 (`SpeedDefault` / 'auto') runs at around 300-1500 MB/s and achieves approximately the same compression ratios as Level 1.
* Level 7-8 (`SpeedBetterCompression`) runs at around 50-100 MB/s and achieves 10-30% better compression ratios than Level 1.
* Level 19 (I think this is equivalent to `SpeedBestCompression` / 'max') runs at around 0.8-3 MB/s and achieves 20-70% better compression ratios on compressible files than Level 1. 

These are all pretty significant differences, especially with respect to speed, and since people use restic with a variety of backends on a variety of hardware, it seems like more control over the compression levels would be useful. 

I tested the two extra modes ("better" and "fastest") inside restic by adding the appropriate enums and switch cases in internal/repository/repository.go and they seem to produce backup speeds and repo sizes proportional to what would be expected given my zstd benchmarks above. 

Would exposing these additional compression levels be useful to the restic community?  Does doing so jive with the general philosophy of the project?  I'm sure this was considered at some point when compression was initially added, but I had trouble finding relevant discussions.

This is the first time I've touched golang or restic code.  If there's interest, I'm happy to take a stab at a proper PR, but I wanted to get a sense for whether this was worth pursuing first.  I imagine this is something an experienced restic contributor could knock out in 20 minutes (which is why I think there might be philosophical issues with the proposed changes or something else I'm not considering). 

I'm pretty new to restic, but I'm currently pretty pumped about it and I especially like using it with resticprofile to automate and organize all of my backups.  It also did my taxes for me today and I even saw it help an old lady cross the street.

konidev20:
Added support for `fastest` and `better` compression levels exposed by klaupost/compress library.

Closes #4728 

Checklist
---------

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -73,7 +73,9 @@ const (
 	CompressionAuto    CompressionMode = 0
 	CompressionOff     CompressionMode = 1
 	CompressionMax     CompressionMode = 2
-	CompressionInvalid CompressionMode = 3
+	CompressionFastest CompressionMode = 3
+	CompressionBetter  CompressionMode = 4
+	CompressionInvalid CompressionMode = 5
 )
 
 // Set implements the method needed for pflag command flag parsing.
@@ -85,9 +87,13 @@ func (c *CompressionMode) Set(s string) error {
 		*c = CompressionOff
 	case "max":
 		*c = CompressionMax
+	case "fastest":
+		*c = CompressionFastest
+	case "better":
+		*c = CompressionBetter
 	default:
 		*c = CompressionInvalid
-		return fmt.Errorf("invalid compression mode %q, must be one of (auto|off|max)", s)
+		return fmt.Errorf("invalid compression mode %q, must be one of (auto|off|max|fastest|better)", s)
 	}
 
 	return nil
@@ -101,6 +107,10 @@ func (c *CompressionMode) String() string {
 		return "off"
 	case CompressionMax:
 		return "max"
+	case CompressionFastest:
+		return "fastest"
+	case CompressionBetter:
+		return "better"
 	default:
 		return "invalid"
 	}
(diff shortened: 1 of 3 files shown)


internal/repository/repository.go
```diff
@@ -85,9 +87,13 @@ func (c *CompressionMode) Set(s string) error {
...
 	case "max":
 		*c = CompressionMax
+	case "fastest":
+		*c = CompressionFastest
+	case "better":
+		*c = CompressionBetter
 	default:
 		*c = CompressionInvalid
-		return fmt.Errorf("invalid compression mode %q, must be one of (auto|off|max)", s)
+		return fmt.Errorf("invalid compression mode %q, must be one of (auto|off|max|fastest|better)", s)
```
MichaelEischer, line 96: Please sort the levels from off to max (also in cmd/restic/global.go)

MichaelEischer (changes requested):
Just one small nit

internal/repository/repository.go
konidev20: I had the same thought. This is done.

MichaelEischer (approved):
LGTM.

Issue #5254, opened 2025-02-21
restic exit code incoherency on forget operation


deajan:
Output of `restic version`
--------------------------
restic 0.17.3 compiled with go1.23.3 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
S3 Minio server

Problem description / Steps to reproduce
----------------------------------------

When using forget command on a S3 repository, failures (because of S3 policies that don't allow deletions) will still make restic exit with exit code 0, rendering scripting difficult.

Example:
```
# /usr/local/bin/restic forget 09866f87
repository abcdedf opened (version 2, compression level auto)
found 3 old cache directories in /root/.cache/restic, run `restic cache --cleanup` to remove them
Remove(<snapshot/09866f8748>) failed: client.RemoveObject: Access Denied.
unable to remove snapshot/09866f8748d3c5fd70bc7bfa2c846253c6b0c76acddfab4559e27820be71b693 from the repository
[0:00] 100.00%  1 / 1 files deleted
# echo $?
0
```

Expected behavior
-----------------

Guess the exit code should not be 0.

Actual behavior
---------------

Looks like restic still thinks that the transaction with the S3 server succeed, even if said otherwise.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Still using restic alot with about 20TB of data and hundreds / thousands of snapshots. Best decision I took to give restic a test drive ^^ 

Btw, I do have a file where I record restic CLI inconsistencies I encounter while developping around restic. Perhaps you could have a look at https://github.com/netinvent/npbackup/blob/main/RESTIC_CMDLINE_CONSISTENCY.md

Thank you.

konidev20:
This is a duplicate of #5233

deajan:
> This is a duplicate of [#5233](https://github.com/restic/restic/issues/5233)

Indeed. Still, I have posted a file containing all the restic CLI incoherences I've found in a year. Maybe this could help.

MichaelEischer:
> Perhaps you could have a look at https://github.com/netinvent/npbackup/blob/main/RESTIC_CMDLINE_CONSISTENCY.md

Not all commands support JSON so far. Other than someone implementing JSON support for those commands, we can let restic err with exit code 1 if `--json` isn't supported if that's more useful. Commands with JSON support and their output format are listed at https://restic.readthedocs.io/en/stable/075_scripting.html#json-output . There's https://github.com/restic/restic/issues/5247 by now.

[...]

PR #5322, opened 2025-03-30
forget: return exit code 3 on partial removal of snapshots

### LINKED ISSUE #5233: `forget` currently exits with an exit code 0 even if some snapshots couldn't be removed
Output of `restic version`
--------------------------
restic 0.17.3

restic must define an appropriate exit code when it is not able to remove all the snapshots when using `forget` command along with the `--prune` flag

`forget` currently exits with an exit code 0 even if some snapshots couldn't be removed. 

This has been discussed in #5140 

Quoting @MichaelEischer:

> forget currently does exit with exit code 0 even if some snapshots couldn't be removed (see [restic/cmd/restic/cmd_forget.go](https://github.com/restic/restic/blob/536ebefff44a3278fda42550137ac0f15df8d2fe/cmd/restic/cmd_forget.go#L313)
> That is definitely a problem when using forget --prune and somewhat questionable when called without that option.

konidev20:
Before this change, `forget` command returns exit code of 0 even if it's not able to remove all the selected snapshots, when using it with the `--prune` flag.

With this change, the `forget` command will return an exit code of 3 if it's not able to remove all the selected snapshots.

Closes #5233 

Also, discussed in #5140, #5254 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5233
+++ b/changelog/unreleased/issue-5233
@@ -1,0 +1,8 @@
+Bugfix: forget command returns exit code 3 on partial removal of snapshots
+
+The `forget` command now returns exit code 3 when it fails to remove one or
+more snapshots. Previously, it returned exit code 0, which could lead to
+confusion if the command was used in a script.
+
+https://github.com/restic/restic/issues/5233
+https://github.com/restic/restic/pull/5322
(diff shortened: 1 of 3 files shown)


MichaelEischer (approved):
LGTM. Thanks!

PR #5331, opened 2025-03-31
docu: Add possible pack file padding to repository design

### LINKED ISSUE #5328: Harden / replace CDC algorithm
The immediate [attack on the CDC algorithm](https://www.daemonology.net/blog/chunking-attacks.pdf) used by restic has been mitigated by https://github.com/restic/restic/issues/5291 . But we need a permanent solution.

This issue is for tracking the necessary work to harden or replace the algorithm.

Most obvious options:
- Increase key size
- Switch to a different CDC. What are other backup solutions using?
- Throw a strong cryptographic cipher into the mix (e.g. encrypt the hash before using it for the cutting decision)

Relevant considerations:
- The CDC is already a massive performance bottleneck. This shouldn't get any worse.
- Should aim for 128+ bit key size
- Ideally both the computational effort and the number of known cutpoints to derive the CDC key is large. Requiring a large number of known cutpoints increases the effectiveness of padding exponentially.

Other aspects:
- Add or at least add the option to additionally pad the size of pack files, for example, using https://lbarman.ch/blog/padme/ . As most pack files are slightly above 16MB, this would remove nearly all signal from the pack size. Adding a single padding blobs per packfile would be the most straightforward and flexible option here.

aawsome:

Enhances the repository format to support padding in order to tackle length guessing attacks, see
#5291.
May close #5328.

Yes, see #5291 or #5328

Checklist
---------

- [N/A] I have added tests for all code changes.
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)). -> Not necessary, may follow with an implementation of the specification

### DIFF
--- a/doc/design.rst
+++ b/doc/design.rst
@@ -267,6 +267,7 @@ of a JSON document like the following:
       "packs": [
         {
           "id": "73d04e6125cf3c28a299cc2f3cca3b78ceac396e4fcf9575e34536b26782413c",
+          "padding": 12354,
           "blobs": [
             {
               "id": "3ec79977ef0cf5de7b08cd12b874cd0f62bbaf7f07f3497a5b1bbcc8cb39b1ce",
@@ -301,6 +302,9 @@ corresponds to ``Length(encrypted_blob)`` in the pack file header.
 Field ``uncompressed_length`` is only present for compressed blobs and
 therefore is never present in version 1 of the repository format. It is
 set to the value of ``Length(blob)``.
+The field ``padding`` is optional and describes the number of bytes used
+as padding in the pack file. If it is not present or ``0``, no padding is
+used.
 
 The field ``supersedes`` lists the storage IDs of index files that have
 been replaced with the current index file. This happens when index files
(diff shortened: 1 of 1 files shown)


doc/design.rst
```diff
@@ -158,7 +158,7 @@ A Pack's structure is as follows:
 
 ::
 
-    EncryptedBlob1 || ... || EncryptedBlobN || EncryptedHeader || Header_Length
+    EncryptedBlob1 || ... || EncryptedBlobN || Random bytes padding || EncryptedHeader || Header_Length
```
MichaelEischer, line 161: This breaks the file size check in `prune` and `check` and therefore isn't an option.

aawsome:
rethinking it, we could also simply add a unused blob with given number of zero bytes. By knowing that this is at the end of a pack, we can even use that information and don't show those "last" blobs as unused, but as padding blobs. Then older versions will (wrongly) show them as unused, but new versions know, they are padding blobs.. Maybe this is what you meant by "add padding blob"? Anyway, I'd prefer to first add it to the specification and then have it implemented ;-)

aawsome:
Closing this as the padding blob (or other chunking strategies) are better solutions.

Issue #5341, opened 2025-04-02
restic 0.18.0 on NetBSD 10.1 throws xattr errors when filesystem doesn't do xattrs....


gregoster:
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on netbsd/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
rest-server

Problem description / Steps to reproduce
----------------------------------------
Every file on the filesystem that is being considered for backup throws an error like:
...
error: incomplete metadata for /dev/rwd2l: xattr.list /dev/rwd2l: operation not supported                          16:11:13 [1418/1891]
error: incomplete metadata for /dev/rwd2m: xattr.list /dev/rwd2m: operation not supported                                              error: incomplete metadata for /dev/rwd2n: xattr.list /dev/rwd2n: operation not supported                                              
error: incomplete metadata for /dev/rwd2o: xattr.list /dev/rwd2o: operation not supported                                              error: incomplete metadata for /dev/rwd2p: xattr.list /dev/rwd2p: operation not supported                                              
error: incomplete metadata for /dev/rwd3a: xattr.list /dev/rwd3a: operation not supported                                              error: incomplete metadata for /dev/rwd3b: xattr.list /dev/rwd3b: operation not supported                                              
...
error: incomplete metadata for /var/yp: xattr.list /var/yp: operation not supported
error: incomplete metadata for /var/yp/Makefile.main: xattr.list /var/yp/Makefile.main: operation not supported
error: incomplete metadata for /var/yp/Makefile.yp: xattr.list /var/yp/Makefile.yp: operation not supported
error: incomplete metadata for /var/yp/binding: xattr.list /var/yp/binding: operation not supported
error: incomplete metadata for /var/yp/nicknames: xattr.list /var/yp/nicknames: operation not supported
...
Basically if the file is on the filesystem, there is an error produced for it.
Version 0.17.3 worked just fine.  Version 0.18.0 spews errors.
I don't think RESTIC_REPOSITORY and RESTIC_PASSWORD variable content are overly relevant here, as those worked fine with 0.17.3 and are consistent with what are used for similar backups with Linux hosts.  The back-ends are running rest-server (I see the issue on multiple NetBSD boxes, with two different rest-server back-ends).

The filesystems on the NetBSD boxes looks like (just one example, the others are similar):
file system: /dev/rraid1a
format  FFSv1
endian  little-endian
[...]

PR #5344, opened 2025-04-03
EOPNOTSUPP can be returned if the filesystem does not support xattrs

gregoster:
EOPNOTSUPP is returned on NetBSD on non-FFSv2ea filesystems, and possibly on other OSes with filesystems which do not support xattrs.  This patch ignores EOPNOTSUPP as an error for xattr.

https://github.com/restic/restic/issues/5341

Checklist
---------

### DIFF
--- a/internal/fs/node_xattr.go
+++ b/internal/fs/node_xattr.go
@@ -53,8 +53,7 @@ func handleXattrErr(err error) error {
 
 	case *xattr.Error:
 		// On Linux, xattr calls on files in an SMB/CIFS mount can return
-		// ENOATTR instead of ENOTSUP.  Ignore "Operation not supported"
-		// as well.
+		// ENOATTR instead of ENOTSUP.  BSD can return EOPNOTSUPP.
 		switch e.Err {
 		case syscall.ENOTSUP, syscall.EOPNOTSUPP, xattr.ENOATTR:
 			return nil
--- a/internal/fs/node_xattr.go
+++ b/internal/fs/node_xattr.go
@@ -54,8 +54,7 @@ func handleXattrErr(err error) error {
 	case *xattr.Error:
 		// On Linux, xattr calls on files in an SMB/CIFS mount can return
 		// ENOATTR instead of ENOTSUP.  BSD can return EOPNOTSUPP.
-		switch e.Err {
-		case syscall.ENOTSUP, syscall.EOPNOTSUPP, xattr.ENOATTR:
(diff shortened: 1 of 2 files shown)


internal/fs/node_xattr.go
```diff
@@ -53,9 +53,10 @@ func handleXattrErr(err error) error {
 
 	case *xattr.Error:
 		// On Linux, xattr calls on files in an SMB/CIFS mount can return
-		// ENOATTR instead of ENOTSUP.
+		// ENOATTR instead of ENOTSUP.  Ignore "Operation not supported"
+		// as well.
```
greatroar: ENOTSUP is "Operation not supported"; EOPNOTSUPP is "Operation not supported on socket". So I think this comment should call out NetBSD for being weird:

```suggestion
		// ENOATTR instead of ENOTSUP.  NetBSD can return EOPNOTSUPP.
```

MichaelEischer: From the Linux errno manpage xD . Although, I agree that NetBSD shows some rather unusual behavior here.

> (ENOTSUP and EOPNOTSUPP have the same value on Linux, but according to POSIX.1 these error values should be distinct.)

MichaelEischer (changes requested):
See comment above

internal/fs/node_xattr.go
gregoster: The xattr package for go:
https://github.com/pkg/xattr/blob/master/xattr_test.go
uses both ENOTSUP and EOPNOTSUPP as well.   FreeBSD (I think) will be the same as NetBSD in terms of returning EOPNOTSUPP.  Not sure about Darwin.

MichaelEischer: EOPNOTSUPP seems indeed to be a BSD thing: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/acl_set.3.html . Although, restic has supported xattrs on darwin, freebsd, linux and solaris for quite a while, and there haven't been any reports regarding that error code. But feel free to replace `NetBSD` with `BSD` in the above suggestion.

MichaelEischer:
@gregoster are you still interested in completing the PR?

gregoster:
> @gregoster are you still interested in completing the PR?

Yes, I'd still like this completed.   I've merged in the suggested wording changes.  Is there anything else required?

MichaelEischer (approved):
LGTM.

I've rebased the PR to get rid of the merge commit.

Issue #5342, opened 2025-04-02
"chmod: operation not supported" errors on WebDAV and SMB


mikix:
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
Local (but behind the scenes, it is using SMB or WebDAV via GVfs FUSE mounts).

Problem description / Steps to reproduce
----------------------------------------
If I use a FUSE mount to make a WebDAV remote locally accessible (I have not yet tried this with a fully remote backend like rclone), I get errors like the following:

```
$ restic init --repo blarg2 --insecure-no-password
Save(<key/eb4be4df8e>) returned error, retrying after 1.438621768s: chmod myrepo/keys/eb4be4df8e6b3e81bba4072b5f2f126c0e1953f200c04ebffce01ecff0372ba6: operation not supported
Save(<key/eb4be4df8e>) returned error, retrying after 1.99187847s: chmod myrepo/keys/eb4be4df8e6b3e81bba4072b5f2f126c0e1953f200c04ebffce01ecff0372ba6: operation not supported
Save(<key/eb4be4df8e>) returned error, retrying after 4.295487722s: chmod myrepo/keys/eb4be4df8e6b3e81bba4072b5f2f126c0e1953f200c04ebffce01ecff0372ba6: operation not supported
```

This is new behavior with 0.18.0 (i.e. it does not happen on 0.17.3).

Things to try still:
- Using rclone backend on these

Expected behavior
-----------------
Not stress about the lack of a functioning chmod on systems that don't really support it.

Actual behavior
---------------

The command takes forever while it retries. Not sure if it actually ends after a certain number or not. I get impatient.

Do you have any idea what may have caused this?
-----------------------------------------------
I'm guessing restic added some error checking, but maybe it can be relaxed somewhat in these cases.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Restic is great. I maintain the backup program [Déjà Dup](https://dejadup.org/) and I have started defaulting to restic instead of our previous backend. It's been going well so far!

mikix:
OK using an rclone backend, this error doesn't happen. So it's only on local FUSE filesystems, I guess. Restic is thinking it should be a fully capable filesystem, but I'm tricking it.

I can look at how to fix this, unless you folks know of the obvious solution.

mikix:
[...]

PR #5345, opened 2025-04-04
backend/local: ignore chmod "not supported" errors

mikix:
This fixes a regression in 0.18.0, where local filesystems that don't support chmod would throw errors during writes.

This change is the lightest touch option. But other thoughts if we wanted to go bigger:
- I'm uncertain whether the `os.IsPermission` check on the same line could be removed. Presumably the change in Go versions with 0.18.0 introduced this bug? And it was being caught with `os.IsPermission` before, in which case we could remove it?
- I see that a very similar ENOTSUP check is being done in the `chmod` wrapper in `file_unix.go`. But it felt like a bigger refactor than it was worth to try to unify the chmod-calling code points.

https://github.com/restic/restic/issues/5342

Checklist
---------

- [ ] I have added tests for all code changes. (I'm a little unsure how best to test this - presumably injecting /mocking a low level error up.)

### DIFF
--- a/changelog/unreleased/issue-5342
+++ b/changelog/unreleased/issue-5342
@@ -1,0 +1,7 @@
+Bugfix: Ignore "chmod not supported" errors when writing files
+
+Restic 0.18.0 introduced a bug that caused "chmod xxx: operation not supported"
+errors to appear when writing to a local file repository that did not support
+chmod (like CIFS or WebDAV mounted via FUSE). Restic now ignores those errors.
+
+https://github.com/restic/restic/issues/5342
--- a/internal/backend/local/local_unix.go
+++ b/internal/backend/local/local_unix.go
@@ -43,5 +43,12 @@ func isMacENOTTY(err error) bool {
 
 // set file to readonly
 func setFileReadonly(f string, mode os.FileMode) error {
-	return os.Chmod(f, mode&^0222)
+	err := os.Chmod(f, mode&^0222)
+
+	// ignore the error if the FS does not support setting this mode (e.g. CIFS with gvfs on Linux)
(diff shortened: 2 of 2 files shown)


internal/backend/local/local.go
```diff
@@ -193,7 +193,7 @@ func (b *Local) Save(_ context.Context, h backend.Handle, rd backend.RewindReade
 	// ignore if the operation fails as some filesystems don't allow the chmod call
 	// e.g. exfat and network file systems with certain mount options
 	err = setFileReadonly(finalname, b.Modes.File)
-	if err != nil && !os.IsPermission(err) {
+	if err != nil && !os.IsPermission(err) && !errors.Is(err, errors.ErrUnsupported) {
```
MichaelEischer: I'd prefer to let `setFileReadonly` perform the same `isNotSupported` check as `internal/fs/file_unix.go`. (Just copy&paste it)

MichaelEischer (changes requested):
The backends previously were using some of the functions from `fs`. I've switched that to directly using the `os` functions in restic 0.18. to decouple the fs (intended for use by backup/restore) and the backend code. But apparently some of the hacks there are actually required for the backends. Let's just copy the `setFileReadonly` function to the `local` backend.

Regarding testing it don't see a good way to do so. And mocking everything is more work than it's worth.

[Edit]For reference: this was regressed by https://github.com/restic/restic/pull/5024/commits/24f4e780f1019b49fd9f0c6644d3925a1e754056 . Go isn't to blame here.[/Edit]

internal/backend/local/local.go
mikix: Done. :+1:

MichaelEischer (approved):
LGTM

PR #5347, opened 2025-04-05
backup: return exit code 3 if not all targets are available

### LINKED ISSUE #4467: Exit code 0 instead of 3 when directory given by --files-from is not readable
Output of `restic version`
--------------------------
restic 0.16.0 compiled with go1.20.6 on linux/amd64

I used 0.14 on the same infrastructure before which had the same problem.

The command was used to reproduce the problem:

/usr/bin/restic -r rclone:storagebox:Backups/BigNas -p /home/jb/Documents/restic-storagebox-bignas-pwd.txt backup --files-from /home/jb/Documents/test-dir-list.txt

rclone

Expected behavior
-----------------
echo $? returns 3

Reason:

The text file test-dir-list.txt (parameter for --files-from listing the dirs to backup) contains the following line:

/mnt/Daten/Backups/Gogs-not-there

This dir does not exist. This is done to simulate a non reachable NFS share.

https://restic.readthedocs.io/en/latest/040_backup.html#exit-status-codes states:

Restic returns one of the following exit status codes after the backup command is run:

0 when the backup was successful (snapshot with all source files created)
1 when there was a fatal error (no snapshot created)
3 when some source files could not be read (incomplete snapshot with remaining files created)

Program output: 

repository e8bb90c9 opened (version 2, compression level auto)
using parent snapshot 58fa48ca

Files:           0 new,     0 changed,  2270 unmodified
Dirs:            0 new,     0 changed,    16 unmodified

processed 2270 files, 711.119 MiB in 0:03
snapshot 6a5fa242 saved

Actual behavior: 
echo $? returns 0

Expected behavior: 
echo $? returns 3

I hope I did not understand anything wrong and waste your time with this bug report. If I am correct though this bug is critical IMHO since it allows incomplete backups without notice. 

The actual backup in my case is done by a golang progam I wrote. The important lines are:

cmd := exec.Command(program, repo, repoParam, pwd, pwdParam, command, files, filesParam, tls, cacert, cacertParam)
if err := cmd.Run(); err != nil {
	return fmt.Sprintf("Error creating restic snapshot into repo:%s: %s\n", repoName, err.Error())
}

This did not catch a NFS share disconnect.

Use restic with --files-from that contains a non reachable dir.

No

No

Yes, it helps me every day. I am glad restic exists and I am impressed with the work the restic team is doing over such a long period of time.

konidev20:
Restic `backup` command will also exit with code 3 if any one of the targets selected for backup are inaccessible. Previously, the `backup` command would exit with code 3 only if a child directory or file was inaccessible.

I have modified the `collectTargets` function in the `cmd_backup.go` command to check if all the targets collected from command line args, `--files-from`, `--files-from-verbatim` and `--files-from-raw` are accessible. If any of the targets are inaccessible then, we return the `ErrInvalidSourceData` towards the end; this will result in a exit code `3` (partial snapshot)

Closes #4467 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4467
+++ b/changelog/unreleased/issue-4467
@@ -1,6 +1,6 @@
 Bugfix: Exit with code 3 when some `backup` source files do not exist
 
-Restic used to exit with code 0 even when some targets did not exist. Restic 
+Restic used to exit with code 0 even when some backup sources did not exist. Restic
 would exit with code 3 only when child directories or files did not exist. This
 could cause confusion and unexpected behavior in scripts that relied on the exit
 code to determine if the backup was successful.
--- a/changelog/unreleased/issue-4467
+++ b/changelog/unreleased/issue-4467
@@ -1,0 +1,11 @@
+Bugfix: Exit with code 3 when some targets do not exist
+
+Restic used to exit with code 0 even when some targets did not exist. Restic 
+would exit with code 3 only when child directories or files did not exist. This
+could cause confusion and unexpected behavior in scripts that relied on the exit
+code to determine if the backup was successful.
+
+Restic now exits with code 3 when some targets do not exist.
+
+https://github.com/restic/restic/issues/4467
+https://github.com/restic/restic/pull/5347
--- a/changelog/unreleased/issue-4467
+++ b/changelog/unreleased/issue-4467
@@ -5,7 +5,7 @@ would exit with code 3 only when child directories or files did not exist. This
 could cause confusion and unexpected behavior in scripts that relied on the exit
 code to determine if the backup was successful.
 
-Restic now exits with code 3 when some targets do not exist.
+Restic now exits with code 3 when some backup sources do not exist.
 
 https://github.com/restic/restic/issues/4467
 https://github.com/restic/restic/pull/5347
--- a/changelog/unreleased/issue-4467
+++ b/changelog/unreleased/issue-4467
@@ -1,4 +1,4 @@
-Bugfix: Exit with code 3 when some targets do not exist
+Bugfix: Exit with code 3 when some `backup` source files do not exist
 
 Restic used to exit with code 0 even when some targets did not exist. Restic 
 would exit with code 3 only when child directories or files did not exist. This
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -157,6 +157,9 @@ var backupFSTestHook func(fs fs.FS) fs.FS
 // ErrInvalidSourceData is used to report an incomplete backup
 var ErrInvalidSourceData = errors.New("at least one source file could not be read")
 
+// ErrNoSourceData is used to report that no source data was found
+var ErrNoSourceData = errors.Fatal("all source directories/files do not exist")
+
 // filterExisting returns a slice of all existing items, or an error if no
 // items exist at all.
 func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -437,9 +437,17 @@ func collectTargets(opts BackupOptions, args []string, warnf func(msg string, ar
 		return nil, errors.Fatal("nothing to backup, please specify source files/dirs")
 	}
 
+	countTargets := len(targets)
+
 	targets, err = filterExisting(targets, warnf)
 	if err != nil {
 		return nil, err
+	}
+
+	countExisting := len(targets)
+
+	if countExisting < countTargets {
+		return nil, ErrInvalidSourceData
 	}
 
 	return targets, nil
@@ -496,9 +504,14 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 		return err
 	}
 
+	success := true
 	targets, err := collectTargets(opts, args, printer.E, term.InputRaw())
 	if err != nil {
-		return err
+		if errors.Is(err, ErrInvalidSourceData) {
+			success = false
+		} else {
+			return err
+		}
 	}
 
 	timeStamp := time.Now()
@@ -632,7 +645,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 	arch.SelectByName = selectByNameFilter
 	arch.Select = selectFilter
 	arch.WithAtime = opts.WithAtime
-	success := true
+
 	arch.Error = func(item string, err error) error {
 		success = false
 		reterr := progressReporter.Error(item, err)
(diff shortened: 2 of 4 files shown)


konidev20:
Hi @MichaelEischer, bubbling this PR up since I see you are reviewing some PRs.

cmd/restic/cmd_backup.go
```diff
@@ -438,11 +438,19 @@ func collectTargets(opts BackupOptions, args []string, warnf func(msg string, ar
 		return nil, errors.Fatal("nothing to backup, please specify source files/dirs")
 	}
 
+	countTargets := len(targets)
+
 	targets, err = filterExisting(targets, warnf)
```
MichaelEischer: Please move those checks into `filterExisting`, which already checks that the result array is not empty. So the new check also belongs there.

changelog/unreleased/issue-4467
```diff
@@ -0,0 +1,11 @@
+Bugfix: Exit with code 3 when some targets do not exist
```
MichaelEischer:
```suggestion
Bugfix: Exit with code 3 when some `backup` source files do not exist
```

The terminology has change to refer to the backup data as `source directories/files` instead of `targets`, which was  rather confusing. The internal naming in the code unfortunately still uses the old naming. Please update that in the changelog

MichaelEischer (changes requested):
Thanks, see below for comments and sorry for yet another merge conflict :see_no_evil:  .

changelog/unreleased/issue-4467
konidev20: Completed and Committed

cmd/restic/cmd_backup.go
konidev20: Completed. Moved the check in to the `filterExisting` function.

changelog/unreleased/issue-4467
```diff
@@ -0,0 +1,11 @@
+Bugfix: Exit with code 3 when some `backup` source files do not exist
+
+Restic used to exit with code 0 even when some targets did not exist. Restic
```
MichaelEischer:
```suggestion
Restic used to exit with code 0 even when some backup sources did not exist. Restic
```

changelog/unreleased/issue-4467
```diff
@@ -0,0 +1,11 @@
+Bugfix: Exit with code 3 when some `backup` source files do not exist
+
+Restic used to exit with code 0 even when some targets did not exist. Restic 
+would exit with code 3 only when child directories or files did not exist. This
+could cause confusion and unexpected behavior in scripts that relied on the exit
+code to determine if the backup was successful.
+
+Restic now exits with code 3 when some targets do not exist.
```
MichaelEischer:
```suggestion
Restic now exits with code 3 when some backup sources do not exist.
```

cmd/restic/cmd_backup.go
```diff
@@ -160,6 +160,8 @@ var ErrInvalidSourceData = errors.New("at least one source file could not be rea
 // filterExisting returns a slice of all existing items, or an error if no
```
MichaelEischer, line 163:
```suggestion
// ErrNoSourceData is used to report that no source data was found
var ErrNoSourceData = errors.Fatal("all source directories/files do not exist")

// filterExisting returns a slice of all existing items, or an error if no
```

cmd/restic/cmd_backup.go
```diff
@@ -160,6 +160,8 @@ var ErrInvalidSourceData = errors.New("at least one source file could not be rea
...
 // items exist at all.
 func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
+	countTargets := len(items)
+
```
MichaelEischer:
```suggestion
```

cmd/restic/cmd_backup.go
```diff
@@ -170,6 +172,12 @@ func filterExisting(items []string, warnf func(msg string, args ...interface{}))
...
 
+	countExisting := len(result)
+
+	if countExisting < countTargets {
+		return nil, ErrInvalidSourceData
+	}
+
 	if len(result) == 0 {
 		return nil, errors.Fatal("all source directories/files do not exist")
 	}
```
MichaelEischer:
```suggestion
        if len(result) == 0 {
                return nil, ErrNoSourceData
        } else if len(result) < len(items) {
                return result, ErrInvalidSourceData
        }
```

cmd/restic/cmd_backup.go
```diff
@@ -160,6 +160,8 @@ var ErrInvalidSourceData = errors.New("at least one source file could not be rea
 // filterExisting returns a slice of all existing items, or an error if no
 // items exist at all.
 func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
```
MichaelEischer, line 165: The callsite of `filterExisting` must be changed to:
```
return filterExisting(targets, warnf)
```

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -271,7 +272,9 @@ func TestBackupNonExistingFile(t *testing.T) {
 
 	opts := BackupOptions{}
 
-	testRunBackup(t, "", dirs, opts, env.gopts)
+	err := testRunBackupAssumeFailure(t, "", dirs, opts, env.gopts)
+	rtest.Assert(t, err != nil, "expected error for non-existing file")
+	rtest.Assert(t, !errors.Is(err, ErrInvalidSourceData), "expected ErrInvalidSourceData; got %v", err)
```
```suggestion
        rtest.Assert(t, errors.Is(err, ErrInvalidSourceData), "expected ErrInvalidSourceData; got %v", err)

        // only non-existing file
        dirs = []string{
                filepath.Join(p, "nonexisting"),
        }
        err = testRunBackupAssumeFailure(t, "", dirs, opts, env.gopts)
        rtest.Assert(t, err != nil, "expected error for non-existing file")
        rtest.Assert(t, errors.Is(err, ErrNoSourceData), "expected ErrNoSourceData; got %v", err)
```

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -271,7 +272,9 @@ func TestBackupNonExistingFile(t *testing.T) {
 
 	opts := BackupOptions{}
 
-	testRunBackup(t, "", dirs, opts, env.gopts)
+	err := testRunBackupAssumeFailure(t, "", dirs, opts, env.gopts)
```
MichaelEischer, line 276:
```suggestion
 // mix of existing and non-existing files
	err := testRunBackupAssumeFailure(t, "", dirs, opts, env.gopts)
```

cmd/restic/cmd_backup_test.go
```diff
@@ -71,6 +71,9 @@ func TestCollectTargets(t *testing.T) {
 	rtest.OK(t, err)
 	sort.Strings(targets)
 	rtest.Equals(t, expect, targets)
+
+	_, err = collectTargets(opts, []string{filepath.Join(dir, "cmdline arg"), filepath.Join(dir, "non-existing-file")}, t.Logf, nil)
```
MichaelEischer, line 75: Check that the valid targets are still returned

MichaelEischer (changes requested):
The PR in its current form is completely broken, see comments and suggestions below.

MichaelEischer (approved):
LGTM.

PR #5348, opened 2025-04-06
mount: prevent mounting over repository

### LINKED ISSUE #5234: restic system freeze
Hello!

Mounting restic repository to same restic folder causing system to be unsable and reboot takes 10 minutes to recover.

```
1. create restic mount
2. mount it to it self
3. try to 'ls' or unmount
4. system will delay, reboot takes a lot time
```

```
restic --insecure-no-password -r  /media/axet/64GB/Archives/ mount /media/axet/64GB/Archives/
repository d8a55f2c opened (version 2, compression level auto)
[0:00] 100.00%  3 / 3 index files loaded
Now serving the repository at /media/axet/64GB/Archives/
Use another terminal or tool to browse the contents of this folder.
When finished, quit with Ctrl-c here or umount the mountpoint.
unable to umount (maybe already umounted or still in use?): exit status 1: fusermount: failed to unmount /media/axet/64GB/Archives: Device or resource busy
```

* restic                                              0.17.3-1

Would be better to have openat implemented and allow self-mount #5230

konidev20:
`mount` command will prevent mount if the mountpoint and the repository path are the same.

I have used Claude Code Opus 4.7 to propose these fixes. AI has been used to generate this PR.

Closes #5234 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -13,6 +14,8 @@ import (
 	"github.com/spf13/pflag"
 	"golang.org/x/sys/unix"
 
+	"github.com/restic/restic/internal/backend/local"
+	"github.com/restic/restic/internal/backend/location"
 	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
@@ -149,6 +152,19 @@ func runMount(ctx context.Context, opts MountOptions, gopts global.Options, args
 		return errors.Fatal("inaccessible mountpoint")
 	}
 
+	// Refuse to mount onto (or under, or over) the local repository directory.
+	// Doing so makes the FUSE server read its own backend files through the
+	// mount it just created, deadlocking the kernel (GH #5234).
+	loc, err := location.Parse(gopts.Backends, gopts.Repo)
+	if err != nil {
+		return err
+	}
+	if loc.Scheme == "local" {
+		if err := checkMountpointOverlap(loc.Config.(*local.Config).Path, mountpoint); err != nil {
+			return err
+		}
+	}
+
 	debug.Log("start mount")
 	defer debug.Log("finish mount")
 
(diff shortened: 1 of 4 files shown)


cmd/restic/cmd_mount.go
```diff
@@ -127,6 +130,31 @@ func runMount(ctx context.Context, opts MountOptions, gopts GlobalOptions, args
 
 	mountpoint := args[0]
 
+	r, err := ReadRepo(gopts)
```
MichaelEischer: Determining the repository location would by now have to live in `internal/global/global.go`.

cmd/restic/cmd_mount.go
```diff
@@ -127,6 +130,31 @@ func runMount(ctx context.Context, opts MountOptions, gopts GlobalOptions, args
...
+		localConfig := loc.Config.(*local.Config)
+		m, err := filepath.Abs(mountpoint)
+		if err != nil {
+			return err
+		}
+		l, err := filepath.Abs(localConfig.Path)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(m, l) {
```
MichaelEischer: This will break when the repo is at folder `repo` and the mountpoint is called `r`.

MichaelEischer (changes requested):
Thanks for giving this a try. I'm not yet convinced this is the right way, at the very least the code to determine the local repository path has to move to `internal/global/global.go` or maybe `internal/backend/location`.

MichaelEischer:
@konidev20 are you still interested in finishing this PR?

konidev20:
Yes @MichaelEischer. This weekend or the next.

konidev20:
Hi @MichaelEischer, 

This is ready for review. On MacOS and Windows, I don't know what the behavior would be. I haven't tested it fully. 

[Edit]
I tested on my MacOS laptop, it seems to be working. 
[/Edit]

jtru (approved):
These changes look good to me and work fine on my Linux box. I consider the feature a welcome safeguard.

Since Windows doesn't implement `mount` presently, I don't think there's a valid case to test at this time.


Both `resolvePath` and `isInside` look generally useful enough to justify having them live in a less `mount`-specific place, I guess? (I am not familiar enough with all of the codebase yet to suggest a location, but maybe someone else has a concrete idea :))

MichaelEischer (approved):
LGTM. I'm somewhat unhappy about having to mess with the local.Config but don't see a good alternative either.

> Both resolvePath and isInside look generally useful enough to justify having them live in a less mount-specific place, I guess?

There's no real place for such helpers. The `fs` package is primarily intended for `backup` and `restore`. And I don't see another use of those functions so far, so let's just keep them here.

PR #5351, opened 2025-04-10
Important PR: Enhancement - restic snapshot filters based on time

wplapper:
This PR introduces time based filtering along the lines of https://github.com/restic/restic/issues/5193.

All commands which use a ``SnapshotFiler`` gain three additional options: ``--lower-time-limit`` and ``--upper-time-limit`` and ``--relative-to``. These options limit the time range the user is interested in. Since #5193 suggests three methods of defining a time reference, one of them is a relative time, something like "two and a half months ago", a reference time is needed. That is what ``--relative-to`` will provide.

A new concept ``DurationTime`` is introduced, which can be one of three things: a ``time.Time``, a ``restic.Duration`` related to a reference Time (``RelativeTo``) or a snapshot ID, including 'latest'. All these will eventually be converted to a ``time.Time``, so the ``matches()`` function can do time based filtering ``--lower-time-limit <= sn.Time <= --upper-time-limit``. 

The main part of the new code is located in ``internal/restic/snapshot_find.go``, which is handling the interface to the command line (Set()), which deals with interpretation of the user's input by using regexp() for the definition of a ``restic.Duration`` and the definition of a date with an optional time. The strings ``now`` and ``latest`` are interpreted in the correct context. 

The methods ``Empty()`` and ``matches()`` are extended so they can deal with time based filters.

If a time based filter is detected which references a snapshot or ``latest``, the function ``setTimeFilters()`` is called to convert these snapshot IDs to their equivalent time values. Finally ``setTimes()`` is called to check the sanity of the filter values and converts a ``restic.Duration`` to its equivalent time.Time, based on the value of ``--relative-to`` option which is either set explicitly or implicitly by using the snapshot ID of ``latest``. All of the above happens in function ``FindAll()``.

The method String() for a ``DurationTime`` reads the state of the ``DurationTime`` element and creates a string representation which matches one of a ``snapshot ID``, a ``restic.Duration`` or a ``time.Time``.

https://github.com/restic/restic/issues/2078 // More options for "restic tag"
https://github.com/restic/restic/issues/5136 // Ability to filter snapshots by date
https://github.com/restic/restic/issues/5193 // Allow forget to relate policy to current time and not just latest snapshot
https://github.com/restic/restic/pull/5209   // Add snapshot filtering --lower-time-limit and --upper-time-limit

Alternative names for time based filters could be ``--filter-before`` and ``--filter-after``; ``rustic`` is using these names.

Checklist
---------

  see separate PR https://github.com/restic/restic/pull/21936 for tests.


wplapper:
Some tests still fail - will try to re-run it again later.

wplapper:
temporary close to trigger rerun of test jobs.

rawtaz:
@wplapper You don't have to write a comment for every action you perform to manage the code on this branch.

E.g. things like pushing commits is common practice, even if you force-push sometimes. And closing and reopening a PR is presumed to be something you do intentionally or unintentionally, but what matters is if you keep it open at the end of the day, since that means you still want it to be active.

My point here is that all these comments are just creating noise, so maybe try to focus on comments that add to the discussion instead – we have faith in that you're taking actions that are for the benefit of the PR, there's no need to prove that :)

wplapper:
> @wplapper You don't have to write a comment for every action you perform to manage the code on this branch.

Understood. I removed all "noisy comments".

wplapper:
I split off cmd/restic/cmd_snapshots_integration_test.go and internal/data/duration_time_test.go to make the review package smaller. The rest will follow in a separate PR.

wplapper:
The tests are in PR https://github.com/restic/restic/pull/21936.

PR #5358, opened 2025-04-14
Fix tty capture when started in background

### LINKED ISSUE #5354: rclone:protondrive cannot be run in background, fails with ENOTTY
Output of `restic version`
--------------------------

```
restic 0.18.0 compiled with go1.24.1 on linux/arm64
```

```
rclone:protondrive
```

**Problem description:** On (at least) arm64 - in my case Raspbian on a Raspberry Pi 4 - attempting to start a background sync to a protondrive-backed restic respository will lead to an ENOTTY error and fail, causing the current TTY to be logged out at the same time.

**Steps to reproduce**

Initialise a repository on a protondrive remote via rclone:

```
restic -r rclone:protondrive:backups/restic init
```

Attempt to backup in background:
```
nohup restic -r rclone:protondrive:backups/restic --verbose backup $HOME/Documents &
```
Output:

```
nohup: ignoring input and redirecting stderr to stdout

logout
Connection to HOSTNAME closed.
```

Contents of nohup.out:
```
Fatal: unable to open repository at rclone:protondrive:backups/restic: error moving process to background: inappropriate ioctl for device
```

Expected behavior
-----------------

Restic should run in the background and dump its output to `nohup.out`, or wherever it's piped to.

Actual behavior
---------------

Restic fails and logs out the current TTY session.

No - it doesn't happen when running the same steps with local media as the backup location, for example. Cause of failure is probably somewhere in the rclone to protondrive or restic to rclone integration.

Restic works great for local backups so far.
-->

MichaelEischer:

The rclone and sftp backends require starting a child process. It is first moved into the foreground and back into the background after the initial startup is done.

However, this behavior is also active if restic itself is started in the background. In this case, restic changing the foreground process may confuse the shell and in case of bash causes it to exit. Thus, disable modification of the controlling process group of the tty if restic is run in the background.

Fixes https://github.com/restic/restic/issues/5354

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5354
+++ b/changelog/unreleased/issue-5354
@@ -1,0 +1,12 @@
+Bugfix: Allow use of rclone/sftp backend when running restic in background
+
+When starting restic in the background, this could result in unexpected behavior
+when using the rclone or sftp backend.
+
+For example running `restic -r rclone:./example --insecure-no-password init &`
+could cause the calling `bash` shell to exit unexpectedly.
+
+This has been fixed.
+
+https://github.com/restic/restic/issues/5354
+https://github.com/restic/restic/pull/5358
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
LGTM.

PR #5359, opened 2025-04-15
PR retired: git prune - prune more aggresively - issue 5293

### LINKED ISSUE #5293: prune: improve automatic handling of small pack files
Output of `restic version`
--------------------------
0.17.3

`prune` should clean up small files more aggressively by default. Currently only pack files smaller than 4% of the default pack size are cleaned up. Setting `--repack-small` increases that threshold to 80%. The low default of 4% was chosen to not accidentally repack the whole repository when users switch from the 4MB default pack size of old restic versions to the new maximum of 128MB. However, this results in every `backup` run effectively leaving a small pack file behind that is not cleaned up by default.

In parallel to https://github.com/restic/restic/pull/5183, @rawtaz and I discussed options to improve the current default behavior. What should work is the following:

Retrieve the size of all pack files. Then let `prune` consider everything smaller than 80% of the 1st percentile (or maybe the 3rd or 5th percentile) as "small". That variant will work independently of the packsize with which the repository was created initially. This will completely solve the use case of getting rid of small files for normal repositories.

Then the only remaining use case is migrating a repository to a higher pack size. There it is up to the users to decide which file sizes to upgrade. The most common scenario will still be upgrading everything to the new pack size (although a lot of users probably won't even do that). This can be handled by the new `--repack-smaller-than` option. `--repack-small` would then be deprecated.

wplapper:
Changes: cmd_prune.go: removed option `--repack-small` and marked it deprecated.

Add changelog note.

Added code to filter out the processing of all packfiles early in the ``prune`` process, so that map can be reused for calculating a ``targetPackSize`` value based on the chosen percentile of the sorted packfiles list.

Added test to validate modified prune behaviour.

This PR addresses the issue https://github.com/restic/restic/issues/5293.

https://github.com/restic/restic/issues/5293

closes #5293

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5293
+++ b/changelog/unreleased/issue-5293
@@ -9,4 +9,4 @@ marked as deprecated.
 
 https://github.com/restic/restic/issues/5293
 https://github.com/restic/restic/pull/5183
-https://github.com/restic/restic/pull/new
+https://github.com/restic/restic/pull/5359
--- a/changelog/unreleased/issue-5293
+++ b/changelog/unreleased/issue-5293
@@ -1,0 +1,12 @@
+Enhancement: prune small packfiles more aggressively
+
+prune reads all packfiles and their sizes, sorts them by ascending size order.
+It picks the first percentile value from the sorted list of packfile sizes and
+applies the scale factor of 80% to it and uses it for its prune decisions.
+
+As a consequence, the option `--repack-small` is not longer needed and has been
+marked as deprecated.
+
+https://github.com/restic/restic/issues/5293
+https://github.com/restic/restic/pull/5183
+https://github.com/restic/restic/pull/new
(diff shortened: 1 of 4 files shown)


wplapper:
This PR is getting out of sync with modern restic. Revert to Draft status.

wplapper:
I have decided to revert this PR to closed status to ease the pressure on the
maintainers of restic. Since nothing has really happened with this PR in the last 360 days, it cannot be too important after all. Once the number of PRs has dropped considerably, I might move this PR back to the queue to be reviewed. In case you consider this PR
important, please contact me via the usual channels. Thank you!

wplapper:
I think there is no need for this PR at all: 

Choose sensible defaults for 
- ``--repack-smaller-than``, something between 50% to 80% of ``repo.PackSize()``
- ``--max-unused``, lower than the current 5%. The sorting of packfile sizes happens anyway in prune.go, where the actual load factor of individual packfiles are taken into account. This determines how much of the repository actually gets pruned.
 
Since both options can be controlled by the user, non default settings can be applied easily. 

At the same time ``--repack-small`` can be dropped as well as ``--repack-smaller-than`` gives the user better control for her selected target packsize if she wants to overwrite the default settings.

wplapper:
This PR will be closed because I messed up git. Continued in PR#21803.

PR #5360, opened 2025-04-16
doc: typo & minor rewording in 'Removing files from snapshots'

makuhama:

fixed typo: that => than
better wording: differentiate => distinguish
fixed misleading adjective: rewritten => untouched (indeed the original snapshot is NOT rewritten)

No. I think the changes are trivial and obvious.

Checklist
---------

- [X ] I'm done! This pull request is ready for review.


rawtaz (approved):
This makes sense. Thank you!

Issue #5135, opened 2024-11-11
.,kmj,


lissom:
Output of `restic version`
--------------------------
restic 0.16.3 compiled with go1.21.6 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
backup --tags-success
Tags that only apply if the exit code is zero.
We want all backups, but we also need to know which ones are complete vs exit code 3. 
I can be very difficult to tell that files are missing until something larger goes wrong down the road.

(Storing if the errors in the snapshot data could work too.)

What are you trying to do? What problem would this solve?
---------------------------------------------------------
Detect good vs bad backups.  We want all of the backup, good or not.  So ignoring the snapshot on failure isn't an ideal solution, though this would be preferable to the current situation.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Restic is great.

MichaelEischer:
An alternative would be to extend the snapshot summary [`SnapshotSummary`](https://restic.readthedocs.io/en/stable/075_scripting.html#snapshots). That information is available for example via `snapshots --json`. I somewhat prefer that variant as can always be enabled and doesn't require yet another option for the backup command.

> (Storing if there are errors in the snapshot data could work too. Something as granular as the files that couldn't be read would be great.)

Storing the backup log would require a whole separate mechanism as logs can be quite big.

lissom:
That would work too. However, unless it's searchable (like a tag) then I would still need to tag it to get the correct snapshot in the usual way.

Gaibhne:
I agree, unless metadata gets made searchable/filterable through the CLI, not adding a regular command line flag for it would mean reading back and manually/via script parsing all existing snapshots' metadata, which does not seem like a reliable/dependable path for something as critical as handling/dealing with incomplete snapshots, which essentially imply lost or potentially lost data.

I would argue that data integrity is the absolute number one focus of `restic` (and backup tools in general), not a nice convenience feature that can be hidden behind user scripts shared on forum posts. As such, I would definitely think it reasonable to extend the backup options.

My suggestion would be an opt-out solution that tags incomplete snapshots. Incomplete snapshots are the exception, and people noticing the tag by surprise in their repo would be a good thing. In that spirit, a `--no-tag-unsuccessful` or `--no-tag-with-errors` or similar would be my ideal way forward.

lissom:
#5362 would be great, even better if it can be searched for like tags are.

rawtaz:
[...]

PR #5362, opened 2025-04-17
Enhanced snapshot summary

jtru:
This PR adds statistics for Archiver errors encountered during the creation of a snapshot into that snapshot's summary. It also reorganizes all Err* values/codes that restic can throw around in a new file in the `restic` package, which should make it easier to understand the codebase and maybe do more fine-grained bookkeeping of errors in the future (which I intend to think about more, if this PR gets merged).

I touched briefly upon this on IRC (planning to record the exit code of restic in these stats), with @MichaelEischer suggesting that keeping track of error counts would make more sense, which I agree with. The inspiration of tracking this stat in snapshot summaries originally came from #5135.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -156,9 +156,6 @@ func (opts *BackupOptions) AddFlags(f *pflag.FlagSet) {
 
 var backupFSTestHook func(fs fs.FS) fs.FS
 
-// ErrInvalidSourceData is used to report an incomplete backup
-var ErrInvalidSourceData = errors.New("at least one source file could not be read")
-
 // filterExisting returns a slice of all existing items, or an error if no
 // items exist at all.
 func filterExisting(items []string) (result []string, err error) {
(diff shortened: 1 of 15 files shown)


jtru:
I agree with @lissom that automatic tagging on certain events/results would be a very useful thing to have (as long as one can opt out, or agrees to opt in to this kind of automatic tagging), but I think there should be something like an agreed-upon tag namespace/tag name prefix to minimize impact on users, and make these tags very likely unambiguous.

I would prefer to discuss that in more detail in another issue, however :)

lissom:
I would like it if we ended this ticket with something searchable.  Versus having to list and iterate through the results looking for a good one.  The problem with tagging after the fact (maybe the prefix is "restic-") is that it requires an exclusive lock.  Which if you have a singe slow backup, or a single backup with network errors, etc anything that causes a backup not to complete, it prevents the post-hoc tagging of other images.

jtru:
This PR's title refers to an internal data structure that is called the "snapshot summary", which I proposed to extend (and to get my feet wet with golang in general and restic in particular, tbh).

I was not aware that post-snapshot-creation tagging requires an exclusive lock, but it makes perfect sense it does - which, I think, is a *great* reason for that auto-tagging mechanism that you propose to exist, I think. Still, getting it implemented should be treated in the course of another PR/ticket ;)

lissom:
The other thing that would be great is if the snapshot summary is searchable (also not this PR, but maybe that informs how this PR is implemented).  Taking the limit of things, it would seem that it's reasonable to assume more and more data gets added to the summary as time goes on (lets assume restic is around forever).

PR #5364, opened 2025-04-19
forget: add `--keep-unique` flag to keep one snapshot per tree

### LINKED ISSUE #5157: Provide ability to remove unchanged snapshots
*Coming from [Could it be possible to remove unchanged snapshots?](https://forum.restic.net/t/could-it-be-possible-to-remove-unchanged-snapshots/8759)...*

Output of `restic version`
--------------------------

* `restic 0.17.0 compiled with go1.23rc2 on linux/amd64`

We now have `--skip-if-unchanged`, which prevents empty snapshots from being created, but we still have older empty snapshots that are there from before we started using it.

Given that there’s currently no option to remove duplicate snapshots (besides telling `forget` which specific snapshots to remove), let's find a way to remove these, perhaps with a `prune` option.

This would significantly remove the number of unnecessary snapshots in the repository that aren't needed.  Besides helping folks that haven't been using  `--skip-if-unchanged` until recently (everyone, because it didn't exist), it will also help folks who didn't discover the option until later, or those who didn't realize they wanted it.

If it weren't for Restic, I wouldn't be able to work on [BackUpScale](https://backupscale.com/) so I'm especially thankful to the maintainers and community for making it possible.

konidev20:
* Modified `forget` command to include `--keep-unique` flag
* Modified `snapshot_policy` to remove snapshot if the tree hasn't changed from the parent snapshot

Closes #5157 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5157
+++ b/changelog/unreleased/issue-5157
@@ -1,0 +1,7 @@
+Enhancement: Added `--keep-unique` flag to `forget` command
+
+Restic `forget` command can now remove duplicate snapshots with the
+`--keep-unique` flag set.
+
+https://github.com/restic/restic/issues/5157
+https://github.com/restic/restic/pull/5364
--- a/cmd/restic/cmd_forget.go
+++ b/cmd/restic/cmd_forget.go
@@ -112,6 +112,7 @@ type ForgetOptions struct {
 	WithinMonthly data.Duration
 	WithinYearly  data.Duration
 	KeepTags      data.TagLists
+	Unique        bool
 
 	UnsafeAllowRemoveAll bool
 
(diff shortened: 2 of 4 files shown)


konidev20:
@MichaelEischer , @rawtaz I have tried my hand to solve for this enhancement request. Please review.

internal/restic/snapshot_policy.go
```diff
@@ -288,6 +298,17 @@ func ApplyPolicy(list Snapshots, p ExpirePolicy) (keep, remove Snapshots, reason
...
+		if p.Unique {
+			if cur.Parent != nil {
+				parent := findParentSnapshot(keep, *cur.Parent)
+				if parent != nil {
+					if parent.Tree == cur.Tree {
+						keepSnap = false
+					}
+				}
+			}
+		}
```
konidev20, line 310: Not sure, if I have to do it before the keep and keep within retention passes or after. 

The more I think about it, I feel it should be before, but I don't know how that would interact with the other passes. 

I am open for thoughts.

CircuitCoder:
Thank you! My restic repo desperately needed a GC, since it was set up way before `--skip-if-unchanged` happened.

Some nitpicks:
1. `findParentSnapshot` is O(n) and runs n times, which is unsuitable for repos with large amount of snapshots (maybe I shouldn't have waited for it to grow into ~1e5 snapshots). Using a hashmap solves this. See: https://github.com/CircuitCoder/restic/commit/7d2f9fe7e843d1ad7e9aa6ed95c9a2f1cb23da08
2. Each other type of `--keep-*` options only adds additional kept snapshots, but `--keep-unique` removes snapshots, which I find a little bit surprising. I personally prefer a different name, maybe `--drop-unchanged`?
  Alternatively, align the behavior of `--keep-unique` better with other `--keep-*` options, so that it additionally keep all snapshots that are oldest owners to their respectively trees.

konidev20:
@CircuitCoder Thank you for checking this PR, will review these comments on the weekend and update it.

PR #5373, opened 2025-04-28
bugfix: have `--{cpu,mem,...}-profile` work even if Restic exits with error code

ilyagr:
Before this, if `restic backup --cpu-profile dir/ backup-dir/` couldn't read some of the input files (e.g. they weren't readable by the user restic was running under), the `cpu.pprof` file it outputs would be empty.

https://github.com/spf13/cobra/issues/1893

No, I was going to file a bug, but it was simpler to illustrate
the problem with a PR. This could be mergeable, or it could just be
an illustration/proof-of-concept.

Checklist
---------

- (n/a?) I have added tests for all code changes.
- (n/a?) I have added documentation for relevant changes (in the manual).
- [x] Follow [the other commits in the repository](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits)


MichaelEischer (approved):
LGTM. Thanks a lot for finding this!

I've dropped the changelog entry as we usually only mention changes relevant for end users and this is a developer only change.

ilyagr:
Thank you for the review!

PR #5375, opened 2025-04-29
Snapshot description

Laborratte5:
This PR adds the ability to add a description text to snapshots.
When creating a new snapshots with `restic backup` the `--description`
and `--description-file` option allows setting a description text.

The description of existing snapshots can be changed using the
`restic description <snapshotID> [--set <desc> | --set-file <file>]` subcommand.
To remove a description `restic description <snapshotID> --remove-description` can be used.

https://github.com/restic/restic/issues/2376

Checklist
---------


Laborratte5:
Note: I intentionally didn't add support for changing the description of multiple snapshots add once
as I think descriptions a somewhat unique per snapshot. This prevents users from accidentally overwriting
the description of many snapshots.

Laborratte5:
Another idea that just came to my mind is the following.
As multi-line descriptions look kind of weird in the snapshot table, maybe some formatting like git commit messages
could be used, where the first line is the heading, displayed in the snapshots table. And after a free line the longer
description would follow, which is only displayed with the `description` command.

The usage of the description command would then change to `restic description [snapshotID ...]` to show the complete description for all listed snapshots. (I would then also add the snapshot filter options like `--tag`, `--host` etc.)
To edit the description a `--edit` option would be added to the description subcommand.

```sh
Laborratte5@ARCH-MEDION ~/P/P/restic > ./restic snapshots --repo test-repo
ID        Time                 Host         Tags        Paths                                     Size         Description
--------------------------------------------------------------------------------------------------------------------------------------------
658d9f2e  2025-04-23 12:25:06  ARCH-MEDION              /home/Laborratte5/Projekte/Programmieren/restic  256.546 MiB  this is the first description
7d534f72  2025-04-23 13:31:35  ARCH-MEDION              /home/Laborratte5/Projekte/Programmieren/restic  532.816 MiB  my description
41a40da2  2025-04-23 13:32:21  ARCH-MEDION              /home/Laborratte5/Projekte/Programmieren/restic  695.289 MiB  this is a third description
afcaed29  2025-04-25 23:16:25  ARCH-MEDION              /home/Laborratte5/Projekte/Programmieren/restic  888.552 MiB  this is the first description
--------------------------------------------------------------------------------------------------------------------------------------------

Laborratte5@ARCH-MEDION ~/P/P/restic > ./restic description 41a40da2  --repo test-repo
this is a third description

longer description of the purpose and rational behind the snapshot.
lorem ipsum
yada yada
Laborratte5@ARCH-MEDION ~/P/P/restic > 
```

However in this case the description should be read via a file since multi line input with a cli may be a bad user experience.

Wonderer0:
> However in this case the description should be read via a file since multi line input with a cli may be a bad user experience.

I definitely agree.  It would be a good idea to allow a `--description-file` option to the `backup` command, with the first line shown in the list of snapshots, similar to Git as you describe above.  This allows users to write the descriptions using their favourite text editor, and then make the backup using the same command every time.

If the user always puts the description file in the same place near the top of the directory being packed up it would be possible to write a script to extract the full details the backups and build a backup history text file from them. 

I personally would prefer a shorter name like `--notes` or `--comment` (like Borg), but that's a very minor issue.

aawsome:
Having this in the restic design document would be nice!
About the JSON name: Please keep `description` - this is the name rustic also already uses!

Laborratte5:
> Having this in the restic design document would be nice!

I forgot about that. I'll add this to the doc/design.rst file under Snapshots.

> About the JSON name: Please keep description

Do you mean as the `message_type` of the json output?

aawsome:
> Do you mean as the `message_type` of the json output?

I mean the name used in JSON structure of the snapshot file. rustic already supports descriptions and saves them in the `description` key of the snapshot file. By using the same name, we ensure that descriptions added by rustic can be read by restic and vice versa.

Laborratte5:
That should already be the case
https://github.com/restic/restic/blob/24ecf02480eb25a7b61c38c2e88d79bc1b4be9bf/internal/restic/snapshot.go#L25
or am I missing something?

aawsome:
Yes it is already the correct name. I just wanted to emphasize that this should not be changed! ;-)

Laborratte5:
I now added this to the design document. Let me know if this is good enough or if I should add more info. (Although I am unsure what else to add)

internal/ui/table/table.go
```diff
@@ -25,6 +25,24 @@ type Table struct {
...
 	"join": strings.Join,
+	"trunc": truncateStr,
+}
+
+// Truncates the first line in `s` to the given `maxLength`
+func truncateStr(maxLen int, s string) string {
+	headline := strings.Split(s, "\n")[0]
+
+	if len(headline) > maxLen {
+		if len(headline) >= 3 {
```
MichaelEischer:
```suggestion
		if len(headline) > 3 {
```

internal/ui/table/table.go
```diff
@@ -25,6 +25,24 @@ type Table struct {
...
+			headline = headline[:maxLen-3]
+		} else {
+			headline = headline[:maxLen]
+		}
+
+		headline += "..."
```
MichaelEischer: Shouldn't the `...` only be appended when the headline is shortened to `maxLen-3`? That way the resulting string actually has the requested length.

cmd/restic/cmd_snapshots.go
```diff
@@ -256,6 +266,10 @@ func PrintSnapshots(stdout io.Writer, list restic.Snapshots, reasons []restic.Ke
 			data.Size = ui.FormatBytes(sn.Summary.TotalBytesProcessed)
 		}
 
+		if sn.Description != "" {
+			data.Description = sn.Description
+		}
```
MichaelEischer:
```suggestion
		data.Description = sn.Description
```
The if is redundant.

cmd/restic/cmd_backup.go
```diff
@@ -301,6 +303,8 @@ func (opts BackupOptions) Check(gopts GlobalOptions, args []string) error {
 		}
 	}
 
+	opts.DescriptionOptions.Check()
+
 	return nil
```
MichaelEischer:
```suggestion
	return opts.DescriptionOptions.Check()
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+func newDescriptionCommand() *cobra.Command {
+	var opts changeDescriptionOptions
+
+	cmd := &cobra.Command{
+		Use:   "description snapshotID [--set description | --set-file description]",
+		Short: "View or modify the description of snapshots",
+		Long: `
+The "description" command allows you to view or modify the description on an existing snapshot.
+
+The special snapshotID "latest" can be used to restore the latest snapshot in the
```
MichaelEischer: Copy&paste typo

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/restic/restic/internal/ui/termstatus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func newDescriptionCommand() *cobra.Command {
+	var opts changeDescriptionOptions
+
+	cmd := &cobra.Command{
+		Use:   "description snapshotID [--set description | --set-file description]",
```
MichaelEischer: Those option names don't match the naming in `descriptionOptions`.

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+	f.BoolVar(&opts.removeDescription, "remove-description", false, "remove the description from a snapshot")
+	opts.descriptionOptions.AddFlags(f)
+}
+
+func (opts *changeDescriptionOptions) Check() error {
+	err := opts.descriptionOptions.Check()
+	if err != nil {
+		return err
+	}
+	return nil
```
MichaelEischer:
```suggestion
	return opts.descriptionOptions.Check()
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+		return errors.Fatal("--description and --description-file cannot be used together")
+	}
+
+	return nil
+}
+
+// readDescription returns the description text specified by either the
+// `--description` option or the content of the `--description-file`
+func readDescription(opts descriptionOptions) (string, error) {
+	descriptionScanner := bufio.NewScanner(strings.NewReader(opts.Description))
```
MichaelEischer: What is the bufio construction used for when `opts.Description != ""`? The function could just return the `opts.Description`.

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+	if len(opts.DescriptionFile) > 0 {
+		// Read snapshot description from file
+		data, err := textfile.Read(opts.DescriptionFile)
+		if err != nil {
+			return "", err
+		}
+		descriptionScanner = bufio.NewScanner(bytes.NewReader(data))
```
MichaelEischer: It would make sense to somewhat limit the size of the description somewhat. As the description is a string, it must also be valid utf-8. This should also be checked early on.

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+			err := changeDescription(ctx, repo, sn, "", printFunc)
+			if err != nil {
+				Warnf("unable to remove the description of snapshot ID %q, ignoring: %v'\n", sn.ID(), err)
+				continue
+			}
+		}
+		return nil
+	}
+
+	if descriptionChange {
```
MichaelEischer:
```suggestion
	} else if descriptionChange {
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+		for sn := range FindFilteredSnapshots(ctx, repo, repo, &restic.SnapshotFilter{}, args) {
+			err := changeDescription(ctx, repo, sn, description, printFunc)
+			if err != nil {
+				Warnf("unable to modify the description for snapshot ID %s, ignoring: %v'\n", sn.ID().Str(), err)
+				continue
+			}
+		}
+	}
+
+	if !(opts.removeDescription || descriptionChange) {
```
MichaelEischer:
```suggestion
	else {
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if opts.removeDescription {
+		for sn := range FindFilteredSnapshots(ctx, repo, repo, &restic.SnapshotFilter{}, args) {
+			err := changeDescription(ctx, repo, sn, "", printFunc)
+			if err != nil {
+				Warnf("unable to remove the description of snapshot ID %q, ignoring: %v'\n", sn.ID(), err)
+				continue
+			}
+		}
+		return nil
```
MichaelEischer:
```suggestion
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+		for sn := range FindFilteredSnapshots(ctx, repo, repo, &restic.SnapshotFilter{}, args) {
+			fmt.Fprintf(gopts.stdout, "Description of snapshot %s:\n%s\n", sn.ID().Str(), sn.Description)
+		}
+	}
+
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	return nil
```
MichaelEischer:
```suggestion
	return ctx.Err()
```

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func runDescription(ctx context.Context, opts changeDescriptionOptions, gopts GlobalOptions, term *termstatus.Terminal, args []string) error {
+	// check arguments
+	switch {
+	case len(args) < 1:
+		return errors.Fatal("no snapshot ID specified")
+	case len(args) > 1:
+		return errors.Fatalf("more than one snapshot ID specified: %v", args)
```
MichaelEischer: Setting or removing a description also works for multiple snapshots, just like the `tags` command.

cmd/restic/cmd_description.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	defer unlock()
+
+	printFunc := func(c changedSnapshot) {
+		Verboseff("old snapshot ID: %v -> new snapshot ID: %v\n", c.OldSnapshotID, c.NewSnapshotID)
+	}
+
+	if gopts.JSON {
+		printFunc = func(c changedSnapshot) {
+			term.Print(ui.ToJSONString(c))
```
MichaelEischer: The JSON output should also be documented at `doc/075_scripting.rst`.

MichaelEischer (changes requested):
I somewhat doubt that we need a `description` command. The `--set` and `--remove` options could just as well be added to the `rewrite` command (I'd expect rather little usage of those options). And the `snapshots` command should IMHO default to showing the full description (just like `git log` does by default). If `--compact` is passed, then it would be fine to truncate the description output.

What do you think?

The merge conflicts are most likely due to https://github.com/restic/restic/pull/5518 and https://github.com/restic/restic/pull/5525 . Just let me know me if you need help with rebasing the PR.

Laborratte5:
> I somewhat doubt that we need a `description` command. The `--set` and `--remove` options could just as well be added to the `rewrite` command (I'd expect rather little usage of those options).

Yeah that is fair. I will move it to the `rewrite` command when I have time.

cmd/restic/cmd_description.go
Laborratte5: The rational behind this was that I thought of descriptions as somewhat unique per snapshot. Therefor the use-case of changing the description of multiple snapshots at once is most likely an accident and to avoid this I added this limit.
Although I do realize now that it still would not be that big of a problem if a user changed the description of multiple snapshots on accident since the old snapshot description is not lost due to snapshots being read-only. Moreover since I am moving this to the `rewrite` command to keep the `rewrite` command consistent changing multiple snapshot descriptions at once should be supported.

internal/ui/table/table.go
Laborratte5: The idea was to make it obvious that the string got truncated. Otherwise it may be that a string is truncated without adding the dots. ("ab" truncated to 1 would become just "a"). But when I change the `snapshot` command to only truncated with the `--compact` option I don't think this is an issue anymore, since the user knows that things may be truncated and it is an edge case anyways.

cmd/restic/cmd_description.go
Laborratte5: > As the description is a string, it must also be valid utf-8

This should already be addressed by `textfile.Read` if I understand it correctly.

> It would make sense to somewhat limit the size of the description somewhat

I addressed this now by checking the description size after reading it. This still means that a large file would be read completely before the error is thrown, but it at least prevents user from saving a snapshot with a too large description into the repository.
I arbitrarily chose 4096 as a size since it looked big enough for description text in my terminal.

Laborratte5:
I rebased this branch now and changed my code to use the new Terminal/Print system.
Since I had a hard time judging which messages should use which verbosity level it would be good if someone could take a look at this. (Especially the usage of `P` vs. `PT`)

cmd/restic/cmd_description.go
Laborratte5: I removed the JSON output for now, since `rewrite` does not have JSON output currently.

Laborratte5:
All things should have been addressed now.

changelog/unreleased/issue-2376
```diff
@@ -0,0 +1,16 @@
+Enhancement: Add per snapshot description
+
+Restic can now store user supplied descriptions for snapshots.
```
MichaelEischer:
```suggestion
Restic can now store user-supplied descriptions for snapshots.
```

internal/data/snapshot.go
```diff
@@ -23,6 +23,7 @@ type Snapshot struct {
 	UID      uint32     `json:"uid,omitempty"`
 	GID      uint32     `json:"gid,omitempty"`
 	Excludes []string   `json:"excludes,omitempty"`
+	Description string    `json:"description,omitempty"`
```
MichaelEischer: That's not properly formatted.

internal/ui/table/table.go
```diff
@@ -25,6 +25,23 @@ type Table struct {
...
 var funcmap = template.FuncMap{
 	"join": strings.Join,
+	"trunc": truncateStr,
+}
+
+// Truncates the first line in `s` to the given `maxLength`
+func truncateStr(maxLen int, s string) string {
+	headline := strings.Split(s, "\n")[0]
+
+	if len(headline) > maxLen {
```
MichaelEischer: This is incorrect when dealing with UTF-8. Just use `internal/ui.Truncate` instead.

doc/040_backup.rst
```diff
@@ -646,6 +646,20 @@ the pipe and act accordingly (e.g., remove the last backup). Refer to the
...
+It is possible to add arbitrary text to a snapshot using ``--description``
+or ``--description-file``.
+The current maximum length of a description is 4096 characters.
+
+.. code-block:: console
+
+    $ restic -r /srv/restic-repo backup --description "finished important article" ~/work
+    [...]
+
+The description for each snapshot can be viewed with the ``snapshot`` command.
```
MichaelEischer, line 757: Should also explain the following: How can I see the full description? The text output only contains the truncated description.

cmd/restic/cmd_snapshots.go
```diff
@@ -191,6 +197,9 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
 		if hasSize {
 			tab.AddColumn("Size", `{{ .Size }}`)
 		}
+		if hasDescription {
+			tab.AddColumn("Description", `{{ trunc 20 .Description }}`)
```
MichaelEischer: I'm wondering whether this is too short or not. But I have no experience value what a typical description length could be.

cmd/restic/cmd_rewrite.go
```diff
@@ -100,13 +110,66 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
 	return &snapshotMetadata{Hostname: sma.Hostname, Time: timeStamp}, nil
 }
 
+// changeDescriptionOptions collects all options for description changing
+type changeDescriptionOptions struct {
+	descriptionOptions
+	removeDescription bool
```
MichaelEischer: As the description is just a string field in the snapshot, it would also be possible to just use `--description ""`. I'm somewhat undecided whether that would be simpler to use or not.

cmd/restic/cmd_rewrite.go
```diff
@@ -100,13 +110,66 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
...
+}
+
+func (opts *changeDescriptionOptions) AddFlags(f *pflag.FlagSet) {
+	f.BoolVar(&opts.removeDescription, "remove-description", false, "remove the description from a snapshot")
+	opts.descriptionOptions.AddFlags(f)
+}
+
+func (opts *changeDescriptionOptions) Check() error {
+	if opts.removeDescription && !opts.descriptionOptions.empty() {
+		return errors.Fatal("cannot set and remove description simultanious")
```
MichaelEischer:
```suggestion
		return errors.Fatal("cannot set and remove description at the same time")
```

cmd/restic/cmd_rewrite.go
```diff
@@ -100,13 +110,66 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
...
+	return opts.descriptionOptions.Check()
+}
+
+func (opts *changeDescriptionOptions) empty() bool {
+	return !opts.removeDescription && opts.descriptionOptions.empty()
+}
+
+const maxDescriptionLength = 4096
+
+var descriptionTooLargeErr = errors.New(fmt.Sprintf("The provided descriptions exceeds the maximum length of %d bytes.", maxDescriptionLength))
```
MichaelEischer:
```suggestion
var descriptionTooLargeErr = fmt.Errorf("the provided description exceeds the maximum length of %d bytes", maxDescriptionLength)
```

cmd/restic/cmd_rewrite.go
```diff
@@ -100,13 +110,66 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
...
+	description := opts.Description
+	if opts.DescriptionFile != "" {
+		// Read snapshot description from file
+		data, err := textfile.Read(opts.DescriptionFile)
+		if err != nil {
+			return "", err
+		}
+		descriptionScanner := bufio.NewScanner(bytes.NewReader(data))
+		var builder strings.Builder
+		for descriptionScanner.Scan() {
```
MichaelEischer: What is the point of this loop?

cmd/restic/cmd_rewrite.go
```diff
@@ -186,11 +255,11 @@ func rewriteSnapshot(ctx context.Context, repo *repository.Repository, sn *data.
 	}
 
 	return filterAndReplaceSnapshot(ctx, repo, sn,
-		filter, opts.DryRun, opts.Forget, metadata, "rewrite", printer)
+		filter, opts.DryRun, opts.Forget, metadata, description, "rewrite", printer)
 }
 
 func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *data.Snapshot,
-	filter rewriteFilterFunc, dryRun bool, forget bool, newMetadata *snapshotMetadata, addTag string, printer progress.Printer) (bool, error) {
+	filter rewriteFilterFunc, dryRun bool, forget bool, newMetadata *snapshotMetadata, description string, addTag string, printer progress.Printer) (bool, error) {
```
MichaelEischer: A new description should be passed as part of `newMetadata *snapshotMetadata`. The description must also be passed as string pointer. Otherwise, there's no way to distinguish between "keep the current description" and "remove description".

cmd/restic/cmd_rewrite.go
```diff
@@ -242,6 +313,8 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *d
 			printer.P("would set hostname to %s", newMetadata.Hostname)
 		}
 
+		printer.P("would set description to %s", description)
```
MichaelEischer: Should only be printed if a change was requested.

cmd/restic/cmd_rewrite.go
```diff
@@ -266,6 +339,9 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *d
 		sn.Hostname = newMetadata.Hostname
 	}
 
+	printer.P("setting description to %s", description)
+	sn.Description = description
```
MichaelEischer: Should only be changed if a change was requested.

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -467,6 +467,41 @@ func TestIncrementalBackup(t *testing.T) {
...
+	testSetupBackupData(t, env)
+	opts := BackupOptions{}
+
+	testRunBackup(t, "", []string{env.testdata}, opts, env.gopts)
+	testRunCheck(t, env.gopts)
+	newest, _ := testRunSnapshots(t, env.gopts)
+
+	if newest == nil {
+		t.Fatal("expected a backup, got nil")
+	}
```
```suggestion
	rtest.Assert(t, newest != nil, "expected a backup, got nil")
```

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -467,6 +467,41 @@ func TestIncrementalBackup(t *testing.T) {
...
+
+	rtest.Assert(t, newest.Description == "",
+		"expected no description, got %v", newest.Description)
+	parent := newest
```
MichaelEischer: I don't really see the point of setting a parent snapshot here.

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -467,6 +467,41 @@ func TestIncrementalBackup(t *testing.T) {
...
 
+func TestBackupDescription(t *testing.T) {
+	env, cleanup := withTestEnvironment(t)
+	defer cleanup()
+
+	testSetupBackupData(t, env)
+	opts := BackupOptions{}
+
+	testRunBackup(t, "", []string{env.testdata}, opts, env.gopts)
+	testRunCheck(t, env.gopts)
```
MichaelEischer: Can omit the call to `check`.

cmd/restic/cmd_rewrite_integration_test.go
```diff
@@ -154,6 +161,46 @@ func TestRewriteMetadata(t *testing.T) {
 	}
 }
 
+func TestDescription(t *testing.T) {
+
```
MichaelEischer: Unnecessary whitespace

cmd/restic/cmd_rewrite_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestReadDescription(t *testing.T) {
+	wantDescription := "This is a short test description."
+	opts := descriptionOptions{
+		Description: wantDescription,
+	}
+	gotDescription, err := readDescription(opts)
+
+	rtest.OK(t, err)
+	rtest.Assert(t, wantDescription == gotDescription, "Wanted '%s' description, got '%s'", wantDescription, gotDescription)
+}
```
MichaelEischer: Please also add a test case for reading a description from a file. Use `t.TempDir()` to store the temporary file.

MichaelEischer (changes requested):
Thanks for the changes so far! I've found a bunch of small issues and a larger bug in the `rewrite` command. The current code always overwrites the description, such that it would overwrite an existing description even if not requested by the user.

We could also use `--description ""` instead of `--remove-description`. Although, I'm not sure whether that would be simpler to use.

rawtaz:
> We could also use --description "" instead of --remove-description. Although, I'm not sure whether that would be simpler to use.

The first one is better IMO. Since the description field will be there anyway, it's much better to just set the description to be empty, than to add additional options to act on it. That's one single interface to set/edit the description.

Laborratte5:
As you have noticed I haven't come around working on this PR lately. I fixed some of the mentions above before realizing you reassigned this issue. Do you still want me to push those changes or should I leave it to you now to avoid more overhead?

MichaelEischer:
@Laborratte5 the assignee status just used to track who takes care of the PR reviews and follow ups.

cmd/restic/cmd_rewrite.go
Laborratte5: `Scanner` tokenizes the input line by line. The loop ensures all lines of the file are read.

cmd/restic/cmd_rewrite.go
Laborratte5: I do like the idea of the `--description ""`. However, it seems like there is no way to avoid the default value with `pflag`. This means there is no way of distinguishing if the default value of the flag was used because the flag was not set by the user (in which case the description should not be changed) or if the user provided an empty value in order to remove the description.
If there is a solution to this I am happy to implement it since I do like the simpler interface. However, I have not found a good way to solve this issue.

The only solution that came to my mind would be using a random string as a default value and then comparing the flag value with this string, but this feels too hacky to me.

MichaelEischer:
Are you still interested in finishing the PR?

Laborratte5:
Yes, I am. I was waiting for feedback regarding this (sorry for the communication problems):

> I do like the idea of the `--description ""`. However, it seems like there is no way to avoid the default value with `pflag`. This means there is no way of distinguishing if the default value of the flag was used because the flag was not set by the user (in which case the description should not be changed) or if the user provided an empty value in order to remove the description. If there is a solution to this I am happy to implement it since I do like the simpler interface. However, I have not found a good way to solve this issue.
> 
> The only solution that came to my mind would be using a random string as a default value and then comparing the flag value with this string, but this feels too hacky to me.

I did push my current state. The only thing that is missing is the direction regarding the `--remove-decription` flag and the documentation which I try to finish this week.

cmd/restic/cmd_rewrite.go
MichaelEischer: It's possible to check whether a flag was explicitly specified on the command line by checking the flag itself. Search for `readConcurrencyFlag` for an example.

cmd/restic/cmd_rewrite.go
```diff
@@ -97,7 +129,59 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
...
+	return !opts.removeDescription && opts.descriptionOptions.empty()
+}
+
+const maxDescriptionLength = 4096
+
+var descriptionTooLargeErr = fmt.Errorf("the provided description exceeds the maximum length of %d bytes", maxDescriptionLength)
+
+// readDescription returns the description text specified by either the
+// `--description` option or the content of the `--description-file`
+func readDescription(opts descriptionOptions) (string, error) {
```
MichaelEischer: Should move to `cmd_backup` next to the definition `descriptionOptions`. Also change this to a method of `descriptionOptions`. If you want, you can also extract the `descriptionOptions` into a separate file like `description.go` or similar.

cmd/restic/cmd_rewrite.go
```diff
@@ -97,7 +129,59 @@ func (sma snapshotMetadataArgs) convert() (*snapshotMetadata, error) {
 		}
 		timeStamp = &t
 	}
-	return &snapshotMetadata{Hostname: sma.Hostname, Time: timeStamp}, nil
+	return &snapshotMetadata{Description: description, Hostname: sma.Hostname, Time: timeStamp}, nil
+}
+
+// changeDescriptionOptions collects all options for description changing
+type changeDescriptionOptions struct {
```
MichaelEischer: Without `--remove-description` it should be possible to remove this struct. From what I can see `cmd_backup` shouldn't have problems if we explicitly track whether description was specified.

cmd/restic/cmd_rewrite.go
MichaelEischer: Doesn't `description, _ = strings.CutSuffix(string(data), "\n")` do the same as this loop?

cmd/restic/cmd_rewrite.go
```diff
@@ -285,15 +387,18 @@ func filterAndReplaceSnapshot(ctx context.Context, repo restic.Repository, sn *d
 
 func runRewrite(ctx context.Context, opts RewriteOptions, gopts global.Options, args []string, term ui.Terminal) error {
 	if !opts.SnapshotSummary && opts.ExcludePatternOptions.Empty() && opts.Metadata.empty() {
-		return errors.Fatal("Nothing to do: no excludes provided and no new metadata provided")
+		return errors.Fatal("Nothing to do: no excludes provided, no new metadata provided and no description manipulation flag provided")
```
MichaelEischer:
```suggestion
		return errors.Fatal("Nothing to do: no excludes provided, no new metadata provided and no description modification flag provided")
```

MichaelEischer (changes requested):
> The only thing that is missing is the direction regarding the `--remove-decription` flag

I've completely missed your question for that flag. I've answered below https://github.com/restic/restic/pull/5375/changes#r3311748124 .

PR #5384, opened 2025-05-03
check: add percentage of repository checked

### LINKED ISSUE #5377: Minor: add percentage to `restic check --read-data-subset 2G` output
Output of `restic version`
--------------------------

restic 0.18.0 compiled with go1.24.1 on darwin/arm64

When I run ` restic -r $RESTIC_LOCAL_REPOSITORY check --read-data-subset 2G`, I get the following as part of the output:

```
read 2147483648 bytes of data packs
```

I'd like it to also include the percentage, e.g. `read 2147483648 bytes (2.5%) of data packs`.

Just printing the bytes made sense before specifying the subset size in GB was implemented (https://github.com/restic/restic/issues/3490), but now it'd be nice to have both.

See above

Yes! I really appreciate it!

konidev20:
* Minor change to calculate and print the percentage of the repository checked

Closes #5377 

Checklist
---------
~- [ ] I have added tests for all code changes.~
~- [ ] I have added documentation for relevant changes (in the manual).~
~- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).~

### DIFF
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -425,7 +425,8 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args 
 				subsetSize = repoSize
 			}
 			packs = selectRandomPacksByFileSize(chkr.GetPacks(), subsetSize, repoSize)
-			printer.P("read %d bytes of data packs\n", subsetSize)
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. Thanks!

PR #5385, opened 2025-05-03
docs: updated installation docs for Windows

Silvenga:

This PR updates the [installation docs](https://restic.readthedocs.io/en/stable/020_installation.html) around Windows. It makes two changes:

- Adds documentation on using WinGet to install restic, noting that `Machine` scope is likely wanted by end-users.
- Removes references to installing the restic binary into `%SystemRoot%\System32`, because it's just kind of weird (partly due to security). The Linux `%SystemRoot%\System32` equivalent is `/sbin/` (where restic should typically be installed into `/usr/bin` or `/usr/local/bin`). `%SystemRoot%\System32` just happens to be one of those paths that is included in the `SYSTEM` `PATH` by default.

Nope.

Checklist
---------

### DIFF
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -175,15 +175,19 @@ restic can be installed from the official repo of Solus via the ``eopkg`` packag
 Windows
 =======
 
-restic can be installed using `Scoop <https://scoop.sh/>`__:
+restic can be installed using either `Scoop <https://scoop.sh/>`__ or `WinGet <https://learn.microsoft.com/en-us/windows/package-manager/>`__.
+
+Regardless of the method, the ``restic.exe`` binary will be added to your ``PATH`` automatically, making the ``restic`` command accessible in Powershell or CMD.
 
 .. code-block:: console
 
     scoop install restic
 
-Using this installation method, ``restic.exe`` will automatically be available
-in the ``PATH``. It can be called from cmd.exe or PowerShell by typing ``restic``.
+.. code-block:: console
 
+    winget install --exact --id restic.restic --scope Machine
+
+By default, WinGet will install restic into the ``User`` scope, which is typically in your user's ``%LOCALAPPDATA%`` directory. This behavior may be undesirable for system-wide backups, so specifying ``--scope Machine`` is recommended so that restic is installed into ``%ProgramFiles%``. This requires elevation.
 
 .. _official_binaries:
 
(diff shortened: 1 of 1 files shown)


rawtaz:
I like the addition of WinGet. But it's a bit sad if one has to use a third party package manager to install restic on Windows. Is there really no substitute for `%SystemRoot%\System32` that we can suggest instead, in order to keep the "just drop the binary here" part as well for simplicity? After all, restic being a single binary build with Go that you can just run is one of the great things about it.

Silvenga:
@rawtaz WinGet is a package manager from Microsoft (and OSS contributions) for Windows.

> WinGet the Windows Package Manager is available on Windows 11, modern versions of Windows 10, and Windows Server 2025 as a part of the App Installer. The App Installer is a System Component delivered and updated by the Microsoft store on Windows Desktop versions, and via Updates on Windows Server 2025 - [Docs](https://learn.microsoft.com/en-us/windows/package-manager/winget/)

The only default PATH locations on Windows are `SYSTEM` owned, and really shouldn't be modified by end-users. The standard approach in the land of Windows is to build an installer that registered the new folders in PATH (so Windows Installer can automatically heal installations + cleanup assets as needed).

Shipped single binaries aren't that common in Windows (for lot's of reasons, but the main one is security, Windows Ops dislikes prefers dynamically linked binaries) - it was a big effort to add support for these so called "portable" installations in WinGet (and indirectly, CLI apps shipped via the Windows Store) just because the structure of how programs are typically managed is radically different from Linux philosophies.

Users can use things like the `setx` built-in to set the the user's PATH (by modifying the registry hive) - but I suspect most users just use the UI for this because the UI does directory validation and ensures the PATH format is correct (searching for `Edit environment variables for your account` in the Start menu).

TBH, users that don't know how to manage their `PATH`, likely shouldn't be modifying the core system directories (possibly breaking Windows security in the processes). The "weirdness" I mentioned before.

So I'm not sure if docs are really needed for running restic without a path, e.g., there's no directions for MacOS.

rawtaz:
@Silvenga Thank you for such an elaborate response, and for educating me about WinGet in Windows :)

Seeing as the "Official binaries" section already mentions "Just download and run the one matching your system", I'm totally fine with the proposed changes. I'll leave it for @MichaelEischer to have a look too though, just in case he feels differently about it. It will probably take a while as I know he's busy now.

rawtaz (approved):
LTGM

PR #5389, opened 2025-05-10
Update passwords section of faq

fronesis47:
This section was out of date: it contained a dead link to an old issue, and did not mention the RESTIC_PASSWORD_COMMAND variable. 

I also added some basic info on using built-in keyring/keychains to store and retrieve env variables, including passwords.

Yes, it was discussed here, concluding with the suggestion that I try to update this faq:
https://forum.restic.net/t/why-cant-restic-save-credentials-directly-to-keychain-keyring/9637/8

Checklist
---------

_No code changed, just docs_

I think it's ready to review, but I've never done this before, so please be patient with me

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -110,9 +110,8 @@ How can I specify encryption passwords automatically?
 When you run ``restic backup``, you need to enter the passphrase on
 the console. This is not very convenient for automated backups, so you
 can also provide the password through the ``--password-file`` option, or one of
-the environment variables ``RESTIC_PASSWORD`` or ``RESTIC_PASSWORD_FILE``.
-A discussion is in progress over implementing unattended backups happens in
-:issue:`533`.
+the environment variables: ``RESTIC_PASSWORD``, ``RESTIC_PASSWORD_FILE``,
+or ``RESTIC_PASSWORD_COMMAND``.
 
 .. important:: Be careful how you set the environment; using the env
                command, a `system()` call or using inline shell
@@ -124,9 +123,32 @@ A discussion is in progress over implementing unattended backups happens in
                `accessible only to that user`_. Please make sure that
                the permissions on the files where the password is
                eventually stored are safe (e.g. `0600` and owned by
-               root).
+               root). Note also that ``RESTIC_PASSWORD_COMMAND`` is
+               safe because it does not export the password itself to
+               the environment.
 
 .. _accessible only to that user: https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009
+
+On platforms with an available keychain, keyring or similar secret store, a
+user can add and then dynamically retrieve passwords, cloud credentials,
+repository paths, or any other data deemed sensitive. Here's an example of
+part of a shell script using the `built-in`_ ``security`` command on macOS
+to retrieve credentials from the system's Keychain before running various
+``restic`` commands:
+
+.. _built-in: https://ss64.com/mac/security.html
+
+::
+
+    export GOOGLE_PROJECT_ID=$(security find-generic-password -a resticGCS -s restic_project_ID -w)
+
+    export GOOGLE_APPLICATION_CREDENTIALS=$(security find-generic-password -a resticGCS -s restic_key -w)
+
+    export RESTIC_REPOSITORY=$(security find-generic-password -a resticGCS -s restic_repo_path -w)
+
+    export RESTIC_PASSWORD_COMMAND='security find-generic-password -a resticGCS -s restic_pwd -w' 
+
+
 
 How to prioritize restic's IO and CPU time
 ------------------------------------------


doc/faq.rst
```diff
@@ -110,24 +110,28 @@ How can I specify encryption passwords automatically?
...
 
 .. _accessible only to that user: https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009
 
+On platforms with an available keychain/keyring, various environment variables can first be added to the keychain, and then retrieved dynamically. Both actions use the `built-in`_ ``security`` command. A user can add/retrieve not only passwords but also cloud credentials, repository paths, or any other variables the user would prefer to store in the keyring/keychain. Here's an example of part of a shell script that a restic user on MacOS could use to retrieve credentials from the system Keychain – all before running various ``restic`` commands.
```
rawtaz, line 119: I had too many small comments on this so I'll just suggest a changed text instead:

~~~
On platforms with an available keychain, keyring or similar secret store, a user can add and then dynamically retrieve passwords, cloud credentials, repository paths, or any other data deemed sensitive. Here's an example of part of a shell script using the `built-in`_ ``security`` command on macOS to retrieve credentials from the system's Keychain before running various ``restic`` commands:
~~~

Also, please wrap the text to be as wide as the existing text in the document, thanks!

rawtaz (changes requested):
Very good addition to make users aware of this possibility!

doc/faq.rst
fronesis47: That’s a fantastic rewrite – thanks! I’ll incorporate this change and rewrap the text when I’m back in front of a computer.

doc/faq.rst
```diff
@@ -110,24 +110,28 @@ How can I specify encryption passwords automatically?
...
-               directly and they will be readable to all users on a
-               system. Using export in a shell script file should be
-               safe, however, as the environment of a process is
-               `accessible only to that user`_. Please make sure that
-               the permissions on the files where the password is
-               eventually stored are safe (e.g. `0600` and owned by
-               root).
+the environment variables: ``RESTIC_PASSWORD``, ``RESTIC_PASSWORD_FILE``, or ``RESTIC_PASSWORD_COMMAND``.
+
+Be careful how you set the environment; using the env command, a `system()` call or using inline shell scripts (e.g. `RESTIC_PASSWORD=password restic ...`) might expose the credentials in the process list directly and they will be readable to all users on a system. Using export in a shell script file should be safe, however, as the environment of a process is `accessible only to that user`_. Please make sure that the permissions on the files where the password is eventually stored are safe (e.g. `0600` and owned by root). Note also that ``RESTIC_PASSWORD_COMMAND`` is safe because it does not export the password itself to the environment.
```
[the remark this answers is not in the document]
fronesis47: - Thanks!
- Sorry, I’m missing something, because I don’t see the “little line break” before “or” in the original text. So I’m not sure what to do on that one. 
- I’m happy to add the ```important``` back in. I removed it because looking at the previous version, it felt like this was a notice that came at the _end_ of the section. When I revised it came in the middle, and so it felt more organic to just transition to the “Be careful” line. But I can see that there are other ```important``` flags throughout, so I can add it back in if that seems better.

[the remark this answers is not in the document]
fronesis47: UPDATE: Ok, I think I understand that the line break was related to the ```important``` tag. I've added the tag back and I _think_ I got the lines breaking correctly, but you'll need to double-check me.

fronesis47:
I believe I've now made all requested changes.

doc/faq.rst
```diff
@@ -110,24 +110,37 @@ How can I specify encryption passwords automatically?
...
-               root).
+the environment variables: ``RESTIC_PASSWORD``, ``RESTIC_PASSWORD_FILE``, or ``RESTIC_PASSWORD_COMMAND``.
+
+.. important:: Be careful how you set the environment; using the env command, a `system()` call or using 
+                inline shell scripts (e.g. `RESTIC_PASSWORD=password restic ...`) might expose the credentialsin 
+                the process list directly and they will be readable to all users on a system. Using export in a shell
+                script file should be safe, however, as the environment of a process is `accessible only to that user`_. 
+                Please make sure that the permissions on the files where the password is eventually stored are safe 
+                (e.g. `0600` and owned by root). Note also that ``RESTIC_PASSWORD_COMMAND`` is safe because it does not 
+                export the password itself to the environment.
```
rawtaz, line 121:
```suggestion
the environment variables: ``RESTIC_PASSWORD``, ``RESTIC_PASSWORD_FILE``,
or ``RESTIC_PASSWORD_COMMAND``.

.. important:: Be careful how you set the environment; using the env
               command, a `system()` call or using inline shell
               scripts (e.g. `RESTIC_PASSWORD=password restic ...`)
               might expose the credentials in the process list
               directly and they will be readable to all users on a
               system. Using export in a shell script file should be
               safe, however, as the environment of a process is
               `accessible only to that user`_. Please make sure that
               the permissions on the files where the password is
               eventually stored are safe (e.g. `0600` and owned by
               root). Note also that ``RESTIC_PASSWORD_COMMAND`` is
               safe because it does not export the password itself to
               the environment.
```

doc/faq.rst
```diff
@@ -110,24 +110,37 @@ How can I specify encryption passwords automatically?
...
 
 .. _accessible only to that user: https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009
 
+On platforms with an available keychain, keyring or similar secret store, a user can add and then dynamically retrieve passwords, 
+cloud credentials, repository paths, or any other data deemed sensitive. Here's an example of part of a shell script using the 
+`built-in`_ ``security`` command on macOS to retrieve credentials from the system's Keychain before running various ``restic`` 
+commands:
```
rawtaz, line 128:
```suggestion
On platforms with an available keychain, keyring or similar secret store, a
user can add and then dynamically retrieve passwords, cloud credentials,
repository paths, or any other data deemed sensitive. Here's an example of
part of a shell script using the `built-in`_ ``security`` command on macOS
to retrieve credentials from the system's Keychain before running various
``restic`` commands:
```

rawtaz (changes requested):
Thank you for your edits. I made some nit-picking changes in a couple of suggestions.

It's mainly just fixing whitespace to align with the 80 character recommendation in the original Docutils project that introduced reStructuredText. Also, in the "important" notice, since you have only added a sentence at the end, I restored the original formatting so that the change in the commit becomes only the relevant part.

fronesis47:
OK, I committed both those changes. It's clear to me now that I don't know how line breaks work in .rst files, because I can't really tell what I did wrong. When it looked at my code it seemed ok, but clearly the breaks were all wrong. If you can point me to a primer on how to do this correctly, I'll try to figure it out for next time.

Thanks!

rawtaz:
Looks good to me. Can you please also squash the commits into one commit? Then we can merge this :)

So, before I made the latest suggestions, the "important" notice had been reformatted entirely, so that every line in it appeared as changed. Since the *actual* changes you wanted to make to it was just to add a single sentence at the end of it, there's little point in reformatting more text than that. It's cleaner and nicer if the commits you make only change the lines they are intended to change.

About the line breaks, all I initially meant was that the text with/about the additional environment variable you added should be line wrapped like the already existing text, instead of spanning a much wider line width which makes it stand out from the rest of the text and also isn't in line with the [recommendations for reStructuredText](https://devguide.python.org/documentation/markup/#use-of-whitespace) which recommends 80 characters.

fronesis47:
If you remove the merge block, I should then be able to select "squash and merge" at the same time, right? 

I can also try to squash the commits through the command-line, but I'll need to set that up, as right now my current computer is not configured to access GitHub from the terminal. (Sorry, newbie!).

fronesis47:
OK, I follow you. I'll give it a go this afternoon and then let you know how I made out. :)

fronesis47:
OK, I think I successfully squashed the commits without breaking anything. :)

rawtaz (approved):
I'm a bit unsure about a few points of grammar but this is totally good enough so let's merge this!

rawtaz:
Thanks again for your contribution, and congratulations on the first PR and squash :>

fronesis47:
Thanks for your very patient help!

PR #5396, opened 2025-05-25
restic list pack snapshotID - create lists of packfiles for a given snapshot as text 

### LINKED ISSUE #5372: Feature request: Obtain the list of blocks in a snapshot
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on darwin/amd64

Have a way to list the blocks present in a snapshot; even better, the blocks in a snapshot in a file path

An alternative would be a way to try to read fist few bytes in a block and report a list of failures (which would save a lot of time as well).

We keep all backups in AWS S3. We have a lifecycle policy to transition to IA after 90 days and to DEEP_ARCHIVE after 390 days.

The deduplication feature is awesome, but it is really annoying when trying to provide developers with a 3 weeks old backup and you notice 2, 3 maybe 9 blocks in that backup are in DEEP_ARCHIVE.

The only way to do it is to attempt a restore, capture all those 

```
Load(<data/7f7096ecc6>, 1362710, 3273655) returned error, retrying after 1m3.303469605s: The operation is not valid for the object's storage class
```

then make a unique list of failed block ids and feed the list into a glacier restore script.

Yes, restic is awesome!

wplapper:
Command to show packfiles which belong to a snapshot or list of snapshots.

packfilelist gathers all snapshots from the the standard restic.SnapshotFilter, collects the trees which belong to these snapshots and calls restic.FindUsedBlobs to collect all blobs participating in these snapshots. These blobs get translated to their containing packfiles and these lists are printed out either as standard text of in JSON format. The `--detail` options allows for more insights to be looked at.

This PR addresses issues raised in #5372.

https://github.com/restic/restic/issues/5372
https://github.com/restic/restic/issues/5751

Closes #5372

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5372
+++ b/changelog/unreleased/issue-5372
@@ -1,0 +1,8 @@
+Enhancement: command to list packfiles belonging to a snapshot
+
+This enhancement creates a new option `packfiles` for command `restore`. It shows the
+packfiles which are related to a snapshot or a set of snapshots. Output is either
+plain text or JSON. Different levels of details can be activated.
+
+https://github.com/restic/restic/issues/5372
+https://github.com/restic/restic/pull/5396
--- a/changelog/unreleased/issue-5372
+++ b/changelog/unreleased/issue-5372
@@ -1,8 +1,8 @@
-Enhancement: command to list packfiles belonging to a snapshot
+Enhancement: list packfiles belonging to a snapshot
 
-This enhancement creates a new option `packfiles` for command `restore`. It shows the
-packfiles which are related to a snapshot or a set of snapshots. Output is either
-plain text or JSON. Different levels of details can be activated.
+This enhancement uses command ``list packs snapshotID`` to show the packfiles
+belonging to the given snapshot. It can be use to apply S3 object lock
+to all packs of the given snapshot.
 
 https://github.com/restic/restic/issues/5372
 https://github.com/restic/restic/pull/5396
--- a/changelog/unreleased/issue-5372
+++ b/changelog/unreleased/issue-5372
@@ -1,8 +1,7 @@
-Enhancement: list packfiles belonging to a snapshot
+Enhancement: Add command to list packfiles belonging to a snapshot
 
-This enhancement uses command ``list packs snapshotID`` to show the packfiles
-belonging to the given snapshot. It can be use to apply S3 object lock
-to all packs of the given snapshot.
+This enhancement adds command `list packs snapshotID` to show the packfiles
+belonging to the given snapshot.
 
 https://github.com/restic/restic/issues/5372
 https://github.com/restic/restic/pull/5396
--- a/cmd/restic/cmd_list.go
+++ b/cmd/restic/cmd_list.go
@@ -98,15 +98,10 @@ func runList(ctx context.Context, gopts global.Options, args []string, term ui.T
 }
 
 // packfileLIst runs the list packs <snapshotID>
-// it creates a sorted list of packfiles belonging to this snapshot
+// it prints a sorted list of packfiles belonging to this snapshot
 func packfileLIst(ctx context.Context, repo restic.Repository, snapshotID string, printer progress.Printer) error {
 
-	snapshotLister, err := restic.MemorizeList(ctx, repo, restic.SnapshotFile)
-	if err != nil {
-		return err
-	}
-
-	sn, _, err := (&data.SnapshotFilter{}).FindLatest(ctx, snapshotLister, repo, snapshotID)
+	sn, _, err := (&data.SnapshotFilter{}).FindLatest(ctx, repo, repo, snapshotID)
 	if err != nil {
 		return fmt.Errorf("required snapshot ID %q not found", snapshotID)
 	}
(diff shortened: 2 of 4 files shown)


wplapper:
I have decided to revert this PR to closed status to ease the pressure on the 
maintainers of restic. Once the number of PRs has dropped considerably, I might
moved this PR back to the queue to be reviewed. In case you consider this PR
important, please contact me via the usual channels. Thank you!

MichaelEischer (changes requested):
I seriously dislike adding an extra command for something that is conceptually a part of the `restore` command. Clarify the use case and sketch how a cleanly integrated solution could look like before making any further code changes.

The remarks below highlight additional problems, in addition to the overall design issue.

Please only add options that where it's clear what their use case is. Exposing arbitrary internal details of the repository format is a bad idea as this will hinder future evolution of the repository format.

As the original use case for this command was to get a list of `packfiles` that must be warmed up prior to a `restore` operation, it should focus on that. With that limited focus it is probably a much better fit as an extension to an existing command. Btw, the use case of getting a list of all pack files containing tree blobs should be solved via a `restic list treepacks` command.

Some further comments on the options:
- `--short-id`: I don't see where this would ever be useful. The command is highly technical so there's nothing to gain from short ids.
- `--summary`: Not useful for the mentioned use cases. Some of the information _might_ make sense for the `stats` command, but definitely not here.
- `--detail`: just pick one level and don't provide the others at all
- `--detail 2`: I don't even remotely see a use case for that. It also exposes too many internals.
- `--orphan`: Too many internal details
- `--no-header`: Nope. If that's a problem for someone then they just have to use the JSON output.

wplapper:
I have moved part of the existing code as an extension of the ``restic restore``command. I have taken all existing options away and control the level of output detail via the ``--verbose`` option. I hope that fits better with your future development plans.

wplapper:
There is a need for such type of command / sub-command. See https://github.com/restic/restic/issues/5751.

cmd/restic/cmd_restore.go
```diff
@@ -1,16 +1,26 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"crypto/aes"
+	"encoding/binary"
+	"encoding/json"
+	"golang.org/x/crypto/poly1305"
```
MichaelEischer: No code outside `./internal/repository` may interact with this kind of repository internals (not up for discussion). After https://github.com/restic/restic/pull/21848 this is also most likely not possible anymore anyways. If it were _absolutely_ necessary, we could discuss exposing `pack.Size` (which performs the same calculations as you do below). But I don't see that this is the case.

cmd/restic/cmd_restore.go
```diff
@@ -315,3 +353,159 @@ func getXattrSelectFilter(opts RestoreOptions, printer progress.Printer) (func(x
...
+	if gopts.JSON {
+		return produceJSONOutput(packfilesSort, snapPacks, gopts.Term.OutputWriter())
+	}
+
+	produceTextOutput(packfilesSort, snapPacks, gopts, printer)
+	return nil
+}
+
+// produceJSONOutput generates JSON output
+func produceJSONOutput(packfilesSort []restic.ID, snapPacks map[restic.ID]PacklistInfo, stdout io.Writer) error {
```
MichaelEischer: Drop all output functionality beyond printing a plain list/array of pack file names. For the given use case of knowing which pack files need an Object Lock, it's neither important to learn the pack file type or size. The current approach to calculate the pack file size is also not compatible with my plans to encapsulate the restic internal anyways.

MichaelEischer (changes requested):
Sorry for the late review. It took me a while to figure out what I don't like about the current approach and how to address it. The main problem is that we have two use cases all mixed up:
- apply S3 object lock to all packs of some/all snapshots
- warm up files on cold storage

At first glance the warm-up use case just differs by only handling one snapshot at a time. But the difference is more fundamental: when only restoring part of a snapshot it is useful to only warm up that specific part. There's no equivalent for that in the S3 object lock case; one always has to protect the whole snapshot not just a subset.

The current PR is a better fit of the S3 object lock case, but that use case does not belong into the `restore` command. So different from what I've written initially, this code belongs somewhere else. We already have the `restic list packs` command. My feeling is that we should extend it to optionally accept a list of snapshots as parameters. That command only prints a plain text list of pack IDs, so there's also no need for extra information like type or size and also no need for a JSON output format.

As a result the `restic list packs <snapshots>` command will be usable to lock pack files of a snapshot and as a crutch to warm up the data packs for snapshot. Fine-grained warm ups for S3 already already implemented separately.

cmd/restic/cmd_restore.go
wplapper: code removed.

cmd/restic/cmd_restore.go
wplapper: no JSON output anymore.

MichaelEischer (approved):
LGTM. I've cleaned up the code a bit and simplified the test skaffolding. The check for the number of parameters passed to `list` is now also stricter again. Passing too many parameters now also returns an error.

PR #5400, opened 2025-05-29
docs: clarify ** wildcard must me between path separators

hollowhemlock:

Clarify ** wildcard needs to be placed between path separators.

[How do I specify a wildcard string for the end/trailing of a folder - Getting Help - restic forum](https://forum.restic.net/t/how-do-i-specify-a-wildcard-string-for-the-end-trailing-of-a-folder/9710/10?u=unequal8761)

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -346,9 +346,10 @@ A trailing ``/`` is ignored, a leading ``/`` anchors the pattern at the root dir
 This means, ``/bin`` matches ``/bin/bash`` but does not match ``/usr/bin/restic``.
 
 Regular wildcards cannot be used to match over the directory separator ``/``,
-e.g. ``b*ash`` matches ``/bin/bash`` but does not match ``/bin/ash``. For this,
-the special wildcard ``**`` can be used to match arbitrary sub-directories: The
-pattern ``foo/**/bar`` matches:
+e.g. ``b*ash`` matches ``/bin/bash`` but does not match ``/bin/ash``. To match
+across an arbitrary number of subdirectories, use the special ``**`` wildcard.
+The ``**`` must be positioned between path separators. The pattern 
+``foo/**/bar`` matches:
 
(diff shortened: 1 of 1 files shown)


doc/040_backup.rst
```diff
@@ -346,9 +346,10 @@ A trailing ``/`` is ignored, a leading ``/`` anchors the pattern at the root dir
 This means, ``/bin`` matches ``/bin/bash`` but does not match ``/usr/bin/restic``.
 
 Regular wildcards cannot be used to match over the directory separator ``/``,
-e.g. ``b*ash`` matches ``/bin/bash`` but does not match ``/bin/ash``. For this,
-the special wildcard ``**`` can be used to match arbitrary sub-directories: The
-pattern ``foo/**/bar`` matches:
+e.g. ``b*ash`` matches ``/bin/bash`` but does not match ``/bin/ash``. To match
+across an arbitrary number of subdirectories, use the special ``**`` wildcard.
+The ``**`` must be positioned between path separators (e.g. ``/**/``).
```
MichaelEischer:
```
(e.g. ``/**/``)
```
That sample is somewhat misleading as it should rather be written as `**`. Using `**` as prefix is rather redundant: `**/path` is equivalent to just `path`.

Similarly `path/**` is rarely what one wants. So the `foo/**/bar` example is already rather representative of the intended use case. So let's just drop the `e.g.` part.

MichaelEischer (changes requested):
.

hollowhemlock:
Amended the last commit to drop the e.g. part.

doc/040_backup.rst
hollowhemlock: Amended the last commit to drop the e.g. part.

rawtaz (approved):
Looks fine, thanks!

rawtaz (approved):

MichaelEischer (approved):
LGTM.

PR #5408, opened 2025-06-02
walker: fix error handling if tree cannot be loaded

MichaelEischer:
A tree that cannot be loaded is a fatal error when walking the tree. Thus, return the error and exit the tree walk. This could cause the `stats` command to crash when a tree blob is missing.

No. But there was a crash report in some forum thread a while ago.

Checklist
---------


greatroar (approved):

MichaelEischer (commented):
LGTM

PR #5410, opened 2025-06-03
repository/pack: Pad packs to further mitigate chunking attack

greatroar:

This is a backwards compatible way of introducing padding of packs. For each new pack, we determine if it needs padding with [Padmé](https://lbarman.ch/blog/padme/). If so, we add a data blob of all zeros. The user-visible result, aside from bigger pack files, is that restic cat will find these all-zero blobs.

Still to do is to teach prune about this change, because right now it will just try to remove those unused blobs, then add them back in when uploading new packs.

I think we should document clearly that pruning a repo with padding with an old version of restic will destroy the padding.

For #5328. Replaces #5407.

Checklist
---------

### DIFF
--- a/internal/repository/pack/pack.go
+++ b/internal/repository/pack/pack.go
@@ -9,11 +9,10 @@ import (
 	"math/bits"
 	"sync"
 
+	"github.com/restic/restic/internal/crypto"
 	"github.com/restic/restic/internal/debug"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/restic"
-
-	"github.com/restic/restic/internal/crypto"
 )
 
 // Packer is used to create a new Pack.
@@ -86,8 +85,8 @@ func (p *Packer) Finalize() error {
 	// in a compatible way by inserting a blob of all zeros. This only works,
 	// and probably only matters, for data packs.
 
-	padding := padmé(p.bytes)
-	if p.blobs[0].Type == restic.DataBlob && padding > 0 {
+	if p.blobs[0].Type == restic.DataBlob {
+		padding := padmé(p.bytes)
 		padding -= int(plainEntrySize)
 		padding = max(padding, 0)
 
--- a/internal/repository/pack/pack.go
+++ b/internal/repository/pack/pack.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"math/bits"
 	"sync"
 
 	"github.com/restic/restic/internal/debug"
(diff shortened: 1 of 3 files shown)


internal/repository/pack/pack.go
```diff
@@ -75,15 +79,31 @@ func (p *Packer) Finalize() error {
...
 
+	// Add some padding to mitigate an attack that looks for a known file
+	// by checking the lengths of blobs.
+	//
+	// The pack format was not designed with padding in mind, so we add it
+	// in a compatible way by inserting a blob of all zeros. This only works,
+	// and probably only matters, for data packs.
+
+	padding := padmé(p.bytes)
+	if p.blobs[0].Type == restic.DataBlob && padding > 0 {
```
aawsome: IMO this dangerous. Note that there is a difference between not writing a padding blob and potentially writing a 0-byte padding blob (which however could pose other problems): An adversary who is able to see the pack file can read the pack size header and therefore determine the number ob blobs within a pack. So, they get information if a padding was added or not.

aawsome:
FYI, I did a stochastical computation to get an estimate of the average overhead needed by Padmé and got an total overhead of about 1%.

I think this is problematic because:
- there are users where theoretical chunking attack does not apply, e.g. because they don't have adversaries with access to the backend in their threat model. For them this is 1% unneeded overhead. For this reason, the padding should be at least made optional.
- the only reason to use CDC is the idea that this saves space compared to e.g. fixed size chunking. And this is mainly the case when inserting bytes into existing files. Now, I think we can question a) how many times this situation happens in real user scenarios; and b) how many space is really saved in those scenarios - given that we typically also use compression. When we need to add 1% of padded space, the CDC not only needs to theoretically save space but IMO must proof that it saves in most real-day scenarios more than will be lost by adding the padding overhead...

About the 2nd point, one could rethink if a simple-and-stupid Padmé really is the best choice here or if we can maybe find a better padding which adds less overhead.

internal/repository/pack/pack.go
greatroar: I suppose you're right. Switched to always padding.

greatroar:
I find 1% acceptable, but we could round up to a constant block size instead, since storage systems will do that anyway. The common objections to doing so don't really affect restic, since pack sizes tend to be pretty uniform.

aawsome:
@greatroar Sorry for the late reply. The point is, as chunk sizes may vary between 512kiB and 8MiB, every rounding to less than 7.5 MiB - in theory - leaks some information. But so does also Padmé...
I have two suggestions how we could pad, but I'm not competent enough to judge if they are fine or which one is better:
- fill up to the next 64kiB resulting in 32kiB padding on average, alternatively 128kiB/64kiB. This is far less than the 160kiB (1%) of a standard 16 MiB pack and especially gives low padding percentage for even larger files
- take the size modulo 8MiB (or 7.5MiB) and use Padmé on this size for the padding size. This should result in ~80kiB average padding.

I personally must say that I think filling up to the next 64kiB is easy and hopefully safe enough.

aawsome:
Just another thing: You should also think about statistics, i.e. in which statistics the padding should or should not occur. I'm specially thinking about:
- `data_blobs`. Here IMO both adding/ignoring padding blobs makes sense; we have to decide what to do.
- `data_added_files`: I think padding should not contribute to this
- `data_addad_files_packed`: Here I think padding should contribute to this.

aawsome:
FWIW, I already implemented the padding blob for rustic in this PR: https://github.com/rustic-rs/rustic_core/pull/413

Issue #5391, opened 2025-05-11
First Backup on Windows : "panic: runtime error: invalid memory address or nil pointer dereference"


issackelly:
Output of `restic version`
--------------------------

(windows host, backup client)
PS C:\Users\Issac Kelly\Projects> restic version
restic 0.18.0 compiled with go1.24.1 on windows/amd64

(via apt-get install restic, backup target host)
```
[issac@kazad-dum pool]$ restic version
restic 0.16.4 compiled with go1.22.2 on linux/amd64
```

What backend/service did you use to store the repository?
---------------------------------------------------------

sftp on linux host on same network

Problem description / Steps to reproduce
----------------------------------------

I ran the following on the linux host:

```
[issac@kazad-dum pool]$ restic init --repo /pool/restic-repository/
enter password for new repository:
enter password again:
created restic repository fe5fc0e0f9 at /pool/restic-repository/
```

And on the windows machine I get the following:

```
PS C:\Users\Issac Kelly\Projects> restic -r sftp:issac@kaz:/pool/restic-repository backup C:\Users\"Issac Kelly"\ --iexclude-file="C:\Users\Issac Kelly\backup-excludes" --use-windows-vss --verbose 
Kelly\backup-excludes --use-windows-vss --verbose does not exist, skipping
enter password for repository: 
repository fe5fc0e0 opened (version 2, compression level auto)
no parent snapshot found, will read all files
[0:00]          0 index files loaded
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x30 pc=0x10ccb6d]

goroutine 46 [running]:
github.com/restic/restic/internal/archiver.(*Archiver).nodeFromFileInfo(0xc0002340a0, {0xc00010a300, 0x28}, {0xc00010a0f0, 0x28}, {0x23e64cedcf8?, 0xc00048e7e0?}, 0x0?)
        /restic/internal/archiver/archiver.go:268 +0x6d
github.com/restic/restic/internal/archiver.(*Archiver).dirPathToNode(0xc0002340a0, {0xc00010a300, 0x28}, {0xc00010a0f0, 0x28})
        /restic/internal/archiver/archiver.go:745 +0x1d8
github.com/restic/restic/internal/archiver.(*Archiver).saveTree(0xc0002340a0, {0x1f111b0, 0xc0002b02d0}, {0xc00010a300, 0x28}, 0xc000580580, 0x0, 0xc0005805c0)
[...]

PR #5421, opened 2025-06-12
backup: fix possible crash if directory disappears before metadata collection

MichaelEischer:

Fix a crash during the backup in nodeFromFileInfo in two cases:
- if a directory is deleted between restic stating it and trying to list  its directory content.
- when restic tries to list the parent directory of a backup target, but  the parent directory has been deleted.

A backup also no longer completely fails if a parent folder becomes inaccessible during the backup. This only becomes relevant with the above crash fix as this scenario most likely results in a crash in restic 0.18.0.
The general idea is to also handle errors for parent directories of backup directories in the same way as all other file access errors during a backup.

Partially addresses https://github.com/restic/restic/issues/5391 .

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5421
+++ b/changelog/unreleased/pull-5421
@@ -1,0 +1,8 @@
+Bugfix: Fix rare crash if directory is removed during backup
+
+In restic 0.18.0, the `backup` command could crash if a directory is removed
+inbetween reading its metadata and listing its directory content.
+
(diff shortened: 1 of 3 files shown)


MichaelEischer (commented):
LGTM.

MichaelEischer (commented):
LGTM again

PR #5424, opened 2025-06-20
Windows Backup Privilege Tweaks

Crazycatz00:

This PR enables backup semantics for file EA opens and attempts to enable `SeBackupPrivilege` when first beginning a backup.

When enabled, `SeBackupPrivilege` allows read-only opening that bypasses ACLs; this can permit a restricted account with the proper privilege to back up all users' files without changing file-system permissions. Go supports backup semantics for read-only opens since [405275 / v1.20](<https://go-review.googlesource.com/c/go/+/405275>) and restic currently enables the privilege to backup security descriptors, so such backups _mostly_ work already, with two issues:
1. Reading extended attributes fails for files when `SeBackupPrivilege` allows access which would have been denied without the privilege.
   This is because `windows.CreateFile` is used directly without the appropriate flag.
   (Directories already have the flag as they require it to get a handle at all.)
2. `SeBackupPrivilege` is only enabled upon reading the first security descriptor; accesses before this are subject to normal ACLs and may fail. This always affects the scanner, and the actual backup will error & skip initial items until one has permissive enough ACLs.

An example of both issues:
```console
> rem Running in cmd.exe as Administrator
> set RESTIC_PASSWORD=0000
> restic init -r ".\repo"
> mkdir ".\data"

> rem Create directory "b" accessible only to the SYSTEM account
> mkdir ".\data\b" && echo Test > ".\data\b\b.txt" && icacls ".\data\b" /inheritance:r /grant SYSTEM:F

> restic backup -r ".\repo" -nvv ".\data\b"
open repository
repository 5403b32f opened (version 2, compression level auto)
load index files
start scan on [.\data\b]
start backup on [.\data\b]
scan: openfile for readdirnames failed: open \\?\D:\Test\data\b: Access is denied.
scan finished in 0.003s: 0 files, 0 B
error: incomplete metadata for data\b\b.txt: get EA failed while opening file handle for path data\b\b.txt, with: Access is denied.
new       /data/b/b.txt, saved in 0.003s (8 B added)
new       /data/b/, saved in 0.004s (0 B added, 0 B stored, 412 B metadata)
new       /data/, saved in 0.007s (0 B added, 0 B stored, 429 B metadata)

Data Blobs:      1 new
Tree Blobs:      3 new
Would add to the repository: 1.778 KiB (1.380 KiB stored)

processed 1 files, 8 B in 0:00

> cd ".\data" && restic backup -r "..\repo" -nvv "b"
open repository
repository 5403b32f opened (version 2, compression level auto)
load index files
start scan on [b]
start backup on [b]
scan: openfile for readdirnames failed: open \\?\D:\Test\data\b: Access is denied.
scan finished in 0.001s: 0 files, 0 B
error: openfile for readdirnames failed: open \\?\D:\Test\data\b: Access is denied.
Fatal: unable to save snapshot: snapshot is empty
```

No.

Checklist
---------

- [x] I have added tests.
- ~[ ] I have added documentation for relevant changes (in the manual).~

### DIFF
--- a/changelog/unreleased/pull-5424
+++ b/changelog/unreleased/pull-5424
@@ -1,6 +1,6 @@
-Enhancement: Enable file system privileges before access
+Enhancement: Enable file system privileges on Windows before access
 
-Restic attempted to enable (Windows) file system privileges when
+Restic attempted to enable Windows file system privileges when
 reading or writing security descriptors - after potentially being wholly
 denied access to previous items. It also read file extended attributes without
 using the privilege, possibly missing them and producing errors.
--- a/changelog/unreleased/pull-5424
+++ b/changelog/unreleased/pull-5424
@@ -1,0 +1,11 @@
+Enhancement: Enable file system privileges before access
+
+Restic attempted to enable (Windows) file system privileges when
+reading or writing security descriptors - after potentially being wholly
+denied access to previous items. It also read file extended attributes without
+using the privilege, possibly missing them and producing errors.
+
+Restic now attempts to enable all file system privileges before any file
+access. It also requests extended attribute reads use the backup privilege.
+
+https://github.com/restic/restic/pull/5424
--- a/internal/fs/priv_windows_test.go
+++ b/internal/fs/priv_windows_test.go
@@ -1,0 +1,61 @@
+//go:build windows
+// +build windows
+
+package fs
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/restic/restic/internal/errors"
+	"github.com/restic/restic/internal/test"
+	"golang.org/x/sys/windows"
+)
+
+func TestBackupPrivilegeBypassACL(t *testing.T) {
+	testPath := testGetRestrictedFilePath(t)
+
+	// Read-only Open/OpenFile should automatically use FILE_FLAG_BACKUP_SEMANTICS since Go v1.20.
+	testfile, err := os.Open(testPath)
+	test.OK(t, errors.Wrapf(err, "failed to open file for reading: %s", testPath))
+	test.OK(t, testfile.Close())
+}
+
+func TestRestorePrivilegeBypassACL(t *testing.T) {
+	testPath := testGetRestrictedFilePath(t)
+
+	// Writable OpenFile needs explicit FILE_FLAG_BACKUP_SEMANTICS.
+	// Go with issue #73676 merged would allow: os.OpenFile(testPath, os.O_WRONLY|windows.O_FILE_FLAG_BACKUP_SEMANTICS, 0)
+	utf16Path := windows.StringToUTF16Ptr(testPath)
+	handle, err := windows.CreateFile(utf16Path, windows.GENERIC_WRITE, 0, nil, windows.OPEN_EXISTING, windows.FILE_ATTRIBUTE_NORMAL|windows.FILE_FLAG_BACKUP_SEMANTICS, 0)
+	test.OK(t, errors.Wrapf(err, "failed to open file for writing: %s", testPath))
+	test.OK(t, windows.Close(handle))
+}
+
+func testGetRestrictedFilePath(t *testing.T) string {
+	// Non-admin is unlikely to have needed privileges.
+	isAdmin, err := isAdmin()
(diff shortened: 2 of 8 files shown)


Crazycatz00:
Thinking on this, *should* restic automatically try to enable privileges that cause it to bypass access permissions? Should it be enabled via flag, like VSS is by `--use-fs-snapshot`?
It's using the privilege for its intended purpose (backup), but it may be unexpected that restic can access files the running (administrator) user normally can't.

jtru:
This is a great feature that I would very much like to have asap, but I *think* it would be better to have it as an explicit opt-in. Otherwise, I expect all kinds of "endpoint protection" and anti-virus snake oil  to trip over `restic` executables in the near future, which would be a real hassle. Better let users explicitly enable the backup privilege in case they know they may actually attain it.

ProactiveServices:
> Thinking on this, _should_ restic automatically try to enable privileges that cause it to bypass access permissions? 

It's not bypassing anything, it's using the access paradigms as they were intended and I'd say it is not violating the principle of least surprise. Backup software is the intended use of this privilege.

cmd/restic/cmd_backup.go
```diff
@@ -482,6 +482,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts GlobalOptions, ter
 	var err error
 
 	if runtime.GOOS == "windows" {
+		fs.EnableBackupPrivilege()
```
MichaelEischer: This doesn't compile on non-windows systems. I'd prefer to let `internal/fs/fs_local.go` handle the privilege request automatically. It should request the privileges when the fs instance is created.

internal/fs/sd_windows.go
```diff
@@ -42,7 +42,7 @@ var lowRestoreSecurityFlags windows.SECURITY_INFORMATION = windows.DACL_SECURITY
 // This needs admin permissions or SeBackupPrivilege for getting the full SD.
 // If there are no admin permissions, only the current user's owner, group and DACL will be got.
 func getSecurityDescriptor(filePath string) (securityDescriptor *[]byte, err error) {
-	onceBackup.Do(enableBackupPrivilege)
+	onceBackup.Do(EnableBackupPrivilege)
```
MichaelEischer: If `EnableBackupPrivilege` is already called by the backup command, then this call here is redundant.

MichaelEischer (changes requested):
I'd like to keep the privilege management hidden in the fs package, see below for a suggestion how to achieve that.

Crazycatz00:
This should be a bit more proper. Contained to the `internal/fs` module, running just once on `init()`, and made sure it builds for at least Linux via `GOOS`. Also rebased to latest and added two tests to verify the behavior.

internal/fs/priv_windows_test.go
```diff
@@ -0,0 +1,61 @@
...
+	// Create temporary file.
+	tempDir := t.TempDir()
+	testPath := filepath.Join(tempDir, "testfile.txt")
+
+	testfile, err := os.Create(testPath)
+	test.OK(t, errors.Wrapf(err, "failed to create temporary file: %s", testPath))
+	test.OK(t, testfile.Close())
+
+	// Set restricted permissions.
+	sd, err := windows.SecurityDescriptorFromString("D:PAI(D;;FRFWFX;;;WD)(A;;SD;;;WD)")
```
MichaelEischer, line 54: Please add a short comment what the security descriptor does. I just see a wild collection of characters ^^

changelog/unreleased/pull-5424
```diff
@@ -0,0 +1,11 @@
+Enhancement: Enable file system privileges before access
```
MichaelEischer:
```suggestion
Enhancement: Enable file system privileges on Windows before access
```

MichaelEischer (changes requested):
LGTM, I just have two small nits

internal/fs/priv_windows_test.go
Crazycatz00: Actually had this commented at first, then I saw the base64 blobs the security descriptor tests uses... 😅
Re-added.

MichaelEischer (approved):
LGTM

PR #5448, opened 2025-07-14
feat: add integrated nice and ionice options for docker

ferringb:
The intended usage here is to basically kick restic as a background "do it, but don't bother my normal load" process.

This allows passing the following environment variables in to influence scheduling:

- NICE: usual CPU nice.  Defaults to 0.
- IONICE_CLASS: usual ionice class.  Note that setting realtime requires CAP_SYS_ADMIN.  Also note the actual ionice default is "none".
- IONICE_PRIORITY: set the priority within the given class.

This simplifies setting CPU and IO priority, doing it directly via the container.  The expectation is any user of this- as I wished- want to kick restic into the background.

This is possible already with the given containers, however this requires entrypoint override, or manipulating docker CLI args (or k8s args) to achieve it.  It's doable, but if the common case is just "do it in the background", providing environment variables to do it is the lowest friction for user.

This integrates 3 environment variables, described above.  Further detail:

- NICE: this is always passed into the `nice` command, but defaulted to 0- no change.
- IONICE_CLASS: this is more complex; it's unset.  If set, it must be a valid `ionice` class.  Realtime is not supported without the user adding the SYS_NICE capability.  I consider that an exceptional case rather than the expected common case of "just background restic".  Note also that if the value is unset, we *must not touch the class*.  The default class for processes is "none", but busybox ionice refuses `ionice -c0 -nN` since a priority for no class makes no sense.
- IONICE_PRIORITY: if a class is given, this value is honored as an argument to `ionice`.  It defaults to 4- which should be the default priority for any given class.

The `nice` suggestion has appeared in multiple tickets where the end users CPU scheduler was making non-optimal choices; I don't think I've seen the `ionice` suggestion however.

To my knowledge, there is no explicit asking for integrating controllable `nice` or `ionice` however.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5448
+++ b/changelog/unreleased/pull-5448
@@ -1,0 +1,11 @@
+Enhancement: Allow nice and ionice configuration for restic containers
+
+The official restic docker now supports the following environment variables:
+
+`NICE`: set the desired nice scheduling.  See `man nice`.
+`IONICE_CLASS`: set the desired I/O scheduling class.  See `man ionice`.  Note that real time support requires the invoker to manually add the `SYS_ADMIN` capability.
+`IONICE_PRIORITY`: set the prioritization for ionice in the given `IONICE_CLASS`.  This does nothing without `IONICE_CLASS`, but defaults to `4` (no priority, no penalties).
+
+See https://restic.readthedocs.io/en/stable/020_installation.html#docker-container for further details.
+
+https://github.com/restic/restic/pull/5448
--- a/changelog/unreleased/pull-5448
+++ b/changelog/unreleased/pull-5448
@@ -3,7 +3,7 @@ Enhancement: Allow nice and ionice configuration for restic containers
 The official restic docker now supports the following environment variables:
 
 `NICE`: set the desired nice scheduling.  See `man nice`.
-`IONICE_CLASS`: set the desired I/O scheduling class.  See `man ionice`.  Note that real time support requires the invoker to manually add the `SYS_ADMIN` capability.
+`IONICE_CLASS`: set the desired I/O scheduling class.  See `man ionice`.  Note that real time support requires the invoker to manually add the `SYS_NICE` capability.
 `IONICE_PRIORITY`: set the prioritization for ionice in the given `IONICE_CLASS`.  This does nothing without `IONICE_CLASS`, but defaults to `4` (no priority, no penalties).
 
 See https://restic.readthedocs.io/en/stable/020_installation.html#docker-container for further details.
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -289,6 +289,23 @@ Restic relies on the hostname for various operations. Make sure to set a static
 hostname using `--hostname` when creating a Docker container, otherwise Docker
 will assign a random hostname each time.
 
+The container additionally honors traditional ``nice`` `(man page) <https://man7.org/linux/man-pages/man1/nice.1.html>`__ and ``ionice`` `(man page) <https://man7.org/linux/man-pages/man1/ionice.1.html#OPTIONS>`__ directives via the following environment variables.
+This allows restic to be scheduled as a background process to reduce latency for the system while running.
+
+* ``NICE``: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment- standard CPU priority.
+* ``IONICE_CLASS``: If set, this adjusts the IO prioritization of the restic process via ``ionice -c`` for the available classes.
+  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see capabilities manpage <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow assuming this IO class.
+* ``IONICE_PRIORITY``: this defaults to 4 (no prioritization or penalties); this is the prioritization within the given ``IONICE_CLASS``.
+
+The following example runs restic such that other CPU and IO requests have higher priority. This effectively perturbs the system as minimally as possible while ``restic`` runs.
+
+.. code-block:: console
+
+    # docker run -e NICE=20 -e IONICE_CLASS=2 -e IONICE_PRIORITY=7 ghcr.io/restic/restic
+
+*Remember* that this invocation is explicitly telling your CPU and IO scheduler to deprioritize restic.  This typically will result in a longer runtime.  For a system with heavy load, this can be drastically longer.
+
+
 From Source
 ***********
 
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -292,10 +292,10 @@ will assign a random hostname each time.
 The container additionally honors traditional ``nice`` `(man page) <https://man7.org/linux/man-pages/man1/nice.1.html>`__ and ``ionice`` `(man page) <https://man7.org/linux/man-pages/man1/ionice.1.html#OPTIONS>`__ directives via the following environment variables.
 This allows restic to be scheduled as a background process to reduce latency for the system while running.
 
-* ``NICE``: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment- standard CPU priority.
+* ``NICE``: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment - standard CPU priority.
 * ``IONICE_CLASS``: If set, this adjusts the IO prioritization of the restic process via ``ionice -c`` for the available classes.
-  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see capabilities manpage <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow assuming this IO class.
-* ``IONICE_PRIORITY``: this defaults to 4 (no prioritization or penalties); this is the prioritization within the given ``IONICE_CLASS``.
+  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see capabilities manpage <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow using this IO class.
+* ``IONICE_PRIORITY``: This defaults to 4 (no prioritization or penalties); this is the prioritization within the given ``IONICE_CLASS``.
 
 The following example runs restic such that other CPU and IO requests have higher priority. This effectively perturbs the system as minimally as possible while ``restic`` runs.
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,11 +8,17 @@ RUN go mod download
 
 COPY . .
 RUN go run build.go
-
 FROM alpine:latest AS restic
-
 RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
+COPY ./docker/entrypoint.sh /entrypoint.sh
 
-ENTRYPOINT ["/usr/bin/restic"]
+# IO class default is "none"- 0, however busybox reject ionice `-c0 -n<something>`
+# since priority has no meaning for no scheduler.
+# Thus the set logic below is necessary.
+ENV IONICE_CLASS=
+ENV IONICE_PRIORITY=4
+ENV NICE=0
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@ COPY ./docker/entrypoint.sh /entrypoint.sh
 
 # IO class default is "none"- 0, however busybox reject ionice `-c0 -n<something>`
 # since priority has no meaning for no scheduler.
-# Thus the set logic below is necessary.
+# Thus the entrypoint script below is necessary
 ENV IONICE_CLASS=
 ENV IONICE_PRIORITY=4
 ENV NICE=0
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -1,6 +1,6 @@
 # the official binaries are cross-built from Linux running on an AMD64 host
 # other architectures also seem to generate identical binaries but stay on the safe side
-FROM --platform=linux/amd64 restic/builder:latest as helper
+FROM --platform=linux/amd64 restic/builder:latest AS helper
 
 ARG TARGETOS
 ARG TARGETARCH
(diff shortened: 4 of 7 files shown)


ferringb:
Note: I inlined the nice/ionice into a `/bin/sh -c` invocation.  Yeah, it's mildly dirty; it can be moved into an explicit entrypoint script, I just have a dislike for that approach since they inevitably become overly complex requiring the user to pull the image and extract the entrypoint script to look at it.

Moving that `-c` blurb into a script would reduce the annoyance of having to escape `"`, but that's about all it adds.  I'll do so if you wish.

docker/Dockerfile.release
```diff
@@ -1,6 +1,6 @@
 # the official binaries are cross-built from Linux running on an AMD64 host
 # other architectures also seem to generate identical binaries but stay on the safe side
-FROM --platform=linux/amd64 restic/builder:latest as helper
+FROM --platform=linux/amd64 restic/builder:latest AS helper
```
ferringb, line 3: This is unrelated, I'm just resolving a docker build complaint while in here.  Note `Dockerfile` has this change already.

docker/Dockerfile
```diff
@@ -15,4 +15,12 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
-ENTRYPOINT ["/usr/bin/restic"]
+ENV IONICE_CLASS=
```
MichaelEischer, line 20: Judging from the man page the default is `0`. So let's set this by default and then get rid of the magic `set --` stuff.

docker/Dockerfile
```diff
@@ -15,4 +15,12 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
...
+ENV IONICE_PRIORITY=4
+ENV NICE=0
+
+ENTRYPOINT ["/bin/sh", "-ec", "\
+    set -- nice -n\"${NICE:-0}\" /usr/bin/restic; \
```
MichaelEischer: Won't this drop the parameters passed to restic?

docker/Dockerfile.release
```diff
@@ -12,7 +12,16 @@ RUN mv /output/restic_${TARGETOS}_${TARGETARCH} /output/restic
...
+
 COPY --from=helper /output/restic /usr/bin
 RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
-ENTRYPOINT ["/usr/bin/restic"]
+ENV IONICE_CLASS=
+ENV IONICE_PRIORITY=4
+ENV NICE=0
+
+ENTRYPOINT ["/bin/sh", "-xec", "\
```
MichaelEischer: Remove the `-x`.

doc/020_installation.rst
```diff
@@ -277,6 +277,23 @@ Restic relies on the hostname for various operations. Make sure to set a static
...
+The usage of this is principally to allow restic to be scheduled as a background process reducing latency for the system while running.
+
+* NICE: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment- standard CPU priority.
+* IONICE_CLASS: If set, this adjusts the IO prioritization of the restic process via ``ionice -c`` for the available classes.
+  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see here <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow assuming this IO class.
+* IONICE_PRIORITY: this defaults to 4 (no prioritization or penalties); this is the prioritization within the given `IONICE_CLASS`.
+
+The following example runs restic such that other CPU and IO requests have higher priority. Effectively perturb the system as minimally as possible while getting the ``restic`` operation finished.
```
MichaelEischer:
```suggestion
The following example runs restic such that other CPU and IO requests have higher priority. This effectively perturbs the system as minimally as possible while getting the ``restic`` operation finished.
```

changelog/unreleased/pull-5448
```diff
@@ -0,0 +1,11 @@
+Enhancement: Allow nice and ionice configuration for restic containers
+
+ghcr.io/restic/restic containers now support the following environment variables:
```
MichaelEischer: `Dockerfile.release` is also used for the docker.io/restic/restic container.

MichaelEischer (changes requested):

ferringb:
*cough* there's a lot of WTF's in that PR of mine. I'm just going to assume I was half awake at the time.  Everything you flagged is valid/correct- pardon.

One thing I do recall- I was wary of CAP_SYS_NICE interaction, specifically if it was dropped, thus the set trickery.  Checking the manage page, CAP_SYS_NICE is required to screw with other processes, but isn't required for downgrading the process you're launching- whether ionice or nice.  It's only required for increasing the priority.  So yeah, lot of WTFs.

docker/Dockerfile
ferringb: In reworking the PR, I found why I did the set crap- busybox refuses an invocation like `ionice -c 0 -n 4 sh`

repro:
```
ferringb@hrair ~/restic $ docker run -it alpine:latest ionice -c 0 -n 4 sh
ionice: ioprio_set: Invalid argument
ferringb@hrair ~/restic $ echo $?
1
```
Non busybox at least still executes it, albeit with a complaint.
```
ferringb@hrair ~/restic $ ionice -c 0 -n 4 sh
ionice: données de classe ignorées pour la priorité « aucune »
sh-5.3$ 
```

So the set crap has to stay, but I'll make it cleaner.

ferringb: I've left this particular logic in, but adjusted the changelog and the commit, and added an explicit comment in to explain 'why' the set logic is necessary.

docker/Dockerfile
```diff
@@ -15,4 +15,15 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
 COPY --from=builder /go/src/github.com/restic/restic/restic /usr/bin
 
-ENTRYPOINT ["/usr/bin/restic"]
+# IO class default is "none"- 0, however busybox ionice `-c0 -n<something>`
```
MichaelEischer:
```suggestion
# IO class default is "none"- 0, however busybox rejects `ionice -c0 -n<something>`
```

doc/020_installation.rst
```diff
@@ -287,6 +287,23 @@ Restic relies on the hostname for various operations. Make sure to set a static
 hostname using `--hostname` when creating a Docker container, otherwise Docker
 will assign a random hostname each time.
 
+The container additionally honors traditional ``nice`` `(man page) <https://man7.org/linux/man-pages/man1/nice.1.html>`__ and ``ionice`` `(man page) <https://man7.org/linux/man-pages/man1/ionice.1.html#OPTIONS>`__ directives via the following environment variables.
+The usage of this is principally to allow restic to be scheduled as a background process reducing latency for the system while running.
```
MichaelEischer:
```suggestion
This allows restic to be scheduled as a background process to reduce latency for the system while running.
```

doc/020_installation.rst
```diff
@@ -287,6 +287,23 @@ Restic relies on the hostname for various operations. Make sure to set a static
...
+
+* NICE: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment- standard CPU priority.
+* IONICE_CLASS: If set, this adjusts the IO prioritization of the restic process via ``ionice -c`` for the available classes.
+  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see here <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow assuming this IO class.
+* IONICE_PRIORITY: this defaults to 4 (no prioritization or penalties); this is the prioritization within the given `IONICE_CLASS`.
```
MichaelEischer:
```suggestion
* ``NICE``: If set, this adjusts the CPU prioritization via ``nice -n``. This defaults to no adjustment- standard CPU priority.
* ``IONICE_CLASS``: If set, this adjusts the IO prioritization of the restic process via ``ionice -c`` for the available classes.
  Note: if you attempt to set `real-time`, you *will* have to add the ``SYS_NICE`` capability (`see capabilities manpage <https://man7.org/linux/man-pages/man7/capabilities.7.html#DESCRIPTION>`__) to allow assuming this IO class.
* ``IONICE_PRIORITY``: this defaults to 4 (no prioritization or penalties); this is the prioritization within the given ``IONICE_CLASS``.
```

doc/020_installation.rst
```diff
@@ -287,6 +287,23 @@ Restic relies on the hostname for various operations. Make sure to set a static
...
+
+The following example runs restic such that other CPU and IO requests have higher priority. This effectively perturbs the system as minimally as possible while getting the ``restic`` operation finished.
```
MichaelEischer:
```suggestion
The following example runs restic such that other CPU and IO requests have higher priority. This effectively perturbs the system as minimally as possible while ``restic`` runs.
```

docker/Dockerfile.release
```diff
@@ -21,4 +21,15 @@ LABEL org.opencontainers.image.source="https://github.com/restic/restic"
 COPY --from=helper /output/restic /usr/bin
 RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
 
-ENTRYPOINT ["/usr/bin/restic"]
+# IO class default is "none"- 0, however busybox ionice `-c0 -n<something>`
```
MichaelEischer:
```suggestion
# IO class default is "none"- 0, however busybox rejects `ionice -c0 -n<something>`
```

changelog/unreleased/pull-5448
MichaelEischer:
```suggestion
The official restic docker now supports the following environment variables:
```

changelog/unreleased/pull-5448
```diff
@@ -0,0 +1,11 @@
...
+
+NICE: set the desired nice scheduling.  See `man nice`.
+IONICE_CLASS: set the desired I/O scheduling class.  See `man ionice`.  Note that real time support requires the invoker to manually add the `SYS_ADMIN` capability.
+IONICE_PRIORITY: set the prioritization for ionice in the given `IONICE_CLASS`.  This does nothing without `IONICE_CLASS`, but defaults to `4` (no priority, no penalties).
```
MichaelEischer:
```suggestion
- `NICE`: set the desired nice scheduling.  See `man nice`.
- `IONICE_CLASS`: set the desired I/O scheduling class.  See `man ionice`.  Note that real time support requires the invoker to manually add the `SYS_ADMIN` capability.
- `IONICE_PRIORITY`: set the prioritization for ionice in the given `IONICE_CLASS`.  This does nothing without `IONICE_CLASS`, but defaults to `4` (no priority, no penalties).
```

MichaelEischer (changes requested):
I have a few more nits, see suggestions below.

> cough there's a lot of WTF's in that PR of mine. I'm just going to assume I was half awake at the time. Everything you flagged is valid/correct- pardon.

No worries, it's the job of a PR review to find those issues

docker/Dockerfile
```diff
@@ -15,4 +15,15 @@ RUN apk add --no-cache ca-certificates fuse openssh-client tzdata jq
...
+# Thus the set logic below is necessary.
+ENV IONICE_CLASS=
+ENV IONICE_PRIORITY=4
+ENV NICE=0
+
+ENTRYPOINT ["/bin/sh", "-ec", "\
+    set -- nice -n \"${NICE}\" /usr/bin/restic \"$@\";  \
+    [ -n \"${IONICE_CLASS}\" ] && set -- ionice -c \"${IONICE_CLASS}\" -n \"${IONICE_PRIORITY}\" \"$@\"; \
+    exec \"$@\"; \
+    "]
```
MichaelEischer: This still loses the command arguments:
```
docker/build.sh
docker run --rm docker.io/restic/restic:latest version
```

Prints the help output and not the version.

```
docker run --rm docker.io/restic/restic:latest version version
```

Prints the version, but that's obviously not how this should look like. So, you'll have to add `$0` in addition to `$@`.

At this point I'd prefer to just add an entrypoint script that handles everything necessary, the current inline calls to `sh` is already nearly unreadable.

doc/020_installation.rst
ferringb: For my knowledge- why are you using ``` `` ``` here for literals (`NICE` for example), but in changelog entries usiong ``` `NICE` ``` ?  I thought ``` `` ``` was for commands.  Is there something particular to this render pathway?

docker/Dockerfile
ferringb: Funny sidenote, if one is testing these changes, it helps to not use `--help` as what you were passing to validate that arguments passed through.  Go fucking figure.  I'm moving all of this to script per your request, although for an FYI, the sole thing wrong with that is I wasn't passing a last arg of `--` to sh.

ferringb:
All english fixes- which I suggest you reverify, I'm native, but I just don't see that shit- should be resolved.  As to the arg passing, as I mentioned, I was using `--help` so that test was... stupid.  To save some time:

main build:
```
ferringb@hrair ~/restic $ docker/build.sh --no-cache &> /dev/null && echo "built"
built
ferringb@hrair ~/restic $ docker run -e IONICE_CLASS=2 -e IONICE_PRIORITY=7 -e NICE=5 -it docker.io/restic/restic:latest backup
Fatal: nothing to backup, please specify source files/dirs
ferringb@hrair ~/restic $ docker run --entrypoint /bin/sh -e IONICE_CLASS=2 -e IONICE_PRIORITY=7 -e NICE=5 -it docker.io/restic/restic:latest -x /entrypoint.sh backup
+ set -- /usr/bin/restic backup
+ '[' -n 2 ]
+ set -- ionice -c 2 -n 7 /usr/bin/restic backup
+ exec nice -n 5 ionice -c 2 -n 7 /usr/bin/restic backup
Fatal: nothing to backup, please specify source files/dirs
```
release
```
ferringb@hrair ~/restic $ docker build -t test -f docker/Dockerfile.release . --no-cache  &> /dev/null && echo "built"
built
ferringb@hrair ~/restic $ docker run -e IONICE_CLASS=2 -e IONICE_PRIORITY=7 -e NICE=5 -it test backup
Fatal: nothing to backup, please specify source files/dirs
ferringb@hrair ~/restic $ docker run --entrypoint /bin/sh -e IONICE_CLASS=2 -e IONICE_PRIORITY=7 -e NICE=5 -it test -x /entrypoint.sh backup
+ set -- /usr/bin/restic backup
+ '[' -n 2 ]
+ set -- ionice -c 2 -n 7 /usr/bin/restic backup
+ exec nice -n 5 ionice -c 2 -n 7 /usr/bin/restic backup
Fatal: nothing to backup, please specify source files/dirs
```

MichaelEischer (approved):
LGTM. I've fixed a few more typos, rebased the PR and adjusted the recently added .dockerignore file.

Please take a look.

ferringb:
LGTM.

Issue #4509, opened 2023-10-09
Backup file owner and ACL on windows


Sch-Tim:
Output of `restic version`
--------------------------

restic 0.16.0 compiled with go1.20.6 on windows/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

According to the docs, restic isn't able to backup/restore file owners and ACL on windows. After a simple test, I can confirm that. 
I think that this would be an important and useful feature. 

The reasons for that are that as long as you aren't storing only user data on a single user machine, restoring after a disaster could be more time consuming than it needs to be as every user has to restore the data on himself. 
If it could be done by the person who reinstalls the system, it's a one step and even possibly automatic process. 

The other use case would be to restore data owned by the system or administrator user. 

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I didn't try anything since the documentation says it's not possible. 

The question how to keep this kind of information when copying a file is not new and the standard answer seems to be using "robocopy.exe". 

I have to admit that I'm rather new to go, but I tried to understand the backup process a bit and it seems that the file isn't simply copied into the repository but somehow bitwise read and rewritten, so using robocopy won't be a solution. 

Extra information
---------------------------------------------------------
This is not meant as a simple feature request. I'd like to participate to this project and out of egoistic reasons this could be my first try. 
My question is if there are already ideas how one could tackle this or even failed attempts. 
Looking through the open issues, I couldn't find anything pointing in that direction.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

I'm using restic on my Linux server since a while and I'm pretty happy with it. 
Now I want to convert from my current somehow sh'tty backup software on windows to restic and I have some trouble to adjust it to my needs.

hgraeber:
I am interested in this, too, and willing to help. Here some hints:

My idea is to store the SDDL for each directory and file item (File taken from the small-repo test data file):
```pwsh
❯ get-acl .\testdir\bar | Format-List -Property PSPath, Sddl
PSPath : Microsoft.PowerShell.Core\FileSystem::C:\Users\Myself\restore\testdir\bar
Sddl   : O:S-1-5-21-2235380413-24826354-127642980-1104G:DUD:AI(A;ID;FA;;;S-1-5-21-2235480413-24
         826354-127632980-1104)(A;ID;FA;;;SY)(A;ID;FA;;;BA)
```
The SDDL contains owners (as SID) and the ACL in a compact text form.

[...]

PR #5449, opened 2025-07-16
feat(internal/fs/node): Restore ownership by name

### LINKED ISSUE #3572: Also support restoring ownership by name
Output of `restic version`
--------------------------

restic 0.12.0 compiled with go1.16.6 on linux/amd64

It would be great, if `restic restore` would support an option for restoring ownership information from the symbolic user/group names instead of just restoring the uid/gid.

Also, it would be extra great, if restic would warn, by default, if the uid/gids of restored files map to different user/group names on the target system than on the backed up system.

I was doing a disaster recovery of a system, i.e. I installed it from scratch, installed and configured some services via a playbook and restored some data file with restic. It turns out that many system users such as 'nginx' have dynamically assigned uid/gids on modern Linux distributions. Thus, they likely change if the system is re-deployed from scratch.

Since restic stores/restored raw uids/gids, I had to chown the restored files which was quite tedious.

See also https://forum.restic.net/t/restore-with-numeric-uid-or-with-username/3811/2 for a similar description of this use case.

Restic recently helped me backing up some old Intel Core 1 host with just 2 GB RAM with reasonable speed which was nice.

provokateurin:

When doing a restore the UID and GID for the chown are now looked up on the current system, so that the ownership will actually match the expected user/group names.

I am not sure if this needs extra tests, since the existing ones all pass which at least validates that it didn't break existing behavior.
I am also not sure if this should be behind a CLI flag and if this needs additional documentation.
Please let me know if any of those things are needed, I'm happy to put more work into this feature.

This is my first PR to restic, but I've been a very happy user so far and this issues has been biting me quite a bit. I went so far to even create a script that would be put into every snapshot, in order to correctly restore the ownership, but it would be great to have a solution directly in restic.

Fixes https://github.com/restic/restic/issues/3572

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3572
+++ b/changelog/unreleased/issue-3572
@@ -1,0 +1,9 @@
+Enhancement: Support restoring ownership by name on UNIX systems
+
+Restic restore used to restore file ownership on UNIX systems by UID and GID.
+It now allows restoring the file ownership by user name and group name with `--ownership-by-name`.
+This allows restoring snapshots on a system where the UID/GID are not the same as they were on the system where the snapshot was created.
+However it does not include support for POSIX ACLs, which are still restored by their numeric value.
+
+https://github.com/restic/restic/issues/3572
+https://github.com/restic/restic/pull/5449
--- a/cmd/restic/cmd_restore.go
+++ b/cmd/restic/cmd_restore.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"path/filepath"
+	"runtime"
 	"time"
 
 	"github.com/restic/restic/internal/data"
@@ -33,6 +34,8 @@ repository.
 
 To only restore a specific subfolder, you can use the "snapshotID:subfolder"
 syntax, where "subfolder" is a path within the snapshot.
+
+POSIX ACLs are always restored by their numeric value, while file ownership can optionally be restored by name instead of numeric value.
 
 EXIT STATUS
 ===========
@@ -68,6 +71,7 @@ type RestoreOptions struct {
 	Delete              bool
 	ExcludeXattrPattern []string
 	IncludeXattrPattern []string
+	OwnershipByName     bool
 }
 
 func (opts *RestoreOptions) AddFlags(f *pflag.FlagSet) {
@@ -85,6 +89,9 @@ func (opts *RestoreOptions) AddFlags(f *pflag.FlagSet) {
 	f.BoolVar(&opts.Verify, "verify", false, "verify restored files content")
 	f.Var(&opts.Overwrite, "overwrite", "overwrite behavior, one of (always|if-changed|if-newer|never)")
 	f.BoolVar(&opts.Delete, "delete", false, "delete files from target directory if they do not exist in snapshot. Use '--dry-run -vv' to check what would be deleted")
+	if runtime.GOOS != "windows" {
+		f.BoolVar(&opts.OwnershipByName, "ownership-by-name", false, "restore file ownership by user name and group name (except POSIX ACLs)")
+	}
 }
 
 func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
@@ -164,11 +171,12 @@ func runRestore(ctx context.Context, opts RestoreOptions, gopts GlobalOptions,
 
 	progress := restoreui.NewProgress(printer, ui.CalculateProgressInterval(!gopts.Quiet, gopts.JSON, term.CanUpdateStatus()))
 	res := restorer.NewRestorer(repo, sn, restorer.Options{
-		DryRun:    opts.DryRun,
-		Sparse:    opts.Sparse,
-		Progress:  progress,
-		Overwrite: opts.Overwrite,
-		Delete:    opts.Delete,
+		DryRun:          opts.DryRun,
+		Sparse:          opts.Sparse,
+		Progress:        progress,
+		Overwrite:       opts.Overwrite,
+		Delete:          opts.Delete,
+		OwnershipByName: opts.OwnershipByName,
 	})
 
 	totalErrors := 0
--- a/internal/fs/node.go
+++ b/internal/fs/node.go
@@ -131,6 +131,39 @@ func lookupUsername(uid uint32) string {
 }
 
 var (
+	userNameLookupCache      = make(map[string]uint32)
+	userNameLookupCacheMutex = sync.RWMutex{}
+)
+
+// Cached uid lookup by user name. Returns 0 when no id can be found.
+//
+//nolint:revive // captialization is correct as is
+func lookupUid(userName string) uint32 {
+	userNameLookupCacheMutex.RLock()
+	uid, ok := userNameLookupCache[userName]
+	userNameLookupCacheMutex.RUnlock()
+
+	if ok {
+		return uid
+	}
+
+	u, err := user.Lookup(userName)
+	if err == nil {
+		var s int
+		s, err = strconv.Atoi(u.Uid)
+		if err == nil {
+			uid = uint32(s)
+		}
+	}
+
+	userNameLookupCacheMutex.Lock()
+	userNameLookupCache[userName] = uid
+	userNameLookupCacheMutex.Unlock()
+
+	return uid
+}
+
+var (
 	gidLookupCache      = make(map[uint32]string)
 	gidLookupCacheMutex = sync.RWMutex{}
 )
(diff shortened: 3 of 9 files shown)


zumm:
I think it should be optinal. Setting it as default is breaking change.

Also any chance we could get it for Windows?

provokateurin:
While I can see why you say it is a breaking change, I don't think it is one since the original behavior should stay on most systems where the uid and gid mappings are the same. It just fixes the behavior for systems where this isn't the case.

I don't know about Windows, but it didn't have any chown-like thing before on a restore. Also I just don't know much about Windows in general, do they even have the concept of file ownership with uid and gid?

zumm:
There is concept of Security Descriptors on Windows. Support was introduced in #4509 (and this is the example why features should be optional even if they seems like good default).

provokateurin:
@zumm I added a CLI flag to optionally enable the new behavior.

MichaelEischer:
> There is concept of Security Descriptors on Windows. Support was introduced in #4509 (and this is the example why features should be optional even if they seems like good default).

Windows will have to be handled separately. The security descriptors rely on unique IDs which can't be easily translated between systems.

cmd/restic/cmd_restore.go
```diff
@@ -86,6 +87,7 @@ func (opts *RestoreOptions) AddFlags(f *pflag.FlagSet) {
 	f.BoolVar(&opts.Verify, "verify", false, "verify restored files content")
 	f.Var(&opts.Overwrite, "overwrite", "overwrite behavior, one of (always|if-changed|if-newer|never)")
 	f.BoolVar(&opts.Delete, "delete", false, "delete files from target directory if they do not exist in snapshot. Use '--dry-run -vv' to check what would be deleted")
+	f.BoolVar(&opts.OwnershipByName, "ownership-by-name", false, "restore file ownership by uid and gid")
```
MichaelEischer: That option doesn't work on Windows for now and thus shouldn't be offered there.

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,7 @@
+Enhancement: Restore ownership by name
+
+Restic now optionally restores file ownership on UNIX systems by the names of the user and group.
+This allows restoring snapshots on a system where the UID/GID are not the same as they were on the system where the snapshot was created.
+Previously, on a system where the UID/username mapping or the GID/group name mapping did not match the expected values, restoring a snapshot would result in incorrect ownership of the files.
+
+https://github.com/restic/restic/issues/3572
```
MichaelEischer, line 8: Include a link to this PR.

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,7 @@
+Enhancement: Restore ownership by name
+
+Restic now optionally restores file ownership on UNIX systems by the names of the user and group.
+This allows restoring snapshots on a system where the UID/GID are not the same as they were on the system where the snapshot was created.
+Previously, on a system where the UID/username mapping or the GID/group name mapping did not match the expected values, restoring a snapshot would result in incorrect ownership of the files.
```
MichaelEischer: The "previous behavior" part should come first in the description.

internal/fs/node_unix.go
```diff
@@ -9,8 +9,17 @@ import (
 	"github.com/restic/restic/internal/restic"
 )
 
-func lchown(name string, uid, gid int) error {
-	return os.Lchown(name, uid, gid)
+func lchown(name string, node *restic.Node, lookupByName bool) error {
```
MichaelEischer: This definitely needs a test.

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,7 @@
+Enhancement: Restore ownership by name
```
MichaelEischer:
```suggestion
Enhancement: Support restoring ownership by name
```

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,7 @@
...
+
+Restic now optionally restores file ownership on UNIX systems by the names of the user and group.
```
MichaelEischer: Please mention the necessary parameter and that it is for the `restore` command.

provokateurin:
Rebased to resolve conflicts

MichaelEischer (changes requested):
While taking a closer look, I noticed a rather ugly problem that cannot be solved by this PR. Restic stores POSIX ACLs in a backup, but only in their numeric on-disk format. Thus, we have no way of restoring those based on their name.

However, fixing that is out of scope for this PR. Instead let's document the limitation somewhere (ideally in the CLI parameter). Besides that the PR only needs tests for the new codepaths.

internal/fs/node.go
```diff
@@ -130,6 +130,37 @@ func lookupUsername(uid uint32) string {
 	return username
 }
 
+var (
+	userNameLookupCache      = make(map[string]uint32)
+	userNameLookupCacheMutex = sync.RWMutex{}
+)
+
+// Cached uid lookup by user name. Returns 0 when no id can be found.
+func lookupUid(userName string) uint32 {
```
MichaelEischer, line 141: This will require a `//nolint:revive // captialization is correct as is` comment

changelog/unreleased/issue-3572
rawtaz: See https://github.com/restic/restic/blob/master/changelog/TEMPLATE for a guiding example on how to structure the changelog file :)

provokateurin:
> Instead let's document the limitation somewhere (ideally in the CLI parameter)

What exactly would you write there? I feel like it does not really belong into the usage description of the new CLI flag, but maybe for the restore command itself.

internal/fs/node_unix.go
provokateurin: I tried that, but unfortunately chown needs root permission, so it would add the requirement to run the tests with some kind of root access which is a really bad idea.
What do you think is the right way to test this?

MichaelEischer: We definitely don't want to require running the tests as root. But a basic sanity check should be possible. A user is is allowed to chown files owned by them to their own uid. This is effectively a no-op, but should be enough to sanity check the overall username / uid translation steps.

Other than that the tests in `internal/fs/node_test.go` should also be run with `lookupByName=true` at least once. There will be quite a few corner cases which won't be covered by those tests, but a basic smoke test is better than nothing at all.

MichaelEischer:
> What exactly would you write there? I feel like it does not really belong into the usage description of the new CLI flag, but maybe for the restore command itself.

The CLI flag could get a `(except POSIX ACLs)` suffix. Or add a somewhat longer description to the restore command itself. Or probably better: add both.

cmd/restic/cmd_restore.go
```diff
@@ -84,6 +86,9 @@ func (opts *RestoreOptions) AddFlags(f *pflag.FlagSet) {
 	f.BoolVar(&opts.Verify, "verify", false, "verify restored files content")
 	f.Var(&opts.Overwrite, "overwrite", "overwrite behavior, one of (always|if-changed|if-newer|never)")
 	f.BoolVar(&opts.Delete, "delete", false, "delete files from target directory if they do not exist in snapshot. Use '--dry-run -vv' to check what would be deleted")
+	if runtime.GOOS != "windows" {
+		f.BoolVar(&opts.OwnershipByName, "ownership-by-name", false, "restore file ownership by uid and gid")
```
MichaelEischer:
```suggestion
		f.BoolVar(&opts.OwnershipByName, "ownership-by-name", false, "restore file ownership by username and groupname")
```

provokateurin:
@MichaelEischer anything missing from my side or something I can do to get it merged? No pressure though, I see you are doing a lot of work for restic already! :heart:

provokateurin:
Sorry to ping again, @MichaelEischer any chance you could have another look? I resolved all review comments.

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,8 @@
+Enhancement: Support restoring ownership by name
+
+Restic restore used to restore file ownership on UNIX systems by UID and GID.
+It now allows restoring the file ownership by user name and group name with `--ownership-by-name`.
+This allows restoring snapshots on a system where the UID/GID are not the same as they were on the system where the snapshot was created.
```
MichaelEischer, line 5: Should also mention that this does not work for POSIX ACLs so far.

changelog/unreleased/issue-3572
```diff
@@ -0,0 +1,8 @@
+Enhancement: Support restoring ownership by name
```
MichaelEischer:
```suggestion
Enhancement: Support restoring ownership by name on UNIX systems
```

MichaelEischer (changes requested):
Sorry for the long delay. I only have two small nits regarding the changelog left, then this is ready for merging.

provokateurin:
Done! Thanks for reviewing!

MichaelEischer (approved):
LGTM.

For handling POSIX ACLs correctly, I've created a minimal issue: https://github.com/restic/restic/issues/5598

PR #5451, opened 2025-07-18
Concurrency simplifications

greatroar:

Simplifies some concurrent code by dropping the "sender + pool" pattern for the "spawn when needed" pattern. I just made those names up, but the idea is to spawn as many goroutines as there are jobs to do, limiting concurrency with a semaphore (built into errgroup.Group). This saves dozens of lines of code. It also has the benefit of smaller stack dumps in the case of a panic.

Also uses that same semaphore in internal/dump, where I previously used some tricky logic based on channel buffer sizes.

Can be considered a follow-up to #5074.

Checklist
---------

### DIFF
--- a/internal/dump/common.go
+++ b/internal/dump/common.go
@@ -104,9 +104,11 @@ func (d *Dumper) WriteNode(ctx context.Context, node *restic.Node) error {
 
 func (d *Dumper) writeNode(ctx context.Context, w io.Writer, node *restic.Node) error {
 	wg, ctx := errgroup.WithContext(ctx)
-	limit := d.repo.Connections() - 1 // See below for the -1.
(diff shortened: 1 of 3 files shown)


greatroar:
<s>I removed the change to restic.ParallelRemove, because it broke the prune integration tests. I think the tests are at fault, because they rely in a subtle way on runPrune returning an error when its context gets canceled in some circumstances but not others (without checking the returned error, which will be context.Canceled).</s> Scrap that, the problem is actually in ParallelList, called by ForAllLocks.

greatroar:
Ok, without the ParallelList changes, all tests pass. I declare this PR finished and will leave ParallelList for later.

MichaelEischer (approved):
LGTM.

PR #5462, opened 2025-08-02
Correct Wasabi link

TheAlchemistOf42:

This change updates the link for finding the right Wasabi service URL from the no longer existing Zendesk help center to Wasabi's documentation site.

Before:
https://wasabi-support.zendesk.com/hc/en-us/articles/360015106031-What-are-the-service-URLs-for-Wasabi-s-different-regions-

After:
https://docs.wasabi.com/v1/docs/what-are-the-service-urls-for-wasabi-s-different-storage-regions

No, I have not previously discussed this change, but when I looked through the issues and forums to see if this was previously reported, I noticed the link was reported as broken back in #4245 (which is already closed).

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -349,7 +349,7 @@ Wasabi
 
 S3 storage from `Wasabi <https://wasabi.com>`__ can be used as follows.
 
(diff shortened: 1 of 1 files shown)


rawtaz (approved):
Very nice, thank you!

PR #5464, opened 2025-08-04
 restic copy - add more status counters - issue #5175

### LINKED ISSUE #5175: restic copy - give more information in summary, like backup
Output of `restic version`
--------------------------

restic 0.17.3 compiled with go1.23.3 on windows/amd64

When running `restic backup`, you get a summary of objects added to the repository.

Example:
```
Files:         113 new,     0 changed,  2391 unmodified
Dirs:            0 new,     4 changed,     9 unmodified
Data Blobs:    253 new
Tree Blobs:      5 new
Added to the repository: 416.738 MiB (90.187 MiB stored)

processed 2504 files, 242.568 GiB in 29:37
snapshot 67a32f3e saved
```

When running `restic copy`, on the other hand, information is sparse:
```
[55:44] 100.00%  5655 / 5655 packs copied
snapshot 6d12443e saved
```

I would like `restic copy` to give more information.
If file and dir count is not easily obtained from the packs being copied, at least info on the amount of data added/stored to the target repository would be useful.

I am trying to keep track of how packs copied from a source repository changes a target repository, space wise.
If restic could report this, I would not have to check disk consumption of the repo "by hand".
Aditionally, this would give info on deduplication factor, which usually brightens my day.

Restic helps me and makes me happy every day. Keep up the good work!

wplapper:
'copyTree()' now counts and sizes the blobs in 'copyBlobs' and prints them out via 'Verbosef()'.

This PR addresses the issue raised in https://github.com/restic/restic/issues/5175

code has been added to count and size the changes collected in copyBlobs. These counts are printed out via Verbosef().

closes https://github.com/restic/restic/issues/5175

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5319
+++ b/changelog/unreleased/pull-5319
@@ -2,7 +2,8 @@ Enhancement: add more status counters to `restic copy`
 
 `restic copy` now produces more status counters in text format. The new counters
 are the number of blobs to copy, their size on disk and the number of packfiles
-used from the source repository.
+used from the source repository. The additional statistics is only produced when
+the `--verbose` option is apecified.
 
 https://github.com/restic/restic/issues/5175
 https://github.com/restic/restic/pull/5319
--- a/changelog/unreleased/pull-5319
+++ b/changelog/unreleased/pull-5319
@@ -1,0 +1,8 @@
+Enhancement: add more status counters to `restic copy`
+
+`restic copy` now produces more status counters in text format. The new counters
+are the number of tree/data blobs, their sizes and the number of packfiles
+they are attachedc to.
+
+https://github.com/restic/restic/issues/5175
+https://github.com/restic/restic/pull/5319
--- a/changelog/unreleased/pull-5319
+++ b/changelog/unreleased/pull-5319
@@ -1,8 +1,8 @@
 Enhancement: add more status counters to `restic copy`
 
 `restic copy` now produces more status counters in text format. The new counters
-are the number of tree/data blobs, their sizes and the number of packfiles
-they are attachedc to.
+are the number of blobs to copy, their size on disk and the number of packfiles
+used from the source repository.
 
 https://github.com/restic/restic/issues/5175
 https://github.com/restic/restic/pull/5319
--- a/changelog/unreleased/pull-5319
+++ b/changelog/unreleased/pull-5319
@@ -3,7 +3,7 @@ Enhancement: add more status counters to `restic copy`
 `restic copy` now produces more status counters in text format. The new counters
 are the number of blobs to copy, their size on disk and the number of packfiles
 used from the source repository. The additional statistics is only produced when
-the `--verbose` option is apecified.
+the `--verbose` option is specified.
 
 https://github.com/restic/restic/issues/5175
 https://github.com/restic/restic/pull/5319
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -242,10 +242,27 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
 		return err
 	}
 
+	copyStats(srcRepo, copyBlobs, packList, printer)
 	bar := printer.NewCounter("packs copied")
 	err = repository.Repack(ctx, srcRepo, dstRepo, packList, copyBlobs, bar, printer.P)
 	if err != nil {
 		return errors.Fatalf("%s", err)
 	}
 	return nil
+}
+
+func copyStats(srcRepo restic.Repository, copyBlobs restic.BlobSet, packList restic.IDSet, printer progress.Printer) {
+	// count and size
+	countBlobs := 0
+	sizeBlobs := uint64(0)
+	for blob := range copyBlobs {
+		for _, blob := range srcRepo.LookupBlob(blob.Type, blob.ID) {
+			countBlobs++
+			sizeBlobs += uint64(blob.Length)
+		}
+		break
+	}
+
+	printer.V("  %7d all  blobs with a size %11s in %7d packfiles\n",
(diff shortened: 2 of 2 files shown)


cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+				treePackfiles.Insert(blob.PackID)
+			} else if blob.Type == restic.DataBlob {
+				countDataBlobs++
+				sizeDataBlobs += blob.Length
+				dataPackfiles.Insert(blob.PackID)
+			}
+		}
+	}
+
+	Verbosef("  %7d tree blobs with a size %11s in %7d packfiles\n",
```
```suggestion
	printer.P("  %7d tree blobs with size %11s in %7d packfiles\n",
```

This is necessary after https://github.com/restic/restic/pull/5518

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
 		return err
 	}
 
+	// count and size
+	countTreeBlobs := 0
```
MichaelEischer: The statistics code must be moved into a helper function. While I'm not particularly opinionated regarding function size, I'd still like to keep functions to a point where the whole function fits on the screen. Yes, I'm aware that this rule of thumb is ignored in a lot of places in the codebase, but I'd like to not worsen the problem even further.

cmd/restic/cmd_copy.go
MichaelEischer: `Verbosef` already prints information when neither `--quiet` or `--verbose` are specified. As the extended statistics of the `backup` command are only visible, when setting `--verbose`, let's be consistent and use `printer.V` here.

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+	countDataBlobs := 0
+	treePackfiles := restic.NewIDSet()
+	dataPackfiles := restic.NewIDSet()
+	sizeTreeBlobs := uint(0)
```
```suggestion
	sizeTreeBlobs := uint64(0)
```
This will overflow on 32bit systems otherwise.

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+	sizeDataBlobs := uint(0)
+	for blob := range copyBlobs {
+		for _, blob := range srcRepo.LookupBlob(blob.Type, blob.ID) {
```
MichaelEischer, line 261: That's not correct. Blobs that exist multiple times would be counted multiple times although they are only copied once. Just check the first `PackedBlob` instance returned.

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+			if blob.Type == restic.TreeBlob {
+				countTreeBlobs++
+				sizeTreeBlobs += blob.Length
```
MichaelEischer: `Length` is the size on disk. Is that what we're interested in here?

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +239,35 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+				dataPackfiles.Insert(blob.PackID)
+			}
+		}
+	}
+
+	Verbosef("  %7d tree blobs with a size %11s in %7d packfiles\n",
+		countTreeBlobs, ui.FormatBytes(uint64(sizeTreeBlobs)), len(treePackfiles))
+	Verbosef("  %7d data blobs with a size %11s in %7d packfiles\n",
+		countDataBlobs, ui.FormatBytes(uint64(sizeDataBlobs)), len(dataPackfiles))
+	Verbosef("  %7d all  blobs with a size %11s in %7d packfiles\n",
```
MichaelEischer: This feels rather redundant. Maybe just print how much data will be copied in total?

MichaelEischer (changes requested):

cmd/restic/cmd_copy.go
wplapper: I rebased my PR, so that's done.

cmd/restic/cmd_copy.go
wplapper: done.

cmd/restic/cmd_copy.go
wplapper: done.

cmd/restic/cmd_copy.go
wplapper: insert ``break`` in inner loop.

cmd/restic/cmd_copy.go
wplapper: I personally am interested in the length on the disk, since this determined the runtime of the copy command.

cmd/restic/cmd_copy.go
wplapper: Print only the the sum of all blobs.

changelog/unreleased/pull-5319
```diff
@@ -0,0 +1,8 @@
+Enhancement: add more status counters to `restic copy`
+
+`restic copy` now produces more status counters in text format. The new counters
```
MichaelEischer, line 3: The stats are only shown in verbose mode, aka. `copy -v`. This should be mentioned.

MichaelEischer (commented):
I just have one more small nit. I like the compact statistics output, if there's interest in more details we can still add those later on.

MichaelEischer (approved):
LGTM.

I've rebased the PR to fix the merge conflict.

PR #5465, opened 2025-08-04
Fix: Correctly restore ACL inheritance state

### LINKED ISSUE #5427: Backup & Restore "IsInherited" Property on Windows
Output of `restic version`
--------------------------
```
PS C:\Program Files (x86)\restic> ./restic version
```

The addition of backing up SecurityDescriptors on Windows #4708 is fantastic, however I have noticed that restic does not restore the `IsInherited` property on files/directories.

Example:

**Output prior to backup**

```
PS C:\Windows\system32> $ACL = Get-Acl "E:\restic_test_root\level1\level2\level3"
$ACL.Access

FileSystemRights  : FullControl
AccessControlType : Allow
IdentityReference : BUILTIN\Administrators
IsInherited       : True
InheritanceFlags  : ContainerInherit, ObjectInherit
PropagationFlags  : None

FileSystemRights  : FullControl
AccessControlType : Allow
IdentityReference : NT AUTHORITY\SYSTEM
IsInherited       : True
InheritanceFlags  : ContainerInherit, ObjectInherit
PropagationFlags  : None

```
**Output after restore**

```
PS C:\Program Files (x86)\restic> $ACL = Get-Acl "E:\restic_test_root\level1\level2\level3"
$ACL.Access

FileSystemRights  : FullControl
AccessControlType : Allow
IdentityReference : BUILTIN\Administrators
IsInherited       : False
InheritanceFlags  : ContainerInherit, ObjectInherit
PropagationFlags  : None

FileSystemRights  : FullControl
AccessControlType : Allow
IdentityReference : NT AUTHORITY\SYSTEM
IsInherited       : False
InheritanceFlags  : ContainerInherit, ObjectInherit
PropagationFlags  : None


```

Having the original Inheritance properties is important when restoring, to retain the flow of permissions on child objects.

Always. Your hard work maintaining this excellent project is well appreciated!

aneesh-n:

When restoring a file or directory on Windows, the `IsInherited` property of its Access Control Entries (ACEs) was always being set to `False`, even if the ACEs were inherited in the original backup.

This was caused by the restore process calling the `SetNamedSecurityInfo` API without providing context about the object's inheritance policy. By default, this API applies the provided Discretionary Access Control List (DACL) as an explicit set of permissions, thereby losing the original inheritance state.

This commit fixes the issue by inspecting the `Control` flags of the saved Security Descriptor during restore. Based on whether the `SE_DACL_PROTECTED` flag is present, the code now adds the appropriate `PROTECTED_DACL_SECURITY_INFORMATION` or `UNPROTECTED_DACL_SECURITY_INFORMATION` flag to the `SetNamedSecurityInfo` API call.

By providing this crucial inheritance context, the Windows API can now correctly reconstruct the ACL, ensuring the `IsInherited` status of each ACE is preserved as it was at the time of backup.

Closes https://github.com/restic/restic/issues/5427

Checklist
---------

### DIFF
--- a/internal/fs/sd_windows.go
+++ b/internal/fs/sd_windows.go
@@ -12,14 +12,17 @@ import (
 
 var lowerPrivileges atomic.Bool
 
-// Flags for backup and restore with admin permissions
-var highSecurityFlags windows.SECURITY_INFORMATION = windows.OWNER_SECURITY_INFORMATION | windows.GROUP_SECURITY_INFORMATION | windows.DACL_SECURITY_INFORMATION | windows.SACL_SECURITY_INFORMATION | windows.LABEL_SECURITY_INFORMATION | windows.ATTRIBUTE_SECURITY_INFORMATION | windows.SCOPE_SECURITY_INFORMATION | windows.BACKUP_SECURITY_INFORMATION | windows.PROTECTED_DACL_SECURITY_INFORMATION | windows.PROTECTED_SACL_SECURITY_INFORMATION | windows.UNPROTECTED_DACL_SECURITY_INFORMATION | windows.UNPROTECTED_SACL_SECURITY_INFORMATION
+// Flags for backup with admin permissions. Includes protection flags for GET operations.
+var highBackupSecurityFlags windows.SECURITY_INFORMATION = windows.OWNER_SECURITY_INFORMATION | windows.GROUP_SECURITY_INFORMATION | windows.DACL_SECURITY_INFORMATION | windows.SACL_SECURITY_INFORMATION | windows.LABEL_SECURITY_INFORMATION | windows.ATTRIBUTE_SECURITY_INFORMATION | windows.SCOPE_SECURITY_INFORMATION | windows.BACKUP_SECURITY_INFORMATION | windows.PROTECTED_DACL_SECURITY_INFORMATION | windows.PROTECTED_SACL_SECURITY_INFORMATION | windows.UNPROTECTED_DACL_SECURITY_INFORMATION | windows.UNPROTECTED_SACL_SECURITY_INFORMATION
+
+// Flags for restore with admin permissions. Base flags without protection flags for SET operations.
+var highRestoreSecurityFlags windows.SECURITY_INFORMATION = windows.OWNER_SECURITY_INFORMATION | windows.GROUP_SECURITY_INFORMATION | windows.DACL_SECURITY_INFORMATION | windows.SACL_SECURITY_INFORMATION | windows.LABEL_SECURITY_INFORMATION | windows.ATTRIBUTE_SECURITY_INFORMATION | windows.SCOPE_SECURITY_INFORMATION | windows.BACKUP_SECURITY_INFORMATION
 
 // Flags for backup without admin permissions. If there are no admin permissions, only the current user's owner, group and DACL will be backed up.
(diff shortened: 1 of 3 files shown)


internal/fs/sd_windows.go
```diff
@@ -150,8 +157,31 @@ func getNamedSecurityInfoLow(filePath string) (*windows.SECURITY_DESCRIPTOR, err
...
 // setNamedSecurityInfoHigh sets the higher level SecurityDescriptor which requires admin permissions.
-func setNamedSecurityInfoHigh(filePath string, owner *windows.SID, group *windows.SID, dacl *windows.ACL, sacl *windows.ACL) error {
-	return windows.SetNamedSecurityInfo(fixpath(filePath), windows.SE_FILE_OBJECT, highSecurityFlags, owner, group, dacl, sacl)
+func setNamedSecurityInfoHigh(filePath string, owner *windows.SID, group *windows.SID, dacl *windows.ACL, sacl *windows.ACL, control windows.SECURITY_DESCRIPTOR_CONTROL) error {
+	securityInfo := highSecurityFlags
+
+	// Remove all protection flags first to ensure a clean state before adding the correct one.
+	// The highSecurityFlags variable contains all of them, which is fine for GET operations,
+	// but for SET operations, we must specify only one of PROTECTED or UNPROTECTED.
+	securityInfo &^= (windows.PROTECTED_DACL_SECURITY_INFORMATION | windows.UNPROTECTED_DACL_SECURITY_INFORMATION)
```
MichaelEischer: lowRestoreSecurityFlags also contains `PROTECTED_DACL_SECURITY_INFORMATION`. Do we need similar handling there?

internal/fs/sd_windows.go
```diff
@@ -150,8 +157,31 @@ func getNamedSecurityInfoLow(filePath string) (*windows.SECURITY_DESCRIPTOR, err
 }
 
 // setNamedSecurityInfoHigh sets the higher level SecurityDescriptor which requires admin permissions.
-func setNamedSecurityInfoHigh(filePath string, owner *windows.SID, group *windows.SID, dacl *windows.ACL, sacl *windows.ACL) error {
-	return windows.SetNamedSecurityInfo(fixpath(filePath), windows.SE_FILE_OBJECT, highSecurityFlags, owner, group, dacl, sacl)
+func setNamedSecurityInfoHigh(filePath string, owner *windows.SID, group *windows.SID, dacl *windows.ACL, sacl *windows.ACL, control windows.SECURITY_DESCRIPTOR_CONTROL) error {
+	securityInfo := highSecurityFlags
```
MichaelEischer: What about splitting the `highSecurityFlags` into a backup and restore variant? Then we don't need to drop the flags below. The `highBackupSecurityFlags` would then be based just add those flags to the `highRestoreSecurityFlags`.

MichaelEischer (commented):
Thanks! I just have two questions/suggestion below.

MichaelEischer (changes requested):
@aneesh-n can you take a look at the two comments above? Feel free to just close those comments if they aren't relevant.

internal/fs/sd_windows.go
aneesh-n: Thanks. Made the fix

internal/fs/sd_windows.go
aneesh-n: Thanks, makes sense. Updated

MichaelEischer (approved):
LGTM. Thanks. I've rebased the PR to fix the merge conflict near

```
var lowerPrivileges atomic.Bool
```

PR #5466, opened 2025-08-05
Added Gentoo install info

y0n3d4:

Simply adding instruction to install restic on Gentoo Linux

No: it's just an info add in

Checklist
---------

- [ ] I have added tests for all code changes. ``(Not applies: no code changes)``
- [ ] I have added documentation for relevant changes (in the manual). ``(Not applies: no code changes)``

### DIFF
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -82,7 +82,7 @@ package from the official repos using ``emerge``:
 
 .. code-block:: console
 
-    # emerge -av restic
+    # emerge restic
 
 macOS
 =====
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -74,6 +74,16 @@ avoid any conflicts:
 
    $ dnf copr remove copart/restic
 
+Gentoo Linux
+============
+
+On `Gentoo Linux <https://www.gentoo.org/>`__, you can install the ``restic``
+package from the official repos using ``emerge``:
+
+.. code-block:: console
+
+    # emerge -av restic
+
 macOS
 =====
 
--- a/doc/020_installation.rst
+++ b/doc/020_installation.rst
@@ -74,16 +74,6 @@ avoid any conflicts:
 
    $ dnf copr remove copart/restic
 
-Gentoo Linux
-============
-
-On `Gentoo Linux <https://www.gentoo.org/>`__, you can install the ``restic``
-package from the official repos using ``emerge``:
-
-.. code-block:: console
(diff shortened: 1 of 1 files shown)


doc/020_installation.rst
```diff
@@ -74,6 +74,16 @@ avoid any conflicts:
...
 
+Gentoo Linux
+============
+
+On `Gentoo Linux <https://www.gentoo.org/>`__, you can install the ``restic``
+package from the official repos using ``emerge``:
+
+.. code-block:: console
+
+    # emerge -av restic
```
rawtaz: I'm thinking that we shouldn't use `-av`, as it should be up to the user to decide what options to provide to their commands. The other install instructions here don't use verbose mode by default either. Does this make sense?

y0n3d4: You're absolutely right
Years and years of Gentoo have deformed my mind

rawtaz (approved):
Thanks!

doc/020_installation.rst
```diff
@@ -74,6 +74,16 @@ avoid any conflicts:
...
 
+Gentoo Linux
+============
+
+On `Gentoo Linux <https://www.gentoo.org/>`__, you can install the ``restic``
+package from the official repos using ``emerge``:
+
+.. code-block:: console
+
+    # emerge restic
```
rawtaz, line 85: One more thing - must you be root? I noticed the other examples in this documentation have just `$` as their prompt, indicating a non-root user (although `$` could be a root user as well). Is `#` the common prompt in Gentoo?

y0n3d4: Emerge requires root privileges when installing packages, due to adding/updating files in system directories like `/usr/bin`, `/etc` , `/usr/lib64` and so on

Under cover, (downloading sources | compiling them | linking all stuff) it executed using the unprivileged `portage` user, but you need superuser privilege to perform all install phases, otherwise `emerge` will complain if you try to install a package without root privileges:

```
$ emerge restic
emerge: superuser access is required
```

I have used the root prompt '#' just to remark this requirement, but you can quietly using the unprivileged user prompt '$', as it will be `emerge` to remind distracted users of it

Many user-friendly Linux distributions automatically add the first user created during installation to the sudo group (or equivalent), giving him administrative privileges, so he can install packages of his distro without superuser privileges. This is done to simplify administrative tasks without requiring the user to switch to the root account (imho, I find this a really bad habit :-P)

PR #5213, opened 2025-01-22
PR closed - wrong email: restic check - enable --read-data-* with snapshot filter

wplapper:
Enable restic check based on snapshots. The standard snapshot filtering routine  "FindFilteredSnapshots" is used for snapshots selection. However if no snapshots are specified, the subcommand behaves as before.

What does this PR change? What problem does it solve?
-----------------------------------------------------
The PR solves issue 3326. 

Add standard snapshot filtering to subcommand.
Add code to cmd/restic/cmd_check.go to detect snapshot filtering 
Add code to internal/checker/checker.go to collect data blobs based on tree visitor. The functions CountPacks() and GetPacks() have been modified, so that snapshot filtering creates a subset of packfiles on which all size calculations will be based.

When snapshot filtering is activated, the complete tree for this snapshot is read and the data blobs - and their packfiles - are extracted.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
https://github.com/restic/restic/issues/3326

Closes #3326

Checklist
---------

- [x] I have added tests for all code changes.
- [x] I have added documentation for relevant changes (in the manual).
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I'm done! This pull request is ready for review.

PR #5469, opened 2025-08-07
restic check with snapshot filters (continuation of PR #5213)

wplapper:
This is the continuation of PR #5213. I decided to use a git provided e-mail address instead of my private e-mail address. All changes up-to 2025-03-07 have been incorporated. Details to follow here:

changelog/unreleased/issue-3326:
Introduction of snapshot based filter for command `restic check`

cmd/restic/cmd_check.go:
Integration of snapshot based filtering. Some messages had to be changed so that the text does not refer to 'all packfiles' but to 'selected packfiles'.

A new test for filtering via snapshots has been added.

doc/077_troubleshooting.rst:
A short paragraph in the manual has been added about snapshot based filtering.

internal/checker/checker.go:
A new func CheckWithSnapshots() has been added which collects the needed packfiles via restic.FindUsedBlobs().

internal/checker/checker_test.go:
A new test TestCheckRepoSnapshot() has been added to check out the new functionality.

This PR addresses snapshot based filters in command ``restic check``.

https://github.com/restic/restic/issues/3326

https://github.com/restic/restic/issues/3326

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-3326
+++ b/changelog/unreleased/issue-3326
@@ -1,0 +1,8 @@
+Enhancement: enable --read-data-subset and --read-data for specified snapshot(s)
+
+Snapshots can now be specified on the command line via the standard snapshot filter,
+(`--tag`, `--host`, `--path` or specifying snapshot IDs directly) and will be used
+for checking the packfiles used by these snapshots.
+
+https://github.com/restic/restic/issues/3326
+https://github.com/restic/restic/pull/5213
--- a/changelog/unreleased/issue-3326
+++ b/changelog/unreleased/issue-3326
@@ -1,8 +1,8 @@
-Enhancement: enable --read-data-subset and --read-data for specified snapshot(s)
+Enhancement: `restic check` for specified snapshot(s) via snapshot filtering
 
-Snapshots can now be specified on the command line via the standard snapshot filter,
-(`--tag`, `--host`, `--path` or specifying snapshot IDs directly) and will be used
-for checking the packfiles used by these snapshots.
+Snapshots can now be specified for the command `restic check` on the command line
+via the standard snapshot filter, (`--tag`, `--host`, `--path` or specifying
+snapshot IDs directly) and will be used for checking the packfiles used by these snapshots.
 
 https://github.com/restic/restic/issues/3326
 https://github.com/restic/restic/pull/5213
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -15,9 +15,11 @@ import (
 
 	"github.com/restic/restic/internal/backend/cache"
 	"github.com/restic/restic/internal/checker"
+	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/global"
 	"github.com/restic/restic/internal/repository"
+	"github.com/restic/restic/internal/repository/pack"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/ui"
 	"github.com/restic/restic/internal/ui/progress"
@@ -34,6 +36,9 @@ finds. It can also be used to read all data and therefore simulate a restore.
 
 By default, the "check" command will always load all data directly from the
 repository and not use a local cache.
+
+The "check" command can now check packfiles for specific snapshots. The snapshots
+are filtered via the standard SnapshotFilter.
 
 EXIT STATUS
 ===========
@@ -71,6 +76,7 @@ type CheckOptions struct {
 	ReadDataSubset string
 	CheckUnused    bool
 	WithCache      bool
+	data.SnapshotFilter
 }
 
 func (opts *CheckOptions) AddFlags(f *pflag.FlagSet) {
@@ -84,6 +90,7 @@ func (opts *CheckOptions) AddFlags(f *pflag.FlagSet) {
 		panic(err)
 	}
 	f.BoolVar(&opts.WithCache, "with-cache", false, "use existing cache, only read uncached data from repository")
+	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
 }
 
 func checkFlags(opts CheckOptions) error {
@@ -220,20 +227,12 @@ func prepareCheckCache(opts CheckOptions, gopts *global.Options, printer progres
 
 func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args []string, term ui.Terminal) (checkSummary, error) {
 	summary := checkSummary{MessageType: "summary"}
-	if len(args) != 0 {
-		return summary, errors.Fatal("the check command expects no arguments, only options - please see `restic help check` for usage and flags")
-	}
 
 	var printer progress.Printer
 	if !gopts.JSON {
 		printer = ui.NewProgressPrinter(gopts.JSON, gopts.Verbosity, term)
 	} else {
 		printer = newJSONErrorPrinter(term)
-	}
-
-	readDataFilter, err := buildPacksFilter(opts, printer)
-	if err != nil {
-		return summary, err
 	}
 
 	cleanup := prepareCheckCache(opts, &gopts, printer)
@@ -331,6 +330,11 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 		return summary, ctx.Err()
 	}
 
+	readDataFilter, err := buildPacksFilter(ctx, repo, opts, args, printer)
+	if err != nil {
+		return summary, err
+	}
+
 	printer.P("check snapshots, trees and blobs\n")
 	errChan = make(chan error)
 	var wg sync.WaitGroup
@@ -416,11 +420,26 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 	return summary, nil
 }
 
-func buildPacksFilter(opts CheckOptions, printer progress.Printer) (func(packs map[restic.ID]int64) map[restic.ID]int64, error) {
+func buildPacksFilter(ctx context.Context, repo *repository.Repository, opts CheckOptions,
+	args []string, printer progress.Printer) (func(packs map[restic.ID]int64) map[restic.ID]int64, error) {
+
+	// avaluate if snapshot filtering is active
+	packSnapshots, active, err := selectPacks(ctx, repo, opts, args, printer)
+	if err != nil {
+		return nil, err
+	}
+	which := ""
+	if active {
+		which = "selected "
+	}
+
 	switch {
 	case opts.ReadData:
 		return func(packs map[restic.ID]int64) map[restic.ID]int64 {
-			printer.P("read all data\n")
+			printer.P("read %sdata\n", which)
+			if active {
+				packs = packSnapshots
+			}
 			return packs
 		}, nil
 	case opts.ReadDataSubset != "":
@@ -429,9 +448,12 @@ func buildPacksFilter(opts CheckOptions, printer progress.Printer) (func(packs m
 			bucket := dataSubset[0]
 			totalBuckets := dataSubset[1]
 			return func(packs map[restic.ID]int64) map[restic.ID]int64 {
+				if active {
+					packs = packSnapshots
+				}
 				packCount := uint64(len(packs))
 				packs = selectPacksByBucket(packs, bucket, totalBuckets)
-				printer.P("read group #%d of %d data packs (out of total %d packs in %d groups)\n", bucket, len(packs), packCount, totalBuckets)
+				printer.P("read group #%d of %d %sdata packs (out of total %d packs in %d groups)\n", bucket, len(packs), which, packCount, totalBuckets)
 				return packs
 			}, nil
 		} else if strings.HasSuffix(opts.ReadDataSubset, "%") {
@@ -440,13 +462,19 @@ func buildPacksFilter(opts CheckOptions, printer progress.Printer) (func(packs m
 				return nil, err
 			}
 			return func(packs map[restic.ID]int64) map[restic.ID]int64 {
-				printer.P("read %.1f%% of data packs\n", percentage)
+				if active {
+					packs = packSnapshots
+				}
+				printer.P("read %.1f%% of %sdata packs\n", percentage, which)
 				return selectRandomPacksByPercentage(packs, percentage)
 			}, nil
 		}
 
 		repoSize := int64(0)
 		return func(packs map[restic.ID]int64) map[restic.ID]int64 {
+			if active {
+				packs = packSnapshots
+			}
 			for _, size := range packs {
 				repoSize += size
 			}
@@ -556,3 +584,58 @@ func (*jsonErrorPrinter) P(_ string, _ ...interface{})  {}
 func (*jsonErrorPrinter) PT(_ string, _ ...interface{}) {}
 func (*jsonErrorPrinter) V(_ string, _ ...interface{})  {}
 func (*jsonErrorPrinter) VV(_ string, _ ...interface{}) {}
+
+func selectPacks(ctx context.Context, repo *repository.Repository, opts CheckOptions,
+	args []string, printer progress.Printer) (map[restic.ID]int64, bool, error) {
+
+	selectedTrees := []restic.ID{}
+	allPacks, err := pack.Size(ctx, repo, false)
+	if err != nil {
+		return nil, false, err
+	}
+
+	filterActive := false
+	// check snapshot filter
+	if len(args) > 0 || !opts.SnapshotFilter.Empty() {
+		snapshotLister, err := restic.MemorizeList(ctx, repo, restic.SnapshotFile)
+		if err != nil {
+			return nil, false, err
+		}
+
+		err = (&opts.SnapshotFilter).FindAll(ctx, snapshotLister, repo, args, func(_ string, sn *data.Snapshot, err error) error {
+			if err != nil {
+				return err
+			}
+
+			selectedTrees = append(selectedTrees, *sn.Tree)
+			return nil
+		})
+
+		if err != nil {
+			return nil, false, err
+		}
+		if len(selectedTrees) == 0 {
+			return nil, false, errors.Fatal("snapshotfilter active but no snapshot selected")
+		}
+
+		// convert selecteTrees to a list of used blobs
+		filterActive = true
+		usedBlobs := restic.NewBlobSet()
+		err = data.FindUsedBlobs(ctx, repo, selectedTrees, usedBlobs, nil)
+		if err != nil {
+			return nil, false, err
+		}
+
+		// convert these blobs to their encompassing packfiles
+		selectedPacks := make(map[restic.ID]int64)
+		for bh := range usedBlobs {
+			for _, pb := range repo.LookupBlob(bh.Type, bh.ID) {
+				selectedPacks[pb.PackID] = allPacks[pb.PackID]
+				break
+			}
+		}
+		return selectedPacks, filterActive, nil
+	}
+
+	return nil, false, nil
+}
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -340,7 +340,7 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 		defer wg.Done()
 		bar := printer.NewCounter("snapshots")
 		defer bar.Done()
-		chkr.Structure(ctx, bar, errChan)
+		chkr.Structure(ctx, bar, errChan, &printer)
 	}()
 
 	for err := range errChan {
@@ -377,7 +377,6 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 	}
 
 	if readDataFilter != nil {
-
 		if chkr.FilterStatus() {
 			printer.P("Snapshot filtering is active")
 		}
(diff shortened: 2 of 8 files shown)


cmd/restic/cmd_check.go
```diff
@@ -245,6 +247,31 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 	}
 	defer unlock()
 
+	// check snapshot filter
+	selectedTrees := []restic.ID{}
+	if len(args) > 0 || !opts.SnapshotFilter.Empty() {
```
MichaelEischer: Please move this into a helper function.

cmd/restic/cmd_check.go
```diff
@@ -392,9 +428,14 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts GlobalOptions, args
 		p.Done()
 	}
 
+	whichSelection := "data"
+	if filterBySnapshot {
+		whichSelection = "selected data"
```
MichaelEischer: It would be sufficient to just inject `"selected "` here.

internal/checker/checker.go
```diff
@@ -15,7 +16,6 @@ import (
 	"github.com/restic/restic/internal/repository/pack"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/ui/progress"
-	"golang.org/x/sync/errgroup"
 )
 
 // Checker runs various checks on a repository. It is advisable to create an
```
MichaelEischer, line 18: Checker.Structure should IMHO also only check the specified snapshots.

internal/checker/checker.go
```diff
@@ -522,3 +521,33 @@ func (c *Checker) ReadPacks(ctx context.Context, packs map[restic.ID]int64, p *p
...
+	// convert blobs to packfile IDs
+	snapPacks := map[restic.ID]int64{}
+	for blob := range usedBlobs {
+		for _, res := range c.repo.LookupBlob(blob.Type, blob.ID) {
+			snapPacks[res.PackID] = c.packs[res.PackID]
+		}
+	}
+
+	if len(snapPacks) > 0 {
+		c.packs = snapPacks
```
MichaelEischer: https://github.com/restic/restic/pull/5532 has split the checker code into a repository-internal part and the part handling the restic data format. As such it is no longer possible to overwrite `c.packs` from here. I've opened https://github.com/restic/restic/pull/5550 to cleanup how selecting the packs to read work.

My suggestion would be to extend `Structure()` to also build a list of accessed pack files. Then with the changes from #5550, this list can be returned via the filter callback to `Packs(...)`.

MichaelEischer (changes requested):
The changes from https://github.com/restic/restic/pull/5532 require untangling how the pack file selection works here, see the comments for details. `Structure(...)` should also adhere to the snapshot selection. That method can also generate the pack file list, making `CheckWithSnapshots` unnecessary and also avoiding the massive performance penalty of iterating a second time over the snapshot trees.

Edit:
Just integrate the snapshot selection into `checker.LoadSnapshots`. Then `checker.Snapshots` naturally also adheres to the snapshot filter. The `checker.Checker` can the also wrap the `ReadPacks` method such that it replaces the pack list passed to the filter callback with the one only containing the packs relevant for the snapshots.

cmd/restic/cmd_check.go
MichaelEischer: On second thought, please just integrate this into `checker.LoadSnapshots(...)`.

wplapper:
Hi [MichaelEischer](https://github.com/MichaelEischer),

looking at the existing changes after #5532 and the proposed changed in #5550 I have decided to wait until #5550 has been commited. Then I will start from a ``restic 0.18.1`` base and adopt my changes and your recommendations accordingly. I will place my PR into "draft" status.

internal/checker/checker.go
wplapper: ``Structure()``has been extended and a wrapper for ``ReadPacks()`` has been created.

internal/checker/checker.go
wplapper: Has been implemented.

cmd/restic/cmd_check.go
wplapper: done.

cmd/restic/cmd_check.go
wplapper: moved copying of parameters into `checker.LoadSnapshots(...)``. The actual test of this condition happens in ``Structure()``.

wplapper:
I have no idea what this Windows test ``TestDirAttributeCombinationsOverwrite` does, I have seen it before, it seems to be transient, and I have no idea of how to fix it. I need some solid Windows advice. Thanks!

wplapper:
I have removed these failing windows tests from the this PR. I will have to wait anyway until this issue has been resolved.

changelog/unreleased/issue-3326
```diff
@@ -0,0 +1,8 @@
+Enhancement: enable --read-data-subset and --read-data for specified snapshot(s)
```
MichaelEischer: Should mention that this change applies to the `check` command.

internal/restorer/restorer_windows_test.go
```diff
@@ -61,7 +61,8 @@ func TestFileAttributeCombination(t *testing.T) {
 	testFileAttributeCombination(t, false)
 }
 
-func TestEmptyFileAttributeCombination(t *testing.T) {
+//func TestEmptyFileAttributeCombination(t *testing.T) {
+func tTestEmptyFileAttributeCombination(t *testing.T) {
```
MichaelEischer: Please revert that change. At least at the moment it seems like those tests aren't as flaky anymore. If necessary, we could think about adjusting those tests to retry a an access denied error. But I'm against removing them.

internal/checker/checker.go
```diff
@@ -127,8 +137,32 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
...
+	if len(c.args) > 0 || !c.snapshotFilter.Empty() {
+		err := (&c.snapshotFilter).FindAll(ctx, c.snapshots, c.repo, c.args, func(_ string, sn *data.Snapshot, _ error) error {
+			if sn != nil {
+				trees = append(trees, *sn.Tree)
+			}
+			return nil
+		})
+		if err != nil {
+			if printer != nil {
+				(*printer).E("could not load snapshots: %v", err)
```
MichaelEischer: This error MUST be sent back via the `errChan`. Otherwise, the calling code won't notice that something went wrong.

internal/checker/checker.go
```diff
@@ -127,8 +137,32 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
...
 // subtrees are available in the index. errChan is closed after all trees have
 // been traversed.
-func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error) {
-	trees, errs := loadSnapshotTreeIDs(ctx, c.snapshots, c.repo)
+func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error, printer *progress.Printer) {
+	var errs []error
+	var trees []restic.ID
+
+	if len(c.args) > 0 || !c.snapshotFilter.Empty() {
+		err := (&c.snapshotFilter).FindAll(ctx, c.snapshots, c.repo, c.args, func(_ string, sn *data.Snapshot, _ error) error {
```
MichaelEischer: This error must be sent back via the errChan.

internal/checker/checker.go
```diff
@@ -127,8 +137,32 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
 // Structure checks that for all snapshots all referenced data blobs and
 // subtrees are available in the index. errChan is closed after all trees have
 // been traversed.
-func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error) {
-	trees, errs := loadSnapshotTreeIDs(ctx, c.snapshots, c.repo)
+func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error, printer *progress.Printer) {
+	var errs []error
+	var trees []restic.ID
+
+	if len(c.args) > 0 || !c.snapshotFilter.Empty() {
```
MichaelEischer: Please move the whole snapshot selection into a helper method.

internal/checker/checker.go
```diff
@@ -127,8 +137,32 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
 // Structure checks that for all snapshots all referenced data blobs and
 // subtrees are available in the index. errChan is closed after all trees have
 // been traversed.
-func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error) {
-	trees, errs := loadSnapshotTreeIDs(ctx, c.snapshots, c.repo)
+func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error, printer *progress.Printer) {
```
MichaelEischer:
```suggestion
func (c *Checker) Structure(ctx context.Context, p *progress.Counter, errChan chan<- error, printer progress.Printer) {
```

`progress.Printer` is an interface. So no need add another pointer.

internal/checker/checker.go
```diff
@@ -127,8 +137,32 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
...
+			return
+		}
+		c.filterActive = true
+		c.trees = trees
+	} else {
+		trees, errs = loadSnapshotTreeIDs(ctx, c.snapshots, c.repo)
+	}
+
+	if len(trees) == 0 && printer != nil {
+		(*printer).E("\n=== no snapshots selected - expect surprising messages==\n\n")
```
MichaelEischer: Why not just return in this case? Restic should still behave reasonably, even if a user did something incorrect.

internal/checker/checker.go
```diff
@@ -259,3 +293,46 @@ func (c *Checker) UnusedBlobs(ctx context.Context) (blobs restic.BlobHandles, er
...
+	}
+
+	allPacks, err := pack.Size(ctx, c.repo, false)
+	if err != nil {
+		errChan <- err
+		return
+	}
+	// convert selected trees to a set of used blobs
+	usedBlobs := restic.NewBlobSet()
+	err = data.FindUsedBlobs(ctx, c.repo, c.trees, usedBlobs, nil)
```
MichaelEischer: This is rather slow. Please just set `trackUnused` to true in the checker (must already happen in LoadSnapshots). Then `c.blobRefs.M` will already contain all used blobs.

internal/checker/checker.go
```diff
@@ -259,3 +293,46 @@ func (c *Checker) UnusedBlobs(ctx context.Context) (blobs restic.BlobHandles, er
...
+	// convert these blobs into their encompassing packfiles
+	selectedPacks := make(map[restic.ID]int64)
+	for bh := range usedBlobs {
+		for _, pb := range c.repo.LookupBlob(bh.Type, bh.ID) {
+			selectedPacks[pb.PackID] = allPacks[pb.PackID]
+			break
+		}
+	}
+
+	c.Checker.ReadPacks(ctx, func(_ map[restic.ID]int64) map[restic.ID]int64 {
```
MichaelEischer: `_ map[restic.ID]int64` contains the same data as `allPacks`. Thus, drop the call to `pack.Size` and move the loop that calculated `selectedPacks` here.

internal/checker/checker.go
```diff
@@ -51,9 +58,12 @@ func New(repo checkerRepository, trackUnused bool) *Checker {
 	return c
 }
 
-func (c *Checker) LoadSnapshots(ctx context.Context) error {
+func (c *Checker) LoadSnapshots(ctx context.Context, snapshotFilter data.SnapshotFilter, args []string) error {
 	var err error
 	c.snapshots, err = restic.MemorizeList(ctx, c.repo, restic.SnapshotFile)
+	c.snapshotFilter = snapshotFilter
+	c.args = args
```
MichaelEischer, line 61: Should already set `filterActive`.

cmd/restic/cmd_check_integration_test.go
```diff
@@ -34,3 +35,87 @@ func testRunCheckOutput(t testing.TB, gopts global.Options, checkUnused bool) (s
 	})
 	return buf.String(), err
 }
+
+func testRunCheckOutputWithOpts(t testing.TB, gopts global.Options, opts CheckOptions, args []string) (string, error) {
+	buf, err := withCaptureStdout(t, gopts, func(ctx context.Context, gopts global.Options) error {
+		gopts.Verbosity = 2
+		gopts.Verbose = 2
```
MichaelEischer:
```suggestion
```
`gopts.Verbose` is unnecessary. Sorry for that mess.

doc/077_troubleshooting.rst
```diff
@@ -82,6 +82,12 @@ If ``check`` detects damaged pack files, it will show instructions on how to rep
 them using the ``repair pack`` command. Use that command instead of the "Repair the
 index" section in this guide.
 
+If you are interested to check the repository via snapshots, you can now
```
MichaelEischer:
```suggestion
If you are interested to check only specific snapshots, you can now
```

MichaelEischer (changes requested):
Thanks for the changes so far. This is definitely moving into the right direction, but needs a few further cleanups. The current `ReadPacks` extension is also _far_ slower than it has to be (see comments).

internal/checker/checker.go
wplapper: I appended the error to ``errs``. From there it will be taken to ``errChan``.

internal/checker/checker.go
wplapper: see above.

internal/checker/checker.go
wplapper: done as recpmmended (``loadActiveTrees``).

internal/checker/checker.go
wplapper: made it a ``progress.Printer``.

internal/checker/checker.go
wplapper: removed test completely.

internal/checker/checker.go
wplapper: done as recommended.

internal/checker/checker.go
wplapper: done as recommended.

internal/checker/checker.go
wplapper: Yep, done as recommended.

doc/077_troubleshooting.rst
wplapper: done as recommended.

cmd/restic/cmd_check_integration_test.go
wplapper: removed ``gopts.Verbose = 2``.

cmd/restic/cmd_check.go
```diff
@@ -365,6 +361,12 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 		return summary, ctx.Err()
 	}
 
+	opts.filteredStatus = chkr.FilterStatus()
```
MichaelEischer: Let's move it below the `if opts.CheckUnused {` block. Then the call is directly next to the place where it's used.

cmd/restic/cmd_check.go
```diff
@@ -365,6 +361,12 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
...
+	readDataFilter, err := buildPacksFilter(opts, printer)
```
MichaelEischer: `filteredStatus` should be passed as parameter to `buildPacksFilter(...)`. It doesn't really belong into the opts struct.

internal/checker/checker.go
```diff
@@ -51,9 +57,13 @@ func New(repo checkerRepository, trackUnused bool) *Checker {
...
+	c.trackUnused = true
```
MichaelEischer: That should only be set if the snapshotFilter is actually used. Aka. `c.filterActive` must also be set here. Or even better, get completely rid of `filterActive` and just use a helper function that determines whether there's an active filter.

internal/checker/checker.go
```diff
@@ -259,3 +297,39 @@ func (c *Checker) UnusedBlobs(ctx context.Context) (blobs restic.BlobHandles, er
...
+// packfile set and submits them to repository.ReadPacks()
+func (c *Checker) ReadPacks(ctx context.Context, filter func(packs map[restic.ID]int64) map[restic.ID]int64, p *progress.Counter, errChan chan<- error) {
+	// no snapshot filtering, pass through
+	if !c.FilterStatus() {
+		c.Checker.ReadPacks(ctx, filter, p, errChan)
+		return
+	}
+
+	// convert used blobs into their encompassing packfiles
+	selectedPacks := restic.NewIDSet()
```
MichaelEischer: This should also only be computed within `filteredPackfiles`. The same applies for the variable definition of `packsWithSize`.

internal/checker/checker.go
```diff
@@ -259,3 +297,39 @@ func (c *Checker) UnusedBlobs(ctx context.Context) (blobs restic.BlobHandles, er
...
+			packsWithSize[packID] = allPacks[packID]
+		}
+		return filter(packsWithSize)
+	}
+
+	c.Checker.ReadPacks(ctx, filteredPackfiles, p, errChan)
+}
+
+// FilterStatus returns the filtering status
+func (c *Checker) FilterStatus() bool {
```
MichaelEischer: Should move right below `LoadSnapshots()`.

internal/checker/checker.go
```diff
@@ -124,13 +134,41 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
...
+				return er
+			}
+			return nil
+		})
+		if err != nil {
+			errs = append(errs, err)
+			return
+		}
+		c.filterActive = true
+		c.trees = trees
```
MichaelEischer: Should just return the trees instead of storing them in `c.trees`

internal/checker/checker.go
```diff
@@ -124,13 +134,41 @@ func loadSnapshotTreeIDs(ctx context.Context, lister restic.Lister, repo restic.
 	return ids, errs
 }
 
+func (c *Checker) loadActiveTrees(ctx context.Context) (errs []error) {
+	trees := []restic.ID{}
+	errs = []error{}
+
+	if len(c.args) > 0 || !c.snapshotFilter.Empty() {
+		err := (&c.snapshotFilter).FindAll(ctx, c.snapshots, c.repo, c.args, func(_ string, sn *data.Snapshot, er error) error {
```
MichaelEischer:
```suggestion
		err := (&c.snapshotFilter).FindAll(ctx, c.snapshots, c.repo, c.args, func(_ string, sn *data.Snapshot, err error) error {
```

There's no problem with shadowing the outer `err` variable

cmd/restic/cmd_check_integration_test.go
```diff
@@ -34,3 +35,86 @@ func testRunCheckOutput(t testing.TB, gopts global.Options, checkUnused bool) (s
...
+func testRunCheckOutputWithOpts(t testing.TB, gopts global.Options, opts CheckOptions, args []string) (string, error) {
+	buf, err := withCaptureStdout(t, gopts, func(ctx context.Context, gopts global.Options) error {
+		gopts.Verbosity = 2
+		_, err := runCheck(context.TODO(), opts, gopts, args, gopts.Term)
+		return err
+	})
+	return buf.String(), err
+}
+
+func TestCheckFullOutput(t *testing.T) {
```
MichaelEischer: Please merge the tree tests into a table test (see for example `TestPrepareCheckCache`; you bascially create a struct that contains the expected inputs and output and then use a loop to iterate over all of them). Almost all of their code is shared.

MichaelEischer (changes requested):
Thanks for keeping up with all the review comments! I have a few further comments, although most of them are nits this time.

internal/checker/checker.go
wplapper: done.

internal/checker/checker.go
wplapper: done.

internal/checker/checker.go
wplapper: moved setting of ``c.trackUnused`` into ``loadActiveTrees()``.

cmd/restic/cmd_check.go
wplapper: done.

cmd/restic/cmd_check.go
wplapper: moved the call to ``buildPacksFilter()`` directly before its use.

internal/checker/checker.go
wplapper: Your suggestion encapsulates the code better :)

cmd/restic/cmd_check_integration_test.go
wplapper: bult testCase table.

MichaelEischer (approved):
LGTM. Thanks!

I've added a few small cleanups, see the extra commits for details. In particular, only `packsWithSize` is built but not `selectedPacks`. As sets in go are under the hood also maps, we can just directly build the final map. `FilterStatus()` is now the authoritative source whether filtering happens and is less sensitive to whether it's called after certain other methods.

MichaelEischer:
And a trivial rebase + fix to work with the changed datatype for `c.blobRefs.M`

PR #5472, opened 2025-08-11
restic copy --stream: run one large copy operation crossing snapshot boundaries - issue #5453

### LINKED ISSUE #5453: Spool packs across snapshots during `restic copy`
Output of `restic version`
--------------------------

```
$ restic version
restic 0.18.0 compiled with go1.24.4 on linux/amd64
```

`restic copy` takes a `--pack-size` argument which is very helpful, and provides a hint to restic of the target pack size in the remote. `restic copy` takes either a list of snapshots or a `--from-repo` argument. When restic copy takes a long list of snapshots or a `--from-repo` with many small diffs it treats each snapshot separately and is unable to respect the `--pack-size` argument.

Feature request: Spool packs across snapshots during `restic copy`. Rather than treating each snapshot as an individual entity, go through the snapshot list tracking needed data until hitting the requested pack size. Then send the data in a large pack, before sending the metadata for all of the snapshots that now have their required data on the remote.

I'm not yet familiar with this section of Restic's source code, but it would be great to get some thoughts on whether this is possible. With minor searches it appears safe as the worst case is you spend longer copying a larger pack that fails and that data remains unreferenced - this seems to be the expected behaviour when requesting a large pack size.

I have many small backup sources that backup text files that change very little regularly. This creates many small packs in the repo when `restic backup` is used - this is expected behaviour, and when `restic forget --prune --repack-small` comes around they get grouped up into easier to manage files. All is well so far.

When I use `restic copy` from this repo with many snapshots and many large packs, it appears to be splitting things back into small packs because it moves the data _each snapshot_ is responsible for at once. My expected behaviour is that it works to generate packs of the requested `--pack-size` instead.

For write once/read rarely (or expensive read) remote repos, this makes `rsync` or `rclone` more effectively than `restic copy`. That's because I can repack the local repo to get the pack size I want then `rsync` this somewhere and maintain that pack size. I cannot see a way to achieve this with `restic copy` without repacking on the remote.

Always :) I'm moving a bunch of my backups around, found the new `--pack-size` and flawlessly updated my core repos. Really nice stuff.

wplapper:
restic copy now allows to optimize the repository packsize across multiple snapshots when copying from one repository to another repository.

changelog/unreleased/issue-5453:
announce new option `--batch`.

add stream test

cmd/restic/cmd_copy.go:
new functions collectAllSnapshots(), copyTreeBatched()  to perfrom the batch copy.

cmd/restic/cmd_copy_integration_test.go:
a new streaming test

doc/045_working_with_repos.rst:
a short note about option `--batch`

This PR modifies the way the ``copy`` operation works. With the ``--batch`` option, one large request list is built and sent to repository.Repack(), instead of one request list per snapshot to be copied. This allows optimizing the repository packsize while copying ``restic backup``with only minor changes between different snapshots. Normally, one new index file is created per snapshots, together with at least two packfiles for each new snapshot. When changes between backups are small, this can result in a lot of smallish packfiles. 

Yes, this can always be fixed by running ``restic prune``, but this is an extra step.  

closes #5453

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5453
+++ b/changelog/unreleased/issue-5453
@@ -1,12 +1,10 @@
-Enhancement: `restic copy` can now spool packfiles across muliple snapshots
+Enhancement: `copy` copies snapshots in batches
 
-When using `restic copy` used to save all newly created packfiles when finishing one snapshot,
-even when the actual packfile size was quite small. This applied particularly to
-incremental backups, when there was only small changes between individual backups.
+The `copy` command used to copy snapshots individually, even if this resulted in creating pack files
+smaller than the target pack size. In particular, this resulted in many small files
+when copying small incremental snapshots.
 
-When using the new option `--batch`, `restic copy` now creates one large request list
-which contains all blobs from all snapshots to be copied and then executes the
-copy operation.
+Now, `copy` copies multiple snapshots at once to avoid creating small files.
 
 https://github.com/restic/restic/issues/5175
 https://github.com/restic/restic/pull/5464
--- a/changelog/unreleased/issue-5453
+++ b/changelog/unreleased/issue-5453
@@ -1,0 +1,12 @@
+Enhancement: `restic copy` can now spool packfiles across muliple snapshots
+
+When using `restic copy` used to save all newly created packfiles when finishing one snapshot,
+even when the actual packfile size was quite small. This applied particularly to
+incremental backups, when there was only small changes between individual backups.
+
+When using the new option `--batch`, `restic copy` now creates one large request list
+which contains all blobs from all snapshots to be copied and then executes the
+copy operation.
+
+https://github.com/restic/restic/issues/5175
+https://github.com/restic/restic/pull/5464
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -218,7 +218,7 @@ func similarSnapshots(sna *data.Snapshot, snb *data.Snapshot) bool {
 }
 
 func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
-	visitedTrees restic.IDSet, rootTreeID restic.ID, printer progress.Printer, uploader restic.BlobSaver, seenBlobs restic.IDSet) error {
+	visitedTrees restic.IDSet, rootTreeID restic.ID, printer progress.Printer, uploader restic.BlobSaver) error {
 
 	wg, wgCtx := errgroup.WithContext(ctx)
 
@@ -232,15 +232,11 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
 	packList := restic.NewIDSet()
 
 	enqueue := func(h restic.BlobHandle) {
-		if seenBlobs.Has(h.ID) {
-			return
-		}
 		pb := srcRepo.LookupBlob(h.Type, h.ID)
 		copyBlobs.Insert(h)
 		for _, p := range pb {
 			packList.Insert(p.PackID)
 		}
-		seenBlobs.Insert(h.ID)
 	}
 
 	wg.Go(func() error {
@@ -322,19 +318,15 @@ func copyTreeBatched(ctx context.Context, srcRepo restic.Repository, dstRepo res
 	visitedTrees restic.IDSet, selectedSnapshots []*data.Snapshot, opts CopyOptions,
 	printer progress.Printer) error {
 
-	// seenBlobs is necessary in about 1 of 10000 blobs, in the other 99.99% the check
-	// dstRepo.LookupBlobSize() is working
-	seenBlobs := restic.NewIDSet()
-	// dependent on opts.batch the package Uploader is started either for
+	// dependent on opts.batch the pack uploader is started either for
 	// each snapshot to be copied or once for all snapshots
-
 	if opts.batch {
 		// call WithBlobUploader() once and then loop over all selectedSnapshots
 		err := dstRepo.WithBlobUploader(context.TODO(), func(ctx context.Context, uploader restic.BlobSaver) error {
 			for _, sn := range selectedSnapshots {
 				printer.P("\n%v", sn)
 				printer.P("  copy started, this may take a while...")
-				err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader, seenBlobs)
+				err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader)
 				if err != nil {
 					return err
 				}
@@ -360,7 +352,7 @@ func copyTreeBatched(ctx context.Context, srcRepo restic.Repository, dstRepo res
 		printer.P("\n%v", sn)
 		printer.P("  copy started, this may take a while...")
 		err := dstRepo.WithBlobUploader(context.TODO(), func(ctx context.Context, uploader restic.BlobSaver) error {
-			if err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader, seenBlobs); err != nil {
+			if err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader); err != nil {
 				return err
 			}
 			debug.Log("tree copied")
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -63,11 +63,9 @@ Exit status is 12 if the password is incorrect.
 type CopyOptions struct {
 	global.SecondaryRepoOptions
 	data.SnapshotFilter
-	batch bool
 }
 
 func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
-	f.BoolVar(&opts.batch, "batch", false, "batch all snapshots to be copied into one step to optimize use of packfiles")
 	opts.SecondaryRepoOptions.AddFlags(f, "destination", "to copy snapshots from")
 	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
 }
@@ -161,7 +159,7 @@ func runCopy(ctx context.Context, opts CopyOptions, gopts global.Options, args [
 
 	selectedSnapshots := collectAllSnapshots(ctx, opts, srcSnapshotLister, srcRepo, dstSnapshotByOriginal, args, printer)
 
-	if err := copyTreeBatched(ctx, srcRepo, dstRepo, selectedSnapshots, opts, printer); err != nil {
+	if err := copyTreeBatched(ctx, srcRepo, dstRepo, selectedSnapshots, printer); err != nil {
 		return err
 	}
 
@@ -187,35 +185,33 @@ func similarSnapshots(sna *data.Snapshot, snb *data.Snapshot) bool {
 	return true
 }
 
-// copyTreeBatched: copy multiple snapshot trees in one go, using calls to
-// repository.RepackInner() for all selected snapshot trees and thereby packing the packfiles optimally.
-// Usually each snapshot creates at least one tree packfile and one data packfile.
+// copyTreeBatched copies multiple snapshots in one go. Snapshots are written after
+// data equivalent to at least 10 packfiles was written.
 func copyTreeBatched(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
-	selectedSnapshots []*data.Snapshot, opts CopyOptions, printer progress.Printer) error {
+	selectedSnapshots []*data.Snapshot, printer progress.Printer) error {
 
 	// remember already processed trees across all snapshots
 	visitedTrees := restic.NewIDSet()
-	batchSize := 1
-	if opts.batch {
-		batchSize = len(selectedSnapshots)
-	}
 
 	for len(selectedSnapshots) > 0 {
 		var batch []*data.Snapshot
+		batchSize := uint64(0)
+		targetSize := uint64(dstRepo.PackSize()) * 10
 		// call WithBlobUploader() once and then loop over all selectedSnapshots
 		err := dstRepo.WithBlobUploader(ctx, func(ctx context.Context, uploader restic.BlobSaver) error {
-			for len(selectedSnapshots) > 0 && len(batch) < batchSize {
+			for len(selectedSnapshots) > 0 && batchSize < targetSize {
 				sn := selectedSnapshots[0]
 				selectedSnapshots = selectedSnapshots[1:]
 				batch = append(batch, sn)
 
 				printer.P("\n%v", sn)
 				printer.P("  copy started, this may take a while...")
-				err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader)
+				sizeBlobs, err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader)
 				if err != nil {
 					return err
 				}
 				debug.Log("tree copied")
+				batchSize += sizeBlobs
 			}
 
 			return nil
@@ -237,7 +233,7 @@ func copyTreeBatched(ctx context.Context, srcRepo restic.Repository, dstRepo res
 }
 
 func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
-	visitedTrees restic.IDSet, rootTreeID restic.ID, printer progress.Printer, uploader restic.BlobSaver) error {
+	visitedTrees restic.IDSet, rootTreeID restic.ID, printer progress.Printer, uploader restic.BlobSaver) (uint64, error) {
 
 	wg, wgCtx := errgroup.WithContext(ctx)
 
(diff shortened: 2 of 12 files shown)


cmd/restic/cmd_copy.go
```diff
@@ -109,7 +114,8 @@ func runCopy(ctx context.Context, opts CopyOptions, gopts GlobalOptions, args []
 	}
 
 	dstSnapshotByOriginal := make(map[restic.ID][]*restic.Snapshot)
-	for sn := range FindFilteredSnapshots(ctx, dstSnapshotLister, dstRepo, &opts.SnapshotFilter, nil) {
+	nullFilter := &restic.SnapshotFilter{}
+	for sn := range FindFilteredSnapshots(ctx, dstSnapshotLister, dstRepo, nullFilter, nil) {
```
MichaelEischer: What is the purpose of this change?

cmd/restic/cmd_copy.go
```diff
@@ -120,9 +126,13 @@ func runCopy(ctx context.Context, opts CopyOptions, gopts GlobalOptions, args []
 		return ctx.Err()
 	}
 
+	if opts.streamAll {
```
MichaelEischer: Duplicating most of the code of the check command is completely unmaintainable. Instead let's do something else: extend `copyTree` that is can copy multiple root trees at once. Then all that's needed for switching between both modes is a switch that control whether a single or multiple trees are passed to the function. `copyTree` then ideally should also take on the responsibility of saving the new snapshot, that is it should become a `copySnapshots` function.

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +248,124 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+	}
+
+	visitedTrees := restic.NewIDSet()
+	return copyTreeStreamed(ctx, srcRepo, dstRepo, visitedTrees, rootTreeIDs, selectedSnapshots, quiet, opts.dryRun)
+}
+
+// copyTreeStreamed: copy multiple snapshot trees in one go, using one call to
+// repository.Repack() for all selected snapshot trees and thereby packing the packfiles optimally.
+// Usually each snapshot creates one index file and at least one tree packfile.
+func copyTreeStreamed(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
```
MichaelEischer: This MUST be unified with `copyTree`.

cmd/restic/cmd_check_integration_test.go
```diff
@@ -29,8 +29,8 @@ func testRunCheckOutput(gopts GlobalOptions, checkUnused bool) (string, error) {
 	gopts.stdout = buf
 	err := withTermStatus(gopts, func(ctx context.Context, term *termstatus.Terminal) error {
 		opts := CheckOptions{
-			ReadData:    true,
-			CheckUnused: checkUnused,
+			ReadData: true,
+			//CheckUnused: checkUnused,
```
MichaelEischer: Revert all change in this file.

changelog/unreleased/issue-5453
```diff
@@ -0,0 +1,15 @@
...
+
+When using the new option `--stream`, `restic copy` now creates one large request list
+which contains all blobs from all snapshots to be copied and then executes the
+copy operation.
+
+A summary of how many blobs, their sizes and the packfiles they are contained in is printed.
+
+https://github.com/restic/restic/issues/5453
+https://github.com/restic/restic/issues/5175
+https://github.com/restic/restic/pull/nnnn
```
MichaelEischer: Should contain the actual number of this PR.

cmd/restic/cmd_copy.go
```diff
@@ -238,6 +248,124 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+	err := wg.Wait()
+	if err != nil {
+		return err
+	}
+
+	printStats(srcRepo, copyBlobs, packList, selectedSnapshots)
+
+	if dryRun {
+		for _, sn := range selectedSnapshots {
+			Verbosef("would have saved snapshot %s saved\n", sn.ID().Str())
```
```suggestion
			Verbosef("would have copied snapshot %s\n", sn.ID().Str())
```

cmd/restic/cmd_copy.go
```diff
@@ -252,5 +380,54 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+			return err
+		}
+		Verbosef("snapshot %s saved\n", newID.Str())
+	}
+
 	return nil
 }
+
+// printStats: generate short statistic for blob / packs / snapshots to be copied
+func printStats(srcRepo restic.Repository, copyBlobs restic.BlobSet, packList restic.IDSet, selectedSnapshots []*restic.Snapshot) {
```
MichaelEischer: Those stats are also relevant for the snapshot-by-snapshot mode.

cmd/restic/cmd_copy.go
```diff
@@ -57,12 +58,16 @@ Exit status is 12 if the password is incorrect.
...
 	secondaryRepoOptions
+	streamAll bool
+	dryRun    bool
 	restic.SnapshotFilter
 }
 
 func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
 	opts.secondaryRepoOptions.AddFlags(f, "destination", "to copy snapshots from")
 	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
+	f.BoolVar(&opts.streamAll, "stream", false, "stream all snapshots to be copied in one large copy step")
```
MichaelEischer: What about `--batch` as option name? `--stream` is somewhat weird as the copy operation itself is streamed in either variant.

MichaelEischer (changes requested):
The `--stream` and non-stream code paths MUST be unified or this will become a maintenance nightmare.

MichaelEischer:
The merge conflicts are due to https://github.com/restic/restic/pull/5518 and https://github.com/restic/restic/pull/5525 . Just ping me if you need help with rebasing the PR.

cmd/restic/cmd_copy.go
```diff
@@ -252,5 +380,54 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+		if err != nil {
+			return err
+		}
+		Verbosef("snapshot %s saved\n", newID.Str())
+	}
+
 	return nil
 }
+
+// printStats: generate short statistic for blob / packs / snapshots to be copied
```
MichaelEischer: Remove the stats function here and let's discuss it in https://github.com/restic/restic/pull/5464 .

cmd/restic/cmd_copy.go
wplapper: removed change

cmd/restic/cmd_copy.go
wplapper: reworked copy logic completely.

cmd/restic/cmd_copy.go
wplapper: done.

cmd/restic/cmd_check_integration_test.go
wplapper: done.

changelog/unreleased/issue-5453
wplapper: fixed.

cmd/restic/cmd_copy.go
wplapper: will be integrated later.

cmd/restic/cmd_copy.go
wplapper: done.

cmd/restic/cmd_copy.go
wplapper: removed function for now.

changelog/unreleased/pull-5472
MichaelEischer, line 1: This  changelog file is redundant.

internal/repository/repack.go
```diff
@@ -48,20 +48,33 @@ func Repack(
...
+	// moved flushing of new packs to here
+	if err := dstRepo.Flush(wgCtx); err != nil {
 		return nil, err
 	}
+
 	return obsoletePacks, nil
 }
 
+// RepackInner is a visible wrapper around repack()
+func RepackInner(ctx context.Context,
```
MichaelEischer: This is a horrible hack, but I get that `Repack` currently doesn't allow implementing this PR. So, let's do it properly instead.

Due to https://github.com/restic/restic/pull/5558 it's no longer possible to separately start the uploader and call flush. This is instead handled automatically via `WithBlobUploader`. 

But I don't see why we couldn't extract the `WithBlobUploader` call out of `Repack`. Then the call has to pass in the `restic.BlobSaver` but that is fine IMHO. I'd then also rename `Repack` to `CopyBlobs`.

cmd/restic/cmd_copy.go
```diff
@@ -203,10 +180,13 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
 	packList := restic.NewIDSet()
 
 	enqueue := func(h restic.BlobHandle) {
-		pb := srcRepo.LookupBlob(h.Type, h.ID)
-		copyBlobs.Insert(h)
-		for _, p := range pb {
-			packList.Insert(p.PackID)
+		if !seenBlobs.Has(h.ID) {
```
MichaelEischer: I don't think we need `seenBlobs` at all. Once `repository.Repack(Inner)` has run, all copied blobs will immediately also be returned by `dstRepo.LookupBlobSize`, even if the repository was not flushed yet.

cmd/restic/cmd_copy.go
```diff
@@ -243,10 +223,110 @@ func copyTree(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Rep
...
+	visitedTrees restic.IDSet, selectedSnapshots []*data.Snapshot, opts CopyOptions,
+	printer progress.Printer) error {
+
+	// dependent on opts.batch the package Uploader is started either for
+	// each snapshot to be copied or once for all snapshots
+	var wg *errgroup.Group
+	var wgCtx context.Context
+	var seenBlobs restic.IDSet
+
+	if opts.batch {
```
MichaelEischer: With `repository.RepackInner` (or it's replacement) we can actually let copy automatically do the right thing. Now, that `copyTree` no longer flushes automatically, `copyTreeBatched` can just check that a reasonable amount of data has already been copied before flushing the repository. This should only add a minimal amount of extra code but has the benefit of offering optimal performance out of the box. 

In case you're wondering how to implement this flexible flush control with `WithBlobUploader`, take a look at the double loop at https://github.com/restic/restic/blob/c854338ad152098928c49136a2a2149db8243224/internal/repository/repack_test.go#L21 .

MichaelEischer (changes requested):
This will require further rework to be compatible with the changes from https://github.com/restic/restic/pull/5558 . While looking closer at `copyTreeBatched` I've noticed that it's actually possible to let it automatically batch multiple snapshots together. That should be possible with limited extra complexity and would have to benefit of just working out of the box.

wplapper:
In principle, one could make ``--batch`` the default option and have ``--no-batch`` as a backup for some unusual use cases.

cmd/restic/cmd_copy.go
```diff
@@ -268,3 +302,81 @@ func copyStats(srcRepo restic.Repository, copyBlobs restic.BlobSet, packList res
...
+				err := copyTree(ctx, srcRepo, dstRepo, visitedTrees, *sn.Tree, printer, uploader, seenBlobs)
+				if err != nil {
+					return err
+				}
+				debug.Log("tree copied")
+			}
+
+			// save all the snapshots
+			for _, sn := range selectedSnapshots {
+				err := copySaveSnapshot(ctx, sn, dstRepo, printer)
```
MichaelEischer: Saving the snapshot must happen _outside_ of `WithBlobUploader`, otherwise some blobs might still be missing.

cmd/restic/cmd_copy.go
```diff
@@ -268,3 +302,81 @@ func copyStats(srcRepo restic.Repository, copyBlobs restic.BlobSet, packList res
...
+}
+
+// copyTreeBatched: copy multiple snapshot trees in one go, using calls to
+// repository.RepackInner() for all selected snapshot trees and thereby packing the packfiles optimally.
+// Usually each snapshot creates at least one tree packfile and one data packfile.
+func copyTreeBatched(ctx context.Context, srcRepo restic.Repository, dstRepo restic.Repository,
+	visitedTrees restic.IDSet, selectedSnapshots []*data.Snapshot, opts CopyOptions,
+	printer progress.Printer) error {
+
+	// seenBlobs is necessary in about 1 of 10000 blobs, in the other 99.99% the check
```
MichaelEischer: This is a problem that should be properly investigated instead of adding workaround. Stating that something randomly doesn't work basically admits defeat, when something important could be broken.

MichaelEischer (changes requested):
Looks like I should have replied here a bit earlier. Adding a `--all` option is completely the opposite direction of what this PR should move into, sorry. I actually think that even a `--batch` option is already slightly too much. In particular, there's an apparent way that let's `copy` perform well even without additional configuration.

I might partially change my mind if a relevant performance benefit or other benefit can be demonstrated, but this will be an uphill battle.

As I want to merge this PR before I introduce even more merge conflicts (the next 2 or 3 are already prepared), I'll take over cleaning up the code here (including my comments).

cmd/restic/cmd_copy.go
MichaelEischer: Well, that wasn't actually the case and will be fixed by https://github.com/restic/restic/pull/5614 .

cmd/restic/cmd_copy_integration_test.go
```diff
@@ -85,6 +107,178 @@ func TestCopy(t *testing.T) {
...
+		return err
+	}))
+
+	var repo3 restic.Repository
+	var unlock3 func()
+	rtest.OK(t, withTermStatus(t, env3.gopts, func(ctx context.Context, gopts global.Options) error {
+		printer := ui.NewProgressPrinter(gopts.JSON, gopts.Verbosity, gopts.Term)
+		_, repo3, unlock3, err = openWithReadLock(ctx, gopts, false, printer)
+		rtest.OK(t, err)
+		defer unlock3()
```
MichaelEischer: unlock3 has already been called once withTermStatus returns. After that point the `repository` must not be used anymore.

cmd/restic/cmd_copy.go
wplapper: Sorry, I trusted your statement about the pending blobs. I tried to investigate the issue (since I knew that ``seenBlobs``was a circumvention), but I could not see an apparent pattern. So - in order to deliver a correct result - I used ``seenBlobs``. I never suspected that the master index was the culprit.

wplapper:
Michael, thank you for cleaning up my messes :)

wplapper:
May I point out, after copying your modified code to my git repository, that we still have a long tail of small packfiles after running ``copy (implicit --batch)``. For the copied repository it looks like this:
```
0:00] 100.00%  83 / 83 index files loaded
  compressed tree blobs        11006    11.189 MiB
  compressed data blobs        72923     8.396 GiB
  compressed ALL  blobs        83929     8.407 GiB
uncompressed tree blobs        11006    52.090 MiB
uncompressed data blobs        72923    15.084 GiB
uncompressed All  blobs        83929    15.135 GiB
tree packfiles                    83    11.622 MiB
data packfiles                   561     8.399 GiB
ALL  packfiles                   644     8.411 GiB

index files                       83     3.851 MiB
snapshot files                    83    37.402 KiB
```
whereas the original repository looks like this:
```
  compressed tree blobs        11006    11.189 MiB
  compressed data blobs        72923     8.396 GiB
  compressed ALL  blobs        83929     8.407 GiB
uncompressed tree blobs        11006    52.090 MiB
uncompressed data blobs        72923    15.084 GiB
uncompressed All  blobs        83929    15.135 GiB
tree packfiles                     1    11.619 MiB
data packfiles                   528     8.399 GiB
ALL  packfiles                   529     8.411 GiB

index files                        2     3.846 MiB
snapshot files                    83    37.157 KiB
```

In total about 110 more packfiles, and many more index files. This can easily be remedied with a ``restic prune``, but this is an extra step. What would you like me to do?
PS: The output is an overview command I have written for my pleasure. It just summarizes the the key ingredients of a restic repository.

MichaelEischer:
> In total about 110 more packfiles, and many more index files. This can easily be remedied with a `restic prune`, but this is an extra step.

Thanks for testing! Looks like I've effectively broken the PR for the snapshots you've tested with :see_no_evil: . The latest commit makes the batching quite a bit more aggressive, which should be enough to yield a single digit number of indexes and get rid of most of the difference in the pack files number. Can you give that another try?

wplapper:
Hi Michael,

I just saw your note and I have tested again. Excellent news! ``restic copy`` is now quite a bit faster and the long tail of small packfiles has completely disappeared! Here is overview summary again:
```
  compressed tree blobs        11006    11.189 MiB
  compressed data blobs        72923     8.396 GiB
  compressed ALL  blobs        83929     8.407 GiB
uncompressed tree blobs        11006    52.090 MiB
uncompressed data blobs        72923    15.084 GiB
uncompressed All  blobs        83929    15.135 GiB
tree packfiles                     1    11.619 MiB
data packfiles                   528     8.399 GiB
ALL  packfiles                   529     8.411 GiB

index files                        2     3.846 MiB
snapshot files                    83    37.157 KiB
count trees                       83
``` 

Excellent work!

MichaelEischer (approved):
LGTM. Thanks!

PR #5478, opened 2025-08-26
Refactor to use reflect.TypeFor

cuiweixie:

Checklist
---------


MichaelEischer (changes requested):
I don't get the point of this refactoring. `TypeFor` internally also calls `TypeOf`. And here the change has the downside that datatype is now hardcoded at the variable declaration of `windowsAttributes` but now also at the `TypeFor` call.

Issue #5488, opened 2025-09-06
RFE: Allow restic init to support Group accessible repositories, explicitly and implicity


brad2014:
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------
1. A --group-access argument to restic init: `restic init --group-access=restic ...` creates all repo directories in group `restic` with mode `2770`, and all files in group `restic` with mode `0440`. 
2. Implicit group access: `restic init ...` (at least for the local backend) should detect if the repo root directory or its parent is SETGUID, and if so, runs as if --group-access=*detected-group*. 

The output messages printed by restic init should note when either of these occurs.  For example,
```
root@resticrepo# restic init -r /var/restic/test --group-access=restic
enter password for new repository: 
enter password again: 
Created restic repository 72a677ae83 at /home/cache/restic/test.
**Group access enabled for group restic.**

Please note that knowledge of your password is required to access
the repository. Losing your password means that your data is
irrecoverably lost.
```

restic init would report an error if the user does not have permission to change the group as specified.  I'd expect some repo backends would support this (e.g. local and sftp), and some would not (s2 and similar).

What are you trying to do? What problem would this solve?
---------------------------------------------------------
Keying file permissions off of the mode of the config file is problematic, since:

a) it doesn't yet exist prior to init, and
b) it necessitates a manual step after init to correct all the file and directory permissions. 

It is an enhancement to avoid having users manually hack the internal repository structure and content.

The expectation is that the directory SETGUID bit is actually a signal to the universe that files below it are *intended* to be group accessible.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
restic makes me happy every day, and twice on sunday!

brad2014:
Keying off the mode of the `config` file is problematic because:

1) only files created after the config group read permissions is applied are created with group read permission.  Others (such as the key file) are not.  One ends up with a repo that has some files with group read and some without, and that must be fixed manually.
2) it doesn't fix the group and mode of directory files.

That said, the affect of the explicit or implicit group access flag is to set the `config` file mode and group so that the recent 0.14.0 revs running against the repo directory created with the --group-access flag might work (but for bugs like #5487).

MichaelEischer:
[...]

PR #5489, opened 2025-09-06
docs: fix permission setup for group-accessible repo

MichaelEischer:

The group always needs execute access for the directories. In addition, files should be always set to read-only for everyone as restic never modifies files.

Noticed while looking at https://github.com/restic/restic/issues/5488 .

Checklist
---------

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -798,12 +798,13 @@ permission bit to all repository files with ``chmod``:
 
 .. code-block:: console
 
-    $ chmod -R g+rX /srv/restic-repo
+    $ find /srv/restic-repo -type f -exec chmod 440 '{}' \;
+    $ find /srv/restic-repo -type d -exec chmod 750 '{}' \;
 
 This serves two purposes: 1) it sets the read permission bit on the
 repository config file triggering restic's logic to create new files as
 group accessible and 2) it actually allows the group read access to the
-files.
+files and directories.
 
 .. note:: By default files on Unix systems are created with a user's
           primary group as defined by the gid (group id) field in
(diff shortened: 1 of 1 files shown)


brad2014 (commented):
This sounds much better, but I wanted to mention (as a group access repo creator and user) that I was confused by the original read-only vs read-write conversation.  It may sound odd (and I hope its not inappropriate), but I think the entire section (lines 791-836) might want to be simply something like:

--- suggestion start ---

Since restic version 0.14, both local and SFTP repositories can be made accessible to all the members of a given UNIX group on the repository host.

To permit multiple users to use a repository, first run "restic init" to create it, if it doesn't already exist. Then, some manual intervention is currently required. Run the following commands over the repository files themselves, which give the required permissions (and hints to restic). Thereafter, restic commands can be run against that repository by any member of a given UNIX group.

```
# chgrp -R restic-users /srv/restic-repo
# find /srv/restic-repo -type f -exec chmod 440 '{}' \;
# find /srv/restic-repo -type d -exec chmod 2770 '{}' \;
```

Of course, you'll change `/srv/restic-repo` to the path to your repo and `restic-users` to the name of your UNIX group whose users are allowed access.

(Internally, the group read permission on the `config` file tells restic to create all future ordinary files inside the repo with group-read permission, and the setgid mode bit on directories causes restic to set the group of each newly created file to the group of its parent directory. They thus remain accessible to all group members, regardless of which user created them. Note that restic never creates or deletes directories within an initialized repo, only regular files.)

~~There are a few restic commands that don't create *any* files in the repo, and if you want to further lock down the repo to prevent group members from modifying the repo in any way, just use mode `2750` instead of `2770` above.  This disables group-write access to directories, so files can only be created or deleted by the owner.  Of course, restic subcommands that require write access will fail.~~

For remote repository access via SFTP, note that the user running the restic command on a client machine may be different from the account SFTP is logged in to.  
```
# restic backup -r sftp:restic@repohost:/srv/restic-repo
```
In the example, the local user `root` can read all the files on the client host, and send them for backup to a remote user `restic` to add them to the repository.  In this example, `restic` could be a user in the `restic-users` group on `repohost`.


--- end suggestion ---

(I hope this kind of suggestion, or this forum for it, is not inappropriate. if so, my apologies.)

MichaelEischer:
@brad2014 PRs are the ideal place to discuss changes, so the suggestion is exactly in the right place. I've incorporated a significant part of your suggestion. Please take a look.

I've noticed that the "readonly variant" described in the docs isn't actually represented in the backend implementation. Instead, it relies on the backend creating all necessary directories right at the start. This isn't something that can actually be guaranteed and is likely to break at some point in the future. As the existing docs in that regard are completely broken and nobody has complained in years, I'll just assume that it's safe to just drop it.

brad2014:
Thank you.  Looks good :+1:

PR #5490, opened 2025-09-06
docs: sync compatibility section with website

### LINKED ISSUE #5445: Inconsistency between website docs vs Sphinx docs on compatibility
It looks like the docs in `doc/` in restic's git repository (which are used to generate the Sphinx docs at https://restic.readthedocs.io/en/stable/090_participating.html#compatibility) weren't updated for https://github.com/restic/restic/issues/2588.

The website was changed in https://github.com/restic/restic.net/pull/22 but not the similar copy of the docs themselves in the main repo at https://github.com/restic/restic/blob/v0.18.0/doc/090_participating.rst#compatibility.

i.e. The [website](https://restic.net/#compatibility) currently says:
>Compatibility
>
>Backward compatibility for backups is important so that our users are always able to restore saved data. Therefore restic follows [Semantic Versioning](http://semver.org/) to clearly define which versions are compatible. The repository and data structures contained therein are considered the “Public API” in the sense of Semantic Versioning.
>
>Once version 1.0.0 is released, we guarantee backward compatibility of all repositories within one major version; as long as we do not increment the major version, data can be read and restored. We strive to be fully backward compatible to all prior versions.
>
>During initial development (versions prior to 1.0.0), maintainers and developers will do their utmost to keep backwards compatibility and stability, although there might be breaking changes without increasing the major version.

while the docs, including the [Sphinx](https://restic.readthedocs.io/en/stable/090_participating.html#compatibility) ones say:
>Backward compatibility for backups is important so that our users are always able to restore saved data. Therefore restic follows [Semantic Versioning](https://semver.org/) to clearly define which versions are compatible. The repository and data structures contained therein are considered the "Public API" in the sense of Semantic Versioning. This goes for all released versions of restic, this may not be the case for the master branch.
>
>We guarantee backward compatibility of all repositories within one major version; as long as we do not increment the major version, data can be read and restored. We strive to be fully backward compatible to all prior versions.

MichaelEischer:

Sync the text of the compatibility section in the sphinx docs with that on the website. This is no change in policy, just a more precise description of the status quo.

There are no plans for major deprecations in the repository format so far, although further minor deprecations (like the S3 legacy layout or the very old index format) may happen if necessary.

Fixes https://github.com/restic/restic/issues/5445

Checklist
---------

### DIFF
--- a/doc/090_participating.rst
+++ b/doc/090_participating.rst
@@ -130,14 +130,17 @@ Backward compatibility for backups is important so that our users are
 always able to restore saved data. Therefore restic follows `Semantic
 Versioning <https://semver.org>`__ to clearly define which versions are
 compatible. The repository and data structures contained therein are
-considered the "Public API" in the sense of Semantic Versioning. This
-goes for all released versions of restic, this may not be the case for
-the master branch.
+considered the "Public API" in the sense of Semantic Versioning.
 
-We guarantee backward compatibility of all repositories within one major
-version; as long as we do not increment the major version, data can be
-read and restored. We strive to be fully backward compatible to all
-prior versions.
+Once version 1.0.0 is released, we guarantee backward compatibility of
+all repositories within one major version; as long as we do not
+increment the major version, data can be read and restored. We strive
+to be fully backward compatible to all prior versions.
+
+During initial development (versions prior to 1.0.0), maintainers and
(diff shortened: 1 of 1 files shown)


MichaelEischer (commented):
LGTM.

PR #4339, opened 2023-05-25
backend, termstatus: fix ioctl calls for 64-bit big-endian platforms

greatroar:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Changes the way we call ioctl to get and set terminal info. The old code didn't work on some 64-bit big-endian platforms, notably linux/s390x.

See commit messages for details. This code already has tests.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

Fixes #4223.

Checklist
---------

- [x] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [x] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I have run `gofmt` on the code in all commits.
- [x] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [x] I'm done! This pull request is ready for review.

Issue #4223, opened 2023-02-25
 restic does not print information when doing backups on s390x


skriesch:
Output of `restic version`
--------------------------
```
restic version
restic 0.15.1 compiled with go1.19.5 on linux/s390x
```

How did you run restic exactly?
-------------------------------
I have installed restic on openSUSE Tumbleweed for s390x in the [LinuxONE Commutity Cloud](https://developer.ibm.com/articles/get-started-with-ibm-linuxone/), which is free for open source projects. I have chosen SLES and [upgraded](https://en.opensuse.org/ZSystems/Upgrade_SLES_to_openSUSE) the system to openSUSE as a first step.

Then I have installed restic:
`sudo zypper in restic`

When doing backups on s390x, restic only prompts for password, but does not print any more information, regardless of the verbosity. The backup seems to work though. This has been evaluated with restic-0.15.1-0.s390x.rpm. On x86_64, there is quite some additional information by default.

The following commands can be used to reproduce:

`restic init --repo /tmp/testrepo`
enter password for new repository: 
enter password again: 
created restic repository 49ad167248 at /tmp/testrepo

Please note that knowledge of your password is required to access
the repository. Losing your password means that your data is
irrecoverably lost.
`restic -r /tmp/testrepo backup ~/`
enter password for repository: 
` echo $?`
0

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
Mainframe VM in the LinuxONE Community Cloud. You can receive one for free (see link above).

Expected behavior
-----------------
You are receiving the same additional information as on x86 by default.

Actual behavior
---------------
You are asked only for the password and you are receiving the information about the created repository.

Steps to reproduce the behavior
-------------------------------
Install restic on a s390x based system or create a vm emulation with qemu.
Then use the commands above to reproduce less information in comparison to x86 as a hardware.

The single "go build" option not used on s390x is "buildmode=pie " in comparison to x86.

Do you have any idea what may have caused this?
-----------------------------------------------
Exclusion of information output for hardware specific configurations.

Do you have an idea how to solve the issue?
-------------------------------------------

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

skriesch:
This issue has been reported in our openSUSE Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1208665

PR #5493, opened 2025-09-07
backend,termstatus: Unify foreground/background detection

greatroar:

PR #5358 reintroduced a version of the TIOCGPGRP ioctl call that works on all Unix platforms except Linux, due to a bug/inconsistency in x/sys/unix. This commit fixes that by introducing termstatus.Tcgetpgrp.

It also introduces termstatus.Getpgrp to deal with the different signature of unix.Getpgrp in Solaris vs. all other Unix platforms, so that some AIX/Solaris-specific code can be removed elsewhere and foreground/background detection is done the same everywhere except on Windows.

I tested this on Linux and checked that AIX, Darwin, FreeBSD and Solaris continue to compile.

The bug in x/sys/unix was first noticed in #4223, which was fixed by #4339.

Checklist
---------


MichaelEischer:
I see you've also noticed the bug in https://github.com/restic/restic/pull/5358 . While starting to fix it, I got a bit carried away and refactored most terminal-related code into a separate package: https://github.com/restic/restic/pull/5494 .

My suggestion would be to drop the `tcgetpgrp()` hack in my PR and use the approach from here. It's probably better to sort out the ioctl mess first and then I can rebase my PR afterwards.

internal/ui/termstatus/getpgrp_sysv.go
```diff
@@ -0,0 +1,7 @@
+//go:build aix || solaris
+
+package termstatus
+
+import "golang.org/x/sys/unix"
+
+func Getpgrp() (int, error) { return unix.Getpgrp() }
```
MichaelEischer: AIX apparently doesn't return an error

greatroar: Oops. Fixed.

MichaelEischer (approved):
LGTM. Thanks for fixing this once again!

PR #5494, opened 2025-09-07
Refactor terminal background handling

MichaelEischer:
Consolidate most code that directly interacts with a terminal into `internal/terminal`.

Follow up to https://github.com/restic/restic/pull/5358 and https://github.com/restic/restic/pull/5493

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5354
+++ b/changelog/unreleased/issue-5354
@@ -10,3 +10,5 @@ This has been fixed.
 
 https://github.com/restic/restic/issues/5354
 https://github.com/restic/restic/pull/5358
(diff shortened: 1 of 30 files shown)


internal/terminal/shell_split.go
```diff
@@ -1,4 +1,4 @@
-package backend
```
greatroar: This isn't really related to terminals.

MichaelEischer: Dropped the commit.

MichaelEischer (commented):
LGTM

Issue #5467, opened 2025-08-05
Possibly erroneus "Fatal: repository contains errors" from check


Fjodor42:
Output of `restic version`
--------------------------
`restic 0.18.0 compiled with go1.24.1 on linux/amd64`

What backend/service did you use to store the repository?
---------------------------------------------------------
Rclone (rclone v1.69.3), Jottacloud backend

Problem description / Steps to reproduce
----------------------------------------

I have switched cloud storage providers, moving repository data with rclone.

I don't know if the following is an artefact of the initial transfer, problems with the repository as such, or a bug in Restic.

What I note is that:

1. The Rclone part give occasional errors, prompting retries, after which Restic reports `check successful on second attempt` and the "original error", which seems to indicate that all is well after all, but
2. In the end, Restic reports `The repository is damaged and must be repaired. Please follow the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html .`

It's the mismatch between `check successful on second attempt` in part 1. and `The repository is damaged and must be repaired. Please follow the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html .` that confuses me.

It is possible that part 2. stems from an actual error with the data, but as far as I read the output, the only errors reported are the "original errors" from the attempt where Rclone errored out, but I don't feel confident stating whether the final error message is correct or not.

Command:
```
restic -r rclone:jottacloud:///<redacted> -o rclone.args='serve restic --stdio --retries 10 --transfers 8' -check --read-data -verbose --password-file <redacted>
```

Output:
```
using temporary cache in /tmp/restic-check-cache-2026371542
create exclusive lock for repository
load indexes
check all packs
check snapshots, trees and blobs
[0:12] 100.00%  17 / 17 snapshots
read all data
rclone: 2025/08/05 15:32:43 ERROR : data/9c/9ca4ca33715e9be047c89fedbe724435e8d8f3a1cee7ea0b2a6d524aebfede3e: Didn't finish writing GET request (wrote 30901086/33842161 bytes): read tcp [2a06:4005:608b:1:6a05:caff:feb0:7367]:51618->[2a0a:cc80:0:c2a8:2::35]:443: read: connection reset by peer
Load(<data/9ca4ca3371>, 33842161, 0) returned error, retrying after 1.082796111s: readFull: unexpected EOF
[...]

PR #5495, opened 2025-09-07
check: fix error reporting on download retry

MichaelEischer:
Reset all blob decoding errors in a pack file when retrying the download. We're only interested in the result for the last attempt as otherwise the overall pack file hash check and the blob errors are inconsistent with each other.

For a perfect backend, this shouldn't make any difference as the backend would only ever return incomplete data but never garbage data. Hence the checks inside the `Load(...)` call wouldn't report errors. But if the first incomplete attempt returned garbage data but the second attempt was successful, then `check` would report an error for individual blobs, but not the overall pack file. This is highly confusing.

Partially addresses https://github.com/restic/restic/issues/5467

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5467
+++ b/changelog/unreleased/issue-5467
@@ -1,0 +1,11 @@
+Bugfix: Improve handling of download retries in `check` command
+
+In very rare cases, the `check` command could unnecessarily report a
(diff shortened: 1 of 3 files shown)


MichaelEischer (commented):
LGTM

PR #5496, opened 2025-09-07
docs: add info about ssd wear in backend connections

DoS007:
It occured to me while reading the docs, that more concurrent requests lead potentially to more ssd wear (and not only pack size). I wanted this information to be added to the docs for others to see. (My OS e.g. writes to SSD after ~30/35 sec and longer uploading times increase the potential for that - and not only pack size - if i understand correctly, do I ***?***).

I did a quick search and didn't see anything relevant.

Checklist
---------

This is a *very* simple pull request. I'm here just a restic end user. So please bear with me.

- [x] I have added tests for all code changes.: *I added no program logic.*
- [x] I have added documentation for relevant changes (in the manual).:  *Not for relevant changes in the code*
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).: *I think i need a pull request number before i can add the Changelog entry. So i can only do that, after making this request?*

### DIFF
--- a/changelog/unreleased/pull-5496
+++ b/changelog/unreleased/pull-5496
@@ -1,0 +1,9 @@
+Enhancement: Docs about backend connections indicate potentiel effect on ssd wear
+
+In the docs, in chapter `Tuning Backup Parameters`, `backend connections` potential effect 
+for increased upload time (per single pack), when using more connections, is now pointed out.
+And also the effect for potential SSD wear, because the temporary pack files might stay 
+longer in TMP while uploading.
+
+
+https://github.com/restic/restic/pull/5496
--- a/changelog/unreleased/pull-5496
+++ b/changelog/unreleased/pull-5496
@@ -1,9 +1,0 @@
-Enhancement: Docs about backend connections indicate potentiel effect on ssd wear
-
-In the docs, in chapter `Tuning Backup Parameters`, `backend connections` potential effect 
-for increased upload time (per single pack), when using more connections, is now pointed out.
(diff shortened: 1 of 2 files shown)


DoS007:
The link syntax in the docs should be hopefully correct (the link to Pack Size). But would be good if someone could veryify just to be sure and also the truth of said info the doc; so that more connections may lead to less spead for each connection, which leads to longer upload times, which leads potentially to more ssd writes.

changelog/unreleased/pull-5496
```diff
@@ -0,0 +1,9 @@
+Enhancement: Docs about backend connections indicate potentiel effect on ssd wear
```
MichaelEischer: We generally don't a changelog entries for documentation changes.

doc/047_tuning_backup_parameters.rst
```diff
@@ -39,7 +39,9 @@ the REST backend the parameter would be ``-o rest.connections=5`` or for the loc
 except for the local backend which uses a limit of ``2``. The defaults should work well in
 most cases. For high-latency backends it can be beneficial to increase the number of 
 connections. Please be aware that this increases the resource consumption of restic and 
-that a too high connection count *will degrade performance*.
+that a too high connection count *will degrade performance* and can also result in longer
```
MichaelEischer: Please split this into two sentences.

MichaelEischer (changes requested):
Thanks, I just have two small comments.

MichaelEischer (approved):
LGTM.

Issue #5487, opened 2025-09-05
Incorrect file permissions after backup via sftp


brad2014:
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
sftp 

Problem description / Steps to reproduce
----------------------------------------
Similar to #4447, but affecting backup files.

"restic backup" is creating data files with user (and for my case group) read-write permission when the repo is accessed via sftp:restic@repohost, while the local files (on repohost) are all read only.

For example, the first of these data files was created during a local backup (as root on repohost), and the other during a sftp backup (to sftp:restic@repohost:...), both to the same repo.

```
1887814808    628 -r--r-----   1 root     restic     640063 Sep  5 22:39 ./index/5bebf47942e1d47fd25453f42c3d53dcd43959cc4898e7f27230307b9ea93405
1612200155     28 -rw-rw----   1 restic   restic      26374 Sep  5 22:58 ./index/9587e9696220eda5f2611ca508e6e5b5ac7c27faca544ee84e96f66434abb61c
```

Expected behavior
-----------------
All repo files uploaded by sftp should have the same permission as if the repo file was created by a local backup.

Actual behavior
---------------
Repo files have read-write permission instead of read-only.

Do you have any idea what may have caused this?
-----------------------------------------------
Probably something similar to #4447.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
restic makes me happy every day (at 3am).

greatroar:
Are the restic versions used to make the files the same?

This can hardly be the same problem as #4447, because in this instance, restic does an explicit chmod after creating the file. I'm not sure why it adds write permission; this happens in internal/backend/util/paths.go, but the discussion surrounding that code in #3419 doesn't mention it (or I'm glossing over it...).

brad2014:
## Reproducible test case

Context: 

Both restic server (below, `repohost`) and restic client (`clienthost`) are using restic 0.18.0.

The test case initializes repository /var/restic/test, adjusts it for [Group Accessible Repository](https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#group-accessible-repositories) 
(though its instructions are outdated), then performs a backup on server, then on client.

Parent directory repohost:/var/restic has group restic mode g+s
Repohost is Rocky 9, with its packaged ssh/sftp server.
 
[...]

PR #5497, opened 2025-09-07
Bugfix(sftp): fix loose permissions on sftp backend.

ImTheCurse:
- make file permissions to read-only on SFTP back-end.

this fixes the issue of loose permission, an unnecessary write permission after writing the file onto the SFTP back-end.

yes, it was discussed on #5487 

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5487
+++ b/changelog/unreleased/issue-5487
@@ -1,8 +1,8 @@
-
 Bugfix: fixed loose permissions of the sftp backend.
 
 Restic now restricts the file permissions on sftp backend to readonly.
-In restic 0.18.0 using backup would allow group write permissions on files.
+Now, using backup in restic would allow group write permissions on files.
 This change only has an effect for sftp servers with support for the chmod operation.
 
 https://github.com/restic/restic/issues/5487
+https://github.com/restic/restic/pull/5497
--- a/changelog/unreleased/issue-5487
+++ b/changelog/unreleased/issue-5487
@@ -1,0 +1,8 @@
+
+Bugfix: fixed loose permissions of the sftp backend.
+
+Restic now restricts the file permissions on sftp backend to readonly.
+In restic 0.18.0 using backup would allow group write permissions on files.
+This change affects sftp servers that support the chmod operation, otherwise keeps it the same as before.
+
+https://github.com/restic/restic/issues/5487
--- a/changelog/unreleased/issue-5487
+++ b/changelog/unreleased/issue-5487
@@ -3,6 +3,6 @@ Bugfix: fixed loose permissions of the sftp backend.
 
 Restic now restricts the file permissions on sftp backend to readonly.
 In restic 0.18.0 using backup would allow group write permissions on files.
-This change affects sftp servers that support the chmod operation, otherwise keeps it the same as before.
+This change only has an effect for sftp servers with support for the chmod operation.
 
 https://github.com/restic/restic/issues/5487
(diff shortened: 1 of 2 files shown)


internal/backend/sftp/sftp.go
```diff
@@ -287,6 +287,20 @@ func tempSuffix() string {
 	return hex.EncodeToString(nonce[:])
 }
 
+func setReadOnly(client *sftp.Client, path string, mode os.FileMode) error {
```
brad2014, line 290: A minor thing, but could we call it setFileReadOnly to match the function name used in backend/local, so that filename search finds the two corresponding functions together (as with the other function names here)?

internal/backend/sftp/sftp.go
```diff
@@ -287,6 +287,20 @@ func tempSuffix() string {
...
+	// clear owner/group/other write bits
+	readonlyMode := mode &^ 0o222
```
brad2014, line 292: 0o222 -> 0222

ImTheCurse:
@MichaelEischer mind reviewing this?

changelog/unreleased/issue-5487
```diff
@@ -0,0 +1,8 @@
+
+Bugfix: fixed loose permissions of the sftp backend.
+
+Restic now restricts the file permissions on sftp backend to readonly.
+In restic 0.18.0 using backup would allow group write permissions on files.
```
MichaelEischer: That behavior isn't limited to a specific restic version.

changelog/unreleased/issue-5487
```diff
@@ -0,0 +1,8 @@
...
+This change affects sftp servers that support the chmod operation, otherwise keeps it the same as before.
```
MichaelEischer:
```suggestion
This change only has an effect for sftp servers with support for the chmod operation.
```

changelog/unreleased/issue-5487
```diff
@@ -0,0 +1,8 @@
...
+
+https://github.com/restic/restic/issues/5487
```
MichaelEischer, line 7: Also add a link to this PR.

changelog/unreleased/issue-5487
```diff
@@ -0,0 +1,8 @@
+
```
MichaelEischer: Unexpected newline

internal/backend/sftp/sftp.go
```diff
@@ -287,6 +287,20 @@ func tempSuffix() string {
...
+	err := client.Chmod(path, readonlyMode)
+
+	if err != nil {
```
MichaelEischer: The outer if is redundant. `errors.Is` can handle `err == nil`.

internal/backend/sftp/sftp.go
MichaelEischer:
```suggestion
func setFileReadonly(client *sftp.Client, path string, mode os.FileMode) error {
```

MichaelEischer: Let's keep the function name of the function this is based on.

MichaelEischer (changes requested):
The approach is fine in general, but I have a few nits that have to be addressed.

ImTheCurse:
fixed nitpicks, should be good now @MichaelEischer

MichaelEischer (approved):
LGTM. Thanks!

MichaelEischer:
Fixed the CI failure due to the changelog.

PR #5510, opened 2025-09-15
Replace Printf/Verbosef/Warnf with termstatus

MichaelEischer:
Restic currently has two ways to print to stdout. The first is using the global functions Printf/Verbosef/Warnf (+variants). The newer way is to use a `termstatus.Terminal`, which provides proper support for status bars. This PR converts all users of the global functions to use termstatus. A follow-up PR will further reduce the input/output handling by removing stdout/stderr from globalOptions and merging those into termstatus.Terminal along with the corresponding helpers.

`Printf` is replaced with `printer.S`, `Verbosef` with `printer.P` and `Verboseff` with `printer.V`. Those functions don't need a trailing newline character, so it's removed except for cases where multiple newlines are intended. Progress bars are now consistently created using termstatus. The changes are mostly mechanical, such that there should be no major change in behavior for users. The mechanical changes were made using Cursor.

There is a new `OutputRaw()` method for `termstatus.Terminal`, which is also no longer buffered. The new output method is intended for commands that need precise control over how data is printed on stdout. This primarily applies to the `cat`, `dump` and `find` commands.

Test cases now consistently use `withTermstatus` to setup the necessary `termstatus.Terminal` object. While `setupTermstatus()` could also be used for now, it will become mandatory with the follow-up PR.

Commands that do not support JSON on any meaningful way, do not pass the JSON flag to `newTerminalProgressPrinter`. All others do.

TODOs:
- [x] check JSON output behavior of functions
- [x] basic manual test of all modified commands

No. I wanted to unify the output code for quite a while, but didn't get around to it.

Checklist
---------

### DIFF
--- a/cmd/restic/cleanup.go
+++ b/cmd/restic/cleanup.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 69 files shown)


MichaelEischer (commented):
LGTM.

PR #3189, opened 2020-12-22
ui/progress: Use mutex instead of atomic

fd0:
The counter value needs to be aligned to 64 bit in memory for the
atomic functions to work on some platform (such as 32 bit ARM).

The atomic package says in its documentation:

> These functions require great care to be used correctly. Except for
> special, low-level applications, synchronization is better done with
> channels or the facilities of the sync package.

This commit replaces the atomic functions with a simple sync.Mutex, so
we don't have to care about alignment.

What does this PR change? What problem does it solve?
-----------------------------------------------------

Fixes the crash on arm32.

Was the change discussed in an issue or in the forum before?
------------------------------------------------------------

Closes #3185

Alternative implementation to #3187

Checklist
---------

- [x] I'm done, this Pull Request is ready for review

Issue #3185, opened 2020-12-22
prune --dry-run on ARM fails with segmentation violation


demlak:
Hardware:
--------------------------
"cubietruck 2" with 1GB RAM

Output of `restic version`
--------------------------
restic 0.11.0-dev (compiled manually) compiled with go1.15.6 on linux/arm

How did you run restic exactly?
--------------------------

```
# sudo restic_v0.11.0-207-gf7c7c2f7_linux_arm -r repo prune --dry-run
enter password for repository: 
repository 551da140 opened successfully, password is correct
loading indexes...
loading all snapshots...
finding data that is still in use for 11 snapshots
panic: runtime error: invalid memory address or nil pointer dereference
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11ce8]

goroutine 14 [running]:
runtime/internal/atomic.goLoad64(0x1824294, 0x0, 0x1a6e200)
        /usr/local/go/src/runtime/internal/atomic/atomic_arm.go:131 +0x1c
github.com/restic/restic/internal/ui/progress.(*Counter).get(...)
        /home/beta/restic.git/internal/ui/progress/counter.go:69
github.com/restic/restic/internal/ui/progress.(*Counter).run.func1(0x1824270)
        /home/beta/restic.git/internal/ui/progress/counter.go:75 +0x24
panic(0x88b8b8, 0xf0e838)
        /usr/local/go/src/runtime/panic.go:969 +0x158
runtime/internal/atomic.goLoad64(0x1824294, 0x1af773c, 0x3)
        /usr/local/go/src/runtime/internal/atomic/atomic_arm.go:131 +0x1c
github.com/restic/restic/internal/ui/progress.(*Counter).get(...)
        /home/beta/restic.git/internal/ui/progress/counter.go:69
github.com/restic/restic/internal/ui/progress.(*Counter).run(0x1824270)
        /home/beta/restic.git/internal/ui/progress/counter.go:90 +0x6c
created by github.com/restic/restic/internal/ui/progress.New
        /home/beta/restic.git/internal/ui/progress/counter.go:46 +0x10c
```

What backend/server/service did you use to store the repository?
--------------------------
I used restic rest-server to store the repo

fd0:
@greatroar can you please have a look? I had a quick look at the code but it looks correct. IMHO we should switch to using a mutex instead of using the `atomic` package. Thoughts? :)

[...]

PR #5511, opened 2025-09-16
ui/progress: Restore atomics in Counter

greatroar:

We switched ui/progress.Counter from atomics to a mutex in #3189 because of an alignment bug, but the new-style atomic types don't need manual alignment.

Also included is a patch for the related ui/progress.Updater, which had a comment pointing out a potential bug. The comment has become a test and the code being tested is simpler than it was.

#3185 was the alignment bug.

Checklist
---------

### DIFF
--- a/internal/ui/progress/counter.go
+++ b/internal/ui/progress/counter.go
@@ -1,7 +1,7 @@
 package progress
 
 import (
-	"sync"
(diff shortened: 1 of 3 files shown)


MichaelEischer (approved):
LGTM. Thanks!

PR #5514, opened 2025-09-17
ui: collect Quote and Truncate helpers

MichaelEischer:
Collect ui formatting helpers in the ui package

Minor refactoring. It makes little sense to have functionality like termstatus.Truncate but also ui.DisplayWidth in different packages. Thus, collect them all in one place.

No.

Checklist
---------

### DIFF
--- a/internal/ui/backup/text.go
+++ b/internal/ui/backup/text.go
@@ -8,7 +8,6 @@ import (
 	"github.com/restic/restic/internal/archiver"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/ui"
-	"github.com/restic/restic/internal/ui/termstatus"
 )
 
 // TextProgress reports progress for the `backup` command.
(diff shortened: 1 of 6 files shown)


greatroar:
LGTM.

Note that there are two definitions of a rune's display width in a single file after this change: DisplayWidth treats ambiguous runes as narrow, while Truncate treats them as wide. If DisplayWidth's semantics can be changed, the two can be merged and DisplayWidth can reuse the faster code of Truncate.

MichaelEischer (commented):
LGTM.

Let's keep the separate wide rune interpretations. Several of the test patterns from `TestDisplayWidth` have length 3 but would increase to 4 when considering ambiguous runes as wide. However, all of them are displayed with width 3 when pasting them into my terminals. So at least for me this is the more correct interpretation.

For `Truncate`, however, I don't see a problem with being pessimistic. There the important thing is to limit the display width to at most the given width, which is guaranteed by the current behavior that treats characters as wide if in doubt.

PR #5525, opened 2025-09-23
Extract snapshot data types from restic package

MichaelEischer:
The restic package currently contains data types for two relatively independent parts of restic: the repository interface and the snapshot data types. Quite a while ago, I extracted the backend types from the restic package and untangled the remaining code. To wrap up this refactoring, move the snapshot data types into a new package called `data`. This ensures that the different layers of restic remain decoupled from each other.

The change itself is unfortunately one mega-commit, but slicing the change into smaller parts introduces unnecessary churn as it requires temporary import changes that have to be reverted later on.

Follow up to #4499 .

Checklist
---------


MichaelEischer (commented):
LGTM.

Sorry in advance to everyone caught in the incoming merge conflict fallout. 95% of all cases can be solved by keeping your change and just replacing `restic` with `data` in all places where the compiler/linter complains.

Issue #4736, opened 2024-03-26
Feature Request: include 'bytes added to repository' in progress line while backing up


mhansen:
```
$ restic version
restic 0.16.4 compiled with go1.21.6 on darwin/arm64
```

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Restic has a curses-style CLI status line that updates a few times a second as the backup is happening, with time, total scanned size, total directory size. e.g.:

```
[0:04] 160 files 142.814 MiB, total 429 files 295.206 MiB, 0 errors
```

I propose adding the 'total bytes added to repository so far' statistic to this line.

This is printed out at the end:

```
Added to the repository: 5.677 GiB (5.569 GiB stored)
```

But it would be reassuring to also print this during the backup.

As an extension, most of the other statistics would be great to show during the backup too:

```
Files:       168196 new,     0 changed,     0 unmodified
Dirs:          318 new,     0 changed,     0 unmodified
Data Blobs:   9439 new
Tree Blobs:    319 new
Added to the repository: 5.677 GiB (5.569 GiB stored)
```

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I'm backing up my Google Photos with Google Takeout every 2 months. Any given backup has very little new data. I'd like to see while I'm backing up how much new data has been added to the repository, so (mostly) I can be confident that it's working, and (less likely) Ctrl-C if it's backing up too much new stuff.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yeah, restic rules. Love that it's open-source, deduping, cross-platform, and fast. It's how I'm backing up my Google Photos: https://www.markhansen.co.nz/google-photos-backup/

Code pointers
---

Here's the code that prints the status line: https://github.com/restic/restic/blob/d1d773cfcd3115aecbbd6ad6be2bc5e11f395b29/internal/ui/backup/text.go#L51

I see this doesn't have easy access to the total bytes added to the repository; it would need to be plumbed from https://github.com/restic/restic/blob/master/internal/ui/backup/progress.go#L88 (and even further back).

I see this is used by both backup and restore commands. Maybe not all statistics make sense for both commands.

MichaelEischer:
After https://github.com/restic/restic/pull/4705 is merged, this shouldn't be too hard to implement. The console output of `backup` and `restore` are independent enough to not pose a problem here.

However, I can imagine that the output could be too verbose by default, but it is probably okay for `backup -v`.

Issue #3431, opened 2021-06-16
Show on-disk file size in addition to 'added to repo' in backup summary


OJFord:
Output of `restic version`
--------------------------

restic 0.12.0 compiled with go1.15.8 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

A successful `restic backup` outputs a summary that ends with something like:
> Added to the repo: 12.345 MiB

It'd be nice if this showed both the 'real' size of added/changed/removed files, as on local disk, in addition to the amount 'added to the repo', which could be significantly less.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

It'd just be a nice at-a-glance summary of how well the added files, in particular, were covered by extant blocks.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Just a suggestion from yet another happy user.

calr0x:
I'd love to see this as well!
Thanks for the great project!

OJFord:
I'm really confused now - I thought I was experiencing this the other way around, but vaguely recalled this issue (didn't think I'd opened it! very short memory..).

Either way I think it'd be useful to show both, but which is it currently, what does 'Added to repo' mean:
1. The cumulative file size of new/modified/deleted files in the snapshot; or
2. The additional remote storage required by the snapshot, after block-wise de-duplication, etc.

? I think it could be clearer which it means even if it's not wanted (or not possible/too difficult) to add the 'other one'.

MichaelEischer:
Restic reports how much deduplicated data was uploaded to the repository, without accounting for overheads due to encryption and other repository metadata. That is its roughly option 2.

OJFord:
Thanks @MichaelEischer. I just noticed though, after `restic backup`:
> Added to the repo: 2.115 GiB

and then `restic diff <parent> <new>`:
>   Added:   2.115 GiB
>   Removed: 1.727 GiB

is the reason 'added to the repo' is not net of the 'removed' figure that all of what was removed is still referenced by at least the parent snapshot, so we added 2.115GiB worth of new (de-duplicated) blocks, removed references to 1.727GiB worth of old ones, but until a prune, `restic backup` itself will only ever _add_, and always equal to the 'added' figure in the resulting diff?

Many thanks.

MichaelEischer:
> > Added:   2.115 GiB
> > Removed: 1.727 GiB
> 
[...]

PR #5529, opened 2025-09-25
Rechunk copy feature

### LINKED ISSUE #5473: Rechunk copy
Output of `restic version`
--------------------------

restic 0.18.0 compiled with go1.24.1 on linux/amd64

Currently, `restic copy` executes blob-to-blob copy without touching the blob content. So, if you `copy` snapshots across repositories with different chunking parameters, then deduplication by content-defined chunking is [impaired](https://restic.readthedocs.io/en/stable/045_working_with_repos.html#copying-snapshots-between-repositories). 

Restic's recommended usage is that you init your repos with `--copy-chunker-params` so that all your repositories generate identical set of chunk blobs for given data. That way, there is no problem with `restic copy`. However, there are cases that you have to merge repositories which were originally intended for separate use. Or, you may not have been aware of `--copy-chunker-params` when you first started using restic.

In these cases, current workaround solution to ensure deduplication is doing re-backup, where you restore snapshots to somewhere and backup them to the dst repository. Although the method is possible, there are many inconveniences. You need additional space; the process is inefficient (files of identical contents are processed multiple times); metadata are tampered with; and last but not least, it is bothersome.

To help with those cases, I suggest `rechunk-copy` feature.
It will work as follows: First, it gathers file contents that are included in a snapshot (in src repo). Then a rechunk worker streamreads each (unique) file content and feeds it to the chunker (with dst repo's chunker params), and uploads the output chunks to the dst repo. It keeps the map of (srcContent -> dstContent) while doing this. After the rechunking is done, it rewrites the tree (change Content of each file node) with the map it keeps.

It will help users who are going to merge the repositories with different chunking parameters, or migrate the repository to ones with unified chunking parameters.

I have implemented my own version of `rechunk-copy` and been using it in my personal repositories. But I'm not sure if it meets restic project's quality standard yet. Still, it would be happy if I can contribute to the project.

Yes! Restic makes managing my data much easier and efficient.
Well-designed architecture, reliable performance, multi platform support (for both client & backend), active community, and it is open source!
All these make me want to stick to restic.

dgkmd:

It supports `copy --rechunk` feature which rechunks data while copying snapshots between repositories, as discussed in #5473.

The code consists of main `rechunker` module and a few submodules as follows: blob cache, scheduler, and worker; along with modification to `cmd_copy.go`.

Main `rechunker` module handles overall workflow, that is, to rechunk-copy snapshots. 

Blob cache acts as a transparent cache layer of the source repo. It is needed because the blobs are scattered across packs non-contiguously while rechunk process needs stream-reading the file.

Scheduler manages file processing order. Specifically, it prioritizes files whose all consisting blobs are ready in the blob cache, thus minimizing redundant pack download.

Worker rechunks each assigned file. It can be run multi-threaded and also can run standalone without upper layer modules, if needed (e.g., to rechunk single file).

The feature can be used with `copy --rechunk` command. It has two additional arguments on top of plain `copy`. `--cache-size` controls maximum blob cache size in MiB, and `--add-tag` adds tags of the names specified to the created snapshots.

For safety, `copy --rechunk` can be run only if `RESTIC_FEATURES=rechunk-copy` environment variable is set.

Closes #5473 
https://forum.restic.net/t/is-it-possible-to-re-chunk-after-a-restic-copy/6072
https://forum.restic.net/t/copy-snapshots-between-repositories-without-copy-chunker-params/7320

Checklist
---------


wplapper:
Hi dgkmd,
Without ``--use-pack-cache`` you are still suffering from the same performance problem which I discovered earlier during my tests. In today's test I use a local (empty) repository as target and rclone:onedrive:xxx as my source repository. I try to copy a small snapshot  with ~4,000 data blobs and a size of uncompressed 41.6 MiB, compressed 10.7 MiB. My read performance is about 7 MiB/s, so the reading of all blobs should take a few seconds.  The normal copy takes all in all ~13 seconds, and I rechunk-copied about 60% after 1,000 seconds. I will have a closer look later where the command gets stuck. ``. 

I will be testing with the option ``--use-pack-cache`` later..

dgkmd:
Hi @wplapper ,
Thank you for the review.

Yes, it is expected that there is a performance problem with remote source repositories when used without `--use-pack-cache`.
I changed use-pack-cache to opt in (default off), because the pack cache may consume huge amount of memory (a few GBs), potentially causing OOM in low resource machines.

Please let me know how is the performance when you tested with `--use-pack-cache`!

wplapper:
Hi dgkmd,

I have tested your code with the option ``--use-pack-cache`` and it works correctly. However the memory consumption is massive. Since I do not wish to go through all your code line by line, I would appreciate if you could add debug.Log() statements liberally. I am particularly interested how you are managing your LRU cache, since could release the cache held by a packfile as soon it is not used anymore. I copied a snapshot with 10874 files, size 4.375 GiB, and the memory consumption was about 5 GiB at the end of the run. I ran the command with DEBUG enabled: at the end of the run you had to reload the same packfile several times, different sections, to complete the outstanding operations. I my opinion this should not happen if you managed the cache properly. 

It would clearly help to mark the different sections of the command execution with debug.Log() statements: NewRechunker(), Plan(), RechunkData(), RewriteTree(). Also add debug.Log() statements when creating/accessing/removing cache entries.

Have you thought about files which contain only a single data blob? in my environment I have plenty of those. You should be able to copy those files without reading the file, rechunking it, only to find out that it is identical to the input blob you just read. And those files do not have to be copied at all if they already exist in the target repository, just check the target repository index for this condition.
PS.: This is not a thorough scientific analysis, I just compared input an output data blobs and found them identical.

Thanks for all your work, Winfried

dgkmd:
Hi @wplapper ,
Thank you for your thorough feedback.

Pack cache works as follows.
1. When a blob `A` is requested and the blob is *not* in the cache:
    1) Locates the pack `P` in which the blob `A` resides.
    2) Downloads all blobs in the pack `P` that are referenced by at least one file, with `Repository.LoadBlobsFromPack`.
    3) Stores the blobs to the cache.
    4) If the number of stored **packs** exceeds LRU_SIZE, it evicts the oldest pack `P'`. Which means, it evicts all blobs loaded from `P'` from the cache.
    5) Return `A` through the async channel.
2. When a blob `A` is requested and the blob *is* in the cache:
    1) Updates LRU recency of the pack `P` in which the blob resides.
    2) Immediately returns the blob `A`.

Additionally, it implements scheduling algorithm where it prioritizes processing the files whose blobs are all ready in the cache.

The memory usage is proportional to the pack cache size, which is hard coded in LRU_SIZE variable. The empirical memory usage estimation is `25 MiB * LRU_SIZE`, which is about 5 GiB in current setting (`LRU_SIZE = 200`). So at least the cache size will not explode with the repo size :)
Determining LRU_SIZE is indeed a trade-off problem. There will be no eviction if LRU_SIZE is very large, so no re-downloading packs, at the cost of huge memory consumption (and vice versa).

On handling single-blob files, we cannot guarantee it will be single-blob when chunked with different chunking parameters. Also we cannot know whether it will be chunked single-blob ahead of time. Therefore I think it's our best to just process all files...

I'll add plenty of `debug.Log()` for traceability, following your advice.
And also make no-more-needed-blobs evicted immediately and compare the performance.

Thank you again for your considerate review!

Best regards,

dgkmd:
Hi @wplapper ,

I updated the code so that it prints `debug.Log` when the PackCache downloads or evicts a pack. Also, at the end of run, it summarizes relevant information, such as (i) the list of more-than-once downloaded packs, (ii) the number of those redundant packs, (iii) how many times overall were they (redundantly downloaded packs) downloaded, and additionaly (iv) how many blob processings were saved by ChunkDict.

The number of redundant downloads is undeterministic, but there are major factors that affect the number:
1) file distribution of the repository
2) pack cache size (`LRU_SIZE`): the **larger**, the fewer redundant downloads
3) `numWorkers`: the **smaller**, the fewer redundant downloads

You could find the number of redundant downloads change as you change those two parameters (`LRU_SIZE`, `numWorkers`).
It would be ideal if restic can recognize the hardware spec at runtime and intelligently determine the two parameters, but I found that another challenging task... (do you have any good idea? :))
So as of now I just declared those two values heuristically, i.e., `numWorkers := GOMAXPROCS(0)` and `LRU_SIZE := 200`. But even in current settings, the number of redundant downloads (relative to repo size) should not be that much (in my environment was less than 10% of repo size).

Best regards,

P.S. You have to specify both rechunker.go and rechunk_cache.go in DEBUG_FILES for complete debug logs.

dgkmd:
I refined the debug logging code in my latest commit. You can get exhaustive debug logs by setting the following:
```
export RESTIC_FEATURES=rechunk-copy DEBUG_FILES=rechunker.go,rechunk_cache.go,cmd_rechunk_copy.go
```

wplapper:
Hi [dgkmd](https://github.com/dgkmd),
 I have looked at your code and it works as I expected it to work. I would like to make the following comments:

1.) Instead of evicting whole packfiles, would it make sense to evict individual data blobs, since they are smaller in size. By tracking how many times a particular blob is used, you can evict it as soon as the all occurrences have been used. That might lighten the memory load.

2.) I observed consistently fluctuating download rates, whereas I would expect a more or less constant download rate as long as there is more work to do. I use rclone:onedrive repositories as my input repository. I have had no time yet to stady this behaviour in more detail.

3.) I strongly recommend to break up the large functions into smaller, more manageable sub functions. The prime example is ``RechunkData()``which is over 500 lines long. @MichaelEischer will not like this. I think each of the Go func's would be a good candidate for such a function.

Keep up the good work!

dgkmd:
Hi wplapper, 
Thanks for the comment.

I updated the code, where I split `RechunkData()` to smaller sub-functions. Now I'm working on improving the cache considering your comments. It may take a while.
Any further feedbacks are appreciated!

wplapper:
Hi [dgkmd](https://github.com/dgkmd),
I copied your code and added more debugging output to your code. I wanted to understand what is happening with all these many parallel pipelines. What I have seen before and and saw again that all these pipelines get stuck once in a while and nothing happens during this time. I would like to attach a compressed restic-debug.log.xz, but github won't let me do that. Could you please send me your email address and I will forward these files to you. 

But I will give you one of many more examples in the log:
```
2025/10/18 18:35:41.855006 index/index.go:118	index.init.func1	91	index 0xc01cdd2388 only has 3989 blobs and is too young (29.225628461s)
2025/10/18 18:35:41.855017 index/master_index.go:219	index.(*MasterIndex).finalizeFullIndexes	91	index 0xc01cdd2388 not full
2025/10/18 18:35:41.855026 index/master_index.go:223	index.(*MasterIndex).finalizeFullIndexes	91	return 0 indexes
2025/10/18 18:35:41.958217 debug/round_tripper.go:110	debug.loggingRoundTripper.RoundTrip	199	------------  HTTP RESPONSE ----------
HTTP/2.0 206 Partial Content
Content-Length: 17359408
Accept-Ranges: bytes
Content-Range: bytes 0-17359407/17360838
Content-Type: application/octet-stream
Date: Sat, 18 Oct 2025 17:35:41 GMT
Last-Modified: Mon, 25 Aug 2025 02:35:02 GMT
Server: rclone/v1.71.1

2025/10/18 18:35:43.632228 logger/log.go:47	logger.(*Backend).Load	198	  load err <nil>
2025/10/18 18:35:43.632280 repository/repository.go:1165	repository.(*packBlobIterator).Next	198	  process blob <data/73625e26>, skipped 0, <Blob (data) 73625e26, offset 0, length 341562, uncompressed length 647512>
2025/10/18 18:35:43.637437 repository/repository.go:1165	repository.(*packBlobIterator).Next	198	  process blob <data/93e955c7>, skipped 0, <Blob (data) 93e955c7, offset 341562, length 86482, uncompressed length 93442>
2025/10/18 18:35:43.637735 repository/repository.go:1165	repository.(*packBlobIterator).Next	198	  process blob <data/a3ff028d>, skipped 0, <Blob (data) a3ff028d, offset 428044, length 2206576, uncompressed length 2314775>
2025/10/18 18:35:43.643707 repository/repository.go:1165	repository.(*packBlobIterator).Next	198	  process blob <data/a4992818>, skipped 0, <Blob (data) a4992818, offset 2634620, length 86425, uncompressed length 93428>
```




There is a time difference of about 1.5 seconds before the system wakes up again. I feel, this is not right. You seem to put all your eggs in one basket, and you have to wait until this HTTP request is satisfied. Everything else came to a standstill.

Since some of the pipeline elements are i/o bound, you should not start more than `dstRepo.Connections()`` parallel processes, and not ``runtime.GOMAXPROCS(0)``.

I will attach the modified source code
[rechunker_wpl.txt](https://github.com/user-attachments/files/22986683/rechunker_wpl.txt)
rename file to .go again ...

dgkmd:
Hi @wplapper ,
Could you send the debug log to kimdonggyu.dev@gmail.com?

The current code lazily requests download when it needs the blob data. So with current implementation, it is expected that the whole pipeline waits while the pack is being downloaded (especially when i/o bandwidth is the bottleneck).
To that end, I think some kind of prefetch should be implemented to eliminate the bottleneck. I'll take that into account while working on improving the blob cache.

Also, on i/o bound jobs, the number of pack downloaders are limited to 4 (check `numDownloaders` variable) since it directly calls download method (`LoadBlobsFromPack`), but I did not set such limit to uploading, because I believe the backend would handle the job (with `PackUploader`)...
Please correct me if something is wrong.

Best regards,

dgkmd:
Hi @wplapper ,
I reimplemented the blob cache and also updated other code. Major changes are as follows:
1) Blobs are loaded by the pack, but evicted by each blob.
2) Rechunker traces blob usage and notifies the cache to ignore no longer needed blob (makes it evict the blob immediately and don't download it anymore)
3) Files for rechunking are ordered by their length (chunk count) -longest first- as you had commented in the email.
4) The cache is now enabled by default, and you can set its size by giving an argument `--cache-size` in MiB unit.

From my observation, redundant blob download to some extent were inevitable, especially if packs were shaped by interleaving multiple files. Redundant downloads were minimized when I set a huge cache size (>10 GiB), but currently I set default cache size to 4096 MiB to make it safe for majority of users.

Please let me know if the code works as you expect.

Best regards,

wplapper:
Hi @dgkmd,

Thank you very much for your most recent update to your rechunker code. On the whole it looks very good, but you still have a problem with dispatching the ``rc.priorityFilesList`` quickly enough. Dispatching these readily available files will remove a lot of memory pressure from the execution of your command. In the example snapshot I used, here the summary
```
   3823 unique files in selected snapshots
      6 regular file downloads to start
   3817 fastTrack files to process
     30 data packfiles to load
   4222 used blobs with size 471.714 MiB

```

there is a build-up of over 2,500 ready files before ``runDispatcher()``can get hold of them because all the other files stuck in the ``chDispatch`` queue. In my opinion, you should have two Go routines in ``runDispatcher()``, one for the normal stuff and one which exclusively deals with new arrivals in ``rc.priorityFilesList``.

The other comment I would like to make: leave the debug.Log() messages in your code, so one can actually see what is happening and when it is happening. Only by using debug.Log() records I could detect your scheduling issue.

I will upload my modification to your code onto my restic github repository. I will leave another comment where to find the code once it has been uploaded.

The attached file abbr_``restic-debug.txt` contains a restic debug log filtered by ``rechunker/rechunk`` records. You can see when new files arrive for dispatch, when the packfiles start to download and when they become ready. Once a file is finished completely, you will see the message ``ended fileID caffeeee``. The priority queue message is ``the prio queue now contains 2565 files``.

I also looked at the number of parallel Go routines. For the download, you don't need more than ``srcRepo.Connections()``, in my case 5, beceause the file download is limited by (network) i/o. You have chosen a fixed '4' in your code.

[abbr_restic-debug.txt](https://github.com/user-attachments/files/23325290/abbr_restic-debug.txt)

wplapper:
Hi @dgkmd,

you can find my modifications to your code here. I also have added a toy to simulate part of the workings of your real code. 

https://github.com/wplapper/restic/tree/rechunker

dgkmd:
Hi @wplapper,

Thank you for your review. I looked into your code and debug log. It seems the reason for thousands of `fastTrack` files stacking is that all rechunk workers are still working on non-prioritized large files at that point. So I reimplemented dispatcher logic to serve two channels, as you have commented, one for regular files and one for prioritized files (fast track). Then I added a dedicated rechunker worker that only listens to prioritized files channel. It will start on processing prioritized files as soon as a file is sent to `chPriority`, even when all other workers are busy processing regular files.

Until now I focused on implementing the code, so I didn't take enough care for my code to be readable and easily analyzable. Provided the core logic will not change much from now on, I will add sufficient comments and logs so that anyone can easily analyze the code. Thanks for the advice.

Please let me know if it works as you expected.

Best regards,

dgkmd:
Also, on the number of downloaders among your comments: I will test with `numDownloaders := min(numWorkers, srcRepo.Connections())`, and if it works without issue, I will change to that value.

wplapper:
Hi @dgkmd,

I have tested your new code and it works as I expected it to work. You might consider lowering the ``numWorkers`` to ``srcRepo.Connections()``, since in a normal case with a network connection involved all these workers sitting around doing nothing useful. 

A few comments:
1.) As I said earlier, put some useful ``debug.Log()`` statement in your code, so that the reviewer can actually see what your code is doing. My personal ``debug.Log()`` statements focus on packfile loading and file processing. You might have other preferences.
 
2.) I think you should check that ``srcRepo.Config().ChunkerPolynomial != dstRepo.Config().ChunkerPolynomial``, otherwise use the normal copy command.

3.) Try to limit the size of your functions to about 50 lines, so that they are easy to read and easy to understand. If a function gets too long, a reviewer might loose the context of this function. Usually loops are good candidates for helper functions, if a function gets too long.

4.) Memory usage is an area which I have watched with external monitoring tools, but it might be helpful to dedicate some `debug.Log()`` statements on it, like max storage reserved/currently used. In one of my test examples the maximum usage was above 1GB, and later dropped down below 500 MB, which indicates that memory eviction actually works (as expected).

5.) Consider a summary section, like number of snapshots used, number of packfiles used and their sizes, blobs used and their sizes. This is a question which is regularly asked: see amongst others: #4736, #3431, #5464

dgkmd:
Hi @wplapper,

Thank you for your comments. I'm currently updating the code, considering your former and latest comments.

Changes made until current latest commit:
1. Changed `numWorkers` to `min(runtime.GOMAXPROCS(0), srcRepo.Connections())`
2. Made 'no longer needed blob' eviction mechanism more efficient; now it is evicted as soon as the blob has been lastly loaded, instead of evicting at completion of the corresponding file. You can observe the improvement in peak memory usage, especially if there are plenty of large files.
3. It prints peak memory usage by blob cache (measured internally) at summary report.
4. Code refactoring is ongoing. I'm moving long segments in functions into sub-functions, by rule-of-thumb you mentioned (about 50 lines). With that I'm placing `debug.Log()` where relevant. Also, I placed top level functions above, and lower level functions below, to make readers easily understand what is going on.

The code update is not yet complete, but meanwhile any advices or comments are appreciated!

Best regards,

wplapper:
Hi @dgkmd,

Thanks for the latest update. The first time I ran your new version, I got panic dump. I cannot reproduce it on will, but I have seen a similar one before. Here is the the screen capture from the terminal: 
[panic-001.txt](https://github.com/user-attachments/files/23610111/panic-001.txt)

Otherwise the code looks good, meaningful debug.Log() output!

dgkmd:
Hi @wplapper,

Thank you for letting me the issue. I looked into the log, and I suspect the error is due to blob cache callback writes to the debugNote *after* all the rechunking job is done and then the summary is being printed. 
I could not reproduce the error, but it seems to be certain that the error occurs due to concurrent r/w to the internal map of debugNote. So I changed print summary function to do its job with a (thread-safe) dump copy of debugNote.

dgkmd:
Hi @wplapper,

It's been a while since last update.
While refactoring the overall code, I felt that the existence of ChunkDict feature makes the entire code significantly more complex, although it can boost performance in specific scenarios.
I thought that it would harm long-term maintainability of the code, so I got rid of ChunkDict from the code.
Also, the worker pipeline is simplified; it now consists of only reader and writer.

It is quite an overhaul of the code, but I focused on simplifying and disentangling the code. Please have a look when you have time.

Best regards,

wplapper:
Hi dgkmd,
I will have a look next week. We are currently visiting our youngest son and his family.

dgkmd:
It's nice. Have a good time!

dgkmd:
Code update status: Tracking blob load status for prioritized process is now done for prefixes of all files, instead of being done only for small files. 
There is about 10% performance improvement, when tested with a local, small sized repository(<100 GiB).

dgkmd:
I squashed the commit history when refactoring the code, because the code structure has been largely changed. You can find old commits in feature/rechunker-old branch of dgkmd/restic repository.

But the logic remains largely the same. `Dispatcher` determines which file to process next, `BlobCache` serves a cache layer of the source repository, `Worker` does rechunk computation job, then `Rechunker` and `eventTracker` orchestrates the main task (to rechunk snapshots) and handles comminication and notification among the submodules respectively.

wplapper:
Hi,

My home internet is broken, therefore a short note: I looked at the source code and noticed that 
- not all chan's are getting closed.
- some debug.Log calls are a bit too noisy:  ``file %v was visited by another dispatcher``  - pretty obvious when you have 2 dispatchers. IMHO just remove these log statements.

Merry Christmas and Happy New Year to you.

Winfried

PS: someone ran into a telegraph pole and brought our internet down.

wplapper:
Hi,

here is a test which failed (``go test -race``):

```
WARNING: DATA RACE
Write at 0x00c000034268 by goroutine 28:
  github.com/restic/restic/internal/rechunker.(*BlobCache).startDownloaders.func1()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:153 +0xd85
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/wplapper/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:93 +0x86

Previous read at 0x00c000034268 by goroutine 27:
  github.com/restic/restic/internal/rechunker.(*BlobCache).startDownloaders.func1()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:173 +0x5b6
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /home/wplapper/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:93 +0x86

Goroutine 28 (running) created at:
  golang.org/x/sync/errgroup.(*Group).Go()
      /home/wplapper/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:78 +0x11c
  github.com/restic/restic/internal/rechunker.(*BlobCache).startDownloaders()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:101 +0x192
  github.com/restic/restic/internal/rechunker.NewBlobCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:68 +0x6ac
  github.com/restic/restic/internal/rechunker.WrapWithCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:330 +0xa4
  github.com/restic/restic/internal/rechunker.(*Rechunker).setupCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker.go:279 +0x270
  github.com/restic/restic/internal/rechunker.(*Rechunker).Rechunk()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker.go:244 +0x6af
  github.com/restic/restic/internal/rechunker.TestRechunker()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker_test.go:200 +0xb7e
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1934 +0x21c
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1997 +0x44

Goroutine 27 (running) created at:
  golang.org/x/sync/errgroup.(*Group).Go()
      /home/wplapper/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:78 +0x11c
  github.com/restic/restic/internal/rechunker.(*BlobCache).startDownloaders()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:101 +0x192
  github.com/restic/restic/internal/rechunker.NewBlobCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:68 +0x6ac
  github.com/restic/restic/internal/rechunker.WrapWithCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/blob_cache.go:330 +0xa4
  github.com/restic/restic/internal/rechunker.(*Rechunker).setupCache()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker.go:279 +0x270
  github.com/restic/restic/internal/rechunker.(*Rechunker).Rechunk()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker.go:244 +0x6af
  github.com/restic/restic/internal/rechunker.TestRechunker()
      /home/wplapper/restic/dgkmd/restic/internal/rechunker/rechunker_test.go:200 +0xb7e
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1934 +0x21c
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1997 +0x44

--- FAIL: TestRechunker 

```

It looks like that ``c.free`` gets read by a `debug.Log()` statement in unlocked state.

dgkmd:
Hi wplapper,

> PS: someone ran into a telegraph pole and brought our internet down.

Wow, that's such an unusual (and unfortunate) accident. I hope it get fixed quickly!

> some debug.Log calls are a bit too noisy:  ``file %v was visited by another dispatcher``  - pretty obvious when you have 2 dispatchers. IMHO just remove these log statements.

I think you are right, I removed those debug.Log lines.

> not all chan's are getting closed.

Is it regarding dispatcher's priority chan? If it is, the priority channel may close after the end of the regular channel goroutine; because at the end of the regular channel it will trigger `dispatcher.Close()`, which in turn terminates all goroutines from the dispatcher, including the priority channel goroutine. But I think all this process is not straightforward, so if this is what you've meant, I'll try to think of a better (simpler) implementation.

> It looks like that c.free gets read by a debug.Log() statement in unlocked state.

Thank you very much for letting me identify the bug. I fixed the code so that no concurrent R/W to that variable will occur.

Merry Christmas and Happy New Year to you, too!

Best regards,

internal/rechunker/dispatcher.go
```diff
@@ -0,0 +1,236 @@
...
+			// check if the file was visited by another dispatcher;
+			// if it was, skip the file.
+			if visited != nil && visited(file.hashval) {
+				continue
+			}
+
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-d.done:
```
wplapper: You can never come here since d.Close() is run from this go-routine

internal/rechunker/dispatcher.go
```diff
@@ -0,0 +1,236 @@
...
+	return l
+}
+
+func (d *Dispatcher) Close() {
+	if d == nil {
+		return
+	}
+
+	select {
+	case <-d.done:
```
wplapper: you will never come here, since d.Close() is run in this go-routine.

internal/rechunker/dispatcher.go
```diff
@@ -0,0 +1,236 @@
...
+	default:
+		close(d.done)
```
wplapper: you are posting to d.done here.

internal/rechunker/dispatcher.go
dgkmd: It is a trick to avoid double-closing d.done.

codekow:
Really looking forward to this feature. I'll test when I can.

dgkmd:
@codekow Thank you for encouraging me! It will be greatly appreciated if you test the code and give me a feedback.

MichaelEischer:
Thanks for working on this feature! I just had a very, very brief look at this PR and I've already noticed a few larger problems:
- Why is this a separate `rechunk-copy` command and not integrated with `copy --rechunk`? This just clutters the CLI.
- The PR is far larger than I'd like. It also lacks any integration with existing code and adds yet another pipeline to download and upload files. As part of https://github.com/restic/restic/issues/2446, I'm currently working on adding `Object{Reader,Writer}` types which would allow rechunking an individual file using just `io.Copy(objectWriter, objectReader)`. I've seen that `worker.go` contains some related code, which will have to be replaced once I'm done. My hope is also that we can unify the code of `ObjectReader`, `internal/dumper` and the `BlobCache` from this PR. Please don't make any major changes in that regard until my PR is far enough.
- I hope we can get rid of the dispatcher, although I haven't looked into the code far enough to tell whether that's actually an option.
- `Rechunker.RewriteTree(...)` at first glance appears to copy most of the code from `walker.TreeRewriter`. Unnecessarily duplicating code is a massive maintenance burden.

dgkmd:
Hi @MichaelEischer,
Thank you very much for your review!

> - Why is this a separate `rechunk-copy` command and not integrated with `copy --rechunk`? This just clutters the CLI.

The CLI interface choice was rather arbitrary. I just didn't want to mess with existing `copy` command while developing the feature. If you like, I will integrate the CLI command to `copy --rechunk`.

> - The PR is far larger than I'd like. It also ...

Thanks for letting me know the context. I will keep my code structure as-is; I will also look into the issue and PR you mentioned.

> - I hope we can get rid of the dispatcher, although I haven't looked into the code far enough to tell whether that's actually an option.

The role of `dispatcher` is to dynamically schedule file order to maximize 'cache hits'. It was originally inside the main `rechunker` module, but I split it into a separate module because `rechunker` got too huge. Please let me know later that which way would be better (whether to leave it as a separate submodule vs to integrate into the main module).

> - `Rechunker.RewriteTree(...)` at first glance appears to copy most of the code from `walker.TreeRewriter`. Unnecessarily duplicating code is a massive maintenance burden.

You are right that it is mostly copied from `walker.TreeRewriter` code; it is copied because it needed minor modification to the original code when I implemented the first draft. But now I think that it can be implemented using `walker.TreeRewriter`, so I will refactor `Rechunker.RewriteTree()` to reuse `walker.TreeRewriter`.

I will wait for your mentioned PR to be ready. Meanwhile, it would be appreciated if you have any additional advices or comments.

Best regards,

changelog/unreleased/issue-5473
```diff
@@ -0,0 +1,25 @@
+Enhancement: Add rechunk-copy feature
+
+Restic didn't rechunk data blobs when copying snapshots between repositories
+with different chunker parameters. Instead, it copied the blobs as-is,
+which impaired deduplication powered by [Content Defined Chunking](https://restic.net/blog/2015-09-12/restic-foundation1-cdc/). 
+To mitigate this issue, users had to manually restore the snapshots somewhere,
+and then backup them again to the new repository. This workaround was 
+inefficient, prone to tamper with the original metadata, and bothersome.
+
+It now supports `rechunk-copy` command, in which the data are rechunked while copying.
```
wplapper: This file  needs some rework as well, since you have moved the code into the ``restic copy --rechunk`` command.

cmd/restic/cmd_rechunk_copy.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/restic/restic/internal/ui"
+	"github.com/restic/restic/internal/ui/progress"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// Reference: cmd_copy.go (v0.18.0)
+
+func newRechunkCopyCommand(globalOptions *global.Options) *cobra.Command {
```
wplapper: According to my linter, newRechunkCopyCommand() is unused. So I assume that there will be more unused functions in here.

Actually I removed cmd_rechunk_copy.go completely with any negative effect, except the integration test fails now. You probably want to rework the integration_test in the future anyway.

wplapper:
The ``restic copy --rechunk (--force)`` option is the perfect opportunity to create a fresh copy of a repository, leaving all unused and duplicates blobs behind which are filling up the source repository. I would call this "spring cleaning". Highly recommended on top of all the other features ``--rechunk`` offers.

dgkmd:
Hi @wplapper,
I'm glad that you find it useful! Please let me know if something does not run as you expected.

Currently, I'm focusing on refactoring/compacting the code, especially reusing existing code in other packages (following MichaelEischer's comment).
You can find that `Rechunker.RewriteTrees` is reimplemented with `walker.TreeRewriter` and `rechunker_test.go` is simplified with structs from `{archiver, data, walker}` etc.

wplapper:
Hi @dgkmd,

I ran a ``restic copy --rechunk --force`` from an existing repository on local disk to an empty repository on local disk with debugging enabled. Here are the basic parameters, as taken for the output ftom the copy command:
```
>/tmp/restic-debug.log; DEBUG_LOG=/tmp/restic-debug.log  ./restic -r /media/user/internal/restic_home_rechunk_copy/ copy --from-repo /media/user/internal/restic_home_master --rechunk --force -vv
debug log file /tmp/restic-debug.log
debug enabled
repository 42fcc4ec opened (version 2, compression level better)
repository ab554fe5 opened (version 2, compression level better)
Gathering snapshots...
Scanning files to process... 

[Pre-run Summary]
Number of snapshots: 114
Number of distinct files to process: 63691
  - Total size (including duplicate blobs): 16.239 GiB
Number of packs to download: 589

[1:31] 63691/63691 distinct files processed
100.00% 16.239 GiB/16.239 GiB

Rewriting trees...
Rewriting done.

Writing snapshots
snapshot ba4982a6 saved, copied from source snapshot 5fb727a5
... repeat 113 times ...
Additional data stored to the repository: 9.050 GiB
```

The timings (extracted from RESTIC_DEBUG log):
```
2026/04/28 11:18:19 === start ===
2026/04/28 11:18:21 restic/cmd_copy.go:409	main.rechunkCopy	1	Running Plan()
2026/04/28 11:19:52 rechunker/rechunker.go:239  rechunker.(*Rechunker).Rechunk  1   Rechunk done. 
2026/04/28 11:19:52 restic/cmd_copy.go:433  main.rechunkCopy.func2  128176  Running RewriteTree() for tree ID 3f0748bf
2026/04/28 11:21:00 repository/packer_manager.go:66 repository.(*packerManager).Flush   128176  manually flushing pending pack
2026/04/28 11:21:01 ... saving snapshots ...
2026/04/28 11:21:01 === ended ===

time for startup, opening repos, planning 00:02
time to rechunk                           01:31
time to rewrite trees                     01:08
time to save snapshots                    00:01
```

I was surprised to find that the rewriting of the trees takes nearly as long as the rechunking. Would it be possible to improve the speed of the ``rewrite trees`` part? Could one transform all trees once and then use ``data.StreamTrees()`` instead of walking all the snapshots individually one after the other? Just a question and an idea. Alternatively you could think about running these `rechnker.RewriteTree()`` concurrently.

The number of the ``rechunk``part is about 1 million lines, and the ``rewrite`` part is a another 1.1 million lines of debugging output. So there is a lot going on while rewriting the trees. The debug output in total is about 333 MiB in size.

dgkmd:
Hi @wplapper,
Sorry for the late response, and thank you for the feedback!

It is indeed unusual that it takes long time rewriting trees. I could not reproduce the issue with my repos.

Current `Rechunker.RewriteTrees()` uses `walker.TreeRewriter`, and it remembers previously processed trees in `idMap` so it should not take so long time even when processing multiple snapshots. 
And as rewriting of the upper directory depends on the rewriting result of subdirectories, performance gain by concurrent running might not be much compared with added complexity to handle the dependency.

There are a few things I would like to ask you for debugging:
(1) Is the local cache enabled for the source repo? Restic supports local cache for tree blobs, and TreeRewriter depends on that cache. So if local cache is disabled, it can take long time like when running `Rechunk()` with `BlobCache` disabled.
(2) Did the issue appear after the code is changed to use `walker.TreeRewriter` (commit 342769db0807f96a0e00d792e0dd0e2747711550)? Could you test with commit 8ab9d44f6d934b8a08e1466e3aac489aafeb284f (which does not use `walker.TreeRewriter`) and let me know the result?
(3) Can you let me know the approximate number of directories (corresponding to the number of tree blobs) in all your snapshots? Mine has no more than 1/10th lines of log for `RewriteTrees()` than for `Rechunk()`, so maybe snapshot characteristic might be different between us.

Thanks again for the feedback!

wplapper:
Hi dgkmd,

I messed up your copy of the restic code. Sorry about that. I now have downloaded (git clone) a new copy of your repository, compiled it and everything is working as expected. 

Sorry for the mess I have created for you.

dgkmd:
Hi @wplapper,
No worries :)
If you find another issue or have some suggestions, please let me know.

dgkmd:
Hi @MichaelEischer,
Thank you for reviewing.

Please let me know if there is something I can assist you with regarding this PR.

MichaelEischer:
> Please let me know if there is something I can assist you with regarding this PR.

I still haven't got around to make the changes I mentioned a while back, the review for this PR will unfortunately have to wait until then.

dgkmd:
Hi @MichaelEischer ,

I'm tracking your PR #21943.
Please let me know if your feature is ready or there is something I can do for you.

MichaelEischer:
@dgkmd I'll let you know when I'm far enough to get to this PR, sorry for taking so long

WhyNotHugo:
I tested this using:

```
RESTIC_FEATURES=rechunk-copy ~/src/github.com/restic/restic/restic copy \
    --rechunk \
    --from-password-command "$RESTIC_PASSWORD_COMMAND" \
    --from-repo ../restic-old
```

I fed all the lines in the style of `snapshot e7fa2530 saved, copied from source snapshot 1edd3606` to this [restic-validate-rechunk](https://git.sr.ht/~whynothugo/dotfiles/tree/4920f82495b16db3ab18ec16c76c758c2e76dd1a/item/home/.local/bin/restic-validate-rechunk) script to validate that both snapshots are identical (basically using `diff -rq` on each pair of snapshots).

Both snapshots are identical. `restic check` (the stable release) also reports the repository as stable.

TLDR: worked for me for re-chunking a repository with a few dozen snapshots.

(_obviously I tried this with a spare copy first; don't do this with production data!_)

WhyNotHugo:
When trying the above procedure over a final repository, I got the error:

> Fatal: source repo and destination repo have same chunker polynomials; run without `--rechunk`, or set `--force` flag to proceed with rechunk anyway

This repository shares the same polynomial as the target one, but it has lots of snapshots which were copied from other repositories with different polynomials.

I don't think this Fatal error is correct: it should be checking the polynomials used for the snapshots being copied, and not the one for the repository itself.

PR #5532, opened 2025-09-28
Replace Repository.SetIndex with internal helper

MichaelEischer:
`Repository.SetIndex` is only used by the `Checker` code and `RepairIndex`. To allow converting `SetIndex` into a repository internal API, first split the checker into two parts: one that checks packs+index and another that checks the snapshot, tree and blob structure. Then `SetIndex` is replaced by `loadIndexWithCallback` which allows ignoring index load errors and inspecting the loaded indexes.

No.

Checklist
---------

- [ ] ~~I have added tests for all code changes.~~ No new tests. Keeping the tests in the `checker` package, but moving a lot of code to the `repository` package isn't ideal, but I didn't find a good line along which to split the code.


MichaelEischer (commented):
LGTM.

PR #5541, opened 2025-10-01
feat: allow override env RESTIC_HOST with flag to filter all snapshots

### LINKED ISSUE #5440: Provide a way to specify "all hosts" to restic snapshots when $RESTIC_HOST is set
Output of `restic version`
--------------------------
`restic 0.18.0 compiled with go1.24.3 on linux/amd64`

When I run `restic snapshots` with $RESTIC_HOST set, I can't get it to show the list of all snapshots across all hosts.

Basically I think this issue comes about cause restic is using the same parameter name for 2 different purposes. For some commands host is metadata to be stored. For others it's a filter. Conceptually the same issue exists with tags, however there is no environment variable for that so it's only theoretical there.

Initial thoughts are that if a command accepts multiple --host args (because host acts as a filter), then it *should not* default to $RESTIC_HOST, but instead default to all hosts. Only commands that accept a single instance of the host parameter (because it is treated as metadata to be written) should default to $RESTIC_HOST. Potentially there should be a $RESTIC_HOST_FILTER env variable that is used when host is a filter rather than metadata.

However this is probably a breaking change for some users. So a more palatable solution may be a flag (maybe --all-hosts) or special hostname (maybe '*') that ignores previously set host filters.

There are workarounds, but they are clunky:
`env -u RESTIC_HOST restic snapshots`
or
`(unset RESTIC_HOST; restic snapshots)`
These are obviously for "bourne"ish shells. I can see this being more awkward for python scripts, not sure how it would be done for powershell, etc.

In general when I run restic, I source a file which sets many restic related env variables, including $RESTIC_HOST. This makes running commands cleaner and avoids errors where I forget to specify host (or pack size, or compression, etc.).

What I'm then trying to do is list all snapshots in a repo across all hosts. I will also have the same issue for `restic forget` and possibly others I haven't found.

The generalised version of the problem is that everything set by env variables should be overridable by command line args (a common convention across command line tools), and that's not true for $RESTIC_HOST.

Usually restic makes backups simple. Overall I'm very grateful restic exists, though today it surprised me in a way that was frustrating to debug and workaround.

konidev20:
When the `RESTIC_HOST` environment variable was set, there was no way to list or operate on snapshots from all hosts, as the environment variable would always filter to that specific host. 

With these changes specifying `--host=""` will filter snapshots from all hosts.

To maintain consistency I have updated the following commands that support snapshot filtering: 
`snapshots`, `forget`, `find`, `stats`, `copy`, `tag`, `repair snapshots`, `rewrite`, `mount`, `restore`, `dump`, and `ls`.

Disclaimer:
I have used the AI IDE Cursor to help me with the repetitive changes across all the commands and the unit tests for these changes.

Closes #5440 

Checklist
---------


cmd/restic/find.go
```diff
@@ -11,31 +11,39 @@ import (
...
 // initMultiSnapshotFilter is used for commands that work on multiple snapshots
 // MUST be combined with restic.FindFilteredSnapshots or FindFilteredSnapshots
+// MUST be followed by finalizeSnapshotFilter after flag parsing
 func initMultiSnapshotFilter(flags *pflag.FlagSet, filt *restic.SnapshotFilter, addHostShorthand bool) {
 	hostShorthand := "H"
 	if !addHostShorthand {
 		hostShorthand = ""
 	}
-	flags.StringArrayVarP(&filt.Hosts, "host", hostShorthand, nil, "only consider snapshots for this `host` (can be specified multiple times) (default: $RESTIC_HOST)")
+	flags.StringArrayVarP(&filt.Hosts, "host", hostShorthand, nil, "only consider snapshots for this `host` (can be specified multiple times, use empty string to override $RESTIC_HOST) (default: $RESTIC_HOST)")
```
MichaelEischer:
```suggestion
	flags.StringArrayVarP(&filt.Hosts, "host", hostShorthand, nil, "only consider snapshots for this `host` (can be specified multiple times, use empty string to unset default value) (default: $RESTIC_HOST)")
```
"override" is somewhat besides the point as it's more about undoing the effects of `RESTIC_HOST` instead of replacing it.

cmd/restic/find.go
```diff
@@ -11,31 +11,39 @@ import (
...
+}
 
-	// set default based on env if set
-	if host := os.Getenv("RESTIC_HOST"); host != "" {
-		filt.Hosts = []string{host}
+// finalizeSnapshotFilter applies RESTIC_HOST default only if --host flag wasn't explicitly set.
+// This allows users to override RESTIC_HOST by providing --host="" or --host with explicit values.
+func finalizeSnapshotFilter(flags *pflag.FlagSet, filt *restic.SnapshotFilter) {
+	// Only apply RESTIC_HOST default if the --host flag wasn't changed by the user
+	if !flags.Changed("host") {
```
MichaelEischer: There's no need to look at `flags` to detect whether `--host` was specified. `filt.Host` is now `nil` by default and has a different value if and only if `--host` was passed. That is, `filt.Hosts == nil` is sufficient here.

cmd/restic/cmd_forget_test.go
```diff
@@ -96,7 +96,43 @@ func TestForgetOptionValues(t *testing.T) {
...
+			args: []string{"--host", "flaghost"},
+			want: []string{"flaghost"},
+		},
+		{
+			name: "empty flag clears env",
+			args: []string{"--host", ""},
+			want: nil,
+		},
+		{
+			name: "empty flag clears env",
```
MichaelEischer: That test case is duplicated

cmd/restic/find_test.go
```diff
@@ -39,6 +39,30 @@ func TestSnapshotFilter(t *testing.T) {
...
 			"def",
 		},
+		{
+			"env set, empty flag overrides",
+			[]string{"--host", ""},
+			nil, // empty host filter means all hosts
+			"envhost",
+		},
+		{
+			"env set, explicit value overrides",
```
MichaelEischer: already tested by `"both"`

MichaelEischer (changes requested):
LGTM in general, except for the nits below:

cmd/restic/find.go
konidev20: Completed. Updated the doc string.

cmd/restic/find.go
konidev20: Completed.

cmd/restic/cmd_forget_test.go
konidev20: Removed the duplicate test case.

cmd/restic/find_test.go
konidev20: Oversight by me. I will review thoroughly before committing next time.

MichaelEischer:
Please rebase the PR to get rid of the merge commit.

MichaelEischer (approved):
LGTM. Thanks!

PR #5544, opened 2025-10-02
azure: use PutBlob API for uploads instead of PutBlock API + PutBlock List API

### LINKED ISSUE #5531: Update AzureBlob upload API to use PutBlob instead of PutBlock+PutBlockList
Output of `restic version`
--------------------------

Azure Blob Storage

In Azure Blob Storage, there is 2 ways to upload "blobs" (in the case of BlockBlobs Objects. PageBlobs and AppendBlobs Objects are not considered here):
https://docs.microsoft.com/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs

- With PutBlob API => Send a full file in one transaction (limit to 5000Mo)
https://docs.microsoft.com/rest/api/storageservices/put-blob

- With PutBlock + PutBlockList API => Send one or multiple blocks (limit to 4000Mo each block) of a file then send the blocklist to validate the file (which could be up to 190To). A transaction is made for every PutBlock then one is made for PutBlockLIst
https://docs.microsoft.com/rest/api/storageservices/put-block
https://docs.microsoft.com/rest/api/storageservices/put-block-list

Restic use exclusively the second method and as Restic's Packs files are by default of 16Mo, for each pack uploaded there is a PutBlock and a PutBlockList transaction, which lead to double the transactions and the fees.

To show the effect, I've uploaded a 5To file with --pack-size=4M

From Azure Portal :

Image

In this exemple, the 2M transactions, which are the sum of PutBlock and PutBlockList transactions, have been billed ~8€ (Hot Tier). With PutBlob method, the bill would be half.

Expected behavior
-----------------

Use of PutBlob API to have only one transaction per Pack file's upload.

Actual behavior
---------------

Every Pack files are sent with a PutBlock transaction then a PutBlockList transaction, which leads to increase unnecessary the number of transactions and so the bill. 

As Restic's Pack files are far away from the 5000Mo upper limit of the PutBlob API, the code from 
https://github.com/restic/restic/blob/master/internal/backend/azure/azure.go
should be reworked to at least use PutBlob API by default.

Here you can find some info from Azure SDK for Go (from line 164 and 203):
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/blockblob/client.go

Really a good tool. I'm like a kid with a new toy :)

konidev20:
The Azure Storage backend will now support uploading pack files in a single API call using the `PutBlob` API. This will reduce the cost implications of the current approach by 50% where a combination of `PutBlock` and `Put Block List` API is used. (Azure charges per operation)

Closes #5531 

Checklist
---------
- ~[ ] I have added tests for all code changes.~ Existing test cases cover the changes.
- ~[ ] I have added documentation for relevant changes (in the manual).~ The changes are not user facing. It's an internal optimization.

### DIFF
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -42,7 +42,7 @@ type Backend struct {
 }
 
 const singleUploadMaxSize = 256 * 1024 * 1024
-const singleBlockMaxSize = 32 * 1024 * 1024
+const singleBlockMaxSize = 100 * 1024 * 1024
 const defaultListMaxItems = 5000
 
 // make sure that *Backend implements backend.Backend
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -41,8 +41,8 @@ type Backend struct {
 	accessTier blob.AccessTier
 }
 
-const singleBlobMaxSize = 5000 * 1024 * 1024  // 5000 MiB - max size for Put Blob API in service version 2019-12-12+
-const singleBlockMaxSize = 2000 * 1024 * 1024 // 2000 MiB - max size for a stage block API since on 32-bit systems, the max size for an int is 2047 MiB
+const singleUploadMaxSize = 256 * 1024 * 1024
+const singleBlockMaxSize = 32 * 1024 * 1024
 const defaultListMaxItems = 5000
 
 // make sure that *Backend implements backend.Backend
@@ -261,7 +261,7 @@ func (be *Backend) Save(ctx context.Context, h backend.Handle, rd backend.Rewind
 
 	// If the file size is less than or equal to the max size for a single blob, use the single blob upload
 	// otherwise, use the block-based upload
-	if fileSize <= singleBlobMaxSize {
+	if fileSize <= singleUploadMaxSize {
 		err = be.saveSingleBlob(ctx, objName, rd, accessTier)
 	} else {
 		err = be.saveLarge(ctx, objName, rd, accessTier)
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -42,7 +42,7 @@ type Backend struct {
 }
 
 const singleBlobMaxSize = 5000 * 1024 * 1024  // 5000 MiB - max size for Put Blob API in service version 2019-12-12+
-const singleBlockMaxSize = 4000 * 1024 * 1024 // 4000 MiB - max size for StageBlock API in service version 2019-12-12+
+const singleBlockMaxSize = 2000 * 1024 * 1024 // 2000 MiB - max size for a stage block API since on 32-bit systems, the max size for an int is 2047 MiB
 const defaultListMaxItems = 5000
 
 // make sure that *Backend implements backend.Backend
--- a/internal/backend/azure/azure.go
+++ b/internal/backend/azure/azure.go
@@ -41,8 +41,8 @@ type Backend struct {
 	accessTier blob.AccessTier
 }
 
-const saveLargeSize = 256 * 1024 * 1024
-const singleBlobMaxSize = 5000 * 1024 * 1024 // 5000 MiB - max size for Put Blob API in service version 2019-12-12+
+const singleBlobMaxSize = 5000 * 1024 * 1024  // 5000 MiB - max size for Put Blob API in service version 2019-12-12+
+const singleBlockMaxSize = 4000 * 1024 * 1024 // 4000 MiB - max size for StageBlock API in service version 2019-12-12+
 const defaultListMaxItems = 5000
 
 // make sure that *Backend implements backend.Backend
@@ -54,11 +54,6 @@ func NewFactory() location.Factory {
 
 func open(cfg Config, rt http.RoundTripper) (*Backend, error) {
 	debug.Log("open, config %#v", cfg)
-
-	// Validate configuration
-	if err := cfg.Validate(); err != nil {
-		return nil, err
-	}
 
 	var client *azContainer.Client
 	var err error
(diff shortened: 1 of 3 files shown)


MichaelEischer (changes requested):
I don't see a reason to keep any method other than `auto`. So let's get rid of old upload path and just use the `auto` variant. More variants just mean more code that can cause problems.

R23-Git:
Hi @konidev20 ,

Thanks again for your time. This new method is working.

After some analysis, please consider this:
 I really think that the original `saveSmall` and `saveLarge` functions are already there to differentiate `PutBlob `and `PutBlock`, considering the limits of the service `Versions 2016-05-31 through 2019-07-07`. But `saveSmall` do not use PutBlob in this implementation (maybe it was not implemented in the Azure SDK back then).

 So maybe you just need to make some changes in `azure.go`:
- `const saveLargeSize = 256 * 1024 * 1024` to `const saveLargeSize = 5000 * 1024 * 1024`  (line 44)
- Adapt `saveSmall` to use the correct `Upload ` Azure SDK subfunction (for PutBlob API)
- `buf := make([]byte, 100*1024*1024)` to `buf := make([]byte, 4000*1024*1024)` into `saveLarge` function (line 299)

As you can see, values are matching the service versions table:

| Service version| Maximum block size (via Put Block)| Maximum blob size (via Put Block List)| Maximum blob size via single write operation (via Put Blob)|
|--------|--------|--------|--------|
| Version 2019-12-12 and later| 4,000 mebibytes (MiB)| Approximately 190.7 TiB (4,000 MiB × 50,000 blocks) | 5,000 MiB|
| Versions 2016-05-31 through 2019-07-07| 100 MiB| Approximately 4.75 TiB (100 MiB × 50,000 blocks)| 256 MiB|
| Versions earlier than 2016-05-31| 4 MiB| Approximately 195 GiB (4 MiB × 50,000 blocks)| 64 MiB|

Also, it could be great to let the end user to choose the size (under the limits of course) with 2 options: (seems Rclone do it like that):
- A cut size option to choose when we want to use PutBlob / PutBlock, 
ie: `-o azure.cutsize=512M` would force to use PutBlob if file is under 512MiB, and PutBlock if over. 
This option can't be over 5000MiB (saveLargeSize variable)
- A chunk size option to choose how to split block via the PutBlock API, 
ie: -o `azure.chunksize=16M` would force 16MiB blocks via PutBlock API, so cumulated with the previous option, a pack file over 512MiB will be split into 16MiB blocks then a PutBlockList will be emitted to produce the whole file into the storage.
This option can't be over 4000MiB.

This will ensure that anyone can set the values considering there needs and/or limits (seems that old Service Version of AzureStorage are still running. Also limits are not the same for the Emulator:
image

konidev20:
The Windows test failing in unrelated. 

> Also, it could be great to let the end user to choose the size (under the limits of course) with 2 options: (seems Rclone do it like that):
A cut size option to choose when we want to use PutBlob / PutBlock,
ie: -o azure.cutsize=512M would force to use PutBlob if file is under 512MiB, and PutBlock if over.
This option can't be over 5000MiB (saveLargeSize variable)
A chunk size option to choose how to split block via the PutBlock API,
ie: -o azure.chunksize=16M would force 16MiB blocks via PutBlock API, so cumulated with the previous option, a pack file over 512MiB will be split into 16MiB blocks then a PutBlockList will be emitted to produce the whole file into the storage.
This option can't be over 4000MiB.

@R23-Git 
I think this would be a neat feature to implement in a new feature request maybe? To limit the scope of changes in this PR. This PR is essentially just replacing the implementation of the saveSmall function.

MichaelEischer:
What would be the actual benefit of adding these options? Users already have the option to set the pack file size target via `--pack-size`. Almost all files created by restic remain near the pack file size target (currently limited to 128MB). While I plan to increase the pack file size in the future, it will never be larger than either 1 or 2 GB. In a few exceptional cases the pack files can be larger than that limit. But that isn't something we need an extra configuration option for.

According to https://learn.microsoft.com/en-us/rest/api/storageservices/versioning-for-the-azure-storage-services all regions use a 2025 version by now.

Keeping support for the emulator sounds reasonable though, so what about only cutting files into smaller parts that are larger than 2000MB?

> I think this would be a neat feature to implement in a new feature request maybe?

I'm against adding this feature unless someone can provide a reason why it's actually necessary.

internal/backend/azure/azure.go
```diff
@@ -41,7 +41,8 @@ type Backend struct {
 	accessTier blob.AccessTier
 }
 
-const saveLargeSize = 256 * 1024 * 1024
+const singleBlobMaxSize = 5000 * 1024 * 1024  // 5000 MiB - max size for Put Blob API in service version 2019-12-12+
+const singleBlockMaxSize = 2000 * 1024 * 1024 // 2000 MiB - max size for a stage block API since on 32-bit systems, the max size for an int is 2047 MiB
```
MichaelEischer: This MUST remain at 256. Otherwise, 32-bit system will instantly run out of memory when uploading large files.

internal/backend/azure/azure.go
```diff
@@ -255,48 +257,46 @@ func (be *Backend) Save(ctx context.Context, h backend.Handle, rd backend.Rewind
...
+// saveSingleBlob uploads data using a single Put Blob operation.
+// This method is more efficient for files under 5000 MiB as it requires only one API call
+// instead of the two calls (StageBlock + CommitBlockList) required by the block-based approach.
+func (be *Backend) saveSingleBlob(ctx context.Context, objName string, rd backend.RewindReader, accessTier blob.AccessTier) error {
 	blockBlobClient := be.container.NewBlockBlobClient(objName)
 
-	// upload it as a new "block", use the base64 hash for the ID
-	id := base64.StdEncoding.EncodeToString(rd.Hash())
-
 	buf := make([]byte, rd.Length())
```
MichaelEischer, line 279: We need a way to get rid of this allocation. Keeping multi gigabyte packs in memory is a bad idea.

internal/backend/azure/azure.go
konidev20:
```
const singleUploadMaxSize = 256 * 1024 * 1024
const singleBlockMaxSize = 32 * 1024 * 1024
```

This is what I propose. 

We will limit single blob upload size to 256 MiB.

internal/backend/azure/azure.go
konidev20: For pack file larger than 256 MiB. We will create blocks of 32 MiB and upload. With the default 5 connections, only 160 MiB would be allocated. But this would be again and again. 

Do we need a buffer pool here?

Since, restic's pack file size is limited to 128 MiB right now, I don't think this path will be used at all. 

Do we just remove the saveLarge option completely?

@MichaelEischer thoughts?

MichaelEischer: Let's just start without a buffer pool.
We have to keep it. In some cases tree blobs can become very large and result in pack files that are a few GB in size. I'd also like to bump the maximum pack size to roughly 1GB sometime soon.

internal/backend/azure/azure.go
```diff
@@ -41,7 +41,8 @@ type Backend struct {
 	accessTier blob.AccessTier
 }
 
-const saveLargeSize = 256 * 1024 * 1024
+const singleUploadMaxSize = 256 * 1024 * 1024
+const singleBlockMaxSize = 32 * 1024 * 1024
```
MichaelEischer: Let's keep the old value of `100*1024*1024`. I don't recall anyone complaining about memory usage problems with it.

MichaelEischer (commented):
Let's keep the old blob size of 100MB for large uploads. Other than that the change looks fine.

konidev20:
> Let's keep the old blob size of 100MB for large uploads. Other than that the change looks fine.

I will need to update the docs and add a changelog. Will do that and mark this PR ready for review. 

Thanks for the review!

MichaelEischer (approved):
LGTM. Thanks!

PR #5546, opened 2025-10-03
CI: group dependency updates for golang.org/x/*

MichaelEischer:
Dependabot currently opens individual PRs for each `golang.org/x/*` dependency. However, most of those updates are interrelated, thus this just creates noise. Group them to reduce the number of PRs.

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5550, opened 2025-10-03
check: refactor pack selection for read data

MichaelEischer:
Drop the `packs` map from the internal state of the checker. Instead the Packs(...) method now calls a filter callback that can select the packs intended for checking.

Triggered by https://github.com/restic/restic/pull/5469 while looking for way to adapt it to the changes from https://github.com/restic/restic/pull/5532 .

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5551, opened 2025-10-04
Reduce terminal progress fps to 10

### LINKED ISSUE #5383: Restic updates screen way too often causing huge GPU and energy usage for no reason
Output of `restic version`
--------------------------

SFTP

When running a backup the screen refresh is happening at an unreasonable speed. The files being backed up are multiple gigabytes so there's no need to update filenames, but the ETA and the same filenames are refreshed multiple times per second. This causes the GPU of my laptop to be used **50%**! This is completely unnecessary. The refresh rate should be much less, or an option added to disable the ETA, or something. This causes unnecessary power usage.

I am using Windows Terminal and Powershell.

I cannot use quiet mode since the backup may take a couple of hours and I want to see the progress at some level at least.

Reproduce: run a backup of large files and see the screen be updated way too often and Windows Terminal using a lot of (integrated) GPU.

Expected behavior
-----------------
Screen is not updated unless filenames change. ETA might be updated every couple of seconds if the same files are being backed up. Even when files change there might not be a need to update screen at every file if it happens multiple times in a second. User cannot see the filenames anyway at that speed.

Actual behavior
---------------
Screen is updated way too many times per second with unnecessary ETA calculations causing unnecessary GPU usage and the same filenames that are already on screen are rewritten over and over again.

This was probably not considered when building the UI routines.

Yes, I am very happy about Restic otherwise. It works wonders on my systems and keeps files safe. I highly appreciate your work!

MichaelEischer:
Progress bars were updated with 60fps which can cause high CPU or GPU usage
for some terminal emulators. Reduce it to 10fps to conserve energy.
In addition, this lower frequency seem to be necessary to allow selecting
anything in the terminal with certain terminal emulators.

Sort of fixes https://github.com/restic/restic/issues/5383

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5383
+++ b/changelog/unreleased/issue-5383
@@ -1,0 +1,9 @@
+Enhancement: Reduce progress bar refresh rates to reduce energy usage
+
+Progress bars were updated with 60fps which can cause high CPU or GPU usage
+for some terminal emulators. Reduce it to 10fps to conserve energy.
+In addition, this lower frequency seem to be necessary to allow selecting
+anything in the terminal with certain terminal emulators.
+
+https://github.com/restic/restic/issues/5383
+https://github.com/restic/restic/pull/5551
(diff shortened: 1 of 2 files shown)


greatroar:
+1. I've had RESTIC_PROGRESS_FPS set to 15 for years.

MichaelEischer (commented):
LGTM.

PR #5552, opened 2025-10-04
add a dockerignore

ferringb:
This change adds ignore patterns for docker, limiting what it transfers into the build context.  This is strictly to make the resultant container more hygenic in regards to what the build process can truly rely on.  Due to `.git` having to be transferred in, layer caching is impossible even if the VCS changes between the last image and current are just non code changes.

Originally this was intended to make rebuilds of the docker image faster, but due to `.git` being included, this is now just for a more hygenic docker build.

Big ole nope on that one.  This is a QoL offshoot of #5448  .

Checklist
---------

The checklist for docs and tests is fully not applicable for this change.  This is a release process visible change, only.
- [X] not applicable: I have added tests for all code changes.
- [X] not applicable: I have added documentation for relevant changes (in the manual).
- [X] not applicable: There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+# Actual layer caching is impossible due to .git, but
+# that must be included for provenance reasons.  These ignores
+# are strictly for hygenic build.
+*
+!/*.go
+!/go.*
+!/cmd/*
+!/internal/*
+!/helpers/*
+!/VERSION
+!/.git/


.dockerignore
```diff
@@ -0,0 +1,7 @@
+*
```
MichaelEischer, line 4: I'd like to keep the `.git` folder. Go automatically extracts VCS information from it and adds those to the resulting binary. See https://github.com/restic/restic/blob/481fcb9ca78d009bea9b19aed6ac22f323838c57/.github/workflows/docker.yml#L59

ferringb: Yeah, I noticed that youre release process rips the `restic` binary out of the container, so this absolutely has to stay for `debug.BuildInfo`.  Fixed.

MichaelEischer (approved):
LGTM.

Issue #5480, opened 2025-08-27
"Invalid file size" errors when restoring Windows backup, on MacOS, specifically to an exFAT volume


akrabu:
Output of `restic version`
--------------------------
restic 0.18.0 compiled with go1.24.1 on darwin/arm64

What backend/service did you use to store the repository?
---------------------------------------------------------
Local

Problem description / Steps to reproduce
----------------------------------------

When I try to restore a snapshot of a Windows user profile, on a Mac, with the target formatted as exFAT, (seemingly) random "Invalid file size" errors crop up.  If I restore on top of it, it will fill in some files, error out on others.  If you do it enough times, you actually get everything restored without verification errors.

I am restoring like this:
`restic restore eeffba10:/E/Users/username/Desktop/Test -t /Volumes/1TB_Secure/Test --verify`

**To an exFAT volume**.  Note it completes successfully on HFS+/APFS.

Expected behavior
-----------------

To restore all files that exFAT supports natively on a Mac (I realize many links and other special files may fail, but this is random user files such as images and documents)

Actual behavior
---------------

Many errors like this:
`ignoring error for /Volumes/1TB_Secure/test/Alexa488_20xCA1Snap.tif: Invalid file size for /Volumes/1TB_Secure/test/Alexa488_20xCA1Snap.tif: expected 4237380, got 3940355`

See more here: https://forum.restic.net/t/restore-windows-snapshot-on-macos-results-in-invalid-file-size/10226/3?u=akrabu

Do you have any idea what may have caused this?
-----------------------------------------------

Seems to be a bug

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Restic is a vital part of my daily job, <3 you guys

MichaelEischer:
I've created a possible fix based on the analysis from https://github.com/restic/restic/issues/5543 in https://github.com/restic/restic/pull/5553 . Can you test whether it solves your issue?

akrabu:
Didn't seem to fix the problem.  I DM'd you on the forum with details!

MichaelEischer:
Can you test whether setting the environment variable `GODEBUG=asyncpreemptoff=1` and telling restic to only use a single connection solves the issue?

akrabu:
Ahh unfortunately that didn't do it ☹️ 

It's so weird how if you run the restore multiple times, with --verify and --overwrite=if-changed that eventually you'll get everything, with fewer and fewer errors each time.

akrabu:
[...]

PR #5553, opened 2025-10-04
restore: serialize file writes if preallocation is not possible

MichaelEischer:

My current assumption regarding the bug in https://github.com/restic/restic/issues/5543 is that this is related to concurrently calling `WriteAt(...)` with different offsets after the current file end in the restorer. If the underlying filesystem does not support sparse files, then both calls have to extend the file up to the given offset and fill it with zeros. My assumptions is that this creates a race condition in which both calls can overwrite each other.

Avoid this case by simply serializing all writes to a given file if it failed to be preallocated.

The current implementation is a rather ugly proof of concept, but should be good enough to test whether it solves the issue.

Hopefully fixes https://github.com/restic/restic/issues/5543 and https://github.com/restic/restic/issues/5480 .

Checklist
---------

### DIFF
--- a/cmd/fsstress/main.go
+++ b/cmd/fsstress/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"math/rand"
+	"os"
+)
+
+func main() {
+	r := rand.New(rand.NewSource(123456))
+
+	var filenames []string
+	for i := 0; i < 10; i++ {
+		filenames = append(filenames, fmt.Sprintf("file-%d.txt", r.Int()))
+	}
+	for _, filename := range filenames {
+		// create file but make sure to not overwrite existing files
+		if _, err := os.OpenFile(filename, os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0600); err != nil {
+			panic(err)
+		}
+	}
+
+	// prepare chunks
+	var chunks [][]byte
+	for i := 0; i < 1000; i++ {
+		buf := make([]byte, r.Intn(10000)+10_000)
+		if _, err := r.Read(buf); err != nil {
+			panic(err)
+		}
+		chunks = append(chunks, buf)
+	}
+
+	type task struct {
+		filename string
+		offset   int64
+		blob     []byte
+	}
+
+	// schedule to files
+	var tasks []task
+	for _, filename := range filenames {
+		offset := int64(0)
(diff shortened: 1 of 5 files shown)


akrabu:
@MichaelEischer Just tested - doesn't seem to have changed anything.  I'll PM you the error message on the forum.

MichaelEischer:
As the original assumption apparently wasn't correct, let's try something else. The PR currently contains two different things:
- the `restorer` now truncates (read extend) files to their final size before writing actual data. This might help solve the problem at the cost of writing  the whole file twice.
- There is a new command `fsstress` which throws a lot of write commands at the filesystem that maybe are able to reproduce the problem. It can be compiled using `go build ./cmd/fsstress` and when run with `./fsstress` it creates a few files in the local folder, fills them with random data with pretty much random write requests and verifies the data afterwards. I'd be interested in what happens when running that command on the problematic filesystems.

MichaelEischer:
And the debug log will now contain which blobs were written to which file. To get the entries set the environment variables  `DEBUG_LOG=debug.log` and filter for lines containing `writeToFile`. As this creates a log of debug output it's probably only suitable for a small test repo. If that test repo has a verification error, then please check whether a blob for the problematic file and offset was written.

Edit:
Use `DEBUG_FUNCS='restorer.(*fileRestorer).downloadBlobs.func1.1' ` instead of `DEBUG_LOG` for more targeted logging. Those logs will be printed to stdout.

MichaelEischer:
It would also be interesting to know whether using https://github.com/restic/restic/pull/5627 changes anything about the restore problems. That PR should reduce how of `restore` opens and closes files, which could have an influence here.

akrabu:
> As the original assumption apparently wasn't correct, let's try something else. The PR currently contains two different things:
> 
> * the `restorer` now truncates (read extend) files to their final size before writing actual data. This might help solve the problem at the cost of writing  the whole file twice.
> * There is a new command `fsstress` which throws a lot of write commands at the filesystem that maybe are able to reproduce the problem. It can be compiled using `go build ./cmd/fsstress` and when run with `./fsstress` it creates a few files in the local folder, fills them with random data with pretty much random write requests and verifies the data afterwards. I'd be interested in what happens when running that command on the problematic filesystems.

I ran `fsstress` several times, it worked fine.

#5553 did not work

#5627 DID work!  🥳

MichaelEischer:
Thanks a lot for testing. Then let's hope this is permanently solved by #5627.

PR #5558, opened 2025-10-10
repository: enforce correct usage of SaveBlob

MichaelEischer:
Uploading new blobs requires quite bit of boilerplate code. To reduce it replace StartPackUploader+Flush with WithBlobUploader . The latter handles the setup and flushing automatically.

Then move the SaveBlob method into the helper returned by WithBlobUploader. This ensures that SaveBlob cannot be called while the repository is not ready for it (well, strictly speaking the callback passed to WithBlobUploader could leak a reference to SaveBlob. However, there's no way to statically enforce this in Go).

No. Noticed this while making preparations for https://github.com/restic/restic/issues/2446 .

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_debug.go
+++ b/cmd/restic/cmd_debug.go
@@ -352,7 +352,7 @@ func loadBlobs(ctx context.Context, opts DebugExamineOptions, repo restic.Reposi
 		return err
 	}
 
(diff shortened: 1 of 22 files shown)


MichaelEischer (commented):
LGTM.

PR #5560, opened 2025-10-12
index: port to  modern Go iterators

MichaelEischer:
Convert MasterIndex, Index and AssociatedSet to use modern Go iterators. I've kept the Repository interface unchanged for now. The change was originally intended to allow removing the `packs` argument from `repository.Repack`. However, to not negatively affect `Prune`, this requires more work than I currently want to spend. Thus, this PR only contains the independently useful index refactorings.

Refactoring made using Cursor AI.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5567, opened 2025-10-22
Improve example for forget --keep-daily

### LINKED ISSUE #5156: Improve example for `forget --keep-daily`
The `forget` options `--keep-daily` and `--keep-within-daily` have been a constant source of questions. Currently the [documentation](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) contains an example with snapshots created at regular intervals. However, this does not explain the differences between `--keep-daily` and `--keep-within-daily`.

To make things easier to understand, the example could instead show a case where snapshots were created on a daily basis but not on the weekend. In addition, some of the days should contain multiple snapshots. This would both showcase the "daily" aspect as well as the difference between both options.

Paulomen2712:

This PR improves the documentation for the forget options `--keep-daily` and `--keep-within-daily` by providing a clearer example. The current example uses snapshots taken at regular intervals, which doesn’t clearly illustrate the difference between these two options. The new example will demonstrate snapshots taken daily except on weekends, with some days containing multiple snapshots and with one day missing a backup. This setup better highlights the "daily" concept and clarifies how `--keep-daily` and `--keep-within-daily` behave differently. Also it reorders the section to first introduce the `--keep-within-*` options before the example that now includes them.

Closes #5156

Checklist
---------

### DIFF
--- a/doc/060_forget.rst
+++ b/doc/060_forget.rst
@@ -260,67 +260,6 @@ the tag.
 
    $ restic forget --tag '' --keep-last 1
 
-Let's look at a simple example: Suppose you have only made one backup every
-Sunday for 12 weeks:
-
-.. code-block:: console
-
-   $ restic snapshots
-   repository f00c6e2a opened successfully
-   ID        Time                 Host        Tags        Paths
-   ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. Thanks!

PR #5570, opened 2025-10-25
backends: configurable maxElapsedTime for retry operations

devonhk:

Add --retry-max-elapsed-time flag to configure the maximum elapsed time for retry operations. The default remains 15 minutes.

Previously, the timeout was hardcoded at 15 minutes, forcing users to wait the full duration when backends were down or misconfigured.

Many of us automate our backups using scripts or other tools. A common operation is to check if a repo is initialized
before performing backups. This is usually done with `restic cat config`. However, if the backend is down or misconfigured, it's ideal for the operation to fail fast and return an error. Unfortunately, the retry mechanism is hardcoded to run for 15mins.
 

use "Closes #5571 "

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5571
+++ b/changelog/unreleased/issue-5571
@@ -1,0 +1,9 @@
+Enhancement: Configurable retry timeout
+
+Restic's retry operations used a hardcoded timeout of 15 minutes,
+forcing users to wait the full duration when backends were down or misconfigured.
+
+Restic now allows setting the retry timeout via the --retry-max-elapsed-time flag.
+This permits commands like `restic cat config` to fail quickly when backends are unreachable.
+
+https://github.com/restic/restic/issues/5571
--- a/internal/global/global.go
+++ b/internal/global/global.go
@@ -44,23 +44,24 @@ type BackendWrapper func(r backend.Backend) (backend.Backend, error)
 
 // Options hold all global options for restic.
 type Options struct {
-	Repo               string
-	RepositoryFile     string
-	PasswordFile       string
-	PasswordCommand    string
-	KeyHint            string
-	Quiet              bool
-	Verbose            int
-	NoLock             bool
-	RetryLock          time.Duration
-	JSON               bool
-	CacheDir           string
-	NoCache            bool
-	CleanupCache       bool
-	Compression        repository.CompressionMode
-	PackSize           uint
-	NoExtraVerify      bool
-	InsecureNoPassword bool
+	Repo                string
+	RepositoryFile      string
+	PasswordFile        string
+	PasswordCommand     string
+	KeyHint             string
+	Quiet               bool
+	Verbose             int
+	NoLock              bool
+	RetryLock           time.Duration
+	JSON                bool
+	CacheDir            string
+	NoCache             bool
+	CleanupCache        bool
+	Compression         repository.CompressionMode
+	PackSize            uint
+	NoExtraVerify       bool
+	InsecureNoPassword  bool
+	RetryMaxElapsedTime time.Duration
 
 	backend.TransportOptions
 	limiter.Limits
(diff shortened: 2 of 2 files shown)


devonhk:
Here's an example of a quick timeout on a rest backend (5 seconds).

```shell
vagrant@testing:/vagrant$ RESTIC_REST_USERNAME=user RESTIC_REST_PASSWORD=password /vagrant/restic --retry-max-elapsed-time=5s  -r rest:http://localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80 cat config 
Stat(<config/>) returned error, retrying after 849.484141ms: Head "http://user:***@localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80/config": dial tcp [::1]:8000: connect: connection refused
Stat(<config/>) returned error, retrying after 1.966898395s: Head "http://user:***@localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80/config": dial tcp [::1]:8000: connect: connection refused
Stat(<config/>) failed: Head "http://user:***@localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80/config": dial tcp [::1]:8000: connect: connection refused
Fatal: unable to open config file: Head "http://user:***@localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80/config": dial tcp [::1]:8000: connect: connection refused
rest:http://localhost:8000/6d521db2-3e72-4149-b2d0-65fee9810f80/

```

michaelbeaumont:
This PR doesn't solve the issue I created. In order to solve that issue the timeout must be able to be turned off and the previous behavior restored.

devonhk:
Simply reducing the amount of time spent retrying solves the problem for me :man_shrugging:.
Anyway, I updated the description to point to a new issue.

devonhk:
@MichaelEischer Could I get a review please? 

The PR introduces a small change that's backwards compatible.

internal/global/global.go
```diff
@@ -110,6 +111,7 @@ func (opts *Options) AddFlags(f *pflag.FlagSet) {
 	f.StringSliceVarP(&opts.Options, "option", "o", []string{}, "set extended option (`key=value`, can be specified multiple times)")
 	f.StringVar(&opts.HTTPUserAgent, "http-user-agent", "", "set a http user agent for outgoing http requests")
 	f.DurationVar(&opts.StuckRequestTimeout, "stuck-request-timeout", 5*time.Minute, "`duration` after which to retry stuck requests")
+	f.DurationVar(&opts.RetryMaxElapsedTime, "retry-max-elapsed-time", 15*time.Minute, "max elapsed time for retry operations")
```
MichaelEischer, line 114: Before https://github.com/restic/restic/pull/4784 the timeout was much shorter. Judging from repeated complaints about the longer retries, I think we want to move closer to the pre-0.17.0 behavior.  The old retry timeout was roughly one minutes, which is maybe a bit short, so let's default to 2 minutes. (It's no problem that the stuck-request-timeout is larger as restic will always perform at least one retry independent of the configured timeout).

The main reason to increase the timeout up to 15 minutes was to not have to add a tuning know, but that led to other problems.

changelog/unreleased/issue-5571
```diff
@@ -0,0 +1,9 @@
+Enhancement: Configurable retry timeout
```
MichaelEischer, line 1: Mark this as `Change` as we're also going to decrease the default timeout.

internal/global/global.go
MichaelEischer: I'd also go with a shorter name for the CLI parameter like `--retry-timeout`.

internal/global/global.go
```diff
@@ -597,7 +599,8 @@ func wrapBackend(be backend.Backend, gopts Options, printer progress.Printer) (b
 	success := func(msg string, retries int) {
 		printer.E("%v operation successful after %d retries", msg, retries)
 	}
-	be = retry.New(be, 15*time.Minute, report, success)
+
+	be = retry.New(be, gopts.RetryMaxElapsedTime, report, success)
```
MichaelEischer, line 603: This also needs some basic sanity check that the timeout is not negative or too high (maybe an hour as upper limit?).

changelog/unreleased/issue-5571
```diff
@@ -0,0 +1,9 @@
...
+
+Restic's retry operations used a hardcoded timeout of 15 minutes,
+forcing users to wait the full duration when backends were down or misconfigured.
+
+Restic now allows setting the retry timeout via the --retry-max-elapsed-time flag.
+This permits commands like `restic cat config` to fail quickly when backends are unreachable.
```
MichaelEischer, line 7: Please add a reference to the `--retry-timeout` option to the "Checking if a repository is already initialized" section in the docs and also possibly elsewhere.

The changelog should also mention that `--retry-timeout 0s` can be use to minimize retries.

MichaelEischer (changes requested):
Sorry for the late review. I'd like to use this as a chance to reduce the default timeout again. The 15min timeout was chosen to also cover longer connection disruptions that occur for some users when, for example, their internet connection reconnects at night. But with the new option they can configure it themselves, so there's no need for everyone else to wait for the long timeout.

@michaelbeaumont Why does this not solve https://github.com/restic/restic/issues/5517 ? The parameter that's changed here only determines the duration during which additional retries might be started, it does _not_ bound the duration of an individual request. There's also always at least one retry. This should be enough to accommodate a slow repository without delaying things too much.

PR #5588, opened 2025-11-06
snapshots: Show timezone in non-compact output

seqizz:

Restic lists snapshots in user's current timezone. It's very convenient but might be confusing in case of multiple users accessing the same repository without realizing they're on different timezones.

This change is an attempt to solve it by showing one-line information to the bottom of the snapshot list (plus, it is a best effort trying to get the timezone info, since I am not sure how reliable would this method be on non-Linux operating systems).

https://forum.restic.net/t/timezone-question/10228

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/global"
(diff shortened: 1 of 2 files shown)


cmd/restic/cmd_snapshots.go
```diff
@@ -265,7 +266,17 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
 		tab.AddRow(data)
 	}
 
-	tab.AddFooter(fmt.Sprintf("%d snapshots", len(list)))
+	// Add timezone information to prevent confusion (non-compact mode)
+	// Each snapshot can be registered in different timezones,
+	// but we display them all in local timezone on this output.
+	footer := fmt.Sprintf("%d snapshots", len(list))
+	if !compact {
```
MichaelEischer: Is there any particular reason for not showing the timezone in compact mode? Right now `compact` only influences the width of the output.

cmd/restic/cmd_snapshots.go
```diff
@@ -265,7 +266,17 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
...
+		zoneName, _ := time.Now().Local().Zone()
+		if zoneName != "" {
+			footer = fmt.Sprintf("(Timestamps shown in %s timezone)\n%s", zoneName, footer)
```
MichaelEischer: I think I'd somewhat prefer to omit the brackets:
```
---------------------------------------------------------------------------------------------------------------------------
Timestamps shown in CET timezone
65 snapshots
```

MichaelEischer (changes requested):
Nice, I like the overall approach, I just have to nits commented below. Please add a short changelog entry (see `changelog/TEMPLATE`)

seqizz:
Thanks! Did the changes and added a changelog entry.

MichaelEischer (approved):
LGTM. Thanks!

PR #5590, opened 2025-11-07
Add configurable HTTP range request size

angcl:
# Add configurable HTTP range request size

## Summary

This PR makes the HTTP Range request size configurable to allow users to reduce the number of HTTP requests when using REST and other HTTP-based backends.

Previously, restic used a hardcoded maximum chunk size of 32 MB (2 × DefaultPackSize) when loading blobs from pack files via HTTP Range requests. This meant that any blob retrieval operation would be split into multiple 32 MB range requests. For backends with high per-request overhead (latency, API rate limits, connection costs), this could be inefficient.

This change introduces a new `--http-range-size` flag and `RESTIC_HTTP_RANGE_SIZE` environment variable to configure this limit, allowing users to increase it and reduce the number of HTTP requests at the cost of potentially transferring more unused data.

## Changes

- **`internal/global/global.go`**: Added `HTTPRangeSize` field to `Options` struct with command-line flag and environment variable support
- **`internal/repository/repository.go`**:
  - Added `HTTPRangeSize` field to repository `Options` struct
  - Modified `LoadBlobsFromPack()` to use configurable range size with a default of 32 MB when not specified
  - Updated `streamPack()` function signature to accept `maxChunkSize` as a parameter instead of using a hardcoded constant
- **`internal/repository/repository_internal_test.go`**: Updated all test calls to `streamPack()` to include the default chunk size parameter

## Usage

```bash
# Use 64 MB range requests instead of the default 32 MB
restic --http-range-size 64 restore latest

# Or set via environment variable
export RESTIC_HTTP_RANGE_SIZE=64
restic backup /data
```

## Backward Compatibility

The default behavior (32 MB) is maintained when the option is not specified, ensuring full backward compatibility.

## Testing

- All existing repository tests pass
- Build succeeds without errors
- New flag appears correctly in `--help` output
- Default behavior preserved when flag is not specified

### DIFF
--- a/internal/repository/repository.go
+++ b/internal/repository/repository.go
@@ -62,6 +62,7 @@ type Options struct {
 	Compression   CompressionMode
 	PackSize      uint
 	NoExtraVerify bool
+	HTTPRangeSize uint
 }
 
 // CompressionMode configures if data should be compressed.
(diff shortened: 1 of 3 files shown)


internal/repository/repository.go
```diff
@@ -955,10 +956,15 @@ const maxUnusedRange = 1 * 1024 * 1024
 // then LoadBlobsFromPack will abort and not retry it. The buf passed to the callback is only valid within
 // this specific call. The callback must not keep a reference to buf.
 func (r *Repository) LoadBlobsFromPack(ctx context.Context, packID restic.ID, blobs []restic.Blob, handleBlobFn func(blob restic.BlobHandle, buf []byte, err error) error) error {
-	return streamPack(ctx, r.be.Load, r.LoadBlob, r.getZstdDecoder(), r.key, packID, blobs, handleBlobFn)
+	httpRangeSize := r.opts.HTTPRangeSize
+	if httpRangeSize == 0 {
+		// Default to 2x the pack size (matching the old constant behavior)
+		httpRangeSize = 2 * DefaultPackSize
```
MichaelEischer, line 962: That defaulting should already happen in the `New(...)` method.

MichaelEischer (changes requested):
@angcl Please always use the PR template. That also applies when using some AI coding assistant!

I also don't quite understand the use case. Please provide some benchmark numbers that this _actually_ provides a performance benefit and not just adds one more knob that nobody knows how to set properly.

The option name is also misleading as it applies for any backend type not just the HTTP-based ones. But we can discuss this once the use case has been demonstrated.

MichaelEischer:
I'll close this PR for now as I currently don't understand the use case.

PR #5592, opened 2025-11-08
Fail fast for invalid RESTIC_PACK_SIZE env values

gunar:
## What does this PR change? What problem does it solve?

- previously a typo like RESTIC_PACK_SIZE=64MiB let a backup run to completion while silently falling back to the default pack size, so you only learn much later that the wrong pack size was used; now we return a fatal error as soon as the env var can’t be parsed, matching the CLI flag’s validation
- document the rationale inline so future readers know why we guard the env path
- add tests that cover both the failure case (64MiB) and a valid numeric env value

## Was the change previously discussed in an issue or on the forum?

- small bugfix, no prior discussion

## Checklist

- [x] I have added tests for all code changes
- [x] I have added documentation for relevant changes (not needed)
- [x] There's a new file in changelog/unreleased/ (not needed for this small bugfix)
- [x] I'm done! This pull request is ready for review

## Tests

- GOCACHE=$(pwd)/.gocache go test ./internal/global

### DIFF
--- a/changelog/unreleased/pull-5592
+++ b/changelog/unreleased/pull-5592
@@ -1,0 +1,7 @@
+Bugfix: Return error if `RESTIC_PACK_SIZE` contains invalid value
+
+If the environment variable `RESTIC_PACK_SIZE` could not be parsed, then
+restic ignored its value. Now, the restic commands fail with an error, unless
+the command-line option `--pack-size` was specified.
+
+https://github.com/restic/restic/pull/5592
--- a/cmd/restic/main.go
+++ b/cmd/restic/main.go
@@ -49,6 +49,10 @@ The full documentation can be found at https://restic.readthedocs.io/ .
 		DisableAutoGenTag: true,
 
 		PersistentPreRunE: func(c *cobra.Command, _ []string) error {
+			switch c.Name() {
+			case "__complete", "__completeNoDesc":
+				return nil
+			}
 			return globalOptions.PreRun(needsPassword(c.Name()))
 		},
 	}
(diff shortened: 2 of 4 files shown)


gunar:
For context, what I _really_ wanted to do was to refactor all environment variable parsers to lean on Cobra/`pflag` so that the same parsing/validation is used for both CLI flags and environment variables. Alas, that's too big a refactor for my first PR here :-)

gunar:
Seeing the discussion in #4759 just now.

I’ve updated this PR so RESTIC_PACK_SIZE is parsed during PreRun (and skipped for the completion helpers), then ignored whenever --pack-size was provided. That way we still fail fast for real commands, but autocompletion and CLI overrides continue to work.

We could also add tests for the completion helper (by calling Cobra __complete). It seems to me that is something we're missing. Implementing it looks doable but it's a bigger change.

MichaelEischer (changes requested):
Thanks! Please add a short changelog entry of type "Bugfix", then this PR can be merged.

> Seeing the discussion in https://github.com/restic/restic/issues/4759 just now.

The refactoring I've mentioned there is completed by now. What is remaining is a proper pattern for more easily handling and validating environment variables. 

> For context, what I really wanted to do was to refactor all environment variable parsers to lean on Cobra/pflag so that the same parsing/validation is used for both CLI flags and environment variables.

That sounds like you already have something specific in mind?

MichaelEischer (approved):
LGTM. Thanks!. I've added the changelog and slightly simplified the pattern such that the environment variable is just read later on. `PreRun` already reads other environment variable, so this doesn't matter much.

PR #5596, opened 2025-11-16
backend/local: fix "operation not supported" when unlocking

mikix:
If the repo is on a mounted folder that doesn't support chmod (like SMB), it was causing an "operation not supported" error when trying to chmod 666 a file before deleting it.

But it isn't generally needed before deleting a file (the folder permissions matter there, not the file permissions). So, just drop it.

This code path I'm deleting was introduced [ten years ago](https://github.com/restic/restic/commit/264472219804ac7c62ab1f90d3cf4a156b89495b) without comment. Is there a platform that Restic supports that *does* care about the file permissions before deleting? If so, I can fix this another way - by ignoring ErrUnsupported for Unix (like the [other recent fix](https://github.com/restic/restic/pull/5345/commits/a8535aba58c3fe5056d35e5fd14d0400e757ca38) I made did)

The only remaining `chmod` call I can see that might be a problem is in ResetPermissions in file.go (which could call the file_xxx.go's `chmod` abstraction, but I think there we actually do want to raise an issue if we can't correctly reset permissions - it's only used when there's an access denied error already. So the fact that chmod is unsupported is relevant when explicitly trying to reset permissions.)

Closes #5595

Checklist
---------

### DIFF
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -255,13 +255,7 @@ func (b *Local) Stat(_ context.Context, h backend.Handle) (backend.FileInfo, err
 func (b *Local) Remove(_ context.Context, h backend.Handle) error {
 	fn := b.Filename(h)
 
-	// reset read-only flag
-	err := os.Chmod(fn, 0666)
-	if err != nil && !os.IsPermission(err) {
-		return errors.WithStack(err)
-	}
-
-	return os.Remove(fn)
+	return removeFile(fn)
 }
 
 // List runs fn for each file in the backend which has the type t. When an
--- a/changelog/unreleased/issue-5595
+++ b/changelog/unreleased/issue-5595
@@ -1,0 +1,8 @@
+Bugfix: Fix "chmod not supported" errors when unlocking
+
+Restic 0.18.0 introduced a bug that caused "chmod xxx: operation not supported"
+errors to appear when unlocking with a stale lock, on a local file repository
+that did not support chmod (like CIFS or WebDAV mounted via FUSE). Restic now
+just doesn't bother calling chmod in that case on Unix, as it is unnecessary.
+
+https://github.com/restic/restic/issues/5595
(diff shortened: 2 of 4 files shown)


internal/backend/local/local.go
```diff
@@ -255,12 +255,6 @@ func (b *Local) Stat(_ context.Context, h backend.Handle) (backend.FileInfo, err
 func (b *Local) Remove(_ context.Context, h backend.Handle) error {
 	fn := b.Filename(h)
 
-	// reset read-only flag
-	err := os.Chmod(fn, 0666)
```
MichaelEischer, line 259: This is actually necessary for windows support. From https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilea#remarks :

> To delete a read-only file, first you must remove the read-only attribute.

MichaelEischer (changes requested):
We cannot remove the chmod call (see comment). Please just filter out `ErrUnsupported` as already done by `setFileReadonly`. (Feel free to create a unified helper for setting file read-only and writeable if you want)

mikix:
Or maybe - much like we have platform specific wrappers for lots of stuff, make a `setFileDeletable` wrapper that is a no-op on Unix but calls chmod on windows.

MichaelEischer:
> Or maybe - much like we have platform specific wrappers for lots of stuff, make a `setFileDeletable` wrapper that is a no-op on Unix but calls chmod on windows.

Then I'd just turn that into a `deleteFile` wrapper that does the correct thing on each platform.

internal/backend/local/local.go
mikix: OK, I tried the alternative fix, to add platform-specific `removeFile` functions.

MichaelEischer (approved):
LGTM. Thanks!

MichaelEischer:
> The only remaining chmod call I can see that might be a problem is in ResetPermissions in file.go (which could call the file_xxx.go's chmod abstraction, but I think there we actually do want to raise an issue if we can't correctly reset permissions - it's only used when there's an access denied error already.

Yes, that error must be checked. `ResetPermissions` is already only called if `restore` couldn't open a file for writing.

Issue #5577, opened 2025-11-01
Panic on error with restic prune.


jdesanto:
Output of `restic version`
--------------------------

❯ restic version
restic 0.18.1 compiled with go1.25.1 on darwin/arm64

What backend/service did you use to store the repository?
---------------------------------------------------------

Terramaster F8+ SSD TOS v6 over SMB

Problem description / Steps to reproduce
----------------------------------------

Prune a repository.  Password variable is set. No other parameters.

Expected behavior
-----------------

Successful prune rather than panic. I'd expect some sort of error check on the existence of the id

Actual behavior
---------------

❯ restic -v -r mainbackup prune
repository 5fc605f4 opened (version 2, compression level auto)
loading indexes...
[0:02] 100.00%  28 / 28 index files loaded
loading all snapshots...
finding data that is still in use for 5 snapshots
[0:00] 20.00%  1 / 5 snapshots
id 3e6215b5ee484a0e949d0f3e788a79c350d764d02ab41039d4e73c84f2af9416 not found in repository
github.com/restic/restic/internal/repository.(*Repository).LoadBlob
	github.com/restic/restic/internal/repository/repository.go:235
github.com/restic/restic/internal/restic.LoadTree
	github.com/restic/restic/internal/restic/tree.go:115
github.com/restic/restic/internal/restic.loadTreeWorker
	github.com/restic/restic/internal/restic/tree_stream.go:36
github.com/restic/restic/internal/restic.StreamTrees.func1
	github.com/restic/restic/internal/restic/tree_stream.go:176
golang.org/x/sync/errgroup.(*Group).Go.func1
	golang.org/x/sync@v0.16.0/errgroup/errgroup.go:93
runtime.goexit
	runtime/asm_arm64.s:1268

Do you have any idea what may have caused this?
-----------------------------------------------

Corruption after loss of power to the NAS.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yes. Every day. Terrific piece of software.

wplapper:
Try `restic check` on your repository first. I think you have corrupted data in your repo. There is at least one snapshot, where the tree structure is not in sync with the data stored in the repository. Removing the offending snapshot will allow you to use the rest of the repository. See more details on ``https://restic.readthedocs.io/en/stable/077_troubleshooting.html``.

MichaelEischer:
https://github.com/restic/restic/pull/5599 will fix the output to show an error instead of a stacktrace.

[...]

PR #5599, opened 2025-11-16
prune: return proper error if blob cannot be found

MichaelEischer:
Print a proper error message if a blob could not be found instead of a stacktrace, see https://github.com/restic/restic/issues/5577

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5600, opened 2025-11-16
only suggest TMP as tmp dir variable on windows

### LINKED ISSUE #5572: specifying temp directory
Output of `restic version`
--------------------------

local directory

I need to set the location where restic stores the temporary pack files.
The documentation states in multiple places

> On Windows, use either the TMP or TEMP environment variable.

I set TEMP to the desired directory but it had no effect.

Expected behavior
-----------------
I do suggest two things
- state in the documentation that TMP takes precedence or don't mention TEMP at all.
- add Global Flags command line option --temp-dir to set the location for the temporary files, similar to --cache-dir. It's odd that this important setting can only be controlled by an environment variable.

Actual behavior
---------------
As it turns out, Windows sets TEMP and TMP by default and TMP seems to have precedence in restic.

The documentation is misleading.

I'm new to restic. My first impression was quite good, especially compared to other comparable backup tools. However there are some problems that need to be addressed.

MichaelEischer:
TMP takes precedence over TEMP.

Fixes https://github.com/restic/restic/issues/5572

Checklist
---------

### DIFF
--- a/doc/047_tuning_backup_parameters.rst
+++ b/doc/047_tuning_backup_parameters.rst
@@ -104,7 +104,7 @@ pack size in MiB.  Restic currently defaults to a 16 MiB pack size.
 The side effect of increasing the pack size is requiring more disk space for temporary pack
 files created before uploading.  The space must be available in the system default temp
 directory, unless overwritten by setting the ``$TMPDIR`` (except Windows) environment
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
LGTM

PR #5607, opened 2025-11-18
Work around broken Readdirnames

vasi:

FUSE on Linux sometimes returns dir-entries with st_ino == 0, which causes Readdirnames() to filter those items out. This break `restic check` on those filesystems. Instead, use our own `readdirnames()` that includes those items.

This affects potentially all uses of Restic in Deja Dup, see this ticket: https://gitlab.gnome.org/World/deja-dup/-/issues/623#note_2606051

This is probably a bug in FUSE, see ticket: https://github.com/libfuse/libfuse/issues/1338 . But there's no fix yet, and versions with the bug have been released into the wild, so we should workaround it when possible.

Discussed in the Deja Dup issue tracker: https://gitlab.gnome.org/World/deja-dup/-/issues/623#note_2606051

Checklist
---------

     - It's not clear that there's a good way to test this, it relies on the version of libfuse installed
    - Should not be user-visible

### DIFF
--- a/changelog/unreleased/pull-5607
+++ b/changelog/unreleased/pull-5607
@@ -1,0 +1,7 @@
+Bugfix: Work around FUSE readdir bug on Linux
+
+Restic used to use the Go standard library Readdirnames() function, which ignored files with a zero inode. With certain versions of FUSE, this caused `restic check` to fail on some FUSE filesystems, being unable to find some pack files. This applied particularly to usage of Restic by Deja Dup.
+
+Now, we use our own readdirnames() implementation that doesn't ignore any files, and `restic check` should succeed even with buggy versions of FUSE.
+
+https://github.com/restic/restic/pull/5607
--- a/internal/backend/local/local.go
+++ b/internal/backend/local/local.go
@@ -286,7 +286,7 @@ func visitDirs(ctx context.Context, dir string, fn func(backend.FileInfo) error)
 		return err
 	}
 
-	sub, err := d.Readdirnames(-1)
+	sub, err := readdirnames(d)
 	if err != nil {
 		// ignore subsequent errors
 		_ = d.Close()
--- a/internal/backend/local/local_linux.go
+++ b/internal/backend/local/local_linux.go
@@ -1,0 +1,47 @@
+package local
+
+import (
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+// Read all dir entries, including even entries with st_ino == 0
+// Works around bugs in FUSE's synthetic inode number generation.
+
+func parseDirent(buf []byte) (consumed int, name string, err error) {
+	dirent := (*syscall.Dirent)(unsafe.Pointer(&buf[0]))
+	nameBytes := (*[256]byte)(unsafe.Pointer(&dirent.Name[0]))
+	for i := 0; i < 256; i++ {
+		if nameBytes[i] == 0 {
+			name = string(nameBytes[:i])
+			break
+		}
+	}
+	return int(dirent.Reclen), name, nil
+}
+
+func readdirnames(f *os.File) ([]string, error) {
+	buf := make([]byte, 8192)
+	names := []string{}
+	for {
+		bytes, err := syscall.Getdents(int(f.Fd()), buf)
+		if err != nil {
+			return nil, err
+		}
+		if bytes == 0 {
+			break
+		}
+		for n := 0; n < bytes; {
+			consumed, name, err := parseDirent(buf[n:])
+			if err != nil {
+				return nil, err
+			}
+			if name != ".." && name != "." {
+				names = append(names, name)
+			}
+			n += consumed
+		}
+	}
+	return names, nil
+}
--- a/internal/backend/local/local_nonlinux.go
+++ b/internal/backend/local/local_nonlinux.go
@@ -1,0 +1,10 @@
+//go:build !linux
+// +build !linux
+
+package local
+
+import "os"
+
+func readdirnames(f *os.File) ([]string, error) {
+	return f.Readdirnames(-1)
+}


MichaelEischer:
I have to say I _seriously_ dislike rolling our readdirnames implementation. There's a reason such things are part of the standard library. As inode numbers are expected to be unique within a filesystem, returning zero for multiple independent entries clearly a libfuse/gvfs bug.

At the very least I want to be sure that there's no other better fix that can be implemented reasonably fast. Does that bug only affect directory entries for directories or also for files? If it's just directories then it would also be an option to hardcode the 256 foldernames in the `data` directory.

vasi:
> I have to say I _seriously_ dislike rolling our readdirnames implementation. There's a reason such things are part of the standard library. As inode numbers are expected to be unique within a filesystem, returning zero for multiple independent entries clearly a libfuse/gvfs bug.

I hear that. It's really unfortunate the broken libfuse was released.

Unfortunately, I feel very uncomfortable trying to fix libfuse because of the lack of a clear specification for high-level FUSE filesystem readdir. It's clearly allowed for filesystems to return a NULL `*struct stat`, implying fuse should generate inode numbers; and it's allowed to return a real struct stat with positive st_ino. But it's not documented what should happen if it returns real struct stat with zero st_ino. I believe it used to always generate inodes in this case,  it still does _sometimes_, and I think that's the correct behavior. But it's not written down anywhere.
 
> At the very least I want to be sure that there's no other better fix that can be implemented reasonably fast. Does that bug only affect directory entries for directories or also for files? If it's just directories then it would also be an option to hardcode the 256 foldernames in the `data` directory.

That's a good idea, but I think it may affect both files and directories. As I understand it, it may happen anytime we need multiple calls to the `getdents` syscall for a single directory. So it's more likely on directories with many entries, which is why we're seeing it here. But we may also see it in the `Readdir(-1)` later in this file! Ugh.

One better way to do this might be calling `syscall.ReadDirent` with a really large buffer size, and then using `syscall.ParseDirent` on the result. It's still a bit dirty, but at least we wouldn't have to roll our own parsing like I do here. Would that feel ok to you?

MichaelEischer:
> But it's not documented what should happen if it returns real struct stat with zero st_ino.

At least the linux kernel header files say the following https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/fs.h?h=v6.17.8#n3365 :

>  * Userspace may rely on the inode number being non-zero. For example, glibc
> * simply ignores files with zero i_ino in unlink() and other places.

That was ensured in https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.17.8&id=2adc376c551943a07170cbe70f43e6d6065f8906, by checking that the linux kernel doesn't generate zero inodes.

On the other hand glibc recently has changed their behavior in regard to zero inodes: https://sourceware.org/bugzilla/show_bug.cgi?id=12165 .

But those changes are about filesystems consistently returning zero inodes or not.

> I believe it used to always generate inodes in this case, it still does sometimes, and I think that's the correct behavior. But it's not written down anywhere.

 Judging from the description above libfuse manages to return different inodes for a directory entry depending on the buffer size used to read from the directory! This is obviously completely broken. This clearly makes it a regression in libfuse. Based on the above, using a libc before 2.37 together with the problematic libfuse version will probably also result in missing files.

I wonder whether it would be a good idea to ask the go what they think about picking up the changed behavior of glibc.

MichaelEischer:
> One better way to do this might be calling `syscall.ReadDirent` with a really large buffer size, and then using `syscall.ParseDirent` on the result.

That will then also skip directory entries with zero inodes. I mean we could use a patched version of the code from the go standard library, while that would be a bit verbose it would definitely be better than trying to roll our own dirent parser.

vasi:
> > But it's not documented what should happen if it returns real struct stat with zero st_ino.
> 
> At least the linux kernel header files say the following https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/fs.h?h=v6.17.8#n3365 :

We're talking about two different specs here! There's the "Linux VFS/POSIX" specification (what you're referring to) and the "internal-to-libfuse FUSE-high-level-filesystem" behavior (what I  meant). You're absolutely right that Linux VFS/POSIX allows ignoring files with a zero st_ino.

But libfuse tries to make things easier for FUSE filesystem developers by offering a "high-level" filesystem API that is different from the Linux VFS. Eg: it allows filesystems to return a NULL struct-stat from readdir, and then fills in synthetic inode numbers. (It allows lots of other simplifications too, eg: allowing one-shot readdir instead of multiple calls, or looking up a complete path instead of one component at a time.) The question here is about whether this synthetic inode number behavior should apply _within libfuse_ to a FUSE filesystem returning zero st_ino, and that's the bit I find under-specified.

> On the other hand glibc recently has changed their behavior in regard to zero inodes: https://sourceware.org/bugzilla/show_bug.cgi?id=12165 .

This is a great find. I like the idea of asking the Go developers about this, I'll go file a ticket with them.

> Judging from the description above libfuse manages to return different inodes for a directory entry depending on the buffer size used to read from the directory! This is obviously completely broken. This clearly makes it a regression in libfuse. Based on the above, using a libc before 2.37 together with the problematic libfuse version will probably also result in missing files.

Agreed that this seems like very silly behavior.

vasi:
Filed with Go here: https://github.com/golang/go/issues/76428

vasi:
> > One better way to do this might be calling `syscall.ReadDirent` with a really large buffer size, and then using `syscall.ParseDirent` on the result.
> 
> That will then also skip directory entries with zero inodes.

This is what I meant: https://github.com/restic/restic/compare/master...vasi:restic:readdir2

It does work, but I'm really not sure if I prefer it!

MichaelEischer:
Thanks a lot for the investigation into this bug!

> This is what I meant: [master...vasi:restic:readdir2](https://github.com/restic/restic/compare/master...vasi:restic:readdir2)
> 
> It does work, but I'm really not sure if I prefer it!

Ah, reading the directory listing at once could work. Although I somewhat doubt that it will work reliably. With the async preemption feature of Go, long running syscalls will be interrupted from time to time and especially for larger repositories that seems likely to happen. (The current code in this PR would also need to correctly handle EINTR errors returned by the syscalls)

How urgent do you think the bugfix is? As it looks like Go upstream is interest in aligning the zero inode handling, I'd somewhat prefer to just wait until a Go version is released that includes the fix. It will still take several weeks to a few months until a new restic version is ready for a release.

vasi:
I also sent a PR to libfuse: https://github.com/libfuse/libfuse/pull/1372

Hopefully one or both of Go/libfuse takes my PR, and then we're out of the woods and can close this. Let's give them a couple weeks and revisit?

MichaelEischer:
> Let's give them a couple weeks and revisit?

Sounds good.

vasi:
Update after a couple of weeks:

* libfuse took my patch, and plans a new release next week
* Go took my patch, but 1.26 isn't due to release until Feb. My patch will be in the next 1.25.x release, but there's no fixed schedule for minor releases. Maybe it'll drop in Jan?

I think the only thing left for Resic is to remember to make a new build when the next Go release or patch comes out. Should that be tracked somehow? Then closing this feels appropriate.

mikix:
If [historical trends](https://go.dev/doc/devel/release) continue, I agree that Go is likely to release 1.25.6 in the first week or two of Jan. Once it does drop, I'd be very interested in a Restic 0.18.2 release. This issue and a waiting [chmod fix](https://github.com/restic/restic/commit/7f3e3b77cea69ac3edf92c865b24d3f60dfb4912) would both really help Deja Dup when it's working with fuse mounted backup locations.

No pressure, but if it happens to fit with your existing plans to cut a release in January built with the new Go release, that would be swell. I could then drop a new Deja Dup release using it. A little cascading chain of releases. :smile:

Regardless, I agree that this PR is not necessary any more - but huge thanks to @vasi for doing the research and legwork to get this fixed in Go. (And the fuse fix which got released in [3.18.0](https://github.com/libfuse/libfuse/releases/tag/fuse-3.18.0).)

mikix:
Go released 1.25.6, so all that needs to be done now is rebuild with the new release 👍

MichaelEischer:
Thanks for the work on fixing the underlying issues. I'll close this PR for now as it is no longer required. My plan is to get a new release out in the next few weeks.

PR #5610, opened 2025-11-19
check/copy/diff/stats: reduce memory usage

MichaelEischer:
The `index.AssociatedSet` is a very memory efficient way to track BlobHandles. Up to now its usage was limited to the `prune` command. However, after https://github.com/restic/restic/pull/5560 it's now easily possible to use it in place of a `restic.BlobSet`. Replace all BlobSet usages where it is expected that a large number of BlobHandles are stored there.

Test generation was AI assisted.

No. I had the idea while reviewing https://github.com/restic/restic/pull/5469 .

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5611, opened 2025-11-20
docs: correct the schema provided for tag summary

insertish:

Updates the documentation page about scripting with a small fix for the tag command's summary message.

Actual schema in the code: https://github.com/restic/restic/blob/master/cmd/restic/cmd_tag.go#L76

No, just a small doc fix.

Checklist
---------

N/A


MichaelEischer (approved):
LGTM. Thanks!

PR #5612, opened 2025-11-21
repository: add async blob upload method

MichaelEischer:
Integrate the blobUploader used by the archiver into the repository. The new method has signature `SaveBlobAsync(ctx context.Context, t restic.BlobType, buf []byte, id restic.ID, storeDuplicate bool, cb func(newID restic.ID, known bool, size int, err error))`. In contrast to the old blobUploader it is independent of a buffer pool implementation.

This will make it easier to use async blob uploading in other parts of the code base.

The bufferPool in the archiver code is now based on `sync.Pool` and will automatically shrink and grow. This can especially reduce memory usage on systems with many CPU cores.

Preparation for https://github.com/restic/restic/issues/2446

Checklist
---------


MichaelEischer (commented):
LGTM.

PR #5613, opened 2025-11-22
data: replace Tree struct with an iterator

MichaelEischer:

In order to eventually be able to chunk trees (https://github.com/restic/restic/issues/2446), it is necessary to replace the `data.Tree` struct with an iterator for reading such a tree and also for writing.

The writing part is mostly implemented already. The PR converts the last remaining pieces that built a whole tree and serialized it using `json.Marshal`. There is also a new `TreeWriter` struct which slightly simplifies writing a trees.

For reading, the `data.Tree` struct is completely replaced with a `TreeNodeIterator`. As the underlying type is `iter.Seq[data.NodeOrError]`, it is composable can be easily mocked for tests. The `TreeNodeIterator` decodes nodes while iterating over a tree blob. This should reduce peak memory usage as now only the serialized tree blob and a single node have to alive at the same time. Using the iterator has implications for the error handling however. Now it is necessary that all loops that iterate through a tree check for errors before using the node returned by the iterator. While it would be possible to omit the error handling changes now, they are fundamentally necessary to eventually implement chunked tree blobs as the latter have to loaded while decoding the tree.

The other change is that it is no longer possible to iterate over a tree multiple times. Instead it must be loaded a second time. This only affects the tree rewriting code. `data.StreamTree` now uses synchronous callbacks. As the `TreeNodeIterator` to deserializes nodes during iteration, the processing moves from `StreamTrees` into the callback. Schedule them onto the workers used by StreamTrees for proper load distribution.

`diff` now uses a dedicated `DualTreeIterator`. The previous implementation stored the whole tree in a map and used it for checking overlap between trees. This is now replaced with the `DualTreeIterator`, which iterates over two trees in parallel and returns the merged stream in order. In case of overlap between both trees, it returns both nodes at the same time. Otherwise, only a single node is returned.

The `TreeNodeIterator` hardcodes certain assumptions about the structures of a tree blobs. I've added support to skip unknown top-level keys for future proofing.

Contains misc test fixes and cleanups:
- data.TestCreateSnapshot now creates actually valid trees and no longer uses duplicate filenames
- repository.WithBlobUploader no longer deadlocks if a test throws a fatal error within the callback
- `rtest.Equals` now uses generics instead of `any` for its parameters. This allows `rtest.Equals(t, value, nil)` to just work.
- `walker.TestTreeMap` has moved to the `data` package

Drive-by fix:
- `dump` did not handle errors when walking a tree of a snapshot failed

DualTreeIterator tests and skipping unknown keys are primarily AI generated using Cursor.

Prerequisite for https://github.com/restic/restic/issues/2446

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5614, opened 2025-11-23
repository: fix LookupBlobSize to also return pending blobs

MichaelEischer:
Extracted from https://github.com/restic/restic/pull/5472 as this is relevant independent of that PR. This change isn't really user visible. At the moment this could only result in a few warnings less from the `backup`  command when it reuploads missing blobs. All other places that write to a repository don't rely on knowing about pending blobs so far.

Extracted from #5472

Checklist
---------


MichaelEischer (commented):
LGTM.
Should have been merged before https://github.com/restic/restic/pull/5472 . But let's merge the tests anyways.

PR #5616, opened 2025-11-24
Backup from s3

BataevDaniil:

I want to do a backup with s3 in order to: 
- data storage in another data center in case of a failure in the current one 
- if a bug in the application erases a lot of files by mistake, then I will be able to restore the data at the time of the bug absence. 
- I'm doing a database backup at the same time, then I can restore the database and file storage data to consistency. 
I use restic because can take increment snapshots and deduplicate, which allows you to save storage space and roll back to the right place.

Previously, I tried to mount s3 as a file system using s3fs. But 
- it works very slowly 
- background memory 10% 
- if you accidentally perform any operation and cancel it, it will still be completed 
- use 100% CPU and a lot of memory when running 
I tested it on a storage with 500,000 files and a capacity of 150 GB.

In this pull request, I added a new format for the backup command for the path to s3 bucket and any of its internal parts of the tree. To do this, instead of the file path, specify the list `s3://bucket_name/maybe_deep_folder_or_file` 
`s3://` indicates that all files will be taken from s3. 
`bucket_name` is a required parameter. 
`maybe_deep_folder_or_file` is the path in the hierarchy whose internal tree we want to save

To run it at local, you need to:
1. create repository local
```bash
./restic -r ./test-repo init
```
2. build and run
```sh
make restic && RESTIC_PASSWORD=1234 AWS_ENDPOINT_URL=http://localhost:9000 AWS_ACCESS_KEY_ID=access_key AWS_SECRET_ACCESS_KEY=secret_key 
DEBUG_LOG=dubub.log ./restic -r ./test-repo backup s3://bucket_name_1 s3://bucket_name_2
```

Thanks

https://forum.restic.net/t/backup-of-an-s3-bucket/4852
https://forum.restic.net/t/creating-a-restic-backup-of-a-s3-bucket/10335

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5616
+++ b/changelog/unreleased/pull-5616
@@ -1,0 +1,15 @@
+Enhancement: Back up files directly from an S3 source
+
+Restic could only back up files from the local filesystem, or data piped in via
+`--stdin` / `--stdin-from-command`. Backing up the contents of an S3 bucket
+required mounting it first, for example with rclone or s3fs.
+
+Restic can now read backup sources directly from an S3-compatible object store.
+Pass one or more `s3://bucketname/prefix` targets to the `backup` command.
+The endpoint and credentials are read from the
+`AWS_ENDPOINT_URL`, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment
+variables. S3 sources cannot be combined with local paths in the same command.
+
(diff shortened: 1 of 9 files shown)


BataevDaniil:
Good afternoon!
Pull request is ready for review. Can you please take a look? @MichaelEischer

PR #5623, opened 2025-11-27
--skip-if-unchanged skips backup based on targets

Paulomen2712:
Previously, when using the --skip-if-unchanged flag for a backup, restic would only check if the root tree ID remained the same. This would mean if changes were made in unrelated paths, for example in the parent directories, the backup would still be made. If this flag is enabled, restic now goes through each target and checks if the specified base targets have changed.

This PR improves `restic backup --skip-if-unchanged` by making it ignore changes in parent directories outside the specified backup targets. Previously, snapshot creation could be triggered by unrelated metadata changes in ancestor folders, even if the backup paths themselves were unchanged. Now, restic compares only the tree nodes corresponding to the actual backup targets, skipping snapshots when those directories are unchanged.  

Closes #5575

Checklist
---------


cmd/restic/cmd_backup_integration_test.go
```diff
@@ -440,15 +440,17 @@ func TestIncrementalBackup(t *testing.T) {
 
 	rtest.OK(t, appendRandomData(testfile, incrementalFirstWrite))
 
-	opts := BackupOptions{}
+	opts := BackupOptions{SkipIfUnchanged: runtime.GOOS != "windows"}
```
MichaelEischer: Why does this exclude windows? Also please don't modify the `TestIncrementalBackup` test to use `SkipIfUnchanged`.

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -727,6 +730,20 @@ func TestBackupSkipIfUnchanged(t *testing.T) {
...
 	testRunCheck(t, env.gopts)
+
+	for i := 0; i < 3; i++ {
+		err := testRunBackupAssumeFailure(t, filepath.Dir(env.testdata), []string{"testdata/0", "testdata/nonexisting"}, opts, env.gopts)
+		rtest.Assert(t, err != nil && err.Error() == "at least one source file could not be read", "No data error expected")
+		testListSnapshots(t, env.gopts, 2)
+		rtest.OK(t, os.Chtimes(env.testdata, time.Now(), time.Now()))
+	}
+
+	t.Helper()
```
MichaelEischer: The main function of a test must not be marked as helper.

internal/data/tree.go
```diff
@@ -321,6 +321,20 @@ func FindTreeDirectory(ctx context.Context, repo restic.BlobLoader, id *restic.I
 	return id, nil
 }
 
+func FindTreeNode(ctx context.Context, repo restic.BlobLoader, id *restic.ID, nodepath string) (*Node, error) {
```
MichaelEischer, line 328: Please add tests for this function.

MichaelEischer (changes requested):
I've thought quite a while whether there's an alternative to `targetsChanged` that does not require iterating through the snapshot afterwards. But it would be rather complex. The only two suitable points would be `arch.trackItem` (when also passing the information to it which paths were explicitly specified by the user) or `arch.treeSaver` (which would also need to know the explicit paths).

The information about explicitly specified paths is by now part of the `tree` objects in `internal/archiver/tree.go` (when removing the `t.Explicit = false` line in that file). That could then be passed to `trackItem`, which would count modifications of explicit files. Then creating a snapshot would be skipped if that count is zero. But that currently has the problem that excluded items / errors aren't passed to `trackItem`, such that we'd have to chase corner cases.

`arch.treeSaver` also wouldn't know about excluded / errored files such that the problem is somewhat similar.

The main reason why I wanted to integrate with the `tree` datatype is that this would make it easier to correctly handle corner cases like a backup of `[]string{"../example", "example"}`. `targetsChanged` is currently unable to correctly handle such a case.

But as integrating into `saveTree` seems to be rather complex, it's probably better to stick to the `targetsChanged` approach but use a different list of paths. `tree` (`internal/archiver/tree.go` field `Explicit bool`) already tracks which tree elements are derived from explicitly specified paths. That is, by walking through that tree, we can generate a list of the paths at which the explicitly specified backup targets end up in the snapshot. The main change for that to work is removing the `t.Explicit = false` line `internal/archiver/tree.go`. Then, `saveTree` can just collect a list of the paths. With that my `../example, example` corner case should work.

The other corner case elephant in the room is when the old snapshot backed up `a, b, c` and the new one only backs up `a, b`. If neither a and b changed, then the PR would currently not create a new snapshot. That seems to be rather unexpected behavior for me. I'm not yet sure what would be the simplest approach to handle that. Maybe comparing the number of directories entries in all parents of explicit backup target might work, but I haven't thought that through yet. (Getting the number of directory entries of the current and previous snapshot is trivial in `saveTree`, so maybe that would could end up being easier after all). What do you think?

cmd/restic/cmd_backup_integration_test.go
Paulomen2712: That did have a reason, but I removed it now and it worked, so not sure anymore. The reason I modified the test was to put the flag there but to show it doesn't have any effect and the exact same behavior as before happenes. I just added a couple assert Snapshots. If you prefer I can create a new test. My aim is always to add as little code as possible

Paulomen2712: But I did change it to a separate test

Issue #5562, opened 2025-10-14
feature request: don't rewrite unchanged file names under progress bar


tinnamchoi:
Output of `restic version`
--------------------------

```
restic 0.18.1 compiled with go1.25.1 X:nodwarf5 on linux/amd64
```

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

Don't overwrite file paths (the ones under the progress bar) if unchanged.

What are you trying to do? What problem would this solve?
---------------------------------------------------------

Sometimes when restic is spending a lot of time on a large file, I like to look into the file manually to see if I even want to keep this file. It is currently impossible to copy the file path from my terminal emulator `st` as the line is rewritten even if the file hasn't changed.

Relevant discussion: https://github.com/restic/restic/issues/5383#issuecomment-3344370633

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

It's a great tool! Wish I found out about this sooner.

MichaelEischer:
Did you test whether this is already fixed with https://github.com/restic/restic/pull/5551 ? Instead of building restic yourself you can also set `RESTIC_PROGRESS_FPS=10` to achieve the same effect. At least with the Gnome Terminal that has fixed the selection problem.

tinnamchoi:
I did test it, it was not fixed by that PR.

MichaelEischer:
https://github.com/restic/restic/pull/5626 will reduce the redraw further if nothing changes in the progress bar. However, as it usually includes a timer, the progress bar will still be redrawn once per second.

It would definitely be possible to only refresh the status lines that have actually changed. If someone wants to help with that, feel free to open a PR.

dgkmd:
I slightly changed the status bar update code so that it only rewrites changed lines, instead of rewriting every line.
For unchanged lines, it just moves the cursor down instead of writing its content.

I've tested with Ubuntu 24.04.3 and st 0.9.3, the environment that @tinnamchoi had mentioned, and it resolves the problem.

MichaelEischer:
From #5671: the frequent redraws also consume quite a bit of bandwidth when running restic remotely.

PR #5626, opened 2025-11-29
ui: only redraw status bar if it has not changed

MichaelEischer:
Only redraw the status bar is something has changed. As most status bars also include a timer, this will throttle them to one redraw per second unless something else has changed.

Partially improves https://github.com/restic/restic/issues/5562

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5627, opened 2025-11-30
restore: tune fileswriter

MichaelEischer:
Increase the number of buckets in the fileswriter to reduce possible lock contention. In addition, keep file handles cached to not have to open and close each file for every blob written. While both changes should improve performance, my local tests rather looked like they fall into the category of micro optimizations. They might be more useful when restoring to remote filesystems.

Stumbled across the current behavior while investigating https://github.com/restic/restic/pull/5553 .

Checklist
---------

### DIFF
--- a/internal/restorer/filerestorer.go
+++ b/internal/restorer/filerestorer.go
@@ -79,12 +79,10 @@ func newFileRestorer(dst string,
 	workerCount := int(connections)
 
 	return &fileRestorer{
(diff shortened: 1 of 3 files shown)


akrabu:
Woohoo this one worked! I successfully restored the test repo to my exFAT drive on MacOS.  👍

MichaelEischer (commented):
LGTM

PR #5636, opened 2025-12-03
Clarify docs for parameters and move environment variables

MichaelEischer:
Make it slightly more clear that parameters apply to all commands.
And move environment variables to the scripting section. They were pretty much out of place in the `backup` section.

Discussed on IRC.

Checklist
---------


rawtaz:
I agree with the intention and the solution. Just noticing though that existing links to the "Tuning Backup Parameters" document will stop working.

Apparently one can map old paths to new ones in the RTD *project*, not in conf.py. Then again, documentation does evolve so perhaps it's not that big of a problem and it's a good thing if people update their links.

rawtaz (approved):

doc/075_scripting.rst
```diff
@@ -14,8 +14,99 @@
...
+
+Environment Variables
+*********************
+
+In addition to command-line options, restic supports passing various options in
+environment variables. The following lists these environment variables:
+
+.. code-block:: console
+
+    RESTIC_REPOSITORY_FILE              Name of file containing the repository location (replaces --repository-file)
```
kjetilho, line 30: This is not a new issue, but I think this is slightly misleading: the option overrides the environment variable when used, but "replaces" suggests it is the other way around.

It is hard to express this compactly, though.  Perhaps "(alternative to --repository-file)" or "(option --repository-file)".  Another option is to describe behaviour explicitly in a paragraph:

> When both an environment variable and the option it replaces is given, the command line option has precedence.

With the latter paragraph, the texts for each environment variable does not need changing.

rawtaz: I'm gonna go out on a limb here and suggest something else; that we remove the `(replaces --foo)` occurences completely. Reasoning:

- Currently only _some_ of the env vars have this notation, even though a few more of them have a corresponding CLI option (e.g. `RESTIC_CACHE_DIR` which has `--cache-dir`), so it's inconsistent/incomplete as it is now.
- Imagine we add the missing ones – we'll them mostly just repeat the same thing over and over again.
- That "same thing" is pretty obvious – more or less (I didn't check each one) all the env vars have a CLI option named the same thing. So, do we really need to mention this more or less obvious, not to mention conventional, fact on each line?

What we could do is to add a sentence at the bottom which clarifies that most of the env vars have a corresponding CLI option, and as already mentioned the CLI option taking precedence.

I think this would clear up the documentation and make it easier to digest, without much loss of understanding.

rawtaz: If we wanted to keep the occurences, perhaps `(overridden by --foo)` is better.

kjetilho: I think it is very useful to state the connection between option and environment variable explicitly, especially since there are so many exceptions.  I like your `overridden by --foo` suggestion.

rawtaz: Why? If someone is looking for an environment variable, they are hardly looking for a CLI option in the list of environment variables. And they are not going to look for CLI options in the list of environent variables. So what actual use case are you suggesting we solve by having this information here, when it is easily noted that most environment variables have an equivalent option with a 1:1 name?

kjetilho: This is exactly where I would look to see if an option can be replaced by an environment variable.  There really isn't a convenient way to find this information elsewhere in the docs, as far as I can tell.  (perhaps `restic generate` docs should be put somewhere under References?)

Since it is not consistent, it needs to be documented.  And yes, this means RESTIC_PACK_SIZE should get a note that the corresponding option is `--pack-size`.  The alternative is that all these environment variables (at least `RESTIC_*`) should be available as options as well, e.g. `--progress-fps` -- but we really do not want to support `--password`, so there goes the consistency and we need to document it anyway.

rawtaz: > This is exactly where I would look to see if an option can be replaced by an environment variable.

Yes, and the process would go like this:

- You know the name of the option you want to look for an equivalent to, e.g. `--cache-dir`.
- You look in this list of environment variables, and immediately see that the format for these variables is `RESTIC_` followed by a name in capital letters.
- Since you are looking for an environment variable for a "cache dir", you quickly find the `RESTIC_CACHE_DIR` environment variable.

What part of this process would not work this way? And in what part of it would you need a `(overridden by ...)` text?

> but we really do not want to support --password, so there goes the consistency and we need to document it anyway.

Not really, with a mention that *most* variables have a corresponding option. Anyone wanting to know what options there are are for a command expected to look in the CLI for the complete list, e.g. `restic help backup`.

kjetilho: I do not understand why you want to make the documentation harder to use, but if you insist.

rawtaz: I'm not looking to make it harder. I am looking for replies explaining the use cases, hence my questions. Can you answer them?

kjetilho: I already told you.  With your suggestion, I can no longer do a search, I have to do a speculative and domain specific transformation of the option name and search for that.  This is actively making the documentation harder to navigate, with absolutely no benefit.  It feels absurd that I have to explain this.

rawtaz: > I already told you. With your suggestion, I can no longer do a search, I have to do a speculative and domain specific transformation of the option name and search for that.

No, you never mentioned this. If you want to actually search the text for the option name, then that is a use case for keeping the option name. This is the reason I kept asking.

> This is actively making the documentation harder to navigate, with absolutely no benefit.

It's not all black and white. The benefit would be cleaner and less verbose documentation. Just like there is value to your specific use case, there is also value to not having too much text and repetition in e.g. documentation, in particular where 90% of it is easily explained in one common sentence instead.

---

I hope you understand that I was just looking for a more practical explanation as to why it matters.

doc/075_scripting.rst
```diff
@@ -14,8 +14,99 @@
...
+This section provides an overview of available environment variables and how
+certain tasks may be accomplished when you use restic via scripts.
+
+.. _environment-variables:
+
+Environment Variables
+*********************
+
+In addition to command-line options, restic supports passing various options in
+environment variables. The following lists these environment variables:
```
fabien-joubert: These two paragraphs convey the same information. The section introduction is sufficient, so I suggest removing the first paragraph to avoid redundancy.

```suggestion
.. _environment-variables:

Environment Variables
*********************

In addition to command-line options, restic supports passing various options in
environment variables. The following lists these environment variables:
```

fabien-joubert (commented):
Minor documentation cleanup noticed during review

doc/075_scripting.rst
MichaelEischer: I've slightly reduced the duplication, but prefer to keep some initial wording that mentions that the section contains more than just environment variables.

MichaelEischer (commented):
Let's go with those changes for now, although we might want to clean up the `replaces --option` thing in a separate PR.

PR #5643, opened 2025-12-05
Improve documentation for restore

kjetilho:
This tries to clarify some edge cases with how paths work for the `restore` sub-command, especially how to specify an exclude rule which works correctly with `--delete`.

This is based on discussion and exploration on IRC.

### DIFF
--- a/doc/050_restore.rst
+++ b/doc/050_restore.rst
@@ -19,13 +19,13 @@ Restoring from a snapshot
 
 Restoring a snapshot is as easy as it sounds, just use the following
 command to restore the contents of the latest snapshot to
-``/tmp/restore-work``:
+``/tmp/restore``:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore-work
+    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
 Use the word ``latest`` to restore the last backup. You can also combine
 ``latest`` with the ``--host`` and ``--path`` filters to choose the last
@@ -33,22 +33,22 @@ backup for a specific host, path or both.
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore latest --target /tmp/restore-art --path "/home/art" --host luigi
+    $ restic -r /srv/restic-repo restore latest --target /tmp/restore --path "/home/art" --host luigi
     enter password for repository:
-    restoring <Snapshot of [/home/art] at 2015-05-08 21:45:17.884408621 +0200 CEST> to /tmp/restore-art
+    restoring <Snapshot of [/home/art] at 2015-05-08 21:45:17.884408621 +0200 CEST> to /tmp/restore
 
-The files will be restored to ``/tmp/restore-art/home/art``.
+The files will be restored to ``/tmp/restore/home/art``.
 
 Use ``--exclude`` and ``--include`` to restrict the restore to a subset of
 files in the snapshot. For example, to restore a single file:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore-work --include /work/foo
+    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore --include /work/foo
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
-This will restore the file ``/work/foo`` to ``/tmp/restore-work/work/foo``.
+This will restore the file ``/work/foo`` to ``/tmp/restore/work/foo``.
 
 To only restore a specific subfolder, you can use the ``<snapshot>:<subfolder>``
 syntax, where ``snapshot`` is the ID of a snapshot (or the string ``latest``)
@@ -58,11 +58,11 @@ to ``subfolder``.
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore 79766175:/work --target /tmp/restore-work --include /foo
+    $ restic -r /srv/restic-repo restore 79766175:/work --target /tmp/restore --include /foo
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
-This will restore the file ``/work/foo`` to ``/tmp/restore-work/foo``.
+This will restore the file ``/work/foo`` to ``/tmp/restore/foo``.
 
 You can use the command ``restic ls latest`` or ``restic find foo`` to find the
 path to the file within the snapshot. This path you can then pass to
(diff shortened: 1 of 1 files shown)


doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
 Use the word ``latest`` to restore the last backup. You can also combine
 ``latest`` with the ``--host`` and ``--path`` filters to choose the last
 backup for a specific host, path or both.
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore latest --target /tmp/restore-art --path "/home/art" --host luigi
+    $ restic -r /srv/restic-repo restore latest --target /tmp/restore --path "/home/art" --host luigi
```
rawtaz: This seems to be the only place where a simple static value is quoted, so I suggest we take the opportunity to remove the quotes around `/home/art` as they are not needed and such quoting isn't done in other places in this chapter. Can you add that change?

doc/050_restore.rst
```diff
@@ -134,6 +138,14 @@ files that do not exist in the snapshot. For this, pass the ``--delete`` option
...
 exist in the snapshot.
 
+With ``--delete``, ``--exclude`` has a dual purpose:
+
+* It will exclude files from being restored.  The exclude argument is taken
+  relative to the subfolder given with the snapshot (if any; otherwise ``/`` of
+  the snapshot.)
+* It will exclude files in the target directory from being deleted.  The exclude
+  argument is here relative to the target directory.
+
```
rawtaz: - What this addition tries to point out is already mentioned nine rows further down. Instead of having two separate paragraphs addressing the same thing, albeit in slightly different ways, it would be preferable to instead improve the existing one if needed.
- Also, the subfolder syntax has already been explained earlier – if that explanation wasn't good enough, it's better to improve that explanation instead of spreading it out in multiple places :) The point is to keep the different parts of the documentation focusing on explaining one thing each.
- In short, let us improve the other sections where needed, to make their docs so good that this addition isn't needed!

doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
 
-This will restore the file ``foo`` to ``/tmp/restore-work/work/foo``.
+This will restore the file ``/work/foo`` to ``/tmp/restore/work/foo``.
 
 To only restore a specific subfolder, you can use the ``<snapshot>:<subfolder>``
 syntax, where ``snapshot`` is the ID of a snapshot (or the string ``latest``)
-and ``subfolder`` is a path within the snapshot.
+and ``subfolder`` is a path within the snapshot.  This changes the root for the
+arguments to ``--include`` or ``--exclude``, they should be specified relative
+to ``subfolder``.
```
rawtaz: - While reading your suggestion here, it came to my attention that this section and also another mention of subfolders further down in this documentation needs improving.
- After thinking about ways to improve it, I do have some ideas I think are good, but the changes for them become too large to fit as comments or changes in this PR.
- For that reason, I'm thinking of opening a different PR for the sole purpose of improving the documentation for the subfolder feature to the restore command, and letting your improvement reside in this PR.
- I'd like to adjust your suggestion a bit to make it clearer, can you check the following suggestion?

~~~
and ``subfolder`` is a path within the snapshot. Note that the subfolder syntax
also affects options like ``--include`` and ``--exclude``, such that their
arguments should be specified relative to ``subfolder`` (e.g. ``/foo`` instead
of ``/work/foo``).
~~~

doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore latest --target /tmp/restore-art --path "/home/art" --host luigi
+    $ restic -r /srv/restic-repo restore latest --target /tmp/restore --path "/home/art" --host luigi
     enter password for repository:
-    restoring <Snapshot of [/home/art] at 2015-05-08 21:45:17.884408621 +0200 CEST> to /tmp/restore-art
+    restoring <Snapshot of [/home/art] at 2015-05-08 21:45:17.884408621 +0200 CEST> to /tmp/restore
+
+The files will be restored to ``/tmp/restore/home/art``.
```
rawtaz: - If I'm not misunderstanding your addition here, your goal is to point out how the restore snapshot's paths are populating the target directory of the restore, right? That makes sense, as this isn't really explained anywhere.
- However, I think we should make this a more elaborate paragraph, and that it should come earlier in this documentation, before where we introduce `--path`, for the below reason.
- The reason is that this comment is read to belong to the example right above it, and that example unfortunately introduces `--path /home/art`. This might very well confuse the reader, because they might think that `--path /home/art` has something to do with the fact that their files (e.g. if they were mentally looking to restore `/home/art/picture.jpg`) are restored into `/tmp/restore/home/art`, whereas in reality it's just a filter. I'd like to avoid that potential confusion.
- I suggest we:
  - Remove the added line in this change.
  - Keep the `s/restore-art/restore/`) change.
  - Look into the suggestions I wrote further up.

doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
-``/tmp/restore-work``:
+``/tmp/restore``:
 
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore-work
+    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
```
rawtaz, line 29: - I think this would be a good place to add an explanation of how the paths in the snapshot populates the target directory of the restore (see comments below this comment for reasoning – I'm basically suggesting moving and elaborating on your addition further down).
- I'm suggesting the following text that hopefully explains this once and for all:

~~~
You will notice that the paths listed in the snapshot you are restoring will
be the paths populating the target directory of the restore. For example, if
``restic snapshots`` lists the path ``/home/art`` in the snapshot to restore
and you specify ``--target /tmp/restore``, the result will be that ``/home/art``
from the snapshot will be restored as the path ``/tmp/restore/home/art``.
~~~

doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
 .. code-block:: console
 
-    $ restic -r /srv/restic-repo restore 79766175:/work --target /tmp/restore-work --include /foo
+    $ restic -r /srv/restic-repo restore 79766175:/work --target /tmp/restore --include /foo
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
-This will restore the file ``foo`` to ``/tmp/restore-work/foo``.
+This will restore the file ``/work/foo`` to ``/tmp/restore/foo``.
```
rawtaz: - I suggest we change this to the following, as the file itself will not be restored "to" itself, but rather to its parent path *or* "as" itself with a given path:

```suggestion
This will restore the file ``/work/foo`` as ``/tmp/restore/foo``.
```

doc/050_restore.rst
```diff
@@ -19,46 +19,50 @@ Restoring from a snapshot
...
 
-    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore-work
+    $ restic -r /srv/restic-repo restore 79766175 --target /tmp/restore
     enter password for repository:
-    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore-work
+    restoring <Snapshot of [/home/user/work] at 2015-05-08 21:40:19.884408621 +0200 CEST> to /tmp/restore
 
 Use the word ``latest`` to restore the last backup. You can also combine
 ``latest`` with the ``--host`` and ``--path`` filters to choose the last
 backup for a specific host, path or both.
```
rawtaz: - This change will make it more in line with the previous paragraph (saying "to do X, do this:" with a colon at the end), and also make it feel more okay to remove the added line two comments down from here.

```suggestion
backup for a specific host, path or both:
```

rawtaz (dismissed):
Some thoughts, comments and suggestions.

doc/050_restore.rst
andrewsav-bt: I think this is incorrect. If the snapshot was backed up with the relative path, then it won't be restored with the full path shown by `restic snapshots`

MichaelEischer: There's by now https://restic.readthedocs.io/en/latest/040_backup.html#absolute-and-relative-paths . So let's omit this part here.

MichaelEischer (approved):
LGTM

doc/050_restore.rst
MichaelEischer: I've replaced the example snapshot with one that contains two paths such that files are actually restored outside `--path`.

PR #5650, opened 2025-12-27
docs: add warning for capability-based non-root backups

fabien-joubert:

This PR adds a prominent warning to the "Full backup without root" example in the
documentation.

The example relies on the Linux capability `CAP_DAC_READ_SEARCH`, which allows any
process executing the capability-enabled restic binary to bypass standard file
permission checks for reading and directory traversal. This is a powerful privilege
that was not previously made explicit.

The added warning clarifies the security implications of this setup and points
readers to the relevant Linux documentation, helping users better understand the
trade-offs of capability-based non-root backups.

Discussed in #5506, specifically in https://github.com/restic/restic/issues/5506#issuecomment-3694178643.

Checklist
---------

### DIFF
--- a/doc/080_examples.rst
+++ b/doc/080_examples.rst
@@ -319,6 +319,18 @@ Note that when using a systemd unit to run restic, you can use
 Using file capabilities
 =======================
 
+.. warning::
+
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. Good idea, thanks!

PR #5660, opened 2026-01-01
fix for S3 restore failing with 202 Accepted for cold storage

vadimpanin:

S3 Glacier/DEEP_ARCHIVE restore fails with "202 Accepted" instead of waiting for objects to become available.

When using the experimental S3 cold storage restore feature (RESTIC_FEATURES=s3-restore), restoring objects from GLACIER or DEEP_ARCHIVE storage classes fails immediately with the message "202 Accepted" instead of waiting for the objects to become available. With `202 Accepted` "error" originating from `be.client.RestoreObject` (minio.Client) call in s3.go `requestRestore`.

Adding a handler for `StatusCode == 202` solves the issue:
```go
	if err := be.client.RestoreObject(ctx, be.cfg.Bucket, filename, "", opts); err != nil {
		var e minio.ErrorResponse
		if errors.As(err, &e) {
			// HTTP 202 Accepted means the restore request was successfully initiated
			if e.StatusCode == 202 {
				return true, nil
			}
			switch e.Code {
			case "InvalidObjectState":
				return false, nil
			case "RestoreAlreadyInProgress":
				return true, nil
			}
		}
		return false, err
	}
```

Tested with the number of days from 1 to 7, the issue is reproducible. Important note - on each execution one of the packs moves to "RestoreAlreadyInProgress" state. So, with just one pack to restore, a second attempt to restore will start waiting. With two packs to restore, the third attempt to restore will start waiting etc. Gets really difficult when you need to restore 176 packs.

Closes #5659

Checklist
---------

     No, but I can if requested. It will require mocking minio.Client though.
     No, but I don't think it's needed

### DIFF
--- a/changelog/unreleased/issue-5659
+++ b/changelog/unreleased/issue-5659
@@ -1,0 +1,14 @@
+Bugfix: Fix S3 DEEP_ARCHIVE restore failing with "202 Accepted"
+
+When restoring from S3 Glacier or DEEP_ARCHIVE storage classes using the
+experimental `s3-restore` feature, restic incorrectly treated the HTTP 202
+Accepted response as an error. This caused the restore to fail immediately
+with the message "202 Accepted" instead of waiting for the objects to become
+available.
+
+Restic now correctly recognizes HTTP 202 Accepted as a successful restore
+initiation and waits for the objects to transition from cold storage before
+proceeding with the restore.
+
+https://github.com/restic/restic/issues/5659
+https://github.com/restic/restic/pull/5660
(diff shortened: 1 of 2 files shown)


MichaelEischer:
I just took a closer look at the implementation of `RestoreObject` in minio-go: https://github.com/minio/minio-go/blob/38651126d634c1c731cf3c48ad6d153f8053bb82/api-restore.go#L178 is already supposed to handle this case, so this looks like a bug in the library. Could you open an issue at https://github.com/minio/minio-go/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen ?

MichaelEischer (changes requested):
(just for status tracking, see above)

pgreenland:
Stumbled across this PR having just found the same issue myself.

Happens with minio-go's example code too - believe I've identified the cause, raised an issue with them here: https://github.com/minio/minio-go/issues/2223

@vadimpanin's fix is the workaround I was planning on using locally.

Issue #5280, opened 2025-03-12
restic find -O -N does not produce any result when -O > -N.


wplapper:
Output of `restic version`
--------------------------
restic 0.17.3-dev (de9a040d2) compiled with go1.23.3 on linux/amd64

What backend/service did you use to store the repository?
---------------------------------------------------------
local filesystem

Problem description / Steps to reproduce
----------------------------------------

issue `restic  -r .. find xxx -O '2025-01-01' -N '2020-01-01'`. No output. 
Swap -O with -N and the result lines appear!

Expected behavior
-----------------
I expected an error message, since these given dates will never produce any meaningful result. 

I will create an PR and change `restic find` so it will produce an appropriate error message.

Actual behavior
---------------
No output, no error message, return code 0.

Do you have any idea what may have caused this?
-----------------------------------------------
Missing test for pat.oldest >= pat.newest.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Yes, I like restic very much.

wplapper:
Additional problems found:

What is the expected output from `restic -r ... find --pack 0bda2e56`. I receive not output. The debug output is very sparse:

2025/03/12 14:32:24 restic/cmd_find.go:435	main.(*Finder).packsToBlobs	1	Looking for packs...
2025/03/12 14:32:24 restic/cmd_find.go:448	main.(*Finder).packsToBlobs.func1	1	Found pack 0bda2e5615e7b346b28d5b1088d5d405401567edf8772151808577a3868458aa
2025/03/12 14:32:24 restic/cmd_find.go:485	main.(*Finder).packsToBlobs	1	11 blobs found
2025/03/12 14:32:24 restic/cmd_find.go:367	main.(*Finder).findIDs	1	searching IDs in snapshot 4faf7a13
The terminal shows no visible output, return code is 0.

Next problem: JSON output and text output are intermixed:
`restic -r ... find  --show-pack-id --tree efe30152342a7adbc2f01eee3e25d7b80c31e8ed79f291095240a1154a503074 --json`
output is 
`[{"object_type":"tree","id":"efe30152342a7adbc2f01eee3e25d7b80c31e8ed79f291095240a1154a503074","path":"/","snapshot":"4faf7a1363f88adcd37f41affe54c971cf4e56e7371c28ce739402645c8025e2","time":"2025-03-12T09:32:21.405919472Z"}]
[...]

PR #5664, opened 2026-01-10
Bugfix: ``restic find --pack <tree-pack>`` did not produce output for tree packs

wplapper:
cmd/restic/cmd_find.go: now produces output for both data-blob and tree-blob packfiles.
In ``func packsToBlobs()`` the blob type is now checked and the blob is either inserted into ``f.blobIDs`` or ``f.treeIDs``.

This PR fixes a bug in ``restic find --pack``.

Yes, in https://github.com/restic/restic/issues/5280

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_find_integration_test.go
+++ b/cmd/restic/cmd_find_integration_test.go
@@ -3,10 +3,10 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
+	"path/filepath"
 
 	"github.com/restic/restic/internal/global"
 	"github.com/restic/restic/internal/restic"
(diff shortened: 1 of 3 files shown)


cmd/restic/cmd_find_integration_test.go
```diff
@@ -141,3 +144,71 @@ func TestFindInvalidTimeRange(t *testing.T) {
...
+		rtest.Assert(t, len(jsonResult) > 0, "expected at least one tree record in the packfile")
+
+		// look at the last record
+		lastIndex := len(jsonResult) - 1
+		record := jsonResult[lastIndex]
+		rtest.Assert(t, record.ObjectType == "tree" && record.SnapshotID == sn1.String(),
+			"expexted a tree record with known snapshot id, but got type=%s and snapID=%s instead of %s",
+			record.ObjectType, record.SnapshotID, sn1.String())
+		// for windows only: convert \\ to /
+		backupPath = strings.ReplaceAll(backupPath, `\\`, "/")
```
MichaelEischer:
```suggestion
		backupPath = strings.ReplaceAll(backupPath, `\`, "/")
```

That should fix the windows test case. Please enable the assertion again on Windows.

cmd/restic/cmd_find_integration_test.go
```diff
@@ -141,3 +144,71 @@ func TestFindInvalidTimeRange(t *testing.T) {
...
+		// get the json records
+		jsonResult := []JSONOutput{}
+		rtest.OK(t, json.Unmarshal(results, &jsonResult))
+		rtest.Assert(t, len(jsonResult) > 0, "expected at least one tree record in the packfile")
+
+		// look at the last record
+		lastIndex := len(jsonResult) - 1
+		record := jsonResult[lastIndex]
+		rtest.Assert(t, record.ObjectType == "tree" && record.SnapshotID == sn1.String(),
+			"expexted a tree record with known snapshot id, but got type=%s and snapID=%s instead of %s",
```
MichaelEischer:
```suggestion
			"expected a tree record with known snapshot id, but got type=%s and snapID=%s instead of %s",
```

MichaelEischer (changes requested):
LGTM except for a few nits, see below.

cmd/restic/cmd_find_integration_test.go
MichaelEischer: Use `filepath.ToSlash(...)` instead.

wplapper: Windows test failed: reason being the drive mapping: Here is the Windows error message
```		// Windows response:
		//expected "C:/Users/RUNNER~1/AppData/Local/Temp/restic-test-3529440698/testdata/0/0/9" as part of
		//         "/C/Users/RUNNER~1/AppData/Local/Temp/restic-test-3529440698/testdata/0/0/9"
``

wplapper: I shortened the expected path to exclude the drive mapping.

MichaelEischer (approved):
LGTM. Thanks!

PR #5666, opened 2026-01-11
PR dead: Bugfix: restic find --json --show-pack-id: suppress text output in JSON mode

wplapper:
cmd/restic/cmd_find.go:
Suppress "normal" output of f.findObjectsPacks() in JSON mode.

It fixes mixed output in JSON mode. It actually suppresses the execution of the call to ``f.findObjectsPacks()``.

Yes, in https://github.com/restic/restic/issues/5280

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5666
+++ b/changelog/unreleased/pull-5666
@@ -1,0 +1,8 @@
+Bugfix: restic find --json --show-pack-id suppresses some output in JSON mode.
+
+When calling `restic find` with both options `--json --show-pack-id` in the past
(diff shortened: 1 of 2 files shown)


MichaelEischer:
This is IMHO worse than the current behavior, sorry. The whole point of specifying `--show-pack-id` is to get this output, thus silently hiding it will just confuse users. While we could return an error in this case, the problem should be obvious enough. So, I'd prefer to just keep things as is until we add some proper JSON output.

I'll close this PR as the latter suggestion is a rather different endeavor.

PR #5668, opened 2026-01-11
Enhancement: restic diff --content shows actual file differences between two versions of a file

wplapper:
Added code to compare contents of two snapshot files.

This PR allows content comparison of two files coming from two different snapshots. Only files which are marked "M" (by ``restic diff``) will be compared. Binary files will not be compared, but a message will be created. Text files larger than 64 KiB will be truncated at 64 Kib, the default value for option ``--diff-max-size``.

Since a text file of 64 KiB bytes will fit into the first data blob of a file, only the first data blob will ever be loaded for a text based file comparison. The same data blob will also be used to find out if a file is a `binary file`. A file is considered binary if it contains any non utf-8 data in the first 4 KiB of its content. The function `utf8.Valid()`` is used for this test.

Extracting the contents of these files is done via ``dump.WriteNode()``. In an subsequent step the two files will be compared via the package ``github.com/hexops/gotextdiff/myers``.

The file ``cmd/restic/cmd_diff_integration_test.go``contains 4 new tests to check most aspects of the new functionality. 
Also added various helper functions for the creation and modification of text files.

https://github.com/restic/restic/issues/5663

Checklist
---------

  see separate PR https://github.com/restic/restic/pull/21946
  see separate PR https://github.com/restic/restic/pull/21946


kcwu:
`diff --content-diff snap1:PATH snap2:PATH` works. However, `diff --content-diff snap1:PATH/FILE snap2:PATH/FILE` failed with error `path PATH/FILE: not a directory`.

But it is probably a separate issue since `diff` doesn't support file before this PR.

wplapper:
hi @kcwu

> diff --content-diff snap1:PATH snap2:PATH works. However, diff --content-diff snap1:PATH/FILE snap2:PATH/FILE failed with error path PATH/FILE: not a directory.
But it is probably a separate issue since diff doesn't support file before this PR.

Correct, this never worked before.

wplapper:
Hi https://github.com/dgkmd,

if you have some spare time, could you please have a look at the windows issues when comparing two files from different snapshots in a repository. Currently I am bypassing everything which is called `ẁindows``, because I have no idea to compare 2 windows files without investigating their type (binary/text). I also have no idea what the diff equivalent output looks like in windows.

Thanks, wplapper

cmd/restic/cmd_diff.go
```diff
@@ -476,3 +505,122 @@ func runDiff(ctx context.Context, opts DiffOptions, gopts global.Options, args [
...
+				printer.S("\n*** show contents diff for file %q ***", displayName1)
+				printer.S("%s\n\n", stdoutStderr)
+				mu.Unlock()
+			*/
+			mu.Lock()
+			printer.S("\n*** show contents diff for file %q ***", displayName1)
+			printer.S("no idea of how to run a file comparison successfully without knowing the file tyoe beforehand!")
+			mu.Unlock()
+		default:
+			return errors.Fatalf("don't know how tun run a file difference programme in the %q OS", runtime.GOOS)
```
flinteger-code: Thanks for the PR.
How about using a Go package to compute the diff instead of calling a system command? This would avoid external dependencies and inconsistencies across different diff implementations, and it would also simplify the code by removing the need for exec, locking, etc.

Some example diff package:
- https://github.com/sergi/go-diff
- https://github.com/pmezard/go-difflib
- https://www.google.com/search?q=golang+diff+package

dgkmd:
Hi @wplapper,

I looked into your code.

First, I agree with @flinteger-code that it would be a better choice to process `diff` inside Go, instead of calling a system command.
It will improve cross platform compatibility, and it also will be able to run the task without writing files on the disk.

On determining between text and binary files, I surveyed how linux `diff` determines between the two.
I found that it considers a file as a text file if it does *not* contain any 0x00 (null char) byte. This is a crude but valid criterion because linux uses UTF-8 encoding only.

As far as I know, in Windows there are many non-UTF8 encodings being used, such as UTF-16 and local codepages (such as CP949 for Korean), but recently it is moving to UTF-8 standard as well.

If it is okay to regard only UTF-8 as valid text encoding, I think that the easy choice is using [unicode/utf8.Valid()](https://pkg.go.dev/unicode/utf8#Valid) standard library function.

Also I think that it is a good idea to specify a file size threshold, because text files are usually small sized.

dgkmd:
Moreover, I have no experience in using WIndows equivalent of `diff` too. 
I searched the web, and there are similar functions in windows (`fc` for command prompt, `Compare-Object` for powershell), but it seems they are much less commonly used than linux `diff`. And they don't automatically distinguish between text and binary files.

cmd/restic/cmd_diff.go
wplapper: Hi @flinteger-code,
I have taken your advice to heart and I am implementing ``github.com/hexops/gotextdiff/myer`` comparison for text files. Binary files will be checked before comparing text files and will be excluded from any comparison operation.

Thanks for your help!

wplapper:
HI @dgkmd ,

Thanks for your time and advice. I have implemented the "zero" test for binary files. Secondly I am implementing ``github.com/hexops/gotextdiff/myers`` comparison of text strings. More details soon

cmd/restic/cmd_diff.go
flinteger-code: Thanks for making the changes.

cmd/restic/cmd_diff.go
```diff
@@ -1,22 +1,33 @@
...
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/global"
 	"github.com/restic/restic/internal/restic"
 	"github.com/restic/restic/internal/ui"
+	"github.com/restic/restic/internal/ui/progress"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
+const LargeFile = 16 << 20
```
flinteger-code: How about read this limit from command argument, like `--diff-max-size` or read from env like `RESTIC_DIFF_MAX_SIZE`?
16MB is lot of output to console, people may want limit diff size.

wplapper: Good idea, I have to think about it. 
I have a question for you: would you like to see the first ``--diff-max-size`` being diffed or just a plain message that this file is over (your) chosen limit and no diff output al all? 

I this this pretty straight forward to implement: modify the node.Content in such a way that it gets shortened so it just one blob over the chosen size limit and feed that into ``d.WriteNode()``.

flinteger-code: Good question. I think we can show the diff up to a certain limit, then display a message indicating that the file exceeds the limit and that the diff is only shown up to that threshold. The implementation can simply truncate the input when the file is oversized and set a oversize flag to print message. How do you think?

mingchen:
How's `--json` output with content diff? If json can capture the modified diff would be great.
e.g. 
> restic --json diff --content-diff snapshot1 snapshot2

wplapper:
Hi @mingchen,

> How's --json output with content diff? If json can capture the modified diff would be great.
e.g.     restic --json diff --content --json snapshot1 snapshot2

There will be an extended ``Change`` JSON document available, which will contain the content changes to text files.
Here it is:
```
type Change struct {
	MessageType string `json:"message_type"` // "change"
	Modifier    string `json:"modifier"`
	IsOversized bool   `json:"is_oversized,omitempty"`
	IsBinary    bool   `json:"is_binary,omitempty"`
	Path        string `json:"path"`
	Diff        string `json:"diff,omitempty"`
}
```
``Diff``will contain the actual changes in a unified diff format. ``IsOversized`` and ``IsBinary`` are bools which do what their names says: the signal if the file comparison is binary or if the file is larger than option ``--diff-max-size``.

cmd/restic/cmd_diff.go
```diff
@@ -93,6 +109,12 @@ type Change struct {
 	Modifier    string `json:"modifier"`
 }
 
+type ContentDiff struct {
+	node1 *data.Node
+	node2 *data.Node
+	name  string
+}
```
flinteger-code, line 122: The current implementation add a new `message_type` "diff_text_contents" along with "change", it make it hard to parse json.
could add diff as part of `Change` in line 106 as optional for modified files. e.g.

```
type Change struct {
	MessageType string `json:"message_type"` // "change"
	Path        string `json:"path"`
	Modifier    string `json:"modifier"`
    Diff          string `json:"diff"`  // optional, content diff for modified file ("modifier": "M")
}
```

With this change, instead of output two lines of json for modified file:
```
{"message_type":"change","path":"/doc.go","modifier":"M"}
{"message_type":"diff_text_contents","differences":"--- 2be50b5d /doc.go 2026-01-28 10:00:02 -0800 PST\n+++ 8d17bdc4 /doc.go 2026-01-28 10:13:17 -0800 PST\n@@ -2,9 +2,10 @@\n //\n // # Overview\n //\n-// The packages are structured so that cmd/ contains the main package for the\n+// Hello\n+//\n+// The packages are foo structured so that cmd/ contains the main package for the\n // restic binary, and internal/ contains almost all code in library form. We've\n-// chosen to use the internal/ path so that the packages cannot be imported by\n // other programs. This was done on purpose, at the moment restic is a\n // command-line program and not a library. This may be revisited at a later\n // point in time.\n"}
```

we got diff inside "change" line:
```
{"message_type":"change","path":"/doc.go","modifier":"M", "diff": "...the actual diff content..."}
```

wplapper: Hi @flinteger-code 

The two JSON outputs get created at different times. We probable have to defer the first lot (relatively small amount of data), until the content comparison is ready to roll. I will have a look.

It looks like this now:
```
{"message_type":"change","path":"/rclone_onedrive_restic_massive.stat","modifier":"M","oversized":true,"diff":"--- fb1a70e0 /home/wplapper/restic/stats/rclone_onedrive_restic_massive.stat 2025-12-28 10:24:51 +0000 UTC\n+++ dd8bfa7a /home/wplapper/restic/stats/rclone_onedrive_restic_massive.stat 2026-01-04 20:19:10 +0000 UTC\n@@ -6,6 +6,7 @@\n \"000645cbe7f57b8455d744f60adf1ea81b9bdeb46f405131e4b6445e844e36bf\": \"2025 ... etc ....
```

Is it ok? I have not committed it yet.
The ``diff_text_contents`` is gone! I will make the ``diff_binary_contents`` go away as well.

flinteger-code: Thanks for changes. This looks great. 👍

PR #5677, opened 2026-01-19
Enhancement: snapshots: Add --sort-by-time option

flinteger-code:

Current the `restic snapshots` output is always sort by time ASC. I want to view snapshots by time DESC instead.

This change introduce a new option `--sort-by-time asc|desc` for `snapshots` command. This is especially useful for scripting workflows that process output. Built-in sorting reduces the need for additional sorting in scripts.

```
      --sort-by-time string   sort snapshots by time: asc (ascending, oldest first) or desc (descending, newest first) (default "asc") (default: "asc")
```

No. This is a simple and compatible change.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_snapshots.go
+++ b/cmd/restic/cmd_snapshots.go
@@ -51,10 +51,11 @@ Exit status is 12 if the password is incorrect.
 // SnapshotOptions bundles all options for the snapshots command.
 type SnapshotOptions struct {
 	data.SnapshotFilter
-	Compact bool
-	Last    bool // This option should be removed in favour of Latest.
-	Latest  int
-	GroupBy data.SnapshotGroupByOptions
+	Compact    bool
+	Last       bool // This option should be removed in favour of Latest.
+	Latest     int
+	GroupBy    data.SnapshotGroupByOptions
+	SortByTime string
 }
 
 func (opts *SnapshotOptions) AddFlags(f *pflag.FlagSet) {
@@ -68,6 +69,7 @@ func (opts *SnapshotOptions) AddFlags(f *pflag.FlagSet) {
 	}
 	f.IntVar(&opts.Latest, "latest", 0, "only show the last `n` snapshots for each host and path")
 	f.VarP(&opts.GroupBy, "group-by", "g", "`group` snapshots by host, paths and/or tags, separated by comma")
+	f.StringVar(&opts.SortByTime, "sort-by-time", "asc", "sort snapshots by time: asc (ascending, oldest first) or desc (descending, newest first) (default: asc)")
 }
 
 func runSnapshots(ctx context.Context, opts SnapshotOptions, gopts global.Options, args []string, term ui.Terminal) error {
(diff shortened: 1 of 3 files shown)


wplapper:
Hello @flinteger-code,

Thank you four your PR. I have been looking through your list of changes and I was surprised to find all the manual pages changed. 

I think you could leave all the man pages untouched, since they are automatically created when a new release is generated by the restic maintainers, see function ``generateFiles()`` in  ``./helpers/prepare-release/main.go``.

That will take 37 changed files away, which automatically updated anyway.

flinteger-code:
> Hello @flinteger-code,
> 
> Thank you four your PR. I have been looking through your list of changes and I was surprised to find all the manual pages changed.
> 
> I think you could leave all the man pages untouched, since they are automatically created when a new release is generated by the restic maintainers, see function `generateFiles()` in `./helpers/prepare-release/main.go`.
> 
> That will take 37 changed files away, which automatically updated anyway.

Thanks for mention this, didn't notice this part of build script. Removed those changes.

cmd/restic/cmd_snapshots.go
```diff
@@ -163,10 +181,17 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
 		hasSize = hasSize || (sn.Summary != nil)
 	}
 
-	// always sort the snapshots so that the newer ones are listed last
-	sort.SliceStable(list, func(i, j int) bool {
-		return list[i].Time.Before(list[j].Time)
-	})
+	// Sort the snapshots based on sortByTime option
+	if sortByTime == "desc" {
```
MichaelEischer, line 185: The sorting in `printSnapshots` is actually redundant. The list is already sorted by `sort.Sort(sort.Reverse(list))`. Thus, remove the sorting in the print functions and presort the list at that central place.

cmd/restic/cmd_snapshots.go
```diff
@@ -51,10 +51,11 @@ Exit status is 12 if the password is incorrect.
...
 	data.SnapshotFilter
-	Compact bool
-	Last    bool // This option should be removed in favour of Latest.
-	Latest  int
-	GroupBy data.SnapshotGroupByOptions
+	Compact    bool
+	Last       bool // This option should be removed in favour of Latest.
+	Latest     int
+	GroupBy    data.SnapshotGroupByOptions
+	SortByTime string
```
MichaelEischer, line 58: I'd prefer to implement the sort-by using an extra datatype as done for example for `CompressionMode`.

cmd/restic/cmd_snapshots.go
```diff
@@ -78,6 +80,17 @@ func runSnapshots(ctx context.Context, opts SnapshotOptions, gopts global.Option
 	}
 	defer unlock()
 
+	// Set default sort-by-time if not specified
+	if opts.SortByTime == "" {
```
MichaelEischer, line 84: Parameter validation should happen before `openWithReadLock`. By now there's also a SnapshotOptions.Finalize method. Move the validation there.

cmd/restic/cmd_snapshots.go
```diff
@@ -68,6 +69,7 @@ func (opts *SnapshotOptions) AddFlags(f *pflag.FlagSet) {
 	}
 	f.IntVar(&opts.Latest, "latest", 0, "only show the last `n` snapshots for each host and path")
 	f.VarP(&opts.GroupBy, "group-by", "g", "`group` snapshots by host, paths and/or tags, separated by comma")
+	f.StringVar(&opts.SortByTime, "sort-by-time", "asc", "sort snapshots by time: asc (ascending, oldest first) or desc (descending, newest first) (default: asc)")
```
MichaelEischer, line 72: I'd prefer to use a `reverse` flag for consistency with `find` and `ls`.

MichaelEischer (changes requested):
Thanks for the improvement. I'd prefer using a `--reverse` option for consistency with the `find` and `ls` commands. The sorting code also needs some cleanups, see comments.

PR #5678, opened 2026-01-20
Enhancement: restic cat - add type `full-tree`

wplapper:
Create a topologically sorted list of tree nodes ID and names belonging to a snapshot or a subfolder of a snaphot

Added keyword "full-tree" to the list of acceptable types Moved the code validateCatArgs() up to check for simple errors as soon as possible.

Output is JSON only, like the rest of the ``cat`` command.

This PR creates a list of tree blobs which constitute all - or a selected subset of - tree blobs for a given snapshot. In the past, I was always struggling to see these details without doing repeated ``restic list`` and ``restic cat``.

Added the type ``full-tree``and created a function ``buildFullTree()`` which calls ``ẁalker.Walk()`` with appropriate parameters.
The result is gathered in a struct, which is then converted to a JSON structure of the form
```
{
  "snapshot": "774ebacda7f3fb8f92589773c023d8aa0fd21cd04c76d86309baf5e11aaa2f90",
  "root": "17c020f49e3a6f26f8a0a1312042a6a4d93430b956f13beea9426adf4880146a",
  "subdirecories": [
    {
      "subtree": "4f743fe78398e0a8c0ded6ac464df455c319b1a02d6cd2a2313b87c886c1b5f7",
      "path": "/home"
    },
		...
    {
      "subtree": "6409bed28d08898b849ecc4fdf338cdb0d67358619c99e6f6c3b402b1895baf8",
      "path": "/home/wplapper/restic/testdata/0/0/9"
    }
  ]
}
``` 

No.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-5678
+++ b/changelog/unreleased/pull-5678
@@ -1,0 +1,10 @@
+Enhancement: add new type `tree-full` to `restic cat`
+
+`restic cat tree-full <snapshot-id[:subfolder]> generates JSON output for all
+participating tree blobs for the selected snapshot or a subfolder thereof. This
+will help when troubleshooting problems and relating directory names to JSON
+nodes in the repository,
+
+https://github.com/restic/restic/pull/5679
+https://github.com/restic/restic/pull/5678
+https://github.com/restic/restic/issues/5674
--- a/cmd/restic/cmd_cat.go
+++ b/cmd/restic/cmd_cat.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"path"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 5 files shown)


wplapper:
Documentation for the new type ``full-tree`` has to wait until https://github.com/restic/restic/pull/5679 has been integrated.

cmd/restic/cmd_cat.go
```diff
@@ -207,14 +210,68 @@ func runCat(ctx context.Context, gopts global.Options, args []string, term ui.Te
...
+		SnapshotID restic.ID           `json:"snapshot"`
+		Tree       restic.ID           `json:"root"`
+		ItemList   []subdirectoryEntry `json:"subdirecories"`
+	}
+
+	treeStruct := snapshotTreeInfo{SnapshotID: *sn.ID(), Tree: *sn.Tree}
+	err := walker.Walk(ctx, repo, *sn.Tree, walker.WalkVisitor{
+		ProcessNode: func(parentTreeID restic.ID, nodepath string, node *data.Node, err error) error {
+			if err != nil {
+				printer.E("Unable to load tree %s  ... which belongs to snapshot %s", parentTreeID, sn.ID())
```
MichaelEischer, line 253: Shouldn't this rather print the nodepath? The snapshot ID is already known as the command only accepts a single snapshot

doc/view_repository.rst
```diff
@@ -361,6 +361,89 @@ Example::
 
 .. _cat_blob:
 
+full-tree snapshot[:subfolder]
+------------------------------
+
+If one is interested in the blob contents of the subfolder
+``/home/user/restic/testdata/0/for_cmd_ls`` but does not know the tree ID for the
+subfolder, one has to descend from the top of the tree step by step down to the subfolder
```
MichaelEischer, line 369: Isn't that exactly what `restic cat tree snapshotID:/home/user/restic/testdata/0/for_cmd_ls` would do (except for printing the tree ID)?

changelog/unreleased/pull-5678
```diff
@@ -0,0 +1,10 @@
+Enhancement: add new type `tree-full` to `restic cat`
+
+`restic cat tree-full <snapshot-id[:subfolder]> generates JSON output for all
+participating tree blobs for the selected snapshot or a subfolder thereof. This
+will help when troubleshooting problems and relating directory names to JSON
+nodes in the repository,
+
+https://github.com/restic/restic/pull/5679
```
MichaelEischer, line 8: Unrelated PR and issue

MichaelEischer (changes requested):
I understand what the code does, but I don't understand the use case so far. Can you explain that a bit? What is the added value on top of `restic cat tree snapshotID:/some/path`?

wplapper:
I am sorry to say that I have had a misconception how the snapshot:fubfolder syntax works. I cannot see any more reason to keep this PR alive. I will close the PR.

PR #5680, opened 2026-01-21
man: add note about append-only mode to restic-unlock.1

### LINKED ISSUE #5653: Explicitly document whether unlock works in append-only mode
The documentation should state, whether `restic unlock` works in append-only mode.
unlock says 'removed lock', which makes the reader of the documentation wonder, whether that works in append-only mode.

Understand how to correctly set up backups with append-only mode. I.e. can the 'untrusted' client system unlock a locked repo, or do I need to use the more secure client with delete/overwrite-access.

Yes, it's fun to use and makes regular backups easy.

castilma:
Since restic unlock _removes_ locks, a user reading about append-only might wonder whether this mode prevents removal of locks, too. This change explicitly mentions in the man page that the deletion of locks is an allowed exception.

Closes #5653 

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_unlock.go
+++ b/cmd/restic/cmd_unlock.go
@@ -19,6 +19,8 @@ func newUnlockCommand(globalOptions *global.Options) *cobra.Command {
 		Long: `
 The "unlock" command removes stale locks that have been created by other restic processes.
 
+Removing locks works even with repositories served in append-only mode from restic's rest-server.
+
 EXIT STATUS
 ===========
 


wplapper:
Hello @castilma,

Thank you for the PR. You want to consider to move your proposed change to the source code of the unlock command ``cmd/resttic/cmd_unlock.go``. The reason is a very simple one: ``doc/man/restic-unlock.1`` is generated automatically when a new release gets created. In other words, your proposed changes will never make it into the new release. All these man pages are created from their equivalent ``cmd/restic/cmd_xxx.go`` code. For ``cmd/restic/cmd_unlock.go`` it looks like this:
```
	cmd := &cobra.Command{
		Use:   "unlock",
		Short: "Remove locks other processes created",
		Long: `
The "unlock" command removes stale locks that have been created by other restic processes.
```
Add your changes after the last line of the above quote.

Look at ``restic generate --man`` for more details of how this is achieved.

cmd/restic/cmd_unlock.go
```diff
@@ -19,6 +19,8 @@ func newUnlockCommand(globalOptions *global.Options) *cobra.Command {
 		Long: `
 The "unlock" command removes stale locks that have been created by other restic processes.
 
+Removing locks works even with repositories served in append-only mode from Restic's rest-server.
```
MichaelEischer:
```suggestion
Removing locks works even with repositories served in append-only mode from restic's rest-server.
```

MichaelEischer (changes requested):
Just one small nit, see below. Otherwise LGTM.

MichaelEischer (approved):
LGTM. Thanks!

PR #5690, opened 2026-01-31
CI: prevent backends from importing internal/restic package

MichaelEischer:
The architectural layers in restic are `internal/data` above `internal/restic` / `internal/repository`  above `internal/backend`. Make sure that individual backends cannot break that rule. The cache needs an exception as it interacts more significantly with the repository.

Noticed the layering issue while taking a quick look at https://github.com/restic/restic/pull/5362 .

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5699, opened 2026-02-01
doc: describe effects of relative paths on snapshot structure

### LINKED ISSUE #2641: Restic unnecessarily rescans files when referring to them with different paths
Output of `restic version`
--------------------------

`restic 0.9.5 compiled with go1.12.9 on linux/arm`

`restic --repo rest:https://user:pass@foo.blah/user backup /data/Music/CDs`

Restic rest server

Expected behavior
-----------------

Restic should avoid rescanning the same files when they are referred to with a different path, where possible. For instance the contents are checked when backing up `$(pwd)` immediately after backing up `.`.

Actual behavior
---------------

Restic rescans these files. 

- Create a directory `/foo/bar` with many large files
- Navigate to `/foo/bar`
- Backup this directory using a relative path `restic backup .`
  - Observe that this takes a very long time to backup as it has to upload the files
- Backup this directory again using a relative path `restic backup .`
  - Observe that this is very fast as the files do not even need to be rescanned as their inode number and ctime's are the same as what was just uploaded
- Backup this directory using an absolute path `restic backup $(pwd)`
  - Observe that this operation is much slower as restic is reading the files (although not uploading them)

Restic also seems to rescan the files in `/foo/bar` when backup up `/foo` after the example above.

I suppose that restic is using the non-absolute path to the files as part of the determination of whether to rescan them.

I suppose restic could canonicalize the paths before deciding whether to recheck. I was surprised that the filename featured here at all actually, I'd have thought that checking the inode would be sufficient.

Very happy indeed! What a great bit of software! Thank you!

MichaelEischer:
Properly document the directory structure created when a backup uses relative paths.

Replaces https://github.com/restic/restic/pull/4443
Fixes https://github.com/restic/restic/issues/2641

Checklist
---------

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -315,6 +315,10 @@ Note that the snapshot metadata will always contain the absolute path.
     /user
     /user/work.txt
 
+.. note:: When switching between absolute and relative paths, the change detection in restic
+    will not be able to detect unmodified files. This is because the change detection depends
+    on the file path inside the snapshot.
+
 Dry Runs
 ********
 
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -279,6 +279,41 @@ the corresponding folder and use relative paths.
     processed 5307 files, 1.720 GiB in 0:03
     skipped creating snapshot
 
+.. _absolute-and-relative-paths:
+
+Absolute and relative paths
+***************************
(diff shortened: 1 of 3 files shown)


MichaelEischer (commented):
LGTM

PR #5700, opened 2026-02-01
fail if RESTIC_READ_CONCURRENCY or RESTIC_COMPRESSION are invalid

### LINKED ISSUE #4759: invalid environment variable values silently ignored

MichaelEischer:
Return an error if `RESTIC_READ_CONCURRENCY` or `RESTIC_COMPRESSION` cannot be parsed and they haven't been overwritten via their command line option.

Follow-up to https://github.com/restic/restic/pull/5592
Fixes https://github.com/restic/restic/issues/4759

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4759
+++ b/changelog/unreleased/issue-4759
@@ -1,0 +1,10 @@
+Bugfix: Return error if environment variables contain invalid values
+
+If the value of one of the environment variables `RESTIC_COMPRESSION`, `RESTIC_PACK_SIZE`
+or `RESTIC_READ_CONCURRENCY` could not be parsed, then restic ignored their value.
+Now, the restic commands fail with an error, unless they were overwritten with their
+corresponding command-line option.
+
+https://github.com/restic/restic/issues/4759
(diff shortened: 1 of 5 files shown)


MichaelEischer (commented):
LGTM

PR #5706, opened 2026-02-07
Add support for virtual device ids

intentionally-left-nil:
This PR addresses the issue of two backups, identical except for coming from different btrfs snapshots, produce blob changes due to the device_id changing. While the `device-id-for-hardlinks` flag address this for regular files, this behavior still persists for hard links. This PR provides a mechanism that addresses this problem for all files. It can be used with, or without the `device-id-for-hardlinks` flag

## Basic algorithm
The actual device id doesn't matter in itself. It's just used to keep inodes from different devices from conflicting. So, as long as we can keep the device id stable for all files on a device, and stable between runs, we can use any numbering scheme we want. Next, the backup logic is designed to always visit files in the same order for a given set of files to backup. Putting this together, and we can have a very simple algorithm: The first device_id that we see when backing files up gets an id of 1. The next device gets an id of 2, and so on. As long as backing up remains in the same order, every file will always get the same value.

## Implementation
The core logic is quite simple to manage, with a lot of piping to pass the data around. At its core, the Savepoint() logic is single-threaded. It eventually calls down into `nodeFromFileInfo`. Here, we add the logic to rewrite the device id to the virtual device id, and also cache the virtual id for future calls. The only complication is that the file uploader workers also need this mapping. The code is designed such that the workers get a read-only implementation of the mapping, and only the saveTree() enumeration does the writing. This ensures deterministic ordering and greatly increases the performance. sync.Map() is specifically built for this type of scenario and is quite fast (10ns / op).

## What if the virtual device id is wrong?
The algorithm will definitely ensure consistency within a backup. E.g. if the device_id is 142, then all nodes will have a device_id of 1 after the re-mapping. If there is a mismatch between device ids between different backups, this is already what happens today for the snapshot scenarios mentioned, so this is no different than today's behavior (where extra metadata is uploaded every backup for all the nodes with different device ids) This could happen if the user changes backup paths between different backups, or perhaps if the only file in a block device were removed between scans. Even with this user behavior, the system would re-stabilize once the backup paths become consistent across runs, and in the meantime the only real side effect is extra metadata uploading.

## Test plan
1. Existing unit tests pass
2. Created a unit test for the virtual device id behavior
3. Created a unit test to ensure that the code behaves well with `device-id-for-hardlinks`
4. Manually tested the various conditions with a test script (see the first comment in the PR for details)

https://github.com/restic/restic/issues/3041
That thread is very long (and old now), but should hopefully address any remaining difficulties

Checklist
---------

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -249,7 +249,7 @@ func (arch *Archiver) trackItem(item string, previous, current *data.Node, s Ite
 }
 
 // nodeFromFileInfo returns the restic node from an os.FileInfo.
-func (arch *Archiver) nodeFromFileInfo(snPath, filename string, meta ToNoder, ignoreXattrListError bool) (*data.Node, error) {
+func (arch *Archiver) nodeFromFileInfo(snPath, filename string, meta ToNoder, deviceIdMap deviceIdMapper, ignoreXattrListError bool) (*data.Node, error) {
 	node, err := meta.ToNode(ignoreXattrListError, func(format string, args ...any) {
 		_ = arch.error(filename, fmt.Errorf(format, args...))
 	})
@@ -268,6 +268,15 @@ func (arch *Archiver) nodeFromFileInfo(snPath, filename string, meta ToNoder, ig
 			// restic to upload new tree blobs
 			node.DeviceID = 0
 		}
+	}
+
+	if feature.Flag.Enabled(feature.VirtualDeviceId) {
+		virtualDeviceID, ok := deviceIdMap.GetVirtualId(node.DeviceID)
+		if !ok {
+			err = errors.Errorf("A virtual device ID for device %v was not found", node.DeviceID)
+			return node, arch.error(filename, err)
+		}
+		node.DeviceID = virtualDeviceID
 	}
 	// overwrite name to match that within the snapshot
 	node.Name = path.Base(snPath)
@@ -307,10 +316,10 @@ func (arch *Archiver) wrapLoadTreeError(id restic.ID, err error) error {
 
 // saveDir stores a directory in the repo and returns the node. snPath is the
 // path within the current snapshot.
-func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, meta fs.File, previous data.TreeNodeIterator, complete fileCompleteFunc) (d futureNode, err error) {
+func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, meta fs.File, previous data.TreeNodeIterator, deviceIdMap deviceIdMapper, complete fileCompleteFunc) (d futureNode, err error) {
 	debug.Log("%v %v", snPath, dir)
 
-	treeNode, names, err := arch.dirToNodeAndEntries(snPath, dir, meta)
+	treeNode, names, err := arch.dirToNodeAndEntries(snPath, dir, meta, deviceIdMap)
 	if err != nil {
 		return futureNode{}, err
 	}
@@ -334,7 +343,7 @@ func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, me
 			return futureNode{}, err
 		}
 		snItem := join(snPath, name)
-		fn, excluded, err := arch.save(ctx, snItem, pathname, oldNode)
+		fn, excluded, err := arch.save(ctx, snItem, pathname, oldNode, deviceIdMap)
 
 		// return error early if possible
 		if err != nil {
@@ -359,13 +368,13 @@ func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, me
 	return fn, nil
 }
 
-func (arch *Archiver) dirToNodeAndEntries(snPath, dir string, meta fs.File) (node *data.Node, names []string, err error) {
+func (arch *Archiver) dirToNodeAndEntries(snPath, dir string, meta fs.File, deviceIdMap deviceIdMapper) (node *data.Node, names []string, err error) {
 	err = meta.MakeReadable()
 	if err != nil {
 		return nil, nil, fmt.Errorf("openfile for readdirnames failed: %w", err)
 	}
 
-	node, err = arch.nodeFromFileInfo(snPath, dir, meta, false)
+	node, err = arch.nodeFromFileInfo(snPath, dir, meta, deviceIdMap, false)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -449,7 +458,7 @@ func (arch *Archiver) allBlobsPresent(previous *data.Node) bool {
 // Errors and completion needs to be handled by the caller.
 //
 // snPath is the path within the current snapshot.
-func (arch *Archiver) save(ctx context.Context, snPath, target string, previous *data.Node) (fn futureNode, excluded bool, err error) {
+func (arch *Archiver) save(ctx context.Context, snPath, target string, previous *data.Node, deviceIdMap deviceIdMapper) (fn futureNode, excluded bool, err error) {
 	start := time.Now()
 
 	debug.Log("%v target %q, previous %v", snPath, target, previous)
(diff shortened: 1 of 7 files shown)


intentionally-left-nil:
I created the following test harness to play around with backups. Since my /tmp and /home are on different devices it makes it easy to test the new behavior
```
#!/usr/bin/env bash
set -euo pipefail

if [[ $# -lt 1 ]]; then
  echo "Usage: $0 <init|backup|restore>" >&2
  exit 1
fi

RESTIC="restic" # Replace with your local path as need be
CMD="$1"
SAVEME=/tmp/saveme
BACKUP_REPO=/tmp/backup
export RESTIC_PASSWORD=test

case "$CMD" in
  init)
    rm -rf "${SAVEME:?SAVEME must be set}" "${BACKUP_REPO:?BACKUP_REPO must be set}"
    mkdir -p "$SAVEME/nested"
    echo -n "a" > "$SAVEME/a.txt"
    echo -n "b" > "$SAVEME/b.txt"
    ln -f "$SAVEME/b.txt" "$SAVEME/nested/hardlink.txt"
    ln -s ../a.txt "$SAVEME/nested/symlink.txt"
    "$RESTIC" -r "$BACKUP_REPO" init
    ;;
  backup)
    BACKUP_OUT=$("$RESTIC" -r "$BACKUP_REPO" backup ~/.inputrc "$SAVEME" 2>&1)
    echo "$BACKUP_OUT"
    SNAPSHOT_ID=$(echo "$BACKUP_OUT" | sed -n 's/.*snapshot \([a-f0-9]*\) saved.*/\1/p')
    [[ -z "$SNAPSHOT_ID" ]] && { echo "Failed to get snapshot ID from backup output"; exit 1; }

    echo "=== restic stats latest ==="
    "$RESTIC" -r "$BACKUP_REPO" stats latest

    echo ""
    echo "=== restic cat snapshot $SNAPSHOT_ID (snapshot object, includes root tree ID) ==="
    "$RESTIC" -r "$BACKUP_REPO" cat snapshot "$SNAPSHOT_ID" | jq . 2>/dev/null || "$RESTIC" -r "$BACKUP_REPO" cat snapshot "$SNAPSHOT_ID"

    echo ""
    echo "=== restic cat tree $SNAPSHOT_ID (root tree blob – internal nodes, device_id etc.) ==="
    "$RESTIC" -r "$BACKUP_REPO" cat tree "$SNAPSHOT_ID" | jq . 2>/dev/null || "$RESTIC" -r "$BACKUP_REPO" cat tree "$SNAPSHOT_ID"

    echo ""
    echo "=== restic cat tree ${SNAPSHOT_ID}:tmp/saveme/nested (nested dir tree blob) ==="
    "$RESTIC" -r "$BACKUP_REPO" cat tree "${SNAPSHOT_ID}:tmp/saveme/nested" | jq . 2>/dev/null || "$RESTIC" -r "$BACKUP_REPO" cat tree "${SNAPSHOT_ID}:tmp/saveme/nested"

    INPUTRC_PATH=$(echo "$HOME/.inputrc" | sed 's|^/||')
    INPUTRC_DIR=$(dirname "$INPUTRC_PATH")
    echo ""
    echo "=== restic cat tree ${SNAPSHOT_ID}:${INPUTRC_DIR} (tree containing .inputrc node) ==="
    "$RESTIC" -r "$BACKUP_REPO" cat tree "${SNAPSHOT_ID}:${INPUTRC_DIR}" | jq . 2>/dev/null || "$RESTIC" -r "$BACKUP_REPO" cat tree "${SNAPSHOT_ID}:${INPUTRC_DIR}"

    ;;
  restore)
    rm -rf "${SAVEME:?SAVEME must be set}"
    "$RESTIC" -r "$BACKUP_REPO" restore latest --target /
    echo "=== verifying /tmp/saveme structure ==="
    [[ -f "$SAVEME/a.txt" ]] || { echo "Missing $SAVEME/a.txt"; exit 1; }
    [[ "$(cat "$SAVEME/a.txt")" = "a" ]] || { echo "$SAVEME/a.txt content wrong"; exit 1; }
    [[ -f "$SAVEME/b.txt" ]] || { echo "Missing $SAVEME/b.txt"; exit 1; }
    [[ "$(cat "$SAVEME/b.txt")" = "b" ]] || { echo "$SAVEME/b.txt content wrong"; exit 1; }
    [[ -f "$SAVEME/nested/hardlink.txt" ]] || { echo "Missing $SAVEME/nested/hardlink.txt"; exit 1; }
    [[ "$(cat "$SAVEME/nested/hardlink.txt")" = "b" ]] || { echo "$SAVEME/nested/hardlink.txt content wrong"; exit 1; }
    B_INODE=$(stat -c %i "$SAVEME/b.txt")
    HL_INODE=$(stat -c %i "$SAVEME/nested/hardlink.txt")
    [[ "$B_INODE" = "$HL_INODE" ]] || { echo "Hardlink broken: b.txt inode $B_INODE, hardlink.txt inode $HL_INODE"; exit 1; }
    [[ -L "$SAVEME/nested/symlink.txt" ]] || { echo "$SAVEME/nested/symlink.txt is not a symlink"; exit 1; }
    [[ "$(readlink "$SAVEME/nested/symlink.txt")" = "../a.txt" ]] || { echo "Symlink target wrong"; exit 1; }
    [[ "$(cat "$SAVEME/nested/symlink.txt")" = "a" ]] || { echo "Symlink content wrong"; exit 1; }
    echo "OK: structure correct, hardlink shares inode with b.txt"
    ;;
  *)
    echo "Usage: $0 <init|backup|restore>" >&2
    exit 1
    ;;
esac

```

I was able to manually validate the following:
1. The normal behavior without any flags has the device_id on all nodes
2. With device-id-hardlinks only hardlinks have a device id
3. With virtual-device-id, they are all mapped, consistently to 1, 2, etc
4. The mappings remain consistent between runs
5. Restoring works correctly and the hardlink is recovered properly

internal/archiver/archiver.go
```diff
@@ -269,6 +269,15 @@ func (arch *Archiver) nodeFromFileInfo(snPath, filename string, meta ToNoder, ig
...
 		}
 	}
+
+	if feature.Flag.Enabled(feature.VirtualDeviceId) {
+		virtualDeviceID, ok := deviceIdMap.GetVirtualId(node.DeviceID)
+		if !ok {
+			err = errors.Errorf("A virtual device ID for device %v was not found", node.DeviceID)
+			return node, arch.error(filename, err)
+		}
+		node.DeviceID = virtualDeviceID
```
intentionally-left-nil, line 279: This is the only effective change to the backup behavior, and everything else in this PR is just piping

internal/archiver/virtual_device_id.go
```diff
@@ -0,0 +1,80 @@
...
+	m := &deviceIdMap{
+		count: 0,
+		cache: &vIdCache{},
+	}
+	m.cache.set(0, 0)
+	return m
+}
+
+// vIdCache is a concurrent read-only view of realID -> virtualID; GetVirtualId returns (0, false) when missing.
+type vIdCache sync.Map
```
intentionally-left-nil, line 42: Per the go docs:
>The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

Which is exactly the behavior that we have here. Only the one writer is adding entries, and they are all to new keys, so there is essentially zero lock contention, just the synchronization needed

MichaelEischer (changes requested):
Thanks for taking on this topic! As discussed in https://github.com/restic/restic/issues/3041#issuecomment-3929931326, I don't think this way of numbering the deviceIDs is the right way. Please follow the suggestion provided there.

MichaelEischer:
@intentionally-left-nil Are you still interested in reworking this PR?

MichaelEischer:
I'll close this PR for now as it needs a full rewrite and there's been no feedback. Feel free to open a new PR that addresses my review comments.

PR #5718, opened 2026-02-17
mount: check for more requisite mountpoint conditions

jtru:
Esp. with large repositories, `restic mount` may waste a lot of time and resources on preparing for a fuse mount that ultimately cannot succeed due conditions that are not checked for.

In order to be able to mount a repository over a mountpoint target directory via FUSE, that target directory needs to be both writeable and executable for the UID performing the mount, which this patch ensures.

FUSE does allow for mounting over a target path that refers to a regular (writeable) file, but the result is not accessible via chdir(), so we prevent that as well, and accept only directory inodes as the intended target mountpoint path.

No.

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/sys/unix"
 
 	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/debug"
@@ -35,8 +36,8 @@ func newMountCommand(globalOptions *global.Options) *cobra.Command {
 		Use:   "mount [flags] mountpoint",
 		Short: "Mount the repository",
 		Long: `
-The "mount" command mounts the repository via fuse to a directory. This is a
-read-only mount.
+The "mount" command mounts the repository via fuse over a writeable directory.
+The repository will be mounted read-only.
 
 Snapshot Directories
 ====================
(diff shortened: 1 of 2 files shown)


cmd/restic/cmd_mount.go
```diff
@@ -133,9 +134,19 @@ func runMount(ctx context.Context, opts MountOptions, gopts global.Options, args
 
 	// Check the existence of the mount point at the earliest stage to
 	// prevent unnecessary computations while opening the repository.
-	if _, err := os.Stat(mountpoint); errors.Is(err, os.ErrNotExist) {
+	mountpoint_stat, err := os.Stat(mountpoint)
```
MichaelEischer:
```suggestion
	stat, err := os.Stat(mountpoint)
```
Or `mountpointStat`. Go uses camel case.

MichaelEischer (changes requested):
Please add a small changelog entry, see https://github.com/restic/restic/blob/master/changelog/TEMPLATE for the template.

cmd/restic/cmd_mount.go
jtru: Ack, I pushed a change to rectify that :)

I hope the changelog entry is satisfactory!

MichaelEischer (approved):
LGTM. Thanks! I've slightly tweaked the changelog summary, we reserve `change` for breaking changes.

jtru:
Thanks very much for merging!

Apologies for causing the small bit of extra work - I wasn't aware of the special meaning of these words in the changelog context and mistook "change" for the most neutral/innocent of the three options ;)

PR #5720, opened 2026-02-18
mount: avoid duplicate index loading

MichaelEischer:
Only report mountpoint readiness after the snapshots are actually loaded.
The repository index can now be loaded incrementally. Previously, the index was loaded twice. I've opted for properly implementing incremental reloading instead of just getting rid of the initial call to `LoadIndex`. This has the benefit of also providing fast reloads if the snapshots change at runtime.

Performance issues were reported in https://forum.restic.net/t/restic-mount-takes-5-minutes/10645/4

Checklist
---------


MichaelEischer (commented):
LGTM

PR #5721, opened 2026-02-19
clarify changelog expectations

MichaelEischer:
Clarify the changelog description in several aspects that repeatedly come up during PR reviews. The description is somewhat longer than I'd prefer, but it's better to repeat things all the time.

Not specifically.

Checklist
---------

### DIFF
--- a/changelog/TEMPLATE
+++ b/changelog/TEMPLATE
@@ -1,14 +1,20 @@
 # The first line must start with Bugfix:, Enhancement: or Change:,
-# including the colon. Use present tense and the imperative mood. Remove
-# lines starting with '#' from this template.
-Enhancement: Allow custom bar in the foo command
+# including the colon. 'Change:' is for breaking changes only.
+# Documentation-only changes do not get a changelog entry.
+# Include the affected command in the summary if relevant.
+#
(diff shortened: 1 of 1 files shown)


changelog/TEMPLATE
```diff
@@ -1,14 +1,20 @@
...
+# Use present tense and the imperative mood. Remove lines starting
+# with '#' from this template.
+Enhancement: Allow custom bar in the `foo` command
 
 # Describe the problem in the past tense, the new behavior in the present
 # tense. Mention the affected commands, backends, operating systems, etc.
 # If the problem description just says that a feature was missing, then
-# only explain the new behavior.
-# Focus on user-facing behavior, not the implementation.
+# only explain the new behavior. Aim for a compact description.
```
rawtaz: Do you think that `short and concise` is better than `compact`?

rawtaz (approved):
Looks good to me, just one question in case you want to consider it.

changelog/TEMPLATE
MichaelEischer: It definitely sounds smoother. So let's go with it. We can still tweak it if it nudges people too far.

PR #5728, opened 2026-02-25
docs: move backup-without-root section to backup page

AdrianMastronardi:
Move the "Backing up your system without running restic as root" section from
the Examples page (`doc/080_examples.rst`) to the Backing up page
(`doc/040_backup.rst`), where it naturally belongs. Drop the less secure file
capabilities (`setcap`) method, keeping only the ambient capabilities approach
(`setpriv` and systemd), as agreed with @MichaelEischer in #4991.

A stub with a cross-reference is left in the Examples page to preserve
existing anchors/URLs.

Closes #4991

Checklist
---------


MichaelEischer:
I'd prefer to do it the other way around. Keep the backup-without-root-section in the examples, drop the file capabilities, add a reference from a new backup section to the example section and not in the example that this only works for Linux. That last part isn't immediately obvious.

MichaelEischer (changes requested):
See above

PR #5730, opened 2026-02-26
feat: Add support for setting `--ignore-ctime` and `--ignore-inode` via environment variables

### LINKED ISSUE #5738: Add support for setting `--ignore-ctime` and `--ignore-inode` via environment variables
Output of `restic version`
--------------------------

`restic 0.18.1 compiled with go1.25.1 on linux/amd64`

Restic should allow configuring `--ignore-ctime` and `--ignore-inode` via environment variables in addition to CLI args.

I have a relatively large PVC (~1.5TB). I use `volsync`'s `restic` mover to back it up daily.  The PVC data changes very little day-to-day. I'd say on most days it's about 99% the same as the previous day. 

Backups use `volsync`'s `snapshot` copy method, which creates a volume snapshot and mounts that to the `restic` mover pod. 

The problem is that snapshots do not preserve the file `ctime`. This causes `restic` to rescan the whole dataset, taking hours to complete and significantly increasing the cluster I/O load, while only backing up 10-15GB in the end.

`volsync` does not allow passing CLI args to `restic`, thus my only option is to set `--ignore-ctime` via an environment variable

Works great with volsync :))

astro-stan:

`volsync` does not allow passing CLI args to `restic`, thus the only option is to set `--ignore-ctime` and `--ignore-inode` via environment variables

This PR adds environment variables for the `ignore-ctime` and `ignore-inode` options. The environment variables are used as the default value for the CLI option, and can be overwritten by providing the CLI flag directly.

Closes #5738

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -134,22 +134,8 @@ func (opts *BackupOptions) AddFlags(f *pflag.FlagSet) {
 	f.StringVar(&opts.TimeStamp, "time", "", "`time` of the backup (ex. '2012-11-01 22:08:41') (default: now)")
 	f.BoolVar(&opts.WithAtime, "with-atime", false, "store the atime for all files and directories")
 
-	// Use environment to set the default value if available
-	ignoreInodeDefault := false
-	if v := os.Getenv("RESTIC_IGNORE_INODE"); v != "" {
-		if b, err := strconv.ParseBool(v); err == nil {
-			ignoreInodeDefault = b
-		}
-	}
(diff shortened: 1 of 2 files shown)


astro-stan:
Hello @fd0 ,

Apologies for the ping, but could you please take a look at this PR? It's only a few lines, and I'm really keen to get it merged :))

Thank you for your hard work!

MichaelEischer (approved):
LGTM. We have to get away from individually adding environment variables for CLI parameters, but that's not the problem for this PR.

PR #5746, opened 2026-03-10
update golang to 1.26.0

### LINKED ISSUE #5745: Update Golang to 1.26.0 to fix CVE-2025-61732 and CVE-2025-68121
Output of `restic version`
--------------------------

v0.18.1

Nothing should be added.

It would fix critical CVE and improve the security of restic.

nroeske-lab:
The PR updates the golang version to 1.26.0 and fixes CVEs

Closes #5745

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/restic/restic
 
-go 1.24.0
+go 1.26.0
 
 // keep the old behavior for reparse points on windows until handling reparse points has been improved in restic
 // https://forum.restic.net/t/windows-junction-backup-with-go1-23-or-later/8940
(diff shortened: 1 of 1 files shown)


clrossel:
Go version should be bumped to `1.26.1` now that a newer version than `1.26.0` is available.

MichaelEischer:
We don't want to drop compatibility with Go 1.25 just now. Even just compiling with go 1.24.13 would be sufficient to fix the CVEs.

As mentioned in https://github.com/restic/restic/issues/5745#issuecomment-4400773629 those CVEs are just false positives. That is it will be handled as part of the regular dependency updates in the next restic release.

PR #5748, opened 2026-03-12
implement restic backup --exclude-no-dump

### LINKED ISSUE #4208: Support for nodump attribute
We currently configure exclusion list using the `nodump` attribute, a standard file attribute on btrfs/ext* filesystem.

Excluding file using a per-file attribute is different than other supported method:

1. it does not restrict file name
2. it does not force to put a flag-file in a folder
3. it does not require a central "gitignore" file
4. it is already used by other backup tool (ie: Borg Backup).

The point 4 is not so trivial, we plan to use two different backup tool, one for detached disk and a different one for remote storage, so if we found a bug in a tool will be possible to recover data using a different tool.

I think that this feature should be implemented by every Linux backup tool, however if it is considered too specific (or unuseful), it is enough for us implementing an option like `--exclude-cmd`, so we can develop a custom logic ourself.

jim-minter:

This PR implements `restic backup --exclude-no-dump` (Linux only), which may be used to exclude all directories and files that have the "no dump" Linux file attribute set.  See [CHATTR(1)](https://www.man7.org/linux/man-pages/man1/chattr.1.html) for more details on Linux file attributes.

On non-Linux systems, the `--exclude-no-dump` option is not available.

Closes #4208

Checklist
---------

### DIFF
--- a/internal/archiver/attributes_linux.go
+++ b/internal/archiver/attributes_linux.go
@@ -1,0 +1,19 @@
+//go:build linux
+
+package archiver
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// isNoDump returns whether the "no dump" Linux file attribute is set on path.
+// See CHATTR(1) for more information about Linux file attributes.
+func isNoDump(path string) (bool, error) {
+	statx := &unix.Statx_t{}
+
+	if err := unix.Statx(0, path, unix.AT_NO_AUTOMOUNT|unix.AT_SYMLINK_NOFOLLOW, 0, statx); err != nil {
+		return false, err
+	}
+
+	return statx.Attributes&unix.STATX_ATTR_NODUMP != 0, nil
+}
--- a/internal/archiver/exclude.go
+++ b/internal/archiver/exclude.go
@@ -334,3 +334,16 @@ func RejectCloudFiles(warnf func(msg string, args ...interface{})) (RejectFunc, 
 		return false
 	}, nil
 }
+
+// RejectByNoDump returns a func which on Linux rejects files with the "no dump"
+// Linux file attribute is set.  On other OSes it rejects no files.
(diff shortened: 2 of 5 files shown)


internal/archiver/attributes_linux.go
```diff
@@ -0,0 +1,19 @@
...
+
+package archiver
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// isNoDump returns whether the "no dump" Linux file attribute is set on path.
+// See CHATTR(1) for more information about Linux file attributes.
+func isNoDump(path string) (bool, error) {
```
MichaelEischer, line 11: Extracting the `NODUMP` attribute must happen in the `fs` package. Add a `stat_linux.go` there.

As I'd like to avoid introducing duplicate stat calls, can the existing `Lstat` and the `Statx` calls be combined (probably with a fallback path for older linux versions, that just uses the regular Lstat call).

internal/archiver/exclude.go
```diff
@@ -334,3 +334,16 @@ func RejectCloudFiles(warnf func(msg string, args ...interface{})) (RejectFunc,
 		return false
 	}, nil
 }
+
+// RejectByNoDump returns a func which on Linux rejects files with the "no dump"
+// Linux file attribute is set.  On other OSes it rejects no files.
+func RejectByNoDump(warnf func(string, ...any)) (RejectFunc, error) {
+	return func(item string, _ *fs.ExtendedFileInfo, _ fs.FS) bool {
```
MichaelEischer, line 341: Should extract its information from the ExtendedFileInfo. See https://github.com/restic/restic/pull/5748/changes#r3487750658

MichaelEischer (changes requested):
Filesystem-specific code by now has to live in the `fs` package not the `archiver`, see comments for details. Please also add a changelog entry.,

PR #5750, opened 2026-03-14
filter: add Glob function supporting ** wildcard patterns

### LINKED ISSUE #5729: The ** wildcard should work with everything the same way it works with exclusions
Output of `restic version`
--------------------------
`restic 0.18.1 compiled with go1.25.6 X:nodwarf5 on linux/amd64
`

The `**` wildcard should work with everything the same way it works with exclusions.

I'm trying to include many scattered files ending in multiple file extensions (e.g. `.txt`, `.ini`, `.json`).
See this thread for context: https://forum.restic.net/t/file-filtering-problems-why-some-patterns-don-t-match-any-files-on-linux/10710

⬇️This new feature would solve at least this problem: Without the ** wildcard, I have to do something like this in my `--files-from` file ⬇️
```
/home/joseph/.local/share/Steam/*/*.json*
/home/joseph/.local/share/Steam/*/*/*.json*
/home/joseph/.local/share/Steam/*/*/*/*.json*
```

⬇️If the `**` wildcard functions correctly, then the following inclusion should find every file with `.json` in its name, no matter if the files are in `/home/joseph/.local/share/Steam/` or in one of the subdirectories within `/home/joseph/.local/share/Steam/` ⬇️

**`/home/joseph/.local/share/Steam/**/.json*`**

It helped me back up a number of files to Google Drive and to my HDD, and that's how Restic (and Rclone) made me happy. Restic is now my favorite backup tool. To those who made Restic and Rclone, good job creating these programs 🙂

omar-y-abdi:
Fixes #5729

Problem: `--files-from` uses Go's `filepath.Glob()` which does not support `**`
wildcards. Users wanting to include files recursively (e.g., `/home/user/**/*.json`)
must list each subdirectory level manually. Meanwhile, `--exclude` already supports
`**` through restic's own `internal/filter.Match()`.

Solution: Add a `Glob()` function to `internal/filter/` that supports `**` wildcards
by walking the filesystem from the longest static prefix and using the existing
`filter.Match()`/`filter.ChildMatch()` functions for matching and directory pruning.
When no `**` is present, delegates to `filepath.Glob()` for full backward
compatibility. Replace `filepath.Glob(line)` with `filter.Glob(line)` in
`collectTargets()`.

Why: Reuses restic's existing `**` pattern matching infrastructure (already
battle-tested in the exclude system) rather than writing new matching logic.
`staticPrefix()` optimizes by starting the walk from the deepest non-wildcard
directory. `ChildMatch()` prunes directories that cannot match, avoiding
unnecessary traversal.

Test:
- `go build ./...` passes → works
- `go test ./internal/filter/...` — 7 new test cases, all pass → works
- Manual e2e: `restic backup --files-from` with `**/*.txt` pattern backed up
  3 .txt files recursively across nested directories → works
- Without `**`, delegates to `filepath.Glob` (backward compatible) → works
- Non-matching `**` pattern returns empty result → works

Closes #5729

### DIFF
--- a/changelog/unreleased/issue-5729
+++ b/changelog/unreleased/issue-5729
@@ -1,0 +1,8 @@
+Enhancement: Support `**` wildcard in `--files-from` patterns
+
+The `--files-from` option now supports the `**` wildcard pattern for
+matching files across directory boundaries. For example,
+`/home/user/**/*.json` will match all `.json` files in any
+subdirectory. Previously, only single `*` wildcards were supported.
+
+https://github.com/restic/restic/issues/5729
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -403,7 +402,7 @@ func collectTargets(opts BackupOptions, args []string, warnf func(msg string, ar
 			}
 
 			var expanded []string
-			expanded, err := filepath.Glob(line)
+			expanded, err := filter.Glob(line)
 			if err != nil {
 				return nil, fmt.Errorf("pattern: %s: %w", line, err)
 			}
--- a/internal/filter/glob.go
+++ b/internal/filter/glob.go
@@ -1,0 +1,85 @@
+package filter
+
+import (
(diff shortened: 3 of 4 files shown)


omar-y-abdi:
Any reviewer that can look at this PR? 

Possibly merge it if it's okay?

PR #4058, opened 2022-11-29
Reconnect sftp connection

ibash:
Draft PR for https://github.com/restic/restic/issues/353

TODO:
- [ ] Cleanup
- [ ] Thread safety
- [ ] More testing

The retry mechanism indicates that 10 retries should be done. But it seems like more than 10 reconnections are done, I'd like to find out why.

--

What does this PR change? What problem does it solve?
-----------------------------------------------------

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

Checklist
---------

- [ ] I have read the [contribution guidelines](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches).
- [ ] I have [enabled maintainer edits](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added tests for all code changes.
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [ ] I have run `gofmt` on the code in all commits.
- [ ] All commit messages are formatted in the same style as [the other commits in the repo](https://github.com/restic/restic/blob/master/CONTRIBUTING.md#git-commits).
- [ ] I'm done! This pull request is ready for review.

PR #5752, opened 2026-03-19
backend/sftp: add optional reconnect on transient SSH disconnects

### LINKED ISSUE #353: restic perfomance drastically decreases on sftp connection issues
I'm trying to back up data to Hetzner's backup space (http://wiki.hetzner.de/index.php/Backup/en). It starts nicely with ~15-20MiB/sec, but as soon as I'm receiving a `Write failed: Broken pipes` or `Connection to uXXX.your-backup.de closed by remote host.` from Hetzner's backup space then restic is permanently decreasing its performance to <1MiB/s and doesn't seem to be able to recover from it. While this clearly seems to be a problem from Hetzner's backend storage I can't find a ssh configuration (TCPKeepAlive, ServerAliveInterval,...) which would work around this problem. It would be nice if restic could recover from this situation itself.

Thanks for restic! :)

lfrancke:

Adds optional automatic reconnect support for the SFTP backend via `-o sftp.reconnect=N`, where N is the maximum number of reconnect attempts allowed per session (default 0, disabled).

I was facing the same issues as discussed in #353, also with Hetzner. I reached out to their support and they looked in their logs and they see a whole bunch of this:

```
2026-02-26T08:12:13.533402+00:00 [info]: Accepted publickey for <redacted> from
<redacted> port 35188 ssh2: ED25519
SHA256:<redacted>
2026-02-26T08:24:19.269730+00:00 [info]: Timeout, client not responding from
user <redacted> <redacted> port 35188
```

This meant that my initial backup never succeeded (I've now waited for over a month of daily backups).
I've never had this issue before with B2 or anywhere else so....not sure what the root cause is.

On the restic side I just saw that it disconnects and never reconnects, causing the entire backup to fail. With this option enabled, restic detects transient disconnects, reconnects the SSH session, and retries the interrupted operation.

_Full disclosure / AI usage_: While I'm an experienced software developer (or at least I like to think so :) ) I'm not a Go person. I reviewed everything manually to the best of my abilities but in the end I had help by Claude.

I tried to incorporate the early feedback from another older PR:

- Opt-in: Reconnect is disabled by default
- Concurrency concerns: Concurrent uploads that all hit a disconnect trigger exactly one reconnect
- Exponential backoff with jitter
- The reconnect counter resets after 5 minutes of stable operation. My initial backup ran for hours and without this reset it'd still never finish

Save, Load, Stat, and Remove retry once after reconnect. List and Delete trigger a reconnect on transient failure but do not retry internally (List to avoid delivering duplicate entries, Delete because recursive deletion is not idempotent). The outer retry backend handles the next attempt with a fresh connection (I believe).

Yes, it (hopefully) closes #353

This supersedes the draft in #4058, addressing the feedback from that PR (see above)

Checklist
---------       

### DIFF
--- a/internal/backend/sftp/reconnect.go
+++ b/internal/backend/sftp/reconnect.go
@@ -1,0 +1,286 @@
+package sftp
+
+import (
+	"io"
+	"math/rand"
+	"os"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/restic/restic/internal/debug"
+	"github.com/restic/restic/internal/errors"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/pkg/sftp"
+)
+
+// connState holds the per-connection state for an SFTP backend.
+// Each reconnect creates a new connState and tears down the old one.
+type connState struct {
+	client      *sftp.Client
+	cmd         *os.Process
+	result      <-chan error
+	posixRename bool
+	done        chan struct{} // closed on teardown to stop goroutines
+	closeOnce   sync.Once
+}
+
+// close tears down the connection. Safe to call multiple times.
+func (cs *connState) close() {
+	cs.closeOnce.Do(func() {
+		_ = cs.client.Close()
+
+		// Wait for the SSH process to exit. If it doesn't exit within
+		// closeTimeout, kill it and wait for the result.
+		select {
+		case <-cs.result:
+		case <-time.After(closeTimeout):
+			_ = cs.cmd.Kill()
+			<-cs.result
+		}
+
+		// Now that cmd.Wait() has completed, signal auxiliary goroutines
+		// (stderr reader) to stop.
+		close(cs.done)
+	})
+}
+
+// reconnectBudget tracks global reconnect attempts per backend instance.
+type reconnectBudget struct {
+	mu             sync.Mutex
+	attempts       uint
+	lastReconnect  time.Time
+	firstReconnect time.Time
+	maxElapsedTime time.Duration
+	stabilityReset time.Duration
+}
+
+const (
+	defaultMaxElapsedTime = 15 * time.Minute
+	defaultStabilityReset = 5 * time.Minute
+
+	reconnectInitialBackoff = 250 * time.Millisecond
+	reconnectMaxBackoff     = 15 * time.Second
+	reconnectBackoffFactor  = 2.0
+)
+
+var errBackendClosed = backoff.Permanent(errors.New("sftp: backend is closed"))
+
+func newReconnectBudget() reconnectBudget {
+	return reconnectBudget{
+		maxElapsedTime: defaultMaxElapsedTime,
+		stabilityReset: defaultStabilityReset,
+	}
+}
+
+// tryConsume checks if a reconnect attempt is allowed.
+func (b *reconnectBudget) tryConsume(maxAttempts uint) (bool, uint) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	now := time.Now()
+
+	// Reset counter if connection has been stable for stabilityReset.
+	if !b.lastReconnect.IsZero() && now.Sub(b.lastReconnect) > b.stabilityReset {
+		b.attempts = 0
+		b.firstReconnect = time.Time{}
+	}
+
+	if b.attempts >= maxAttempts {
+		return false, 0
+	}
+	if !b.firstReconnect.IsZero() && now.Sub(b.firstReconnect) > b.maxElapsedTime {
+		return false, 0
+	}
+
+	if b.firstReconnect.IsZero() {
+		b.firstReconnect = now
+	}
+	b.attempts++
+	b.lastReconnect = now
+	return true, b.attempts
+}
+
+// reconnectBackoff returns the backoff duration for the given attempt number
+// (1-based), using exponential backoff with full jitter.
+func reconnectBackoff(attempt uint) time.Duration {
+	dur := reconnectInitialBackoff
+	for i := uint(1); i < attempt; i++ {
+		dur = time.Duration(float64(dur) * reconnectBackoffFactor)
+		if dur > reconnectMaxBackoff {
+			dur = reconnectMaxBackoff
+			break
+		}
+	}
+	return time.Duration(rand.Int63n(int64(dur) + 1))
+}
+
+// conn returns the current connection state under lock.
+func (r *SFTP) conn() *connState {
+	r.connMu.Lock()
+	defer r.connMu.Unlock()
+	return r.cs
+}
+
+// ensureConnected reconnects after a transient disconnect. It uses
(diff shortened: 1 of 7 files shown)


internal/backend/sftp/sftp.go
lfrancke, line 1: The changes in this file look bigger than they are. Most are "mechanical": the old code accessed the SFTP client directly via `r.c`. The new code goes through a `connState` struct (`cs`) that wraps the client and SSH process together. This indirection is needed so that reconnect can atomically swap in a new connection. The actual retry/reconnect logic lives in `reconnect.go`.

internal/backend/sftp/reconnect.go
lfrancke, line 1: This is the core of the PR. It has three parts:
1. `connState` (lines 19-48): Holds the per-connection state (SFTP client, SSH process handle, posix-rename capability). The `close()` method tears down the connection with a timeout, killing the SSH process if it doesn't exit cleanly.
2. `reconnectBudget` (lines 50-118): Rate-limits reconnect attempts. Tracks a count against the user-configured max, plus a wall-clock elapsed time limit (15 min). The counter resets after 5 minutes of stable operation  so long-running backups don't exhaust the budget from occasional blips. reconnectBackoff provides exponential backoff with full jitter.
3. `ensureConnected` / `withRetry` (lines 130-286): `ensureConnected` uses `singleflight` so that N concurrent goroutines hitting a disconnect produce exactly one reconnect. A generation counter (`connGen`) prevents stale reconnects. `withRetry` is the main wrapper used by Save, Load, Stat, Remove. It runs the operation, and on transient disconnect, reconnects once and retries. `isTransientDisconnect` classifies errors conservatively (broken pipe, connection reset, SSH exit 255, SFTP connection-lost codes, never permission denied, not found, etc.).

lfrancke (commented):
Just leaving two comments as a kind of "reviewers guide" to the PR

YtvwlD:
As a user, I'm quite conflicted about this PR.

On one side, I've been waiting for this feature for quite some time, but it's become less important to me with faster CPUs and network connections.

On the other side, I don't want LLM-generated code in my backup program, because I want it to work correctly. (Well yes, this feature is opt-in and it just concerns the connection handling and not the actual data.)

lfrancke:
Well.... I will not be able to remove your generic doubts about LLMs I'm afraid, neither do I want to. I can just ask you to judge the code on its merits and not on the author. If you have any concerns about tests, code, design/architecture let me know and I can think about it. I can't really argue against a blanket "no".

On your other comment

> On one side, I've been waiting for this feature for quite some time, but it's become less important to me with faster CPUs and network connections.

I'm on a very fast connection with a very recent CPU and this problem is not tied to either of those factors for me. Your error-case might be different of course.

I've been using a custom version of restic with this for a few weeks now and it works for me. I was finally able to get my initial backup of a few hundred GBs working.

Edit: I did check the repo and other issues before sending this PR. I did not find a LLM policy.

justjanne:
> Well.... I will not be able to remove your generic doubts about LLMs I'm afraid, neither do I want to. I can just ask you to judge the code on its merits and not on the author. If you have any concerns about tests, code, design/architecture let me know and I can think about it. I can't really argue against a blanket "no".

You admitted yourself that you don't fully understand the code, yet the PR makes multiple changes in critical code in one go.

While I'm "just" a random restic user (and developer of some unrelated go projects), I don't think I'd be comfortable with that.

In some places your code handles the retry and reconnection logic deep inside the SFTP backend, in other places your code passes this task entirely to the Retry backend. This means different calls will retry a different amount of times, in different ways, with different backoffs.

I don't believe most of this complexity is necessary. Have you tried whether it might be enough to

1. change startClient to do nothing if the current client is already started
2. disconnect the existing client on transient errors
3. call startClient at the beginning of each public method the backend exposes

That'd let the external Retry backend handle most of the actual retries (with a per-file limit).
Then you could, in a separate commit, introduce a global reconnect limiter:

4. change startClient to track the timestamp of the last N reconnects, and reject permanently if the number of reconnects / time interval exceeds the configured limit

This would IMO keep the changes to the rest of the code small enough that potential mistakes the LLM might make would be easier to spot and review.

MichaelEischer (changes requested):
This PR is rather overcomplicated. If this PR should have any chance of getting merged, then all retries MUST be removed from sftp.go. Nearly every change I've seen in there duplicates functionality that is already part of the retry backend. (Nested retry mechanisms are generally considered bad architecture).

Regarding `reconnect.go`: it only should track the number of consecutive connection failures. Resetting it after a while if fine. But the jitter stuff etc. doesn't belong in here.

`isTransientDisconnect` is an absolute mess. I seriously doubt that even half of the errors checked there can occur and are relevant.

Also check https://github.com/restic/restic/pull/4058 again whether there's a simpler approach.

While an LLM can be useful for assistance, you're still expected to understand the generated code. "I don't fully understand the AI-generated code" is both a problem for the PR review process and a sign that the code has maintainability / understand-ability issues. Depending on whether you think you're up for the challenge or not, we can also abandon the PR.

(I've only assigned myself to better track PRs where I already had a look at)

internal/backend/sftp/sftp.go
```diff
@@ -458,48 +543,60 @@ func (r *SFTP) openReader(_ context.Context, h backend.Handle, length int, offse
...
 
 // List runs fn for each file in the backend which has the type t. When an
 // error occurs (or fn returns an error), List stops and returns it.
+//
+// List does not retry internally (restarting would deliver duplicates to fn).
+// On transient disconnect it triggers a reconnect so that the outer retry
+// backend's next attempt gets a fresh connection.
 func (r *SFTP) List(ctx context.Context, t backend.FileType, fn func(backend.FileInfo) error) error {
-	if err := r.clientError(); err != nil {
+	cs, gen, err := r.getConn()
```
MichaelEischer, line 572: `getConn` should just transparently ensure that it returns a client that appears to be working and has not active clientError.

internal/backend/sftp/sftp.go
```diff
@@ -241,33 +277,44 @@ func buildSSHCommand(cfg Config) (cmd string, args []string, err error) {
...
 	// repurpose existing connection
-	return open(sftp, cfg)
+	return openBackend(be, cfg)
 }
 
 func (r *SFTP) Properties() backend.Properties {
+	cs := r.conn()
+	posixRename := false
+	if cs != nil {
+		posixRename = cs.posixRename
```
MichaelEischer, line 313: `posixRename` should just remain part of `SFTP`. That makes it much simpler to handle.

lfrancke:
Thanks for taking a look. And sorry for not reporting back here. I actually took the [comment](https://github.com/restic/restic/pull/5752#issuecomment-4090555665) to heart and have been running a version of restic with this simplified approach since March and just forgot to report back here. I'll try to take a look at the PR you reference and your comment in detail this week.

PR #5764, opened 2026-03-28
Add benchmark with concurrent archivers

WhyNotHugo:

Current benchmarks all work with a single archiver, and are not suitable for testing changes which might affect contention.

Add a benchmark with multiple small files and multiple concurrent archivers.

Nope

Checklist
---------

### DIFF
--- a/internal/archiver/archiver_test.go
+++ b/internal/archiver/archiver_test.go
@@ -376,6 +376,62 @@ func BenchmarkArchiverSaveFileSmall(b *testing.B) {
 	}
 }
 
+func BenchmarkArchiverSaveFileConcurrent(b *testing.B) {
+	const fileSize = 4 * 1024
+	const numFiles = 64
+
+	d := TestDir{}
+	for i := 0; i < numFiles; i++ {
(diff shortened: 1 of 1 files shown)


internal/archiver/archiver_test.go
```diff
@@ -376,6 +376,62 @@ func BenchmarkArchiverSaveFileSmall(b *testing.B) {
...
+			b.Errorf("archiver error for %v: %v", item, err)
+			return err
+		}
+		b.StartTimer()
+
+		err := repo.WithBlobUploader(context.Background(), func(ctx context.Context, uploader restic.BlobSaverWithAsync) error {
+			wg, ctx := errgroup.WithContext(ctx)
+			arch.runWorkers(ctx, wg, uploader)
+
+			var futures []futureNode
```
MichaelEischer, line 407: Why does this not use `archiver.saveDir`? That would already handle all the futureNode processing and collection.

MichaelEischer (changes requested):
Please rebase the PR, the test no longer compiles with the latest changes from master.

PR #5765, opened 2026-03-28
Don't use locks for counting stats

WhyNotHugo:
Counting statistics uses locks each time an archiver needs to update
stats.

Use lockless stats counting, using locks only for measuring network
speeed (this bit is stateful and can't trivially be made lockless).

To be frank, I was curious if this was even feasible, and went out to
implement it mostly to know if it was. Benchmarks show a mere 1.6%
improvement in performance, with no real additional algorithmic
complexity for it.

Nope

Checklist
---------

Note sure if this kind of internal change warrants any changelog entry.

### DIFF
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/restic/restic/internal/data"
@@ -35,12 +36,12 @@ type ErrorFunc func(file string, err error) error
 
 // ItemStats collects some statistics about a particular file or directory.
 type ItemStats struct {
-	DataBlobs      int    // number of new data blobs added for this item
-	DataSize       uint64 // sum of the sizes of all new data blobs
-	DataSizeInRepo uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
-	TreeBlobs      int    // number of new tree blobs added for this item
-	TreeSize       uint64 // sum of the sizes of all new tree blobs
-	TreeSizeInRepo uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
+	DataBlobs      atomic.Int64  // number of new data blobs added for this item
+	DataSize       atomic.Uint64 // sum of the sizes of all new data blobs
+	DataSizeInRepo atomic.Uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
+	TreeBlobs      atomic.Int64  // number of new tree blobs added for this item
+	TreeSize       atomic.Uint64 // sum of the sizes of all new tree blobs
+	TreeSizeInRepo atomic.Uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
 }
 
 type ChangeStats struct {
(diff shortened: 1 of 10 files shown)


WhyNotHugo:
PR includes https://github.com/restic/restic/pull/5764, that one probably wants to be merged first. The increase in line count is mostly that new benchmark.

internal/archiver/archiver.go
```diff
@@ -35,12 +36,12 @@ type ErrorFunc func(file string, err error) error
 
 // ItemStats collects some statistics about a particular file or directory.
 type ItemStats struct {
-	DataBlobs      int    // number of new data blobs added for this item
-	DataSize       uint64 // sum of the sizes of all new data blobs
-	DataSizeInRepo uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
-	TreeBlobs      int    // number of new tree blobs added for this item
-	TreeSize       uint64 // sum of the sizes of all new tree blobs
-	TreeSizeInRepo uint64 // sum of the bytes added to the repo (including compression and crypto overhead)
+	DataBlobs      atomic.Int64  // number of new data blobs added for this item
```
MichaelEischer, line 39: I don't see how using atomics in `ItemStats` helps. The struct doesn't need locking inside the archiver and once it's passed to the ui parts, its values aren't modified.

internal/ui/backup/text.go
```diff
@@ -32,31 +32,31 @@ func NewTextProgress(term ui.Terminal, verbosity uint) *TextProgress {
...
 		status = fmt.Sprintf("[%s] %v files, %s, %d errors",
 			ui.FormatDuration(time.Since(start)),
-			processed.Files, ui.FormatBytes(processed.Bytes), errors,
+			processed.Files.Load(), ui.FormatBytes(processed.Bytes.Load()), errors,
 		)
 	} else {
 		var eta, percent string
 
-		if secs > 0 && processed.Bytes < total.Bytes {
+		if secs > 0 && processed.Bytes.Load() < total.Bytes.Load() {
```
MichaelEischer, line 46: Fields should only be loaded once to avoid inconsistencies. Otherwise, two `processed.Bytes.Load()` calls can yield different values.

internal/ui/backup/progress.go
```diff
@@ -56,36 +57,34 @@ func NewProgress(printer ProgressPrinter, interval time.Duration) *Progress {
...
 					secondsRemaining = 0
 				} else {
-					todo := float64(p.total.Bytes - p.processed.Bytes)
+					todo := float64(p.total.Bytes.Load() - p.processed.Bytes.Load())
 					secondsRemaining = uint64(todo / rate)
 				}
 			}
 
-			p.printer.Update(p.total, p.processed, p.errors, p.currentFiles, p.start, secondsRemaining)
+			p.mu.Lock()
```
MichaelEischer, line 76: Why is this call synchronized, but e.g. `printer.ReportTotal` isn't?

MichaelEischer (changes requested):
I see at least two other reasons other than locking which this could be faster:
- ItemStats is passed by pointer not value
- `addProcessed` is inlined

Please make the changes in separate steps and use https://pkg.go.dev/golang.org/x/perf/cmd/benchstat for a solid comparison of the benchmark results. (What benchmark did you use?)

You'll have some convincing to do that the change is worth it; having `.Load()` and possible inconsistencies all over the place looks somewhat weird to me.

PR #21771, opened 2026-04-01
Enhancement: Allow rewriting snapshot paths with `--path-from` and `--path-to`

nuz014:

This PR adds `--path-from` and `--path-to` flags to `restic rewrite`, allowing
users to change the paths stored in snapshot metadata.

- `--path-to /new/path` replaces all paths in the snapshot
- `--path-from /old/path --path-to /new/path` replaces only the matching path
- `--path-from` without `--path-to` produces an error

This is useful when a backup source has been moved or renamed and the user
wants the snapshot metadata to reflect the new location without re-creating
the backup.

Example usage:

restic rewrite --path-to /new/path --forget latest
restic rewrite --path-from /old/path --path-to /new/path --forget ec2d5d46
restic rewrite --path-to /new/path --dry-run latest

Closes #5723

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5723
+++ b/changelog/unreleased/issue-5723
@@ -1,0 +1,8 @@
+Enhancement: Allow rewriting snapshot paths with `rewrite --path-to`
+
+The `rewrite` command now supports `--path-from` and `--path-to` options
+to replace paths stored in a snapshot. When both options are used, only
+the matching path is replaced. When only `--path-to` is specified, all
+paths are replaced with the new path.
+
+https://github.com/restic/restic/issues/5723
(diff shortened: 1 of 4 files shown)


MilitaoLucas:
I don't know if this exactly solves #5723 completely. In metadata it is changed, but in the chunks itself, the absolute path is still the older one. There doesn't seem to be a way to change the chunks path implemented in restic... 
If you try to do this and browse the files in the backup, the root differs in newer snapshots, so you can't have a totally incremental view...

itzonban:
Does this also support the removal of a path, like `--path-to ''`? 

I just ran into the following issue, that I removed a path from all existing snapshots but it is still shown in the snapshot list

```
/path1
/path2
/path3
```

I removed `/path3` from all snapshots by using

`rewrite --exclude /path3`.

The path is gone from the Backup Data but is still listed at the snapshot paths.

PR #21781, opened 2026-04-16
Fix --host and --user flags in the key passwd command

ghost:
Fixes the --host and --user flags in the key passwd command. The AddKey function call is missing the opts.Username and opts.Hostname variables.

No

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5757
+++ b/changelog/unreleased/issue-5757
@@ -1,0 +1,8 @@
+Bugfix: respect `--user` and `--host` in `key passwd`
+
+The `key passwd` command silently ignored the `--user` and `--host` flags
+and always recorded the new key with the current user and host name.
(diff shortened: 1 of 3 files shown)


MichaelEischer:
Thanks, can you add a basic tests to make sure this won't break again? The approach would be some combination of what's done in `testRunKeyAddNewKeyUserHost` and `testRunKeyPasswd`. (I can also take over if you want).

And please add a minimal changelog entry using the template from https://github.com/restic/restic/blob/master/changelog/TEMPLATE .

ghost:
Happy for you to take over if you're happy to. I've never written a go test before.

MichaelEischer (approved):
LGTM

MichaelEischer (approved):

Issue #4382, opened 2023-06-22
Restic Mount of Windows backup repository shows folders but no files under SAMBA


durcom2:
Output of `restic version`
--------------------------

How did you run restic exactly?
-------------------------------
restic backup --use-fs-snapshot --no-scan --quiet --iexclude-file restic.exclude -p r1pw.txt --repo rest:http://restuser:restpw8765@ptdbackup.dynns.com:8000/personal/ed/ c:\

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
rest server

Expected behavior
-----------------
ability to see files via samba

show the files
Actual behavior
---------------
folders shown, but no files

Steps to reproduce the behavior
-------------------------------
Run restic backup from windows client computer.  Mount the repo with restic mount on an ubuntu 22 box, and share it via SAMBA
try to access the files in the backup from the windows computer via the SAMBA share

Note that backed up files show up fine on the ubuntu box when accessed on the linux box restic mountpoint.

Also, files backed up with restic from another ubuntu box show up fine when using restic mount in BOTH the windows box AND when accessing them via a SAMBA share, so it seems to be something with the windows version of restic.  Linux version of restic seems to work fine.

Do you have any idea what may have caused this?
-----------------------------------------------

Do you have an idea how to solve the issue?
-------------------------------------------

nope

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

rawtaz:
You might want to change the password that you wrote in the command line example.

MichaelEischer:
@durcom2 What restic version do you use? What is the output of `stat /path/to/some/problematic/file` on Ubuntu? My guess is that the problem here is related to file permission. How exactly did you mount the repository on Ubuntu? Did you pass `--allow-other` to `restic mount`?

durcom2:
/root/restic mount --cache-dir /mnt/other --allow-other --no-lock --repo /mnt/backups/restic/personal/ed --time-template "2006 Jan 02 03-04 PM" --password-file /root/resticpw.txt /mnt/restic_mnt

permissions on mounted repo are 777 for folders and 644 for files which means they should be visible...

JsBergbau:
I tried some time ago the same and also didn't manage to share the fuse mounted files via samba. So I gave up on this.

jkirk:
Seems to be related to #2034.

@durcom2 The mount point of `restic mount` is `/mnt/restic_mnt`, right? How does your Samba share configuration look like?

Issue #2034, opened 2018-10-10
Windows client snapshot -> FUSE mount -> Samba share -> cant copy files


dmi-github:
I have found a similar issue which is already closed: https://github.com/restic/restic/issues/988
However, I am not quite sure if this issue has been pinned down correctly. It seems to me, that it requires a particular scenario to be reproducible.

Output of `restic version`
--------------------------
Servers - responsible for providing repository with rest-serv, for fuse mount and samba share, 
Tested with two different servers:
* Server A:
  * restic 0.9.2 compiled with go1.10.1 on linux/amd64
  * rest-server 0.9.7 compiled with go1.10.1 on linux/amd64
  * Samba Version 4.7.6-Ubuntu
* Server B:
  * restic 0.9.2 compiled with go1.10.3 on linux/arm64
  * rest-server 0.9.7 compiled with go1.10.1 on linux/arm64
  * Samba Version Version 4.5.12-Debian

Clients - responsible for creating snapshots
* Windows: restic 0.9.2 compiled with go1.10.3 on windows/amd64
* Linux:  client version corresponds to server version(creating snapshots directly on the server - Although the repository is available in the local filesystem, I also access it through rest-serv to ensure comparable conditions)

How did you run restic exactly?
-------------------------------
creating snapshots without any fancy options.
mounting the repository with:
* RESTIC_REPOSITORY=/mnt/backup/live/restic_live
* RESTIC_PASSWORD = restic
* restic mount --allow-other --snapshot-template '2006_01_02__15_04'

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
local repository
* mounting: accessed directly
* creating snapshots: accessed through rest-serv

Expected behavior
-----------------
The mounted repository is shared by samba. I expect to be able to browse and copy files out of this samba share.

Actual behavior
---------------
Two different hosts backup to my repository
* windows client
* linux client(the 'server' itself)

I have configured a samba share which includes the mountpoint served by restic's fuse. When I access this share I observe the following behaviour:

* I can browse all snapshot files without problems
* I can copy files from snapshots which have been created by the linux client.
* I can not copy files from snapshots which have been created by the windows client.(This file is no longer located in...)
* For the Windows snapshots, I can see the filesize in folder view in Windows Explorer. However, If I open the 'File Properties' window, it is a 50/50 chance to get the actual file size or to get 0 bytes as a result.
* It does not matter how I access the samba share(e.g. Windows Explorer, Chrome, ...)
[...]

PR #21784, opened 2026-04-19
mount: Ensure a hard link count > 0 for all files

jtru:
When using `restic mount` to serve a repository via a POSIX host's file system, all files backed up from Windows systems will show up on the mounting host with a (hard)link count of 0.

While this is not a problem in general and most programs do not even register this strange value, some others (such as Samba's smbd(8)) will go the extra mile and check a file's stat() results for various properties (such as a positive non-zero link count), and refuse to operate if any of the reported values appear off.

Since other inode properties absent from non-POSIX backup sources are also "faked up" (e.g., the inode number) during mount and work fine in general, the chances of this change to be in any way harmful are probably rather slim.

This makes it possible to serve up a `restic mount`-provided repository via the Samba project's `smbd`. Without it, accessing a regular file backed up from a Windows source host will cause the smb client to croak with the NT status equivalent of ENOENT.

To work around this with this patch, the fuse layer will ensure a minimal link count of 1 for all files it presents to the mounting host.

No, merely on IRC.

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-21784
+++ b/changelog/unreleased/pull-21784
@@ -1,0 +1,9 @@
+Bugfix: Support exporting a `restic mount` of a Windows system via Samba
+
+A repository mounted using `restic mount` on a POSIX system, could not use
+Samba to export files from restic backups of Windows systems. Backups of
+other systems were not affected. This has been fixed.
+
+https://github.com/restic/restic/pull/21784
+https://github.com/restic/restic/issues/2034
+https://github.com/restic/restic/issues/4382
--- a/internal/fuse/file.go
+++ b/internal/fuse/file.go
@@ -55,7 +55,8 @@ func (f *file) Attr(_ context.Context, a *fuse.Attr) error {
 	a.Size = f.node.Size
 	a.Blocks = (f.node.Size + blockSize - 1) / blockSize
 	a.BlockSize = blockSize
-	a.Nlink = uint32(f.node.Links)
+	// present a link count > 0 to keep over-eager stat() .st_nlink checks (e.g., from Samba's smbd) happy
+	a.Nlink = max(uint32(1), uint32(f.node.Links))
 
 	if !f.root.cfg.OwnerIsRoot {
 		a.Uid = f.node.UID
--- a/internal/fuse/file.go
+++ b/internal/fuse/file.go
@@ -55,7 +55,9 @@ func (f *file) Attr(_ context.Context, a *fuse.Attr) error {
 	a.Size = f.node.Size
 	a.Blocks = (f.node.Size + blockSize - 1) / blockSize
 	a.BlockSize = blockSize
-	// present a link count > 0 to keep over-eager stat() .st_nlink checks (e.g., from Samba's smbd) happy
+	// Windows (and other non-POSIX) backups may store a link count of 0.
+	// FUSE must still report a positive nlink so tools that validate stat()
+	// (e.g. Samba) accept the file.
 	a.Nlink = max(uint32(1), uint32(f.node.Links))
 
 	if !f.root.cfg.OwnerIsRoot {
--- a/internal/fuse/fuse_test.go
+++ b/internal/fuse/fuse_test.go
@@ -5,6 +5,7 @@ package fuse
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"math/rand"
 	"os"
 	"strings"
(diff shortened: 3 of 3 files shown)


jtru:
I was asked on IRC to elaborate on the setup that made me find this problem, so here goes.

Below is restic 0.18.1 having mounted a repository at `/tmp/tmpsmb` with the `mount --time-template '2006-01-02T15_04_05Z07_00' --allow-other --no-default-permissions` flags specified on my Linux host, with `stat(1)` displaying file metadata as reported from data backed up from a Windows 10 host's NTFS-formatted C:\ drive:

```
root@foex-test01:/tmp/tmpsmb/snapshots/2026-04-12T20_08_02+02_00/C/Users/Johannes/Downloads/Rufus# ls -l
total 80
-rw-rw-rw- 0 root root 81579 Oct 18  2025 rufus.log
root@foex-test01:/tmp/tmpsmb/snapshots/2026-04-12T20_08_02+02_00/C/Users/Johannes/Downloads/Rufus# stat rufus.log 
  File: rufus.log
  Size: 81579           Blocks: 160        IO Block: 512    regular file
Device: 0,56    Inode: 879487326304452967  Links: 0
Access: (0666/-rw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-10-18 18:20:10.060522300 +0000
Modify: 2025-10-18 18:20:10.060522300 +0000
Change: 2025-10-18 18:20:10.060522300 +0000
 Birth: -
```

The crucial part is the **Links: 0** data right after the (faked up, since Windows does not have those) **Inode:** number.

When sharing the contents of `/tmp/tmpsmb/` via Samba's smbd it (and potentially other software) will claim this file does not exist when trying to serve it to any authorized client, due to the 0-valued link count. The problem manifests like this over `smbclient`:

```
smb: \> cd snapshots\2026-04-12T20_08_02+02_00\C\Users\Johannes\Downloads\Rufus\
smb: \snapshots\2026-04-12T20_08_02+02_00\C\Users\Johannes\Downloads\Rufus\> ls
  .                                   D        0  Sun Sep 14 14:11:44 2025
  ..                                  D        0  Sat Apr 11 14:44:48 2026
  rufus.log                                81579  Sat Oct 18 20:20:10 2025

                0 blocks of size 1024. 0 blocks available
smb: \snapshots\2026-04-12T20_08_02+02_00\C\Users\Johannes\Downloads\Rufus\> get rufus.log 
NT_STATUS_OBJECT_NAME_NOT_FOUND opening remote file \snapshots\2026-04-12T20_08_02+02_00\C\Users\Johannes\Downloads\Rufus\rufus.log
```

With the patch from this PR applied on top of current master and using it to mount the repo, the result from `stat(2)` changes, and the same repository/snapshot looks like this instead:

```
root@foex-test01:/tmp/tmpsmb/snapshots/2026-04-12T20_08_02+02_00/C/Users/Johannes/Downloads/Rufus# ls -l
total 80
-rw-rw-rw- 1 root root 81579 Oct 18  2025 rufus.log
root@foex-test01:/tmp/tmpsmb/snapshots/2026-04-12T20_08_02+02_00/C/Users/Johannes/Downloads/Rufus# stat rufus.log 
  File: rufus.log
  Size: 81579           Blocks: 160        IO Block: 512    regular file
Device: 0,56    Inode: 879487326304452967  Links: 1
Access: (0666/-rw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-10-18 18:20:10.060522300 +0000
Modify: 2025-10-18 18:20:10.060522300 +0000
Change: 2025-10-18 18:20:10.060522300 +0000
 Birth: -
```

This change alone fixes smbd's odd behavior (and potentially other software) that cannot or is unwilling to deal with a link count of 0.

Since a link count of 0 would normally indicate a deleted file (which probably should not be observable on a POSIX system outside of Linux' `/proc/` for fds pointing towards open file that have been unlinked in their containing file system), I think it would be a good idea to change the mimimum link count of files indicated by the fuse mount to be at least 1.

MichaelEischer (approved):
LGTM. Thanks a lot for finding this. After reading through the PR description twice, I'm very sure that this is finally the fix for https://github.com/restic/restic/issues/2034 and https://github.com/restic/restic/issues/4382 .

I've added a basic test (AI coded) to prevent regressing this in the future and a short changelog entry that explains this without going too much into technical details although that's surprisingly tricky here. Please take a look.

internal/fuse/file.go
```diff
@@ -55,7 +55,8 @@ func (f *file) Attr(_ context.Context, a *fuse.Attr) error {
 	a.Size = f.node.Size
 	a.Blocks = (f.node.Size + blockSize - 1) / blockSize
 	a.BlockSize = blockSize
-	a.Nlink = uint32(f.node.Links)
+	// present a link count > 0 to keep over-eager stat() .st_nlink checks (e.g., from Samba's smbd) happy
```
clrossel: I would prefer replacing this comment with the comment from the test code as it more clearly explains the problem.

Test code comment below:

```
// Windows (and other non-POSIX) backups may store a link count of 0.
// FUSE must still report a positive nlink so tools that validate stat()
// (e.g. Samba) accept the file.
```

clrossel (commented):
I would prefer updating the comment in ```‎internal/fuse/file.go``` as noted.

internal/fuse/file.go
jtru: I concur, this is a better explanation providing more required context. Future readers will benefit from the information that Windows (and potentially other OS?) do not store a positive link count for their backed-up files in the repo in the first place :)

changelog/unreleased/pull-21784
```diff
@@ -0,0 +1,9 @@
+Bugfix: Support exporting a `restic mount` of a Windows system via Samba
+
+A repository mounted using `restic mount` on a POSIX system, could not use
+Samba to export restic backup of Windows systems. Backups of other systems
```
jtru: I think "could not use  Samba to export **files** from restic backup**s** of Windows systems" would be more appropriate, since (at least in my pretty extensive testing), I could browse the mounted snapshots (i.e., readdir() works fine and shows all files with metadata) and traverse the directory tree, but actually opening/reading file data failed with a confusing SMB/NT status message.

MichaelEischer: Thanks!

jtru:
Changes (including the test) look great - I'd be very happy if this fix were merged this way :)

PR #21788, opened 2026-05-01
Add macOS backup exclusion support

### LINKED ISSUE #2180: Feature request: Support macOS Backup Core excludes
Output of `restic version`
--------------------------
restic 0.9.4 compiled with go1.11.4 on darwin/amd64

Check files to see if they’ve been excluded from inclusion in user backups in Backup Core (Time Machine). There is an API at [CSBackupIsItemExcluded](https://developer.apple.com/documentation/coreservices/1443602-csbackupisitemexcluded).

macOS app developers can use Backup Core to exclude files from backups. Usually non-essential files, caches, and the like. Technically this sets an extended file system attribute called `com.apple.metadata:com_apple_backup_excludeItem`.

Users can manually mark files and directories from exclusion from the Time Machine UI. This doesn’t set an extended attribute on the file. However, the above API (or `tmutil isexcluded`) will correctly return a file’s backup inclusion status.

What are you trying to do?
--------------------------
Avoid backing up ephemeral files by relying on app developers who know their apps’ internal well and know what should be excluded from backups.

Deduplicated backups is enough to make anyone happy.

elohmeier:
## What

Adds a macOS-only backup option, `--exclude-macos-backup`, to exclude files and directories marked by macOS as excluded from backups via Backup Core.

This uses a Darwin-only purego bridge to `CSBackupIsItemExcluded`, preserving restic's `CGO_ENABLED=0` release build model.

Fixes #2180.

## Testing

- `CGO_ENABLED=0 go test ./internal/archiver`
- `CGO_ENABLED=0 go test ./cmd/restic -run 'TestBackupExclude|TestBackupWithRelativePath|TestCollectTargets|TestReadFilenamesRaw'`
- `GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go build ./cmd/restic`
- `GOOS=darwin GOARCH=amd64 CGO_ENABLED=0 go build ./cmd/restic`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build ./cmd/restic`
- `GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build ./cmd/restic`

### DIFF
--- a/changelog/unreleased/issue-2180
+++ b/changelog/unreleased/issue-2180
@@ -1,0 +1,8 @@
+Enhancement: Support excluding files marked by macOS as excluded from backups
+
+The `backup` command now supports `--exclude-macos-backup` on macOS. When set,
+restic excludes files and directories marked by macOS as excluded from system
+backups via Backup Core, such as application cache data that Time Machine would
+skip.
+
+https://github.com/restic/restic/issues/2180
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -77,31 +77,32 @@ Exit status is 12 if the password is incorrect.
 type BackupOptions struct {
 	filter.ExcludePatternOptions
 
-	Parent            string
-	GroupBy           data.SnapshotGroupByOptions
-	Force             bool
-	ExcludeOtherFS    bool
-	ExcludeIfPresent  []string
-	ExcludeCaches     bool
-	ExcludeLargerThan string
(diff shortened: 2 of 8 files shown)


awolman:
I find the option naming "--exclude-macos-backup" a little bit counter intuitive.

How about "--use-macos-backupcore-exclusion-list" instead?  Its wordy, but clear.

kasper9n:
I would expect this to be part of the existing `--exclude-caches` flag

PR #21795, opened 2026-05-08
Enhancement: add version-specific docs URLs to help output and improve test error handling

Turmaxx:
This PR adds the `docs` command to `restic` as well as a `user` and `dev` sub-command

Previous to this addition, users had to either copy and paste the documentation URL link to their browser or type it themselves.

now with a simple `restic docs` , they can open the documentation from the terminal onto the browser on Linux, Macos & Windows. **_(tested on linux since I dont have a macos or windows laptop)_**, 

But I have added tests in `cmd_docs_test.go`

I don't think so.

Checklist
---------


Turmaxx:
Is there a place where subcommands are documented?

MichaelEischer:
https://restic.readthedocs.io/en/stable/manual_rest.html#usage-help contains the output of `restic --help`.

I'm not particularly sure what benefit the command provides compare to the link in `restic --help` which you can ctrl+click on most terminals. There's also the downside that this command doesn't work on all platforms supported by restic. What I could imagine as useful is a command that automatically points users to the correct version of the documentation for their restic version. That is, `v0.18.0` should point to those docs, but a developer build should point to `latest`.

Turmaxx:
Thanks for taking a look at the PR, @MichaelEischer.

I see your point that this overlaps with the existing help output and the documentation link there. My intention with this change was mainly to add a small quality-of-life improvement for quickly accessing the docs, rather than introducing new functionality.

That said, I understand the concern about added surface area and maintenance complexity, especially if it doesn’t provide meaningful benefit beyond what’s already available. If that outweighs the value here, I’m happy to close or rework the PR so that it instead points to the documentation matching the specific binary version the user is running.

Turmaxx:
@MichaelEischer 

I've made a couple of changes as suggested.

So when the user is running a particular vesrion of restic and invokes the `restic docs` command it sends them to that [specific version documenation](https://restic.readthedocs.io/en/v0.18.1/).

Using the `restic docs user` and `restic docs dev` sends them to [stable](https://restic.readthedocs.io/en/stable/) and [latest](https://restic.readthedocs.io/en/latest) respectively.

And the final thing I've add is that when you are using a manually compiled version of restic,  `restic docs` sends them to the [latest](https://restic.readthedocs.io/en/latest) version of the documentation since its a bleeding edge version.

rawtaz:
I'd like to add my thoughts on this topic. First, thanks for thinking about this. It's nice of you to suggest improvements for users trying to get help!

I don't think that adding one new command and a couple of subcommands is the way to go here. It's simply too much added surface, when the fact is that both `restic`, `restic help` and `restic --help/-h` already provide the URL to the documentation as almost the very first thing they do in their output.

The `help` and `--help/-h` options are arguably the most established thing you can find in regular CLI software. If a user wants documentation for the CLI, they run `--help/-h` or as is usual in Go and similar binaries, `help`. I think this is so established that there's no need for an additional command or two.

That said, the suggestion of making the URL shown be version specific (i.e. for `vX.Y.Z` or for `latest`) isn't a bad idea! I'd like that. That's what I'd take away from this PR :)

MichaelEischer (changes requested):
After looking at this again, I have to agree with @rawtaz . The benefit / overhead ratio isn't good enough to justify an extra `docs` command let alone multiple subcommands.

Let's limit this PR to automatically adjust the docs link in `cmd/restic/main.go` to point to the release-specific docs for a released version and the `latest` docs version in all other case.

Turmaxx:
Hey @MichaelEischer @rawtaz,

Thanks for the feedback guys. I've updated the PR and limited the scope accordingly.

The extra docs command and subcommands have been removed. The PR now only updates the docs link in `cmd/restic/main.go` to point to the release-specific documentation for released versions and to the latest documentation in all other cases as suggested.

Turmaxx:
Also while we are at this PR. I noticed that my IDE had a warning in `cmd/restic/cmd_list_integration_test.go`, an error wasn't being checked, so I added a check after the scan loop. This makes sure any scanning or read errors are surfaced and cause the test to fail instead of being silently ignored.

Turmaxx:
@MichaelEischer What is the state of this PR?

PR #21798, opened 2026-05-11
windows: try to fix flaky tests

MichaelEischer:
The unit tests on Windows fail from time to time with error:
```
2026-05-10T19:22:59.8628469Z         testing.go:123: [31munexpected error: Access is denied.
2026-05-10T19:22:59.8628846Z             github.com/restic/restic/internal/repository.(*packerManager).newPacker
2026-05-10T19:22:59.8629196Z             	D:/a/restic/restic/internal/repository/packer_manager.go:206
```

The line in `packer_manager.go` is the error log directly after the call to `fs.TempFile`. I see two main culprits here: the used RNG could generate the same filename if called in roughly the same microsecond. As this happens a few thousand times during the TestEmptyFileAttributeCombinationsOverwrite and related test cases, the chances of a collision aren't particularly low.

In addition, in case of a name collision, the retry would not work. The temp file is created without sharing and uses `FILE_FLAG_DELETE_ON_CLOSE` which prevents opening new handles to that file. However, this triggers an access denied error. Thus, ignore that particular error code.

No. But see the PR build history for examples of the test failures.

Checklist
---------

### DIFF
--- a/internal/fs/file_windows.go
+++ b/internal/fs/file_windows.go
@@ -76,6 +76,10 @@ func TempFile(dir, prefix string) (f *os.File, err error) {
 		if os.IsExist(err) {
 			continue
 		}
+		// Access denied error can occur if the tmp files conflict with each other.
+		if isAccessDeniedError(err) {
+			continue
+		}
 		return os.NewFile(uintptr(h), path), err
 	}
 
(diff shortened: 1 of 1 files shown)


MichaelEischer (commented):
LGTM

PR #21800, opened 2026-05-12
Enhancement: Add WebDAV backend

### LINKED ISSUE #277: Storage Backend: WebDAV
Please add the possibility to add to a Webdav server

mcrute:
This change adds a WebDAV backend that is similar to the REST backend but is compatible with standard WebDAV servers. For the most part this means replacing the custom JSON indexing expected from the remote server with support for calling `PROPFIND` on the remote server and parsing its output. Tolerance of the on-spec HTTP status codes was added too (the REST backend supports only `200` but WebDAV uses several other `2xx` codes based on method). This backend also uses `PUT` instead of `POST` to create files because that is the required method in WebDAV. Otherwise this is very structurally similar to the REST backend.

This backend can act as a simpler alternative to using rclone which is rather clunky and has a bit of configuration overhead for a protocol as simple as WebDAV.

I tested this against Caddy V2 with the WebDAV plugin that uses golang.org/x/net/webdav as a backend but this should work against any minimally spec compliant server that supports `MKCOL` (only needed for new repo creation), and `PROPFIND`.

Closes #277 (well I guess technically this solves the issue, it's already closed) 

Checklist
---------
- [X] I have added tests for all code changes, see [writing tests](https://restic.readthedocs.io/en/stable/090_participating.html#writing-tests)

### DIFF
--- a/changelog/unreleased/pull-21800
+++ b/changelog/unreleased/pull-21800
@@ -1,0 +1,7 @@
+Enhancement: Add WebDAV backend
+
+Adds a WebDAV backend that is similar to the REST backend but supports
+RFC 4918 compatible WebDAV servers. On-disk repository layout is the
+same as the local backend. Uses the `webdav:` prefix for URLs.
+
+https://github.com/restic/restic/pull/21800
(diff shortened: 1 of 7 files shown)


mcrute:
@MichaelEischer is this something you'd be willing to review and merge? WebDAV support would be super helpful in my backup setup. I'm happy to make any changes if any are needed to get this ready to merge. Thanks!

PR #21802, opened 2026-05-14
docs: clarify backup step in troubleshooting guide

### LINKED ISSUE #21799: Incomplete documentation on troubleshooting
[Actual version of documentation](https://restic.readthedocs.io/en/latest/077_troubleshooting.html#remove-missing-data-from-snapshots) leaves a couple of unanswered questions. I think it's much better to answer them right in the documentation rather than leaving them buried in the depths of internet forums, so I will ask them in this issue

So, let's take this piece of information:

> With a correct index, the backup command guarantees that newly created snapshots can be restored successfully. It can also heal older snapshots, if the missing data is also contained in the new snapshot

> Therefore, it is recommended to run all your backup tasks again. In some cases, this is enough to fully repair the repository

So... How can I actually know if some data is still inaccessible? **Is there a command that can just list these packs**? Or should I repeat the whole long `restic check --read-data=true`? Is there a faster way than the latter?

> If your repository is still missing data, then you can use the repair snapshots command to remove all inaccessible data from the snapshots

Let's assume I know there is still some missing data, because I got an answer to the previous question. Then...

Deleting something is the last thing a user should try — so it should be the last thing recommended by the documentation

Clearly there is a more safe and important step you can do first. It's finding an undamaged version of corrupted files and repeating the backup command again. But in order to do this, you need to somehow know exactly which files in exactly which snapshots are damaged. So the question is... **how can a user identify the file names (and their snapshots) associated with these still-damaged-packs?**

Yes, you can get them in the output of `repair snapshots`. But it's still a destructive operation — _even without `--forget`_, you create a mess in the repository, making the problem harder to analyze

MichaelEischer:
Add further explanations how to check whether a repository is still damaged after the running the backup step of the troubleshooting guide and how to find out which files and snapshots are damaged.

Fixes https://github.com/restic/restic/issues/21799

Checklist
---------

### DIFF
--- a/doc/077_troubleshooting.rst
+++ b/doc/077_troubleshooting.rst
@@ -157,6 +157,11 @@ if the missing data is also contained in the new snapshot.
 Therefore, it is recommended to run all your ``backup`` tasks again. In some
 cases, this is enough to fully repair the repository.
 
+To check if the repository is fully repaired, you can run ``restic check``
+(without options) again.
+
+To get a list of still damaged files, you can run ``restic repair snapshots --dry-run``.
+Look for ``would save new snapshot`` messages to find affected snapshots.
 
 5. Remove missing data from snapshots
 *************************************


MichaelEischer (commented):
LGTM.

PR #21803, opened 2026-05-14
prune more aggresively

### LINKED ISSUE #5293: prune: improve automatic handling of small pack files

wplapper:
prune more aggressively. This PR is the continuation of PR#5359.

issue https://github.com/restic/restic/issues/5293.

close https://github.com/restic/restic/issues/5293

Checklist
---------


wplapper:
This is the continuation of PR#5359. Since I messed up git somehow, I could not update PR#5359 anymore. Therefore a new PR.

cmd/restic/cmd_prune.go
```diff
@@ -78,12 +77,19 @@ func (opts *PruneOptions) AddFlags(f *pflag.FlagSet) {
...
+	var unused bool
 	f.StringVar(&opts.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&opts.MaxRepackSize, "max-repack-size", "", "stop after repacking this much data in total (allowed suffixes for `size`: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&opts.RepackCacheableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
-	f.BoolVar(&opts.RepackSmall, "repack-small", false, "repack pack files below 80% of target pack size")
+	f.BoolVar(&unused, "repack-small", false, "repack pack files below 80% of target pack size")
 	f.BoolVar(&opts.RepackUncompressed, "repack-uncompressed", false, "repack all uncompressed data")
 	f.StringVar(&opts.SmallPackSize, "repack-smaller-than", "", "pack `below-limit` packfiles (allowed suffixes: k/K, m/M)")
+
+	err := f.MarkDeprecated("repack-small", "ignored.")
```
MichaelEischer:
```suggestion
	err := f.MarkDeprecated("repack-small", "small files are automatically repacked. Use --repack-smaller-than to specify a minimum size")
```

cmd/restic/cmd_prune.go
```diff
@@ -78,12 +77,19 @@ func (opts *PruneOptions) AddFlags(f *pflag.FlagSet) {
 }
 
 func (opts *PruneOptions) AddLimitedFlags(f *pflag.FlagSet) {
+	var unused bool
 	f.StringVar(&opts.MaxUnused, "max-unused", "5%", "tolerate given `limit` of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in % or the word 'unlimited')")
 	f.StringVar(&opts.MaxRepackSize, "max-repack-size", "", "stop after repacking this much data in total (allowed suffixes for `size`: k/K, m/M, g/G, t/T)")
 	f.BoolVar(&opts.RepackCacheableOnly, "repack-cacheable-only", false, "only repack packs which are cacheable")
-	f.BoolVar(&opts.RepackSmall, "repack-small", false, "repack pack files below 80% of target pack size")
+	f.BoolVar(&unused, "repack-small", false, "repack pack files below 80% of target pack size")
```
MichaelEischer: The flag also has to be removed from `doc/060_forget.rst`.

cmd/restic/cmd_prune_integration_test.go
```diff
@@ -68,7 +68,7 @@ func testPruneVariants(t *testing.T, unsafeNoSpaceRecovery bool) {
 		testPrune(t, opts, checkOpts)
 	})
 	t.Run("Small", func(t *testing.T) {
```
MichaelEischer: This test case no longer tests anything new. Remove it.

changelog/unreleased/issue-5293
```diff
@@ -0,0 +1,8 @@
+Change: prune small packfiles more aggressively
+
+prune now packs more small packfiles by default. The option `--repack-small`
+is not longer needed and has been marked as deprecated.
+
+https://github.com/restic/restic/issues/5293
+https://github.com/restic/restic/pull/5359
```
MichaelEischer: I'd prefer to omit that PR as it has not been merged and the based on a quick glance the main discussion has happened in the issue not the PR.

internal/repository/prune.go
```diff
@@ -328,13 +329,32 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
...
+	slices.SortFunc(toSort, func(a, b ToSort) int {
+		return cmp.Compare(a.size, b.size)
+	})
+
+	index := len(indexPack) * 3 / 100
+	if index > 1 {
+		// right hand side of 3rd percentile, start counting from 0
+		index -= 2
+	}
+	targetPackSize := uint(max(4*(1<<20), toSort[index].size)) * 4 / 5
```
MichaelEischer: This will crash for an empty repository.

internal/repository/prune.go
```diff
@@ -328,13 +329,32 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
 	var repackCandidates []packInfoWithID
 	var repackSmallCandidates []packInfoWithID
 	repoVersion := repo.Config().Version
-	// only repack very small files by default
-	targetPackSize := repo.PackSize() / 25
+
+	// calculate 3rd percentile from sorted packfiles
```
MichaelEischer: The targetPackSize calculation should happen in a helper function.

internal/repository/prune.go
```diff
@@ -328,13 +329,32 @@ func decidePackAction(ctx context.Context, opts PruneOptions, repo *Repository,
...
+		toSort = append(toSort, ToSort{packID, uint64(ip.usedSize + ip.unusedSize)})
+	}
+	slices.SortFunc(toSort, func(a, b ToSort) int {
+		return cmp.Compare(a.size, b.size)
+	})
+
+	index := len(indexPack) * 3 / 100
+	if index > 1 {
+		// right hand side of 3rd percentile, start counting from 0
+		index -= 2
```
MichaelEischer: I'm not quite sure why this is necessary. The `index` calculation already rounds towards zero. And with the totally scientifically 😅 exactly chosen 3rd percentile, any exactness of the percentile calculation isn't relevant.

internal/repository/prune.go
MichaelEischer: Please add a short description why the calculation looks like this. (I can also help here if necessary).

MichaelEischer (changes requested):
Thanks for the PR, I've found a few nits and bugs, see below.

cmd/restic/cmd_prune.go
wplapper: done.

cmd/restic/cmd_prune_integration_test.go
wplapper: done.

changelog/unreleased/issue-5293
wplapper: removed the old PR reference.

internal/repository/prune.go
wplapper: fixed.

internal/repository/prune.go
wplapper: done.

internal/repository/prune.go
wplapper: I removed the index manipulation, since we are dealing with approximates.

cmd/restic/cmd_prune.go
wplapper: done.

internal/repository/prune.go
wplapper: I hope I managed a reasonable description.

cmd/restic/cmd_prune.go
```diff
@@ -237,40 +243,40 @@ func runPruneWithRepo(ctx context.Context, opts PruneOptions, repo *repository.R
 
 // printPruneStats prints out the statistics
 func printPruneStats(printer progress.Printer, stats repository.PruneStats) error {
-	printer.V("\nused:         %10d blobs / %s\n", stats.Blobs.Used, ui.FormatBytes(stats.Size.Used))
+	printer.V("\nused:         %10d blobs / %s", stats.Blobs.Used, ui.FormatBytes(stats.Size.Used))
```
MichaelEischer, line 246: For the future: such refactorings should at least be separate commits.

MichaelEischer (approved):
LGTM.

I've tweaked the changelog a bit and reverted the limitation on `--repack-smaller-than` being larger than a packfile size. That way the parameter can be used to control rewriting small packfiles if  the new behavior turns out to be too aggressive.

internal/repository/prune.go
MichaelEischer: I've changed the description to focus more on why use the 3rd percentile and the overall calculation. Sorry for the extra work.

PR #21813, opened 2026-05-16
Return helpful error if subfolder syntax fails on Windows

### LINKED ISSUE #21806: relative path notation is non-standard
Output of `restic version`
--------------------------

rest server

It took me ages to restore a certain folder from a backup. There is some knowledge necessary that isn't obvious or widely propagated.

I tried to select the folder with different methods --path/--include as well as latest:[path]

I tried many things, relative paths and absolute paths, won't list them all. 

Finally this worked for ls:
```bash
restic --repo rest:http://thomas:somewhereoverthe@restbow:8000/thomas ls --path C:\Users\myuser latest:/Work/MyProject
# or
restic --repo rest:http://thomas:somewhereoverthe@restbow:8000/thomas restore --target .\Temp --dry-run --path C:\Users\myuser latest:/Work/MyProject
```
Two things stood out, that triggered facepalms:

The notation of Absolute paths under Windows: 
  - /C/Users/myuser rather than C:\Users\myuser
  - having absolute path notation with a leading '/' after the snapshot where should be a relative path, e.g. latest:myuser rather than latest:/myuser

Expected behavior
-----------------

see examples above

Please make the experience consistent for Windows users. It is very confusing that some paths allow backslashes like in --target while others need slashes. It should be easy to implement a translation layer to internally convert C:\ to /C/ and back. At least output a hint how to 'correctly' use drive letters and that restic needs relative paths to start with '/' rather than only output "path not found".

Even for users of unixesque operating systems, the '/' prefixing a relative path is counterintuitive. Allow ommiting it.

Actual behavior
---------------

see above

see above

Today was the first time I had to restore something, so yes, it helped, but it took some time. I even looked for a GUI frontend, found only non-foss ones. A decent GUI tool could have saved me the time researching the obscure subpath notation rules.

Conceptually the best backup tool I found -- with a UX worth improving.

MichaelEischer:
Windows users expect to use Windows paths in the `snapshot:subfolder` syntax. However, currently only Unix style paths are supported.

Extend the "not found" error with a helpful error on Windows if the user used a backslash.

Test written by Cursor.

Fixes https://github.com/restic/restic/issues/21806

Checklist
---------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -707,7 +707,7 @@ restic users. The changes are ordered by importance.
    correctly.
 
    Specifying volume names is now handled correctly. To restore snapshots created
-   before this bugfix, use the <snapshot>:<subpath> syntax. For example, to restore
+   before this bugfix, use the <snapshot>:<subfolder> syntax. For example, to restore
    a snapshot with ID `12345678` that backed up `C:`, use the following command:
 
    ```
--- a/changelog/0.17.1_2024-09-05/issue-2004
+++ b/changelog/0.17.1_2024-09-05/issue-2004
@@ -6,7 +6,7 @@ resulting snapshot would result in an error. Note that using `C:\`
 as backup target worked correctly.
 
 Specifying volume names is now handled correctly. To restore snapshots
-created before this bugfix, use the <snapshot>:<subpath> syntax. For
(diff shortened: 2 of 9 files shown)


MichaelEischer (commented):
LGTM

PR #21817, opened 2026-05-20
sftp: Use mode 0700 for repository directories

### LINKED ISSUE #4447: Incorrect directory permissions after init via SFTP

eyupcanakman:

When a repository is initialized over SFTP, its directories were created with the SFTP server's default permissions, often 0755, instead of the 0700 used for local repositories.
The SFTP backend creates them with pkg/sftp's Mkdir and MkdirAll, which take no mode argument.
It now sets each created directory to 0700.

Closes #4447

Checklist
---------

### DIFF
--- a/internal/backend/sftp/sftp_test.go
+++ b/internal/backend/sftp/sftp_test.go
@@ -1,12 +1,15 @@
 package sftp_test
 
 import (
+	"context"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/restic/restic/internal/backend"
 	"github.com/restic/restic/internal/backend/sftp"
 	"github.com/restic/restic/internal/backend/test"
 	"github.com/restic/restic/internal/errors"
@@ -27,6 +30,14 @@ func findSFTPServerBinary() string {
 
 var sftpServer = findSFTPServerBinary()
 
+func testConfig(path string) sftp.Config {
+	return sftp.Config{
+		Path:        path,
+		Command:     fmt.Sprintf("%q -e", sftpServer),
+		Connections: 5,
+	}
+}
+
 func newTestSuite(t testing.TB) *test.Suite[sftp.Config] {
 	return &test.Suite[sftp.Config]{
 		// NewConfig returns a config for a new temporary backend that will be used in tests.
@@ -34,12 +45,8 @@ func newTestSuite(t testing.TB) *test.Suite[sftp.Config] {
 			dir := rtest.TempDir(t)
 			t.Logf("create new backend at %v", dir)
 
-			cfg := &sftp.Config{
-				Path:        dir,
-				Command:     fmt.Sprintf("%q -e", sftpServer),
-				Connections: 5,
-			}
-			return cfg, nil
+			cfg := testConfig(dir)
+			return &cfg, nil
 		},
 
 		Factory: sftp.NewFactory(),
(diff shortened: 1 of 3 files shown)


internal/backend/sftp/sftp_test.go
```diff
@@ -67,3 +70,61 @@ func BenchmarkBackendSFTP(t *testing.B) {
...
+		Connections: 5,
+	}
+
+	be, err := sftp.Create(context.Background(), cfg, func(string, ...interface{}) {})
+	rtest.OK(t, err)
+	defer func() { rtest.OK(t, be.Close()) }()
+
+	// Remove a data subdirectory so that Save has to recreate it.
+	subdir := filepath.Join(cfg.Path, "data", "01")
+	rtest.OK(t, os.RemoveAll(subdir))
```
MichaelEischer: The repository is empty at that point such that `os.Remove(...)` should also be sufficient.

internal/backend/sftp/sftp_test.go
```diff
@@ -67,3 +70,61 @@ func BenchmarkBackendSFTP(t *testing.B) {
 
 	newTestSuite(t).RunBenchmarks(t)
 }
+
+func TestCreateSetsDirPermissions(t *testing.T) {
+	if sftpServer == "" {
+		t.Skip("sftp server binary not found")
+	}
+
+	cfg := sftp.Config{
```
MichaelEischer: Please deduplicate the `sftp.Config` generation in the new and existing tests.

MichaelEischer (changes requested):
Thanks for the PR, I have two minor remarks. Besides that LGTM.

eyupcanakman:
Done, switched to `os.Remove` and pulled the shared config into a `testConfig` helper.

jtru:
I'm not convinced that changing this default and set 0700 permissions "in stone" is a good idea. (Although I have to admit I was not aware of the 0700 default for local repos either, and don't think that to be a great choice as well ;))

Taking away the opportunity to have system-wide and user-set defaults via umask affect a number of use cases that might not be immediately obvious, but still worthwhile - such as having a dedicated user account with read-only access to all repo data be able to rsync/copy all repo data over to another host.

Since restic's threat model does allow for unauthorized parties to read repository content (because the attacker cannot learn anything about snapshot contents that way), I don't think the potential gain (keeping ciphertext inaccessible via DAC) is necessarily worth any potential breakage.

MichaelEischer:
> I'm not convinced that changing this default and set 0700 permissions "in stone" is a good idea. (Although I have to admit I was not aware of the 0700 default for local repos either, and don't think that to be a great choice as well ;))

The sftp backend already enforces specific permissions on files, thus extending this to directories does not introduce new problems. Besides, there's a mechanism for group-accessible repositories for the local and sftp backend: https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#group-accessible-repositories

MichaelEischer (approved):
LGTM. Thanks!

jtru:
> > I'm not convinced that changing this default and set 0700 permissions "in stone" is a good idea. (Although I have to admit I was not aware of the 0700 default for local repos either, and don't think that to be a great choice as well ;))
> 
> The sftp backend already enforces specific permissions on files, thus extending this to directories does not introduce new problems. Besides, there's a mechanism for group-accessible repositories for the local and sftp backend: https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#group-accessible-repositories

Ah, thanks for the pointer - I was not aware of this feature/subtlety until now. After having learned these facts, I wholeheartedly support the suggested change ;)

PR #21818, opened 2026-05-21
docs: amend `restic find` documentation

wplapper:
Add more documentation for the ``find`` command, particularly how to filter by snapshot.

I tripped several times over the fact that I had (wrongly) specified a snapshot for the ``restic find`` command and was wondering why the filtering did not work as expected. 

This addition to the ``restic find --snapshot``documentation tries to explain a bit more about snapshot filtering for the ``find`` command, since the syntax is different from all the other commands using a ``SnapshotFilter``.  

no.

Checklist
---------


MichaelEischer (approved):
LGTM.

I've slightly tweaked the wording, dropped the unmarked headline and moved the warning to quote spaces above the snapshot filtering as the warning also applies without snapshot filters.

PR #21822, opened 2026-05-25
cli: add test-pattern sub-command to test path matching

thebigbone:
Adds restic test-pattern command using the same `filter.Match` and `filter.ChildMatch` logic as find/backup/restore, with optional -i/--ignore-case flag. Requires no repository, matching version/cache/generate commands.

#5708 

Checklist
---------


thebigbone:
@MichaelEischer can you review it?

PR #21823, opened 2026-05-25
backend/retry: stop retrying no-space errors

### LINKED ISSUE #4558: Restic check: no space left on device
See also https://github.com/restic/restic/issues/2453

Output of `restic version`
--------------------------
restic 0.16.2 compiled with go1.21.3 on freebsd/amd64

sftp to openssh on Windows.

```
using temporary cache in /tmp/restic-check-cache-1115144068
load indexes
check all packs
Load(<data/e9135e72b2>, 0, 0) returned error, retrying after 685.855528ms: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/e9/tmp-1265412558: no space left on device
Load(<data/3c88192929>, 0, 0) returned error, retrying after 356.154156ms: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/3c/tmp-3420658088: no space left on device
Load(<data/3c88192929>, 0, 0) returned error, retrying after 917.507365ms: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/3c/tmp-3818159037: no space left on device
Load(<data/e9135e72b2>, 0, 0) returned error, retrying after 1.002076989s: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/e9/tmp-2851860172: no space left on device
...
Load(<data/3c88192929>, 0, 0) returned error, retrying after 15.046602629s: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/3c/tmp-3495162652: no space left on device
Load(<data/3c88192929>, 0, 0) returned error, retrying after 10.828795999s: Copy: write /tmp/restic-check-cache-1115144068/8319f755b24bb716efbe3eebf650617403f1275ccdccdc77a50807a3631ff8eb/data/3c/tmp-207687818: no space left on device

(for about 4 hours, I'm not really sure why it stopped at that point.)
```
Expected behavior
-----------------
I understand how check works and that it needs tmp space to do that. I also know that it can be fixed with changing tmp dir (or making tmp larger, or whatever).
But it would be great - if possible - ~restic checks first if there is enough space to do that or exits with error. Or~ that it stops during check with an error if no space is left instead of retrying (forever?)

Actual behavior
---------------
It printed the error above for about 4 hours and finally exited for reasons unknown.

I really was out of space in /tmp. I set TMPDIR to a different location and was able to complete the check promptly.

Restic makes me happy every day (even today)!

Easy to set up and I hardly ever have to think about it. The backups have saved me more than once.

Montana:
## Summary

Treat local disk-full errors as permanent retry failures in the retry backend.

When an operation returns a wrapped no-space error, the retry loop now returns the original error immediately instead of continuing through retries.

## Motivation

The retry backend is meant for failures that may succeed on another attempt. A local disk-full condition is different: retrying the same operation usually repeats the same failure.

One visible case is `restic check`, where restic may write to a temporary cache directory. If that directory runs out of space, retrying the same `Load` operation does not make progress and delays the actionable error message.

Failing fast lets the user free space, change the cache directory, or adjust their temporary directory.

Fixes #4558.

## Changes

- Add platform-specific no-space error detection.
- Treat `ENOSPC` / disk-full errors as permanent retry errors.
- Add a regression test confirming `Load` is attempted only once for a wrapped no-space error.

## Testing

```bash
go test ./internal/backend/retry

### DIFF
--- a/internal/backend/retry/no_soace.go
+++ b/internal/backend/retry/no_soace.go
@@ -1,0 +1,14 @@
+//go:build !windows
+
+package retry
+
+import (
+	"errors"
+	"syscall"
+)
+
+var errNoSpace = syscall.ENOSPC
+
+func isNoSpaceError(err error) bool {
+	return errors.Is(err, errNoSpace)
+}
--- a/internal/backend/retry/no_space_windows.go
+++ b/internal/backend/retry/no_space_windows.go
@@ -1,0 +1,13 @@
+package retry
+
+import (
+	"errors"
+	"syscall"
+)
+
+var errNoSpace = syscall.ERROR_DISK_FULL
+
+func isNoSpaceError(err error) bool {
+	return errors.Is(err, errNoSpace) ||
+		errors.Is(err, syscall.ERROR_HANDLE_DISK_FULL)
+}
--- a/internal/backend/retry/backend_retry.go
+++ b/internal/backend/retry/backend_retry.go
@@ -107,7 +107,6 @@ func (be *Backend) retry(ctx context.Context, msg string, f func() error) error 
 
 	bo := backoff.NewExponentialBackOff()
 	bo.MaxElapsedTime = be.MaxElapsedTime
-
 	if feature.Flag.Enabled(feature.BackendErrorRedesign) {
 		bo.InitialInterval = 1 * time.Second
 		bo.Multiplier = 2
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Review by MichaelEischer — CHANGES_REQUESTED** [review]
The handling of no-space errors is indeed a problem, but the implementation is in the wrong place. The `IsPermanentError(...)` method of backends already signals which error must not be retried. For the local backend that method is at https://github.com/restic/restic/blob/f000da3b354d6071fb824db9116ae94b9f2bb534/internal/backend/local/local.go#L105 . This also fixes the current problem that the PR adds `local` backend-specific error handling to the generic retry backend.

P.S.: Please always use the PR template.

File: internal/backend/retry/no_space_windows.go [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"errors"
+	"syscall"
+)
+
+var errNoSpace = syscall.ERROR_DISK_FULL
+
+func isNoSpaceError(err error) bool {
+	return errors.Is(err, errNoSpace) ||
```
MichaelEischer, line 11: Omit the `errNoSpace` variable and just hardcode it here. (Also applies for internal/backend/retry/no_soace.go )

internal/backend/retry/no_soace.go
MichaelEischer, line 1: There's a typo in the filename.

PR #21829, opened 2026-05-30
backup: prevent hang using --stdin-from-command if upload fails

MichaelEischer:
When using `--stdin-from-command`, the `backup` command could hang until
manually cancelled if the backup stopped before all subprocess output was
consumed, for example after a failed upload to the repository. This has been
fixed.

Fixes https://github.com/restic/restic/issues/5683

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21830, opened 2026-05-31
tests - more test cases for internal/repository/check.go

wplapper:
internal/repository/check.go is the workhorse for the `restic check --read-data` command.

Here are some more tests, some of them developed with the help of claude.ai, sonnet 4.6. 
Some of the error exits have not been covered by tests before. Since the health of a repository is such an important part of the package, I thought it might be good to improve code and test coverage.

More test coverage for repository/check.go

No.

Checklist
---------

- [X] I have added tests for all code changes, see [writing tests](https://restic.readthedocs.io/en/stable/090_participating.html#writing-tests)


wplapper:
The only places in the internal/repository/check.go which are not covered by the tests are very difficult to simulate. Here is the copverage profile as .html
[coverage.html](https://github.com/user-attachments/files/28433564/coverage.html)


```
(1) if len(blobs) == 0 {
      return &ErrPackData{PackID: id, errs: []error{errors.New("pack is empty or not indexed")}}
    }

(2) if ctx.Err() != nil {
      return ctx.Err()
     }
(3) if minHdrStart > curPos {
       _, err := bufRd.Discard(minHdrStart - curPos)
       if err != nil {
          return &partialReadError{err}
       }
       curPos += minHdrStart - curPos
     }
(4) blobs, hdrSize, err := pack.List(r.Key(), bytes.NewReader(hdrBuf), int64(len(hdrBuf)))
    if err != nil {
      return &ErrPackData{PackID: id, errs: append(errs, err)}
     }
(5) if !idxHas {
      errs = append(errs, errors.Errorf("blob %v is not contained in index or position is incorrect", blob.ID))
      continue
    }
```
case (1): probably intercepted by the restic index check
case (2): when the transaction is blown out of the water
case (3): when there is a gap between the end of the payload and the beginning of the the header part of the packfile
case (4): parse error
case (5): index mismatch, should probably be covered in index tests

MichaelEischer:
Would it be possible to move those tests to `internal/repository/...` where the tested code lives? `TestGapInBlobs`, `TestCheckPackHashMismatch`, `TestCheckPackDownloadError` and `TestCheckPackPartialDownloadError` all sound like they only test code in `internal/repository/checker.go`.

wplapper:
> Would it be possible to move those tests to internal/repository/... where the tested code lives? TestGapInBlobs, TestCheckPackHashMismatch, TestCheckPackDownloadError and TestCheckPackPartialDownloadError all sound like they only test code in internal/repository/checker.go.

The only test I can move out is ``TestGapInBlobs()``. The other tests cannot be moved, since they rely on living in the package ``checker_test``. Moving them into ``ìnternal/repository`` generates an import cycle which is not allowed. The tests rely on ``checkData(ckr)`` being accessible. 

The actual function tested is ``internal/repository/check.go``. ``internal/repository/checker.go`` is just a set of wrappers around LoadIndex(), ``Packs()`` and ``ReadPacks()``.

MichaelEischer:
> The tests rely on checkData(ckr) being accessible.

`checkData` internally calls `chkr.ReadPacks`. Without a pack filter, that is a direct call of `c.Checker.ReadPacks`. That is, you can copy `collectErrors` and the `checkData` can easily be adapted to work within the `repository` package.

MichaelEischer:
The same basically applies to `setupCheckRepo`. There we just have to omit the call to `mustLoadSnapshots`.

wplapper:
Hi @MichaelEischer, 
I applied the code changes you suggested. They worked (as expected) :)

MichaelEischer (approved):
LGTM. Thanks for the extra tests!

I've dropped a few comments that likely to get outdated over time or aren't particularly relevant

wplapper:
Michael,

Thanks for all your hard work!

PR #21841, opened 2026-06-01
prevent imports of repository internals

MichaelEischer:
Add a golangci-lint rule to prevent importing repository internal packages from outside the repository package.

This required a few changes:
- convert a rewrite index integration test to a unit test as it requires modifying `index.Full`
- move a large part of the debug command into the repository package. As this code works with pack internal, that's the correct place anyways. (can be reviewed by copy pasting the original functions and then checking the resulting minimal diff)
- change a `check` test to no longer import `internal/repository/hashing` and simplify it in the process
- `list index` now has it's own dedicated helper in the repository package

While there's no strict need for this refactoring, it will help ensure that new code won't introduce dependencies to repository internals.

AI-assisted using Cursor.

No.

Checklist
---------

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,6 +40,13 @@ linters:
               desc: "internal/restic should not be imported to keep the architectural layers intact"
             - pkg: "github.com/restic/restic/internal/repository"
               desc: "internal/repository should not be imported to keep the architectural layers intact"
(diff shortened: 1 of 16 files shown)


MichaelEischer (commented):
LGTM

PR #21845, opened 2026-06-04
`restic check` and `restic repair packs`: treat missing packfiles the same as damaged and truncated packfiles

wplapper:
Missing packfiles should be treated the same as truncated or damaged packfiles.

This is the first commit of multiple commits:

(1) internal/repository/checker.go:
* add field 'Missing' to repository.PackError
* activate 'Missing' field when a packfile was detected as missing

(2) cmd/restic/cmd_check.go:
* check the repository.PackError.Missing field and skips the error return for this condition

(3) cmd/restic/cmd_repair_packs.go:
* checks for given packfiles in master index and skips the error return for a missing packfile

(4) new integration test cases for ``restic repair packs``

(5) additional documentation

Missing packfiles were not treated well by ``restic check``and ``restic repair packs``. The documentation also has no useful hints how to deal with lost packfiles.

No.

Checklist
---------


changelog/unreleased/issue-2334
```diff
@@ -0,0 +1,14 @@
+Enhancement: `restic check`, `restic repair packs` now  heal lost packfiles
```
MichaelEischer:
```suggestion
Enhancement: `restic repair packs` can handle missing packfiles
```

internal/repository/repair_pack.go
```diff
@@ -10,7 +10,9 @@ import (
 	"github.com/restic/restic/internal/restic"
 )
 
-func RepairPacks(ctx context.Context, repo *Repository, ids restic.IDSet, printer restic.Printer) error {
+func RepairPacks(ctx context.Context, repo *Repository, ids restic.IDSet,
```
MichaelEischer: Either place all parameters in one line or place each one on a separate line. Also applies to `reuploadBlobsFromPack`.

internal/repository/repair_pack.go
```diff
@@ -25,9 +27,11 @@ func RepairPacks(ctx context.Context, repo *Repository, ids restic.IDSet, printe
 		// examine all data the indexes have for the pack file
 		for b := range repo.listPacksFromIndex(ctx, ids) {
 			indexBlobs := b.Blobs
-			err := reuploadBlobsFromPack(ctx, repo, b.PackID, indexBlobs, printer, uploader)
-			if err != nil {
-				return err
+			if len(packToBlobs[b.PackID]) > 0 {
```
MichaelEischer: `loadBlobsFromPack` already skips empty blob lists. So there's nothing to gain by a length check here.

internal/repository/repair_pack.go
```diff
@@ -91,10 +95,14 @@ func resolveBlobsForPacks(ctx context.Context, repo *Repository, ids restic.IDSe
...
 }
 
-func reuploadBlobsFromPack(ctx context.Context, repo *Repository, packID restic.ID, blobs pack.Blobs, printer restic.Printer, uploader restic.BlobSaverWithAsync) error {
+func reuploadBlobsFromPack(ctx context.Context, repo *Repository, packID restic.ID, blobs pack.Blobs,
+	printer restic.Printer, uploader restic.BlobSaverWithAsync, skipMissingPackfileErrors bool,
+) error {
 	err := repo.loadBlobsFromPack(ctx, packID, blobs, func(blob restic.BlobHandle, buf []byte, err error) error {
 		if err != nil {
-			printer.E("failed to load blob %v: %v", blob.ID, err)
+			if !skipMissingPackfileErrors {
```
MichaelEischer: This is just an error message. Why hide it?

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
-		if err != nil {
-			return err
-		}
-		if _, err := io.Copy(f, bytes.NewReader(buf)); err != nil {
-			_ = f.Close()
-			return err
-		}
-		if err := f.Close(); err != nil {
-			return err
+		if err == nil {
```
MichaelEischer: The `if buf == nil` check is fully intentional and NOT a typo. If a pack file is damaged, then the whole idea of that check was to create a backup copy anyways. That behavior MUST be kept.

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
+			}
+			if _, err := io.Copy(f, bytes.NewReader(buf)); err != nil {
+				_ = f.Close()
+				return err
+			}
+			if err := f.Close(); err != nil {
+				return err
+			}
+		} else {
+			if !packsFromIndex.Has(id) {
```
MichaelEischer: Swap both branches of the if. Fewer negations make the code easier to understand.

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
 		return errors.Fatalf("%s", err)
 	}
 
+	// load all packs from master index
+	packsFromIndex := restic.NewIDSet()
+	err = repo.ListBlobs(ctx, func(blob restic.PackBlob) {
```
MichaelEischer: This loop rather belongs into `RepairPacks`, if necessary, but not in `cmd`. In `cmd`, ideally only CLI parameters parsing and wiring up the necessary pieces happens. But relevant logic should be located in `internal/`. (Yes, that rule is ignored by varying degrees, but `RepairPacks` is a case that already does things mostly correctly).

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
+				// not in index an does not load
+				printer.E("%v not a valid packfile", id)
+				ids.Delete(id)
+			} else {
+				skipMissingPackfileErrors = true
+			}
 		}
 	}
+	if len(ids) == 0 {
+		return fmt.Errorf("found no valid packfiles to delete")
```
MichaelEischer: That's not what this command is about.

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
+	})
+	if err != nil {
+		return err
+	}
+
 	printer.P("saving backup copies of pack files to current folder")
+	skipMissingPackfileErrors := false
 	for id := range ids {
 		buf, err := repo.LoadRaw(ctx, restic.PackFile, id)
 		// corrupted data is fine
```
MichaelEischer: The long form of this comment is `// corrupted data is fine. Only exit here if no data could be read at all`.

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
+When one of these errors occur, ``restic check`` will also recommend a
+``restic repair packs`` command.
+
+When duplicate blobs are found in the index, the recommendation will be to use  ``restic repair index``
+
+When mixed packs are detected (in very old repositories), the recommendation will be to
+use ``restic prune``. Modern repositories will separate tree blobs from data blobs and
+keep these blobs in separate packfiles.
+
+If someone is interested to check only specific snapshots, one can
```
MichaelEischer: The docs actively refer to a user as `you`.

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
 somewhere. Please include the check output and additional information that might
 help locate the problem.
 
-If ``check`` detects damaged pack files, it will show instructions on how to repair
-them using the ``repair packs`` command. Use that command instead of the "Repairing the
+When ``restic check`` detects damaged packfiles, it will show instructions on how to repair
+them using the ``repair packs`` command. Use that command instead of the "Repair the
```
MichaelEischer: All section titles in here us the "ing" form.

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,28 +67,48 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
+				return err
+			}
+			if err := f.Close(); err != nil {
+				return err
+			}
+		} else {
+			if !packsFromIndex.Has(id) {
+				// not in index an does not load
+				printer.E("%v not a valid packfile", id)
+				ids.Delete(id)
```
MichaelEischer: `RepairPacks` is already mostly able to handle non-existent packfiles. Thus, keep that logic in there.

changelog/unreleased/issue-2334
```diff
@@ -0,0 +1,14 @@
...
+
+In the past, deleted packfiles could not be repaired with the `restic check` and
+`restic repair packs` command, since a missing packfile triggered an error exit
+and not the repair procedure.
+
+these commands now treat missing packfiles the same way as damaged packfiles are treated.
+Missing packfiles can now "removed" correctly, by updating the master index.
+
+https://github.com/restic/restic/issues/2334
```
MichaelEischer: This issue is completely unrelated to this PR.

changelog/unreleased/issue-2334
```diff
@@ -0,0 +1,14 @@
...
+https://github.com/restic/restic/issues/3273
```
MichaelEischer: Do not link already resolved issues.

changelog/unreleased/issue-2334
```diff
@@ -0,0 +1,14 @@
...
+https://github.com/restic/restic/issues/5642
```
MichaelEischer: How does this PR relate to that issue? (The fix for #5642 is to better handle retries for truncated files. But that is not a matter for this PR).

changelog/unreleased/issue-2334
```diff
@@ -0,0 +1,14 @@
...
+https://github.com/restic/restic/pull/2876
```
MichaelEischer: Already closed PR

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
 index" section in this guide.
 
-If you are interested to check only specific snapshots, you can now
+``restic check`` treats lost=deleted packfiles identical to damaged packfiles. A
+packfile is considered damaged when one of the following conditions are met:
+
+- the packfile does not exist, but has entries in the master index
```
MichaelEischer: Why would that be a damaged packfile and not a missing one?

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
 help locate the problem.
 
-If ``check`` detects damaged pack files, it will show instructions on how to repair
-them using the ``repair packs`` command. Use that command instead of the "Repairing the
+When ``restic check`` detects damaged packfiles, it will show instructions on how to repair
+them using the ``repair packs`` command. Use that command instead of the "Repair the
 index" section in this guide.
 
-If you are interested to check only specific snapshots, you can now
+``restic check`` treats lost=deleted packfiles identical to damaged packfiles. A
```
MichaelEischer:
```suggestion
``restic check`` treats missing packfiles identical to damaged packfiles. A
```

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
+packfile is considered damaged when one of the following conditions are met:
+
+- the packfile does not exist, but has entries in the master index
+- the packfile has an incorrect length: the recorded size (in the packfile header) and actual size differ.
+
+When these damages occur, ``restic check`` will recommend a ``restic repair packs`` command.
+
+If ``check --read-data`` has been specified, ``Packdata`` errors can occur.
```
MichaelEischer: What are `Packdata` errors? How would I recognize one as user?

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
+``Packdata`` errors are defined as:
+
+- partial download errors or total download errors
+- sha256sum mismatches
+
+When one of these errors occur, ``restic check`` will also recommend a
+``restic repair packs`` command.
+
+When duplicate blobs are found in the index, the recommendation will be to use  ``restic repair index``
```
MichaelEischer: That's too unprecise to be correct. What you are referring to is cases where a pack file is listed in multiple indexes.

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
...
+
+When mixed packs are detected (in very old repositories), the recommendation will be to
```
MichaelEischer: What are "mixed packs"?

doc/077_troubleshooting.rst
```diff
@@ -80,16 +80,38 @@ Similarly, if a repository is repeatedly damaged, please open an `issue on GitHu
 somewhere. Please include the check output and additional information that might
 help locate the problem.
 
-If ``check`` detects damaged pack files, it will show instructions on how to repair
-them using the ``repair packs`` command. Use that command instead of the "Repairing the
+When ``restic check`` detects damaged packfiles, it will show instructions on how to repair
```
MichaelEischer:
```suggestion
When ``restic check`` detects damaged or missing pack files, it will show instructions on how to repair
```

doc/077_troubleshooting.rst
MichaelEischer: This case is not relevant for this troubleshooting guide.

doc/077_troubleshooting.rst
MichaelEischer: I'm not sure that this is the best place to explain the different errors that check can return in detail. A reference feels out of place for a troubleshooting guide.

MichaelEischer (changes requested):
I think this should be implemented rather differently. We should avoid adding more logic that digs around in repository internals to `cmd`. Instead I'd suggest the following:
- In `cmd_repair_packs.go` change `if buf == nil { return err }` to instead print a warning and continue.
- In `RepairPacks` the `ParallelRemove` needs a function for the `report` parameter that ignores deletion errors if `backend.IsNotExist(err)`. (or we could just let `ParallelRemove` do that by default.)
- Keep the check/checker changes.

wplapper:
I have moved all repair code from ``cmd/restic/cmd_repair_packs.go`` to ``internal/repository/repair_pack.go``:

``runRepairPacks()`` now opens the repository, loads the index and then calls `repository.RepairPacks()``. 

I will look at the troubleshooting guide later.

cmd/restic/cmd_repair_packs.go
```diff
@@ -41,18 +38,6 @@ Exit status is 12 if the password is incorrect.
 }
 
 func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal, args []string) error {
-	ids := restic.NewIDSet()
-	for _, arg := range args {
```
MichaelEischer, line 45: You've moved far too much. Parsing arguments still belongs into the `cmd` package (see https://github.com/restic/restic/pull/21845#discussion_r3477157456 and https://github.com/restic/restic/pull/21845#pullrequestreview-4574280557). My comments were specifically about the _new_ code added not about the existing code. We can discuss about the loop creating backup copies below, but the ID parsing MUST stay here.

cmd/restic/cmd_check.go
```diff
@@ -401,12 +402,13 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
...
 			}
 		}
+		p.Done()
 	}
 
 	if len(salvagePacks) > 0 {
 		printer.E("\nThe repository contains damaged pack files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.\n\n")
 		for id := range salvagePacks {
-			summary.BrokenPacks = append(summary.BrokenPacks, id.String())
+			summary.BrokenPacks = append(summary.BrokenPacks, id.Str())
```
MichaelEischer: `check` should always return full IDs unless this makes the output completely unreadable. But the latter isn't the case here, so change it back.

internal/repository/checker.go
```diff
@@ -71,7 +71,7 @@ type ErrPackData struct {
 }
 
 func (e *ErrPackData) Error() string {
-	return fmt.Sprintf("pack %v contains %v errors: %v", e.PackID, len(e.errs), e.errs)
+	return fmt.Sprintf("pack %v contains %v errors: %v", e.PackID.Str(), len(e.errs), e.errs)
```
MichaelEischer: `check` should always print full IDs unless there's a very good reason not to.

internal/repository/repair_pack.go
```diff
@@ -91,7 +154,8 @@ func resolveBlobsForPacks(ctx context.Context, repo *Repository, ids restic.IDSe
 	return packToBlobs, nil
 }
 
-func reuploadBlobsFromPack(ctx context.Context, repo *Repository, packID restic.ID, blobs pack.Blobs, printer restic.Printer, uploader restic.BlobSaverWithAsync) error {
+func reuploadBlobsFromPack(ctx context.Context, repo *Repository, packID restic.ID, blobs pack.Blobs, printer restic.Printer, uploader restic.BlobSaverWithAsync,
+) error {
```
MichaelEischer: Please fix the formatting

cmd/restic/cmd_repair_packs.go
```diff
@@ -61,33 +46,17 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
 	}
 	defer unlock()
 
-	err = repo.LoadIndex(ctx, printer)
+	snapshotLister, err := restic.MemorizeList(ctx, repo, restic.SnapshotFile)
```
MichaelEischer: Remove the snapshotLister and all other traces of short IDs in this PR. `restic repair packs` MUST always use full IDs. The only reasonable way to use `restic repair packs` is by copy&pasting the command from the `check` output. Thus, there's no need for short IDs. In addition, short IDs are prone to collisions that make the non-unique.

internal/repository/repair_pack.go
```diff
@@ -1,26 +1,89 @@
...
+	}
+
+	return result
+}
+
+// RepairPacks fixes broken / non-existing packfiles (but known to the master index)
+// there are two input parameters possible:
+// args = []string which is used in the mainline code and in the integration tests
+// ids = restic.IDSet for internal/repository/repair_pack_test.
+// If ids is empty, the args slice is used.
```
MichaelEischer: The `args` / `ids` duality is bad interface design. Having multiple arguments that roughly do the same results in more complexity, less test coverage and ultimately bugs.

Revert the function signature back to what it was before.

internal/repository/repair_pack.go
```diff
@@ -1,26 +1,89 @@
...
+	}
+	if len(ids) == 0 {
+		return errors.Fatal("no ids specified")
+	}
+
+	printer.P("saving backup copies of pack files to current folder")
+	for id := range ids {
+		buf, err := repo.LoadRaw(ctx, restic.PackFile, id)
+		// corrupted data is fine
+		if err == nil || buf != nil {
```
MichaelEischer: I'd prefer to write this as
```
if err != nil && buf == nil {
printer.E("skipping unreadable pack file %v", id)
continue
}
// rest of the if clause below
```

internal/repository/repair_pack.go
```diff
@@ -1,26 +1,89 @@
...
+}
+
+// RepairPacks fixes broken / non-existing packfiles (but known to the master index)
+// there are two input parameters possible:
+// args = []string which is used in the mainline code and in the integration tests
+// ids = restic.IDSet for internal/repository/repair_pack_test.
+// If ids is empty, the args slice is used.
+func RepairPacks(ctx context.Context, repo *Repository, snapshotLister restic.Lister, args []string, ids restic.IDSet, printer restic.Printer,
+) error {
 	printer.P("salvaging intact data from specified pack files")
```
MichaelEischer, line 14: This message should also move down along with the `bar := printer.NewCounter("pack files")` progress bar.

changelog/unreleased/pull-21845
```diff
@@ -0,0 +1,10 @@
+Enhancement: `restic repair packs` can handle missing packfiles
+
+In the past, deleted packfiles could not be repaired with the `restic check` and
```
MichaelEischer: I'd just drop the "missing feature" paragraph and merge the relevant parts into the next paragraph.

changelog/unreleased/pull-21845
```diff
@@ -0,0 +1,10 @@
...
+`restic repair packs` command, since a missing packfile triggered an error exit
+and not the repair procedure.
+
+these commands now treat missing packfiles the same way as damaged packfiles are treated.
+Missing packfiles can now "removed" correctly, by updating the master index.
```
MichaelEischer:
```suggestion
Missing packfiles are now removed from the index.
```

Users don't know what a "master index" is.

MichaelEischer (changes requested):
I've checked the code changes except the tests. See below for comments.

When addressing PR reviews please do not introduce new behavior (like handling short IDs, which shouldn't be supported here at all) that wasn't asked for unless it is actually necessary. This will just result in additional work for both of us.

internal/repository/repair_pack.go
wplapper: done so, slightly modified.

internal/repository/repair_pack.go
wplapper: gone completely.

cmd/restic/cmd_repair_packs.go
wplapper: gone completely.

changelog/unreleased/pull-21845
```diff
@@ -0,0 +1,5 @@
+Enhancement: `restic repair packs` can handle missing packfiles
+
+Missing packfiles will now "remove" missing packfiles by updating the index.
```
MichaelEischer:
```suggestion
`restic repair packs` will now handle missing packfiles by removing the from the repository index.
```

internal/repository/repository.go
```diff
@@ -1350,3 +1350,17 @@ func ZeroChunk() restic.ID {
 	})
 	return zeroChunkID
 }
+
+// PacksFromIndexAsSet generates a set of packfiles from the index
+func PacksFromIndexAsSet(ctx context.Context, repo *Repository) (result restic.IDSet, err error) {
```
MichaelEischer: This loop does _exactly_ the same as the `repo.ListBlobs(...)` call which I've complain about before, thus the comment from https://github.com/restic/restic/pull/21845#discussion_r3477157456 still applies.

wplapper:
to summarize the changes:
```
git diff master --numstat 
5       0       changelog/unreleased/pull-21845    // changelog
1       1       cmd/restic/cmd_check.go            // activate check for missing packfile
17      2       cmd/restic/cmd_repair_packs.go     // check for missing packfile
125     0       cmd/restic/cmd_repair_packs_integration_test.go // new tests
2       3       doc/077_troubleshooting.rst        // update doc
2       1       internal/repository/checker.go     // signal missing packfile, when detected
4       2       internal/repository/repair_pack.go // formatting changes only
14      0       internal/repository/repository.go  // access index and create set of packfile IDs
```

cmd/restic/cmd_repair_packs.go
```diff
@@ -66,12 +66,24 @@ func runRepairPacks(ctx context.Context, gopts global.Options, term ui.Terminal,
...
+	// an ID is considered valid if it is either
+	// found in the index or found in the packfile list
+	// the check buf == nil -> raise error needs to be replaced with the logic
+	// just below
 	for id := range ids {
 		buf, err := repo.LoadRaw(ctx, restic.PackFile, id)
 		// corrupted data is fine
-		if buf == nil {
-			return err
+		if err != nil && buf == nil && !packsFromIndex.Has(id) {
```
MichaelEischer: This is not what I've asked for in https://github.com/restic/restic/pull/21845#discussion_r3483400585 . For a missing pack file that is still in the index, we have`err == notFound`, `buf == nil` and `packsFromIndex.Has(id)`. As a result the missing file is saved as a zero byte file to disk . If `id` refers to a non-existent file that also does not exist in the index, then it is skipped here. But what for? `RepairPacks` has to be able to deal with missing files anyways.

MichaelEischer (changes requested):
I'll take over finishing this PR. We're moving in circles without getting [edit]much[/edit] closer to the target.

MichaelEischer (approved):
LGTM. I've fully dropped `PacksFromIndexAsSet` as it's not necessary IMO. The tests are also slightly simplified, with the largest change being that I've move the test helpers around as they can also be useful for other tests.

Sorry for taking over the PR, but that was simpler than commenting in full detail what I had in mind.

(I'll merge the PR tomorrow or so)

PR #21852, opened 2026-06-06
backup: skip source paths that cannot be accessed

### LINKED ISSUE #5667: `backup` creates snapshot even if all source files could not be read
**Output of `restic version`**

**What backend/service did you use to store the repository?**
local

**Problem description / Steps to reproduce**
- The source path is quoted (in the "backup" command)
- Backslashes are not escaped and the path has a final backslash.

Correctly escaping the backslashes (or even just removing the trailing backslash) seems to fix the problem.

Example:
`restic -r my_repo backup "F:\source\"`
`restic -r my_repo backup "F:\source\" --exclude-file "path\to\file"`

**Expected behavior**
restic should either report that the command line is ill-formed, or report that the source does not exist before opening the repository.

I would expect something like:
```
F:\source" --exclude-file path does not exist, skipping
Fatal: all source directories/files do not exist
```

**Actual behavior**

`restic -r my_repo backup "F:\source\"`

```
[0:00] 100.00%  1 / 1 index files loaded
error: CreateFile F:\source": La syntaxe du nom de fichier, de répertoire ou de volume est incorrecte.

Dirs:            1 new,     0 changed,     0 unmodified
Added to the repository: 296 B (255 B stored)

processed 0 files, 0 B in 0:03
snapshot 6f42a439 saved
```

`restic -r my_repo backup "F:\source\" --exclude-file "path\to\file"`

```
error: /F/source" --exclude-file path: CreateFile F:\source" --exclude-file path: La syntaxe du nom de fichier, de répertoire ou de volume est incorrecte.

Dirs:            1 new,     0 changed,     0 unmodified

processed 0 files, 0 B in 0:02
snapshot 70790799 saved
```

**Do you have any idea what may have caused this?**
It looks like quoted paths are not parsed the same way by restic and by Powershell. This issue seems to be related to #5391.

**Did restic help you today? Did it make you happy in any way?**
Yes!

eyupcanakman:

The `backup` command only skipped source paths that did not exist. A path that could not be accessed for another reason, such as a malformed path on Windows, was kept and led to an empty snapshot. It now skips any source path it cannot access and aborts with `all source directories/files do not exist` when none are left.

Closes #5667

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-5667
+++ b/changelog/unreleased/issue-5667
@@ -1,0 +1,9 @@
+Bugfix: Skip inaccessible `backup` source paths
+
+The `backup` command only skipped source paths that did not exist. A path that
+could not be accessed for another reason, such as a malformed path on Windows,
+was kept and produced an empty snapshot. Restic now skips any such path and
+aborts if none remain.
+
+https://github.com/restic/restic/issues/5667
+https://github.com/restic/restic/pull/21852
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -171,13 +171,17 @@ var ErrInvalidSourceData = errors.New("at least one source file could not be rea
 // ErrNoSourceData is used to report that no source data was found
 var ErrNoSourceData = errors.Fatal("all source directories/files do not exist")
 
-// filterExisting returns a slice of all existing items, or an error if no
-// items exist at all.
+// filterExisting returns the items that exist and can be accessed. It returns
+// ErrNoSourceData if none remain, or ErrInvalidSourceData if some were skipped.
 func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
 	for _, item := range items {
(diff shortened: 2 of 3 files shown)


MichaelEischer (approved):
LGTM

At some point we should replace `filterExisting` with a fixed version of the `errors.New("snapshot is empty")` check in the archiver. But this fix is fine for now.

eyupcanakman:
The exit-code split that `filterExisting` does (all-missing `ErrNoSourceData` exit 1 vs some-missing `ErrInvalidSourceData` exit 3) would need to move into the archiver, which only sees the all-empty case today. I could open a PR for that if you want to go that way.

MichaelEischer:
What I had in mind was removing `filterExisting` completely and instead adding proper tracking as part of `saveTree` / `unrollTree`. That would add quite a bit of complexity in the archiver, so I'm not sure we want to even do that before reworking the concept of backup sets / snapshots is done.

PR #21853, opened 2026-06-06
Add testcontainer

BataevDaniil:
I replaced the manual MinIO download with [testcontainers-go](https://github.com/testcontainers/testcontainers-go).

Testcontainers uses Docker to start MinIO and waits for it to become ready.

Benefits:
- every test gets new environment.
- each container uses a different port, etc.
- we don't need to start MinIO manually.
- we don't need to download the MinIO binary.
- it works in 14 seconds instead of 34
- testcontainers automatically handles container lifecycle (start, wait, stop, and remove after tests).

This is needed for tests in PR #5616.


wplapper:
Hi @BataevDaniil,

You are not using the standard template for creating this PR. Please use always the standard template!



Checklist
---------



See also: .github/PULL_REQUEST_TEMPLATE.md

internal/backend/s3/s3_test.go
```diff
@@ -20,101 +14,43 @@ import (
...
-	err := os.MkdirAll(dir, 0700)
-	if err != nil {
+func runMinio(ctx context.Context, t testing.TB) (*s3.Config, func()) {
+	container, err := tcminio.Run(ctx, "minio/minio:RELEASE.2025-09-07T16-13-09Z")
+	if err != nil || container == nil {
 		t.Fatal(err)
 	}
-}
-
-func runMinio(ctx context.Context, t testing.TB, dir, key, secret string) func() {
```
wplapper, line 32: You normally don't delete in existing tests - unless you found a glaring error. Normally you add more functions to suite your needs and your new environment.

MichaelEischer:
The test suite worked without docker so I don't see why we should introduce that as dependency.

> it works in 14 seconds instead of 34

The existing test using MinIO only takes 9 seconds in this run: https://github.com/restic/restic/actions/runs/27906392228/job/82575699781

PR #21848, opened 2026-06-05
move restic.{Blob,Blobs,PackedBlob} types to `./internal/repository/pack` package 

MichaelEischer:
What does this PR change? What problem does it solve?
-----------------------------------------------------
Based on https://github.com/restic/restic/pull/21846

Move restic.{Blob,Blobs,PackedBlob} types to `./internal/repository/pack` package and replace them with more limited datatypes in the repository interface. This improves the encapsulation of the repository implementation. In particular, it makes it even harder to implement any processing of pack file internals outside the `repository` package.

`restic.PackBlob` is an interface implemented by `pack.PackedBlob` that exposes most details about a blob in a pack file, except for the exact offset. 

Most changes are trivial datatype / naming changes. The more interesting changes include:
- repository interface
  - `LookupBlob` returns a `PackBlob`
  - `ListBlobs` passes a `PackBlob` to its callback
  - `ListPacksFromIndex` is unexported as it only had `repository` package internal users
  - `ListPack` has been replaced by `ListPackHandles` which only returns `BlobHandle`s.
  - `LoadBlobsFromPack` accepts `BlobHandle`s instead of `Blobs`. The Blobs are looked up internally. repository-internal usage can use `loadBlobsFromPack` with the old interface.
- CopyBlob now requires a `*Repository` as source. The repack test therefore uses a `oneConnectionBackend` instead of wrapping the repository.
- The filerestorer test includes a fake repository, which now has to replicate the Blob datatypes internally.

Changes assisted by Cursor. I've drastically cleaned up the resulting commits.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------
No.

Checklist
---------

- [ ] I have added tests for all code changes, see [writing tests](https://restic.readthedocs.io/en/stable/090_participating.html#writing-tests)
- [ ] I have added documentation for relevant changes (in the manual).
- [ ] There's a new file in `changelog/unreleased/` that describes the changes for our users (see [template](https://github.com/restic/restic/blob/master/changelog/TEMPLATE)).
- [x] I'm done! This pull request is ready for review.

PR #21857, opened 2026-06-07
repository: change LoadBlob, LookupBlob and LookupBlobSize to BlobHandles

MichaelEischer:
https://github.com/restic/restic/pull/21848 will change `LoadBlobsFromPack` to accept BlobHandles. The other Load and Lookup method of a repository still expected separate type and id arguments. As we already have a proper type use that instead. `ListBlob` and `ListPack` also indirectly return BlobHandles. There a few downsides though: several places now have to first construct the BlobHandle which amounts to a little bit more code than before. But the possible simplifications due to less datatype splitting and the explicit BlobHandle constructions are somewhat balanced such that making this change is not a problem.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21860, opened 2026-06-08
backend/cache: use backend types not restic types

MichaelEischer:
Backends are not supposed to import the `restic` package. The cache backend was the last exception to this rule. Switch over to the backend types to address this.

`Backend.List` does not need to parse the filename into IDs as this is already handled in `Repository.List`.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21861, opened 2026-06-08
restic: introduce Counter interface to decouple from ui/progress

MichaelEischer:
Pulling in the `restic` package currently also includes the `ui/progress` package along with its dependencies. Especially for tests this is sort of overkill. Switch the restic package to use a Counter interface instead. To not have to litter the code with nil checks, also introduce a NoopCounter.

Changes made using Cursor.

No.

Checklist
---------


MichaelEischer:
LGTM

PR #21862, opened 2026-06-08
Move Lock type to repository package and clean it up a bit

MichaelEischer:
Historically the `Lock` struct has lived in the `restic` package. By now it could only be constructed by the repository package. As it is an internal API of the repository package move it there. To make room for the Lock struct, rename the `Lock()` function to `LockRepo()`.

I've reduced unnecessary exports a bit, merged the Unlocker and lockInfo structs, split the Lock struct into a pure Lock DTO and a lockHandle type and removed the mutex from the lockHandle.

AI assisted changes using Cursor with significant manual cleanup.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21867, opened 2026-06-10
backup --time option - allow different time formats

### LINKED ISSUE #2065: timezone in TimeFormat variable?

wplapper:
Instead of only allowing one time format global.TimeFormat="2006-01-02 15:04:05" for the ``restic backup --time`` option allow multiple time formats as in the ``restic find`` command for the options ``--oldest`` and ``--newest``.

This PR fixes the old issue https://github.com/restic/restic/issues/2065

Fix: replace the call to time.ParseInLocation() with a call to parseTime() in cmd_find.go. Add the requested format to list of allowed formats.

closes https://github.com/restic/restic/issues/2065

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-2065
+++ b/changelog/unreleased/issue-2065
@@ -1,0 +1,9 @@
+Change: Allow more flexible formats of the ``restic backup --time`` option
+
+Restic backup --time allowed in the past only the format "2006-01-02 15:04:05".
+
+backup --time can now use the full range of time formats as used in the
+``restic find`` command for the options ``--oldest`` and ``--newest``.
+
+https://github.com/restic/restic/issues/2065
+https://github.com/restic/restic/pull/21867
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -529,7 +529,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts global.Options, te
 	timeStamp := time.Now()
 	backupStart := timeStamp
 	if opts.TimeStamp != "" {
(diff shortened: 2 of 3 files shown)


wplapper:
- I am not sure if the documentation needs updating. ``restic backup`` does not document the ``--time`` option
- I am not sure if there is a need for testing a new time format. The time formats are not extensively covered in any test scenario.
- I am not touching the thorny issue of how to display time stamps in different environments: e.g. ``restic snapshot``, ``restic mount``, JSON.

Issue #3658, opened 2022-02-27
Restic won't find/restore files using --include containing square brackets in name (Windows)


JsBergbau:
Output of `restic version`
--------------------------
restic 0.12.1 compiled with go1.16.6 on windows/amd64

How did you run restic exactly?
-------------------------------
`restic restore -t C:\temp --include "/G/Restictest/Folder 2 [square brackets]"`

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
local file system

Expected behavior
-----------------
restic restores the folder `Folder 2 [square brackets]`

Actual behavior
---------------
nothing is restored

Steps to reproduce the behavior
-------------------------------
Add the files of the attachment into a repo and try to restore with the steps above.

Do you have any idea what may have caused this?
-----------------------------------------------
No

Do you have an idea how to solve the issue?
-------------------------------------------
It works with `restic restore -t C:\temp --include "/G/Restictest/Folder 2 *" latest`

However it took me some minutes to figure it out that the problem is obviously caused by square brackets in file name.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------
Restic makes me feel happy and save every day because for me it is the best backup program on the market right now. I trust my data is safely backuped with restic. It makes me happy to improve restic.

[Restictest.zip](https://github.com/restic/restic/files/8148761/Restictest.zip)

MichaelEischer:
The path matching is done as described in https://pkg.go.dev/path/filepath#Match. Unfortunately, in windows there's no way to escape square brackets.

JsBergbau:
`[[]`
Works as expected, meaning it will all find/restore all files with on opening square bracket. However for closing squarebracket `[]]` it won't work since it is not recognized as character class.

MichaelEischer:
Shouldn't `[[]]` work? `[[]` results in an opening square bracket. And `]` on its own does not require escaping.

JsBergbau:
`[[]]` finds only `[]`. So only files that have an opening square bracket followed by a closing one.

MichaelEischer:
Yes, so in summary this means that you only have to escape an opening square bracket in a filename as `[[]` and that's it. Closing square brackets don't have to be changed if the opening square brackets are properly escaped.

MichaelEischer:
Closing in favor of https://github.com/restic/restic/issues/4867 .

pryon:
[...]

Issue #3005, opened 2020-10-09
Don't treat entries in --files-from as glob definitions


jim-collier:
I know (e.g. [2448](https://github.com/restic/restic/issues/2448)) that there is already disagreement on whether this is a "bug", or a "feature request", and in fact globbing was explicitly introduced as a "feature" at some point into `--files-from`.

My argument is that, given that no other software on Earth that I could find treats lines in a `--files-from` (or analogous feature) as glob specifications, Restic shouldn't either, and it's a bug.

But if you'd rather consider it a new feature request, fine with me - something like:

- A new flag that doesn't glob, e.g. `--files-raw-from`, and/or
- An environment variable that overrides `--files-from` default behavior so as to not glob

Why?

1. As far as I know and have been able to determine through google (surely with some obscure exceptions), all other software with such a flag expects the list to be a plain 1:1 raw dump of _exact_ file path specifications (whether absolute or relative).
1. The writers of such software know that such lists are generated by _some other tool_. Not by hand-crafting glob specifications.
    - Hand-crafted wildcard specifications are literally what a completely different set of related features are for: pattern-based "exclude" and "include" arguments and file pointers. Rsync, for example, supports at least _six_ variations of specifying pattern-based includes, excludes, and mixed definitions.)
1. The Linux way is "Do One Thing Well", and chain together the tools you need.
1. Treating `--files-from` entries as glob specifications, introduces unexpected errors, violating the principle of "least surprise". As I discovered, with brackets in filenames - which I now have to jump through additional hoops to programmatically process and escape.
    - There's deep personal irony here. The reason I have so many brackets in filenames, is because long ago when using primarily Windows, parentheses in filenames caused (and still does) errors in Windows CMD processor's half-baked attempt at a modern-ish "if exist fileX ( thenY )" branching syntax. So now with brackets instead of parentheses, I'm getting bugs in Windows. (Well at least these are solvable by escaping.)

As an example of what programs typically expect, when using `rsync`, `rclone`, and numerous other tools, users and admins often:

- Generate a list of regular (non-symlink/socket/named pipe) files, filtered by date or size, with something like `find . -f | sort > filelist1.txt`.
- Filter that list with something like `grep -iPv ".*\/local\/Trash\/.*" filelist1.txt > filelist2.txt` (etc.) As many times as necessary.
- `rsync --files-from filelist2.txt source/ dest/`

[...]

Issue #2448, opened 2019-10-20
Paths in --files-from files that include brackets aren't backed up


tomwassenberg:
Output of `restic version`
--------------------------
`restic 0.9.5 compiled with go1.12.4 on windows/amd64`

How did you run restic exactly?
-------------------------------
To circumvent #233, I'm using the `--files-from` flag to pass a file containing a pre-filtered list of paths to backup. This file consists of absolute paths to files/dirs on an NTFS disk, without any escaping performed on them. I use the following (simplified) command to do this:
```
D:\> restic -r local:E:\ backup --files-from path-test.txt
```

What backend/server/service did you use to store the repository?
----------------------------------------------------------------
A local repository.

Expected behavior
-----------------
Restic should parse file paths in a `--files-from` file that include brackets, and include these in the backup if they exist.

Actual behavior
---------------

The contents of a `--files-from` file seem to be evaluated as glob patterns. Paths containing brackets, that are allowed in paths on NTFS volumes, are never matched.

Steps to reproduce the behavior
-------------------------------
```
D:\> dir /B D:\[example].txt
[example].txt

D:\> type path-test.txt
D:\[example].txt

D:\> restic -r local:E:\ backup --files-from path-test.txt
pattern "D:\\[example].txt" does not match any files, skipping
Fatal: nothing to backup, please specify target files/dirs
```

Do you have any idea what may have caused this?
-----------------------------------------------
Looking at the git history, the desired behavior seems to have been present until #1892 was merged. Parsing asterisks as wildcards isn't a problem for me, since these are not allowed in filenames on most filesystems. However `filepath.Glob` treats more characters as special, which seems to cause the problem at hand as a side-effect.

Do you have an idea how to solve the issue?
-------------------------------------------
Restic should parse all valid file paths from a `files-from` file literally, which is implied by the [docs](https://restic.readthedocs.io/en/latest/040_backup.html#including-and-excluding-files). That makes is possible to pass all filenames that are valid in the containing filesystem, without having to do any escaping.

`--files-from` could also remain using `filepath.Glob` patterns, but that would be counter-intuitive to the use case that is mentioned in the docs. That use case states that `--files-from` is useful to pass a _pre-filtered_ list of file paths, using e.g. `find | grep`.

[...]

PR #21872, opened 2026-06-12
feat(restore): add --include-from-raw for literal-path includes

### LINKED ISSUE #4867: Restic restore --files-from feature
Output of `restic version`
--------------------------
restic 0.16.4 compiled with go1.21.6 on linux/amd64

It would be helpful if the --files-from feature that exists in "backup" function to backup a list of files saved as paths in a text file also existed for the "restore" function to restore a list of files based on their paths that are read from a text file and restored from a snapshot to a target location.

This would allow me to create a list of files paths in a snapshot using the restic "ls" function and save the output list to a text file, then filter those files to wanted ones to restore daily from the latest snapshot using a restore script.
This can somehow already be achieved by the recently added restic "--include-file" option to the "restore" function but the --include and --exclude expect patterns not plain file names and that causes a problem when trying to restore files which have pattern interpretable characters in their paths like "WR-7200o[.jpg" for example.

--------------------------
I use it in all my production servers and by far it's the most efficient, fast, yet reliable backup solution that we have worked with. I genuinely can't give the developers enough praise for their efforts in creating a production-ready and reliable open-source backup solution that's actively maintained and improved more than even the other proprietary solutions out there.

mayerwin:

`restore --include` / `--include-file` match every entry through Go's `filepath.Match` glob engine. This makes brackets, asterisks and question marks unrepresentable as literal characters, and it has bitten users multiple times when restoring single files whose names contain `[` or `]`.

This PR adds `--include-from-raw FILE` to `restore` (and `rewrite`, which shares `IncludePatternOptions`). The file contains NUL-terminated absolute paths; each one is matched literally against the snapshot tree. **No glob semantics are changed for existing flags.** This is a direct mirror of [#3017](https://github.com/restic/restic/pull/3017), which added `--files-from-raw` to `backup` for exactly the same reason on the input side.

### Why this matters

There is no clean workaround today. Brackets in filenames are common (Windows: `Document [draft v2].pdf`, macOS: `Photo [2024-06-10].heic`, Linux: `[archive].tar.gz`). On Windows the situation is strictly worse because Go's `filepath.Match` treats `\` as a path separator (not an escape) AND `ValidatePatterns` rejects the POSIX class self-escape `[[]` / `[]]`, so there is no way to express a literal `[` as a match character at all.

The only current workaround is to broaden the `--include` pattern with a wildcard — e.g. `--include 'Q3 *.pdf'` to fetch one file named `Q3 [draft].pdf`. That over-restores every sibling matching the relaxed pattern. For users with exact lists (file managers, GUI wrappers, recovery scripts that already know the paths) this means downloading from the repo backend and reading from disk every same-prefix sibling for no functional reason — wasted bandwidth, wasted disk, wasted energy. Backup-side users have had `--files-from-raw` since #3017 specifically to avoid this; the restore side has been the missing half.

### What changes

- New `--include-from-raw FILE` flag on `restore` and (via the shared `IncludePatternOptions`) `rewrite`. Can be specified multiple times. `FILE` may be `-` for stdin.
- Each line in `FILE` is terminated by a NUL byte; the empty filename is an error (matches the backup-side `--files-from-raw` semantics).
- Patterns from this file are matched **literally**: `*`, `?`, `[`, `]`, and `**` lose their special meaning. The flag is additive — entries combine with any `--include` / `--iinclude` / `--include-file` already given.
- Existing `--include` semantics are unchanged. No existing test was modified.

### Implementation

- `internal/filter/literal.go` adds `LiteralPattern` and `ParseLiteralPatterns`. The key insight is that the existing `match` function already takes an `==` equality branch when `patternPart.isSimple == true` (line `filter.go:204-206`); `LiteralPattern` constructs `Pattern` values with every part forced to `isSimple = true`, so `filepath.Match` is never invoked for these patterns. No changes to the existing matcher. There is deliberately no separate validation step: a literal pattern cannot be a malformed glob, and the only invalid entry — the empty string — is already rejected while reading the file.
- `internal/filter/include.go` adds `IncludeFilesRaw` to `IncludePatternOptions`, registers the flag, reads paths via a small NUL-terminated parser equivalent to the backup-side `readFilenamesFromFileRaw`, and feeds the result through `IncludeByLiteralPattern`. An empty listing adds no constraint, the same as an empty `--include-file`, rather than selecting nothing.
- `cmd/restic/cmd_restore.go` is unchanged — `IncludePatternOptions.CollectPatterns` was already the chokepoint, so adding `IncludeFilesRaw` lights the flag up on every command that consumes the struct (today: `restore` and `rewrite`).
- `doc/050_restore.rst` documents the flag, the use case, and the Windows-specific motivation.
- `changelog/unreleased/pull-restore-include-from-raw` per `changelog/TEMPLATE` conventions.

### Test coverage

- **Unit:** `TestIncludeByLiteralPattern` in `internal/filter/include_test.go` covers brackets matched literally, the negative case (would-match-as-class file does NOT match), `*` matched literally, plain entries alongside metachar-bearing entries, `childMayMatch` propagation for parent dirs (needed so the restorer keeps descending), and unrelated paths.
- **Integration:** `TestRestoreIncludeFromRaw` in `cmd/restic/cmd_restore_integration_test.go` backs up a tree containing `reports/Q3 [draft].pdf`, `reports/Q3 d.pdf` (the class-interpretation false-positive), `reports/summary.md`, and `reports/internal.txt`. It restores via `--include-from-raw` requesting only the bracket file and `summary.md`, then asserts: the bracket file IS restored, `summary.md` IS restored, the false-positive sibling and the unrelated file are NOT.

Both pass on Windows 11 (where the underlying glob limitation bites hardest). The pre-existing `TestRewriteIncludeNothing` and `TestBackendRetryConfig` Windows failures reproduce on pristine master (tar.exe + symlink permission issues unrelated to this change).

```
$ go test -run 'TestRestoreIncludeFromRaw|TestRestoreIncludes|TestRestoreFilter|TestRestoreMustFail|TestIncludeBy' ./internal/filter/... ./cmd/restic/...
ok  github.com/restic/restic/internal/filter  0.021s
ok  github.com/restic/restic/cmd/restic       0.974s
```

- #4867 — original feature request asking for verbatim `--include` / `--exclude` on restore, motivated by exactly the same `WR-7200o[.jpg` case this PR enables.
- #5565 — Windows-specific reproducer (`1_text test format (cover) [vocals]_(no reverb).mp3`), closed as a duplicate of #4867.
- Prior closed threads on the same gap: #2276, #2448, #3005, #3658.

The earlier issues asked for a change to `--include` semantics or for restic to escape brackets implicitly, and were closed because that would break existing users who rely on glob behavior. This PR doesn't change glob semantics — it adds a separate, explicitly literal flag that the user opts into by name, and it touches no existing matcher. The precedent for this exact shape is `--files-from-raw` (PR #3017), which added the same opt-in literal mode on the backup side and has been part of restic since 0.13.

Closes #4867 and #5565.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-4867
+++ b/changelog/unreleased/issue-4867
@@ -21,4 +21,6 @@ and `--include-file` (any matching pattern selects the file). The same
 flag is registered on the `rewrite` command, which shares the include
 infrastructure.
 
+https://github.com/restic/restic/issues/4867
+https://github.com/restic/restic/issues/5565
 https://github.com/restic/restic/pull/21872
--- a/changelog/unreleased/pull-21872
+++ b/changelog/unreleased/pull-21872
@@ -21,4 +21,4 @@ and `--include-file` (any matching pattern selects the file). The same
 flag is registered on the `rewrite` command, which shares the include
 infrastructure.
 
-https://github.com/restic/restic/pull/REPLACE_ME
+https://github.com/restic/restic/pull/21872
--- a/changelog/unreleased/pull-restore-include-from-raw
+++ b/changelog/unreleased/pull-restore-include-from-raw
@@ -1,0 +1,24 @@
+Enhancement: Add `--include-from-raw` option to the `restore` command
+
+The `restore` command's `--include` and `--include-file` options matched
+each entry as a `filepath.Match` glob, which gives some characters a
+special meaning (`*`, `?`, `[`, `]`, `\`). For paths whose names contain
+brackets in particular, there was no portable way to request a single
+file: Go's `filepath.Match` rejects the POSIX class self-escape `[[]`
+and on Windows the backslash escape is unavailable because `\` is a
+path separator. The only workaround was to broaden the pattern with a
+wildcard like `Q3 *.pdf`, which over-restores every sibling matching
+the relaxed pattern — wasted bandwidth and disk for callers that
+already had exact paths.
+
+The `restore` command now accepts `--include-from-raw FILE`, mirroring
+the existing `backup --files-from-raw`: the file contains
+NUL-terminated absolute paths, each one matched literally against the
+snapshot tree. Glob metacharacters lose their special meaning in this
+mode, so paths containing brackets, asterisks, or question marks
+round-trip cleanly. The flag composes with `--include`, `--iinclude`,
+and `--include-file` (any matching pattern selects the file). The same
+flag is registered on the `rewrite` command, which shares the include
+infrastructure.
+
+https://github.com/restic/restic/pull/REPLACE_ME
--- a/cmd/restic/cmd_restore_integration_test.go
+++ b/cmd/restic/cmd_restore_integration_test.go
@@ -75,6 +75,15 @@ func testRunRestoreExcludesFromFile(t testing.TB, gopts global.Options, dir stri
 	rtest.OK(t, testRunRestoreAssumeFailure(t, snapshotID.String(), opts, gopts))
 }
 
+func testRunRestoreIncludesFromRawFile(t testing.TB, gopts global.Options, dir string, snapshotID restic.ID, includesRawFile string) {
+	opts := RestoreOptions{
+		Target: dir,
+	}
+	opts.IncludeFilesRaw = []string{includesRawFile}
+
+	rtest.OK(t, testRunRestoreAssumeFailure(t, snapshotID.String(), opts, gopts))
+}
+
 func TestRestoreMustFailWhenUsingBothIncludesAndExcludes(t *testing.T) {
 	env, cleanup := withTestEnvironment(t)
 	defer cleanup()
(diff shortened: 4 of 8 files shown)


wplapper:
Please fix the PR description to follow the PR template at https://raw.githubusercontent.com/restic/restic/refs/heads/master/.github/PULL_REQUEST_TEMPLATE.md . This is NOT optional when using AI-agents.

mayerwin:
I've reformatted the description to follow the repository's pull request template, the change/problem write-up, the prior-discussion links (#4867, #5565, and the earlier closed threads), and the checklist are now in their respective sections. No code changes; this is purely a description edit.

Ready for review whenever a maintainer has a moment, happy to iterate on the flag name, doc wording, or test layout.

mayerwin:
Pushed a small follow-up (6583ec9) tightening the implementation after a self-review:

- **Removed `ValidateLiteralPatterns`.** A literal pattern can't be a malformed glob, and the empty entry is already rejected while reading the NUL-separated file, so the only thing the validator did was reject whitespace-only paths — which diverges from `backup --files-from-raw` (the flag this mirrors) and would make a file literally named `" "` impossible to restore.
- **An empty listing now adds no constraint**, the same as an empty `--include-file`, instead of registering a matcher that selects nothing.
- Dropped a doc comment that referenced a not-yet-existing `--exclude-from-raw` flag.
- The changelog now lists the closed issues and is named `issue-4867` per `changelog/TEMPLATE`.

Verified on Windows 11 (go1.26): `go build` / `go vet` clean, and `go test -run 'TestIncludeByLiteralPattern|TestRestoreIncludeFromRaw|TestRestoreIncludes' ./internal/filter/... ./cmd/restic/...` passes. The integration test restores exactly the bracket file plus one literal sibling and correctly leaves behind the sibling that a character-class interpretation would have matched.

PR #21877, opened 2026-06-13
snapshots: show location name instead of current timezone

MichaelEischer:
The snapshots command footer includes the current timezone since restic 0.19.0. However the time zone change change during the year, e.g. between `CET` and `CEST` for Germany. The printed snapshot time are in local time, that is they correctly use either `CET` or `CEST` in the example. Therefore the correct timezone is actually `Europe/Berlin`. Fix this by switching to the location name. Note that if not `TZ` environment variable is set on Unix systems, then the output will just state `Local`, but that's hopefully also sufficient.

Fixes https://forum.restic.net/t/possible-bug-in-timezone-naming/10867

Checklist
---------

### DIFF
--- a/changelog/unreleased/pull-21876
+++ b/changelog/unreleased/pull-21876
@@ -1,0 +1,8 @@
+Bugfix: Show timezone location in `snapshots` output
+
+Since restic 0.19.0, the `snapshots` command printed the current timezone when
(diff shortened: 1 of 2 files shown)


MichaelEischer (commented):
LGTM

PR #21882, opened 2026-06-14
Untangle imports of repository, restorer, fs

MichaelEischer:
- Extract `TempFile` and `PreallocateFile` into internal/fileio from`internal/fs`. The latter primarily focuses on converting between data.Node and the actual filesystem state. But those two functions implement unrelated fs helpers.
- repository.LockRepo returns a plain unlock `func ()` instead of an interface.
- Move `progress.Printer` to `restic` package. This makes the `repository` package independent of the `ui` package.
- Extract `HardlinkIndex` into `internal/data` from `internal/restorer`. The index was also used by `cmd_stats`. Thus it is a shared helper for the data model.
- Introduce an `restorer.ProgressReporter` interface to invert the dependency direction between `restorer` and `ui/restorer`.
- As drive-by fix let cmd_mount use `fs.HasPathPrefix` instead of it's own helper.

Most changes made using Cursor.

No. Hopefully wraps up the recent refactoring spree.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21883, opened 2026-06-14
find: improve blob resolution for broken pack files

MichaelEischer:
find` would only look up completely missing pack files in the index. A still existing but broken pack file would result in an error.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

PR #21886, opened 2026-06-14
archiver: pass ItemAction to progress reporter instead of data.Node

MichaelEischer:
The progress reporter only ever accessed a very limited part of the data.Node. As a clean replacement calculate the action description and only pass that. This also makes the ui package independent of the data package.

No.

Checklist
---------


MichaelEischer (commented):
LGTM

Issue #3366, opened 2021-04-18
restic -x (one-file-system) stats all entries of mountpoints which contain a new directory


sithglan:
Output of restic version

restic 0.12.0 compiled with go1.15.8 on linux/amd64

How did you run restic exactly?

        export AWS_ACCESS_KEY_ID='user'
        export AWS_SECRET_ACCESS_KEY='password'
        export RESTIC_PASSWORD='aeskey'
        export RESTIC_REPOSITORY='s3:http://172.17.0.1:9199/kvm-01'
        strace -o /run/shm/restic.log -f restic -q backup -x / > /dev/null

What backend/server/service did you use to store the repository?

        minio

Expected behavior

        I expect that restic does not traverse into nfs mounted filesystems.

Actual behavior

        While restic does not backup the files in nfs mounted directories, it
        stats them and also complains when one of them disappear during backup.

(tomcat-06) [/run/shm] mount | grep /gcl
172.19.0.253:/srv/gcl/export on /gcl type nfs (rw,relatime,vers=3,rsize=32768,wsize=32768,namlen=255,hard,proto=udp,timeo=11,retrans=3,sec=sys,mountaddr=172.19.0.253,mountvers=3,mountport=59574,mountproto=udp,local_lock=none,addr=172.19.0.253)
(tomcat-06) [/run/shm] grep '"/gcl/'
.tmpfs      restic.log
(tomcat-06) [/run/shm] grep '"/gcl/' restic.log | wc -l
636408
(tomcat-06) [/run/shm] grep '"/gcl/' restic.log | head
12496 newfstatat(AT_FDCWD, "/gcl/100002_VOLVO_7098824_1414664811204.gcl",  <unfinished ...>
12496 newfstatat(AT_FDCWD, "/gcl/100003_MERCEDES_7548233_1419801055271.gcl",  <unfinished ...>
12496 newfstatat(AT_FDCWD, "/gcl/100003_VOLVO_7097724_1419797980090.gcl", {st_mode=S_IFREG|0644, st_size=166550, ...}, AT_SYMLINK_NOFOLLOW) = 0
12496 newfstatat(AT_FDCWD, "/gcl/100003_VOLVO_7097771_1420839619094.gcl",  <unfinished ...>
12496 newfstatat(AT_FDCWD, "/gcl/100003_VOLVO_7097771_1421359438826.gcl",  <unfinished ...>
[...]

PR #21887, opened 2026-06-14
backup - show excluded files and directories in verbose mode - text/JSON

### LINKED ISSUE #21870: Log excluded files/directories

Restic should output skipped directories (because of exclusions) in JSON output.

During backup when JSON output (and I think verbose mode) is enabled, restic already outputs a line for each file which is not excluded. This line contains either `"action":"unchanged"` or `"action":"modified"`.

It would be very nice, if it could also output a line with `"action":"excluded"` if an item has been skipped. If a directory has been skipped, it is not neccessary to recurse into that directory, but it would be very helpful if the directory itself would be logged as excluded.

An additional nice-to-have feature would be if the reason for the exclusion would be logged. (i.e. the exclusion rule, or something like "cache-directory", "cloud-file", ...)

It would make it much easier to find out if I excluded too much.

Everyday. I trust restic more than commercial solutions (v...m)

wplapper:
This a minimal implementation of showing excluded files in verbose mode while running backup.

This PR adds the capability of tracing excluded files/directories in verbose mode, in text and JSON modes.

So far no counting of excluded files/directories in the summary.

This is an minimal invasive implementation which shows excluded files directories as 
"exclude >>path<<" (text mode) or
{
  "message_type": "excluded_item",
  "item": >>path<<
} // JSON mode
  
closes https://github.com/restic/restic/issues/21870

Checklist
---------


fkusche:
Wow great that was quick!
I compiled your branch (856a4bd9d) and tried it out.

I now get the following messages on stderr: (This is the full output - nothing more)
```
could not access exclusion tagfile: lstat /media/hidrive/_nobackup.txt: permission denied
could not access exclusion tagfile: lstat /media/hidrive/CACHEDIR.TAG: permission denied
could not access exclusion tagfile: lstat /media/hidrive/users/_nobackup.txt: permission denied
could not access exclusion tagfile: lstat /media/hidrive/users/CACHEDIR.TAG: permission denied
```
Interestingly, those files don't exist. This is a remote SMB share which I mounted using these options: `_netdev,vers=3.11,seal,noserverino,uid=2000,gid=2000,forceuid,forcegid,file_mode=0660,dir_mode=0770,credentials=...` However, according to the JSON logs, the files below `/media/hidrive` have been backed up.

Also I don't get any output with `"action": "excluded"` in the logs.

This is the commandline (I removed a lot of additional exclude-statements because it would be very long otherwise)
```bash
/usr/local/bin/restic backup --compression max --limit-upload 3000 --json -v --no-scan --one-file-system --exclude-if-present=_nobackup.txt --exclude-caches --iexclude=tmp --iexclude=temp --iexclude=__pycache__ [...] /media/hidrive / /home /boot
```

Thank you very much for your effort!

wplapper:
Hi,

I have checked that ``backup -v --json`` will produce messages as described above: I took part of my home directory, which contains python files and excluded ``__pycache__`` as you did in your example. It generates the exclude messages. I assume that you are at the latest version of the restic repository, commit 7d36449e. This is the base for my PR - but you should habe got compilation errors otherwise. 

To definitely get this issue resolved: run your restic command with ``>/tmp/restic-debug.log; DEBUG_LOG=/tmp/restic-debug.log /usr/local/bin/restic backup ...`` but reduce the amount of files you backup. You should find messages like this in /tmp/restic-debug.log:

``archiver.(*Archiver).save	73	/home/user/.../__pycache__ is excluded by path`` which should go hand in hand with JSON messages
```
{
  "message_type": "exclude_file",
  "item": "/home/user/.../__pycache__"
} 
```

fkusche:
Apologies, you are of course completely right. I checked out your repository but forgot to switch to the correct branch. (When I copied the commit id above, I even copied it from the `git log` output but did not notice, that there was no `HEAD` marker.)

I now re-ran it and it already satisfies what I need. I even found an exclude which was too broad, already! Thank you!

I have found one minor thing (I can live with it - but I want to provide the feedback to you because I am really glad that you implemented this so quickly):

If a directory contains a marker file / `CACHEDIR.TAG` or is excluded because of `--one-file-system`, the contents directly in this directory are all logged as `excluded`. For example this logs all entries in `/dev`, `/run`, `/proc`, ... too. This is a bit unexpected, because one would think that restic just won't read the directory at all. (Interestingly, the marker file or `CACHEDIR.TAG` is logged twice. Once as `unchanged` and once as `excluded`.)

internal/archiver/archiver.go
```diff
@@ -183,6 +186,7 @@ func New(repo archiverRepo, filesystem fs.FS, opts Options) *Archiver {
 		CompleteItem: func(string, *data.Node, *data.Node, ItemStats, time.Duration) {},
 		StartFile:    func(string) {},
 		CompleteBlob: func(uint64) {},
+		ExcludedItem: func(string, string) {},
```
MichaelEischer: The `fileType` parameter is never read and hence should be removed. The values `saveDir` and `saveTree` are also implementation details that shouldn't even be exposed to the ui code let alone the user.

internal/archiver/archiver.go
```diff
@@ -348,6 +352,7 @@ func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, me
 		}
 
 		if excluded {
+			arch.ExcludedItem(pathname, "saveDir")
```
MichaelEischer: I'd prefer to move those calls inside `archiver.save`. That's also the place where currently the debug logs regarding excludes are emitted.

changelog/unreleased/issue-21870
```diff
@@ -0,0 +1,7 @@
+Change: show excluded items during backup in verbose mode
+
+``restic backup -v`` now shows excluded items in verbose mode. Both text and JSON
```
MichaelEischer: The changelog is in Markdown format. That is, single backticks should be used (unlike the docs).

The text output is only shown for `-vv` (which is the correct level). JSON is more verbose in that regard.

changelog/unreleased/issue-21870
```diff
@@ -0,0 +1,7 @@
+Change: show excluded items during backup in verbose mode
```
MichaelEischer: According to the rules for JSON output parsing, this is just an enhancement and not a change of behavior.

internal/ui/backup/progress.go
```diff
@@ -179,3 +180,9 @@ func (p *Progress) Finish(snapshotID restic.ID, summary *archiver.Summary, dryru
 	p.Updater.Done()
 	p.printer.Finish(snapshotID, summary, dryrun)
 }
+
+func (p *Progress) ExcludedItem(path string, fileType string) {
+	p.mu.Lock()
```
MichaelEischer: A plain call to the printer does not require locking.

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -730,3 +741,47 @@ func TestBackupSkipIfUnchanged(t *testing.T) {
...
+	env.gopts.Verbosity = 2
+	output, err := testRunBackupOutput(t, backupOptions, env.gopts, []string{filepath.Join(env.testdata, "0", "for_cmd_ls")})
+	rtest.OK(t, err)
+
+	type MessageType struct {
+		MessageType string `json:"message_type"` // "verbose_status"
+	}
+	type MessageAction struct {
+		MessageType string `json:"message_type"` // "verbose_status"
+		Action      string `json:"action"`       // any action
```
```suggestion
		Action      string `json:"action,omitempty"`       // any action
```
Then you can omit the `MessageType` struct and the redundant parsing.

cmd/restic/cmd_backup_integration_test.go
```diff
@@ -730,3 +741,47 @@ func TestBackupSkipIfUnchanged(t *testing.T) {
...
+	}
+
+	// drill down into JSON line
+	var mType MessageType
+	var mAction MessageAction
```
MichaelEischer: Place those variables right next to where they are used for the first time.

doc/075_scripting.rst
```diff
@@ -311,6 +311,17 @@ Verbose status provides details about the progress, including details about back
 | ``total_files``           | Total number of files                                    | uint64  |
 +---------------------------+----------------------------------------------------------+---------+
 
+Excluded backup items are detailed in this format:
+
++---------------------------+----------------------------------------------------------+---------+
+| ``message_type``          | Always "verbose_status"                                  | string  |
```
MichaelEischer: The datatype is different so this should use a different `message_type`. That makes it easier to parse the output.

internal/archiver/archiver.go
MichaelEischer: Also please pass `abstarget` to `ExcludedItem` then the log output shows exactly the filename against which restic matched the exclude patterns.

MichaelEischer (changes requested):
Thanks for the PR! It requires a few small cleanups, see comments for details. The somewhat larger item is that I'd prefer to use a different `message_type`. For type-safe parsing it's already annoying to have to check the `message_type` first and I don't think we want to introduce another such dependency on the `action` field.

MichaelEischer:
> If a directory contains a marker file / `CACHEDIR.TAG` or is excluded because of `--one-file-system`, the contents directly in this directory are all logged as `excluded`. For example this logs all entries in `/dev`, `/run`, `/proc`, ... too.

That's how excludes currently work, see https://github.com/restic/restic/issues/3366 .

> Interestingly, the marker file or `CACHEDIR.TAG` is logged twice. Once as `unchanged` and once as `excluded`

I couldn't reproduce that behavior. My test:
```
mkdir simple
touch simple/work.txt
touch simple/ignore
restic backup -vv --exclude-if-present ignore simple
```

This logs `work.txt` as excluded and `ignore` as backed up.

fkusche:
> > If a directory contains a marker file / `CACHEDIR.TAG` or is excluded because of `--one-file-system`, the contents directly in this directory are all logged as `excluded`. For example this logs all entries in `/dev`, `/run`, `/proc`, ... too.
> 
> That's how excludes currently work, see #3366 .

Thanks for the reference. As I said, it's not a big issue. And if it is known and not a subtly introduced bug, then even better.

> > Interestingly, the marker file or `CACHEDIR.TAG` is logged twice. Once as `unchanged` and once as `excluded`
> 
> I couldn't reproduce that behavior. My test:

I'm currently on vacation and can't check again in the next days. Anyway it was only a minor finding which I wanted to tell.

doc/075_scripting.rst
wplapper: Implemented all the above recommendations.

doc/075_scripting.rst
```diff
@@ -311,6 +311,15 @@ Verbose status provides details about the progress, including details about back
 | ``total_files``           | Total number of files                                    | uint64  |
 +---------------------------+----------------------------------------------------------+---------+
 
+Excluded backup items are detailed in this format:
+
++---------------------------+----------------------------------------------------------+---------+
+| ``message_type``          | Always "exclude_file"                                    | string  |
```
MichaelEischer: What about `excluded_item` instead? `exclude_file` feels somewhat strange as even the struct type includes a `// file or directory name` comment. `excluded_item` would also neatly align with the name of the internal progress callback.

MichaelEischer (changes requested):
Thanks for the changes. I have one final remark regarding the `message_type` name, see below.

MichaelEischer (approved):
LGTM. Thanks!

doc/075_scripting.rst
wplapper: fixed it.

PR #21893, opened 2026-06-18
cmd/mount: add --open-file-manager flag

bgurmendi:
**What does this PR change? What problem does it solve?**

Add a `--open-file-manager` flag to the `mount` command. When set, restic launches the system file manager at the mountpoint once the repository is ready to browse. This uses `xdg-open` on Linux and FreeBSD, and `open` on macOS. The file manager is launched in a goroutine so it does not block the mount process. If launching fails (e.g. no desktop environment), the error is logged at debug level only and does not affect the mount.

**Was the change previously discussed in an issue or on the forum?**

No.

**Checklist**

### DIFF
--- a/changelog/unreleased/pull-21893
+++ b/changelog/unreleased/pull-21893
@@ -1,0 +1,8 @@
+Enhancement: Add `--open-file-manager` flag to `mount` command
+
+The `mount` command now supports the `--open-file-manager` flag, which
+automatically opens the system file manager at the mountpoint once the
+repository is ready to browse. This uses `open` on macOS and `xdg-open`
+on Linux and FreeBSD.
+
+https://github.com/restic/restic/pull/21893
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -6,7 +6,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
 
@@ -101,6 +103,7 @@ type MountOptions struct {
 	OwnerRoot            bool
 	AllowOther           bool
 	NoDefaultPermissions bool
+	OpenFileManager      bool
 	data.SnapshotFilter
 	TimeTemplate  string
 	PathTemplates []string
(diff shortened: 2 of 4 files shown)


doc/man/restic-mount.1
```diff
@@ -78,6 +78,10 @@ Exit status is 12 if the password is incorrect.
 \fB--no-default-permissions\fP[=false]
```
wplapper, line 78: Thank you for your PR. Please remove doc/man/restic-mont.1 and doc/man/restic.1 from your list of changed files. The are auto-generated when a new release of restic gets build. See ``restic generate --help`` for further details.

wplapper:
Please use the standard template for opening a new PR. See .github/PULL_REQUEST_TEMPLATE.md for details. This template will ask you to provide a changelog entry, provide a test for the new fucntionality and in your case a short update to the ``restic mount`` documentation.

rawtaz:
I'm sorry to be blunt, but what actual problem does this solve? Why can't the user just add an extra word on the command line to open the file browser?

bgurmendi:
I appreciate the direct question. It's true I haven't explained the use case correctly.

I have several people, both individuals and organizations, using Restic for backups of their work computers. It's a simple system of cron jobs and scripts that perform the backups. I also configure a script for them to mount and view the data. This way, they are self-sufficient for most of their needs when recovering a single file.

The problem lies in how this mount script performs the mount: it creates the mount directory, opens the file manager in that directory, and then executes Resti's mount command. For the user, a file manager opens that displays nothing; clicking "refresh" also shows nothing, until Resti finally completes the mount, and then, if they click "refresh," the files become visible.

I've tried to make the script more complex by reading Resti's output and waiting for it to indicate that the mount is complete before opening the file manager.  But this is quite complex to do well; it requires launching background processes and being careful not to get stuck reading the output of `restic`, etc. I think many inexperienced users can benefit from this parameter, gaining peace of mind and clarity.

I've already received some panicked calls saying the backup is empty, when all that was needed was a little patience and updating the file manager again. I'll update the pull request to better explain the use case.

Thanks for your comment.

bgurmendi:
Current solution:

```bash
#!/bin/bash
set -euo pipefail  # exit on error, undefined vars, pipe failures

MOUNTPOINT="/tmp/restic-remote"

mkdir -p $MOUNTPOINT

coproc script -q -c "restic mount \"$MOUNTPOINT\" --time-template \"2006/01/2006-01-02_15-04\"" /dev/null 2>&1

# Trap para pasar Ctrl+C al coprocess
trap "kill $COPROC_PID 2>/dev/null" SIGINT SIGTERM

# Lee desde el descriptor del coprocess
while IFS= read -r line <&${COPROC[0]}; do
  echo "$line"  # Mostrar en pantalla
  [[ "$line" == *"Now serving the repository at"* ]] && break
done

xdg-open "$MOUNTPOINT" >/dev/null 2>&1 &

# Esperar a que termine (o Ctrl+C lo mata)
wait $COPROC_PID

```

With this pull request, the script is like this:

```bash
#!/bin/bash
set -euo pipefail  # exit on error, undefined vars, pipe failures
MOUNTPOINT="/tmp/restic-remote"
mkdir -p $MOUNTPOINT
restic mount "$MOUNTPOINT" --time-template "2006/01/2006-01-02_15-04" --open-file-manager
```

PR #21900, opened 2026-06-20
archiver: ignore duplicate but excluded directory entry

### LINKED ISSUE #21899: Synology @eaDir backup fails after 0.19 upgrade despite exclude rules
Output of `restic version`
--------------------------

restic 0.19.0 compiled with go1.26.4 on linux/amd64

Backblaze B2

I run restic in Docker on a Synology NAS.

The backup source inside the container is `/source`.

The backup started failing after upgrading restic to the 0.19.x series. The same backup job worked before the upgrade.

The command is:

```bash
docker-compose run --rm restic backup \
  -v \
  /source \
  --exclude-file=/source/docker/restic/excludes.txt \
  --skip-if-unchanged \
  --exclude-caches \
  --tag synology
```
The source contains Synology metadata directories named `@eaDir`.
The exclude file contains explicit rules for those directories:

```text
**/@eaDir
**/@eaDir/**
/source/data/@eaDir
/source/data/@eaDir/**
/source/**/@eaDir
/source/**/@eaDir/**
```
The exclude file is visible inside the container at:

```text
/source/docker/restic/excludes.txt
```
Even with these exclude rules, restic still reports an error for `/source/data/@eaDir`.

Relevant output:

```text
open repository
using parent snapshot 50f625f5
load index files
start scan on [/source]
start backup on [/source]
error: /source/data/@eaDir: name "@eaDir" is not greater than last name "@eaDir"
scan finished in 175.846s: 351902 files, 375.466 GiB

Files:           1 new,     1 changed, 351900 unmodified
Dirs:            2 new,     3 changed, 37592 unmodified
Data Blobs:      2 new
Tree Blobs:      5 new
Added to the repository: 6.610 KiB (3.102 KiB stored)

processed 351902 files, 375.466 GiB in 3:57
snapshot ab0a35a6 saved
```

The Synology scheduled task reports this as:

```text
Current status: 3 (Interrupted)
```

Steps to reproduce on my system:

1. Run restic backup in Docker on Synology DSM.
2. Use `/source` as the backup source.
3. Have a Synology metadata directory such as `/source/data/@eaDir`.
4. Add exclude rules for `@eaDir`, including both glob and absolute-path rules.
5. Run `restic backup` or `restic backup --dry-run`.
6. restic still prints:

```text
error: /source/data/@eaDir: name "@eaDir" is not greater than last name "@eaDir"
```

Expected behavior
-----------------

The excluded `@eaDir` directory should not cause a backup warning or error.

I would expect restic either:

1. not to enter `/source/data/@eaDir` at all because it is excluded, or
2. at least not to fail the backup with exit status 3 because of an explicitly excluded directory.

Actual behavior
---------------

The snapshot is created, but restic exits with status 3.

It prints:

```text
error: /source/data/@eaDir: name "@eaDir" is not greater than last name "@eaDir"
```

and then:

```text
```

This causes the Synology scheduled task to be marked as interrupted/failed.

This looks like a regression after upgrading restic to the 0.19.x series.

The same backup setup worked before the upgrade.

MichaelEischer:
Refactoring the archiver to use `TreeNodeIterator` has led to a subtle behavior change in regard to duplicate directory entries. Previously, when they were excluded, the error was hidden as well. This changed and could result in unnecessary exit code 3 errors.

Explicitly skip duplicate directory entries if the first instance was excluded.

Test is written by Cursor Agent (or rather 80% copy&pasted from the surrounding code)

Fixes https://github.com/restic/restic/issues/21899

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-21899
+++ b/changelog/unreleased/issue-21899
@@ -1,0 +1,9 @@
+Bugfix: Fix excludes of duplicate directory entries during `backup`
+
+Since restic 0.19.0, creating a backup of a directory containing
+duplicate directory entries always resulted in "Warning: at least
+one source file could not be read" even if the files in question
+were excluded. This has been fixed.
+
+https://github.com/restic/restic/issues/21899
+https://github.com/restic/restic/pull/21900
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -320,11 +320,19 @@ func (arch *Archiver) saveDir(ctx context.Context, snPath string, dir string, me
 	finder := data.NewTreeFinder(previous)
 	defer finder.Close()
 
+	var lastExcluded string
+
 	for _, name := range names {
 		// test if context has been cancelled
 		if ctx.Err() != nil {
 			debug.Log("context has been cancelled, aborting")
 			return futureNode{}, ctx.Err()
+		}
+
+		if name == lastExcluded {
+			// Skip duplicate directory entry if it was already excluded.
+			// This avoids printing errors about duplicate directory entries even though the entry in question is ignored.
+			continue
 		}
 
 		pathname := arch.FS.Join(dir, name)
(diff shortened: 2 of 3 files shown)


MichaelEischer (commented):
LGTM

PR #21907, opened 2026-06-21
Fix broken snapshots via `restic repair snapshots --forget`

### LINKED ISSUE #21892: restic repair snapshots cannot remove broken snapshot file
Output of `restic version`
--------------------------
restic 0.19.0-dev (compiled manually) compiled with go1.26.4 on linux/amd64

local filesystem, rclone:onedrive

I manage to create a snapshotfile with length zero while copying snapshots from one repository to another. I interrupted the copy process when it started creating new snapshots in the target repository.
The simple reproduction step is to touch a file in a local repository at 
``./snapshots/33b18e677cf0b071c304dd04c9555584ac72c27ee1d010d3dfb57a369bc2a6f7`` and use this repository for for further error testing.

Expected behavior
-----------------

I would expect to be able to heal the broken snapshot file by deleting it with restic methods. Currently this is not possible. 

Actual behavior
---------------

``restic check`` reports the error correctly:

```
error: failed to load snapshot 33b18e67: LoadRaw(<snapshot/33b18e677c>): invalid data returned

The repository is damaged and must be repaired. Please follow the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html .

```

``restic snapshots`` produces the following output:
`Ignoring "Ignoring "33b18e677cf0b071c304dd04c9555584ac72c27ee1d010d3dfb57a369bc2a6f7": failed to load snapshot 33b18e67: LoadRaw(<snapshot/33b18e677c>): invalid data returned": failed to load snapshot 33b18e67: LoadRaw(<snapshot/33b18e677c>): invalid data returned`` 
followed by the valid snapshot output.

``restic repair snapshots 33b18e677cf0b071c304dd04c9555584ac72c27ee1d010d3dfb57a369bc2a6f7`` produces:

```
repository 3cea7a41 opened (version 2, compression level better)
[0:00] 100.00%  3 / 3 index files loaded
Ignoring "33b18e677cf0b071c304dd04c9555584ac72c27ee1d010d3dfb57a369bc2a6f7": failed to load snapshot 33b18e67: LoadRaw(<snapshot/33b18e677c>): invalid data returned

no snapshots were modified
```

repair snapshots is not geared up to intercept an error from a snapshot file failure. It uses higher level methods ``FindFilteredSnapshots()`` which intercept the error and ignore it.

I am happy to use and develop restic.I will develop a PR to fix this issue.

wplapper:
Currently there is no method in restic to remove broken snapshot files. This PR enables the fixing=removing these broken snapshot files.

The main culprit is ``FindFilteredSnapshots()`` in cmd/restic/find.go, which acts as an interface to ``(*data.SnapshotFilter).FindAll()``. It sees the broken snapshot, sends a message to the error printer and happily ignores it. In future, all calls to  ``FindFilteredSnapshots()`` will be replaced with calls to ``(*data.SnapshotFilter).FindAll()``, but in a separate PR.

Added function ``scanAndRemoveBrokenSnapshot()`` which will run through all snapshots and will memorize the broken snapshot files. If the list of broken snapshot files is matched by one of the explicit arguments, this snapshot will be removed.

close https://github.com/restic/restic/issues/21892

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-21892
+++ b/changelog/unreleased/issue-21892
@@ -1,13 +1,11 @@
-Enhancement: `restic repair snapshots --remove-ids` remoives broken snapshots
+Enhancement: `restic repair snapshots --remove-ids` removes broken snapshots
 
 In the past restic had no command to remove broken snapshot files.
 
 `restic repair snapshots --remove-ids` can now remove broken snapshot files,
-which have been found previously by `restic check` or
-by earlier error messages from restic.
+which have been found previously by `restic check` or by earlier error messages from restic commands.
 
-`restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot
-file if such a snapshot exists. Otherwise an error will be raised.
+For example, `restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot file with id `1d204771`
 
 https://github.com/restic/restic/issues/21892
 https://github.com/restic/restic/pull/21907
--- a/changelog/unreleased/issue-21892
+++ b/changelog/unreleased/issue-21892
@@ -1,0 +1,13 @@
+Enhancement: `restic repair snapshots --remove-ids` remoives broken snapshots
+
+In the past restic had no command to remove broken snapshot files.
+
+`restic repair snapshots --remove-ids` can now remove broken snapshot files,
+which have been found previously by `restic check` or
+by earlier error messages from restic.
+
+`restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot
+file if such a snapshot exists. Otherwise an error will be raised.
+
+https://github.com/restic/restic/issues/21892
+https://github.com/restic/restic/pull/21907
--- a/changelog/unreleased/issue-21892
+++ b/changelog/unreleased/issue-21892
@@ -1,11 +1,11 @@
-Enhancement: `restic repair snapshots --remove-ids` removes broken snapshots
+Enhancement: `restic repair snapshots --forget` removes broken snapshots
 
 In the past restic had no command to remove broken snapshot files.
 
-`restic repair snapshots --remove-ids` can now remove broken snapshot files,
+`restic repair snapshots --forget` can now remove broken snapshot files,
 which have been found previously by `restic check` or by earlier error messages from restic commands.
 
-For example, `restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot file with id `1d204771`
+For example, `restic repair snapshots --forget 1d204771` will remove a broken snapshot file with id `1d204771`.
 
 https://github.com/restic/restic/issues/21892
 https://github.com/restic/restic/pull/21907
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -339,6 +339,7 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 
 	printer.P("check snapshots, trees and blobs\n")
 	errChan = make(chan error)
+	var brokenSnapshots []string
 	var wg sync.WaitGroup
 
 	wg.Add(1)
@@ -351,13 +352,17 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 
 	for err := range errChan {
 		errorsFound = true
-		if e, ok := err.(*checker.TreeError); ok {
+		switch e := err.(type) {
+		case *checker.TreeError:
 			printer.E("error for tree %v:\n", e.ID.Str())
 			for _, treeErr := range e.Errors {
 				summary.NumErrors++
 				printer.E("  %v\n", treeErr)
 			}
-		} else {
+		case *checker.SnapshotError:
+			printer.E("snapshot error %v: %v", e.ID, e.Message)
+			brokenSnapshots = append(brokenSnapshots, e.ID)
+		default:
 			summary.NumErrors++
 			printer.E("error: %v\n", err)
 		}
@@ -414,12 +419,18 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 		printer.E("Damaged pack files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
 	}
 
+	if len(brokenSnapshots) > 0 {
+		printer.E("\nThe repository contains damaged snapshot files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.\n\n")
+		printer.E("restic repair snapshots --remove-ids %s", strings.Join(brokenSnapshots, " "))
+		printer.E("Damaged pack files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
+	}
+
 	if ctx.Err() != nil {
 		return summary, ctx.Err()
 	}
 
 	if errorsFound {
-		if len(salvagePacks) == 0 {
+		if len(salvagePacks) == 0 && len(brokenSnapshots) == 0 {
 			printer.E("\nThe repository is damaged and must be repaired. Please follow the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html .\n\n")
 		}
 		return summary, errors.Fatal("repository contains errors")
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -421,8 +421,8 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 
 	if len(brokenSnapshots) > 0 {
 		printer.E("\nThe repository contains damaged snapshot files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.\n\n")
-		printer.E("restic repair snapshots --remove-ids %s", strings.Join(brokenSnapshots, " "))
-		printer.E("Damaged pack files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
+		printer.E("restic repair snapshots --forget %s", strings.Join(brokenSnapshots, " "))
+		printer.E("Damaged snapshot files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
 	}
 
 	if ctx.Err() != nil {
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -421,7 +421,7 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 
 	if len(brokenSnapshots) > 0 {
 		printer.E("\nThe repository contains damaged snapshot files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.\n\n")
-		printer.E("restic repair snapshots --forget %s", strings.Join(brokenSnapshots, " "))
+		printer.E("restic repair snapshots --forget %s\n\n", strings.Join(brokenSnapshots, " "))
 		printer.E("Damaged snapshot files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
 	}
 
--- a/cmd/restic/cmd_repair_snapshots.go
+++ b/cmd/restic/cmd_repair_snapshots.go
@@ -2,7 +2,10 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"regexp"
 	"slices"
+	"strings"
 
 	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/errors"
@@ -65,8 +68,9 @@ Exit status is 12 if the password is incorrect.
 
 // RepairOptions collects all options for the repair command.
 type RepairOptions struct {
-	DryRun bool
-	Forget bool
+	DryRun    bool
+	Forget    bool
+	removeIDs bool
 
 	data.SnapshotFilter
 }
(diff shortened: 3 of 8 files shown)


changelog/unreleased/issue-21892
```diff
@@ -0,0 +1,13 @@
+Enhancement: `restic repair snapshots --remove-ids` remoives broken snapshots
```
MichaelEischer:
```suggestion
Enhancement: `restic repair snapshots --remove-ids` removes broken snapshots
```

changelog/unreleased/issue-21892
```diff
@@ -0,0 +1,13 @@
...
+
+In the past restic had no command to remove broken snapshot files.
+
+`restic repair snapshots --remove-ids` can now remove broken snapshot files,
+which have been found previously by `restic check` or
+by earlier error messages from restic.
+
+`restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot
```
MichaelEischer:
```suggestion
For example, `restic repair snapshots --remove-ids 1d204771` will remove a broken snapshot file with id `1d204771`
```

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -65,19 +68,88 @@ Exit status is 12 if the password is incorrect.
...
 	f.BoolVarP(&opts.DryRun, "dry-run", "n", false, "do not do anything, just print what would be done")
 	f.BoolVarP(&opts.Forget, "forget", "", false, "remove original snapshots after creating new ones")
+	f.BoolVar(&opts.removeIDs, "remove-ids", false, "remove given snapshotfiles from repository")
 
 	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
 }
 
+// scanAndRemoveBrokenSnapshot scans every snapshot file in the repository
+// (regardless of any filter, This is distinct from the tree/content repair below:
+// developed with the help of claude.ai, Sonnet 4.6
```
MichaelEischer: Drop the inline comments which ai agent was used to write some code. This rather belongs in the PR description. After a proper review, the comment doesn't hold much valuable information anymore.

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -65,19 +68,88 @@ Exit status is 12 if the password is incorrect.
...
+func scanAndRemoveBrokenSnapshot(ctx context.Context, repo restic.Repository,
```
MichaelEischer: dryRun is part of RepairOptions.

Also either keep all parameters in a single line or place all parameters on separate lines:
```
func name(
   param1,
   param2,
   ....
) {
```

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -65,19 +68,88 @@ Exit status is 12 if the password is incorrect.
...
+	dryRun bool, snapshotLister restic.Lister, printer restic.Printer,
+) error {
+
+	if !opts.SnapshotFilter.Empty() {
+		return fmt.Errorf("no filtering allowed for `repair snapshot --remove-ids`")
+	}
+
+	var broken restic.IDs
+	// walk through all the snapshots, find broken ones
+	_ = opts.SnapshotFilter.FindAll(ctx, snapshotLister, repo, nil, func(id string, _ *data.Snapshot, err error) error {
```
MichaelEischer: Ignoring the error here looks wrong. After all it means that listing the snapshots has failed, which definitely should be reported.

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -65,19 +68,88 @@ Exit status is 12 if the password is incorrect.
 
 // RepairOptions collects all options for the repair command.
 type RepairOptions struct {
-	DryRun bool
-	Forget bool
+	DryRun    bool
+	Forget    bool
+	removeIDs bool
```
MichaelEischer:
```suggestion
	RemoveIDs bool
```

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -65,19 +68,88 @@ Exit status is 12 if the password is incorrect.
...
+		return ctx.Err()
+	}
+
+	if len(broken) == 0 {
+		printer.P("found no broken snapshot files")
+		return nil
+	}
+
+	reHex := regexp.MustCompile(`^[0-9a-fA-F]+$`)
+	for _, removeID := range brokenIDs {
```
MichaelEischer: Correctly resolving a shortID should be done using `internal/restic/backend_find.go`. (the `find` command is bad example). But as this is a rather technical command and users would get the command they should run from `check` anyways, I don't see why we should expand snapshot ids here at all.

cmd/restic/cmd_check.go
```diff
@@ -414,12 +419,18 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 		printer.E("Damaged pack files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
 	}
 
+	if len(brokenSnapshots) > 0 {
+		printer.E("\nThe repository contains damaged snapshot files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.\n\n")
+		printer.E("restic repair snapshots --remove-ids %s", strings.Join(brokenSnapshots, " "))
+		printer.E("Damaged pack files can be caused by backend problems, hardware problems or bugs in restic. Please open an issue at https://github.com/restic/restic/issues/new/choose for further troubleshooting!\n")
```
MichaelEischer: This should say `snapshot files`

cmd/restic/cmd_snapshots.go
```diff
@@ -90,12 +90,18 @@ func runSnapshots(ctx context.Context, opts SnapshotOptions, gopts global.Option
 	defer unlock()
 
 	var snapshots data.Snapshots
-	for sn := range FindFilteredSnapshots(ctx, repo, repo, &opts.SnapshotFilter, args, printer) {
-		snapshots = append(snapshots, sn)
-	}
+	_ = opts.SnapshotFilter.FindAll(ctx, repo, repo, args, func(_ string, sn *data.Snapshot, err error) error {
```
MichaelEischer: The change in this file belongs into https://github.com/restic/restic/pull/21912 .

MichaelEischer (changes requested):
After reading through the PR, I think we can simplify it quite a bit. From what I've seen so far, we could simplify the CLI to `restic repair snapshots --forget $ids` (yes, that does not require new CLI parameters).

`restic repair snapshots` can already handle snapshots with a missing tree in the snapshot. So instead of adding a completely separate code path, it should just delete unreadable snapshots when told so by the user. That is, `restic repair snapshot --forget` should also remove all broken snapshots. To not have to process all other snapshots, we can use `restic repair snapshots --forget $ids`.

Users probably don't really care why a snapshot is broken such that having all snapshot fixing behind the same flow is much easier to handle.

cmd/restic/cmd_snapshots.go
```diff
@@ -297,7 +303,7 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
 	if zoneName == "Local" {
 		zoneName = "local time"
 	}
-	tab.AddFooter(fmt.Sprintf("Timestamps shown in %s\n%s", zoneName, footer))
+	tab.AddFooter(fmt.Sprintf("Timestamps shown in %s timezone\n%s", zoneName, footer))
```
MichaelEischer: Revert that unintentional change.

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -146,7 +147,26 @@ func runRepairSnapshots(ctx context.Context, gopts global.Options, opts RepairOp
...
+			if err2 != nil {
+				return err2
+			}
+
+			if opts.DryRun {
+				printer.P("would remove broken snapshot %v", brokenID.Str())
+				return nil
+			}
+
+			if err := repo.RemoveUnpacked(ctx, restic.WriteableSnapshotFile, brokenID); err != nil {
```
MichaelEischer: There's no check for the `--forget` flag anywhere.

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -146,7 +147,26 @@ func runRepairSnapshots(ctx context.Context, gopts global.Options, opts RepairOp
 	})
 
 	changedCount := 0
-	for sn := range FindFilteredSnapshots(ctx, snapshotLister, repo, &opts.SnapshotFilter, args, printer) {
+	errOuter := opts.SnapshotFilter.FindAll(ctx, snapshotLister, repo, args, func(id string, sn *data.Snapshot, err error) error {
+		if err != nil {
+			brokenID, err2 := restic.Find(ctx, repo, restic.SnapshotFile, id)
```
MichaelEischer: At this point, `id` is either a full ID or could not be resolved. So there's no reason to call `restic.Find` here. What would be a good idea though is to use `restic.ParseID(s)` and only handle parsable IDs.

I'm wondering whether we should always check that `repo.LoadRaw` returns data or a `restic.ErrInvalidData` error. The downside of that would be that only files can be deleted that can be downloaded without throwing an error.

A middle ground might be to only enable this "broken snapshots deletion path" if args is not empty. WDYT?

cmd/restic/cmd_repair_snapshots.go
```diff
@@ -159,10 +179,15 @@ func runRepairSnapshots(ctx context.Context, gopts global.Options, opts RepairOp
 		if changed {
 			changedCount++
 		}
-	}
+		return nil
+	})
+
 	if ctx.Err() != nil {
```
MichaelEischer: The context check is redundant after the `snapshot_find.go` changes.

changelog/unreleased/issue-21892
```diff
@@ -0,0 +1,11 @@
+Enhancement: `restic repair snapshots --remove-ids` removes broken snapshots
```
MichaelEischer: The changelog entry and the troubleshooting guide need updating.

cmd/restic/cmd_snapshots.go
```diff
@@ -90,12 +90,18 @@ func runSnapshots(ctx context.Context, opts SnapshotOptions, gopts global.Option
 	defer unlock()
 
 	var snapshots data.Snapshots
-	for sn := range FindFilteredSnapshots(ctx, repo, repo, &opts.SnapshotFilter, args, printer) {
-		snapshots = append(snapshots, sn)
-	}
-	if ctx.Err() != nil {
-		return ctx.Err()
+	err = opts.SnapshotFilter.FindAll(ctx, repo, repo, args, func(_ string, sn *data.Snapshot, err error) error {
```
MichaelEischer: I don't understand why `cmd_snapshots.go` is changed in this PR. From my point of view, it belongs into https://github.com/restic/restic/pull/21912 . Either explain why or revert **all** changes in this file.

cmd/restic/cmd_repair_snapshots_integration_test.go
```diff
@@ -136,3 +138,44 @@ func TestRepairSnapshotsIntact(t *testing.T) {
...
+	testRunBackup(t, filepath.Dir(env.testdata), []string{"testdata"}, BackupOptions{}, env.gopts)
+
+	// create zero length file in "snapshots"
+	// will fail with
+	// failed to load snapshot 1d204771: LoadRaw(<snapshot/1d20477115>): invalid data returned
+	handle, err := os.Create(filepath.Join(env.repo, "snapshots", "1d20477115fb872069a28a80ffb95a82cb8b1b1920de046a68c0195da63f30cf"))
+	rtest.OK(t, err)
+	rtest.OK(t, handle.Close())
+
+	createRandomFile(t, env, "just_a_name", 42)
```
MichaelEischer: Use `test.Random(seed, count int)` instead of creating a temporary file.

cmd/restic/cmd_repair_snapshots_integration_test.go
```diff
@@ -136,3 +138,44 @@ func TestRepairSnapshotsIntact(t *testing.T) {
...
+
+	// run repair snapshots
+	repairOpts := RepairOptions{}
+	env.gopts.BackendTestHook = nil
+	_, err = withCaptureStdout(t, env.gopts, func(ctx context.Context, gopts global.Options) error {
+		return runRepairSnapshots(ctx, gopts, repairOpts, []string{"1d204771", target}, gopts.Term)
+	})
+	rtest.OK(t, err)
+
+	// verfiry that there are no snapsho errors: ``restic check``
```
MichaelEischer:
```suggestion
	// verify that there are no snapshot errors: ``restic check``
```

MichaelEischer (changes requested):
The current state would also remove snapshots for `restic repair snapshot $ids` (without --forget) which is too aggressive IMHO. I'm somewhat undecided how much safeguards we need to ensure that we don't wipe snapshots if a backend temporarily misbehaves or whether the fact that the user has called `restic repair snapshots` is already enough.

cmd/restic/cmd_repair_snapshots.go
wplapper: ``--forget``is now tested before removal.

cmd/restic/cmd_repair_snapshots.go
wplapper: Sorry, I had used ``restic.ParseID(s)`` here before and it created error message
``invalid length for ID: "1d204771"``.

cmd/restic/cmd_repair_snapshots.go
wplapper: Oversight: forget to remove this check.

cmd/restic/cmd_snapshots.go
wplapper: see above ...

cmd/restic/cmd_repair_snapshots.go
wplapper: > I'm wondering whether we should always check that repo.LoadRaw returns data or a restic.ErrInvalidData error. The downside of that would be that only files can be deleted that can be downloaded without throwing an error.

I suggest that the code should catch all snapshot errors, but will guard against unintentional removals by checking for ``brokenID``in ``args``.

cmd/restic/cmd_snapshots.go
wplapper: I checked out cmd/restic/cmd_snapshots.go from master.

cmd/restic/cmd_repair_snapshots_integration_test.go
wplapper: done so.

cmd/restic/cmd_repair_snapshots_integration_test.go
wplapper: fixed

MichaelEischer (approved):
LGTM. I've made a bunch of fixes and cleanups. Relevant changes include
- Common error message if either `--forget` is missing or snapshot ID not specified on command line
- Prefer full snapshot IDs everywhere
- The args check can now handle both short and full snapshot IDs
- "Removing broken snapshots" is now step 4 in the troubleshooting guide as `backup` won't work before the snapshots are fixed.

cmd/restic/cmd_repair_snapshots.go
MichaelEischer: You're right, the expanded snapshot ID is only passed on if a snapshot can be loaded without errors.

I've changed the `args` check to check for the `id` forwarded by `FindAll`. In case of a snapshot loading error, that ID is always the same as what a user specified in the args. This also ensures that short and long IDs and anything inbetween will work.

MichaelEischer:
@wplapper did you intentionally close the PR? It wasn't merged yet.

wplapper:
@MichaelEischer, that was unintentionally. I reopened it.

PR #21909, opened 2026-06-23
replace FindFilteredSnapshots() with (*data.SnapshotFilter).FindAll()

wplapper:
This PR removes all calls to `FindFilteredSnapshots()`` and replaces them with calls to
` (*data.SnapshotFilter).FindAll()``. 
This PR is dependent on https://github.com/restic/restic/pull/21907.

The error handling in FindFilteredSnapshots() is not fit for purpose and needs to be replaced
with something better. Most command will terminate on a snapshot file error. ``restic snapshots`` will ignore the error since it is used in a lot of tests - and would therefore break existing behaviour if it were to fail.

No.

Checklist
---------

### DIFF
--- a/changelog/unreleased/issue-21892
+++ b/changelog/unreleased/issue-21892
@@ -1,0 +1,13 @@
+Enhancement: `restic repair snapshots --remove-ids` remoives broken snapshots
+
+In the past restic had no command to remove broken snapshot files.
+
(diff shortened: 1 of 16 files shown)


wplapper:
I realized that this PR carries too much ballast from https://github.com/restic/restic/pull/21907 around. I will close this PR and open a new one with only minimal duplication of code.

PR #21912, opened 2026-06-23
replace FindFilteredSnapshots() with (*data.SnapshotFilter).FindAll() 

wplapper:
Continuation of https://github.com/restic/restic/pull/21909.

replace remove FindFilteredSnapshots() with equivalent call to opts.FindAll()

mostly trivial changes except
cmd/restic/cmd_copy.go:
had to replace iter.Seq with iter.Seq2 to propagate the error in case it exists.

internal/data/snapshot_find.go:
add call to restic.MemorizeList()
added the code to check for transaction errors to FindAll().

cmd/restic/find.go:
remove func FindFilteredSnapshots()

This PR removes all calls to FindFilteredSnapshots()`` and replaces them with calls to (*data.SnapshotFilter).FindAll()``.
This PR is dependent on https://github.com/restic/restic/pull/21907.

The error handling in FindFilteredSnapshots() is not fit for purpose and needs to be replaced
with something better. Most command will terminate on a snapshot file error. restic snapshots will ignore the error since it is used in a lot of tests - and would therefore break existing behaviour if it were to fail.

No.

Checklist
---------

No visible changes for the user.

### DIFF
--- a/cmd/restic/cmd_copy.go
+++ b/cmd/restic/cmd_copy.go
@@ -72,20 +72,14 @@ func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
 	initMultiSnapshotFilter(f, &opts.SnapshotFilter, true)
 }
 
-var errSentinelEndIteration = errors.New("end iteration")
-
 // collectAllSnapshots: select all snapshot trees to be copied
 func collectAllSnapshots(ctx context.Context, opts CopyOptions,
 	srcSnapshotLister restic.Lister, srcRepo restic.Repository,
 	dstSnapshotByOriginal map[restic.ID][]*data.Snapshot, args []string, printer restic.Printer,
-) iter.Seq2[*data.Snapshot, error] {
-	return func(yield func(*data.Snapshot, error) bool) {
-		err := opts.SnapshotFilter.FindAll(ctx, srcSnapshotLister, srcRepo, args, func(_ string, sn *data.Snapshot, err error) error {
+) iter.Seq[*data.Snapshot] {
+	return func(yield func(*data.Snapshot) bool) {
+		for sn := range FindFilteredSnapshots(ctx, srcSnapshotLister, srcRepo, &opts.SnapshotFilter, args, printer) {
 			// check whether the destination has a snapshot with the same persistent ID which has similar snapshot fields
-			if err != nil {
-				yield(nil, err)
-				return errSentinelEndIteration
-			}
 			srcOriginal := *sn.ID()
 			if sn.Original != nil {
 				srcOriginal = *sn.Original
@@ -101,17 +95,12 @@ func collectAllSnapshots(ctx context.Context, opts CopyOptions,
 					}
 				}
 				if isCopy {
-					return nil
+					continue
 				}
 			}
-			if !yield(sn, nil) {
-				return errSentinelEndIteration
+			if !yield(sn) {
+				return
 			}
-			return nil
-		})
-		if err != nil && !errors.Is(err, errSentinelEndIteration) {
-			yield(nil, err)
-			return
 		}
 	}
 }
@@ -159,19 +148,15 @@ func runCopy(ctx context.Context, opts CopyOptions, gopts global.Options, args [
 	}
 
 	dstSnapshotByOriginal := make(map[restic.ID][]*data.Snapshot)
-	err = opts.SnapshotFilter.FindAll(ctx, dstSnapshotLister, dstRepo, nil, func(_ string, sn *data.Snapshot, err error) error {
-		if err != nil {
-			return err
-		}
+	for sn := range FindFilteredSnapshots(ctx, dstSnapshotLister, dstRepo, &opts.SnapshotFilter, nil, printer) {
 		if sn.Original != nil && !sn.Original.IsNull() {
 			dstSnapshotByOriginal[*sn.Original] = append(dstSnapshotByOriginal[*sn.Original], sn)
 		}
 		// also consider identical snapshot copies
 		dstSnapshotByOriginal[*sn.ID()] = append(dstSnapshotByOriginal[*sn.ID()], sn)
-		return nil
-	})
-	if err != nil {
-		return err
+	}
+	if ctx.Err() != nil {
+		return ctx.Err()
 	}
 
 	selectedSnapshots := collectAllSnapshots(ctx, opts, srcSnapshotLister, srcRepo, dstSnapshotByOriginal, args, printer)
@@ -205,7 +190,7 @@ func similarSnapshots(sna *data.Snapshot, snb *data.Snapshot) bool {
 // copyTreeBatched copies multiple snapshots in one go. Snapshots are written after
 // data equivalent to at least 10 packfiles was written.
 func copyTreeBatched(ctx context.Context, srcRepo *repository.Repository, dstRepo restic.Repository,
-	selectedSnapshots iter.Seq2[*data.Snapshot, error], printer restic.Printer) error {
+	selectedSnapshots iter.Seq[*data.Snapshot], printer restic.Printer) error {
 
 	// remember already processed trees across all snapshots
 	visitedTrees := srcRepo.NewAssociatedBlobSet()
@@ -214,7 +199,7 @@ func copyTreeBatched(ctx context.Context, srcRepo *repository.Repository, dstRep
 	minDuration := 1 * time.Minute
 
 	// use pull-based iterator to allow iteration in multiple steps
-	next, stop := iter.Pull2(selectedSnapshots)
+	next, stop := iter.Pull(selectedSnapshots)
 	defer stop()
 
 	for {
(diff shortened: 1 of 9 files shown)


cmd/restic/cmd_find.go
```diff
@@ -678,12 +678,19 @@ func runFind(ctx context.Context, opts FindOptions, gopts global.Options, args [
...
-	for sn := range FindFilteredSnapshots(ctx, snapshotLister, repo, &opts.SnapshotFilter, opts.Snapshots, printer) {
+	err = opts.SnapshotFilter.FindAll(ctx, snapshotLister, repo, opts.Snapshots, func(_ string, sn *data.Snapshot, err error) error {
+		if err != nil {
+			return err
+		}
 		filteredSnapshots = append(filteredSnapshots, sn)
-	}
+		return nil
+	})
 	if ctx.Err() != nil {
```
MichaelEischer: Instead of checking for both `ctx.Err()` and `err` everywhere, we should let `snapshotFilter.FindAll` handle that. From what I've found, the only code path there that doesn't guarantee a context cancellation check is in `ParallelList`. There the `<-ctx.Done()` should return an error instead of nil.

cmd/restic/cmd_copy.go
```diff
@@ -76,10 +76,14 @@ func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
...
-) iter.Seq[*data.Snapshot] {
-	return func(yield func(*data.Snapshot) bool) {
-		for sn := range FindFilteredSnapshots(ctx, srcSnapshotLister, srcRepo, &opts.SnapshotFilter, args, printer) {
+) iter.Seq2[*data.Snapshot, error] {
+	return func(yield func(*data.Snapshot, error) bool) {
+		err := opts.SnapshotFilter.FindAll(ctx, srcSnapshotLister, srcRepo, args, func(_ string, sn *data.Snapshot, err error) error {
 			// check whether the destination has a snapshot with the same persistent ID which has similar snapshot fields
+			if err != nil {
+				yield(nil, err)
+				return err
```
MichaelEischer: The decision whether to end the iteration should be left to the caller. Thus this should look like the following:
```
if !yield(nil, err) {
   return sentinelError
}
return nil
```

cmd/restic/cmd_copy.go
```diff
@@ -95,12 +99,17 @@ func collectAllSnapshots(ctx context.Context, opts CopyOptions,
...
 				}
 				if isCopy {
-					continue
+					return nil
 				}
 			}
-			if !yield(sn) {
-				return
+			if !yield(sn, nil) {
+				return nil
```
```
return sentinelError
```

cmd/restic/cmd_copy.go
```diff
@@ -76,10 +76,14 @@ func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
 func collectAllSnapshots(ctx context.Context, opts CopyOptions,
 	srcSnapshotLister restic.Lister, srcRepo restic.Repository,
 	dstSnapshotByOriginal map[restic.ID][]*data.Snapshot, args []string, printer restic.Printer,
-) iter.Seq[*data.Snapshot] {
-	return func(yield func(*data.Snapshot) bool) {
-		for sn := range FindFilteredSnapshots(ctx, srcSnapshotLister, srcRepo, &opts.SnapshotFilter, args, printer) {
+) iter.Seq2[*data.Snapshot, error] {
+	return func(yield func(*data.Snapshot, error) bool) {
```
MichaelEischer, line 82: add
```
sentinelError := errors.New("end iteration")
```

cmd/restic/cmd_copy.go
```diff
@@ -95,12 +99,17 @@ func collectAllSnapshots(ctx context.Context, opts CopyOptions,
...
 				}
 			}
-			if !yield(sn) {
-				return
+			if !yield(sn, nil) {
+				return nil
 			}
+			return nil
+		})
+		if err != nil {
```
```
if err != nil && !errors.Is(err, sentinelError) {
```

MichaelEischer (changes requested):

wplapper:
Please note: cmd/restic/cmd_snapshots.go is used here as well as in PR #21907. I try to keep the file synchronized in both PRs.
The same applies to internal/data/snapshot_find.go.

cmd/restic/cmd_copy.go
wplapper: The sentinel token has been applied and tested.

cmd/restic/cmd_find.go
wplapper: I move the checking of transaction error in ``snapshotFilter.FindAll()``.

MichaelEischer:
I'll check the PR again after https://github.com/restic/restic/pull/21907 is completed.

wplapper:
> I'll check the PR again after https://github.com/restic/restic/pull/21907 is completed.

Thank you, Michael

MichaelEischer:
Not sure whether you see my comment in https://github.com/restic/restic/pull/21907, but you've closed that PR too early.

wplapper:
> Not sure whether you see my comment in #21907, but you've closed that PR too early.

Is now open again. Sorry, my mistake.

MichaelEischer:
I've merged #21907. Please rebase this PR accordingly.

cmd/restic/cmd_copy.go
```diff
@@ -72,14 +72,20 @@ func (opts *CopyOptions) AddFlags(f *pflag.FlagSet) {
...
-) iter.Seq[*data.Snapshot] {
-	return func(yield func(*data.Snapshot) bool) {
-		for sn := range FindFilteredSnapshots(ctx, srcSnapshotLister, srcRepo, &opts.SnapshotFilter, args, printer) {
+) iter.Seq2[*data.Snapshot, error] {
+	return func(yield func(*data.Snapshot, error) bool) {
+		err := opts.SnapshotFilter.FindAll(ctx, srcSnapshotLister, srcRepo, args, func(_ string, sn *data.Snapshot, err error) error {
 			// check whether the destination has a snapshot with the same persistent ID which has similar snapshot fields
+			if err != nil {
+				yield(nil, err)
+				return errSentinelEndIteration
```
MichaelEischer: This must be
```suggestion
				if !yield(nil, err) {
				return errSentinelEndIteration
				}
```

It is up to the iterator user to decide whether a broken snapshot is a problem.

cmd/restic/cmd_copy.go
```diff
@@ -95,12 +101,17 @@ func collectAllSnapshots(ctx context.Context, opts CopyOptions,
...
-			if !yield(sn) {
-				return
+			if !yield(sn, nil) {
+				return errSentinelEndIteration
 			}
+			return nil
+		})
+		if err != nil && !errors.Is(err, errSentinelEndIteration) {
+			yield(nil, err)
+			return
```
MichaelEischer: Redundant return statement

cmd/restic/cmd_snapshots.go
```diff
@@ -294,10 +300,7 @@ func PrintSnapshots(stdout io.Writer, list data.Snapshots, reasons []data.KeepRe
 	// but we display them all in local timezone on this output.
 	footer := fmt.Sprintf("%d snapshots", len(list))
 	zoneName := time.Now().Local().Location().String()
-	if zoneName == "Local" {
```
MichaelEischer, line 297: This looks like a rebase problem. Revert this particular change.

cmd/restic/cmd_snapshots.go
```diff
@@ -90,12 +90,18 @@ func runSnapshots(ctx context.Context, opts SnapshotOptions, gopts global.Option
...
-	for sn := range FindFilteredSnapshots(ctx, repo, repo, &opts.SnapshotFilter, args, printer) {
-		snapshots = append(snapshots, sn)
-	}
-	if ctx.Err() != nil {
-		return ctx.Err()
+	err = opts.SnapshotFilter.FindAll(ctx, repo, repo, args, func(_ string, sn *data.Snapshot, err error) error {
+		if err == nil {
+			snapshots = append(snapshots, sn)
+		} else {
+			printer.E("%v", err)
```
MichaelEischer: Why does `snapshots` differ from the other commands here and does not use `if err != nil { return err }`?

MichaelEischer (changes requested):
Thanks, I only have a few small findings, see below.

cmd/restic/cmd_copy.go
wplapper: done so.

cmd/restic/cmd_copy.go
wplapper: removed ``return``

cmd/restic/cmd_snapshots.go
wplapper: At one time I had a problem with listing snapshots in test mode with ``testRunList()``. But the problem is not there anymore. So I made it identical to all the other commands.

MichaelEischer (approved):
LGTM. I've added a fix for a theoretical nil-pointer crash in `collectAllSnapshots`, see commit description for details.

Issue #1181, opened 2017-08-27
Store exclude list in repository 


dionorgua:
Probably will never happens. But it's better to propose than not.

Just want to be able to bring single restic binary on arbitrary machine and quickly call `restic backup --profile linux-default` or `restic backup --profile win32` to backup without writing all `--exclude' rules.

The idea is to just store these options as presets somewhere in repository like FileSet resource in Bacula/Bareos.

fd0:
Thanks for the proposal, that is indeed a nice idea!

PR #21930, opened 2026-07-03
backup - show exclude lines in `snapshot --json`

wplapper:
This is a 3 character fix!

internal/filter/exclude.go:
change signature of CollectPatterns() from (opts ExcludePatternOptions) to (opts *ExcludePatternOptions)

cmd/restic/cmd_backup.go:
change signature of collectRejectByNameFuncs(opts BackupOptions, ...) to collectRejectByNameFuncs(opts *BackupOptions, ...)

by doing so, the excluded items from opts.ExcludeFiles show up as well.

This PR now shows the items from opts.Excludes as well as the items from opts.ExcludeFiles. All the code was already there, but it was not using pointers. Therefore the excluded lists could never show up.

closes nearly 9 year old issue https://github.com/restic/restic/issues/1181

Checklist
---------

- no documentation changes necessary


wplapper:
The changelog file and the tests will be larger than 3 characters. ~~And so will be the documentation changes :-) .~~

The case insensitive variant of the exclude list will come as separate PR.

PR #21937, opened 2026-07-08
Reduce in-memory index entry size by storing packIndex as uint32

brainexe:

We're running into memory issues in restic pushs. 
In our remote repo we have a lot of hosts, so maaany snapshots (over 4k, for each server only <20). But doing operations requires loading the index data.   

Each blob in a repository requires one index entry in memory, which is one of the largest contributors to restic's memory usage. The indexEntry.packIndex field, an index into the per-Index packs slice, was stored as a machine-word int (8 bytes on amd64) although the value is always a small array index.

Store it as a uint32 instead. This shrinks indexEntry from 64 to 56 bytes on 64-bit platforms (56 % 8 == 0, so no padding is added), reducing index memory by roughly 12% -- about 100 MB for a repository with ~13 million blobs, as confirmed by profiling. uint32 allows 4.29 billion packs (>30 billion blobs at the assumed 8 blobs/pack), far beyond any real repository; addToPacks guards against overflow defensively. The on-disk repository format is unaffected.

```
  ┌───────────────────────────────────┬──────────────┬────────────────┬──────────────────┐
  │              metric               │ before (int) │ after (uint32) │        Δ         │
  ├───────────────────────────────────┼──────────────┼────────────────┼──────────────────┤
  │ Live index heap (gctrace -> peak) │ ~814 MB      │ ~713 MB        │ −101 MB (−12.4%) │
  ├───────────────────────────────────┼──────────────┼────────────────┼──────────────────┤
  │ Peak RSS (tree_rss_mb)            │ 1252 MB      │ 1129 MB        │ −123 MB (−9.8%)  │
  └───────────────────────────────────┴──────────────┴────────────────┴──────────────────┘
```

Checklist
---------


changelog/unreleased/pull-XXXXX
```diff
@@ -0,0 +1,14 @@
...
+entry referenced its containing pack using a machine-word-sized integer,
+even though only a small array index is needed.
+
+The pack reference is now stored as a 32-bit integer. This shrinks each index
+entry from 64 to 56 bytes on 64-bit platforms, reducing the memory required
+for the index by roughly 12% (about 100 MB for a repository with ~13 million
+blobs). Repositories on disk are unaffected; only the in-memory representation
+changed.
+
+https://github.com/restic/restic/pull/XXXXX
```
MichaelEischer: Please fill in the correct pull request number here. And also update the filename.

changelog/unreleased/pull-XXXXX
```diff
@@ -0,0 +1,14 @@
+Enhancement: Reduce memory usage of the in-memory index
+
+For most operations restic keeps one index entry per blob of a repository in
+memory, which is one of the largest contributors to its memory usage. Each
+entry referenced its containing pack using a machine-word-sized integer,
+even though only a small array index is needed.
+
+The pack reference is now stored as a 32-bit integer. This shrinks each index
```
MichaelEischer: The restic changelog is intended for end users. This is too technical for that. Let's just say that the index requires roughly 12% less memory.

internal/repository/index/index.go
```diff
@@ -70,7 +73,13 @@ func NewIndex() *Index {
 // This procedere allows to use pack IDs which can be easily garbage collected after.
 func (idx *Index) addToPacks(id restic.ID) int {
```
MichaelEischer: I'd prefer to directly return a `uint32` here and pass that all the way through until `indexmap.add`. Casting back and forth is fragile (although I can't pinpoint any problems right now)

internal/repository/index/index.go
MichaelEischer, line 488: Should check that `idx.packs` is within the size limit after the merge.

MichaelEischer (changes requested):
Thanks a lot for the PR! I'm somewhat surprised that we've missed this pretty basic optimization.

I'd like to consistently use the `packIdx` as `uint32` starting from `addToPacks` and completely remove the casts between `int` and `uint32`. At least at first glance it appeared to be possible with a reasonable amount of effort.

MichaelEischer:
> In our remote repo we have a lot of hosts, so maaany snapshots (over 4k, for each server only <20). But doing operations requires loading the index data.

You might want to split that repository. The optimization doesn't fundamentally change the scaling limit.

brainexe:
> You might want to split that repository. The optimization doesn't fundamentally change the scaling limit.

Yes, we're absolutely aware of this. It requires some larger changes on our side, but it's still planned to split the repos

MichaelEischer (approved):
LGTM. Thanks!

PR #21939, opened 2026-07-09
Match --path snapshot filters with trailing path separators

### LINKED ISSUE #2377: restic forget --path with final slash doesn't match snapshots
Output of `restic version`
--------------------------
restic 0.9.5 compiled with go1.12.1 on darwin/amd64

    # using s3 configured with envs
    $ restic backup "/Users/abathur/.bash_sessions/"
    repository X opened successfully, password is correct

    Files:           3 new,     0 changed,     0 unmodified
    Added to the repo: 1.407 KiB

    processed 3 files, 301 B in 0:01
    snapshot 0f5e0556 saved

    $ restic backup "/Users/abathur/.bash_sessions/"
    repository X opened successfully, password is correct

    Files:           0 new,     0 changed,     3 unmodified
    Dirs:            0 new,     0 changed,     2 unmodified
    Added to the repo: 0 B  

    processed 3 files, 301 B in 0:01
    snapshot 1e0d42cb saved

    $ restic forget --keep-last 1 --path "/Users/abathur/.bash_sessions/"
    repository X opened successfully, password is correct
    Applying Policy: keep the last 1 snapshots snapshots

    $ restic forget --keep-last 1 --path "/Users/abathur/.bash_sessions"
    repository X opened successfully, password is correct
    Applying Policy: keep the last 1 snapshots snapshots
    keep 1 snapshots:
    ID        Time                 Host        Tags        Reasons        Paths
    ---------------------------------------------------------------------------------------------------
    1e0d42cb  2019-08-18 11:32:50  Y                       last snapshot  /Users/abathur/.bash_sessions
    ---------------------------------------------------------------------------------------------------
    1 snapshots

    remove 1 snapshots:
    ------------------------------------------------------------------------------------
    0f5e0556  2019-08-18 11:32:42  Y                       /Users/abathur/.bash_sessions
    ------------------------------------------------------------------------------------
    1 snapshots

s3

Expected behavior
-----------------
restic backup & restic forget treat paths the same way

Actual behavior
---------------
restic forget sees no matching snapshots for the same path that restic backup used to create snapshots

    restic backup "/some_path/"
    restic backup "/some_path/"
    restic forget --keep-last 1 --path "/some_path/"
    restic forget --keep-last 1 --path "/some_path"

Backup is probably normalizing the path while forget is just doing a naive string match.

Normalizing the forget paths in the same way seems "right", but I'm not sure how multiple path arguments and Windows paths might factor in.

If normalizing is problematic, perhaps restic forget could also raise an error or at least print a warning when it has a path ending on a slash.

immensely ;)

TowyTowy:
**What does this PR change? What problem does it solve?**

Snapshot paths are stored without trailing path separators, but the `--path` filter used by `forget`, `snapshots`, `stats`, `tag`, `copy`, `rewrite` and `find` compared filter values verbatim. A value with a trailing slash such as `restic forget --path /home/user/` therefore silently matched no snapshots, while `ls latest --path /home/user/` worked (that code path already normalizes via `filepath.Clean`).

This strips trailing path separators from filter paths in `SnapshotFilter.FindAll`. Following the discussion in the issue, it deliberately does only that rather than a full `filepath.Clean`, to avoid altering `/`-separated paths when filtering Unix-created snapshots from Windows. A trailing backslash is only stripped on Windows (it is a valid filename character elsewhere), and root paths like `/` and `C:\` are kept intact.

Verified with a new unit test (fails on master: a filter path with a trailing slash matches 0 snapshots; passes with the change) and end-to-end with a built binary: `restic forget --keep-last 1 --path <dir>/ --dry-run` now applies the policy where it previously matched nothing.

**Was the change previously discussed in an issue or on the forum?**

Closes #2377. The trailing-slash-only approach follows the maintainer comments there.

**Checklist**

- [ ] I have added documentation for relevant changes (in the manual). (n/a — behavior now matches what the manual already implies)
- [x] There's a new file in `changelog/unreleased/` that describes the changes for our users.
- [x] All commit messages are formatted in the same style as the other commits in the repo.

---

Disclosure: this change was developed with the assistance of Claude (Anthropic), and was reviewed and tested (unit + end-to-end) before submission.

### DIFF
--- a/internal/data/snapshot_find.go
+++ b/internal/data/snapshot_find.go
@@ -3,6 +3,7 @@ package data
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -31,6 +32,31 @@ func (f *SnapshotFilter) Empty() bool {
 
 func (f *SnapshotFilter) matches(sn *Snapshot) bool {
 	return sn.HasHostname(f.Hosts) && sn.HasTagList(f.Tags) && sn.HasPaths(f.Paths)
+}
+
+// normalizePaths removes trailing path separators from the path filters.
+// Snapshot paths are stored without trailing separators (except for root
+// directories), such that filter paths with trailing separators would
+// otherwise never match.
+func (f *SnapshotFilter) normalizePaths() {
+	for i, path := range f.Paths {
+		f.Paths[i] = stripTrailingSeparators(path)
+	}
+}
+
+// stripTrailingSeparators removes trailing path separators from path. Root
+// directories like "/" or `C:\` are kept intact. A trailing backslash is only
+// removed on Windows, as it is a valid filename character on other operating
+// systems.
+func stripTrailingSeparators(path string) string {
+	for len(path) > 1 && os.IsPathSeparator(path[len(path)-1]) {
+		if len(path) == 3 && path[1] == ':' {
+			// keep Windows drive roots like `C:\` intact
+			break
+		}
+		path = path[:len(path)-1]
+	}
+	return path
 }
 
 // findLatest finds the latest snapshot with optional target/directory,
@@ -128,6 +154,8 @@ var ErrInvalidSnapshotSyntax = errors.New("<snapshot>:<subfolder> syntax not all
 
 // FindAll yields Snapshots, either given explicitly by `snapshotIDs` or filtered from the list of all snapshots.
 func (f *SnapshotFilter) FindAll(ctx context.Context, be restic.Lister, loader restic.LoaderUnpacked, snapshotIDs []string, fn SnapshotFindCb) error {
+	f.normalizePaths()
+
 	if len(snapshotIDs) != 0 {
 		var err error
 		usedFilter := false
--- a/internal/data/snapshot_find.go
+++ b/internal/data/snapshot_find.go
@@ -3,7 +3,6 @@ package data
 import (
 	"context"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/data/snapshot_find.go
```diff
@@ -33,6 +34,31 @@ func (f *SnapshotFilter) matches(sn *Snapshot) bool {
 	return sn.HasHostname(f.Hosts) && sn.HasTagList(f.Tags) && sn.HasPaths(f.Paths)
 }
 
+// normalizePaths removes trailing path separators from the path filters.
+// Snapshot paths are stored without trailing separators (except for root
+// directories), such that filter paths with trailing separators would
+// otherwise never match.
+func (f *SnapshotFilter) normalizePaths() {
+	for i, path := range f.Paths {
+		f.Paths[i] = stripTrailingSeparators(path)
```
orangecms: Can you just use `filepath.Clean()` from `"path/filepath"`?
I proposed that in #22016 , then found this here.

orangecms: Okay, take a closer look:
`filepath.Clean()` is already in use, but _only in `findLatest`_.
The iteration over the paths just needs to be moved to `FindAll()`. Claude couldn't figure.

orangecms: Or rather, copied/factored out, because it's also used in `FindLatest()`.

TowyTowy: Good call, that's much better. filepath.Clean already strips trailing separators and leaves `/` and `C:\` alone, so the hand-rolled helper was 20 lines doing what the stdlib does.

Factored findLatest's existing loop into normalizePaths() and called it from both, in c5854c8. Net -13 lines.

One thing worth knowing before you merge: because the shared helper keeps the filepath.Abs step, FindAll now resolves relative --path filters to absolute too, which it didn't before. I think that's a fix rather than a regression, since snapshot paths are stored absolute so a relative filter matched nothing at all previously. But it is behaviour beyond the trailing-separator fix, so say the word if you'd rather I split it out and have FindAll only Clean.

orangecms: please check out my proposed fixup:
https://github.com/orangecms/restic/tree/snapshots-paths-fixup

internal/data/snapshot_find.go
```diff
@@ -33,26 +33,37 @@ func (f *SnapshotFilter) matches(sn *Snapshot) bool {
...
+	}
+	f.Paths = paths
+	return nil
+}
+
+// findLatest finds the latest snapshot with optional target/directory,
+// tags, hostname, and timestamp filters.
+func (f *SnapshotFilter) findLatest(ctx context.Context, be restic.Lister, loader restic.LoaderUnpacked) (*Snapshot, error) {
+
+	if err := f.normalizePaths(); err != nil {
```
orangecms: Suggestion: move this to `FindLatest()`, see https://github.com/orangecms/restic/tree/snapshots-paths-fixup

internal/data/snapshot_find.go
TowyTowy: Took your fixup as-is and dropped mine, so the commit is yours with your authorship. You're right that FindLatest() is the better place: findLatest() is reached from both FindLatest() and FindAll(), so normalising at the two public entry points covers every path and avoids the double pass my version did on the FindAll route.

Branch is now my original commit plus your `clean file paths for both FindLatest and FindAll`. Tests pass, and the regression test still fails on master and passes here.

PR #21949, opened 2026-07-12
ls: only keep fields required for sorting in memory

MichaelEischer:
Instead of keeping the full `data.Node` object in memory, convert it to a `lsNodeOutput` struct. It only contains the fields necessary for the different output.

This can reduce the memory usage, in particular for nodes that contain large attributes.

No. Found this while working on tree chunking.

Checklist
---------


MichaelEischer (commented):
LGTM

Issue #5245, opened 2025-02-10
Some sort of progress indicator when running under SystemD


Fjodor42:
Output of `restic version`
--------------------------

restic 0.17.3 compiled with go1.23.3 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

For good reason, progress indication doesn't show when running under SystemD.

Long-running tasks, however, would be nice to check the progress of with `journalctl`, however.

One suggestion could be to spit out something like, in this case for `verify`, changed accordingly for other commands with the interval either being op en for discussion, or, perhaps, settable via a `restic` argument:

```
read all data: 10% complete
read all data: 20% complete
...
read all data: 100% complete
```

What are you trying to do? What problem would this solve?
---------------------------------------------------------

I am proposing giving some sort of indication of progress when running under SystemD, while not just spitting out the rapidly-changing progress indication shown when running in an interactive shell.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

Absolutely!

I use it on all of my machines, and have, generally, found it stable, and in the few cases of problems (due to corruption), I have been able to find documentation to help me fix them.

Keep up the good work :-)

MichaelEischer:
Setting `RESTIC_PROGRESS_FPS=0.0166` should work to e.g. get one line every minute.

MichaelEischer:
Closing as the question seems to be resolved.

PR #21954, opened 2026-07-13
doc: describe the cadence of JSON status messages

junkerderprovinz:
What does this PR change? Why is this change needed?

The scripting guide (`doc/075_scripting.rst`) documents the JSON lines message formats in detail, but not *when* `status` messages are emitted. While integrating restic into a backup UI we initially believed (from older behaviour and third-party posts) that piped `--json` output only carries the final summary, and worked around it with `RESTIC_PROGRESS_FPS` - only to find out from the source that backup/restore have emitted status lines over pipes by default for a long time, and that only `--quiet` suppresses them.

This PR adds one short paragraph to the general "JSON lines" section stating the actual behaviour:

- status messages are emitted roughly ten times per second by default, terminal or pipe alike,
- `--quiet` disables them,
- `RESTIC_PROGRESS_FPS` overrides both (including re-enabling them under `--quiet`), and fractional values like `0.0166` give one message per minute for log capture (as recommended in #5245).

Documentation-only change, so no changelog entry per the contribution guidelines.

Was the change previously discussed in an issue?

No separate issue; it codifies maintainer statements from #3194 and #5245 and the behaviour introduced in #3264.

Checklist
- [x] I have added documentation for relevant changes (in the docs subdirectory).

### DIFF
--- a/doc/075_scripting.rst
+++ b/doc/075_scripting.rst
@@ -243,6 +243,13 @@ Several commands, in particular long running ones or those that generate a large
 use a format also known as JSON lines. It consists of a stream of new-line separated JSON
 messages. You can determine the nature of the message using the ``message_type`` field.
 
+Messages with type ``status`` are emitted at a regular interval while the command runs,
+by default about ten times per second, regardless of whether the output is a terminal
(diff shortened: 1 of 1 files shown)


MichaelEischer (approved):
LGTM. Thanks!

PR #21969, opened 2026-07-16
Ignore future-dated snapshots when selecting latest

### LINKED ISSUE #21956: *data.SnapshotFilter.FindLatest() does not filter out future-dated snapshots.
Output of `restic version`
--------------------------
restic 0.19.1-dev (compiled manually) compiled with go1.26.4 on linux/amd64

restic should either consistently ignore future-dated snapshots or flag them as "in error". 

Steps to reproduce:
```
restic -r ... rewrite >any-snapshot< --forget --new-time "2026-12-31 23:59:59"
restic -r ... ls latest
```
restic is happy to use a future-dated snapshot.

Future-dated snapshots are happily ignored by ''restic forget --keep-*``. See data.ApplyPolicy() and data.findLatestTimestamp() for details.
I am am wondering why this fix did not make it into data.FindLatest().

See also 
https://forum.restic.net/t/is-a-safe-forget-command-possible-when-the-append-only-client-was-compromised/4543
https://github.com/restic/restic/issues/3498

I am happy to create a PR for this issue, once I know what the creators / maintainers advice how to to deal with future-dated snapshots.

tianrking:
## Summary

- ignore snapshots dated after the selection start when resolving `latest`
- add a regression test covering a future-dated snapshot alongside a valid snapshot
- add the required unreleased changelog entry

## Root cause

`SnapshotFilter.findLatest` honored an explicit timestamp limit but otherwise
accepted any later snapshot. This differed from the retention policy's existing
behavior, which excludes future timestamps.

## Validation

- `go test ./internal/data -count=1`
- `gofmt -d internal/data/snapshot_find.go internal/data/snapshot_find_test.go`
- `git diff --check`

`go test ./...` was also run on Windows. The modified package passed; unrelated
cross-platform tests expect POSIX helper commands such as `true`, `false`, and
`sleep`, which are not available in the default Windows environment.

Fixes #21956

### DIFF
--- a/changelog/unreleased/issue-21956
+++ b/changelog/unreleased/issue-21956
@@ -1,0 +1,7 @@
+Bugfix: Ignore future-dated snapshots when selecting `latest`
+
+When a repository contained a future-dated snapshot, commands that select the
+`latest` snapshot could use it. Restic now ignores those snapshots while
+selecting `latest`.
+
+https://github.com/restic/restic/issues/21956
--- a/internal/data/snapshot_find.go
+++ b/internal/data/snapshot_find.go
@@ -51,13 +51,14 @@ func (f *SnapshotFilter) findLatest(ctx context.Context, be restic.Lister, loade
 	f.Paths = absTargets
 
 	var latest *Snapshot
+	now := time.Now()
 
 	err = ForAllSnapshots(ctx, be, loader, nil, func(id restic.ID, snapshot *Snapshot, err error) error {
 		if err != nil {
 			return errors.Errorf("Error loading snapshot %v: %v", id.Str(), err)
 		}
 
-		if !f.TimestampLimit.IsZero() && snapshot.Time.After(f.TimestampLimit) {
+		if snapshot.Time.After(now) || (!f.TimestampLimit.IsZero() && snapshot.Time.After(f.TimestampLimit)) {
 			return nil
 		}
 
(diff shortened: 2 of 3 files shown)


wplapper:
> Fixes https://github.com/restic/restic/issues/21956

I would like to question the above statement. The reasons are as follows:
1. You cannot know what the creators / maintainers think of how to deal with future dated snapshots. I personally would create an error when a future dated snapshot is found, but that might create too many compatibility problems.
2. There are more interfaces to retrieve snapshots: 
- data.FindLatest(), the one you modify by this PR, used by ``ls``, ``dump``, ``list``, ``restore``, ``backup`` and by all command which used the ``latest`` as  snapshot specifier.  
- data.FindAll() is used by most of the commands which deal with snapshots using a data.SnapshotFilter.
- data.ForAllSnapshots(), used by ``prune`` and ``recover`` and implicitely used by data.FindAll()
- data.FindSnapshot(), used by ``cat`` and ``diff``.
- ``rewrite --new-time`` can currently create future dated snapshots. IMHO ``rewrite`` should be prevented from creating such snapshots.

3. ``restic check`` should throw an hint/error when future dated snapshots are found in the repository. A note about ``rewrite --new-time`` to fix this situation could be attached.

Please don't start changing more than what you have done, I think we need a decision from the creators / maintainers. Thank you!

MichaelEischer:
We've decided against adding such a check in https://github.com/restic/restic/issues/21956 . Thanks for working on it though.

PR #21970, opened 2026-07-19
idiomatic go modernization

udf2457:
* **Adopt `any` everywhere**  
  Replaced all remaining `interface{}` with the `any` alias, including variadic callback signatures like `func(string, ...interface{})` → `func(string, ...any)`.

* **Modernize loops**  
  * Use Go 1.22 integer range loops: `for i := range N` instead of `for i := 0; i < N; i++`.
  * Use Go 1.23 iterator functions to avoid intermediate slices:  
    `for v := range strings.SplitSeq(...)` and `for line := range bytes.SplitSeq(...)` instead of `strings.Split` / `bytes.Split` followed by `for _, v := range`.

* **Leverage new standard library helpers (Go 1.21+)**  
  * `slices.Contains` replaces manual search loops.
  * `maps.Copy` replaces manual key-by-key map copying.
  * `min` / `max` builtins replace hand-written clamping logic.
  * `fmt.Appendf` replaces `[]byte(fmt.Sprintf(...))` conversions.
  * `reflect.TypeFor[T]()` replaces `reflect.TypeOf` workarounds for generic types.

* **Simplify test context handling (Go 1.21+)**  
  Replaced the repetitive `ctx, cancel := context.WithCancel(context.Background()); defer cancel()` pattern in tests with `ctx := t.Context()`, which is automatically canceled when the test finishes.

* **Performance & style nits**  
  * Used `strings.Builder` instead of `+=` concatenation inside loops.
  * Streamlined goroutine coordination in a few backend/test helpers.

* **JSON consistency fix**  
  Removed `omitempty` from JSON tags on `time.Time` fields (e.g., `mtime`, `atime`, `ctime`, `time`, `date`). This ensures zero-value timestamps are serialized explicitly rather than omitted, producing more stable and predictable JSON output.

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -181,7 +182,7 @@ var ErrNoSourceData = errors.Fatal("all source directories/files do not exist")
 
 // filterExisting returns the items that exist and can be accessed. It returns
 // ErrNoSourceData if none remain, or ErrInvalidSourceData if some were skipped.
-func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
+func filterExisting(items []string, warnf func(msg string, args ...any)) (result []string, err error) {
 	for _, item := range items {
 		_, err := fs.Lstat(item)
 		if err != nil {
@@ -304,10 +305,8 @@ func (opts BackupOptions) Check(gopts global.Options, args []string) error {
 		}
 
 		filesFrom := append(append(opts.FilesFrom, opts.FilesFromVerbatim...), opts.FilesFromRaw...)
-		for _, filename := range filesFrom {
-			if filename == "-" {
-				return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
-			}
+		if slices.Contains(filesFrom, "-") {
+			return errors.Fatal("unable to read password from stdin when data is to be read from stdin, use --password-file or $RESTIC_PASSWORD")
 		}
 	}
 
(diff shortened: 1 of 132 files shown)


wplapper:
Hi, I am pretty sure that @MichaelEischer does not like this sort of monster change. Please break it up into manageable pieces - and generate a PR for each of these changes, using the standard template https://github.com/restic/restic/blob/master/.github/PULL_REQUEST_TEMPLATE.md - the use of the template  is NOT optional! 

Group the changes according to your list above. Changes (additions+deletions) should not be more than 300 lines per PR. One subject per change. Keep iterating as necessary.

MichaelEischer:
This kind of change is actually one that needs a easy way for me to verify. This very much looks like a `go fix` run based on the description. So the command and all steps necessary to reproduce must be include.

udf2457:
@MichaelEischer it is indeed `go fix ./...` with nothing else, no manual additions, no manual removals.

Obviously I read through the output and to my eyes there's nothing controversial in there and nothing that makes functional changes. its all, well, modernization as the subject says. 😉 

That is why I did not bother with the PR template.   And that is also why I was not jumping on the opportunity to break it into tiny PRs because its all pretty basic stuff that can be reviewed in 5–10 minutes despite the number of lines it touches.

Hence why I closed it.  I was not going to enter into a debate over it, I just personally felt the above that the amount of extra work being asked of me was disproportional to the nature of the PR.

MichaelEischer:
> That is why I did not bother with the PR template. And that is also why I was not jumping on the opportunity to break it into tiny PRs because its all pretty basic stuff that can be reviewed in 5–10 minutes despite the number of lines it touches.

I'd nevertheless prefer to have one commit per change even if the changes are rather uncontroversial. The 5-10 minutes number only holds when scrolling through the PR. This is probably safe for most conversions, hence the idea to split by fix type.

But I actually have a half-finished go fix branch lying around somewhere, so I'd suggest that I'll just take over here.

MichaelEischer:
But thanks for the PR anyways :-)

PR #21971, opened 2026-07-19
`TestNodeRestoreAt` & `TestOverwriteXattr` – filter out `com.apple.provenance`

udf2457:
On recent macOS versions, the OS automatically stamps a `com.apple.provenance` extended attribute onto files created by managed/sandboxed processes (part of the App Management / TCC privacy framework). Your test binary gets this stamp, so:

- The **node** (what restic backed up) has e.g. `[{user.foo [98 97 114]}]`
- The **file on disk after restore** has `[{com.apple.provenance [...]} {user.foo [98 97 114]}]`

The test does an exact comparison and fails. This is purely a test-comparison issue — restic's restore logic is correct; macOS is silently adding system metadata after the write.

This PR is a suggestion to filter out `com.apple.provenance`

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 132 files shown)


udf2457:
I pushed the wrong commit to this PR, so closed and will open fresh

(see #21972)

PR #21978, opened 2026-07-22
Run `go fix ./...`

MichaelEischer:

Converted using the following script. Also checked that the overall diff matches that of a plain `go fix ./...` call (for some reason has to be called twice). Full review is still pending.
```
set -e
analyzers=(
  any
  buildtag
  fmtappendf
  forvar
  hostport
  inline
  mapsloop
  minmax
  newexpr
  omitzero
  plusbuild
  rangeint
  reflecttypefor
  slicescontains
  slicessort
  stditerators
  stringsbuilder
  stringscut
  stringscutprefix
  stringsseq
  testingcontext
  waitgroup
)

for a in "${analyzers[@]}"; do
  echo "======== Applying go fix $a ========"
  go fix -"$a" ./... 2>&1 || true
  # Check for changes
  if git diff --quiet && git diff --cached --quiet; then
    echo "No changes for $a"
  else
    echo "Changes found for $a:"
    git diff --stat | tail -20
    git add -A
    git commit -m "$(cat <<EOF
Apply go fix -$a ./...

EOF
)"
    echo "Committed: Apply go fix $a"
  fi
done
```

Script is AI agent generated.

Replaces https://github.com/restic/restic/pull/21970

Checklist
---------

- ~~[ ] I have added tests for all code changes, see [writing tests](https://restic.readthedocs.io/en/stable/090_participating.html#writing-tests)~~

### DIFF
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -181,7 +181,7 @@ var ErrNoSourceData = errors.Fatal("all source directories/files do not exist")
 
 // filterExisting returns the items that exist and can be accessed. It returns
 // ErrNoSourceData if none remain, or ErrInvalidSourceData if some were skipped.
-func filterExisting(items []string, warnf func(msg string, args ...interface{})) (result []string, err error) {
+func filterExisting(items []string, warnf func(msg string, args ...any)) (result []string, err error) {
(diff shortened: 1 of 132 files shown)


MichaelEischer (commented):
LGTM

PR #21979, opened 2026-07-24
restic rewrite --exclude does not create empty directories when these are vacated using --(i)exclude* rules

### LINKED ISSUE #21929: add --exclude-empty-directories to `restic rewrite`
Output of `restic version`
--------------------------
restic 0.19.0-dev (compiled manually) compiled with go1.26.4 on linux/amd64

While I was working on https://github.com/restic/restic/pull/21928 I noticed that empty directories were created when excluding all files in a directory. This is how the current production code works. It would be nice to be able to suppress these empty directories if the user could say so.

The implementation of ``gatherExcludeFilters()``could reuse the existing ``keepEmptyDirectory()`` function and would return ``false`` if ``--exclude-empty-directories`` has been set.

wplapper:
This PR prevents the creation of empty sub directories which became empty while applying the
--exclude rules.

    1.) rewrite - exclude - check empty directories
    
    This PR does not create empty directories any more while using some --exclude variant,
    if the directory got vacated by activating the exclude rules.
    Therefore existing empty directories will be preserved.
    
    The modified code counts the number of entries in the unmodified tree. If the
    modified tree count is 0 and the unmodified tree count is non zero, then this directory
    can be skipped, otherwise it will be preserved. 
    Non-empty directories will be pushed onto the list of (changed) trees anyway.

    2.) rewrite - exclude - functional tests
    
    Two of the tests which deal with collapsed empty directory trees need to be modified, since
    empty directories trees can now collapse, up to the top of the tree.

    3.) rewrite - integration test
    
    cmd/restic/testdata/restic-test-base.tar.gz: add a directory structure with real names,
    including extensions: I choose a slimmed down version if the restic repository,
    leaving out changelog, .git and actual data below testdata/* and a few more bits.
    
    cmd/restic/integration_helpers_test.go:
    added testSetupBackupDataChoose() where one can select the directory structure
    to be used for backup.
    
    cmd/restic/cmd_rewrite_integration_test.go:
    add one test to prove that existing empty directories will not disappear.

closes https://github.com/restic/restic/issues/21929

Checklist
---------


wplapper:
I am not sure if the documentation needs to be amended.

wplapper:
This could be made optional with a new option ``--remove-empty-dirs``. Then this option would be passed onto ``RewriteOpts.RemoveEmptyDirectories`` as a bool and checked as 
``if t.opts.RemoveEmptyDirectories && tb.Count() == 0 && countCurrentNodes > 0 {`` ...

PR #21982, opened 2026-07-25
restic check >non-existent-snapshot< - fix error messages 

### LINKED ISSUE #21981: restic check 12345678 produces some unintelligent messages
Output of `restic version`
--------------------------
restic 0.19.1-dev (compiled manually) compiled with go1.26.4 on linux/amd64

local file system, but any backend will fail.

restic check >non-existent snapshot<

Expected behavior
-----------------
``restic check dcc60411`` (non-existent snapshot) should say 
snapshot error dcc60411: no matching ID found for prefix "dcc60411" and probably ignore it.

Actual behavior
---------------
```
snapshot error dcc60411: no matching ID found for prefix "dcc60411"

The repository contains damaged snapshot files. These damaged files must be removed to repair the repository. This can be done using the following commands. Please read the troubleshooting guide at https://restic.readthedocs.io/en/stable/077_troubleshooting.html first.

restic repair snapshots --forget dcc60411
```

``restic repair snapshots --forget dcc60411`` is incorrect advice.

in cmd/restic/cmd_check.go there needs to be a new check for the error being a ``restic.NoIDByPrefixError``.

-

wplapper:
internal/checker/checker.go:
in function loadActiveTrees() handle call results to snapshotFilter.FindAll() differently in case of an *restic.NoIDByPrefixError: just push the error straight through instead of encapsulating it as an *checker.SnapshotError. The non-existent snapshot is just ignored anyway.

cmd/restic/cmd_check.go:
check for new error type *restic.NoIDByPrefixError and issue message on printer.E()

modify test cases slightly for stderr output as well and add error for a non-existent snapshot iD.

unintelligent error messages when using ``restic check >some-ID<``.

closes https://github.com/restic/restic/issues/21981

Checklist
---------

### DIFF
--- a/cmd/restic/cmd_check.go
+++ b/cmd/restic/cmd_check.go
@@ -351,7 +351,7 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 	}()
 
 	for err := range errChan {
-		errorsFound = true
+		//errorsFound = true
 		switch e := err.(type) {
 		case *checker.TreeError:
 			printer.E("error for tree %v:\n", e.ID.Str())
@@ -359,12 +359,18 @@ func runCheck(ctx context.Context, opts CheckOptions, gopts global.Options, args
 				summary.NumErrors++
 				printer.E("  %v\n", treeErr)
 			}
+			errorsFound = true
+
 		case *checker.SnapshotError:
 			printer.E("snapshot error %v: %v", e.ID, e.Message)
 			brokenSnapshots = append(brokenSnapshots, e.ID)
+			errorsFound = true
+		case *restic.NoIDByPrefixError:
+			printer.E("%v, ignored", err)
 		default:
 			summary.NumErrors++
 			printer.E("error: %v\n", err)
+			errorsFound = true
 		}
 	}
 
--- a/cmd/restic/cmd_check_integration_test.go
+++ b/cmd/restic/cmd_check_integration_test.go
@@ -37,13 +37,13 @@ func testRunCheckOutput(t testing.TB, gopts global.Options, checkUnused bool) (s
 	return stdout.String(), stderr.String(), err
 }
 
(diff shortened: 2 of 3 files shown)


wplapper:
The problem was introduced with PR https://github.com/restic/restic/pull/21907 on 2026-06-25. This fixes a corner case for ``restic check``. PR #21907 is actively looking for broken snapshots, but a wrong snapshot ID should not be considered being a broken snapshot.

user01010111:
@wplapper I checked feecbc110611. A missing prefix now gets the not-found message without repair advice, and an existing unreadable snapshot still gets the damaged-snapshot warning and `repair snapshots --forget` advice.

The new test doesn't yet assert that the warning and repair command are absent. I prepared a 22-line test-only addition using your existing helpers: two negative assertions in the filter test, plus a damaged-snapshot control using both short and full IDs. Temporarily reintroducing the incorrect advice makes the new assertions fail while the existing test still passes.

There is also a remaining full-ID case: on an empty disposable repository, `restic check 1d20477115fb872069a28a80ffb95a82cb8b1b1920de046a68c0195da63f30cf` still exits 1 and recommends forgetting that nonexistent snapshot. `FindSnapshot` sends a valid full ID directly to `LoadSnapshot`, bypassing the prefix lookup and `NoIDByPrefixError`. I reproduced this with the binary built from this PR.

The prepared tests pass with Go 1.26.6 on Linux, as do `RESTIC_TEST_FUSE=0 go test ./cmd/restic ./internal/checker ./internal/data -count=1`, `go vet` for those packages, and the `build.go` build. FUSE was disabled because `fusermount` is unavailable.

Please let me know whether you'd like the test-only addition sent to your branch or kept for an agreed follow-up. It's prepared locally; I haven't opened another PR.

wplapper:
Hi @user01010111,
Thanks for checking my PR, and in particular the tests I have added (or didn't add).

I am currently on holidays. I will contact you when I am back home. Then I can check in peace the missing bits you mentioned in your post.

PR #21986, opened 2026-07-29
Add mount tree metadata preload

yonathan-ashebir:

Delay problem when exploring a mount of remote repository

At an opt-in flag for caching tree blobs before establishing a mount. Thus, incurring an initial load time to have more interactive exploration (and also to become `find`-friendly), which can be desired for high latency backends (cloud ...) for reasonably sized repos.

Although I believe the matter has been raised before and there is a commit that reverted the behaviour, I still believe it is good to have as an opt-in. 

Checklist
---------


wplapper:
Thank you for your PR!

Please remove files ``man/restic.mount.1`` and ``m̀an/bash_completion.sh`` from your modified file list. These files will be auto-generated by ``restic generate`` for every new published release.

yonathan-ashebir:
Thanks, removed the generated doc changes from doc/bash-completion.sh and doc/man/restic-mount.1 in 63176889e.

PR #22003, opened 2026-08-09
doc: clarify new/changed/unmodified in backup summary

### LINKED ISSUE #4565: Correct the conflicting language of summaries in verbose output
Output of `restic version`
--------------------------
restic 0.16.2 compiled with go1.21.3 on linux/386

When performing a backup for the first time, e.g.:
```
restic backup --cleanup-cache --verbose -p /root/backup.p --compression max \
home root opt \
var/{smb,www,local}
```

The following output is generated:
```
Files:       1085848 new,     0 changed,     0 unmodified
Dirs:        16015 new,     0 changed,     0 unmodified
```
Reading that first line, zero files were "changed".
But, at the same time, zero files were also "unmodified". Since "unmodified" means "not modified", and since 0 files were not modified, it seems apparent that every file was modified.
*Another way of saying "0 UNmodified", is "1085848 modified". Or: none were changed, and all were changed.*
Thus, in effect, the above line is saying:
```
Files:       1085848 new,     0 changed, 1085848 modified
```

How can no files be changed, but also every file be modified?
Same for directories.

While I know the important number here is "0 changed", having "0 unmodified" (meaning "every file modified") at the same time, is unnecessarily confusing. It's also a double negative--"none not modified".

**At the very least, this should say "0 changed, 1085848 modified". That still wouldn't make sense, but at least it wouldn't involve a double negative.**

It's a great backup utility, and has all the features I need for handling Linux server and workstation backups.
I even introduced it to a colleague who maintains Windows systems, who finds it quite satisfying.

locker95:
First-time backups print something like `5307 new, 0 changed, 0 unmodified`, which reads oddly if you treat “unmodified” as the opposite of “changed”.

Added a short note under the first `--verbose` example: the three numbers are mutually exclusive (new / changed / unmodified vs the parent snapshot).

Fixes #4565

### DIFF
--- a/doc/040_backup.rst
+++ b/doc/040_backup.rst
@@ -41,6 +41,18 @@ As you can see, restic created a backup of the directory and was pretty
 fast! The specific snapshot just created is identified by a sequence of
 hexadecimal characters, ``40dc1520`` in this case.
 
+The ``Files:`` / ``Dirs:`` summary line splits items into three **mutually
+exclusive** buckets relative to the parent snapshot (if any):
+
+* **new** — path did not exist in the parent snapshot
+* **changed** — path existed but content or metadata differs
+* **unmodified** — path matches the parent snapshot
+
+So ``5307 new, 0 changed, 0 unmodified`` on a first backup is expected: every
+file is new, nothing can be “changed” or “unmodified” yet. Later backups use
+the same three words; they are not “modified vs unmodified” with double
+negatives — “unmodified” means “same as parent”, not “not new”.
+
 You can see that restic processed 1.720 GiB of data; this is the
 size of the files and directories in ``~/work`` on the local file system. It
 also reports that only 1.200 GiB was added to the repository. This means that


doc/040_backup.rst
```diff
@@ -41,6 +41,18 @@ As you can see, restic created a backup of the directory and was pretty
 fast! The specific snapshot just created is identified by a sequence of
 hexadecimal characters, ``40dc1520`` in this case.
 
+The ``Files:`` / ``Dirs:`` summary line splits items into three **mutually
+exclusive** buckets relative to the parent snapshot (if any):
+
+* **new** — path did not exist in the parent snapshot
```
MichaelEischer, line 47: Let's keep the AI hypens out of the docs and stick to this style:
```
* ``--force``: turn off change detection and rescan all files.
```

doc/040_backup.rst
```diff
@@ -41,6 +41,18 @@ As you can see, restic created a backup of the directory and was pretty
...
+* **changed** — path existed but content or metadata differs
+* **unmodified** — path matches the parent snapshot
+
+So ``5307 new, 0 changed, 0 unmodified`` on a first backup is expected: every
+file is new, nothing can be “changed” or “unmodified” yet. Later backups use
```
MichaelEischer, line 52: Just use double backticks

doc/040_backup.rst
```diff
@@ -41,6 +41,18 @@ As you can see, restic created a backup of the directory and was pretty
...
+the same three words; they are not “modified vs unmodified” with double
+negatives — “unmodified” means “same as parent”, not “not new”.
```
MichaelEischer, line 54: Drop that sentence. It's rather hard to understand.

MichaelEischer (changes requested):

PR #22007, opened 2026-08-13
fix(swift): store APIKey as SecretString to prevent debug log leak

Marcel-MTL:

The OpenStack password (OS_PASSWORD / ST_KEY) was declared as a plain string in Config, so debug.Log("config %#v", cfg) in Open() wrote it in clear text to the DEBUG_LOG file. Change APIKey to options.SecretString, consistent with AuthToken and ApplicationCredentialSecret, and call.Unwrap() when constructing the swift.Connection.

To reproduce before fix:
export DEBUG_LOG=/tmp/restic.log
_run any restic command against a swift backend_
grep APIKey /tmp/restic.log  # shows password in clear text

No

Checklist
---------

- [ ] No new test added: the protection is enforced by the type system (.Unwrap() in swift.go won't compile if APIKey reverts to string), and TestSecretStringStruct in internal/options/secret_string_test.go already covers the redaction behavior for structs containing SecretString fields.


changelog/unreleased/pull-XXXXX
```diff
@@ -0,0 +1,11 @@
+Bugfix: Redact Swift backend password in debug log output
```
MichaelEischer: Drop the duplicate changelog file

changelog/unreleased/pull-22007
```diff
@@ -0,0 +1,11 @@
+Bugfix: Redact Swift backend password in debug log output
+
+The OpenStack password (set via OS_PASSWORD or ST_KEY) was stored as a
```
MichaelEischer: The changelog description is for restic users not developers. Just shorten it down an mention that it's now ensured that it no longer ends up in a debug log file.

MichaelEischer (changes requested):

Marcel-MTL:
@MichaelEischer I've addressed both comments, removed the duplicate changelog file and shortened the description. Ready for another look.
Thanks

MichaelEischer (approved):
LGTM. Thanks!

PR #22008, opened 2026-08-14
Store stable (virtual) device IDs for hardlinked files

### LINKED ISSUE #3041: Restic 0.10.0 always reports all directories "changed", adds duplicate metadata, when run on ZFS snapshots

hashhar:

restic stores `stat.st_dev` in each node so that the restorer can recreate hardlinks. btrfs and ZFS assign a new device ID whenever a subvolume or a filesystem snapshot is mounted so every hardlinked file differs from the parent snapshot and restic uploads a duplicate set of tree blobs on each run. The alpha feature `device-id-for-hardlinks` already omits the ID for directories and files without hardlinks. Hardlinked files still store it.

This PR maps the reported device IDs to virtual ones that stay the same across backup runs. The ID is taken from the parent snapshot if available and otherwise derived from the snapshot-relative path at which the device is first seen. Only the fact that files on one device share an ID within a snapshot matters to the restorer, so the reported value is not required.

Users who already run with the feature see no rewrite: the first backup after the upgrade adopts the IDs stored by the parent snapshot. Enabling the feature on an existing repository rewrites the directory metadata once because directories no longer store a device ID at all, which was already the case before this PR.

## Impl choices and decisions

* `storedDeviceID` decides the device ID for every node. The tree walker calls it and the fileSaver workers copy the result into the node. The mapping must run on one goroutine because the first node seen on a device decides the ID of the whole device. `saveDir` processes directory entries in lexicographic order so the same path determines the ID of a device in every run. The fileSaver workers run in a non-deterministic order hence the device ID is passed to them through `saveFileJob`.
* Paths are hashed with FNV-1a 64. `hash/maphash` is seeded randomly per process and would produce different IDs on every run.
* Zero means "no device ID" and is never returned. A hash that is zero or already taken is re-rolled.
* An ID inherited from the parent snapshot is checked against the IDs already in use. If a device is split across two mountpoints between runs both parent nodes carry the same stored ID. Reusing it would make the restorer hardlink unrelated files with matching inode numbers.
* Only nodes that actually store a device ID may claim an entry otherwise a directory could take the entry for a device using its own path hash and the hardlinked files behind it could never inherit the parent value.

`--one-file-system` keeps using the device IDs reported by the filesystem, those decide which parts of the filesystem are traversed. `restic ls --ncdu` exposes the device ID as `dev` and now shows the virtual value.

Closes #3041

The implementation follows the design in https://github.com/restic/restic/issues/3041#issuecomment-3929931326.

An earlier attempt exists at #5706.

Checklist
---------


hashhar:
note that first commit is a pre-existing bugfix i noticed as part of this work and is related to this PR.

hashhar:
I was doing a self-review and noticed that we can centralize the decision of assigning device IDs so I've made that change. It still follows the same design as earlier but with easier to reason about code.

Issue #22005, opened 2026-08-10
improve `restic find` speed


wplapper:
Output of `restic version`
--------------------------
restic 0.19.1-dev (compiled manually) compiled with go1.26.4 on linux/amd64

What should restic do differently? Which functionality do you think we should add?
----------------------------------------------------------------------------------

I am looking at the time it takes to run `restic find` to find things in the repository and compare it with some code which does the same by using ``data.StreamTrees()`` instead of using ``walker.Walk()``.

``find`` uses ``walker.Walk()`` in two distinct functions for two different inputs:
1. ``findInSnapshot()`` to find patterns in pathnames
2. ``findIDs()`` is used for options ``--tree`` and ``--blob``

From my tests I can definitely conclude that the streaming version runs between 3 and 10 times faster than the walking version. The more snapshots a repository contains, the greater the performance difference will be, since the new algorithm saves on recursing the full tree every time a snapshot is analyzed. 
My repository only contains about 80 snapshots, about 12,500 tree blobs, 50,000 unique files in 11 GiB of data.

Outline of the new algorithm:
-----------------------------

Run data.StreamTrees(), collect parent to child (=sub-directory) information per tree element. Build the complete top-down directory structure for each sbapshot.

Then run a second data.StreamTrees() to collect data for the ``--blob`` option or to find patterns in the pathname list. This actually pretty fast and does not use much memory.

Sort the output, according to existing ``find`` rules and generate output.

Replace ``findInSnapshot()`` and ``findIDs()`` with the new code.

Fixes:
------

The new code also addresses issue raised in https://github.com/restic/restic/pull/5666 regarding the options ``--show-pack-id`` together with ``--json`` which now generates clean JSON output.

Did restic help you today? Did it make you happy in any way?
------------------------------------------------------------

I am happy to submit a PR.

wplapper:
On a fundamental level, the calls to ``walker.Walk()`` are gone, together with the calling functions ``findInSnapshot()`` and ``findIDs()``.

This will be replaced by one call to ``data.StreamTrees()`` which will collect all the parent->child relationships for the selected set of snapshots. Once this fucntion completes, a call to ``buildDirectoryTree()`` will convert theses parent->child relationships into a list of pairs (pathname and snapshot) for each tree for the selected snapshots. 

[...]

PR #22009, opened 2026-08-15
find: fast version using StreamTrees

wplapper:
This is supposed to be a drop-in replacement of the currect implementation of ``restic find``.

Instead of calling ``walker.Walk()`` multiple times, call ``data.StreamTrees()`` once and use the results from ``StreamTrees`` in the subsequent processing of the find steps

Changes:
--------
remove findInSnapshot() and findIDs() completely and replace them with
```
streamTrees:        build parent->child relationship by using *data.StreamTrees()
buildDirectoryTree: build map directoryNames[restic.ID][]DirectoryInfo]
                    which contains for each tree a list = aof pairs(=DirectoryInfo) (pathname, sn)
streamBlobs:        takes over the task of findIDs
printSelectedBlobs: prints the results from streamBlobs()
streamPatterns:     takes over the task of findInSnapshot
printPatterns:      prints the results from streamPatterns()
```
Additions:
----------
using the option ``--show-pack-id`` is now fully integrated in the ``*statefulOutput`` functions and can be safely used in JSON and text outputs. See https://github.com/restic/restic/pull/5666 for more details.

Note: 
-----
The output order of the files is not strictly identical compared to the current output order, but treated as a set, the outputs are identical. The reason is as follows:
``walker.Walk()`` walks the tree in topological order and generates the output as it walks along. The new implementation collects all the output raw data in one go and then sorts them into the necessary order.

I makes ``find``run faster, between 3 - 10 times faster. I have observed special cases where I saw a speedup of ~20. The more calls to ``walker.Walk()`` are needed to get the result set, the greater the speedup will be. I would not be surprised if you were to search in, lets say 500 snapshots, you could experience speedups of ~20 or more. But I don't have any repository with this many snapshots.

https://github.com/restic/restic/issues/22005

Checklist
---------

 tests can be found at #22010


wplapper:
Commit 734532e needs doc changes and certainly new tests - to follow.

PR #22016, opened 2026-08-18
snapshots: remove trailing slashes for paths

orangecms:
Shell completion would add a trailing slash for a directory, and then one would currently get an empty list for `snapshots --path /foo/bar/`, because the `paths` are stored without the trailing slash.

Checklist
---------


orangecms:
It might be better to fix this elsewhere, any hint is welcome; I've only looked into the codebase for the first time now, am not as familiar with it yet.

orangecms:
After some digging, I see that #21939 already has the better solution - neat! :)

PR #22024, opened 2026-08-21
index: replace encoding/json with a purpose-built decoder

ton31337:
Loading the repository index dominates the cost of every restic command
on a large repository. A system-wide `perf record` on a shared hosting
node running 5-10 concurrent `restic backup` invocations, against
repositories holding tens of millions of blobs, put roughly 97% of
restic's CPU in the index load and under 1% in the backup itself. Three
captures taken minutes apart agreed, and so did every sampled process.

Self cost, restic processes only:

     8.25%  encoding/json.checkValid
     7.60%  index.(*indexMap).foreachWithID
     7.46%  encoding/json.stateInString
     7.07%  [kernel]
     5.19%  encoding/json.(*decodeState).rescanLiteral
     4.16%  encoding/json.(*decodeState).object
     3.63%  zstd.sequenceDecs_decodeSync_safe_bmi2
     3.37%  hash/maphash.(*Hash).Write
     3.01%  encoding/json.stateEndValue
     2.63%  encoding/json.indirect
     2.17%  encoding/json.unquoteBytes
     1.89%  reflect.(*rtype).Name

and cumulatively:

    75.68%  index.ForAllIndexes.func1
    63.07%    index.DecodeIndex
    59.71%      encoding/json.Unmarshal
    39.75%        encoding/json.(*decodeState).unmarshal

Note what the top of that profile is doing. checkValid walks the entire
buffer once purely to confirm it is well-formed, before any value is
extracted; stateInString and stateEndValue are the per-byte scanner
state machine it drives; rescanLiteral then walks each literal a second
time during the decode proper. Those four are ~24% of restic's total CPU
and are all scanning the input rather than converting it. indirect,
unquoteBytes and reflect.(*rtype).Name below them are the reflection
overhead of dispatching into packJSON/blobJSON.

None of that is avoidable through encoding/json, but none of it is
necessary either. The index schema is fixed and trivial -- hex IDs,
small integers and a two-valued type enum -- and on top of the scanning
cost every pack and blob is materialised into a packJSON/blobJSON slice
that is copied into the Index and then discarded.

Decode index files with a parser written for this schema, storing
entries into the Index as they are read. On BenchmarkDecodeIndex
(200,000 pack files, ~1.6 million blobs), median of five runs:

                            before       after
    DecodeIndex           3087 ms/op   806 ms/op    3.8x faster
    DecodeIndexParallel    448 ms/op   121 ms/op    3.7x faster
    allocated              575.3 MB    216.8 MB     -62%
    allocations             902,704       2,948     -306x

The parser recognises only what restic's own encoder emits. Escapes in
strings, fractional or exponent numbers, duplicate keys and malformed
input all make it bounce, and DecodeIndex re-parses the buffer with
encoding/json. That keeps encoding/json as the reference implementation
for unusual input and as the single source of the error messages users
see, so no index that could be read before becomes unreadable.


CAOShurong:
Reviewed at exact head `0504452d` (base `a80be14`, go1.26.7 windows/amd64). This is a hot core path, so I verified beyond build+tests:

**Local gates**

- `go build ./internal/repository/...` and `go vet ./internal/repository/index/...` clean.
- Full index package suite passes, including `FuzzDecodeIndex` seeds and the committed regression corpus entry (`a4f5f503f1c2568b`) — good to see the differential harness already caught at least one real input class during development.
- I also ran an independent 90s fuzz session (`go test -fuzz FuzzDecodeIndex -fuzztime 90s`): ~17.3M execs against the seed corpus, zero crashes and zero fast-vs-encoding/json divergences.
- `./internal/repository/...` and `./internal/restic/...` all green at this head.

**Independent benchmark reproduction** (same machine, `-benchtime 1x`, so treat as order-of-magnitude):

| | master | PR branch |
|---|---|---|
| `BenchmarkDecodeIndex` | ~2092 ms/op | ~583–648 ms/op |
| allocations | — | 216.8 MB/op, 2,950 allocs/op |

That is a ~3.3–3.6× single-run speedup locally vs the claimed 3.8×, and the allocation numbers reproduce the PR's claim (216.8 MB / 2,948→2,950 allocs) essentially exactly. The win direction and magnitude check out.

**Code-level notes**

- The `skippable` fold-match fallback (leaving anything `bytes.EqualFold`-compatible with a known key to encoding/json, because of its case-insensitive field matching) is exactly the subtlety I'd expect a hand-written parser to get wrong, and it's handled correctly here with a comment explaining why.
- Bounce-don't-guess discipline is consistent throughout: escapes, leading zeros, fractions/exponents, values above `MaxUint32`, repeated keys, and trailing garbage all fall back rather than approximate encoding/json behavior.

**One gap: missing changelog entry**

`CONTRIBUTING.md` step 7 asks for a `changelog/unreleased/pull-22024` file for any change users should be aware of — and a multi-x index decode speedup definitely qualifies (it's the kind of thing release announcements highlight). The PR currently adds only source and test files. Could you add one? Something like:

```
Enhancement: Speed up repository index decoding

Decoding repository index files now uses a parser written for the index
format instead of encoding/json. On large repositories this makes index
loading roughly 3-4x faster and reduces memory allocations by ~60%.

https://github.com/restic/restic/pull/22024
```

(Also noting the PR has no CI checks reported yet — likely just needs a maintainer workflow approval for first-time contributors.)

MichaelEischer:
Go 1.27 includes a much faster JSON decoder. How do it's performance compare to this PR? There are also more optimized interfaces in that new JSON decoder that should help with performance. This needs to be clarified before I'll even _consider_ looking at this PR. Handrolling a complex parser is a quick way to cause problems.

CAOShurong:
Thanks — I reran the existing benchmark on the exact PR head `0504452d` and exact base `a80be14`, using the official Go 1.27.0 toolchain on Windows/amd64 (Intel Core Ultra 5 125H). Go 1.27's default `encoding/json` is backed by the new v2 implementation.

Five-run medians on the same generated 200,000-pack input:

- `BenchmarkDecodeIndex`: base `encoding/json` **1,346.3 ms/op**, 575.3 MB, 902,711 allocs; PR **668.3 ms/op**, 216.8 MB, 2,948 allocs — about **2.01x faster**, **62.3% fewer allocated bytes**, and **306x fewer allocations**.
- `BenchmarkDecodeIndexParallel`: base **1,572.8 ms/op** vs PR **114.7 ms/op** in this Windows `RunParallel` run. I would treat that figure as runtime/hardware-specific rather than compare it directly with the Linux number in the PR description.

For reference, the same serial benchmark with Go 1.26.7 was 2,125.6 ms/op on base vs 706.8 ms/op on the PR. So Go 1.27 materially improves the standard-decoder baseline (about 1.58x here), but it does not remove the PR's serial advantage or allocation reduction.

I recorded the exact commands and raw results in a local evidence file; no repository source was changed. This tests the current `encoding/json.Unmarshal` path rather than a temporary port to the `encoding/json/v2` API. Benchmark orchestration used OpenAI Codex assistance; the commands and results were executed and checked locally.

MichaelEischer:
There are now also MarshalJSONTo and UnmarshalJSONFrom interfaces. As the index contains lots of BlobType and ID elements, using those could help performance a lot. Please also give those a try.

ton31337:
According to the analysis:

config | sec/op | MiB/op | allocs/op | vs its own Go's encoding/json
-- | -- | -- | -- | --
Go 1.26 · encoding/json | 3.133 | 548.6 | 902,704 | 1.00×
Go 1.27 · encoding/json (uses UnmarshalJSONFrom) | 1.782 | 548.6 | 902,702 | 1.00×
Go 1.26 · own parser | 0.833 | 206.8 | 2,948 | 3.76×
Go 1.27 · own parser | 0.784 | 206.8 | 2,948 | 2.27×

With 1.27, it's better (because of UnmarshalJSONFrom()), but still not sweet enough.

MichaelEischer:
I've just tested a few variants myself. `MarshalJSONTo and UnmarshalJSONFrom` indeed don't provide any performance benefit over the existing methods (without any chance for improvements)

However, I still seriously dislike handrolling everything from decoding numbers to strings ourselves. That's a recipe for problems sooner or later. Instead it's possible to use the new jsontext package from Go 1.27, which can tokenize the input without unnecessary allocations. (see https://pkg.go.dev/encoding/json/jsontext#Decoder). This allows implementing a Go 1.27+ hand-rolled decoder with minimal boilerplate that can also robustly skip unknown object keys.

For older Go versions, just fall back to the existing `json.Unmarshal`. It's not worth the maintenance effort to add a separate optimized decoder.

P.S.: Please fill out the PR template.
P.S.2: (I've only assigned myself to track PRs for review)

ton31337:
>However, I still seriously dislike handrolling everything from decoding numbers to strings ourselves.

Speaking as one of the maintainers of [FRRouting](https://github.com/frrouting/frr) project, I can say that we are using json-c for most of the JSON stuff, but there are special cases (like this for restic too), where we do "handcrafted" JSONs just because of the performance, otherwise it's horrible. And yes, these custom encoders are for very huge JSONs only.

PR #22027, opened 2026-08-24
mount: honor --repository-file when checking local overlap

### LINKED ISSUE #22025: Failure to mount repository while using command line option --repository-file
Output of `restic version`
--------------------------
restic 0.19.1 compiled with go1.26.4 on linux/amd64

rest-server started in a docker container

I try to use `restic --insecure-no-password --repository-file repository-file mount ./mount` but this gives an error

```Fatal: mountpoint .../restic-bug/mount is inside the local repository directory .../restic-bug; refusing to mount to avoid deadlocking the FUSE server```

With version 19.0 the above command works as expected.

It also works, if I specify the repository on the command line: `restic --insecure-no-password -r "rest:http://john:1234@localhost:8000" mount ./mount`

Generally, I don't call restic directly, but by using resticprofile, which uses internally `--repository-file`. I first thought it might be a problem with resticprofile, but the problem also appears with restic alone.

I used the following bash script to reproduce the problem:

```
#!/usr/bin/env bash

DIRECTORY_TO_BACKUP=directory-to-backup
REST_USER=john
REST_PASSWORD=1234

generate_random_string() {
    local length=${1:-8}
    tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c "$length"
}

if [ ! -d "$DIRECTORY_TO_BACKUP" ]; then
  mkdir "$DIRECTORY_TO_BACKUP"
  for (( i=1; i<=10; i++ )); do
    FILE_PATH="$DIRECTORY_TO_BACKUP"/file_$(generate_random_string 6)
    generate_random_string 1000 > "$FILE_PATH"
    echo "" >> "$FILE_PATH"
  done
fi

if [ ! -f repository-file ]; then
  echo "rest:http://$REST_USER:$REST_PASSWORD@localhost:8000" > repository-file
fi

test -d mount || mkdir mount

if [ ! -d data ]; then
  mkdir data
  htpasswd -c -B -b "data/.htpasswd" "$REST_USER" "$REST_PASSWORD"
fi

docker pull restic/rest-server:latest

cat <<EOF

Execute the following commands in another terminal:

restic --repository-file repository-file --insecure-no-password init
restic --repository-file repository-file --insecure-no-password backup "$DIRECTORY_TO_BACKUP"
restic --repository-file repository-file --insecure-no-password mount ./mount

EOF

docker run --rm -p 8000:8000 -v ./data:/data --name rest_server  restic/rest-server
```

Expected behavior
-----------------
restic successfully executes mount command.

Actual behavior
---------------
restic refuses to create a mount point.

No - probably a change between version 19.0 and 19.1. Might be related to pull request #5348 (issue #5234)

I use restic daily privately and also at work and I am very happy with it!

### PR DESCRIPTION
**Pull request by AshSgDe29071999**
`validateMountpoint` parsed `gopts.Repo` to decide whether the mountpoint overlaps a local repository. That field is empty when the location comes from `--repository-file`, and an empty location is treated as a local path (cwd).

So this failed even for a rest-server backend:

```
restic --repository-file repository-file mount ./mount
Fatal: mountpoint .../mount is inside the local repository directory ...; refusing to mount to avoid deadlocking the FUSE server
```

The same command with `-r rest:http://...` worked. The overlap check now reads the repository-file the same way `OpenRepository` does, and still rejects a real local path that overlaps the mountpoint.

Fixes #22025

## Testing

`go test ./cmd/restic -run 'TestCheckMountpointOverlap|TestValidateMountpointRepositoryFile' -count=1`

### DIFF
--- a/cmd/restic/cmd_mount.go
+++ b/cmd/restic/cmd_mount.go
@@ -21,6 +21,7 @@ import (
 	"github.com/restic/restic/internal/errors"
 	"github.com/restic/restic/internal/fs"
 	"github.com/restic/restic/internal/global"
+	"github.com/restic/restic/internal/textfile"
 	"github.com/restic/restic/internal/ui"
 	"github.com/restic/restic/internal/ui/progress"
 
@@ -242,7 +243,23 @@ func validateMountpoint(mountpoint string, gopts global.Options) error {
 	// Refuse to mount onto (or under, or over) the local repository directory.
 	// Doing so makes the FUSE server read its own backend files through the
 	// mount it just created, deadlocking the kernel (GH #5234).
-	loc, err := location.Parse(gopts.Backends, gopts.Repo)
+	//
+	// Resolve -r / --repository-file the same way OpenRepository does. Parsing
+	// the empty gopts.Repo string as a local path treats cwd as the repo, so
+	// `restic --repository-file ... mount ./mount` refused every mount under
+	// the working directory even for rest: (and other remote) backends (#22025).
+	repo := gopts.Repo
+	if repo == "" && gopts.RepositoryFile != "" {
+		s, err := textfile.Read(gopts.RepositoryFile)
+		if err != nil {
+			return err
+		}
+		repo = strings.TrimSpace(string(s))
+	}
+	if repo == "" {
+		return nil
+	}
+	loc, err := location.Parse(gopts.Backends, repo)
 	if err != nil {
 		return err
 	}
--- a/cmd/restic/cmd_mount_integration_test.go
+++ b/cmd/restic/cmd_mount_integration_test.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	systemFuse "github.com/anacrolix/fuse"
+	"github.com/restic/restic/internal/backend/all"
 	"github.com/restic/restic/internal/data"
 	"github.com/restic/restic/internal/global"
 	"github.com/restic/restic/internal/restic"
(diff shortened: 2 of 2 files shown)


yangfan-yf-yf (commented):
I found two issues that should be addressed before merge:

1. `validateMountpoint` reads `RepositoryFile`, but `runMount` then passes the unchanged `gopts` to `openWithReadLock`, and `global.OpenRepository` reads the file again. If the file changes from a remote URL to an overlapping local path between those reads, validation approves the remote URL while the mount opens the local repository, bypassing the deadlock safeguard. Could we resolve the repository location once before validation and pass that exact value to both validation and `OpenRepository` (for example, set `gopts.Repo` to the resolved value and clear `gopts.RepositoryFile`)? A regression test should also change the file after resolution and verify that the captured location is used. Sharing the resolution path would also preserve the existing missing-file and conflicting-option error handling.

2. `CONTRIBUTING.md` asks user-visible bug fixes to include an entry under `changelog/unreleased`, using `changelog/TEMPLATE`. Please add `changelog/unreleased/issue-22025` and include the issue and PR URLs.

MichaelEischer (changes requested):
This implementation in unmaintainable. Instead turn `func readRepo(gopts Options) (string, error)` in `global.go` into a method on the `global.Options` struct that can be used to resolve the repository location.

P.S.: Using an AI agent is no excuse to ignore the PR template

Oh and ignore the race condition mentioned by jangfan-yf-yf. It's a user's own problem if they try to break the safeguard.

PR #22028, opened 2026-08-25
docs: fix two dead links (PGP key URL, MinIO docs URL)

CAOShurong:
Found while crawling the external links in `doc/*.rst` (98 unique URLs live-checked):

1. **`doc/090_participating.rst` (Security section):** the MIT PGP keyserver lookup URL for the security-contact key (`https://pgp.mit.edu/pks/lookup?op=get&search=0xCF8F18F...`) no longer resolves. Replaced with `https://restic.net/gpg-key-alex.asc`, which serves HTTP 200 and contains exactly the same key - verified by importing it with GnuPG: fingerprint `CF8F18F2844575973F79D4E191A6868BD3F7A907` (key id `91A6868BD3F7A907`, Alexander Neumann <alexander@bumpern.de>), matching the key block printed below the link in the doc.

2. **`doc/030_preparing_a_new_repo.rst` (Minio Server section):** `https://min.io/docs/minio/linux/` returns 404. Replaced with `https://min.io/docs/minio/linux/index.html` which serves 200 and is the same documentation landing page.

Both replacements were verified with live HTTP checks before committing.

### DIFF
--- a/doc/030_preparing_a_new_repo.rst
+++ b/doc/030_preparing_a_new_repo.rst
@@ -281,36 +281,6 @@ after the bucket name like this:
           be converted to path-style URLs instead, for example ``s3.us-west-2.amazonaws.com/bucket_name``.
           See below for configuration options for S3-compatible storage from other providers.
 
-Minio Server
-************
-
-`Minio <https://min.io/>`__ is an Open Source Object Storage,
-written in Go and compatible with Amazon S3 API.
-
--  Download and Install `Minio Download <https://min.io/download#/linux>`__.
--  You can also refer to `Minio Docs <https://min.io/docs/minio/linux/index.html>`__ for step by step guidance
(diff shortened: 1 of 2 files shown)


doc/030_preparing_a_new_repo.rst
```diff
@@ -288,7 +288,7 @@ Minio Server
 written in Go and compatible with Amazon S3 API.
 
 -  Download and Install `Minio Download <https://min.io/download#/linux>`__.
--  You can also refer to `Minio Docs <https://min.io/docs/minio/linux/>`__ for step by step guidance
+-  You can also refer to `Minio Docs <https://min.io/docs/minio/linux/index.html>`__ for step by step guidance
```
MichaelEischer: This redirects to minio aistor which isn't freely available. Let's just drop the whole Minio Server section.

MichaelEischer (approved):

PR #22031, opened 2026-08-27
prune: add per-pack unused prune options

LRFLEW:

Currently, restic's prune options only allow you to tune the repacking behavior based on whole-repository unused data statistics. Users can use `--max-unused` and `--max-repack-data` to effectively set an upper and lower bound to control how much data is repacked. However, there currently aren't any similar options to control how packs are chosen for repacking on a per-pack level. This PR adds two new options to try and make the prune process more tunable.

- `--max-pack-unused` sets a percentage (default `100%`). If the percentage of an individual pack file that is unused is more than this percentage, then the pack file will be repacked to remove the unused data. This ignores the value for `--max-unused`, which can allow it to keep repacking past that threshold. However, this value respects any limit set for `--max-repack-size`, and it will stop repacking as soon as that limit is reached.

- `--ignore-pack-unused` sets a percentage (default `0%`). If the percentage of an individual pack file that is unused is less than or equal to this percentage, then the pack file will be kept without repacking. This ignores the value for `--max-unused`, and can be used to set a lower-bound to avoid repacking files where the benefit of repacking is potentially too low to bother with. The pack files may still be considered for repacking for reasons other than unused data, such as compression and total size.

If both `--max-pack-unused` and `--ignore-pack-unused` are specified, then the percentage for `--ignore-pack-unused` must be less than or equal than the percentage for `--max-pack-unused`. If the values are different, then any pack files that fall between the two values will be handled according to `--max-unused`. Note that any unused data from packs kept because of `--ignore-pack-unused` still contributes to the total unused that `--max-unused` uses. If both values are the same, then `--max-unused` has no effect.

Examples

- Repack any packs with at least 20% unused data, then repack packs until under 5% unused.
```
$ restic prune --max-pack-unused 20% --max-unused 5%
```

- Repack packs until under 5% unused, ignoring any pack files with 10% or less unused.
```
$ restic prune --ignore-pack-unused 10% --max-unused 5%
```

- Repack only packs with more than 10% unused.
```
$ restic prune --max-pack-unused 10% --ignore-pack-unused 10%
```

- Repack any packs with at least 20% unused data, then repack packs until under 5% unused, ignoring any packs with less than 10% or less unused.
```
$ restic prune --max-pack-unused 20% --ignore-pack-unused 10% --max-unused 5%
```

I am not aware of any issues or discussions related to this or similar feature requests. If you're aware of any, let me know, and I will add them here.

Checklist
---------


MichaelEischer:
The PR description describes what this PR does, but never state what goal it should achieve. "Add more tunables" is not a goal, it is additional complexity that needs to be justified.

LRFLEW:
I didn't include the rationale in the PR to try and keep the main description cleaner, and keep it to just what was needed to understand *what* the code is trying to do. The rationale I had behind this PR comes down to trying to tune for prune "efficiency". I'm not entirely sure if that term makes sense, so I'll try to explain what I'm talking about.

When you run prune, you'll get a stats output that looks something like this:

```
to repack:        217208 blobs / 2.547 GiB
this removes:       3008 blobs / 402.756 MiB
...
```

These first two lines correlate to the "efficiency" I'm referring to. The first line indicates how much pack data is getting repacked, and the second line indicates how much space saving that repacking will produce. We can take a ratio of these two values to get that the repacking efficiency of this particular prune was about 402.8 MiB / 2608 MiB = 15.44%.

This sort of efficiency can be practical to try and optimize for. Since pruning requires an exclusive lock on the repository, it may not be a best use of resources to repack for low-efficiency data savings on a shared repository, as it would prevent other devices from being able to perform backups. Similarly, it may not be practical for a portable device to do this sort of low-efficiency data savings, especially when such operations may be preventing sleeping (or other power-saving states) on battery-powered devices. As such, tuning the pruning for efficiency can be a useful tool to have.

The current options tuning options can *influence* the efficiency, since restic will prioritize repacking the highest efficiency packs, but there's no option to **directly** control this attribute. Any threshold set for `--max-unused` or `--max-repack-size` can still trigger low-efficiency repacking or block high-efficiency repacking, depending on the state of the repository. Adding options to directly set efficiency thresholds allows for directly tuning for efficiency, which is the "goal" of this PR.

As for why this PR works on a per-pack level, the idea is to make the threshold "stable" across multiple prunes. Calling multiple prunes with `--max-repack-size` set to a fixed value allows splitting repacks into multiple transactions, so it seems practical to ensure the other tuning options will work the same whether called as a single prune call or as multiple sized-size prune calls. `--max-unused` already works this way, since it's tracking a full-repository stat, so I wanted these new options to also work this way. The new options could have worked by looking at the rolling average efficiency across the prune call, but doing so would mean that which packs get repacked would depend on if the prune gets split into multiple calls. Doing it by looking at each pack's individual efficiency contribution makes it "stable" in this regard.

Also on the topic of goals, one of my main goals when writing this PR was that I wanted to ensure that the default behavior remained unchanged to avoid unintended consequences. That's a large part of why I opted to make it two options to define three separate ranges (with the middle range behaving as before). In that way, the goal was to provide a way for power-users attempting to optimize their setup to tune this parameter (in a similar way that `--pack-size` and `--no-cache` exist to do) without impacting the average user's experience.

PR #22063, opened 2026-09-16
Allow skipping folders based on an xattr

Paulomen2712:

At CERN, we run one of our backup appliances using restic, and we currently backup 4 billions of files to more than 50,000 restic repositories, making a total of 10PB handled by the system.

Right now, `restic` needs to crawl the entire filesystem when performing a backup in order to check which entries have changed and need to be stored. Having one single change on a 17 million dataset implies traversing all the entries, which adds noticeable metadata load on mounted filesystems backed by distributed storage systems. 

Some storage platforms like CephFS recursively update extended attributes on parent directories whenever a file changes. We can use this to skip directories when we know that nothing inside them has changed since the previous snapshot, regardless of how deeply nested the files are.

Therefore, this PR aims to leverage those recursive mtimes in filesystems that support them to avoid traversing directory trees that are unmodified, improving backup speed and reducing the impact on production backend.

For files, we already do an initial `stat` to compare metadata and avoid an expensive read of the file. The implementation basically looks like this:

```go
if previous != nil && !fileChanged(...) {
	node, err := arch.nodeFromFileInfo(...) // metadata info
	node.Content = previous.Content         // copy content from previous snapshot
}
```

The logic would be similar for directories, where the main benefit would be skipping an entire branch and avoiding all the unnecessary `stat` calls below it.

```go
if previous != nil && !arch.dirChanged(node, previous) {
	node, err := arch.nodeFromFileInfo(...)
	node.Subtree = previous.Subtree // reuse subtree from previous snapshot
}
```

Here, `dirChanged` compares an xattr specified through the new `--compare-xattr` flag. This leaves the storage-specific xattr selection to the user. For example, with CephFS, the backup could be run as:

```bash
restic backup --compare-xattr ceph.dir.rctime
```

For this to work, a few small changes had to be made to how nodes handle extended attributes. On these storage systems, restic currently does not store these xattrs because `xattr.llist` returns an empty list, as the attributes are not exposed through the normal listing mechanism.

We therefore need to know in advance which attribute names we want to track. This was addressed by appending a list of xattrs specified in the `OS_XATTR` environment variable to the list returned by `xattr.llist`.

Another improvement this PR provides is to parallelise the fetching of extended attributes. On a local filesystem this does not make much difference, but distributed storage systems typically separate metadata from the actual file data. In the case of CephFS, requests are handled by Metadata Servers (MDS), which can process multiple metadata queries concurrently. As a result, there is a significant performance benefit when these `getfattr` requests are issued in parallel. Since before restic wouldn't track the extended attributes at all, this was never a concern. In our testing, this improvement offers a 30% speed-up while fetching the extended attributes.

How was this PR tested? 

Decided to add this section because extensive tests of this feature were done.

Testing was done by continuously backing up the git repo of the linux kernel. It was chosen because it has around 100k files over 6k directories. A benefit of choosing a git repo source is that we can simulate valid modifications by simply walking the commit tree and also brings full reproducibility.

Thus, a test iteration would use two copies of the same repository: one running the latest `restic`, and one running these changes.

| Step | Modified `restic` | Latest `restic` |
|---|---|---|
| 1. Reset repository | `git checkout HEAD~N` |  |
| 2. Backup | `restic backup --compare-xattr ...` | `restic backup ...` |
| 3. Restore | `restic restore ...` | `restic restore ...` |
| 4. Check Git state | `git status` | `git status` |
| 5. Check Git objects | `git fsck` | `git fsck` |
| 6. Compare restores | `rsync -lrciIn A/ B/` | |

All integrity checks should return no errors or differences. In particular, the final `rsync` comparison should produce no output.

Here is a snippet of the last month of testing. Each test takes around 40 mins, because of the large git repositories.

What you can see in this plot is that we ran more than 1000 tests that were all successful and show that the changes still lead to the exact same restic state, i.e we get the exact same amount of data added, files/dirs new and changed and that all are still valid. But we see a massive gain of 20x performance both in terms of backup duration and amount of files processed.

Screenshot From 2026-09-16 11-44-22
After a succesfull testing period, this feature is currently in production at CERN. 

A long time ago this was discussed in #3648. Here, two problems were raised:
 - The statistics won't line up. For example now running `restic snapshots`, the size will not be correct. This is because the size we compute = `total_bytes_processed`. Since now we are processing a lot less data by skipping, these numbers no longer match. Technically these storages also expose recursive sizes, which could be displayed instead, but this adds a lot of bookkeeping and storage specific implementation. For CephFS it would be knowing that this number corresponds exactly to `ceph.dir.rbytes`
 - The scanner does not skip anything, but since it is run only for statistics in theory this can be toggled off when runnign the backup that skips, since the statistics of the backup itself wouldn't be entirely accurate in the first place
 - In general, we conclude that are two assumible caveats, taking into account the massive improvement this feature brings. 

Checklist
---------

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - add-skip-dirs
 
   # run tests for all pull requests
   pull_request:
--- a/cmd/restic/cmd_backup.go
+++ b/cmd/restic/cmd_backup.go
@@ -96,7 +96,7 @@ type BackupOptions struct {
 	ReadConcurrency   uint
 	NoScan            bool
 	SkipIfUnchanged   bool
-
+	CompareXattr      string
 	readConcurrencyFlag *pflag.Flag
 }
 
@@ -140,6 +140,7 @@ func (opts *BackupOptions) AddFlags(f *pflag.FlagSet) {
 		f.BoolVar(&opts.ExcludeCloudFiles, "exclude-cloud-files", false, "excludes online-only cloud files (such as OneDrive, iCloud drive, …)")
 	}
 	f.BoolVar(&opts.SkipIfUnchanged, "skip-if-unchanged", false, "skip snapshot creation if identical to parent snapshot")
+	f.StringVar(&opts.CompareXattr, "compare-xattr", "", "skip based on unchanged xattr")
 
 	opts.readConcurrencyFlag = f.Lookup("read-concurrency")
 
@@ -655,6 +656,7 @@ func runBackup(ctx context.Context, opts BackupOptions, gopts global.Options, te
 	arch.SelectByName = selectByNameFilter
 	arch.Select = selectFilter
 	arch.WithAtime = opts.WithAtime
+	arch.CompareXattr = opts.CompareXattr
 
 	arch.Error = func(item string, err error) error {
 		success = false
--- a/internal/archiver/archiver.go
+++ b/internal/archiver/archiver.go
@@ -1,6 +1,7 @@
 package archiver
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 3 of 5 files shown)


Paulomen2712:
It seems that `https://dl.minio.io/server/minio` no longer is valide:

```
410 Gone

The open-source MinIO Server, MinIO Client (mc) and MinIO KES projects are
archived and no longer maintained. MinIO does not provide product support,
security updates, or security advisories for them, and does not accept or
process vulnerability reports concerning them.

These files are no longer served from this site. This applies to all
community releases and to all hotfix builds for them.
```

I changed the `.github.com/workflows/tests.yml` to install the minio with `go install github.com/minio/minio@latest` instead

PR #22054, opened 2026-09-06
doc: clarify environment variable precedence

user01010111:
What does this PR change? What problem does it solve?
-----------------------------------------------------

Clarify that explicit command-line options override their corresponding environment variables, with `-r` / `--repo` as an example. Replace all 13 ambiguous “replaces” entries in the scripting documentation with “overridden by”.

Validated with a Sphinx HTML build with warnings treated as errors, a restic build, the existing focused flag/environment tests, and disposable local repositories confirming environment fallback and command-line overrides for repository locations and password files.

Was the change previously discussed in an issue or on the forum?
----------------------------------------------------------------

Closes #22040.

Checklist
---------

- [x] I have added documentation for relevant changes (in the manual).
- [x] I'm done! This pull request is ready for review.

No code changes or generated man-page edits. Tests and a changelog entry are not added for this wording correction.

PR #22067, opened 2026-09-20
backend/s3: switch tests from minio to seaweedfs

MichaelEischer:
The CI tests depend on the minio server binary, which by now has been removed from the upstream servers. The minio server at https://github.com/minio/minio is archived for nearly a year now and the only option to get a binary would be to use `go install` to build it ourselves. However, this takes ages and will become more dysfunctional over time.

Instead completely replace minio with seaweedfs. While seaweedfs is normally intended for larger deployments, there's also a `weed mini` mode that just offers what we need for testing.

Built using cursor agent, reviewed and slightly cleaned up myself afterwards.

Tests are currently broken: https://github.com/restic/restic/pull/22054

Checklist
---------


MichaelEischer (commented):
LGTM

